PR #3949, opened 2020-02-20
Include `dep ensure` step in `pulumi new` output.

lukehoban:
This change adds the ", run `dep ensure`, then` part to the output below:

Screen Shot 2020-02-20 at 12 27 19 PM


lukehoban:
Note that we could have chosen to go further and automatically run `dep ensure` for the user like we do `npm install`.  I avoided this for now as we hope to remove the need for this - and I don't want to break the ability to fix the templates in the future.  Plus, `dep` is so slow this would be more painful.  But once we land on our long term approach, we likely want to do that.

stack72:
This made me happy :) Thanks!!

PR #3881, opened 2020-02-06
improve PAC error output

jkisk:
resolves #2872

before:
image

after with policy violations:
image

after with no policy violations:
image

PR #3960, opened 2020-02-21
Fix panic on failed stack policy

justinvp:
Avoid panic during stack validations. This regressed recently with https://github.com/pulumi/pulumi/pull/3881/files#diff-1484cdf02b75d8b12012f9474273ec54R815 when the PAC error output was modified.

Fixes https://github.com/pulumi/pulumi-policy/issues/200

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,12 @@
 CHANGELOG
 =========
 
-- Fix a regression for CustomTimeouts in Python SDK
+## HEAD (unreleased)
+- Fix a regression for CustomTimeouts in Python SDK.
   [#3964](https://github.com/pulumi/pulumi/pull/3964)
+
+- Avoid panic when displaying failed stack policies.
+  [#3960](https://github.com/pulumi/pulumi/pull/3960)
 
 ## 1.11.0 (2020-02-19)
 - Allow oversize protocol buffers for Python SDK.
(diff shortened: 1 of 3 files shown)


justinvp:
Would it be more appropriate to fix this "upstream", such that any other future downstream uses of `PolicyViolationEventPayload.ResourceURN` don't have to bother doing such checks?

Basically, change this:

https://github.com/pulumi/pulumi/blob/24a1cc433a119f04b1d067dad0f32b336a5b0e51/pkg/resource/deploy/step_generator.go#L1323-L1331

To convert URNs with a value of `""` to the stack's URN?

@pgavlin, @chrsmith, thoughts?

justinvp:
> +1 to moving the logic upstream,

Moved upstream.

> at the point the policy event is created seems arbitrary

The source of this URN is someone calling `reportViolation` in their policy code. `reportViolation` takes two parameters: (1) the violation message and (2) an optional URN of the resource to associate the violation with. If the URN isn't specified, it comes across as `""`. I don't think I've seen any policies in the wild actually specifying a URN here, but someone could pass any random string. So while it might be an arbitrary place to do this, it is still an appropriate place to do such validation.

> why not fix up URNs of "" to be the stack resource whenever a resource is first created?

I was wondering the same, but I'm not sure there's a good way to do it. In order to construct the URN associated with the stack resource, we have to know the stack name and project name, and I'm not sure we have that information in all cases where a URN is used.

In any case, I'm going to merge this to address the panic, but happy to make subsequent improvements here.

PR #3961, opened 2020-02-22
use version tag

ekrengel:
Related to https://github.com/pulumi/pulumi-policy/issues/41


pkg/backend/display/progress.go
```diff
@@ -776,11 +776,22 @@ func (display *ProgressDisplay) printPolicyViolations() bool {
 			eventJ.PolicyPackName); packNameCmp != 0 {
 			return packNameCmp < 0
 		}
-		if packVerCmp := strings.Compare(
-			eventI.PolicyPackVersion,
-			eventJ.PolicyPackVersion); packVerCmp != 0 {
-			return packVerCmp < 0
+		if eventI.PolicyPackVersionTag != "" && eventJ.PolicyPackVersionTag != "" {
```
chrsmith: Nit: You might want to consider something like introducing `iTag, jTag :=  eventI.PolicyPackVersionTag, eventJ.PolicyPackVersionTag`. Since that might allow you to put avoid splitting the code across multiple lines without sacrificing much in terms of readability.

pkg/backend/httpstate/client/client.go
```diff
@@ -40,6 +42,12 @@ import (
 	"github.com/pulumi/pulumi/pkg/workspace"
 )
 
+// policyPackVersionTagPattern is the regex pattern for allowable version tags
+// of a Policy Pack.
+const policyPackVersionTagPattern = "^[a-zA-Z0-9-_.:]{1,150}$"
```
chrsmith: Why did you settle on this particular regular expression? Is it designed to match NPM or some other common package manager? If not, I'd suggest we use the more strict `"^[a-zA-Z0-9-_.]{1,100}$"` that we use for most other user-provided identifiers in Pulumi, like stack names.

We can always relax the naming requirements later. (But having a tag 100+ characters in length seems like it could only lead to problems, IMHO.)

pkg/backend/httpstate/client/client.go
```diff
@@ -546,22 +553,36 @@ func (pc *Client) PublishPolicyPack(ctx context.Context, orgName string,
...
 
-	fmt.Printf("Publishing %s to %s\n", analyzerInfo.Name, orgName)
+	// Print a publishing message. We have to handle the case where an older version of pulumi/policy
+	// is in use, which does not provide  a version tag.
+	publishingMsg := fmt.Sprintf("Publishing %q - version %s to %q organization\n",
+		analyzerInfo.Name, analyzerInfo.VersionTag, orgName)
+	if analyzerInfo.VersionTag == "" {
+		publishingMsg = fmt.Sprintf("Publishing %q to %q organization\n", analyzerInfo.Name, orgName)
+	}
+	fmt.Print(publishingMsg)
```
chrsmith: This seems problematic. We shouldn't have the Pulumi API client also deal with writing output messages directly to STDOUT. This makes it impossible to replicate the same output on the Pulumi Service, since the message isn't part of the engine event stream.

So at the very least, we should have the caller (or someone even lower in the callstack) call `fmt.Print`. And even then, we would want to avoid calling `fmt.Print` directly, but instead emit a diagnostic event.

pkg/engine/events.go
```diff
@@ -70,14 +70,15 @@ type DiagEventPayload struct {
...
-	Prefix            string
+	ResourceURN          resource.URN
+	Message              string
+	Color                colors.Colorization
+	PolicyName           string
+	PolicyPackName       string
+	PolicyPackVersion    string
+	EnforcementLevel     apitype.EnforcementLevel
+	Prefix               string
+	PolicyPackVersionTag string
```
chrsmith: Nit: Would we want to move this field up, so that it is next to the other `PolicyPack*` fields? Would it make sense to avoid the repetition and create a `type PolicyPackInfo struct` to bundle the data?

Same for other structs where we add the new `PolicyPackVersionTag` field, like `AnalyzeDiagnostic`.

sdk/proto/analyzer.proto
```diff
@@ -116,13 +116,15 @@ message AnalyzeDiagnostic {
...
     string urn = 8;                        // URN of the resource that violates the policy.
+    string policyPackVersionTag = 9;       // Version Tag of the Policy Pack.
 }
 
 // AnalyzerInfo provides metadata about a PolicyPack inside an analyzer.
 message AnalyzerInfo {
 	string name = 1;                  // Name of the PolicyPack.
 	string displayName = 2;           // Pretty name for the PolicyPack.
 	repeated PolicyInfo policies = 3; // Metadata about policies contained in PolicyPack.
+    string versionTag = 4;            // Version of the Policy Pack.
```
chrsmith: Nit: The alignment looks like its off by two spaces here.

sdk/proto/analyzer.proto
```diff
@@ -116,13 +116,15 @@ message AnalyzeDiagnostic {
     repeated string tags = 6;              // Keywords/terms to associate with a policy, e.g., "cost".
     EnforcementLevel enforcementLevel = 7; // Severity of the policy violation.
     string urn = 8;                        // URN of the resource that violates the policy.
+    string policyPackVersionTag = 9;       // Version Tag of the Policy Pack.
```
justinvp: Why is it necessary to add this? Why couldn't the existing `policyPackVersion` be used (which effectively was never really hooked up in the policy SDK)?

sdk/proto/analyzer.proto
ekrengel: I noticed this but on my VS code it looks aligned 🤔

pkg/apitype/events.go
```diff
@@ -47,13 +47,12 @@ type DiagnosticEvent struct {
...
-	PolicyName           string `json:"policyName"`
-	PolicyPackName       string `json:"policyPackName"`
-	PolicyPackVersion    string `json:"policyPackVersion"`
-	PolicyPackVersionTag string `json:"policyPackVersionTag"`
+	ResourceURN       string `json:"resourceUrn,omitempty"`
+	Message           string `json:"message"`
+	Color             string `json:"color"`
+	PolicyName        string `json:"policyName"`
+	PolicyPackName    string `json:"policyPackName"`
+	PolicyPackVersion string `json:"policyPackVersion"`
```
ekrengel: - [ ] I need to check what the service is doing w the event's `PolicyPackVersion` vs `PolicyPackVersionTag`.

pkg/backend/display/progress.go
ekrengel: Going to use version now, so this new check is not longer needed.

pkg/backend/httpstate/client/client.go
ekrengel: I was matching some other regex. Happier to go with stricter regex. 

- [ ] I will also update the regex in the service.

pkg/engine/events.go
ekrengel: I ended up removing it from a lot of places. But I will double check these.. The reason I put it at the bottom was because this mimics the protobuf schema which is numbered and since we dont have any empty numbers, I had to put it at the end to remain backwards compatible...

My preference is that this stays in the same order at the protobuf so we keep consistent order.

pkg/backend/httpstate/client/client.go
ekrengel: Agreed -- although this is how it was initially implemented. I thought we had an open issue to address this, but alas I dont see it. If you think it's critical, please open another issue.

pkg/backend/httpstate/policypack.go
```diff
@@ -156,37 +167,41 @@ func (pack *cloudPolicyPack) Publish(
...
 		return result.FromError(err)
 	}
 
-	fmt.Printf("\nPermalink: %s/policypacks/%s/%d\n", pack.Backend().URL(), pack.ref.Name(), version)
+	// Handle older versions of pulumi/policy that do not provide the version tag.
+	if analyzerInfo.Version == "" {
+		analyzerInfo.Version = strconv.Itoa(version)
+	}
 
+	fmt.Printf("\nPermalink: %s/policypacks/%s/%s\n", pack.Backend().URL(), pack.ref.Name(), analyzerInfo.Version)
```
justinvp: Should we change `PublishPolicyPack` to return a string rather than int? Then the if check here could be removed.

PR #3967, opened 2020-02-24
Improve error message for policy pack not found

### LINKED ISSUE #3945: Bad error message if policy pack isn't found
If you try to run a policy pack, sometimes you get a pretty awful error message "failed to load plugin ... fork/exec ... no such file or directory"

```zsh
$ PULUMI_DEBUG_COMMANDS=true pulumi up \
    --cwd ./iam-user-reader \
    --policy-pack ./policy-pack \
    --skip-preview

error: policy pack "/Users/chris/go/src/github.com/chrsmith/pulumi-iam-user-validator/iam-user-reader/policy-pack/bin/index.js" failed to start: failed to load plugin /Users/chris/.pulumi/bin/pulumi-analyzer-policy: fork/exec /Users/chris/.pulumi/bin/pulumi-analyzer-policy: no such file or directory
```

The relevant information is in there, but it's hard to tease out. For example, in my case, I needed to set the `--policy-pack` flag to `../policy-pack`.

Ultimately we should detect this specific error and provide a more use-friendly message, since it's probably going to be a common occurrence.

chrsmith:
Improve an error message when trying to load a policy pack that isn't found.

Currently, the error is accurate... but unwieldy:

``` 
error: policy pack "/Users/chris/go/src/github.com/pulumi/pulumi-service/examples/resources/does/not/exist" failed to start: failed to load plugin /Users/chris/.pulumi/bin/pulumi-analyzer-policy: fork/exec /Users/chris/.pulumi/bin/pulumi-analyzer-policy: no such file or directory
```

With this change, we look for the `syscall.ENOENT` error and just provide a shorter message:

```
error: policy pack plugin not found at "/Users/chris/go/src/github.com/pulumi/pulumi-service/examples/resources/does/not/exist"
```

... to be honest though, I honestly don't know why we are seeing this error in the first place.

From the Golang code, `cmd.Start()` is documented as "Start starts the specified command but does not wait for it to complete." So I'm not sure why this would actually fail since the plugin (the `pulumi-analyzer-policy` file) _doest_ exist.

Also, how is the actual policy pack -- the thing that the `pulumi-analyzer-policy` script will end up bootstrapping -- passed? It looks like the arguments passed to the shell script are `[127.0.0.1:60160 .]`.

... so I would guess that the exit code isn't from the `sh`/shell process, but instead the _last exit code_ that was executed. Which would be from `node`, when the `run-policy-pack` module isn't found?

```
#!/bin/sh
# pulumi-analyzer-policy
PULUMI_RUN_SCRIPT_PATH=$(node -e "try { console.log(require.resolve('@pulumi/pulumi/cmd/run-policy-pack')) } catch (e) { console.error(e.message) }")
if [ ! -z "$PULUMI_RUN_SCRIPT_PATH" ]; then
    node "$PULUMI_RUN_SCRIPT_PATH" $@
fi
```

Anyways, let me know if I can clean up the code more and/or if there is a better way to go about this.

Fixes #3945

### DIFF
--- a/pkg/resource/plugin/analyzer_plugin.go
+++ b/pkg/resource/plugin/analyzer_plugin.go
@@ -108,9 +108,14 @@ func NewPolicyAnalyzer(
 	plug, err := newPlugin(ctx, pwd, pluginPath, fmt.Sprintf("%v (analyzer)", name),
 		[]string{host.ServerAddr(), "."}, env)
 	if err != nil {
-		if err == errRunPolicyModuleNotFound {
+		// The original error might have been wrapped before being returned from newPlugin. So we look for
+		// the root cause of the error. This won't work if we switch to Go 1.13's new approach to wrapping.
+		if errors.Cause(err) == errRunPolicyModuleNotFound {
 			return nil, fmt.Errorf("it looks like the policy pack's dependencies are not installed; "+
 				"try running npm install or yarn install in %q", policyPackPath)
+		}
+		if errors.Cause(err) == errPluginNotFound {
+			return nil, fmt.Errorf("policy pack not found at %q", name)
 		}
 		return nil, errors.Wrapf(err, "policy pack %q failed to start", string(name))
 	}
--- a/pkg/resource/plugin/plugin.go
+++ b/pkg/resource/plugin/plugin.go
@@ -22,6 +22,7 @@ import (
 	"strconv"
 	"strings"
 	"sync/atomic"
+	"syscall"
 	"time"
 
 	multierror "github.com/hashicorp/go-multierror"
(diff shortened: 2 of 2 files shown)


justinvp:
> I honestly don't know why we are seeing this error in the first place.

Thanks, Chris! I'd like to better understand this as well. Let me do some of my own spelunking, and I'll report back soon.

pkg/resource/plugin/analyzer_plugin.go
```diff
@@ -108,10 +108,15 @@ func NewPolicyAnalyzer(
 	plug, err := newPlugin(ctx, pwd, pluginPath, fmt.Sprintf("%v (analyzer)", name),
 		[]string{host.ServerAddr(), "."}, env)
 	if err != nil {
-		if err == errRunPolicyModuleNotFound {
+		// The original error might have been wrapped before being returned from newPlugin. So we look for
+		// the root cause of the error. This won't work if we switch to Go 1.13's new approach to wrapping.
+		if errors.Cause(err) == errRunPolicyModuleNotFound {
```
ekrengel: OMG TIL about `Cause` 🥇

justinvp: Same.

Issue #3958, opened 2020-02-21
Caching of Policy Packs to be org scoped


ekrengel:
The local caching of Policy Packs needs to be scoped by organization.

PR #3971, opened 2020-02-25
add orgName as subdir to policies cache dir

ekrengel:
Fixes https://github.com/pulumi/pulumi/issues/3958

Previously we were caching all policies under `.pulumi/policies` which meant there could be collisions between orgs. This adds the orgName as a subdir (e.g. `.pulumi/policies/ekrengel`.

### DIFF
--- a/pkg/workspace/plugins.go
+++ b/pkg/workspace/plugins.go
@@ -351,14 +351,14 @@ func HasPluginGTE(plug PluginInfo) (bool, error) {
 }
 
 // GetPolicyDir returns the directory in which policies on the current machine are managed.
-func GetPolicyDir() (string, error) {
-	return GetPulumiPath(PolicyDir)
+func GetPolicyDir(orgName string) (string, error) {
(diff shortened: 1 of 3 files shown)


pkg/workspace/plugins.go
```diff
@@ -351,14 +351,14 @@ func HasPluginGTE(plug PluginInfo) (bool, error) {
 }
 
 // GetPolicyDir returns the directory in which policies on the current machine are managed.
-func GetPolicyDir() (string, error) {
-	return GetPulumiPath(PolicyDir)
+func GetPolicyDir(orgName string) (string, error) {
+	return GetPulumiPath(PolicyDir, orgName)
 }
 
 // GetPolicyPath finds a PolicyPack by its name version, as well as a bool marked true if the path
```
praneetloke: Sorry for the late comment, but do you think the comment should be updated now that this func requires the org name as well?

ekrengel: Address this in another PR I have open in this repo. Added you as a reviewer there!

PR #3977, opened 2020-02-25
Remove use of `github.com/pkg/errors` from Go SDK

jen20:
In preparation for publishing a separate module of the Go SDK for Pulumi on which providers can depend, we should reduce the dependency footprint so as to cause end users as few issues as possible with transitive dependency versioning.

This commit removes all use of `github.com/pkg/errors` from the Go SDK to that end, replacing it with the standard library `errors` package and `fmt` for error formatting where appropriate.

### DIFF
--- a/sdk/go/pulumi/context.go
+++ b/sdk/go/pulumi/context.go
@@ -17,6 +17,9 @@
 package pulumi
 
 import (
+	"context"
+	"errors"
+	"fmt"
 	"reflect"
 	"sort"
 	"strings"
@@ -24,8 +27,6 @@ import (
 
 	structpb "github.com/golang/protobuf/ptypes/struct"
 	multierror "github.com/hashicorp/go-multierror"
-	"github.com/pkg/errors"
-	"golang.org/x/net/context"
 	"google.golang.org/grpc"
 
 	"github.com/pulumi/pulumi/pkg/resource"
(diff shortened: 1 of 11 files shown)


sdk/go/pulumi/run.go
```diff
@@ -53,7 +53,7 @@ func RunErr(body RunFunc, opts ...RunOption) error {
 
 	// Validate some properties.
 	if info.Project == "" {
-		return errors.Errorf("missing project name")
+		return fmt.Errorf("missing project name")
```
stack72: errors.New

sdk/go/pulumi/context.go
```diff
@@ -70,7 +71,7 @@ func NewContext(ctx context.Context, info RunInfo) (*Context, error) {
 	if addr := info.EngineAddr; addr != "" {
 		conn, err := grpc.Dial(info.EngineAddr, grpc.WithInsecure())
 		if err != nil {
-			return nil, errors.Wrap(err, "connecting to engine over RPC")
+			return nil, fmt.Errorf("connecting to engine over RPC %s", err)
```
stack72: nit: missing `:`

sdk/go/pulumi/run.go
jen20: Good catch

sdk/go/pulumi/context.go
jen20: Good catch

jen20:
Pushed a new commit covering both of them - good catches1

sdk/go/pulumi/context.go
```diff
@@ -70,7 +71,7 @@ func NewContext(ctx context.Context, info RunInfo) (*Context, error) {
 	if addr := info.EngineAddr; addr != "" {
 		conn, err := grpc.Dial(info.EngineAddr, grpc.WithInsecure())
 		if err != nil {
-			return nil, errors.Wrap(err, "connecting to engine over RPC")
+			return nil, fmt.Errorf("connecting to engine over RPC: %s", err)
```
lukehoban: Should these be `%w` to maintain the wrapping semantics?  Not sure how important that is - but potentially keeps things closer to where they were before?

jen20: Yes - assuming we're happy with supporting only Go 1.13 and above (which I am), using `%w` here is better.

PR #3938, opened 2020-02-19
Add support for secrets in Go SDK

EvanBoyle:
Adds secret state to `pulumi.Output`, support for `additionalSecretOutputs`, and `config.Get/Try/RequireSecret*()`.

I uncovered some test bugs along the way in RPC tests both in pkg/resource and in the go sdk, and fixed them as a part of this change. 

Fixes https://github.com/pulumi/pulumi/issues/2820

Issue #3846, opened 2020-01-30
Define and document behavior for go ResourceOption merging


EvanBoyle:
We should deliberately define and add tests for the resource options merging behavior in the Go SDK. Right now there is some discrepancy between the way that options are merged between providers (merged) and ignore changes arrays (overwritten):

ReadResource in go takes in a list of options and merges them
https://github.com/pulumi/pulumi/blob/f168bdc1c2757f27e99d859a10044e4328d758a6/sdk/go/pulumi/context.go#L266-L269

But different options have different behaviors:
https://github.com/pulumi/pulumi/blob/f168bdc1c2757f27e99d859a10044e4328d758a6/sdk/go/pulumi/resource.go#L263-L268

Note that other languages expose a `mergeOptions` functionality: 
https://github.com/pulumi/pulumi/blob/9abcca345a6c024087b35152b0d6fca92baef191/sdk/nodejs/resource.ts#L865

PR #3978, opened 2020-02-26
Support transformations in Go

### LINKED ISSUE #3490: Go: support `transformations`
Support this API in the Go SDK.

tasiah:
Closes #3490.

A few things to note:

- All the tests are in `lifecycle_test.go` because Go currently doesn't support dynamic providers.
    - I based the tests off `transformations_test.go`. ~~At the time of writing, #3938 hadn't been merged, so I chose `Aliases` as our `ResourceOption` for testing. I could/might change all the instances of `Aliases` to `AdditionalSecretOutputs` if desired.~~ The tests in `lifecycle_test.go` should be roughly comparable to those in `transformations_test.go`.
    - I spent a considerable amount of time on trying to convert the fifth test in `transformations_test.go` (which tests ["cross-resource transformations that inject dependencies on one resource into another](https://github.com/pulumi/pulumi/blob/master/tests/integration/transformations/python/simple/__main__.py#L99)) without much success, likely due to the fact that I don't have much experience with Go channels or Goroutines. I would super appreciate some insight about how to go about creating this test and/or whether it is even necessary.
- Because we need to modify the root stack in [`context.go`](https://github.com/pulumi/pulumi/compare/tasiah/go-transforms?expand=1#diff-0f443da88b4eb884910f6486f57cedf9R1006), I needed to change `ctx.stack` from a struct to a [pointer to a struct](https://github.com/pulumi/pulumi/compare/tasiah/go-transforms?expand=1#diff-234c3947c3219c4d13a9156702f12eb8R87). In theory, the only thing that this changed is the fact that the interface type `Resource` now expects a pointer instead of a struct. (See: `lifecycle_test.go`, but the parts that changed that have nothing to do with transformations.)


tasiah:
I would switch this PR to a draft if I [could](https://github.com/isaacs/github/issues/1547), but since I can't, please ignore the testing stuff temporarily! I decided just to swap out all instances of `pulumi.Aliases` for `pulumi.AdditionalSecretOutputs` for consistency with `transformations_test.go`.

sdk/go/pulumi/context.go
```diff
@@ -277,6 +277,36 @@ func (ctx *Context) ReadResource(
...
+				o.applyResourceOption(resOptions)
+			}
+
+			if resOptions.Parent != nil && resOptions.Parent.URN() != options.Parent.URN() {
+				return errors.New("transformations cannot currently be used to change the `parent` of a resource")
+			}
+			props = res.Props
+			options = resOptions
+		}
+	}
```
lukehoban: The fact that we do this on Read was actually surprising to me - I’m not sure it was intentional that we apply transformations to Reads for other languages! - but this does appear to be consistent.

sdk/go/pulumi/context.go
```diff
@@ -955,3 +1018,13 @@ func (ctx *Context) RegisterResourceOutputs(resource Resource, outs Map) error {
 func (ctx *Context) Export(name string, value Input) {
 	ctx.exports[name] = value
 }
+
+// RegisterStackTransformation adds a transformation to all future resources constructed in this Pulumi stack.
+func (ctx *Context) RegisterStackTransformation(t ResourceTransformation) error {
+	if ctx.stack == nil {
+		return errors.New("the root stack resource was referenced before it was initialized")
+	}
```
lukehoban: Is it possible for this to happen in the Go SDK?

sdk/go/pulumi/resource.go
```diff
@@ -159,6 +175,10 @@ type resourceOptions struct {
 	Aliases []Alias
 	// AdditionalSecretOutputs is an optional list of output properties to mark as secret.
 	AdditionalSecretOutputs []string
+	// Transformations is an optional list of transformations to apply to this resource during construction.
+	// The transformations are applied in order, and are applied prior to transformation and to parents
+	// walking from the resource up to the stack.
```
lukehoban: This comment is a little awkward - is it the same as in other languages (might be my poor writing that’s getting copied in here!).

pkg/engine/lifecycle_test.go
```diff
@@ -5216,6 +5216,197 @@ func TestSingleResourceDefaultProviderGolangLifecycle(t *testing.T) {
...
+			MonitorAddr: info.MonitorAddress,
+		})
+		assert.NoError(t, err)
+
+		return pulumi.RunWithContext(ctx, func(ctx *pulumi.Context) error {
+			// Scenario #1 - apply a transformation to a CustomResource
+			res1Transformation := func(args *pulumi.ResourceTransformationArgs) *pulumi.ResourceTransformationResult {
+				return &pulumi.ResourceTransformationResult{
+					Props: args.Props,
+					Opts:  append(args.Opts, pulumi.AdditionalSecretOutputs([]string{"output"})),
```
lukehoban: Ideally we would use a `mergeOptions` style API here - but I think we don’t have it yet in Go. Not critical - but possibly worth a TODO pointing at the issue tracking that.

lukehoban:
>  I would super appreciate some insight about how to go about creating this test and/or whether it is even necessary.

I think it is fine to not add this test for the Go SDK. It was mostly added initially to ensure the design of Transformations could support that use case - not as much to test a correct implementation of that design.

sdk/go/pulumi/resource.go
tasiah: Yup, that comment is taken from the python implementation: https://github.com/pulumi/pulumi/blob/master/sdk/python/lib/pulumi/resource.py#L355-#L359

sdk/go/pulumi/context.go
tasiah: I don't think so, but having the extra safety check here probably doesn't hurt.

sdk/go/pulumi/transformation.go
```diff
@@ -0,0 +1,33 @@
...
+// ResourceTransformationArgs is the argument bag passed to a resource transformation.
+type ResourceTransformationArgs struct {
+	// The resource instance that is being transformed.
+	Resource Resource
+	// The type of the resource.
+	Type string
+	// The name of the resource.
+	Name string
+	// The original properties passed to the resource constructor.
+	Props Input
```
EvanBoyle: Is this the appropriate type? Both python and node use a mapping of str-> input:

https://github.com/pulumi/pulumi/blob/9abcca345a6c024087b35152b0d6fca92baef191/sdk/nodejs/resource.ts#L557
https://github.com/pulumi/pulumi/blob/33b232adc4737144c9e8626343050059cec032a7/sdk/python/lib/pulumi/resource.py#L232

sdk/go/pulumi/context.go
tasiah: Going to remove this check as it seems unnecessary.

tasiah: Going to remove this check since it seems unnecessary.

sdk/go/pulumi/context.go
```diff
@@ -399,6 +429,36 @@ func (ctx *Context) RegisterResource(
...
+			}
+
+			if resOptions.Parent != nil && resOptions.Parent.URN() != options.Parent.URN() {
+				return errors.New("transformations cannot currently be used to change the `parent` of a resource")
+			}
+			props = res.Props
+			options = resOptions
+		}
+	}
+
```
EvanBoyle: Looks like this block of code is the same in both `ReadResource` and `RegisterResource`. Perhaps we could factor it out into a private method and share this code between them.

sdk/go/pulumi/context.go
```diff
@@ -277,6 +277,36 @@ func (ctx *Context) ReadResource(
 		options.Parent = ctx.stack
 	}
 
+	// Before anything else, if there are transformations registered, give them a chance to run to modify the
+	// user-provided properties and options assigned to this resource.
+	transformations := options.Transformations
+	if options.Parent != nil {
```
EvanBoyle: How are stack transformations handled? Does the engine automatically apply them to all resources, or do we need to copy them to children resources as well?

lukehoban: The stack is the parent of all un-parented resources. See https://www.pulumi.com/docs/intro/concepts/programming-model/#parent.

sdk/go/pulumi/transformation.go
lukehoban: In Go, the concrete type here will typically be an Args struct - so definitely this is intentionally not the same as TypeScript/Go.

I was actually wondering something similar though.  It might be nice to have a test case that actually reads the props and creates a new one.  I assume it would need to cast from this type down into the concrete type expected as input for the resource?   I didn't see any tests covering that - but it will be a reasonably common usage mode of this feature.

pkg/engine/lifecycle_test.go
```diff
@@ -5216,6 +5216,199 @@ func TestSingleResourceDefaultProviderGolangLifecycle(t *testing.T) {
...
+
+				return &pulumi.ResourceTransformationResult{
+					Props: props,
+					Opts:  args.Opts,
+				}
+			}
+			assert.NoError(t, ctx.RegisterStackTransformation(res3Transformation))
+			assert.NoError(t, newResource(ctx, "res3"))
+
+			// Scenario #4 - transformations are applied in order of decreasing specificity
```
EvanBoyle: This is consistent with other languages (I see that the comment was copied) but I find this surprising. The most specific transformations can get clobbered by the least specific. Any background on this @lukehoban @pgavlin?

lukehoban: Yes - this feature is a "dependency injection" style feature, which is all about giving consumers the power to overwrite what a component or abstraction they are using does by default.  So for this feature in particular, we want to give ultimate authority to the outermost caller.

sdk/go/pulumi/transformation.go
tasiah: @pgavlin was trying to help me modify a `pulumi.Input`, but we ran into some issues when trying to run this snippet:
```
props := pulumi.All(args.Props).ApplyT(func(arr []interface{}) *testResourceArgs {
					props := arr[0].(*testResourceArgs)
					props.Foo = "baz"
					return props
				})
```
He thinks it's a bug in `awaitInputs` and started a debugging branch [here](https://github.com/pulumi/pulumi/compare/pgavlin/debugAwait). Simply put, I don't think the suggested test is possible until we figure out what's wrong with `awaitInputs`, which is why I straight up replace `props` in my tests rather than modifying them. If there's any alternative ways to do this, please let me know!

lukehoban: Interesting.  Is it possible to at least read a value out of an `Input` and then construct a new `Input` that includes part of the initial one? 

This is morally similar to the following from the TypeScript tests which doesn't mutate, but does "copy over" properties from the incoming properties.  (There is of course no equivalent to the `...` in Go, we would have to enumerate all the properties of the Args type).

```ts
                 return {
                    props: { optionalInput: "newDefault", ...props },
                    opts: pulumi.mergeOptions(opts, { additionalSecretOutputs: ["output"] }),
                };
```

tasiah: To be honest, I'm not super clear on how `pulumi.Input` works, so I don't really know how to extract values from it, outside of what Pat showed me with that snippet above. Maybe @pgavlin or @EvanBoyle can give us more insight?

lukehoban: @pgavlin I would have hoped there was an easy way to cast type `Props` had here directly to an `Args` object.  Is that not the case?  Similar to how in .NET this is:

```cs
public ResourceArgs Args { get; }
```

Which is a base type for all Args objects such that casting to the known actual arg type is straightforward.

The `pulumi.All(args.Props).ApplyT(func(arr []interface{}) *testResourceArgs {` dance above does not look ideal here.

tasiah:
> We should also update the programming model docs as a part of this with a go example

I'm wondering if we should wait to update the docs until we finish the `mergeOptions` equivalent for Go (#3846) for consistency with the other languages. Thoughts?

pkg/engine/lifecycle_test.go
```diff
@@ -5216,6 +5216,199 @@ func TestSingleResourceDefaultProviderGolangLifecycle(t *testing.T) {
...
+				// TODO[pulumi/pulumi#3846] We should use a mergeOptions-style API here.
+				return &pulumi.ResourceTransformationResult{
+					Props: args.Props,
+					Opts:  append(args.Opts, pulumi.AdditionalSecretOutputs([]string{"output"})),
+				}
+			}
+			assert.NoError(t, newResource(ctx, "res1",
+				pulumi.Transformations([]pulumi.ResourceTransformation{res1Transformation})))
+
+			// Scenario #2 - apply a transformation to a Component to transform it's children
```
pgavlin:
```suggestion
			// Scenario #2 - apply a transformation to a Component to transform its children
```

sdk/go/pulumi/transformation.go
pgavlin: So there's no way I can see for the type system to enforce that the `Input` to `RegisterResource` is not an `Output`, but IIRC we do have runtime checks to ensure that it is a pointer to a struct, and we could extend those to also ensure that it is not an `Output`. This would allow users to assume that the type is a top-level `Args` type, and we could write this:

```
props := args.Props.(*testResourceInputs)
props.Foo = pulumi.String("baz")
```

@tasiah can you try changing your test to that approach?

lukehoban:
:tada:

PR #3984, opened 2020-02-26
Add support for extracting schemas from providers.

pgavlin:
These changes add a new method to the resource provider gRPC interface,
`GetSchema`, that allows consumers of these providers to extract
JSON-serialized schema information for the provider's types, resources,
and functions.


sdk/go/pulumi/config/try.go
```diff
@@ -27,7 +27,8 @@ import (
 func Try(ctx *pulumi.Context, key string) (string, error) {
 	v, ok := ctx.GetConfig(key)
 	if !ok {
-		return "", fmt.Errorf("missing required configuration variable '%s'; run `pulumi config` to set", key)
+		return "",
+			fmt.Errorf("missing required configuration variable '%s'; run `pulumi config` to set", key)
```
EvanBoyle: This change is surprising. I wonder how it got out of sync with the generated code https://github.com/pulumi/pulumi/blob/master/sdk/go/pulumi/templates/config-try.go.template

pgavlin: Oh, that is weird. I'll roll that bit back.

PR #3986, opened 2020-02-26
fix panic when dest is struct and input is ptr

EvanBoyle:
Fixes https://github.com/pulumi/pulumi-azure/issues/459

Minimal repro:

```go
package main

import (
	"github.com/pulumi/pulumi-aws/sdk/go/aws/kms"
	"github.com/pulumi/pulumi-aws/sdk/go/aws/s3"
)

func main() {
		// Create a KMS Key for S3 server-side encryption
		key, err := kms.NewKey(ctx, "my-key", nil)
		if err != nil {
			return err
		}

		bucket, err := s3.NewBucket(ctx, "my-bucket", &s3.BucketArgs{
			ServerSideEncryptionConfiguration: s3.BucketServerSideEncryptionConfigurationArgs{
				Rule: &s3.BucketServerSideEncryptionConfigurationRuleArgs{
					ApplyServerSideEncryptionByDefault: s3.BucketServerSideEncryptionConfigurationRuleApplyServerSideEncryptionByDefaultArgs{
						SseAlgorithm:   pulumi.StringInput(pulumi.String("aws:kms")),
						KmsMasterKeyId: key.ID(),
					},
				},
			},
		})
		if err != nil {
			return err
		}

		return nil
```

This update works without changes if you pass `Rule: s3.BucketServerSideEncryptionConfigurationRuleArgs{` instead of the pointer. 

This change properly handles awaiting inputs when resolved is a concrete struct and the value is a pointer. 

There may be a cleaner way to achieve this (I'm not very familiar with the reflection library) but this works.

### DIFF
--- a/sdk/go/pulumi/types.go
+++ b/sdk/go/pulumi/types.go
@@ -572,6 +572,15 @@ func awaitInputs(ctx context.Context, v, resolved reflect.Value) (bool, bool, er
 
 		v, isInput = reflect.ValueOf(input), true
 
+		// We require that the kind of an `Input`'s `ElementType` agrees with the kind of the `Input`'s underlying value.
+		// This requirement is trivially (and unintentionally) violated by `*T` if `*T` does not define `ElementType`,
+		// but `T` does (https://golang.org/ref/spec#Method_sets).
+		// In this case, dereference the pointer to get at its actual value.
+		if v.Kind() == reflect.Ptr && valueType.Kind() != reflect.Ptr {
+			v = v.Elem()
+			contract.Assert(v.Interface().(Input).ElementType() == valueType)
+		}
+
 		// If we are assigning the input value itself, update the value type.
 		if assignInput {
 			valueType = v.Type()


sdk/go/pulumi/types.go
```diff
@@ -603,8 +603,12 @@ func awaitInputs(ctx context.Context, v, resolved reflect.Value) (bool, bool, er
 		}
 	case reflect.Ptr:
 		if !v.IsNil() {
-			resolved.Set(reflect.New(resolved.Type().Elem()))
-			return awaitInputs(ctx, v.Elem(), resolved.Elem())
+			if resolved.Kind() == reflect.Ptr {
+				resolved.Set(reflect.New(resolved.Type().Elem()))
+				return awaitInputs(ctx, v.Elem(), resolved.Elem())
+			}
+			// resolved is a struct
```
pgavlin: I'm surprised that this is possible: I'd have thought that the assert on line 589 (https://github.com/pulumi/pulumi/pull/3986/files#diff-b79899f0b74b310adc08a01fbcb77e8eR589) would have failed in this case. Do you know why it doesn't?

EvanBoyle: The error isn't because they aren't assignable. The panic is due to `resolved.Type().Elem()` which panics if the kind is not one of a specified list:

from libexec/src/reflect/type.go
```go
func (t *rtype) Elem() Type {
	switch t.Kind() {
	case Array:
		tt := (*arrayType)(unsafe.Pointer(t))
		return toType(tt.elem)
	case Chan:
		tt := (*chanType)(unsafe.Pointer(t))
		return toType(tt.elem)
	case Map:
		tt := (*mapType)(unsafe.Pointer(t))
		return toType(tt.elem)
	case Ptr:
		tt := (*ptrType)(unsafe.Pointer(t))
		return toType(tt.elem)
	case Slice:
		tt := (*sliceType)(unsafe.Pointer(t))
		return toType(tt.elem)
	}
	panic("reflect: Elem of invalid type")
}
```

pgavlin: > The error isn't because they aren't assignable.

A pointer value is not assignable to a struct value, which is why I would have thought that assertion would fire.

EvanBoyle: Ah, yes. The original panic was before the assertion:

https://github.com/pulumi/pulumi/pull/3986/files#diff-b79899f0b74b310adc08a01fbcb77e8eR581

EvanBoyle:
Here's a branch with some logs https://github.com/pulumi/pulumi/compare/evan/StructPtrPanicLogs?expand=1

The output when running the repro above is:

```
Evans-MBP:go_aws_qs evanboyle$ pulumi up
Previewing update (dev):
     Type                 Name           Plan     Info
     pulumi:pulumi:Stack  go_aws_qs-dev           1 error; 16 messages
 
Diagnostics:
  pulumi:pulumi:Stack (go_aws_qs-dev):
    panic: reflect: Elem of invalid type
    goroutine 52 [running]:
    reflect.(*rtype).Elem(0x1c2d500, 0xc0001d6e70, 0x16)
    	/usr/local/Cellar/go/1.13.8/libexec/src/reflect/type.go:920 +0x143
    github.com/pulumi/pulumi/sdk/go/pulumi.awaitInputs(0x1ef4200, 0xc0000380e8, 0x1c2d600, 0xc00031f1e0, 0x94, 0x1c2d500, 0xc000168c80, 0x199, 0x1f00000, 0x0, ...)
    	/Users/evanboyle/go/src/github.com/pulumi/pulumi/sdk/go/pulumi/types.go:584 +0x1b8d
    github.com/pulumi/pulumi/sdk/go/pulumi.awaitInputs(0x1ef4200, 0xc0000380e8, 0x1c8eb00, 0xc00031f1e0, 0x99, 0x1c2d400, 0xc000168c80, 0x199, 0x1380000, 0x0, ...)
    	/Users/evanboyle/go/src/github.com/pulumi/pulumi/sdk/go/pulumi/types.go:618 +0xcfd
    github.com/pulumi/pulumi/sdk/go/pulumi.toOutputWithContext.func1(0x1f07a80, 0x1c2d400, 0x1ef4200, 0xc0000380e8, 0x1c8eb00, 0xc00031f1e0, 0xc000118c00, 0x1f015e0, 0xc00025c620)
    	/Users/evanboyle/go/src/github.com/pulumi/pulumi/sdk/go/pulumi/types.go:705 +0x13f
    created by github.com/pulumi/pulumi/sdk/go/pulumi.toOutputWithContext
    	/Users/evanboyle/go/src/github.com/pulumi/pulumi/sdk/go/pulumi/types.go:702 +0x1e8
    exit status 2
 
    valueType: s3.BucketServerSideEncryptionConfigurationRule
    v.Type(): *s3.BucketServerSideEncryptionConfigurationRuleArgs
    resolved.Type(): s3.BucketServerSideEncryptionConfigurationRule
 


```

pgavlin:
Thanks! This makes it much clearer what exactly is going on. From [the Go spec](https://golang.org/ref/spec#Method_sets):

> The method set of the corresponding pointer type *T is the set of all methods declared with receiver *T or T (that is, it also contains the method set of T)

This fact interacts with our type system in a confusing manner. It means that if `T` is an `Input`, `*T` will _also_ be an input, even if `*T` does not define `ElementType` itself. Furthermore, it generally means that `*T` will be a `TInput`, not a `TPtrInput`. In fact, with today's codegen, `*T` will _never_ be a `TPtrInput`, even if it could: instead, the `*T` can be converted to a `TPtrInput` using a `TPtr` function exposed by the package that defines `T`. To be completely frank, I don't remember the rationale behind this design, and we should consider whether or not we can have `*T` always implement `TPtrInput` (which would hide `T`'s implementation of `TInput`).

All that said, even if we do update the codegen, we should better handle the case where `(*T).ElementType` returns `(T).ElementType`. I'd prefer that we write the fix in terms of this situation by adding something like the following between the code on lines 532 and 533:

```go
// We require that the kind of an `Input`'s `ElementType` agrees with the kind of the `Input`'s underlying value. This requirement is trivially (and unintentionally) violated by `*T` if `*T` does not define `ElementType`, but `T` does (link to spec). In this case, dereference the pointer to get at its actual value.
if v.Kind() == reflect.Ptr && valueType.Kind() != reflect.Ptr {
    v = v.Elem()
    contract.Assert(v.(Input).ElementType() == valueType)
}
```

EvanBoyle:
Thanks for the explanation. This fix needs to happen at a slightly different point. The kinds at that point are actually: 

```
v.kind(): interface
valueType.kind(): string
```
But the fix works inserted a few lines down after we manipulate the interface into a prt.

pgavlin:
> Thanks for the explanation. This fix needs to happen at a slightly different point. The kinds at that point are actually:
> 
> ```
> v.kind(): interface
> valueType.kind(): string
> ```
> 
> But the fix works inserted a few lines down after we manipulate the interface into a ptr.

This is surprising as well... What is the value of `v.String()` at that point?

EvanBoyle:
@pgavlin :

```
    valueType.Kind(): struct
    valueType.String(): s3.BucketServerSideEncryptionConfigurationRule
    v.Kind(): interface
    v.String(): <s3.BucketServerSideEncryptionConfigurationRuleInput Value>

```

pgavlin:
Okay, that helps.

PR #3990, opened 2020-02-27
Update gRPC library used in Go SDKs and regenerate code with latest tools

jen20:
In order to generate modern Go code using (for example) the standard library context package, we need to upgrade our GRPC generators. Historically generation was done in a container built from a dockerfile in this repo, but this PR moves to using the standard build container.


sdk/go/pulumi/config/try.go
```diff
@@ -27,7 +27,8 @@ import (
 func Try(ctx *pulumi.Context, key string) (string, error) {
 	v, ok := ctx.GetConfig(key)
 	if !ok {
-		return "", fmt.Errorf("missing required configuration variable '%s'; run `pulumi config` to set", key)
+		return "",
```
pgavlin: cc @EvanBoyle

EvanBoyle: If you run `make gen` in the go SDK you can see that these files are actually out of sync on master. We should check this in.

Surprised that a dirty git branch doesn't fail any of our builds.

jen20: I was also somewhat surprised that the dirty checking did not pick this up, but it is in fact a correct change so far as I can tell.

PR #3996, opened 2020-02-28
ignore internal properties when unmarshaling

EvanBoyle:
Fixes https://github.com/pulumi/pulumi-azure/issues/460

We are supposed to ignore unmarshaling any properties that begin with a double underscore `__`.

We have similar behavior in dotnet, node, etc. Not quite sure if this fix is in the right place. Will need @pgavlin to confirm.


EvanBoyle:
cc @elsesiy

elsesiy:
LGTM, thanks for the quick fix!

Issue #3671, opened 2019-12-17
Improve startup performance


lukehoban:
The shortest possible `pulumi preview` or `pulumi up` is currently around 4 seconds.  We can likely improve this significantly

A typical preview using TypeScript and AWS looks as follows:
* **750ms-1000ms**: talking to Pulumi Service
* **500ms**: booting NodeJS
* **500ms**: initializing Pulumi inside NodeJS (including TS compilation, but I recall V8 hooks stuff was actually a bigger part of this)
* **2500ms**: configuring AWS provider.

<img width="960" alt="Screen Shot 2019-12-16 at 9 17 49 AM" src="https://user-images.githubusercontent.com/223467/70956195-203ef900-2028-11ea-8c3c-0c1de06b0bf3.png">
<img width="974" alt="Screen Shot 2019-12-16 at 9 02 20 AM" src="https://user-images.githubusercontent.com/223467/70956196-20d78f80-2028-11ea-925e-7eb4c4abbac2.png">

There are some cases where things are overall much worse - on a fresh tethered network connection and without having run Pulumi recently I saw overall worse performance across many of these steps, including much longer waits on network calls.  On more limited machines or networks, these results may be more representative, and may lead to further optimization opportunities.

<img width="1345" alt="Screen Shot 2019-12-16 at 9 01 30 AM" src="https://user-images.githubusercontent.com/223467/70956194-203ef900-2028-11ea-838e-5a6ee7ee7984.png">

From the above - some potential opportunities:
- [ ] Optimize provider configuration (this is mostly inherited from the upstream providers for AWS/Azure/GCP, but we could contribute general performance improvements - like avoiding network calls that are made multiple times).
- [ ] Attempt to improve NodeJS process launch
- [ ] Attempt to improve Pulumi NodeJS bootstrap (like v8hooks load, potentially moving it off critical path)
- [ ] Move some of the network calls to the Pulumi service backend off the critical path (run concurrently with other work)
- [ ] Improve `api/getStack` performance in the Pulumi service backend

pgavlin:
> Move some of the network calls to the Pulumi service backend off the critical path (run concurrently with other work)

Which calls were you thinking of here?

lukehoban:
> > Move some of the network calls to the Pulumi service backend off the critical path (run concurrently with other work)

> Which calls were you thinking of here?

The 5 calls that happen synchronously at the beginning of every preview/update:
- `api/getStack`
- `api/createUpdate`
[...]

PR #4004, opened 2020-02-28
Avoid configuring providers twice during preview

lukehoban:
In the very common case where provider configuration does not change, during preview we were calling `Configure` on the cloud provider twice - once for the "old" configuration, and once for the "new" configuration.

This is not necessary, and we can just avoid using the new provider when configuration has not changed, since we will have configured the old provider very early so if we can use that we should.

Note that this technically doesn't prevent the second call to `Configure` from being made, but it prevents us from ever waiting on it.  We may want to go further and avoid even calling `Configure` on the provider in this case.

Part of #3671.


pkg/resource/deploy/providers/registry.go
```diff
@@ -298,12 +298,12 @@ func (r *Registry) Diff(urn resource.URN, id resource.ID, olds, news resource.Pr
 
 	// If the diff requires replacement, unload the provider: the engine will reload it during its replacememnt Check.
 	//
-	// If the diff does not require replacement and we are running a preview, register it under its current ID so that
-	// references to the provider from other resources will resolve properly.
+	// If the diff does not require replacement but does have changes, and we are running a preview, register it under
+	// its current ID so that references to the provider from other resources will resolve properly.
 	if len(diff.ReplaceKeys) != 0 {
```
pgavlin:
```suggestion
	if diff.Replace() != 0 {
```

pkg/resource/deploy/providers/registry.go
```diff
@@ -298,12 +298,12 @@ func (r *Registry) Diff(urn resource.URN, id resource.ID, olds, news resource.Pr
...
 		closeErr := r.host.CloseProvider(provider)
 		contract.IgnoreError(closeErr)
-	} else if r.isPreview {
+	} else if r.isPreview && len(diff.ChangedKeys) != 0 {
```
pgavlin:
```suggestion
	} else if r.isPreview && diff.Changes == plugin.DiffSome {
```

Issue #3939, opened 2020-02-19
Run dep ensure as part of Go tests in integration framework


stack72:
When we are running our Go integration tests, none of our code actually installs the code that is required to actually install the go dependencies

We need to make `dep ensure` run as part of the prepareGoProject

This means there is now a dependency on dep as part of our setup - this needs to be added to our common tool chain

lukehoban:
Note that we'll most likely have to run `pulumi plugin install` manually as well for now.

stack72:
@lukehoban do we need the plugins if we have the sdks installed?

lukehoban:
Yes - as of now we don't automatically acquire plugins as part of SDK installation.  We hope to enable this in the (near) future with https://github.com/pulumi/pulumi/issues/1549.

lukehoban:
We've decided to hold off on this to see if we can land Go modules support first.  Dep doesn't support "npm link"-style dependencies, which are critical for correctness of our testing framework - so there is more work needed if we do intend to keep `dep` much longer.

If we land do modules support by 3/13, we will close this out and move ahead with Go modules-based testing.  If not, we should move ahead with this and put in the work to ensure we have solid test support until we can get to Go modules.

PR #4005, opened 2020-02-29
Run Dep Ensure when testing Go integration projects

stack72:
Fixes: #3939

The work to create a temporary directory in the GOPATH will be removed when we move to Go modules but for now, this is a requirement for running the tests so it's a short(ish) term pain we need to take

This work now allows the Go language executor to be used on Windows. This was not initially possible due to the missing `.exe` on windows


pkg/testing/integration/program.go
```diff
@@ -1448,28 +1458,40 @@ func (pt *ProgramTester) copyTestToTemporaryDirectory() (string, string, error)
...
+			usr, userErr := user.Current()
+			if userErr != nil {
+				return "", "", errors.New("Couldn't get the current user")
+			}
+			gopath = filepath.Join(usr.HomeDir, "go")
+		}
+		folder := fmt.Sprintf("stackName-%d", time.Now().UnixNano())
+		tmpdir = filepath.Join(gopath, "src", folder)
+		err := os.Mkdir(tmpdir, 0700)
+		if err != nil {
```
lukehoban: This looks like the same logic as added above - should we share this?

stack72: Yeah I will refactor this to make it cleaner

.appveyor.yml
```diff
@@ -54,6 +54,8 @@ install:
     set PULUMI_LOCAL_NUGET=C:\Pulumi\nuget
     
     set PULUMI_TEST_OWNER=moolumi
+
+    choco install dep
```
lukehoban: This won’t necessarily get the same version we use for testing on other platforms. Should we force getting a specific version?

pkg/engine/project.go
```diff
@@ -21,7 +21,6 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
-
```
lukehoban: This separation into three blocks is intentional in our style across the codebase.

pkg/testing/integration/program.go
```diff
@@ -1422,6 +1433,10 @@ func (pt *ProgramTester) copyTestToTemporaryDirectory() (string, string, error)
 		return "", "", err
 	}
 
+	//if projinfo.Proj != nil && projinfo.Proj.Main != "" {
+	//	sourceDir = filepath.Join(sourceDir, projinfo.Proj.Main)
+	//}
+
```
lukehoban: Remove?

pkg/testing/integration/program.go
```diff
@@ -1710,13 +1732,34 @@ func (pt *ProgramTester) prepareGoProject(projinfo *engine.Projinfo) error {
...
+		err = pt.runCommand("dep-ensure", []string{goDepBin, "ensure", "-v"}, cwd)
+		if err != nil {
+			return fmt.Errorf("error installing go dependencies: %w", err)
+		}
 	}
 
+	// In our go tests, there seems to be an issue where we *need* to make a build before we
+	// actually run the tests. If a build isn't made, then we get an issue along the lines of
+	// `cannot import absolute path`. We will investigate this and add back in the respect
+	// for the runBuild option
```
lukehoban: I’d love to identify what this issue is and open a bug to track it.

tests/integration/config_basic/go/Gopkg.toml
```diff
@@ -0,0 +1,7 @@
+[[constraint]]
+  version = "v1.11.1"
+  name = "github.com/pulumi/pulumi"
```
lukehoban: Does this mean we will be testing against 1.11.1 - not against what is actually being built?

We definitely need to test against the current code built locally for the Pulumi/Pulumi dep.

.appveyor.yml
stack72: This is now redundant as this is going away since we moved to GHA

pkg/engine/project.go
stack72: Fixed!

pkg/testing/integration/program.go
stack72: Fixed :)

pkg/testing/integration/program.go
stack72: Yeah I will open an issue to track this for sure

tests/integration/config_basic/go/Gopkg.toml
jen20: This is not going to be easy since dep is not module aware. I’d recommend not doing this until we have modules in the Go SDK (soon).

lukehoban: You mean running tests that don’t test the thing under test?  Or you mean not running tests at all?

If we wanted, I assume we could use `Dependencies` to whack the vendor tree after running dep so that we can test the right thing?

stack72: @lukehoban it seems that Dep doesn't support relative imports :/

https://golang.github.io/dep/docs/FAQ.html#does-dep-support-relative-imports

I am not sure of a way around this for now - I know we can do it when we move to go modules because we can replace the module with the local path

Thoughts on what you want to do? Can we suffer the short term pain here just to ensure something is being tested for go?

lukehoban: I'm not sure - but can't we just overwrite `vendor` after running `dep`?  We used to play this trick when we used `dep` more widely of deleteing `vendor/something`, then copying into `vendor/something`, then deleting `vendor/soemthing/vendor`.  

At least for the tests in this repo - it is a regression in correctness of the tests to do what this PR currently does.  Previously, the test would have picked up the locally built and installed Go SDK - now they will vendor in an old one.  That won't work.

For tests in `examples` though - they may benefit from the behaviour here as they don't need anything built locally.

jen20: I’m proposing we do nothing here until Go modules are fully supported. We’d need to merge the vendor tree to do what you propose, but that only works assuming only limited features of modules are used anyway.

stack72:
@lukehoban as suggested, all of the Gopkg.toml files have been removed and now we are building from the local version of the SDK

I still understand that this is throwaway work and that this will be superceded by the Go mod work, but at least we can now effectively test our Go applications

I'd rather we merged this so we get that ability and then I will happily change the PR to move to Go mod later

tests/integration/config_basic/go/Gopkg.toml
stack72: This is no longer the case, we are building from the local SDK and don't need any Gopkg.toml files

PR #4015, opened 2020-03-03
Initial support for configuring policies

justinvp:
Apologies for the larger PR. It's broken up into more digestible commits.

This adds support for the `--policy-pack-config` flag and underlying implementation to apply configuration to policies. It also includes support for sending config schema to the service and using configuration from the service.

The `pulumi/pulumi-policy` side of the change is here: https://github.com/pulumi/pulumi-policy/pull/207

Fixes https://github.com/pulumi/pulumi-policy/issues/185
Fixes https://github.com/pulumi/pulumi-policy/issues/186
Fixes https://github.com/pulumi/pulumi-policy/issues/189
Fixes https://github.com/pulumi/pulumi-policy/issues/190

Part of https://github.com/pulumi/pulumi-policy/issues/187
Part of https://github.com/pulumi/pulumi-policy/issues/188


pkg/engine/update.go
```diff
@@ -229,10 +245,30 @@ func installAndLoadPolicyPlugins(plugctx *plugin.Context, policies []RequiredPol
...
+		// If we have configuration from the API, pass it to the analyzer.
+		if len(configFromAPI) > 0 {
+			analyzerInfo, err := analyzer.GetAnalyzerInfo()
+			if err != nil {
+				return err
+			}
+			config, err := plugin.ReconcilePolicyPackConfig(analyzerInfo.Policies, configFromAPI)
+			if err != nil {
+				return errors.Wrapf(err, "reconciling configuration for analyzer %q", analyzerInfo.Name)
+			}
```
justinvp: Not: An alternative to retrieving the metadata and reconciling the config here would be to just pass the config directly from the service (or from a `--policy-pack-config` specified file) to the Policy Pack. However, then each language SDK for policies would have to handle reconciling its default values with the specified config. Looking ahead to implementing the SDK for Python, .NET, and Go, it seemed "better" to have a single implementation here for doing this reconciliation, rather than writing it N times for each language.

pkg/resource/plugin/analyzer_config.go
```diff
@@ -0,0 +1,280 @@
...
+			resultConfig.Properties = make(map[string]interface{})
+		}
+		for k, v := range givenConfig.Properties {
+			resultConfig.Properties[k] = v
+		}
+	}
+
+	// Validate the resulting config.
+	if config != nil {
+		if err := ValidatePolicyPackConfig(policies, result); err != nil {
```
justinvp: Validation errors will be things like if a required config value is missing, or if the specified value is the incorrect type, etc.

But I actually don't think this is the right long-term place to be outputting such errors. This is what output currently looks like; it's just an error that stops the preview/update:

```bash
$ pulumi preview --policy-pack ~/path/to/policypack --policy-pack-config ~/path/to/policypack/config.json
Previewing update (dev):
 
Permalink: https://app.pulumi.com/<org>/<proj>/dev/previews/ae0e97ff-cd01-4071-9f71-746b1dca0695
error: error reconciling configuration for analyzer "aws-typescript" from path "/Users/user/path/to/policypack: 1 error occurred:
	* Policy "s3-no-public-read": foo is required
```

Would it be better if these were outputted similar to other policy diagnostic events?

pkg/resource/plugin/analyzer_plugin.go
```diff
@@ -311,6 +329,42 @@ func (a *analyzer) GetPluginInfo() (workspace.PluginInfo, error) {
...
+
+	_, err := a.client.Configure(a.ctx.Request(), &pulumirpc.ConfigureAnalyzerRequest{
+		PolicyConfig: c,
+	})
+	if err != nil {
+		rpcError := rpcerror.Convert(err)
+		// Handle the case where the policy pack doesn't implement a recent enough
+		// AnalyzerService to support the Configure method. Ignore the error as it
+		// just means the analyzer isn't capable of being configured in this way.
+		if rpcError.Code() == codes.Unimplemented {
```
ekrengel: It seems like this would be an error case to me. I would think that we may want to return early if the `policyConfig` is null or len == 0. This should handle the case where the analyzer is older and does not support config. If the analyzer is older and we somehow have config, something has gone wrong... right?

pkg/resource/plugin/analyzer_config.go
ekrengel: I think yes it probably should be similar to a diagnostic event, but maybe a new kind since this wont be tied to a particular resource 🤔 ... if its not a diagnostic event the service wont know about it!

I am also wondering if we should be validating the nil config. What happens if the config has an empty required field?

pkg/backend/httpstate/client/client.go
```diff
@@ -552,11 +552,22 @@ func (pc *Client) PublishPolicyPack(ctx context.Context, orgName string,
 		return "", err
 	}
 
+	policies := make([]apitype.Policy, len(analyzerInfo.Policies))
```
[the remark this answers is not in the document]
justinvp: Yep, I ended up changing it in a later commit to not specify the length and always append.

sdk/python/lib/pulumi/runtime/proto/analyzer_pb2_grpc.py
```diff
@@ -39,6 +39,11 @@ def __init__(self, channel):
         request_serializer=google_dot_protobuf_dot_empty__pb2.Empty.SerializeToString,
         response_deserializer=plugin__pb2.PluginInfo.FromString,
         )
+    self.Configure = channel.unary_unary(
```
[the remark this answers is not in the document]
justinvp: Python and go still get generated when changes are made to the .proto files.

pkg/resource/plugin/analyzer_config.go
```diff
@@ -0,0 +1,280 @@
...
+
+	"github.com/hashicorp/go-multierror"
+	"github.com/pkg/errors"
+	"github.com/pulumi/pulumi/pkg/apitype"
+	"github.com/pulumi/pulumi/pkg/util/contract"
+	"github.com/xeipuuv/gojsonschema"
+)
+
+// LoadPolicyPackConfigFromFile loads the JSON config from a file.
+func LoadPolicyPackConfigFromFile(file string) (map[string]AnalyzerPolicyConfig, error) {
```
lukehoban: Somehow it feels odd that these are in the `plugin` package.  What was the reason for that?

pkg/resource/plugin/analyzer_config.go
```diff
@@ -0,0 +1,280 @@
...
+		}
+		enforcementLevel = el
+		// Remove enforcementLevel from the map.
+		delete(props, "enforcementLevel")
+	}
+	return enforcementLevel, nil
+}
+
+// ValidatePolicyPackConfig validates the policy pack's configuration.
+func ValidatePolicyPackConfig(policies []AnalyzerPolicyInfo, config map[string]AnalyzerPolicyConfig) error {
```
lukehoban: Is this called?

pkg/resource/plugin/analyzer_plugin.go
```diff
@@ -311,6 +329,42 @@ func (a *analyzer) GetPluginInfo() (workspace.PluginInfo, error) {
...
 }
 
+func (a *analyzer) Configure(policyConfig map[string]AnalyzerPolicyConfig) error {
+	label := fmt.Sprintf("%s.Configure(...)", a.label())
+	logging.V(7).Infof("%s executing", label)
+
+	c := make(map[string]*pulumirpc.PolicyConfig)
+
+	for k, v := range policyConfig {
+		var el pulumirpc.EnforcementLevel
```
lukehoban: We are defaulting into the zero value - is that right?  Should we be explicit about what the default is here?

pkg/resource/plugin/analyzer_plugin.go
```diff
@@ -361,6 +415,117 @@ func marshalProvider(provider *AnalyzerProviderResource) (*pulumirpc.AnalyzerPro
...
+	case apitype.Mandatory:
+		return pulumirpc.EnforcementLevel_MANDATORY
+	case apitype.Disabled:
+		return pulumirpc.EnforcementLevel_DISABLED
+	}
+	contract.Failf("Unrecognized enforcement level %s", el)
+	return 0
+}
+
+func marshalMap(m map[string]interface{}) *structpb.Struct {
```
lukehoban: A shame we have to write all this - isn't there something already that does this?

pkg/backend/httpstate/client/client.go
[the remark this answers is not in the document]
justinvp: Changed all cases to specify the length up-front and insert via index instead of appending.

pkg/resource/plugin/analyzer_plugin.go
```diff
@@ -259,21 +262,36 @@ func (a *analyzer) GetAnalyzerInfo() (AnalyzerInfo, error) {
...
 		enforcementLevel, err := convertEnforcementLevel(p.EnforcementLevel)
 		if err != nil {
 			return AnalyzerInfo{}, err
 		}
 
-		policies = append(policies, apitype.Policy{
+		schema := convertConfigSchema(p.GetConfigSchema())
+
+		// Inject `enforcementLevel` into the schema.
+		if schema != nil {
```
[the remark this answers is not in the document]
justinvp: > Do we want to support configurable enforcement levels even if there is no specific configuration for the policy? I would think yes, but then I imagine it may be hard to distinguish btw a policy pack that supports config but has no config and a policy pack that does not support config.

Yes, we definitely want configurable enforcement levels even if there is no config schema for a policy. I removed the nil check and am now always injecting enforcementLevel into the schema, even if the policy did not have it's own config schema defined.

From the service's perspective, is not seeing any schema for any of the policies how it is determining whether a policy pack is configurable?

If so, I'll need some other way in the CLI to determine if the policy pack supports config. I could add a new property `supportsConfig` to the metadata it gets back over the GRPC interface, which would default to false for older policy packs. Or add a new `SupportsConfig()` RPC that could be called to check. Or, since we haven't released a new `@pulumi/policy` that uses the version tag, I could check the version property you added. If that's a non-empty string, then the policy pack supports config. Leaning towards this latter approach.

pkg/backend/httpstate/client/client.go
```diff
@@ -552,11 +552,28 @@ func (pc *Client) PublishPolicyPack(ctx context.Context, orgName string,
 		return "", err
 	}
 
+	var policies []apitype.Policy
```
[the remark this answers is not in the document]
justinvp: Done.

pkg/resource/plugin/analyzer_config.go
justinvp: Instead of returning these as a single combined error, I added a commit that emits these as individual diagnostic messages. This allows these errors to be seen in the Pulumi Console.

Screen Shot 2020-03-06 at 9 11 20 AM

I also fixed it so all validation errors for all policy packs are returned, instead of returning early when the first policy pack had validation errors. 

We should still consider adding a new diagnostic event for these, which would allow improved grouping and display, but that requires changes to both the CLI and service and can be done as a subsequent improvement.

justinvp: > I am also wondering if we should be validating the nil config. What happens if the config has an empty required field?

Yep, fixed.

pkg/backend/httpstate/client/client.go
```diff
@@ -608,6 +625,27 @@ func (pc *Client) PublishPolicyPack(ctx context.Context, orgName string,
 	return version, nil
 }
 
+// convertPolicyConfigSchema converts a policy's schema from the analyzer to the apitype.
+func convertPolicyConfigSchema(s *plugin.AnalyzerPolicyConfigSchema) (*apitype.PolicyConfigSchema, error) {
```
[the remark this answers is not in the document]
justinvp: Done.

pkg/resource/plugin/analyzer_plugin.go
justinvp: Right now, the implementation always creates initial config values based on the declared enforcement level for each policy... This happens even for older policy packs that don't support config. I'll need to somehow determine the policy pack doesn't support config (using one of the approaches I mention at https://github.com/pulumi/pulumi/pull/4015#discussion_r389030415), and if that's the case, avoid doing anything with config (i.e. don't send any config schema and don't call the `Configure` rpc).

pkg/resource/plugin/analyzer_config.go
justinvp: Yea, felt a bit odd to me as well. Moved all this out of `plugin` to it's own `analyzer` package.

pkg/resource/plugin/analyzer_config.go
justinvp: Yes, but inside the package. Changed to lowercase.

pkg/resource/plugin/analyzer_plugin.go
justinvp: Changed it to require a valid enforcement level, rather than defaulting like this. In practice, all current callers are going to pass valid enforcement levels given how we build-up the config.

pkg/resource/plugin/analyzer_plugin.go
justinvp: Yes, it's unfortunate there are no built-in helpers for this. https://github.com/golang/protobuf/issues/370

pkg/resource/plugin/analyzer_plugin.go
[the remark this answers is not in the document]
justinvp: I went ahead and added another commit that adds the `supportsConfig` bool property to the metadata returned from the policy pack. For older policy packs, this will come back as the default value of `false`, in which case we do not send any config schema for such policy packs so the service can distinguish whether a policy pack is configurable or not based on whether it has config schema.

pkg/resource/plugin/analyzer_plugin.go
justinvp: Fixed.

Issue #3593, opened 2019-12-01
FileArchive does not preserve executable / file permission bits


ryankoppenhaver:
The `pulumi.asset.FileArchive` constructor accepts a path to either an archive file or a directory.  If the directory form is used, or if Pulumi needs to transcode the file (e.g. converting a `tar.gz` to a `zip`), the files in the resulting artifact do not have the executable bit set, even if the original files do.

For example, given an executable file at `./data/example`, the following code creates an AWS Lambda function that should run that file:

```javascript
const testLambda = new aws.lambda.Function("testLambda", {
  role: defaultRole.arn,
  code: new pulumi.asset.FileArchive("./data"),
  handler: "example",
  runtime: "go1.x"
})
```

Instead, lambda execution fails with:

```json
{
  "errorMessage": "fork/exec /var/task/example: permission denied",
  "errorType": "PathError"
}
```
(Note: you can reproduce this without a Go lambda binary handy; any Linux executable, including a shell script, will do, in which case `Process exited before completing request` would be the "success" case.)

Full reproduction: [index.js.txt](https://github.com/pulumi/pulumi/files/3908813/index.js.txt)

squaremo:
I think the executable bit is no longer required. I can create a lambda using this program:
https://gist.github.com/squaremo/b7e7e078d06186b5a245500dd28cd2b2

When invoked, it exits with

    {"errorMessage":"RequestId: 9c34ae9b-99f2-45b0-a9fb-9f79dc6a5759 Error: Runtime exited with error: exit status 1","errorType":"Runtime.ExitError"}

which I take to be the equivalent of the success case mentioned above. If I provide a nonsense file (e.g., with a leading newline), I get

    {"errorMessage":"fork/exec /var/task/handler: exec format error","errorType":"PathError"}

If I compile a Go program, `chmod a-x`, ZIP it, and use that ZIP file from the program, I can invoke the resulting lambda and get the expected output rather than an error.

The issue is titled "FileArchive does not preserve executable / file permission bits", which is still true; but, the actual problem for which that's a solution seems to have gone away. I'm going to close this. If there are other problem arising from file archive permissions, or evidence it's not fixed after all, please reopen.

PR #4018, opened 2020-03-05
set default archive format to ZIP, set appropriate executable permissions

EvanBoyle:
Closes https://github.com/pulumi/pulumi/issues/3593

This change sets the default archive format to ZIP over tar, as lambda only accepts zip artifacts. Not sure if this change in behavior is breaking or needs more care, but hopefully I can get some input from the group. 

It also sets the appropriate file permissions on the included files (executable).

With this change, we can skip `zip` steps in our examples.

something like:
```go
		args := &lambda.FunctionArgs{
			Handler: pulumi.String("handler"),
			Role:    role.Arn,
			Runtime: pulumi.String("go1.x"),
			Code:    pulumi.NewFileArchive("./handler/handler.zip"),
		}
```
can just be:
```go
		args := &lambda.FunctionArgs{
			Handler: pulumi.String("handler"),
			Role:    role.Arn,
			Runtime: pulumi.String("go1.x"),
			Code:    pulumi.NewFileArchive("./handler"),
		}
```
ref: https://github.com/pulumi/examples/blob/master/aws-go-lambda/main.go#L47-L52


lukehoban:
> This change sets the default archive format to ZIP over tar, as lambda only accepts zip artifacts

What issue exactly is this fixing?  Archives worked with Lambda previously - right?

EvanBoyle:
@lukehoban as far as I'm aware archives have never worked with lambda if you specify a directory. If you specify a pre-existing zip file it works just fine. 

This fixes two things:

1. Use zip instead of tar for archives by default. Lambda has never accepted tar files, so this behavior as never worked in the first place.
2. Set the appropriate file permissions so that the files in the archive are executable in the lambda runtime.

lukehoban:
I don't quite follow. We use `FileArchive(<folder>)` all the time already today with Lambda, Google Cloud Functions and Azure Functions: https://github.com/pulumi/examples/search?q=FileArchive&unscoped_q=FileArchive.

EvanBoyle:
@lukehoban just to clarify, the reason why this works with other languages on AWS (python, nodejs, dotnet) is that none of them produce executable files. 

The (AWS Go Lambda docs)[https://docs.aws.amazon.com/lambda/latest/dg/golang-package.html] are very explicit about requiring a zip file. In addition, the property in CloudFormation and the AWS CLI is called `zip-file`. I have been unable to find any working examples of uploading a code bundle as a `tar` on AWS lambda and having it work for executables. All of the examples I've found using `tar` in lambda use other languages to extract the archive into `/tmp` and then `chmod` the resulting files.

My suggestion here is that we make a __technically breaking__ change for 2.0 to move the default format of `FileArchive` from `tar` to `zip`.

sdk/go/common/resource/asset.go
```diff
@@ -1108,9 +1110,9 @@ func (a *Archive) EnsureHash() error {
 				return err
 			}
 		} else {
-			// Otherwise, it's not an archive; we'll need to transform it into one.  Pick tar since it avoids
-			// any superfluous compression which doesn't actually help us in this situation.
-			err := a.Archive(TarArchive, hash)
+			// Otherwise, it's not an archive; we'll need to transform it into one.
+			// Pick ZIP as default for compat with lambda.
+			err := a.Archive(ZIPArchive, hash)
```
lukehoban, line 1115: I'm still confused why this is needed.

This code doesn't actually produce an archive, it just computes the hash.  Why does it matter whether we use TarArchive or ZipArchive for computing the hash?

We do ultimately provide a ZIP to Lambda per:

https://github.com/pulumi/pulumi-terraform-bridge/blob/25b33bb489e0075e7733a7ee5a4c6bf33a08b8e6/pkg/tfbridge/assets.go#L164-L164

Are you *sure* this change is necessary as part of the fix?

EvanBoyle: @lukehoban I'm sure that if you remove this line, that we get the same error as without this change when running the lambda:
```sh
Evans-MBP:aws-go-lambda evanboyle$ aws lambda invoke --function-name $(pulumi stack output lambda) --region $(pulumi config get aws:region) --payload '"foo"' output.json
{
    "StatusCode": 200,
    "FunctionError": "Unhandled",
    "ExecutedVersion": "$LATEST"
}

Evans-MBP:aws-go-lambda evanboyle$ cat output.json 
{"errorMessage":"fork/exec /var/task/handler: permission denied","errorType":"PathError"}
```

With this fix:

```sh
Evans-MBP:aws-go-lambda evanboyle$ aws lambda invoke --function-name $(pulumi stack output lambda) --region $(pulumi config get aws:region) --payload '"foo"' output.json
{
    "StatusCode": 200,
    "ExecutedVersion": "$LATEST"
}
Evans-MBP:aws-go-lambda evanboyle$ cat output.json 
"FOO"
Evans-MBP:aws-go-lambda evanboyle$ 
```

Feel free to try it out locally. I'm not sure why this is necessary, but `EnsureHash` does store the resulting hash on the archive object. 

We do write out the hash when serializing as well: https://github.com/pulumi/pulumi/blob/84759ce01dbc1717a499a92dba924a780db12ce6/sdk/go/common/resource/asset.go#L227

MarcoPolo:
Looking forward to this! I ran into this issue today when trying to get a Rust lambda live: https://github.com/MarcoPolo/hello-rust-pulumi/blob/master/index.ts#L25.

squaremo:
After reading through code and looking at AWS documentation, I had the very strong impression that an executable bit was required on binaries.
I tried to recreate the problem, using the basic example of a Go handler program (in function/main.go) and a (Go) Pulumi program. I got these results:

| When I do everything as usual, except ... | the result is ... |
|----|----|
| nothing | a working function |
| Upload a ZIP file through the console | a working function |
| chmod a-x function/main then upload a ZIP file | a working function |
| chmod a-x function/main then run `pulumi up` | a working function |
| modify the Pulumi SDK to set file mode 0777 on all files added to the ZIP, and recompile the pulumi-aws provider, then any of the above | a working function |

There were many things I tripped over on the way:
 - I used `GOOS=linux go build` to compile the handler, but forgot I needed `GOARCH=amd64` as well (since I'm on an M1 Mac)
 - I got the handler path wrong when uploading the ZIP, because I used `zip main.zip function/main` and the path within the ZIP file for the exe is then `function/main`, whereas using the Pulumi program, it's just `main`
 - tried using `Code: pulumi.NewFileArchive("function/main")` expecting it to create a ZIP file as in the above step, but it fails

...but it seems to work as advertised, now. I rather suspect the go1.x runtime has been updated to implicitly make the handler binary executable, but I couldn't find documentary evidence with a quick search. In any case: I don't think this PR is necessary any longer.

Issue #4013, opened 2020-03-03
Creating an Azure C# project with the name "Pulumi" causes a Cycle Error


andreasnauta:
Steps to reproduce: 

1. Type `pulumi new`
2. Select `azure-csharp`
3. Type `Pulumi` as the project name.
4. Select default values.

Then you get  "error NU1108: Cycle detected."

![Pulumi](https://user-images.githubusercontent.com/8932820/75776292-02924d80-5d54-11ea-927c-8773e21fe99c.jpg)

mikhailshilkov:
Thanks for reporting this. We should probably prevent people from using project names `Pulumi` and `Pulumi.<Something>`.

PR #4022, opened 2020-03-05
Stop new projects starting with the word Pulumi

stack72:
Fixes: #4013

Without this, we are in a situation where we have a cyclic import

### DIFF
--- a/pkg/workspace/templates.go
+++ b/pkg/workspace/templates.go
@@ -522,6 +522,12 @@ func ValidateProjectName(s string) error {
 		return errors.New("A project name may only contain alphanumeric, hyphens, underscores, and periods")
 	}
 
+	// This is needed to stop cyclic imports in DotNet projects
+	if strings.ToLower(s) == "pulumi" || strings.HasPrefix(strings.ToLower(s), "pulumi.") {
+		return errors.New("A project name must not be `Pulumi` and must not start with the prefix `Pulumi.` " +
+			"to avoid collision with standard libraries")
+	}
+
 	return nil
 }
 
--- a/pkg/workspace/templates_test.go
+++ b/pkg/workspace/templates_test.go
@@ -241,3 +241,58 @@ func TestRetrieveFileTemplate(t *testing.T) {
 		})
 	}
 }
+
+func TestProjectNames(t *testing.T) {
+	tests := []struct {
+		testName    string
+		projectName string
+		expectError bool
+	}{
+		{
+			testName:    "Correct Project Name",
+			projectName: "SampleProject",
+			expectError: false,
+		},
+		{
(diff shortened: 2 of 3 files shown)


pkg/workspace/templates.go
```diff
@@ -518,6 +518,11 @@ func ValidateProjectName(s string) error {
 		return errors.New("A project name must be 100 characters or less")
 	}
 
+	// This is needed to stop cyclic imports in DotNet projects
+	if strings.HasPrefix(strings.ToLower(s), "pulumi") {
```
mikhailshilkov: I feel like this condition is too prohibitive. I expect people to do things like `pulumi-test` and I would love to avoid breaking them. I would prohibit literally `Pulumi` and starting with `Pulumi.`, case sensitive.

pkg/workspace/templates_test.go
```diff
@@ -241,3 +241,43 @@ func TestRetrieveFileTemplate(t *testing.T) {
...
+		{
+			testName:    "Project Name starting with the word Pulumi",
+			projectName: "PulumiProject",
+			expectError: true,
+		},
+		{
+			testName:    "Project Name greater than 100 characters",
+			projectName: "cZClTe6xrjgKzH5QS8rFEPqYK1z4bbMeMr6n89n87djq9emSAlznQXXkkCEpBBCaZAFNlCvbfqVcqoifYlfPl11hvekIDjXVIY7m1",
+			expectError: true,
+		},
```
mikhailshilkov: Consider adding a test for the name `Pulumi`

pkg/workspace/templates.go
stack72: Fixed

pkg/workspace/templates_test.go
stack72: Done!

pkg/workspace/templates.go
```diff
@@ -522,6 +522,11 @@ func ValidateProjectName(s string) error {
 		return errors.New("A project name may only contain alphanumeric, hyphens, underscores, and periods")
 	}
 
+	// This is needed to stop cyclic imports in DotNet projects
+	if strings.ToLower(s) == "pulumi" || strings.HasPrefix(strings.ToLower(s), "pulumi.") {
+		return errors.New("A project name must not start with the prefix `Pulumi` or `Pulumi.`")
```
mikhailshilkov: How about
```suggestion
		return errors.New("A project name must not be `Pulumi` and must not start with the prefix `Pulumi.` to avoid collision with standard libraries")
```
or maybe split into two checks and two error messages if that's too long.

stack72:
Thanks @mikhailshilkov - updated the error message and will merge on a green. build

Issue #4026, opened 2020-03-05
 panic: reflect: Elem of invalid type


EvanBoyle:
Panic in the marshaling code in the go SDK.

```go
package main

import (
	"github.com/pulumi/pulumi-aws/sdk/go/aws/eks"
	"github.com/pulumi/pulumi/sdk/go/pulumi"
)

func main() {
	pulumi.Run(func(ctx *pulumi.Context) error {
		// Create an AWS resource (S3 Bucket)
		_, err := eks.NewCluster(ctx, ctx.Stack(), &eks.ClusterArgs{
			Name:    pulumi.String("sfc-dev-k8s-3"),
			RoleArn: pulumi.String("arn:aws:iam::******:role/eks-cluster-role-5344e11"),
			Version: pulumi.String("1.14"),
			VpcConfig: &eks.ClusterVpcConfigArgs{
				ClusterSecurityGroupId: pulumi.String("sg-0da30ab129869c5b3"),
				PublicAccessCidrs:      pulumi.StringArray{pulumi.String("0.0.0.0/0")},
				SecurityGroupIds:       pulumi.StringArray{pulumi.String("sg-fdskljfklsjdf")},
				SubnetIds:              pulumi.StringArray{pulumi.String("subnet-jfkldsajklfjkdslajf"), pulumi.String("subnet-jkljsdkjafka")},
				VpcId:                  pulumi.String("vpc-lkjklfdsakjfkdsjaj"),
			},
		})
		return err
	})
}

```
    panic: reflect: Elem of invalid type
    goroutine 14 [running]:
    reflect.(*rtype).Elem(0x1b41760, 0xc00018d420, 0x199)
    	/usr/local/Cellar/go/1.13.8/libexec/src/reflect/type.go:920 +0x143
    github.com/pulumi/pulumi/sdk/go/pulumi.marshalInputAndDetermineSecret(0x1ae4080, 0xc00018d420, 0x1d4c3a0, 0x1b41760, 0xc00011b501, 0x10903f2, 0x1abdea0, 0x3, 0xc00011b600, 0x108ea2d, ...)
    	/Users/evanboyle/go/src/github.com/pulumi/pulumi/sdk/go/pulumi/rpc.go:270 +0xd5
[...]

PR #4027, opened 2020-03-05
fix reflect panic

EvanBoyle:
Closes https://github.com/pulumi/pulumi/issues/4026

We're trying to dereference a struct, assuming the dest is a pointer and getting a panic as a result. Some logs right before this location:

```
 rv.Type(): *eks.ClusterVpcConfigArgs
 rv.Type().kind(): ptr
 destType: eks.ClusterVpcConfig
 destType.kind(): struct

```

### DIFF
--- a/sdk/go/pulumi/rpc.go
+++ b/sdk/go/pulumi/rpc.go
@@ -267,7 +267,10 @@ func marshalInputAndDetermineSecret(v interface{},
 			if rv.IsNil() {
 				return resource.PropertyValue{}, deps, secret, nil
 			}
-			v, destType = rv.Elem().Interface(), destType.Elem()
+			if destType.Kind() == reflect.Ptr {
+				destType = destType.Elem()
+			}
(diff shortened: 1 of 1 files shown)


sdk/go/pulumi/rpc.go
```diff
@@ -267,7 +267,10 @@ func marshalInputAndDetermineSecret(v interface{},
 			if rv.IsNil() {
 				return resource.PropertyValue{}, deps, secret, nil
 			}
-			v, destType = rv.Elem().Interface(), destType.Elem()
+			if destType.Kind() == reflect.Ptr {
+				destType = destType.Elem()
+			}
+			v = rv.Elem().Interface()
```
lukehoban: Any test case we can/should add here?

EvanBoyle: Happy to circle back and add one if we have a good idea on how to construct the case. I don't, so I'm going to just merge for now.

PR #4035, opened 2020-03-07
Update schema-based docs generator

### LINKED ISSUE #3995: Complete implementation of new schema-based docs generator
Related to #2664.

The work done so far supports generating docs for the NodeJS SDK for any given provider. We need to fix-up the implementation to do the generation for all language SDKs.

Some of the remaining work entails the following tasks:
* [ ] Use Go's templating package instead of inline markdown
* [ ] Declarations in each language
	* Call into individual language generators to generate these declarations in each language
* [ ] Emit nested types inline
* [ ] Linking of types within declarations
	* Make the declaration gen smarter to include links
	* Linking within the page (e.g. anchor links to nested structures)
	* Linking within the site (e.g. linking types like `pulumi.Input` to their associated SDK doc link https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/pulumi/#Input)
* [ ] Show how to read an existing resource (i.e. static `get` function)
* [ ] Emit deprecation messages

praneetloke:
Resolves #3995.

Apologies for the generic PR title, as there are too many changes to list in a succinct way.

Here are some of the important changes introduced with PR:
* Use Go templates to generate the markdown files
  * Templates are embedded during build. The static bundle must be generated prior to that using `go generate` (documented in the README file added with this PR) and is only intended to be run by the `docs` repo. So for all other builds, the templates won't be bundled into the Go binary.
* Errors. I am not sure what the best way is to bubble-up errors for this generator, but I used panics to stop generating a doc if something unexpected (for example, error executing a template) which could result in a doc with non-conformant content. It seemed ok to do this for now. Would love to find out if there is a different pattern we follow for generators.
* ...all other changes are related to the tasks in the linked issue.

Here's a quick **preview** of what the resource doc page looks like with this generator for each language:

> **Note**: The layout and look-and-feel of the page is **not** final. It is just a starter. There is a separate effort to improve the client-side UX of the layout itself. I am providing these screenshots simply as a way to see what effect the changes in this PR have on the doc.

I am providing the images as links rather than inlining them, as they full-page screenshots. I also chose a resource that wasn't super-long just as an example. I can take screenshots for another resource if you would like to take a look at a different view of it.

#### TS
https://user-images.githubusercontent.com/1466314/76147463-3ec5f680-6051-11ea-8559-506fcfc16602.png
#### Python
https://user-images.githubusercontent.com/1466314/76147467-4ab1b880-6051-11ea-9a77-b3e5f3050cf2.png
#### Go
https://user-images.githubusercontent.com/1466314/76147470-4f766c80-6051-11ea-9fa2-7958b7571faf.png
#### C#
https://user-images.githubusercontent.com/1466314/76147473-543b2080-6051-11ea-926b-e5360c351c3e.png


praneetloke:
I added the "no-changelog-required" label since this PR doesn't add any features to the core CLI functionality itself.

pkg/codegen/docs/templates/examples.tmpl
```diff
@@ -0,0 +1,25 @@
+{{ define "examples" }}
+## Example Usage
+
+{{ range . }}
+### {{ .Heading }}
+
+{{ htmlSafe "{{< langchoose nojavascript >}}" }}
```
justinvp: For a subsequent change, are we going to want to show the JavaScript tab (and just use the TypeScript content for it)?

pkg/codegen/docs/templates/resource.tmpl
```diff
@@ -0,0 +1,108 @@
...
+> See the <a href="{{ .APIDocLinks.go }}">input</a> API doc.
+{{ htmlSafe "{{% /lang %}}" }}
+
+{{ htmlSafe "{{% lang csharp %}}" }}
+> See the <a href="{{ .APIDocLinks.csharp }}">input</a> API doc.
+{{ htmlSafe "{{% /lang %}}" }}
+
+{{ template "properties" .Properties }}
+{{ end }}
+{{ end }}
```
justinvp: Nit: add blank line

pkg/codegen/dotnet/doc.go
```diff
@@ -0,0 +1,54 @@
...
+var _ codegen.DocLanguageHelper = DocLanguageHelper{}
+
+// GetDocLinkForResourceType returns the .NET API doc URL for a type belonging to a resource provider.
+func (d DocLanguageHelper) GetDocLinkForResourceType(packageName, _, typeName string) string {
+	typeName = strings.ReplaceAll(typeName, "?", "")
+	var packageNamespace string
+	if packageName != "" {
+		packageNamespace = "." + title(packageName)
+	}
+	return fmt.Sprintf("https://www.pulumi.com/docs/reference/pkg/dotnet/Pulumi%s/%s.html", packageNamespace, typeName)
```
justinvp:
```suggestion
	return fmt.Sprintf("/docs/reference/pkg/dotnet/Pulumi%s/%s.html", packageNamespace, typeName)
```

pkg/codegen/nodejs/doc.go
```diff
@@ -0,0 +1,73 @@
...
+// DocLanguageHelper is the NodeJS-specific implementation of the DocLanguageHelper.
+type DocLanguageHelper struct{}
+
+var _ codegen.DocLanguageHelper = DocLanguageHelper{}
+
+// GetDocLinkForResourceType returns the NodeJS API doc for a type belonging to a resource provider.
+func (d DocLanguageHelper) GetDocLinkForResourceType(packageName, modName, typeName string) string {
+	path := fmt.Sprintf("%s/%s", packageName, modName)
+	typeName = strings.ReplaceAll(typeName, "?", "")
+	return fmt.Sprintf("https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/%s/#%s", path, typeName)
```
justinvp:
```suggestion
	return fmt.Sprintf("/docs/reference/pkg/nodejs/pulumi/%s/#%s", path, typeName)
```

pkg/codegen/nodejs/doc.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+// GetDocLinkForInputType returns the doc link for an input type.
+func (d DocLanguageHelper) GetDocLinkForInputType(packageName, modName, typeName string) string {
+	typeName = strings.ReplaceAll(typeName, "?", "")
+	return fmt.Sprintf("https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/%s/types/input/#%s", packageName, typeName)
```
justinvp:
```suggestion
	return fmt.Sprintf("/docs/reference/pkg/nodejs/pulumi/%s/types/input/#%s", packageName, typeName)
```

pkg/codegen/nodejs/doc.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+// GetDocLinkForBuiltInType returns the godoc URL for a built-in type.
```
justinvp:
```suggestion
// GetDocLinkForBuiltInType returns the URL for a built-in type.
```

pkg/codegen/docs/README.md
```diff
@@ -0,0 +1,40 @@
+# Docs generator
+
+This generator generates resource-level docs by utilizing the Pulumi schema generated by [tfbridge](https://github.com/pulumi/pulumi-terraform-bridge/).
+
+## Crash course on templates
+
+The templates use Go's built-in `html/template` package to process templates with data. The driver for this doc generator (in tfbridge) then persists each file from memory onto the disk as `.md` files in the output directory specified by `tfbridge`.
```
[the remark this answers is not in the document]
praneetloke: > Then, in the docs repo, create a new go docgen tool that has pulumi/pulumi as a dependency, that can take the Pulumi schema as input (standard input?) and an argument specifying the output directory, and "drive" the doc gen from the input schema?
>
> The scripts in the docs gen for kicking this all off for each repo, would be coded to call tfbridge to get the schema for TF-based providers, and whatever command line tool is required to get the schema for Kubernetes.

I like this idea, and in fact, in a branch for the `docs` repo I have something like this. See this [Makefile](https://github.com/pulumi/docs/blob/praneetloke/docs/prototype/Makefile#L40) target and this [script](https://github.com/pulumi/docs/blob/praneetloke/docs/prototype/scripts/gen_resource_docs.sh), that it calls. I do think that the `docs` repo should be the only "driver" of the docs generator in this repo. I think it makes it easier to understand the direct dependencies too, rather than transitively calling this tool.

> Or instead of a new go-based docgen tool in the docs repo, we could consider just adding another hidden command to the Pulumi CLI itself, similar to the hidden gen-markdown command we have to generate docs for the CLI commands. I could see a hidden gen-docs command, that takes the Pulumi schema from standard input and outputs to a specified directory.

Yeah this is fine too I guess. Not sure if there is a strong reason to do this over the above approach, though? And if not, I'd prefer to do the above. What do you think?

pkg/codegen/docs/utils.go
```diff
@@ -0,0 +1,82 @@
...
+	return prev == rune('.') && unicode.IsUpper(next)
+}
+
+func isPythonTypeNameBoundary(prev rune, next rune) bool {
+	// For Python, names are snake_cased (Duh?).
+	return (prev == rune('_') && unicode.IsLower(next))
+}
+
+// wbr inserts HTML <wbr> in between case changes, e.g. "fooBar" becomes "foo<wbr>Bar".
+func wbr(s string) string {
```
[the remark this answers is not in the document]
praneetloke: Ah yes. Will do!

pkg/codegen/docs/templates/examples.tmpl
praneetloke: I am actually not sure of it at the moment. It is possible that, for the examples section we may (or must?) show the JS tab. Once the examples are available in the Pulumi schema, I will know for certain.

pkg/codegen/docs/templates/resource.tmpl
praneetloke: Done

pkg/codegen/docs/README.md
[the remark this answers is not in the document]
justinvp: The above approach sounds good to me.

PR #4042, opened 2020-03-09
Fix terminal keypad mode

### LINKED ISSUE #3480: Terminal gets into a state where UP/DOWN don't work with prompts
Strangely, if I launch PowerShell Core (`pwsh`) on iTerm2 3.3.4 (macOS 10.14), the UP and DOWN keys subsequently don't interact nicely with our interactive CLI prompts, like `pulumi new` template selection. The act of running `pwsh` is changing my shell in some persistent way.

This is the result of hitting DOWN for example:

```
$ pulumi new
Please choose a template:
> aws-javascript             A minimal AWS JavaScript Pulumi program
  aws-python                 A minimal AWS Python Pulumi program
  aws-typescript             A minimal AWS TypeScript Pulumi program
  azure-javascript           A minimal Azure JavaScript Pulumi program
  azure-python               A minimal Azure Python Pulumi program
  azure-typescript           A minimal Azure TypeScript Pulumi program
  gcp-javascript             A minimal Google Cloud JavaScript Pulumi program
  gcp-python                 A minimal Google Cloud Python Pulumi program
  gcp-typescript             A minimal Google Cloud TypeScript Pulumi program
  kubernetes-javascript      A minimal Kubernetes JavaScript Pulumi program
  kubernetes-python          A minimal Kubernetes Python Pulumi program
  kubernetes-typescript      A minimal Kubernetes TypeScript Pulumi program
  Show additional templates
error: no template selected; please use `pulumi new` to choose one
```

I can deterministically reproduce this by creating a fresh shell -- works before running `pwsh`, does not work afterwards. I've run into this previously but never was able to get a simple repro (in fact, I think there may already be an issue tracking, but couldn't find it).

@EvanBoyle Am I remembering correctly that you encountered a similar issue just recently?

mikhailshilkov:
Fix #3480 by printing `\033[?1l` before every survey call, which does the same thing as `tput rmkx` mentioned in the issue.

I briefly looked into fixing https://github.com/AlecAivazis/survey/issues/228 upstream. Unfortunately, we are using some older version, and there's 2.0+ already, so I guess we would need to update to the latest version first. It has breaking changes, so it sounds like a bigger task that I wanted to take right now.

I'm not totally sure which side effect we may trigger by switching the keypad mode, so I tried to call it as late as possible: right before the survey calls, to avoid unneeded switches.


pkg/util/cmdutil/console.go
```diff
@@ -95,6 +95,16 @@ func RemoveTrailingNewline(s string) string {
 	return s
 }
 
+// EndKeypadTransmitMode switches the terminal out of the keypad transmit 'application' mode back to 'normal' mode.
+func EndKeypadTransmitMode() {
+	if runtime.GOOS != "windows" && Interactive() {
```
lukehoban: On the “Windows” part - is that the right test, or is actually something about whether a certain kind of terminal is attached?  Are there Windows environments where this same issue will still occur?

mikhailshilkov: So I double-checked and `GOOS` returns `linux` on WSL. Cmd/PowerShell terminals have no notion of normal/application mode and tput commands, AFAIK.

PR #4043, opened 2020-03-09
Ensure python overlays work as expected

stack72:
Fixes: https://github.com/pulumi/pulumi-terraform-bridge/issues/119

This allows us to specify an overlays block e.g.

```
Overlay: &tfbridge.OverlayInfo{
	DestFiles: []string{
		"pulumi_docker/docker.py",
		"pulumi_docker/image.py",
	},
},
```

The overlays files are treated differently to normal module files
as they are not generated. This structure means that we will emit
the correct entries in the __init__.py file

Without this structure (ie. pulumi_pkgname), the generator actually
copies the file (i.e. docker.py) to the root of the Python SDK. This
is because the structure of the Python SDK has a sub-folder than that
of the NodeJS SDK

I tested this using PR https://github.com/pulumi/pulumi-docker/pull/141
and this now works as expected and we can take advantage of the new
Python overlays for Docker

### DIFF
--- a/pkg/codegen/python/gen.go
+++ b/pkg/codegen/python/gen.go
@@ -136,7 +136,7 @@ func (mod *modContext) gen(fs fs) error {
 			d = ""
 		}
 		if d == dir {
-			exports = append(exports, p[:len(p)-len(".py")])
+			exports = append(exports, strings.TrimSuffix(path.Base(p), ".py"))
 		}
 	}
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ CHANGELOG
   [#4013](https://github.com/pulumi/pulumi/pull/4013)
 - Ensure we can locate Go created application binaries on Windows
   [#4030](https://github.com/pulumi/pulumi/pull/4030)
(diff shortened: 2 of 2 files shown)


pkg/codegen/python/gen.go
```diff
@@ -136,7 +136,15 @@ func (mod *modContext) gen(fs fs) error {
...
-			exports = append(exports, p[:len(p)-len(".py")])
+			// this ensures that we can look at nested files
+			// without this, we don't understand where the destFiles are
+			rawNameParts := strings.Split(p, "/")
+			if len(rawNameParts) == 1 {
+				exports = append(exports, p[:len(p)-len(".py")])
+			} else {
+				fileName := rawNameParts[len(rawNameParts)-1]
+				exports = append(exports, fileName[:len(fileName)-len(".py")])
+			}
```
lukehoban: I suspect this won't work on Windows (`"/"`).  But if I'm reading it right, all of this code can be replaced with:

```go
exports = append(exports, strings.TrimSuffix(path.Base(p), ".py"))
```

stack72: nice suggestion - pushed! will merge on green

PR #4055, opened 2020-03-11
Updates to resource doc generator for generating docs for Functions

### LINKED ISSUE #4024: Complete the docs generator for Functions
Related to https://github.com/pulumi/docs/issues/2264.

These were previously known as "Data Sources". In the previous milestone, a new codegen for "docs" was implemented under `pkg/codegen/docs`, which uses the Pulumi schema to generate the markdown files used by our docs site. The "Functions" part was left as a TODO.

praneetloke:
Closes #4024.

This PR does the following:
* Implement the `genFunction` method to process the relevant doc template for generating a doc for Functions using the new resource-based doc generator
* Generate links for a nested type for its input and output type variants.

Example screenshots:

TS:
https://user-images.githubusercontent.com/1466314/76449047-9c777d00-6388-11ea-8d16-e3a58f4e8a58.png

Python:
https://user-images.githubusercontent.com/1466314/76449052-9f726d80-6388-11ea-860a-680568e8f11e.png

Go:
https://user-images.githubusercontent.com/1466314/76449062-a26d5e00-6388-11ea-9ff9-321bbd700c0c.png

C#:
https://user-images.githubusercontent.com/1466314/76449076-a6997b80-6388-11ea-996b-71613e971547.png

TS:
https://user-images.githubusercontent.com/1466314/76449106-b6b15b00-6388-11ea-9282-e8f466d71d58.png

Python:
https://user-images.githubusercontent.com/1466314/76449115-ba44e200-6388-11ea-82ca-273db73cb1ec.png

Go:
https://user-images.githubusercontent.com/1466314/76449121-be70ff80-6388-11ea-8e9d-228c240b52bd.png

C#:
https://user-images.githubusercontent.com/1466314/76449125-c0d35980-6388-11ea-9288-b75871a56bc8.png

### DIFF
--- a/pkg/codegen/docs/bundler.go
+++ b/pkg/codegen/docs/bundler.go
@@ -96,14 +96,14 @@ func main() {
 	contents := make(map[string][]byte)
 	for _, f := range files {
 		if f.IsDir() {
-			fmt.Println(fmt.Sprintf("%q is a dir. Skipping...", f.Name()))
+			fmt.Printf("%q is a dir. Skipping...\n", f.Name())
 		}
 		b, err := ioutil.ReadFile(docsTemplatesPath + "/" + f.Name())
 		if err != nil {
 			log.Fatalf("Error reading file %s: %v", f.Name(), err)
 		}
 		if len(b) == 0 {
-			fmt.Println(fmt.Sprintf("%q is empty. Skipping...", f.Name()))
+			fmt.Printf("%q is empty. Skipping...\n", f.Name())
 			continue
 		}
 		contents[f.Name()] = b
--- a/pkg/codegen/docs/bundler.go
+++ b/pkg/codegen/docs/bundler.go
@@ -96,11 +96,15 @@ func main() {
 	contents := make(map[string][]byte)
 	for _, f := range files {
 		if f.IsDir() {
-			fmt.Printf("%q is a dir. Skipping...", f.Name())
+			fmt.Println(fmt.Sprintf("%q is a dir. Skipping...", f.Name()))
 		}
 		b, err := ioutil.ReadFile(docsTemplatesPath + "/" + f.Name())
 		if err != nil {
 			log.Fatalf("Error reading file %s: %v", f.Name(), err)
+		}
+		if len(b) == 0 {
+			fmt.Println(fmt.Sprintf("%q is empty. Skipping...", f.Name()))
+			continue
 		}
 		contents[f.Name()] = b
 	}
--- a/pkg/codegen/docs/gen.go
+++ b/pkg/codegen/docs/gen.go
@@ -97,7 +97,7 @@ type ConstructorParam struct {
 	DefaultValue string
 }
 
-type resourceArgs struct {
+type resourceDocArgs struct {
 	Header
 
 	Comment  string
@@ -485,7 +485,7 @@ func (mod *modContext) getConstructorResourceInfo(resourceTypeName string) map[s
 
 // genResource is the entrypoint for generating a doc for a resource
 // from its Pulumi schema.
-func (mod *modContext) genResource(r *schema.Resource) resourceArgs {
+func (mod *modContext) genResource(r *schema.Resource) resourceDocArgs {
 	// Create a resource module file into which all of this resource's types will go.
 	name := resourceName(r)
 
@@ -515,7 +515,7 @@ func (mod *modContext) genResource(r *schema.Resource) resourceArgs {
 		}
 	}
 
-	data := resourceArgs{
+	data := resourceDocArgs{
 		Header: Header{
 			Title: name,
 		},
@@ -536,13 +536,6 @@ func (mod *modContext) genResource(r *schema.Resource) resourceArgs {
 	}
 
 	return data
-}
-
-func (mod *modContext) genFunction(w io.Writer, fun *schema.Function) {
-	fmt.Fprintf(w, "%s\n\n", fun.Comment)
-
-	// TODO: Emit the page for functions, similar to the page for resources.
-	fmt.Fprintf(w, "TODO\n\n")
 }
 
 func visitObjectTypes(t schema.Type, visitor func(*schema.ObjectType)) {
@@ -670,11 +663,13 @@ func (mod *modContext) gen(fs fs) error {
 
 	// Functions
 	for _, f := range mod.functions {
+		data := mod.genFunction(f)
+
 		buffer := &bytes.Buffer{}
-		mod.genHeader(buffer, tokenToName(f.Token))
-
-		mod.genFunction(buffer, f)
-
+		err := templates.ExecuteTemplate(buffer, "function.tmpl", data)
+		if err != nil {
+			panic(err)
+		}
 		addFile(lower(tokenToName(f.Token))+".md", buffer.String())
 	}
 
--- a/pkg/codegen/docs/gen.go
+++ b/pkg/codegen/docs/gen.go
@@ -71,9 +71,10 @@ type Property struct {
 
 // DocNestedType represents a complex type.
 type DocNestedType struct {
-	Name        string
-	APIDocLinks map[string]string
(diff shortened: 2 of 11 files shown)


praneetloke:
Once again applied the "no-changelog-required" label, because there are no user facing features/impacts with this PR.

pkg/codegen/docs/templates/function.tmpl
```diff
@@ -0,0 +1,59 @@
...
+
+<!-- Python -->
+<div class="highlight"><pre class="chroma"><code class="language-python" data-lang="python"><span class="k">function </span> get_{{ pyName .ResourceName }}(</span>{{ htmlSafe .FunctionArgs.python }}<span class="p">)</span></code></pre></div>
+
+<!-- Go -->
+<div class="highlight"><pre class="chroma"><code class="language-go" data-lang="go"><span class="k">func </span>Lookup{{ .ResourceName }}<span class="p">(</span>{{ htmlSafe .FunctionArgs.go }}<span class="p">) (*{{ template "linkify_param" .FunctionResult.go }}, error)</span></code></pre></div>
+
+<!-- C# -->
+<div class="highlight"><pre class="chroma"><code class="language-csharp" data-lang="csharp"><span class="k">public static </span>Task<{{ template "linkify_param" .FunctionResult.csharp }}> <span class="p">Get{{ .ResourceName }}(</span>{{ htmlSafe .FunctionArgs.csharp }}<span class="p">)</span></code></pre></div>
```
justinvp: This may be changing for 2.0, just so you're aware: https://github.com/pulumi/pulumi-terraform-bridge/issues/125

In which case, we'd probably want to render the static class along with the `InvokeAsync` signature.

```cs
public static class GetVpc
{
    public static Task<GetVpcResult> InvokeAsync(GetVpcArgs? args = null, InvokeOptions? options = null);
}
```

pkg/codegen/docs/templates/function.tmpl
```diff
@@ -0,0 +1,59 @@
...
+
+{{ if ne (len .InputProperties) 0 }}
+
+The following arguments are supported:
+{{ template "properties" .InputProperties }}
+
+{{ end }}
+
+<!-- Output properties -->
+## {{.Header.Title}} Output Properties
```
justinvp: Instead of "Output Properties" terminology, should we just use the name of the result type, and display it similar to the other Supporting Types? Or otherwise say something like `{{ .Header.Title }} Result`? The reason is, this is a bit different than output properties of resources. In that case, it's properties on the resource type itself. Whereas here, we're talking about properties on a result struct/class/interface.

pkg/codegen/docs/gen_function.go
```diff
@@ -0,0 +1,309 @@
...
+	// in the Function.
+	FunctionArgs map[string]string
+	// FunctionResult is a map per language property types
+	// that is returned as a result of calling a Function.
+	FunctionResult map[string]propertyType
+
+	// InputProperties is a map per language and the corresponding slice
+	// of input properties accepted by the Function.
+	InputProperties map[string][]property
+	// InputProperties is a map per language and the corresponding slice
```
justinvp: Nit:

```suggestion
	// OutputProperties is a map per language and the corresponding slice
```

pkg/codegen/docs/bundler.go
```diff
@@ -96,12 +96,16 @@ func main() {
 	contents := make(map[string][]byte)
 	for _, f := range files {
 		if f.IsDir() {
-			fmt.Printf("%q is a dir. Skipping...", f.Name())
+			fmt.Println(fmt.Sprintf("%q is a dir. Skipping...", f.Name()))
```
justinvp: Nit: Why not the following?

```suggestion
			fmt.Printf("%q is a dir. Skipping...\n", f.Name())
```

We do that pretty pervasively in the CLI.

pkg/codegen/docs/gen.go
```diff
@@ -41,53 +41,77 @@ import (
...
 }
 
-// PropertyType represents the type of a property.
-type PropertyType struct {
+// docNestedType represents a complex type.
+type docNestedType struct {
+	Name         string
+	IsOutputType bool
+	APIDocLinks  map[string]apiTypeDocLinks
+	Properties   map[string][]property
```
justinvp: One general thing I totally forgot to mention. I believe this is still the case, but @pgavlin, please correct me if I am wrong or if we're planning on changing this for 2.0... For nested types only, and for Python only, the lang-specific names of the properties should be `camelCase` rather than `snake_case` like everywhere else in Python.

pkg/codegen/docs/gen.go
```diff
@@ -97,30 +121,65 @@ type ConstructorParam struct {
...
-	StateParam       string
+	ConstructorResource map[string]propertyType
+	// ArgsRequired is a flag indicating if the args param is required
+	// when creating a new resource.
+	ArgsRequired bool
+
+	// InputProperties is a map per language and a corresponding slice of
+	// input properties accepted as args while creating a new resource.
+	InputProperties map[string][]property
+	// InputProperties is a map per language and a corresponding slice of
```
justinvp:
```suggestion
	// OutputProperties is a map per language and a corresponding slice of
```

pkg/codegen/docs/gen.go
```diff
@@ -97,30 +121,65 @@ type ConstructorParam struct {
...
+type appearsIn struct {
+	Input  bool
+	Output bool
 }
 
-type stringSet map[string]struct{}
+// stringSet is a type-alias for a map of type tokens
+// and whether or not the type appears in input as well
+// as output properties.
+type stringSet map[string]appearsIn
```
justinvp: Should the type be renamed since it's now more than just a `stringSet`?

pkg/codegen/docs/bundler.go
praneetloke: Only because I preferred this way over writing strings with new-line chars. I will change this to use `\n` instead of the enveloped call. I don't have strong ties with the current format.

pkg/codegen/docs/templates/function.tmpl
praneetloke: Ah thanks for the heads-up. If what is being proposed in the issue is implemented, I could see this being changed to `GetVpc.InvokeAsync(...)` rather than render the static class along with the InvokeAsync signature here. Is there a reason to want to render the class and the static method in its entirety. One reason I could of is that it would be a little more close to what the API doc would show.

pkg/codegen/docs/templates/function.tmpl
praneetloke: > Or otherwise say something like `{{ .Header.Title }} Result`?

Great point! And I like this suggestion.

pkg/codegen/docs/gen_function.go
praneetloke: Oops...will fix it

pkg/codegen/docs/gen.go
praneetloke: Ohh. Do let me know. Though, I think I'll write-up an issue for it to be fixed in a follow-up PR rather than in this one.

pkg/codegen/docs/gen.go
praneetloke: ha yes!

praneetloke:
> Now that Pulumify should be able to handle all the new files, what do you think of opening a draft PR in the docs repo that includes the generated docs for AWS with these changes? Then I can provie any additional feedback from viewing the preview.

I am actually working on a change to the `docs` repo that would make it the driver for generating the resource-based docs. If you would like me to hold merging this until you can see a preview in Pulumify, let me know and I'll open a draft PR with generated docs for that purpose.

justinvp:
> If you would like me to hold merging this until you can see a preview in Pulumify, let me know and I'll open a draft PR with generated docs for that purpose.

No need to hold this PR.

PR #4056, opened 2020-03-11
Stabilize Python codegen.

pgavlin:
Make sure all property casing conversions are computed prior to codegen.

### DIFF
--- a/pkg/codegen/python/gen.go
+++ b/pkg/codegen/python/gen.go
@@ -173,21 +173,6 @@ func (mod *modContext) gen(fs fs) error {
 		}
 	}
 
-	// Calculate casing tables. We do this up front because our docstring generator (which is run during
-	// genResource) requires them.
-	for _, r := range mod.resources {
-		for _, prop := range r.Properties {
-			if err := mod.recordProperty(prop); err != nil {
-				return err
-			}
-		}
-		for _, prop := range r.InputProperties {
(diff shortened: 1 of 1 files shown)


pgavlin:
Fixes https://github.com/pulumi/pulumi-terraform-bridge/issues/129.

pgavlin:
Note that this is not quite a complete fix for the linked issue: it does not fix the instability in the version that is embedded in the schema. It does fix the Python codegen, though.

pkg/codegen/python/gen.go
```diff
@@ -1199,14 +1185,32 @@ func GeneratePackage(tool string, pkg *schema.Package, extraFiles map[string][]b
 		_ = getMod(":config/config:")
 	}
 
-	scanResource := func(r *schema.Resource) {
+	scanResource := func(r *schema.Resource) error {
 		mod := getMod(r.Token)
 		mod.resources = append(mod.resources, r)
+
+		// Calculate casing tables. We do this up front because our docstring generator (which is run during
+		// genResource) requires them.
```
lukehoban: This comment doesn't appear to be correct anymore - and does the comment point to any potential issue with this change?

pgavlin: > This comment doesn't appear to be correct anymore

The comment is still correct. Which part seems wrong?

>  - and does the comment point to any potential issue with this change?

It pointed to the issue that these changes are fixing.

PR #4057, opened 2020-03-11
Support for running Python policy packs

justinvp:
These changes enable running policy packs written in Python. I'll be opening a subsequent PR in the `pulumi-policy` repo that adds the Python policy SDK.

Part of https://github.com/pulumi/pulumi-policy/issues/208

### DIFF
--- a/pkg/resource/plugin/analyzer_plugin.go
+++ b/pkg/resource/plugin/analyzer_plugin.go
@@ -18,6 +18,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"path/filepath"
 	"reflect"
 	"sort"
 	"strings"
@@ -78,11 +79,23 @@ func NewAnalyzer(host Host, ctx *Context, name tokens.QName) (Analyzer, error) {
 	}, nil
 }
 
-const policyAnalyzerName = "policy"
-
 // NewPolicyAnalyzer boots the nodejs analyzer plugin located at `policyPackpath`
 func NewPolicyAnalyzer(
 	host Host, ctx *Context, name tokens.QName, policyPackPath string, opts *PolicyAnalyzerOptions) (Analyzer, error) {
+
+	proj, err := workspace.LoadPolicyPack(filepath.Join(policyPackPath, "PulumiPolicy.yaml"))
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to load Pulumi policy project located at %q", policyPackPath)
+	}
+
+	var policyAnalyzerName string
+	if strings.EqualFold(proj.Runtime.Name(), "nodejs") {
+		policyAnalyzerName = "policy"
+	} else if strings.EqualFold(proj.Runtime.Name(), "python") {
+		policyAnalyzerName = "policy-python"
+	} else {
+		return nil, fmt.Errorf("unsupported policy runtime %s", proj.Runtime.Name())
+	}
 
 	// Load the policy-booting analyzer plugin (i.e., `pulumi-analyzer-${policyAnalyzerName}`).
 	_, pluginPath, err := workspace.GetPluginPath(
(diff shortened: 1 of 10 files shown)


pkg/resource/plugin/analyzer_plugin.go
```diff
@@ -78,12 +79,25 @@ func NewAnalyzer(host Host, ctx *Context, name tokens.QName) (Analyzer, error) {
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to load Pulumi policy project located at %q", policyPackPath)
+	}
+
+	var policyAnalyzerName string
+	switch proj.Runtime.Name() {
+	case "nodejs":
+		policyAnalyzerName = "policy"
+	case "python":
+		policyAnalyzerName = "policy-python"
```
justinvp: Having to introduce `pulumi-analyzer-policy-python` is kind of lame, but it's not really user-facing, aside from the fact that we're now distributing an additional `pulumi-analyzer-policy-python` file (`pulumi-analyzer-policy-python.cmd` on Windows).

The alternative would to modify the existing `pulumi-analyzer-policy` script to know whether it should be launching node vs. python (vs. other languages).

ekrengel: hmm what if the Policy yaml file had the runtime?

ekrengel:
Took a look and nothing jumps out as me as incorrect.. but I'll differ to @pgavlin here since I am unfamiliar with this part of the codebase.

pkg/resource/plugin/analyzer_plugin.go
lukehoban: The approach taken here is reasonably similar to our language hosts, so I think it's reasonable to use the same approach here, even though it leads to a bit of duplication.  Curious if @pgavlin has strong feelings either way on this.  

For language hosts, I believe we just append the language name to the root and then look for that on the path.  Would doing the same here make this more fundamentally multi-language? (allow a 3rd part to create `pulumi-analyzer-policy-go` themselves?)

pkg/resource/plugin/analyzer_plugin.go
```diff
@@ -606,16 +619,23 @@ func constructEnv(opts *PolicyAnalyzerOptions) ([]string, error) {
 		}
 	}
 
+	langSpecificVar := func(k string) string {
+		if runtime == "nodejs" {
+			return strings.Replace(k, "PULUMI_", "PULUMI_NODEJS_", 1)
+		}
+		return k
+	}
```
lukehoban: Possibly not now - but I wonder if we want to transition to using consistent names, and pass both to Node.js for now, and start reading from both in the Node.js implementation so that we can get to a consistent pattern here soon and not need Node specific code at this layer?

sdk/python/lib/pulumi/policy/__main__.py
```diff
@@ -0,0 +1,55 @@
+# Copyright 2016-2020, Pulumi Corporation.  All rights reserved.
```
lukehoban: Don't we use full apache headers on all these files?

pkg/resource/plugin/analyzer_plugin.go
lukehoban: In fact - would doing just this unblock things like https://github.com/pulumi/pulumi-policy-opa?  Couldn't that project just name itself `pulumi-analyzer-policy-opa` and then mark `runtime: opa` in it's `Pulumipolicy.yaml`?  Users would have to distribute that binary to machines, but actual policies that used it could in principle just work?

justinvp: > Would doing the same here make this more fundamentally multi-language?

> In fact - would doing just this unblock things like https://github.com/pulumi/pulumi-policy-opa? Couldn't that project just name itself `pulumi-analyzer-policy-opa` and then mark `runtime: opa` in it's `Pulumipolicy.yaml`? Users would have to distribute that binary to machines, but actual policies that used it could in principle just work?

Yes, great point! Doing it like this should unblock other languages and OPA.

pgavlin: What @lukehoban said. Seems like we should just take the same strategy the language hosts use.

PR #3830, opened 2020-01-28
WIP: GetRequiredPlugins for Go

pgavlin:
These changes implement `GetRequiredPlugins` for Go using a registry
mechanism and an alternate entry point for `pulumi.Run`. Packages that
require plugins are expected to register themselves with the Pulumi SDK.
When `pulumi.Run` is used and the `PULUMI_PLUGINS` envvar is truthy, the
program will dump a JSON-encoded description of its required plugins to
stdout. The language host then uses this description to respond to
`GetRequiredPlugins`.

Fixes #1549.

PR #4060, opened 2020-03-11
Go plugin acquisition

EvanBoyle:
Revamping the reverted PR from here: https://github.com/pulumi/pulumi/pull/3830

It turns out that if we don't append existing environment variables in the `go run` case, the command fails with:

```
failed to discover plugin requirements: failed to execute program cmd: exit status 1
```

The fact that the previous CI run from `f652d93` passed is slightly concerning, as doing a pulumi up locally fails if I don't have the executable pre-build `go build -o ...`. 

We should definitely make sure that we're testing the go run path in CI since we're finding discrepancies like this cc @stack72


pkg/codegen/go/gen.go
```diff
@@ -600,7 +600,7 @@ func (pkg *pkgContext) genResource(w io.Writer, r *schema.Resource) error {
 		fmt.Fprintf(w, "// Get%[1]s gets an existing %[1]s resource's state with the given name, ID, and optional\n", name)
 		fmt.Fprintf(w, "// state properties that are used to uniquely qualify the lookup (nil if not required).\n")
 		fmt.Fprintf(w, "func Get%s(ctx *pulumi.Context,\n", name)
-		fmt.Fprintf(w, "\tname string, id pulumi.IDInput, state *%[1]sState, opts ...pulumi.ResourceOption) (*%[1]s, error) {\n", name)
```
pgavlin: Let's revert this change.

PR #4063, opened 2020-03-12
Support "link" style dependencies in Go tests

lukehoban:
The intergration test framework currently supports using `dep` for dependency management.

However, `dep` has no native ability to manage "yarn link"-style dependencies on locally available packages.

This is a necessary scenario for testing in most repos though, as (e.g.) examples in the Kubernetes repo need to test against the locally available version of `pulumi-kubernetes`.

The best we can do is a trick of (a) deleting the vendored copy of the locally available dependency (b) copying the locally available dependency into the vendor folder (c) deleting the nested vendor folder in the new copy of the locally available dependency.


lukehoban:
See https://github.com/pulumi/pulumi-kubernetes/pull/1029/files#diff-4ddb3da55b97a0ef5f25e1113b305a51 for an example of using this.

PR #4068, opened 2020-03-12
Shorten Python package descriptions.

pgavlin:
These descriptions should only be one line, and should not contain
markdown.

### DIFF
--- a/pkg/codegen/python/gen.go
+++ b/pkg/codegen/python/gen.go
@@ -685,7 +685,7 @@ func genPackageMetadata(tool string, pkg *schema.Package, requires map[string]st
 	fmt.Fprintf(w, "setup(name='%s',\n", pyPack(pkg.Name))
 	fmt.Fprintf(w, "      version='${VERSION}',\n")
 	if pkg.Description != "" {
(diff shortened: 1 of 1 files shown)


pgavlin:
> I guess this doesn't ensure "no markdown" but the first lines look fine for the three major providers.

That's right.

pkg/codegen/python/gen.go
```diff
@@ -677,7 +685,7 @@ func genPackageMetadata(tool string, pkg *schema.Package, requires map[string]st
 	fmt.Fprintf(w, "setup(name='%s',\n", pyPack(pkg.Name))
 	fmt.Fprintf(w, "      version='${VERSION}',\n")
 	if pkg.Description != "" {
-		fmt.Fprintf(w, "      description=\"\"\"%s\"\"\",\n", pkg.Description)
+		fmt.Fprintf(w, "      description=\"%s\",\n", sanitizePackageDescription(pkg.Description))
```
lukehoban: Are we sure these sanitized descriptions are going to be safe to embed inside quotes?  

Do we want instead this?

```suggestion
		fmt.Fprintf(w, "      description=%q,\n", sanitizePackageDescription(pkg.Description))
```

pgavlin: Yep, that's better

PR #4069, opened 2020-03-12
Expose engine logging to Go SDK

### LINKED ISSUE #3801: Expose the engine's logging interface from the Go SDK.
Just what it says on the tin.

tasiah:
Fixes #3801 

To allow users to do something like `ctx.Log.Info("Starting docker build and push...", logResource, 0, false)`, I needed to add another field to `pulumi.Context` as I describe [here](https://github.com/pulumi/pulumi/issues/3801#issuecomment-598328191). The implementation isn't the prettiest, but the end result is pretty similar to the other languages'. An alternative could be to  instead add new functions directly onto `pulumi.Context` like `ctx.Info(...)` and `ctx.Debug(...)`.

After this has been merged we should (1) update the [docs](https://www.pulumi.com/docs/intro/concepts/programming-model/#logging) and (2) update the logging for pulumi/pulumi-docker.

### DIFF
--- a/sdk/go/pulumi/context.go
+++ b/sdk/go/pulumi/context.go
@@ -50,6 +50,7 @@ type Context struct {
 	rpcsDone    *sync.Cond  // an event signaling completion of RPCs.
 	rpcsLock    *sync.Mutex // a lock protecting the RPC count and event.
 	rpcError    error       // the first error (if any) encountered during an RPC.
+	Log         Log
 }
 
 // NewContext creates a fresh run context out of the given metadata.
@@ -83,6 +84,10 @@ func NewContext(ctx context.Context, info RunInfo) (*Context, error) {
 	}
 
 	mutex := &sync.Mutex{}
+	log := &logState{
+		engine: engine,
+		ctx:    ctx,
+	}
 	return &Context{
 		ctx:         ctx,
 		info:        info,
@@ -94,6 +99,7 @@ func NewContext(ctx context.Context, info RunInfo) (*Context, error) {
 		rpcs:        0,
 		rpcsLock:    mutex,
 		rpcsDone:    sync.NewCond(mutex),
+		Log:         log,
 	}, nil
 }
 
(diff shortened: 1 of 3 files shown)


sdk/go/pulumi/log.go
```diff
@@ -0,0 +1,76 @@
...
+func (log *logState) Warn(msg string, resource Resource, streamID int32, ephemeral bool) error {
+	return _log(log.ctx, log.engine, pulumirpc.LogSeverity_WARNING, msg, resource, streamID, ephemeral)
+}
+
+// Logs a fatal error to indicate that the tool should stop processing resource
+func (log *logState) Error(msg string, resource Resource, streamID int32, ephemeral bool) error {
+	return _log(log.ctx, log.engine, pulumirpc.LogSeverity_ERROR, msg, resource, streamID, ephemeral)
+}
+
+func _log(ctx context.Context, engine pulumirpc.EngineClient, severity pulumirpc.LogSeverity,
```
lukehoban: We don't need to use the `_` prefix in Go - this will already be private to this package due to being `lowercase`.

sdk/go/pulumi/context.go
```diff
@@ -50,6 +50,7 @@ type Context struct {
 	rpcsDone    *sync.Cond  // an event signaling completion of RPCs.
 	rpcsLock    *sync.Mutex // a lock protecting the RPC count and event.
 	rpcError    error       // the first error (if any) encountered during an RPC.
+	Log         Log
```
lukehoban: Since this one is a public API - possibly add a comment on it - and perhaps even add a newline between this and the private fields?

sdk/go/pulumi/log.go
tasiah: If I remove the underscore, I get the error `log already declared through import of package log ("log")`. Do you have any other ideas for alternative function names?

lukehoban: `logImpl`?  But honestly fine to leave as-is as well.

EvanBoyle: You can alias the import as well https://www.golang-book.com/books/intro/11

PR #4078, opened 2020-03-13
Update Go logging to use LogArgs

tasiah:
See discussion here https://github.com/pulumi/docs/pull/2643#discussion_r391962960.

Once this has been merged, we should also merge:
- pulumi/docs#2643
- pulumi/pulumi-docker#149 (with updated `go.mod`)
- pulumi/examples#594 (with updated `Gopkg.toml`)

### DIFF
--- a/sdk/go/pulumi/log.go
+++ b/sdk/go/pulumi/log.go
@@ -36,30 +36,41 @@ type logState struct {
 	ctx    context.Context
 }
 
+// LogArgs may be used to specify arguments to be used for logging.
+type LogArgs struct {
+	Resource  Resource
+	StreamID  int32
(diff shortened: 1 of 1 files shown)


sdk/go/pulumi/log.go
```diff
@@ -25,41 +25,52 @@ import (
...
 }
 
 type logState struct {
 	engine pulumirpc.EngineClient
 	ctx    context.Context
 }
 
+// LogArgs may be used to specify arguments to be used for logging.
+type LogArgs struct {
+	Resource  Resource
```
lukehoban: We should add doc comments to these public API properties.

PR #4085, opened 2020-03-16
Fixes for resource docs generator

praneetloke:
This PR contains the following changes/fixes:
* Fix the generation of formal parameters of the function shown in the "Lookup a resource" section by individually generating the formal parameters string.
* "Linkify" the formal parameters in the code snippet of the "Lookup a resource" section.
* Generate the constructor params for Python in code and render the string in the template, so we can control the insertion of word-break tags in the names. This essentially brings it in-sync with how the rest of the constructors were being generated.
* Fix issue with the Go-based Function argument not having the right prefix based on whether or not it was a resource or package-level Function.

The result of these changes can be previewed using the [preview link](http://content-bucket-de682d9.s3-website-us-east-1.amazonaws.com/) provided by Pulumify [here](https://github.com/pulumi/docs/pull/2641#issuecomment-599140931).

### DIFF
--- a/pkg/codegen/python/doc.go
+++ b/pkg/codegen/python/doc.go
@@ -66,7 +66,7 @@ func (d DocLanguageHelper) GetLanguageTypeString(pkg *schema.Package, moduleName
 		case *schema.ObjectType:
 			return getDictWithTypeName(tokenToName(dictionaryTy.Token))
 		default:
-			return "Dict[Any, Any]"
+			return "Dict[str, Any]"
 		}
 	}
 	return name
@@ -106,14 +106,14 @@ func getDictWithTypeName(t string) string {
 }
 
 // getMapWithTypeName returns the Python representation of a dictionary
-// with a key of type `t` and a value of Any.
+// with a string keu and a value of type `t`.
 func getMapWithTypeName(t string) string {
 	switch t {
 	case "string":
-		return "Dict[str, Any]"
+		return "Dict[str, str]"
 	case "any":
-		return "Dict[Any, Any]"
+		return "Dict[str, Any]"
 	default:
-		return fmt.Sprintf("Dict[%s, Any]", t)
+		return fmt.Sprintf("Dict[str, %s]", PyName(t))
 	}
 }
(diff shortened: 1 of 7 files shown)


pkg/codegen/python/doc.go
```diff
@@ -91,11 +92,28 @@ func elementTypeToName(el string) string {
...
+// with a key of type `t` and a value of Any.
+func getMapWithTypeName(t string) string {
+	switch t {
+	case "string":
+		return "Dict[str, Any]"
+	case "any":
+		return "Dict[Any, Any]"
+	default:
+		return fmt.Sprintf("Dict[%s, Any]", t)
+	}
```
justinvp: I don't think this is correct. Schema.MapType is defined as:

https://github.com/pulumi/pulumi/blob/ee62f38ab998d7aaaefd16742f8b94912a37c57b/pkg/codegen/schema/schema.go#L101-L105

> MapType represents maps from strings to particular element types. 

So the dictionary's key should always be `str` and the value should be what `t` is specifying.

pkg/codegen/python/doc.go
```diff
@@ -63,9 +62,11 @@ func (d DocLanguageHelper) GetLanguageTypeString(pkg *schema.Package, moduleName
 		switch dictionaryTy := t.(type) {
 		case *schema.MapType:
 			elType := dictionaryTy.ElementType.String()
-			return getDictWithTypeName(elementTypeToName(elType))
+			return getMapWithTypeName(elementTypeToName(elType))
 		case *schema.ObjectType:
 			return getDictWithTypeName(tokenToName(dictionaryTy.Token))
+		default:
+			return "Dict[Any, Any]"
```
justinvp: Shouldn't this be:

```suggestion
			return "Dict[str, Any]"
```

?

praneetloke: Yes, absolutely! Based on your other comment, the key for this should always be of type string.

pkg/codegen/python/doc.go
praneetloke: Good catch! Fixed it in the latest commit.

PR #4092, opened 2020-03-17
Display resource type in PAC violation output.

### LINKED ISSUE #4061: PAC violations should display resource type
@clstokes brought it to my attention that PAC violation messages no longer include resource type information.

jkisk:
Displays the resource type along with the name of the resource in PAC violation error output.

image

Resolves #4061.


justinvp:
I've lost track...how did it look previously when we showed the resource type?

jkisk:
> I've lost track...how did it look previously when we showed the resource type?

image

jkisk:
> This change looks fine to me, but to opine here we are displaying six different values which can be very confusing to know what you are looking at.
> 
> Would it make sense to split the "policy pack info" on one line, and the "resource information" on the second? Or perhaps use a different color for the resource name/type vs. the policy pack info?
> 
> I don't feel too strongly about it, so if you can't think of something that's strictly better we can just start with this and see how much we like it.

It makes sense to continue to improve this, I will go ahead and merge so the info is surfaced for now but keep thinking about the next iteration.

justinvp:
@jkisk, can we emit the type in the same way as we did before:

```
aws:s3:Bucket
```

Instead of:

```
aws:s3/bucket:Bucket
```

justinvp:
BTW, when we adopt the CLI with these changes in the `pulumi-policy` repo, we need to update all the integration tests for the new output: https://github.com/pulumi/pulumi-policy/blob/master/tests/integration/integration_test.go

PR #4094, opened 2020-03-17
Fix 'pulumi stack ls' on Windows

### LINKED ISSUE #3504: Pulumi Stack ls behavior is incorrect
Note - originally commented here and was asked to file as a new issue: https://github.com/pulumi/pulumi/issues/2679#issuecomment-551992056 

I have a situation where `pulumi stack ls` and `pulumi stack ls --all` both return empty. I have logged in using `pulumi login --local`
```
pulumi stack ls --all
NAME  LAST UPDATE  RESOURCE COUNT
```
If I just do `pulumi stack`  I get some output as below 
```
 pulumi stack
Current stack is nrdev:
    Managed by xxxx
    Last updated: 2 minutes ago (2019-11-08 22:09:02.3442582 +0100 CET)
    Pulumi version: v1.5.1
Current stack resources (4):
    TYPE                                           NAME
    pulumi:pulumi:Stack                            pulumitest-nrdev
    ├─ azure:core/resourceGroup:ResourceGroup      resourceGroup
    ├─ azure:storage/account:Account               storage
    └─ pulumi:providers:azure                      default_1_3_1

Current stack outputs (1):
    OUTPUT            VALUE
    connectionString  DefaultEndpointsProtocol=https;AccountName=xxx;AccountKey=xxxx;EndpointSuffix=core.windows.net

Use `pulumi stack select` to change stack; `pulumi stack ls` lists known ones
```
Current directory structure is as below
```
Mode                LastWriteTime         Length Name
----                -------------         ------ ----
d-----        02-11-2019    20:20                node_modules
-a----        08-11-2019    21:41             23 .gitignore
-a----        08-11-2019    21:41            636 index.ts
-a----        08-11-2019    21:41            208 package.json
-a----        02-11-2019    20:20          82513 package-lock.json
-a----        08-11-2019    21:41             85 Pulumi.dev.yaml
-a----        08-11-2019    21:42            146 Pulumi.nrdev.yaml
-a----        08-11-2019    21:41             88 Pulumi.yaml
-a----        08-11-2019    21:41            456 tsconfig.json
```
Some other outputs as requested earlier

```
> pulumi version
v1.5.1

> pulumi whoami -v
User: <domain>\<user>
Backend URL: file://~

❯ cat ~/.pulumi/credentials.json                           
{                                                          
    "current": "file://~",                                 
    "accessTokens": {                                      
        "file://~": ""                                     
    },                                                     
    "accounts": {                                          
        "file://~": {                                      
            "lastValidatedAt": "0001-01-01T00:00:00Z"      
        }                                                  
    }                                                      
}                                                          
```

chrsmith:
The command `pulumi stack ls` didn't work on Windows when using the local file backend. The reason boils down to the [`blob.Bucket` type](https://godoc.org/gocloud.dev/blob#Bucket) we use, to abstract file IO for the { local disk, Azure, GCP, S3, etc. }

By design, it only supports forward slashes as directory separator characters. So we canonicalize file paths using `file path.ToSlash` before we call its methods, using the `wrappedBucket` interface.

However, we missed the canonicalization for the data we pass to `ListObjects(...)`. Which was why `pulumi stack ls` didn't work on Windows. (But `pulumi stack` did, because we use a different method, which does the path canonicalization.)

This PR also runs the `pkg/backend/` unit tests on Appveyor too, since previously we didn't have any Windows-coverage for these file bucket methods.

Fixes #3504

### DIFF
--- a/pkg/backend/filestate/bucket_test.go
+++ b/pkg/backend/filestate/bucket_test.go
@@ -1,0 +1,94 @@
+package filestate
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"gocloud.dev/blob"
+)
+
+func mustNotHaveError(t *testing.T, context string, err error) {
+	t.Helper()
+	if err != nil {
+		t.Fatalf("Error in testcase %q, aborting: %v", context, err)
+	}
+}
+
+// The wrappedBucket type exists so that when we use the blob.Bucket type we can present a consistent
+// view of file paths. Since it will assume that backslashes (file separators on Windows) are part of
+// file names, and this causes "problems".
+func TestWrappedBucket(t *testing.T) {
+	// wrappedBucket will only massage file paths IFF it is needed, as filepath.ToSlash is a noop.
+	if filepath.Separator == '/' {
+		assert.Equal(t, `foo\bar\baz`, filepath.ToSlash(`foo\bar\baz`))
+		t.Skip("Skipping wrappedBucket tests because file paths won't be modified.")
+	}
+
+	// Initialize a filestate backend, using the default Pulumi directory.
+	cloudURL := FilePathPrefix + "~"
+	b, err := New(nil, cloudURL)
+	if err != nil {
+		t.Fatalf("Initializing new filestate backend: %v", err)
+	}
+	localBackend, ok := b.(*localBackend)
+	if !ok {
+		t.Fatalf("backend wasn't of type localBackend?")
+	}
+
+	wrappedBucket, ok := localBackend.bucket.(*wrappedBucket)
+	if !ok {
+		t.Fatalf("localBackend.bucket wasn't of type wrappedBucket?")
+	}
+
+	ctx := context.Background()
+	// Perform basic file operations using wrappedBucket and verify that it will
+	// successfully handle both "/" and "\" as file separators. (And probably fail in
+	// exciting ways if you try to give it a file on a system that supports "\" or "/" as
+	// a valid character in a filename.)
+	t.Run("SanityCheck", func(t *testing.T) {
+		randomData := []byte("Just some random data")
+
+		err := wrappedBucket.WriteAll(ctx, ".pulumi/bucket-test/foo", randomData, &blob.WriterOptions{})
+		mustNotHaveError(t, "WriteAll", err)
+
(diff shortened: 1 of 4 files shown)


leezen:
🎉 for more Windows test coverage

pkg/backend/filestate/bucket_test.go
```diff
@@ -0,0 +1,94 @@
...
+	if err != nil {
+		t.Fatalf("Error in testcase %q, aborting: %v", context, err)
+	}
+}
+
+// The wrappedBucket type exists so that when we use the blob.Bucket type we can present a consistent
+// view of file paths. Since it will assume that backslashes (file separators on Windows) are part of
+// file names, and this causes "problems".
+func TestWrappedBucket(t *testing.T) {
+	// wrappedBucket will only massage file paths IFF it is needed, as filepath.ToSlash is a noop.
```
jkisk: IFF > IF

pkg/backend/filestate/bucket_test.go
```diff
@@ -0,0 +1,94 @@
...
+	wrappedBucket, ok := localBackend.bucket.(*wrappedBucket)
+	if !ok {
+		t.Fatalf("localBackend.bucket wasn't of type wrappedBucket?")
+	}
+
+	ctx := context.Background()
+	// Perform basic file operations using wrappedBucket and verify that it will
+	// successfully handle both "/" and "\" as file separators. (And probably fail in
+	// exciting ways if you try to give it a file on a system that supports "\" or "/" as
+	// a valid character in a filename.)
```
jkisk: Are there systems that do that? Just curious...

pkg/backend/filestate/bucket_test.go
```diff
@@ -0,0 +1,94 @@
...
+		// Verify the leading slash isn't necessary.
+		err = wrappedBucket.Delete(ctx, ".pulumi/bucket-test/foo")
+		mustNotHaveError(t, "Delete", err)
+
+		exists, err := wrappedBucket.Exists(ctx, ".pulumi/bucket-test/foo")
+		mustNotHaveError(t, "Exists", err)
+		assert.False(t, exists, "Deleted file still found?")
+	})
+
+	// Verify ListObjects / listBucket works with regard to differeing file separators too.
```
jkisk: differeing > differing

pkg/backend/filestate/bucket_test.go
chrsmith: Not to my knowledge.

PR #4098, opened 2020-03-18
Fix rename stack message when attempting to move organizations

leezen:


pkg/backend/httpstate/backend.go
```diff
@@ -791,7 +791,8 @@ func (b *cloudBackend) RenameStack(ctx context.Context, stack backend.Stack, new
 	// for general use yet.
 	if stackID.Owner != newIdentity.Owner {
 		errMsg := "You cannot transfer stack ownership via a rename. If you wish to transfer ownership\n" +
-			"of a stack to another organization, please contact support@pulumi.com."
+			"of a stack to another organization, you can do so in the Pulumi Console by going to the\n" +
+			"\"Settings\" page of the stack and then clicking the \"Transfer Stack\" button."
```
lukehoban: Note that I believe we have all the information we need to synthesize the actual URL to send the user to.

leezen: Ohhh... that's a good point. I'll update the PR.

pkg/backend/httpstate/backend.go
```diff
@@ -787,11 +787,16 @@ func (b *cloudBackend) RenameStack(ctx context.Context, stack backend.Stack, new
 		return err
 	}
 
-	// Transferring stack ownership is available to Pulumi admins, but isn't quite ready
-	// for general use yet.
 	if stackID.Owner != newIdentity.Owner {
+		url := "."
+
+		if consoleURL, err := b.StackConsoleURL(stack.Ref()); err == nil {
```
mikhailshilkov: Not a big deal, but maybe it would make sense to extend `StackConsoleURL` function to accept extra paths as additional parameters?

leezen: As a way to ensure that the path actually exists? Otherwise, not sure how it helps beyond acting like another way to concatenate?

leezen: Going to merge for now, but happy to come back to this if you feel otherwise. Thanks for the review!

Issue #4075, opened 2020-03-12
Update the Function signature for C# invokes in the resource docs generator


praneetloke:
See https://github.com/pulumi/pulumi-terraform-bridge/pull/126 for details on the changes to the way Invokes are done in C#.

praneetloke:
Should we also use the `propertyInfo` object to retrieve the C# name for a property? See this [comment](https://github.com/pulumi/pulumi/pull/4111#discussion_r395052663).

PR #4111, opened 2020-03-19
Update the resource docs generator to use the correct casing for nested property names in Python

### LINKED ISSUE #4095: Resource docs generator must use camelCase name for nested properties in Python
See this [comment](https://github.com/pulumi/docs/pull/2641#discussion_r393293399) for more info.

praneetloke:
Fixes #4095.

This PR is mostly to fix some issues with the Python docs generation.

* This PR also includes the change in the template to temporarily block the resource docs from being discoverable to crawlers
* Use PascalCase for property types of the underlying element types in lists and dictionaries.
* Correctly render the property name based on the property case map similar to how the Python code generator does it.

This docs [PR](https://github.com/pulumi/docs/pull/2675) contains regenerated docs with these changes.


pkg/codegen/docs/gen.go
```diff
@@ -482,8 +491,28 @@ func (mod *modContext) getProperties(properties []*schema.Property, lang string,
...
+			pyName := python.PyName(prop.Name)
+			if snakeCase, ok := camelCaseToSnakeCase[prop.Name]; ok {
+				propLangName = snakeCase
+			} else if camelCase, ok := snakeCaseToCamelCase[pyName]; ok {
+				propLangName = camelCase
+			} else {
+				propLangName = prop.Name
+			}
+		case "go", "csharp":
+			propLangName = strings.Title(prop.Name)
```
justinvp: For C#, do we need to respect the C# name?

https://github.com/pulumi/pulumi/blob/5a730d6ebcf0f134a12817ad90a4f56719580620/pkg/codegen/dotnet/gen.go#L1186-L1203

praneetloke: Hmm I will need to check that. I have a made a comment in https://github.com/pulumi/pulumi/issues/4075 to handle C#-specific fixes. So I'll tackle this in a PR for that. I will need to fix #4075 since the PR for changing the way we do Invokes in C# was merged recently.

justinvp: 👍

pkg/codegen/docs/gen.go
```diff
@@ -479,15 +488,40 @@ func (mod *modContext) getProperties(properties []*schema.Property, lang string,
...
+		propLangName := prop.Name
+		switch lang {
+		case "python":
+			pyName := python.PyName(prop.Name)
+			// The default casing for a Python property name is snake_case unless
+			// it is a property of a nested object, in which case, we should check the property
+			// case maps.
+			propLangName = pyName
+
+			if nested {
```
praneetloke: @justinvp I changed the logic here to only check the property case maps for Python when the properties that are passed in are nested properties. This makes it so that top-level properties use snake_case, by default, except for nested properties. This makes it so that for Functions, for which we do not build a separate property case map (neither does the Python code gen), the properties are snake_case'd except for nested properties, which will honor the case maps.

PR #4126, opened 2020-03-20
Add test for resource docs to confirm Python casing

praneetloke:
This PR also contains a commit to fix property type string generation for Go. Please see this [commit](https://github.com/pulumi/pulumi/commit/fc3f6ec6500b3abeb0d2757b2ef8d529a427440d). The issue was caused by a recent PR (#4110) that was merged today. With that PR merged, the resource docs generator also needs to build a map of Go packages to correctly build the type names. Good thing was that I caught this while writing the test for this PR :D


pkg/codegen/go/gen.go
```diff
@@ -1067,6 +1060,19 @@ func GeneratePackage(tool string, pkg *schema.Package) (map[string][]byte, error
 		}
 	}
 
+	return packages
+}
+
+func GeneratePackage(tool string, pkg *schema.Package) (map[string][]byte, error) {
+	var goInfo GoInfo
```
praneetloke: Nothing really changed in this function other than factoring out some stuff that the resource docs generator also needs access to via the `DocLangHelper` interface.

Issue #3756, opened 2020-01-15
Update Policy gestures to support configuration


ekrengel:
Related to: https://github.com/pulumi/pulumi-policy/issues/91

We will be improving the UI experience to support configuration. The CLI now allows you to enable/disable Policy Packs, so the CLI will also need to be able to support configuring them.

justinvp:
> The CLI now allows you to enable/disable Policy Packs, so the CLI will also need to be able to support configuring them.

@ekrengel, @sean1588, what changes are we planning to make to the commands?

ekrengel:
I think we just have one additional need which is that enabling a PP should check if config is required & support passing in config / checking that its valid. Checking its valid is not a hard requirement thought bc the svc does this.

PR #4127, opened 2020-03-20
Policy pack configuration using policy enable cmd

sean1588:
Issue #3756 

Support adding policy pack configuration through CLI by specifying a policy configuration json file.

Example cmd:
```
pulumi policy enable {orgName}/{policyPackName} --config config.json --version={version}
```

Right now all policy pack config validation is being done server side. Maybe if we feel there would be an advantage to doing config validation on the client we can move to that at some point, but I believe we would have to request the schema from the server anyway, so either way we would need to make an additional api call. I feel we can revisit this at some point if need be. We can always do a follow up PR if we feel doing validation client side is necessary.


pkg/cmd/util.go
```diff
@@ -702,3 +703,20 @@ func updateFlagsToOptions(interactive, skipPreview, yes bool) (backend.UpdateOpt
 		SkipPreview: skipPreview,
 	}, nil
 }
+
+// loadJSONConfigFile loads file with json configuration data and unmarshals into a map[string]*json.RawMessage object.
+func loadJSONConfigFile(filePath string) (map[string]*json.RawMessage, error) {
```
justinvp: Rather than having this function here, I am wondering if we should call the same function we use to load local config specified to the `--policy-pack-config` flag of `pulumi up/preview/watch`.

https://github.com/pulumi/pulumi/blob/61928f04e0524893266e091b4c509f9900bcd4c8/pkg/resource/analyzer/config.go#L31-L37

That way we have consistent behavior of loading local policy config from a file.

This existing function does some validation of the `EnforcementLevel` values, and also, as a convenience, allows specifying enforcement level for a policy like:

```json
{
    "some-policy": "advisory"
}
```

Rather than having to specify it like:

```json
{
    "some-policy": {
        "enforcementLevel": "advisory"
    }
}
```

This would require adding some code to convert `map[string]plugin.AnalyzerPolicyConfig` to `map[string]*json.RawMessage`. Probably the easiest way to do that is to loop through the map and `json.Marshal` each `plugin.AnalyzerPolicyConfig` value to bytes. The bytes can then be cast to `json.RawMessage`. Similar to what's done here:

https://github.com/pulumi/pulumi/blob/c28c6022472572c71875ef2e7e3d5a35f7ea3c96/pkg/backend/httpstate/client/client.go#L633-L641

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ CHANGELOG
   [#4141](https://github.com/pulumi/pulumi/pull/4141)
 - Add missing builtin `MapArray` to Go SDK.
   [#4144](https://github.com/pulumi/pulumi/pull/4144)
+- Add support for policy pack configuration using CLI.
```
ekrengel: Since the changelog is actually used for our releases, we may want to clarify this. "Add support for enabling Policy Packs with configuration" ... or something to that effect.

tests/integration/policy/test_policy_pack/package.json
```diff
@@ -2,9 +2,9 @@
     "name": "aws-policy",
     "version": "0.0.1",
     "dependencies": {
-        "@pulumi/pulumi": "^0.17.28",
-        "@pulumi/aws": "0.18.16",
-        "@pulumi/policy": "^0.2.0"
+        "@pulumi/pulumi": "1.13.0",
+        "@pulumi/aws": "1.27.0",
+        "@pulumi/policy": "0.4.1-dev.1584625475"
```
ekrengel: I think we talked about have one Policy Pack using an older version of the policy repo to ensure backwards compatibility.. is there a reason you changed your mind?

tests/integration/policy/test_policy_pack/index.ts
```diff
@@ -16,6 +16,15 @@ if (!packName) {
                 name: "s3-no-public-read",
                 description: "Prohibits setting the publicRead or publicReadWrite permission on AWS S3 buckets.",
                 enforcementLevel: "mandatory",
+                configSchema: {
+                    properties: {
+                        message: {
+                            "type": "string",
```
ekrengel: nit: I think type, minLength and maxLength do not need to be quotes & may make this look cleaner.

pkg/cmd/policy_enable.go
```diff
@@ -48,15 +50,32 @@ func newPolicyEnableCmd() *cobra.Command {
...
+				})
 		}),
 	}
 
 	cmd.PersistentFlags().StringVar(
 		&args.policyGroup, "policy-group", "",
 		"The Policy Group for which the Policy Pack will be enabled; if not specified, the default Policy Group is used")
 
+	cmd.PersistentFlags().StringVar(
+		&args.configFile, "config-file", "",
```
ekrengel: I think when running a policy-pack locally there is a different flag name used for the config file. Should we use the same here for consistency?

pkg/cmd/util.go
```diff
@@ -702,3 +703,20 @@ func updateFlagsToOptions(interactive, skipPreview, yes bool) (backend.UpdateOpt
 		SkipPreview: skipPreview,
 	}, nil
 }
+
+// loadJSONConfigFile loads file with json configuration data and unmarshals into a map[string]*json.RawMessage object.
```
ekrengel: Since this is in a generic util file, maybe be more specific about its use in the comment? Or just move it to the pkg/cmd/policy_enable.go file. I dont think its used anywhere else?

pkg/cmd/policy_enable.go
justinvp: For `pulumi preview/up/watch`, the flag is `--policy-pack-config`. Since this command is already about policies, I think we can leave off the `policy-pack` prefix, and just use `--config`.

tests/integration/policy/test_policy_pack/configs/invalid-config.json
```diff
@@ -0,0 +1,6 @@
+{
+    "s3-no-public-read":{
```
justinvp: Nit:

```suggestion
    "s3-no-public-read": {
```

tests/integration/policy/test_policy_pack/configs/invalid-config.json
```diff
@@ -0,0 +1,6 @@
...
+        "enforcementLevel":"mandatory",
```
justinvp: Nit:

```suggestion
        "enforcementLevel": "mandatory",
```

tests/integration/policy/test_policy_pack/configs/invalid-config.json
```diff
@@ -0,0 +1,6 @@
...
+        "message":"this message is invalid because it is way more than 10 characters......"
```
justinvp: Nit:

```suggestion
        "message": "this message is invalid because it is way more than 10 characters......"
```

tests/integration/policy/test_policy_pack/configs/valid-config.json
```diff
@@ -0,0 +1,10 @@
+{
+    "all":{
```
justinvp: Nit:

```suggestion
    "all": {
```

tests/integration/policy/test_policy_pack/configs/valid-config.json
```diff
@@ -0,0 +1,10 @@
...
+        "enforcementLevel":"advisory"
```
justinvp: Nit:

```suggestion
        "enforcementLevel": "advisory"
```

tests/integration/policy/test_policy_pack/configs/valid-config.json
```diff
@@ -0,0 +1,10 @@
...
+    },
+    "s3-no-public-read":
+    {
+        "enforcementLevel":"mandatory",
```
justinvp: Nit:

```suggestion
        "enforcementLevel": "mandatory",
```

tests/integration/policy/test_policy_pack/configs/valid-config.json
```diff
@@ -0,0 +1,10 @@
...
+        "message":"test msg"
```
justinvp: Nit:

```suggestion
        "message": "test msg"
```

pkg/cmd/util.go
sean1588: Ok. I will leverage that function so we can use the validations it has and then just convert  the types.

pkg/cmd/util.go
sean1588: Ok, I went ahead and relocated to policy_enable.co and then leveraged the function justin is suggesting since it does some validations.

pkg/cmd/policy_enable.go
```diff
@@ -48,15 +52,52 @@ func newPolicyEnableCmd() *cobra.Command {
...
+func loadPolicyConfigFromFile(file string) (map[string]*json.RawMessage, error) {
+	analyzerPolicyConfig, err := resourceanalyzer.LoadPolicyPackConfigFromFile(file)
+	if err != nil {
+		return nil, err
+	}
+
+	// Convert type map[string]plugin.AnalyzerPolicyConfig to map[string]*json.RawMessage.
+	config := map[string]*json.RawMessage{}
+	for k, v := range analyzerPolicyConfig {
+		bytes, err := json.Marshal(v)
```
justinvp: I don't think we can just marshal `v` directly.

Doing so results in JSON like:

```json
{"EnforcementLevel":"mandatory","Properties":{"foo":"bar"}}
```

The value needs to be converted into the format expected by the service. Something like:

```go
func marshalAnalyzerPolicyConfig(v AnalyzerPolicyConfig) (*json.RawMessage, error) {
	m := make(map[string]interface{})
	for k, v := range v.Properties {
		m[k] = v
	}
	if v.EnforcementLevel != "" {
		m["enforcementLevel"] = v.EnforcementLevel
	}
	b, err := json.Marshal(m)
	if err != nil {
		return nil, err
	}
	raw := json.RawMessage(b)
	return &raw, nil
}
```

Which will result in:

```
{"enforcementLevel":"mandatory","foo":"bar"}
```

I created a Go Playground to demonstrate: https://play.golang.org/p/L8GYTzTAsOR

sean1588: Correct, I noticed it structured the configurable values under the Properties key. Thanks for adding that code snippet btw. I updated to convert the values.

tests/integration/policy/test_policy_pack/package.json
sean1588: nope, was actually just a miss on my part. the only way I could get this to work was just to add another test policy pack directory with the old version of the policy sdk. Not sure how far back in versions you want to test, so I just used the version that we were using for the tests before I made these changes. I also just made a completely different test case with all the old existing tests, since the set of commands we were using in the test I believe are only backwards compatible not forwards compatible with the changes, so I had to keep a lot of the old commands that we were testing with previously as well, for example the way we were specifying pp versions.

sean1588:
OK, this is ready for a re-review. I made a few changes since last review:

1. I updated the code to load the policy config to use the function in the resource analyzer package that Justin suggested so that we could leverage some of the validations it had. This also involved doing some conversion to convert the type being returned from that function to what the api service was expecting.

2. I also added the tests to test for backwards compatibility with other versions of the policy sdk. The only way I could reasonably do this is by adding another test policy pack directory with the old version. We could probably do this in a way that we could dynamically manipulate this through code at runtime and have only one test policy pack, but it ended up being a PITA and not worth the time because we need to also generate the index.ts file as well that is compatible with that particular version of the sdk. I also a specific test case with all the old tests for that specific version and just copied over all the old tests we were running prior to the change. I needed to retain the old commands because they are only backwards compatible not forwards. It looks like a lot of changes and the test code could be dry-ed up to some degree but would probably make them less readable and hard to follow, so I think its best not to IMO and just leave as is.

tests/integration/policy/policy_pack_v0-4-1-dev/package.json
```diff
@@ -0,0 +1,14 @@
+{
```
ekrengel: wondering if it makes more sense to call this dir `policy_pack_w_config` since the version is probably meaningless to most people?

tests/integration/policy/policy_test.go
```diff
@@ -14,9 +14,61 @@ import (
 )
 
 // TestPolicy tests policy related commands work.
-func TestPolicy(t *testing.T) {
-	t.Skip("Temporarily skipping test that is causing unrelated tests to fail - pulumi/pulumi#4149")
+func TestPolicyVersion0_4_1_dev(t *testing.T) {
```
ekrengel: Same with naming here.

tests/integration/policy/policy_test.go
```diff
@@ -14,9 +14,61 @@ import (
...
+
+	// Enable and Disable without specifying the version number.
+	e.RunCommand("pulumi", "policy", "enable", fmt.Sprintf("%s/%s", orgName, policyPackName), "latest")
+	e.RunCommand("pulumi", "policy", "disable", fmt.Sprintf("%s/%s", orgName, policyPackName))
+
+	e.RunCommand("pulumi", "policy", "rm", fmt.Sprintf("%s/%s", orgName, policyPackName), "0.0.1")
+	e.RunCommand("pulumi", "policy", "rm", fmt.Sprintf("%s/%s", orgName, policyPackName), "all")
+}
+
+func TestPolicyVersion0_2_0(t *testing.T) {
```
ekrengel: Lets change naming here too

pkg/cmd/policy_enable.go
```diff
@@ -48,15 +53,69 @@ func newPolicyEnableCmd() *cobra.Command {
...
 		}),
 	}
 
 	cmd.PersistentFlags().StringVar(
 		&args.policyGroup, "policy-group", "",
 		"The Policy Group for which the Policy Pack will be enabled; if not specified, the default Policy Group is used")
 
+	cmd.PersistentFlags().StringVar(
+		&args.config, "config", "",
+		"The file path for the policy pack configuration file")
```
ekrengel: `Policy Pack`

pkg/cmd/policy_enable.go
```diff
@@ -48,15 +53,69 @@ func newPolicyEnableCmd() *cobra.Command {
...
 }
+
+func loadPolicyConfigFromFile(file string) (map[string]*json.RawMessage, error) {
+	analyzerPolicyConfigMap, err := resourceanalyzer.LoadPolicyPackConfigFromFile(file)
+	if err != nil {
+		return nil, err
+	}
+
+	// Convert type map[string]plugin.AnalyzerPolicyConfig to map[string]*json.RawMessage.
+	config := map[string]*json.RawMessage{}
```
ekrengel: Hmm not sure how this is working, but don't you normally have to use `make` for a map? 🤔

tests/integration/policy/policy_pack_v0-4-1-dev/package.json
sean1588: Agreed, that's a more meaningful name. will update

pkg/cmd/policy_enable.go
sean1588: It is equivalent to initalizing an empty map using make. Only difference is not including the curly braces at the end when using the make function.  `make(map[string]*json.RawMessage)`.

pkg/cmd/policy_enable.go
```diff
@@ -48,15 +53,69 @@ func newPolicyEnableCmd() *cobra.Command {
 				version = &cliArgs[1]
 			}
 
+			// Load the configuration from the user-specified JSON file into config object.
+			var config map[string]*json.RawMessage
+			if args.config != "" {
+				// config, err = loadJSONConfigFile(args.configFile)
```
justinvp:
```suggestion

```

pkg/cmd/policy_enable.go
sean1588: Using `make` seems to be the more consistent pattern for us from a quick glance in our code base. So went and just changed to using make while I was in there.  😃

pkg/cmd/policy_enable.go
sean1588: lol... sorry. completely missed that. will make change.

PR #4134, opened 2020-03-20
Fix the signature for C# invokes in the resource docs generator

### LINKED ISSUE #4075: Update the Function signature for C# invokes in the resource docs generator

praneetloke:
Fixes #4075.

This PR fixes two things:
* The `function.tmpl` template has been updated for C# to reflect the new way of using Functions using function classes with a static async invoke function.
* A new interface function has been added to DocLanguageHelper to get the appropriate property name for a language.
  * This was done to enable properties to use any language-level info that may be attached to the property in the schema.

(The above changes are split by commit.)

Here's the docs [PR](https://github.com/pulumi/docs/pull/2687) (also split by commit) to show the effect of these changes.

### DIFF
--- a/pkg/codegen/docs/templates/function.tmpl
+++ b/pkg/codegen/docs/templates/function.tmpl
@@ -19,7 +19,12 @@
 <div class="highlight"><pre class="chroma"><code class="language-go" data-lang="go"><span class="k">func </span>Lookup{{ .ResourceName }}<span class="p">(</span>{{ htmlSafe .FunctionArgs.go }}<span class="p">) (*{{ template "linkify_param" .FunctionResult.go }}, error)</span></code></pre></div>
 
 <!-- C# -->
-<div class="highlight"><pre class="chroma"><code class="language-csharp" data-lang="csharp"><span class="k">public static </span>Task<{{ template "linkify_param" .FunctionResult.csharp }}> <span class="p">Get{{ .ResourceName }}(</span>{{ htmlSafe .FunctionArgs.csharp }}<span class="p">)</span></code></pre></div>
+<div class="highlight">
+<pre class="chroma">
+<code class="language-csharp" data-lang="csharp"><span class="k">public static class </span><span class="nx">Get{{ .ResourceName }} </span><span class="p">{</span>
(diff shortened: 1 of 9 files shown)


pkg/codegen/docs/templates/function.tmpl
```diff
@@ -19,7 +19,12 @@
 <div class="highlight"><pre class="chroma"><code class="language-go" data-lang="go"><span class="k">func </span>Lookup{{ .ResourceName }}<span class="p">(</span>{{ htmlSafe .FunctionArgs.go }}<span class="p">) (*{{ template "linkify_param" .FunctionResult.go }}, error)</span></code></pre></div>
 
 <!-- C# -->
-<div class="highlight"><pre class="chroma"><code class="language-csharp" data-lang="csharp"><span class="k">public static </span>Task<{{ template "linkify_param" .FunctionResult.csharp }}> <span class="p">Get{{ .ResourceName }}(</span>{{ htmlSafe .FunctionArgs.csharp }}<span class="p">)</span></code></pre></div>
+<div class="highlight">
+<pre class="chroma">
+<code class="language-csharp" data-lang="csharp"><span class="k">public static class </span><span class="nx">Get{{ .ResourceName }} </span><span class="p">{</span>
+    <span class="k">public static </span>Task<{{ template "linkify_param" .FunctionResult.csharp }}> <span class="p">InvokeAsync(</span>{{ htmlSafe .FunctionArgs.csharp }}<span class="p">)</span>
```
justinvp: I noticed the return type is fully qualified, for example:

http://content-bucket-b916193.s3-website-us-east-1.amazonaws.com/docs/reference/pkg/aws/s3/getbucket/

Screen Shot 2020-03-20 at 12 54 11 PM

Can we remove the namespace (`GetBucketResult` rather than `Pulumi.Aws.S3.GetBucketResult`) similar to what we do for the parameters?

praneetloke: Done. The latest [commit](https://github.com/pulumi/docs/pull/2687/commits/3ab7964b5bfeaac0a78a4ddd4db87a553a4420ef) in the companion docs PR I have open, reflects this change.

praneetloke: Oh also note that, I removed the namespace from property type names throughout as well. Let me know if you would like me to revert to showing the full namespace for the property type names, but just not for the invoke result.

Issue #4139, opened 2020-03-20
Go programs with errors fail with no report on the errors


lukehoban:
If I have any error in my Go program, then `pulumi preview` now just fails with:

```
$ pulumi preview
Previewing update (dev):
 
Permalink: https://app.pulumi.com/lukehoban/gok8s/dev/previews/a660e10e-b64c-447e-a96f-befcce53ea0c
error: failed to discover plugin requirements: failed to execute program cmd: exit status 2
```

This is because we try to `go run` the program in `GetRequiredPlugins`, but don't report any of the output.

PR #4141, opened 2020-03-20
capture std err when getting required plugins for go

EvanBoyle:
Fixes https://github.com/pulumi/pulumi/issues/4139

When go programs have compile time errors, we aren't capturing the error output. We obscure the error behind a generic "get required plugins" failure. 

The new output for a program with a compile-time error looks like:

```
Evans-MBP:go_aws_qs evanboyle$ pulumi up
Previewing update (dev):
     Type                 Name           Plan     Info
     pulumi:pulumi:Stack  go_aws_qs-dev           2 messages
 
Diagnostics:
  pulumi:pulumi:Stack (go_aws_qs-dev):
    # go_aws_qs
    ./main.go:19:30: cannot use name (type interface {}) as type string in argument to s3.NewBucket: need type assertion
 
error: failed to discover plugin requirements: failed to execute program cmd: exit status 2

```

Note that we're already capturing stdout  in line 172, so we don't redirect it here.


sdk/go/pulumi-language-go/main.go
```diff
@@ -167,6 +167,7 @@ func (host *goLanguageHost) GetRequiredPlugins(ctx context.Context,
 
 	cmd.Env = os.Environ()
 	cmd.Env = append(cmd.Env, "PULUMI_PLUGINS=true")
+	cmd.Stderr = os.Stderr
```
lukehoban: Do we intentionally not want to redirect SdOut as well?

EvanBoyle: from the description: 

> Note that we're already capturing stdout in line 172, so we don't redirect it here.

If you try to capture stdout, `cmd.Output()` will return an error saying that stdout is already bound.

Issue #4135, opened 2020-03-20
Go SDK missing builtin pulumi.MapArrayOutput


EvanBoyle:
In the AWS sdk we're generating a type `MapArrayOutput` https://github.com/pulumi/pulumi-aws/blob/e9783ed59f49c6bdaddfd5b7d100fca13fb10d71/sdk/go/aws/autoscaling/group.go#L161

based on this chunk of schema:

```json
                "tagsCollection": {
                    "type": "array",
                    "items": {
                        "type": "object",
                        "additionalProperties": {
                            "$ref": "pulumi.json#/Any"
                        }
                    },
                    "description": "A list of tag blocks (maps). Tags documented below.\n"
                },
```

Given what I see in the schema, the expected type seems correct, we just need to generate another builtin similar to what we did in https://github.com/pulumi/pulumi/pull/4086 for `ArrayMapOutput`

PR #4144, opened 2020-03-20
Add missing builtin MapArray to the Go SDK

EvanBoyle:
fixes https://github.com/pulumi/pulumi/issues/4135


EvanBoyle:
@pgavlin by stdlib you mean `type_builtins`, and code generator is what runs on the json schema in the provider repos? 

How would the code gen know how to synthesize these missing types?

PR #4150, opened 2020-03-21
Temporarily skip test causing unrelated failures in master

justinvp:
@ekrengel, I know you generally prefer to not disable tests, but let's temporarily disable this test to get master green, until it's been updated to not use the default policy group.


justinvp:
> There’s also a policy failure test that idk if you want to disable as well?

Which one?

ekrengel:
JK... that’s in the service repo 🤦‍♀️

PR #4167, opened 2020-03-24
Another attempt at removing the need for vendoring

EvanBoyle:
Key things that changed:

1. The original pr forgot to move github.com/pulumi/pulumi/main.go. This has been moved to pkg/cmd/pulumi and the makefile was updated accordingly (build install commands)
2. Test commands had to be updated to execute from the correct directories (examples, tests).
3. Integration testing framework was updated to initialize a go.mod for go programs, and do a local replace if an enlistment of pulumi/pulumi is detected. This makes sure we're testing local bits and not downloading a previously released tag of pulumi/pulumi in the `go mod download` step that runs for each go integration test.


Makefile
```diff
@@ -42,25 +43,26 @@ generate::
...
 test_fast::
-	$(GO_TEST_FAST) ${PROJECT_PKGS}
+	cd pkg && $(GO_TEST_FAST) ${PROJECT_PKGS}
 
 test_all::
-	$(GO_TEST) ${PROJECT_PKGS}
-	$(GO_TEST) -v -p=1 ${EXTRA_TEST_PKGS}
+	cd pkg && $(GO_TEST) ${PROJECT_PKGS}
+	cd examples && $(GO_TEST) -v -p=1 ${EXAMPLES_PKGS}
+	cd tests && $(GO_TEST) -v -p=1 ${TESTS_PKGS}
```
lukehoban: This will slow down the build by not running these two sets of tests in parallel.  Is it the case that we have no other option since these no need to be `go test` runs in different folders with different `go.mod`s?

We may want to move `examples` into `tests` as follow up (the examples are really tests not examples anyway, and are misleading as a top level `examples` as they are not helpful for users).

pkg/testing/integration/program.go
```diff
@@ -1728,6 +1728,37 @@ func (pt *ProgramTester) prepareGoProject(projinfo *engine.Projinfo) error {
...
+
+	// if we have a local pulumi enlistment, test against that
+	_, err = os.Stat(pulumiSdkPath)
+	if err == nil {
+		editStr := fmt.Sprintf("github.com/pulumi/pulumi/sdk=%s", pulumiSdkPath)
+		err = pt.runCommand("go-mod-edit", []string{goBin, "mod", "edit", "-replace", editStr}, cwd)
+		if err != nil {
+			return err
+		}
+	}
```
lukehoban: We should generalize this to use the same mechanism we use everywhere else - which is to use the `Dependencies` field to specify dependencies that should use "linked" versions (locally built).  There's code that does this for Dep a few lines further down here.

We should not need to special case `pulumi/pulumi/sdk` here - as we should be able to ensure that all tests correctly include `pulumi/pulumi/sdk` in their `Dependencies` list in this repro (and in other repos, they will need to include other  modules, like `pulumi/pulumi-aws/sdk/go/aws` in the AWS package.

pkg/testing/integration/program.go
```diff
@@ -1728,6 +1728,37 @@ func (pt *ProgramTester) prepareGoProject(projinfo *engine.Projinfo) error {
...
+
+	err = pt.runCommand("go-mod-download", []string{goBin, "mod", "download"}, cwd)
+	if err != nil {
+		return err
+	}
+
 	// We only need to run dep-ensure if there is a Gopkg.toml file
 	_, err = os.Stat(filepath.Join(cwd, "Gopkg.toml"))
```
lukehoban: It feels like the code below will interact awkwardly with the new code in the case that there is a `Gopkg.toml` but not a `go.mod`.  We'll end up preparing in both ways?  Seems we should either drop support for Gopkg.toml (and deal with the break thaat causes for existing tests as we bring in these updates), or else switch on whether there is a `Gopkg.toml` or `go.mod` and perhaps fail if there is neither?

Makefile
EvanBoyle: Sounds good. I'll move examples as a part of this change.

pkg/testing/integration/program.go
EvanBoyle: I'll remove the dep path and we can deal with upgrading tests incrementally as this gets rolled out.

pkg/testing/integration/program.go
EvanBoyle: Good point. I'll make this change.

Makefile
EvanBoyle: I'm actually going to do this as a part of another change given that we're having some issues with `make build` that this changeset will resolve.

PR #4175, opened 2020-03-25
Fix versions of dependencies in policy tests

lukehoban:
These were causing *very* old versions of `@pulumi/pulumi` to get pulled in, and were seeing failures on Node 13 due to missnig gRPC native modules.

Aside - we've tried to not have any dependencies in `@pulumi/pulumi` on `@pulumi/aws` and other higher-level libraries, to avoid layering violation issues.  Would love to see if we can reasonably simplify the testing at this layer to not have this dependency.


sean1588:
@justinvp yea I can look into removing the dependency.

Issue #4158, opened 2020-03-23
Process examples short-code in the language generators


praneetloke:
Related to https://github.com/pulumi/pulumi-terraform-bridge/issues/122.

Remove non-relevant code examples from the doc comments of each language generator.

**Note**: The PR that introduced the examples short-code in the tfgen tool had to be temporarily [reverted](https://github.com/pulumi/pulumi-terraform-bridge/pull/136). It will need to be added back to that tool in-parallel to working on this issue.

* [x] Update schema-based language code generators in pulumi/pulumi to strip the non-relevant examples
* [x] Do the same in the old code generators in the pulumi-terraform-bridge repo
* [x] Update the provider repos with the latest version pulumi-terraform-bridge and regenerate the schema (blocked by https://github.com/pulumi/pulumi-terraform-bridge/issues/144)

praneetloke:
It appears that we have now updated most provider repos to use the latest version of tfgen from `pulumi-terraform-bridge`. I will keep this open until we have regenerated the docs to include the examples short codes.

praneetloke:
Examples short codes are making their way into the providers as and when the providers are updated to take the latest version of the `pulumi-terraform-bridge` dependency. I'll close this out for now.

PR #4184, opened 2020-03-26
Remove non-relevant examples and empty examples section from docs comments

praneetloke:
Part of #4158.

This PR introduces a package-level function in `codegen` for the language generators to use in order to strip out the examples short code and also any examples that are in languages not relevant to a specific language. For example, for a TS SDK, the Python and the C# examples are unnecessary and irrelevant and so on...

There is a companion PR for this in the pulumi-terraform-bridge to be opened. I have the changes there that re-introduce the examples short code (as well as updates to the old language code generators), but will open that PR once this is merged.

### DIFF
--- a/pkg/codegen/docs.go
+++ b/pkg/codegen/docs.go
@@ -30,3 +30,12 @@ type DocLanguageHelper interface {
 	// an existing resource.
 	GetResourceFunctionResultName(resourceName string) string
 }
+
+// StripNonRelevantExamples strips the non-relevant language snippets from a resource's description.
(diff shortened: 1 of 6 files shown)


praneetloke:
> We'll be using the Pulumi-schema-based code generators for all languages?

Not yet. We currently use it for Go and Python for most of the providers, except for K8s, which only uses the Go generator from here. For C# and NodeJS, the generators in the pulumi-terraform-bridge repo are used. I have changes to those generators to call the same function introduced in this PR, for which I'll open a PR shortly.

PR #4186, opened 2020-03-26
Add cmd to support policy pack config validation

sean1588:
Add command to support policy pack config validation to enable users to validate pp config against config schema.

CMD example:
```
pulumi policy validate {org}/{policyPack} {version} --config {configFile}
```

This will cmd will fetch the version of the json schema from the service associated with the pp and run a validation from the specified config file against the schema.

I "stole" the validation logic directly from the backend service. There appears to be an existing schema validation func, that looks to be a bit different and only accpeted the type AnalyzerPolicyConfig, so I would need to do some type of conversion to leverage it, which I fought with for a bit, but in the end I copied what was being done server side, so at least the way we are validating is consistent. Maybe we can move this func to some pkg that can be imported for use by both the server and the cli at some point... Maybe we can add more tests for the validation, but since this is more of a convenience function I think it is fine since the service will have the final say anyway before applying, so I am willing to call it good enough in the interest of time.


pkg/resource/analyzer/config.go
```diff
@@ -192,6 +192,38 @@ func validatePolicyConfig(schema plugin.AnalyzerPolicyConfigSchema, config map[s
 	return errors, nil
 }
 
+// ValidatePolicyPackConfig validates a policy pack configuration against the specified config schema.
+func ValidatePolicyPackConfig(schemaMap map[string]apitype.PolicyConfigSchema,
```
sean1588: I took this from the pulumi-service, so it is identical for the most part. Maybe we can move to a common pkg where we can import for both the cli and service at some point. There is a func above this that also seems to be doing some schema validation....  but only accpeted the type AnalyzerPolicyConfig, so I would need to do some type of conversion to leverage it, which I fought with for a bit, but in the end I just copied what was being done server side, so at least the way we are validating is consistent with what will be done when the pp is enabled.

pkg/cmd/policy_validate.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/pulumi/pulumi/pkg/backend"
+	"github.com/pulumi/pulumi/sdk/go/common/util/cmdutil"
+	"github.com/spf13/cobra"
+)
+
+func newPolicyValidateCmd() *cobra.Command {
+	var argConfig string
+
+	var cmd = &cobra.Command{
+		Use:   "validate <org-name>/<policy-pack-name> <version>",
```
ekrengel: Hmm I am wondering if this should be "validate-config" ... I think that just "validate" may get confusing, like does it validate the whole PP? @justinvp thoughts?

ekrengel: We should support "latest" for this like we do for the enable command

pkg/cmd/policy_validate.go
```diff
@@ -0,0 +1,72 @@
...
+		Args:  cmdutil.ExactArgs(2),
+		Short: "[PREVIEW] Validate Policy Pack configuration against config schema",
```
ekrengel: "Validate a Policy Pack configuration" --> not sure if we need to expose details of the schema here... that being said if we want to I would say "...against its configuration schema"

pkg/cmd/policy_validate.go
```diff
@@ -0,0 +1,72 @@
...
+		Long:  "Validate Policy Pack configuration against the specified version of its configuration schema.",
```
ekrengel: Maybe: "Validate a Policy Pack configuration against the configuration schema of the specified version."

pkg/cmd/policy_validate.go
```diff
@@ -0,0 +1,72 @@
...
+			if err != nil {
+				return err
+			}
+
+			// Get version from cmd argument
+			version := &cliArgs[1]
+
+			// Load the configuration from the user-specified JSON file into config object.
+			var config map[string]*json.RawMessage
+			if argConfig != "" {
```
ekrengel: If argConfig is empty, this will just validate an empty config? Seems like a waste, but maybe fine since the config could require values and then an empty config would be invalid... wondering if we should output some warning thought like BTW you didnt give us config? Seems like it could be a very easy mistake to make from a user perspective.

pkg/cmd/policy_validate.go
```diff
@@ -0,0 +1,72 @@
...
+				return err
+			}
+			fmt.Println("Policy Pack configuration is valid.")
+			return nil
+		}),
+	}
+
+	cmd.Flags().StringVar(&argConfig, "config", "",
+		"The file path for the Policy Pack configuration file")
+	cmd.MarkFlagRequired("config") // nolint: errcheck
```
ekrengel: What if you want to validate an empty config is okay? I think it may be okay to no support this.. just bc if this field isnt required then things get weird / easy to make a mistake... and you can always just pass a file with an empty `{}` to validate the empty config is ok

pkg/cmd/policy_validate.go
sean1588: I had tried supporting latest, but api call was failing. From what I can tell, it's because the api isnt configured to support latest in the call path `api/orgs/{orgName}/policypacks/{policyPackName}/versions/{version}/schema`. When sending latest the call just came back saying it couldn't find the version. So I think this may involve some work on service side to support this functionality.....

ekrengel: ah yes this handler is def not implemented.. kk lets not worry about that for now then (if we get a request then we can add it)

pkg/cmd/policy_validate.go
sean1588: agreed. i think that sounds better!

pkg/cmd/policy_validate.go
sean1588: I marked the flag as required, because at least the way it is written now, the command requires a config file or will error out. So I just made it required up front. You can however pass a file with "{}" if you want. Not sure if this is what you are asking though.... maybe I am misunderstanding.

pkg/cmd/policy_validate.go
sean1588: Good point! yea, for whatever reason idk why I was assuming this might be covered by marking the flag as required, but now coming to think of it, it probably would allow to run with not specifying anything and be a bad user experience... will address. thanks.

sean1588: ok. just tested it. the way it is written, it actually will not let you proceed without specifying a file and print a message telling the user they need to specify a config file, which I believe is the behavior we want. I did however go and add an integration test though to verify behavior with file that is empty and file that has `{}`.

pkg/resource/analyzer/config.go
```diff
@@ -192,6 +192,38 @@ func validatePolicyConfig(schema plugin.AnalyzerPolicyConfigSchema, config map[s
...
+	config map[string]*json.RawMessage) (err error) {
+	for property, schema := range schemaMap {
```
ekrengel: Hmm so I was just thinking about this as I do the documentation.. I'm not sure this respects the top level "required" field.. does it? We may need to rethink this logic.

ekrengel: Which would also be the case in the service -- and also another good argument for centralizing the logic

pkg/cmd/policy_validate.go
justinvp: +1 to `validate-config`.

PR #4189, opened 2020-03-26
Apply initial config in ReconcilePolicyPackConfig

jkisk:


pkg/resource/analyzer/config_test.go
```diff
@@ -820,7 +820,7 @@ func TestReconcilePolicyPackConfigSuccess(t *testing.T) {
 
 	for _, test := range tests {
 		t.Run(test.Test, func(t *testing.T) {
-			result, validationErrors, err := ReconcilePolicyPackConfig(test.Policies, test.Config)
```
justinvp: Mind adding some tests to validate that the initial config is taken into account?

jkisk: sure thing!

PR #4195, opened 2020-03-26
 Move Windows release builds to happen in GHA + on master

stack72:
The windows release was appveyor calling a specific build target when
a tag was present. This is now called on a push to any of the branches
in the push category

This means it will not trigger on the pushes to the PR builds


.appveyor.yml
```diff
@@ -1,74 +0,0 @@
-version: 1.0.{build}
```
lukehoban: :tada::tada::tada:

.github/workflows/windows-build.yml
```diff
@@ -2,7 +2,10 @@ on:
   push:
     branches:
       - master
-  pull_request:
+      - /feature.*/
+    tags:
+      - v1.*
+      - v2.*
```
lukehoban: In Travis we do this for `tag IS present` - any reason to be different here?

.github/workflows/windows-build.yml
```diff
@@ -2,7 +2,10 @@ on:
   push:
     branches:
       - master
-  pull_request:
+      - /feature.*/
```
lukehoban: Should this be `/^feature.*/`?  Should it more precisely match Travis which includes `^feature-` and `^features/`?

.github/workflows/windows-pr.yml
```diff
@@ -0,0 +1,91 @@
...
+        go mod download
+        popd
+    - name: Build Pulumi
+      run: |
+        cd src\github.com\${{ github.repository }}
+        dotnet msbuild /t:Build /v:Detailed build.proj /p:PulumiRoot="D:\\Pulumi"
+    - name: Run Pulumi Tests
+      run: |
+        cd src\github.com\${{ github.repository }}
+        dotnet msbuild /t:Tests /v:Detailed build.proj /p:PulumiRoot="D:\\Pulumi"
```
lukehoban: It's a real shame we have to duplicate all of this.   Do we have any option to reuse more here across PR/Build?  Or even have a single action and parameterize it somehow?

.github/workflows/windows-build.yml
stack72: we can loosen the restrictions here if you would rather

.github/workflows/windows-build.yml
stack72: this regex will grab both 'features/' and 'feature-'

.github/workflows/windows-pr.yml
stack72: I have a question out to the PM of GHA about this - there *should* be a way to do this in 1 conditional but the docs are so ambigious - I thought it better to duplicate and then make it better rather than risk making a release on a PR

.github/workflows/windows-build.yml
lukehoban: I guess I'm just suggesting we do exactly the same thing in both Travis and GHA.  No reason to be subtly different.

stack72: no worries!

stack72:
@lukehoban I matched the runner to be that of travis - any tag + `master` or `features/` or `feature-`

stack72:
FWIW, the duplication of the files here is temporary while I get an answer about how we can have a condition to run the publish on anything other than a PR

PR #4196, opened 2020-03-26
Move examples tests to /test/examples

EvanBoyle:
This gets us back to the same test parallelism that we had pre-modules.


EvanBoyle:
Not sure why, but in order to get the engine perf tests to pass on Windows I had to run `examples` and `integration` tests separately. Note that the windows build takes 17 mins where the travis build takes ~40 (aka this is still acceptable as unix builds are still the limiting factor).

Issue #4118, opened 2020-03-19
Ever-changing `pulumiManifest.go`


mikhailshilkov:
When used in a provider, the latest master generates `pulumiManifest.go` files. The files have a version in them, so the files change on every generation and cause build failures.

lukehoban:
We have decided to just have individual providers add all `pulumiManifest.go` files to their `.gitignore`.  This will be part of the process of adopting the latest Go code-generator, and so we can close this issue out.

PR #4210, opened 2020-03-27
make go codegen for plugin registration deterministic

EvanBoyle:
This updates the generated code in our providers to shell out and dynamically acquire a version, rather than writing the version into the file. This fixes the fact that `pulumiManifest.go` changes on every build: https://github.com/pulumi/pulumi/issues/4118

When we gitignored the pulumiManifest files, we didn't consider that that would mean that they wouldn't be distributed via `go mod download` commands run by end users. As a result, plugin registration is currently a noop.

This change takes the following approach:

1. embed https://github.com/pulumi/pulumi/blob/master/scripts/get-version as a string in `pulumiManifest.go`
2. when the `init()` function is called, create a temp file, write the contents of the script into it, and then execute the script via the temp file, from the directory of the pulumiManifest.go file. 

Without a windows machine handy (and with an intuition that go windows users are a smaller fraction of our user base), I didn't implement this for windows. 

If there are any errors in this path, or we fail to acquire a version, we just skip plugin acquisition altogether which is the current behavior today. 

In addition, this hack isn't as performant as something natively implemented in go. We should do this (which would make it trivial to support windows as well), but in the interest of time and 2.0 deadlines, I think we shouldn't let perfect be the enemy of done. I've followed this issue to follow up on this work Post 2.0 or sooner if time allows https://github.com/pulumi/pulumi/issues/4211

The generated pulumiManifest.go:

```go
// *** WARNING: this file was generated by the Pulumi Terraform Bridge (tfgen) Tool. ***

// nolint: lll
package aws

import (
	"errors"
	"io/ioutil"
	"os"
	"os/exec"
	"path/filepath"
	"runtime"

)

func init() {
	version, err := getVersion()
	if err != nil || version == "" {
		return
	}
	pulumi.RegisterPackage(pulumi.PackageInfo{Name: "aws", Version: version})
}

func getVersion() (string, error) {
	// TODO support windows plugin acquisition
	if runtime.GOOS == "windows" {
		return "", errors.New("automatic windows plugin acquisition is unsupported at this time")
	}

	// write out get version script to a tmp file, execute it to acquire version
	f, err := ioutil.TempFile("", "get-version")
	defer f.Close()
	if err != nil {
		return "", err
	}
	_, err = f.WriteString(getVersionBash)
	if err != nil {
		return "", err
	}

	// determine location of this file
	_, filename, _, ok := runtime.Caller(1)
	if !ok {
		return "", errors.New("cannot determine location of pulumiManifest.go")
	}
	execDir := filepath.Join(filename, "..")

	cmd := exec.Command("/bin/bash", f.Name(), "HEAD")
	cmd.Dir = execDir
	cmd.Env = os.Environ()

	stdout, err := cmd.Output()
	if err != nil {
		return "", err
	}
	return string(stdout), nil
}

const getVersionBash = `
#!/bin/bash
set -o nounset -o errexit -o pipefail

if [ $# -eq 0 ]; then
    echo "No arguments provided. Pass appropriate git commit-ish value get version of (i.e. 'HEAD')"
    exit 1
fi

COMMITISH=$1
DIRTY_TAG=""

# Figure out if the worktree is dirty, we run update-index first
# as we've seen cases in Travis where not doing so causes git to
# treat the worktree as dirty when it is not.
git update-index -q --refresh
if ! git diff-files --quiet; then
    DIRTY_TAG="dirty"
fi

# If we have an exact tag, just use it.
if git describe --tags --exact-match "${COMMITISH}" >/dev/null 2>&1; then
    echo -n "$(git describe --tags --exact-match "${COMMITISH}")"
    if [ ! -z "${DIRTY_TAG}" ]; then
        echo -n "+${DIRTY_TAG}"
    fi

    echo ""
    exit 0
fi

# Otherwise, increment the minor version version (if the package is 1.X or later) or the
# patch version (if the package is pre 1.0), add the -alpha tag and some
# commit metadata. If there's no existing tag, pretend a v0.0.0 was
# there so we'll produce v0.0.1-dev builds.
if git describe --tags --abbrev=0 "${COMMITISH}" > /dev/null 2>&1; then
    TAG=$(git describe --tags --abbrev=0 "${COMMITISH}")
else
    TAG="v0.0.0"
fi

# Strip off any pre-release tag we might have (e.g. from doing a -rc build)
TAG=${TAG%%-*}

MAJOR=$(cut -d. -f1 <<< "${TAG}")
MINOR=$(cut -d. -f2 <<< "${TAG}")
PATCH=$(cut -d. -f3 <<< "${TAG}")

if [ "${MAJOR}" = "v0" ]; then
    PATCH=$((${PATCH}+1))
else
    MINOR=$((${MINOR}+1))
    PATCH=0
fi

# if we're in a features/xxx branch and caller passed --embed-feature-branch then append "-xxx" to
# the version as well.
FEATURE_TAG=""
for arg in "$@"
do
    if [[ "$arg" == "--embed-feature-branch" ]]; then
        if [[ "${TRAVIS_BRANCH:-}" == features/* ]]; then
            FEATURE_TAG=$(echo "${TRAVIS_BRANCH}" | sed -e 's|^features/|-|g')
        fi
        if [[ "${TRAVIS_BRANCH:-}" == feature-* ]]; then
            FEATURE_TAG=$(echo "${TRAVIS_BRANCH}")
        fi
    fi
done

# We want to include some additional information. To the base tag we
# add a timestamp and commit hash. We use the timestamp of the commit
# itself, not the date it was authored (so it will change when someone
# rebases a PR into master, for example).
echo -n "${MAJOR}.${MINOR}.${PATCH}-alpha${FEATURE_TAG}.$(git show -s --format='%ct+g%h' ${COMMITISH})"
if [ ! -z "${DIRTY_TAG}" ]; then
    echo -n ".${DIRTY_TAG}"
fi

echo ""
`

```

Invoking this code path via a locally linked pulumi go program 
```sh
Evans-MBP:aws-go-s3-folder evanboyle$ PULUMI_PLUGINS=true go run main.go
{"plugins":[{"name":"aws","version":"v1.29.0-alpha.1585334195+g6cfba9da2\n"}]}
```


EvanBoyle:
Turns out `go mod download` doesn't pull the `.git` directory, so this approach won't work.

PR #4212, opened 2020-03-27
Add DisplayName to Package

tasiah:
CAUTION: Due to time constraints, I haven't thoroughly tested this. You would still need to go into each provider manually to add in the proper display name, but this PR at least provides the ability to extract the display name from the package.


praneetloke:
I will close out this PR. We ended up creating a lookup map in the resource docs generator itself for now. We can revisit adding something like this to the schema later, if need be.

Issue #4177, opened 2020-03-25
Resource docs generator bug with double-quotes in markdown


praneetloke:
Double-quotes in the markdown of a property comment should not be encoded. See https://github.com/pulumi/docs/pull/2641#issuecomment-600321461.

praneetloke:
For an example pf the bug, see `volumeType` in https://www.pulumi.com/docs/reference/pkg/aws/ec2/instance/#instanceebsblockdevice.

tasiah:
Fixed [accidentally](https://pulumi.slack.com/archives/CBUP505U7/p1585165244004600?thread_ts=1585165175.004500&cid=CBUP505U7) with [this commit](https://github.com/pulumi/pulumi/commit/bdefc557af755f90a700c295f6f5ccbba102ecd8).

PR #4214, opened 2020-03-28
Update the resource-docs templates

cnunciato:
This change includes modifications made for generating the revisions in https://github.com/pulumi/docs/pull/2746.


pkg/codegen/docs/gen.go
```diff
@@ -1011,11 +1015,13 @@ func (mod *modContext) genIndex(exports []string) string {
 	w := &bytes.Buffer{}
 
 	name := mod.mod
+	menu := false
 	if name == "" {
 		name = mod.pkg.Name
+		menu = true
```
praneetloke: It'd be nice to have a comment here, why this flag is only set when the package name is blank.

pkg/codegen/docs/templates/properties.tmpl
```diff
@@ -1,48 +1,22 @@
...
+<dl class="resources-properties">
+{{ range . }}
+    <dt class="property-{{- if .IsInput -}}{{- if .IsRequired -}}required{{- else -}}optional{{- end -}}{{- end -}}{{ if .DeprecationMessage }} property-deprecated{{- end -}}"
+            title="{{ if .IsInput -}}{{- if .IsRequired -}}Required{{- else -}}Optional{{- end -}}{{- end -}}{{- if .DeprecationMessage -}}, Deprecated{{- end -}}">
+        <span>{{- htmlSafe .DisplayName -}}</span>
+        <span class="property-indicator"></span>
+        <span class="property-type">{{- if eq .Type.Link "#" "" -}}{{- htmlSafe .Type.Name -}}{{- else -}}{{ template "linkify" .Type }}{{- end -}}</span>
+    </dt>
+    <dd>
+        {{- print "{{% md %}}" -}}{{- .Comment -}}{{- print "{{% /md %}}" -}}
```
praneetloke: I think this should be:
```suggestion
        {{- print "{{% md %}}" -}}{{- htmlSafe .Comment -}}{{- print "{{% /md %}}" -}}
```

...this is needed to fix this issue: https://github.com/pulumi/pulumi/issues/4177

praneetloke:
Sorry, but I just noticed two things in this PR (one of them I added a comment inline.), and the other is:

https://github.com/pulumi/pulumi/pull/4214/files#diff-0676a3018568af956dacf76b6e246e20R128-R138

I realize you didn't change anything there, but I think you might need to change it to use the new lang chooser component? The above link is showing the lines in the `resource.tmpl` that controls the rendering of the language-specific doc links for input and output types of a nested type.

PR #4217, opened 2020-03-28
fix: up with refresh with targets not limiting refresh

### LINKED ISSUE #4174: pulumi up with `--refresh` and `--target` does not refresh only the target
When running `pulumi up --refresh --target $urn`, everything is refreshed as the `--target` is not applied to the refresh.

nesl247:
Closes #4174


stack72:
Manually merging this so that you don't need to rebase!

stack72:
Merged manually

PR #4218, opened 2020-03-28
fix: logging out with filestate when url is deleted

### LINKED ISSUE #4054: Can't log-out of a local state folder if it's been deleted
If you do `pulumi login file://my_folder` and then delete that folder, it's impossible to use any of the CLI commands to clear it. 

```
c:\> pulumi logout
error: unable to open bucket file://deleted_folder
```

I had to manually edit `%USERPROFILE%/.pulumi/credentials.json` to clear the `"current": "file://deleted_folder"` setting to get anywhere.

Thanks

nesl247:
I'm not 100% sure if this is the best solution or not, but it makes sense to me. If the directory doesn't exist anymore, I don't see the reason the user couldn't be logged out. If it was due to a network issue, they should be able to login easily without any repercussions.

Fixes #4054

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@ CHANGELOG
 =========
 
 ## HEAD (unreleased)
-_(none)_
+- Fix logout with file backend when state is deleted
+  [#4218](https://github.com/pulumi/pulumi/pull/4218)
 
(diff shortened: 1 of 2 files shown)


stack72:
Manually merged

PR #4220, opened 2020-03-28
fix coverage scripts to work with submodules

EvanBoyle:


EvanBoyle:
@luke this just fixes the build, and restores the previous state which was to run coverage over pkg. (technically some of the code in pkg moved to sdk so this isn't exactly the same). 

Yes ideally we should run coverage over all of our go files, but this seems like more of a net-new feature given that this build is effectively a no-op. It just runs and generates a coverage report, but doesn't do anything with the information or fail if we don't maintain a certain threshold. 

I filed an issue so that we can decide if we care: https://github.com/pulumi/pulumi/issues/4221

For now, I'm going to merge this so we don't have a red cron build.

PR #4223, opened 2020-03-29
Adding support for Azure Storage backend.

davidobrien1985:
This change will add support for Azure Storage account backend for GitHub Actions.
Use by setting the following environment variables.

```
PULUMI_BACKEND: "azurestorage"
AZURE_STORAGE_KEY: ${{ secrets.AZURE_STORAGE_KEY }}
AZURE_STORAGE_ACCOUNT: ${{ secrets.AZURE_STORAGE_ACCOUNT }}
```

### DIFF
--- a/dist/actions/entrypoint.sh
+++ b/dist/actions/entrypoint.sh
@@ -45,12 +45,7 @@ if [ ! -z "$PULUMI_CI" ]; then
     # is missing, the caller of this script will need to pass `-s <stack-name>` to specify the stack explicitly.
     if [ ! -z "$BRANCH" ]; then
         if [ ! -z "$PULUMI_BACKEND" ]; then
-            if [ "$PULUMI_BACKEND" == "azurestorage"]; then
-                pulumi login -c azblob://$AZURE_STORAGE_ACCOUNT
-            else
-                echo -e "Please set the '$PULUMI_BACKEND' variable to 'azurestorage'."
-                exit 1
-            fi
+            pulumi login -c $PULUMI_BACKEND
         else
             echo "Using Pulumi Service backend."
         fi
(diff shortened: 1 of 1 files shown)


lukehoban:
I was thinking we would just allow a `PULUMI_BACKEND` and pass that literally to the login command. Then it could be used for any backend you need (Azure blob, S3, GCS, etc). The user could construct the appropriate url shape for a Pulumi backend. Is that possible (and perhaps even easier?).

davidobrien1985:
Right, yeah, that might make it easier for the user.

```
PULUMI_BACKEND: ${{ secrets.PULUMI_BACKEND }}
AZURE_STORAGE_KEY: ${{ secrets.AZURE_STORAGE_KEY }}
AZURE_STORAGE_ACCOUNT: ${{ secrets.AZURE_STORAGE_ACCOUNT }}
```

For `PULUMI_BACKEND` values follow https://www.pulumi.com/docs/intro/concepts/state/#self-managed-backend

davidobrien1985:
How is this released @lukehoban ? The CI build for this seems to be tied to the actual pulumi binary release. Can this be released independently?

lukehoban:
> How is this released @lukehoban ? The CI build for this seems to be tied to the actual pulumi binary release. Can this be released independently?

Currently it’s released with all versioned releases of the Pulumi CLI. We want versions to stay aligned between these two.

davidobrien1985:
@lukehoban fair enough. Do you expect me to update the changelog or will that be done by someone on your end?

praneetloke:
@davidobrien1985 I'll open a separate PR with your changes for this. That way, the PR build can run. Thank you for opening this PR!

praneetloke:
Closing this in favor of #4243.

Issue #4227, opened 2020-03-29
get-version scripts now returning invalid versions with new tags


EvanBoyle:
./scripts/get-version is now returning `sdk/v1.13.1` which fails semver.parseTolerant. These scripts need to be updated across all of our repos to ignore anything before the `/` in a tag.

EvanBoyle:
https://travis-ci.com/github/pulumi/pulumi-aws/builds/156375379

stack72:
This is going to become a P1 as we need to release a 2.0.0 beta tomorrow or Tuesday

PR #4229, opened 2020-03-29
update get-version to understand submodule tags

EvanBoyle:
Fixes https://github.com/pulumi/pulumi/issues/4227

### DIFF
--- a/scripts/get-version
+++ b/scripts/get-version
@@ -36,7 +36,7 @@ fi
 # e.g: submodule/submodule/.../v.X.X.X
 # sdk/v1.1.1 -> v1.1.1
 IFS='/'
-read -a tagsplit <<< "$TAG"
+read -a tagsplit <<< "${TAG}"
 
 if echo "${#tagsplit[@]}" > /dev/null 2>&1 -gt 1; 
 then
@@ -47,7 +47,7 @@ fi
 # after we have removed any submodule prefixes.
 if [ $EXACT -eq 1 ]
 then
-    echo -n "$TAG"
+    echo -n "${TAG}"
     if [ ! -z "${DIRTY_TAG}" ]; then
         echo -n "+${DIRTY_TAG}"
     fi
--- a/scripts/get-version
+++ b/scripts/get-version
@@ -38,6 +38,17 @@ else
     TAG="v0.0.0"
 fi
 
+# Check to see if the latest tag is for a submodule and transform to semver tolerant
+# e.g: submodule/submodule/.../v.X.X.X
+# sdk/v1.1.1 -> v1.1.1
+IFS='/'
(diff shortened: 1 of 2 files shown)


scripts/get-version
```diff
@@ -19,25 +20,42 @@ fi
...
+IFS='/'
+read -a tagsplit <<< "$TAG"
+
+if echo "${#tagsplit[@]}" > /dev/null 2>&1 -gt 1; 
+then
+    TAG=${tagsplit[${#tagsplit[@]}-1]}
+fi
+
+# if we found an exact match, return it at this point
+# after we have removed and submodule prefixes.
```
justinvp: Nit:

```suggestion
# If we found an exact match, return it at this point
# after we have removed any submodule prefixes.
```

scripts/get-version
```diff
@@ -19,25 +20,42 @@ fi
...
     TAG=$(git describe --tags --abbrev=0 "${COMMITISH}")
 else
     TAG="v0.0.0"
 fi
 
+# Check to see if the latest tag is for a submodule and transform to semver tolerant
+# e.g: submodule/submodule/.../v.X.X.X
+# sdk/v1.1.1 -> v1.1.1
+IFS='/'
+read -a tagsplit <<< "$TAG"
```
justinvp: Nit: I guess it's not strictly necessary, but throughout this script we always use braces. Should we just use them here for consistency?

```suggestion
read -a tagsplit <<< "${TAG}"
```

scripts/get-version
```diff
@@ -19,25 +20,42 @@ fi
...
+if echo "${#tagsplit[@]}" > /dev/null 2>&1 -gt 1; 
+then
+    TAG=${tagsplit[${#tagsplit[@]}-1]}
+fi
+
+# if we found an exact match, return it at this point
+# after we have removed and submodule prefixes.
+if [ $EXACT -eq 1 ]
+then
+    echo -n "$TAG"
```
justinvp: Nit: Similarly use braces here for consistency?

```suggestion
    echo -n "${TAG}"
```

EvanBoyle:
Working on powershell updates. Need to spin up a windows WM to test them.

EvanBoyle:
Tested the windows updates:

```
// before
PS C:\Users\Administrator\pulumi> .\scripts\get-version HEAD
pkg/v1.13.2-dev.1585432158+g6a945294
// after
PS C:\Users\Administrator\pulumi> .\scripts\get-version HEAD
v1.13.2-dev.1585522789+g4b8361d0
```

Also tried it with the script applied dirty on the tag commit:

```
PS C:\Users\Administrator\pulumi> .\scripts\get-version HEAD
v1.13.1+dirty
```

PR #4232, opened 2020-03-30
remove dep ensure message from pulumi new *go*

EvanBoyle:
We'll want this change out with the next release. We're planning on having all examples and templates up to date by that point.

### DIFF
--- a/pkg/cmd/new.go
+++ b/pkg/cmd/new.go
@@ -672,13 +672,6 @@ func printNextSteps(proj *workspace.Project, originalCwd, cwd string, generateOn
 
 		// Install dependencies within the virtualenv
 		commands = append(commands, "pip3 install -r requirements.txt")
-	} else if strings.EqualFold(proj.Runtime.Name(), "go") {
-		// If we're generating a Go project, we (currently) will be placing a `Gopkg.toml`, and the
-		// user will need to `dep ensure` before running an update.  So we will recommend this to
-		// them explicitly.
-		//
-		// TODO[pulumi/pulumi#3817] When we move to Go modules, we will want to update this.
-		commands = append(commands, "dep ensure")
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/cmd/new.go
```diff
@@ -672,13 +672,6 @@ func printNextSteps(proj *workspace.Project, originalCwd, cwd string, generateOn
 
 		// Install dependencies within the virtualenv
 		commands = append(commands, "pip3 install -r requirements.txt")
-	} else if strings.EqualFold(proj.Runtime.Name(), "go") {
-		// If we're generating a Go project, we (currently) will be placing a `Gopkg.toml`, and the
-		// user will need to `dep ensure` before running an update.  So we will recommend this to
-		// them explicitly.
-		//
-		// TODO[pulumi/pulumi#3817] When we move to Go modules, we will want to update this.
-		commands = append(commands, "dep ensure")
```
lukehoban: I actually *do* think it would be nice to automatically `go mod download` if there is a go.mod, just like we automatically `npm install`, just to shift that cost into the `Pulumi new`. 

But not a huge thing, and more important to remove the current text which will be wrong soon.

stack72: @lukehoban it does go mod download by default - @EvanBoyle tested this :)

EvanBoyle: If I understand correctly, I think Luke is suggesting that we automatically run `go mod download` on behalf of the user like we do with npm install. This way, our startup time feels snappier since the first run won't have the additional cost of populating the module cache.

lukehoban: I know - but this still applies:

> just to shift that cost into the `pulumi new`

EvanBoyle: I'm going to land the change as is and follow up on @lukehoban's suggestion at the end of the week https://github.com/pulumi/pulumi/issues/4236

PR #4237, opened 2020-03-30
Fix csharp doc links generated by resource docs generator

### LINKED ISSUE #4215: .NET API docs links for input and output types in the resource docs are incorrect
(The type name seems to be incorrect, as well as the namespace needs to have either `Inputs` or `Outputs` in the url.)

praneetloke:
Fixes #4215.

This PR fixes the resource docs generator to specify the appropriate input and output type qualifier when generating type strings.

It also fixes an issue with generating doc links for Pulumi types in nodejs. Those types are used in the constructor and function params. For example, `CustomResourceOptions`. (cc @cnunciato)

### DIFF
--- a/pkg/codegen/docs/gen.go
+++ b/pkg/codegen/docs/gen.go
@@ -260,19 +260,12 @@ func (mod *modContext) typeString(t schema.Type, lang string, characteristics pr
 	}
 
 	// Strip the namespace/module prefix for the type's display name.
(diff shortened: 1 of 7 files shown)


praneetloke:
PR build passed. I will merge go ahead and merge this PR. Please do leave any comments, if any, and I'd be happy to address them in a separate PR.

Issue #4136, opened 2020-03-20
Re-enable container entrypoint tests.


pgavlin:
These tests are consistently failing with the following in CI and blocking automated releases:
```
--- FAIL: TestPulumiActionsImage (51.31s)
    environment.go:83: Created new test environment:  /tmp/test-env259219729
    containers_test.go:98: Running command docker images pulumi/actions:latest --quiet
    --- FAIL: TestPulumiActionsImage/dotnet (26.74s)
        environment.go:83: Created new test environment:  /tmp/test-env858295100
        containers_test.go:133: Running command pulumi stack init moolumi/container-dotnet-15fd8d0359662ea2
        containers_test.go:134: Running command pulumi config set runtime dotnet
        containers_test.go:137: Running command docker run --env PULUMI_ACCESS_TOKEN=[secure] --volume /tmp/test-env858295100:/src --workdir /src --rm pulumi/actions:latest up --stack moolumi/container-dotnet-15fd8d0359662ea2
        containers_test.go:127: Running command pulumi stack rm --force --yes
        containers_test.go:128: 
            	Error Trace:	environment.go:103
            	            				containers_test.go:128
            	            				containers_test.go:155
            	            				containers_test.go:105
            	Error:      	Received unexpected error:
            	            	unlinkat /tmp/test-env858295100/bin/Debug/netcoreapp3.0/Grpc.Core.Api.dll: permission denied
            	Test:       	TestPulumiActionsImage/dotnet
            	Messages:   	cleaning up the test directory
    --- FAIL: TestPulumiActionsImage/nodejs (15.80s)
        environment.go:83: Created new test environment:  /tmp/test-env722228587
        containers_test.go:133: Running command pulumi stack init moolumi/container-nodejs-15fd8d0993050a7b
        containers_test.go:134: Running command pulumi config set runtime nodejs
        containers_test.go:137: Running command docker run --env PULUMI_ACCESS_TOKEN=[secure] --volume /tmp/test-env722228587:/src --workdir /src --rm pulumi/actions:latest up --stack moolumi/container-nodejs-15fd8d0993050a7b
        containers_test.go:127: Running command pulumi stack rm --force --yes
        containers_test.go:128: 
            	Error Trace:	environment.go:103
            	            				containers_test.go:128
            	            				containers_test.go:155
            	            				containers_test.go:109
            	Error:      	Received unexpected error:
[...]

PR #4240, opened 2020-03-30
Refactor build-and-publish-docker

chrsmith:
This PR does some refactoring to make it easier to address #4136. It refactors `build-and-publish-docker` so that testing and publishing can be performed independently. The file has been renamed to just `build-docker.sh` to reflect this, and now requires a "--publish" flag to actually perform the publish like before.

Previously we were running some of the container tests (but not `TestPulumiActionsImage`). And with this change we no longer run any tests when we invoke `build-docker.sh` from within `publish_packages.sh`.

Instead, we run `build-docker.sh --test` ~daily as part of the Travis CI cron jobs. We could do this on every commit to master, too. But I figured it would add more CI execution time than we'd like.

Note that the PR isn't quite ready yet, and has a couple of problems:

(1) The way I'm getting the "CLI version" to use when building the Docker containers isn't right. It's the command we use in `publish_packages.sh` (set [here](https://github.com/pulumi/pulumi/blob/master/scripts/publish_packages.sh#L8), used [here](https://github.com/pulumi/pulumi/blob/master/scripts/publish_packages.sh#L67)) but when ran on CI we get an error:

```
Building Pulumi containers
- Building pulumi/pulumi
invalid argument "pulumi/pulumi:v1.14.0-alpha.1585599244+gbbae83c5" for "-t, --tag" flag: invalid reference format
See 'docker build --help'.
```

`scripts/get-version HEAD` returns "v1.14.0-alpha.1585599244+gbbae83c5" when I think it needs to just be "v1.14.0-alpha.1585599244". @pgavlin do you have any idea why there is a discrepancy here? 

(2) There are some underlying test failures that need to be addressed from #4136, which appear to be related to deleting the files on disk while they are still referenced in a Docker image.


pgavlin:
> scripts/get-version HEAD returns "v1.14.0-alpha.1585599244+gbbae83c5" when I think it needs to just be "v1.14.0-alpha.1585599244". @pgavlin do you have any idea why there is a discrepancy here?

`v1.14.0-alpha.1585599244+gbbae83c51` is expected. The `+gbbae83c51` is the commit, and is ignored by semver 2.0.

chrsmith:
Closing this out so I can open a PR from the `features/test-container-build`, which will allow me to actually confirm that the tests are working in CI. (Since Travis CI `push` jobs are only ran on `master` or e.g. branches with `features` in the name.)

PR #4255, opened 2020-04-01
Only show the update version warning message in interactive mode

### LINKED ISSUE #4082: `pulumi stack output --json` has invalid json because of warnings
When i run `pulumi --non-interactive stack output --json` with a newer version of pulumi available i get
```
{
  "services": [ ... ]
}
warning: A new version of Pulumi is available. To upgrade from version '1.12.0' to '1.12.1', run 
   $ brew upgrade pulumi
or visit https://pulumi.com/docs/reference/install/ for manual instructions and release notes.
```

The warning is in stdout and invalidates the json and i cannot decode the output anymore
It would be cool to disable the warning when `--non-interactive` is on or put the warning in stderr.

leezen:
Fixes #4082

### DIFF
--- a/pkg/cmd/pulumi.go
+++ b/pkg/cmd/pulumi.go
@@ -131,7 +131,9 @@ func NewPulumiCmd() *cobra.Command {
 		}),
 		PersistentPostRun: func(cmd *cobra.Command, args []string) {
 			// Before exiting, if there is a new version of the CLI available, print it out.
-			if checkVersionMsg := <-updateCheckResult; checkVersionMsg != nil && cmdutil.Interactive() {
+			jsonFlag := cmd.Flag("json")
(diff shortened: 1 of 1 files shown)


leezen:
@justinvp Yes, that would be the case. I do agree it'd be ideal to have people using CI systems see the warning and upgrade as a result, but I think it's probably better to respect the `DisableInteractive` flag. We could have a different check for just `DisableInteractive` and interactive terminals.

lukehoban:
I had sort of assumed we would fix #4082 by suppressing this during `--json` output, instead of suppressing during non-interactive.  Is there a reason it's better to tie this to non-interactive?

leezen:
@lukehoban I think the way I was thinking about this was it seems odd to push out a warning in a non-interactive setting. Who is reading it? But the CI aspect makes sense and narrowing this to just `json` is fine. I'll update this PR.

PR #4256, opened 2020-04-01
Add package details and atttribution to schema

lukehoban:
Part of https://github.com/pulumi/home/issues/734.

### DIFF
--- a/pkg/codegen/go/gen.go
+++ b/pkg/codegen/go/gen.go
@@ -854,7 +854,6 @@ func (pkg *pkgContext) genHeader(w io.Writer, goImports []string, importedPackag
 		pkgName = path.Base(pkg.mod)
 	}
 
-	fmt.Fprintf(w, "// nolint: lll\n")
 	fmt.Fprintf(w, "package %s\n\n", pkgName)
 
 	var imports []string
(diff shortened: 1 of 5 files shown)


pkg/codegen/go/gen.go
```diff
@@ -854,7 +854,6 @@ func (pkg *pkgContext) genHeader(w io.Writer, goImports []string, importedPackag
 		pkgName = path.Base(pkg.mod)
 	}
 
-	fmt.Fprintf(w, "// nolint: lll\n")
```
lukehoban, line 857: This is mostly unrelated to other changes here - but part of overall cleaning up codegen.

pkg/codegen/schema/schema.go
```diff
@@ -431,6 +433,8 @@ type PackageSpec struct {
 	Homepage string `json:"homepage,omitempty"`
 	// License indicates which license is used for the package's contents.
 	License string `json:"license,omitempty"`
+	// Attribution allows freeform text attribution of derived work, if needed.
+	Attribution string `json:"attribution,omitempty"`
```
lukehoban, line 437: Given how we're using this - I might just rename this to `Notes` to be more generic.

praneetloke: To me `Notes` sounds specific to how we show this information. If the information in this property is really attribution, then I'd suggest leaving it as-is, otherwise, yeah `Notes` would be good here.

PR #4259, opened 2020-04-01
Backport some of the fixes regarding builds back to master

stack72:


scripts/get-version.ps1
```diff
@@ -2,7 +2,7 @@
 $ErrorActionPreference="Stop"
 
 git update-index -q --refresh
-git diff-files --quiet | Out-Null
+git diff-files -- . ':!**/go.mod' ':!**/go.sum' --quiet | Out-Null
```
EvanBoyle, line 5: Surprised these matching expressions don't need to change for windows. Would've expected something more like `':!**\\go.mod'`

stack72: 🤷‍♂ but it works - I was able to publish a windows binary just fine

Issue #4115, opened 2020-03-19
Add Python support to `pulumi policy new`


justinvp:
`pulumi policy new` currently always attempts to install `npm` dependencies. It needs to be fixed to be like `pulumi new`, to not use `npm` for non-Node.js templates and for Python to show instructions on how to install dependencies (create virtual env, install deps, activate env).

PR #4263, opened 2020-04-01
Add Python support to `pulumi policy new`

justinvp:
Instead of always running `npm install`, if the Policy Pack is Python, emit instructions on how to install dependencies. Also minor cleanup/refactoring.

## Node.js

Screen Shot 2020-04-01 at 10 57 51 AM

## Node.js w/ `-g`

Screen Shot 2020-04-01 at 10 58 23 AM

## Python

Screen Shot 2020-04-01 at 10 58 54 AM

Fixes https://github.com/pulumi/pulumi/issues/4115

### DIFF
--- a/pkg/cmd/new.go
+++ b/pkg/cmd/new.go
@@ -233,7 +233,7 @@ func runNew(args newArgs) error {
 	fmt.Println()
 
 	// Load the project, update the name & description, remove the template section, and save it.
-	proj, _, err := readProject()
+	proj, root, err := readProject()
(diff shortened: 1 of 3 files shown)


justinvp:
> The only thing of note is the the nodejs instructions tell you how to publish the pack while the python ones do not. Can you not publish a python pack?

Yes, by design. You cannot currently publish a Python pack. We deferred this to post GA. It is tracked here: https://github.com/pulumi/pulumi-policy/issues/211

Issue #4269, opened 2020-02-27
Generate Kuberenetes docs based on the schema-based docs generation process


praneetloke:
(Previous notes from Justin)
  * My understanding is we will be able to generate a Pulumi schema for the Kubernetes provider (https://github.com/pulumi/pulumi-kubernetes/issues/70), so we should be able to generate docs from that.
  * Will need to spend a little time enabling this and hooking it up into the doc gen scripts, etc.

----
- [x] Normalize the module names
- [x] Merge multiple versions of a module into a single one (parity with the current API docs)
- [x] Show the `provider` resource as the Provider resource like we do with TF-based providers
- [x] Show the resource-level `DeprecationMessage`
- [x] Lengthy module/resource names are truncated in the index pages
- [x] Don’t show the `api_version` and `kind` properties for the Python language as those are hard-coded internally in the Python SDK for every resource.
- [x] Fix broken `customresourcedefinition` page due to html-encoding issue

PR #4270, opened 2020-04-02
Update resource docs generator to handle some k8s cases

praneetloke:
Part of https://github.com/pulumi/pulumi/issues/4269.

This is still sort of a WIP PR. The docs do generate, but there are some minor issues with regards to how the Kubernetes resources are named and laid out, that need to be worked out to call the docs good-to-go. (See checklist in the related issue.)

This PR is generally to fix the way the generator was using the module name in each language. It now honors the language-specific info that is in the schema. This is a detail that is very important for k8s specifically as nested types can be referenced across "modules"; a set of core types if you will.

In addition to that, I also introduced logging using glog. Note that I looked at `sdk/go/common/util/logging.go` and will likely switch to using that in a follow-up PR. Let me know if you think I should do that in this PR instead.

The companion docs PR is https://github.com/pulumi/docs/pull/2820.


praneetloke:
Rebased with the latest from `master`.

pkg/codegen/docs/gen.go
```diff
@@ -193,26 +204,35 @@ func (ss nestedTypeUsageInfo) add(s string, input bool) {
...
+	if !ok {
+		return false
+	}
+
+	if input && a.Input {
+		return true
+	} else if !input && a.Output {
+		return true
+	}
+	return false
```
justinvp, line 226: Nit: This is fine, but it could be simplified...

Something like:

```suggestion
	if !ok {
		return false
	}
	return (input && a.Input) || (!input && a.Output)
```

or even just:

```suggestion
	return ok && ((input && a.Input) || (!input && a.Output))
```

pkg/go.mod
```diff
@@ -22,6 +22,7 @@ require (
 	github.com/djherbis/times v1.2.0
 	github.com/docker/docker v0.0.0-20170504205632-89658bed64c2
 	github.com/dustin/go-humanize v1.0.0
+	github.com/golang/glog v0.0.0-20160126235308-23def4e6c14b
```
justinvp, line 25: @EvanBoyle, are we switching away from glog to klog? I've lost track.

EvanBoyle: We're sticking with glog right now. The klog change has been reverted.

PR #4272, opened 2020-04-02
Enable features in mock monitor

### LINKED ISSUE #4145: Mock resource monitor does not support secrets
When writing tests using the new mock feature (https://github.com/pulumi/pulumi/pull/3738), we're unable to run a test suite if any of the resources are marked as secret outputs. Is this a feature that could be added to `setMocks`? We're using pulumi v1.12.1 and node v12.16.1.

Example:

```typescript
import * as assert from "assert";

pulumi.runtime.setMocks({
    newResource: function(resourceType, name, inputs) {
        return { id: "", state: {} };
    },
    call: function(token, args) {
        return {};
    },
});

    bucket: pulumi.secret("my-secret-bucket-name"),
});

describe("bucket", () => {
    it("should have a secret name", async () => {
        const name = await (bucket.bucket as any).promise();
        assert.equal(name, "my-secret-bucket-name");
    });
});
```

Output:

```
(node:83084) UnhandledPromiseRejectionWarning: TypeError: monitorRef.supportsFeature is not a function
    at example/node_modules/@pulumi/pulumi/runtime/settings.js:427:20
    at new Promise (<anonymous>)
    at Object.monitorSupportsSecrets (example/node_modules/@pulumi/pulumi/runtime/settings.js:426:12)
    at example/node_modules/@pulumi/pulumi/runtime/rpc.js:266:47
    at Generator.next (<anonymous>)
    at fulfilled (example/node_modules/@pulumi/pulumi/runtime/rpc.js:18:58)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
(node:83084) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 2)
(node:83084) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
  bucket
    1) should have a secret name
```

Thanks for taking a look at this!

mikhailshilkov:
Enables secret support in mock-based tests. Go and C# seem to be fine as-is.

Fixes #4145

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ CHANGELOG
 ## HEAD (unreleased)
 - Fix handling of `nil` values in Outputs in Go.
   [#4268](https://github.com/pulumi/pulumi/pull/4268)
+  
 - Include usage hints for Input types in Go SDK
   [#4279](https://github.com/pulumi/pulumi/pull/4279)
 
@@ -12,6 +13,9 @@ CHANGELOG
 
 - Fix the `call` mock in Python.
   [#4274](https://github.com/pulumi/pulumi/pull/4274)
+  
+- Fix handling of secret values in mock-based tests.
+  [#4272](https://github.com/pulumi/pulumi/pull/4272)
 
 ## 1.14.0 (2020-04-01)
 - Fix error related to side-by-side versions of `@pulumi/pulumi`.
--- a/sdk/nodejs/runtime/mocks.ts
+++ b/sdk/nodejs/runtime/mocks.ts
@@ -112,6 +112,12 @@ export class MockMonitor {
     public registerResourceOutputs(req: any, callback: (err: any, innerResponse: any) => void) {
         callback(null, {});
     }
+
+    public supportsFeature(req: any, callback: (err: any, innerResponse: any) => void) {
+        callback(null, {
+            getHassupport: () => true,
+        });
+    }
 }
 
 /**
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ CHANGELOG
 ## HEAD (unreleased)
 - Fix handling of `nil` values in Outputs in Go.
   [#4268](https://github.com/pulumi/pulumi/pull/4268)
+  
+- Fix handling of secret values in mock-based tests.
+  [#4272](https://github.com/pulumi/pulumi/pull/4272)
```
lukehoban, line 9: Do we have any tests that validate mocks working with secrets?

mikhailshilkov: I had the same question but it looks like we don't have any tests for mocks. Probably worth adding them for all languages / methods / parameter permutations, but sounds a bit out of the scope for this issue?

lukehoban: We did add tests for Python mocks recently - https://github.com/pulumi/pulumi/pull/4074. 

But happy to split out a new issue on getting test coverage broadly on mocking.

mikhailshilkov: Awesome! I missed that one. Now I augmented the Python test to use a secret. Also created #4276 to add tests for other runtimes.

PR #4273, opened 2020-04-02
Propagate secretness correctly in Python `apply`

### LINKED ISSUE #4163: Secretness is not propagating correctly in Python `apply`
In the following example:
1. The export is displayed in plain text
2. The export and the `hostedZoneId` output are stored in the state file in plain text.

However, the "Something" tag is correctly stored in cipher text in the state file.

```py
secret = pulumi.Output.secret("42")

bucket = s3.Bucket(
    'my-bucket',
    tags={ 'Something': secret}, 
    opts=pulumi.ResourceOptions(additional_secret_outputs=['hostedZoneId'])
)

pulumi.export('bucket_name', secret)
```

This appears to be due to `apply` not propagating secretness correctly, and was introduced in the changes in https://github.com/pulumi/pulumi/pull/3327/files#diff-1467e73b3e668641ccb747c246f32132.

Let's also review test coverage to make sure we have appropriate test coverage around secrets in the Python (and other language) SDKs.

justinvp:
Note: I didn't add any new tests for Go -- I'll need to look into that a bit more and see if it could use any additional coverage.

Fixes #4163

### DIFF
--- a/sdk/python/lib/test/test_next_serialize.py
+++ b/sdk/python/lib/test/test_next_serialize.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 import asyncio
 import unittest
+from typing import Any, Optional
 
 from google.protobuf import struct_pb2
 from pulumi.resource import CustomResource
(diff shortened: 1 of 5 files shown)


sdk/python/lib/test/test_next_serialize.py
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
```
justinvp, line 13: I added tests to this file as this is where some similar/related tests already were -- though, these new tests (and some of the existing ones) would prob make more sense in a new `test_output.py`.

Issue #4149, opened 2020-03-21
TestPolicy is causing unrelated test failures in master


justinvp:
https://travis-ci.com/pulumi/pulumi/builds/154362492

```
7196=== RUN   TestUntargetedCreateDuringTargetedUpdate
7197--- FAIL: TestUntargetedCreateDuringTargetedUpdate (9.67s)
7198    environment.go:83: Created new test environment:  /tmp/test-env075983751
7199    targets_test.go:34: Running command pulumi stack init test-f054f65d
7200    targets_test.go:35: Running command yarn link @pulumi/pulumi
7201    targets_test.go:36: Running command pulumi up --non-interactive --skip-preview --yes
7202    targets_test.go:36: Ran command pulumi args [up --non-interactive --skip-preview --yes] and expected success. Instead got failure.
7203    targets_test.go:36: Run Error: exit status 255
7204    targets_test.go:36: STDOUT: Updating (test-f054f65d):
7205         
7206        Permalink: https://app.pulumi.com/pulumi-bot/untargeted_create/test-f054f65d/updates/1
7207    targets_test.go:36: STDERR: error: policy pack "test-policy-pack-15fe2beeecf9fd40" failed to start: test-policy-pack-15fe2beeecf9fd40 (analyzer) plugin [/opt/pulumi/bin/pulumi-analyzer-policy] wrote a non-numeric port to stdout ('no policy name provided'): strconv.Atoi: parsing "no policy name provided": invalid syntax
7208    targets_test.go:37: Running command pulumi stack output urn
7209    targets_test.go:37: Ran command pulumi args [stack output urn] and expected success. Instead got failure.
7210    targets_test.go:37: Run Error: exit status 255
7211    targets_test.go:37: STDOUT: 
7212    targets_test.go:37: STDERR: error: current stack does not have output property 'urn'
7213    targets_test.go:46: Running command pulumi up --target  --non-interactive --skip-preview --yes
7214    targets_test.go:46: Ran command pulumi args [up --target  --non-interactive --skip-preview --yes] and expected success. Instead got failure.
7215    targets_test.go:46: Run Error: exit status 255
7216    targets_test.go:46: STDOUT: Updating (test-f054f65d):
7217         
7218        Permalink: https://app.pulumi.com/pulumi-bot/untargeted_create/test-f054f65d/updates/2
[...]

PR #4278, opened 2020-04-02
Skip policy tests that cause concurrent Travis jobs to fail

justinvp:


sean1588:
Will try and get around to #4149 soon as I can. Been working on other issues...

Issue #4071, opened 2020-03-12
Go SDK - Create standard doc string templates for T TArgs TInput TOutput that clarify usage


EvanBoyle:
Right now it can be very hard to determine what form of `T` to construct when using nested types. This cloud trail argument:

https://github.com/pulumi/pulumi-aws/blob/c29dcf1b7f3209874af908b4fa94c81991b5adde/sdk/go/aws/cloudtrail/trail.go#L155

Leads to this overwhelming type file with no clarity on what the proper form of the type should be:

https://github.com/pulumi/pulumi-aws/blob/c29dcf1b7f3209874af908b4fa94c81991b5adde/sdk/go/aws/cloudtrail/pulumiTypes.go

We should be able to come up with a standard doc string for each form of `T` `TArgs` `TInput` and `TOutput` that is specific about the intended usage, for instance: 

```go
// T may be used to specify detailed instructions about how to build a container.
// T is the raw representation of the type.

// TArgs is the pulumified version of T where are members are outputs. 
// Construct TArgs as an input to NewDockerBuild(...)
```

The work here is to:

1. Define the standard template for each form of T
2. Update our code generators to put these in in our SDK types for go (aws, azure, gcp). 

Came up in https://github.com/pulumi/pulumi-docker/pull/147#discussion_r391787251

PR #4279, opened 2020-04-02
Update go codegen to include usage hints on Input types

EvanBoyle:
Fixes https://github.com/pulumi/pulumi/issues/4071

This change adds some very basic usage comments on Input types. It gives users a hint on construction of concrete instances. The tests are illustrative, but here's an example in context for pulumi-aws: 

https://github.com/pulumi/pulumi-aws/compare/evan/goCodeGenUsage?expand=1

### DIFF
--- a/pkg/codegen/go/gen.go
+++ b/pkg/codegen/go/gen.go
@@ -311,11 +311,31 @@ func printComment(w io.Writer, comment string, indent bool) {
 }
 
 func genInputInterface(w io.Writer, name string) {
+	printComment(w, getInputUsage(name), false)
 	fmt.Fprintf(w, "type %sInput interface {\n", name)
 	fmt.Fprintf(w, "\tpulumi.Input\n\n")
 	fmt.Fprintf(w, "\tTo%sOutput() %sOutput\n", title(name), name)
 	fmt.Fprintf(w, "\tTo%sOutputWithContext(context.Context) %sOutput\n", title(name), name)
 	fmt.Fprintf(w, "}\n\n")
+}
+
+func getInputUsage(name string) string {
+	if strings.HasSuffix(name, "Array") {
+		baseTypeName := name[:strings.LastIndex(name, "Array")]
+		return fmt.Sprintf("Construct a concrete instance of %sInput via:\n\t%s{ %sArgs{...} }", name, name, baseTypeName)
+	}
+
+	if strings.HasSuffix(name, "Map") {
+		baseTypeName := name[:strings.LastIndex(name, "Map")]
+		return fmt.Sprintf("Construct a concrete instance of %sInput via:\n\t%s{ \"key\": %sArgs{...} }", name, name, baseTypeName)
+	}
+
+	if strings.HasSuffix(name, "Ptr") {
+		baseTypeName := name[:strings.LastIndex(name, "Ptr")]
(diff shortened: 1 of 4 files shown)


pkg/codegen/go/gen.go
```diff
@@ -311,13 +311,33 @@ func printComment(w io.Writer, comment string, indent bool) {
...
 func genInputInterface(w io.Writer, name string) {
+	printComment(w, getInputUsage(name), false)
 	fmt.Fprintf(w, "type %sInput interface {\n", name)
 	fmt.Fprintf(w, "\tpulumi.Input\n\n")
 	fmt.Fprintf(w, "\tTo%sOutput() %sOutput\n", title(name), name)
 	fmt.Fprintf(w, "\tTo%sOutputWithContext(context.Context) %sOutput\n", title(name), name)
 	fmt.Fprintf(w, "}\n\n")
 }
 
+func getInputUsage(name string) string {
```
lukehoban, line 322: Do we want some more descriptive text at the beginning of all of these ?

For the builtins we have a comment like:

```go
// StringPtrInput is an input type that accepts StringPtr and StringPtrOutput values.
```

Should we here have:

```go
// FooPtrInput is an input type that accepts FooArgs, FooPtr and FooPtrOutput values.
// You can construct a concrete instance of `FooPtrInput` with either:
//
//    FooArgs{...}
//
// or:
// 
//    nil
```

pkg/codegen/go/gen.go
```diff
@@ -311,13 +311,33 @@ func printComment(w io.Writer, comment string, indent bool) {
...
+	}
+
+	if strings.HasSuffix(name, "Map") {
+		baseTypeName := name[:strings.LastIndex(name, "Map")]
+		return fmt.Sprintf("Construct a concrete instance of %sInput via:\n\t%s{ \"key\": %sArgs{...} }", name, name, baseTypeName)
+	}
+
+	if strings.HasSuffix(name, "Ptr") {
+		baseTypeName := name[:strings.LastIndex(name, "Ptr")]
+		return fmt.Sprintf("Construct a concrete instance of %sInput via:\n\t%sArgs{...}.To%sOutput()", name, baseTypeName, name)
```
lukehoban, line 335: This can just be `FooArgs{..}`, right?

We always codegen an implementation of the `PtrInput` interface (as well as the plain `Input` interface) for the `Args` struct I believe?

If we wanted to be uniform, I think it would be `FooPtr(&FooArgs{...})` or `FooPtr(nil)`.  But I think we want to promote just passing the plain Args struct (or nil) here.

pkg/codegen/go/gen.go
pgavlin: Note that we'll also need to begin these comments with `FooPtrInput` or we won't be lint-clean.

EvanBoyle: Good call, updated

pkg/codegen/go/gen.go
EvanBoyle: 👍

EvanBoyle:
Regenerated pulumi-aws with the suggested improvements. Looks great!

https://github.com/pulumi/pulumi-aws/compare/evan/goCodeGenUsage?expand=1

PR #4294, opened 2020-04-03
fix: remove --show-* as persistent flags

### LINKED ISSUE #4290: `--show-urns` is in Global Flags for `pulumi stack export`
When running `pulumi stack export --help`, `--show-urns` shows up in the `Global Flags` section. This is specific to `stack`.

This also applies to `--show-ids`, `--show-secrets`.

nesl247:
Fixes #4290

### DIFF
--- a/pkg/cmd/stack.go
+++ b/pkg/cmd/stack.go
@@ -154,9 +154,9 @@ func newStackCmd() *cobra.Command {
 	cmd.PersistentFlags().StringVarP(
 		&stackName, "stack", "s", "",
 		"The name of the stack to operate on. Defaults to the current stack")
-	cmd.PersistentFlags().BoolVarP(
+	cmd.Flags().BoolVarP(
 		&showIDs, "show-ids", "i", false, "Display each resource's provider-assigned unique ID")
-	cmd.PersistentFlags().BoolVarP(
+	cmd.Flags().BoolVarP(
 		&showURNs, "show-urns", "u", false, "Display each resource's Pulumi-assigned globally unique URN")
 	cmd.PersistentFlags().BoolVar(
 		&showSecrets, "show-secrets", false, "Display stack outputs which are marked as secret in plaintext")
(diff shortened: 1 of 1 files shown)


pkg/cmd/stack.go
```diff
@@ -154,9 +154,9 @@ func newStackCmd() *cobra.Command {
 	cmd.PersistentFlags().StringVarP(
 		&stackName, "stack", "s", "",
 		"The name of the stack to operate on. Defaults to the current stack")
-	cmd.PersistentFlags().BoolVarP(
+	cmd.Flags().BoolVarP(
 		&showIDs, "show-ids", "i", false, "Display each resource's provider-assigned unique ID")
-	cmd.PersistentFlags().BoolVarP(
+	cmd.Flags().BoolVarP(
 		&showURNs, "show-urns", "u", false, "Display each resource's Pulumi-assigned globally unique URN")
 	cmd.PersistentFlags().BoolVar(
```
justinvp, line 161: What about `show-secrets`?

```suggestion
	cmd.Flags().BoolVar(
```

nesl247: I didn't change that as I wasn't sure if it affected `stack output` or not.

nesl247: I actually went back and looked at the reason for this. `--show-ids` and `--show-urns` are flags for `pulumi stack` itself, which `--show-secrets` I believe to affect `export` and such.

justinvp: I looked into this in more detail.

There is a `pulumi stack output` command which also accepts `--show-secrets`, but it redefines this flag for itself.

There's no other commands under `stack` that refers to its `showSecrets` bool variable or is otherwise looking for "show-secrets", so we should be able to change it here from `PersistentFlags` to `Flags`.

stack72:
@nesl247 I am going to merge this manually since the tests don't run on a community PR (which we will be fixing soon!!!)

Thanks for the work here - it's very much appreciated

Issue #4211, opened 2020-03-27
Improve go plugin acquisition


EvanBoyle:
Currently, our go plugin acquisition code is a little bit of a hack. We don't support windows, and we're shelling out and executing a shell script written into a temporary file. This probably isn't performant, and rather than shelling out to a script, we should implement the functionality of that script natively in go. This would unblock windows support. See https://github.com/pulumi/pulumi/pull/4210 for more details on the approach and current implementation.

EvanBoyle:
#4210 was closed due to the fact that go mod download doesn't retrieve git info. 

We think there's a path forward using `go list -deps --json` similar to the approach we use for dotnet.

Issue #1549, opened 2018-06-20
Pulumi Go programs have no natural way of acquiring plugins


swgillespie:
We rely on package managers to do this on install for other languages, but naturally Go can't rely on this. We should either come up with something here or (at minimum) update templates and documentation to ensure that 1) people know how to download plugins if they are missing and 2) people know to reach for `pulumi plugin install` when using a new provider with Go.

swgillespie:
Go support is out of scope for M21.

lukehoban:
@pgavlin @EvanBoyle Do we have any thoughts on what we'll do here?  This seems to be a fairly deep remaining question mark for Go support.

EvanBoyle:
Let's make sure we update the getting started docs to remove plugin installation once we ship this. AWS example here, but we need to do this for each cloud's getting started guide:

https://github.com/pulumi/docs/pull/2480#discussion_r381648889

EvanBoyle:
closed via https://github.com/pulumi/pulumi/pull/4060

PR #4297, opened 2020-04-03
Reimplement getRequiredPlugins for go sdk

EvanBoyle:
This uses `go list -m -json all`. We must use -m `module` mode as this is the only way to get the version info as far as I can tell. This means that this approach only works for users using modules, not glide, dep, etc. This is the direction the community is going, and the direction we've moved our docs/examples so this should be fine. But as a result, I've made the implementation lenient. It doesn't fail if we run into errors along the way, just logs some messages saying we failed to acquire plugins. This is strictly better than what we have today, and we can always revisit this decision in the future as the ecosystem evolves. 

```
Evans-MBP:aws-go-webserver evanboyle$ pulumi preview
Previewing update (dcgo):
[resource plugin aws-1.28.0] installing
Downloading plugin: 64.58 MiB / 64.58 MiB [=========================] 100.00% 7s
Moving plugin... done.
     Type                      Name               Plan       
 +   pulumi:pulumi:Stack       go-webserver-dcgo  create     
 +   ├─ aws:ec2:SecurityGroup  web-secgrp         create     
 +   └─ aws:ec2:Instance       web-server-www     create     
 
Resources:
    + 3 to create

Permalink: https://app.pulumi.com/EvanBoyle/go-webserver/dcgo/previews/a761b512-4136-4caa-b504-6843e72c0193

```

Fixes https://github.com/pulumi/pulumi/issues/1549 
Fixes https://github.com/pulumi/pulumi/issues/4211

### DIFF
--- a/sdk/go/common/util/buildutil/semver_test.go
+++ b/sdk/go/common/util/buildutil/semver_test.go
@@ -36,3 +36,14 @@ func TestVersions(t *testing.T) {
 		assert.Equal(t, expected, p, "failed parsing '%s'", ver)
 	}
 }
+
+func TestPseduoVersion(t *testing.T) {
+	pseudoVersion := "v1.29.1-0.20200403140640-efb5e2a48a86"
+	assert.True(t, IsPseudoVersion(pseudoVersion))
+
+	tagVersion := "v1.29.0"
+	assert.False(t, IsPseudoVersion(tagVersion))
+
+	betaVersion := "v.1.29.0-beta.1"
+	assert.False(t, IsPseudoVersion(betaVersion))
+}
--- a/sdk/go/common/util/buildutil/semver_test.go
+++ b/sdk/go/common/util/buildutil/semver_test.go
@@ -44,6 +44,6 @@ func TestPseduoVersion(t *testing.T) {
 	tagVersion := "v1.29.0"
 	assert.False(t, IsPseudoVersion(tagVersion))
 
-	betaVersion := "v.1.29.0-beta.1"
+	betaVersion := "v1.29.0-beta.1"
 	assert.False(t, IsPseudoVersion(betaVersion))
 }
--- a/sdk/go/pulumi-language-go/main.go
+++ b/sdk/go/pulumi-language-go/main.go
@@ -33,6 +33,7 @@ import (
 	"github.com/pkg/errors"
 	"google.golang.org/grpc"
 
+	"github.com/pulumi/pulumi/sdk/go/common/util/buildutil"
 	"github.com/pulumi/pulumi/sdk/go/common/util/cmdutil"
 	"github.com/pulumi/pulumi/sdk/go/common/util/logging"
 	"github.com/pulumi/pulumi/sdk/go/common/util/rpcutil"
@@ -181,17 +182,30 @@ func (m *modInfo) getPlugin() (*pulumirpc.PluginDependency, error) {
 	if err != nil {
 		return nil, errors.New("module does not have semver compatible version")
 	}
-	// downgrade to zero for the patch
-	// pinning to a specific commit can result in a minor version that doesn't exist
-	// v1.29.1-0.20200403140640-efb5e2a48a86 (first commit after 1.29.0 release)
-	version := fmt.Sprintf("v%v.%v.%v", v.Major, v.Minor, 0)
+	version := m.Version
+
+	// psuedoversions are commits that don't have a corresponding tag at the specified git hash
+	// https://golang.org/cmd/go/#hdr-Pseudo_versions
+	// pulumi-aws v1.29.1-0.20200403140640-efb5e2a48a86 (first commit after 1.29.0 release)
+	if buildutil.IsPseudoVersion(version) {
+		// no prior tag means there was never a release build
+		if v.Major == 0 && v.Minor == 0 && v.Patch == 0 {
+			return nil, errors.New("invalid pseduoversion with no prior tag")
+		}
+		// patch is typically bumped from the previous tag when using pseudo version
+		// downgrade the patch by 1 to make sure we match a release that exists
+		patch := v.Patch
+		if patch > 0 {
+			patch--
+		}
+		version = fmt.Sprintf("v%v.%v.%v", v.Major, v.Minor, patch)
+	}
 
 	plugin := &pulumirpc.PluginDependency{
 		Name:    name,
 		Version: version,
 		Kind:    "resource",
 	}
-	logging.V(5).Infof("GetRequiredPlugins: Found plugin name: %s, version: %s", plugin.Name, plugin.Version)
 
 	return plugin, nil
 }
@@ -238,6 +252,13 @@ func (host *goLanguageHost) GetRequiredPlugins(ctx context.Context,
 		if err == nil {
 			logging.V(5).Infof("GetRequiredPlugins: Found plugin name: %s, version: %s", plugin.Name, plugin.Version)
 			plugins = append(plugins, plugin)
+		} else {
+			logging.V(5).Infof(
+				"GetRequiredPlugins: Ignoring dependency: %s, version: %s, error: %s",
+				m.Path,
+				m.Version,
+				err.Error(),
+			)
 		}
 	}
 
--- a/sdk/go/pulumi-language-go/main.go
+++ b/sdk/go/pulumi-language-go/main.go
@@ -15,14 +15,17 @@
 package main
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"flag"
 	"fmt"
+	"io"
 	"os"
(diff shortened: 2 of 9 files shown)


sdk/go/pulumi-language-go/main.go
```diff
@@ -157,10 +160,71 @@ func newLanguageHost(engineAddress, tracing string) pulumirpc.LanguageRuntimeSer
...
+// with respect to plugin acquisition
+type modInfo struct {
+	Path    string
+	Version string
+}
+
+func getPluginName(modName string) string {
+	// github.com/pulumi/pulumi-aws/sdk/... => aws
+	pluginPart := strings.Split(modName, "/")[2]
+	return strings.SplitN(pluginPart, "-", 2)[1]
```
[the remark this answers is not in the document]
lukehoban: Oh - I see that the caller ensures they can't.

sdk/go/pulumi-language-go/main.go
```diff
@@ -157,10 +160,71 @@ func newLanguageHost(engineAddress, tracing string) pulumirpc.LanguageRuntimeSer
...
+}
+
 // GetRequiredPlugins computes the complete set of anticipated plugins required by a program.
+// We're lenient here as this relies on the `go list` command and the use of modules.
+// If the consumer insists on using some other form of dependency management tool like
+// dep or glide, the list command fails with "go list -m: not using modules"
 func (host *goLanguageHost) GetRequiredPlugins(ctx context.Context,
```
lukehoban, line 180: Can we add any test coverage of this?

sdk/go/pulumi-language-go/main.go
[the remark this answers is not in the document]
EvanBoyle: Correct. I refactored this to make the relationship clear in case we change this code in the future.

sdk/go/pulumi-language-go/main.go
```diff
@@ -157,10 +160,71 @@ func newLanguageHost(engineAddress, tracing string) pulumirpc.LanguageRuntimeSer
...
+				break
+			}
+			logging.V(5).Infof("GetRequiredPlugins: Error parsing list output: %s", err.Error())
+			return &pulumirpc.GetRequiredPluginsResponse{}, nil
+		}
+
+		if strings.HasPrefix(m.Path, "github.com/pulumi/pulumi-") {
+			plugin := &pulumirpc.PluginDependency{
+				Name:    getPluginName(m.Path),
+				Version: m.Version,
```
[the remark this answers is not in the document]
EvanBoyle: Good point. I updated this to handle non-tag commits (we normalize the patch to `0` to be compatible).  The only way we can get something like `v0.0.0` is when the module is new and has no associated tags, which should never happen now that we've upgraded to the new module structure. I suppose we could filter out 0.0.0 if we want to be extra careful. I'm open to suggestions here.

sdk/go/pulumi-language-go/main.go
EvanBoyle: I refactored to test the code to test the filtering and conversion of the module result from `go list` into a plugin description. 

This seems sufficient to me. Any further validation would require an integration test with a dependency on pulumi-aws.

sdk/go/pulumi-language-go/main.go
```diff
@@ -157,10 +160,71 @@ func newLanguageHost(engineAddress, tracing string) pulumirpc.LanguageRuntimeSer
...
+
+	logging.V(5).Infof("GetRequiredPlugins: Determining pulumi packages")
+
+	gobin, err := findExecutable("go")
+	if err != nil {
+		return nil, errors.Wrap(err, "couldn't find go binary")
+	}
+
+	// don't wire up stderr so non-module users don't see error output from list
+	cmd := exec.Command(gobin, "list", "-m", "-json", "all")
```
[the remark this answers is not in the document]
EvanBoyle: Yup, tried this initially and learned about the internal restriction.

sdk/go/pulumi-language-go/main.go
```diff
@@ -157,10 +161,89 @@ func newLanguageHost(engineAddress, tracing string) pulumirpc.LanguageRuntimeSer
...
+	// pinning to a specific commit can result in a minor version that doesn't exist
+	// v1.29.1-0.20200403140640-efb5e2a48a86 (first commit after 1.29.0 release)
+	version := fmt.Sprintf("v%v.%v.%v", v.Major, v.Minor, 0)
+
+	plugin := &pulumirpc.PluginDependency{
+		Name:    name,
+		Version: version,
+		Kind:    "resource",
+	}
+	logging.V(5).Infof("GetRequiredPlugins: Found plugin name: %s, version: %s", plugin.Name, plugin.Version)
```
lukehoban, line 194: This is duplicated and can be removed here.

sdk/go/pulumi-language-go/main.go
```diff
@@ -157,10 +161,89 @@ func newLanguageHost(engineAddress, tracing string) pulumirpc.LanguageRuntimeSer
...
+		if err := dec.Decode(&m); err != nil {
+			if err == io.EOF {
+				break
+			}
+			logging.V(5).Infof("GetRequiredPlugins: Error parsing list output: %s", err.Error())
+			return &pulumirpc.GetRequiredPluginsResponse{}, nil
+		}
+
+		plugin, err := m.getPlugin()
+		if err == nil {
```
lukehoban, line 238: Should we log in the err case?

sdk/go/pulumi-language-go/main.go
```diff
@@ -157,10 +161,89 @@ func newLanguageHost(engineAddress, tracing string) pulumirpc.LanguageRuntimeSer
...
+	name := strings.SplitN(pluginPart, "-", 2)[1]
+
+	v, err := semver.ParseTolerant(m.Version)
+	if err != nil {
+		return nil, errors.New("module does not have semver compatible version")
+	}
+	// downgrade to zero for the patch
+	// pinning to a specific commit can result in a minor version that doesn't exist
+	// v1.29.1-0.20200403140640-efb5e2a48a86 (first commit after 1.29.0 release)
+	version := fmt.Sprintf("v%v.%v.%v", v.Major, v.Minor, 0)
```
lukehoban, line 187: In the case that a precise version was specified - this seems wrong. 

Also in general - I’m not positive this won’t work - for Kubernetes for example I think dev builds are prerelease for major version bumps?  And is dependent on some very specific things about how we currently version dev tags.  Anything more reliable we could do here?  

Perhaps no and this is still the best we can do heuristically and is worth it?

EvanBoyle: Good point. I'll lift some of the pseduo version parsing code out of internal to make this more robust.

EvanBoyle:
@lukehoban Please take another look. I took some code out of internal that detects pseudo versions so we can special case their handling. This allows us to leave patch versions alone in the normal case where our version matches a tag. This also allows us to handle beta/dev tags. I added more unit test cases to document the behavior of more of these corner cases. 

I manually tested the `v2.0.0-beta.1` release of AWS:

```
Evans-MBP:aws-go-webserver evanboyle$ pulumi preview
Previewing update (dcgo):
[resource plugin aws-2.0.0-beta.1] installing
Downloading plugin: 64.56 MiB / 64.56 MiB [=========================] 100.00% 3s
Moving plugin... done.
     Type                      Name               Plan       Info
 +   pulumi:pulumi:Stack       go-webserver-dcgo  create     2 messages
 +   ├─ aws:ec2:SecurityGroup  web-secgrp         create     
 +   └─ aws:ec2:Instance       web-server-www     create     
 
Diagnostics:
  pulumi:pulumi:Stack (go-webserver-dcgo):
    go: downloading github.com/pulumi/pulumi-aws/sdk/v2 v2.0.0-beta.1
    go: extracting github.com/pulumi/pulumi-aws/sdk/v2 v2.0.0-beta.1
 

Permalink: https://app.pulumi.com/EvanBoyle/go-webserver/dcgo/previews/47722fda-8bc9-48d6-8776-ba2c3f4e7558
```

sdk/go/common/util/buildutil/semver_test.go
```diff
@@ -36,3 +36,14 @@ func TestVersions(t *testing.T) {
...
 }
+
+func TestPseduoVersion(t *testing.T) {
+	pseudoVersion := "v1.29.1-0.20200403140640-efb5e2a48a86"
+	assert.True(t, IsPseudoVersion(pseudoVersion))
+
+	tagVersion := "v1.29.0"
+	assert.False(t, IsPseudoVersion(tagVersion))
+
+	betaVersion := "v.1.29.0-beta.1"
```
lukehoban, line 47:
```suggestion
	betaVersion := "v1.29.0-beta.1"
```

?

PR #4301, opened 2020-04-05
Use `function` css class in doc gen rather than `datasource`

justinvp:
Depends on https://github.com/pulumi/docs/pull/2848


justinvp:
> I have this change in a PR I am working on. I have actually moved all of this to templates as well rather than using print statements in Go.

Ah, ok. 👍 I'll close this PR.

PR #4308, opened 2020-04-06
Move resource docs index file generation to Go templates

praneetloke:
Part of #4269.

...and also the following issues that are specific to generating docs for the Kubernetes provider:

* Normalize the module names, i.e. stripe `.k8s.io` (and other domain names) from the module names in the index page.
* Merge multiple versions of a module into a single one (parity with the current API docs)
* Show the provider resource as the Provider resource like we do with TF-based providers
* Show the resource-level `DeprecationMessage`
* Don’t show the `api_version` and `kind` properties for the Python language as those are hard-coded internally in the Python SDK for every resource.


praneetloke:
Levi just made me aware of https://github.com/pulumi/pulumi/pull/4283. I am going to look at removing the code (introduced in this PR) that was manually excluding `apiVersion` and `kind` if the package was kubernetes to use the new schema property.

praneetloke:
Unfortunately, https://github.com/pulumi/pulumi/pull/4283 hasn't been consumed in the `pulumi-kubernetes` repo yet (and that is the repo where it matters to be used), so I'll use that property in a separate PR when it is ready and merge this one as-is for now.

PR #4309, opened 2020-04-06
HCL2 updates

pgavlin:
Pulumi HCL2 IR:
- Add support for invokes
- Add support for resource options, incl. ranged resources
- Allow the apply rewriter to ignore promise-typed values
- Add tests for the binder
- Add support functions for TF: entries and range

NodeJS codegen:
- Simplify for expression codegen
- Add support for invoke codegen
- Add support for entries and range functions
- Add tests

Python codegen:
- Implement codegen for most expression types
- Add support for invoke codegen
- Add tests

### DIFF
--- a/pkg/codegen/internal/test/testdata/aws-fargate.pp
+++ b/pkg/codegen/internal/test/testdata/aws-fargate.pp
@@ -1,0 +1,108 @@
+// Read the default VPC and public subnets, which we will use.
+vpc = invoke("aws:ec2:getVpc", {
+	default = true
+})
+subnets = invoke("aws:ec2:getSubnetIds", {
+	vpcId = vpc.id
+})
+
+// Create a security group that permits HTTP ingress and unrestricted egress.
+resource webSecurityGroup "aws:ec2:SecurityGroup" {
+	vpcId = vpc.id
+	egress = [{
+		protocol = "-1"
+		fromPort = 0
+		toPort = 0
+		cidrBlocks = ["0.0.0.0/0"]
+	}]
+	ingress = [{
+		protocol = "tcp"
+		fromPort = 80
+		toPort = 80
+		cidrBlocks = ["0.0.0.0/0"]
+	}]
+}
+
+// Create an ECS cluster to run a container-based service.
+resource cluster "aws:ecs:Cluster" {}
+
+// Create an IAM role that can be used by our service's task.
+resource taskExecRole "aws:iam:Role" {
+	assumeRolePolicy = {
+		Version = "2008-10-17"
+		Statement = [{
+			Sid = ""
+			Effect = "Allow"
+			Principal = {
+				Service = "ecs-tasks.amazonaws.com"
+			}
+			Action = "sts:AssumeRole"
+		}]
+	}
+}
+resource taskExecRolePolicyAttachment "aws:iam:RolePolicyAttachment" {
+	role = taskExecRole.name
+	policyArn = "arn:aws:iam::aws:policy/service-role/AmazonECSTaskExecutionRolePolicy"
+}
+
+// Create a load balancer to listen for HTTP traffic on port 80.
+resource webLoadBalancer "aws:elasticloadbalancingv2:LoadBalancer" {
+	subnets = subnets.ids
+	securityGroups = [webSecurityGroup.id]
+}
+resource webTargetGroup "aws:elasticloadbalancingv2:TargetGroup" {
+	port = 80
+	protocol = "HTTP"
+	targetType = "ip"
+	vpcId = vpc.id
+}
+resource webListener "aws:elasticloadbalancingv2:Listener" {
+	loadBalancerArn = webLoadBalancer.arn
+	port = 80
+	defaultActions = [{
+		type = "forward"
+		targetGroupArn = webTargetGroup.arn
+	}]
+}
+
+// Spin up a load balanced service running NGINX
+resource appTask "aws:ecs:TaskDefinition" {
+	family = "fargate-task-definition"
+	cpu = "256"
+	memory = "512"
+	networkMode = "awsvpc"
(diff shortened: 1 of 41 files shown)


pgavlin:
The test data may be most interesting: https://github.com/pulumi/pulumi/pull/4309/files#diff-9c8bce59d5a829a62f68be2d5ac7a684

pkg/codegen/internal/test/testdata/aws-fargate.pp.ts
```diff
@@ -0,0 +1,97 @@
...
+        Action: "sts:AssumeRole",
+    }],
+}});
+const taskExecRolePolicyAttachment = new aws.iam.RolePolicyAttachment("taskExecRolePolicyAttachment", {
+    role: taskExecRole.name,
+    policyArn: "arn:aws:iam::aws:policy/service-role/AmazonECSTaskExecutionRolePolicy",
+});
+// Create a load balancer to listen for HTTP traffic on port 80.
+const webLoadBalancer = new aws.elasticloadbalancingv2.LoadBalancer("webLoadBalancer", {
+    subnets: subnets.ids,
```
lukehoban, line 46: This doesn’t work does it?

pkg/codegen/internal/test/testdata/aws-s3-folder.pp.ts
```diff
@@ -0,0 +1,35 @@
...
+import * as aws from "@pulumi/aws";
+
+// Create a bucket and expose a website index document
+const siteBucket = new aws.s3.Bucket("siteBucket", {website: {
+    indexDocument: "index.html",
+}});
+const siteDir = "www";
+// For each file in the directory, create an S3 object stored in `siteBucket`
+const files: aws.s3.BucketObject[];
+for (const [__key, __value] of (() => throw new Error("FunctionCallExpression: readDir (aws-s3-folder.pp:12,11-27)"))()) {
```
lukehoban, line 11: I guess this is a TODO?

pgavlin: Yep

pkg/codegen/internal/test/testdata/aws-fargate.pp.ts
pgavlin: Hm, it does not. I'll have to file a bug :)

pkg/codegen/internal/test/testdata/aws-fargate.pp
```diff
@@ -0,0 +1,108 @@
...
+		cidrBlocks = ["0.0.0.0/0"]
+	}]
+}
+
+// Create an ECS cluster to run a container-based service.
+resource cluster "aws:ecs:Cluster" {}
+
+// Create an IAM role that can be used by our service's task.
+resource taskExecRole "aws:iam:Role" {
+	assumeRolePolicy = {
```
mikhailshilkov, line 31: Is `toJSON` call missing here or is it not required?

pgavlin: It's not required. The `assumeRolePolicy` property accepts arbitrary objects and serializes them transparently.

lukehoban: @pgavlin That is true in Node.js, but we don't support that yet in other languages do we?  Do we handle code generation correctly for this in other languages?

pgavlin: We don't generate code for this: it's implemented in the provider itself.

mikhailshilkov: It's just `string` in the schema, so I wonder how C# should handle this situation. Default to `toJSON` when anything is trying to assign to a string? Anyway, one more use case for type annotations, it seems.

pgavlin: > It's just string in the schema

The input property has this type:

```
                "assumeRolePolicy": {
                    "oneOf": [
                        {
                            "type": "string"
                        },
                        {
                            "type": "string",
                            "$ref": "#/types/aws:iam/documents:PolicyDocument"
                        }
                    ],
                    "description": "The policy that grants an entity permission to assume the role.\n"
                },
```

lukehoban: We don't codegen these union types of `string | JSON-object` in anything except Node.js - right?

mikhailshilkov: Ah, it's actually `string | PolicyDocument` in the schema, but C# doesn't support oneOf I guess.

pgavlin: > We don't codegen these union types of string | JSON-object in anything except Node.js - right?

Yeah, I was typing that out. The lang-specific code generators will need to recognize that.

lukehoban: (we don't even have anyway to codegen JSON object types in any othr language - https://github.com/pulumi/pulumi-terraform-bridge/issues/132)

lukehoban: I *think* supporting this is literally just pulumi/pulumi-terraform-bridge#132.

Technically we might want to avoid code-generating example code that won't work in other languages till that is fixed.  But since this is used so rarely - might be better to just prioritize fixing 132.

pgavlin: > (we don't even have anyway to codegen JSON object types in any othr language - pulumi/pulumi-terraform-bridge#132)

The Pulumi schema and associated code generators can do this. We don't have support for adding additional object types to the Pulumi schemas generated from TF schemas.

PR #4316, opened 2020-04-07
Define merge behavior for go resource options

EvanBoyle:
Fixes https://github.com/pulumi/pulumi/issues/3846

I based these changes on the typescript behavior defined here, and follow the same set of principles: https://github.com/pulumi/pulumi/blob/9abcca345a6c024087b35152b0d6fca92baef191/sdk/nodejs/resource.ts#L865

To summarize:
1. always append/merge anything that is/can be a collection/array/map.
2. Otherwise, the last value specified wins.

(note that there isn't quite as much flexibility in the types as there is in typescript, so the behavior may vary slightly)

We were mostly doing the right thing here. I just adjusted the behavior for a few array properties that weren't being appended together and added a private `func merge` that can be shared between `ReadResource`, `RegisterResource` and the unit tests to ensure that the merging behavior can't drift from what's in the tests. 

Note that we don't expose a user-facing `MergeOptions` function as options are variadic, and we handle the merging appropriately on behalf of the user in Read/RegiserResource functions. We demonstrate the appropriate behavior in the transformations documentation: https://www.pulumi.com/docs/intro/concepts/programming-model/#transformations


EvanBoyle:
Looks like this change caught a legit bug:

https://travis-ci.com/github/pulumi/pulumi/builds/158918220
https://github.com/pulumi/pulumi/pull/4316/commits/a7d6268266aeab90bcb823673fd8d0e66ea41e34

sdk/go/pulumi/resource.go
```diff
@@ -217,6 +217,17 @@ func (o resourceOrInvokeOption) applyInvokeOption(opts *invokeOptions) {
 	o(nil, opts)
 }
 
+// merging is handled by each functional options call
+// properties that are arrays/maps are always appened/merged together
+// last value wins for non-array/map values and for conflicting map values (bool, struct, etc)
+func merge(opts ...ResourceOption) *resourceOptions {
```
lukehoban, line 223: I believe a desired part of this was to expose a public API equivalent to `mergeOptions` in NodeJS. This is frequently used by user code on (a) transformations (b) some component scenarios that need to mix in options to what the user provided. The existing tests for (a) could/should probably be updated to use this API. Related - is https://www.pulumi.com/docs/intro/concepts/programming-model/#transformations “correct” wrt these changes?  If so, perhaps you his merging doesn’t need to be exposed to users?

EvanBoyle: Yes, the transformations example is still correct. With the way we implemented functional options, we don't expose the underlying `resourceOptions` struct. This implementation more or less enables equivalent behavior with nodejs. The only difference is that you have no way to introspect the existing options, delete options, etc. 

If we're fine with that caveat, variadic args allows the omission of an explicit `mergeOptions` API.

PR #4318, opened 2020-04-07
Replace IDeployment with a sealed class

### LINKED ISSUE #4241: Should Deployment.Instance return Deployment rather than IDeployment?
Currently, `Deployment.Instance` returns `IDeployment`:

https://github.com/pulumi/pulumi/blob/7efca23beb6557b9e7777fb9143df9cf41388c91/sdk/dotnet/Pulumi/Deployment/Deployment.cs#L42

https://github.com/pulumi/pulumi/blob/7efca23beb6557b9e7777fb9143df9cf41388c91/sdk/dotnet/Pulumi/Deployment/IDeployment.cs#L7-L42

What if we want to add more information in the future, such as the _qualified_ stack name that includes the org (https://github.com/pulumi/pulumi/issues/2800)?

Changing the interface would be a breaking change, so how and where would we expose it?

Should `Deployment.Instance` instead return `Deployment` so that we can add members in the future?

mikhailshilkov:
Wrap our previously public `IDeployment` interface into the sealed class `DeploymentInstance`. This way we can avoid breaking changes in case we want to extend the instance with an extra member. `IDeployment` used to be the only public interface except `IMocks`.

Resolves #4241

### DIFF
--- a/sdk/dotnet/Pulumi/Deployment/DeploymentInstance.cs
+++ b/sdk/dotnet/Pulumi/Deployment/DeploymentInstance.cs
@@ -7,7 +7,7 @@ namespace Pulumi
     /// </summary>
     public sealed class DeploymentInstance : IDeployment
     {
-        private IDeployment _deployment;
+        private readonly IDeployment _deployment;
         
         internal DeploymentInstance(IDeployment deployment)
         {
--- a/sdk/dotnet/Pulumi/Deployment/DeploymentInstance.cs
+++ b/sdk/dotnet/Pulumi/Deployment/DeploymentInstance.cs
@@ -1,25 +1,33 @@
-﻿// Copyright 2016-2019, Pulumi Corporation
-
 using System.Threading.Tasks;
 
 namespace Pulumi
 {
-    internal interface IDeployment
+    /// <summary>
+    /// Metadata of the deployment that is currently running. Accessible via <see cref="Deployment.Instance"/>.
+    /// </summary>
+    public sealed class DeploymentInstance : IDeployment
     {
+        private IDeployment _deployment;
+        
+        internal DeploymentInstance(IDeployment deployment)
+        {
+            _deployment = deployment;
+        }
+
         /// <summary>
         /// Returns the current stack name.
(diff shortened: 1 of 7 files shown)


sdk/dotnet/Pulumi/Deployment/DeploymentInstance.cs
```diff
@@ -0,0 +1,54 @@
+using System.Threading.Tasks;
+
+namespace Pulumi
+{
+    /// <summary>
+    /// Metadata of the deployment that is currently running. Accessible via <see cref="Deployment.Instance"/>.
+    /// </summary>
+    public sealed class DeploymentInstance : IDeployment
```
justinvp, line 8: Any reason why we can't use the existing `Deployment` class instead of introducing a new public type?

mikhailshilkov: Great question!
First, [this line](https://github.com/pulumi/pulumi/blob/a95a4d11954fb28b80f585cd399e570a1a6f80ea/sdk/dotnet/Pulumi.Tests/PulumiTest.cs#L23) in `PulumiTest` would be hard to re-implement.
Second, `Deployment` is currently effectively a static class: it implemented `IDeployment` only explicitly, and not directly. So, I'd have to change the `Deployment` class to add those methods to the class level.
I thought it would make sense to have this separation of the deployment "engine" and its state.
Would love to hear your opinion!

justinvp: Ah, bummer it wasn't easier to tease these apart, but makes sense!

sdk/dotnet/Pulumi/Deployment/DeploymentInstance.cs
```diff
@@ -0,0 +1,54 @@
...
+    {
+        private IDeployment _deployment;
```
justinvp, line 10: Nit: readonly ?

```suggestion
        private readonly IDeployment _deployment;
```

PR #4324, opened 2020-04-07
Move pulumi cmd sources into cmd/pulumi.

pgavlin:
Just what it says on the tin.

We may or may not want to take this--there is some value in implementing
the CLI as what is essentially a library--but if we do not, we should
find an alternative organization for these sources. Leaving them
directly under `cmd/` feels unusual.


EvanBoyle:
Given the clean builds, I guess this doesn't require any updates to any of our build or release scripts? 

I'm a little wary of potentially churning our build right now with everything we have in flight, but if you feel confident and are willing to shepherd these changes through I have no objection. 

I'm not sure I totally understand the motivation in the first place, would you mind elaborating on what you mean by "feels unusual"? Some sort of code smell?

pgavlin:
> Given the clean builds, I guess this doesn't require any updates to any of our build or release scripts?

Correct. There should be no changes required.

> I'm not sure I totally understand the motivation in the first place, would you mind elaborating on what you mean by "feels unusual"? Some sort of code smell?

See e.g. the `cmd` directory for Go itself: https://github.com/golang/go/tree/master/src/cmd

All of the sources that comprise the a tool are either under that tool's subdirectory or in a package outside of the `cmd` directory.

EvanBoyle:
Morally, I understand your point. Practically speaking it doesn't really bother me as we only have a single entry point.

pgavlin:
> Morally, I understand your point. Practically speaking it doesn't really bother me as we only have a single entry point.

At the moment. In the near-to-medium term I plan on adding some code-generation-oriented tools that will be peers of `pulumi`.

PR #4337, opened 2020-04-08
Handle k8s cases in the resource doc generator

### LINKED ISSUE #4269: Generate Kuberenetes docs based on the schema-based docs generation process

praneetloke:
Fixes #4269.

This PR can be previewed using this docs [PR](https://github.com/pulumi/docs/pull/2890).

*  Fix broken customresourcedefinition page in k8s due to html-encoding issue (The fix was to use the html-encoded property `DisplayName)
* Show the resource-level `DeprecationMessage` (The css class is introduced in the associated docs PR)
* Issue with rendering a union type string in Python.

There is still one more thing to add to the generator, which is the generator must pass the `moduleToPackage` from the `nodejs` language info object to the nodejs code gen so that the type strings are properly generated. The codegen support for considering the `moduleToPackage` is actually in-flight in another [PR](https://github.com/pulumi/pulumi/pull/4325). Once that is merged, we will be able to use that.

### DIFF
--- a/pkg/codegen/docs/gen.go
+++ b/pkg/codegen/docs/gen.go
@@ -353,7 +353,7 @@ func (mod *modContext) cleanTypeString(t schema.Type, langTypeString, lang, modN
 				if schema.IsPrimitiveType(e) {
 					continue
 				}
-				return mod.cleanTypeString(e.(*schema.ObjectType), langTypeString, lang, modName, isInput)
+				return mod.cleanTypeString(e, langTypeString, lang, modName, isInput)
 			}
 		case *schema.ObjectType:
 			objTypeModName := mod.pkg.TokenToModule(t.Token)
--- a/pkg/codegen/docs/gen.go
+++ b/pkg/codegen/docs/gen.go
@@ -769,7 +769,8 @@ func (mod *modContext) getConstructorResourceInfo(resourceTypeName string) map[s
 		case "nodejs", "go":
 			// Intentionally left blank.
 		case "csharp":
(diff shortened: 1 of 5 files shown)


pkg/codegen/docs/gen.go
```diff
@@ -1324,40 +1336,18 @@ func generateModulesFromSchemaPackage(tool string, pkg *schema.Package) map[stri
...
 	}
 
 	scanK8SResource := func(r *schema.Resource) {
 		mod := getK8SMod(pkg, r.Token, modules, tool)
 		mod.resources = append(mod.resources, r)
 
-		// For k8s, all nested properties will use a snake_case,
-		// so we'll just add them all to the respective property
-		// case maps.
-		for _, p := range r.Properties {
```
praneetloke, line 1348: I fixed the underlying issue that originally prevented me from using the same logic for generating the Python property case maps. So this special logic for k8s wasn't needed anymore.

PR #4339, opened 2020-04-09
Create mapping of display names for resource providers

sean1588:
Issue https://github.com/pulumi/docs/issues/2869

Create mapping of display names for resource providers in the API Reference TOC menu.

This should probably ideally be set in the resource provider repos themselves, but that would require a ton of changes across many repos, so for time sake we are adding this map for now.

### DIFF
--- a/pkg/codegen/docs/gen.go
+++ b/pkg/codegen/docs/gen.go
@@ -61,6 +61,39 @@ var (
 	// langModuleNameLookup is a map of module name to its language-specific
 	// name.
 	langModuleNameLookup map[string]string
+	// titleLookup is a map to map module package name to the desired display name
+	// for display in the TOC menu under API Reference.
+	titleLookup = map[string]string{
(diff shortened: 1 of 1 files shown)


pkg/codegen/docs/gen.go
```diff
@@ -1207,7 +1207,47 @@ func (mod *modContext) genIndex() indexData {
 	title := modName
 	menu := false
 	if title == "" {
-		title = mod.pkg.Name
+		// titleLookup is a map to map module package name to the desired display name
+		// for display in the TOC menu under API Reference.
+		titleLookup := map[string]string{
```
praneetloke, line 1212: Let's move this map to the top of the file, even though by the logic of `title` being empty, this will only be initialized once. It'll be easier to read this function.

sean1588: ok. cool. I moved it to the top where we delare the vars.

PR #4345, opened 2020-04-09
feat: use node 12 in docker image

### LINKED ISSUE #4344: Use node 12 or newer in Dockerfile
The version of node in the `Dockerfile` is version 11, which isn't supported at all anymore.

nesl247:
Fixes #4344


stack72:
I am testing this locally

Issue #3838, opened 2020-01-29
[feat] Deno TypeScript runtime support


brandonkal:
I'm going to throw this one out there because it is a natural fit for configuration code.

Pros:
- No package manager. Let's face it. It's annoying to download a bunch of modules just to "Hello World" a Pulumi Program. For simple things, one will just reach for YAML instead. Because Deno uses standard HTTPS imports you can just run a file and forget about packages.
- You can run via URL. `deno https://example.com/a-pulumi-program.ts` could be the Pulumi equivalent of `kubectl apply -f https://example.com/some-manifest.yaml`
- Easier to lock down (i.e. a Deno program cannot read environment variables by default so config is more hermetic). It is still an option though if the user specifies.
- Simple to support.

To support Deno Pulumi would need to
- Generate TypeScript code that includes the file extension in the imports.
- Ensure TypeScript code doesn't access modules variable and other legacy CJS things.
- Distribute code on a nice HTTP endpoint or let raw.githubusercontent.com perform that function. So an import could be
```ts
import * as pulumi from 'https://pulumi.com/pulumi@1.4.5/mod.ts'
```
- The pulumi CLI would handle installing plugin binaries as there is no postinstall scripts like npm.

pgavlin:
cc @lukehoban

gabemeola:
This would be great, and more easily allow for mixing matching languages (say Java for the server and Typescript for Pulumi infra) without needing to pull in the whole NodeJS ecosystem.  
I think https://github.com/pulumi/pulumi/issues/3901 might solve this a bit better.

Graham-Beer:
Wanted to add my support for Deno as well. Much more light weight that NodeJS and makes use of the new Javascript features under the hood where NodeJS doesn't. Built in support for TypeScript as well is a much added bonus.

brandonkal:
Closing as I no longer use Pulumi

leezen:
Re-opening as the issue is still something the community wants.

pecigonzalo:
I think Deno greatly lowers the barrier to entry for TS as you dont have to bring all the "noise" of the JS/TS ecosystem.

nuvanti:
We are a Deno shop, it'd be great if you could add Deno support.

rafaribe:
What's the current state on this? Any progress?

gaui:
Would love this!

algoflows:
The waited has been so long that the thumbs have gone down from 40 to 7 🥲

[...]

Issue #3586, opened 2019-11-26
yarn pnp support


brandonkal:
On my machine, a kubernetes-typescript install takes:
200 seconds with a standard install
20 seconds with yarn PNP

First-class Yarn PNP support would be appreciated as it would mean much faster project startup time.

Error:
```
brandon@dev:~/pulumi-project/01-first$ pulumi up
Enter your passphrase to unlock config/secrets
    (set PULUMI_CONFIG_PASSPHRASE to remember): 
Previewing update (test):
     Type                 Name                 P
     pulumi:pulumi:Stack  pulumi-project-test   
 
Diagnostics:
  pulumi:pulumi:Stack (pulumi-project-test):
    error: It looks like the Pulumi SDK has not been installed. Have you run npm install or yarn install?
 
error: failed to load language plugin nodejs: could not read plugin [/home/brandon/.pulumi/bin/pulumi-language-nodejs] stdout: EOF
```

Reproduction
```sh
pulumi new kubernetes-typescript
mkdir 01-first && mv package.json 01-first/
mv index.ts 01-first/
echo 'config: .pulumi' >> Pulumi.yaml
```
Create a package.json
```json
{
  "private": true,
  "workspaces": [
    "01-first"
  ]
}
```
```
yarn policies set-version v2
yarn install
cd 01-first
pulumi stack init test
```
The tree should now look like this:
tree
```
.
|____yarn.lock
|____tsconfig.json
|____.yarn
  ...snip...
|____.gitignore
|____.pnp.js
|____01-first
| |____index.ts
| |____package.json
|____.pulumi
| |____Pulumi.test.yaml
|____Pulumi.yaml
|____package.json
|____.yarnrc
```
```
pulumi up
// logs error noted above
```

pgavlin:
Interesting! This certainly sounds like a promising improvement.

ringods:
A first stable version of Yarn 2 has been announced on 24 January 2020:

https://dev.to/arcanis/introducing-yarn-2-4eh1

One of the biggest changes is that [Plug'n'Play](https://yarnpkg.com/features/pnp) is enabled by default. You can expect that more and more people will start using it. I just tried it too, only to see it failing and eventually find this GH issue.

[...]

Issue #3520, opened 2019-11-17
Skip dep install on `pulumi new` (prefer yarn)


brandonkal:
I prefer yarn with pnp for managing dependencies for TS/JS. Having the option to skip the auto-install of JS dependencies or use yarn would be nice.

When I ran `pulumi new` @types/node@8` was installed and I would prefer the option to change that to version 12 before the installation of dependencies occurs.

lukehoban:
There is the following flag to `pulumi new` that allows skipping this step so you can customize dependency installation. 

I’m not sure we would change the default here as a key goal of `pulumi new` is to be a simple way to get a fully initialized project. 

Any other options that would have helped you here?

From https://www.pulumi.com/docs/reference/cli/pulumi_new/: 

> -g, --generate-only             Generate the project only; do not create a stack, save config, or install dependencies

brandonkal:
Thank you for the tip. I looked through the flags before opening so I'm surprised I missed `-g`!

I'm not asking for the default to be changed. Though on that note, I would enjoy having `pulumi new` use yarn when it is in the PATH (with an environment variable option for those who wish to make the preference explicit).

pgavlin:
> I would enjoy having pulumi new use yarn when it is in the PATH (with an environment variable option for those who wish to make the preference explicit).

This is an interesting suggestion. I think that should be a relatively small change. @justinvp ?

justinvp:
Yep, this has been requested before and we are fixing as part of https://github.com/pulumi/pulumi-policy/issues/102. PR here: https://github.com/pulumi/pulumi/pull/3556

Issue #2980, opened 2019-07-25
Automatically generated function bundles don't like `yarn link`


eropple:
I'm working on shared infra for our company/for eventual open-source release and one of the things I've realized is that Pulumi has trouble with `yarn link`. I can't share the code right now but what I've discovered is that creating a `CallbackFunction` in a linked package fails with an odd error:

```
  pulumi:pulumi:Stack (data-api-v2-deploy-dev.data-api-v2):
    error: Running program 'consumer-dir' failed with an unhandled exception:
    TypeError: Cannot read property 'apply' of undefined
        at new ELKTransport (linked-dir/src/orchestration/elk-transport/index.ts:116:45)
        at consumer-dir/index.ts:60:28
        at Generator.next (<anonymous>)
        at fulfilled (consumer-dir/index.ts:4:58)
    error: Running program 'consumer-dir' failed with an unhandled exception:
    Error: failed to register new resource elk-shipper-lambda-lambda [aws:lambda/function:Function]: 2 UNKNOWN: failed to compute asset hash: asset path 'node_modules/@hydrow-infra/pulumi-tools' is a directory; try using an archive
        at Object.registerResource (linked-dir/node_modules/@pulumi/pulumi/runtime/resource.js:112:27)
        at new Resource (linked-dir/node_modules/@pulumi/pulumi/resource.js:172:24)
        at new CustomResource (linked-dir/node_modules/@pulumi/pulumi/resource.js:270:9)
        at new Function (linked-dir/node_modules/@pulumi/lambda/function.ts:221:9)
        at new CallbackFunction (linked-dir/node_modules/@pulumi/lambda/lambdaMixins.ts:325:9)
        at new NotifyingCallbackFunction (linked-dir/src/orchestration/notifying-callback.ts:71:24)
        at new ELKTransport (linked-dir/src/orchestration/elk-transport/index.ts:95:32)
        at consumer-dir/index.ts:60:28
        at Generator.next (<anonymous>)
        at fulfilled (consumer-dir/index.ts:4:58)

    Could not include required dependency '@pulumi/aws' in 'consumer-dir'.
    Could not include required dependency '@pulumi/pulumi' in 'consumer-dir'.
    Could not include required dependency '@pulumi/terraform' in 'consumer-dir'.
    Could not include required dependency 'lodash' in 'consumer-dir'.
    Could not include required dependency 'runtypes' in 'consumer-dir'.
    Could not include required dependency 'utility-types' in 'consumer-dir'.
[...]

PR #4349, opened 2020-04-09
[WIP] Initial Yarn v2 support

ringods:
# Current situation

Pulumi still has the following assumptions regarding the NodeJS language run time:

* the main entry point to invoke is `node`
* the Pulumi program is a single NodeJS project with the `node_modules` at the same level

Below are the commands as defined currently in the code:

Setup | What is executed? | Package install & pack
-------|------|-----
Node + NPM | `node` | `npm <command>`
Node + Yarn (PULUMI_PREFER_YARN) | `node` | `yarn <command>`

The commands from the column **What is executed?** are executed by the NodeJS language run time, not by core Pulumi. The commands from the column **Package install & pack** are executed by core Pulumi.

Looking at the current code, the package install is done by core Pulumi, checking for the existence of the `node_modules` folder afterwards.

# Proposal

This document describes a proposal to have full Yarn v2 support in Pulumi. While working on a solution, we will get better Yarn v1 support as a result too.

In general, an abstraction is needed around the concept of a Javascript runtime + a package manager for Javascript/Typescript modules. This abstraction should support the setups mentioned above. But the current Yarn support in Pulumi doesn't handle the following cases:

* [Yarn workspaces](https://yarnpkg.com/features/workspaces) (v1 & v2): multiple Javascript projects can be stored within a Git monorepo, usually as sub folders under a `packages` folder. Via Yarn, all the dependencies are stored in a single shared place to conserve on disk space. It is also much easier to have dependencies between the Javascript projects during development.
* [Yarn v2 Plug'n'Play](https://yarnpkg.com/features/pnp) (v2): PnP implements a totally new mechanism to retrieve and store the dependencies of a Javascript project. Gone is the `node_modules` folder. It is replaced by a cache location containing archives. At run time, modules are loaded via a composite ZIP file system.

Both features are orthogonal to each other and can be easily combined.

The assumptions from the first table are a problem for better NodeJS support. The NodeJS community has multiple package managers besides `npm`: Yarn v1, Yarn v2 and pnpm. I mention the two Yarn versions separately as v2 is substantially different.

With the proper abstraction around a Javascript run time and package manager, Pulumi is more ready for deno support as requested in #3838.

Zooming out from the current implementation, here are the real needs:

* the Pulumi NodeJS language run time must be able to start with the correct
dependencies
* check if an install of the Javascript dependencies was done correctly

For the implementation of the checks above, we should relay to the tools at hand, hence this overview of the commands to run:

Setup | What should be executed? | Package install & pack | Verify installation + parse output
----|----|----|----
Node + NPM | `node` | `npm <command>` | `npm list <pkg>`
Node + Yarn v1 (PULUMI_PREFER_YARN) | `node` | `yarn <command>` | `yarn why <pkg>`
Node + Yarn v2 (PULUMI_PREFER_YARN) | `yarn node` | `yarn <command>` | `yarn why <pkg>`

Yarn v2 introduced its way to start `node` with the correct dependencies: `yarn node`. This has support for the most simple case up to workspaces and Plug'n'Play support.

To verify the installation of the dependencies, we should refrain from hard coding a check for the `node_modules` folder. We should use the correct commands from the package manager at hand and parse the specific output. First, this supports the use of workspaces both in Yarn v1 and v2. Second, it is also the correct way to support PnP.

# Implementation

We know what to execute to get the Yarn v2 support working. The open question is how to implement this. Here is where some help or guide is needed.

Here are some findings from the code and some suggestions to implement the requirements:

* Go dependencies go from `github.com/pulumi/pulumi/pkg/**` to `github.com/pulumi/pulumi/sdk/**`
* I would place the NodeJS runtime abstraction under `sdk/nodejs/cmd/pulumi-language-nodejs`. Core Pulumi as well as the runtime can both use it.
* Refactor the current level of Yarn support to be fully backward compatible. The checks for `PULUMI_PREFER_YARN` and `PULUMI_EXPERIMENTAL` should be moved from `pkg` to `sdk` to become accessible.
* Add the Yarn v2 support behind the `PULUMI_EXPERIMENTAL` flag for the initial implementation.

# Not yet solved

An implementation of the above would integrate Pulumi and Yarn v2 better. There are still things to solve in subsequent tickets:

* Support for calculating the closure of an embedded Lambda function. This requires integration with the [Yarn PnP API](https://yarnpkg.com/advanced/pnpapi)
* The [`dist`](https://github.com/pulumi/pulumi/tree/master/sdk/nodejs/dist) scripts still invoke `node` directly, e.g. for dynamic providers. To be seen how to detect the correct environment.

# Related:

* #3586 
* #1868
* #3520
* #2980

This proposal and implementation has been discussed earlier today with @justinvp on a Zoom call.


ringods:
@justinvp @stack72 today I have been working on getting that abstraction in. While working on it, I continue to bump into more and more code that has intricate assumption on `node_modules`. E.g. finding the Pulumi plugins is iterating over the `node_modules` folder in Go code to find the plugin binaries. I see the amount of work for this Yarn support increasing beyond the time I have available. 

The key message is: **I have to halt the work here as I my own priorities are in danger. **

I hope someone at Pulumi can take over and get this implemented. Yarn 2 will be released in a final version real soon now and I suspect that more people will start to use it. It is a major improvement over `npm` and `yarn v1`.

justinvp:
> continue to bump into more and more code that has intricate assumption on `node_modules`

@ringods, Bummer! I was afraid that was going to be the case, but was still hopeful you could get a subset of things working that you care most about. Thanks for the initial work here!

crubier:
Any news on this ?

leezen:
If someone in the community would like to pick this up, that would be great! For now, I'm going to close out this PR given its age.

Issue #4330, opened 2020-04-08
Program codegen: Go


pgavlin:
- [ ] Resource variables
	- [ ] Basic instantiations
	- [ ] Ranged instantiations
		- [ ] Eventually-typed ranges (i.e. ranges that are promises or outputs)
	- [ ] Resource options
- [ ] Config variables
	- [ ] Typed config
	- [ ] Required config
	- [ ] Optional config
- [ ] Local variables
- [ ] Output variables
- [ ] Component declarations
- [ ] Component variables
- [ ] Expressions
	- [ ] Anonymous function expressions
	- [ ] Binary operator expressions
	- [ ] Conditional expressions
	- [ ] For expressions
	- [ ] Function call expressions
		- [ ] Apply intrinsic
		- [ ] Interpolate intrinsic (where necessary)
		- [ ] `entries`
		- [ ] `fileArchive`, `fileAsset`
		- [ ] `invoke`
		- [ ] `mimeType`
		- [ ] `range`
		- [ ] `readDir`
		- [ ] `toJSON`
		- [ ] etc.
	- [ ] Index expressions
	- [ ] Literal value expressions
		- [ ] Multi-line string literals
	- [ ] Object construction expressions
	- [ ] Relative traversal expressions
	- [ ] Scope traversal expressions
	- [ ] Splat expressions
	- [ ] Template expressions
	- [ ] Template join expressions
	- [ ] Tuple construction expressions
	- [ ] Unary operator expressions
- [ ] Applies
	- [ ] Proxied apply lowering
	- [ ] Interpolate lowering
- [ ] Trivia
	- [ ] Comments
	- [ ] Whitespace

EvanBoyle:
Some sub-issues to track some of the more concrete outstanding work:

https://github.com/pulumi/pulumi/issues/4760
https://github.com/pulumi/pulumi/issues/4761
https://github.com/pulumi/pulumi/issues/4762
https://github.com/pulumi/pulumi/issues/4763
https://github.com/pulumi/pulumi/issues/4764
https://github.com/pulumi/pulumi/issues/4767
https://github.com/pulumi/pulumi/issues/4768
https://github.com/pulumi/pulumi/issues/4769
https://github.com/pulumi/pulumi/issues/4770
https://github.com/pulumi/pulumi/issues/4771

PR #4352, opened 2020-04-09
Fix treatment of trivia in template expressions.

pgavlin:
And improve test coverage for the same.


pgavlin:
> Would be fun to get an issue to look at here myself to start gathering more context.

Yeah, that's a good idea. I'll see about finding something for you.

lukehoban:
> Would be fun to get an issue to look at here myself to start gathering more context.

There is of course https://github.com/pulumi/tf2pulumi/issues/148 and https://github.com/pulumi/pulumi/issues/4330 :smile:.

EvanBoyle:
@lukehoban happy to talk more about picking those up.

PR #4354, opened 2020-04-10
Refactor script for building, testing, publishing Docker containers

chrsmith:
This change renames and then refactor `scripts/build-and-publish-docker` so that it now takes a `--test` and `--publish` flag, allowing us to decouple the two operations. (It also runs tests _after_ we publish the Docker containers so that a flaky test doesn't derail the CLI release.)

This doesn't actually fix the underlying issues we have with some of the Golang unit tests which verify the entry point script of our GitHub Actions container however. It will just make it easier to reproduce any test issues, since we can just run `scripts/build-docker.sh <last published SDK> --test`.

Successfully building containers requires running `scripts/build-sdk.sh`. Because that will copy the `.tar.gz` to S3, which will then be served by https://get.pulumi.com, which is where we [download Pulumi from in our containers](https://github.com/pulumi/pulumi/blob/master/dist/pulumi/Dockerfile#L68) from.

So in the steady state, if we want to run the container tests more regularly, it will probably need to be as part of the `travis_cron` build target in the Makefile. (Which will run ~daily from `master`, which will have ran `build-sdk.sh` via `publish_packages.sh` on its last commit.)

Does that make sense? Does that sound like reasonable next steps (checking this in, addressing the reliability issues in the tests, then running all the container tests nightly via `travis_cron`)?

Part of #4136


chrsmith:
@pgavlin Friendly ping.

scripts/build-docker.sh
```diff
@@ -0,0 +1,111 @@
...
+        -c "pip install pipenv && /src/pulumi-test-containers -test.parallel=1 -test.v -test.run TestPulumiDockerImage"
+
+    # The actions container should fetch program dependencies from NPM, PIP, etc. before
+    # executing. These tests just shell out to docker run to confirm that.
+    # Disabled due to https://github.com/pulumi/pulumi/issues/4136
+    # echo_header "Executing container entrypoint tests"
+    # RUN_CONTAINER_TESTS=true go test ${ROOT}/tests/containers/... -test.run TestPulumiActionsImage -test.v
+
+    # In case there are any other unit tests defined in the module, run those as well.
+    GOOS=linux go test -c -o /tmp/pulumi-test-containers ${ROOT}/tests/containers/...
```
stack72, line 51:
```suggestion
	pushd ${ROOT}/tests
    GOOS=linux go test -c -o /tmp/pulumi-test-containers ${ROOT}/tests/containers/...
    popd
```

PR #4355, opened 2020-04-10
Support the binary option for .NET

### LINKED ISSUE #3509: Dotnet: adopt consistent approach for running programs across compiled languages
For the go SDK we look for an executable with the same name as the program before running it via 'go run' on behalf of the user. The behavior is as follows:

1. Look for an executable:
a. in `$CWD`
b. in `$GOPATH/bin`
c. somewhere on the `$PATH`
2. Fallback to executing via `go run`

We should adopt similar behavior in dotnet for consistency amongst our compiled languages.

See https://github.com/pulumi/pulumi/pull/3503 for go implementation.

mikhailshilkov:
The `binary` option was added to the Go plugin. Do the same for the .NET plugin.

Pulumi yaml example:

```
name: azure-cs
runtime:
    name: dotnet
    options:
        binary: netcoreapp3.1/Infra.dll
description: A minimal Azure C# Pulumi program
```

Resolves #3509

An error message if the given path is not found:

```
Diagnostics:
  pulumi:pulumi:Stack (azure-cs-dev):
    Could not execute because the specified command or file was not found.
    Possible reasons for this include:
      * You misspelled a built-in dotnet command.
      * You intended to execute a .NET Core program, but dotnet-netcoreapp3.1/Infra2.dll does not exist.
      * You intended to run a global tool, but a dotnet-prefixed executable with this name could not be found on the PATH.
 
```

I'm on the fence whether we should check the path ourselves. The standard error does give the name of the DLL and says it's not found, so I decided to keep it simple for now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,10 @@ CHANGELOG
 
 - Make prebuilt executables opt-in only for the Go SDK
  [#4338](https://github.com/pulumi/pulumi/pull/4338)
- 
+
+- Support the `binary` option (prebuilt executables) for the .NET SDK
+ [#4355](https://github.com/pulumi/pulumi/pull/4355)
+
 - Add helper methods for stack outputs in the Go SDK
   [#4341](https://github.com/pulumi/pulumi/pull/4341)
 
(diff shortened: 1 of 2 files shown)


mikhailshilkov:
PR in docs: https://github.com/pulumi/docs/pull/2906

PR #4356, opened 2020-04-10
Do not require setters on stack output properties in .NET

### LINKED ISSUE #4077: Do not require public setters on .NET Components / Stacks
A small but annoying implementation detail. The following code:

```cs
class ValidStack : Stack
 {
    [Output]
    public Output<string> Foo { get; }

    public ValidStack()
    {
       this.Foo = Output.Create("bar");
    }
}
```

is perfectly valid C# code but will fail at runtime because there's no public setter on `Foo`. That's a reasonable requirement for Custom Resources but not for Component Resources or Stacks.

Relax the check.

mikhailshilkov:
Resolves #4077

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,9 @@ CHANGELOG
 - Automate execution of `go mod download` for `pulumi new` Go templates
   [#4353](https://github.com/pulumi/pulumi/pull/4353)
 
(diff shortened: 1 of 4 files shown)


mikhailshilkov:
This is less straightforward than I anticipated, so I won't pursue this before the release

PR #4358, opened 2020-04-10
fix: error when setting config without value in non-interactive mode

### LINKED ISSUE #4336: `pulumi config set` command in non interacting mode ask for values
The command `pulumi config --non-interactive set <key>` without value is asking for it.  
An error should be throw.

pulumi v1.12.1

nesl247:
Fixes #4336

### DIFF
--- a/pkg/cmd/pulumi/config.go
+++ b/pkg/cmd/pulumi/config.go
@@ -288,6 +288,8 @@ func newConfigSetCmd(stack *string) *cobra.Command {
 					return readerr
 				}
 				value = cmdutil.RemoveTrailingNewline(string(b))
+			case !cmdutil.Interactive():
+				return errors.New("config value must be specified in non-interactive mode")
 			case secret:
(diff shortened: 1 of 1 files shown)


pkg/cmd/pulumi/config.go
```diff
@@ -282,6 +282,8 @@ func newConfigSetCmd(stack *string) *cobra.Command {
 			switch {
 			case len(args) == 2:
 				value = args[1]
+			case !cmdutil.Interactive():
+				return errors.New("config value must be specified in non-interactive mode")
```
justinvp, line 286: This breaks the following:

```bash
$ pulumi config set foo < value.txt 
error: config value must be specified in non-interactive mode
```

The new check needs to be moved _after_ the `case !terminal.IsTerminal(int(os.Stdin.Fd())):` check.

nesl247: Fixed

stack72:
I manually merged this @nesl247 

Thanks again!! :)

PR #4359, opened 2020-04-10
fix: don't hang with invalid --cwd

### LINKED ISSUE #4002: Pulumi hangs when given bogus --cwd argument
Passing the `--cwd` argument to a relative folder that does not exist appears to just hang. (At least on v1.11.1 on macOS.)

```
$ pulumi --cwd ./totally-bogus stack ls
```

nesl247:
This fixes the problem. I read the docs that say that all Pulumi commands should make use of `cmdutil.RunFunc`, but given that this is executed before any other commands, I'm not sure of the consequences to not having pulumi handle this itself.

To clarify on why this is the fix, Cobra doesn't handle errors in `PersistentPreRun` and Pulumi wasn't handling it either.

Fixes #4002

### DIFF
--- a/pkg/cmd/pulumi/pulumi.go
+++ b/pkg/cmd/pulumi/pulumi.go
@@ -81,7 +81,7 @@ func NewPulumiCmd() *cobra.Command {
 			"    - pulumi destroy  : Tear down your stack's resources entirely\n" +
 			"\n" +
 			"For more information, please visit the project page: https://www.pulumi.com/docs/",
-		PersistentPreRun: cmdutil.RunFunc(func(cmd *cobra.Command, args []string) error {
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
 			// We run this method for its side-effects. On windows, this will enable the windows terminal
 			// to understand ANSI escape codes.
 			_, _, _ = term.StdStreams()
@@ -128,7 +128,7 @@ func NewPulumiCmd() *cobra.Command {
 			}
 
 			return nil
-		}),
+		},
 		PersistentPostRun: func(cmd *cobra.Command, args []string) {
 			// Before exiting, if there is a new version of the CLI available, print it out.
 			jsonFlag := cmd.Flag("json")


stack72:
@chrsmith I'd love to get your thoughts on this PR as you originally opened the issue this fixes

chrsmith:
I cannot say for certain that removing the call to `cmdutil.RunFunc` isn't going to break something subtle, so I'd definitely want to keep wrapping the `PersistentPreRun` command if at all possible. And if we do remove the call to `cmdutil.RunFunc`, understand why we need to do it.

Is the specific problem that our `PersistentPreRun` command is returning an `error`, but it is getting ignored? If that is the case, it seems like we should instead create a variant of `cmdutil.RunFunc` that will return an `error` value?

Would it make more sense to update `pulumi/pulumi.go` to call a new `cmdutil.RunFuncE` that doesn't swallow any errors? e.g.

```go
...
    PersistentPreRunE: cmdutil.RunFuncE(func(cmd *cobra.Command, args []string) error {
        ...
```
However, I'm not sure why this would be needed. Looking through the code in `sdk/go/common/util/cmdutil/exit.go`, it looks like `RunFunc` calls `RunResultFunc` which calls `ExitError` if any errors are encountered. So I would have guessed that `cmdutil.RunFunc` would properly shutdown/kill the process if it ever encountered a problem. (And therefore, we wouldn't need to use Cobra's `PersistentPreRunE` field...)

In other words, why exactly do we need to use `PersistentPreRunE` instead of `PersistentPreRun`? Is this a bug somewhere in `cmdutil.RunFunc`?

nesl247:
@chrsmith I can't really answer those questions to be honest. From what I can see, it's that nothing was handling the error being thrown. Making this change seemed to work just fine, and given the very little code running in `PersistentPreRun` I figured it safe to change.

If you do feel that it makes more sense to have a custom function `RunFuncE`, then please feel free to close this PR out.

Though in my opinion, I always opt for less custom code.

chrsmith:
> Though in my opinion, I always opt for less custom code.

Totally agree :)

I had some time to take another look at this, and I tracked down the root cause. It has nothing to do with error handling, but rather us blocking until we see a value sent on a channel... which will never be sent, since we exited the `PersistentPreRun` callback early.

So I'll close this PR out. But thanks for sending out the change, since it did help in tracking down the root cause.

PR #4360, opened 2020-04-10
Remove input props from output properties list

sean1588:
Issue: https://github.com/pulumi/docs/issues/2872

In the resource docs we show the input props (resource arguments) as output props as well because they are implicitly output props as well. This is kind of redundant so we are removing the input props from the output props list output.

### DIFF
--- a/pkg/codegen/docs/gen.go
+++ b/pkg/codegen/docs/gen.go
@@ -964,6 +964,21 @@ func (mod *modContext) genLookupParams(r *schema.Resource, stateParam string) ma
 	return lookupParams
 }
 
+// filterOutputProperties removes the input properties from the properties list, returning only output props. 
+func filterOutputProperties(inputProps []property, props []property) []property {
+	var outputProps []property
+	inputMap := make(map[string]bool, len(inputProps))
+	for _, prop := range inputProps {
+		inputMap[prop.Name] = true
+	}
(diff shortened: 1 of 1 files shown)


pkg/codegen/docs/gen.go
```diff
@@ -985,6 +1001,7 @@ func (mod *modContext) genResource(r *schema.Resource) resourceDocArgs {
 		if r.StateInputs != nil {
 			stateInputs[lang] = mod.getProperties(r.StateInputs.Properties, lang, true, false)
 		}
+		outputProps[lang] = filterOutputProperties(inputProps[lang], outputProps[lang])
```
praneetloke, line 1004: Instead of filtering the output properties after calling `getProperties` for the full list of output properties, could we do it before? That way, we just call `getProperties` with the filtered list of output properties? `getProperties` calls out to each lang code gen to generate a property's type string name for every language, including some string processing to strip namespaces etc. So in terms of time spent, it would be better to avoid doing all that for _all_ the output properties and just do that for only the properties we need to.

sean1588: ahh ok. I think I get what you are saying.

sean1588: Ok. I made the adjustment.... I think that is what you were thinking of.

PR #4361, opened 2020-04-10
Remove duplicate deprecation notices.

jkisk:
This change adds logic that removes the printing of the comment field if a deprecationMessage field is present, to avoid duplication.  This approach is a tradeoff because in some places the comment field correctly adds information, see second example.  See also list of affected files in AWS docs for further reference.

Example 1: https://www.pulumi.com/docs/reference/pkg/aws/glue/job/
before:
image

after:
image

Example 2: https://www.pulumi.com/docs/reference/pkg/aws/ec2/instance/
before:
image

after:
image

List of affected files in AWS docs:
image

refers to https://github.com/pulumi/docs/issues/2867

### DIFF
--- a/pkg/codegen/docs/templates/properties.tmpl
+++ b/pkg/codegen/docs/templates/properties.tmpl
@@ -11,8 +11,8 @@
         <span class="property-type">{{- if eq .Type.Link "#" "" -}}{{- htmlSafe .Type.Name -}}{{- else -}}{{ template "linkify" .Type }}{{- end -}}</span>
     </dt>
     <dd>
-        {{- print "{{% md %}}" -}}{{- htmlSafe .Comment -}}{{- print "{{% /md %}}" -}}
-        {{- if .DeprecationMessage -}}<p class="property-message">Deprecated: {{ print "{{% md %}}" -}}{{- .DeprecationMessage -}}{{- print "{{% /md %}}" -}}</p>{{- end -}}
+        {{- if .DeprecationMessage -}}<p class="property-message">Deprecated: {{ print "{{% md %}}" -}}{{- .DeprecationMessage -}}{{- print "{{% /md %}}" -}}</p>
+        {{- else}}{{- print "{{% md %}}" -}}{{- htmlSafe .Comment -}}{{- print "{{% /md %}}" -}}{{- end -}}
     </dd>
 {{ end }}
 </dl>


pkg/codegen/docs/templates/properties.tmpl
```diff
@@ -11,8 +11,8 @@
         <span class="property-type">{{- if eq .Type.Link "#" "" -}}{{- htmlSafe .Type.Name -}}{{- else -}}{{ template "linkify" .Type }}{{- end -}}</span>
     </dt>
     <dd>
-        {{- print "{{% md %}}" -}}{{- htmlSafe .Comment -}}{{- print "{{% /md %}}" -}}
-        {{- if .DeprecationMessage -}}<p class="property-message">Deprecated: {{ print "{{% md %}}" -}}{{- .DeprecationMessage -}}{{- print "{{% /md %}}" -}}</p>{{- end -}}
+        {{- if .DeprecationMessage -}}<p class="property-message">Deprecated: {{ print "{{% md %}}" -}}{{- .DeprecationMessage -}}{{- print "{{% /md %}}" -}}</p>
```
praneetloke, line 14: @cnunciato just to be on the same page, should we do render this information as an "either or" type of thing or render both, but make sure to strip off the deprecation from the comment itself? I guess I thought we wanted to do the latter? Looking at the issue itself, I guess it is not clear to me which one you would recommend here?

cnunciato: I was hoping we'd be able to apply some sort of a regex to strip out the duplicate content somehow, yeah. Having duplicates seems slightly better to me than not having any description at all. Thoughts?

praneetloke: Yeah especially in some cases the deprecation message may not actually be duplicated in the property comment, in which case, we would want to show both. For example, here's one such case where having both is necessary:

This is on https://www.pulumi.com/docs/reference/pkg/aws/ec2/instance/.
Screenshot_2020-04-13 Instance

I do think it would be good to look at both cases to understand why the deprecation message is duplicated in the comment in one, but not the other.

jkisk: I'll keep looking into this, thanks for the feedback @praneetloke and @cnunciato .

praneetloke: Great! Thank you, @jkisk !

PR #4369, opened 2020-04-12
Properly propagate unknowns in go SDK marshaling operations

EvanBoyle:
We weren't handling this properly. It was resulting in a panic during preview with a secret that was the result of wrapping multiple apply/all calls:

```
 
    error: program failed: resolving options: malformed RPC secret: missing value
    exit status 1
```


EvanBoyle:
> I assume these changes align with what Node (and Python and .NET ) do for equivalent calls?

As far as I can tell, yes. 

> Can we add tests for this? 

We don't have unit tests that are set up to exercise this path. Happy to add them after the fact if you have ideas.

PR #4370, opened 2020-04-12
fix stack reference helpers to handle nil values

EvanBoyle:
Stack outputs can be nil and these helper functions need to return a zero value in that instance.


EvanBoyle:
Good point. Updated accordingly.

PR #4377, opened 2020-04-13
fix unknown status for secrets wrapping unknowns

EvanBoyle:
This manifested as a failed preview due to trying to configure a plugin with an unknown value:

https://github.com/pulumi/pulumi/blob/7ff46cb4fac8b4f594d9bb706757be5c191ebd75/sdk/go/common/resource/plugin/provider_plugin.go#L408

### DIFF
--- a/sdk/go/common/resource/properties_test.go
+++ b/sdk/go/common/resource/properties_test.go
@@ -159,3 +159,12 @@ func TestCopy(t *testing.T) {
 	src["c"] = NewNumberProperty(99.99)
 	assert.Equal(t, 2, len(dst))
 }
+
+func TestSecretUnknown(t *testing.T) {
+	o := NewOutputProperty(Output{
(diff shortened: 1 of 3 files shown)


sdk/go/common/resource/properties_test.go
```diff
@@ -159,3 +159,12 @@ func TestCopy(t *testing.T) {
 	src["c"] = NewNumberProperty(99.99)
 	assert.Equal(t, 2, len(dst))
 }
+
+func TestSecretUnknown(t *testing.T) {
+	o := NewOutputProperty(Output{
```
pgavlin, line 164: We should add tests for `Computed` as well. AFAIK `Output` values are pretty much unused; we generally use `Computed` instead.

EvanBoyle: added

PR #4378, opened 2020-04-13
Fix various printing issues.

pgavlin:
- Parentheses were not handled properly
- Literals inside of template control sequences were not handled
  properly

These changes also improve test coverage for the printers.


pgavlin:
Most of this change is about the handling of parentheses. Parentheses are attachable to any expression, and are semantically meaningful in the case of object construction keys, where surrounding a key in parentheses indicates that it should be evaluated as an expression rather than a literal.

The tests are here: https://github.com/pulumi/pulumi/pull/4378/files#diff-c6783b3538c6054ba0c3332489b4b9d8

PR #4382, opened 2020-04-13
fix link to be lowercase

ekrengel:
This makes sure links are lowercase.

Fixes https://github.com/pulumi/docs/issues/2837

### DIFF
--- a/pkg/codegen/docs/gen.go
+++ b/pkg/codegen/docs/gen.go
@@ -1332,7 +1332,7 @@ func (mod *modContext) genIndex() indexData {
 		parts := strings.Split(modName, "/")
 		modName = parts[len(parts)-1]
 		modules = append(modules, indexEntry{
-			Link:        modName + "/",
+			Link:        strings.ToLower(modName) + "/",
 			DisplayName: modName,
 		})
 	}


justinvp:
There are other related bugs. For example, on this page:

https://www.pulumi.com/docs/reference/pkg/gcp/serviceaccount/account/

Screen Shot 2020-04-13 at 2 59 25 PM

`Account` links to:

https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/gcp/serviceAccount/#Account

When it should link to:

https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/gcp/serviceaccount/#Account

praneetloke:
> There are other related bugs. For example, on this page:
> 
> `Account` links to:
> https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/gcp/serviceAccount/#Account
>
> When it should link to:
> https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/gcp/serviceaccount/#Account

That is actually in a different part of the generator where we deal with generating the constructor params for each language. I don't think module names can always be lower-cased for links to their API docs, though? Links within the same page as the resource docs are always lower-cased, so this PR is good for that.

ekrengel:
I will look into this as well @justinvp as a follow on

justinvp:
> I don't think module names can always be lower-cased for links to their API docs, though?

For Node.js, the paths are always lowercase. So any links to Node.js API docs need to have the path lowercased. And note, it's just the path that needs to be lowercase -- any anchors, such as `#Account`, should remain uppercase.

Links to Python API docs probably also need to have their path lowercased, because Hugo is managing the published URLs and automatically lowercasing, but in this particular case, the module is snake_cased as [`service_account`](https://www.pulumi.com/docs/reference/pkg/python/pulumi_gcp/service_account/), so it isn't actually an issue.

Go is hosted externally and can have mixed casing, so shouldn't require any changes.

C# is using static files, so Hugo isn't changing the paths to lowercase, so I believe we're good there (mixed casing is allowed).

praneetloke:
> For Node.js, the paths are always lowercase.

Ah ok. I didn't think this was true. But if that's the case, then yeah applying the same fix for that also makes sense.

PR #4383, opened 2020-04-13
Add support for stack output to Github Action step output

amichel:
Add support for stack output to Github Action step output. This works as standalone step (e.g. if stack output is passed in args) as well as after stack up, so no additional step is required in such case.
Currently only "all" value is supported in step input, which causes all stack outputs to be mapped to step output.
Usage:

```YAML

 - name: Pulumi Stack
        id: pulumi
        uses: docker://pulumi/actions
        with:
          args: up
          MAP_OUTPUT: all
        env:
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           PULUMI_ACCESS_TOKEN: ${{ secrets.PULUMI_ACCESS_TOKEN }}
           PULUMI_CI: up
           PULUMI_ROOT: my-website-infra

 - name: Show Output
        run: |
          echo ${{steps.pulumi.outputs.targetBucket}}
          echo ${{steps.pulumi.outputs.targetBucketArn}}

 - name: S3 Sync
        id: s3
        uses: awact/s3-action@0.1.1
        with:
          args: --follow-symlinks --delete
        env:
          AWS_S3_BUCKET: ${{ steps.pulumi.outputs.targetBucket }}
          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
          AWS_REGION: "eu-west-1"
          SOURCE_DIR: "./my-website-pages"
```


amichel:
Build failed in Travis on some GCP credentials problem. I don't see how it can be related my commit.

leezen:
@amichel Don't worry about that -- it's an issue with community PRs we have as non-members of the organization don't have the right permissions to decrypt the secrets that are required for a build.

chrsmith:
Thank you for the PR @amichel , I'll need to do some testing/tuning of this PR on my own. (For example, we probably need to adjust where in the `CHANGELOG.md` file you mention this.)

I'll try to take a close look at the change and how you hooked things up soon. For reference, here is some more information on how to "output" data from a GitHub Actions step:
https://github.com/actions/toolkit/blob/master/docs/commands.md#set-outputs

chrsmith:
Sorry for taking so long to get back to this! Over at Pulumi, we've been having a lot of conversations about how to approach the various CI/CD integrations we maintain.

This extension to our GitHub Actions `entrypoint.sh` is perfectly reasonable, but at the same time, we're already seeingthe pain associated with trying to have a "one size fits all" script for the GitHub Actions container. (For example, if you only need the NodeJS runtimes, you could pick a much smaller container for running Pulumi instead. Or if you aren't using AWS or GCP, one without their SDKs and toolchains, etc.)

So thank you very much for the pull request, but I'm inclined to not merge it for the time being. And instead, just encourage developers to customize their own GitHub Actions or deployment scripts to use the snippet you wrote:

```
 pulumi stack output -j \
    | jq --raw-output 'to_entries map("::set-output name=" + .key+"::" + (.value | tostring)+"^") | .[]' \
    | xargs -d '^' echo 
```

[Relevant documentation on GitHub](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-6)

I wouldn't be too surprised if we change our mind, and end up adding this in the future. But for now let's hold off on adding it. Again, thank you for the pull request; it's certainly a cool and useful capability.

PR #4384, opened 2020-04-13
[codegen/docs] Fix Kubernetes resource links for NodeJS and C# constructor params  

praneetloke:
This PR adds some special cases for generating Kubernetes resource docs:

* For constructor params in NodeJS, the current k8s code gen does not add the suffix `Args` to the args type and creates that type inside the input types namespace, whereas the schema code gen does not create the high-level resource input argument type in the input type namespace. For example, note the module path for [`BucketArgs`](https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/aws/s3/#BucketArgs) of the `aws` provider and note the path for [`DaemonSet`](https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/kubernetes/types/input/#DaemonSet) (which is the args property for creating a [`DaemonSet`](https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/kubernetes/apps/v1/#DaemonSet-constructor) resource.)

* For constructor params in C#, the input types namespace is actually `Pulumi.Kubernetes.Types.Inputs.ModuleName.ResourceName` whereas the schema-based code gen would generate `Pulumi.Kubernetes.ModuleName.Inputs.ResourceName`.

* The Python property name casing for Kubernetes is now special-cased to always use a snake_case for nested properties as input properties can be specified using either case, but output properties can only be accessed using snake_case.

### DIFF
--- a/pkg/codegen/docs/gen.go
+++ b/pkg/codegen/docs/gen.go
@@ -445,15 +445,26 @@ func (mod *modContext) typeString(t schema.Type, lang string, characteristics pr
 
 func (mod *modContext) genConstructorTS(r *schema.Resource, argsOptional bool) []formalParam {
 	name := resourceName(r)
-	argsType := name + "Args"
+	docLangHelper := getLanguageDocHelper("nodejs")
+	// Use the NodeJS module to package lookup to transform the module name to its normalized package name.
+	modName := getLanguageModuleName(mod.pkg, mod.mod, "nodejs")
+
+	var argsType string
+	var argsDocLink string
+	// The non-schema-based k8s codegen does not apply a suffix to the input types.
+	if isKubernetesPackage(mod.pkg) {
+		argsType = name
+		argsDocLink = docLangHelper.GetDocLinkForResourceInputOrOutputType(mod.pkg.Name, modName, argsType, true)
+	} else {
+		argsType = name + "Args"
+		argsDocLink = docLangHelper.GetDocLinkForResourceType(mod.pkg.Name, modName, argsType)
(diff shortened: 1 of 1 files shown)


pkg/codegen/docs/gen.go
```diff
@@ -1036,19 +1066,22 @@ func (mod *modContext) genResource(r *schema.Resource) resourceDocArgs {
...
 	name := resourceName(r)
 
-	// TODO: Unlike the other languages, Python does not have a separate Args object for inputs.
-	// The args are all just named parameters of the constructor. Consider injecting
-	// `resource_name` and `opts` as the first two items in the table of properties.
 	inputProps := make(map[string][]property)
 	outputProps := make(map[string][]property)
 	stateInputs := make(map[string][]property)
+
+	var filteredOutputProps []*schema.Property
```
praneetloke, line 1073: These changes don't have any effect on the generated resource docs. I just happened to notice that we don't need to calculate the filtered properties per-language.

pkg/codegen/docs/gen.go
```diff
@@ -809,6 +834,11 @@ func (mod *modContext) getConstructorResourceInfo(resourceTypeName string) map[s
 		case "nodejs", "go":
 			// Intentionally left blank.
 		case "csharp":
+			if mod.mod == "" {
```
praneetloke, line 837: This actually fixes a bug for _all_ providers right now. The `Provider` resource will never have a module name and as such its link is always at the package-level.

PR #4385, opened 2020-04-14
Annotate .ctor args in docs

chrsmith:
For the generated resource documentation, we were previously hard-coding the constructor arguments (in `constructor_args.tmpl`). This however isn't all that great.

- Golang needs to add a `ctx` parameter.
- The constructor args were missing their type information (i.e. https://github.com/pulumi/docs/issues/2865)

This PR fixes https://github.com/pulumi/docs/issues/2865 and tries to improve things if only slightly. It does this by:

- Adding a `ConstructorParamsTyped` (type `map[string][]formalParam`) so that we can pass along the actual set of param names and types.
- Add a new `Comment` field to `formalParam` so we can describe each ctor param.

Happy to do any refactoring or cleanup to leave the code a little better than when I found it. e.g. having the pre-rendered `ConstructorParams map[string]string` seems like a code smell. But I would guess that that's simplifying `resource.tmpl` quite a bit. (But if it's been bugging you and seems like something we'd want to merge into this PR, let me know.)

http://localhost:1313/docs/reference/pkg/kafka/acl/ **(new)**

image
image
image
image

https://www.pulumi.com/docs/reference/pkg/kafka/acl/ **(current)**

image
image

### DIFF
--- a/pkg/codegen/docs/gen.go
+++ b/pkg/codegen/docs/gen.go
@@ -553,6 +553,7 @@ func (mod *modContext) genConstructorGo(r *schema.Resource, argsOptional bool) [
 				Name: "Context",
 				Link: docLangHelper.GetDocLinkForPulumiType(mod.pkg, "Context"),
 			},
+			Comment: "Context object for the current deployment",
 		},
 		{
 			Name: "name",
--- a/pkg/codegen/docs/gen.go
+++ b/pkg/codegen/docs/gen.go
@@ -173,6 +173,9 @@ type formalParam struct {
 	OptionalFlag string
 
 	DefaultValue string
+
+	// Comment is an optional description of the parameter.
+	Comment string
 }
 
 type packageDetails struct {
@@ -188,7 +191,11 @@ type resourceDocArgs struct {
 	Comment            string
 	DeprecationMessage string
 
+	// ConstructorParams is a map from language to the rendered HTML for the constructor's
+	// arguments.
 	ConstructorParams map[string]string
+	// ConstructorParamsTyped is the typed set of parameters for the constructor, in order.
+	ConstructorParamsTyped map[string][]formalParam
 	// ConstructorResource is the resource that is being constructed or
 	// is the result of a constructor-like function.
 	ConstructorResource map[string]propertyType
@@ -467,6 +474,14 @@ func cleanOptionalIdentifier(s, lang string) string {
 	return s
 }
 
+// Resources typically take the same set of parameters to their constructors, and these
+// are the default comments/descriptions for them.
+const (
+	ctorNameArgComment = "The unique name of the resource."
+	ctorArgsArgComment = "The arguments to resource properties."
+	ctorOptsArgComment = "Bag of options to control resource's behavior."
+)
+
 func (mod *modContext) genConstructorTS(r *schema.Resource, argsOptional bool) []formalParam {
 	name := resourceName(r)
 	docLangHelper := getLanguageDocHelper("nodejs")
@@ -496,6 +511,7 @@ func (mod *modContext) genConstructorTS(r *schema.Resource, argsOptional bool) [
 				Name: "string",
 				Link: docLangHelper.GetDocLinkForBuiltInType("string"),
 			},
+			Comment: ctorNameArgComment,
 		},
 		{
 			Name:         "args",
@@ -504,6 +520,7 @@ func (mod *modContext) genConstructorTS(r *schema.Resource, argsOptional bool) [
 				Name: argsType,
 				Link: argsDocLink,
 			},
+			Comment: ctorArgsArgComment,
 		},
 		{
 			Name:         "opts",
@@ -512,6 +529,7 @@ func (mod *modContext) genConstructorTS(r *schema.Resource, argsOptional bool) [
 				Name: "CustomResourceOptions",
 				Link: docLangHelper.GetDocLinkForPulumiType(mod.pkg, "CustomResourceOptions"),
 			},
+			Comment: ctorOptsArgComment,
 		},
 	}
 }
@@ -542,6 +560,7 @@ func (mod *modContext) genConstructorGo(r *schema.Resource, argsOptional bool) [
 				Name: "string",
 				Link: docLangHelper.GetDocLinkForBuiltInType("string"),
 			},
+			Comment: ctorNameArgComment,
 		},
 		{
 			Name:         "args",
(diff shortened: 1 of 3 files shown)


pkg/codegen/docs/gen.go
```diff
@@ -469,14 +485,16 @@ func (mod *modContext) genConstructorTS(r *schema.Resource, argsOptional bool) [
...
 				Link: docLangHelper.GetDocLinkForResourceType(mod.pkg.Name, modName, argsType),
 			},
+			Comment: ctorArgsArgComment,
 		},
 		{
 			Name:         "opts",
 			OptionalFlag: "?",
 			Type: propertyType{
-				Name: "CustomResourceOptions",
+				Name: "pulumi.CustomResourceOptions",
```
praneetloke, line 494: Is there a reason to add the `pulumi.` back here?

praneetloke:
> Happy to do any refactoring or cleanup to leave the code a little better than when I found it. e.g. having the pre-rendered ConstructorParams map[string]string seems like a code smell. But I would guess that that's simplifying resource.tmpl quite a bit.

Please feel free to cleanup as needed. As you said, the pre-rendering is to avoid doing some complicated things in the template as it can be hard to follow/do. Trying to mix injected variables inside markup was a little hard for these constructor params specifically due to the kind of markup it required.

pkg/codegen/docs/templates/constructor_args.tmpl
```diff
@@ -1,19 +1,15 @@
 {{ define "constructor_args" }}
 <dl class="resources-properties">
-    <dt class="property-required" title="Required">
-        <span>name</span>
+  {{ range $params := . }}
+    <dt class="property-xxx" title="xxx">
```
praneetloke, line 4: Is that class name legit? And that `title` attribute value too.

pkg/codegen/docs/gen.go
```diff
@@ -1080,7 +1110,9 @@ func (mod *modContext) genResource(r *schema.Resource) resourceDocArgs {
 		Comment:            r.Comment,
 		DeprecationMessage: r.DeprecationMessage,
 
-		ConstructorParams:   mod.genConstructors(r, allOptionalInputs),
+		ConstructorParams:      renderedCtorParams,
+		ConstructorParamsTyped: typedCtorPArams,
```
praneetloke, line 1114:
```suggestion
		ConstructorParamsTyped: typedCtorParams,
```

pkg/codegen/docs/gen.go
```diff
@@ -1071,6 +1099,8 @@ func (mod *modContext) genResource(r *schema.Resource) resourceDocArgs {
 		Notes:      mod.pkg.Attribution,
 	}
 
+	renderedCtorParams, typedCtorPArams := mod.genConstructors(r, allOptionalInputs)
```
praneetloke, line 1102:
```suggestion
	renderedCtorParams, typedCtorParams := mod.genConstructors(r, allOptionalInputs)
```

pkg/codegen/docs/gen.go
```diff
@@ -567,15 +589,17 @@ func (mod *modContext) genConstructorCS(r *schema.Resource, argsOptional bool) [
...
 			},
+			Comment: ctorArgsArgComment,
 		},
 		{
 			Name:         "opts",
 			OptionalFlag: "?",
 			DefaultValue: " = null",
 			Type: propertyType{
-				Name: "CustomResourceOptions",
+				Name: "Pulumi.CustomResourceOptions",
```
praneetloke, line 599: Same here. Any reason to add it back?

pkg/codegen/docs/gen.go
chrsmith: Nothing in particular, just to make it clearer that the type `CustomResourceOptions` comes from the `pulumi` package/namespace/module, rather than whatever the current resource provider is. (e.g. not `aws.CustomResourceOptions`.) But if you feel strongly about it, happy to change it back.

pkg/codegen/docs/templates/constructor_args.tmpl
chrsmith: It is not! Let me fix that 🙃

chrsmith: Updated the PR to now mark params as optional/required, e.g.

image

pkg/codegen/docs/gen.go
praneetloke: Yeah we dropped those prefixes intentionally in favor of not showing namespaces for types. FWIW, I, too, felt the same way as you did and had those prefixes in the beginning. :)

pkg/codegen/docs/templates/resource.tmpl
```diff
@@ -28,30 +33,38 @@
...
+        <span>resource_name</span>
         <span class="property-indicator"></span>
+        <span class="property-type">str</span>
     </dt>
     <dd>The unique name of the resource.</dd>
     <dt class="property-optional" title="Optional">
         <span>opts</span>
         <span class="property-indicator"></span>
+        <span class="property-type">
+            <a href="#resource-arguments">object</a>
```
justinvp, line 55: This doesn't look right. `opts` in Python is typed as [`ResourceOptions`](https://www.pulumi.com/docs/reference/pkg/python/pulumi/#pulumi.ResourceOptions) and it should not be linking to the `resource-arguments` section, but rather to https://www.pulumi.com/docs/reference/pkg/python/pulumi/#pulumi.ResourceOptions

pkg/codegen/docs/gen.go
cnunciato: We did decide collectively to remove these, but as I recall we didn't spend a lot of time thinking about it (I believe the intent was to conserve space), and looking at them now, they do make things clearer. I'm personally okay with putting them back in, but if others would prefer to save that kind of change for another PR, we could do that, too. A tooltip or hover (if we did want to handle it in another PR) could be a good way to strike a balance.

chrsmith:
@justinvp thanks for the feedback, PTAL.

pkg/codegen/docs/gen.go
```diff
@@ -535,13 +553,15 @@ func (mod *modContext) genConstructorGo(r *schema.Resource, argsOptional bool) [
 				Name: "Context",
 				Link: docLangHelper.GetDocLinkForPulumiType(mod.pkg, "Context"),
 			},
+			Comment: "Context object for the current deployment",
```
justinvp, line 556: @chrsmith, super nitpick... All the other comments end with a period.

justinvp: I had a spare moment: https://github.com/pulumi/pulumi/pull/4440

PR #4386, opened 2020-04-14
Generate links for primitive types.

sean1588:
Issue: https://github.com/pulumi/docs/issues/2874

Add links for primitive types. Did a small bit of refactoring to convert `GetDocLinkForBuiltInType` to be part of the DocHelperInterface so that we can just refer to the concrete instance of `DocLanguageHelper` which is already tied to the language needed when it is initially instantiated. The only other way was to create a switch statement and make a different case for each language which felt counter-intuitive to the pattern we were already trying to establish.

### DIFF
--- a/pkg/codegen/docs/gen.go
+++ b/pkg/codegen/docs/gen.go
@@ -419,8 +419,10 @@ func (mod *modContext) typeString(t schema.Type, lang string, characteristics pr
 		// Links to anchor tags on the same page must be lower-cased.
 		href = "#" + strings.ToLower(tokenName)
 	default:
-		// Assume primitive/built-in type if no match for cases listed above.
-		href = docLanguageHelper.GetDocLinkForBuiltInType(t.String())
+		// Check if type is primitive/built-in type if no match for cases listed above.
+		if schema.IsPrimitiveType(t) {
+			href = docLanguageHelper.GetDocLinkForBuiltInType(t.String())
+		}
 	}
 
 	// Strip the namespace/module prefix for the type's display name.
--- a/pkg/codegen/docs/gen.go
+++ b/pkg/codegen/docs/gen.go
@@ -418,6 +418,9 @@ func (mod *modContext) typeString(t schema.Type, lang string, characteristics pr
 		tokenName := tokenToName(t.Token)
 		// Links to anchor tags on the same page must be lower-cased.
 		href = "#" + strings.ToLower(tokenName)
+	default:
+		// Assume primitive/built-in type if no match for cases listed above.
+		href = docLanguageHelper.GetDocLinkForBuiltInType(t.String())
 	}
 
 	// Strip the namespace/module prefix for the type's display name.
@@ -489,7 +492,7 @@ func (mod *modContext) genConstructorTS(r *schema.Resource, argsOptional bool) [
 			Name: "name",
 			Type: propertyType{
 				Name: "string",
-				Link: nodejs.GetDocLinkForBuiltInType("string"),
+				Link: docLangHelper.GetDocLinkForBuiltInType("string"),
 			},
 		},
 		{
@@ -535,7 +538,7 @@ func (mod *modContext) genConstructorGo(r *schema.Resource, argsOptional bool) [
 			Name: "name",
 			Type: propertyType{
 				Name: "string",
-				Link: go_gen.GetDocLinkForBuiltInType("string"),
+				Link: docLangHelper.GetDocLinkForBuiltInType("string"),
 			},
 		},
 		{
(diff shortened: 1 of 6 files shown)


pkg/codegen/python/doc.go
```diff
@@ -47,6 +47,11 @@ func (d DocLanguageHelper) GetDocLinkForFunctionInputOrOutputType(packageName, m
 	return ""
 }
 
+// GetDocLinkForBuiltInType returns the godoc URL for a built-in type.
+func (d DocLanguageHelper) GetDocLinkForBuiltInType(typeName string) string {
+	return ""
```
praneetloke, line 52: What do you think about creating a switch-case expression here to return the appropriate link to the [built-in types](https://docs.python.org/3/library/stdtypes.html) in Python? For example, `string` type in our schema would link to [`str`](https://docs.python.org/3/library/stdtypes.html#text-sequence-type-str).

pkg/codegen/dotnet/doc.go
```diff
@@ -43,6 +43,15 @@ func (d DocLanguageHelper) GetDocLinkForResourceType(packageName, _, typeName st
 	return fmt.Sprintf("/docs/reference/pkg/dotnet/Pulumi%s/%s.html", packageNamespace, typeName)
 }
 
+// GetDocLinkForBuiltInType returns the godoc URL for a built-in type.
```
praneetloke, line 46:
```suggestion
// GetDocLinkForBuiltInType returns the C# URL for a built-in type.
```

pkg/codegen/dotnet/doc.go
```diff
@@ -43,6 +43,15 @@ func (d DocLanguageHelper) GetDocLinkForResourceType(packageName, _, typeName st
...
+// Currently not using the typeName parameter because the returned link takes to a general page
```
praneetloke, line 47: Yep this is why I hard-coded the link and postponed the decision on how far we want to take the linking. I suggest adjusting the comment, though, to just say that it returns a link to the top-level built-in types page regardless of the type, and just leave it at that.

pkg/codegen/docs/gen.go
```diff
@@ -418,6 +418,9 @@ func (mod *modContext) typeString(t schema.Type, lang string, characteristics pr
 		tokenName := tokenToName(t.Token)
 		// Links to anchor tags on the same page must be lower-cased.
 		href = "#" + strings.ToLower(tokenName)
+	default:
+		// Assume primitive/built-in type if no match for cases listed above.
+		href = docLanguageHelper.GetDocLinkForBuiltInType(t.String())
```
praneetloke, line 423: Actually I don't think we can assume that it would be a primitive type. The schema has other types besides just Array and Object types.

I think what we should do instead is to check if the type is in-fact a primitive type by calling `schema.IsPrimitiveType(t)` and then calling `GetDocLinkForBuiltInType` to get try and get the link for it.

pkg/codegen/dotnet/doc.go
sean1588: Agreed!

pkg/codegen/python/doc.go
sean1588: I went ahead and added a link for the built in types. Seeems like there is only one top-level page for all built in types similar to how the .NET docs are structured, so I just return the link like we do there.

pkg/codegen/python/doc.go
```diff
@@ -52,6 +52,13 @@ func (d DocLanguageHelper) GetDocLinkForFunctionInputOrOutputType(pkg *schema.Pa
 	return ""
 }
 
+// GetDocLinkForBuiltInType returns the Python URL for a built-in type.
+// Currently not using the typeName parameter because the returned link takes to a general
+// top-level page containing info for all built in types.
+func (d DocLanguageHelper) GetDocLinkForBuiltInType(typeName string) string {
+	return "https://docs.python.org/3/library/stdtypes.html"
```
justinvp, line 59: @sean1588, why can't we be more specific? For example, for string, I'd expect to link directly to https://docs.python.org/3/library/stdtypes.html#text-sequence-type-str. For numeric types, link directly to https://docs.python.org/3/library/stdtypes.html#numeric-types-int-float-complex. Etc.

pkg/codegen/dotnet/doc.go
```diff
@@ -50,6 +50,13 @@ func (d DocLanguageHelper) GetDocLinkForResourceType(pkg *schema.Package, _, typ
 	return fmt.Sprintf("/docs/reference/pkg/dotnet/Pulumi%s/%s.html", packageNamespace, typeName)
 }
 
+// GetDocLinkForBuiltInType returns the C# URL for a built-in type.
+// Currently not using the typeName parameter because the returned link takes to a general
+// top -level page containing info for all built in types.
+func (d DocLanguageHelper) GetDocLinkForBuiltInType(typeName string) string {
+	return "https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/built-in-types"
```
justinvp, line 57: @sean1588, Why can't we be more specific for C#? For strings, I'd expect to link to https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/reference-types#the-string-type. For bool, I'd expect to link directly to https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/bool. Etc.

PR #4396, opened 2020-04-14
Remove contextful methods from the Go SDK.

pgavlin:
These methods are not necessary: callers that need to call contextful
APIs can close over an appropriate context in their callback. These
methods simply add noise and complexity to the SDK.


pgavlin:
Closing this for the time being.

Issue #4375, opened 2020-04-13
fish shell completions


nesl247:
Now that https://github.com/spf13/cobra released 1.0.0 with fish support, it would be great if pulumi would be updated to use this and support fish.

PR #4401, opened 2020-04-15
Enable fish completions for pulumi cli

stack72:
Fixes: #4375

This upgrades cobra to 1.0.0 (we were on 0.0.6 and only 0.0.7 was
available before 1.0.0). This enables us to allow users to generate
fish completions. We have defaulted to including the fish completion
description by default


stack72:
Post-2.0 @justinvp

nesl247:
This was merged, but they don't seem to be installed automatically when installing from homebrew. This really should be handled automatically like other applications.

gerrywastaken:
@stack72 While you added this I doubt many people know it exists: https://www.pulumi.com/docs/reference/cli/#command-line-completion

I only ended up guessing and getting lucky.

stack72:
Thanks @gerrywastaken I will get that added to the guide!

PR #4403, opened 2020-04-15
Avoid replace on second update with import applied

### LINKED ISSUE #3657: Unexpected replace on resource with `import` applied on second update
Deploying this code:

```ts
import * as pulumi from '@pulumi/pulumi'
import * as gcp from '@pulumi/gcp'

const config = new pulumi.Config()
const gcpConfig = new pulumi.Config('gcp')
const project = gcpConfig.require('project')
const zone = gcpConfig.require('zone')
const clusterName = config.require('clusterName')
const nodePoolName = config.require('nodePoolName')

const clusterId = `${project}/${zone}/${clusterName}`
const cluster = new gcp.container.Cluster(clusterName, { name: clusterName }, { import: clusterId })

const nodePoolId = `${clusterId}/${nodePoolName}`
const nodePool = new gcp.container.NodePool(
  nodePoolName,
  { cluster: cluster.name, name: nodePoolName },
  { import: nodePoolId },
)
```

In the first run, here’s `pulumi up` output:

```
Previewing update (org/development):

     Type                       Name                 Plan
 +   pulumi:pulumi:Stack        cluster-development  create
 =   ├─ gcp:container:Cluster   cosmos               import
 =   └─ gcp:container:NodePool  cosmos-node-pool     import

Resources:
    + 1 to create
    = 2 to import
    3 changes

Updating (org/development):

     Type                       Name                 Status
 +   pulumi:pulumi:Stack        cluster-development  created
 =   ├─ gcp:container:Cluster   cosmos               imported
 =   └─ gcp:container:NodePool  cosmos-node-pool     imported

Resources:
    + 1 created
    = 2 imported
    3 changes

Duration: 5s
```

Here’s the second run of `pulumi up` immedately after the first without modifying the program:

```
Previewing update (org/development):

     Type                       Name                 Plan
     pulumi:pulumi:Stack        cluster-development
 +-  ├─ gcp:container:Cluster   cosmos               replace
 +-  └─ gcp:container:NodePool  cosmos-node-pool     replace

Resources:
    +-2 to replace
    1 unchanged

Do you want to perform this update? no
confirmation declined, not proceeding with the update
```

The verbose logs include:

```
I1211 21:42:13.954057   72460 step_executor.go:321] StepExecutor worker(2): applying step import-replacement on urn:pulumi:development::cluster::gcp:container/cluster:Cluster::cosmos (preview true)
I1211 21:42:13.954068   72460 registry.go:146] GetProvider(urn:pulumi:development::cluster::pulumi:providers:gcp::default_1_7_0::14b5dd08-c2af-42f3-af0d-4277dc26002c)
I1211 21:42:13.954083   72460 provider_plugin.go:671] Provider[gcp, 0xc001316a80].Read(org-development/us-east1-c/cosmos,urn:pulumi:development::cluster::gcp:container/cluster:Cluster::cosmos) executing (#inputs=0, #state=0)
I121
...
I1211 21:42:15.476406   72460 provider_plugin.go:573] Provider[gcp, 0xc001316a80].Diff(urn:pulumi:development::cluster::gcp:container/cluster:Cluster::cosmos,cosmos) success: changes=1 #replaces=[] #stables=[region authenticatorGroupsConfig subnetwork location nodeConfig databaseEncryption zone nodePools name network project ipAllocationPolicy initialNodeCount clusterIpv4Cidr enableKubernetesAlpha defaultMaxPodsPerNode releaseChannel enableTpu description] delbefrepl=false, diffs=#[], detaileddiff=map[]
I1211 21:42:15.476996   72460 step_executor.go:321] StepExecutor worker(2): step import-replacement on urn:pulumi:development::cluster::gcp:container/cluster:Cluster::cosmos retired
```

It is unclear to me why we are deciding to do an `import-replacement` well before diffing the resource.

justinvp:
After importing some resources, and running a second update with the import still applied, an unexpected replace would occur. This wouldn't happen for the vast majority of resources, but for some it would.

It turns out that the resources that trigger this are ones that use a different format of identifier for the import input than they do for the ID property.

Before this change, we would trigger an import-replacement when an existing resource's ID property didn't match the import property, which would be the case for the small set of resources where the input identifier is different than the ID property.

To avoid this, we now store the `importID` in the statefile, and compare that to the import property instead of comparing the ID.

Fixes #3657

### DIFF
--- a/sdk/go/common/apitype/core.go
+++ b/sdk/go/common/apitype/core.go
@@ -287,6 +287,8 @@ type ResourceV3 struct {
 	Aliases []resource.URN `json:"aliases,omitempty" yaml:"aliases,omitempty"`
 	// CustomTimeouts is a configuration block that can be used to control timeouts of CRUD operations
 	CustomTimeouts *resource.CustomTimeouts `json:"customTimeouts,omitempty" yaml:"customTimeouts,omitempty"`
+	// ImportID is the import input used for imported resources.
+	ImportID resource.ID `json:"importID,omitempty" yaml:"importID,omitempty"`
 }
 
 // ManifestV1 captures meta-information about this checkpoint file, such as versions of binaries, etc.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,8 @@ CHANGELOG
 - Switch .NET projects to .NET Core 3.1
   [#4400](https://github.com/pulumi/pulumi/pull/4400)
 
+- Avoid unexpected replace on resource with `import` applied on second update.
+  [#4403](https://github.com/pulumi/pulumi/pull/4403)
 
 ## 1.14.1 (2020-04-13)
 - Propagate `additionalSecretOutputs` opt to Read in NodeJS.
--- a/pkg/backend/display/json.go
+++ b/pkg/backend/display/json.go
@@ -85,7 +85,8 @@ func stateForJSONOutput(s *resource.State, opts Options) *resource.State {
 
 	return resource.NewState(s.Type, s.URN, s.Custom, s.Delete, s.ID, inputs,
 		outputs, s.Parent, s.Protect, s.External, s.Dependencies, s.InitErrors, s.Provider,
-		s.PropertyDependencies, s.PendingReplacement, s.AdditionalSecretOutputs, s.Aliases, &s.CustomTimeouts)
+		s.PropertyDependencies, s.PendingReplacement, s.AdditionalSecretOutputs, s.Aliases, &s.CustomTimeouts,
+		s.ImportID)
 }
 
 // ShowJSONEvents renders engine events from a preview into a well-formed JSON document. Note that this does not
--- a/pkg/engine/lifecycle_test.go
+++ b/pkg/engine/lifecycle_test.go
@@ -4144,6 +4144,103 @@ func TestImport(t *testing.T) {
 	assert.Nil(t, res)
 }
 
+// TestImportWithDifferingImportIdentifierFormat tests importing a resource that has a different format of identifier
+// for the import input than for the ID property, ensuring that a second update does not result in a replace.
+func TestImportWithDifferingImportIdentifierFormat(t *testing.T) {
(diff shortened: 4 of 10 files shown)


sdk/go/common/apitype/core.go
```diff
@@ -287,6 +287,8 @@ type ResourceV3 struct {
 	Aliases []resource.URN `json:"aliases,omitempty" yaml:"aliases,omitempty"`
 	// CustomTimeouts is a configuration block that can be used to control timeouts of CRUD operations
 	CustomTimeouts *resource.CustomTimeouts `json:"customTimeouts,omitempty" yaml:"customTimeouts,omitempty"`
+	// ImportID is the import input used for imported resources.
+	ImportID resource.ID `json:"importID,omitempty" yaml:"importID,omitempty"`
```
pgavlin, line 291: We may want to bump the statefile version: if a user updates their stack with a version of the CLI that stores the import ID, then updates again with an older version, the second update will either lose the import ID (if it succeeds), or fail due to a replace (i.e. the situation this bug fixes). Personally I'm in favor of bumping the version for this, esp. since it aligns with 2.0.

justinvp: Should we just bump `DeploymentSchemaVersionCurrent` from `3` to `4` and add `4` to the following case here (along with some comments):

https://github.com/pulumi/pulumi/blob/77969c83e50f557534ee1773572ce659730eb002/pkg/resource/stack/deployment.go#L167-L170

Or would it be more clear to go through the ceremony of adding a `ResourceV4`, adding a migration (a migration that does nothing?)?

pgavlin: > Or would it be more clear to go through the ceremony of adding a ResourceV4, adding a migration (a migration that does nothing?)?

I think this would be clearer.

justinvp: I am going to merge this without the statefile version bump. https://github.com/pulumi/pulumi/issues/4409 tracks following-up post 2.0 launch.

PR #4407, opened 2020-04-15
Add API types for the v4 checkpoint

justinvp:
Resource gains a new `ImportID` field, which will be used to avoid
unexpected replaces on some resources with `import` applied on second
updates.


justinvp:
We're going to hold off on this for now. https://github.com/pulumi/pulumi/issues/4409

PR #4410, opened 2020-04-16
Add support for a PULUMI_CONSOLE_DOMAIN env var

### LINKED ISSUE #4888: AWS MSK cluster updates fail
Hi, I successfully imported an AWS MSK cluster into my Pulumi stack.

Now when I try to change any AWS MSK cluster attributes (eg tags or volume size), I get the following error:
```
Diagnostics:
  pulumi:pulumi:Stack (bos-production):
    error: update failed

    panic: fatal: An assertion has failed: Expected diff to not require deletion or replacement during Update of urn:pulumi:production::bos::aws:msk/cluster:Cluster::kafka
    goroutine 99 [running]:
    	/home/runner/go/pkg/mod/github.com/pulumi/pulumi/sdk/v2@v2.4.0/go/common/util/contract/failfast.go:23
    github.com/pulumi/pulumi/sdk/v2/go/common/util/contract.Assertf(0xc001104000, 0x67997d6, 0x48, 0xc001173790, 0x1, 0x1)
    	/home/runner/go/pkg/mod/github.com/pulumi/pulumi/sdk/v2@v2.4.0/go/common/util/contract/assert.go:33 +0x197
    github.com/pulumi/pulumi-terraform-bridge/v2/pkg/tfbridge.(*Provider).Update(0xc0007621a0, 0x717cee0, 0xc00117a660, 0xc000522000, 0xc0007621a0, 0x5ab3c01, 0xc00136c300)
    	/home/runner/go/pkg/mod/github.com/pulumi/pulumi-terraform-bridge/v2@v2.4.1-0.20200611172440-d3f6b9ed94d9/pkg/tfbridge/provider.go:835 +0x81d
    github.com/pulumi/pulumi/sdk/v2/proto/go._ResourceProvider_Update_Handler.func1(0x717cee0, 0xc00117a660, 0x64df980, 0xc000522000, 0x64d6100, 0xaaf8268, 0x717cee0, 0xc00117a660)
    	/home/runner/go/pkg/mod/github.com/pulumi/pulumi/sdk/v2@v2.4.0/proto/go/provider.pb.go:1920 +0x89
    github.com/grpc-ecosystem/grpc-opentracing/go/otgrpc.OpenTracingServerInterceptor.func1(0x717cee0, 0xc0011414d0, 0x64df980, 0xc000522000, 0xc00176a160, 0xc00176a180, 0x0, 0x0, 0x70c6680, 0xc000521290)
    	/home/runner/go/pkg/mod/github.com/grpc-ecosystem/grpc-opentracing@v0.0.0-20180507213350-8e809c8a8645/go/otgrpc/server.go:57 +0x2eb
    github.com/pulumi/pulumi/sdk/v2/proto/go._ResourceProvider_Update_Handler(0x6590a40, 0xc0007621a0, 0x717cee0, 0xc0011414d0, 0xc0004c6960, 0xc000750180, 0x717cee0, 0xc0011414d0, 0xc001168a80, 0x9b8)
    	/home/runner/go/pkg/mod/github.com/pulumi/pulumi/sdk/v2@v2.4.0/proto/go/provider.pb.go:1922 +0x14b
    google.golang.org/grpc.(*Server).processUnaryRPC(0xc000762000, 0x719dee0, 0xc00075c600, 0xc00031e800, 0xc00082ef90, 0xaabc498, 0x0, 0x0, 0x0)
    	/home/runner/go/pkg/mod/google.golang.org/grpc@v1.28.0/server.go:1082 +0x50a
    google.golang.org/grpc.(*Server).handleStream(0xc000762000, 0x719dee0, 0xc00075c600, 0xc00031e800, 0x0)
    	/home/runner/go/pkg/mod/google.golang.org/grpc@v1.28.0/server.go:1405 +0xcc9
    google.golang.org/grpc.(*Server).serveStreams.func1.1(0xc00004e9d0, 0xc000762000, 0x719dee0, 0xc00075c600, 0xc00031e800)
    	/home/runner/go/pkg/mod/google.golang.org/grpc@v1.28.0/server.go:746 +0xa1
    created by google.golang.org/grpc.(*Server).serveStreams.func1
    	/home/runner/go/pkg/mod/google.golang.org/grpc@v1.28.0/server.go:744 +0xa1
```

(I am on the latest versions of the binaries & SDKs: pulumi 2.4.0, pulumi-aws 2.10.1)

Thank you in advance for your support.

chrsmith:
A Pulumi "backend" is identified by a URL, such as `https://api.pulumi.com`, `file:///Users/chris/pulumi-checkpoints`, or `s3://my-pulumi-stacks/`. And `filestate.IsFileStateBackendURL` is used to determine which backend to use ("file state" or "http state") based on the URL's scheme.

This has served us well, however it poses a problem when you try to couple the Pulumi CLI with the Pulumi Console. The Pulumi Service is served from two different domains: api.pulumi.com (REST API) and app.pulumi.com (web frontend).

The CLI is currently wired so that `pulumi login` is the same as if you had written `pulumi login --cloud-url https://api.pulumi.com`. (Meaning you specify the REST API's domain as the "cloud backend URL".) And so whenever we go to generate a URL to the Pulumi Console, we have a simple heuristic for guessing at what the Pulumi Console domain should be.

The problem is that when you are using an on-prem version of the Pulumi Service, and not using the same `api.*` and `app.*` convention for the domains, we end up generating incorrect URLs. (e.g. if you have `api.pulumi.contoso.com` and `console.pulumi.contoso.com`, at the end of every update we will link you to https://app.pulumi.contoso.com which would be wrong.)

This PR adds a small hack, which is to honor a `PULUMI_CONSOLE_DOMAIN` environment variable that would allow you to set the domain we use for links to the Pulumi Console.

A better solution would be to somehow couple the "backend URL" with its "console URL". (e.g. when you login, we pass around something like `type HTTPBackendURLs struct { RESTAPIDomain string; ConsoleDomain string }`. However, that would be a big undertaking for how we persist credentials and accounts in the `workspace` package.

Another approach would be to have you "log in" using the Pulumi Console domain, and then figure out how to have app.pulumi.com serve both API and web traffic. (That's a reasonable idea, but would require a bunch of work server-side.)

So that leads us to just honoring a new environment variable for now.

Fixes [#4888](https://github.com/pulumi/pulumi-service/issues/4888)

### DIFF
--- a/pkg/backend/httpstate/console.go
+++ b/pkg/backend/httpstate/console.go
@@ -16,10 +16,23 @@ package httpstate
 
 import (
 	"net/url"
+	"os"
 	"path"
 	"strings"
 )
 
+const (
+	// PulumiCloudURL is the Cloud URL used if no environment or explicit cloud is chosen.
+	PulumiCloudURL = "https://" + defaultAPIDomainPrefix + "pulumi.com"
+
+	// defaultAPIDomainPrefix is the assumed Cloud URL prefix for typical Pulumi Cloud API endpoints.
+	defaultAPIDomainPrefix = "api."
+	// defaultConsoleDomainPrefix is the assumed Cloud URL prefix typically used for the Pulumi Console.
+	defaultConsoleDomainPrefix = "app."
+)
+
+// cloudConsoleURL returns a URL to the Pulumi Cloud Console, rooted at cloudURL. If there is
+// an error, returns "".
 func cloudConsoleURL(cloudURL string, paths ...string) string {
 	u, err := url.Parse(cloudURL)
 	if err != nil {
@@ -27,12 +40,22 @@ func cloudConsoleURL(cloudURL string, paths ...string) string {
 	}
 
 	switch {
+	case os.Getenv("PULUMI_CONSOLE_DOMAIN") != "":
+		// Honor a PULUMI_CONSOLE_DOMAIN environment variable to override the
+		// default behavior. Since we identify a backend by a single URI, we
+		// cannot know what the Pulumi Console is hosted at...
+		u.Host = os.Getenv("PULUMI_CONSOLE_DOMAIN")
 	case strings.HasPrefix(u.Host, defaultAPIDomainPrefix):
+		// ... but if the cloudURL (API domain) is "api.", then we assume the
+		// console is hosted at "app.".
 		u.Host = defaultConsoleDomainPrefix + u.Host[len(defaultAPIDomainPrefix):]
 	case u.Host == "localhost:8080":
+		// ... or when running locally, on port 3000.
 		u.Host = "localhost:3000"
 	default:
-		return "" // We couldn't figure out how to convert the api hostname into a console hostname
+		// We couldn't figure out how to convert the api hostname into a console hostname.
+		// We skip the host and just print the relative path.
+		return path.Join(paths...)
 	}
 
 	u.Path = path.Join(paths...)
--- a/pkg/backend/httpstate/console.go
+++ b/pkg/backend/httpstate/console.go
@@ -22,6 +22,11 @@ import (
 )
 
 const (
+	// ConsoleDomainEnvVar overrides the way we infer the domain we assume the Pulumi Console will
+	// be served from, and instead just use this value. e.g. so links to the stack update go to
+	// https://pulumi.example.com/org/project/stack/updates/2 instead.
+	ConsoleDomainEnvVar = "PULUMI_CONSOLE_DOMAIN"
+
 	// PulumiCloudURL is the Cloud URL used if no environment or explicit cloud is chosen.
 	PulumiCloudURL = "https://" + defaultAPIDomainPrefix + "pulumi.com"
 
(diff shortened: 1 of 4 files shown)


pkg/backend/httpstate/console.go
```diff
@@ -16,23 +16,46 @@ package httpstate
...
+		// console is hosted at "app.".
 		u.Host = defaultConsoleDomainPrefix + u.Host[len(defaultAPIDomainPrefix):]
 	case u.Host == "localhost:8080":
+		// ... or when running locally, on port 3000.
 		u.Host = "localhost:3000"
 	default:
-		return "" // We couldn't figure out how to convert the api hostname into a console hostname
+		// We couldn't figure out how to convert the api hostname into a console hostname.
+		// We skip the host and just print the relative path.
+		return path.Join(paths...)
```
chrsmith, line 58: This is a behavior change, but it seems like a better way to go. If we want to generate a URL to the Pulumi Console with some paths as a suffix, it seems odd to return "" in that case.

justinvp: Don't we want it to return ""? There are places in the codebase that look at the result of calls to `cloudConsoleURL` and do something different if it returns an empty string.

pkg/backend/httpstate/console.go
```diff
@@ -16,23 +16,46 @@ package httpstate
...
+// cloudConsoleURL returns a URL to the Pulumi Cloud Console, rooted at cloudURL. If there is
+// an error, returns "".
 func cloudConsoleURL(cloudURL string, paths ...string) string {
 	u, err := url.Parse(cloudURL)
 	if err != nil {
 		return ""
 	}
 
 	switch {
+	case os.Getenv("PULUMI_CONSOLE_DOMAIN") != "":
```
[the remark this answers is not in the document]
chrsmith: I debated it, and went with hard-coding it because it was only used in one place. But I'll make the change, since it's probably a better idea to export it.

chrsmith:
> Don't we want it to return ""? There are places in the codebase that look at the result of calls to cloudConsoleURL and do something different if it returns an empty string.

Great catch. Yeah, I hadn't thought about that. I'll revert that change.

justinvp:
Mind opening an issue in the docs repo so we don't forget to add this to https://github.com/pulumi/docs/blob/master/content/docs/reference/cli/environment-variables.md ? Thanks!

PR #4411, opened 2020-04-16
Fix hang if PersistentPreRun fails

### LINKED ISSUE #4002: Pulumi hangs when given bogus --cwd argument

chrsmith:
`pulumi` would hang if you gave it a bogus directory for the `--cwd` flag. As it turns out this was because of the update version check.

We create a channel (`updateCheckResult`) and make a request to see if a newer version of the CLI is available. However, if the `PersistentPreRun` callback returns an error early, then we never start the Goroutine that will write a value into `updateCheckResult`... so when we execute the `PersistentPostRun` callback, we block indefinitely trying to read a value from the channel that will never come.

https://github.com/pulumi/pulumi/blob/master/pkg/cmd/pulumi/pulumi.go#L125

With this change, we no longer block until we have received a response to the version check. This might not be what we want, since if the command is super-fast, it's possible we shutdown before the update check is complete. Let me know what you think.

Fixes #4002

### DIFF
--- a/pkg/cmd/pulumi/pulumi.go
+++ b/pkg/cmd/pulumi/pulumi.go
@@ -133,8 +133,15 @@ func NewPulumiCmd() *cobra.Command {
 			// Before exiting, if there is a new version of the CLI available, print it out.
 			jsonFlag := cmd.Flag("json")
 			isJSON := jsonFlag != nil && jsonFlag.Value.String() == "true"
-			if checkVersionMsg := <-updateCheckResult; checkVersionMsg != nil && !isJSON {
-				cmdutil.Diag().Warningf(checkVersionMsg)
+
+			select {
+			case checkVersionMsg := <-updateCheckResult:
+				if checkVersionMsg != nil && !isJSON {
+					cmdutil.Diag().Warningf(checkVersionMsg)
+				}
+			default:
+				// If the channel doesn't have anything in it yet, just ignore it
+				// and continue shutting down.
 			}
 
 			logging.Flush()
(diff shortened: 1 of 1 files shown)


justinvp:
> With this change, we no longer block until we have received a response to the version check. This might not be what we want, since if the command is super-fast, it's possible we shutdown before the update check is complete. Let me know what you think.

Alternatively, we could keep blocking and modify the couple places in `PersistentPreRun` where we return an `err` to call `close(updateCheckResult)` first, right?

chrsmith:
+1 to closing the channel since we know we won't send anything on it. Will send out an update.

pkg/cmd/pulumi/pulumi.go
```diff
@@ -133,8 +142,15 @@ func NewPulumiCmd() *cobra.Command {
...
-				cmdutil.Diag().Warningf(checkVersionMsg)
+
+			select {
+			case checkVersionMsg := <-updateCheckResult:
+				if checkVersionMsg != nil && !isJSON {
+					cmdutil.Diag().Warningf(checkVersionMsg)
+				}
+			default:
+				// If the channel doesn't have anything in it yet, just ignore it
+				// and continue shutting down.
```
justinvp, line 153: Not sure if you are still working on it, but I think these changes should now be reverted.

chrsmith: Yes, that code needed to be updated. I had to review things myself, but I think this snippet has the behavior we want:

```go
checkVersionMsg, ok := <-updateCheckResult
 if ok && checkVersionMsg != nil && !isJSON {
    ...
}
```

- If the channel is closed, it will immediately return and `ok` will be `false`.
- If the channel is opened, then it will block until there is a value to be read from the channel.

PR #4420, opened 2020-04-16
Upgrade the version stamping of pulumi to match the providers

stack72:


EvanBoyle:
Why are we removing the feature branch info from tags? I thought this was a feature not a bug?

My understanding is that this enables us to publish binaries for branches prefixed with `feature`. It gives us a convenient way to distribute early bits, or even one-off fixes with customers.

PR #4424, opened 2020-04-16
Various codegen updates.

pgavlin:
- Define `null` in Pulumi HCL2
- Bind Pulumi HCL2 in topological order s.t. variable types can be
  properly computed
- Fix resources that range over bools and numbers
- Add element, length, lookup, readFile, and split functions
- Do not rewrite function signatures with input types during binding
- Fix splat expression binding for non-lists
- Add support for evaluating expressions
- Add support for operator precedence to code generators
- Add support for constants to the HCL2 IR
- Add support for generating ranged resources in Python
- Add support for generating conditional resource in Node and Python
- Fix various naming issues in Python


pgavlin:
I'm going to go ahead and merge this to unblock https://github.com/pulumi/tf2pulumi/pull/160. I'll address any feedback after the fact.

Issue #4406, opened 2020-04-15
[codegen/docs] Generate resource docs for k8s overlays


praneetloke:
* [x] The Helm and YAML resources are `ComponentResource` type resources that need to have resource docs generated.

* [x] The `CustomResource` resource under the `apiextensions` module is also to be considered as an overlay.

praneetloke:
The code changes to support the overlay resources are complete. I'll close this issue once the docs have been regenerated and they are merged.

praneetloke:
Closing this. The overlay docs were included with https://github.com/pulumi/docs/pull/3044.

PR #4426, opened 2020-04-16
[codegen/docs] Allow specifying the languages to show in the lang chooser

praneetloke:
Related to #4406.

The real reason for doing this is to hide the `go` language for k8s overlay resources. Rather than showing "Coming soon!" I chose to hide it completely for the overlay resources, since parts of the page have static text that doesn't flow well when an unsupported language is chosen. I floated the idea of extending the `choosable` component to support disabling a language, but still showing it and perhaps showing a tooltip when the user hovers over it. But we are not going to do that right now. Maybe something to do after 2.0 launch.


praneetloke:
cc @cnunciato

pkg/codegen/docs/gen.go
```diff
@@ -184,6 +184,11 @@ type packageDetails struct {
 type resourceDocArgs struct {
 	Header header
 
+	Tool string
+	// LangChooserLanguages is a comma-separated list of languages to pass to the
+	// language-chooser short-code.
```
justinvp, line 189: Nit

```suggestion
	// language chooser shortcode.
```

praneetloke:
@justinvp I added a couple of more commits based on the discussion from the companion PR https://github.com/pulumi/docs/pull/2980 that will actually use these changes. Could you please take another look?

PR #4432, opened 2020-04-17
[codegen/docs] Improve titles tags and meta descriptions

sean1588:
Improve titles tags and meta descriptions for improved search indexing. Also remove the block_external_search_index  as well.

Issue: https://github.com/pulumi/docs/issues/2856


pkg/codegen/docs/gen.go
```diff
@@ -1129,10 +1132,20 @@ func (mod *modContext) genResource(r *schema.Resource) resourceDocArgs {
...
+	}
+
 	stateParam := name + "State"
+
 	data := resourceDocArgs{
 		Header: header{
-			Title: name,
+			Title:    name,
+			TitleTag: fmt.Sprintf("Resource %s | Package %s", name, titleTag),
+			MetaDesc: regexp.MustCompile(`(?m)^.*$`).FindString((r.Comment)),
```
praneetloke, line 1148: Let's pull this regexp to be a file-level `var` instead of compiling for each resource. It would also be helpful to have a comment to indicate what the regexp is trying to do.

pkg/codegen/docs/templates/header.tmpl
```diff
@@ -1,7 +1,8 @@
 {{ define "header" }}
 ---
 title: "{{ .Title }}"
-block_external_search_index: true
+title_tag: "{{ .TitleTag }}"
```
praneetloke, line 4: @cnunciato are we ready to remove the `block_external_search_index`?

cnunciato: With this PR, we should be, yes. I'll take a look at this this morning.

praneetloke: Ok in that case, @sean1588 we'll need to remove it from the `index.tmpl` template as well. 

Speaking of `index.tmpl`, I am also wondering if we should do a similar improvement to the meta description there as well? If it is ok to punt on that for now, I am ok with that. Just wondering...

pkg/codegen/docs/gen.go
sean1588: ahh.. good point. done.

pkg/codegen/docs/gen.go
```diff
@@ -1129,10 +1132,20 @@ func (mod *modContext) genResource(r *schema.Resource) resourceDocArgs {
...
+		titleTag = mod.pkg.Name
+	}
+
 	stateParam := name + "State"
+
 	data := resourceDocArgs{
 		Header: header{
-			Title: name,
+			Title:    name,
+			TitleTag: fmt.Sprintf("Resource %s | Package %s", name, titleTag),
```
cnunciato, line 1150: Could we get a `Module %s | ` in there as well, so that what you end up with is `Resource Bucket | Module s3 | Package AWS` here?

pkg/codegen/docs/templates/header.tmpl
cnunciato: I was assuming we would do that, so would prefer it if it didn't require too much more effort.

cnunciato: At the very least, we do need to remove the `block` properties from the frontmatter in the index files, though, yeah.

pkg/codegen/docs/gen.go
sean1588: Yea, just noticed this myself when I was adjusting the index pages.... thank you.

pkg/codegen/docs/templates/header.tmpl
sean1588: OK, @cnunciato and @praneetloke , added the changes for the title and description for the index pages. The meta_description on the modules index page is not the greatest, right now it is just essentially a list of the resources in that module, not sure what else we can give it as there is no human readable description at the module level.... The meta_description on the package index page looks fine though.

cnunciato: When I run the generator, I see this fairly often, which doesn't seem right:

image

Looking at the `meta_desc`s that do render, though, I do think we could probably still be a bit more descriptive. In the issue, I mention referring to the API docs as a guide. They currently render:

```
<meta name="description" content="Explore members of the s3 module in the @pulumi/aws package.">
```

So I think beyond just "Provides a S3 bucket resource", we could do something similar:

```
<meta name="description" content="Explore the resources and functions of the s3 module in the aws package.">
```

And at the resource level, maybe something like:

```
<meta name="description" content="Explore the Bucket resource of the aws/s3 module, including examples, input properties, output properties, lookup functions, and supporting types.">
```

It's just a bit of extra text, but this is what gets used in previews and possibly search results, so it's good to try to put something informative in that space. For example:

image

pkg/codegen/docs/gen.go
```diff
@@ -1130,9 +1138,16 @@ func (mod *modContext) genResource(r *schema.Resource) resourceDocArgs {
 	}
 
 	stateParam := name + "State"
+
+	baseDescription := fmt.Sprintf("Explore the %s resource of the %s module,"+
+		" including examples, input properties, output properties,"+
+		" lookup functions, and supporting types. ", name, mod.mod)
```
praneetloke, line 1209: I suggest removing the space at the end of the sentence here and instead, make the whitespace character between the two descriptions explicit on line 1150. So `baseDescription +" "+ metaDescriptionRegexp.FindString((r.Comment))`. Just so someone doesn't accidentally try to remove the space from this sentence.

pkg/codegen/docs/gen.go
```diff
@@ -1514,6 +1545,26 @@ func (mod *modContext) genIndex() indexData {
 	return data
 }
 
+<<<<<<< HEAD
```
praneetloke, line 1548: Looks like you have some merge conflict. Pat just merged some changes related to decoding the language info from the schema.

pkg/codegen/docs/gen.go
```diff
@@ -1514,6 +1539,23 @@ func (mod *modContext) genIndex() indexData {
...
 
+func formatTitleText(title string) string {
+	// If title not found in titleLookup map, default back to title given.
+	if val, ok := titleLookup[title]; ok {
+		return val
+	}
+	return title
+}
+
+func decodeLangSpecificInfo(pkg *schema.Package, lang string, obj interface{}) error {
```
praneetloke, line 1550: I think you may have made a slight error in resolving the conflict here. This function was actually removed with Pat's recent PR as it is no longer needed. See https://github.com/pulumi/pulumi/pull/4454/files#diff-ffd570ac00ee5f55462913956726d9d9L1518.

pkg/codegen/docs/gen.go
sean1588: good thinking

stack72:
Merging as this has gone green - @praneetloke has been waiting on this

PR #4433, opened 2020-04-17
Add version to package details

sean1588:
Issue: https://github.com/pulumi/docs/issues/2984

### DIFF
--- a/pkg/codegen/docs/templates/package_details.tmpl
+++ b/pkg/codegen/docs/templates/package_details.tmpl
@@ -8,7 +8,7 @@
     {{ if ne .Notes "" -}}
 	<dt>Notes</dt>
 	<dd>{{ htmlSafe .Notes }}</dd>
-    {{- end }}
+	{{- end }}
 	{{ if ne .Version "" -}}
 	<dt>Version</dt>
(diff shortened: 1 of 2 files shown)


pkg/codegen/docs/templates/package_details.tmpl
```diff
@@ -9,5 +9,9 @@
 	<dt>Notes</dt>
 	<dd>{{ htmlSafe .Notes }}</dd>
     {{- end }}
+	{{ if ne .Version "" -}}
```
praneetloke, line 12: The indentation looks a bit off here? Could you please take a look?

sean1588: yea, it was really weird. I also noticed this when I was initially reviewing the diffs. On my machine it was showing up as being perfectly aligned when opened in VSCode, must be some weird white space issue. Let me fool with it and see if it will cooperate.

sean1588: OK. I think it's correct now. Appears one line was using tabs and the other had spaces.... lol. 🙄

praneetloke: Ahh I see. Thank you for fixing it :)

PR #4436, opened 2020-04-17
Fix header anchor ids in resource docs

justinvp:
1. Add ids to headers so they can be linked to
2. Make header ids more general
    - Rather than including the identifier in the name of the id, use a simpler more generic id.

@praneetloke, @cnunciato, what do you think about (2)?

Fixes https://github.com/pulumi/docs/issues/2987

### DIFF
--- a/pkg/codegen/docs/templates/index_categories.tmpl
+++ b/pkg/codegen/docs/templates/index_categories.tmpl
@@ -1,5 +1,5 @@
 {{- define "index_modules" }}
-<h3>Modules</h3>
+<h3 id="modules">Modules</h3>
 <ul class="api">
 {{- range . }}
     <li><a href="{{ .Link }}" title="{{ .DisplayName }}"><span class="symbol module"></span>{{ .DisplayName }}</a></li>
@@ -8,7 +8,7 @@
 {{- end }}
 
 {{- define "index_resources" }}
-<h3>Resources</h3>
+<h3 id="resources">Resources</h3>
 <ul class="api">
 {{- range . }}
     <li><a href="{{ .Link }}" title="{{ .DisplayName }}"><span class="symbol resource"></span>{{ .DisplayName }}</a></li>
@@ -17,7 +17,7 @@
 {{- end }}
 
 {{- define "index_functions" }}
-<h3>Functions</h3>
+<h3 id="functions">Functions</h3>
 <ul class="api">
 {{- range . }}
     <li><a href="{{ .Link }}" title="{{ .DisplayName }}"><span class="symbol function"></span>{{ .DisplayName }}</a></li>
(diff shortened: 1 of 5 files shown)


pkg/codegen/docs/templates/index_categories.tmpl
```diff
@@ -1,5 +1,5 @@
 {{- define "index_modules" }}
-<h3>Modules</h3>
+<h3 id="modules">Modules</h3>
```
praneetloke, line 2: Actually now that I am look at this separately. I don't know why these don't just use `###`?

praneetloke: ..here and in other similar places where the text content is static. The reason we switch to the `<h*>` tag is because adding `<wbr>` tags in the title was not rendering with the markdown equivalent for the header tags.

cnunciato:
> @cnunciato what do you think about (2)?

 I like it! 👍

pkg/codegen/docs/templates/index_categories.tmpl
justinvp: Yeah, these could be `###` here, but I am just going to leave them `<h3>` for now.

The `<wbr>` thing is a good catch. Will fix.

justinvp: Separate related issue I noticed. Now that these have ids, they show up on the right-hand "ON THIS PAGE"...

The package details template gets rendered both here and on individual resource/function pages, and since these are all h3, so is the package details header.

But on the individual resource/function pages, the top-level headers are all h2, which ends up making it look like "Package Details" is a subheading under "Supporting Types".

See bottom of this screenshot from the Bucket page:

Screen Shot 2020-04-17 at 11 39 45 AM

To address, I'm going to promote these h3s to h2.

praneetloke: > To address, I'm going to promote these h3s to h2.

sounds good! Great that you noticed this!

justinvp:
OK, added a couple more commits. @praneetloke, @cnunciato, mind taking another look?

For reference, here's an example diff of regenerated AWS docs with these changes: https://github.com/pulumi/docs/commit/1d154281f74379dd843c2022e96ebb2ad4db3432 (Ignore the changes in the examples; I must have a mismatched `tf2pulumi` version).

cnunciato:
@justinvp looks great. 👍

cnunciato:
Actually, sorry -- I just realized theres a conflict here with this PR, which modifies these headings as well: https://github.com/pulumi/pulumi/pull/4436/files

Apologies for not noticing this sooner. How do you want to reconcile these?

justinvp:
@cnunciato, you mean https://github.com/pulumi/pulumi/pull/4408? I can wait until after @sean1588's changes have merged, and then apply these changes (w/ necessary tweaks) on top.

cnunciato:
Yes, bad copy-paste, sorry. 🙄 That sounds good, thanks.

PR #4443, opened 2020-04-18
Inject `id` resource output property in docs

justinvp:
All resources have an implicit `id` output property that must be injected in docs (per https://github.com/pulumi/docs/issues/2975#issuecomment-615411715). (We don't need to do this for TF data sources, as they already have an id).

Here's how this change looks in the docs for AWS and Kubernetes:
- AWS: https://github.com/pulumi/docs/commit/4c45a5828c2575337f319c0347168b124fc94064
- Kubernetes: https://github.com/pulumi/docs/commit/6c8fccdbbfab02694f866d3c04f7b8b976ef738b

Fixes https://github.com/pulumi/docs/issues/2884
Fixes https://github.com/pulumi/docs/issues/2975

### DIFF
--- a/pkg/codegen/docs/gen.go
+++ b/pkg/codegen/docs/gen.go
@@ -819,9 +819,15 @@ func (mod *modContext) getProperties(properties []*schema.Property, lang string,
 		})
 	}
 
-	// Sort required props to move them to the top of the properties list.
+	// Sort required props to move them to the top of the properties list, then by name.
 	sort.SliceStable(docProperties, func(i, j int) bool {
-		return docProperties[i].IsRequired && !docProperties[j].IsRequired
+		pi, pj := docProperties[i], docProperties[j]
+		switch {
+		case pi.IsRequired != pj.IsRequired:
+			return pi.IsRequired && !pj.IsRequired
+		default:
+			return pi.Name < pj.Name
+		}
 	})
 
 	return docProperties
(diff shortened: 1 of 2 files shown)


pkg/codegen/docs/templates/resource.tmpl
```diff
@@ -86,15 +86,10 @@ The {{ .Header.Title }} resource accepts the following [input]({{ htmlSafe "{{<
 
 <!-- Output properties -->
 ### Outputs
-{{ if ne (len .OutputProperties) 0 }}
```
praneetloke, line 89: Is there a reason this was removed?

pkg/codegen/docs/gen.go
```diff
@@ -1145,15 +1151,21 @@ func (mod *modContext) genResource(r *schema.Resource) resourceDocArgs {
 	if !r.IsProvider {
 		filteredOutputProps = filterOutputProperties(r.InputProperties, r.Properties)
 	}
-	hasFilteredOutputProps := len(filteredOutputProps) > 0
+
+	// All resources have an implicit `id` output property, that we must inject into the docs.
+	filteredOutputProps = append(filteredOutputProps, &schema.Property{
+		Name:       "id",
```
praneetloke, line 1157: The css class name for this property's `<dt>` tag is being generated as `property-`. I guess it's fine, yeah? Or should we decorate the property somehow uniquely?

pkg/codegen/docs/templates/resource.tmpl
justinvp: Yes. Since we're injecting an output property, there should always be at least one.

pkg/codegen/docs/gen.go
justinvp: I noticed this as well, and it's already true for all the output properties. I was thinking about cleaning that up in a separate change, but since it has no visible impact, seemed like a lower priority change.

pkg/codegen/docs/templates/resource.tmpl
praneetloke: Ah yes. That's what this PR does. Duh!

..carry on, sir :)

PR #4445, opened 2020-04-19
Codegen: collect imports from function calls.

pgavlin:
Just what it says on the tin.


pgavlin:
@lukehoban @EvanBoyle merging to unblock further work. Happy to respond to feedback later.

Issue #3645, opened 2019-12-11
Crash in runtime/resource.ts when creating resources inside `apply` 


lukehoban:
The following example leads to an exception in core Pulumi:

```
  pulumi:pulumi:Stack (cert-demo-cert-demo):
    TypeError: Cannot read property 'concat' of undefined
        at C:\Users\Justin\Source\cert-demo\node_modules\@pulumi\pulumi\runtime\resource.js:356:29
        at Generator.next (<anonymous>)
        at fulfilled (C:\Users\Justin\Source\cert-demo\node_modules\@pulumi\pulumi\runtime\resource.js:18:58)
```

When setting `SHOW_BUG` to true the issue repros:

```ts
import * as pulumi from "@pulumi/pulumi";
import * as aws from "@pulumi/aws";
const SHOW_BUG = true
const apex = process.env.APEX_DOMAIN || "example.com"
const cert = new aws.acm.Certificate("cert", {
  domainName: `v.${apex}`,
  subjectAlternativeNames: [`vv.${apex}`, `vvv.${apex}`],
  validationMethod: "DNS",
});
const zone = aws.route53.getZone({
  name: apex,
  privateZone: false,
});
if(SHOW_BUG) {
  const domains = cert.domainValidationOptions.apply((opts) => 
    opts.map((opt) => 
      new aws.route53.Record(opt.resourceRecordName, {
        name: opt.resourceRecordName,
        records: [opt.resourceRecordValue],
        ttl: 60,
        type: opt.resourceRecordType,
        zoneId: zone.id,
      }, { dependsOn: cert })
    )
  )
  new aws.acm.CertificateValidation(`cert-${apex}`, { 
    certificateArn: cert.arn,
    validationRecordFqdns: domains.apply((doms) => doms.map((dom) => dom.fqdn))
  }, { dependsOn: domains })
}
else {
  let domains: Array<aws.route53.Record> = []
  for(let i = 0; i < 3; i++) {
    domains.push(new aws.route53.Record(`val-${i}`, {
      name: cert.domainValidationOptions[i].resourceRecordName,
      records: [cert.domainValidationOptions[i].resourceRecordValue],
      ttl: 60,
      type: cert.domainValidationOptions[i].resourceRecordType,
      zoneId: zone.id,
    }, { dependsOn: cert }))
  }
  new aws.acm.CertificateValidation(`cert-${apex}`, { 
    certificateArn: cert.arn,
[...]

PR #4446, opened 2020-04-19
Allows conditional dependencies and Resources from apply

Fryuni:
When creating a Resource inside an apply the Output might result in `undefined` during previews. This was reported in #3645 I added the use case that got me to this problem there

I edited this directly on the JS on my local `node_modules` just to get the deployment working and then replicated it here from the Github editor. No idea where to add a test for this, but I'll gladly add as soon as I find out :sweat_smile:


pgavlin:
I've tested this locally and everything looks good. Going to go ahead and merge.

PR #4447, opened 2020-04-19
[codegen/docs] Fix the anchor link used in resource docs template.

praneetloke:
The text is talking about inputs, so it should link to the inputs section of the programming model docs.

### DIFF
--- a/pkg/codegen/docs/templates/resource.tmpl
+++ b/pkg/codegen/docs/templates/resource.tmpl
@@ -80,7 +80,7 @@ To learn more about resource properties and how to use them, see [Inputs and Out
 
 ### Inputs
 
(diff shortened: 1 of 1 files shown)


justinvp:
I was initially confused by this as well, but I don't think there is an `inputs` id on that page to anchor link to.

The section about _both_ Inputs and Outputs has an explicit id of just `outputs`:

https://github.com/pulumi/docs/blame/master/content/docs/intro/concepts/programming-model.md#L1781

```markdown
### Inputs and Outputs {#outputs}
```

So I think this PR can be closed.

praneetloke:
> The section about _both_ Inputs and Outputs has an explicit id of just `outputs`:

Ahh. Yeah I'll close this.

Issue #4444, opened 2020-04-18
Stack command or option to just print the stacks name


mitchellmaler:
Currently there is no command or option to have the stack name printed out to the console. You have to run the `stack` command which prints out a bunch of information along with the name or you can run `stack ls` and grep for the selected one. 

Would be great to have a stack command like `stack current`, or `stack name`, or`stack ls --name` that just prints the current name.

stack72:
I sorted of experimented today with “Pulumi stack —show-name”

Does this work for you?

mitchellmaler:
Sure that works! Lines up with the other --show- flags.

stack72:
Ok I’ll open the PR in the morning and tag you in it for reference

PR #4450, opened 2020-04-19
Add flag to `pulumi stack` to output only the stack name

stack72:
Fixes: #4444

Before:
```
$ pulumi stack
Current stack is 47BE2956-D665-4EC3-9AE6-4D4A1C417074:
    Managed by demo-mbp
    No updates yet; run 'pulumi up'
Current stack resources (0):
    No resources currently in this stack

Use `pulumi stack select` to change stack; `pulumi stack ls` lists known ones
```

After:
```
$ pulumi stack --show-name
47BE2956-D665-4EC3-9AE6-4D4A1C417074
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ CHANGELOG
   
 - Protect against panic when unprotecting non-existant resources
   [#4441](https://github.com/pulumi/pulumi/pull/4441)
+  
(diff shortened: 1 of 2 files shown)


stack72:
@mitchellmaler FYI

mitchellmaler:
Looks great!

PR #4453, opened 2020-04-20
Ensure yes is passed when creating a new template in a container

stack72:
I fixed this locally when creating the new docker containers so it's
time to push this back to master

### DIFF
--- a/tests/containers/containers_test.go
+++ b/tests/containers/containers_test.go
@@ -59,7 +59,7 @@ func TestPulumiDockerImage(t *testing.T) {
 			}()
 
 			stackName := fmt.Sprintf("%s/container-%s-%x", stackOwner, template, time.Now().UnixNano())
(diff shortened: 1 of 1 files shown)


tests/containers/containers_test.go
```diff
@@ -59,7 +59,7 @@ func TestPulumiDockerImage(t *testing.T) {
 			}()
 
 			stackName := fmt.Sprintf("%s/container-%s-%x", stackOwner, template, time.Now().UnixNano())
-			e.RunCommand("pulumi", "new", template, "-f", "-s", stackName)
+			e.RunCommand("pulumi", "new", template, "yes", "-f", "-s", stackName)
```
justinvp, line 62: This needs to be `--yes` or `-y`:

```suggestion
			e.RunCommand("pulumi", "new", template, "-y", "-f", "-s", stackName)
```

Issue #4155, opened 2020-03-23
grpc: received message larger than max


nrf110:
I'm getting the following error when trying to pass the string contents of a large file (webpack bundles) into a resource input.: `Error: failed to register new resource static_css_2-dc0caae3-chunk-css-map [cloudflare:index/workersKv:WorkersKv]: 8 RESOURCE_EXHAUSTED: grpc: received message larger than max (4595819 vs. 4194304)`

This is the code I'm using to reproduce the issue:
```typescript
pulumi.output(args.contentDir).apply((contentDir) => {
            glob.sync(`${contentDir}/**`)
            .filter((file) => fs.statSync(file).isFile())
            .map((file) => {
                const key = path.relative(contentDir, file);
                const kvName = key
                    .toLowerCase()
                    .replace(/\//g, '_')
                    .replace(/\.|\s/g, '-');
                return new cf.WorkersKv(kvName, {
                    namespaceId: namespace.id,
                    key,
                    value: fs.readFileSync(file, { encoding: 'utf8' }),
                }, { parent: this });
            });
        });
```

danielrbradley:
Had a similar issue when uploading blobs to azure using the `StringAsset` class, but not when using a `FileAsset`.

```
grpc: received message larger than max (6900224 vs. 4194304)
```

Also interesting was that the failure caused some state corruption where it then assumed it had created the blobs and then got stuck trying to delete them which resulted in "not found" errors.

Issue #3811, opened 2020-01-26
Ensure all integration tests run on Windows


stack72:
As a follow up to #3656, there are a number of integration tests that currently fail on Windows:

* [ ] TestConfigCaptureNodeJS
* [ ] TestStackReferenceNodeJS
* [ ] TestStackReferencePython
* [ ] TestStackReferenceDotnet
* [ ] TestStackReferenceSecretsNodejs
* [ ] TestStackReferenceSecretsDotnet
* [ ] TestDynamicPython
* [ ] TestPartialValuesPython
* [ ] TestStackBadParenting
* [ ] TestLargeResourceNode

We need to ensure these are not skipped on each test run

Related: #3656

mikhailshilkov:
This task has been open for two years and I'm pretty sure the todo list is out of date. I'll close it, we should improve our CI setup independently of this list.

PR #4455, opened 2020-04-20
GH-4155 - Increase gRPC message size

komalali:
Fixes: #4155 

Related: https://github.com/pulumi/pulumi-policy/pull/239

Comments:
* I tried to minimize the number of individual places where that `1024 * 1024 * 400` number is referenced by creating a util function, but that only helps for the Go code. Is there a preferred way of adding such a widespread setting?
* Is the dotnet code adequate? Haven't ever worked in that lang before.
* For #4155 to truly be fixed, providers will need to be regenerated with pulumi-terraform-bridge that pulls this updated code and the related pulumi-policy PR must be merged.


sdk/dotnet/Pulumi/Deployment/GrpcEngine.cs
```diff
@@ -12,7 +13,9 @@ internal class GrpcEngine : IEngine
 
         public GrpcEngine(string engine)
         {
-            this._engine = new Engine.EngineClient(new Channel(engine, ChannelCredentials.Insecure));
+            var maxRpcMessageSize = 400 * 1024 * 1024;
```
lukehoban, line 16: Perhaps add a comment explaining this constant?  Here and in all the other similar locations.

sdk/go/pulumi/context.go
```diff
@@ -27,13 +27,13 @@ import (
...
 	multierror "github.com/hashicorp/go-multierror"
-	"google.golang.org/grpc"
-
 	"github.com/pulumi/pulumi/sdk/v2/go/common/resource"
 	"github.com/pulumi/pulumi/sdk/v2/go/common/resource/plugin"
 	"github.com/pulumi/pulumi/sdk/v2/go/common/util/contract"
 	"github.com/pulumi/pulumi/sdk/v2/go/common/util/logging"
+	"github.com/pulumi/pulumi/sdk/v2/go/common/util/rpcutil"
 	pulumirpc "github.com/pulumi/pulumi/sdk/v2/proto/go"
+	"google.golang.org/grpc"
```
lukehoban, line 36: Very minor - but we've generally maintained three blocks of imports (1) builtins (2) external dependencies (3) internal dependencies.

sdk/nodejs/runtime/settings.ts
```diff
@@ -25,6 +26,9 @@ const resrpc = require("../proto/resource_grpc_pb.js");
 const resproto = require("../proto/resource_pb.js");
 const structproto = require("google-protobuf/google/protobuf/struct_pb.js");
 
+const maxRPCMessageSize: number = 1024 * 1024 * 400;
+const grpcChannelOptions = <ChannelOptions>{"grpc.max_receive_message_length": maxRPCMessageSize};
```
lukehoban, line 30: Is the cast needed?  Or is this an option?

```suggestion
const grpcChannelOptions: ChannelOptions = {"grpc.max_receive_message_length": maxRPCMessageSize};
```

sdk/go/pulumi/context.go
komalali: That's what those 3 blocks represent! Okay good to know. Will fix.

sdk/nodejs/runtime/settings.ts
komalali: ah yup totally, typescript n00b moment.

sdk/nodejs/tests/runtime/langhost/run.spec.ts
```diff
@@ -1136,6 +1136,24 @@ describe("rpc", () => {
...
+                assert.deepEqual(res, { "largeStringProp": longString });
+                return {
+                    urn: makeUrn(t, name),
+                    id: name,
+                    props: {
+                        "largeStringProp": "a".repeat(1024 * 1024 * 5),
+                    },
+                };
+            },
+        },
```
lukehoban, line 1156: Do these tests actually go through the RPC layer?  I don't recall off the top of my head - but I feel like they do not?  cc @pgavlin 

One place we could add tests to be sure they hit the RPC is in `/test/integration`.  Those will run the full "normal" CLI/Engine/LangHost setup.

lukehoban: BTW - same likely applies to other languages.

komalali: @lukehoban yes they do. I made sure by checking that they fail on master with the same resource exhausted error. I can definitely add integration tests too but I am certain that these tests go through the langhost <-> engine RPC layer.

komalali:
Okay! I added in the integration tests here, added the skipped Node.js test in Windows to the list in https://github.com/pulumi/pulumi/issues/3811 and I think we're good to go here.

PR #4456, opened 2020-04-21
[codegen/go] Fix accessors on struct ptr outputs

lukehoban:
The accesor methods on nestred struct Ptr outputs were previously not accepting pointer typed inputs to their applier callbacks as they should, and would thus always panic if used.

The "simple" fix would be to just accept the pointer type and blindly dereference it.  But this doesn't seem like the right experience - it would make these accessors very unsafe to use in practice.

Instead, this PR implements the accessors on pointer-typed outputs as nil-coaslescing, always lifting the output type into a pointer type and flowing a nil value into the result type.  This ensures the accessor will not nil-deref, and that user code can handle the `nil` value itself (or use `.Apply` directly to implement more specialized behaviour).

Before:

```go
// Name of your S3 bucket.
func (o BuildStorageLocationPtrOutput) Bucket() pulumi.StringOutput {
	return o.ApplyT(func(v BuildStorageLocation) string { return v.Bucket }).(pulumi.StringOutput)
}
```

After:

```go
// Name of your S3 bucket.
func (o BuildStorageLocationPtrOutput) Bucket() pulumi.StringPtrOutput {
	return o.ApplyT(func(v *BuildStorageLocation) *string {
		if v == nil {
			return nil
		}
		return &v.Bucket
	}).(pulumi.StringPtrOutput)
}
```

However, due to the decision to have this more usable behaviour, this is a breaking change, as some/many accessors now return a pointer type when they previously did not.

Fixes pulumi/pulumi-azure#530.

### DIFF
--- a/pkg/codegen/go/gen.go
+++ b/pkg/codegen/go/gen.go
@@ -512,10 +512,19 @@ func (pkg *pkgContext) genOutputTypes(w io.Writer, t *schema.ObjectType, details
 
 		for _, p := range t.Properties {
 			printComment(w, p.Comment, false)
-			outputType, applyType := pkg.outputType(p.Type, !p.IsRequired), pkg.plainType(p.Type, !p.IsRequired)
+			outputType, applyType := pkg.outputType(p.Type, true), pkg.plainType(p.Type, true)
+			deref := ""
+			if p.IsRequired {
+				deref = "&"
+			}
 
 			fmt.Fprintf(w, "func (o %sPtrOutput) %s() %s {\n", name, Title(p.Name), outputType)
-			fmt.Fprintf(w, "\treturn o.ApplyT(func (v %s) %s { return v.%s }).(%s)\n", name, applyType, Title(p.Name), outputType)
+			fmt.Fprintf(w, "\treturn o.ApplyT(func (v *%s) %s {\n", name, applyType)
+			fmt.Fprintf(w, "\t\tif v == nil {\n")
+			fmt.Fprintf(w, "\t\t\treturn nil\n")
+			fmt.Fprintf(w, "\t\t}\n")
+			fmt.Fprintf(w, "\t\treturn %sv.%s\n", deref, Title(p.Name))
+			fmt.Fprintf(w, "\t}).(%s)\n", outputType)
 			fmt.Fprintf(w, "}\n\n")
 		}
 	}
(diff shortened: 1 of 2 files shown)


elsesiy:
Thanks @lukehoban, this must be a new record 🥇

lukehoban:
> However, due to the decision to have this more usable behaviour, this is a breaking change, as some/many accessors now return a pointer type when they previously did not.

Notably though, the existing methods were guaranteed to panic if called, so I don’t think this is actually a breaking change in practice.

pkg/codegen/go/gen.go
```diff
@@ -512,10 +512,19 @@ func (pkg *pkgContext) genOutputTypes(w io.Writer, t *schema.ObjectType, details
 
 		for _, p := range t.Properties {
 			printComment(w, p.Comment, false)
-			outputType, applyType := pkg.outputType(p.Type, !p.IsRequired), pkg.plainType(p.Type, !p.IsRequired)
+			outputType, applyType := pkg.outputType(p.Type, true), pkg.plainType(p.Type, true)
```
pgavlin, line 515: The API looks good, but `p.Type` must be recorded as requiring a pointer output type or we may generate references to missing types. This is tracked by `typeDetails.ptrElement`, which is set here: https://github.com/pulumi/pulumi/blob/master/pkg/codegen/go/gen.go#L1081-L1085. That code should be rewritten to recurse when it sets `ptrElement`.

lukehoban: Indeed - that was indeed leading to references to types that didn't exist - good catch!

pkg/codegen/go/gen.go
```diff
@@ -1078,11 +1087,21 @@ func generatePackageContextMap(tool string, pkg *schema.Package, goInfo GoPackag
 			pkg := getPkg(t.Token)
 			pkg.types = append(pkg.types, t)
 
-			for _, p := range t.Properties {
-				if obj, ok := p.Type.(*schema.ObjectType); ok && !p.IsRequired {
-					getPkg(obj.Token).details(obj).ptrElement = true
+			// For any optional properties, we must generate a pointer type for the corresponding property type.
+			// In addition, if the optional property's type is itself an object type, we also need to generate pointer
+			// types corresponding to all of it's nested properties, as our accessor methods will lift `nil` into
+			// those nested types.
```
pgavlin, line 1093: I think you also need these changes [here](https://github.com/pulumi/pulumi/pull/4456/files#diff-b02dc03ec48c179b1b6609cd77eb12a2R1122-R1131)

lukehoban:
Here's an example of codegen differences in AWS.  

https://github.com/pulumi/pulumi-aws/compare/lukehoban/goaccessor#diff-1f1b5ecc107eb8868597adfb597e7e16

Manually confirmed  that this builds cleanly there.

Issue #4457, opened 2020-04-21
Existing Github Actions Workflow fails on missing --yes after upgrade to 2.0


amichel:
After upgrade to 2.0 existing workflows fail with:
**error: --yes must be passed in to proceed when running in non-interactive mode**
Current workaround is to pass **up --yes** in step args.
I think this should be passed in entrypoint.sh by default

PR #4458, opened 2020-04-21
Pass yes as a parameter to the actions entrypoint

stack72:
Fixes: #4457

This is required since 2.0

### DIFF
--- a/dist/actions/Dockerfile
+++ b/dist/actions/Dockerfile
@@ -25,6 +25,9 @@ COPY ./entrypoint.sh /usr/bin/pulumi-action
 WORKDIR /app
 VOLUME ["/app"]
 
+# We need to pass this environment variable as the confirmation to `--non-interactive` in the CLI
+ENV PULUMI_SKIP_CONFIRMATIONS=true
+
 # This image uses a thin wrapper over the Pulumi CLI as its entrypoint. As a result, you may run commands
 # simply by running `docker run pulumi/pulumi up` to run the program mounted in the `/app` volume location.
 ENTRYPOINT ["/usr/bin/pulumi-action", "--non-interactive"]


dist/actions/Dockerfile
```diff
@@ -27,4 +27,4 @@ VOLUME ["/app"]
 
 # This image uses a thin wrapper over the Pulumi CLI as its entrypoint. As a result, you may run commands
 # simply by running `docker run pulumi/pulumi up` to run the program mounted in the `/app` volume location.
-ENTRYPOINT ["/usr/bin/pulumi-action", "--non-interactive"]
+ENTRYPOINT ["/usr/bin/pulumi-action", "--non-interactive", "--yes"]
```
leezen, line 30: I don't think we can blanket pass in `--yes` since it's not accepted on all commands. You'd need to modify the `entrypoint.sh` to determine which command is being run and if it's one of `refresh`, `destroy`, `up`, `update`, or `new` pass in --yes.

chrsmith: Yeah, if we merged this as-is it would probably cause problems.

For reference, this is where we just pass all the arguments along to Pulumi.
https://github.com/pulumi/pulumi/blob/master/dist/actions/entrypoint.sh#L112-L118

Would it make more sense to change the logic in the CLI to better handle this situation? Since `--non-interactive` could reasonably be interpreted as "no seriously, just confirm everything". Or alternatively, a `PULUMI_FORCE=true` environment variable could be used to automatically bypass confirmation prompts (like `--yes` does).

Do either of those seem like good approaches?

Alternatively, we could just opportunistically look at the first argument to `pulumi` and like Lee was saying, automatically add `--yes` there.

Something like:

```bash
case $ARGS[0] in
    "up" | "update" | "refresh" | "destroy")
        # Automatically tack on --yes to confirm the command, since we are running on non-interactive
        # session.
       ARGS += "--yes"
    *)
        # Do nothing.
        ;;
esac
```

leezen: @chrsmith the CLI was actually changed from `--non-interactive` being an implicit 'yes' to requiring `--yes` in https://github.com/pulumi/pulumi/pull/3793 (and was a breaking change as part of `2.0`)

I think an env var might make sense, as also suggested in https://github.com/pulumi/pulumi/issues/4459

PR #4471, opened 2020-04-22
[codegen/go] Handle recursive types when marking optional properties to use Ptr types in Go

praneetloke:
An issue was introduced with https://github.com/pulumi/pulumi/pull/4456 that specifically affects the Kubernetes package due to the recursive nature of the `JSONSchemaProps` nested type. I have updated the test in the docs codegen to cover this scenario.

### DIFF
--- a/pkg/codegen/go/gen.go
+++ b/pkg/codegen/go/gen.go
@@ -1079,21 +1079,25 @@ func generatePackageContextMap(tool string, pkg *schema.Package, goInfo GoPackag
 	// In addition, if the optional property's type is itself an object type, we also need to generate pointer
 	// types corresponding to all of it's nested properties, as our accessor methods will lift `nil` into
 	// those nested types.
-	var markOptionalPropertyTypesAsRequiringPtr func(parent *schema.ObjectType, props []*schema.Property, parentOptional bool)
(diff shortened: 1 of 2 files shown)


pkg/codegen/go/gen.go
```diff
@@ -1079,12 +1079,17 @@ func generatePackageContextMap(tool string, pkg *schema.Package, goInfo GoPackag
...
-	markOptionalPropertyTypesAsRequiringPtr = func(props []*schema.Property, parentOptional bool) {
+	var markOptionalPropertyTypesAsRequiringPtr func(parent *schema.ObjectType, props []*schema.Property, parentOptional bool)
+	markOptionalPropertyTypesAsRequiringPtr = func(parent *schema.ObjectType, props []*schema.Property, parentOptional bool) {
 		for _, p := range props {
 			if obj, ok := p.Type.(*schema.ObjectType); ok && (!p.IsRequired || parentOptional) {
+				// Skip recursive recursive types. For example, JSONSchemaProps in the Kubernetes
+				// package can have properties with a type of JSONSchemaProps.
+				if parent != nil && parent.Token == obj.Token {
+					continue
+				}
```
lukehoban, line 1090: This is probably fine - but it might be more "correct" to pass a `seen` map into this recursive function and then check that.  That would ensure that if the recursion is two layers deep, it still gets caught.

PR #4473, opened 2020-04-22
Fix Python module names.

pgavlin:
In particular, ensure that they are keyword-safe. This affects
`aws:lambda:*` in particular: prior to these changes, we were generating
code into `pulumi_aws.lamdba`, which is not referencable due to its use
of the `lamdba` keyword. With these changes, we generate code into
`pulumi_aws.lambda_`.

There is also a small fix for multi-line non-formatted strings: these
strings do not need escaped braces.


pgavlin:
I've verified manually that this properly emits code into `lambda_` in `pulumi-aws`.

Issue #4459, opened 2020-04-21
Add env var to skip interactive steps in CLI commands


migueloller:
Following a conversation with @lukehoban on Slack, I'm suggesting a feature is added to the CLI that allows for skipping interactive steps—during CI, for example.

In Pulumi v1 the desired behavior was there but it was implicit. For example, running `pulumi up` in CI would behave as running `pulumi up --non-interactive --yes`. In Pulumi v2, this implicit behavior was removed, and for good reasons. Updates are potentially destructive and implicitly doing this is thus potentially dangerous.

That being said, sometimes it might be desirable to avoid interactive steps, and while this is configurable per command (i.e., `--non-interactive` and `--yes`), having a global way to do this would be a great QoL improvement.

My suggestion is to add support for an env var that explicitly turns on the behavior from v1. Here are some suggested variable names:
- `PULUMI_NON_INTERACTIVE`
- `PULUMI_ALWAYS_YES`

I might also be worth thinking if the behavior should provide additional functionality instead of just implementing v1's behavior as is. For example, maybe the new behavior also skips previews? This might be out of scope, though.

leezen:
Related https://github.com/pulumi/pulumi/issues/4457 and https://github.com/pulumi/pulumi/issues/4458

PR #4477, opened 2020-04-23
Add PULUMI_SKIP_CONFIRMATIONS env var

chrsmith:
Add a new environment variable `PULUMI_SKIP_CONFIRMATIONS` which will act as if you had added the `--yes` flag to applicable commands. This provides a global switch for disabling interactive prompts.

I can't say I'm especially happy with the design here, but I'm not sure what would look better since we are already treating `--non-interactive` to mean "don't prompt for input, but still require `--yes` to confirm intent".

What do you think? Does this seem like the right way to fix https://github.com/pulumi/pulumi/issues/4459 and related issues? Or is there a better or more appropriate way to disable prompting?

```
// DisableInteractive may be set to true in order to disable prompts. This is useful when running in a non-attended
// scenario, such as in continuous integration, or when using the Pulumi CLI/SDK in a programmatic way.
var DisableInteractive bool

// Interactive returns true if we should be running in interactive mode. That is, we have an interactive terminal
// session, interactivity hasn't been explicitly disabled, and we're not running in a known CI system.
func Interactive() bool {
	return !DisableInteractive && InteractiveTerminal() && !ciutil.IsCI()
}
```


lukehoban:
> Reading through the PR and issue, I think this change is inline with the intent we had (which is to remove implicitly assuming --yes). Today, we only have one way to explicitly say --yes which is via the args. Having an env var seems like a nice convenience for power users. The only negative here would be someone who sets the env var in a non-CI environment, forgets they did so, and gets unintended behavior. But this seems like the right balance to me.

> I'd love to get the opinions from the folks who were on the original PR to see if I'm missing anything in the reasoning above /cc @lukehoban @justinvp @pgavlin

Yes - that all seems in line with the original intent.  The reason for the change was just that the default here is dangerous for manual usage at a terminal (things like piping to a file inadvertently auto-accepting a prompt).  For CI use cases I think it's reasonable to opt-in to this auto-accept default, as they have more well-defined use cases encoded in the actual CI workflows.

pkg/cmd/pulumi/util.go
```diff
@@ -65,6 +65,16 @@ func useLegacyDiff() bool {
...
 }
 
+// skipConfirmations returns whether or not confirmation prompts should
+// be skipped. This should be used by pass any requirement that a --yes
+// parameter has been set for non-interactive scenarios.
+//
+// This should NOT be used to bypass protections for destructive
+// operations, such as those that will fail without a --force parameter.
+func skipConfirmations() bool {
+	return cmdutil.IsTruthy(os.Getenv("PULUMI_SKIP_CONFIRMATIONS"))
```
justinvp, line 75: Reminder to open issue/PR to update: https://github.com/pulumi/docs/blob/master/content/docs/reference/cli/environment-variables.md

stack72:
Going to merge this so we can proceed for the 2.2.0 release

pkg/cmd/pulumi/util.go
stack72: https://github.com/pulumi/docs/pull/3313

PR #4488, opened 2020-04-24
Anchoring properties

sean1588:


sean1588:
So.... this has been very flaky, as in some links work and others do not. I have fiddled with this for quite some time. Some of the anchors to properties work and some do not. These links and anchors I am generating on the html end. The other anchor links existing in the page are generated from the markdown side. 

Not sure if it is because the props are nested in that table structure has anything to do with it. Not sure if anything sticks out at you @cnunciato or @praneetloke.

Here is a sample of the source....
```

<dt class="property-optional" title="Optional">
   <span id="kmskeyid"><a href="#kmskeyid">Kms<wbr>Key<wbr>Id</a></span>
   <span class="property-type"><a href="https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/built-in-types">string</a></span>
</dt>
```

pkg/codegen/docs/gen.go
```diff
@@ -816,12 +817,14 @@ func (mod *modContext) getProperties(properties []*schema.Property, lang string,
 		}
 
 		docProperties = append(docProperties, property{
+			ID:                 strings.ToLower(propLangName),
```
justinvp, line 820: It's possible the same property name is going to show up on the same page (e.g. on the resource itself and on one of the supporting types). Should we prefix the type name to help ensure the id is unique?

sean1588: yea, I think that may be a good idea.

sean1588:
Ok. This should be working well now. I figured out the issue. There were some collisions with the prop names across some of the languages because they are all rendered on the screen whether they are shown or not. Silly me. So I just added the language to the prop id. Should be good now! 😄

lukehoban:
> So I just added the language to the prop id. 

A little unfortunate that we have to do that. Is there any way to avoid this (have the ids dynamically swap in/out based on language-choosers?) What does one of these anchors end up looking like? Are they friendly enough to cleanly share?

pkg/codegen/docs/gen.go
```diff
@@ -815,13 +816,17 @@ func (mod *modContext) getProperties(properties []*schema.Property, lang string,
 			propLangName = name
 		}
 
+		propID := strings.ToLower("prop" + lang + propLangName)
```
praneetloke, line 819: Is the `prop` prefix still needed, though?

sean1588:
> A little unfortunate that we have to do that. Is there any way to avoid this (have the ids dynamically swap in/out based on language-choosers?) 

Agreed @lukehoban, yeah I think that would be more ideal, will look to see what the level of effore would be for that. As an example this is what the anchor would look like: `/docs/reference/pkg/aws/s3/bucket/#gobucketname` or `/docs/reference/pkg/aws/s3/bucket/#pythonbucketname`, maybe I could add a delimeter like a hyphen or something to make it look a little better if we end up needing to go this route.

pkg/codegen/docs/gen.go
sean1588: Don't think this is still needed.

praneetloke:
> maybe I could add a delimeter like a hyphen or something to make it look a little better if we end up needing to go this route.

Yeah I was just going to suggest the same as well. I also think we should use the language as the suffix rather than as a prefix to the prop name. I think it would look cleaner (not sure if others agree, though) and also handle URLs without a language suffix, but still navigate to the property in the currently selected language.

Aside from this, I also think we might need some sort of client-side JS handling here as well, because if the URL being loaded is for a language that is currently _not_ the selected language, then we should change the language to _that_. Not sure what the behavior is today, but I don't think the language chooser is dynamic today, i.e. it does not react to information in the URL fragments. Also does the anchor tag would work when the DOM element is hidden (due to the language being something else)?

I suspect that because this is all just plain IDs on DOM elements it would almost require us to have some sort of JS to handle the URL magic. For example, @cnunciato recently introduced Stencil.js into our main site -- I wonder if something could be used there? In fact, the current language chooser component is a Stencil.js component. Perhaps, we could wire-up some logic in it to parse out the hash fragment to pre-select the language and then once the page is loaded, because the right language is selected the respective DOM elements for the properties would be visible, and so the routing would work properly.

sean1588:
> I think it would look cleaner (not sure if others agree, though) and also handle URLs without a language suffix, but still navigate to the property in the currently selected language.

I updated it to add the language as the suffix. I agree, definitely looks better. As far as dynamically generating the ids using the language chooser, I believe it we are re-using it on other pages, not just the resource docs page. Also, like I think you were pointing out, not sure how this would work as those elements would only exist based on the language that was selected, which is not routable up until this point as far as I know. It looks like we are just setting the language in local storage at the moment, so not sure how we would handle this.

pkg/codegen/docs/gen.go
```diff
@@ -825,13 +826,17 @@ func (mod *modContext) getProperties(properties []*schema.Property, lang string,
 			propLangName = name
 		}
 
+		propID := strings.ToLower(propLangName + "~" + lang)
```
praneetloke, line 829: I feel like this is an important detail with this separator. Perhaps we can make this a file-level constant and also add a comment about its role?

pkg/codegen/docs/gen.go
```diff
@@ -137,15 +137,16 @@ type header struct {
...
 	DisplayName        string
 	Name               string
 	Comment            string
 	Type               propertyType
 	DeprecationMessage string
-
-	IsRequired bool
-	IsInput    bool
+	Link               string
+	IsRequired         bool
```
praneetloke, line 148: Not a big deal but what do you think about restoring the new-line character before this property to separate the `bool`s?

pkg/codegen/docs/gen.go
```diff
@@ -137,15 +137,16 @@ type header struct {
 
 // property represents an input or an output property.
 type property struct {
+	ID string
```
praneetloke, line 140: Can you please add a comment for this property? Feel like it is an important detail as to how this is used in the template.

sean1588: good idea.

praneetloke:
Some of the builds failed due to an email change for the bot account. It should be ok to restart the failed builds now.

praneetloke:
Some of the builds failed due to an email change for the bot account. It should be ok to restart the failed builds now.

PR #4489, opened 2020-04-24
Order secret outputs in stack references

### LINKED ISSUE #4176: Secret Output Names of StackReferences are shown as different.
I have a Stack Reference 
```
const infraStack = new pulumi.StackReference(`infra-${env}`);
```

to a stack that exports secrets something like this:
```
exports.Secret1 = config.requireSecret('Secret1')
exports.Secret2 = config.requireSecret('Secret2')
exports.Secret3 = config.requireSecret('Secret3')
```

if i run `pulumi up` on the stack containing the reference i see often changes like this:
```
~ secretOutputNames: [
          ~ [0]: "Secret1" => "Secret2"
          ~ [1]: "Secret2" => "Secret1"
            [2]: "Secret3"
        ]
```

Eventhough the secret contents did not change. Also both stacks didn't change.
I can do a `pulumi refresh` multiple times and it keeps displaying this change in my project.

jaxxstorm:
When referencing `secretOutputNames` in from another stack, spurious
diffs can often be created because the secret output slice was not
ordered.

This PR orders the slice before it's added to the propertymap, ensuring
the order always remains the same

Fixes #4176

Before:

```
~ stack_one: [
      ~ [0]: "Alpha" => "Beta"
      ~ [1]: "Beta" => "Delta"
      ~ [2]: "Delta" => "Epsilon"
      ~ [3]: "Epsilon" => "Alpha"
    ]
```

After:

```
Outputs:
  ~ stack_one: [
      ~ [0]: "Delta" => "Alpha"
      ~ [1]: "Epsilon" => "Beta"
      ~ [2]: "Alpha" => "Delta"
      ~ [3]: "Beta" => "Epsilon"
    ]
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ CHANGELOG
 ## HEAD (unreleased)
 
 - Order secretOutputNames when used in stack references
-  [#4481](https://github.com/pulumi/pulumi/pull/4481)
+  [#4489](https://github.com/pulumi/pulumi/pull/4489)
 
 - Add support for a `PULUMI_CONSOLE_DOMAIN` environment variable to override the
   behavior for how URLs to the Pulumi Console are generated.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ CHANGELOG
 
 ## HEAD (unreleased)
 
+- Order secretOutputNames when used in stack references
+  [#4481](https://github.com/pulumi/pulumi/pull/4481)
+
 - Add support for a `PULUMI_CONSOLE_DOMAIN` environment variable to override the
   behavior for how URLs to the Pulumi Console are generated.
   [#4410](https://github.com/pulumi/pulumi/pull/4410)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,9 @@ CHANGELOG
 
 ## HEAD (unreleased)
 
+- Order secretOutputNames when used in stack references
+  [#4481](https://github.com/pulumi/pulumi/pull/4481)
```
justinvp, line 7:
```suggestion
  [#4489](https://github.com/pulumi/pulumi/pull/4489)
```

pkg/resource/deploy/builtins.go
```diff
@@ -209,6 +210,11 @@ func (p *builtinProvider) readStackReference(inputs resource.PropertyMap) (resou
 		}
 	}
 
+	// Sort the secret outputs
```
justinvp, line 213: Say why we're sorting? Maybe something like the following?

```suggestion
	// Sort the secret outputs so the order is deterministic, to avoid spurious diffs during updates.
```

jaxxstorm: Yep, this is much better!

PR #4490, opened 2020-04-24
[codegen/docs] Fix the generation of Function names, args, and result type names

praneetloke:
Related to https://github.com/pulumi/docs/issues/3075.

This PR uses the Go codegen's Function name overrides map to use the correct Function name, as well as its argument type name and result type name. For all other languages, I have made sure that the docs generator is using the same logic to generate the function name. This was done by factoring-out the logic used by each code gen into a separate function that the doc language helper interface can access.

~~(in the process of creating the docs PR; will post the link here once that is complete)~~
Docs PR: https://github.com/pulumi/docs/pull/3097


praneetloke:
I found an issue while looking at the docs preview. Going to convert this to draft mode and mark it ready once it is fixed.

praneetloke:
Alright. I have verified the diffs of the docs with the fix (the fix was to use the module name, not the Function's token, to lookup the relevant Go code gen package context) and it looks good now.

justinvp:
I added a comment to docs PR: https://github.com/pulumi/docs/pull/3097#discussion_r416170342

PR #4494, opened 2020-04-24
Retry bucket writes on error

### LINKED ISSUE #4258: Persistent 429 Rate limit errors from GCS bucket when using GCS State backend for Pulumi
Since yesterday (31 March 2020), we have been seeing persistent 429 rate limit errors with the GCS state backend for pulumi. See this error irrespective of the size of pulumi stacks (even on newly created stacks) and even when only a single object from the stack is being created or modified.  Also tried `pulumi up --parallel 1` for single threaded execution, but still see this error.

```
Diagnostics:
  gcp:kms:CryptoKeyIAMBinding (xxx-permissions):
    error: post-step event returned an error: failed to save snapshot: An IO error occurred during the current operation: blob (key ".pulumi/stacks/<stack-name>.json") (code=Unknown): googleapi: Error 429: The rate of change requests to the object <gcs-bucket-name>/.pulumi/stacks/<stack-name>.json exceeds the rate limit. Please reduce the rate of create, update, and delete requests., rateLimitExceeded

  pulumi:pulumi:Stack (<pulumi-project-name>-<stack-name>):
    error: update failed
```

Sometimes, the state file gets deleted in this process too when 429 errors are received, which is weird.

We may have to give up on using GCS buckets entirely for storing pulumi state. Does anybody know about what could be causing this issue or any workarounds? Thanks.

jaxxstorm:
When writing the snapshot to the filestate bucket, we can retry in the
event of an error, which helps users who are experiencing issues around
write rates to GCS

Fixes #4258


pkg/backend/filestate/state.go
```diff
@@ -185,7 +186,28 @@ func (b *localBackend) saveStack(name tokens.QName, snap *deploy.Snapshot, sm se
 
 	// And now write out the new snapshot file, overwriting that location.
 	if err = b.bucket.WriteAll(context.TODO(), file, byts, nil); err != nil {
-		return "", errors.Wrap(err, "An IO error occurred while writing the new snapshot file")
+
+		b.mutex.Lock()
+		defer b.mutex.Unlock()
+
+		// Retry the write 10 times in case of upstream bucket errors
+		_, _, err = retry.Until(context.TODO(), retry.Acceptor{
```
leezen, line 194: I think you'll need to tune the `Delay`, `Backoff`, and `MaxDelay` options on this to match the GCS expectations. Maybe something like 1s, backoff of 1.2, max 30s?

pkg/backend/filestate/state.go
```diff
@@ -185,7 +186,28 @@ func (b *localBackend) saveStack(name tokens.QName, snap *deploy.Snapshot, sm se
...
+			Accept: func(try int, nextRetryTime time.Duration) (bool, interface{}, error) {
+				// And now write out the new snapshot file, overwriting that location.
+				err := b.bucket.WriteAll(context.TODO(), file, byts, nil)
+				if err != nil {
+					logging.V(7).Infof("Error while writing snapshot to: %s (attempt=%d, error=%s)", file, try, err)
```
leezen, line 199: Can't we check if this is a 429 and only retry then? I thought the error is wrapped and bubbled up here, in which case we can still inspect the message?

pkg/backend/filestate/state.go
jaxxstorm: Yep, my intention is to tune this quite significantly. I wanted to get a feature branch out so that affected users could test the patch first, then we'll get ready for release from there. Does that sound like a reasonable approach?

pkg/backend/filestate/state.go
jaxxstorm: I thought about this as well, but I actually think this could improve the experience in other situations such as transient network errors or `5xx` errors during provider outages. Thoughts? Assuming we're reasonable about our backoff/retry I think it's a good idea to retry in general

leezen: Good point -- that makes sense -- though, I'm not sure there's a one-sized fits all in terms of the retry frequency and backoff for `429` vs. `5xx` at that point. Also, this will end up retrying on other `400s` where we really shouldn't retry at all. So I'd probably still prefer seeing some logic around what errors we actually retry against.

pkg/backend/filestate/state.go
jaxxstorm: I've added this as suggested!

pkg/backend/filestate/state.go
jaxxstorm: So I went down this path and looked at unwrapping the error. Unfortunately, the go-cloud upstream library here doesn't make this easy.

One option (which I think is reasonable to pursue) was to update the `ErrorCode` method to check the `http` status code from google cloud. I've added this to our fork here:
https://github.com/pulumi/go-cloud/blob/83fac77233017cefe3a2c2aee58e457d7ea34c4a/blob/gcsblob/gcsblob.go#L262-L263

I could also unwrap the error completely and check for `googleapi.Error.Code` but that begs the question - should we do this for all bucket writes? What about other cloud providers?

I almost started considering creating my own unwrap function, but after a discussion with @leezen I think it's best if we just retry here regardless of the response code, I don't think the UI is affected that much. With that in mind, I've left it as it is

PR #4506, opened 2020-04-28
Don't try decrypt values on community PRs

### LINKED ISSUE #4508: Community PRs fail because of gcloud credential encryption
Currently, we have an `before_install` script that decrypts our gcloud credentials

We should check the source of the PR and then skip this step, it'll help with community PRs

jaxxstorm:
When a community member sends a pull request to the repo, travis can't
decrypt some environment variables which include decrypting the GCP
credentials.
In addition to this, there are a few tests that run that require the
`PULUMI_ACCESS_TOKEN` which can't be used.

This PR does 2 things:

- Adds a check to the GCP credentials decryption which checks the source
  of the PR (whether it's from within the local repo, or a fork)
- Sets an environment variable: `COMMUNITY_PRS` which is used downstream
  during some tests. If the var exists, some tests are skipped

This should enhance the experience for community pull requests
significantly

Fixes #4508

For an example of this in action, see #4507

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,16 @@ sudo: true
 git:
   depth: false
 before_install:
-- openssl aes-256-cbc -K $encrypted_342342ee5e49_key -iv $encrypted_342342ee5e49_iv
-  -in gcp-credentials.json.enc -out gcp-credentials.json -d
+- |
+  if [ "$TRAVIS_PULL_REQUEST_SLUG" != "pulumi/pulumi" ]; 
+  then 
+    echo "Not decrypting gcp credentials";
+    export COMMUNITY_PR="true"
+  else 
+    echo "Decrypting gcp credentials"
+    openssl aes-256-cbc -K $encrypted_342342ee5e49_key -iv $encrypted_342342ee5e49_iv -in gcp-credentials.json.enc -out gcp-credentials.json -d
+    export COMMUNITY_PR="false"
+  fi
 - git clone https://github.com/pulumi/scripts ${GOPATH}/src/github.com/pulumi/scripts
 - source ${GOPATH}/src/github.com/pulumi/scripts/ci/prepare-environment.sh
 - source ${PULUMI_SCRIPTS}/ci/keep-failed-tests.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,12 +18,10 @@ before_install:
 - |
   if [ "$TRAVIS_PULL_REQUEST_SLUG" != "pulumi/pulumi" ]; 
   then 
-    echo "Not decrypting gcp credentials";
-    export COMMUNITY_PR="true"
+    echo "Community PR - Not decrypting gcp credentials";
   else 
     echo "Decrypting gcp credentials"
     openssl aes-256-cbc -K $encrypted_342342ee5e49_key -iv $encrypted_342342ee5e49_iv -in gcp-credentials.json.enc -out gcp-credentials.json -d
-    export COMMUNITY_PR="false"
   fi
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -15,8 +15,16 @@ sudo: true
 git:
   depth: false
 before_install:
-- openssl aes-256-cbc -K $encrypted_342342ee5e49_key -iv $encrypted_342342ee5e49_iv
-  -in gcp-credentials.json.enc -out gcp-credentials.json -d
+- |
+  if [ "$TRAVIS_PULL_REQUEST_SLUG" != "pulumi/pulumi" ]; 
+  then 
+    echo "Not decrypting gcp credentials";
+    export COMMUNITY_PR="true"
```
stack72, line 22: do we need to export an environment variable? The tests won't fail if it's a community PR - the tests will skip due to not being able to read PULUMI_ACCESS_TOKEN

jaxxstorm: Some of the tests aren't skipped even if that value isn't set, see here:

https://travis-ci.com/github/pulumi/pulumi/builds/162335371#L3290

I'm happy to set the skip to be if `PULUMI_ACCESS_TOKEN` isn't set if we think that's better?

pkg/cmd/pulumi/new_test.go
```diff
@@ -807,6 +807,10 @@ func removeStack(t *testing.T, name string) {
 }
 
 func skipIfShort(t *testing.T) {
+	_, ok := os.LookupEnv("PULUMI_ACCESS_TOKEN")
+	if !ok {
+		t.Skipf("Skipping: PULUMI_ACCESS_TOKEN is not set")
+	}
```
lukehoban, line 813: It's not immediately clear to me why this is right to do in `skipIfShort` - how is this check related to whether the tests are running in "short" mode?

Perhaps just the name of this function needs to change to `skipIfShortOrNoPulumiAccessToken`?

pkg/npm/npm_test.go
```diff
@@ -33,6 +33,10 @@ func TestYarnInstall(t *testing.T) {
 }
 
 func testInstall(t *testing.T, expectedBin string) {
+	// Skip for community PRs
+	if os.Getenv("COMMUNITY_PR") == "true" {
+		t.Skip("Skipped for community PRs")
+	}
```
lukehoban, line 39: Related to the other feedback - it would be great if we could keep the codebase itself independt of the notion of "community PR" (which is really a test/CI construct).  Ideally the tests refer to semantically meaningful environment variables or parameterization - and the CI environment maps things that make sense at that layer into these in a way that makes sense.

What exactly is it that this test depends on?  Can we make the skip behaviour here more directly tied to that?

.travis.yml
jaxxstorm: I've switched this to use `PULUMI_ACCESS_TOKEN` instead

pkg/npm/npm_test.go
jaxxstorm: It's tied to the `PULUMI_ACCESS_TOKEN`, as per the other feedback, which I think makes more sense. Does that make more sense?

pkg/cmd/pulumi/new_test.go
jaxxstorm: It's probably not, I was really just trying to take advantage of the existing function being available.

I've updated the name as suggested

pkg/npm/npm_test.go
jaxxstorm: Actually, I'm not even sure why I added this, I've removed it.

PR #4509, opened 2020-04-28
Enable container runtime tests for cron jobs

chrsmith:
This PR adds a new `make test_containers_cron` target that will execute as part of Travis CI's Cron jobs.  (Which are [configured](https://travis-ci.com/github/pulumi/pulumi/settings) to run on `master` once a day.)

Running the container tests locally works fine on my machine, but it's likely there are still problems with the test that needs to be worked out when ran on Linux. (i.e. #4136) But with this PR, it will be much easier to (1) confirm these tests are still failing on CI and if so, (2) debug/fix them.

I'll keep an eye out on the #builds channel and Travis CI to make sure these start passing again. Let me know what you think about this approach.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,9 +76,21 @@ publish_packages:
 	$(call STEP_MESSAGE)
 	./scripts/publish_packages.sh
 
+# Run the integration tests for our DockerHub containers. We do so only via the
+# "Travis Cron" job type, because (1) the tests can only be ran _after_ we publish
+# the current SDK version, since it is required by the Docker build. And (2) the
+# tests (currently) aren't reliable enough to run as part of every push to master.
+#
+# So instead we run the ~daily on master. Where we know the current SDK version
+# will have been published.
+.PHONY: test_containers_cron
+test_containers:
+	$(call STEP_MESSAGE)
+	./scripts/build-docker.sh $$(./scripts/get-version HEAD) --test
+
 # The travis_* targets are entrypoints for CI.
 .PHONY: travis_cron travis_push travis_pull_request travis_api
-travis_cron: all
+travis_cron: all test_containers_cron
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -76,9 +76,21 @@ publish_packages:
...
+# "Travis Cron" job type, because (1) the tests can only be ran _after_ we publish
+# the current SDK version, since it is required by the Docker build. And (2) the
+# tests (currently) aren't reliable enough to run as part of every push to master.
+#
+# So instead we run the ~daily on master. Where we know the current SDK version
+# will have been published.
+.PHONY: test_containers_cron
+test_containers:
+	$(call STEP_MESSAGE)
+	./scripts/build-docker.sh $$(./scripts/get-version HEAD) --test
```
stack72, line 89:
```suggestion
	./scripts/build-docker.sh ${VERSION} --test
```

Passing ${VERSION} here should work

Makefile
```diff
@@ -76,9 +76,21 @@ publish_packages:
...
+
 # The travis_* targets are entrypoints for CI.
 .PHONY: travis_cron travis_push travis_pull_request travis_api
-travis_cron: all
+travis_cron: all test_containers_cron
```
stack72, line 93: Is there any reason we only want this on cron? Now that it's only doing a test, then we should be able to put this in `make_push` and `make_pull_request` as well?

chrsmith: Yes, we do want to run this for `push` and `pull_request` jobs too. But currently the tests have proven to be flaky when ran on Travis. (I cannot repro the specific issues on my Mac.) So the plan is to just run them for cron jobs, root out any stability issues, and then graduate them to run more regularly after that.

Makefile
chrsmith: Nice catch.

PR #4513, opened 2020-04-28
Include case when attempting to escape the `.` in a name.

danfhernandez:
https://github.com/pulumi/docs/issues/2917

### DIFF
--- a/pkg/cmd/pulumi/config.go
+++ b/pkg/cmd/pulumi/config.go
@@ -257,10 +257,12 @@ func newConfigSetCmd(stack *string) *cobra.Command {
 			"If a value is not present on the command line, pulumi will prompt for the value. Multi-line values\n" +
 			"may be set by piping a file to standard in.\n\n" +
 			"The `--path` flag can be used to set a value inside a map or list:\n\n" +
-			"    - `pulumi config set --path outer.inner value` " +
-			"will set the value of `outer` to a map `inner: value`.\n" +
 			"    - `pulumi config set --path names[0] a` " +
-			"will set the value to a list with the first item `a`.",
+			"will set the value to a list with the first item `a`.\n" +
+			"    - `pulumi config set --path parent.nested value` " +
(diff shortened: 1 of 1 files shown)


pkg/cmd/pulumi/config.go
```diff
@@ -260,7 +260,9 @@ func newConfigSetCmd(stack *string) *cobra.Command {
 			"    - `pulumi config set --path outer.inner value` " +
 			"will set the value of `outer` to a map `inner: value`.\n" +
 			"    - `pulumi config set --path names[0] a` " +
-			"will set the value to a list with the first item `a`.",
+			"will set the value to a list with the first item `a`.\n" +
+		        "    - `pulumi config set --path '[\"outer.name\"].[\"inner.name\"]' value`" +
+		        "will set the of `outer.name` to a map `inner.name: value`. Useful if the names of outer and inner include a `.`.",
```
praneetloke, line 265:
```suggestion
		        "will set the value of `outer.name` to a map `inner.name: value`. Useful if the names of outer and inner include a `.`.",
```

pkg/cmd/pulumi/config.go
```diff
@@ -260,7 +260,9 @@ func newConfigSetCmd(stack *string) *cobra.Command {
 			"    - `pulumi config set --path outer.inner value` " +
 			"will set the value of `outer` to a map `inner: value`.\n" +
 			"    - `pulumi config set --path names[0] a` " +
-			"will set the value to a list with the first item `a`.",
+			"will set the value to a list with the first item `a`.\n" +
+		        "    - `pulumi config set --path '[\"outer.name\"].[\"inner.name\"]' value`" +
+		        "will set the value of `outer.name` to a map `inner.name: value`. Useful if the names of outer and inner include a `.`.",
```
lukehoban, line 265: I think we've aimed to hard-wrap lines in this CLI help text output at 100 characters to ensure they look "good" in most CLI contexts.

PR #4518, opened 2020-04-28
Allow `pulumi.export` calls from unit tests

### LINKED ISSUE #4478: Python tests fail on exports
Not sure if this is a Pulumi bug or just lack of documentation...?

I'm following this tutorial: https://www.pulumi.com/docs/guides/testing/unit/
But adding `pulumi.export` to the stack causes the tests to fail hard.

Is there a solution for this?

```
(.venv) cmclaughlin-local@cm in /tmp
$ git clone https://github.com/pulumi/examples.git
Cloning into 'examples'...
remote: Enumerating objects: 86, done.
remote: Counting objects: 100% (86/86), done.
remote: Compressing objects: 100% (70/70), done.
Receiving objects: 100% (8466/8466), 9.34 MiB | 71.00 KiB/s, done.
remote: Total 8466 (delta 26), reused 37 (delta 14), pack-reused 8380
Resolving deltas: 100% (5094/5094), done.

(.venv) cmclaughlin-local@cm in /tmp
$ cd examples/testing-unit-py/

(.venv) cmclaughlin-local@cm in /tmp/examples/testing-unit-py
on master
$ echo "pulumi.export('hello', 'world')" >> infra.py

(.venv) cmclaughlin-local@cm in /tmp/examples/testing-unit-py
on master
$ python -m unittest
E
======================================================================
ERROR: test_ec2 (unittest.loader._FailedTest)
----------------------------------------------------------------------
ImportError: Failed to import test module: test_ec2
  File "/Users/cmclaughlin-local/.pyenv/versions/3.7.4/lib/python3.7/unittest/loader.py", line 436, in _find_test_path
    module = self._get_module_from_name(name)
  File "/Users/cmclaughlin-local/.pyenv/versions/3.7.4/lib/python3.7/unittest/loader.py", line 377, in _get_module_from_name
    __import__(name)
  File "/private/tmp/examples/testing-unit-py/test_ec2.py", line 13, in <module>
    import infra
  File "/private/tmp/examples/testing-unit-py/infra.py", line 20, in <module>
    pulumi.export('hello', 'world')
  File "/Users/cmclaughlin-local/repos/shopstyle/pulumi/stacks/pypy/.venv/lib/python3.7/site-packages/pulumi-2.1.0a1587474541-py3.7.egg/pulumi/resource.py", line 884, in export
    raise Exception("Failed to export output. Root resource is not an instance of 'Stack'")
Exception: Failed to export output. Root resource is not an instance of 'Stack'


----------------------------------------------------------------------
Ran 1 test in 0.000s

FAILED (errors=1)
```

justinvp:
This change allows importing modules with calls to `pulumi.export` in unit tests.

I'm not certain this is the best way to address this. Also, we could decide to simply not make this change at all, and just require structuring the program in a way that avoids `pulumi.export` calls from being imported in tests (as is the case today).

## Before

Previously, you'd have to structure the Python program in a way that avoids the `pulumi.export` from being called from unit tests, such as:

```python
# __main__.py

import pulumi

import infra

pulumi.export('group', infra.group)
pulumi.export('server', infra.server)
pulumi.export('publicIp', infra.server.public_ip)
pulumi.export('publicHostName', infra.server.public_dns)
```

```python
# infra.py

import pulumi
from pulumi_aws import ec2

group = ec2.SecurityGroup('web-secgrp', ingress=[
    # Uncomment to fail a test:
    #{ "protocol": "tcp", "from_port": 22, "to_port": 22, "cidr_blocks": ["0.0.0.0/0"] },
    { "protocol": "tcp", "from_port": 80, "to_port": 80, "cidr_blocks": ["0.0.0.0/0"] },
])

user_data = '#!/bin/bash echo "Hello, World!" > index.html nohup python -m SimpleHTTPServer 80 &'

server = ec2.Instance('web-server-www;',
    instance_type="t2.micro",
    security_groups=[ group.name ], # reference the group object above
    # Comment out to fail a test:
    tags={'Name': 'webserver'},     # name tag
    ami="ami-c55673a0")             # AMI for us-east-2 (Ohio)
    # Uncomment to fail a test:
    #user_data=user_data)           # start a simple web server
```

```python
# test_ec2.py

import unittest
import pulumi

class MyMocks(pulumi.runtime.Mocks):
    def new_resource(self, type_, name, inputs, provider, id_):
        return [name + '_id', inputs]
    def call(self, token, args, provider):
        return {}

pulumi.runtime.set_mocks(MyMocks())

# Now actually import the code that creates resources, and then test it.
import infra

class TestingWithMocks(unittest.TestCase):
    # Test if the service has tags and a name tag.
    @pulumi.runtime.test
    def test_server_tags(self):
        def check_tags(args):
            urn, tags = args
            self.assertIsNotNone(tags, f'server {urn} must have tags')
            self.assertIn('Name', tags, 'server {urn} must have a name tag')

        return pulumi.Output.all(infra.server.urn, infra.server.tags).apply(check_tags)
```

This is the approach we use today in our [example](https://github.com/pulumi/examples/tree/master/testing-unit-py) and [test](https://github.com/pulumi/pulumi/tree/master/sdk/python/lib/test/langhost/testing_with_mocks).

## After

With this change, you can now call `pulumi.export` directly from `infra.py`:

```python
# __main__.py

import infra
```

```python
# infra.py

import pulumi
from pulumi_aws import ec2

group = ec2.SecurityGroup('web-secgrp', ingress=[
    # Uncomment to fail a test:
    #{ "protocol": "tcp", "from_port": 22, "to_port": 22, "cidr_blocks": ["0.0.0.0/0"] },
    { "protocol": "tcp", "from_port": 80, "to_port": 80, "cidr_blocks": ["0.0.0.0/0"] },
])

user_data = '#!/bin/bash echo "Hello, World!" > index.html nohup python -m SimpleHTTPServer 80 &'

server = ec2.Instance('web-server-www;',
    instance_type="t2.micro",
    security_groups=[ group.name ], # reference the group object above
    # Comment out to fail a test:
    tags={'Name': 'webserver'},     # name tag
    ami="ami-c55673a0")             # AMI for us-east-2 (Ohio)
    # Uncomment to fail a test:
    #user_data=user_data)           # start a simple web server

pulumi.export('group', group)
pulumi.export('server', server)
pulumi.export('publicIp', server.public_ip)
pulumi.export('publicHostName', server.public_dns)
```

Fixes #4478

### DIFF
--- a/sdk/python/lib/pulumi/runtime/mocks.py
+++ b/sdk/python/lib/pulumi/runtime/mocks.py
@@ -26,7 +26,7 @@ from . import rpc
 from .settings import Settings, configure, get_stack, get_project
 from .sync_await import _sync_await
 from ..runtime.proto import engine_pb2, engine_pb2_grpc, provider_pb2, resource_pb2, resource_pb2_grpc
-from ..runtime.stack import run_pulumi_func
+from ..runtime.stack import _run_test
 from ..output import Output
 
 if TYPE_CHECKING:
@@ -35,7 +35,7 @@ if TYPE_CHECKING:
 
 def test(fn):
     def wrapper(*args, **kwargs):
-        _sync_await(run_pulumi_func(lambda: _sync_await(Output.from_input(fn(*args, **kwargs)).future())))
+        _sync_await(_run_test(lambda: _sync_await(Output.from_input(fn(*args, **kwargs)).future())))
     return wrapper
 
 class Mocks(ABC):
@@ -105,13 +105,17 @@ class MockMonitor:
         return resource_pb2.ReadResourceResponse(urn=urn, properties=props_proto)
 
     def RegisterResource(self, request):
+        urn = self.make_urn(request.parent, request.type, request.name)
+
+        if request.type == "pulumi:pulumi:Stack":
+            return resource_pb2.RegisterResourceResponse(urn=urn)
+
         inputs = rpc.deserialize_properties(request.object)
 
         id_, state = self.mocks.new_resource(request.type, request.name, inputs, request.provider, request.importId)
 
         obj_proto = _sync_await(rpc.serialize_properties(state, {}))
 
-        urn = self.make_urn(request.parent, request.type, request.name)
         return resource_pb2.RegisterResourceResponse(urn=urn, id=id_, object=obj_proto)
 
     def RegisterResourceOutputs(self, request):
--- a/sdk/python/lib/test/langhost/testing_with_mocks/resources.py
+++ b/sdk/python/lib/test/langhost/testing_with_mocks/resources.py
@@ -29,3 +29,6 @@ myinstance = Instance("instance",
                       name="myvm",
                       value=pulumi.Output.secret("secret_value"))
 invoke_result = do_invoke()
+
+
+pulumi.export("instance_public_ip", myinstance.public_ip)
--- a/sdk/python/lib/test/langhost/testing_with_mocks/resources.py
+++ b/sdk/python/lib/test/langhost/testing_with_mocks/resources.py
@@ -30,5 +30,5 @@ myinstance = Instance("instance",
                       value=pulumi.Output.secret("secret_value"))
 invoke_result = do_invoke()
 
-
-pulumi.export("instance_public_ip", myinstance.public_ip)
+pulumi.export("outprop", mycomponent.outprop)
+pulumi.export("public_ip", myinstance.public_ip)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,9 @@ CHANGELOG
 - Increase the MaxCallRecvMsgSize for all RPC calls. 
   [#4455](https://github.com/pulumi/pulumi/pull/4455)
 
+- Allow `pulumi.export` calls from Python unit tests.
+  [#4518](https://github.com/pulumi/pulumi/pull/4518)
+
 ## 2.0.0 (2020-04-16)
 
 - CLI behavior change.  Commands in non-interactive mode (i.e. when `pulumi` has its output piped to
--- a/sdk/python/lib/pulumi/runtime/stack.py
+++ b/sdk/python/lib/pulumi/runtime/stack.py
@@ -80,6 +80,16 @@ async def run_in_stack(func: Callable):
     """
     await run_pulumi_func(lambda: Stack(func))
 
+async def _run_test(func: Callable):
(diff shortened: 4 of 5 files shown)


sdk/python/lib/pulumi/runtime/mocks.py
```diff
@@ -105,13 +105,17 @@ def ReadResource(self, request):
         return resource_pb2.ReadResourceResponse(urn=urn, properties=props_proto)
 
     def RegisterResource(self, request):
+        urn = self.make_urn(request.parent, request.type, request.name)
+
+        if request.type == "pulumi:pulumi:Stack":
+            return resource_pb2.RegisterResourceResponse(urn=urn)
```
justinvp, line 111: This is consistent with what we do for Go:

https://github.com/pulumi/pulumi/blob/7b17463031d3e4a7d31764c8d5811985855820c1/sdk/go/pulumi/mocks.go#L108-L112

Interestingly, C# does not have a check like this. Currently, `IMocks.NewResourceAsync` is called for `"pulumi:pulumi:Stack"`. I think this is a bug that we should fix to be consistent with the other languages. If you agree, I'll open an issue.

mikhailshilkov: Yes, I think you are right

sdk/python/lib/test/langhost/testing_with_mocks/resources.py
```diff
@@ -29,3 +29,6 @@ def do_invoke():
                       name="myvm",
                       value=pulumi.Output.secret("secret_value"))
 invoke_result = do_invoke()
+
+
```
mikhailshilkov, line 33: Uber-nit: two empty lines

mikhailshilkov:
FWIW, I don't see a good reason to keep this weird behavior of not allowing exports. As a user, I would expect to be able to test any program in its fullness.

sdk/python/lib/test/langhost/testing_with_mocks/resources.py
justinvp: Fixed. I added a commit that moves the `pulumi.export` calls from `__main__.py` here, and fixed up the two empty lines.

sdk/python/lib/pulumi/runtime/mocks.py
justinvp: https://github.com/pulumi/pulumi/issues/4524

PR #4521, opened 2020-04-29
Support map-typed inputs in Go SDK RegisterResource

lukehoban:
Adds support for `RegisterResource` to accept map-typed implementations if `Input` as well as the existing struct-typed implementations.  Currently these must be fully untyped - but both `map[string]pulumi.Input` and `map[string]interface{}` are allowed.  In the future, it's plausible that a mode where the data itself is a map, but the `ElementType` implementation returns a struct could be supported, with the struct used to provide type information over the untyped map.

Two examples of using this:

```go
var vpc ec2.Vpc
args := map[string]pulumi.Input{"cidrBlock": pulumi.String("10.0.0.0/16"), "enableDnsSupport": pulumi.Bool(false)}
err := ctx.RegisterResource("aws:ec2/vpc:Vpc", "myvpc", pulumi.Map(args), &vpc)
if err != nil {
	return err
}
ctx.Export("id", vpc.OwnerId)
```

```go

type UntypedArgs map[string]interface{}

func (UntypedArgs) ElementType() reflect.Type {
	return reflect.TypeOf((*map[string]interface{})(nil)).Elem()
}

...

var vpc2 ec2.Vpc
args2 := map[string]interface{}{"cidrBlock": "10.0.0.0/16", "enableDnsSupport": false}
err = ctx.RegisterResource("aws:ec2/vpc:Vpc", "myvpc2", UntypedArgs(args2), &vpc)
if err != nil {
	return err
}
ctx.Export("id2", vpc2.OwnerId)
```

Fixes https://github.com/pulumi/pulumi/issues/4522.


sdk/go/pulumi/rpc.go
```diff
@@ -122,12 +95,65 @@ func marshalInputs(props Input) (resource.PropertyMap, map[string][]URN, []URN,
...
+			if tag == "" {
+				continue
+			}
+			err := marshalProperty(tag, pv.Field(i).Interface(), destField.Type)
+			if err != nil {
+				return nil, nil, nil, err
+			}
+		}
+	case reflect.Map:
+		if rt.Key().Kind() != reflect.String {
```
pgavlin, line 144: Let's move this check up to the check in `context.go`

sdk/go/pulumi/rpc.go
```diff
@@ -122,12 +95,65 @@ func marshalInputs(props Input) (resource.PropertyMap, map[string][]URN, []URN,
...
+			return nil, nil, nil, fmt.Errorf("cannot marhsal Input that is map with non-string keys, saw type %s", rt.String())
```
pgavlin, line 145:
```suggestion
			return nil, nil, nil, fmt.Errorf("cannot marshal Input that is map with non-string keys, saw type %s", rt.String())
```

PR #4527, opened 2020-04-29
Don't call IMocks.NewResourceAsync for the root stack resource

### LINKED ISSUE #4524: IMocks.NewResourceAsync is called for the "pulumi:pulumi:Stack"
Currently, `IMocks.NewResourceAsync` is called for `"pulumi:pulumi:Stack"`, which isn't how the other languages behave.

We should add a check like the one being added for Python in #4518 and already have for Go:

https://github.com/pulumi/pulumi/blob/7b17463031d3e4a7d31764c8d5811985855820c1/sdk/go/pulumi/mocks.go#L108-L112

justinvp:
Do not call `IMocks.NewResourceAsync` for the root stack resource. Also add a basic mocks test. (And fix a typo I noticed while working on this.)

Fixes #4524

### DIFF
--- a/sdk/dotnet/Pulumi.Tests/Mocks/MocksTests.cs
+++ b/sdk/dotnet/Pulumi.Tests/Mocks/MocksTests.cs
@@ -1,0 +1,69 @@
+// Copyright 2016-2020, Pulumi Corporation
+
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using System.Threading.Tasks;
+using Pulumi.Testing;
+using Xunit;
+
+namespace Pulumi.Tests.Mocks
+{
+    class MyMocks : IMocks
+    {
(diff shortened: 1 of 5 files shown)


sdk/dotnet/Pulumi.Tests/Mocks/MocksTests.cs
```diff
@@ -0,0 +1,69 @@
...
+
+            switch (type)
+            {
+                case "aws:ec2/instance:Instance":
+                    return Task.FromResult<(string, object)>(("i-1234567890abcdef0", new Dictionary<string, object> {
+                        { "publicIp", "203.0.113.12" },
+                    }));
+
+                default:
+                    return Task.FromResult<(string, object)>(("", new Dictionary<string, object>()));
```
mikhailshilkov, line 31: Should we fail by default if anything except the Instance resource is newed-up? That seems like a stronger test.

PR #4541, opened 2020-04-30
Makefile: properly encode dependencies

pgavlin:
This allows for the use of `make -j`, which speeds up a full build
dramatically at the cost of rather incomprehensible logs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,13 +45,13 @@ generate::
 	echo "Generate static assets bundle for docs generator"
 	cd pkg && go generate ./codegen/docs/gen.go
 
-build::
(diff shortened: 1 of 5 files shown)


pgavlin:
On my machine, `make -j` runs in ~5 minutes as compared to ~11 minutes for `make`. The build logs are absolute garbage--and my machine is mostly unusable for the first 30 seconds or so--but it's worth the tradeoff.

pgavlin:
(note that we should _not_ use `make -j` in CI without additional work due to the issues it causes with interleaved logs)

Issue #4545, opened 2020-05-01
Pulumi.Tests.Mocks.MocksTests.TestCustom failing in master


justinvp:
```
4665A total of 1 test files matched the specified pattern.
4666[xUnit.net 00:00:01.66]     Pulumi.Tests.Mocks.MocksTests.TestCustom [FAIL]
4667  X Pulumi.Tests.Mocks.MocksTests.TestCustom [4ms]
4668  Error Message:
4669   System.NotSupportedException : Multiple executions of TestAsync must run serially. Please configure your unit test suite to run tests one-by-one.
4670  Stack Trace:
4671     at Pulumi.Deployment.TestAsync[TStack](IMocks mocks, TestOptions options) in /Users/travis/gopath/src/github.com/pulumi/pulumi/sdk/dotnet/Pulumi/Deployment/Deployment_Run.cs:line 127
4672   at Pulumi.Tests.Mocks.MocksTests.TestCustom() in /Users/travis/gopath/src/github.com/pulumi/pulumi/sdk/dotnet/Pulumi.Tests/Mocks/MocksTests.cs:line 33
4673--- End of stack trace from previous location where exception was thrown ---
4674
4675Test Run Failed.
4676Total tests: 86
4677     Passed: 85
4678     Failed: 1
4679 Total time: 3.6855 Seconds
```

justinvp:
Still seeing this in master.

PR #4547, opened 2020-05-01
.NET: Disable parallel xUnit tests

justinvp:
Despite having the `[assembly: CollectionBehavior(DisableTestParallelization = true)]` attribute, it appears `dotnet test` is still running tests in parallel. To address, use a configuration file to disable parallelization.

Based on: https://github.com/xunit/xunit/issues/1169#issuecomment-287521557

Fixes https://github.com/pulumi/pulumi/issues/4545 🤞

### DIFF
--- a/sdk/dotnet/Pulumi.Tests/Pulumi.Tests.csproj
+++ b/sdk/dotnet/Pulumi.Tests/Pulumi.Tests.csproj
@@ -18,4 +18,9 @@
     <ProjectReference Include="..\Pulumi\Pulumi.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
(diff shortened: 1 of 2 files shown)


justinvp:
> But I'm confused at why that attribute isn't working like it should.

I am confused by this as well. Despite the attribute, I've observed tests running in parallel locally and specifying the configuration file seems to fix it.

https://xunit.net/docs/running-tests-in-parallel.html does mention:

> Runners are also allowed to override some of the behavior within a test framework (like number of threads, whether an assembly should run tests within itself in parallel, etc.); when they do so, it overrides whatever behavior has been otherwise specified (via code or configuration).

I am not sure sure why `dotnet test` is overriding the attribute, but respecting the configuration.

PR #4553, opened 2020-05-02
Fixed ResourceOption deepcopy bug in stack_reference.py

solidDoWant:
Replaced copy statements with `ResourceOptions.merge` in `stack_reference.py`. If a set of `ResourceOptions` is passed with `parent` set then the resource creation fails due to `deepcopy` failing.

Minimum code to reproduce:
```
from typing import Optional
from pulumi import ResourceOptions, StackReference, ComponentResource

class SomeComponent(ComponentResource):
    def __init__(self, name: str, stack_name: str, opts: Optional[ResourceOptions] = None):
        super().__init__("some_component", name, {}, opts)

        self.stack_reference: StackReference = StackReference(
            name = f"{name}-stack_reference",
            opts = ResourceOptions(parent=self),
            stack_name = stack_name
        )

stack_name: str = "REDACTED"

SomeComponent("example_of_problem", stack_name)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@ CHANGELOG
 
 ## HEAD (Unreleased)
 - Fixed ResourceOptions issue with stack references in Python SDK
- [#4553](https://github.com/pulumi/pulumi/pull/4553)
+ [#4553](https://github.com/pulumi/pulumi/pull/4553) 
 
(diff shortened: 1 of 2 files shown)


solidDoWant:
Anybody know why Travis is failing the pre-install tasks? Looks like it's trying to download something from a site with an expired cert.

Issue #4421, opened 2020-04-16
Create a CI target to test pkg/codegen downstream dependencies


EvanBoyle:
pkg/codegen has a bunch of downstream dependencies including `pulumi-terraform-bridge`, all of the providers, and docs. We don't have a way to trigger and E2E test from pulumi to verify new codegen changes. This probably isn't something we want to run on every build, given that CI currently takes ~40 minutes in pulumi/pulumi. Having something that we can run ad-hoc to bring more confidence to codegen specific changes would be helpful here.

pgavlin:
Fixed with #4567.

PR #4567, opened 2020-05-05
Add GH actions for downstream codegen testing.

pgavlin:
These changes port the downstream codegen testing actions from
https://github.com/pulumi/pulumi-terraform-bridge to this repository.
`github.com/pulumi/pulumi/pkg/v2` is replaced with the PR HEAD.

### DIFF
--- a/.github/workflows/codegen-test.yml
+++ b/.github/workflows/codegen-test.yml
@@ -1,0 +1,173 @@
+name: Downstream Codegen Testing
+on: [pull_request]
+
+jobs:
+  downstream-aws:
+    name: Test AWS Downstream
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install Go
+        uses: actions/setup-go@v1
+        with:
+          go-version: 1.13.x
+      - name: Install Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: '10.x'
+      - name: Install Python
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.6.10
+      - name: Install Pulumi CLI
+        uses: pulumi/action-install-pulumi-cli@releases/v1
+
+      - name: Check out source code
+        uses: actions/checkout@master
+      - name: Test Downstream
+        uses: pulumi/action-test-provider-downstream@releases/v4
+        env:
+          GOPROXY: "https://proxy.golang.org"
+        with:
+          replace: github.com/pulumi/pulumi/pkg/v2
+          replace-with: pulumi/pkg
+          downstream-name: pulumi-aws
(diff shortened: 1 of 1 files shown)


lukehoban:
I guess this will run on all PRs?  Given that 90% of PRs to this repo don't modify anything that would impact those, feels like this might be quite noisy/heavyweight (and quite likely to cause PR "failures" that aren't "real"?)

pgavlin:
> I guess this will run on all PRs? Given that 90% of PRs to this repo don't modify anything that would impact those, feels like this might be quite noisy/heavyweight

Yep. I'm open to suggestions, but I'd love to avoid making this a manually-triggered run.

> (and quite likely to cause PR "failures" that aren't "real"?)

Why do you say this? While I'm sure this will happen, my guess would be that 80% of PRs wouldn't hit this.

stack72:
We should be able to use this mechanism - https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-paths

To check for changes to the codegen paths

pgavlin:
> We should be able to use this mechanism - https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-paths
> 
> To check for changes to the codegen paths

This is reasonable. My only concern is that changes outside the codegen paths can affect codegen, so we could get false passes.

stack72:
> > We should be able to use this mechanism - https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-paths
> > To check for changes to the codegen paths
> 
> This is reasonable. My only concern is that changes outside the codegen paths can affect codegen, so we could get false passes.

so let's go with a black list - anything that doesn't affect codegen, eg. acceptance tests, README, cli etc ?

pgavlin:
> so let's go with a black list - anything that doesn't affect codegen, eg. acceptance tests, README, cli etc ?

I think it's fine to start with an allow list and move to a disallow list if necessary.

pgavlin:
Note to self: it is probably worth doing triggering testing for the docs repo as well.

pgavlin:
OK, this is ready to go.

The test jobs are failing b/c the repos under test haven't yet been updated with the latest from https://github.com/pulumi/pulumi-terraform-bridge. Once that has happened, this should light up nicely.

stack72:
> OK, this is ready to go.
> 
> The test jobs are failing b/c the repos under test haven't yet been updated with the latest from [pulumi/pulumi-terraform-bridge](https://github.com/pulumi/pulumi-terraform-bridge). Once that has happened, this should light up nicely.

This is awesome :) This actually closes out #4421 right?

pgavlin:
Yup!

PR #4570, opened 2020-05-05
Escape `"""` in Python doc comments.

### LINKED ISSUE #4568: Codegen for Python examples
The codegen for Python examples is not being escaped correctly - this is causing the tests to fail where it is being used

https://travis-ci.com/github/pulumi/pulumi-aws/builds/163897314#L4128

pgavlin:
With the addition of Python examples, doc comments may now contain
triple quotes. These must be escaped in order to avoid malformed source
files.

Fixes #4568.


pgavlin:
I've verified locally that this works as expected: https://github.com/pulumi/pulumi-aws/compare/pgavlin/bridge-v2.3.0-fixed

PR #4577, opened 2020-05-05
Add UTM tags to README links

zchase:
This PR adds UTM tagging to `pulumi.com` links in the README.md. This will help us determine what links people are finding useful and help us breakdown traffic from GitHub a bit more granularly.


zchase:
@stack72 We are going to want to update the README's of all our repos. I created an issue for tracking here: https://github.com/pulumi/home/issues/804

PR #4589, opened 2020-05-07
[codegen/hcl2] Fix iterator apply rewrites.

pgavlin:
There are two expressions in HCL2 that are used to iterate over
collections:

- Splat expressions, e.g. `foo.*.bar`, and
- For expressions, e.g. `[for v in foo: v.bar]`

In both of these cases, the parts of the expression that are not the
collection being iterated behave like callbacks, and must be treated as
such by the apply rewriter.


pkg/codegen/hcl2/rewrite_apply_test.go
```diff
@@ -16,6 +16,7 @@ func (nameInfo) Format(name string) string {
 	return name
 }
 
+//nolint: lll
```
justinvp, line 19: Nit:

```suggestion
// nolint: lll
```

pgavlin: The lack of a space is proper style for machine-readable comments in Go. See https://github.com/golangci/golangci-lint#nolint for canonical examples of this particular comment--I'm trying to dig up the general rule, but haven't yet.

pgavlin: > Use //nolint instead of // nolint because machine-readable comments should have no space by Go convention.

justinvp: Ah, we have a bunch of uses with the space that we should cleanup, then.

PR #4591, opened 2020-05-07
[codegen/docs] Add a new template for examples section

praneetloke:
This PR extracts the (if any) examples in the description and turns it into a template-consumable structure so we can render it any way we want to.

Note that this does rely on the fact that the examples are generated in the schema with the appropriate short-codes. In the event that that does not happen, instead of not showing any examples, we would render the description as-is, because the resource docs generator does not know how to process them. From what I have observed this is only a problem for some smaller packages where some pages have the problem with the title.

### DIFF
--- a/pkg/codegen/docs/gen_test.go
+++ b/pkg/codegen/docs/gen_test.go
@@ -123,7 +123,7 @@ func initTestPackageSpec(t *testing.T) {
 		Resources: map[string]schema.ResourceSpec{
 			"prov:module/resource:Resource": {
 				ObjectTypeSpec: schema.ObjectTypeSpec{
-					Description: "This is a module-level resource called Resource.",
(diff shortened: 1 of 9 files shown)


pkg/codegen/docs/gen_test.go
```diff
@@ -123,7 +123,7 @@ func initTestPackageSpec(t *testing.T) {
...
...w Spaces Bucket\nconst mybucket = new digitalocean.SpacesBucket(\"mybucket\", {\n    region: \"sfo2\",\n    acl: \"public-read\",\n});\n// Create a DigitalOcean managed Let's Encrypt Certificate\nconst cert = new digitalocean.Certificate(\"cert\", {\n    type: \"lets_encrypt\",\n    domains: [\"static.example.com\"],\n});\n// Add a CDN endpoint with a custom sub-domain to the Spaces Bucket\nconst mycdn = new digitalocean.Cdn(\"mycdn\", {\n    origin: mybucket.bucketDomainName,\n    customDomain: \"static.example.com\",\n    certificateId: cert.id,\n});\n```\n```python\nimport pulumi\nimport pulumi_digitalocean as digitalocean\n\n# Create a new Spaces Bucket\nmybucket = digitalocean.SpacesBucket(\"mybucket\",\n    region=\"sfo2\",\n    acl=\"public-read\")\n# Create a DigitalOcean managed Let's Encrypt Certificate\ncert = digitalocean.Certificate(\"cert\",\n    type=\"lets_encrypt\",\n    domains=[\"static.example.com\"])\n# Add a CDN endpoint with a custom sub-domain to the Spaces Bucket\nmycdn = digitalocean.Cdn(\"mycdn\",\n    origin=mybucket.bucket_domain_name,\n    custom_domain=\"static.example.com\",\n    certificate_id=cert.id)\n```\n\n{{% /example %}}\n{{% /examples %}}\n",
```
justinvp, line 126: Any reason not to use a raw string literal here, to make this more readable rather than one long line? e.g.:

```go
					Description: `This is a module-level resource called Resource. 

{{% examples %}}
## Example Usage

{{% example %}}
### Basic Example
...`,
```

pkg/codegen/docs/templates/examples.tmpl
```diff
@@ -0,0 +1,15 @@
+{{ define "examples" -}}
+{{ print "{{% examples %}}" }}
+## Example Usage
+
+{{- range . }}
+{{ .Title }}
```
justinvp, line 6: Are the titles going to be the same in each language? If so, do we need generate a unique id so they can be anchor linked?

praneetloke: Great point, @justinvp ! Yes, the title, if present, will be the same across all languages for a specific example. So yes, having a unique id would be great. Thankfully, @sean1588's work to handle languages in anchor tags will make it all work, just as long as I ensure there is a unique ID for these titles.

pkg/codegen/docs/gen_test.go
praneetloke: No reason. I copy/pasted the line as-is from a schema. I'll format it here. Since this is a test, it'll be good to make this more readable for future edits.

pkg/codegen/docs/templates/examples.tmpl
praneetloke: Actually, I just realized that the way this title is placed currently, it is _not_ inside the language snippet, so generating an ID per-language isn't needed. However, that also means someone sharing a link to the example wouldn't be able to pin to a specific language. Hmm curious what @cnunciato thinks as well.

cnunciato: > I just realized that the way this title is placed currently, it is not inside the language snippet, so generating an ID per-language isn't needed.

Correct, and intentional, yeah. If we want to link to a specific language, we should be able to use the `?language=` querystring approach. (Which IMO we can do in a follow-up.)

praneetloke: > Which IMO we can do in a follow-up.

Yeah that's what I am thinking too.

PR #4594, opened 2020-05-08
[codegen/hcl2] Add a conversion insertion pass.

pgavlin:
Add a rewriter that reifies implicit conversions into a call to the
`__convert` intrinsic. Code generators can recognize this intrinsic and
use it to generate appropriate conversion code.

### DIFF
--- a/pkg/codegen/hcl2/binder_schema.go
+++ b/pkg/codegen/hcl2/binder_schema.go
@@ -222,8 +222,8 @@ func GetSchemaForType(t model.Type) (schema.Type, bool) {
 		schemas := codegen.Set{}
 		for _, t := range t.ElementTypes {
 			if s, ok := GetSchemaForType(t); ok {
-				if s, ok := s.(*schema.UnionType); ok {
-					for _, s := range s.ElementTypes {
+				if union, ok := s.(*schema.UnionType); ok {
+					for _, s := range union.ElementTypes {
 						schemas.Add(s)
 					}
 				} else {
--- a/pkg/codegen/hcl2/binder_schema.go
+++ b/pkg/codegen/hcl2/binder_schema.go
@@ -198,3 +198,52 @@ func (b *binder) schemaTypeToType(src schema.Type) (result model.Type) {
 		}
 	}
 }
+
+// GetSchemaForType extracts the schema.Type associated with a model.Type, if any.
+//
+// The result may be a *schema.UnionType if multiple schema types are associaged with the input type.
+func GetSchemaForType(t model.Type) (schema.Type, bool) {
+	switch t := t.(type) {
+	case *model.ObjectType:
+		if len(t.Annotations) == 0 {
+			return nil, false
+		}
+		for _, a := range t.Annotations {
+			if t, ok := a.(schema.Type); ok {
+				return t, true
+			}
+		}
+		return nil, false
+	case *model.OutputType:
+		return GetSchemaForType(t.ElementType)
+	case *model.PromiseType:
+		return GetSchemaForType(t.ElementType)
+	case *model.UnionType:
+		schemas := codegen.Set{}
+		for _, t := range t.ElementTypes {
+			if s, ok := GetSchemaForType(t); ok {
+				if s, ok := s.(*schema.UnionType); ok {
+					for _, s := range s.ElementTypes {
+						schemas.Add(s)
+					}
+				} else {
+					schemas.Add(s)
+				}
+			}
+		}
+		if len(schemas) == 0 {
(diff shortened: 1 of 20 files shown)


pgavlin:
@mikhailshilkov I believe that you could use this to obtain contextual information about the types required by C# code generation. I suspect that there are bugs--in particular, I think the [annotations check](https://github.com/pulumi/pulumi/pull/4594/files#diff-db5aa6f58554ebef897fd702d2d02437R80-R83) needs to recurse into types--but this would allow the code generator to see the desired type of an `ObjectConsExpression`. For example, you could do something like this:

```go
func (g *generator) GenFunctionCallExpression(w io.Writer, expr *model.FunctionCallExpression) {
	switch expr.Name {
	case hcl2.IntrinsicConvert:
		switch arg := args[0].(type) {
			g.genObjectConsExpression(w, arg, expr.Type())
		default:
			g.Fgenf(w, "%.v", expr) // <- probably wrong w.r.t. precedence
		}
	}
}

func (g *generator) genObjectConsExpression(w io.Writer, expr *model.ObjectConsExpression, destType model.Type) {
	// Extract object schema from type annotations
	var objectType *schema.ObjectType
	if annotations := destType.GetAnnotations(); len(annotations) == 1 {
		objectType = annotations[0].(*schema.ObjectType)
	}

	// Use object schema (if present) to inform codegen, left as an exercise to the reader
}

func (g *generator) GenObjectConsExpression(w io.Writer, expr *model.ObjectConsExpression) {
	g.genObjectConsExpression(w, expr, expr.Type())
}
```

mikhailshilkov:
`RewriteConversions` is never called yet. Is that coming separately or what's the idea here?

pgavlin:
> RewriteConversions is never called yet. Is that coming separately or what's the idea here?

It's a pass that is intended for use by downstream consumers, so you should call this from within the C# code generator:

```
func genResource(g *generator) genResource(w io.Writer, r *hcl2.Resource) {
	...


	// Prior to codegen, add conversions to input properties
	for _, input := range r.Inputs {
		destType, diags := r.InputType.Traverse(hcl.TraverseAttr{Name: input.Name})
		contract.Ignore(diags)
		input.Value = hcl2.RewriteConversions(input.Value, destType.(model.Type))
	}

	...
}
```

mikhailshilkov:
@pgavlin Yeah, I tried something like this (now changed to exactly this) but I don't see annotations happening. It seems to go through the tree but e.g. looking in the debugger `sameAnnotations` never returns false throughout all examples.

Is there a simple thing I should check? Otherwise, I'll come back to debugging on Monday.

pgavlin:
> Is there a simple thing I should check? Otherwise, I'll come back to debugging on Monday.

You don't see annotations even with `model.ResolveOutputs(r.InputType)` ?

mikhailshilkov:
No... maybe take a look at `mikhailshilkov/cs-programgen2` at what I'm doing wrong? Thanks a ton!

pgavlin:
@mikhailshilkov it looks like this is a phase ordering issue. Here's the IR I see in `GenFunctionCallExpression` for a call to `__convert` that originally wrapped an `ObjectConsExpression`:

```
__convert( [
	__convert(__constructor(Aws.Ecs.ServiceArgs,{
		targetGroupArn = webTargetGroup.arn
		containerName = "my-app"
		containerPort = __convert(80)
	}))
])
```

The call to `__constructor` prevents the logic in `GenFunctionCallExpression` from seeing the `ObjectConsExpression`.

mikhailshilkov:
I don't think so? `RewriteConversions` are called before "constructors" are added, and I didn't see the rewrites working as I would expect. I removed constructors and still saw no annotations. Should I try your latest commit that revamped `GetAnnotations()`?

pkg/codegen/hcl2/binder_schema.go
```diff
@@ -198,3 +198,52 @@ func (b *binder) schemaTypeToType(src schema.Type) (result model.Type) {
...
+	case *model.UnionType:
+		schemas := codegen.Set{}
+		for _, t := range t.ElementTypes {
+			if s, ok := GetSchemaForType(t); ok {
+				if s, ok := s.(*schema.UnionType); ok {
+					for _, s := range s.ElementTypes {
+						schemas.Add(s)
+					}
+				} else {
+					schemas.Add(s)
```
mikhailshilkov, line 230: `s` is overridden here to the value of `UnionType` conversion, which will be `nil`. I think we should change the above `if` not to redeclare `s` to add the original result of `GetSchemaForType` to `schemas`.

PR #4597, opened 2020-05-09
[codegen/docs] Generate a table with links to the language-specific module pages

praneetloke:
Part of pulumi/docs#3178.

This PR adds a new template for injecting a table with links to the module's packages in each supported language. The table is added at the bottom of every module's page under the Package Details header.

Along with those changes, this PR also introduces a way to automatically generate a draft PR in the docs repo based on the generator changes made in a PR in this repo (like this PR). For this, a new workflow file was added to `.github/workflows`. This is the [docs PR](https://github.com/pulumi/pulumi/pull/4597#ref-pullrequest-616144478) that was generated for the changes in this PR.


pkg/codegen/docs/gen.go
```diff
@@ -1536,6 +1537,58 @@ func sortIndexEntries(entries []indexEntry) {
...
+		case "go":
+			langTitle = "Go"
+			if mod.mod == "" {
+				title = fmt.Sprintf("%s", mod.pkg.Name)
+			} else {
+				title = fmt.Sprintf("%s/%s", mod.pkg.Name, mod.mod)
+			}
+			link = docLangHelper.GetDocLinkForResourceType(mod.pkg, mod.mod, "")
+		case "nodejs":
+			langTitle = "NodeJS"
```
cnunciato, line 1568:
```suggestion
			langTitle = "Node.js"
```

pkg/codegen/docs/templates/language_sdk_links.tmpl
```diff
@@ -0,0 +1,8 @@
+{{ define "language_sdk_links" }}
+<dl class="resources-properties">
```
cnunciato, line 2: This can just be `tabular` now:

```suggestion
<dl class="tabular">
```

pkg/codegen/docs/gen.go
```diff
@@ -1536,6 +1537,35 @@ func sortIndexEntries(entries []indexEntry) {
...
+	languageLinks := map[string]string{}
+	for _, lang := range supportedLanguages {
+		var link string
+		var title string
+		var langTitle string
+
+		docLangHelper := getLanguageDocHelper(lang)
+		switch lang {
+		case "csharp":
+			langTitle = "C#"
```
justinvp, line 1552: Consider using ".NET" here, since one could use C#, F#, or VB (or any .NET language, really). This would align with how this change uses "Node.js" rather than "TypeScript" and "JavaScript".

```suggestion
			langTitle = ".NET"
```

praneetloke:
Ok I think this is finally ready to be merged. I hooked-up the resource docs test workflow to also generate docs for kubernetes (since it is a native provider). With that I found some issues that needed to be addressed. Namely, generating a link for k8s modules is not 1:1. Modules in k8s are versioned. So in .NET specifically, we don't have top-level pages to link to. So I am omitting the .NET link for such index pages, but when the user navigates to a versioned module, then the .NET link is shown which links to the right page.

In any case, we now have the new resource docs test workflow generating docs for AWS and k8s in the draft docs PR that it generates due to generator changes.

PR #4611, opened 2020-05-12
Initial support for C# program gen

mikhailshilkov:


pkg/codegen/internal/test/testdata/aws-fargate.pp
```diff
@@ -28,7 +28,7 @@
 
 // Create an IAM role that can be used by our service's task.
 resource taskExecRole "aws:iam:Role" {
-	assumeRolePolicy = {
+	assumeRolePolicy = toJSON({
```
mikhailshilkov, line 31: I haven't found a good way to fix this yet. It should probably be a rewrite that would insert `toJSON` automatically based on type mismatch of the right part vs. left part, but that sounds a bit too complicated for this original version.

pgavlin: That's reasonable. Can you file a bug to track enabling this and tag this line with a comment that references the issue?

pkg/codegen/dotnet/gen_program_expressions.go
```diff
@@ -0,0 +1,555 @@
...
+		expr = g.awaitInvokes(expr)
+	} else {
+		expr = g.outputInvokes(expr)
+	}
+	return expr
+}
+
+// outputInvokes wraps each call to `invoke` with a call to the `output` intrinsic. This rewrite should only be used if
+// resources are instantiated within a stack constructor, where `await` operator is not available. We want to avoid the
+// nastiness of working with raw `Task` and wrap it into Pulumi's Output immediately to be able to `Apply` on it.
```
pgavlin, line 54: Could we use leave these as `Task<>`s and use [`ContinueWith<>`](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.task-1.continuewith?view=netcore-3.1#System_Threading_Tasks_Task_1_ContinueWith__1_System_Func_System_Threading_Tasks_Task__0____0__) here instead? That seems more idiomatic.

mikhailshilkov: FTW, all our examples (in the examples repo) convert to outputs immediately. IMO, `ContinueWith` is quite crappy and most people aren't familiar with it either, so I'd rather encourage them to learn outputs instead.

pkg/codegen/dotnet/gen_program.go
```diff
@@ -0,0 +1,428 @@
...
+	}
+
+	return rootNamespace, fmt.Sprintf("%s%s.%s", rootNamespace, namespace, Title(member))
+}
+
+// argumentTypeName computes the C# argument class name for the given expression and model type.
+func (g *generator) argumentTypeName(expr model.Expression, destType model.Type) string {
+	if schemaType, ok := hcl2.GetSchemaForType(destType.(model.Type)); ok {
+		if objType, ok := schemaType.(*schema.ObjectType); ok {
+			token := objType.Token
```
pgavlin, line 301: Using early returns here is a bit more idiomatic:

```suggestion
	schemaType, ok := hcl2.GetSchemaForType(destType.(model.Type))
	if !ok {
		return ""
	}
	objType, ok := schemaType.(*schema.ObjectType)
	if !ok {
		return ""
	}
	
	token := objType.Token
```

pkg/codegen/dotnet/gen_program_expressions.go
```diff
@@ -0,0 +1,560 @@
...
+)
+
+type nameInfo int
+
+func (nameInfo) Format(name string) string {
+	return validIdentifier(name)
+}
+
+// rewriteExpression amends the expression with intrinsics for C# generation.
+func (g *generator) rewriteExpression(expr model.Expression, typ model.Type) model.Expression {
```
pgavlin, line 40: Let's call this `lowerExpression` : the rewrites we're doing are consistent with the general definition of lowering as a process of rewrites that prepare a program for code generation, and `lowerExpression` is used in the other backends.

```suggestion
func (g *generator) lowerExpression(expr model.Expression, typ model.Type) model.Expression {
```

pkg/codegen/dotnet/gen_program.go
```diff
@@ -0,0 +1,428 @@
...
+	} else {
+		g.Fgenf(w, "%sthis.%s", g.Indent, propertyName(v.Name()))
+	}
+	g.Fgenf(w, " = %.3v;\n", g.rewriteExpression(v.Value, v.Type()))
+}
+
+func (g *generator) genOutputProperty(w io.Writer, v *hcl2.OutputVariable) {
+	// TODO(pdg): trivia
+	g.Fgenf(w, "%s[Output(\"%s\")]\n", g.Indent, v.Name())
+	g.Fgenf(w, "%spublic Output<string> %s { get; set; }\n", g.Indent, propertyName(v.Name()))
```
pgavlin, line 423: The element type of the `Output` should be derived from the type of its value instead of being hardcoded to `string`, right?

pkg/codegen/dotnet/utilities.go
```diff
@@ -0,0 +1,77 @@
...
+
+// isLegalIdentifierPart returns true if it is legal for c to be part of a C# identifier (besides the first character)
+// as per https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/lexical-structure
+func isLegalIdentifierPart(c rune) bool {
+	return isLegalIdentifierStart(c) || unicode.In(c, unicode.Mn, unicode.Mc, unicode.Nd, unicode.Pc, unicode.Cf)
+}
+
+// validIdentifier replaces characters that are not allowed in C# identifiers with underscores. A reserved word is
+// prefixed with @. No attempt is made to ensure that the result is unique.
+func validIdentifier(name string) string {
```
pgavlin, line 55:
```suggestion
func makeValidIdentifier(name string) string {
```

You might also call this `cleanName`, which I believe is consistent with the other backends.

pkg/codegen/dotnet/gen_program_expressions.go
pgavlin: Let's add this comment from `awaitInvokes` to this as well, then:

```
// Note that this depends on the fact that invokes are the only way to introduce promises
// in to a Pulumi program; if this changes in the future, this transform will need to be applied in a more general way
// (e.g. by the apply rewriter).
```

pkg/codegen/dotnet/gen_program_expressions.go
```diff
@@ -0,0 +1,560 @@
...
+			expr.SyntaxNode().Range())
+	}
+}
+
+func (g *generator) GenObjectConsExpression(w io.Writer, expr *model.ObjectConsExpression) {
+	g.genObjectConsExpression(w, expr, expr.Type())
+}
+
+func (g *generator) genObjectConsExpression(w io.Writer, expr *model.ObjectConsExpression, destType model.Type) {
+	if len(expr.Items) > 0 {
```
pgavlin, line 413:
```suggestion
	if len(expr.Items) == 0 {
		return
	}
```

pkg/codegen/dotnet/gen_program_expressions.go
mikhailshilkov: Sure, that reminds me to ask you to talk me through what "lower" means here (hopefully I'll be able to join the call later today)

pkg/codegen/dotnet/gen_program.go
mikhailshilkov: Good catch!

pkg/codegen/dotnet/utilities.go
mikhailshilkov: How about renaming `cleanName` to `makeValidIdentifier` in other places? It seems more specific to me?

pgavlin: Sure, that sounds good!

pkg/codegen/dotnet/gen_program.go
mikhailshilkov: I _think_ we are missing required type annotations for this. I added a TODO and will merge as-is to handle this in a follow-up.

PR #4612, opened 2020-05-12
Install golang in pulumi image

### LINKED ISSUE #4599: Docker image pulumi/actions doesn't have golang installed
[The documentation](https://www.pulumi.com/docs/guides/continuous-delivery/github-actions/#demos-and-examples) and [the docker hub page](https://hub.docker.com/r/pulumi/actions) mention that you can use the `pulumi/actions` image with any language. However, when using golang, you'd run into the following:

```
error: failed to discover plugin requirements: couldn't find go binary: unable to find program: go
```

Indeed, golang isn't installed in this image (unlike python or nodejs).

I see two solutions to this problem:
1. Add golang to the image
2. Have one image per language (which would make each image smaller and faster to pull)

Let me know your thoughts and I'd be glad to write a PR for any chosen solution.

jaxxstorm:
Fixes #4599

I don't love this solution, but even stretch only has has Go `1.7`. I've added this as another layer following the helm and dotnet logic, but happy to fold it into the above layer if needed.

### DIFF
--- a/dist/pulumi/Dockerfile
+++ b/dist/pulumi/Dockerfile
@@ -4,6 +4,9 @@ LABEL "repository"="https://github.com/pulumi/pulumi"
 LABEL "homepage"="https://pulumi.com/"
 LABEL "maintainer"="Pulumi Team <team@pulumi.com>"
 
+ENV GOLANG_VERSION 1.13.10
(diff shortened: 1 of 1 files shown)


jaxxstorm:
@chrsmith it's what we're currently using to build pulumi itself, as well as all the providers. I figured it best to stick with that until we "officially" start moving to Go 1.14.

PR #4614, opened 2020-05-12
[codegen/hcl2] Fix descent in RewriteConversions.

pgavlin:
Descend into anonymous function expressions, the operands to conditional
expressions, and the value in for expressions.

### DIFF
--- a/pkg/codegen/hcl2/rewrite_convert.go
+++ b/pkg/codegen/hcl2/rewrite_convert.go
@@ -40,18 +40,35 @@ func sameSchemaTypes(xt, yt model.Type) bool {
 
 func RewriteConversions(x model.Expression, to model.Type) model.Expression {
 	switch x := x.(type) {
+	case *model.AnonymousFunctionExpression:
+		x.Body = RewriteConversions(x.Body, to)
 	case *model.BinaryOpExpression:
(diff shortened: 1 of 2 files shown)


mikhailshilkov:
Is (1) applies (2) converts the blessed order? Or shouldn't it matter?

Had a quick glance, but it looks like this change breaks another example:

```
invoke("aws::getAmi", __convert( {
	filters = [{
		name = "name"
		values = ["amzn-ami-hvm-*-x86_64-ebs"]
	}]
	owners = ["137112412989"] // Amazon
	mostRecent = true
}))
```

while the previous version had an extra convert around the filter

```
invoke("aws::getAmi", __convert( {
	filters = [
        __convert({
		name = "name"
		values = ["amzn-ami-hvm-*-x86_64-ebs"]
	})]
	owners = ["137112412989"] // Amazon
	mostRecent = true
}))
```

PP:

```
ami = invoke("aws:index:getAmi", {
	filters = [{
		name = "name"
		values = ["amzn-ami-hvm-*-x86_64-ebs"]
	}]
	owners = ["137112412989"] // Amazon
	mostRecent = true
})
```

pgavlin:
Ah, damn. There is context that needs to be passed down. Taking another swing at this.

PR #4631, opened 2020-05-13
Generate deprecation comments in Go.

pgavlin:
This is consistent with the convention for marking members as
deprecated: https://github.com/golang/go/wiki/Deprecated.


pgavlin:
Filed https://github.com/pulumi/pulumi/issues/4632 and https://github.com/pulumi/pulumi-terraform-bridge/issues/199 to track the incorrectly-named property references.

PR #4636, opened 2020-05-14
Resolve referenced packages for every TF program

mikhailshilkov:
While testing C# program generation for a provider, I found that namespaces aren't resolved correctly.

For a binder, `b.options.packageCache` has a longer lifecycle than `b.referencedPackages`, so it seems wrong to check the former while populating the latter.

This PR changes `referencedPackages` to a map and populates it with unique packages regardless of the cache state.

### DIFF
--- a/pkg/codegen/hcl2/binder_schema.go
+++ b/pkg/codegen/hcl2/binder_schema.go
@@ -80,13 +80,13 @@ func (b *binder) loadReferencedPackageSchemas(n Node) error {
 	contract.Assert(len(diags) == 0)
 
 	for _, name := range packageNames.SortedValues() {
-		if _, ok := b.options.packageCache.entries[name]; ok {
+		if _, ok := b.referencedPackages[name]; ok {
 			continue
 		}
 		if err := b.loadPackageSchema(name); err != nil {
 			return err
 		}
-		b.referencedPackages = append(b.referencedPackages, b.options.packageCache.entries[name].schema)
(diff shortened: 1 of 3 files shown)


mikhailshilkov:
I believe Python benefits as well, see the diffs above -- they look correct to me now

pkg/codegen/hcl2/binder_schema.go
```diff
@@ -80,13 +80,13 @@ func (b *binder) loadReferencedPackageSchemas(n Node) error {
 	contract.Assert(len(diags) == 0)
 
 	for _, name := range packageNames.SortedValues() {
-		if _, ok := b.options.packageCache.entries[name]; ok {
+		if _, ok := b.referencedPackages[name]; ok {
```
pgavlin, line 83: Huh, good catch. The broken code must have come from a bad refactoring or something--`referencedPackages` is supposed to have a peer set that performs this function. The map works, but can we sort the list returned by `Packages()` by name?

mikhailshilkov: I don't _think_ the old version was sorting them either: packages coming from different resources would be appended at resource order. But I added the sorting in `Packages()`. (OMG, is it really that hard in Go?)

pgavlin: It wasn't sorting them per se, but the old order was deterministic. This looks good.

PR #4643, opened 2020-05-14
Use patched azure gocloud library

### LINKED ISSUE #4642: Unable to use Azure Key Vault Secret Provider in China
When initializing a stack with
```bash
pulumi up --secrets-provider https://<keyvault_name>.vault.azure.cn/keys/<keyname>
```

I got the following error

`error: invalid keyID "https://<keyvault_name>.vault.azure.cn/keys/<keyname>"; must match ^(https://.+\.vault\.azure\.net/)keys/(.+)$ []`

(This was a useful error message. Was very clear.)
This seems like a valid secrets provider if we can change the regex to include `.cn`.

jaxxstorm:
This PR gets us away from an upstream fork of gocloud which we've been using for a while.

Fixes #4642


jaxxstorm:
Want to make sure we get a passed test for this one, so waiting for #4668 to land before merging it

Issue #1334, opened 2018-05-07
Move `pulumi new` "install dependencies" implementation into the language plugins


justinvp:
This is implemented inside `pulumi new` for a simpler initial implementation, but it really should live inside the language plugins long-term. Then `pulumi new` would just call the language plugin (e.g. `InstallDependencies` or `PrepareProject`) to install dependencies, and the language plugin would handle doing whatever it needs to do. That way, if/when we add new languages, `pulumi new` would just work with the new language.

Since there is a gRPC boundary between the CLI and language plugin, the tricky part will be making sure we can show the Stdout and Stderr of any commands (e.g. `npm install`) run by the language plugin.

justinvp:
We have a similar issue for Policy Packs. We need to be able to package a policy pack and then subsequently install it and it's dependencies. Right now we effectively have switch statements, but really we should be deferring to the analyzer plugins to do this.

PR #4644, opened 2020-05-14
Support publishing Python policy packs

justinvp:
This PR adds support for publishing Python policy packs to the service, and downloading/using such policy packs when applied to stacks in an organization.

## Design
Reminder: For Node.js policy packs, we run `npm pack`  to package up the code and upload the `.tgz` to the service. When downloading, we manually extract that `.tgz` file and then run `npm install` to install its dependencies.

I explored a few different approaches to packaging up the Python policy pack code (alternatives discussed below). In the end, I went with a simple and pragmatic approach:

- **Packaging**: We simply `.tgz` the contents of the  Python policy pack, ignoring any `.git` directories and anything specified in `.gitignore` (similar to what `npm pack` does). The policy pack version is specified in `PulumiPolicy.yaml`.

- **Downloading/Installing**: When downloading the policy pack, we manually extract the `.tgz` the same as we do for Node.js policy packs. If it is a Python policy pack, we create a virtual environment for it by running `python -m venv venv` (using the same approach for looking up the `python` binary as we use for running Python programs). If there is a `requirements.txt` file, we run `pip` in the virtual environment to install dependencies in the virtual environment.

* **Running**: When we run the policy pack, we ensure the `python` from the virtual environment is used by prepending the path to the virtual environment’s `bin` directory (`Scripts` on Windows) to the `PATH` environment variable used to execute the analyzer plugin (effectively doing the same thing as the virtual environment’s `activate` script).

## Commits
Summary of commits in this PR:

1. The first commit simply factors out the existing code for running `python` so we can use it in more places than just running Python Pulumi programs.

2. The second commit resurrects some old code we had for archiving a directory of code, for PPCs, that had been deleted. This commit brings back the code as it was before it was deleted (originally at `pkg/util/archive`).

3. The third commit modifies the archive code to use the `.tgz` format rather than `.zip` so we can package Python policy packs in the same format used for Node.js.

4. The fourth commit adds support for specifying the policy pack version in `PulumiPolicy.yaml`. This is how the version is specified for Python policy packs. Node.js policy packs can continue to set the version in `package.json`, but if set in `PulumiPolicy.yaml`, the `PulumiPolicy.yaml` value will be used.

5. The fifth commit builds on top of the former commits to add support for publishing Python policy packs to the service and downloading/using such policy packs when applied to a stack in an organization.

Fixes https://github.com/pulumi/pulumi-policy/issues/211

## Alternatives Considered
I initially went down the path of packaging it via `setuptools` — the same way you’d package something to be published to PyPi. This ended up requiring a lot of additional boilerplate (you’d need a `setup.py` with several parameters set, a `MANIFEST.in` file with an entry to include `PulumiPolicy.yaml` in the package), moving away from using `requirements.txt` to specifying the dependencies in `setup.py`, etc. This all added up to making Python policy packs feel significantly different than regular Python Pulumi programs.

I also briefly explored using [executable Python zip archives](https://docs.python.org/3/library/zipapp.html). The challenge here is around dependencies. According to the [docs](https://docs.python.org/3/library/zipapp.html#creating-standalone-applications-with-zipapp), to create a standalone zipapp you must bundle required dependencies inside the archive — which isn’t going to work for us if someone has dependencies that aren't cross-platform.

### DIFF
--- a/pkg/backend/httpstate/policypack.go
+++ b/pkg/backend/httpstate/policypack.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"path"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -26,6 +25,7 @@ import (
 	"github.com/pulumi/pulumi/sdk/v2/go/common/util/contract"
 	"github.com/pulumi/pulumi/sdk/v2/go/common/util/result"
 	"github.com/pulumi/pulumi/sdk/v2/go/common/workspace"
+	"github.com/pulumi/pulumi/sdk/v2/python"
 )
 
 type cloudRequiredPolicy struct {
@@ -174,24 +174,36 @@ func (pack *cloudPolicyPack) Publish(
 
 	fmt.Println("Compressing policy pack")
 
-	if runtime := op.PolicyPack.Runtime.Name(); !strings.EqualFold(runtime, "nodejs") {
-		return result.Errorf(
-			"failed to publish policies because Pulumi.yaml requests unsupported runtime %s",
-			runtime)
-	}
+	var packTarball []byte
 
 	// TODO[pulumi/pulumi#1307]: move to the language plugins so we don't have to hard code here.
-	packTarball, err := npm.Pack(op.PlugCtx.Pwd, os.Stderr)
-	if err != nil {
-		return result.FromError(
-			errors.Wrapf(err, "could not publish policies because of error running npm pack"))
+	runtime := op.PolicyPack.Runtime.Name()
+	if strings.EqualFold(runtime, "nodejs") {
+		packTarball, err = npm.Pack(op.PlugCtx.Pwd, os.Stderr)
+		if err != nil {
+			return result.FromError(
+				errors.Wrap(err, "could not publish policies because of error running npm pack"))
+		}
+	} else if strings.EqualFold(runtime, "python") {
+		// npm pack puts all the files in a "package" subdirectory inside the .tgz it produces, so we'll do
+		// the same for Python. That way, after unpacking, we can look for the PulumiPolicy.yaml inside the
+		// package directory to determine the runtime of the policy pack.
+		packTarball, err = archive.Tgz(op.PlugCtx.Pwd, "package", true /*useDefaultExcludes*/)
+		if err != nil {
+			return result.FromError(
+				errors.Wrap(err, "could not publish policies because of error creating the .tgz"))
+		}
+	} else {
+		return result.Errorf(
+			"failed to publish policies because PulumiPolicy.yaml specifies an unsupported runtime %s",
+			runtime)
 	}
 
 	//
 	// Publish.
 	//
 
-	fmt.Println("Uploading Policy Pack to Pulumi service")
+	fmt.Println("Uploading policy pack to Pulumi service")
 
 	publishedVersion, err := pack.cl.PublishPolicyPack(ctx, pack.ref.orgName, analyzerInfo, bytes.NewReader(packTarball))
 	if err != nil {
@@ -237,7 +249,7 @@ func (pack *cloudPolicyPack) Remove(ctx context.Context, op backend.PolicyPackOp
 	return pack.cl.RemovePolicyPackByVersion(ctx, pack.ref.orgName, string(pack.ref.name), *op.VersionTag)
 }
 
-const npmPackageDir = "package"
+const packageDir = "package"
 
 func installRequiredPolicy(finalDir string, tarball []byte) error {
 	// If part of the directory tree is missing, ioutil.TempDir will return an error, so make sure
@@ -251,9 +263,9 @@ func installRequiredPolicy(finalDir string, tarball []byte) error {
 		return errors.Wrapf(err, "creating plugin directory %s", tempDir)
 	}
 
-	// npm unpacks into a directory called `package`.
-	tempNPMPkgDir := path.Join(tempDir, npmPackageDir)
-	if err := os.MkdirAll(tempNPMPkgDir, 0700); err != nil {
+	// The policy pack files are actually in a directory called `package`.
+	tempPackageDir := filepath.Join(tempDir, packageDir)
+	if err := os.MkdirAll(tempPackageDir, 0700); err != nil {
 		return errors.Wrap(err, "creating plugin root")
 	}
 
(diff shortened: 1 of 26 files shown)


justinvp:
Things I'm planning to do as follow-up tasks:

1. Document publishing Python policy packs (https://github.com/pulumi/docs/issues/3331)
2. Add integration test for publishing a Python policy pack (https://github.com/pulumi/pulumi-policy/issues/247)
3. Add a Python policy pack applied to the pulumi organization in the service (https://github.com/pulumi/pulumi-service/issues/4972)

pkg/backend/httpstate/policypack.go
```diff
@@ -174,24 +174,36 @@ func (pack *cloudPolicyPack) Publish(
...
+		// package directory to determine the runtime of the policy pack.
+		packTarball, err = archive.Tgz(op.PlugCtx.Pwd, "package", true /*useDefaultExcludes*/)
+		if err != nil {
+			return result.FromError(
+				errors.Wrap(err, "could not publish policies because of error creating the .tgz"))
+		}
+	} else {
+		return result.Errorf(
+			"failed to publish policies because PulumiPolicy.yaml specifies an unsupported runtime %s",
+			runtime)
```
lukehoban, line 199: Not blocking - but seeing this does make me wonder - should we be delegating this logic to the plugins - so that new languages can be added without having to touch this code?  (not sure how far away we are from that being possible generally)

pkg/backend/httpstate/policypack.go
```diff
@@ -303,3 +321,57 @@ func installRequiredPolicy(finalDir string, tarball []byte) error {
...
+	if output, err := pipCmd.CombinedOutput(); err != nil {
+		if len(output) > 0 {
+			os.Stdout.Write(output)
+			fmt.Println()
+		}
+		return errors.Wrap(err, "installing dependencies via `pip install -r requirements.txt`")
+	}
+
+	fmt.Println("Finished installing dependencies")
+	fmt.Println()
```
lukehoban, line 374: There's a fair bit of output here.  Not a big deal - but is the UX of this "good"?

sdk/go/common/resource/plugin/analyzer_plugin.go
```diff
@@ -113,6 +116,28 @@ func NewPolicyAnalyzer(
...
+			}
+			if _, err := os.Stat(virtualenv); os.IsNotExist(err) {
+				return nil, errors.Errorf("virtualenv does not exist: %s", virtualenv)
+			}
+			if !python.IsVirtualEnv(virtualenv) {
+				return nil, errors.Errorf("virtualenv does not appear to be a virtual environment: %s", virtualenv)
+			}
+			env = python.ActivateVirtualEnv(env, virtualenv)
+		}
+	}
```
lukehoban, line 139: It's a slight shame to have to do this here instead of in the python policy analyzer plugin.  I guess because the analyzer is implemented in Python, we don't have any chance to do this in time?  Not sure it's better - but is there any way to move this logic into `pulumi-analyzer-policy-python` so it comes along with the Python-specific code?  Perhaps a more general mechanism to pass `proj.Runtime.Options()` through to the plugin? (do we already do that for analyzers like we do for language hosts)?

sdk/go/common/workspace/project.go
```diff
@@ -116,23 +116,16 @@ func (proj *Project) Save(path string) error {
...
-	return ioutil.WriteFile(path, b, 0644)
+	return save(path, proj, false /*mkDirAll*/)
 }
 
 type PolicyPackProject struct {
 	// Runtime is a required runtime that executes code.
 	Runtime ProjectRuntimeInfo `json:"runtime" yaml:"runtime"`
+	// Version specifies the version of the policy pack. If set, it will override the
+	// version specified in `package.json` for Node.js policy packs.
+	Version string `json:"version,omitempty" yaml:"version,omitempty"`
```
lukehoban, line 127: I didn't quite follow why this needs to be added as part of these changes?

pkg/backend/httpstate/policypack.go
```diff
@@ -303,3 +321,57 @@ func installRequiredPolicy(finalDir string, tarball []byte) error {
 
 	return nil
 }
+
+func completePythonInstall(finalDir, projPath string, proj *workspace.PolicyPackProject) error {
+	// Create virtual environment.
+	fmt.Println("Creating virtual environment for policy pack...")
```
pgavlin, line 327: It's a little odd that these use plain `Println`, but I don't have a better option ¯\_(ツ)_/¯

sdk/go/common/util/archive/archive.go
```diff
@@ -24,11 +23,43 @@ import (
...
 	"path/filepath"
+	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/pulumi/pulumi/sdk/v2/go/common/util/contract"
+	"github.com/pulumi/pulumi/sdk/v2/go/common/util/logging"
 )
 
+// Tgz adds the contents of the provided directory to an in-memory .tar.gz/.tgz and returns the bytes.
+func Tgz(dir, prefixPathInsideTar string, useDefaultExcludes bool) ([]byte, error) {
```
pgavlin, line 34: nit:
```suggestion
func TGZ(dir, prefixPathInsideTar string, useDefaultExcludes bool) ([]byte, error) {
```

sdk/go/common/util/archive/archive.go
```diff
@@ -82,3 +113,104 @@ func Untgz(tarball []byte, dir string) error {
...
+
+func addDirectoryToTar(writer *tar.Writer, root, dir, prefixPathInsideTar string,
+	useDefaultIgnores bool, ignores *ignoreState) error {
+	ignoreFilePath := filepath.Join(dir, gitIgnoreFile)
+
+	// If there is an ignorefile, process it before looking at any child paths.
+	if stat, err := os.Stat(ignoreFilePath); err == nil && !stat.IsDir() {
+		logging.V(9).Infof("processing ignore file in %v", dir)
+
+		ignore, err := newGitIgnoreIgnorer(ignoreFilePath)
```
pgavlin, line 130: `GitIgnoreIgnerer` 💯

pkg/backend/httpstate/policypack.go
justinvp: > should we be delegating this logic to the plugins - so that new languages can be added without having to touch this code?

Ideally, yes. It'd involve defining a new analyzer gRPC method (e.g. `Package`) and then having the Policy SDK for each language provide an implementation.

For Python, this would mean porting the TGZ archiving implementation (with ignore behaviors) of this PR from Go to Python.

For Node.js, it'd mean calling `npm pack` from within Node.js Policy SDK. Though, we'd need to keep the existing Go implementation of shelling out to `npm pack` for previous versions of the Node.js Policy SDK where the `Package` gRPC method doesn't exist.

pkg/backend/httpstate/policypack.go
justinvp: The UX isn't very good. The current output looks like:

```
Downloading policy pack
Unpacking policy pack "/Users/justin/.pulumi/policies/justinvp/pulumi-analyzer-aws-python-v0.0.2.tmp234539259" "/Users/justin/.pulumi/policies/justinvp/pulumi-analyzer-aws-python-v0.0.2"
Creating virtual environment for policy pack...

Finished creating virtual environment

Installing dependencies for policy pack...

Finished installing dependencies

```

We already have https://github.com/pulumi/pulumi-policy/issues/58 tracking improving the experience here, but I'll make a quick improvement while making changes here and change the output to:

```
Installing policy pack aws-python v0.0.2...
Finished installing policy pack
```

The first line would encompass the entire install operation from downloading the policy pack to installing its dependencies. Ideally with https://github.com/pulumi/pulumi-policy/issues/58 we'd add some kind of progress bar, but in lieu of that, the "Installing..." and subsequent "Finished installing" message should suffice.

sdk/go/common/resource/plugin/analyzer_plugin.go
justinvp: > Not sure it's better - but is there any way to move this logic into `pulumi-analyzer-policy-python` so it comes along with the Python-specific code? Perhaps a more general mechanism to pass `proj.Runtime.Options()` through to the plugin? (do we already do that for analyzers like we do for language hosts)?

We don't currently pass `proj.Runtime.Options()` through to the analyzer plugins automatically like we do for language plugins, but we can start doing that. Moving this to `pulumi-analyzer-policy-python` sounds reasonable to me. We'll have to implement it twice for Unix and Windows, but these will effectively just be paired down versions of Python's virtual environment activation scripts, and won't be too large.

sdk/go/common/workspace/project.go
justinvp: We need a place to specify the version for Python and `PulumiPolicy.yaml` seemed like a logical place to do that. For Node.js, you can continue to specify the version in `package.json`, but if it is specified in `PulumiPolicy.yaml` the `PulumiPolicy.yaml` value will be used.

pkg/backend/httpstate/policypack.go
justinvp: Actually, I'm going to hold off on doing this for now. I agree it's the right long-term architecture to defer to the plugin to do this. We already have some related technical debt around installing dependencies for regular programs (which the TODO above attempts to point out, but should be linking to https://github.com/pulumi/pulumi/issues/1334, which I'll fix).

sdk/python/dist/pulumi-analyzer-policy-python
```diff
@@ -1,2 +1,43 @@
...
+
+        # Run python from the virtual environment.
+        "$virtualenv/bin/python" -u -m pulumi.policy "$1" "$2"
+    else
+        echo "\"$virtualenv\" doesn't appear to be a virtual environment"
+        exit 1
+    fi
+else
+    # Otherwise, just run python3.
+    python3 -u -m pulumi.policy "$1" "$2"
```
justinvp, line 42: We look for `-virtualenv` and use it inside these scripts, but don't pass it along (and any other options; yet) to the `pulumi.policy` module because, unfortunately, we currently check for exactly 2 arguments there, instead of a minimum of 2. So we continue to pass the two arguments we passed previously, otherwise older versions of `pulumi-policy` package will fail when it receives more than 2 args.

I've updated the `pulumi.policy` module to look for a minimum of 2 args, so in the future, we could consider passing all the args along (when we don't care as much about older versions of `pulumi` package. (I've also done the same for Node.js, which is currently looking for an exact number of args, rather than a minimum).

sdk/go/common/resource/plugin/analyzer_plugin.go
justinvp: @lukehoban, [here](https://github.com/pulumi/pulumi/pull/4644/commits/e8d1b375747206edd7a9561b07de8da029616127) is what it looks like to do this from the scripts. I've tested these E2E on Windows and Unix. PTAL.

sdk/python/dist/pulumi-analyzer-policy-python.cmd
```diff
@@ -1,5 +1,38 @@
...
+        :: Run python from the virtual environment.
+        "%pulumi_policy_python_virtualenv%\Scripts\python.exe" -u -m pulumi.policy %pulumi_policy_python_engine_address% %pulumi_policy_python_program%
+        exit /B
+    ) else (
+        echo "%pulumi_policy_python_virtualenv%" doesn't appear to be a virtual environment
+        exit 1
+    )
+) else (
+    :: Otherwise, just run python. We use `python` instead of `python3` because Windows
+    :: Python installers install only `python.exe` by default.
```
justinvp: Note: In a subsequent commit, I've converted these style of comments to use `REM` because `::` on this second line is treated as a drive separator.

sdk/go/common/resource/plugin/analyzer_plugin.go
lukehoban: Updates look good to me.  Shame to have to push this into bash, but it does feel like a more flexible interface for future extension.

PR #4646, opened 2020-05-15
[codegen/docs] Various Go-related fixes

praneetloke:
Related to pulumi/docs#3254, pulumi/docs#3192, pulumi/docs#3276.

* Remove all exclusions for conditionally showing the languages in the lang chooser.
  * This was done for k8s overlay resources, so that we could hide the Go language. They are all now supported in Go as of [v2.2.0](https://github.com/pulumi/pulumi-kubernetes/releases/tag/v2.2.0).
* Show a note in the Go "New*" function snippet that lets the user know the function name can be different from other languages and that it is not an error.
* Make the `args` input param optional for Functions in Go.
* Some other minor cleanup.

### DIFF
--- a/pkg/codegen/docs/templates/function.tmpl
+++ b/pkg/codegen/docs/templates/function.tmpl
@@ -29,10 +29,10 @@
 
 <!-- Go -->
 {{ print "{{% choosable language go %}}" }}
-{{- if ne .FunctionName.go .Header.Title }}
-// This function's name in Go will not match the name used in other languages.
-{{ end }}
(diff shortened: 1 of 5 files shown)


pkg/codegen/docs/templates/function.tmpl
```diff
@@ -30,6 +30,9 @@
 <!-- Go -->
 {{ print "{{% choosable language go %}}" }}
 <div class="highlight"><pre class="chroma"><code class="language-go" data-lang="go"><span class="k">func </span>{{ .FunctionName.go }}<span class="p">(</span>{{ htmlSafe .FunctionArgs.go }}<span class="p">) (*{{ template "linkify_param" .FunctionResult.go }}, error)</span></code></pre></div>
+{{- if ne .FunctionName.go .Header.Title }}
+> This function's name was changed for the Go SDK to avoid conflicts with another similarly named function in the package.
```
praneetloke, line 34: I could use some help with here with the wording. :)

justinvp: Perhaps something like the following?

```suggestion
> Note: This function is named `{{ .FunctionName.go }}` in the Go SDK.
```

PR #4647, opened 2020-05-15
Add YouTube video to README.md

zchase:
This PR adds an image that links to one of YouTube videos. This is to help drive views and subscriptions to our YouTube Account. The preview image we can pull from YouTube didn't' have a play button on it so I used one I had created myself. I wasn't sure where I should add the image so I put in the root. Let me know if there is better place for the image to live in.

Fixes: https://github.com/pulumi/home/issues/834

cc// @ericrudder


zchase:
@stack72 It cannot autoplay. Saying this video is embedded is a bit of stretch, it is really just a picture with a link to the video because embedding an `iframe` is not supported. This is similar behavior to the home page where we don't actually load YouTube scripts until the user clicks the "play" button, but in this case, we just link to video instead of trying to embed.

lukehoban:
Somewhat unfortunate that there is now a `youtube_preview_image.png` in the root of the repo.  I wonder if we can/should put this in an `assets` folder or similar to make it less prominent?  (we don't have any existing top level folder that quite makes sense currently)

PR #4656, opened 2020-05-18
Fix NRE in C# mocks and include the stack to result again

### LINKED ISSUE #4655: Regression in .NET unit testing: NullReferenceException on stacks with outputs
#4527 inadvertently broke mock-based unit-testing in C# in two ways:
- The stack object is not returned as part of the resource collection from `TestAsync` anymore
- If the stack has outputs, unit tests will fail with 
```
System.NullReferenceException: Object reference not set to an instance of an object.
   at Pulumi.Deployment.CompleteResourceAsync(Resource resource, ResourceArgs args, ResourceOptions options, ImmutableDictionary`2 completionSources)
```

Affected versions are Pulumi `2.2.0` and `2.2.1`.

mikhailshilkov:
Fix #4655

Fixes both points:
- If response data is null, the NRE is thrown at https://github.com/pulumi/pulumi/blob/30b12cff49c614de67c5388f003d8b65f22fae3c/sdk/dotnet/Pulumi/Deployment/Deployment_ReadOrRegisterResource.cs#L95
- Returning the stack was already something we used in examples in 2.0, so we should keep that behavior

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ CHANGELOG
 - Add support for untagged outputs in Go SDK.
  [#4640](https://github.com/pulumi/pulumi/pull/4640)
 
-- Fix a tegression in .NET unit testing.
+- Fix a Regression in .NET unit testing.
  [#4656](https://github.com/pulumi/pulumi/pull/4656)
 
 ## 2.2.1 (2020-05-13)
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ CHANGELOG
 - Add support for untagged outputs in Go SDK.
  [#4640](https://github.com/pulumi/pulumi/pull/4640)
 
+- Fix a tegression in .NET unit testing.
```
justinvp, line 9:
```suggestion
- Fix a Regression in .NET unit testing.
```

mikhailshilkov:
@justinvp Thanks for the review. I added a test that reveals both problems. Do you mind taking another quick look?

PR #4658, opened 2020-05-18
Don't mark provider token string types with json attribute in C#

mikhailshilkov:
Fix for https://github.com/pulumi/pulumi-aws/issues/966
The `region` property of the AWS provider is a token string type and is currently marked as `json: true` by mistake. That's a regression from when we switced to schema-based C# codegen.

### DIFF
--- a/pkg/codegen/dotnet/gen.go
+++ b/pkg/codegen/dotnet/gen.go
@@ -341,13 +341,10 @@ func (pt *plainType) genInputProperty(w io.Writer, prop *schema.Property, indent
 	}
 	if pt.res != nil && pt.res.IsProvider {
 		json := true
-		switch prop.Type {
-		case schema.StringType:
+		if prop.Type == schema.StringType {
 			json = false
-		default:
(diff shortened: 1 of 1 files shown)


pkg/codegen/dotnet/gen.go
```diff
@@ -339,8 +339,22 @@ func (pt *plainType) genInputProperty(w io.Writer, prop *schema.Property, indent
...
+		switch prop.Type {
+		case schema.StringType:
+			json = false
+		default:
+			switch t := prop.Type.(type) {
+			case *schema.TokenType:
+				if t.UnderlyingType == schema.StringType {
+					json = false
+				}
+			}
```
pgavlin, line 353:
```suggestion
			if t, ok := prop.Type.(*schema.TokenType): ok && t.UnderlyingType == schema.StringType {
				json = false
			}
```

mikhailshilkov: Thanks! I clearly give too much credit to one-case switches 😄 I'll work on that.

PR #4667, opened 2020-05-19
Go basic program gen

EvanBoyle:
Still missing stack outputs, but otherwise most of the way there. Definitely a lot of holes here that need to be plugged, but this establishes an e2e baseline for further development. After this, I'll be shifting towards development driven by expression test as suggested by @pgavlin


pkg/codegen/hcl2/binder_schema.go
```diff
@@ -231,6 +231,8 @@ func (b *binder) schemaTypeToType(src schema.Type) (result model.Type) {
 // The result may be a *schema.UnionType if multiple schema types are associaged with the input type.
 func GetSchemaForType(t model.Type) (schema.Type, bool) {
 	switch t := t.(type) {
+	case *model.ListType:
+		return GetSchemaForType(t.ElementType)
```
pgavlin, line 235: Let's change this to return an appropriate `schema.ArrayType`:

```suggestion
		if element, ok := GetSchemaForType(t.ElementType); ok {
			return &schema.ArrayType{ElementType: element}, true
		}
		return nil, false
```

Maybe do the same for `*model.MapType` while you're in here.

pkg/codegen/go/utilities.go
```diff
@@ -0,0 +1,20 @@
+package gen
+
+import "github.com/pulumi/pulumi/sdk/v2/go/common/util/contract"
+
+// TODO: either fill out rest of providers or find more scalable solution
```
pgavlin, line 5: We should pull this from the package's schema (see `hcl2.Program.Packages()`). In the fullness of time, we might need to link from a schema type back to its package in the event that different versions of the same package are in use).

pkg/codegen/hcl2/binder_schema.go
pgavlin: FWIW, I'm not certain that this is the right approach for composite types in general. I think I'd rather be faithful to the annotations on the type, but this would require allowing all composite types to be annotated. Something to think about for the future; what we have is fine.

pkg/codegen/go/gen_program_expressions.go
```diff
@@ -0,0 +1,320 @@
...
+
+type nameInfo int
+
+func (nameInfo) Format(name string) string {
+	// TODO
+	return name
+}
+
+// rewriteExpression amends the expression with intrinsics for C# generation.
+func (g *generator) rewriteExpression(expr model.Expression, typ model.Type) model.Expression {
```
pgavlin, line 310:
```suggestion
func (g *generator) lowerExpression(expr model.Expression, typ model.Type) model.Expression {
```

pkg/codegen/internal/test/testdata/aws-s3-logging.pp.go
```diff
@@ -1,11 +1,26 @@
...
 )
 
 func main() {
 	pulumi.Run(func(ctx *pulumi.Context) error {
+		logs, err := s3.NewBucket(ctx, "logs", nil)
+		if err != nil {
+			return err
+		}
+		bucket, err := s3.NewBucket(ctx, "bucket", &s3.BucketArgs{
+			Loggings: &s3.BucketLoggingArray{
```
pgavlin, line 15: This should just be `s3.BucketLoggingArray` (no address-of operator).

pkg/codegen/go/gen_program_expressions.go
```diff
@@ -0,0 +1,336 @@
...
+		g.Fgenf(w, "%.20v.Length", expr.Args[0])
+	case "lookup":
+		g.genNYI(w, "Lookup")
+	case "range":
+		g.genNYI(w, "call %v", expr.Name)
+		// g.genRange(w, expr, false)
+	case "readFile":
+		g.genNYI(w, "ReadFile")
+	case "readDir":
+		g.Fgenf(w, "Directory.GetFiles(%.v).Select(Path.GetFileName)", expr.Args[0])
```
mikhailshilkov, line 136: Add a TODO to convert from C# here

pkg/codegen/go/gen_program_expressions.go
```diff
@@ -0,0 +1,336 @@
...
+		case hcl.TraverseAttr:
+			key = cty.StringVal(part.Name)
+		case hcl.TraverseIndex:
+			key = part.Key
+		default:
+			contract.Failf("unexpected traversal part of type %T (%v)", part, part.SourceRange())
+		}
+
+		if model.IsOptionalType(model.GetTraversableType(parts[i])) {
+			g.Fgen(w, "?")
```
mikhailshilkov, line 303: This also looks like C#

pkg/codegen/go/gen_program_expressions.go
```diff
@@ -0,0 +1,333 @@
...
+			token := objType.Token
+			tokenRange := expr.SyntaxNode().Range()
+
+			_, module, member, diags := hcl2.DecomposeToken(token, tokenRange)
+			importPrefix := strings.Split(module, "/")[0]
+			contract.Assert(len(diags) == 0)
+			return fmt.Sprintf("%s.%s", importPrefix, member)
+		} else if tupType, ok := schemaType.(*schema.ArrayType); ok {
+			fmt.Println(tupType)
+		}
```
pgavlin, line 278:
```suggestion
		switch schemaType := schemaType.(type) {
		case *schema.ArrayType:
			token := schemaType.ElementType.(*schema.ObjectType).Token
			tokenRange := expr.SyntaxNode().Range()

			_, module, member, diags := hcl2.DecomposeToken(token, tokenRange)
			importPrefix := strings.Split(module, "/")[0]
			contract.Assert(len(diags) == 0)
			return fmt.Sprintf("%s.%sArray", importPrefix, member)
		case *schema.ObjectType:
			token := schemaType.Token
			tokenRange := expr.SyntaxNode().Range()

			_, module, member, diags := hcl2.DecomposeToken(token, tokenRange)
			importPrefix := strings.Split(module, "/")[0]
			contract.Assert(len(diags) == 0)
			return fmt.Sprintf("%s.%s", importPrefix, member)
		default:
			contract.Failf("unexpected schema type %T", schemaType)
		}
```

pkg/codegen/go/gen_program_expressions.go
```diff
@@ -0,0 +1,333 @@
...
+	}
+
+	return ""
```
pgavlin, line 281: Not sure if you want to pepper your code with `TODO`s, but this method will need additions for plain types, etc.

Issue #4669, opened 2020-05-19
Unable to use stack transformation from unit test


clstokes:
If I call `pulumi.runtime.registerStackTransformation` from my Pulumi application and run a unit test against my application, the tests fail to run with the error `The root stack resource was referenced before it was initialized.` The same Pulumi application runs successfully without error with `pulumi preview`.

#### Steps to reproduce
1. `git clone https://gist.github.com/20db9640bc9f3c8607f6d67d8390fb45 gh-test && cd gh-test`
1. `npm i`
1. `pulumi stack init gh-test`
1. `npm test`
    - Observe error.
1. `pulumi preview`
    - Observe successful preview.

lukehoban:
This happens because we do not currently create a root `Stack` resource automatically in test mode.  It is likely that we should, and that `setMocks` should initialize a root `Stack` automatically .  Coincidentally, we are looking at making that change for Python for other reasons in https://github.com/pulumi/pulumi/pull/4670, but it is likely we should make that same change in Node.js which we fix this issue.

In the meantime, a workaround is to use `pulumi.runtime.runInPulumiStack` to construct a Stack resource.

```ts
pulumi.runtime.setMocks({
    call: x => {
        return {};
    },
    newResource: x => {
        return {id: "", state: {}};
    }
})
pulumi.runtime.runInPulumiStack(async () => {
    pulumi.runtime.registerStackTransformation(args => {
        console.log(`transofming: ${args.name}`)
        return {
            opts: args.opts,
            props: args.props,
        }
    })
    
    // Create an AWS resource (S3 Bucket)
    const bucket = new aws.s3.Bucket("my-bucket");
    
    // Export the name of the bucket
    return {
        bucketName: bucket.id
    };
})
```

amichel:
This problem scope is bigger than just unit tests. I have common transformations that I want to be able to reuse in all projects as npm package. It also causes problem the other way around, when I register stack transformation, it doesn't apply on resources created by modules from shared npm packages.

clstokes:
Related - https://github.com/pulumi/pulumi/issues/6214

arwilczek90:
I agree with @amichel I get the same traceback when running pulumi up on a project with a stack transformation.

PR #4670, opened 2020-05-19
Allow `pulumi.export` calls from unit tests

### LINKED ISSUE #4478: Python tests fail on exports

justinvp:
The [previous attempt](https://github.com/pulumi/pulumi/pull/4518) to allow this didn't actually allow it, so this is take two. As part of the previous attempt, I thought after tweaking the test I had observed the test failing, and then succeeding after making the product changes, but I must have been mistaken.

It turns out that our existing "testing with mocks" tests _weren't running at all_ because of a missing `__init__.py` file. Once the missing `__init__.py` is added, the tests run, but other tests (the "test mode" tests) fail because the code that creates the mocks and resources will run during test discovery, and setting the mocks modifies global state.

To address the test issue, I've moved the mocks tests into their own `test_with_mocks` package that can be run separately from other tests.

To address the actual issue, a root Stack resource is initialized when the mocks are set.

Fixes #4478


lukehoban:
> To address the actual issue, a root Stack resource is initialized when the mocks are set.

Should we be doing this consistently across languages?  I believe this is the root cause of https://github.com/pulumi/pulumi/issues/4669 as well - and quite likely other issues.

justinvp:
> > To address the actual issue, a root Stack resource is initialized when the mocks are set.
> 
> Should we be doing this consistently across languages?

I think we should do this for Node.js too, but would love to hear @pgavlin's, and @mikhailshilkov's thoughts. Notably, .NET and Go don't really have these problems and already have a root stack created as part of running with mocks just by the nature of how mocks are used in those languages.

justinvp:
Confirmed in the Travis logs that the "testing with mocks" tests are now actually running and succeeding:

```
pipenv run python -m unittest discover -s lib/test_with_mocks -v
3258test_component (test_testing_with_mocks.TestingWithMocks) ... ok
3259test_custom (test_testing_with_mocks.TestingWithMocks) ... ok
3260test_invoke (test_testing_with_mocks.TestingWithMocks) ... ok
```

lukehoban:
> Do we still have an issue to track Node.js or should we make one?

#4669 is tracking this.

PR #4673, opened 2020-05-20
Fix C# (and a bit of Node.js) program gen based on errors from AWS

mikhailshilkov:
I generated C# examples in the AWS provider locally and made a tool that compiles all examples and reports errors in an actionable way.

I fixed some errors that were found in C# generation (and one in Node.js). Remaining errors are probably problems in the examples themselves or in our HCL compiler frontend. I'll report those separately.


pgavlin:
> and made a tool that compiles all examples and reports errors in an actionable way

Can you share the source for this somewhere?

PR #4685, opened 2020-05-22
Support secret default values in schema and codegen

lukehoban:
Currently, the secretness is only applied to default values used as input values for resources.  It is not applied to the defaults expoed in the `config` modules.  Doing the latter may also be desired - but (a) it would be a breaking change as the types would need to change from primitives to Outputs and (b) users can always manually wrap value retreived from this module with `pulumi.secret` if desired (a choice they do not have for the default resource inputs case).

Part of pulumi/pulumi-azure#576.

### DIFF
--- a/pkg/codegen/dotnet/gen.go
+++ b/pkg/codegen/dotnet/gen.go
@@ -447,6 +447,9 @@ func (pt *plainType) genInputType(w io.Writer, level int) error {
 				return err
 			}
 			propertyName := pt.mod.propertyName(prop)
+			if prop.DefaultValue.Secret {
+				dv = fmt.Sprintf("Output.CreateSecret(%s)", dv)
+			}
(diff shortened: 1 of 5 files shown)


stack72:
I'm a big fan of this change 🥳

pgavlin:
This change looks good. I'm wondering, though, if we should instead support marking entire input/output properties as secret? Although the capabilities of the two options are slightly different--marking defaults as secret gives the user a bit more control over whether or not their own inputs for a property is secret--I think secret properties gives us a better overall outcome w.r.t. pits of success.

lukehoban:
We ended up abandoning this approach as it doesn't fully address pulumi/pulumi-azure#576.  We may come back to a feature like this for other reasons - and would most likely do what is suggested in https://github.com/pulumi/pulumi/pull/4685#issuecomment-632771718 instead.

PR #4686, opened 2020-05-22
Convert primitive types in C# program gen

mikhailshilkov:
TF examples aren't particularly consistent about using correct primitive types in property assignments. So, a property of type `integer` (e.g. `Port`) may get a `string` value (e.g. `"443"`), or vice versa. AWS examples have instances of these inconsistencies for bool, double, int to string and back.

This PR adds conversions between these primitive types to generated C# programs, either by converting a literal in place (the most common case), or adding a conversion call.

### DIFF
--- a/pkg/codegen/hcl2/rewrite_convert.go
+++ b/pkg/codegen/hcl2/rewrite_convert.go
@@ -2,11 +2,13 @@ package hcl2
 
 import (
 	"github.com/hashicorp/hcl/v2"
+	"github.com/hashicorp/hcl/v2/hclsyntax"
 	"github.com/pulumi/pulumi/pkg/v2/codegen"
 	"github.com/pulumi/pulumi/pkg/v2/codegen/hcl2/model"
 	"github.com/pulumi/pulumi/pkg/v2/codegen/schema"
 	"github.com/pulumi/pulumi/sdk/v2/go/common/util/contract"
 	"github.com/zclconf/go-cty/cty"
+	"github.com/zclconf/go-cty/cty/convert"
 )
 
 func sameSchemaTypes(xt, yt model.Type) bool {
@@ -116,5 +118,78 @@ func RewriteConversions(x model.Expression, to model.Type) model.Expression {
 	if to.AssignableFrom(x.Type()) && sameSchemaTypes(to, x.Type()) {
 		return x
 	}
+	// If we can convert a primitive value in place, do so.
+	if value, ok := convertPrimitiveValues(x, to); ok {
+		return value
+	}
 	return NewConvertCall(x, to)
+}
+
+// convertPrimitiveValues returns a new expression if the given expression can be converted to another primitive type
+// (bool, int, number, string) that matches the target type.
+func convertPrimitiveValues(from model.Expression, to model.Type) (model.Expression, bool) {
+	var expression model.Expression
+	switch {
+	case to.AssignableFrom(model.DynamicType):
+		return nil, false
+	case to.AssignableFrom(model.BoolType):
+		if stringLiteral, ok := extractStringValue(from); ok {
+			if value, err := convert.Convert(cty.StringVal(stringLiteral), cty.Bool); err == nil {
+				expression = &model.LiteralValueExpression{Value: value}
+			}
+		}
+	case to.AssignableFrom(model.IntType), to.AssignableFrom(model.NumberType):
+		if stringLiteral, ok := extractStringValue(from); ok {
+			if value, err := convert.Convert(cty.StringVal(stringLiteral), cty.Number); err == nil {
+				expression = &model.LiteralValueExpression{Value: value}
+			}
+		}
+	case to.AssignableFrom(model.StringType):
+		if stringValue, ok := convertLiteralToString(from); ok {
+			expression = &model.TemplateExpression{
+				Parts: []model.Expression{&model.LiteralValueExpression{
+					Value: cty.StringVal(stringValue),
+				}},
+			}
+		}
+	}
+	if expression == nil {
+		return nil, false
+	}
+
+	expression.SetLeadingTrivia(from.GetLeadingTrivia())
+	expression.SetTrailingTrivia(from.GetTrailingTrivia())
+	return expression, true
+}
+
+// extractStringValue returns a string if the given expression is a template expression containing a single string
+// literal value.
+func extractStringValue(arg model.Expression) (string, bool) {
+	template, ok := arg.(*model.TemplateExpression)
+	if !ok || len(template.Parts) != 1 {
+		return "", false
+	}
+	lit, ok := template.Parts[0].(*model.LiteralValueExpression)
+	if !ok || lit.Type() != model.StringType {
+		return "", false
+	}
+	return lit.Value.AsString(), true
+}
+
+// convertLiteralToString converts a literal of type Bool, Int, or Number to its string representation. It also handles
+// the unary negate operation in front of a literal number.
+func convertLiteralToString(from model.Expression) (string, bool) {
+	switch expr := from.(type) {
+	case *model.UnaryOpExpression:
+		if expr.Operation == hclsyntax.OpNegate {
+			if operandValue, ok := convertLiteralToString(expr.Operand); ok {
+				return "-" + operandValue, true
+			}
+		}
+	case *model.LiteralValueExpression:
+		if stringValue, err := convert.Convert(expr.Value, cty.String); err == nil {
+			return stringValue.AsString(), true
+		}
+	}
+	return "", false
 }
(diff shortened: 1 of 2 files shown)


mikhailshilkov:
One potential question is whether this should happen on an earlier stage during HCL binding

pkg/codegen/dotnet/gen_program_expressions.go
```diff
@@ -252,15 +253,78 @@ func (g *generator) genFunctionUsings(x *model.FunctionCallExpression) []string
...
+
+	return "", false
+}
+
+func (g *generator) genConvert(w io.Writer, expr *model.FunctionCallExpression) {
+	switch arg := expr.Args[0].(type) {
+	case *model.ObjectConsExpression:
+		g.genObjectConsExpression(w, arg, expr.Type())
+	default:
+		if targetType, ok := hcl2.GetSchemaForType(expr.Type().(model.Type)); ok {
```
pgavlin, line 277: You shouldn't need the call to `GetSchemaForType` here. Instead, I'd use something like

```
switch model.ResolveOutputs(expr.Type()) {
case model.BoolType:
    ...
case model.IntType:
    ...
case model.NumberType:
    ...
case model.StringType:
    ...
}
```

mikhailshilkov: Good to know. I thought that only `GetSchemaForType` can disambiguate unions etc. I'll try on Monday!

mikhailshilkov:
> Can you elaborate?

My vague thought was that instead of converting these values in every (statically typed) language, the type checker of IL could make the same transformations. For example, if it finds a literal `80` assigned to a field of type `string`, it could change the literal to `"80"`, so that every language gets it right automatically.

Of course, it can't do likes of `Convert.ToInt32`, but actually all practical examples I found so far were about literals typed incorrectly.

I don't know IL is the correct level to do so, but it sounds like potentially less error-prone, so worth explicitly discussing.

pgavlin:
> My vague thought was that instead of converting these values in every (statically typed) language, the type checker of IL could make the same transformations. For example, if it finds a literal 80 assigned to a field of type string, it could change the literal to "80", so that every language gets it right automatically.

I don't think that this is something we want to do at bind time, as it would cause us to lose information about the original structure of the program. However, this is exactly the sort of thing we can do as a lowering step. The transformation you're specifically suggesting--static conversion of literals from one type to another--is a perfect candidate for a lowering phase exported by the `model` package.

mikhailshilkov:
@pgavlin I re-implemented the PR inside to do primitive conversions in `RewriteConversions`. Let me know what you think.

pkg/codegen/hcl2/rewrite_convert.go
```diff
@@ -116,5 +118,105 @@ func RewriteConversions(x model.Expression, to model.Type) model.Expression {
...
+			if f, err := strconv.ParseFloat(stringLiteral, 64); err == nil {
+				value = cty.NumberFloatVal(f)
+			}
+		}
+	case model.StringType:
+		stringValue := convertLiteralToString(from)
+		if stringValue != "" {
+			value = cty.StringVal(stringValue)
+		}
+	}
```
pgavlin, line 156: I'd suggest that you use the `AssignableFrom` method for these checks. This will properly pick up more complicated unions such as those generated by the use `oneOf` in a Pulumi schema. This will also allow you to remove the `resolveUnderlyingType` method.

```suggestion
	// This check is necessary due to #4703
	to = model.ResolveOutputs(to)
	if to == model.DynamicType {
		return nil, false
	}
	switch {
	case to.AssignableFrom(model.BoolType):
		if stringLiteral, ok := extractStringValue(from); ok {
			if v, err := convert.Convert(cty.StringVal(stringLiteral), cty.Bool); err == nil {
				value = v
			}
		}
	case to.AssignableFrom(model.IntType):
		if stringLiteral, ok := extractStringValue(from); ok {
			if v, err := convert.Convert(cty.StringVal(stringLiteral), cty.Bool); err == nil {
				value = v
			}
		}
	case to.AssignableFrom(model.IntType), to.AssignableFrom(model.NumberType):
		if stringLiteral, ok := extractStringValue(from); ok {
			if v, err := convert.Convert(cty.StringVal(stringLiteral), cty.Bool); err == nil {
				value = v
			}
		}
	case to.AssignableFrom(model.StringType):
		if stringValue, ok := convertLiteralToString(from); ok {
			value = cty.StringVal(stringValue)
		}
	default:
		return nil, false
	}
```

pkg/codegen/hcl2/rewrite_convert.go
```diff
@@ -116,5 +118,105 @@ func RewriteConversions(x model.Expression, to model.Type) model.Expression {
...
+	if value == cty.NilVal {
+		return nil, false
+	}
+
+	stringLiteral := model.LiteralValueExpression{Value: value}
```
pgavlin, line 161: String literals in expressions are a bit weird: in most contexts, they are represented as single-part literals. This includes any position in which we will insert a conversion, so this should be:

```suggestion
	stringLiteral := &model.TemplateExpression{
		Parts: []model.Expression{&model.LiteralValueExpression{
			Value: value,
		}},
	}
	stringLiteral.SetLeadingTrivia(from.GetLeadingTrivia())
	stringLiteral.SetTrailingTrivia(from.GetTrailingTrivia())
	return stringLiteral, true
```

pkg/codegen/hcl2/rewrite_convert.go
```diff
@@ -116,5 +118,105 @@ func RewriteConversions(x model.Expression, to model.Type) model.Expression {
...
+	case *model.TemplateExpression:
+		if arg.Type() != model.StringType || len(expr.Parts) != 1 {
+			return "", false
+		}
+		if lit, ok := expr.Parts[0].(*model.LiteralValueExpression); ok && lit.Type() == model.StringType {
+			return lit.Value.AsString(), true
+		}
+	}
+
+	return "", false
```
pgavlin, line 198: The single-element switch is a bit odd here. I'd use a simple type assertion:
```suggestion
	template, ok := arg.(*model.TemplateExpression)
	if !ok || len(template.Parts) != 1 {
		return "", false
	}
	lit, ok := template.Parts[0].(*model.LiteralValueExpression)
	if !ok {
		return "", false
	}
	return lit.Value.AsString(), true
```

pkg/codegen/hcl2/rewrite_convert.go
```diff
@@ -116,5 +118,105 @@ func RewriteConversions(x model.Expression, to model.Type) model.Expression {
...
+}
+
+// convertLiteralToString converts a literal of type Bool, Int, or Number to its string representation. It also handles
+// the unary negate operation in front of a literal number.
+func convertLiteralToString(from model.Expression) string {
```
pgavlin, line 203: Let's have this return `(string, bool)` and be more explicit about the check.

pkg/codegen/hcl2/rewrite_convert.go
```diff
@@ -116,5 +118,87 @@ func RewriteConversions(x model.Expression, to model.Type) model.Expression {
...
+// convertPrimitiveValues returns a new expression if the given expression can be converted to another primitive type
+// (bool, int, number, string) that matches the target type.
+func convertPrimitiveValues(from model.Expression, to model.Type) (model.Expression, bool) {
+	var expression model.Expression
+	switch {
+	case to.AssignableFrom(model.DynamicType):
+		return nil, false
+	case to.AssignableFrom(model.BoolType):
+		if stringLiteral, ok := extractStringValue(from); ok {
+			if b, err := strconv.ParseBool(stringLiteral); err == nil {
```
pgavlin, line 137: I realize now that in my proposed code I forgot to indicate where the `convert.Convert` calls were coming from. These calls are provided by `github.com/zclconf/go-cty/cty/convert`, and are exactly the APIs used when evaluating HCL2 expressions at runtime. We should use them here as well for correctness.

mikhailshilkov: Ah, I see now. Changed to `convert.Convert`.

pkg/codegen/hcl2/rewrite_convert.go
```diff
@@ -116,5 +119,81 @@ func RewriteConversions(x model.Expression, to model.Type) model.Expression {
...
+	case *model.UnaryOpExpression:
+		if expr.Operation == hclsyntax.OpNegate {
+			if operandValue, ok := convertLiteralToString(expr.Operand); ok {
+				return "-" + operandValue, true
+			}
+		}
+	case *model.LiteralValueExpression:
+		switch expr.Type() {
+		case model.BoolType:
+			return strconv.FormatBool(expr.Value.True()), true
```
pgavlin, line 193: Let's use `convert.Convert` in this method as well.

PR #4690, opened 2020-05-23
Add support for publishing/downloading policy packs to/from a service-managed local filesystem

praneetloke:
I was working on something unrelated and needed this functionality, so I implemented this for my own sake since it wasn't too much work. This could be useful for [self-hosted](https://www.pulumi.com/docs/guides/self-hosted/) installations of Pulumi Service that uses an NAS for its object store. 

Creating this as a draft PR because I am happy to help get this PR to its final state, if need be, or defer to someone else who might want to take this over. Just thought this might be useful for someone else working on this to reuse any part of this, if they want to.

I am not sure how helpful these console output statements are, but here they are:

### Publishing the policy pack
```
➜  policy-playground /opt/pulumi/bin/pulumi policy publish saml-org
Obtaining policy metadata from policy plugin
Compressing policy pack
Uploading policy pack to Pulumi service
Publishing "aws-typescript" - version 0.0.3 to "saml-org"

Permalink: http://localhost:3000/saml-org/policypacks/aws-typescript/0.0.3
```

### Running a simple Pulumi app with the policy applied to the default group
```
➜  policy-playground-2 pulumi policy ls saml-org
NAME            VERSIONS
aws-typescript  0.0.3

➜  policy-playground-2 pulumi whoami -v         
User: praneetloke
Backend URL: http://localhost:3000/praneetloke

➜  policy-playground-2 AWS_PROFILE=pulumi-dev-sandbox /opt/pulumi/bin/pulumi preview
Previewing update (saml-org/dev):
Installing policy pack aws-typescript 0.0.3...
[resource plugin aws-2.5.0] installing
Finished installing policy pack

     Type                 Name                     Plan       
 +   pulumi:pulumi:Stack  policy-playground-2-dev  create     
 +   └─ aws:s3:Bucket     my-bucket                create     
 
Policy Violations:
    [advisory]  aws-typescript v0.0.3  s3-no-public-read (my-bucket: aws:s3/bucket:Bucket)
    Prohibits setting the publicRead or publicReadWrite permission on AWS S3 buckets.
    You cannot set public-read or public-read-write on an S3 bucket. Read more about ACLs here: https://docs.aws.amazon.com/AmazonS3/latest/dev/acl-overview.html
    
Permalink: http://localhost:3000/saml-org/policy-playground-2/dev/previews/e25e6748-3bd2-43de-91e6-ad3598a7959c
```

The `saml-org` in the output above is in my instance of self-hosted Pulumi Service.

### DIFF
--- a/pkg/backend/httpstate/client/client.go
+++ b/pkg/backend/httpstate/client/client.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/pulumi/pulumi/pkg/v2/backend/policypack"
 	"io"
 	"net/http"
 	"path"
@@ -31,6 +30,7 @@ import (
 	"github.com/blang/semver"
 	"github.com/pkg/errors"
 
+	"github.com/pulumi/pulumi/pkg/v2/backend/policypack"
 	"github.com/pulumi/pulumi/pkg/v2/engine"
 	"github.com/pulumi/pulumi/pkg/v2/util/validation"
 	"github.com/pulumi/pulumi/sdk/v2/go/common/apitype"
--- a/pkg/backend/httpstate/client/client.go
+++ b/pkg/backend/httpstate/client/client.go
@@ -18,8 +18,8 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/pulumi/pulumi/pkg/v2/backend/policypack"
 	"io"
-	"io/ioutil"
 	"net/http"
 	"path"
 	"regexp"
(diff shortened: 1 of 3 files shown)


praneetloke:
cc @leezen as well for planning purposes.

praneetloke:
Thank you for reviewing this, @justinvp!

> Would we actually return UploadURI with a file scheme in such cases? I was thinking the backing store would be an implementation detail of the service, and that an UploadURI with an https (or http) scheme would always be returned from the API such that the client is always performing an HTTP PUT to upload the bits. This would keep the service API and client simpler.

Great question! What you are suggesting would be great. However, to defer the upload and download to the service does involve a fair bit of a change since the CLI is currently handling both of those responsibilities given a URI. I am not sure what the original intent was either. @chrsmith may know the answer to that. One thing to keep in mind (and my understanding) is that it would only solve our service-managed backend. The way I have it right now would allow for some extensibility and reuse of publishing/downloading policy packs to/from another self-managed object store, aka while using the `filestate` backend.

As for the "merge likeliness" of this PR, I didn't want to throw away the modifications I made in case it is useful to support a file scheme[1] for a service-managed backend. I see that Lee has created https://github.com/pulumi/pulumi-service/issues/5003, which would be good to look into this PR as part of its design work and see if any/no part of this PR can be used.

[1]=note that there is an issue currently where the service does not return the `file://` scheme in the `UploadURI`, so this PR interprets the lack of one to be the same as `file://`

leezen:
Looks like we had https://github.com/pulumi/pulumi-service/issues/4222 already, so that's probably the better issue to reference

PR #4699, opened 2020-05-26
Add F# operators for InputUnion

mastoj:
Having to write `InputUnion.op_Implicit(80)` to create an `InputUnion` isn't very nice in F#, added operators so you should be able to write write:

```
let x: InputUnion<int, string> = inputUnion1Of1 80
let y: InputUnion<int, string> = inputUnion2Of2 "hello world"
```


mikhailshilkov:
That's great! Do you mind adding a short note to the changelog?

PR #4716, opened 2020-05-28
[Go Program Gen] binary, unary, and initial conditional expression support

EvanBoyle:
Go doesn't support ternary expressions so we must convert:

```
true ? 1 : 0
```
into a temp variable and assignment within if/else:

```go
var tmp0 float64
if true {
tmp0 = 1
} else {
tmp0 = 0
}
```

Need more conditional test cases for things like assignment within object cons, and nesting, but this seems like enough to review for now.

### DIFF
--- a/pkg/codegen/go/gen_program_expressions.go
+++ b/pkg/codegen/go/gen_program_expressions.go
@@ -79,8 +79,41 @@ func (g *generator) GenAnonymousFunctionExpression(w io.Writer, expr *model.Anon
 	g.Fgenf(w, "\n}")
 }
 
-// GenBinaryOpExpression generates code for a BinaryOpExpression.
-func (g *generator) GenBinaryOpExpression(w io.Writer, expr *model.BinaryOpExpression) { /*TODO*/ }
+func (g *generator) GenBinaryOpExpression(w io.Writer, expr *model.BinaryOpExpression) {
+	opstr, precedence := "", g.GetPrecedence(expr)
+	switch expr.Operation {
+	case hclsyntax.OpAdd:
+		opstr = "+"
+	case hclsyntax.OpDivide:
+		opstr = "/"
+	case hclsyntax.OpEqual:
+		opstr = "=="
+	case hclsyntax.OpGreaterThan:
+		opstr = ">"
+	case hclsyntax.OpGreaterThanOrEqual:
+		opstr = ">="
+	case hclsyntax.OpLessThan:
+		opstr = "<"
+	case hclsyntax.OpLessThanOrEqual:
+		opstr = "<="
+	case hclsyntax.OpLogicalAnd:
+		opstr = "&&"
+	case hclsyntax.OpLogicalOr:
+		opstr = "||"
+	case hclsyntax.OpModulo:
+		opstr = "%"
+	case hclsyntax.OpMultiply:
+		opstr = "*"
+	case hclsyntax.OpNotEqual:
+		opstr = "!="
+	case hclsyntax.OpSubtract:
+		opstr = "-"
+	default:
+		opstr, precedence = ",", 1
+	}
+
+	g.Fgenf(w, "%.[1]*[2]v %[3]v %.[1]*[4]o", precedence, expr.LeftOperand, opstr, expr.RightOperand)
+}
 
 // GenConditionalExpression generates code for a ConditionalExpression.
 func (g *generator) GenConditionalExpression(w io.Writer, expr *model.ConditionalExpression) {}
--- a/pkg/codegen/go/gen_program_expressions.go
+++ b/pkg/codegen/go/gen_program_expressions.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"math/big"
+	"reflect"
 	"strings"
 
 	"github.com/hashicorp/hcl/v2"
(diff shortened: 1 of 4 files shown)


EvanBoyle:
Added support for nested ternaries and ternaries within object cons expressions.

pkg/codegen/go/gen_program_ternaries.go
```diff
@@ -0,0 +1,67 @@
...
+		t, _ := ta.allocateExpression(x.TrueResult)
+		f, _ := ta.allocateExpression(x.FalseResult)
+		temp = &ternaryTemp{
+			Name:         fmt.Sprintf("tmp%d", len(ta.temps)),
+			VariableType: x.Type(),
+			Condition:    cond,
+			TrueResult:   t,
+			FalseResult:  f,
+		}
+		ta.temps = append(ta.temps, temp)
```
pgavlin, line 50: This representation loses any trivia information associated with the `?` and `:` tokens from the original `ConditionalExpression`. I suggest that we define `ternaryTemp` as

```go
type ternaryTemp struct {
	Name         string
	Value *model.ConditionalExpression
 }
```

And then rewrite the input expression's `Condition`, `True`, and `False` fields:

```suggestion
	case *model.ConditionalExpression:
		x.Condition, _ = ta.allocateExpression(x.Condition)
		x.TrueResult, _ = ta.allocateExpression(x.TrueResult)
		x.FalseResult, _ = ta.allocateExpression(x.FalseResult)
		temp = &ternaryTemp{
			Name:         fmt.Sprintf("tmp%d", len(ta.temps)),
			Value:    x,
		}
		ta.temps = append(ta.temps, temp)
```

pkg/codegen/go/gen_program_expressions.go
```diff
@@ -304,16 +365,28 @@ func (g *generator) argumentTypeName(expr model.Expression, destType model.Type,
...
-	switch destType.(type) {
+	switch destType := destType.(type) {
 	case *model.OpaqueType:
-		return destType.String()
+		switch destType.Name {
+		case "number":
+			return "float64"
+		default:
+			return destType.String()
+		}
```
pgavlin, line 386:
```suggestion
		switch destType {
		case model.NumberType:
			return "float64"
		default:
			return destType.Name
		}
```

pkg/codegen/go/gen_program_expressions.go
```diff
@@ -79,11 +80,47 @@ func (g *generator) GenAnonymousFunctionExpression(w io.Writer, expr *model.Anon
...
 
-// GenConditionalExpression generates code for a ConditionalExpression.
-func (g *generator) GenConditionalExpression(w io.Writer, expr *model.ConditionalExpression) {}
+	g.Fgenf(w, "%.[1]*[2]v %[3]v %.[1]*[4]o", precedence, expr.LeftOperand, opstr, expr.RightOperand)
+}
+
+func (g *generator) GenConditionalExpression(w io.Writer, expr *model.ConditionalExpression) {
+	// Ternary expressions are not supported in go so we need to allocate temp variables in the parent scope.
+	// This is handled by lower expression and rewriteTernaries
+	contract.Failf("BUG: this conditional expression needs to be lowered.")
```
pgavlin, line 122: Nit: we don't usually say "BUG" like this :P

```suggestion
	contract.Failf("unlowered conditional expression @ %v". expr.SyntaxNode().Range())
```

pkg/codegen/go/gen_program_ternaries.go
```diff
@@ -0,0 +1,64 @@
...
+	temps []*ternaryTemp
+}
+
+func (ta *tempAllocator) allocateExpression(x model.Expression) (model.Expression, hcl.Diagnostics) {
+	var temp *ternaryTemp
+	switch x := x.(type) {
+	case *model.ConditionalExpression:
+		cond, _ := ta.allocateExpression(x.Condition)
+		t, _ := ta.allocateExpression(x.TrueResult)
+		f, _ := ta.allocateExpression(x.FalseResult)
```
pgavlin, line 39: These shouldn't be necessary--you should be able to do `x.Condition, _ = ta.allocateExpression(x.Condition)`. Is that not working?

pkg/codegen/go/gen_program_ternaries.go
```diff
@@ -0,0 +1,64 @@
...
+
+func (tt *ternaryTemp) Traverse(traverser hcl.Traverser) (model.Traversable, hcl.Diagnostics) {
+	return tt.Type().Traverse(traverser)
+}
+
+func (tt *ternaryTemp) SyntaxNode() hclsyntax.Node {
+	return syntax.None
+}
+
+type tempAllocator struct {
```
pgavlin, line 29: Nit: technically, what we're doing is spilling, so I'd name this `tempSpiller`
```suggestion
type tempSpiller struct {
```

pkg/codegen/go/gen_program_ternaries.go
EvanBoyle: ah yes, forgot to drop the `:` when I tried to make the change the first time.

pkg/codegen/go/gen_program_ternaries.go
```diff
@@ -0,0 +1,62 @@
...
+
+func (tt *ternaryTemp) SyntaxNode() hclsyntax.Node {
+	return syntax.None
+}
+
+type tempSpiller struct {
+	temps []*ternaryTemp
+}
+
+func (ta *tempSpiller) allocateExpression(x model.Expression) (model.Expression, hcl.Diagnostics) {
```
pgavlin, line 33:
```suggestion
func (ta *tempSpiller) spillExpression(x model.Expression) (model.Expression, hcl.Diagnostics) {
```

PR #4732, opened 2020-05-29
improve dev version detection

jaxxstorm:
Currently the `isDevVersion` function only tests for "dev" versions.
When installing pulumi from Homebrew with the `--HEAD` flag we get an
alpha version locally. This can mean the user gets the upgrade prompt
when it's not valid.

This change changes the logic to check against a regex for all the
possible semver `pre` versions.

It also adds a unit test to ensure we're getting the right result

### DIFF
--- a/pkg/cmd/pulumi/pulumi_test.go
+++ b/pkg/cmd/pulumi/pulumi_test.go
@@ -1,0 +1,35 @@
+// Copyright 2016-2020, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


pkg/cmd/pulumi/pulumi_test.go
```diff
@@ -0,0 +1,35 @@
...
+	"testing"
+
+	"github.com/blang/semver"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestIsDevVersion(t *testing.T) {
+
+	// This function primarily focuses on the "Pre" section of the semver string
+	// So we'll focus on testing that
```
justinvp, line 26: Nit:

```suggestion
	// This function primarily focuses on the "Pre" section of the semver string,
	// so we'll focus on testing that.
```

pkg/cmd/pulumi/pulumi_test.go
```diff
@@ -0,0 +1,35 @@
...
+	stableVer, _ := semver.ParseTolerant("1.0.0")
+	devVer, _ := semver.ParseTolerant("v1.0.0-dev")
+	alphaVer, _ := semver.ParseTolerant("v1.0.0-alpha.1590772212+g4ff08363.dirty")
```
justinvp, line 29: Add cases for `beta` and `rc`?

jaxxstorm: yep, good point, was being lazy. Added!

PR #4736, opened 2020-05-29
[codegen/nodejs/sdk] Split required {in,out}puts.

pgavlin:
These changes are necessary in order to move tfgen over to the schema
code generator.

The schema generator for tfgen has an annoying behavior for nested types
in which it does not separate input and output types. Worse, the shape
of the type that results from a collision of input and output types is
order-dependent: whichever of the two was observed last wins (note that
the shape is still determenistic, as the order in which types are
recorded is always the same). As a result, the NodeJS code generator
needs to know the set of required properties for the other aspect of the
type in order to generate proper code.

Contributes to
https://github.com/pulumi/pulumi-terraform-bridge/issues/179.

### DIFF
--- a/pkg/codegen/nodejs/gen.go
+++ b/pkg/codegen/nodejs/gen.go
@@ -688,8 +688,33 @@ func visitObjectTypes(t schema.Type, visitor func(*schema.ObjectType), seen code
 }
 
 func (mod *modContext) genType(w io.Writer, obj *schema.ObjectType, input bool, level int) {
+	properties := obj.Properties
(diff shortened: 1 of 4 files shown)


pgavlin:
A couple of alternative approaches I considered:

1. Explicitly split input and output properties on object types. This is a much larger change, and I'm not sure it would be possible to accommodate and differences in shape without breaking changes in the other SDKs.
2. Explicitly split input and output types into separate types in tfgen. I think that this is actually where we want to end up, but again, I think we need to make breaking changes to get here.

Issue #4727, opened 2020-05-29
Support storing pulumi state in terraform http state compatible backends


Place1:
Gitlab 13 has added a cool new feature that provides direct support for _Terraform HTTP State_. This new feature means that terraform users can use Gitlab as their state backend with some nice integration with Gitlab CI and the general project/repo stuff in Gitlab.

https://about.gitlab.com/releases/2020/05/22/gitlab-13-0-released/#gitlab-http-terraform-state-backend

I'm interested to scope out support for the Terraform HTTP State provider in pulumi. It would be pretty cool to run a `pulumi login rest://myserver.com/state` and be up and running with pulumi.

This is probably a useful feature for teams working at organisations that already have their own custom terraform HTTP state backends; they could potentially start using pulumi and integrate with this existing state solution.

The usage docs from terraform are here: https://www.terraform.io/docs/backends/types/http.html

Here's an example http state implementation: https://github.com/mikalstill/junkcode/tree/master/terraform/remote_state

If this is something that pulumi would accept a contribution for i'd be happy to put together a working draft, demo and PR to kick this off.

Place1:
I believe i'll be able to implement this fairly quickly on top of the current filestate backend as another gocloud blob driver.

leezen:
Awesome -- thanks for the update!

ryan4yin:
awesome!

logileifs:
Is there any news on this ?

jay13jay:
is anyone still working on this?  I am trying to make the case to use pulumi on some of my projects but being able to use gitlab as a backend state manager would make my attempts a lot more successful...

alexander-blackwell:
I would also be interested in utilizing Pulumi but would like to see backend state management support in Gitlab first.

hbjydev:
This would be really cool to have, because I really wanna switch from Terraform to Pulumi, but we need our state to be managed by us, so being able to use GitLab’s TF state provider to make this work would be fantastic :)

any updates on it?

mattheews:
I'm Also looking forward to enable this feature  :muscle:

rbcb-bedag:
We are also interested in this feature.
Any updates on this ?

yellowhat:
My workaround is to store the state as a gitlab package:

* `.gitlab-ci.yml`

```yaml
stages:
  - infra

infra:
  stage: build
  image: docker.io/python:3.12.3-slim
  variables:
[...]

PR #4738, opened 2020-05-30
Added support for terraform http state compatible backends

Place1:
This PR addresses #4727

## what this adds

with this PR you can `pulumi login resthttp://example.com` where `http://example.com` is a [terraform http state](https://www.terraform.io/docs/backends/types/http.html) compatible server. `resthttps://` can be used for https.

## background 

The terraform http state backend ([docs](https://www.terraform.io/docs/backends/types/http.html)) allows terraform users to store their statefiles on a remote server using a very bare-bones rest API.

A terraform http state compatible backend supports the following operations:

- retrieve `GET /:key`
- delete `DELETE /:key`
- create/update `POST /:key`
    - the http method is configurable in terraform clients.
    - the default is POST but some users may have backends that require PUT or another http method.
- _some endpoints to implement statefile locking which i've omitted from this PR currently_

The api doesn't offer a "List" operation and there is no support for `:key`s that include slash characters (i.e. `/my/key` is invalid but `/my%2Fkey` is valid).

## approach

To support this type of state backend in pulumi i've implemented a new gocloud blob driver and registered it with pulumi's existing filestate backend.

This approach is nice because we don't need to implement and maintain a new pulumi state driver. The gocloud blob driver implementation is free of pulumi specific logic.

Before I started implementing the blob driver I didn't realize that terraform http state backends don't offer a "List" operation - then I thought that might be a blocker because pulumi uses list operations for commands like `pulumi stack ls`.

To support "List" operations the blob driver maintains an index file that records blob keys that have been created. This is implemented in the `BlobIndex` struct.

## concerns

Pulumi's filestate backend runs many reads/writes during the `up` command. #4258 is a recent issue that is relevant as well. There is a reasonable overhead using HTTP to work with files in this way and I think the filestate backend driver could be refactored to reduce IO operations to blobs.

The above is actually a motivating reason to implement this PR as a blob driver; I hope that it will be able to receive performance updates via improvements to the firestate backend driver itself (benefiting s3, gcs, azblob as well)

Another concern is that pulumi creates lots of files over time. From experience using cloud buckets i've seen them get into the gigabytes from history files and backup files piling up.

Terraform http state backends may be more restrictive than cloud buckets about the number of files that can be created.

Finally the index file that this blob driver maintains to support "List" operations may become a bottleneck if the total number of files is massive as every "create" and "delete" operation will require a write to this index file.

I suspect we can refactor how history and backup files are handled in the filestate backend to address these issues though, perhaps auto pruning the history.

---

Overall im happy with the progress so far and I think this has the potential to be a nice new state backend for pulumi.

### DIFF
--- a/pkg/backend/filestate/restblob/restblob.go
+++ b/pkg/backend/filestate/restblob/restblob.go
@@ -1,0 +1,418 @@
+package restblob
+
+import (
+	"bytes"
+	"context"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/pkg/errors"
+	"gocloud.dev/blob"
+	"gocloud.dev/blob/driver"
+	"gocloud.dev/gcerrors"
+)
+
+func init() {
+	blob.DefaultURLMux().RegisterBucket("resthttp", &RestBlob{})
+	blob.DefaultURLMux().RegisterBucket("resthttps", &RestBlob{})
+}
+
+type RestBlob struct{}
+
+func (r *RestBlob) OpenBucketURL(ctx context.Context, u *url.URL) (*blob.Bucket, error) {
+	u.Scheme = strings.TrimPrefix(u.Scheme, "rest")
+	b := &Bucket{
+		address: u,
+		client:  http.DefaultClient,
+	}
+	b.index = &BucketIndex{
+		address: b.address.String(),
+		client:  b.client,
+	}
+	return blob.NewBucket(b), nil
+}
+
+type Bucket struct {
+	address *url.URL
+	client  *http.Client
+	index   *BucketIndex
+	// TODO: add support for the optional
+	// locking feature in terraform http state backends
+	// lockAddress   string
+	// lockMethod    string
+	// unlockAddress string
+	// unlockMethod  string
+}
+
+type RestClient struct {
+	address string
+	client  *http.Client
+}
+
+func (c *RestClient) NewRestClient(address string) *RestClient {
+	return &RestClient{
+		address: address,
+		client:  http.DefaultClient,
+	}
+}
+
+// bloburl takes a blob key and returns it's blob url
+// on the rest backend.
+func (b *Bucket) bloburl(key string) string {
+	// i've tried a few different key encoding.
+	// i initially wanted something that is still
+	// easily human readable because the key names
+	// may be presented to users in some sort of web ui
+	// that their rest http state backend may provide
+	// but that proved troublesom as some http backends
+	// that i tested, namely gitlab, throw 404 errors when
+	// paths include "." characters.
+	// i've chosen base64 encoding so that the code functions
+	// but it'd be best to find a better compatible key encoding.
+	key = strings.TrimPrefix(key, strings.TrimPrefix(b.address.Path, "/"))
+	key = strings.TrimPrefix(key, "/")
+	key = base64.StdEncoding.EncodeToString([]byte(key))
+
+	// unfortuantly we can't just escape the key to put
+	// in in the rest http URL parameter (path segment)
+	// because some backends like gitlab throw 404
+	// errors if the path component contains a "." dot character.
+	// key = url.PathEscape(key)
+
+	return fmt.Sprintf("%s/%s", b.address, key)
+}
+
+type Meta struct {
+	// if attributes become a requirement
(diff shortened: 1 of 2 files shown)


azkore:
@Place1 I want to try that out, but could you please explain if it would be possible how to use Gitlab token authentication with this PR?

Place1:
@azkore you can use a gitlab token this way: `pulumi login resthttps://<user>:<token>@gitlab.com/api/v4/projects/<YOUR-PROJECT-ID>/terraform/state`

azkore:
@Place1 Great, it seems to work with a new stack, thanks!

Is there currently any way to migrate an existing stack (from AWS S3 for example)?

azkore:
Please disregard the previous question, I was able to do that using  'pulumi stack export' and 'pulumi stack import' after creating a new stack in the gitlab backend. Looks good so far.

Place1:
That’s pretty awesome that you got it working!

mitchellmaler:
I'm guessing state locking will be added in the future after this PR? That is something we have had save us in the past and wouldn't be able to take advantage of this until that is added.

TimoKramer:
Do you think it will be possible to have a custom backend that has a list-api-endpoint instead of an index.js? or will this one be generic?

Place1:
The terraform state backend api spec doesn’t include a list api so probably
not.

But perhaps we should try to make file state backends pluggable like
providers so that other solutions can be tried within the community.

salzig:
Hej, any future for this?

pkg/backend/filestate/restblob/restblob.go
```diff
@@ -0,0 +1,418 @@
...
+	"strings"
+
+	"github.com/pkg/errors"
+	"gocloud.dev/blob"
+	"gocloud.dev/blob/driver"
+	"gocloud.dev/gcerrors"
+)
+
+func init() {
+	blob.DefaultURLMux().RegisterBucket("resthttp", &RestBlob{})
```
salzig, line 22: maybe it would be better of calling it `tfremote`, `tfstore` (with respecting `tfremotes` or `tfstores`), as it's not just any rest backend, but following the way terraform expects remote stores to work?

flashspys:
Hey, this is now stale for to months and no milestones are assigned. Can we expect this to get merged any time soon?

kwiky:
Any update on this ?

Place1:
Funny. I was just searching google to see if any non-saas http based state backend have been supported and I totally forgot I made this PR 3 years ago.

😂

nwhistler:
@Frassle Any chance you could review this? This would be a pretty killer feature.

lunaris:
Thanks for contributing this, and sorry it's been left to go stale. I've spent some time reviewing this and our codebase right now, and I think the concerns and limitations as it stands make it difficult for us to look at merging this. Specifically:

#### List operations

As you mention, it's not clear how the lack of direct list operations will affect ongoing maintenance of this backend. The implementation supplied is neat but potentially would need more design discussion before we proceed. E.g.:

* We may need to take account of additional locks around the index in the event that multiple Pulumi state operations run concurrently.
* The structure of that index is a contract that we (Pulumi) would have to own and manage e.g. the backwards/forward compatibility of, rather than this being the bucket provider's problem as it is for all our other backends. 
* Ideally, we'd not be opening ourselves up to needing to think about solutions such as auto-pruning later on in the event that index size/performance is an issue.

#### Differences in APIs

* As you note, Pulumi is very chatty at this point (we could definitely do a lot better here) and it would be good to understand whether or not we'd be opening users to sub-standard performance because of differences in expectations between the Pulumi client and the TF server backend (frequency or number of writes/calls, for instance).
* I think we'd need to consider whether or not the state locking features of the API are necessary for us to make this work safely and, if they are required aggressively (since our `Bucket` interface doesn't cater for locks explicitly), whether or not this affects performance significantly.

This is a very neat implementation and in an ideal world, there'd be a way to plug in blob backends to gocloud without us needing to take this into the core codebase (a bit like a dynamic library) so that we could fill this use case without solving all these issues. However, for us to merge this into Pulumi and own it, I think we'd have to address the above to a high level. New backends/blob drivers come with quite a high cost relative to how much they offer, given that we are fortunate that gocloud covers a large number of use cases up front. Given the age of this and how far it is from completion, I'm going to close this PR. If we want to open a new one, I think we should do so expecting to address the above concerns, with tests and possibly benchmarks to corroborate/help us take on that maintenance burden. Hope this makes sense :pray:

PR #4742, opened 2020-06-01
[codegen/nodejs/sdk] Do not write reserved files.

pgavlin:
If the source file for a resource or function definition would collide
with a reserved filename (e.g. `index.ts`), append an underscore to its
filename (e.g. `index_.ts`).


pgavlin:
Azure failures are unrelated.

PR #4743, opened 2020-06-02
[codegen/hcl2] Fixes for RewriteConversions.

pgavlin:
- Typecheck in all cases where a type may have changed
- Do not perform literal conversions if the type is already correct
- Perform literal conversions before checking to see if a call to
  `__convert` is required. This catches cases such as string literals
  passed where ints are required. Without this change, that form in
  particular generates a bare number literal rather than a number
  literal wrapped in a `__convert`.

### DIFF
--- a/pkg/codegen/hcl2/rewrite_convert.go
+++ b/pkg/codegen/hcl2/rewrite_convert.go
@@ -40,8 +40,13 @@ func sameSchemaTypes(xt, yt model.Type) bool {
 	return true
 }
 
+// rewriteConversions implements the core of RewriteConversions. It returns the rewritten expression and true if the
+// type of the expression may have changed.
 func rewriteConversions(x model.Expression, to model.Type) (model.Expression, bool) {
+	// If rewriting an operand changed its type and the type of the expression depends on the type of that operand, the
+	// expression must be typechecked in order to update its type.
 	var typecheck bool
+
 	switch x := x.(type) {
 	case *model.AnonymousFunctionExpression:
 		x.Body, _ = rewriteConversions(x.Body, to)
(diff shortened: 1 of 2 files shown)


pkg/codegen/hcl2/rewrite_convert.go
```diff
@@ -100,37 +97,52 @@ func RewriteConversions(x model.Expression, to model.Type) model.Expression {
...
 		}
 	case *model.UnaryOpExpression:
-		x.Operand = RewriteConversions(x.Operand, x.OperandType())
+		x.Operand, _ = rewriteConversions(x.Operand, model.InputType(x.OperandType()))
 	}
 
-	// If the expression's type is directly assignable to the destination type, no conversion is necessary.
-	if to.AssignableFrom(x.Type()) && sameSchemaTypes(to, x.Type()) {
-		return x
+	if typecheck {
```
mikhailshilkov, line 118: I guess I still need to learn how type checking works. If an example is broken, and some bogus value is assigned to a property (e.g. a string to a boolean - before we started converting them, or a non-existing value because of the bug in HCL), why do we still generate those examples? Or what are the cases it catches?

pkg/codegen/hcl2/rewrite_convert.go
```diff
@@ -100,37 +97,52 @@ func RewriteConversions(x model.Expression, to model.Type) model.Expression {
 			valueType, diags := to.Traverse(traverser)
 			contract.Ignore(diags)
 
-			item.Key = RewriteConversions(item.Key, model.InputType(model.StringType))
-			item.Value = RewriteConversions(item.Value, model.InputType(valueType.(model.Type)))
+			var valueChanged bool
+			item.Key, _ = rewriteConversions(item.Key, model.InputType(model.StringType))
```
mikhailshilkov, line 101: What is the principle here and elsewhere - when do we discard `typecheck` vs. when we don't?

pkg/codegen/hcl2/rewrite_convert.go
```diff
@@ -40,53 +40,50 @@ func sameSchemaTypes(xt, yt model.Type) bool {
 	return true
 }
 
-func RewriteConversions(x model.Expression, to model.Type) model.Expression {
+func rewriteConversions(x model.Expression, to model.Type) (model.Expression, bool) {
```
mikhailshilkov, line 43: Consider adding a comment about what the `bool` means

pkg/codegen/hcl2/rewrite_convert.go
pgavlin: > If an example is broken, and some bogus value is assigned to a property (e.g. a string to a boolean - before we started converting them, or a non-existing value because of the bug in HCL), why do we still generate those examples?

In general, we don't convert examples that don't typecheck: e.g. we would not accept an example that passes a list where a string is expected. That said, let's break down the two situations you mentioned, as the second is an exception to the rule.

In the first case, the values aren't actually bogus: [HCL2 performs automatic conversion of these values at runtime](https://github.com/hashicorp/hcl/blob/hcl2/spec.md#type-conversions-and-unification). The rewrite you added matches that behavior; we'll eventually need similar codegen for dynamic conversions of non-literal values.

In the second case, we tolerate missing variables because in some cases they are intended: parts of an example that are not relevant to the resource at hand are often elided. In these cases, we treat the missing variables as having the type `dynamic`, which allows them to pass typechecking.

pkg/codegen/hcl2/rewrite_convert.go
pgavlin: We discard `typecheck` when the rewrite can't affect the type of the containing expression. I'll add a note to that effect.

Issue #4741, opened 2020-06-01
Pulumi terminates EC2 instance if security groups are added or removed


hseagle2015:
I'm using Pulumi 2.3.0 and Pulumi Python SDK to deploy an EC2 instance with one security group attached. If I attach additional security group to that same instance, Pulumi terminates previously create EC2 instance and deploys a new instance with 2 security groups.

Is it possible to avoid EC2 instance termination and instruct Pulumi to simply attach additional security group to existing instance?

If I enable EC2 instance termination protection (by passing `disable_api_termination=True` parameter to aws.ec2.Instance() function), then I get the following error from Pulumi:

```
(venv) user@host:~/pulumi-workspace$ pulumi up
Previewing update (myproject-p1-dev):
     Type                 Name                     Plan       Info
     pulumi:pulumi:Stack  myproject-p1-myproject-p1-dev
 ~   └─ aws:ec2:Instance  myproject-ec2-test          update     [diff: ~securityGroups]

Outputs:
  ~ Public IP      : "XXX.XXX.XXX.XXX" => output<string>
  ~ Public hostname: "ec2-XXX-XXX-XXX-XXX.eu-north-1.compute.amazonaws.com" => output<string>

Resources:
    ~ 1 to update
    3 unchanged

Do you want to perform this update? yes
Updating (myproject-p1-dev):
     Type                 Name                     Status                  Info
     pulumi:pulumi:Stack  myproject-p1-myproject-p1-dev  **failed**              1 error; 22 messages
 ~   └─ aws:ec2:Instance  myproject-ec2-test          **updating failed**     [diff: ~securityGroups]; 1 error

Diagnostics:
  pulumi:pulumi:Stack (myproject-p1-myproject-p1-dev):
    panic: fatal: An assertion has failed: Expected diff to not require deletion or replacement during Update of urn:pulumi:myproject-p1-dev::myproject-p1::aws:ec2/instance:Instance::myproject-ec2-test
    goroutine 28 [running]:
    github.com/pulumi/pulumi/sdk/v2/go/common/util/contract.failfast(...)
        /home/travis/gopath/pkg/mod/github.com/pulumi/pulumi/sdk/v2@v2.1.1-0.20200508232528-aa313aecf8a0/go/common/util/contract/failfast.go:23
    github.com/pulumi/pulumi/sdk/v2/go/common/util/contract.Assertf(0xc000be2000, 0x5ddd59b, 0x48, 0xc000ee1728, 0x1, 0x1)
[...]

PR #4744, opened 2020-06-02
Ensuring a property level deprecation warning is set for python codegen

stack72:
Related: #4741

### DIFF
--- a/pkg/codegen/python/gen.go
+++ b/pkg/codegen/python/gen.go
@@ -294,6 +294,12 @@ func (mod *modContext) genAwaitableType(w io.Writer, obj *schema.ObjectType) str
 		fmt.Fprintf(w, "        if %s and not isinstance(%s, %s):\n", pname, pname, ptype)
 		fmt.Fprintf(w, "            raise TypeError(\"Expected argument '%s' to be a %s\")\n", pname, ptype)
 
+		if prop.DeprecationMessage != "" {
+			fmt.Fprintf(w, "        if %s is not None:\n", pname)
+			fmt.Fprintf(w, "            warnings.warn(\"%s\", DeprecationWarning)\n", prop.DeprecationMessage)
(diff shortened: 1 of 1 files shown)


stack72:
@pgavlin this is what I am interested in you looking at - this is property level warnings for deprecations in Pulumi

This is the only way we get deprecation annotations in Python :/

pkg/codegen/python/gen.go
```diff
@@ -294,6 +294,12 @@ func (mod *modContext) genAwaitableType(w io.Writer, obj *schema.ObjectType) str
 		fmt.Fprintf(w, "        if %s and not isinstance(%s, %s):\n", pname, pname, ptype)
 		fmt.Fprintf(w, "            raise TypeError(\"Expected argument '%s' to be a %s\")\n", pname, ptype)
 
+		if prop.DeprecationMessage != "" {
+			fmt.Fprintf(w, "        if %s is not None:\n", pname)
+			fmt.Fprintf(w, "            warnings.warn(\"%s\", DeprecationWarning)\n", prop.DeprecationMessage)
+			fmt.Fprintf(w, "            pulumi.log.warn(\"%s is deprecated: %s\")\n", pname, prop.DeprecationMessage)
```
lukehoban, line 300: Curious - what is the UX of each of these?  And specifically, why do we need both?

stack72: the pulumi will add to the diagnostics if the user hasn't setup a real python environment (e.g. I write Pulumi code, but I don't use auto-complete of pep etc when doing so so I don't get any Pythonisms in my IDE)

stack72: Happy to remove what you feel is unnecessary though :)

lukehoban: As long as they won't both show up together - this sounds fine.

PR #4746, opened 2020-06-02
[codegen/schema] Support default union types.

pgavlin:
Add support for union types that indicate a default type for targets
that do not support unions, or do not support unions in certain
positions (e.g. output properties). The NodeJS backend makes use of this
in combination with a new flag, `disableUnionOutputTypes`, to avoid
generating unions in output types.

These changes also refactor the various module ->
package/module/namespace mapping methods so that these entities can be
fetched by their language name rather than their token.

### DIFF
--- a/pkg/codegen/nodejs/importer.go
+++ b/pkg/codegen/nodejs/importer.go
@@ -45,6 +45,8 @@ type NodePackageInfo struct {
 	ModuleToPackage map[string]string `json:"moduleToPackage,omitempty"`
 	// Toggle compatibility mode for a specified target.
 	Compatibility string `json:"compatibility,omitempty"`
+	// Disable support for unions in output types.
+	DisableUnionOutputTypes bool `json:"disableUnionOutputTypes,omitempty"`
 }
 
 // NodeObjectInfo contains NodeJS-specific information for an object.
--- a/.github/workflows/codegen-test.yml
+++ b/.github/workflows/codegen-test.yml
@@ -170,3 +170,36 @@ jobs:
           use-provider-dir: true
           pulumi-bot-token: ${{ secrets.PULUMI_BOT_TOKEN }}
           github-actions-token: ${{ secrets.GITHUB_TOKEN }}
+
+  downstream-kubernetes:
+    name: Test Kubernetes Downstream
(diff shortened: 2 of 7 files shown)


pgavlin:
With these changes + the latest commit to https://github.com/pulumi/pulumi-terraform-bridge/pull/194, the last semantic diff I can find in `pulumi-aws` is gone.

pkg/codegen/nodejs/importer.go
```diff
@@ -45,6 +45,8 @@ type NodePackageInfo struct {
 	ModuleToPackage map[string]string `json:"moduleToPackage,omitempty"`
 	// Toggle compatibility mode for a specified target.
 	Compatibility string `json:"compatibility,omitempty"`
+	// Disable support for unions in output types.
+	DisableUnionOutputTypes bool `json:"disableUnionOutputTypes,omitempty"`
```
lukehoban, line 49: Curious - Why does this need to be opt-in?  What are the cases where we *do* want to emit union outputs?  And perhaps knowing the answer to that will make clearer why this is a global setting vs. a per-property setting?

pgavlin: `@pulumi/kubernetes` supports unions in output types.

In an ideal world, we wouldn't need this option: instead, the schema would contain separate types for the input and output APIs. Unfortunately, I think that option is too far off at the moment, so I opted for a more pragmatic approach.

lukehoban: > @pulumi/kubernetes supports unions in output types.

Again mostly curious - do you have an example of this?  I'm curious if it is common and usable - it seems in general it would be quite hard to consume these values if they truly could come back one of several unrelated types.  

I think it is quite possible that the code-generator emits unions, but that the actual values that come back are guaranteed to be one of the kinds - in which case I think we would generally be "safe" making a breaking change.

This doesn't need to get in the way of the changes here though.

pgavlin: One example is `targetPort` in the `ServicePort` interface: https://github.com/pulumi/pulumi-kubernetes/blob/master/sdk/nodejs/types/output.ts#L15433-L15441

PR #4749, opened 2020-06-03
[Go Program Gen] Improved type handling and invokes

EvanBoyle:
~~Should be reviewable commit by commit although there is some minor refactoring that occurs along the way.~~

This review mostly focusing on improving type declarations with more complete coverage of object and tuple cons expressions, as well as propagation of the "inputyness" of values so that we can wrap literals in pulumi types where appropriate (`pulumi.String("foo")`). 

In addition I added invoke support to get the fargate example 99% there. There are a few outstanding issues:

1. Resolving between `Lookup` and `Get` functions for invokes. Would be great to hear some suggestions here.
2. ~~Need to add a lowering step for JSON serialization.~~ __Done.__
3. ~~Must extend the input type propagation logic to cover traversal expressions~~ (`VpcId: pulumi.String(vpc.Id)`)__Done.__
4. write a lowering step that detects unused variable/resource declarations and replaces the names with `_`. This may be optional, but without it the code doesn't compile. Curious to hear thoughts on this. 
5. Lowering step for places where we're using plain array and object types that translates them to appropriate `pulumi.Map` and `pulumi.Array` expressions. 
6. Lowering step to replace optional T literals with ptr<T>.

__Update__: 

~~implementing toJSON caused me to run into this bug:~~

https://github.com/pulumi/pulumi/pull/4749/files#diff-ae83e82bc7feedf66a792c8a49a12105R133-R143

~~Notice that for this resource instantiation, all of the values of `TaskDefinitionArgs` are plain types (string) instead of pulumi types (pulumi.String). Turns out that types are singletons and are cached, so every instance of `model.StringType` points to the same instance. I did not realize this upon starting off on this approach. The conclusion is that annotations are shared across all types, so stripping off annotations for the toJSON argument also strips them off for anything else sharing a type.~~

~~Given this I need to rework the approach and will be going with what @pgavlin suggested which is creating an intrinsic Input that wraps literal values such that~~ `__input(foo)` ~~will generate the appropriate~~ `pulumi.String("foo")`. 

__Update__:

I implemented pulumi type propagation in using a new `__input` intrinsic instead of annotations. This was fairly smooth and resolved a few bugs/outstanding issues such as wrapping scope traversal expressions.


EvanBoyle:
@pgavlin @mikhailshilkov this is ready for a look.

pkg/codegen/go/gen_program_expressions.go
```diff
@@ -270,21 +346,51 @@ func (g *generator) GenRelativeTraversalExpression(w io.Writer, expr *model.Rela
...
+		isInput = false
 		if schemaType, ok := hcl2.GetSchemaForType(resource.InputType); ok {
 			objType, _ = schemaType.(*schema.ObjectType)
+			// convert .id into .ID()
+			last := expr.Traversal[len(expr.Traversal)-1]
+			switch last := last.(type) {
+			case hcl.TraverseAttr:
+				if last.Name == "id" {
+					genIDCall = true
+					expr.Traversal = expr.Traversal[:len(expr.Traversal)-1]
```
pgavlin, line 373:
```suggestion
			if attr, ok := last.(hcl.TraverseAttr); ok && attr.Name == "id" {
					genIDCall = true
					expr.Traversal = expr.Traversal[:len(expr.Traversal)-1]
```

pkg/codegen/go/gen_program_expressions.go
```diff
@@ -312,9 +418,19 @@ func (g *generator) GenTupleConsExpression(w io.Writer, expr *model.TupleConsExp
...
 // GenTupleConsExpression generates code for a TupleConsExpression.
 func (g *generator) genTupleConsExpression(w io.Writer, expr *model.TupleConsExpression, destType model.Type) {
-	// TODO derive from ambient context
-	isInput := true
-	argType := g.argumentTypeName(expr, destType, isInput)
+	isInput := isInputty(destType) || containsInputs(expr)
+	argType := argumentTypeName(expr, destType, isInput)
+	if strings.HasSuffix(argType, "Array") {
+		isInput = true
+	}
```
pgavlin, line 425: Shouldn't this be covered by `isInput`?

pkg/codegen/go/gen_program_expressions.go
```diff
@@ -375,20 +496,84 @@ func (g *generator) argumentTypeName(expr model.Expression, destType model.Type,
...
+		return "map[string]interface{}"
+	case *model.MapType:
+		valType := argumentTypeName(nil, destType.ElementType, isInput)
+		return fmt.Sprintf("map[string]%s", valType)
+	case *model.ListType:
+		argTypeName := argumentTypeName(nil, destType.ElementType, isInput)
+		if strings.HasPrefix(argTypeName, "pulumi.") {
+			return fmt.Sprintf("%sArray", argTypeName)
+		}
+		return fmt.Sprintf("[]%s", argTypeName)
```
pgavlin, line 535: I think these will need to handle input-typed elements and generate inputty names (e.g. `FooMap`, `FooArray`, etc.). Not something to fix now, but maybe file an issue for follow-up.

pkg/codegen/go/gen_program_expressions.go
```diff
@@ -375,20 +496,84 @@ func (g *generator) argumentTypeName(expr model.Expression, destType model.Type,
...
+	case *model.TupleType:
+		// attempt to collapse tuple types
```
pgavlin, line 537: You might consider adding a note that this intentionally does not use `model.UnifyTypes`, as the Go semantics require that all element types match exactly in order to generate correct code.

pkg/codegen/go/gen_program_expressions.go
```diff
@@ -375,20 +496,84 @@ func (g *generator) argumentTypeName(expr model.Expression, destType model.Type,
...
+		}
+		return "[]interface{}"
+	case *model.OutputType:
+		isInput = true
+		return argumentTypeName(expr, destType.ElementType, isInput)
+	case *model.UnionType:
+		for _, ut := range destType.ElementTypes {
+			switch ut.(type) {
+			case *model.OpaqueType:
+				return argumentTypeName(expr, ut, isInput)
```
pgavlin, line 569:
```suggestion
			if _, isOpaqueType := ut.(*model.OpaqueType); isOpaqueType {
				return argumentTypeName(expr, ut, isInput)
```

pgavlin: Also, what case is this intended to handle?

pkg/codegen/go/gen_program_expressions.go
```diff
@@ -432,11 +617,24 @@ func (nameInfo) Format(name string) string {
...
 
 // lowerExpression amends the expression with intrinsics for C# generation.
-func (g *generator) lowerExpression(expr model.Expression, typ model.Type) (model.Expression, []*ternaryTemp) {
+func (g *generator) lowerExpression(expr model.Expression, typ model.Type, isInput bool) (
+	model.Expression, []interface{}) {
 	expr, diags := hcl2.RewriteApplies(expr, nameInfo(0), false /*TODO*/)
 	expr = hcl2.RewriteConversions(expr, typ)
-	expr, temps, ternDiags := g.rewriteTernaries(expr)
+	expr, tTemps, ternDiags := g.rewriteTernaries(expr, g.ternaryTempSpiller)
+	expr, jTemps, jsonDiags := g.rewriteToJSON(expr, g.jsonTempSpiller)
```
pgavlin, line 625: These should be run prior to the apply rewriter, otherwise there is a risk that the expressions that are being lifted will be lifted out of an `apply` callback.

pkg/codegen/go/gen_program_ternaries.go
```diff
@@ -53,8 +55,11 @@ func (ta *tempSpiller) spillExpression(x model.Expression) (model.Expression, hc
 	}, nil
 }
 
-func (g *generator) rewriteTernaries(x model.Expression) (model.Expression, []*ternaryTemp, hcl.Diagnostics) {
-	spiller := &tempSpiller{}
+func (g *generator) rewriteTernaries(
+	x model.Expression,
+	spiller *tempSpiller,
+) (model.Expression, []*ternaryTemp, hcl.Diagnostics) {
+	spiller.temps = nil
```
pgavlin, line 62: Nice catch!

pkg/codegen/go/gen_program_test.go
```diff
@@ -112,7 +116,9 @@ func newTestGenerator(t *testing.T, testFile string) *generator {
 		}
 
 		g := &generator{
-			program: program,
+			program:            program,
+			jsonTempSpiller:    &jsonSpiller{count: 0},
+			ternaryTempSpiller: &tempSpiller{count: 0},
```
pgavlin, line 121: Nit: `count` does not need to be zero-initialized; it will automatically be set to the zero value.

pkg/codegen/go/gen_program.go
```diff
@@ -55,23 +60,31 @@ func GenerateProgram(program *hcl2.Program) (map[string][]byte, hcl.Diagnostics,
...
-	imports := g.collectImports(w, program)
+	g.Fprint(w, "package main\n\n")
 	g.Fprintf(w, "import (\n")
+
+	stdImports, pulumiImports := g.collectImports(w, program)
+	for _, imp := range stdImports.SortedValues() {
+		g.Fprintf(w, "\"%s\"\n", imp)
+	}
+
+	g.Fprintf(w, "\n")
```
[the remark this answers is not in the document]
EvanBoyle: gofmt takes care of this.

pkg/codegen/go/gen_program.go
```diff
@@ -154,21 +178,53 @@ func (g *generator) genResource(w io.Writer, r *hcl2.Resource) {
...
+			g.Fgenf(w, "var %s %s\n", t.Name, argumentTypeName(t.Value.TrueResult, t.Type(), isInput))
+			g.Fgenf(w, "if %.v {\n", t.Value.Condition)
+			g.Fgenf(w, "%s = %.v\n", t.Name, t.Value.TrueResult)
+			g.Fgenf(w, "} else {\n")
+			g.Fgenf(w, "%s = %.v\n", t.Name, t.Value.FalseResult)
+			g.Fgenf(w, "}\n")
+		case *jsonTemp:
+			bytesVar := fmt.Sprintf("tmp%s", strings.ToUpper(t.Name))
+			g.Fgenf(w, "%s, err := json.Marshal(", bytesVar)
+			args := stripInputs(t.Value.Args[0])
```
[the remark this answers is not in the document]
EvanBoyle: Yup, will use the EKS example to work on this but here's an issue: https://github.com/pulumi/pulumi/issues/4760

pkg/codegen/go/gen_program.go
```diff
@@ -154,21 +178,53 @@ func (g *generator) genResource(w io.Writer, r *hcl2.Resource) {
...
+	g.genTemps(w, temps)
+	switch expr := expr.(type) {
+	case *model.FunctionCallExpression:
+		switch expr.Name {
+		case hcl2.Invoke:
+			g.Fgenf(w, "%s, err := %.3v;\n", v.Name(), expr)
+			g.Fgenf(w, "if err != nil {\n")
+			g.Fgenf(w, "return err\n")
+			g.Fgenf(w, "}\n")
+		}
```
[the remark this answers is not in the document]
EvanBoyle: 👍  https://github.com/pulumi/pulumi/issues/4761

pkg/codegen/go/gen_program_expressions.go
```diff
@@ -127,12 +125,27 @@ func (g *generator) GenForExpression(w io.Writer, expr *model.ForExpression) { /
...
 		switch arg := expr.Args[0].(type) {
 		case *model.TupleConsExpression:
 			g.genTupleConsExpression(w, arg, expr.Type())
 		case *model.ObjectConsExpression:
 			g.genObjectConsExpression(w, arg, expr.Type())
+		case *model.LiteralValueExpression:
+			g.genLiteralValueExpression(w, arg, expr.Type())
+		case *model.FunctionCallExpression:
+			name := arg.Name
+			fmt.Println(name)
```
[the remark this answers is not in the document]
EvanBoyle: Yup, removed.

pkg/codegen/go/gen_program_expressions.go
```diff
@@ -231,36 +288,55 @@ func (g *generator) GenObjectConsExpression(w io.Writer, expr *model.ObjectConsE
...
+		var temps []interface{}
+		isInput := isInputty(destType)
+		typeName := argumentTypeName(expr, destType, isInput)
+		if strings.HasSuffix(typeName, "Args") {
+			isInput = true
+		}
+		// invokes are not inputty
+		if strings.Contains(typeName, ".Lookup") {
+			isInput = false
+		}
```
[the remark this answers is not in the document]
EvanBoyle: `destType` is a `Union<T, Output<T>> in this case so `isInputty` returns true. I suppose that some of this could be encapsulated or made more robust, but this is a simple workaround for the time being.

pkg/codegen/go/gen_program_expressions.go
```diff
@@ -231,36 +288,55 @@ func (g *generator) GenObjectConsExpression(w io.Writer, expr *model.ObjectConsE
...
+		isMap := strings.HasPrefix(typeName, "map[")
```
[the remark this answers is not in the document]
EvanBoyle: https://github.com/pulumi/pulumi/issues/4762

pkg/codegen/go/gen_program_expressions.go
```diff
@@ -231,36 +288,55 @@ func (g *generator) GenObjectConsExpression(w io.Writer, expr *model.ObjectConsE
...
+
+		// TODO: retrieve schema and propagate optionals to emit bool ptr, etc.
 
 		// first lower all inner expressions and emit temps
 		for i, item := range expr.Items {
-			// TODO keys are also expressions
-			x, xTemps := g.lowerExpression(item.Value, item.Value.Type())
+			// don't treat keys as inputs
```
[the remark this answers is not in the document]
EvanBoyle: Yup. There may also be MapOutput related scenarios that require it as well, this is noted in that issue.

pkg/codegen/go/gen_program_expressions.go
EvanBoyle: Yup, this is leftover from the earliest implementation.

pkg/codegen/go/gen_program_expressions.go
EvanBoyle: https://github.com/pulumi/pulumi/issues/4763

pkg/codegen/go/gen_program_expressions.go
EvanBoyle: `__convert`. The diff if removed:

```
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -31,4 +31,4 @@
            	            	 					Protocol: pulumi.String("-1"),
            	            	-					FromPort: pulumi.Int(0),
            	            	-					ToPort:   pulumi.Int(0),
            	            	+					FromPort: interface{}(0),
            	            	+					ToPort:   interface{}(0),
            	            	 					CidrBlocks: pulumi.StringArray{
            	            	@@ -41,4 +41,4 @@
            	            	 					Protocol: pulumi.String("tcp"),
            	            	-					FromPort: pulumi.Int(80),
            	            	-					ToPort:   pulumi.Int(80),
            	            	+					FromPort: interface{}(80),
            	            	+					ToPort:   interface{}(80),
            	            	 					CidrBlocks: pulumi.StringArray{
            	            	@@ -96,3 +96,3 @@
            	            	 		webTargetGroup, err := elasticloadbalancingv2.NewTargetGroup(ctx, "webTargetGroup", &elasticloadbalancingv2.TargetGroupArgs{
            	            	-			Port:       pulumi.Int(80),
            	            	+			Port:       interface{}(80),
            	            	 			Protocol:   pulumi.String("HTTP"),
            	            	@@ -106,3 +106,3 @@
            	            	 			LoadBalancerArn: webLoadBalancer.Arn,
            	            	-			Port:            pulumi.Int(80),
            	            	+			Port:            interface{}(80),
            	            	 			DefaultActions: elasticloadbalancingv2.ListenerDefaultActionArray{
            	            	@@ -150,3 +150,3 @@
            	            	 			Cluster:        cluster.Arn,
            	            	-			DesiredCount:   pulumi.Int(5),
            	            	+			DesiredCount:   interface{}(5),
            	            	 			LaunchType:     pulumi.String("FARGATE"),
            	            	@@ -164,3 +164,3 @@
            	            	 					ContainerName:  pulumi.String("my-app"),
            	            	-					ContainerPort:  pulumi.Int(80),
            	            	+					ContainerPort:  interface{}(80),
            	            	 				},
```

pkg/codegen/go/gen_program_ternaries.go
EvanBoyle: I didn't fix this for spilling in python. Not sure it would cause issues though as reusing variables should be fine in the linear flow of the program.

pkg/codegen/go/gen_program_expressions.go
```diff
@@ -206,21 +221,63 @@ func (g *generator) GenFunctionCallExpression(w io.Writer, expr *model.FunctionC
...
+				}
+
+			} else {
+				f, _ := bf.Float64()
+				if isPulumiType {
+					g.Fgenf(w, "%s(%g)", argTypeName, f)
+				} else {
+					g.Fgenf(w, "%g", f)
+				}
+			}
```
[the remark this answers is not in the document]
EvanBoyle: https://github.com/pulumi/pulumi/issues/4764

pkg/codegen/go/gen_program_ternaries.go
pgavlin: > as reusing variables should be fine in the linear flow of the program.

Yep, that's why we don't do the same thing in Python.

pkg/codegen/go/gen_program_expressions.go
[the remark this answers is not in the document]
pgavlin: Okay. I think we ought to investigate this a bit more so we can understand where that `destType` is coming from, but that can be done as follow-up.

EvanBoyle:
@pgavlin any other changes requested or is this ready to merge? Filed a slew of issues, and surely there will be more that need to be addressed as I work on the remaining examples, we get bug reports, etc.

pkg/codegen/go/gen_program_expressions.go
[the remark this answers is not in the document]
EvanBoyle: https://github.com/pulumi/pulumi/issues/4767

pkg/codegen/go/gen_program_expressions.go
pgavlin: Ah, okay. This happens because `RewriteConversions` doesn't pick the best dest type out of the available types in a union. Should this just recurse in all cases (not just on `OpaqueType`)?

pkg/codegen/go/gen_program_expressions.go
pgavlin: Eventually, the value expressions for these temps should be passed through `lowerExpression` as well (this will ensure that they pick up the `apply` transform et. al.). No need to fix now, though.

pkg/codegen/go/gen_program_expressions.go
EvanBoyle: Probably, but I'd prefer to wait until I have an example or test case to develop against though. I'm sure I'll end up hitting this when I go to match one of the three remaining test cases or via a bug report.

PR #4753, opened 2020-06-04
[codegen/nodejs/sdk] Fix token type config getter.

pgavlin:
If the type of a configuration variable is a token type (i.e. a type
provided by an overlay) with the underlying type string, call
`config.get` rather than `config.getObject`. This fixes a regression in
the `aws.config.region` variable.


pgavlin:
This is necessary in order to successfully roll out the latest `pulumi-terraform-bridge` to `pulumi-aws`.

PR #4759, opened 2020-06-05
[codegen/python] Fix readme generation

komalali:
Readmes were not being generated for the top-level module but for the submodules instead. This PR corrects that.

### DIFF
--- a/pkg/codegen/python/gen.go
+++ b/pkg/codegen/python/gen.go
@@ -156,9 +156,21 @@ func (mod *modContext) gen(fs fs) error {
 		fs.add(filepath.Join(dir, "utilities.py"), buffer.Bytes())
 
 		// Ensure that the top-level (provider) module directory contains a README.md file.
-		readme := mod.pkg.Description
-		if readme != "" && readme[len(readme)-1] != '\n' {
-			readme += "\n"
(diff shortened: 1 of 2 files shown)


pgavlin:
I don't think that this is the diff we want: https://github.com/pulumi-bot/pulumi-azuread/compare/a56fcd1..dd4e90a#diff-df92345e6e4f46638fbb1aa9e504b646

In this case, the `README.md` is hand-written. We handle this in the NodeJS code generator by allowing consumers to specify a `README` in the schema rather than using the one we auto-generate: https://github.com/pulumi/pulumi/blob/master/pkg/codegen/nodejs/importer.go#L32-L33

I think we should take a similar approach here.

komalali:
@pgavlin It looks like it's still not picking up the correct README -> https://github.com/pulumi-bot/pulumi-azuread/compare/f30fae4..412121b

Any ideas on what I'm missing here? Seems like maybe the README isn't defined in the azuread schema.json?

pgavlin:
> Any ideas on what I'm missing here? Seems like maybe the README isn't defined in the azuread schema.json?

That's right. I think the code is correct, but `pulumi-terraform-bridge` will also need some fixes to add the `README` to the schema.

PR #4765, opened 2020-06-06
[cli] Add an import command.

### LINKED ISSUE #1635: Add an import command
We have `get` functions that let you read in the existing state of a resource, by its ID, without adding it to your checkpoint file. And we have a `refresh` command that will take existing resources in your checkpoint, and synchronize your checkpoint's state with the live values of the resources.

What we are missing is a way to import existing resources into your checkpoint, essentially "adopting" them for use within Pulumi.  For large existing cloud applications and infrastructure, this capability will be a requirement.

As described, the above idea is actually trivial to achieve.  Import some IDs into the checkpoint, and then refresh their state.  This is probably a fine place to start just for simplicity (it's a little better than just editing the checkpoint by hand to achieve the same thing).

One could imagine something like

```
$ pulumi import aws:elasticloadbalancingv2:Listener mylb-42dc3ff4
```

and having it result in an imported resource in your checkpoint file, with the live resource state.  Presumably running `pulumi update` after this would think the resource had been deleted from your program, until you manually added the source code.

The problem gets complicated by two matters

1. How to discover those IDs, particularly when the mapping of physical resources to logical ones in a Pulumi program are not always evident (and not always even 1-to-1)

2. How to generate the source code for the program that would have created that checkpoint state

Those feel like longer term problems that we'll need to tackle, perhaps as follow-on activities.

I should note that all these issues share a lot in common with our ongoing design discussion of conversion tools.

pgavlin:
These changes add a new top-level command, `import`, to the Pulumi CLI.

Resources that are not managed by Pulumi can be imported into a Pulumi stack
using this command. A definition for each resource will be printed to stdout
in the language used by the project associated with the stack; these definitions
should be added to the Pulumi program. The resources are protected from deletion
by default.

A single resource may be specified in the command line arguments or a set of
resources may be specified by a JSON file. This file must contain an object
of the following form:

    {
        "nameTable": {
            "provider-or-parent-name-0": "provider-or-parent-urn-0",
            ...
            "provider-or-parent-name-n": "provider-or-parent-urn-n",
        },
        "resources": [
            {
                "type": "type-token",
                "name": "name",
                "id": "resource-id",
                "parent": "optional-parent-name",
                "provider": "optional-provider-name",
                "version": "optional-provider-version",
            },
            ...
            {
                ...
            }
        ]
    }

The name table maps language names to parent and provider URNs. These names are
used in the genrated definitions, and should match the corresponding declarations
in the source program. This table is required if any parents or providers are
specified by the resources to import.

The resources list contains the set of resources to import. Each resource is
specified as a triple of its type, name, and ID. The format of the ID is specific
to the resource type. Each resource may specify the name of a parent or provider;
these names must correspond to entries in the name table. If a resource does not
specify a provider, it will be imported using the default provider its type. A
resource that does not specify a provider may specify the version of the provider
that will be used for its import.

Fixes #1635.


pkg/cmd/pulumi/import.go
```diff
@@ -0,0 +1,252 @@
...
+					if pdiff.InputDiff {
+						source = old.Inputs
+					}
+
+					if old, ok := elements.Get(resource.NewObjectProperty(source)); ok {
+						// Ignore failure here.
+						elements.Add(resource.NewObjectProperty(state.Inputs), old)
+					}
+				}
+			}
```
pgavlin, line 221: This is the biggest open question I have.

This code matches the steps a user would need to take to author a resource definition that will successfully import, but it ends up leaving out anything that might have been computed. I _think_ that this is correct (and is probably the only way for this to work successfully, as outputs can be set for properties that would conflict were they inputs), but it can result in surprisingly small resource definitions.

pkg/cmd/pulumi/import.go
```diff
@@ -0,0 +1,252 @@
...
+			if read.Outputs == nil {
+				return errors.Errorf("resource '%v' does not exist", id)
+			}
+			if read.Inputs == nil {
+				return errors.Errorf(
+					"provider does not support importing resources; please try updating the '%v' plugin",
+					urn.Type().Package())
+			}
```
pgavlin, line 183: This is pulled from the implementation of `deploy.ImportStep`.

pkg/cmd/pulumi/import.go
```diff
@@ -0,0 +1,252 @@
...
+			ctx, err := plugin.NewContext(sink, sink, nil, target, cwd, nil, nil)
+			if err != nil {
+				return err
+			}
+			host := ctx.Host
+
+			// Load the default provider.
+			//
+			// TODO: allow a provider to be specified on the command line or pull from the snapshot, as we can't
+			// otherwise store to the state.
```
pgavlin, line 146: This is important even without the bit about storing to the state. We really should generate the definition using the schema for the provider version that is present in the state file, if any.

pkg/codegen/hcl2/functions.go
```diff
@@ -196,6 +196,21 @@ var pulumiBuiltins = map[string]*model.Function{
...
+			}
+
+			return model.StaticFunctionSignature{
+				Parameters: []model.Parameter{{
+					Name: "value",
+					Type: valueType,
+				}},
+				ReturnType: model.NewOutputType(valueType),
+			}, nil
+		})),
```
pgavlin, line 213: I added this function, but apparently I did not add support for it to the code generators. Should be straightforward.

pkg/x/importer/generate.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	contract.Assert(len(parser.Diagnostics) == 0)
+
+	program, diags, err := hcl2.BindProgram(parser.Files, hcl2.PluginHost(host))
+	if err != nil {
+		return err
+	}
+	if len(diags) != 0 {
+		// TODO: do better here
+		return diags[0]
```
pgavlin, line 40: This should be treated as an internal error, and ideally should capture the diagnostics.

pkg/cmd/pulumi/import.go
lukehoban: >  it ends up leaving out anything that might have been computed

I'm not sure if this is the same point - but it seems to also leave out anything that is "optional". 

I just tried for example:

```
➜  tsimporttest PULUMI_EXPERIMENTAL=true /opt/pulumi/bin/pulumi import aws:cloudformation/stack:Stack s example-oidc-iam-sa-13d5faa9               


const s = new aws.cloudformation.Stack("s", {
    name: "example-oidc-iam-sa-13d5faa9",
    tags: {
        Name: "example-oidc-iam-sa-nodes",
    },
});
```

That is enough to import the resource, but is definitely not enough to re-create the resource.  Indeed - trying to create a resources from that fails with:

```
  aws:cloudformation:Stack (s):
    error: Creating CloudFormation stack failed: ValidationError: Either Template URL or Template Body must be specified.
        status code: 400, request id: 503c81d7-ee0e-45d4-adee-edbcf63b9482
```

The `outputs` section in the Pulumi stack does contain the desired `templateBody` property, but it wasn't brought over to the inputs.

pkg/x/importer/generate.go
```diff
@@ -0,0 +1,170 @@
...
+			}
+
+			x, err := generateValue(obj[k])
+			if err != nil {
+				return nil, err
+			}
+			items = append(items, model.ObjectConsItem{
+				Key: &model.LiteralValueExpression{
+					Value: cty.StringVal(string(k)),
+				},
```
lukehoban, line 134: Not sure if it's related to this code or somewhere else - but I hit this a couple times in testing:

```
➜  tsimporttest PULUMI_EXPERIMENTAL=true /opt/pulumi/bin/pulumi import aws:cloudformation/stack:Stack s k8s-aws-cluster-ng-standard-pulumi-1bb83a8b


error: s.pp:11,9-12: undefined variable k8s; 
```

In all cases where a `Tag` has a `.` in the name of one of its keys.

pkg/cmd/pulumi/import.go
pgavlin: The problem here is that `template_body` is optional or computed. It does come back in the inputs, but because it's a computed property, the provider sees no diff if the property is not present in the inputs.

I should point out that the generated definitions are all incomplete because they do not include the `import` resource option.

```go
			"template_body": {
				Type:         schema.TypeString,
				Optional:     true,
				Computed:     true,
				ValidateFunc: validateCloudFormationTemplate,
				StateFunc: func(v interface{}) string {
					template, _ := normalizeCloudFormationTemplate(v)
					return template
				},
			},
			"template_url": {
				Type:     schema.TypeString,
				Optional: true,
			},
```

pkg/cmd/pulumi/import.go
```diff
@@ -0,0 +1,252 @@
...
+	cmd := &cobra.Command{
+		Use:   "import [type] [name] [id]",
+		Args:  cmdutil.ExactArgs(3),
+		Short: "Import a resource into an existing stack",
+		Long: "Import a resource into an existing stack.\n" +
+			"\n" +
+			"A resource that is not managed by Pulumi can be imported into a Pulumi stack\n" +
+			"using this command. A definition for the resource will be printed to stdout\n" +
+			"in the language used by the project associated with the stack; this definition\n" +
+			"should be added to the Pulumi program. The resource is protected from deletion\n" +
```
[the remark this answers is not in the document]
pgavlin: The motivation for marking these resources as protected by default is to make it harder for users to accidentally delete resources if they run an import but forget to copy/paste the definition into their program. This sort of conservative approach feels safer, though it does make importing a resource _for_ deletion harder.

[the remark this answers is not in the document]
lukehoban: My gut would be that this probably shouldn’t be the default as it’s just more complex and not necessarily desired (for example - I don’t think folks doing bulk imports today are choosing to do this ~ever) - but that perhaps a `—protect` flag should be available to opt-in to this?

pkg/cmd/pulumi/import.go
lukehoban: It might be useful to see what the experience is if we don’t leave off Computed. (As long as they are Optional or Required). 

The current experience of underspecifying feels less “correct” than overspecifying.  Though it would be interesting to hand validate a number of concrete cases to better characterize the trade off.

pgavlin: > It might be useful to see what the experience is if we don’t leave off Computed. (As long as they are Optional or Required).

Anything that is optional or required does come back in the inputs; we just remove them when we process the result of the diff. End-to-end the process looks like this:

1. Read the current inputs and state from the provider
2. Diff an empty set of inputs against the current inputs + state
3. If there are diffs, copy the differing properties into the inputs
4. Generate the resource's definition using the inputs from (3)

When I originally wrote this up, I did not do (2) and (3). Instead, I used the inputs from (1) in (4). This resulted in very large definitions that at least in some cases--probably `aws:ec2:Instance` or `aws:s3:Bucket`--did not successfully `Check` due to the presence of conflicting input properties. For example, this can happen if there are two in/out properties A and B where if A is an input B is computed and vice-versa. The experience of copy-pasting a definition that didn't successfully import (and would not successfully create) is what led me to the current approach.

> The current experience of underspecifying feels less “correct” than overspecifying. Though it would be interesting to hand validate a number of concrete cases to better characterize the trade off.

FWIW, the current experience is consistent with what a user would need to do manually in order for an import to succeed.

pkg/cmd/pulumi/import.go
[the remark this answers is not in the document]
pgavlin: > My gut would be that this probably shouldn’t be the default as it’s just more complex and not necessarily desired (for example - I don’t think folks doing bulk imports today are choosing to do this ~ever) - but that perhaps a —protect flag should be available to opt-in to this?

FWIW, if the workflow runs as intended--if a user copy/pastes the definition(s) into their program and runs an update--the resources would no longer be protected. The E2E flow looks something like:

1. Run the `import` command. Resources are protected in the statefile.
2. Copy/paste the definitions into the program. These definitions do not specify the `protect` resource option.
3. Run an update. Resources are no longer protected. The import is now complete.

Basically, adding these resources as protected simply ensures that they're not unintentionally deleted until the import is complete.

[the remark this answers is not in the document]
lukehoban: > FWIW, if the workflow runs as intended--if a user copy/pastes the definition(s) into their program and runs an update--the resources would no longer be protected.

Ahh - I see.  That doesn't seem to be what's currently implemented - but that makes sense.

> Basically, adding these resources as protected simply ensures that they're not unintentionally deleted until the import is complete.

I like the 3 step model you describe.  I think there's an alternate possibility to consider:
1. Run the `import` command, it shows code and prompts whether you would like to proceed to import into state (with option for `--yes`).
2. User pastes code into their program, then accepts the prompt.
3. If they now run update no diff is shown.

Curious your thoughts on those two options?  It feels like maybe this order of encouraging copying into program before modifying state file is more "natural"?

pkg/cmd/pulumi/import.go
```diff
@@ -0,0 +1,252 @@
+// Copyright 2016-2018, Pulumi Corporation.
```
ringods, line 1: @lukehoban Shouldn't you update your license header?

lukehoban:
Just spent some time trying this out locally.  Still trying to gauge whether the full preview/update UX feels "right" here - it definitely feels a bit surprising - but at the same time I can see why we ended up wanting it.  Perhaps if most of the other issues noted below were addressed the core UX would feel more "right".

A few specific things.

### Failure UX

We seem to still present some output even after a failure - which doesn't seem right. 

```
➜  importtest2 ~/go/bin/pulumi import azure:core:ResourceGroup foo main-rg3adcde3b
Previewing update (dev):
     Type                         Name             Plan       Info
 +   pulumi:pulumi:Stack          importtest2-dev  create     1 error
 =   └─ azure:core:ResourceGroup  foo              import     1 error
 
Diagnostics:
  pulumi:pulumi:Stack (importtest2-dev):
    error: preview failed
 
  azure:core:ResourceGroup (foo):
    error: Preview failed: Error building account: Error getting authenticated object ID: Error parsing json result from the Azure CLI: Error waiting for the Azure CLI: exit status 1
 

using Pulumi;

class MyStack : Stack
{
    public MyStack()
    {
    }

}
```

## Success Path 

There's a lot going on with the preview and update steps in the import UX.  That might be a good thing - though it feels like it make it unclear what is really happening - what is the `Stack` row? why is this locked? what "desired state" am I previewing/updating to here?  Also `(same) 🔓` in particular is odd since the protection is actually getting removed (though that is not made clear in the diff).

And then the output of the code is just appended at the end which feels awkward.  For this one - I'd suggest:
1. Include prose explaining what this is and what to do with it.
2. Perhaps emit it to a tmp file as well, and include a link to that tmp file in the prose (with also a flag to pick the name of the file to write to)?

```
➜  importtest2 ~/go/bin/pulumi import azure:core/resourceGroup:ResourceGroup foo /subscriptions/0282681f-7a9e-424b-80b2-96babd57a8a1/resourceGroups/main-rg3adcde3b
Previewing update (dev):
     Type                         Name             Plan       
 +   pulumi:pulumi:Stack          importtest2-dev  create     
 =   └─ azure:core:ResourceGroup  foo              import     
 
Resources:
    + 1 to create
    = 1 to import
    2 changes

Do you want to perform this update? details
+ pulumi:pulumi:Stack: (create)
    [urn=urn:pulumi:dev::importtest2::pulumi:pulumi:Stack::importtest2-dev]
    = azure:core/resourceGroup:ResourceGroup: (import) 🔒
        [id=/subscriptions/0282681f-7a9e-424b-80b2-96babd57a8a1/resourceGroups/main-rg3adcde3b]
        [urn=urn:pulumi:dev::importtest2::azure:core/resourceGroup:ResourceGroup::foo]
        location  : "westus"
        name      : "main-rg3adcde3b"

Updating (dev):
     Type                         Name             Status       
 +   pulumi:pulumi:Stack          importtest2-dev  created      
 =   └─ azure:core:ResourceGroup  foo              imported     
 
Resources:
    + 1 created
    = 1 imported
    2 changes

Duration: 7s

Permalink: https://app.pulumi.com/lukehoban/importtest2/dev/updates/1
using Pulumi;
using Azure = Pulumi.Azure;

class MyStack : Stack
{
    public MyStack()
    {
        var foo = new Azure.Core.ResourceGroup("foo", new Azure.Core.ResourceGroupArgs
        {
            Location = "WestUS",
            Name = "main-rg3adcde3b",
        });
    }

}
```


## Diff on update immediately following import

The diff is really weird (`~__defaults,protect,provider`), and actually is shown as a `same` despite having differences, and two of the three things shown in the summary aren't shown in the details, and the diff in the provider is really odd (I've never seen anything like that).  Feels like we *really* want there to be no immediate diff after an import.

```
➜  importtest2 pulumi up
Previewing update (dev):
     Type                         Name             Plan     Info
     pulumi:pulumi:Stack          importtest2-dev           
     └─ azure:core:ResourceGroup  foo                       [diff: ~__defaults,protect,provider]
 
Resources:
    2 unchanged

Do you want to perform this update? details
  pulumi:pulumi:Stack: (same)
    [urn=urn:pulumi:dev::importtest2::pulumi:pulumi:Stack::importtest2-dev]
      azure:core/resourceGroup:ResourceGroup: (same) 🔓
        [id=/subscriptions/0282681f-7a9e-424b-80b2-96babd57a8a1/resourceGroups/main-rg3adcde3b]
        [urn=urn:pulumi:dev::importtest2::azure:core/resourceGroup:ResourceGroup::foo]
        [provider: urn:pulumi:dev::importtest2::pulumi:providers:azure::default::c413cb56-cba1-4050-b3fd-791ee29297e9 => urn:pulumi:dev::importtest2::pulumi:providers:azure::default_3_12_1::output<string>]
        location  : "WestUS"
        name      : "main-rg3adcde3b"

Updating (dev):
     Type                         Name             Status     Info
     pulumi:pulumi:Stack          importtest2-dev             
     └─ azure:core:ResourceGroup  foo                         [diff: ~__defaults,protect,provider]
 
Resources:
    2 unchanged

Duration: 9s

Permalink: https://app.pulumi.com/lukehoban/importtest2/dev/updates/2
```

## Importing existing resource

Re-importing an existing resource just shows nothing changed.  Which in theory is "right" - but in practice I just kind of expect a UX more specific to the import action - that warns me the resource already exists in my project?

```
➜  importtest2 ~/go/bin/pulumi import azure:core/resourceGroup:ResourceGroup foo /subscriptions/0282681f-7a9e-424b-80b2-96babd57a8a1/resourceGroups/main-rg3adcde3b
Previewing update (dev):
     Type                 Name             Plan     
     pulumi:pulumi:Stack  importtest2-dev           
 
Resources:
    2 unchanged
```

## .NET codegen boilerplate

It's not clear to me that we should add all this "ceremony" in the import codegen case (honestly - not clear we should have it in resource docs either).  But one additional specifically odd thing is that the `using Azure = Pulumi.Azure;` form used here isn't part of the default .NET templates, so the actual usage code isn't copy/pastable without edits.  Feels like we should probagbly just qualify `Pulumi.Azure....` everywhere for maximum copy/pastability (and let IDE tooling simplify?).  Curious @mikhailshilkov thoughts (most of this applies to .NET codegen more generally).

```csharp
using Pulumi;
using Azure = Pulumi.Azure;

class MyStack : Stack
{
    public MyStack()
    {
        var foo = new Azure.Core.ResourceGroup("foo", new Azure.Core.ResourceGroupArgs
        {
            Location = "WestUS",
            Name = "main-rg3adcde3b",
        });
    }

}
```

## Crash

For this (yes - my `type`s are wrong):

```json
{
    "resources": [
        {
            "type": "type-token",
            "name": "PulumiCleanup",
            "id": "/subscriptions/0282681f-7a9e-424b-80b2-96babd57a8a1/resourceGroups/Azure-Account-Cleanup/providers/Microsoft.Automation/automationAccounts/PulumiCleanup"
        },
        {
            "type": "type-token",
            "name": "PulumiCleanup/ResourceGroupCleanup",
            "id": "/subscriptions/0282681f-7a9e-424b-80b2-96babd57a8a1/resourceGroups/Azure-Account-Cleanup/providers/Microsoft.Automation/automationAccounts/PulumiCleanup/runbooks/ResourceGroupCleanup"
        },
        {
            "type": "type-token",
            "name": "aerewtmp",
            "id": "/subscriptions/0282681f-7a9e-424b-80b2-96babd57a8a1/resourceGroups/Azure-Account-Cleanup/providers/Microsoft.ContainerRegistry/registries/aerewtmp"
        }
    ]
}
```

I get this:

```
➜  importtest2 ~/go/bin/pulumi import --file resources.json                                                                                                                              
Previewing update (dev):
     Type                 Name             Plan        
     pulumi:pulumi:Stack  importtest2-dev  running     
panic: fatal: A precondition has failed for urn

goroutine 88 [running]:
        /Users/lukehoban/go/src/github.com/pulumi/pulumi/sdk/go/common/util/contract/failfast.go:23
github.com/pulumi/pulumi/sdk/v2/go/common/util/contract.Require(0xc00169a600, 0x5750185, 0x3)
        /Users/lukehoban/go/src/github.com/pulumi/pulumi/sdk/go/common/util/contract/require.go:26 +0x13c
github.com/pulumi/pulumi/pkg/v2/resource/deploy/providers.(*Registry).Check(0xc00062b270, 0xc00169a640, 0x37, 0x0, 0xc000bc0990, 0x0, 0x7, 0xc00169a640, 0x37, 0x0, ...)
        /Users/lukehoban/go/src/github.com/pulumi/pulumi/pkg/resource/deploy/providers/registry.go:225 +0x8a
github.com/pulumi/pulumi/pkg/v2/resource/deploy.(*importer).registerProviders(0xc000bb1af8, 0x5c06da0, 0xc00065e000, 0xc000b500a0, 0x41, 0x20300000000100, 0x82fffff)
        /Users/lukehoban/go/src/github.com/pulumi/pulumi/pkg/resource/deploy/import.go:218 +0x4d7
github.com/pulumi/pulumi/pkg/v2/resource/deploy.(*importer).importResources(0xc000bb1af8, 0x5c06da0, 0xc00065e000, 0xc000b30090, 0x5c07360)
        /Users/lukehoban/go/src/github.com/pulumi/pulumi/pkg/resource/deploy/import.go:264 +0x100
github.com/pulumi/pulumi/pkg/v2/resource/deploy.(*planExecutor).importResources(0xc000e96000, 0x5c06e60, 0xc00071eea0, 0x5c07360, 0xc0005e3a40, 0x7fffffff, 0x0, 0x0, 0x0, 0x0, ...)
        /Users/lukehoban/go/src/github.com/pulumi/pulumi/pkg/resource/deploy/plan_executor.go:446 +0x164
github.com/pulumi/pulumi/pkg/v2/resource/deploy.(*planExecutor).Execute(0xc000e96000, 0x5c06e60, 0xc00071eea0, 0x5c07360, 0xc0005e3a40, 0x7fffffff, 0x0, 0x0, 0x0, 0x0, ...)
        /Users/lukehoban/go/src/github.com/pulumi/pulumi/pkg/resource/deploy/plan_executor.go:137 +0x11ed
github.com/pulumi/pulumi/pkg/v2/resource/deploy.(*Plan).Execute(...)
        /Users/lukehoban/go/src/github.com/pulumi/pulumi/pkg/resource/deploy/plan.go:352
github.com/pulumi/pulumi/pkg/v2/engine.(*planResult).Walk.func1(0x5c07360, 0xc0005e3a40, 0xc0005e32c0, 0x5c06e60, 0xc00071eea0, 0xc00071ee01, 0xc000fdfff0, 0xc0000aed80)
        /Users/lukehoban/go/src/github.com/pulumi/pulumi/pkg/engine/plan.go:203 +0x24a
created by github.com/pulumi/pulumi/pkg/v2/engine.(*planResult).Walk
        /Users/lukehoban/go/src/github.com/pulumi/pulumi/pkg/engine/plan.go:189 +0x177
```

## Properties that must be specified

Unclear what I can do in this case?  I continually to be fairly sure we're going to need an option to request that "all" properties be specified as inputs.  I think this will otherwise be very common.

```
➜  importtest2 ~/go/bin/pulumi import --file resources.json
Previewing update (dev):
     Type                                Name             Plan       Info
     pulumi:pulumi:Stack                 importtest2-dev             1 error
 =   ├─ azure:automation:RunBook         cleanup          import     3 errors
 =   ├─ azure:automation:Account         pulumiCleanup    import     
 =   └─ azure:containerservice:Registry  aerewtmp         import     
 
Diagnostics:
  pulumi:pulumi:Stack (importtest2-dev):
    error: preview failed
 
  azure:automation:RunBook (cleanup):
    error: azure:automation/runBook:RunBook resource 'cleanup' has a problem: "content": one of `content,publish_content_link` must be specified
    error: azure:automation/runBook:RunBook resource 'cleanup' has a problem: "publish_content_link": one of `content,publish_content_link` must be specified
    error: Preview failed: one or more inputs failed to validate
 

using Pulumi;

class MyStack : Stack
{
    public MyStack()
    {
    }

}
```

mikhailshilkov:
> .NET codegen boilerplate

Yeah, this feels problematic in this case. I had all the same questions while working on examples codegen. There, I defaulted to 
a. Produce something that compiles (therefore the boilerplate).
b. Keep it shorter (therefore, `Azure = Pulumi.Azure`, which helps a tiny bit.

We could introduce different "modes" for generation and pick different options depending on it. For examples, I'd add `using` statements for every namespace rather than doing `Pulumi.Azure` everywhere, but for import we could do the latter.

pgavlin:
> We seem to still present some output even after a failure - which doesn't seem right.

This was intentional, but the degenerate case that you've hit should be handled specially. We still present output because even if some resources fail to import, we may have definitions for others. However, if no resources imported successfully, we should suppress the output.

> what is the Stack row? why is this locked? what "desired state" am I previewing/updating to here

I can see how these things can be confusing when a user's first update is an import, but I think the concepts are adequately clear for more experienced users. Do you have any ideas on how we could improve the clarity for new users?

> Also (same) 🔓 in particular is odd since the protection is actually getting removed (though that is not made clear in the diff).

The protection is being added as part of the import--it will be removed with the next update.

> The diff is really weird (~__defaults,protect,provider), and actually is shown as a same despite having differences, and two of the three things shown in the summary aren't shown in the details, and the diff in the provider is really odd (I've never seen anything like that). Feels like we really want there to be no immediate diff after an import.

I agree, and that diff is surprising. Have you seen the same behavior with other resource types?

> Unclear what I can do in this case? I continually to be fairly sure we're going to need an option to request that "all" properties be specified as inputs. I think this will otherwise be very common.

This feels like a bug in the importer. The intent is that all required properties are automatically fetched from the current state. Can you provide some details about this repro (e.g. the current state of the `RunBook`)?

lukehoban:
> This feels like a bug in the importer. The intent is that all required properties are automatically fetched from the current state. Can you provide some details about this repro (e.g. the current state of the RunBook)?

I was specifically trying to import the resources in this Azure Resource Group: https://portal.azure.com/#@pulumi.com/resource/subscriptions/0282681f-7a9e-424b-80b2-96babd57a8a1/resourceGroups/Azure-Account-Cleanup/overview.  This was the first multi-resource example I tried.

lukehoban:
> > Also `(same) 🔓 ` in particular is odd since the protection is actually getting removed (though that is not made clear in the diff).

> The protection is being added as part of the import--it will be removed with the next update.

That comment I had was actually in the wrong place - I meant to add it to the following section on what happens for an update following the import.  I've fixed now.

lukehoban:
Trying to use this with Kubernetes resources - it seems to fail on almost any resource due to issues with `labels` and `annotations` (not putting quotes around dictionary key?)

```typescript
const frontend = new kubernetes.apps.v1.Deployment("frontend", {
    apiVersion: "apps/v1",
    kind: "Deployment",
    metadata: {
        annotations: {
            [pulumi.com / autonamed]: "true",
        },
        labels: {
            [app.kubernetes.io / managed_by]: "pulumi",
        },
        name: "frontend-fbm89qck",
    },
    spec: {
        replicas: 3,
        selector: {
            matchLabels: {
                app: "frontend",
            },
        },
        template: {
            metadata: {
                labels: {
                    app: "frontend",
                },
            },
            spec: {
                containers: [{
                    env: [{
                        name: "GET_HOSTS_FROM",
                        value: "dns",
                    }],
                    image: "pulumi/guestbook-php-redis",
                    name: "frontend",
                    ports: [{
                        containerPort: 80,
                    }],
                    resources: {
                        requests: {
                            cpu: "100m",
                            memory: "100Mi",
                        },
                    },
                }],
            },
        },
    },
});
```

My first attempt to fix manually failed due to it needing to be further corrected from what was returned above as `managed_by` to the correct `managed-by`.

If I fix that manually - then I hit:

```typescript
    error: kubernetes:apps/v1:Deployment resource 'frontend' has a problem: invalid use of reserved internal annotation "pulumi.com/autonamed"
```

This will I suppose only be triggered by resources created with Pulumi - but will happen on presumably every resource created via Pulumi.

If I then look at the update diff - I see the `autonamed` annotation being removed (even though it didn't really exist)?  But also I see a diff on the provider which doesn't seem "correct"?

```
Do you want to perform this update? details
  pulumi:pulumi:Stack: (same)
    [urn=urn:pulumi:dev::importtest3-js::pulumi:pulumi:Stack::importtest3-js-dev]
    ~ kubernetes:apps/v1:Deployment: (update) 🔓
        [id=frontend-fbm89qck]
        [urn=urn:pulumi:dev::importtest3-js::kubernetes:apps/v1:Deployment::frontend]
        [provider: urn:pulumi:dev::importtest3-js::pulumi:providers:kubernetes::default::245e5f12-5a0e-4c46-b442-7a05a0d3a78b => urn:pulumi:dev::importtest3-js::pulumi:providers:kubernetes::default_2_4_2::output<string>]
      ~ metadata: {
          - annotations: {
              - pulumi.com/autonamed: "true"
            }
        }
    --outputs:--
```

After that - it works and there are no further diffs on the next update.

lukehoban:
Following on the previous comment - this may just be a fundamental limitation around import - but I tried removing the explicit naming on the resources in code (so that it could be used to deploy a new copy of the infrastructure in a new stack), and I am told it will replace.  Is there any reasonably easy way to use `import` such that it I can deploy additional copies of the stack without replacing the existing resources?

```
 ++kubernetes:apps/v1:Deployment: (create-replacement)
        [id=redis-replica-48v4z9b7]
        [urn=urn:pulumi:dev::importtest3-js::kubernetes:apps/v1:Deployment::redis-replica]
        [provider=urn:pulumi:dev::importtest3-js::pulumi:providers:kubernetes::default_2_4_2::ae7b250e-bd84-437e-ae1c-aeca85b79a4e]
      ~ metadata: {
          - name: "redis-replica-48v4z9b7"
        }
```

lukehoban:
This issue reported earlier still appears to exist - and I expect should be fixed before merging:

This is what I see after a succesful import which should have no diff - on any resource.  It is surprising there is any diff, but then also surprising that two things the user has likely never heard of and certainly didn't consciously interact with in this case, are changing.  Also that there is no indication of whether this is an update of replace (it's just blank).

```
Previewing update (dev):
     Type                        Name                Plan     Info
     pulumi:pulumi:Stack         importtest3-js-dev           
     └─ kubernetes:core:Service  frontend                     [diff: ~protect,provider]
 
Resources:
    5 unchanged

Do you want to perform this update? details
  pulumi:pulumi:Stack: (same)
    [urn=urn:pulumi:dev::importtest3-js::pulumi:pulumi:Stack::importtest3-js-dev]
      kubernetes:core/v1:Service: (same) 🔓
        [id=frontend]
        [urn=urn:pulumi:dev::importtest3-js::kubernetes:core/v1:Service::frontend]
        [provider: urn:pulumi:dev::importtest3-js::pulumi:providers:kubernetes::default::6a5d1020-0254-4c14-9177-7c14aba76fd0 => urn:pulumi:dev::importtest3-js::pulumi:providers:kubernetes::default_2_4_2::ae7b250e-bd84-437e-ae1c-aeca85b79a4e]
        apiVersion: "v1"
        kind      : "Service"
        metadata  : {
            labels: {
                app.kubernetes.io/managed-by: "pulumi"
            }
            name  : "frontend"
        }
        spec      : {
            ports   : [
                [0]: {
                    port      : 80
                    targetPort: 80
                }
            ]
            selector: {
                app: "frontend"
            }
            type    : "LoadBalancer"
        }

Updating (dev):
     Type                        Name                Status     Info
     pulumi:pulumi:Stack         importtest3-js-dev             
     └─ kubernetes:core:Service  frontend                       [diff: ~protect,provider]
 ```

**Comment by lukehoban** [comment]
Collecting the feedback that I think we still need to address:
- [ ] Include prose before printing out the source text that explains what it is and what to do with it.
- [ ] Ensure that the update after an import does not show a diff.
- [ ] Fix codegen for Kubernetes `labels` (all languages)
- [ ] (maybe) Suppress generation of Kubernetes `pulumi.com/autonamed` annotation
- [ ] (maybe) Provide the ability to write output source text to a file.

**Comment by stack72** [comment]
Rebased the PR

**Comment by stack72** [comment]
Writing to a file works:

```
▶ PULUMI_EXPERIMENTAL=true pulumi import aws:s3/bucket:Bucket demo test-bucket -o index.ts
Previewing import (dev)

     Type                 Name           Plan
 +   pulumi:pulumi:Stack  config-cp-dev  create
 =   └─ aws:s3:Bucket     demo           import

Resources:
    + 1 to create
    = 1 to import
    2 changes

Do you want to perform this import? yes
Importing (dev)

     Type                 Name           Status
 +   pulumi:pulumi:Stack  config-cp-dev  created
 =   └─ aws:s3:Bucket     demo           imported

Resources:
    + 1 created
    = 1 imported
    2 changes

Duration: 13s


~/code/config-cp
▶ cat index.ts

const demo = new aws.s3.Bucket("demo", {
    acl: "private",
    bucket: "test-bucket",
    forceDestroy: false,
});
```

File: pkg/cmd/pulumi/import.go [hunk 9]
```diff
@@ -427,9 +428,10 @@ func newImportCmd() *cobra.Command {
 			}
 
 			opts.Engine = engine.UpdateOptions{
-				Parallel:      parallel,
-				Debug:         debug,
-				UseLegacyDiff: useLegacyDiff(),
+				Parallel:         parallel,
+				Debug:            debug,
+				UseLegacyDiff:    useLegacyDiff(),
+				ProtectResources: protectResources,
```
pgavlin, line 434: Let's add this as a flag on each import record rather than an `UpdateOption`.

stack72: 👍

pkg/cmd/pulumi/import.go
```diff
@@ -0,0 +1,549 @@
...
+			"The name table maps language names to parent and provider URNs. These names are\n" +
+			"used in the genrated definitions, and should match the corresponding declarations\n" +
+			"in the source program. This table is required if any parents or providers are\n" +
+			"specified by the resources to import.\n" +
+			"\n" +
+			"The resources list contains the set of resources to import. Each resource is\n" +
+			"specified as a triple of its type, name, and ID. The format of the ID is specific\n" +
+			"to the resource type. Each resource may specify the name of a parent or provider;\n" +
+			"these names must correspond to entries in the name table. If a resource does not\n" +
+			"specify a provider, it will be imported using the default provider its type. A\n" +
```
leezen, line 327:
```suggestion
			"specify a provider, it will be imported using the default provider for its type. A\n" +
```

pkg/cmd/pulumi/import.go
```diff
@@ -0,0 +1,549 @@
...
+			"resource that does not specify a provider may specify the version of the provider\n" +
```
leezen, line 328: I assume this should be "that does specify"?

pkg/cmd/pulumi/import.go
```diff
@@ -0,0 +1,549 @@
...
+			}
+
+			if outputFilePath == "" {
+				_, err := output.Write([]byte("Please copy the following code into your Pulumi application. Not doing so\n" +
+					"will cause Pulumi to report that an update will happen on the next update command.\n\n"))
+				if err != nil {
+					return result.FromError(err)
+				}
+				if protectResources {
+					_, err := output.Write([]byte("Please note, as your resources are marked as protected, then to destroy them\n" +
```
leezen, line 463:
```suggestion
					_, err := output.Write([]byte("Please note, as your resources are marked as protected, to destroy them\n" +
```

pkg/cmd/pulumi/import.go
```diff
@@ -0,0 +1,549 @@
...
+						"you will need to remove the protection attribute and run `pulumi update` *before*\n" +
```
leezen, line 464:
```suggestion
						"you will need to remove the `protect` attribute and run `pulumi update` *before*\n" +
```

pkg/cmd/pulumi/import.go
```diff
@@ -0,0 +1,549 @@
...
+			"        ]\n" +
+			"    }\n" +
+			"\n" +
+			"The name table maps language names to parent and provider URNs. These names are\n" +
+			"used in the genrated definitions, and should match the corresponding declarations\n" +
+			"in the source program. This table is required if any parents or providers are\n" +
+			"specified by the resources to import.\n" +
+			"\n" +
+			"The resources list contains the set of resources to import. Each resource is\n" +
+			"specified as a triple of its type, name, and ID. The format of the ID is specific\n" +
```
leezen, line 324: I think we can be a bit clearer about what type and name are. Otherwise, it might be confusing what is meant by 'type' here. Similarly, we should be clear how 'name' is used in the resulting generated program.

pkg/engine/import.go
```diff
@@ -0,0 +1,52 @@
+// Copyright 2016-2018, Pulumi Corporation.
```
leezen, line 1: As mentioned elsewhere, let's update the copyright headers for the new files.

pkg/resource/deploy/plan_executor.go
```diff
@@ -507,7 +545,7 @@ func (pe *planExecutor) rebuildBaseState(resourceToStep map[*resource.State]Step
 	for _, s := range pe.plan.prev.Resources {
 		var old, new *resource.State
 		if step, has := resourceToStep[s]; has {
-			// We produces a refresh step for this specific resource.  Use the new information about
+			// We producesd a refresh step for this specific resource.  Use the new information about
```
leezen, line 548: produced?

leezen (approved):
LGTM. Have a few nits.

pgavlin (commented):
Just a few nits re: wording, but LGTM otherwise.

PR #4766, opened 2020-06-06
[codegen/nodejs/sdk] Export extra files.

pgavlin:
If extra files are present, make sure that appropriate modules are
created as necessary and that extra sources are exported from each
module's `index.ts`.


pgavlin:
This is necessary to fix the test failures in https://github.com/pulumi/pulumi-aws/pull/991.

PR #4773, opened 2020-06-07
[docker/actions] Base64 decode GOOGLE_CREDENTIALS

cobraz:
Hello 👋

This PR make Pulumi Github Action allow using Google credentials that are base64 encoded.

Many workflows and Github Actions, such as [setup-gcloud action](https://github.com/GoogleCloudPlatform/github-actions/blob/master/setup-gcloud/README.md), use a base64 encoded string of Google credentials. This PR will make Pulumi Github Action work without an extra decoding step for users that combine steps requiring base64 encoded strings. There is a simple regex matcher so that it would work with both base64 and raw JSON strings and so it wouldn't be a breaking change.

Thanks for Pulumi ❤️

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ CHANGELOG
 - Export `CustomTimeouts` in the Python SDK
  [#4747](https://github.com/pulumi/pulumi/pull/4747)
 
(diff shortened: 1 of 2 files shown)


chrsmith:
Hey @cobraz , I did some sanity checking on my own shell, and things look good to me. Thank you for the PR!

PR #4775, opened 2020-06-08
Install and use dependencies automatically for new Python projects

### LINKED ISSUE #4645: Install dependencies automatically for new Python projects
Consider making it easier to get started using Python with Pulumi by automatically creating a virtual environment and installing dependencies as part of `pulumi new` (like we do with other languages), and automatically using the virtual environment when running `pulumi up`.

Basically avoid the need to show 3 separate commands that must be copy/pasted/run before running `pulumi up`.

Here's one way this could work:

[asciicast](https://asciinema.org/a/SaTwxisZ684t61t6Zbg4GSftp)

1. When creating a *new* Python project with `pulumi new`, we run `python -m venv venv` on your behalf (using the same lookup logic we use to invoke `python` today, including using `PULUMI_PYTHON_CMD` if set or looking for both  `python3` and  `python`).

2. We then update `Pulumi.yaml` to tell it to use the virtual environment:

    ```yaml
    runtime:
      name: python
      options:
        virtualenv: venv
    ```

    `virtualenv` is a new runtime option for Python. It is the path to a virtual environment for Pulumi to use. If set, we will effectively “activate” this virtual environment before invoking Python commands from Pulumi, by setting the appropriate environment variables, such that any `exec.Command` runs of `python` or `pip` from Pulumi will use the virtual environment.

3. We then install dependencies by running `pip install -r requirements.txt` (which will use the `pip` from the virtual environment), just like how we automatically install dependencies for other languages.

4. We no longer need to output commands that must be copy/pasted/run before running `pulumi up`. We simply indicate the project is ready to go and you can run `pulumi up` to deploy your project.  Since the `Pulumi.yaml` has `virtualenv` set, we run the Python program using the virtual environment when `pulumi up` is run.

If you don’t want Pulumi to automatically use the virtual environment, e.g. you want to use your own approach to virtual environments (like using Pipenv; or installing dependencies globally on the system), then simply remove the `virtualenv` option that was added to `Pulumi.yaml` (and optionally delete the `venv` directory), and then use whatever approach you want before running `pulumi` just like before. 

## Additional things to think about:

- `cdk8s init python` appears to be doing something similar (looks like it uses [Pipenv](https://pipenv.pypa.io)).

- If this only works with `pulumi new`, how can we make it easy to use for existing projects that don't already have the virtual environment created (e.g. clone git repo and trying to get started with an example).

justinvp:
Automatically create a virtual environment and install dependencies in it with `pulumi new` and `pulumi policy new` for Python templates.

This will save a new `virtualenv` runtime option in `Pulumi.yaml` (`PulumiPolicy.yaml` for policy packs):

```yaml
runtime:
  name: python
  options:
    virtualenv: venv
```

`virtualenv` is the path to a virtual environment that Pulumi will use when running `python` commands.

Existing projects are unaffected and can opt-in to using this by setting `virtualenv`, otherwise, they'll continue to work as-is.

### Commits in this PR:

1. Create virtual environment, install deps, and set `virtualenv` runtime option with `pulumi new` and `pulumi policy new`
2. Use the virtual environment from the Python language plugin
3. Use the virtual environment for Python dynamic providers
4. Add some tests

Fixes #4645

---

When Luke and I chatted about this a while back, he was wondering how we could make it easy to use for existing projects that don't already have the virtual environment created (e.g. git clone examples repo and trying to get started with an example). As a separate change, I think we should consider an additional CLI command or gesture to make this easier. Something like `pulumi stack prepare` or `pulumi install` that would "prepare a project". Or potentially have `pulumi stack init` prompt to create the virtual environment and install dependencies automatically (if interactive). Opened https://github.com/pulumi/pulumi/issues/4777 to track this.

### DIFF
--- a/sdk/go/common/resource/plugin/provider_plugin.go
+++ b/sdk/go/common/resource/plugin/provider_plugin.go
@@ -64,7 +64,8 @@ type provider struct {
 
 // NewProvider attempts to bind to a given package's resource plugin and then creates a gRPC connection to it.  If the
 // plugin could not be found, or an error occurs while creating the child process, an error is returned.
-func NewProvider(host Host, ctx *Context, pkg tokens.Package, version *semver.Version) (Provider, error) {
+func NewProvider(host Host, ctx *Context, pkg tokens.Package, version *semver.Version,
+	options map[string]interface{}) (Provider, error) {
 	// Load the plugin's path by using the standard workspace logic.
 	_, path, err := workspace.GetPluginPath(
 		workspace.ResourcePlugin, strings.Replace(string(pkg), tokens.QNameDelimiter, "_", -1), version)
@@ -78,8 +79,12 @@ func NewProvider(host Host, ctx *Context, pkg tokens.Package, version *semver.Ve
 		})
 	}
 
-	plug, err := newPlugin(ctx, ctx.Pwd, path, fmt.Sprintf("%v (resource)", pkg),
-		[]string{host.ServerAddr()}, nil /*env*/)
+	args := []string{host.ServerAddr()}
+	for k, v := range options {
+		args = append(args, fmt.Sprintf("-%s=%v", k, v))
+	}
+
+	plug, err := newPlugin(ctx, ctx.Pwd, path, fmt.Sprintf("%v (resource)", pkg), args, nil /*env*/)
 	if err != nil {
 		return nil, err
 	}
--- a/sdk/go/common/resource/plugin/provider_plugin.go
+++ b/sdk/go/common/resource/plugin/provider_plugin.go
@@ -18,6 +18,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"os"
 	"strings"
 
 	"github.com/blang/semver"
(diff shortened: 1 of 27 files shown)


sdk/go/common/resource/plugin/provider_plugin.go
```diff
@@ -78,8 +79,12 @@ func NewProvider(host Host, ctx *Context, pkg tokens.Package, version *semver.Ve
 		})
 	}
 
-	plug, err := newPlugin(ctx, ctx.Pwd, path, fmt.Sprintf("%v (resource)", pkg),
-		[]string{host.ServerAddr()}, nil /*env*/)
+	args := []string{host.ServerAddr()}
+	for k, v := range options {
+		args = append(args, fmt.Sprintf("-%s=%v", k, v))
+	}
```
justinvp, line 85: To make this work for dynamic providers in Python, I needed some way to pass the `virtualenv` runtime option to the `pulumi-resource-pulumi-python` provider plugin. This change passes along all runtime options as command line flags to providers. Are we OK doing this? (I looked and all _our_ providers should be able to handle the additional command line flags; they'll be ignored). Should we prefix these flag names with `runtime-` or something else to make it more clear where these are coming from?

pgavlin: Environment variables seem like a safer option here--have you given those a try?

justinvp: Environment variables should be able to work for this. I can update the PR to pass them as `PULUMI_RUNTIME_<OPTION>`.

Note that this does mean the provider plugins would be receiving these values differently than the language and analyzer plugins, which receive them as command line flags.

PR #4787, opened 2020-06-09
[codegen/python] Allow import path overrides

komalali:

### DIFF
--- a/pkg/codegen/python/importer.go
+++ b/pkg/codegen/python/importer.go
@@ -30,6 +30,11 @@ type PackageInfo struct {
 	Requires map[string]string `json:"requires,omitempty"`
 	// Readme contains the text for the package's README.md files.
 	Readme string `json:"readme,omitempty"`
+	// Map from module -> package name
+	//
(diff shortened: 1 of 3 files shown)


pkg/codegen/python/importer.go
```diff
@@ -30,6 +30,11 @@ type PackageInfo struct {
 	Requires map[string]string `json:"requires,omitempty"`
 	// Readme contains the text for the package's README.md files.
 	Readme string `json:"readme,omitempty"`
+	// Map from module -> package name
+	//
+	//    { "flowcontrol.apiserver.k8s.io/v1alpha1": "flowcontrol/v1alpha1" }
+	//
+	ModuleToPackage map[string]string `json:"moduleToPackage,omitempty"`
```
pgavlin, line 37: Nit: `Package` refers to Go packages. This should be named something like `ModuleNameOverrides`:

```suggestion
	// Optional overrides for Pulumi modules
	//
	//    { "flowcontrol.apiserver.k8s.io/v1alpha1": "flowcontrol/v1alpha1" }
	//
	ModuleNameOverrides map[string]string `json:"moduleNameOverrides,omitempty"`
```

PR #4796, opened 2020-06-10
[codegen/python] Fix relative import path

komalali:

### DIFF
--- a/pkg/codegen/python/gen_test.go
+++ b/pkg/codegen/python/gen_test.go
@@ -1,0 +1,27 @@
+package python
+
+import "testing"
+
+var pathTests = []struct {
+	input    string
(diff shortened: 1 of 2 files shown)


pkg/codegen/python/gen_test.go
```diff
@@ -0,0 +1,27 @@
...
+	{"", "."},
+	{"../", ".."},
+	{"../..", "..."},
+	{"../../..", "...."},
+	{"something", ".something"},
+	{"../parent", "..parent"},
+	{"../../module", "...module"},
+}
+
+func TestRelPathToRelImport(t *testing.T) {
```
pgavlin, line 18: Awesome!

Issue #4296, opened 2020-04-03
Go config doesn't implicitly use current project


joeduffy:
In other SDKs, I can use simple, unqualified names to set and retrieve config. For instance, if my project's name is `foo`, I can `pulumi config set bar` and retrieve it at runtime simply using the key `"bar"`. It appears that doesn't work in Go, and I need to use the fully qualified name, `"foo:bar"` when retrieving values using the `github.com/pulumi/pulumi/sdk/go/pulumi/config` helper functions.

joeduffy:
As a workaround, it appears that https://godoc.org/github.com/pulumi/pulumi/sdk/go/pulumi/config#New does automatically prepend the project name.

PR #4802, opened 2020-06-11
Set default namespace for go config for configless of get/try/require

EvanBoyle:
Fixes https://github.com/pulumi/pulumi/issues/4296

As far as I can tell, other languages don't expose the config-less form of get/try/require methods. Exposing these in go was probably a bug, but given that removing them would be breaking this is the next best thing.

This change reverses the logic used in the "pulumi config set" command: https://github.com/pulumi/pulumi/blob/8bc3d0e906cc7eed09a4e85a1a70e8f64d8767b3/pkg/cmd/pulumi/config.go#L381

Given that you can't set a "namespace-less" config key, this change seems correct. Updated the integration test coverage to include this behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ CHANGELOG
 
 ## HEAD (Unreleased)
 
+- Set default config namespace for Get/Try/Require methods in Go SDK.
+  [4802](https://github.com/pulumi/pulumi/pull/4802)
+
 - Improve typing for Go SDK secret config values
  [#4800](https://github.com/pulumi/pulumi/pull/4800)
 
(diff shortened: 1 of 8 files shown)


pgavlin:
> Given that you can set a "namespace-less" config key, this change seems correct. 

Should this read "cannot set a namespace-less" config key? Otherwise this is a breaking change, no?

EvanBoyle:
@pgavlin Yes, that is a typo. You __cannot__ set a "namespace-less" key. The code linked here automatically prepends the project if it doesn't find the separator token `:` 

https://github.com/pulumi/pulumi/blob/8bc3d0e906cc7eed09a4e85a1a70e8f64d8767b3/pkg/cmd/pulumi/config.go#L381

Issue #4811, opened 2020-06-12
Github Actions: With GCP:  Configure Docker


jonsherrard:
Hi there,

I've run into an error where Docker is not authenticated to push to the GDP container registry:

In the Github Action entrypoint script, service accounts are enabled using `gcloud` CLI, but there is another step required to authenticate docker (required for pushing images), to the gcr.io

https://github.com/pulumi/pulumi/blob/master/dist/actions/entrypoint.sh#L88

Should https://cloud.google.com/sdk/gcloud/reference/auth/configure-docker the `configure docker` command also be run at this line?

stack72:
Hi @jonsherrard 

Thanks for the report here - let me look into this and get back to you asap!

Paul

stack72:
You were indeed correct @jonsherrard 

PR has been opened to fix this - added the --quiet attribute on the auth to ensure it doesn't flood the logs. We do this on our internal repos for that specific command

jonsherrard:
Thank you very much @stack72 - I'll keep an eye on the PR

PR #4812, opened 2020-06-12
Ensure Authenticated to GCR when GOOGLE_CREDENTIALS supplied to action

stack72:
Fixes: #4811 

This is in line with what we need to do in the examples repo when we supply GCP Credentials


jonsherrard:
Awesome thanks @stack72. When does the docker.hub pulumi bot push these changes?

PR #4758, opened 2020-06-05
Add GitHub Actions detection and update the Pulumi CI escape hatch with more CI vars

praneetloke:
This PR adds the detection logic for GitHub Actions. I have also updated our escape-hatch code for unknown CI systems to allow users to set all CI vars that we support today.

This PR might address the problem reported in https://github.com/pulumi/pulumi/issues/4613.

Issue #4613, opened 2020-05-12
GitHub App not updating PR status/check


clstokes:
We've had multiple users in our Community Slack report that the Pulumi GitHub App is not updating the status/check on pull requests with the results of the `pulumi preview`. I've not been able to reproduce this, so opening this issue on their behalf.

For clarity:
![image](https://user-images.githubusercontent.com/293763/81731528-e857a880-9443-11ea-8632-b5d80b801e21.png)

See https://pulumi-community.slack.com/archives/C84L4E3N1/p1585163434020800 for more context.

praneetloke:
@danfhernandez @clstokes this should now be resolved with Pulumi CLI version 2.5.0. I'll close this out if one of you could also try it out?

Incisive:
Works great on my end, thanks for addressing this !

praneetloke:
Awesome! Thanks for the checking, @Incisive.

PR #4817, opened 2020-06-12
Use the GH webhook payload to parse PR details

praneetloke:
Related to #4758 and #4613.

In a previous PR, I added the CI env vars detection for GH Actions. Unfortunately, it looks like there is some problem with the values of `GITHUB_REF` and `GITHUB_SHA` for a PR. It looks like the example shown in the [doc](https://help.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables#default-environment-variables) is not what we are actually seeing. To demonstrate this, I opened a draft [PR](https://github.com/pulumi/docs/pull/3584/checks?check_run_id=766615693) in our docs repo, which has an extra step to print out the [env vars](https://github.com/pulumi/docs/pull/3584/checks?check_run_id=766615693#step:12:1) and the [event JSON](https://github.com/pulumi/docs/pull/3584/checks?check_run_id=766615693#step:13:7) separately. Note how the value of the `GITHUB_REF` value does not match the documented example of `ref/heads/<branch name>`. And for `GITHUB_SHA` -- I actually can't even tell what this SHA is. It doesn't look like it is the value of anything that I can correlate. Perhaps I read the docs incorrectly and misunderstood what they were supposed to be.

In any case, if you look at the event JSON, the values that we are looking to use are correctly specified in there, including the `pull_request.head.ref`, which has the PR branch name and the SHA value in `pull_request.head.sha`.

In addition to the problem with the values of the two env vars, the PR number extraction was incorrect. In the previous PR, I was trying to unmarshal the value of the `GITHUB_EVENT_PATH` directly, which will never work since it is just the _path_ to the JSON file that actually contains the event. This would have only caused us to _not_ set the `ci.pr.number` env metadata key. I have fixed this as well now.

### DIFF
--- a/sdk/go/common/util/ciutil/github_actions.go
+++ b/sdk/go/common/util/ciutil/github_actions.go
@@ -75,9 +75,9 @@ func (t githubActionsCI) DetectVars() Vars {
 }
 
 // TryGetEvent returns the GitHub webhook payload found in the GitHub Actions environment.
-// GitHub stores the JSON payload of the webhook that triggered the  workflow in a path.
+// GitHub stores the JSON payload of the webhook that triggered the workflow in a path.
 // The path is set as the value of the env var GITHUB_EVENT_PATH. Returns nil if an error
-// is encountered of the GITHUB_EVENT_PATH is not set.
+// is encountered or the GITHUB_EVENT_PATH is not set.
 func (t githubActionsCI) GetPREvent() *githubActionsPullRequestEvent {
 	eventPath := os.Getenv("GITHUB_EVENT_PATH")
 	if eventPath == "" {
--- a/sdk/go/common/util/ciutil/github_actions.go
+++ b/sdk/go/common/util/ciutil/github_actions.go
@@ -63,7 +63,7 @@ func (t githubActionsCI) DetectVars() Vars {
 
 	v.SHA = os.Getenv("GITHUB_SHA")
 	if v.BuildType == "pull_request" {
-		event := t.TryGetEvent()
+		event := t.GetPREvent()
 		if event != nil {
 			prNumber := strconv.FormatInt(event.Number, 10)
 			v.PRNumber = prNumber
@@ -78,18 +78,18 @@ func (t githubActionsCI) DetectVars() Vars {
 // GitHub stores the JSON payload of the webhook that triggered the  workflow in a path.
 // The path is set as the value of the env var GITHUB_EVENT_PATH. Returns nil if an error
 // is encountered of the GITHUB_EVENT_PATH is not set.
-func (t githubActionsCI) TryGetEvent() *githubActionsPullRequestEvent {
+func (t githubActionsCI) GetPREvent() *githubActionsPullRequestEvent {
 	eventPath := os.Getenv("GITHUB_EVENT_PATH")
 	if eventPath == "" {
 		return nil
 	}
 
-	var prEvent githubActionsPullRequestEvent
 	b, err := ioutil.ReadFile(eventPath)
 	if err != nil {
 		return nil
 	}
 
+	var prEvent githubActionsPullRequestEvent
 	if err := json.Unmarshal(b, &prEvent); err != nil {
 		return nil
 	}
--- a/sdk/go/common/util/ciutil/github_actions.go
+++ b/sdk/go/common/util/ciutil/github_actions.go
@@ -17,6 +17,7 @@ package ciutil
 import (
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"strconv"
 )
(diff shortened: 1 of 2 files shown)


chrsmith:
So the behavior from the GitHub Actions sounds like it makes sense to me, or at least seems reasonable.

> Unfortunately, it looks like there is some problem with the values of GITHUB_REF and GITHUB_SHA for a PR. It looks like the example shown in the doc is not what we are actually seeing. To demonstrate this, I opened a draft PR in our docs repo, which has an extra step to print out the env vars and the event JSON separately. 

The `GITHUB_REF` env var was "refs/pull/3584/merge", which while obviously not as helpful as the docs, does imply that it is running on the _merge commit_ that would be generated if the pull request were merged with the target branch.

So that's probably why you couldn't associate the SHA with anything, because it's not the SHA of the pull request branch, but the "pull request branch + target branch".

The [GitHub event(https://github.com/pulumi/docs/pull/3584/checks?check_run_id=766615693#step:13:7) printed a "synchronize" action, which if I recall correctly is what happens whenever a push is made to a branch that is tracked by a pull request.

The Pulumi GitHub app listens for those and writes them into the `GitCommitListeners` table, [here](https://github.com/pulumi/pulumi-service/blob/master/cmd/service/workflow/pull_requests.go#L121).  ([Relevant GitHub docs](https://developer.github.com/webhooks/event-payloads/#pull_request) for context.)

> In any case, if you look at the event JSON, the values that we are looking to use are correctly specified in there, including the pull_request.head.ref, which has the PR branch name and the SHA value in pull_request.head.sha.

That sounds right. And appears to be what the Pulumi GitHub App is doing when it sees the "synchronize" event as well. ([Code](https://github.com/pulumi/pulumi-service/blob/master/cmd/service/workflow/pull_requests.go#L154)).

praneetloke:
> The GITHUB_REF env var was "refs/pull/3584/merge", which while obviously not as helpful as the docs, does imply that it is running on the merge commit that would be generated if the pull request were merged with the target branch.

Correct. The values of the env var as-is makes sense to me as well. I am just not sure when `GITHUB_REF` would ever be `refs/heads/feature-branch` as the docs demonstrate? One thing is for sure -- it doesn't seem to have that value when a PR is opened or updated, which is what matters to the CLI. Moreover, the `GITHUB_SHA` doesn't actually even look like the merge commit SHA, if you look at the webhook payload. See `pull_request.merge_commit_sha`. It's `null` when a PR is opened, but has a value for subsequent PR `synchronize` actions. which does not match the value of `GITHUB_SHA`. I am not sure if I am missing something?

> The [GitHub event](https://github.com/pulumi/docs/pull/3584/checks?check_run_id=766615693#step:13:7) printed a "synchronize" action, which if I recall correctly is what happens whenever a push is made to a branch that is tracked by a pull request.

Correct. I pasted the link to the GHA execution for the `synchronize` action. Here's the [`opened`](https://github.com/pulumi/docs/runs/766530962?check_suite_focus=true#step:3:7) action for that same PR. (I had opened that PR by removing all of the original steps before I restored them in a subsequent `push`.)

> The only suggestion is that maybe we would want to confirm the githubActionsPullRequestEvent action field? Since there are a bunch of different ones, and potentially the way we would interpret the SHA/branch might be different? (But it doesn't look like the case, since I doubt the workflow would trigger in response to the GitHub PR having review_requested or being locked, etc.)

Yeah it doesn't seem to be the case. The CLI would just need to take whatever value is in the PR webhook payload from the workflow environment. I do have a check that we would parse the webhook payload event only if `GITHUB_EVENT_NAME` is `pull_request`.

sdk/go/common/util/ciutil/github_actions.go
```diff
@@ -26,45 +27,72 @@ type githubActionsCI struct {
...
+			v.PRNumber = prNumber
+			v.SHA = event.PullRequest.Head.SHA
+			v.BranchName = event.PullRequest.Head.Ref
 		}
 	}
 	return v
 }
+
+// TryGetEvent returns the GitHub webhook payload found in the GitHub Actions environment.
+// GitHub stores the JSON payload of the webhook that triggered the  workflow in a path.
```
praneetloke, line 78:
```suggestion
// GitHub stores the JSON payload of the webhook that triggered the workflow in a path.
```

sdk/go/common/util/ciutil/github_actions.go
```diff
@@ -26,45 +27,72 @@ type githubActionsCI struct {
...
+// The path is set as the value of the env var GITHUB_EVENT_PATH. Returns nil if an error
+// is encountered of the GITHUB_EVENT_PATH is not set.
```
praneetloke, line 80:
```suggestion
// is encountered or the GITHUB_EVENT_PATH is not set.
```

Issue #3789, opened 2020-01-23
Smaller Docker images, one for each language


Dominik-K:
The current [`pulumi/pulumi` Docker image `v1.8.1.`](https://hub.docker.com/layers/pulumi/pulumi/v1.8.1/images/sha256-e2239928f3dfe52ddb38deda5ddcb41e71e295e4d70fbd1276bd8cfc3e025c88) is **2.44GB** big (compressed: 694.17 MB). Mainly because it contains all frameworks for the different languages. That makes CI builds unnecessary long & using it locally is really uncomfortable, especially when you have small bandwidth available (train, etc.).

I suggest to publish Docker images for each language as a specific image tag. E.g. `pulumi/pulumi:1.8.1-node` for the Typescript version (similar like @Charliekenney23 [suggested](https://github.com/pulumi/pulumi/issues/2891#issue-462439168)). That will reduce the image size down to ca. **200MB** (as @boillodmanuel  [showed it in another issue](https://github.com/pulumi/pulumi/issues/1986#issuecomment-544866074)).

piclemx:
@chrsmith any news on this one?

chrsmith:
No update as of right now, sorry @piclemx . It's something that I've wanted to get to for a while, alas [we've been busy](https://www.pulumi.com/webinars/pulumi-update-2020-04-29/) with a lot of things. However, I will land https://github.com/pulumi/pulumi/pull/4354 soon, which will at least make it much easier to build/test/verify any new containers we publish. So 🤞 this will happen sooner rather than later.

piclemx:
No problem @chrsmith, let me know if you want helps on this. Will be happy to contribue.

jaxxstorm:
@chrsmith after looking at #4599 I had a look at this here:
https://github.com/jaxxstorm/pulumi-docker

Let me know what you think. The images are quite a bit smaller. I'm happy to take this and run with it if you need.

chrsmith:
@jaxxstorm yeah, that's ~exactly what I was thinking we'd want to do! Please do go and run with it :)

There are a few design considerations I think we'd need to answer before we "ship" these, just so that we know how it is we recommend people use them.

[...]

PR #4837, opened 2020-06-16
Adding the langage SDK specific docker images

stack72:
Fixes: #3789


stack72:
It’s worth noticed a few things here:

- these will not replace the existing container but live as well as that
- these containers are language specific to make them lightweight and available for those who want to do some custom work with Pulumi in their pipeline
- any changes will largely be driven by things like security patches

Issue #4632, opened 2020-05-13
[codegen/*] Render property references in comments in the target language.


pgavlin:
e.g. if a deprecation message references the property `b64Std`, we should render this reference as `b64_std` in Python or `B64Std` in Go.

We could go two ways with this:
- Add markup for property references (e.g. `[b64Std](#/resources/random:index%2FrandomId:RandomId/b64Std)`)
- Use a heuristic to detect and rerender property references.

Personally, I'm (unsurprisingly) in favor of the former, as some property names are likely to be overlap with normal vocabulary.

pgavlin:
(cc @praneetloke @lukehoban @cnunciato)

praneetloke:
A while back I filed https://github.com/pulumi/pulumi/issues/4159 for something similar. Might be related?

pgavlin:
Ah yep, that's definitely related

pgavlin:
related: https://github.com/pulumi/pulumi/issues/16099

Issue #4159, opened 2020-03-23
Update the links in the resource description to be language-specific


praneetloke:
Related to #4158.

Two issues:
1. At least in the current TypeScript codegen (haven’t checked other languages), we attempt to fix-up the doc comment to use the correct names.
See https://github.com/pulumi/pulumi-aws/blob/b399e2d36394582be9403feb5c63b086c782432a/sdk/nodejs/emr/cluster.ts#L567-L572
```
/**
 * Use the `coreInstanceGroup` configuration block `instanceCount` argument instead. Number of Amazon EC2 instances used to execute the job flow. EMR will use one node as the cluster's master node and use the remainder of the nodes (`coreInstanceCount`-1) as core nodes. Cannot be specified if `coreInstanceGroup` or `instanceGroup` configuration blocks are set. Default `1`
 *
 * @deprecated use `core_instance_group` configuration block `instance_count` argument instead
 */
readonly coreInstanceCount?: pulumi.Input<number>;
```
2. For TypeScript, we fix-up the doc comment, but it looks like we’re not doing that for the `@deprecated` message which gets appended at the end of the doc comment.

pgavlin:
related: https://github.com/pulumi/pulumi/issues/16099

PR #4838, opened 2020-06-16
[schema/docs] Use a Markdown parser.

pgavlin:
In particular, use the parser to filter and extract examples. This also
sets up support for entity references in documentation that can be used
in order to render language-specific names for resources, functions,
types, and properties.

Related to #4632 and #4159.

### DIFF
--- a/pkg/codegen/docs.go
+++ b/pkg/codegen/docs.go
@@ -17,7 +17,7 @@ package codegen
 import (
 	"regexp"
 
-	"github.com/yuin/goldmark/ast"
+	"github.com/pgavlin/goldmark/ast"
 
 	"github.com/pulumi/pulumi/pkg/v2/codegen/schema"
 	"github.com/pulumi/pulumi/sdk/v2/go/common/util/contract"
--- a/pkg/codegen/docs.go
+++ b/pkg/codegen/docs.go
@@ -18,7 +18,10 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/yuin/goldmark/ast"
+
 	"github.com/pulumi/pulumi/pkg/v2/codegen/schema"
+	"github.com/pulumi/pulumi/sdk/v2/go/common/util/contract"
 )
 
 var (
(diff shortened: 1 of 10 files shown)


pkg/go.mod
```diff
@@ -61,3 +62,5 @@ require (
 	sourcegraph.com/sourcegraph/appdash v0.0.0-20190731080439-ebfcffb1b5c0
 	sourcegraph.com/sourcegraph/appdash-data v0.0.0-20151005221446-73f23eafcf67 // indirect
 )
+
+replace github.com/yuin/goldmark => github.com/pgavlin/goldmark v1.1.33-0.20200616164752-aaa9ff3e1d61
```
pgavlin, line 66: This is breaking the downstream testing. I'm going to move my fork into its own module to avoid this.

pgavlin:
The GCP diff failure is due to an error downloading the resource plugin.

pkg/codegen/docs.go
```diff
@@ -16,9 +16,11 @@ package codegen
...
 	"regexp"
-	"strings"
+
+	"github.com/pgavlin/goldmark/ast"
 
 	"github.com/pulumi/pulumi/pkg/v2/codegen/schema"
+	"github.com/pulumi/pulumi/sdk/v2/go/common/util/contract"
 )
 
 var (
```
praneetloke, line 26: Can we now get rid of the regex's from these [lines](https://github.com/pulumi/pulumi/pull/4838/files#diff-1bed47e82076977fc5ffa09e15aa568dL25-L51)?

pkg/codegen/docs.go
```diff
@@ -118,63 +115,83 @@ func GetAllMatchedGroupsFromRegex(regex *regexp.Regexp, str string) map[string][
...
-		snippet = CSharpCodeSnippetRE.FindString(exampleContent)
-	case "go":
-		snippet = GoCodeSnippetRE.FindString(exampleContent)
-	case "python":
-		snippet = PythonCodeSnippetRE.FindString(exampleContent)
-	case "typescript":
-		snippet = TSCodeSnippetRE.FindString(exampleContent)
+	var examples ast.Node
+	err := ast.Walk(parsed, func(n ast.Node, enter bool) (ast.WalkStatus, error) {
+		if shortcode, ok := n.(*schema.Shortcode); ok && string(shortcode.Name) == "examples" {
```
praneetloke, line 130: Would be good to declare the name of the shortcode as a constant.

pkg/codegen/docs.go
pgavlin: As long as they're not used by any consuming packages, yes. I wasn't sure if e.g. the docs site used them.

praneetloke: Yeah since the functions that were using them have been removed with this PR, we should be fine to remove them assuming the build also passes.

Issue #4836, opened 2020-06-16
[Go Program Gen] Passing test for aws-webserver.pp


EvanBoyle:
It looks like the major item here is multi-line strings.

Issue #4770, opened 2020-06-06
[Go Program Gen] Resolve ambiguity between .Get and .Lookup methods for invoke


EvanBoyle:
Our naming here relies on presence of other functions within the schema and cannot be determined without this context. 

Today we just use the Lookup name, but this is not correct. See: https://github.com/pulumi/pulumi/pull/4749/files#diff-ae83e82bc7feedf66a792c8a49a12105R15-R26

We need to replicate the logic that generates the provider code: https://github.com/pulumi/pulumi/blob/adfa8116fbf61ad328568fe4b53e32c751f30553/pkg/codegen/go/gen.go#L1171

PR #4850, opened 2020-06-18
[Go Program Gen] multiline strings, get/lookup disambiguation, webserver example

EvanBoyle:
Fixes https://github.com/pulumi/pulumi/issues/4836
Fixes https://github.com/pulumi/pulumi/issues/4770

In addition to the title items, I also added support for "namespaceless" invokes for things under the "index" or "" module. This also illuminated an import bug, where we weren't scanning invoke functions for packages so I added support for this.

Pushed another commit to fix https://github.com/pulumi/pulumi/issues/4762 (inputty map types for pulumi.Map (Tags))

### DIFF
--- a/pkg/codegen/go/gen_program.go
+++ b/pkg/codegen/go/gen_program.go
@@ -13,6 +13,7 @@ import (
 	"github.com/pulumi/pulumi/pkg/v2/codegen/hcl2"
 	"github.com/pulumi/pulumi/pkg/v2/codegen/hcl2/model"
 	"github.com/pulumi/pulumi/pkg/v2/codegen/hcl2/model/format"
+	"github.com/pulumi/pulumi/pkg/v2/codegen/schema"
 	"github.com/pulumi/pulumi/sdk/v2/go/common/util/contract"
 )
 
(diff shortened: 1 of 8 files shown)


pkg/codegen/go/gen_program_expressions.go
```diff
@@ -780,6 +797,11 @@ func (g *generator) escapeString(v string) string {
 		if c == '"' || c == '\\' {
 			builder.WriteRune('\\')
 		}
+		if c == '\n' {
```
pgavlin, line 800: It would be more idiomatic here to use raw string literals (which are escaped with `` ` ``) or to put each string literal on its own line and concatenate them, but what you have is simpler. You might file a follow-up issue to track generating the more idiomatic code.

pkg/codegen/go/gen_program.go
```diff
@@ -331,3 +388,23 @@ func (g *generator) genLocalVariable(w io.Writer, v *hcl2.LocalVariable) {
 	}
 
 }
+
+func (g *generator) useLookupInvokeForm(token string) bool {
```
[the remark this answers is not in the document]
EvanBoyle: Good thinking, I added a comment with some links to the AWS SDK to clarify what this does.

pkg/codegen/go/gen_program_expressions.go
EvanBoyle: https://github.com/pulumi/pulumi/issues/4855

PR #4856, opened 2020-06-18
Fix lint errors.

pgavlin:
These errors appear when using golangci-lint 1.27.0.


pgavlin:
Once this is merged, I'll update the version of `golangci-lint` that is installed by the scripts repo.

PR #4873, opened 2020-06-22
[codegen/python] Don't use __all__

komalali:
`__all__` is a list of public objects of a module. Using it overrides the default of hiding everything that begins with an underscore. So if we use `__all__` we must list everything we want to make available publicly, including classes (which we currently do not). - [[source]](https://stackoverflow.com/questions/44834/can-someone-explain-all-in-python)

The way we are currently using `__all__` isn't leveraging this functionality, we are just using it as a list of submodules we want to import. So, renaming the list to `submodules` (or really anything else) means we use the default functionality of making all classes available publicly, and we can get rid of the warning in IDEs like IntelliJ and PyCharm.

Also some whitespace changes and reorganizing to make the linter happy.

Before:
Screen Shot 2020-06-22 at 11 37 34 AM

After:
Screen Shot 2020-06-22 at 4 31 47 PM

### DIFF
--- a/pkg/codegen/python/gen.go
+++ b/pkg/codegen/python/gen.go
@@ -255,7 +255,7 @@ func (mod *modContext) genInit(exports []string) string {
 
 		fmt.Fprintf(w, "import importlib\n")
 		fmt.Fprintf(w, "# Make subpackages available:\n")
-		fmt.Fprintf(w, "__all__ = [")
+		fmt.Fprintf(w, "submodules = [")
 		for i, mod := range mod.children {
 			child := mod.mod
 			if mod.compatibility == kubernetes20 {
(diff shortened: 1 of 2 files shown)


pkg/codegen/python/gen.go
```diff
@@ -298,6 +269,34 @@ func (mod *modContext) genInit(exports []string) string {
...
 	}
 
+	// If there are subpackages, import them with importlib.
+	if len(mod.children) > 0 {
+		sort.Slice(mod.children, func(i, j int) bool {
+			return PyName(mod.children[i].mod) < PyName(mod.children[j].mod)
+		})
+
+		fmt.Fprintf(w, "\n# Make subpackages available:\n")
+		fmt.Fprintf(w, "submodules = [\n")
```
justinvp, line 279: Should we name it with an underscore, e.g. `_submodules` so it is not "public"?

pkg/codegen/python/gen.go
```diff
@@ -298,6 +269,34 @@ func (mod *modContext) genInit(exports []string) string {
 		}
 	}
 
+	// If there are subpackages, import them with importlib.
+	if len(mod.children) > 0 {
```
justinvp, line 273: Should we do an up-front check to see if the length == 1 and only contains `config`, and avoid emitting any code in that case.

See https://github.com/pulumi-bot/pulumi-azuread/compare/cac6eb2..8a632b7#diff-16689a6998d278c7164a5a1c26081cb3R26-R32

```python
# Make subpackages available:
submodules = [
    'config',
]
for pkg in submodules:
    if pkg != 'config':
        importlib.import_module(f'{__name__}.{pkg}')
```

PR #4874, opened 2020-06-23
Fix setup.py

komalali:
A local pip install i.e. `pip install -e path/to/dir` fails because there is no README.md file in the python sdk directory.


komalali:
I was installing with the wrong path. I was doing `pip install -e ~/go/src/github.com/pulumi/pulumi/sdk/python` but the correct path is `pip install -e ~/go/src/github.com/pulumi/pulumi/sdk/python/env/src`

Issue #4902, opened 2020-06-25
Case mapping tables don't account for many-to-one scenarios


komalali:
Both "jsonPath" and "JSONPath" currently map to "json_path" in the camelCase to snake_case conversion which results in one of those properties being mapped back incorrectly. We need this to not happen. 

More context: https://github.com/pulumi/pulumi-kubernetes/pull/1160#discussion_r445205353

justinvp:
Let's also make sure we have a plan for this with https://github.com/pulumi/pulumi/issues/3771

komalali:
@justinvp I think #4789 actually resolves this? Should I reassign to you?

justinvp:
> I think #4789 actually resolves this? Should I reassign to you?

Yeah, I'd be ok treating #4789 as resolving this. I'll reassign to myself.

I suppose we could consider enhancing our mapping table scheme to better accommodate this, but I don't think it's worth it for a couple reasons:

1. Using the input/output classes doesn't have this problem as the snake_case names always map to the correct Pulumi name, and we want folks to adopt this new approach going forward. (We may choose to deprecate the "old way" where this is problematic in the next major provider revision anyway).

2. In this particular case, `JSONPath` is coming from `kubernetes:apiextensions.k8s.io/v1beta1:CustomResourceColumnDefinition` (**v1beta1**), whereas the **v1** is `jsonPath`. We're currently using the v1 casing in our mapping tables, and using that over the v1beta1 casing makes sense to me.

justinvp:
I'm going to close this since #4789 will resolve it.

PR #4907, opened 2020-06-26
[codegen/python] Only add item to map if it doesn't already exist.

komalali:


komalali:
This is really just a bandaid for the actual issue that needs to be addressed here: https://github.com/pulumi/pulumi/issues/4902

PR #4908, opened 2020-06-26
Avoid stack overflow in Python codegen in case of recursive types

mikhailshilkov:
Got a stack overflow exception while generating the Python SDK for the (prototype) Azure RM provider. It has types referencing each other, which kills the casing building function.


komalali:
I actually just made this PR and I wonder if it solves the same problem: https://github.com/pulumi/pulumi/pull/4907

PR #4913, opened 2020-06-26
[python/codegen] - Don't add underscore at the end of module names.

komalali:
Fixes: https://github.com/pulumi/pulumi-aws/issues/1020


komalali:
This is a feature not a bug! We need the underscore, module names can't be keywords 🤦

Issue #4903, opened 2020-06-25
[Go Program Gen] Incorrect types for various pulumi.Map types


EvanBoyle:
In a recent bump of the [AWS sdk](https://github.com/pulumi/pulumi-aws/blame/a9a2c4fa78bd695a201fd3a2219411caf7cbef3a/sdk/go/aws/mediapackage/channel.go#L52) the type for `Tags` changed from `pulumi.Map` to `pulumi.StringMap`. This was the only test coverage we had for pulumi Map types. We need to improve the type detection logic to support pulumi.StringMap, pulumi.BoolMap, etc. These do result in compilation errors:

```go
	"github.com/pulumi/pulumi-aws/sdk/v2/go/aws/acm"
	"github.com/pulumi/pulumi/sdk/v2/go/pulumi"
)

func main() {
	pulumi.Run(func(ctx *pulumi.Context) error {
		_, err = acm.NewCertificate(ctx, "cert", &acm.CertificateArgs{
			DomainName: pulumi.String("example.com"),
			Tags: pulumi.Map{
				"Environment": pulumi.String("test"),
			},
			ValidationMethod: pulumi.String("DNS"),
		})
		if err != nil {
			return err
		}
		return nil
	})
}
```

```go
package main

import (
	"fmt"

	"github.com/pulumi/pulumi-aws/sdk/v2/go/aws/apigateway"
	"github.com/pulumi/pulumi/sdk/v2/go/pulumi"
)

func main() {
	pulumi.Run(func(ctx *pulumi.Context) error {
		myDemoAPI, err := apigateway.NewRestApi(ctx, "myDemoAPI", &apigateway.RestApiArgs{
			Description: pulumi.String("This is my API for demonstration purposes"),
		})
		if err != nil {
			return err
		}
		myDemoResource, err := apigateway.NewResource(ctx, "myDemoResource", &apigateway.ResourceArgs{
			ParentId: myDemoAPI.RootResourceId,
			PathPart: pulumi.String("mydemoresource"),
			RestApi:  myDemoAPI.ID(),
		})
		if err != nil {
			return err
		}
		myDemoMethod, err := apigateway.NewMethod(ctx, "myDemoMethod", &apigateway.MethodArgs{
			Authorization: pulumi.String("NONE"),
			HttpMethod:    pulumi.String("GET"),
			ResourceId:    myDemoResource.ID(),
			RestApi:       myDemoAPI.ID(),
		})
		if err != nil {
			return err
		}
[...]

PR #4914, opened 2020-06-26
[Go Program Gen] Improved handling for pulumi.Map types

EvanBoyle:
Fixes https://github.com/pulumi/pulumi/issues/4903

Also went ahead and refreshed the local aws schema. When doing this, I noticed that the `version` field is no longer populated, and as a result my code was panicking on dereference of `p.Version.Major`. I manually added the version field into the AWS schema, but wanted to double check that this field is populated at runtime @pgavlin

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ CHANGELOG
 
 ## HEAD (Unreleased)
 
+- Go program gen: Improved handling for pulumi.Map types
+  [#491](https://github.com/pulumi/pulumi/pull/4914)
+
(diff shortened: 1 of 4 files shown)


EvanBoyle:
Tags look good in the generated diff across providers. Here's another AWS specific bug that was called out in the issue now generating the proper `pulumi.BoolMap` type:

https://github.com/pulumi-bot/pulumi-aws/blob/950017ade1bf2e7ff58b7244c0f5a806cb86b96b/sdk/go/aws/apigateway/method.go#L97

PR #4925, opened 2020-06-29
[codegen/*] Add support for resource options.

### LINKED ISSUE #4923: [codegen/*] Support basic resource options.
The PCL binder currently supports a subset of resource options, but none of the code generators emit appropriate code.
#1635 in particular needs the parent and provider resource options.

pgavlin:
The PCL binder has supported resource options for some time, but these
options haven't been used or processed by the various code generators.
These options--particularly the parent and provider options0--are
critical for import codegen. These changes implement the basic set of
options, and add a note about fleshing out the rest as necessary.

One component of these changes is a new rewriter that rewrites property
references into property paths that are understood by the Pulumi engine.
This rewriter is used to preprocess the contents of the `ignoreChanges`
resource option.

Fixes #4923.

### DIFF
--- a/pkg/codegen/dotnet/gen_program.go
+++ b/pkg/codegen/dotnet/gen_program.go
@@ -354,7 +354,7 @@ func (g *generator) genResourceOptions(opts *hcl2.ResourceOptions) string {
 	var result bytes.Buffer
 	appendOption := func(name string, value model.Expression) {
 		if result.Len() == 0 {
-			_, err := fmt.Fprint(&result, ", new CustomResourceOptions {")
+			_, err := fmt.Fprintf(&result, ", new CustomResourceOptions\n%s{", g.Indent)
 			g.Indent += "    "
 			contract.IgnoreError(err)
 		}
--- a/pkg/codegen/dotnet/gen_program.go
+++ b/pkg/codegen/dotnet/gen_program.go
@@ -346,6 +346,47 @@ func (g *generator) makeResourceName(baseName, count string) string {
 	return fmt.Sprintf("$\"%s-{%s}\"", baseName, count)
 }
 
+func (g *generator) genResourceOptions(opts *hcl2.ResourceOptions) string {
+	if opts == nil {
+		return ""
+	}
+
+	var result bytes.Buffer
+	appendOption := func(name string, value model.Expression) {
+		if result.Len() == 0 {
+			_, err := fmt.Fprint(&result, ", new CustomResourceOptions {")
+			g.Indent += "    "
+			contract.IgnoreError(err)
+		}
+
+		g.Fgenf(&result, "\n%s%s = %v,", g.Indent, name, g.lowerExpression(value, value.Type()))
+	}
+
+	if opts.Parent != nil {
+		appendOption("Parent", opts.Parent)
+	}
+	if opts.Provider != nil {
+		appendOption("Provider", opts.Provider)
+	}
+	if opts.DependsOn != nil {
+		appendOption("DependsOn", opts.DependsOn)
+	}
+	if opts.Protect != nil {
+		appendOption("Protect", opts.Protect)
+	}
+	if opts.IgnoreChanges != nil {
+		appendOption("IgnoreChanges", opts.IgnoreChanges)
+	}
+
+	if result.Len() != 0 {
+		g.Indent = g.Indent[:len(g.Indent)-4]
+		_, err := fmt.Fprintf(&result, "\n%s}", g.Indent)
+		contract.IgnoreError(err)
+	}
+
+	return result.String()
+}
+
 // genResource handles the generation of instantiations of non-builtin resources.
 func (g *generator) genResource(w io.Writer, r *hcl2.Resource) {
 	qualifiedMemberName := g.resourceTypeName(r)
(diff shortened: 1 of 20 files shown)


pkg/codegen/go/gen_program.go
```diff
@@ -238,11 +243,70 @@ func (g *generator) genNode(w io.Writer, n hcl2.Node) {
...
+	}
+
+	if opts.Parent != nil {
+		appendOption("Parent", opts.Parent, model.DynamicType)
+	}
+	if opts.Provider != nil {
+		appendOption("Provider", opts.Provider, model.DynamicType)
+	}
+	if opts.DependsOn != nil {
+		appendOption("DependsOn", opts.DependsOn, model.NewListType(model.DynamicType))
```
EvanBoyle, line 278: This dynamic type isn't sufficient to generate correct code. You'll probably need to provide some additional metadata to generate `pulumi.Resource[]{...}`

pkg/codegen/internal/test/testdata/aws-fargate.pp.go
```diff
@@ -166,7 +166,9 @@ func main() {
 					ContainerPort:  pulumi.Int(80),
 				},
 			},
-		})
+		}, pulumi.DependsOn([]string{
```
EvanBoyle, line 169: This should be `[]pulumi.Resource`

pkg/codegen/go/gen_program.go
pgavlin: Yep.

pkg/codegen/internal/test/testdata/aws-fargate.pp.go
pgavlin: That's right, yes. I was surprised to see this defaulting to `[]string` instead of `[]interface{}` given that I was passing in `list(dynamic)`.

EvanBoyle: Yeah, just looks like a bug. Try this patch:

```diff
diff --git a/pkg/codegen/go/gen_program_expressions.go b/pkg/codegen/go/gen_program_expressions.go
index a3a61c9b..486d6fe4 100644
--- a/pkg/codegen/go/gen_program_expressions.go
+++ b/pkg/codegen/go/gen_program_expressions.go
@@ -622,7 +622,7 @@ func (g *generator) argumentTypeName(expr model.Expression, destType model.Type,
 				return "pulumi.Float64"
 			}
 			return "float64"
-		case model.StringType, model.DynamicType:
+		case model.StringType:
 			if isInput {
 				return "pulumi.String"
 			}
@@ -632,6 +632,11 @@ func (g *generator) argumentTypeName(expr model.Expression, destType model.Type,
 				return "pulumi.Bool"
 			}
 			return "bool"
+		case model.DynamicType:
+			if isInput {
+				return "pulumi.Any"
+			}
+			return "interface{}"
 		default:
 			return destType.Name
 		}

```

pgavlin: No worries--we can take care of that in a follow-up change.

pkg/codegen/internal/test/testdata/resource-options.pp.py
```diff
@@ -0,0 +1,12 @@
+import pulumi
+import pulumi_aws as aws
+import pulumi_pulumi as pulumi
+
+provider = pulumi.providers.Aws("provider", region="us-west-2")
+bucket1 = aws.s3.Bucket("bucket1", opts=ResourceOptions(provider=provider,
+    dependsOn=[provider],
```
komalali, line 7: in python the resource options are snake_cased, i.e. "depends_on" and "ignore_changes"

pkg/codegen/python/gen_program.go
```diff
@@ -216,6 +216,72 @@ func (g *generator) makeResourceName(baseName, count string) string {
...
+		prefix = "\n" + g.Indent
+	}
+	g.Fprintf(w, ",%sopts=ResourceOptions(", prefix)
+	g.Indented(func() {
+		for i, item := range block.Body.Items {
+			if i > 0 {
+				g.Fprintf(w, ",\n%s", g.Indent)
+			}
+			attr := item.(*model.Attribute)
+			g.Fgenf(w, "%s=%v", attr.Name, attr.Value)
```
komalali, line 279:
```suggestion
			g.Fgenf(w, "%s=%v", PyName(attr.Name), attr.Value)
```

pkg/codegen/internal/test/testdata/resource-options.pp.py
pgavlin: Agh, good catch!

pgavlin: Fixed.

pkg/codegen/dotnet/gen_program.go
```diff
@@ -346,6 +346,47 @@ func (g *generator) makeResourceName(baseName, count string) string {
...
 
+func (g *generator) genResourceOptions(opts *hcl2.ResourceOptions) string {
+	if opts == nil {
+		return ""
+	}
+
+	var result bytes.Buffer
+	appendOption := func(name string, value model.Expression) {
+		if result.Len() == 0 {
+			_, err := fmt.Fprint(&result, ", new CustomResourceOptions {")
```
mikhailshilkov, line 357: We tend to put `{` on a new line in the generated code

pgavlin: Good point; I'll fix this up.

PR #4927, opened 2020-06-30
[codegen/*] Add support for explicit secrets.

### LINKED ISSUE #4924: [codegen/*] Support explicit secret values.
PCL doesn't have any way to mark specific values as secret. We should add support for this via a new builtin function, `secret`, that turns an input of type `T` into a value of type `output(T)` that is marked as secret.

This is related to #1635.

pgavlin:
- Add a new builtin function, secret, that marks its input value as
  secret
- Add support for this function to the various code generators

Note that the Go code generation isn't really correct, as it is missing
a type assertion.

Fixes #4924.

### DIFF
--- a/pkg/codegen/go/gen_program_expressions.go
+++ b/pkg/codegen/go/gen_program_expressions.go
@@ -232,8 +232,11 @@ func (g *generator) GenFunctionCallExpression(w io.Writer, expr *model.FunctionC
 	case "readDir":
 		contract.Failf("unlowered toJSON function expression @ %v", expr.SyntaxNode().Range())
 	case "secret":
-		// TODO: generate an appropriate type assertion
-		g.Fgenf(w, "pulumi.ToSecret(%v)", expr.Args[0])
+		outputTypeName := "pulumi.Any"
+		if model.ResolveOutputs(expr.Type()) != model.DynamicType {
+			outputTypeName = g.argumentTypeName(nil, expr.Type(), false)
+		}
+		g.Fgenf(w, "pulumi.ToSecret(%v).(%sOutput)", expr.Args[0], outputTypeName)
 	case "split":
 		g.genNYI(w, "call %v", expr.Name)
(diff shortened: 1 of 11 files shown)


pkg/codegen/go/gen_program_expressions.go
```diff
@@ -229,6 +229,9 @@ func (g *generator) GenFunctionCallExpression(w io.Writer, expr *model.FunctionC
 		g.genNYI(w, "ReadFile")
 	case "readDir":
 		contract.Failf("unlowered toJSON function expression @ %v", expr.SyntaxNode().Range())
+	case "secret":
+		// TODO: generate an appropriate type assertion
```
pgavlin, line 233: I'll follow an issue to track fixing this.

EvanBoyle: Should be a pretty simple change if you apply that patch I mentioned earlier:

```diff
diff --git a/pkg/codegen/go/gen_program_expressions.go b/pkg/codegen/go/gen_program_expressions.go
index a3a61c9b..486d6fe4 100644
--- a/pkg/codegen/go/gen_program_expressions.go
+++ b/pkg/codegen/go/gen_program_expressions.go
@@ -622,7 +622,7 @@ func (g *generator) argumentTypeName(expr model.Expression, destType model.Type,
 				return "pulumi.Float64"
 			}
 			return "float64"
-		case model.StringType, model.DynamicType:
+		case model.StringType:
 			if isInput {
 				return "pulumi.String"
 			}
@@ -632,6 +632,11 @@ func (g *generator) argumentTypeName(expr model.Expression, destType model.Type,
 				return "pulumi.Bool"
 			}
 			return "bool"
+		case model.DynamicType:
+			if isInput {
+				return "pulumi.Any"
+			}
+			return "interface{}"
 		default:
 			return destType.Name
 		}

```

and then call `argumentTypeName` here.

pgavlin: Done; thanks!

PR #4947, opened 2020-07-02
add plugin download url metadata to package spec

jaxxstorm:
This optional field allows providers and plugins to specify a URL to download a provider plugin from. This will allow community plugins to use automatic plugin acquisition

Currently this is just a noop, the same changes need to be passed through to `pulumi-terraform-bridge` as well.


jaxxstorm:
@stack72 yes it'll only work for those languages. I have a successful test working with the rke provider which i'll send over!

Issue #4931, opened 2020-06-29
_submodules causes IntelliSense to be confused


sindre-nistad:
Hi,
In version 3.11.0, I am unable to get IntelliSense (autocomplete in IntelliJ's PyCharm)  from the `plum_azure` module.
This worked fine in version 3.9.2.

The problem seem related to how `plum_azure/__init__.py` is generated; the module names are listed in `_submodules`, but changing its name to `__all__`, the IntelliSense works as expected.

In Python, `__all__` is a [special variable](https://docs.python.org/3/tutorial/modules.html#importing-from-a-package) in `__init__.py` files.

Is it possible to rename `_submodules` to `__all__` again?

sindre-nistad:
> Can you tell me how you're importing pulumi_azure here? Is it with `import pulumi_azure as azure`?

Yes

komalali:
I dug into this a little further and looked at some popular open-source python packages like django and pandas to see if and how they are using the __all__ keyword. From what I could find, there isn't consistent use of __all__. Additionally, the problem isn't so much that we aren't using __all__ here, it's that IntelliJ/PyCharm doesn't parse the `importlib.import_module()` function correctly, so even though the submodules are correctly imported into the namespace (can be checked by `dir(pulumi_azure)`), they don't show up in autocomplete.

I have a fix locally, will push it up.

sindre-nistad:
Thanks,
Is this something JetBrain's PyCharm team should be made aware of?

komalali:
@sindre-nistad It looks like an issue already exists: https://youtrack.jetbrains.com/issue/PY-38552

PR #4948, opened 2020-07-02
Change how submodules are imported to work with Intellisense

komalali:
IntelliJ/PyCharm is unable to parse the submodules imported using `importlib.import_module()` for autocomplete.
Fixes: https://github.com/pulumi/pulumi/issues/4931

Before:
Screen Shot 2020-07-01 at 9 23 15 PM
Screen Shot 2020-07-01 at 9 34 57 PM

After:
Screen Shot 2020-07-01 at 9 23 58 PM
Screen Shot 2020-07-01 at 9 34 27 PM

### DIFF
--- a/pkg/codegen/python/gen.go
+++ b/pkg/codegen/python/gen.go
@@ -262,10 +262,6 @@ func (mod *modContext) genInit(exports []string) string {
 	w := &bytes.Buffer{}
 	mod.genHeader(w, false, false)
 
-	if mod.submodulesExist() {
-		fmt.Fprintf(w, "import importlib\n")
-	}
-
 	// Import anything to export flatly that is a direct export rather than sub-module.
 	if len(exports) > 0 {
 		sort.Slice(exports, func(i, j int) bool {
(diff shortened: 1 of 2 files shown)


pkg/codegen/python/gen.go
```diff
@@ -301,15 +297,11 @@ func (mod *modContext) genInit(exports []string) string {
 					child = child[match[2]:match[3]]
 				}
 			}
-			if i > 0 {
-				fmt.Fprintf(w, ",\n")
+			if child != "config" {
+				fmt.Fprintf(w, "    %s,\n", PyName(child))
```
justinvp, line 301: @pgavlin, do you know why we skip config? Avoiding side-effects or something else?

pgavlin: I don't know why we skip this, no. I believe that this pattern was present in the old `tfgen` code generator and I copied it here.

justinvp: I did some spelunking and the description in https://github.com/pulumi/pulumi-terraform/pull/356 explains why:

```python
for pkg in __all__:
    if pkg != 'config': # The "config" package is not importable without an engine
        importlib.import_module(f"{__name__}.{pkg}")
```

> The "config" package is not importable without an engine

We may actually be able to do this now given more recent changes to decouple the engine for testing/mocking, but I have not tried it. @komalali, do you want to see if it works without any other changes?

My main feedback is: It'd be nice if we could import `config` too so it also shows up in intellisense, but if it requires more refactoring to be able to work, we could continue to exclude it for now and come back to it later. Unless it's an easy enough change to make now.

komalali: Ooh thanks for digging into that. I'll try that.

komalali: This appears to "just work" without any issues 🤞 🎉 . Tested this in the pulumi-azure provider.

PR #4958, opened 2020-07-02
[codegen/python] pep-8 compliant spacing

komalali:
Making the linter happy by adding adequate newlines between class definitions and functions

### DIFF
--- a/pkg/codegen/python/gen.go
+++ b/pkg/codegen/python/gen.go
@@ -347,7 +347,8 @@ func (mod *modContext) genAwaitableType(w io.Writer, obj *schema.ObjectType) str
 	baseName := pyClassName(tokenToName(obj.Token))
 
 	// Produce a class definition with optional """ comment.
-	fmt.Fprintf(w, "\nclass %s:\n", baseName)
+	fmt.Fprint(w, "\n")
+	fmt.Fprintf(w, "class %s:\n", baseName)
 	printComment(w, obj.Comment, "    ")
 
 	// Now generate an initializer with properties for all inputs.
@@ -378,7 +379,8 @@ func (mod *modContext) genAwaitableType(w io.Writer, obj *schema.ObjectType) str
 	awaitableName := "Awaitable" + baseName
 
 	// Produce an awaitable subclass.
-	fmt.Fprintf(w, "\n\nclass %s(%s):\n", awaitableName, baseName)
+	fmt.Fprint(w, "\n\n")
+	fmt.Fprintf(w, "class %s(%s):\n", awaitableName, baseName)
 
 	// Emit __await__ and __iter__ in order to make this type awaitable.
 	//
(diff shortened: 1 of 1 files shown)


pkg/codegen/python/gen.go
```diff
@@ -347,7 +347,7 @@ func (mod *modContext) genAwaitableType(w io.Writer, obj *schema.ObjectType) str
 	baseName := pyClassName(tokenToName(obj.Token))
 
 	// Produce a class definition with optional """ comment.
-	fmt.Fprintf(w, "class %s:\n", baseName)
+	fmt.Fprintf(w, "\nclass %s:\n", baseName)
```
justinvp, line 350: Nit: Instead of printing the `\n` in front of `class`, what do you think of having a single `fmt.Fprintf(w, "\n")` call before the call to `mod.genAwaitableType` on line 686 below?

pkg/codegen/python/gen.go
```diff
@@ -378,7 +378,7 @@ func (mod *modContext) genAwaitableType(w io.Writer, obj *schema.ObjectType) str
 	awaitableName := "Awaitable" + baseName
 
 	// Produce an awaitable subclass.
-	fmt.Fprintf(w, "class %s(%s):\n", awaitableName, baseName)
+	fmt.Fprintf(w, "\n\nclass %s(%s):\n", awaitableName, baseName)
```
justinvp, line 381: Nit: Instead of the `\n\n` before `class` what do you think of doing something like the following?

```go
}
fmt.Fprintf(w, "\n\n")

awaitableName := "Awaitable" + baseName

// Produce an awaitable subclass.
fmt.Fprintf(w, "class %s(%s):\n", awaitableName, baseName)
```

pkg/codegen/python/gen.go
```diff
@@ -693,7 +693,7 @@ func (mod *modContext) genFunction(fun *schema.Function) (string, error) {
 	}
 
 	// Write out the function signature.
-	fmt.Fprintf(w, "def %s(", name)
+	fmt.Fprintf(w, "\ndef %s(", name)
```
justinvp, line 696: Nit: Similarly, maybe a single `fmt.Fprintf(w, "\n")` before this line instead?

pkg/codegen/python/gen.go
```diff
@@ -723,7 +723,7 @@ func (mod *modContext) genFunction(fun *schema.Function) (string, error) {
 	}
 
 	// Copy the function arguments into a dictionary.
-	fmt.Fprintf(w, "    __args__ = dict()\n\n")
+	fmt.Fprintf(w, "    __args__ = dict()\n")
 	fmt.Fprintf(w, "\n")
```
justinvp, line 727: Any idea why we have this extra line before setting the values in `__args___`? Wondering if we could get rid of this too.

komalali: Yeah there's no apparent reason for it. Will remove.

pkg/codegen/python/gen.go
komalali: Sure. I've made similar changes previously so I'm going to fix them to also follow this pattern.

pkg/codegen/python/gen.go
komalali: I agree with moving `fmt.Fprintf(w, "\n\n")` to a separate line but think it should be under the `// Produce an awaitable subclass.` comment since I'd consider the newlines part of that subclass definition and they wouldn't be there without it.

PR #4959, opened 2020-07-02
[codegen/python] Add underscores to private modules to avoid exporting them.

komalali:
Related: https://github.com/pulumi/pulumi/pull/4948

The tables and utilities modules are only used internally, so adding underscores so they don't show up in consumers' Intellisense.

Before:
Screen Shot 2020-07-01 at 9 34 27 PM

After:
Screen Shot 2020-07-02 at 2 24 04 PM

### DIFF
--- a/pkg/codegen/python/gen.go
+++ b/pkg/codegen/python/gen.go
@@ -128,7 +128,7 @@ func (mod *modContext) genHeader(w io.Writer, needsSDK bool, needsJSON bool) {
 		fmt.Fprintf(w, "import pulumi\n")
 		fmt.Fprintf(w, "import pulumi.runtime\n")
 		fmt.Fprintf(w, "from typing import Union\n")
-		fmt.Fprintf(w, "from %s import utilities, tables\n", relImport)
+		fmt.Fprintf(w, "from %s import _utilities, _tables\n", relImport)
 		fmt.Fprintf(w, "\n")
 	}
 }
@@ -184,7 +184,7 @@ func (mod *modContext) gen(fs fs) error {
 		buffer := &bytes.Buffer{}
 		mod.genHeader(buffer, false, false)
 		fmt.Fprintf(buffer, "%s", utilitiesFile)
-		fs.add(filepath.Join(dir, "utilities.py"), buffer.Bytes())
+		fs.add(filepath.Join(dir, "_utilities.py"), buffer.Bytes())
 
 		// Ensure that the top-level (provider) module directory contains a README.md file.
 		readme := mod.pkg.Language["python"].(PackageInfo).Readme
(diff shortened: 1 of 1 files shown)


justinvp:
How does this look in vscode, with and without Pylance?

> We should also do this for the config module.

Do what? I think we want people to be able to discover and use `config` as it's a nice convenience getting things like `config.region`. How can we make it discoverable?

komalali:
@justinvp oh interesting. that makes sense. I was going off of the fact that it is explicitly not imported into the top-level package. Why is that? https://github.com/pulumi/pulumi/blob/master/pkg/codegen/python/gen.go#L310-L312

komalali:
> How does this look in vscode, with and without Pylance?

Generally the same. The underscored objects show up all the way at the bottom of the list whether or not `__all__` is used. It's slightly better with Pylance than without but also somewhat difficult to judge because VSCode has some code complete caching.

Screen Shot 2020-07-02 at 5 04 57 PM
Screen Shot 2020-07-02 at 5 05 05 PM

justinvp:
I wonder if there is a way to hide (or move to bottom) the lowercase/snake_case `blob`, `container`, `customer_managed_key`, etc. modules from VS Code intellisense, since we already make the classes available from `storage`. Should we consider making those modules private (prefixed with underscore)?

Though, I suppose we can't as that would be a breaking change for anyone referring to those.

Do we care about breaking anyone who might be using `tables` or `utilities` (not sure why anyone would be)? Personally, I lean towards being OK breaking anyone using `tables` or `utilities`. I doubt it would actually affect anyone in practice.

komalali:
> Though, I suppose we can't as that would be a breaking change for anyone referring to those.

Perhaps something we can consider for 3.0

> Do we care about breaking anyone who might be using tables or utilities (not sure why anyone would be)?

I agree with being okay with breaking these as they were never meant to be exposed and I doubt anyone is using them in practice... and the sooner we do this the less likely it is that someone will use it!

justinvp:
> > Do we care about breaking anyone who might be using tables or utilities (not sure why anyone would be)?
> 
> I agree with being okay with breaking these as they were never meant to be exposed and I doubt anyone is using them in practice... and the sooner we do this the less likely it is that someone will use it!

@leezen, @lukehoban, are you ok renaming these "internal" modules? Technically it's a breaking change, but I wouldn't expect it to actually affect anyone in practice as these aren't intended to be used by consumers directly. If we wanted to wait to do this, and follow semver, we'd have to wait to do it in major versions of each provider. Or not do it at all. Or have some kind of flag and only do it for new providers.

komalali:
@justinvp 
> Should we do something to delete those existing files when this change is adopted in the providers

Yes... but what? I would think this is something that would have to be handled manually when the providers are next generated.

komalali:
> since we don't delete files before regenerating

Relatedly, I really believe we should be doing this.

leezen:
I think this is generally fine unless somehow someone might use these in testing? I can't think of a scenario, so I think it makes sense.

justinvp:
> Yes... but what? I would think this is something that would have to be handled manually when the providers are next generated.

I was wondering if we could come up with a cheap way to automate this so @stack72 doesn't have to remember to manually delete `table.py` and `utilities.py` when updating providers that adopt this change that will start producing `_table.py` and `_utilities.py`. Like maybe a temporary change to a common script that deletes the old files if the underscore versions are present? But I suppose it'd be best to focus on https://github.com/pulumi/home/issues/976 rather than spend any effort on this one-off, and leave it to @stack72 to remember or automate himself.

komalali:
Merging this since we have https://github.com/pulumi/pulumi/pull/5018 in the works

PR #4967, opened 2020-07-03
Enable SourceLink with embedded PDB

### LINKED ISSUE #4962: Enable SourceLink and embed PDB symbols for all NuGet packages
[SourceLink](https://github.com/dotnet/sourcelink) is essential and would make it easy to step _into_ Pulumi C# source supplied as NuGet packages.

Happy to submit a PR if someone can point me to how / where these packages are created at build-time.

gitfool:
Fixes #4962.

### DIFF
--- a/.github/workflows/windows-pr.yml
+++ b/.github/workflows/windows-pr.yml
@@ -10,7 +10,7 @@ jobs:
         go-version: [1.14.x]
         node-version: [10.x]
         python-version: [3.7]
-        dotnet: ['3.1.100']
+        dotnet: ['3.1.301']
     runs-on: ${{ matrix.platform }}
     env:
       GOPATH: ${{ github.workspace }}
--- a/.github/workflows/windows-build.yml
+++ b/.github/workflows/windows-build.yml
@@ -14,7 +14,7 @@ jobs:
         go-version: [1.14.x]
         node-version: [10.x]
         python-version: [3.7]
-        dotnet: ['3.1.100']
+        dotnet: ['3.1.301']
     runs-on: ${{ matrix.platform }}
     env:
       GOPATH: ${{ github.workspace }}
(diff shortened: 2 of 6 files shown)


.github/workflows/windows-pr.yml
```diff
@@ -10,7 +10,7 @@ jobs:
         go-version: [1.14.x]
         node-version: [10.x]
         python-version: [3.7]
-        dotnet: ['3.1.100']
+        dotnet: ['3.1.301']
```
mikhailshilkov, line 13: @stack72 @jaxxstorm any objections against this change?

gitfool:
@mikhailshilkov I don't have a full development environment yet, pending the vs code dev container, so I didn't do a full build, just enough to verify the two `nupkg` artifacts had source link symbols.

image

I also bumped the .NET SDK to current latest since it helps with the build order when embedding untracked sources.

.github/workflows/windows-pr.yml
stack72: Not from me - if you are happy with the new version being correct then we can roll that out everywhere

gitfool: See my comment above for why I bumped the .NET SDK.

mikhailshilkov: @stack72 Which versions are we using "everywhere" now? Could you link to the place where it's configured?

gitfool: https://github.com/dotnet/sourcelink/issues/572#issuecomment-583548551

stack72: In all of the providers (from a private repo) the version we have configured is 

```
dotnetversion: ['3.1.201'],
```

gitfool: @mikhailshilkov I just did a manual test and I can confirm I could step debug into Pulumi source from my program.

gitfool: image

Issue #1583, opened 2018-06-29
Gesture to copy configuration from one stack to another


ellismg:
This is a quality of life improvement.

Pulumi makes it very easy to spin up new stacks, which is great.  However, when you create a new stack, you must configure it "from scratch".  The tempting thing to do is to copy an existing `Pulumi.<stack-name>.yaml` file and make the minor edits.  After all, it's just a file right?

Unfortunately, if you are using any encrypted configuration values, things quickly go off the rails.  Since configuration values are encrypted with a per stack key, simply copying the file will not do the right thing.  So, at a minimum, you need to reset all of the secret configuration values, which is a chore.

We should consider a gesture like:

`pulumi config copy [source-stack-name] [destination-stack-name]` which copies all of the configuration values from the source stack to the destination stack.  We could, choose to further augment this with something like a  `--value` flag which would let you specify one or more values to copy.

In the case where we would overwrite values in the destination stack, we could  issue a warning and prompt the user to answer yes or no.

kimdhamilton:
I will use this feature all the time

leezen:
Related: #481

PR #4971, opened 2020-07-06
Add the ability to copy configs between stacks

stack72:
Fixes: #1583

Please note, I specifically tried to keep this code simple. As part of the work to change secrets provider, we will need to refactor some of this code around the building of a list of config to encrypt with a new provider

### DIFF
--- a/pkg/cmd/pulumi/config.go
+++ b/pkg/cmd/pulumi/config.go
@@ -106,7 +106,7 @@ func newConfigCopyCmd(stack *string) *cobra.Command {
 				return err
 			}
 			if currentStack.Ref().Name().String() == destinationStackName {
-				return errors.New("config must be copied to a different destination stack than currently using")
+				return errors.New("current stack and destination stack are the same")
 			}
 			currentProjectStack, err := loadProjectStack(currentStack)
 			if err != nil {
@@ -138,7 +138,7 @@ func newConfigCopyCmd(stack *string) *cobra.Command {
 		&path, "path", false,
 		"The key contains a path to a property in a map or list to set")
 	cpCommand.PersistentFlags().StringVarP(
-		&destinationStackName, "new-stack", "", "",
+		&destinationStackName, "dest", "d", "",
 		"The name of the new stack to copy the config to")
 
 	return cpCommand
@@ -155,7 +155,8 @@ func copySingleConfigKey(configKey string, path bool, currentStack backend.Stack
 
 	v, ok, err := currentProjectStack.Config.Get(key, path)
 	if err != nil {
-		return err
+		return errors.Errorf(
+			"configuration key '%s' not found for stack '%s'", prettyKey(key), currentStack.Ref())
 	}
 	if ok {
 		if v.Secure() {
@@ -167,28 +168,17 @@ func copySingleConfigKey(configKey string, path bool, currentStack backend.Stack
 			decrypter = config.NewPanicCrypter()
 		}
 
-		raw, err := v.Value(decrypter)
+		encrypter, cerr := getStackEncrypter(destinationStack)
+		if cerr != nil {
+			return cerr
+		}
+
+		val, err := v.Copy(decrypter, encrypter)
 		if err != nil {
-			return errors.Wrap(err, "could not decrypt configuration value")
+			return err
 		}
 
-		// now we start the work to copy it - if it was encrypted before then we need to do the same again
-		if v.Secure() {
-			// it was encrypted before, now we need to re-encrypt
-			c, cerr := getStackEncrypter(destinationStack)
-			if cerr != nil {
-				return cerr
-			}
-			enc, eerr := c.EncryptValue(raw)
-			if eerr != nil {
-				return eerr
-			}
-			v = config.NewSecureValue(enc)
-		} else {
-			v = config.NewValue(raw)
-		}
-
-		err = destinationProjectStack.Config.Set(key, v, path)
+		err = destinationProjectStack.Config.Set(key, val, path)
 		if err != nil {
 			return err
 		}
@@ -225,17 +215,24 @@ func copyEntireConfigMap(currentStack backend.Stack,
 		return err
 	}
 
+	var requiresSaving bool
 	for key, val := range newProjectConfig {
 		err = destinationProjectStack.Config.Set(key, val, false)
 		if err != nil {
 			return err
 		}
+		requiresSaving = true
+	}
 
-		err = saveProjectStack(destinationStack, destinationProjectStack)
+	// The use of `requiresSaving` here ensures that there was actually some config
+	// that needed saved, otherwise it's an unnecessary save call
+	if requiresSaving {
+		err := saveProjectStack(destinationStack, destinationProjectStack)
 		if err != nil {
 			return err
 		}
 	}
+
 	return nil
 }
 
--- a/pkg/cmd/pulumi/config.go
+++ b/pkg/cmd/pulumi/config.go
@@ -80,8 +80,163 @@ func newConfigCmd() *cobra.Command {
 	cmd.AddCommand(newConfigRmCmd(&stack))
 	cmd.AddCommand(newConfigSetCmd(&stack))
 	cmd.AddCommand(newConfigRefreshCmd(&stack))
+	cmd.AddCommand(newConfigCopyCmd(&stack))
 
 	return cmd
+}
+
+func newConfigCopyCmd(stack *string) *cobra.Command {
+	var path bool
+	var destinationStackName string
+
+	cpCommand := &cobra.Command{
+		Use:   "cp [key]",
+		Short: "Copy config to another stack",
+		Long: "Copies the config from the current stack to the destination stack. If `key` is omitted,\n" +
+			"then all of the config from the current stack will be copied to the destination stack.",
+		Args: cmdutil.MaximumNArgs(1),
+		Run: cmdutil.RunFunc(func(cmd *cobra.Command, args []string) error {
+			opts := display.Options{
+				Color: cmdutil.GetGlobalColorization(),
+			}
+
+			// Get current stack and ensure that it is a different stack to the destination stack
+			currentStack, err := requireStack(*stack, false, opts, true /*setCurrent*/)
+			if err != nil {
+				return err
+			}
+			if currentStack.Ref().Name().String() == destinationStackName {
+				return errors.New("config must be copied to a different destination stack than currently using")
+			}
+			currentProjectStack, err := loadProjectStack(currentStack)
+			if err != nil {
+				return err
+			}
+
+			// Get the destination stack
+			destinationStack, err := requireStack(destinationStackName, false, opts, false /*setCurrent*/)
+			if err != nil {
+				return err
+			}
+			destinationProjectStack, err := loadProjectStack(destinationStack)
+			if err != nil {
+				return err
+			}
+
+			// Do we need to copy a single value or the entire map
+			if len(args) > 0 {
+				// A single key was specified so we only need to copy that specific value
+				return copySingleConfigKey(args[0], path, currentStack, currentProjectStack, destinationStack,
+					destinationProjectStack)
+			}
+
+			return copyEntireConfigMap(currentStack, currentProjectStack, destinationStack, destinationProjectStack)
+		}),
+	}
+
+	cpCommand.PersistentFlags().BoolVar(
+		&path, "path", false,
+		"The key contains a path to a property in a map or list to set")
+	cpCommand.PersistentFlags().StringVarP(
+		&destinationStackName, "new-stack", "", "",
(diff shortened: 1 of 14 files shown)


pkg/cmd/pulumi/config.go
```diff
@@ -80,10 +80,188 @@ func newConfigCmd() *cobra.Command {
...
+
+func newConfigCopyCmd(stack *string) *cobra.Command {
+	var path bool
+	var newStackName string
+
+	cpCommand := &cobra.Command{
+		Use:   "cp <key>",
+		Short: "Copy config to another stack",
+		Long: "Copies the config from the current stack to the newly specified stack. If a key is omitted,\n" +
+			"then all of the config, of the current stack, will copied to the new stack.",
```
jaxxstorm, line 103:
```suggestion
			"then all of the config of the current stack will be copied to the new stack.",
```

pkg/cmd/pulumi/config.go
```diff
@@ -80,10 +80,188 @@ func newConfigCmd() *cobra.Command {
...
+				Color: cmdutil.GetGlobalColorization(),
+			}
+
+			s, err := requireStack(*stack, false, opts, true /*setCurrent*/)
+			if err != nil {
+				return err
+			}
+
+			if s.Ref().Name().String() == newStackName {
+				return errors.New("config must be copied to a different stack than currently using")
```
jaxxstorm, line 116:
```suggestion
				return errors.New("source and destination stacks must be different when copying config")
```

pkg/cmd/pulumi/config.go
```diff
@@ -80,10 +80,188 @@ func newConfigCmd() *cobra.Command {
...
+
+			return nil
+		}),
+	}
+
+	cpCommand.PersistentFlags().BoolVar(
+		&path, "path", false,
+		"The key contains a path to a property in a map or list to set")
+	cpCommand.PersistentFlags().StringVarP(
+		&newStackName, "new-stack", "", "",
```
justinvp, line 159: Should this be `dest-stack` or `destination-stack`? If so, should we also provide a shorthand? Maybe `d` for `-d`?

pkg/cmd/pulumi/config.go
```diff
@@ -80,10 +80,188 @@ func newConfigCmd() *cobra.Command {
...
+				return saveProjectStack(ns, newStack)
+			}
+
+			return nil
+		}),
+	}
+
+	cpCommand.PersistentFlags().BoolVar(
+		&path, "path", false,
+		"The key contains a path to a property in a map or list to set")
```
justinvp, line 157: Does this even work correctly when `--path` is specified? I wonder if we should just not accept `--path` for this command so that only top-level keys can be copied.

pkg/cmd/pulumi/config.go
```diff
@@ -80,10 +80,188 @@ func newConfigCmd() *cobra.Command {
...
+					return nil, errors.Wrap(err, "could not create a decrypter")
+				}
+			} else {
+				decrypter = config.NewPanicCrypter()
+			}
+			raw, err := v.Value(decrypter)
+			if err != nil {
+				return nil, errors.Wrap(err, "could not decrypt configuration value")
+			}
+			configToCopy.value = raw
```
justinvp, line 198: Does this preserve values that are objects (i.e. when `v.Object()` is true), or will such values end up as JSON strings in the destination stack config? I'd have to checkout these changes to see, but ideally we would preserve these so the destination YAML looks as expected.

justinvp: Just tried this and it doesn't work the way I'd expect.

```
$ pulumi stack init dev
$ pulumi stack init dev2
$ pulumi stack select dev
$ pulumi config set aws:region us-west-2
$ pulumi config set --path data.active true
$ pulumi config set --path data.nums[0] 1
$ pulumi config set --path data.nums[1] 2
$ pulumi config set --path data.nums[2] 3
$ cat Pulumi.dev.yaml
config:
  aws:region: us-west-2
  cpcfgtest:data:
    active: true
    nums:
    - 1
    - 2
    - 3
$ pulumi config cp --new-stack dev2
$ cat Pulumi.dev2.yaml
config:
  aws:region: us-west-2
  cpcfgtest:data: '{"active":true,"nums":[1,2,3]}'
```

I would expect `Pulumi.dev2.yaml` to look like `Pulumi.dev.yaml`:

```
config:
  aws:region: us-west-2
  cpcfgtest:data:
    active: true
    nums:
    - 1
    - 2
    - 3
```

pkg/cmd/pulumi/config.go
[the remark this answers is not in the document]
justinvp: I just tried this and when I don't specify `new-stack` the command runs but appears to do nothing AFAICT. When I specify a random value for `--new-stack` (e.g. `pulumi config cp --new-stack foo`), it runs but doesn't do anything AFAICT.

stack72: Way nicer!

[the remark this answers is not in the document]
stack72: This will now return an error

pkg/cmd/pulumi/config.go
```diff
@@ -80,10 +80,165 @@ func newConfigCmd() *cobra.Command {
...
+	for key, val := range newProjectConfig {
+		err = destinationProjectStack.Config.Set(key, val, false)
+		if err != nil {
+			return err
+		}
+
+		err = saveProjectStack(destinationStack, destinationProjectStack)
+		if err != nil {
+			return err
+		}
```
justinvp, line 237: Any reason why we're saving on each iteration of the loop? Should we save once after the loop?

sdk/go/common/resource/config/value.go
```diff
@@ -240,6 +240,53 @@ func isSecureValue(v interface{}) (bool, string) {
 	return false, ""
 }
 
+func encryptObject(v interface{}, decrypter Decrypter, encrypter Encrypter) (interface{}, error) {
```
justinvp, line 243: Consider renaming this to `reencryptObject` since it's doing more than just _encrypting_ -- it's _decrypting_ and the _re-encrypting_.

```suggestion
func reencryptObject(v interface{}, decrypter Decrypter, encrypter Encrypter) (interface{}, error) {
```

sdk/go/common/resource/config/value.go
```diff
@@ -240,6 +240,53 @@ func isSecureValue(v interface{}) (bool, string) {
...
+	encryptIt := func(val interface{}) (interface{}, error) {
```
justinvp, line 244: And similarly:

```suggestion
	reencryptIt := func(val interface{}) (interface{}, error) {
```

pkg/cmd/pulumi/config.go
```diff
@@ -80,10 +80,165 @@ func newConfigCmd() *cobra.Command {
...
+				return cerr
+			}
+			enc, eerr := c.EncryptValue(raw)
+			if eerr != nil {
+				return eerr
+			}
+			v = config.NewSecureValue(enc)
+		} else {
+			v = config.NewValue(raw)
+		}
```
justinvp, line 189: This doesn't appear to be handling objects correctly.

For example:

```
$ pulumi stack init dev
$ pulumi stack init dev2
$ pulumi stack select dev
$ pulumi config set aws:region us-west-2
$ pulumi config set --path data.active true
$ pulumi config set --path data.nums[0] 1
$ pulumi config set --path data.nums[1] 2
$ pulumi config set --path data.nums[2] 3
$ pulumi config set --path data.pass --secret foo
$ cat Pulumi.dev.yaml
config:
  aws:region: us-west-2
  stackcopytestage:data:
    active: true
    nums:
    - 1
    - 2
    - 3
    pass:
      secure: AAABANIHq5zNQDqx8G/SHeR1aehVjWKUsv9pIMrXG/z9C00=
$ pulumi config cp data --new-stack dev2
$ cat Pulumi.dev2.yaml
config:
  stackcopytestage:data:
    secure: AAABAFFV4QQd9AxeUeC4wSxZ6ST6zYIR0wIr+pmLBQg7Uxi0NYrwV/SZ/rnClAcMOvhqAof0DKbrtHfP6kACT79gPUF7VHhySU8c
```

I would expect `Pulumi.dev2.yaml` to look like `Pulumi.dev.yaml`, just with a different encrypted value for `pass`:

```
config:
  aws:region: us-west-2
  stackcopytestage:data:
    active: true
    nums:
    - 1
    - 2
    - 3
    pass:
      secure: <some other encrypted string>
```

Right now, it's encrypting the whole config value, not the individual `pass` property inside the object, because it's not taking `if c.Object() {` into account.

I wonder if it'd be easier to add a low-level `Copy(decrypter Decrypter, encrypter Encrypter)` to `Value` in `sdk/go/common/resource/config/value.go` that handles this, and then use `Value.Copy` both here and inside `Map.Copy`.

pkg/cmd/pulumi/config.go
```diff
@@ -80,10 +80,165 @@ func newConfigCmd() *cobra.Command {
...
+
+			return copyEntireConfigMap(currentStack, currentProjectStack, destinationStack, destinationProjectStack)
+		}),
+	}
+
+	cpCommand.PersistentFlags().BoolVar(
+		&path, "path", false,
+		"The key contains a path to a property in a map or list to set")
+	cpCommand.PersistentFlags().StringVarP(
+		&destinationStackName, "new-stack", "", "",
```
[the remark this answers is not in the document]
stack72: I like dest better :)

pkg/cmd/pulumi/config.go
```diff
@@ -80,10 +80,165 @@ func newConfigCmd() *cobra.Command {
...
+	key, err := parseConfigKey(configKey)
+	if err != nil {
+		return errors.Wrap(err, "invalid configuration key")
+	}
+
+	v, ok, err := currentProjectStack.Config.Get(key, path)
+	if err != nil {
+		return err
+	}
+	if ok {
```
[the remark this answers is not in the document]
stack72: 👍

pkg/cmd/pulumi/config.go
stack72: This was to ensure that if an error occurred then at least we will have saved some of the config - I guess it's more efficient to do it once - changing now

sdk/go/common/resource/config/value.go
stack72: 👍

sdk/go/common/resource/config/value.go
stack72: 👍

pkg/cmd/pulumi/config.go
stack72: Fixed and working as expected now!

stack72:
Thanks for the review @justinvp - just working on the tests right now - this will be ready again for review soon

pkg/cmd/pulumi/config.go
```diff
@@ -80,10 +80,162 @@ func newConfigCmd() *cobra.Command {
...
+	if err != nil {
+		return errors.Wrap(err, "invalid configuration key")
+	}
+
+	v, ok, err := currentProjectStack.Config.Get(key, path)
+	if err != nil {
+		return errors.Errorf(
+			"configuration key '%s' not found for stack '%s'", prettyKey(key), currentStack.Ref())
+	}
+	if ok {
```
justinvp, line 161: `config.Map.Get` can fail for more reasons that the key not being found (like the path is invalid, etc.).

How about doing something more like the following?

```go
	if err != nil {
		return err
	}
	if !ok {
		return errors.Errorf(
			"configuration key '%s' not found for stack '%s'", prettyKey(key), currentStack.Ref())
	}

	...
```

Or:

```go
	if err != nil {
		return err
	}
	if ok {
		...
	}

	return errors.Errorf(
		"configuration key '%s' not found for stack '%s'", prettyKey(key), currentStack.Ref())
```

sdk/go/common/resource/config/crypt.go
```diff
@@ -208,3 +208,21 @@ func decryptAES256GCM(ciphertext []byte, key []byte, nonce []byte) (string, erro
 
 	return string(msg), err
 }
+
+// Crypter that just adds a prefix to the plaintext string when encrypting,
+// and removes the prefix from the ciphertext when decrypting.
```
justinvp, line 213:
```suggestion
// and removes the prefix from the ciphertext when decrypting, for use in tests.
```

stack72: 👍

pkg/cmd/pulumi/config.go
stack72: Going with the second to keep things as stylistically similar as other code :)

stack72:
Make the suggested changes and will merge on a successful build

PR #4974, opened 2020-07-07
[hcl2] Fix tokenless body item printing.

pgavlin:
In general, each item in an HCL2 body must be followed by a trailing
newline. The printer did not properly insert these newlines for body
items without any associated tokens/trivia, or with trivia that did not
include a trailing new line.

Related to #1635.

### DIFF
--- a/pkg/codegen/hcl2/model/print_test.go
+++ b/pkg/codegen/hcl2/model/print_test.go
@@ -1,0 +1,26 @@
+package model
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/zclconf/go-cty/cty"
+)
+
+func TestPrintNoTokens(t *testing.T) {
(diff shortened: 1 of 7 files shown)


pkg/codegen/hcl2/model/print_test.go
```diff
@@ -0,0 +1,26 @@
...
+				&Attribute{
+					Name: "attribute",
+					Value: &LiteralValueExpression{
+						Value: cty.True,
+					},
+				},
+			},
+		},
+	}
+	expected := "block {\n    attribute = true\n}"
```
EvanBoyle, line 24: Should this have a trailing newline after the closing `}`?

pgavlin: No, this is okay--it's the responsibility of the body printer to print the trailing newline, and this is just printing a block. If we were printing a body, this would get a trailing newline.

Issue #4979, opened 2020-07-08
[sdk/python] - Remove @known_types decorators


komalali:
Related: https://github.com/pulumi/pulumi/issues/4953

These decorators prevent Intellisense from working in VSCode.

PR #4980, opened 2020-07-08
Remove known_types decorators

komalali:
Fixes: https://github.com/pulumi/pulumi/issues/4979

The `@known_types` decorators prevent Intellisense from functioning correctly in VSCode. We were using them as a workaround to avoid circular references between imported packages. We can instead use a combination of [forward references](https://www.python.org/dev/peps/pep-0484/#forward-references) (which we were already doing to some extent) and deferring the imports to when they are needed (which we were also already doing in some cases).

There is follow-up work to standardize docstrings across the SDK which I'll do in a separate PR.

Before:
Screen Shot 2020-07-01 at 3 24 17 PM

After:
Screen Shot 2020-07-08 at 9 51 26 AM

### DIFF
--- a/sdk/python/lib/pulumi/resource.py
+++ b/sdk/python/lib/pulumi/resource.py
@@ -17,6 +17,7 @@ from typing import Optional, List, Any, Mapping, Union, Callable, TYPE_CHECKING,
 
 import copy
 
+from .runtime import known_types
 from .runtime.resource import _register_resource, register_resource_outputs, _read_resource
 from .runtime.settings import get_root_resource
 
@@ -881,9 +882,8 @@ def export(name: str, value: Any):
     :param str name: The name to assign to this output.
     :param Any value: The value of this output.
     """
-    from .runtime.stack import Stack
     res = cast('Stack', get_root_resource())
-    if isinstance(res, Stack):
+    if known_types.is_stack(res):
         res.output(name, value)
     else:
         raise Exception("Failed to export output. Root resource is not an instance of 'Stack'")
--- a/sdk/python/lib/pulumi/resource.py
+++ b/sdk/python/lib/pulumi/resource.py
@@ -25,6 +25,7 @@ from .metadata import get_project, get_stack
 
 if TYPE_CHECKING:
     from .output import Input, Inputs, Output
+    from .runtime.stack import Stack
 
 
 class CustomTimeouts:
@@ -76,7 +77,7 @@ def inherited_child_alias(
 #   * parentAliasName: "app"
 #   * aliasName: "app-function"
 #   * childAlias: "urn:pulumi:stackname::projectname::aws:s3/bucket:Bucket::app-function"
-    from . import Output
+    from . import Output  # pylint: disable=import-outside-toplevel
     alias_name = Output.from_input(child_name)
     if child_name.startswith(parent_name):
         alias_name = Output.from_input(parent_alias).apply(
@@ -159,7 +160,7 @@ class Alias:
     """
 
     # Ignoring type errors associated with the ellipsis constant being assigned to a string value.
-    # We use it as a internal sentinal value, and don't need to expose this in the user facing type system.
+    # We use it as a internal sentinel value, and don't need to expose this in the user facing type system.
     # https://docs.python.org/3/library/constants.html#Ellipsis
     def __init__(self,
                  name: Optional[str] = ..., # type: ignore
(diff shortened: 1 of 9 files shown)


sdk/python/lib/pulumi/runtime/known_types.py
```diff
@@ -12,294 +12,114 @@
...
+"""The type of Stack."""
 
 _output_type: Optional[type] = None
-"""The type of Output. Filled-in as the Pulumi package is initializing."""
+"""The type of Output."""
 
 _unknown_type: Optional[type] = None
-"""The type of unknown. Filled-in as the Pulumi package is initializing."""
-
-def asset(class_obj: type) -> type:
```
justinvp, line 74: Alternatively, adding stronger typing to these may have helped, but I haven't tried it, so maybe not...

```python
if TYPE_CHECKING:
     from .. import Asset

def asset(class_obj: Type['Asset']) -> Type['Asset']:
```

But, I'm in favor of getting rid of the `known_types` decorators, if we can.

sdk/python/lib/pulumi/runtime/known_types.py
```diff
@@ -12,294 +12,114 @@
...
-The implementation of this mechanism is that, for every known type, that type
-is stashed away in a global variable. Whenever the runtime wants to do a type
-test using that type (or instantiate an instance of this type), it uses the
-functions defined in this module to do so.
+This implementation allows for overriding types used as stubs for testing
+(see test/test_next_serialize.py)
 """
 from typing import Any, Optional
 
 _custom_resource_type: Optional[type] = None
```
justinvp, line 32: Any reason why we need to keep these global variables around?

I know `_custom_resource_type` is referenced by one of the tests:
 https://github.com/pulumi/pulumi/blob/dd104a00a76fec807fb5b50e11ceca37508757e1/sdk/python/lib/test/test_next_serialize.py#L50-L54

sdk/python/lib/pulumi/runtime/known_types.py
komalali: I tried this but it doesn't work all the way. It does help with the type-hinting (it'll show that you're missing a required argument) but it doesn't show the docstrings.
  
Screen Shot 2020-07-08 at 11 42 08 AM

And yeah I think the `known_types` decorators are an interesting but non-standard workaround for this particular issue so worth removing for that reason as well.

sdk/python/lib/pulumi/runtime/known_types.py
komalali: @justinvp yeah that's the only reason that I kept them around (see the comment on line 27-28). I couldn't get one of the tests to work without it (test_custom_resource failed with a ValueError because it doesn't expect `FakeCustomResource`).
Is that test particularly useful? Is it worth keeping only the global for `CustomResource` and not the others?

sdk/python/lib/pulumi/resource.py
```diff
@@ -206,9 +204,10 @@ def collapse_alias_to_urn_worker(inner: Union[Alias, str]) -> 'Output[str]':
 
         return create_urn(name, type_, parent, project, stack)
 
-    inputAlias: Output[Union[Alias, str]] = Output.from_input(alias)
+    inputAlias: 'Output[Union[Alias, str]]' = Output.from_input(alias)
```
pgavlin, line 207: Why does this need to be quoted? Haven't we imported `Output` by this point?

pgavlin: (same throughout)

komalali: ah yup totally

sdk/python/lib/pulumi/runtime/known_types.py
justinvp: > Is it worth keeping only the global for `CustomResource` and not the others?

I don't think there's any reason to keep the others, so let's get rid of them. And then as a follow-up, see if we can rework the test to get rid of the `CustomResource` one.

sdk/python/lib/pulumi/runtime/known_types.py
```diff
@@ -12,294 +12,78 @@
...
-    return _output_type(*args) # type: ignore
-
-
-def new_unknown(*args: Any) -> Any:
-    """
-    Instantiates a new Unknown, passing the given arguments to the constructor.
-    """
-    return _unknown_type(*args) # type: ignore
+# We override this global in test/test_next_serialize.py to stub the CustomResource
+# type. TODO: Rework the test to remove the need for this global.
```
justinvp, line 31: Mind opening an issue and adding a link to the issue here?

komalali:
https://github.com/pulumi/pulumi/issues/5000 <- to remove the last known_type

PR #4985, opened 2020-07-08
Remove duplicated methods

komalali:
These functions are throwing a duplicate method error in my IDE because they are already specified in the `printable` interface.


komalali:
@pgavlin ugh you're right. so functionally there's no difference at all? why state it one way or the other?

pgavlin:
> ugh you're right. so functionally there's no difference at all? why state it one way or the other?

That's right. I think we should take your changes, as the language change in 1.14 was primarily intended to allow embedding multiple interfaces from different dependencies that declare identical methods.

komalali:
> the language change in 1.14 was primarily intended to allow embedding multiple interfaces from different dependencies that declare identical methods

that's helpful context thank you!

Issue #4986, opened 2020-07-09
[PANIC] Pulumi v2.6.0 on destroy panics


stack72:
On upgrading from v2.5.0 to v2.6.0 of Pulumi, the CLI panics when trying to destroy any resources:

```
▶ pulumi destroy
Previewing destroy (dev):
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x4c245bb]

goroutine 25 [running]:
github.com/pulumi/pulumi/pkg/v2/backend/display.(*resourceRowData).ColorizedColumns(0xc0009460e0, 0xc0012c5650, 0x0, 0x0)
	/Users/travis/gopath/src/github.com/pulumi/pulumi/pkg/backend/display/rows.go:289 +0x6b
github.com/pulumi/pulumi/pkg/v2/backend/display.(*ProgressDisplay).getOrCreateTreeNode(0xc0000f8f20, 0xc0012c5638, 0x0, 0x0, 0x5740920, 0xc0009460e0, 0xc0012c5650, 0x0)
	/Users/travis/gopath/src/github.com/pulumi/pulumi/pkg/backend/display/progress.go:425 +0x96
github.com/pulumi/pulumi/pkg/v2/backend/display.(*ProgressDisplay).getOrCreateTreeNode(0xc0000f8f20, 0xc0012c5638, 0xc0009100f0, 0x49, 0x5740920, 0xc000946000, 0xc0012c5650, 0x1)
	/Users/travis/gopath/src/github.com/pulumi/pulumi/pkg/backend/display/progress.go:450 +0x407
github.com/pulumi/pulumi/pkg/v2/backend/display.(*ProgressDisplay).generateTreeNodes(0xc0000f8f20, 0xc000921200, 0x10, 0xc00092e420)
	/Users/travis/gopath/src/github.com/pulumi/pulumi/pkg/backend/display/progress.go:469 +0x21f
github.com/pulumi/pulumi/pkg/v2/backend/display.(*ProgressDisplay).refreshAllRowsIfInTerminal(0xc0000f8f20)
	/Users/travis/gopath/src/github.com/pulumi/pulumi/pkg/backend/display/progress.go:580 +0x6f
github.com/pulumi/pulumi/pkg/v2/backend/display.(*ProgressDisplay).processNormalEvent(0xc0000f8f20, 0x536c85c, 0xc, 0x51f91c0, 0xc0009dd980)
[...]

PR #4987, opened 2020-07-09
Revert "Narrow a few interfaces."

stack72:
Reverts pulumi/pulumi#4934

Fixes: #4986


stack72:
It was easier to revert the PR for now so we can investigate the problems with it - we should ensure this gets rolled forward again after v2.6.0 release has been fixed

PR #4997, opened 2020-07-13
Make args optional in Node.js if state inputs aren't defined

mikhailshilkov:
**Problem**: if a native provider defines no state inputs, Node.js generation is broken - the generated code passes `undefined` for `args` where undefined is not expected.

Kubernetes introduces a fix for this behind a compat flag, but I believe this should be generalized. This line:

https://github.com/pulumi/pulumi/blob/736019f7ce63d35266a846a8fe13a70baafa7006/pkg/codegen/nodejs/gen.go#L395

assigns `undefined`, so we should use the same condition to allow `undefined`.


pgavlin:
This is addressed by https://github.com/pulumi/pulumi/pull/4991

mikhailshilkov:
Damn, how did I miss that one! Thanks!

PR #5003, opened 2020-07-14
Protect against engine event mutation.

pgavlin:
Certain operations in `engine/diff` mutate engine events during display.
This mutation can occur concurrently with the serialization of the event
for persistence, which causes a panic in the CLI. These changes fix the
offending code and add code that copies each engine event before
persisteing it in order to guard against future issues.

### DIFF
--- a/pkg/engine/diff.go
+++ b/pkg/engine/diff.go
@@ -232,25 +232,32 @@ func PrintObject(
 	}
 }
 
-func massageStackPreviewAdd(p resource.PropertyValue) {
+func massageStackPreviewAdd(p resource.PropertyValue) resource.PropertyValue {
 	switch {
 	case p.IsArray():
-		for _, v := range p.ArrayValue() {
-			massageStackPreviewAdd(v)
+		arr := make([]resource.PropertyValue, len(p.ArrayValue()))
+		for i, v := range p.ArrayValue() {
+			arr[i] = massageStackPreviewAdd(v)
 		}
+		return resource.NewArrayProperty(arr)
 	case p.IsObject():
-		delete(p.ObjectValue(), "@isPulumiResource")
-		for _, v := range p.ObjectValue() {
-			massageStackPreviewAdd(v)
+		obj := resource.PropertyMap{}
+		for k, v := range p.ObjectValue() {
+			if k != "@isPulumiResource" {
+				obj[k] = massageStackPreviewAdd(v)
+			}
 		}
+		return resource.NewObjectProperty(obj)
+	default:
+		return p
 	}
 }
 
 func massageStackPreviewDiff(diff resource.ValueDiff, inResource bool) {
 	switch {
 	case diff.Array != nil:
-		for _, p := range diff.Array.Adds {
-			massageStackPreviewAdd(p)
+		for i, p := range diff.Array.Adds {
+			diff.Array.Adds[i] = massageStackPreviewAdd(p)
 		}
 		for _, d := range diff.Array.Updates {
 			massageStackPreviewDiff(d, inResource)
@@ -278,8 +285,8 @@ func massageStackPreviewOutputDiff(diff *resource.ObjectDiff, inResource bool) {
 		}
 	}
 
-	for _, p := range diff.Adds {
-		massageStackPreviewAdd(p)
+	for i, p := range diff.Adds {
+		diff.Adds[i] = massageStackPreviewAdd(p)
 	}
 	for k, d := range diff.Updates {
 		if isResource && d.New.IsComputed() && !shouldPrintPropertyValue(d.Old, false) {
--- a/pkg/engine/events.go
+++ b/pkg/engine/events.go
@@ -28,6 +28,7 @@ import (
 	"github.com/pulumi/pulumi/sdk/v2/go/common/resource/plugin"
 	"github.com/pulumi/pulumi/sdk/v2/go/common/tokens"
 	"github.com/pulumi/pulumi/sdk/v2/go/common/util/contract"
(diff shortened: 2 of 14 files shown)


pkg/engine/diff.go
```diff
@@ -232,25 +232,32 @@ func PrintObject(
 	}
 }
 
-func massageStackPreviewAdd(p resource.PropertyValue) {
+func massageStackPreviewAdd(p resource.PropertyValue) resource.PropertyValue {
```
chrsmith, line 235: Nit: It might not be consistent with the other functions in the file, but if you have the context of what this function does right now, would you mind adding a comment on it? TBH I have no idea what `messageStackPreviewAdd` is doing, and what it it is supposed to return. (Does it return the same input? Is it returning a copy of its input? etc.)

Just a suggestion to make it easier for the next person who needs to update this code.

pkg/engine/diff.go
```diff
@@ -232,25 +232,32 @@ func PrintObject(
...
+			arr[i] = massageStackPreviewAdd(v)
 		}
+		return resource.NewArrayProperty(arr)
 	case p.IsObject():
-		delete(p.ObjectValue(), "@isPulumiResource")
-		for _, v := range p.ObjectValue() {
-			massageStackPreviewAdd(v)
+		obj := resource.PropertyMap{}
+		for k, v := range p.ObjectValue() {
+			if k != "@isPulumiResource" {
```
chrsmith, line 246: Nit: Another comment as for why we ignore `"@isPulumiResource"` would be helpful here. Since it isn't a constant referenced from another package, I assume this is only hard-coded in a couple of places?

sdk/go/common/util/deepcopy/copy.go
```diff
@@ -0,0 +1,88 @@
...
+		if elem.CanAddr() {
+			return elem.Addr()
+		}
+		rv := reflect.New(typ.Elem())
+		rv.Set(elem)
+		return rv
+	case reflect.Array:
+		rv := reflect.New(typ).Elem()
+		for i := 0; i < v.Len(); i++ {
+			rv.Index(i).Set(copy(v.Index(i)))
```
chrsmith, line 55: Would you mind writing some unit tests for these `copy` functions? It shouldn't be too cumbersome to create a prototypical "table driven" Golang unit test.

The reason I ask is that I'm not sure if this is correct. (And scanning [the docs](https://golang.org/pkg/reflect/#Value.Index), it isn't super clear.)

So we are creating a new "array type" and then copying every element over individually, great. But does the type information (`typ`) also include the length of the array, too? Or do we need to pre-allocate the length of `rv`, similar to line 61 for the slice-case?

pgavlin:
> Do you know what functions are mutating those engine events and why? This seems like a super-nasty bug waiting to happen.

I only know of the diff functions that I changed. It's definitely a super-nasty bug waiting to happen, hence the addition of the copier.

> In other words, if the mutation is localized to just a certain method in the display package, can we also localize the EE copying to that place too? Since otherwise, we'll inevitably hit the same issue if we add another consumer to the EE stream.
> 
> What do you think? I don't have any problem with us making the deep copy in httpstate, but I just want to make sure I understand what the expected contract is here.

I think that redesigning the `Event` type is probably the best fix: if we change its public API to something like this:

```go
type Event struct {
    typ EventType
    payload interface{}
}

func (e Event) Type() EventType {
    return e.typ
}

func (e Event) Payload() interface{} {
    return deepcopy.Copy(e.payload)
}
```

Then we can nip this in the bud (at some potential performance cost).

chrsmith:
> I think that redesigning the Event type is probably the best fix.

Yeah, @pgavlin strong +1 to making that change to the type itself. (And hopefully updating all the callers won't be too tedious to insert `()` or assign `Payload` to a local variable here and there.

If it does lead to a performance problem, it might be reasonable to add `func (e Event) PayloadNoCopyISwearIWontMutateAnythingUnderPenaltyOfDeath() interface{}`.

pkg/engine/events.go
```diff
@@ -28,14 +28,46 @@ import (
...
+	case ResourceOutputsEvent:
+		_, ok = payload.(ResourceOutputsEventPayload)
+	case ResourceOperationFailed:
+		_, ok = payload.(ResourceOperationFailedPayload)
+	case PolicyViolationEvent:
+		_, ok = payload.(PolicyViolationEventPayload)
+	default:
+		contract.Failf("unknown event type %v", typ)
+	}
+	contract.Assertf(ok, "invalid payload of type %T for event type %v", payload, typ)
```
chrsmith, line 66: This is pretty slick. TBH I wouldn't have considered adding the check for the type of the payload too. 👍

sdk/go/common/util/deepcopy/copy.go
```diff
@@ -0,0 +1,88 @@
+package deepcopy
```
chrsmith, line 1: Nit: We probably need to add a copyright header to the top of the file to be consistent with other files in the source tree, right? Same with the new `_test.go` file, too.

sdk/go/common/util/deepcopy/copy_test.go
```diff
@@ -0,0 +1,60 @@
...
+		[2]byte{42, 24},
+		[]byte{0, 1, 2, 3},
+		[]string{"foo", "bar"},
+		map[string]int{
+			"a": 42,
+			"b": 24,
+		},
+		struct {
+			Foo int
+			Bar map[int]int
```
chrsmith, line 38: Do any of our engine events have opaque `interface{}` values inside of them? Do you think that would be a reasonable corner case to try and add a test scenario for?

PR #5012, opened 2020-07-14
[codegen/*] Add join and toBase64 functions.

pgavlin:
- `join` joins an array of strings using a specified separator
- `toBase64` converts a string to its Base64 representation

These changes also contain fixes for invokes with inputs that are absent
or optional.

### DIFF
--- a/pkg/codegen/go/gen_program.go
+++ b/pkg/codegen/go/gen_program.go
@@ -419,8 +419,7 @@ func (g *generator) genResource(w io.Writer, r *hcl2.Resource) {
 
 	resName := makeValidIdentifier(r.Name())
 	pkg, mod, typ, _ := r.DecomposeToken()
-	mod = strings.ToLower(strings.Replace(mod, "/", ".", -1))
-	if mod == "" {
+	if mod == "" || strings.HasPrefix(mod, "/") || strings.HasPrefix(mod, "index/") {
 		mod = pkg
 	}
 
--- a/pkg/codegen/go/gen_program.go
+++ b/pkg/codegen/go/gen_program.go
@@ -605,7 +605,7 @@ func (g *generator) genLocalVariable(w io.Writer, v *hcl2.LocalVariable) {
 			g.Fgenf(w, "if err != nil {\n")
 			g.Fgenf(w, "return err\n")
 			g.Fgenf(w, "}\n")
-		default:
+		case "join", "toBase64", "mimeType", "fileAsset":
 			g.Fgenf(w, "%s := %.3v;\n", name, expr)
 		}
 	default:
--- a/pkg/codegen/go/gen_program.go
+++ b/pkg/codegen/go/gen_program.go
@@ -419,7 +419,8 @@ func (g *generator) genResource(w io.Writer, r *hcl2.Resource) {
 
 	resName := makeValidIdentifier(r.Name())
 	pkg, mod, typ, _ := r.DecomposeToken()
-	if mod == "" || strings.HasPrefix(mod, "/") || strings.HasPrefix(mod, "index/") {
+	mod = strings.ToLower(strings.Replace(mod, "/", ".", -1))
+	if mod == "" {
 		mod = pkg
 	}
 
--- a/pkg/codegen/go/gen_program.go
+++ b/pkg/codegen/go/gen_program.go
@@ -605,6 +605,8 @@ func (g *generator) genLocalVariable(w io.Writer, v *hcl2.LocalVariable) {
 			g.Fgenf(w, "if err != nil {\n")
 			g.Fgenf(w, "return err\n")
 			g.Fgenf(w, "}\n")
+		default:
+			g.Fgenf(w, "%s := %.3v;\n", name, expr)
 		}
 	default:
 		g.Fgenf(w, "%s := %.3v;\n", name, expr)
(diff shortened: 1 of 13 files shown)


komalali:
Brought this back from the abyss and added a test. The approvals are super old so I'm re-requesting.

komalali:
@pgavlin since this was your PR I can't request your review but I think it'd be good to have your eyes on it to make sure nothing looks off from the test output.

pkg/codegen/go/gen_program.go
```diff
@@ -605,6 +605,8 @@ func (g *generator) genLocalVariable(w io.Writer, v *hcl2.LocalVariable) {
 			g.Fgenf(w, "if err != nil {\n")
 			g.Fgenf(w, "return err\n")
 			g.Fgenf(w, "}\n")
+		default:
```
komalali, line 608: I had to add this. Without it the non-invoke function calls never make it into the output go file

EvanBoyle (approved):

pkg/codegen/internal/test/testdata/functions.pp.ts
```diff
@@ -0,0 +1,7 @@
+import * as pulumi from "@pulumi/pulumi";
+
+const encoded = (new Buffer("haha business")).toString("base64");
```
justinvp, line 3: Is this the Node.js `Buffer`? Apparently, the constructor is deprecated in favor of factory APIs like `Buffer.from()`.

https://github.com/nodejs/node/blob/eaca13ddc9092e831b9b45180c6e0ee1f3292e5e/lib/buffer.js#L258-L267

```suggestion
const encoded = Buffer.from("haha business").toString("base64");
```

pkg/codegen/go/gen_program.go
komalali: 🤔 we are losing a number of go examples and I'm wondering if it's because of this change causing invalid go code. perhaps I should narrow this down.

pkg/codegen/internal/test/testdata/functions.pp.cs
```diff
@@ -0,0 +1,15 @@
+using Pulumi;
+
+class MyStack : Stack
+{
+    public MyStack()
+    {
+        var encoded = Convert.ToBase64String(%v)("haha business");
```
justinvp, line 7: This doesn't look right.


1. The `System` namespace needs to be imported for `Convert` or it needs to be fully qualified like `System.Convert`.
2. The `%v` doesn't look right.
3. We can't just pass a string to `Convert.ToBase64String`. The string needs to be converted to `[]byte` or `ReadOnlySpan<byte>`. How are we encoding the string to bytes? UTF-8 given the other languages?

I think we want this to look something like:

```suggestion
        var encoded = Convert.ToBase64String(new System.Text.UTF8Encoding(false).GetBytes("haha business"));
```

The above assumes a `using System;` is added at the top.

Note, unfortunately the `System.Text.Encoding.UTF8` instance emits a byte order mark, so that's why we have to create a specific instance of `UTF8Encoding` and pass `false` for `encoderShouldEmitUTF8Identifier`.

komalali: Thank you Justin!

justinvp: Actually, I misspoke about the byte order mark. I updated my comment above.

The default `Encoding.UTF8` instance will emit a byte order mark when used with something like `System.IO.StreamWriter`, where the encoding's `GetPreamble()` is called.

If you're just calling `GetBytes` on the encoding, it won't include the byte order mark.

So we can simply do:

```c#
var encoded = Convert.ToBase64String(System.Text.Encoding.UTF8.GetBytes("haha business"));
```

justinvp: And just tried it out to confirm:

```c#
using System;

class Program
{
    static void Main()
    {
        Console.WriteLine(Convert.ToBase64String(new System.Text.UTF8Encoding(false).GetBytes("haha business")));
        Console.WriteLine(Convert.ToBase64String(System.Text.Encoding.UTF8.GetBytes("haha business")));
    }
}
```

Have the same output:

```
$ dotnet run
aGFoYSBidXNpbmVzcw==
aGFoYSBidXNpbmVzcw==
```

pkg/codegen/internal/test/testdata/functions.pp.cs
```diff
@@ -0,0 +1,16 @@
...
+class MyStack : Stack
+{
+    public MyStack()
+    {
+        var encoded = Convert.ToBase64String(System.Text.UTF8.GetBytes("haha business"));
+        var joined = string.Join("-", 
+        {
+            "haha",
+            "business",
+        });
```
justinvp, line 13: One last thing for posterity... I don't it's related to this change specifically, but this doesn't compile.

It'd need to be something like:

```suggestion
        var joined = string.Join("-", new[]
        {
            "haha",
            "business",
        });
```

Or:

```suggestion
        var joined = string.Join("-", new string[]
        {
            "haha",
            "business",
        });
```

Or, since the last argument is a params array, in this specific case it could be:

```suggestion
        var joined = string.Join("-", "haha", "business");
```

justinvp (approved):

t0yv0:
Taking a rain check on reviewing this, need to work with codegen a bit more beforehand 😓

PR #5013, opened 2020-07-15
[codegen/hcl2] Fix nested apply rewrites

pgavlin:
Rewrites that should produce nested applies due to functions that return
eventual types were instead producing a single top-level apply. These
changes fix that by considering a function that produces an eventual
value as inspecting eventual values.


pgavlin:
> Just curious, what is causing you to discover these edge cases? Kicking the tires on import?

Trying out some CFN-related experiments in my spare time.

PR #5015, opened 2020-07-15
Speed up .NET programs with large number of resources

### LINKED ISSUE #5001: Preview/Up Slowness
I have a large stack that always seems to take a long time to process up or even just a preview. I have tried targeting just two resources and I still get the slowness. From the CLI I see if process all the items which show the CRUD operations but it seems to hang/wait for a long time before finishing processing.

The stack I am working on is "mitchell.maler/k8s-cluster-projects/asc-stg-k8s-c01".
Pulumi Version: v2.6.1
Using Pulumi Cloud for stack state.

@lukehoban I have talked to you on slack about this and thought it would be best to document this in an issue for better visibility.

## Example 1
Here I ran an up that targets two URNs which are to be created which took around 8m
```
$ pulumi up \
  --target "urn:pulumi:..." \
  --target "urn:pulumi:..." \
  --skip-preview --tracing=file:./up.trace
Updating (stack_name):
     Type                          Name                   Status      
     pulumi:pulumi:Stack           name                                                           
 +   ├─ kubernetes:core:Secret     secret-name            created     
 +   └─ kubernetes:core:ConfigMap  config-name            created     
Resources:
    + 2 created
    1697 unchanged
Duration: 7m58s
```

## Example 2
I also have ran a preview on the whole stack with `--logtostderr -v=9`. I seem hundreds (maybe thousands) of the evensink.go then a long hang before the langruntime_plugin.go:163 which results in about a 7 min wait. This is most likely my issue and this was only a preview.

```
I0709 20:53:46.756266   23872 eventsink.go:59] Registering resource monitor end: t=vault:kubernetes/authBackendRole:AuthBackendRole, name=omitted, custom=True
I0709 20:53:46.756293   23872 eventsink.go:62] eventSink::Debug(<{%reset%}>Registering resource monitor end: t=vault:kubernetes/authBackendRole:AuthBackendRole, name=omitted, custom=True<{%reset%}>)
....................
I0709 21:00:34.799332   23872 langruntime_plugin.go:163] langhost[dotnet].RunPlan(pwd=/Users/mitch.maler/omitted.csproj,...,dryrun=true) success: progerr=
I0709 21:00:34.801075   23872 plan_executor.go:219] planExecutor.Execute(...): incoming event (nil? true, <nil>)
Resources:
    + 61 to create
    ~ 1 to update
    62 changes. 1698 unchanged
________________________________________________________
Executed in  692.09 secs   fish           external 
   usr time  1236.14 secs  176.61 millis  1235.96 secs 
   sys time  102.84 secs   22.25 millis  102.81 secs 
```

I sent @lukehoban trace files for both of these in slack but he said they are pretty large so might not be easy to explore.

Please let me know if there is anything else I can post here or DM someone. I am still pulling apart and poking around my stack to see if it is anything on my side.

mikhailshilkov:
The current event awaiting loop in the .NET SDK is ~O(N^2), which is definitely noticeable with ~1k resources. Also, we log a lot of excessive debug information, which makes a larger multiplier to N.

This PR speeds up the await loop to ~O(N) and removes the excessive debug output.

Resolves #5001 

cc @CyrusNajmabadi if you are around, in case I miss some original intent of that loop.

### DIFF
--- a/sdk/dotnet/Pulumi/Deployment/Deployment.Runner.cs
+++ b/sdk/dotnet/Pulumi/Deployment/Deployment.Runner.cs
@@ -119,7 +119,9 @@ namespace Pulumi
                             // Log the descriptions of completed tasks.
                             var descriptions = _inFlightTasks[task];
                             foreach (var description in descriptions)
+                            {
                                 Serilog.Log.Information($"Completed task: {description}");
+                            }
 
                             // Check if all the tasks are completed and signal the completion source if so.
                             if (Interlocked.Decrement(ref remaining) == 0)
--- a/sdk/dotnet/Pulumi/Deployment/Deployment.Runner.cs
+++ b/sdk/dotnet/Pulumi/Deployment/Deployment.Runner.cs
@@ -1,7 +1,8 @@
-﻿// Copyright 2016-2019, Pulumi Corporation
+﻿// Copyright 2016-2020, Pulumi Corporation
 
 using System;
 using System.Collections.Generic;
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace Pulumi
@@ -92,6 +93,7 @@ namespace Pulumi
                     {
                         if (_inFlightTasks.Count == 0)
                         {
+                            // No more tasks in flight: exit the loop.
                             break;
                         }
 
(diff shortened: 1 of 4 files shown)


sdk/dotnet/Pulumi/Deployment/Deployment.Runner.cs
```diff
@@ -92,31 +93,62 @@ private async Task<int> WhileRunningAsync()
...
-                        _inFlightTasks.Remove(task);
+                        try
+                        {
+                            // Wait for the task completion.
+                            await task.ConfigureAwait(false);
+
+                            // Log the descriptions of completed tasks.
+                            var descriptions = _inFlightTasks[task];
+                            foreach (var description in descriptions)
+                                Serilog.Log.Information($"Completed task: {description}");
```
pgavlin, line 122: This needs braces, but do we need this logging at all, or was this left over from debugging?

```suggestion
                            foreach (var description in descriptions)
                            {
                                Serilog.Log.Information($"Completed task: {description}");
                            }
```

mikhailshilkov: Yeah, we have a bunch of places where we don't use braces for one-liners. We could standardize on always-braces, although sometimes it gets lengthy.
I believe all serilog calls are "left over from debugging". They do nothing as we don't initialize Serilog sinks unless this line is uncommented https://github.com/pulumi/pulumi/blob/653dcf8f1fe536916f1dd2b0e294de9173f38461/sdk/dotnet/Pulumi/Deployment/Deployment_Run.cs#L152

pgavlin: Okay, got it. As long as they're latent then I'm fine leaving them in.

Though I agree that always-braces can be lengthy, I do think we should standardize on that style.

PR #5018, opened 2020-07-15
Add a helper function to clean the codegen dir

komalali:
Related to: https://github.com/pulumi/home/issues/976

Had to bump the python version because 3.6.10 was removed from github actions.

Related PR in pulumi-terraform-bridge: https://github.com/pulumi/pulumi-terraform-bridge/pull/236

### DIFF
--- a/pkg/codegen/nodejs/gen.go
+++ b/pkg/codegen/nodejs/gen.go
@@ -1571,6 +1571,10 @@ func GeneratePackage(tool string, pkg *schema.Package, extraFiles map[string][]b
 	return files, nil
 }
 
+func PrepareOutDir(outDir string) error {
+	return codegen.CleanDir(outDir, &[]string{"tests"})
+}
+
 const utilitiesFile = `
 export function getEnv(...vars: string[]): string | undefined {
     for (const v of vars) {
--- a/pkg/codegen/nodejs/gen.go
+++ b/pkg/codegen/nodejs/gen.go
@@ -1571,10 +1571,6 @@ func GeneratePackage(tool string, pkg *schema.Package, extraFiles map[string][]b
 	return files, nil
 }
 
-func PrepareOutDir(outDir string) error {
-	return codegen.CleanDir(outDir, &[]string{"tests"})
-}
-
 const utilitiesFile = `
 export function getEnv(...vars: string[]): string | undefined {
     for (const v of vars) {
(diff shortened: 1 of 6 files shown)


pkg/codegen/utilities.go
```diff
@@ -79,3 +82,31 @@ func SortedKeys(m interface{}) []string {
 
 	return keys
 }
+
+// CleanDir removes all existing files from a directory except those in the exclusions list.
+func CleanDir(dirPath string, exclusions *[]string) error {
```
pgavlin, line 87: Let's just pass in the set itself here. It's safe to pass `nil` to indicate an empty set.

```suggestion
func CleanDir(dirPath string, exclusions codegen.StringSet) error {
```

pkg/codegen/python/gen.go
```diff
@@ -1474,6 +1475,11 @@ func GeneratePackage(tool string, pkg *schema.Package, extraFiles map[string][]b
 	return files, nil
 }
 
+func PrepareOutDir(outDir string, pkgName string) error {
```
pgavlin, line 1478: I think this should also take a set of files we want to retain and then pass that to `CleanDir`. Consumers can then just pass in the overlays. Same for node.

pkg/codegen/nodejs/gen.go
```diff
@@ -1571,6 +1571,10 @@ func GeneratePackage(tool string, pkg *schema.Package, extraFiles map[string][]b
 	return files, nil
 }
 
+func PrepareOutDir(outDir string) error {
+	return codegen.CleanDir(outDir, &[]string{"tests"})
```
pgavlin, line 1575: Not all packages have a `tests` directory--the contents of this set should be left to the caller.

pkg/codegen/utilities.go
```diff
@@ -79,3 +82,24 @@ func SortedKeys(m interface{}) []string {
 
 	return keys
 }
+
+// CleanDir removes all existing files from a directory except those in the exclusions list.
```
pgavlin, line 86: One thing that might warrant a note is that the exclusions don't function recursively. For example, you cannot exclude a single file in a subdirectory--instead, you need to exclude the whole subdirectory.

On that note, is that the behavior we want, or do we need to be able to exclude specific files in subdirectories without excluding the whole directory?

komalali: This is true but it hasn't come up yet. In fact, on the tfgen side of this we actually don't even need to add the overlays to the exclusions set because they are read into memory before the directory is cleaned. 

We can solve for that case within this function if we need to exclude files within subdirectories down the line but for now this is okay? Do you want me to add a comment to clarify any of this?

pgavlin: Let's add a comment for clarification and leave the implementation as-is 👍

PR #3498, opened 2019-11-12
Name Transformation

parisholley:
See https://github.com/pulumi/pulumi/issues/3497

Issue #3497, opened 2019-11-12
Support name transformations


parisholley:
Somewhat of an extension to the discussion in #1518, my setup relies heavy on nesting with `parent` to support multi-region deployments. As an example:

```typescript
class Environment extends ComponentResource {
	constructor(name: string, region: string){
		super('my:environment', prefix, {}, {
			transformations: [args => {
				return {
					name: `${name}:${region}:${args.name}`
				};
			}]
		});
	}
}

class Application extends ComponentResource {
	constructor(name: string){
		super('my:application', name);

		const resource = new Resource("resource", {}, {parent: this});
	}
}

const west = new Environment("production", "us-west-1");
const east = new Environment("production", "us-east-1");

const app_west = new Application("app", {parent: west});
const app_east = new Application("app", {parent: east});
```

By using transformations, my nested resources do not have to worry about name conflicts across regions. I can re-use names safely (eg: "database", "loadbalancer") and the rest just works. Without supporting this, I have to leak naming convention down into my components.

PR #5021, opened 2020-07-17
Name Transformation

parisholley:
Closed #3498 in favor of this 2.0 merged PR.

Supports #3497


parisholley:
@leezen @infin8x i've had this PR effectively open since 2019, when can we get this included?

pgavlin:
@lukehoban @mikhailshilkov can you take another look w.r.t the comments you left on @parisholley's original PR?

sdk/dotnet/Pulumi/PublicAPI.Shipped.txt
```diff
@@ -178,11 +178,9 @@ Pulumi.ResourceTransformationArgs
 Pulumi.ResourceTransformationArgs.Args.get -> Pulumi.ResourceArgs
 Pulumi.ResourceTransformationArgs.Options.get -> Pulumi.ResourceOptions
 Pulumi.ResourceTransformationArgs.Resource.get -> Pulumi.Resource
-Pulumi.ResourceTransformationArgs.ResourceTransformationArgs(Pulumi.Resource resource, Pulumi.ResourceArgs args, Pulumi.ResourceOptions options) -> void
```
mikhailshilkov, line 181: We can't take this breaking change in a minor version. Could we provide a new constructor overload instead?

ESipalis:
Would be nice to get this in, so we don't have to manually prefix parent resource name for every resource.

mikhailshilkov:
As this PR is multiple years old now, I'll have to close it. Kudos to @parisholley for this initial version, too bad we haven't brought it over the line. The PR is referenced from the issue, so once we are ready we may re-apply the changes on top of the current code base and find ways to address the existing feedback.

PR #5033, opened 2020-07-21
Python SDK changes to support input/output classes

justinvp:
SDK changes to support strongly-typed input/output "dataclasses".

PR for provider codegen: https://github.com/pulumi/pulumi/pull/5034

Example diffs:
- AWS: https://github.com/pulumi/pulumi-aws/commit/a77712011c3dc7577a2737b5b3004ee529ea5dc6
- Kubernetes: https://github.com/pulumi/pulumi-kubernetes/commit/ea90ce49b0ef9b760b94be47207eb95fd7f2890b

Fixes https://github.com/pulumi/pulumi/issues/4789
Fixes https://github.com/pulumi/pulumi/issues/5140

### DIFF
--- a/sdk/python/lib/pulumi/_types.py
+++ b/sdk/python/lib/pulumi/_types.py
@@ -491,8 +491,7 @@ def output_type(cls: Type[T]) -> Type[T]:
         python_to_pulumi_table = None
         for python_name, pulumi_name, _ in _py_properties(cls):
             if python_name != pulumi_name:
-                if python_to_pulumi_table is None:
-                    python_to_pulumi_table = {}
+                python_to_pulumi_table = python_to_pulumi_table or {}
                 python_to_pulumi_table[python_name] = pulumi_name
         if python_to_pulumi_table is not None:
             setattr(cls, _PULUMI_PYTHON_TO_PULUMI_TABLE, python_to_pulumi_table)
--- a/sdk/python/lib/pulumi/_types.py
+++ b/sdk/python/lib/pulumi/_types.py
@@ -32,6 +32,28 @@
 #
 # When the resource's outputs are resolved, the `Nested` class is instantiated.
 #
+# The resource could alternatively be declared using a Python property for nested_value rather than
+# using a class annotation:
+#
+#   class FooResource(pulumi.CustomResource):
+#       def __init__(self, resource_name, nested_value: pulumi.InputType[NestedArgs]):
+#           super().__init__("my:module:FooResource", resource_name, {"nestedValue": nested_value})
+#
+#       @property
+#       @pulumi.getter(name="nestedValue")
+#       def nested_value(self) -> pulumi.Output[Nested]:
+#           ...
+#
+#
+# Note the `nested_value` property getter function is empty. The `@pulumi.getter` decorator replaces
+# empty function bodies with an actual implementation. In this case, it replaces it with a body that
+# looks like:
+#
+#       @property
+#       @pulumi.getter(name="nestedValue")
+#       def nested_value(self) -> pulumi.Output[Nested]:
+#           pulumi.get(self, "nestedValue")
+#
 #
 # Here's how the `NestedArgs` input class can be declared:
 #
@@ -130,7 +152,8 @@
 #   @pulumi.output_type
 #   class Nested:
 #       def __init__(self, values: Dict[str, Any]):
-#           self._values = values
+#           for k, v in values.items():
+#               self.__dict__[k] = v
 #
 #       @property
 #       @pulumi.getter(name="firstArg")
@@ -155,7 +178,8 @@
 #   @pulumi.output_type
 #   class Nested:
 #       def __init__(self, values: Dict[str, Any]):
-#           self._values = values
+#           for k, v in values.items():
+#               self.__dict__[k] = v
 #
 #       @property
 #       @pulumi.getter(name="firstArg")
@@ -250,7 +274,6 @@ _PULUMI_NAME = "_pulumi_name"
 _PULUMI_INPUT_TYPE = "_pulumi_input_type"
 _PULUMI_OUTPUT_TYPE = "_pulumi_output_type"
 _TRANSLATE_PROPERTY = "_translate_property"
-_VALUES = "_values"
 
 
 def is_input_type(cls: type) -> bool:
@@ -358,7 +381,7 @@ def _add_eq(cls: type):
     # There's no need for a __ne__ method, since Python will call __eq__ and negate it.
     if not issubclass(cls, dict) and "__eq__" not in cls.__dict__:
         def eq(self, other):
-            return type(other) is type(self) and getattr(other, _VALUES, None) == getattr(self, _VALUES, None)
+            return type(other) is type(self) and other.__dict__ == self.__dict__
         setattr(cls, "__eq__", eq)
 
 
@@ -407,7 +430,7 @@ def input_type_to_dict(value: Any) -> Dict[str, Any]:
     Returns a dict for the input type.
     """
     assert is_input_type(type(value))
-    return dict(getattr(value, _VALUES, {}))
+    return value.__dict__.copy()
 
 
 def output_type(cls: Type[T]) -> Type[T]:
@@ -434,7 +457,8 @@ def output_type(cls: Type[T]) -> Type[T]:
         def init(self, value: dict) -> None:
             if not isinstance(value, dict):
                 raise TypeError('Expected value to be a dict')
-            setattr(self, _VALUES, value)
+            for k, v in value.items():
+                self.__dict__[k] = v
         setattr(cls, "__init__", init)
 
     # Create Python properties.
@@ -454,6 +478,9 @@ def getter(_fn=None, *, name: Optional[str] = None):
     name is the Pulumi property name. If not set, the name of the function is used.
     """
     def decorator(fn: Callable) -> Callable:
+        if not callable(fn):
+            raise TypeError("Expected fn to be callable")
+
         # If name isn't specified, use the name of the function.
         pulumi_name = name if name is not None else fn.__name__
         if _utils.is_empty_function(fn):
@@ -486,10 +513,8 @@ def get(self, name: str) -> Any:
     if not isinstance(name, str):
         raise TypeError("Expected name to be a string")
 
-    values: Optional[Dict[str, Any]] = getattr(self, _VALUES, None)
-
     if hasattr(type(self), _PULUMI_INPUT_TYPE):
-        return values.get(name) if values is not None else None
+        return self.__dict__.get(name)
 
     if hasattr(type(self), _PULUMI_OUTPUT_TYPE):
         cls = type(self)
@@ -507,7 +532,12 @@ def get(self, name: str) -> Any:
             # in case the type has a `get` property.
             return getattr(dict, "get")(self, translate(self, name))
 
-        return values.get(translate(self, name)) if values is not None else None
+        return self.__dict__.get(translate(self, name))
+
+    # pylint: disable=import-outside-toplevel
+    from . import Resource
+    if isinstance(self, Resource):
+        return self.__dict__.get(name)
 
     raise AssertionError("get can only be used with classes decorated with @input_type or @output_type")
 
@@ -527,11 +557,7 @@ def set(self, name: str, value: Any) -> None:
     if not hasattr(type(self), _PULUMI_INPUT_TYPE):
         raise AssertionError("set can only be used with classes decorated with @input_type")
 
-    values = getattr(self, _VALUES, None)
-    if values is None:
-        values = dict()
-        setattr(self, _VALUES, values)
-    values[name] = value
+    self.__dict__[name] = value
 
 
 # Use the built-in `get_origin` and `get_args` functions on Python 3.8+,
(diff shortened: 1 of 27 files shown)


justinvp:
There are some remaining TODOs, but this can be reviewed in the meantime.

tests/integration/dynamic/python_types/__main__.py
```diff
@@ -0,0 +1,38 @@
...
+    def __init__(self, *, size: Input[float]):
+        self.size = size
+
+@output_type
+class RandomSize:
+    size: float
+
+class RandomResourceProvider(ResourceProvider):
+    def create(self, props):
+        size = int(props["size"]["size"])
```
lukehoban, line 20: I don't quite follow this example - it seems to have an extra layer of nesting that isn't necessary.  Perhaps thats just because it's a particularly simple case - but it seems surprising that there needs to be an extra layer here.

tests/integration/integration_test.go
```diff
@@ -1514,6 +1515,31 @@ func TestDynamicPythonVenv(t *testing.T) {
...
+			ExtraRuntimeValidation: func(t *testing.T, stack integration.RuntimeValidationStackInfo) {
+				assert.Equal(t, randomVal, stack.Outputs["random_val"].(string))
+				assert.Equal(t, randomSize, stack.Outputs["random_size"].(float64))
+			},
+		}},
+		UseAutomaticVirtualEnv: true,
+	})
+}
+
 func TestResourceWithSecretSerialization(t *testing.T) {
```
lukehoban, line 1543: Is it valuable to have a full integration test that is structurally similar to what real SDKs will do as well?

sdk/python/lib/pulumi/output.py
```diff
@@ -385,3 +387,179 @@ def __init__(self):
...
+    method, an __init__() method is added to the class that accepts a dict
+    representing the outputs.
+    """
+
+    # Get the output properties and mark the class as an output type.
+    props = _process_class(cls, _PULUMI_OUTPUT_TYPE)
+
+    # Add an __init__() method that takes a dict (representing outputs) as an arg
+    # if the class isn't a subclass of dict and doesn't have an __init__() method.
+    if not issubclass(cls, dict) and "__init__" not in cls.__dict__:
```
lukehoban, line 548: Similarly - will things work if the class already had an `__init__`?  Should we error?  Perhaps it's more pythonic to not fail and just leave this so users can customize it if they can reverse engineer the required contract?

sdk/python/lib/pulumi/output.py
```diff
@@ -385,3 +387,179 @@ def __init__(self):
...
+def input_type(cls: Type[T]) -> Type[T]:
+    """
+    Returns the same class as was passed in, but marked as an input type.
+    """
+
+    # Get the input properties and mark the class as an input type.
+    props = _process_class(cls, _PULUMI_INPUT_TYPE)
+
+    # Add a _to_dict func, if needed.
+    if not hasattr(cls, _TO_DICT):
```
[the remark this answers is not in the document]
justinvp: I was patterning this off how the built-in `dataclasses` feature works in Python. If an `__init__()` is defined, it won't add one. Similarly if `__repr__()` exists, it won't add it for you.

In this case, I was thinking of it as an escape hatch if someone needed to customize the "to dictionary" behavior, but there's no obvious scenario for that, so I'd be ok with treating it as an error for now (and we can allow overriding in the future if a scenario comes up).

tests/integration/dynamic/python_types/__main__.py
justinvp: Yeah, this is a contrived example. I took the existing Python dynamic provider test and embellished it to use an input type, just to have some coverage that these would work with dynamic providers. I'll try to come up with a more realistic example of a dynamic provider that would actually use an input type in its design (that doesn't involve creating actual cloud resources).

tests/integration/integration_test.go
justinvp: Yes, definitely. I plan to add more tests, including integration tests. It's one of the remaining tasks.

sdk/python/lib/pulumi/output.py
lukehoban: Got it - if there's precedent in core Python libraries, then happy to align with those.

sdk/python/lib/pulumi/runtime/rpc.py
```diff
@@ -362,23 +373,64 @@ def do_resolve(value_fut: 'asyncio.Future',
...
+        return typ(translated) if typ and _types.is_output_type(typ) else translated
 
     if isinstance(output, list):
-        return [translate_output_properties(res, v) for v in output]
+        element_type: Optional[type] = None
+        if typ:
+            # If typ is a list, get the type for its values, to pass
+            # along for each item.
+            origin = _types.get_origin(typ)
+            if origin is list or List:
```
pgavlin, line 429: Should we report an error if the origin is not a list?

sdk/python/lib/pulumi/runtime/rpc.py
```diff
@@ -362,23 +373,64 @@ def do_resolve(value_fut: 'asyncio.Future',
...
+        if typ and _types.is_output_type(typ):
+            # If typ is an output type, get its types, so we can pass
+            # the type along for each property.
+            types = _types.output_type_types(typ)
+            get_type = lambda k: types.get(k) # pylint: disable=unnecessary-lambda
+        elif typ:
+            # If typ is a dict, get the type for its values, to pass
+            # along for each key.
+            origin = _types.get_origin(typ)
+            if origin is dict or Dict:
```
pgavlin, line 409: Should we report an error if the origin is not a dict?

pgavlin: In general, I wonder if we should error on a type mismatch.

sdk/python/lib/pulumi/_types.py
```diff
@@ -0,0 +1,433 @@
...
+    # Get properties.
+    props = _properties_from_annotations(cls)
+
+    # Clean-up class attributes.
+    for name in props:
+        # If the class attribute (which is the default value for this prop)
+        # exists and is of type 'Property', delete the class attribute so
+        # it is not set at all in the post-processed class.
+        if isinstance(getattr(cls, name, None), _Property):
+            delattr(cls, name)
```
pgavlin, line 105: Can you explain this a bit more? I'll admit I'm a bit lost here :)

sdk/python/lib/pulumi/_types.py
```diff
@@ -0,0 +1,433 @@
...
+        return values.get(name) if values is not None else None
+
+    if hasattr(type(self), _PULUMI_OUTPUT_TYPE):
+        cls = type(self)
+
+        # If the class has a _translate_property() method, use it to translate
+        # property names, otherwise, use an identity function.
+        translate = getattr(cls, _TRANSLATE_PROPERTY, None)
+        if not callable(translate):
+            translate = lambda self, prop: prop
```
pgavlin, line 279: Does this mean that we're translating from Python -> Pulumi names at runtime inside a property getter?

sdk/python/lib/pulumi/runtime/rpc.py
justinvp: Yes, I was wondering about that too. I think we should. I'll address.

sdk/python/lib/pulumi/_utils.py
```diff
@@ -0,0 +1,57 @@
...
+def is_empty_function(fn: typing.Callable) -> bool:
+    """
+    Returns true if the function is empty.
+    """
+    consts = _consts(fn)
+    return (
+        (fn.__code__.co_code == _empty.__code__.co_code and consts == _consts_empty_) or
+        (fn.__code__.co_code == _empty_doc.__code__.co_code and consts == _consts_empty_doc) or
+        (fn.__code__.co_code == _empty_lambda.__code__.co_code and consts == _consts_empty_lambda) or
+        (fn.__code__.co_code == _empty_lambda_doc.__code__.co_code and consts == _consts_empty_lambda_doc)
```
komalali, line 56: This is wild!

sdk/python/lib/pulumi/_utils.py
```diff
@@ -0,0 +1,57 @@
...
+
+def _consts(fn: typing.Callable) -> tuple:
+    """
+    Returns a tuple of the function's constants excluding the docstring.
+    """
+    return tuple(x for x in fn.__code__.co_consts if x != fn.__doc__)
+
+
+# Precompute constants for each of the empty functions.
+_consts_empty_ = _consts(_empty)
```
komalali, line 41: why does this end with an underscore?

sdk/python/lib/pulumi/_types.py
justinvp: It's at runtime, but this is actually translating the other way from Pulumi -> "Python (mapping table)" names, and it's only done for output types.

This is how output types work in this PR:

When unmarshaling the outputs, dictionary keys are translated using the resource's `translate_output_property()` function (as is the case today). If the `dict` is intended to be an output class, we instantiate the class, passing the _translated_ `dict` to the output classes `__init__()`.

The output classes that we generate for our providers inherit from `dict`, so when we pass the _translated_ `dict` to the output class's `__init__()`, it's actually just copying the _translated_ `dict` into itself (if the output class didn't inherit from `dict`, it saves a reference to the _translated_ `dict` as `self._values`). At this point the object has the _translated_ keys, which is what we want to maintain backwards compatibility. Existing programs accessing the values as a `dict` can continue to do so, using the translated key names.

But these objects now have a nicer way of accessing the values, using the new strongly named/typed Python properties that we've added. These properties need to access the values, and since the keys are the translated keys, our provide codegen also generates a `_translate_property()` method on the output class, which does the same things as the resource's `translate_output_property()`: translates Pulumi -> "Python (mapping table)" names. If we see we have a `_translate_property()`, we call it to translate the key to lookup the value.

sdk/python/lib/pulumi/_types.py
justinvp: I can embellish in a comment, but in the meantime, here's an explanation:

### Quick summary of dataclass feature

The design and implementation is inspired and based on the built-in dataclass feature, where you can define classes like:

```python
@dataclass
class C:
    x: int
    y: int = field(repr=False)
    z: int = field(repr=False, default=10)
    t: int = 20
```

Relevant quotes from https://docs.python.org/3/library/dataclasses.html#dataclasses.dataclass:

> The dataclass() decorator examines the class to find `field`s. A `field` is defined as class variable that has a [type annotation](https://docs.python.org/3/glossary.html#term-variable-annotation).
>
> [...]
>
> fields may optionally specify a default value, using normal Python syntax:
>
> ```python
> @dataclass
> class C:
>    a: int       # 'a' has no default value
>    b: int = 0   # assign a default value for 'b'
>```

And https://docs.python.org/3/library/dataclasses.html#dataclasses.field:

> For common and simple use cases, no other functionality is required. There are, however, some dataclass features that require additional per-field information. To satisfy this need for additional information, you can replace the default field value with a call to the provided [`field()`](https://docs.python.org/3/library/dataclasses.html#dataclasses.field) function. For example:
>
> ```python
> @dataclass
> class C:
>     mylist: List[int] = field(default_factory=list)
> 
> c = C()
> c.mylist += [1, 2, 3]
> ```
>
> [...]
>
> If the default value of a field is specified by a call to `field()`, then the class attribute for this field will be replaced by the specified `default` value. If no `default` is provided, then the class attribute will be deleted. The intent is that after the `dataclass()` decorator runs, the class attributes will all contain the default values for the fields, just as if the default value itself were specified. For example, after:
>
> ```python
> @dataclass
> class C:
>     x: int
>     y: int = field(repr=False)
>     z: int = field(repr=False, default=10)
>     t: int = 20
> ```
>
> The class attribute `C.z` will be `10`, the class attribute `C.t` will be `20`, and the class attributes `C.x` and `C.y` will not be set.

## Our design

We're allowing output types to be specified in a similar way:

```
@pulumi.output_type
class Foo:
    my_arg: str = pulumi.property("myArg")
    some_other_value: bool = pulumi.property("someOtherValue")
```

Dataclasses uses "fields" as its terminology, but for us, I'm referring to these as "properties". Our "properties" can similarly be defined as class variables with type annotations. Our `pulumi.property()` function is analogous to the `dataclasses.field()` function, and is used to provide additional information associated with the "property": in this case, the Pulumi camelCase name (pun, omg, 😆).
	
We first determine the class's "properties" in `_properties_from_annotations()` by looking at the class's `__annotations__` (which is the dict `{"my_arg": str, "some_other_value": bool}` for the example above) and then look up the class attribute value for each key in that dict. In this example, the value of the `my_arg` class attribute is an instance of `_Property(name="myArg")`, and the value of `some_other_value` is `_Property(name="someOtherValue")`. We do some more massaging, store the type annotation in the `_Property` instance, and then `_properties_from_annotations()` returns a dictionary of the class attribute names to `_Property` instances.

Then the decorator "cleans up" the class, by removing any of the class attributes that were set to `_Property` via a call to `pulumi.property()`.

Dataclasses does the same thing, cleaning up the class attributes that were set to values returned from calls to `field()`.

justinvp:
@pgavlin, I've addressed your initial feedback, and also added a commit to this PR that instantiates nested output types for calls to invoke. PTAL and let me know if you have any other feedback.

sdk/python/lib/pulumi/_utils.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+    Returns true if the function is empty.
+    """
+    consts = _consts(fn)
+    return (
+        (fn.__code__.co_code == _empty.__code__.co_code and consts == _consts_empty_) or
+        (fn.__code__.co_code == _empty_doc.__code__.co_code and consts == _consts_empty_doc) or
+        (fn.__code__.co_code == _empty_lambda.__code__.co_code and consts == _consts_empty_lambda) or
+        (fn.__code__.co_code == _empty_lambda_doc.__code__.co_code and consts == _consts_empty_lambda_doc)
+    )
```
KentShikama: You could also check if the function is equivalent to returning `None` at the Bytecode level.

```suggestion
def is_empty_function(fn: typing.Callable) -> bool:
    """
    Returns true if the function is empty.
    """
    import dis
    fst, snd, *rest = [i for i in dis.Bytecode(fn)] 
    is_load_none = fst.opname == "LOAD_CONST" and fst.argval == None 
    is_return = snd.opname == "RETURN_VALUE" 
    return is_load_none and is_return and not rest 
```

justinvp: Thanks for the suggestion, @KentShikama! I tried both approaches and found the original approach was faster in microbenchmarks:

```python
import typing
import dis

def _empty():
    ...

def _empty_doc():
    """Empty function docstring."""
    ...

_empty_lambda = lambda: None

_empty_lambda_doc = lambda: None
_empty_lambda_doc.__doc__ = """Empty lambda docstring."""


def _consts(fn: typing.Callable) -> tuple:
    """
    Returns a tuple of the function's constants excluding the docstring.
    """
    return tuple(x for x in fn.__code__.co_consts if x != fn.__doc__)


# Precompute constants for each of the empty functions.
_consts_empty = _consts(_empty)
_consts_empty_doc = _consts(_empty_doc)
_consts_empty_lambda = _consts(_empty_lambda)
_consts_empty_lambda_doc = _consts(_empty_lambda_doc)


def is_empty_function(fn: typing.Callable) -> bool:
    """
    Returns true if the function is empty.
    """
    consts = _consts(fn)
    return (
        (fn.__code__.co_code == _empty.__code__.co_code and consts == _consts_empty) or
        (fn.__code__.co_code == _empty_doc.__code__.co_code and consts == _consts_empty_doc) or
        (fn.__code__.co_code == _empty_lambda.__code__.co_code and consts == _consts_empty_lambda) or
        (fn.__code__.co_code == _empty_lambda_doc.__code__.co_code and consts == _consts_empty_lambda_doc)
    )

def is_empty_function_dis(fn: typing.Callable) -> bool:
    """
    Returns true if the function is empty.
    """
    fst, snd, *rest = [i for i in dis.Bytecode(fn)]
    is_load_none = fst.opname == "LOAD_CONST" and fst.argval == None
    is_return = snd.opname == "RETURN_VALUE"
    return is_load_none and is_return and not rest


def empty():
    pass

def not_empty():
    return 42

if __name__ == '__main__':
    import timeit
    print("Original empty")
    print(timeit.repeat("is_empty_function(empty)",
                        setup="from __main__ import empty, is_empty_function",
                        repeat=3))
    print("Original non-empty")
    print(timeit.repeat("is_empty_function(not_empty)",
                        setup="from __main__ import not_empty, is_empty_function",
                        repeat=3))

    print("Suggestion empty")
    print(timeit.repeat("is_empty_function_dis(empty)",
                        setup="from __main__ import empty, is_empty_function_dis",
                        repeat=3))
    print("Suggestion non-empty")
    print(timeit.repeat("is_empty_function_dis(not_empty)",
                        setup="from __main__ import not_empty, is_empty_function_dis",
                        repeat=3))
```

Results (Python 3.8.5):

```
Original empty
[0.862670518, 0.8678176559999999, 0.8616144619999997]
Original non-empty
[1.2970095930000003, 1.301474218, 1.3091111590000004]
Suggestion empty
[7.6551781210000005, 7.600018379999998, 7.587793077000001]
Suggestion non-empty
[7.4596535589999995, 7.886800770000001, 7.810590937000001]
```

PR #5041, opened 2020-07-23
Add support for serialized resource references.

pgavlin:
Resources are serialized as their URN, ID, and package version. Each
Pulumi package is expected to register itself with the SDK. The package
will be invoked to construct appropriate instances of rehydrated
resources. Packages are distinguished by their name and their version.

This is the foundation of cross-process resources.

Related to #2430.


pgavlin:
I'll fill this in with a more comprehensive note regarding the design in a bit. I'm also working on adding tests for the Go, Node, and .NET SDKs (Python has tests).

sdk/dotnet/Pulumi/Serialization/ResourcePackages.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using Google.Protobuf.Collections;
+using Google.Protobuf.WellKnownTypes;
+
+namespace Pulumi.Serialization
+{
+	public interface IResourcePackage
+	{
+		public Resource Construct(string name, string type, IDictionary<string, object?>? args, string urn);
```
mikhailshilkov, line 14:
```suggestion
		Resource Construct(string name, string type, IDictionary<string, object?>? args, string urn);
```

sdk/dotnet/Pulumi/PublicAPI.Shipped.txt
```diff
@@ -172,6 +172,8 @@ Pulumi.ResourceTransformationResult
 Pulumi.ResourceTransformationResult.Args.get -> Pulumi.ResourceArgs
 Pulumi.ResourceTransformationResult.Options.get -> Pulumi.ResourceOptions
 Pulumi.ResourceTransformationResult.ResourceTransformationResult(Pulumi.ResourceArgs args, Pulumi.ResourceOptions options) -> void
+Pulumi.Serialization.IResourcePackage
+Pulumi.Serialization.IResourcePackage.Construct(string name, string type, System.Collections.Generic.IDictionary<string, object> args, string urn) -> Pulumi.Resource
```
mikhailshilkov, line 176: We likely want to keep all user-facing API in the root `Pulumi` namespace

pkg/resource/stack/deployment.go
```diff
@@ -543,6 +554,32 @@ func DeserializePropertyValue(v interface{}, dec config.Decrypter,
...
+					if !ok {
+						return resource.PropertyValue{}, errors.New("malformed resource value: missing urn")
+					}
+					urn := resource.URN(urnStr)
+
+					var id resource.ID
+					if idV, ok := objmap["id"]; ok {
+						idStr, ok := idV.(string)
+						if !ok {
+							return resource.PropertyValue{}, errors.New("malformed resource value: id must be a string")
```
lblackstone, line 568: Might want to log the id as well

sdk/go/pulumi/rpc.go
```diff
@@ -619,3 +663,22 @@ func unmarshalOutput(v resource.PropertyValue, dest reflect.Value) (bool, error)
...
+	Construct(name, typ string, args map[string]interface{}, urn string) (Resource, error)
+}
+
+var resourcePackages sync.Map // map[string]ResourcePackage
+
+func packageKey(name, version string) string {
+	return fmt.Sprintf("%s@%s", name, version)
+}
+
+// RegisterResourcePackage register a resource pacakge with the Pulumi runtime.
```
lblackstone, line 677:
```suggestion
// RegisterResourcePackage register a resource package with the Pulumi runtime.
```

sdk/nodejs/runtime/rpc.ts
```diff
@@ -334,9 +352,18 @@ export async function serializeProperty(ctx: string, prop: Input<any>, dependent
         // and tracked in a reasonable manner, while not causing us to compute or embed information
         // about it that is not needed, and which can lead to deadlocks.
         if (excessiveDebugOutput) {
-            log.debug(`Serialize property [${ctx}]: component resource urnid`);
+            log.debug(`Serialize property [${ctx}]: component resource urn`);
         }
 
+        if (await monitorSupportsResourceReferences()) {
+            // If we are keeping resources, emit a stronly typed wrapper over the URN
```
lblackstone, line 359:
```suggestion
            // If we are keeping resources, emit a strongly typed wrapper over the URN
```

sdk/dotnet/Pulumi/Deployment/Deployment.cs
```diff
@@ -58,6 +58,8 @@ internal static IDeploymentInternal InternalInstance
         internal IEngine Engine { get; }
         internal IMonitor Monitor { get; }
 
+		internal bool SupportsResourceReferences { get; }
```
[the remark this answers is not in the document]
pgavlin: Ah yeah, I need to fix my tabs/spaces for .cs files.

sdk/dotnet/Pulumi/Deployment/Deployment.cs
```diff
@@ -131,5 +135,11 @@ Stack IDeploymentInternal.Stack
             get => Stack;
             set => Stack = value;
         }
+
+		private static bool monitorSupportsFeature(IMonitor monitor, string feature)
+		{
+			var result = monitor.SupportsFeatureAsync(new Pulumirpc.SupportsFeatureRequest {Id = feature }).Result;
```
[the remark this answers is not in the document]
pgavlin: I intentionally chose to use `Result` here s.t. we can make this call in the constructor. A refactor is definitely possible--we'd take an approach similar to the Node and Python SDKs--do you feel it really warrants that?

EvanBoyle:
> Each Pulumi package is expected to register itself with the SDK

Does this mean provider package, or like a user defined CustomResource? If the former, will this require codegen changes?

How is this supposed to interact with stack exports? I pulled the changes locally and tried an up on a simple program:

```ts
export const bucket = new aws.s3.Bucket("bucket")
```

There was no diff, which is good from a breaking change perspective.

pgavlin:
> Does this mean provider package, or like a user defined CustomResource? If the former, will this require codegen changes?

This does mean provider packages, and this will require codegen changes once there are plugins that produce resource references.

> How is this supposed to interact with stack exports?

Resources in stack exports are JSONified by each SDK prior to serialization, so this should not change the form of exported resources. We might want to take a breaking change in the future and leave these in the reference form in stack outputs and leave it up to the CLI/service to handle their resolution and JSONification when used from `StackReference`, `pulumi stack`, etc.

EvanBoyle:
> We might want to take a breaking change in the future and leave these in the reference form in stack outputs and leave it up to the CLI/service to handle their resolution and JSONification when used from

This leads me to another question about the secretness of these references. Right now they're never secret unless infected by siblings/parents as far as I can tell. Does that make sense? What if a resource contains "secret" outputs, would a resolver handling that reference want to know ahead of time?

pgavlin:
> Right now they're never secret unless infected by siblings/parents as far as I can tell.

Currently a reference would only be secret if it was explicitly made secret, e.g. using `secret(resource)` or `additionalSecretOutputs`. This is identical to other values.

> What if a resource contains "secret" outputs, would a resolver handling that reference want to know ahead of time?

I can't think of a scenario where that would be necessary. Did you have anything in mind?

sdk/dotnet/Pulumi/Deployment/Deployment.cs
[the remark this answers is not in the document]
mikhailshilkov: I believe we have no usage of `.Result` in the SDK yet, and I'd love to keep it this way. At a glance, a refactor doesn't look too bad?

sdk/dotnet/Pulumi/PublicAPI.Shipped.txt
pgavlin: Fixed.

sdk/dotnet/Pulumi/Deployment/Deployment.cs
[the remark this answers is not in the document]
pgavlin: > I believe we have no usage of .Result in the SDK yet, and I'd love to keep it this way

OK, I'll make the change. Can you elaborate on the problems with `.Result`?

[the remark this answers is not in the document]
mikhailshilkov: IMO, it's easier to operate in async-all-the-way mode than to reason whether `Result` is safe in a particular spot. I suspect there could be some hypothetical code in a mocked implementation of `SupportsFeatureAsync` that would deadlock here. It's unlikely we ever hit this - but why even make us think about that?

pgavlin:
## Resource References

These changes introduce resource references as a new type of property value. A resource reference is a triple of (URN, ID, package version) that uniquely names a resource in the reference's stack (we may be able to add support for cross-stack references in the future).

Resource references are intended to enable cross-language components implemented in a plugin-based fashion. In this model, each component resource is managed by a plugin that is authored in a particular language. This is identical to the model we use for custom resources. Experience has shown that it is common for component resources to create child resources and return references to these children in their outputs, and we need a way to represent these references across plugins. Furthermore, it is useful for these references to be rehydrated into strongly-typed resources upon deserialization, which requires the ability to resolve the resource to its state. This capability will be added in a future PR.

In principle, a URN is sufficient for the purpose of uniquely identifying a referenced resource. However, there are other requirements of the platform that lead us to include the resource's ID and the version of the package that created it. We include the ID so the engine can replace a resource reference with either its contained URN or ID when marshaling the reference to a downlevel SDK or provider that does not understand resource references. We include the version so that the deserializer can ensure that the correct version of the SDK is used to rehydrate the resource (this is necessary because resource shapes may change between versions). In both cases we could rely on the resource state for this information instead of including it in the reference (though the version information would involve adding a new field to the resource state); they are included in the reference for ease of implementation.

EvanBoyle:
It seems like we're computing the secretness of the wrong thing. The secretness of the "reference" vs the secretness of the value of the resource. 

Consider an object that we want to serialize:

```ts
const foo = {
 a : "abc",
 b : someResourceWithSecretOutputProperties
}
```

You serialize `foo` and it is not marked as secret as far as I can tell from the implementation. Let's just imagine another client that receives this that isn't able to handle secrets. It would try to deserialize `foo`, and in the process probably make a `ReadResource` call on `someResourceWithSecretOutputProperties` that would now fail as the resource has secret properties.

pgavlin:
> Let's just imagine another client that receives this that isn't able to handle secrets. It would try to deserialize foo, and in the process probably make a ReadResource call on someResourceWithSecretOutputProperties that would now fail as the resource has secret properties.

That's right. And I think this is correct: the resource itself isn't really secret, even if it contains output properties that are.

sdk/dotnet/Pulumi/Serialization/Constants.cs
```diff
@@ -37,6 +42,9 @@ internal static class Constants
         public const string AssetOrArchivePathName = "path";
         public const string AssetOrArchiveUriName = "uri";
 
+    	public const string ResourceUrnName = "urn";
+    	public const string ResourceVersionName = "version";
```
lblackstone, line 46: I think the indentation is wrong here

sdk/dotnet/Pulumi/Serialization/Serializer.cs
```diff
@@ -166,7 +166,15 @@ public Serializer(bool excessiveDebugOutput)
...
 
                 this.DependentResources.Add(customResource);
-                return await SerializeAsync($"{ctx}.id", customResource.Id).ConfigureAwait(false);
+    			if (keepResources)
+    			{
+    				var urn = await SerializeAsync($"{ctx}.urn", customResource.Urn, keepResources).ConfigureAwait(false);
+    				var builder = ImmutableDictionary.CreateBuilder<string, object?>();
+                    builder.Add(Constants.SpecialSigKey, Constants.SpecialResourceSig);
+                    builder.Add(Constants.ResourceUrnName, urn);
+                    return builder.ToImmutable();
```
lblackstone, line 175: This indentation looks off

sdk/dotnet/Pulumi/Serialization/Serializer.cs
```diff
@@ -193,14 +201,22 @@ public Serializer(bool excessiveDebugOutput)
...
                 }
 
-                return await SerializeAsync($"{ctx}.urn", componentResource.Urn).ConfigureAwait(false);
+    			var urn = await SerializeAsync($"{ctx}.urn", componentResource.Urn, keepResources).ConfigureAwait(false);
+    			if (keepResources)
+    			{
+    				var builder = ImmutableDictionary.CreateBuilder<string, object?>();
+                    builder.Add(Constants.SpecialSigKey, Constants.SpecialResourceSig);
+                    builder.Add(Constants.ResourceUrnName, urn);
+                    return builder.ToImmutable();
```
lblackstone, line 210: Same here

sdk/go/pulumi/context.go
```diff
@@ -231,7 +244,7 @@ func (ctx *Context) Invoke(tok string, args interface{}, result interface{}, opt
 	// Otherwsie, simply unmarshal the output properties and return the result.
```
lblackstone, line 244:
```suggestion
	// Otherwise, simply unmarshal the output properties and return the result.
```

sdk/nodejs/runtime/stack.ts
```diff
@@ -67,7 +70,7 @@ class Stack extends ComponentResource<Inputs> {
      *
      * @param init The callback to run in the context of this Pulumi stack
      */
-    async initialize(args: { init: () => Promise<Inputs> }): Promise<Inputs> {
+    async initialize(): Promise<Inputs> {
```
lblackstone, line 73: Is this safe to change?

lblackstone (approved):
Functional changes LGTM

sdk/nodejs/runtime/stack.ts
pgavlin: I'm honestly unsure _why_ this changed. I'll try going without it.

pgavlin: Reverted.

PR #5052, opened 2020-07-27
Handle resource_name properties to avoid conflicts in Python codegen

### LINKED ISSUE #5050: [codegen/python] resourceName property breaks generated SDK code
If a schema resource has a `resourceName` property, the generated Python code is invalid:

```python
class Channel(pulumi.CustomResource):
#...
  def __init__(__self__, resource_name, opts=None, resource_group_name=None, resource_name=None, sku=None, tags=None, __opts__=None):
#...
```

Note two `resource_name` parameters.

I'd love to avoid the requirement to rename the parameter in the schema. For a high-fidelity provider, this would bring more maintenance burden and less match to the underlying API as-is.

A possible solution is to detect the name conflict and rename one of the parameters to `resource_name_` or `__resource_name__` specifically in this `__init__` function. I'm open to other suggestions that would be more idiomatic for Python.

mikhailshilkov:
One way to fix #5050. Alternatively, we could add `resource_name` to `PyName` as another "keyword" or another special case, I believe that would fix the issue at the level of case mappings.

I'm new to the Python SDK, so I'm totally open to other suggestions.

### DIFF
--- a/pkg/codegen/python/gen.go
+++ b/pkg/codegen/python/gen.go
@@ -460,7 +460,7 @@ func (mod *modContext) genResource(res *schema.Resource) (string, error) {
 
 	// If there's an argument type, emit it.
 	for _, prop := range res.InputProperties {
-		fmt.Fprintf(w, ", %s=None", PyName(prop.Name))
+		fmt.Fprintf(w, ", %s=None", initParamName(prop.Name))
 	}
 
 	// Old versions of TFGen emitted parameters named __name__ and __opts__. In order to preserve backwards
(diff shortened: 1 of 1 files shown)


mikhailshilkov:
@justinvp Great questions!

> How common is resource_name?

It's used in 23 resources out of ~500:

> confusing having to specify both new FooResource(resource_name="foo", resource_name_="bar")?

Yes, that's confusing, especially because

> Are they always the same value?

They are NOT the same value. The pattern is that the `resource_name_` property would point to a parent of a sub-resource, e.g. to a bot name for a `BotChannel` resource.

> Should we consider adding some kind of special casing for Azure to provide a more meaningful name,

We should consider this. Yet, so far we avoided any manual overrides, so my plan is to fix the issue on the codegen side, and then consider renaming as a follow-up. It's a bit unclear if/how much we will start adding opinionated overrides to the provider, so it makes sense to keep it to a minimum and collect ideas.

> Does the existing Azure provider have this problem?

Apparently, not. I suppose the TF provider never uses the name `resourceName`.

PR #5065, opened 2020-07-27
switch os/user with luser

jaxxstorm:
We make several calls to `os/user`, which uses CGO and means
cross-compilation is not possible. This replaces `os/user` with the
`luser` package, which is a drop-in replacement which does not use `CGO`


jaxxstorm:
> Don't we still need to maintain the platform-specific VM builds for tests and such?

Yeah we do, but the intent is to encapsulate the build process into a single build step with goreleaser to remove a bunch of the manual/custom shell scripts, which needs cross-compilation to be enabled.

PR #5077, opened 2020-07-29
Support StackOptions in F# functions

UnoSD:
Normally I would prefer a leaner, small change, but I wanted to reuse the same code for all the functions hence I had to change a bit more.

@mikhailshilkov I had a quick look at `StackOptions`, it's quite unfriendly for F#, it has a `System.Collections.Generic.List` property and it is obviously not a record. I may make a separate PR to structure it in a more "F#-y` way as we discussed.


mikhailshilkov:
I guess I messed up with the rebase... will merge in https://github.com/pulumi/pulumi/pull/6376

PR #5100, opened 2020-08-04
Install pulumictl for downstream tests

jaxxstorm:


komalali:
@jaxxstorm I noticed that the azure repo also doesn't have build/common.mk file which contains the only_build target (here it is in pulumi-gcp https://github.com/pulumi/pulumi-gcp/blob/master/build/common.mk)

komalali:
Hmm... I think this actually isn't necessary. The azure downstream is working fine now.

jaxxstorm:
@komalali think if you merge master into your branch now it should work, appreciate the patience!

Issue #5104, opened 2020-08-05
Broken upgrade to 2.8.0 for passphrase secrets provider


thrixton:
Per the title.

```
error: failed to decrypt encrypted configuration value '...': illegal base64 data at input byte 2
This can occur when a secret is copied from one stack to another. Encryption of secrets is done per-stack and
it is not possible to share an encrypted configuration value across stacks.
```

This works fine when running 2.7.1, upgrade to 2.8.0 and it immediately breaks.

There are all sorts of issues with the new options for copying stack config, changing providers as well. More to come.

thrixton:
This is working with a default secrets provider

stack72:
Hi @thrixton 

Apologies this has caused an error for you - I am working on fixing up this issue for the default secrets provider right now

Can you elaborate a little more on the following?

> There are all sorts of issues with the new options for copying stack config, changing providers as well. More to come.

Thanks

Paul

thrixton:
Thanks for looking into this so quickly, and the great work @stack72 and team :)

Regarding other issues mentioned.

All these are with the 2.8.0 release.  

1. `pulumi stack change-secrets-provider passphrase` (on a `default` provider stack) 
  This does nothing, no `encryptionsalt` value is added

2. `pulumi stack init dev2 --copy-config-from dev --secrets-provider passphrase`
2.1 `pulumi config set test test --secret`
At this point, the `encryptionsalt` is removed from the config file, I'm not 100% sure what state it is in as it's part of a larger set of stacks sorry :(
> NOTE: `dev` stack is `passphrase` protected

Quite possibly this has already been fixed in vNext with the changes from #5110.

stack72:
Hey @thrixton 

V2.8.1 is already on its way to being released

I've verified that we don't lose the encryption salt from the config files now and that we can change away from passphrase secret providers

The main question I have from you is, how badly has this affected your stack coding files and are they recoverable from source control? (I'm really hoping that the side effects are not terrible)

thrixton:
Hi @stack72 , 

All is recoverable from SC yes, thanks.

The bigger issue maybe is that 2.8.0 / latest is picked up by any CI/CD extensions (Azure Devops specifically), and there doesn't appear to be a way to "pin" (correct me if i'm wrong here) to a known good version without manual install and configuration.

[...]

PR #5110, opened 2020-08-05
Ensure Passphrase Secret Manager is available to use with new config

stack72:
Fixes: #5104

This reverts a code change that was checking initially for the
SecretsProvider of the currentStack being an empty string. When it
was an empty string, we were checking the backend type and we were
setting a serviceSecretsManager. This wasn't correct, the logic
needed to check for an empty SecretsProvider AND an empty EncryptionSalt

The important part is that it needed to check for the EncryptionSalt
to make sure it wasn't a passphrase secrets manger

I am 99% confident that this bug only affected those users who interacted directly with the configuration of the stack from the CLI i.e ran a command `pulumi config x`

### DIFF
--- a/pkg/cmd/pulumi/crypto.go
+++ b/pkg/cmd/pulumi/crypto.go
@@ -53,12 +53,6 @@ func getStackSecretsManager(s backend.Stack) (secrets.Manager, error) {
 	}
 
 	sm, err := func() (secrets.Manager, error) {
-		if ps.SecretsProvider == "" {
-			switch s.(type) {
-			case httpstate.Stack:
-				return newServiceSecretsManager(s.(httpstate.Stack), s.Ref().Name(), stackConfigFile)
-			}
-		}
 		if ps.SecretsProvider != passphrase.Type && ps.SecretsProvider != "default" && ps.SecretsProvider != "" {
 			return newCloudSecretsManager(s.Ref().Name(), stackConfigFile, ps.SecretsProvider)
 		}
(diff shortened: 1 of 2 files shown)


pkg/cmd/pulumi/crypto.go
```diff
@@ -70,6 +64,8 @@ func getStackSecretsManager(s backend.Stack) (secrets.Manager, error) {
 		switch s.(type) {
 		case filestate.Stack:
 			return newPassphraseSecretsManager(s.Ref().Name(), stackConfigFile)
+		case httpstate.Stack:
```
lukehoban, line 67: I *think* it's okay (a no-op) to keep the check for `ps.EncryptionSalt != ""` above this case - but the overal structure of this code is a little hard to follow.  There are really three cases here - is it possible to structure this to make the checks for each case clearer?  For example - is the line above this (`return newPassphraseSecretsManager`) even reachable currently?

stack72: It is reachable when the code has been called as part of creating a new stack - it has no secretsManager at that point and therefore will have an empty secretsProvider AND an empty EncryptionSalt as they have not been added to the config file

stack72: FWIW, this returns the code in this specific block back to that of what it was before

lukehoban: > FWIW, this returns the code in this specific block back to that of what it was before

Okay - great.

> It is reachable when the code has been called as part of creating a new stack - it has no secretsManager at that point and therefore will have an empty secretsProvider AND an empty EncryptionSalt as they have not been added to the config file

I see - definitely would be helpful to note that in a comment here - though recognize this is an issue with the code as it was before you touched it.

stack72: @lukehoban so we have 2 options here:

1. We work out the test coverage now and risk causing more potential issues for passphrase secrets provider users

2. We ship the patch (as this code is now back to what it was before) and we then work out the testing of the code

Thoughts?

lukehoban: Definitely okay with fixing the serious user-facing bug ASAP - but would love to invest in test coverage here as well.

Issue #5108, opened 2020-08-05
[Go Program Gen] K8s incorrect token normalization, missing import aliases


EvanBoyle:
Related to: #5101

PR #5115, opened 2020-08-05
[Go Program Gen] Fix various Kubernetes issues

EvanBoyle:
Reviewing this commit by commit is best. Commits line up with the following fixes:

1. Use import aliases
2. account for import aliases in expressions
3. Traverse object and tuple cons expressions for imports

(3) is only necessary for Go, as other languages import the package, and then dot into the relevant namespace. The fix for 3 winds up being a little messy as it requires that we collect imports, generate the program, and then collect imports and actually emit the imports. This is because we need to run the collection logic over both the lowered and unlowered program. The unlowered program has easy access to things like JSON builtins that get rewritten to nested applies (would be complex to handle appropriately), the lowered program has all of the __convert intrinsics required to capture the imports for nested object/tuple cons expressions. 

This could be written as one import pass, but it would be far more complicated and unfortunately, I've consumed all the time I have to spend on this issue. 

fixes https://github.com/pulumi/pulumi/issues/5108

### DIFF
--- a/pkg/codegen/go/gen_program.go
+++ b/pkg/codegen/go/gen_program.go
@@ -33,6 +33,8 @@ type generator struct {
 	arrayHelpers        map[string]*promptToInputArrayHelper
 	isErrAssigned       bool
 	importAliases       map[string]map[string]string
+	importBasePaths     map[string]string
+	modToPkg            map[string]map[string]string
 }
 
 func GenerateProgram(program *hcl2.Program) (map[string][]byte, hcl.Diagnostics, error) {
@@ -41,6 +43,8 @@ func GenerateProgram(program *hcl2.Program) (map[string][]byte, hcl.Diagnostics,
 
 	contexts := make(map[string]map[string]*pkgContext)
 	importAliases := make(map[string]map[string]string)
+	modToPkg := make(map[string]map[string]string)
+	importBasePaths := make(map[string]string)
 	for _, pkg := range program.Packages() {
 		contexts[pkg.Name] = getPackages("tool", pkg)
 
@@ -49,8 +53,9 @@ func GenerateProgram(program *hcl2.Program) (map[string][]byte, hcl.Diagnostics,
 			return make(map[string][]byte), nil, err
 		}
 		if langInfo, ok := pkg.Language["go"].(GoPackageInfo); ok {
-			aliases := langInfo.PackageImportAliases
-			importAliases[pkg.Name] = aliases
+			importAliases[pkg.Name] = langInfo.PackageImportAliases
+			importBasePaths[pkg.Name] = langInfo.ImportBasePath
+			modToPkg[pkg.Name] = langInfo.ModuleToPackage
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/codegen/go/gen_program.go
```diff
@@ -54,22 +69,38 @@ func GenerateProgram(program *hcl2.Program) (map[string][]byte, hcl.Diagnostics,
...
 
 	for _, n := range nodes {
-		g.genNode(&index, n)
+		g.genNode(&progPostamble, n)
 	}
 
-	g.genPostamble(&index, nodes)
+	g.genPostamble(&progPostamble, nodes)
+
+	// We must genearte the program first and the preamble second and finally cat the two together.
```
pgavlin, line 97:
```suggestion
	// We must generate the program first and the preamble second and finally cat the two together.
```

pkg/codegen/go/gen_program.go
```diff
@@ -54,22 +69,38 @@ func GenerateProgram(program *hcl2.Program) (map[string][]byte, hcl.Diagnostics,
...
+	// This is because nested object/tuple cons expressions can require imports that aren't
+	// present in resource declarations or invokes alone. Expressions are lowered when the program is generated
+	// and this must happen first so we can access types via __convert intrinsics.
```
pgavlin, line 100: Note that we could also approach this by lowering the entire program prior to generation. I think the approach you've taken here is fine, though.

pkg/codegen/go/gen_program.go
```diff
@@ -32,16 +32,31 @@ type generator struct {
 	scopeTraversalRoots codegen.StringSet
 	arrayHelpers        map[string]*promptToInputArrayHelper
 	isErrAssigned       bool
+	importAliases       map[string]map[string]string
+	importBasePaths     map[string]string
+	modToPkg            map[string]map[string]string
```
pgavlin, line 37: Instead of separating these maps, I think it might be cleaner to use a single `map[string]*schema.Package`. You could then rephrase some of the helpers in terms of `*schema.Package`, which would make it easier to take advantage of the changes in https://github.com/pulumi/pulumi/pull/5114 as a follow-up change.

pkg/codegen/go/gen_program.go
EvanBoyle: Yes, but unfortunately lowering the program has side-effects in the form of producing temp variables that must be emitted inline as code is generated. 

This scenario makes it obvious that this is not a great design, but nothing much to do about it at this point in time.

EvanBoyle:
No unexpected diffs as far as I can tell. The `String -> Any` in some of the providers is a change that hasn't been rolled out. This is actually the accurate type given that the variable referenced does not exist in the program (placeholders).

pkg/codegen/go/gen_program.go
EvanBoyle: Oh, good idea! For now, I've just added a TODO for context next time we're working in this area.

pgavlin: I'd prefer that we replace the maps now if it's not too much trouble. The bit about taking advantage of #5114 can definitely wait, though.

PR #5127, opened 2020-08-06
Use a package map where appropriate

pgavlin:


pkg/codegen/internal/test/testdata/kubernetes-operator.pp.cs
```diff
@@ -0,0 +1,282 @@
...
+                                        Value = "pulumi-kubernetes-operator",
+                                    },
+                                },
+                            },
+                        },
+                    },
+                },
+            },
+        });
+        var pulumi_kubernetes_operatorRole = new Kubernetes.Rbac.authorization.k8s.io.v1.Role("pulumi_kubernetes_operatorRole", new Kubernetes.Rbac.authorization.k8s.io.v1.RoleArgs
```
EvanBoyle, line 96: @mikhailshilkov Looks like the dotnet program gen is not consuming the `namespaces` mapping here

pkg/codegen/internal/test/testdata/kubernetes-operator.pp.py
```diff
@@ -0,0 +1,169 @@
...
+                        {
+                            "name": "OPERATOR_NAME",
+                            "value": "pulumi-kubernetes-operator",
+                        },
+                    ],
+                }],
+            },
+        },
+    })
+pulumi_kubernetes_operator_role = kubernetes.rbac_authorization_k8s_io.v1.Role("pulumi_kubernetes_operatorRole",
```
EvanBoyle, line 60: @komalali it looks like the python program gen is not consuming `moduleNameOverrides` in this instance

Issue #5134, opened 2020-08-07
`history` command should be a sub-command under `stack`


clstokes:
1. `pulumi history` should be `pulumi stack history` to be consistent with other stack-specific commands.
1. The help text for `pulumi history` should be updated. It currently says `Update history for a stack`, but has no _update_ capabilities.

```
clstokes % pulumi history --help
Update history for a stack

This command lists data about previous updates for a stack.

Usage:
  pulumi history [flags]
...
```

stack72:
This test should probably be "History of stack updates" ?

clstokes:
Looks like we're inconsistent between "Display ...", "Show ...", and "Print ...", but I think a more _action_ statement like "Display history for a stack" would be best.

stack72:
Moving the command would be a breaking change - we should actually duplicate the command under the stack command and then we would add a deprecated to the existing use case.

We would then need to remove this from v3.0.0

PR #5158, opened 2020-08-11
The history command should be a subcommand of stack

stack72:
Fixes: #5134

This ensures that `pulumi history` has been deprecated in favor of
the new `pulumi stack history` command. The deprecated command will
be removed in v3.0.0 of Pulumi

### DIFF
--- a/pkg/cmd/pulumi/history.go
+++ b/pkg/cmd/pulumi/history.go
@@ -15,24 +15,15 @@
 package main
 
 import (
-	"encoding/json"
-	"fmt"
-	"sort"
-	"strings"
-	"time"
-
-	"github.com/dustin/go-humanize"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 
-	"github.com/pulumi/pulumi/pkg/v2/backend"
 	"github.com/pulumi/pulumi/pkg/v2/backend/display"
-	"github.com/pulumi/pulumi/sdk/v2/go/common/diag/colors"
 	"github.com/pulumi/pulumi/sdk/v2/go/common/resource/config"
 	"github.com/pulumi/pulumi/sdk/v2/go/common/util/cmdutil"
-	"github.com/pulumi/pulumi/sdk/v2/go/common/util/contract"
 )
 
+// TO-DO: Remove as part of Pulumi v3.0.0
 func newHistoryCmd() *cobra.Command {
 	var stack string
 	var jsonOut bool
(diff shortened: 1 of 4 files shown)


pkg/cmd/pulumi/history.go
```diff
@@ -15,24 +15,15 @@
...
 
-	"github.com/pulumi/pulumi/pkg/v2/backend"
 	"github.com/pulumi/pulumi/pkg/v2/backend/display"
-	"github.com/pulumi/pulumi/sdk/v2/go/common/diag/colors"
 	"github.com/pulumi/pulumi/sdk/v2/go/common/resource/config"
 	"github.com/pulumi/pulumi/sdk/v2/go/common/util/cmdutil"
-	"github.com/pulumi/pulumi/sdk/v2/go/common/util/contract"
 )
 
+// TO-DO: Remove as part of Pulumi v3.0.0
```
justinvp, line 26: Can we add an issue to track and add a link to the TODO comment here? As an aside, I also wonder if we should add an issue tag in GH (or something) so we can more easily track what things we'd like to address in the next major revision. (To make it easier on ourselves when we start planning for such a release.)

Also as an aside, this command is marked as "[PREVIEW]". Since it's a preview, would we necessarily have to wait until the next major version to delete it?

pkg/cmd/pulumi/history.go
```diff
@@ -41,10 +32,11 @@ func newHistoryCmd() *cobra.Command {
 		Use:        "history",
 		Aliases:    []string{"hist"},
 		SuggestFor: []string{"updates"},
-		Short:      "[PREVIEW] Update history for a stack",
-		Long: `Update history for a stack
-
-This command lists data about previous updates for a stack.`,
+		Short:      "[DEPRECATED] Display history for a stack",
```
justinvp, line 35: Should we add `Hidden: true` so the command doesn't show up in `--help`?

pkg/cmd/pulumi/history.go
```diff
@@ -41,10 +32,11 @@ func newHistoryCmd() *cobra.Command {
...
+		Long: "Display history for a stack.\n\n" +
+			"This command displays data about previous updates for a stack.\n\n" +
+			"This command is now DEPRECATED, please use `pulumi stack history`.\n" +
+			"The command will be removed in Pulumi v3.0.0",
```
justinvp, line 39: Should we be more vague about timeframe?

```suggestion
			"The command will be removed in a future release.",
```

pkg/cmd/pulumi/history.go
stack72: I'm going to suggest we leave it in CLI for now so that anyone who has been testing it actually can still use it and get the help text

I will create a follow issue for it's removal - we should discuss the tag with @leezen

pkg/cmd/pulumi/history.go
stack72: good idea :)

pkg/cmd/pulumi/history.go
stack72: Agreed!

PR #5137, opened 2020-08-09
Docgen changes for Python input/output types

justinvp:
Resource doc changes for Python:

- Types are included in constructor/function args
- The property names for input/output types are now always snake_case, regardless of the generated mapping tables, to match the new input/output classes
- Some other minor tweaks to function/constructor signatures (e.g. removed the `__props__` arg, as it's not meant to be used directly; use `@staticmethod` for static `get` methods).

I've opened a [Draft PR](https://github.com/pulumi/docs/pull/3906) in the docs repo that isn't meant to be merged, but demonstrates these changes for the AWS and K8s providers.

- View [this commit](https://github.com/pulumi/docs/pull/3906/commits/5b13017dc5b599191f81993173b383397d1877aa) to see the diff
- See the [preview](http://pulumi-docs-origin-pr-3906-b57027e1.s3-website.us-west-2.amazonaws.com) of the site (e.g. [Bucket](http://pulumi-docs-origin-pr-3906-b57027e1.s3-website.us-west-2.amazonaws.com/docs/reference/pkg/aws/s3/bucket/#create), [GetBucket](http://pulumi-docs-origin-pr-3906-b57027e1.s3-website.us-west-2.amazonaws.com/docs/reference/pkg/aws/s3/getbucket/#using))

Fixes https://github.com/pulumi/pulumi/issues/4791

PR #5165, opened 2020-08-12
[codegen/docs] Fix the case used in links for Node.js and Python modules.

praneetloke:
Related to https://github.com/pulumi/docs/issues/3870.

### DIFF
--- a/pkg/codegen/python/doc.go
+++ b/pkg/codegen/python/doc.go
@@ -194,7 +194,7 @@ func (d DocLanguageHelper) GetModuleDocLink(pkg *schema.Package, modName string)
 	if modName == "" {
 		displayName = fmt.Sprintf("pulumi_%s", pkg.Name)
 	} else {
-		displayName = fmt.Sprintf("pulumi_%s/%s", pkg.Name, strings.ToLower(modName))
+		displayName = fmt.Sprintf("pulumi_%s/%s", pkg.Name, PyName(modName))
 	}
 	link = fmt.Sprintf("/docs/reference/pkg/python/%s", displayName)
 	return displayName, link
--- a/pkg/codegen/python/doc.go
+++ b/pkg/codegen/python/doc.go
@@ -194,7 +194,7 @@ func (d DocLanguageHelper) GetModuleDocLink(pkg *schema.Package, modName string)
 	if modName == "" {
 		displayName = fmt.Sprintf("pulumi_%s", pkg.Name)
 	} else {
-		displayName = fmt.Sprintf("pulumi_%s/%s", pkg.Name, PyName(modName))
+		displayName = fmt.Sprintf("pulumi_%s/%s", pkg.Name, strings.ToLower(modName))
 	}
 	link = fmt.Sprintf("/docs/reference/pkg/python/%s", displayName)
 	return displayName, link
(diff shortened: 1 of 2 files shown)


praneetloke:
Ping. :)

pkg/codegen/python/doc.go
```diff
@@ -194,7 +194,7 @@ func (d DocLanguageHelper) GetModuleDocLink(pkg *schema.Package, modName string)
 	if modName == "" {
 		displayName = fmt.Sprintf("pulumi_%s", pkg.Name)
 	} else {
-		displayName = fmt.Sprintf("pulumi_%s/%s", pkg.Name, strings.ToLower(modName))
+		displayName = fmt.Sprintf("pulumi_%s/%s", pkg.Name, PyName(modName))
```
justinvp, line 197: Are you sure this change is correct? I don't think the Python modules are snake_case, they're lowercase all one word.

praneetloke: Hmm let me double-check that. I did this because the URL for the `pkiSecret` Python module fails when it is just a lower-cased module name. I assumed that every module name in Python needs to be snake_case'd. Coming to think of it we should have had way more broken links because every module link would be broken...

Maybe there is actually some special case for the `pkiSecret` module itself that I am missing.
Found on: https://www.pulumi.com/docs/reference/pkg/vault/pkisecret/
This doesn't work: https://www.pulumi.com/docs/reference/pkg/python/pulumi_vault/pkisecret/
This works: https://www.pulumi.com/docs/reference/pkg/python/pulumi_vault/pki_secret/

It's the same with the `appRole` module.
Found on: https://www.pulumi.com/docs/reference/pkg/vault/approle/
This doesn't work: https://www.pulumi.com/docs/reference/pkg/python/pulumi_vault/approle/
This works: https://www.pulumi.com/docs/reference/pkg/python/pulumi_vault/app_role/

Both of those are in the `vault` package. However, here's one from the Azure DevOps package:

Found on: https://www.pulumi.com/docs/reference/pkg/azuredevops/serviceendpoint/
This doesn't work: https://www.pulumi.com/docs/reference/pkg/python/pulumi_azuredevops/serviceendpoint/
This works: https://www.pulumi.com/docs/reference/pkg/python/pulumi_azuredevops/service_endpoint/

We don't have a module name map for Python SDKs, do we?

justinvp: This is how the mod names are treated in Python:

https://github.com/pulumi/pulumi/blob/b273e1cbb1c4422d146bed16414a3a26d838afda/pkg/codegen/python/gen.go#L1361-L1368

So I would expect us to use `PyName(strings.ToLower("pkiSecret"))` here, but I haven't traced through `PyName` to see if it would actually change anything that's already all lower case.

In the vault schema, it's `pkiSecret` and `PyName(strings.ToLower("pkiSecret"))` returns `pkisecret`, so I'm not sure where `pki_secret` is coming from.

Something odd is going on with the vault provider: https://github.com/pulumi/pulumi-vault/tree/master/sdk/python/pulumi_vault. It has both `pkiSecret` and `pki_secret` dirs, and `appRole` and `app_role` dirs. My guess is we initially were generating the the mod dirs one way, and then changed to some other way, but the old dirs got left behind. There could also be some weird git casing going on where we initially committed with camelCase and now are intending to be using lowercase, but git is keeping it camelCase. I haven't had a chance to dig further into what is going on.

praneetloke: Ah I see. The Python module name overrides is somewhat of a recent addition and I see that you have added that in your [PR](https://github.com/pulumi/pulumi/pull/5137) for the input/output type doc gen changes.

For now, I have reverted this change while I look into this further.

Issue #4103, opened 2020-03-18
Allow to suppress Permalink from preview/up output


omerlh:
Currently, when running preview/up commands the output contains `Permalink`, which (at least on AWS) contains sensitive information - which is an issue when posting the output to GitHub. I would like a flag that allows suppressing this link from the output.

lukehoban:
> which (at least on AWS) contains sensitive information 

Could you clarify this?  Do you mean the URL itself does?  Or do you mean the page it links to (which should be authenticated) does?

> at least on AWS

Do you specifically mean when using the S3 state storage backend?

omerlh:
It's a pre-signed AWS url to the bucket - valid for an hour. It's not horrible, but I don't want to show it there..

lukehoban:
Got it - this is for the S3 state storage backend.

I think quite possibly we just shouldn't show the "permalink" at all for the filestate backends.

glennsarti:
Also applicable to Azure Blob storage

```
Permalink: https://something.blob.core.windows.net/pulumi-state/.pulumi/stacks/dev.json?se=2020-06-17T08%3A17%...SAS KEY
```

scott-the-programmer:
I'll pick this one up if that's OK. 

I'll try get a PR up soon :+1:

lukehoban:
This was fixed with #5191.  There is now a `-suppress-permalink` option you can pass to the CLI.

glennsarti:
🥳   Thanks @lukehoban !

PR #5177, opened 2020-08-15
Added --suppress-permalink option

scott-the-programmer:
## Issue

https://github.com/pulumi/pulumi/issues/4103

We should allow the ability to suppress the permalink output entirely, regardless of the backend type.

## Intent of this PR
`pulumi up/preview/refresh/destroy` will now accept `--suppress-permalink` which will not display the state permalink in the log output. This supports all types of backends

This behavior follows that of `--suppress-output`

## Testing

* Added integration tests `TestPermaLinkSuppressed` and `TestPermaLink` to automatically check the output
* Tested locally by building the `cmd` project and invoking up/preview/refresh/destroy

```console
> pulumi up -y --suppress-permalink

Previewing update (sm-example):
     pulumi:pulumi:Stack  sm-example

Resources:
    2 unchanged

Updating (example-sticky):
     Type                 Name                       Status
     pulumi:pulumi:Stack  sm-example

Resources:
    2 unchanged

Duration: 4s
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@ CHANGELOG
 - Add support for extracting jar files in archive resources
   [#5150](https://github.com/pulumi/pulumi/pull/5150)
 
+- Added `--suppress-permalink` option to suppress the permalink output
+  (fixes [#4103](https://github.com/pulumi/pulumi/issues/4103))
+  [#5177](https://github.com/pulumi/pulumi/pull/5177)
 
 ## 2.8.2 (2020-08-07)
 
--- a/pkg/backend/apply.go
+++ b/pkg/backend/apply.go
@@ -39,7 +39,7 @@ import (
 type ApplierOptions struct {
 	// DryRun indicates if the update should not change any resource state and instead just preview changes.
 	DryRun bool
-	// ShowLink indicates if a link to the update persisted result should be displayed.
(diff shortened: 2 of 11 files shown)


pkg/backend/watch.go
```diff
@@ -35,8 +35,8 @@ import (
 func Watch(ctx context.Context, b Backend, stack Stack, op UpdateOperation, apply Applier) result.Result {
 
 	opts := ApplierOptions{
-		DryRun:   false,
-		ShowLink: false,
+		DryRun:      false,
+		CanShowLink: false,
```
lukehoban, line 39: I’m not sure we need to change the name of this.

scott-the-programmer:
Yeah - I agree with introducing it in non-disruptive way for now.

I can pick this back up in the future (say a month or so?) and change the defaults. I reckon this would be much nicer for the permalink behaviour to be opt-in rather than opt-out

pkg/backend/watch.go
scott-the-programmer: Agree, I thought it could be a bit confusing since suppressPermalink and showLink have conflicting meanings

I.e showLink represents having the ability to produce a permalink whilst suppressPermalink represents whether it should actually show the permalink

Happy to change it back, though!

lukehoban:
Thanks @scott-the-programmer! :tada:

Issue #5142, opened 2020-08-10
Python: Support docstrings in IDE hover tooltips over output properties for resources and function results


justinvp:
## Problem

Today docstrings for resource outputs are specified after class attribute annotations:

```python
class Bucket(pulumi.CustomResource):
    arn: pulumi.Output[str]
    """
    The ARN of the bucket. Will be of format `arn:aws:s3:::bucketname`.
    """
```

The problem with this is that such docstrings aren't actually associated with each class attribute, so when hovering over `arn` in `my_bucket.arn` in IDEs, you don't see the docstring in tooltips.

I suppose the reason these don't show up in tooltips is because Python ignores these docstrings. The language allows them to be declared there (docstrings can be declared pretty much anywhere), but they're ignored by the language and aren't accessible programmatically AFAICT (i.e. they aren't exposed as `__doc__` anywhere, like on classes and functions). They seem to be only used by doc generators like Sphinx. [PEP 224](https://www.python.org/dev/peps/pep-0224/) proposed formally adding these to the language, but the PEP was rejected.

We have a similar issue with function results:

```python
class GetBucketResult:
    """
    A collection of values returned by getBucket.
    """
    def __init__(__self__, ...):
        if arn and not isinstance(arn, str):
            raise TypeError("Expected argument 'arn' to be a str")
        __self__.arn = arn
        """
        The ARN of the bucket. Will be of format `arn:aws:s3:::bucketname`.
        """
        if bucket_domain_name and not isinstance(bucket_domain_name, str):
            raise TypeError("Expected argument 'bucket_domain_name' to be a str")
        __self__.bucket_domain_name = bucket_domain_name
        """
        The bucket domain name. Will be of format `bucketname.s3.amazonaws.com`.
        """
```

Here we declare a docstring after setting the instance variable (again, Python allows it, but ignores them). PyCharm will actually show these in tooltips, but VS Code does not.

## Proposed Solution

If we use Python properties rather than docstrings after attributes, the docstrings declared on the property's getter function do show up in tooltips in PyCharm and VS Code (though, it only works in VS Code when the new Pylance language service is installed & enabled, which is going to be the default lang service for Python in the future).

Here's how this would work for Resources:

We add the following "protected" method to the base `Resource` class:

```python
class Resource:
    def _set_property(self, name: str, value: Any) -> None:
        self.__setattr__(name, value)
```

[...]

Issue #5141, opened 2020-08-10
Python: Automatically generate `__init__` for input classes


justinvp:
Follow-up from https://github.com/pulumi/pulumi/pull/5033

- For input classes, generate an `__init__()` method (if doesn't already exist)
- ~~For input/output classes, generate a `__repr__()` method (if it doesn't already exist)~~ cut for now

justinvp:
Fixed by #5180

PR #5179, opened 2020-08-17
[sdk/python] Use Python properties for resource outputs and input/output type updates

justinvp:
This is a PR into my `justin/pyio_sdk` feature branch. Once any feedback is addressed and these changes are merged, I'll merge the feature branch into master.

Two commits:

### Use Python properties for resource outputs

This enables docstrings to show up when hovering over properties in PyCharm and VS Code (w/ Pylance).

The property values are stored in and retrieved from the object's `__dict__`, which allows existing things like `export("res", res)` to continue to work the same as before, and aligns the input/output types to store values in `__dict__` the same way.

### Changes to input/output types

Two main changes:

1. Previously, if an @output_type class didn't have an __init__ method, one would be added by the decorator that accepted a single dict argument representing the output properties. This commit makes it so that instead of accepting a single dict argument, it accepts individual arguments for each property based on annotations defined on the class. Classes decorated with @input_type now do the same thing.

    Note: I was considering also adding a `__repr__` method as well, but ran out of time, so decided to cut this. It's pretty low priority and we could decide to add it in the future.

2. Values in input/output types are now stored in `self.__dict__` as snake_case Python names, rather than being stored in `self._values` as camelCase Pulumi names, to be consistent with how values are stored for resources. This allows these types to work the same as resources when passed to `pulumi.export`. There are two exceptions for output types: (a) if the class is a subclass of dict, the values are stored in the dict itself (rather than __dict__), and (b) if the output class has a `_translate_property` method, the names will be the translated names from calling that method.

Related PR (provider codegen changes): https://github.com/pulumi/pulumi/pull/5180

Fixes https://github.com/pulumi/pulumi/issues/5142
Fixes https://github.com/pulumi/pulumi/issues/5141

### DIFF
--- a/sdk/python/lib/pulumi/_types.py
+++ b/sdk/python/lib/pulumi/_types.py
@@ -491,8 +491,7 @@ def output_type(cls: Type[T]) -> Type[T]:
         python_to_pulumi_table = None
         for python_name, pulumi_name, _ in _py_properties(cls):
             if python_name != pulumi_name:
-                if python_to_pulumi_table is None:
-                    python_to_pulumi_table = {}
+                python_to_pulumi_table = python_to_pulumi_table or {}
                 python_to_pulumi_table[python_name] = pulumi_name
         if python_to_pulumi_table is not None:
             setattr(cls, _PULUMI_PYTHON_TO_PULUMI_TABLE, python_to_pulumi_table)
--- a/sdk/python/lib/pulumi/_types.py
+++ b/sdk/python/lib/pulumi/_types.py
@@ -32,6 +32,28 @@
 #
 # When the resource's outputs are resolved, the `Nested` class is instantiated.
 #
+# The resource could alternatively be declared using a Python property for nested_value rather than
+# using a class annotation:
+#
+#   class FooResource(pulumi.CustomResource):
+#       def __init__(self, resource_name, nested_value: pulumi.InputType[NestedArgs]):
+#           super().__init__("my:module:FooResource", resource_name, {"nestedValue": nested_value})
+#
+#       @property
+#       @pulumi.getter(name="nestedValue")
+#       def nested_value(self) -> pulumi.Output[Nested]:
+#           ...
+#
+#
+# Note the `nested_value` property getter function is empty. The `@pulumi.getter` decorator replaces
+# empty function bodies with an actual implementation. In this case, it replaces it with a body that
+# looks like:
+#
+#       @property
+#       @pulumi.getter(name="nestedValue")
+#       def nested_value(self) -> pulumi.Output[Nested]:
+#           pulumi.get(self, "nestedValue")
+#
 #
 # Here's how the `NestedArgs` input class can be declared:
 #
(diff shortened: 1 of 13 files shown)


sdk/python/lib/pulumi/_types.py
```diff
@@ -77,44 +98,45 @@
...
 #
-# The former (simpler) declaration is nice syntactic sugar to use when declaring these by hand,
+# The former (simpler) declaration is syntactic sugar to use when declaring these by hand,
 # e.g. when writing a dynamic provider that has nested inputs/outputs. The latter declaration isn't
 # as pleasant to write by hand and is closer to what we emit in our provider codegen. The benefit
 # of the latter (more verbose) form is that it allows docstrings to be specified on the Python
 # property getters, which will show up in IDE tooltips when hovering over the property.
 #
-# Note the property getter/setter functions are empty in the latter (more verbose) declaration.
+# Note the property getter/setter functions are empty in the more verbose declaration.
```
komalali, line 110: I'd like to understand the reason behind leaving this function empty and having it be automatically replaced. Since we are generating the code anyway, why not generate the actual implementation? Are we doing this just to avoid the verbosity?

sdk/python/lib/pulumi/_types.py
```diff
@@ -417,34 +470,43 @@ def output_type(cls: Type[T]) -> Type[T]:
...
+    # the Python name to the Pulumi name, and then pass the Pulumi name to _translate_property() to
+    # convert the Pulumi name to whatever name _translate_property() returns (which, for our
+    # provider codegen, will be the translated name from _tables.CAMEL_TO_SNAKE_CASE_TABLE).
+    # pylint: disable=too-many-nested-blocks
+    if hasattr(cls, _TRANSLATE_PROPERTY):
+        python_to_pulumi_table = None
+        for python_name, pulumi_name, _ in _py_properties(cls):
+            if python_name != pulumi_name:
+                if python_to_pulumi_table is None:
+                    python_to_pulumi_table = {}
```
komalali, line 495:
```suggestion
                python_to_pulumi_table = python_to_pulumi_table or {}
```

sdk/python/lib/pulumi/_types.py
justinvp: Yeah, I was wondering if this was "too much magic": https://github.com/pulumi/pulumi-aws/commit/a77712011c3dc7577a2737b5b3004ee529ea5dc6#r41448321

The motivation for allowing empty bodies was so you wouldn't have to specify names twice (if writing this more verbose form by hand): once in the declaration of the function and again inside the function's body. But I'm definitely open to just emitting the actual implementation in the generated code, where that doesn't matter (and too much "magic" is confusing / hard to debug).

If we did leave the bodies empty, we could theoretically generate faster lookup code in some cases. That's not being done currently, but in the cases where we need to translate the property names a couple times per invocation, we could generate that code ahead-of-time, rather than doing the translations at runtime. But I doubt this would make much of a measurable difference in practice.

komalali: Yeah, it's nice that this option exists for the cases where a provider is being hand-written but for the generated code I vote for emitting the actual implementation for the sake of simplicity and clarity.

justinvp: That sounds good to me. 👍

PR #5180, opened 2020-08-17
[codegen/python] Support docstrings for resource properties / function results

justinvp:
This is a PR into my `justin/pyio_codegen` feature branch.  Once any feedback is addressed and these changes are merged, I'll merge the feature branch into master.

Diffs (just these changes):
- [AWS](https://github.com/pulumi/pulumi-aws/commit/8cd9bbca359252b3d7091e18d7cb8b22c17e972e)
- [Kubernetes](https://github.com/pulumi/pulumi-kubernetes/commit/559b0cb4ffff01d63447d11fb44de327d3b33e59)

Two main changes:

1. Use Python properties for resource outputs and function results. This enables docstrings to show up when hovering over properties in PyCharm and VS Code (w/ Pylance).

2. Updates to input/output classes:

   - Add explicit constructors to output types and remove the need for the private function result class.

   - Use snake_case Python names when storing values to be consistent with casing that resources use for storing values.

Related PR (SDK changes): https://github.com/pulumi/pulumi/pull/5179

Fixes https://github.com/pulumi/pulumi/issues/5142
Fixes https://github.com/pulumi/pulumi/issues/5141

### DIFF
--- a/pkg/codegen/python/gen.go
+++ b/pkg/codegen/python/gen.go
@@ -630,7 +630,7 @@ func (mod *modContext) genAwaitableType(w io.Writer, obj *schema.ObjectType) str
 		if prop.Comment != "" {
 			printComment(w, prop.Comment, "        ")
 		}
-		fmt.Fprintf(w, "        ...\n\n")
+		fmt.Fprintf(w, "        return pulumi.get(self, %q)\n\n", pname)
 	}
 
 	// Produce an awaitable subclass.
@@ -907,7 +907,7 @@ func (mod *modContext) genResource(res *schema.Resource) (string, error) {
 		if prop.Comment != "" {
 			printComment(w, prop.Comment, "        ")
 		}
-		fmt.Fprintf(w, "        ...\n\n")
+		fmt.Fprintf(w, "        return pulumi.get(self, %q)\n\n", pname)
 	}
 
 	// Override translate_{input|output}_property on each resource to translate between snake case and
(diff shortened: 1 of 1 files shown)


komalali:
Can we lump the deprecation message and the setter under the same `if [blah] is not None:`? https://github.com/pulumi/pulumi-aws/compare/justin/python_full#diff-7cab7b553a0ffb44e070d9a3d51ee16dR642-R646

Also, we have `kinesis_analytics` and `kinesisanalytics` and `kinesisanalyticsv2`... is that just an AWS funk or is something going on our end?

pkg/codegen/python/gen.go
```diff
@@ -624,9 +611,26 @@ func (mod *modContext) genAwaitableType(w io.Writer, obj *schema.ObjectType) str
...
+	// Write out Python property getters for each property.
+	for _, prop := range obj.Properties {
+		pname := PyName(prop.Name)
+		ty := mod.typeString(prop.Type, false /*input*/, false /*wrapInput*/, !prop.IsRequired, false /*acceptMapping*/)
+		fmt.Fprintf(w, "    @property\n")
+		if pname == prop.Name {
+			fmt.Fprintf(w, "    @pulumi.getter\n")
+		} else {
+			fmt.Fprintf(w, "    @pulumi.getter(name=%q)\n", prop.Name)
+		}
```
komalali, line 628: Do we still need the decorator if we're emitting the function bodies?

justinvp: Yes. It's needed to provide the mapping between the Python name and Pulumi name.

justinvp:
> Can we lump the deprecation message and the setter under the same `if [blah] is not None:`?

Yeah, there's some opportunity for that, but I'm going to hold off on it for now. The deprecation messages aren't as common, so not too concerned about the extra check, but would definitely be cleaner to have a single check.

> Also, we have `kinesis_analytics` and `kinesisanalytics` and `kinesisanalyticsv2`... is that just an AWS funk or is something going on our end?

Pretty sure this is just AWS funk. This is from the `schema.json`:

```json
"kinesisAnalytics": {
    "type": "string",
    "deprecationMessage": "use `endpoints` configuration block `kinesisanalytics` argument instead"
},
"kinesisanalytics": {
    "type": "string"
},
"kinesisanalyticsv2": {
    "type": "string"
},
```

PR #5187, opened 2020-08-18
[codegen/docs] Link to Pulumi.Input-1.html when generating the doc link for the Pulumi.Input type

praneetloke:
Related to https://github.com/pulumi/docs/issues/3870.


mikhailshilkov (approved):
Looks like a reasonable workaround. Thanks for digging into this!

PR #5191, opened 2020-08-19
Added --suppress-permalink option

scott-the-programmer:
## Issue

https://github.com/pulumi/pulumi/issues/4103

We should allow the ability to suppress the permalink output entirely, regardless of the backend type.

## Intent of this PR
`pulumi up/preview/refresh/destroy` will now accept `--suppress-permalink` which will not display the state permalink in the log output. This supports all types of backends

This behavior follows that of `--suppress-output`

## Testing

* Added integration tests `TestPermaLinkSuppressed` and `TestPermaLink` to automatically check the output
* Tested locally by building the `cmd` project and invoking up/preview/refresh/destroy

```console
> pulumi up -y --suppress-permalink

Previewing update (sm-example):
     pulumi:pulumi:Stack  sm-example

Resources:
    2 unchanged

Updating (example-sticky):
     Type                 Name                       Status
     pulumi:pulumi:Stack  sm-example

Resources:
    2 unchanged

Duration: 4s
```

# ** INTEGRATION TEST FIX ** 

This PR is actually a second attempt - see the first attempt here https://github.com/pulumi/pulumi/pull/5177

In summary, an integration test was failing due to not providing an additional flag for previewing. The message it was matching was also incorrect ("PermaLink" instead of "Permalink"), which is how it passed the CI build first time around

```
		UpdateCommandlineFlags:  []string{"--suppress-permalink"},
		PreviewCommandlineFlags: []string{"--suppress-permalink"}, <---
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,10 @@ CHANGELOG
 
 - SDK changes to support Python input/output classes
   [#5033](https://github.com/pulumi/pulumi/pull/5033)
+  
+- Added `--suppress-permalink` option to suppress the permalink output
+  (fixes [#4103](https://github.com/pulumi/pulumi/issues/4103))
+  [#5191](https://github.com/pulumi/pulumi/pull/5191)
 
 ## 2.8.2 (2020-08-07)
 
--- a/pkg/backend/apply.go
+++ b/pkg/backend/apply.go
@@ -39,7 +39,7 @@ import (
 type ApplierOptions struct {
 	// DryRun indicates if the update should not change any resource state and instead just preview changes.
 	DryRun bool
-	// ShowLink indicates if a link to the update persisted result should be displayed.
+	// ShowLink indicates if a link to the update persisted result can be displayed.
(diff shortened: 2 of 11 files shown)


stack72:
Hi @scott-the-programmer 

Thanks so much for coming back around to this PR - just took this for a spin as follows:

> Pulumi SaaS
https://asciinema.org/a/t324Z7UeI5Rxps9QMDzTn5TYL

> S3 backend
https://asciinema.org/a/FB0KJWSpoyDCekQEU9zlkhoWQ

> local backend
https://asciinema.org/a/NG1yEM3avGvkWEUp3CxXmFTzR

I am running the acceptance tests locally right now to ensure there are no issues - when these are green, I will merge!

Thanks again

Paul

stack72:
Hi @scott-the-programmer 

the acceptance tests work a treat this time and the binary is looking good from the tests above! I am going to merge this locally and fixup the merge conflict

thanks for sending the PR 🥇 

Paul

Issue #5201, opened 2020-08-21
PyName: Remove Panic when all providers have been updated


justinvp:
We're adding an assert to `PyName` to help catch unintended breaking changes due to the `PyName` behavior change. It asserts `PyName(name) == PyNameLegacy(name)` unless the name is in an exclusion list.

Once all the existing providers have been updated to use this version of the codegen (or later), we can then go back and remove the assertion.

PR #5202, opened 2020-08-21
[codegen/python] Fix unintended name changes from PyName and some whitespace cleanup

justinvp:
The first commit cleans up some whitespace.

The second commit makes PyName panic if the result doesn't match PyNameLegacy(name), unless name is in an exclusion list. This is to help catch unintended breaking changes.

I've run these changes on AWS, Azure, AzureAD, GCP, Random, and Kubernetes.

I expect when we run on all the other providers, we may have to add more names to the `excludeFromPanic` set or `useLegacyName` set, depending on the situation.

Once all providers have been updated to use this version of the codegen (and we've added to `useLegacyName` and `excludeFromPanic` as appropriate), we can go back and remove the panic behavior along with the `excludeFromPanic` behavior. This is tracked by https://github.com/pulumi/pulumi/issues/5201

### DIFF
--- a/pkg/codegen/python/python.go
+++ b/pkg/codegen/python/python.go
@@ -15,6 +15,7 @@
 package python
 
 import (
+	"fmt"
 	"strings"
 	"unicode"
 	"unicode/utf8"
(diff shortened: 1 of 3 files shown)


pkg/codegen/python/python.go
```diff
@@ -15,16 +15,60 @@
...
 
+// useLegacyName are names that should return the result of PyNameLegacy from PyName, for compatibility.
+var useLegacyName = codegen.StringSet{
+	// The following property name of a nested type is a case where the newer algorithm produces an incorrect name
+	// (`open_xjson_ser_de`). It should be the legacy name of `open_x_json_ser_de`.
+	// TODO[pulumi/pulumi#5199]: We should see if we can fix this in the algorithm of PyName so it doesn't need to
+	// be special-cased in this set.
+	"openXJsonSerDe": struct{}{}, // AWS
+
+	// The following function name has already shipped with the legacy name (`get_public_i_ps`).
```
komalali, line 34: These notes are so helpful 🙌

PR #5217, opened 2020-08-25
[codegen/python] Add type annotation to resource_name arg

justinvp:
Missed type annotation on the `resource_name` argument 🤦.


justinvp:
Going to merge, happy to address feedback subsequently.

PR #5222, opened 2020-08-25
Increase visibility of console link

jkisk:
This PR will display a link to the Console at the beginning of an update instead of the end, and adds a link at the beginning of a dry run Preview.  The phrase "View Live" provides a call to action to go along with the link.

Before:
image

After:
image

### DIFF
--- a/pkg/backend/httpstate/backend.go
+++ b/pkg/backend/httpstate/backend.go
@@ -901,7 +901,7 @@ func (b *cloudBackend) apply(
 	if !(op.Opts.Display.JSONDisplay || op.Opts.Display.Type == display.DisplayWatch) {
 		// Print a banner so it's clear this is going to the cloud.
 		fmt.Printf(op.Opts.Display.Color.Colorize(
-			colors.SpecHeadline+"%s (%s):"+colors.Reset+"\n"), actionLabel, stack.Ref())
+			colors.SpecHeadline+"%s (%s)"+colors.Reset+"\n\n"), actionLabel, stack.Ref())
 	}
 
 	// Create an update object to persist results.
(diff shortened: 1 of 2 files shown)


lukehoban:
The GUID in the preview link is particularly unfortunate in this context. Something we can fix?

lukehoban:
From a grammar/UX perspective the two lines ending with `:` doesn’t feel quite right. Don’t have a specific suggestion, but I wonder if we can make this fit more “naturally”?

Cc also @justinvp for thoughts.

chrsmith:
I'll admit I'm not a fan of us showing the link three times once the update is completed. As we talked about during standup, if we were able to have the "link stay at the bottom" of the CLI output as it is dynamically generated as resources get updated, that would obviate the need for us to print the link at the "top" of an update once it starts, right? (Happy to hop on a VC to help troubleshoot any issues why that would be super-difficult to add to the `display` package.)

> The GUID in the preview link is particularly unfortunate in this context. Something we can fix?

Yes, definitely. At the very least, we can make it shorter, similar to how Git SHAs are typically abbreviated to 6-8 chars even though technically that isn't accurate. I'll file an issue to track that service-side. (But I assume it isn't something we'd want to gate this PR on.)

lukehoban:
> if we were able to have the "link stay at the bottom" of the CLI output as it is dynamically generated as resources get updated, that would obviate the need for us to print the link at the "top" of an update once it starts, right? 

I’m definitely open to UX proposals along these lines - but I’ll admit I’m a little skeptical this can work “well” given the rest of our current UX. 

I’d be more inclined personally to skip printing it again at the end and rely just on the clear and simple link at the beginning.

justinvp:
> From a grammar/UX perspective the two lines ending with `:` doesn’t feel quite right. Don’t have a specific suggestion, but I wonder if we can make this fit more “naturally”?

Are you talking about the "Previewing update (beverly):" and "Updating (beverly):" lines? Should we simply remove the trailing `:` from those lines?

Also, wondering if we should add some blank lines like:

```
Previewing update (beverly)

View in Console: <link>

    Type                 Name            Plan
    pulumi:pulumi:Stack  pacS3-beverly
```

And:

```
Updating (beverly)

View in Console: <link>

    Type                 Name            Status
    pulumi:pulumi:Stack  pacS3-beverly
```

Lastly, "View in Console" is fine, but I'm wondering if there are any other more enticing phrases? I don't have any great suggestions. Maybe "View Live:"?

jkisk:
Implementing @justinvp suggestions, along with @lukehoban suggestion to remove final link.

image

pkg/backend/httpstate/backend.go
```diff
@@ -912,26 +912,31 @@ func (b *cloudBackend) apply(
...
-			link = b.CloudConsoleURL(base, "previews", update.UpdateID)
-		}
-		if link != "" {
-			defer func() {
-				fmt.Printf(op.Opts.Display.Color.Colorize(
-					colors.SpecHeadline+"Permalink: "+
-						colors.Underline+colors.BrightBlue+"%s"+colors.Reset+"\n"), link)
-			}()
-		}
+		// Print a URL at the beginning and end of the update pointing to the Pulumi Service.
```
justinvp, line 915: Should "and end" be removed?

```suggestion
		// Print a URL at the beginning of the update pointing to the Pulumi Service.
```

jkisk: yes, thank you.

PR #5238, opened 2020-08-27
Avoid raising unexpected type errors with `Any`

### LINKED ISSUE #5237: Python: Unexpected type error raised for outputs annotated as `Any`
The Terraform provider has a [resource declared as](https://github.com/pulumi/pulumi-terraform/blob/38c9e75e508b65cdaef714fe68478ddac4ddeab4/sdk/python/pulumi_terraform/state/remote_state_reference.py#L575-L584):

```python
class RemoteStateReference(pulumi.CustomResource):
    outputs: pulumi.Output[Mapping[str, Any]]
```

The value of the `Mapping` is specified as `Any`.

This results in an `AssertionError` being raised with the message `Unexpected type. Expected 'list' got 'typing.Any'` when a value comes back that is a `list`. The same error would be raised if a value came back that was a `dict`.

The problem is we're not taking `Any` into account before raising this error. If something is typed as `Any` we should treat it as if we have no type information, since it could be anything.

justinvp:
Avoid raising an `AssertionError` due to unexpected types when a type is annotated as `Any`.

Fixes #5237

### DIFF
--- a/sdk/python/lib/pulumi/runtime/rpc.py
+++ b/sdk/python/lib/pulumi/runtime/rpc.py
@@ -399,6 +399,11 @@ def translate_output_properties(output: Any,
     # Unwrap optional types.
     typ = _types.unwrap_optional_type(typ) if typ else typ
 
+    # If the typ is Any, set it to None to treat it as if we don't have any type information,
+    # to avoid raising errors about unexpected types, since it could be any type.
+    if typ is Any:
+        typ = None
+
     if isinstance(output, dict):
         # Function called to lookup a type for a given key.
         # The default always returns None.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,10 @@ CHANGELOG
 =========
 
 ## HEAD (Unreleased)
-_(none)
+
+- Python SDK: Avoid raising an error when an output has a type annotation of Any
+  and the value is a list or dict.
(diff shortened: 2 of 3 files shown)


sdk/python/lib/pulumi/runtime/rpc.py
```diff
@@ -399,6 +399,11 @@ def translate_output_properties(output: Any,
     # Unwrap optional types.
     typ = _types.unwrap_optional_type(typ) if typ else typ
 
+    # If the typ is Any, set it to None to treat it as if we don't have any type information,
+    # to avoid raising errors about unexpected types, since it could be any type.
+    if typ is Any:
+        typ = None
```
komalali, line 405: wow python is weird. so it expects a match? only `Any == Any`? `List` does not qualify as `Any`?

justinvp: Not a Python oddity, we're raising the error:

https://github.com/pulumi/pulumi/blob/910aa96016cd325f4ad5b808ef5075ad68699544/sdk/python/lib/pulumi/runtime/rpc.py#L440-L451

We just weren't taking into account that `typ` may be `Any`, in which case, we shouldn't raise the error.

komalali: ah got it thank you!

Issue #5214, opened 2020-08-24
[codegen/python] Properties named `property` cause errors for subsequent properties


justinvp:
We ran into an issue [adopting](https://github.com/pulumi/pulumi-akamai/pull/4) the new Python improvements in the Akamai provider.

The `PropertyActivation` resource has an output property named `property`. Previously, this wasn't an issue as outputs were stored on resources as instance variables. But now they're exposed as proper Python properties defined like:

```python
    @property
    @pulumi.getter
    def property(self) -> str:
        return pulumi.get(self, "property")

    @property
    @pulumi.getter
    def status(self) -> str:
        return pulumi.get(self, "status")

    @property
    @pulumi.getter
    def version(self) -> Optional[float]:
        return pulumi.get(self, "version")
```

The `@property` is referring to the built-in property decorator function used to declare Python properties.

But then we're redeclaring our own `def property(self)` function, so the next two calls to `@property` for the next two properties (`status` and `version`) are using our redefinition of `property` rather than the built-in `property`.

We'll either need to always fully qualify `@builtins.property` (importing `builtins`) instead of just `@property`, or only do that when there is a property named `property`. Though, will have to check if there are any possibilities of conflicts with `builtins`. Or another possibility is to simply emit any property named `property` last.

PR #5247, opened 2020-08-28
[codegen/python] Emit properties named "property" last

justinvp:
This avoids conflicting with the built-in `@property` decorator function for other properties.

Fixes https://github.com/pulumi/pulumi/issues/5214


mikhailshilkov (approved):
Thanks!

Issue #5250, opened 2020-08-28
Python SDK: Error during preview when setting tags on AWS resources


justinvp:
A couple users have reported this issue. Here's a simple repro:

```python
from pulumi_aws import ec2
vpc = ec2.Vpc(
    'eks-vpc',
    cidr_block='10.100.0.0/16',
    instance_tenancy='default',
    enable_dns_hostnames=True,
    enable_dns_support=True,
    tags={
        'Name' : 'pulumi-eks-vpc'
    }
)
```

This results in an error during preview:

```
$ pulumi up
Previewing update (test):
     Type                 Name             Plan       Info
 +   pulumi:pulumi:Stack  kubernetes-test  create     3 errors; 37 messages
 +   └─ aws:ec2:Vpc       eks-vpc-2        create
Diagnostics:
  pulumi:pulumi:Stack (kubernetes-test):
    error: Program failed with an unhandled exception:
    error: Traceback (most recent call last):
      File "/usr/local/bin/pulumi-language-python-exec", line 85, in <module>
        loop.run_until_complete(coro)
      File "/usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8/lib/python3.8/asyncio/base_events.py", line 616, in run_until_complete
        return future.result()
      File "/Users/<redacted>/venv/lib/python3.8/site-packages/pulumi/runtime/stack.py", line 83, in run_in_stack
        await run_pulumi_func(lambda: Stack(func))
      File "/Users/<redacted>/venv/lib/python3.8/site-packages/pulumi/runtime/stack.py", line 51, in run_pulumi_func
        await RPC_MANAGER.rpcs.pop()
      File "/Users/<redacted>/venv/lib/python3.8/site-packages/pulumi/runtime/rpc_manager.py", line 67, in rpc_wrapper
        result = await rpc
      File "/Users/<redacted>/venv/lib/python3.8/site-packages/pulumi/runtime/resource.py", line 427, in do_register
        await rpc.resolve_outputs(res, resolver.serialized_props, resp.object, resolvers)
      File "/Users/<redacted>/lib/python3.8/site-packages/pulumi/runtime/rpc.py", line 487, in resolve_outputs
        translated_value = translate_output_properties(value, res.translate_output_property, types.get(key))
      File "/Users/<redacted>/lib/python3.8/site-packages/pulumi/runtime/rpc.py", line 434, in translate_output_properties
        return {
[...]

PR #5251, opened 2020-08-28
Python: Ignore internal properties when unmarshaling

justinvp:
Fixes https://github.com/pulumi/pulumi/issues/5250

We recently added some checks to Python that will error if the type annotations don't match the types of the output values from the engine. The engine sometimes returns "internal" keys amongst the values, which cause the error to be raised because the types do not match.

For example, we may have a output property annotated as `Mapped[str, str]`, but during preview the value from the engine is `{"foo": "bar", "__defaults": []}`. The map contains `"__defaults"` with a value of `[]` and `list` and `str` don't match, so we raise the error.

This addresses the issue by filtering out these "internal" properties from the engine, like we do in other languages (e.g. [.NET](https://github.com/pulumi/pulumi/pull/3560), [Go](https://github.com/pulumi/pulumi/pull/3996)).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,10 @@ CHANGELOG
 =========
 
 ## HEAD (Unreleased)
-_(none)_
+
+- Python SDK: Avoid raising an error when internal properties don't match the
+  expected type.
+  [#5251](https://github.com/pulumi/pulumi/pull/5251)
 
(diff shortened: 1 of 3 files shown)


justinvp:
(Included @mikhailshilkov and @EvanBoyle as reviewers as they may recall the similar changes for .NET and Go)

Issue #5054, opened 2020-07-27
[Automation API] expose execution kind via update metadata


EvanBoyle:
We should be able to distinguish between CLI driven, and the two flavors of automation API-driven execution (local and inline src) via metadata. This metadata can be consumed by various backends to add additional context to the operation that was performed.

PR #5258, opened 2020-08-29
expose execution kind via update metadata

EvanBoyle:
Fixes https://github.com/pulumi/pulumi/issues/5054

Adds a new field to UpdateMetadata.Environment `exec.kind` which is set to one of `auto.local` for traditional programs driven by automation API, `cli` for traditional programs that don't use automation API, or `auto.inline` for automation API programs using inline source (host most). 

This field can be picked up by various backends to expose more information about the source of an update (Was this deployed by hand, as a part of an automation API deployment in CI, etc.). 

Implemented as a string flag so that there's the flexibility to expand in the future.

### DIFF
--- a/pkg/cmd/pulumi/destroy.go
+++ b/pkg/cmd/pulumi/destroy.go
@@ -34,6 +34,7 @@ func newDestroyCmd() *cobra.Command {
 	var stack string
 
 	var message string
+	var execKind string
 
 	// Flags for engine.UpdateOptions.
 	var diffDisplay bool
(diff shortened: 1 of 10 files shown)


pkg/cmd/pulumi/destroy.go
```diff
@@ -210,5 +211,11 @@ func newDestroyCmd() *cobra.Command {
 			&eventLogPath, "event-log", "",
 			"Log events to a file at this path")
 	}
+
+	// internal flag
+	cmd.PersistentFlags().StringVar(&execKind, "exec-kind", "", "")
+	// ignore err, only happens if flag does not exist
+	_ = cmd.PersistentFlags().MarkHidden("exec-kind")
```
lukehoban, line 218: Any benefit to making this a flag vs just an env var?

Seems an env var might offer more flexibility?

Also - any reason to constrain the values vs just allow consuming code to set this to whatever they want?

EvanBoyle: This is not an attempt to design an extensible feature. The flag (implementation details) is hidden and can always be swapped out in the future if we get feedback from console users that they'd like to customize this value.

PR #5271, opened 2020-09-01
add build tags to all integration tests

jaxxstorm:
This PR separates all the SDK/language integration tests into their own files and adds build tags to the files.

The purpose of this is so that we can run the tests for each language in separate build pipelines, for example:

`go test -tags=go` will run all tests that are Go based examples.

I've added two tags to each tests file:

- the language specific tag
- an `all` tag

In addition to this, I've updated the `TEST_FAST` and `TEST_ALL` makefile commands to include the `all` tag so that the tests run as normal in travis and locally.


stack72 (approved):
I love this! :D Such an important pre-cursor

Issue #5252, opened 2020-08-29
[codegen/nodejs] New Resource constructor overload breaks PaC type inference


justinvp:
These policies used to work:

```typescript
import * as k8s from "@pulumi/kubernetes";
import { PolicyPack, validateResourceOfType } from "@pulumi/policy";
​
// tslint:disable-next-line: no-unused-expression
new PolicyPack("k8s", {
    policies: [
        {
            name: "replica-count",
            description: "Consider more replicas for high-availability.",
            enforcementLevel: "advisory",
            validateResource: validateResourceOfType(k8s.apps.v1.Deployment, (resource, _, reportViolation) => {
                if (resource.spec!.replicas! <= 1) {
                    reportViolation("Replica count should be greater than 1.");
                }
            }),
        },
        {
            name: "non-secure-http",
            description: "External services should expose via HTTPS.",
            enforcementLevel: "advisory",
            validateResource: validateResourceOfType(k8s.core.v1.Service, (resource, _, reportViolation) => {
                resource.spec!.ports?.forEach(specPort => {
                    if (specPort.port !== 443) {
                        reportViolation(`Service port [${specPort.port}] should be exposed via 443.`);
                    }
                });
            }),
        },
        {
            name: "pin-image-versions",
            description: "Images should be pinned to a specific version",
            enforcementLevel: "mandatory",
            validateResource: validateResourceOfType(k8s.apps.v1.Deployment, (resource, _, reportViolation) => {
                resource.spec!.template!.spec!.containers.forEach(it => {
                    if (it?.image?.endsWith(":latest")) {
                        reportViolation("Image version should not be 'latest'.");
                    }
                });
            }),
        },
    ],
```

But now when the policy pack references `@pulumi/kubernetes` version 2.5.0, the following errors occur during `pulumi up`:

```
    error: [runtime] Running program '/Users/clstokes/.pulumi/policies/demo/pulumi-analyzer-k8s-v0.1.20200825-5' failed with an unhandled exception:
        TSError: ⨯ Unable to compile TypeScript:
    index.ts(12,30): error TS2339: Property 'spec' does not exist on type 'never'.
    index.ts(22,26): error TS2339: Property 'spec' does not exist on type 'never'.
    index.ts(34,26): error TS2339: Property 'spec' does not exist on type 'never'.
[...]

PR #5274, opened 2020-09-02
[codegen/nodejs] Remove unnecessary ctor overload

justinvp:
An extra constructor overload was recently added to pass undefined state from `get` for resources that do not have any state inputs (notably Kubernetes resources). This ended up breaking PaC's `validateResourceOfType`, which relies on type inference in common usage to determine the resource's args type based on the signature of the constructor.

This constructor overload isn't necessary. Instead, we can remove it and modify how the constructor is called inside `get`. This also makes it so we're not exposing details about `get`'s implementation in the public API.

To make it easier to review, [here's the diff](https://github.com/pulumi/pulumi-kubernetes/commit/87274916427e5f309029a8e562a1fe5d9fa5485d) of just these changes for Kubernetes against [Levi's PR](https://github.com/pulumi/pulumi-kubernetes/pull/1285).

Part of https://github.com/pulumi/pulumi/issues/5252

### DIFF
--- a/pkg/codegen/nodejs/gen.go
+++ b/pkg/codegen/nodejs/gen.go
@@ -388,7 +388,7 @@ func (mod *modContext) genResource(w io.Writer, r *schema.Resource) error {
 		fmt.Fprintf(w, "     * @param opts Optional settings to control the behavior of the CustomResource.\n")
 		fmt.Fprintf(w, "     */\n")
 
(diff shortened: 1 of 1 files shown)


justinvp:
@lblackstone, if this looks good, can we consider adopting this in the Kubernetes provider and including it as part of the release?

PR #5275, opened 2020-09-02
feat(autoapi): add workspace scoped envvars to LocalWorkspace and Stack

### LINKED ISSUE #5227: Add Workspace scoped env vars
Sometimes pulumi programs need environment variables for backend auth, pulumi SaaS auth, cloud auth etc. We should add the following methods to Workspace to allow for workspace scoped env vars:

```go
func (w *Workspace) GetEnvVars(ctx context.Context) ([]string, error) {}
func (w *Workspace) SetEvnVars(ctx context.Context) error{}
```
These new methods will need to be implemented for `LocalWorkspace`.

`LocalWorkspace.runPulumiCmdSync` will need to be updated: https://github.com/pulumi/pulumi/blob/c23f26b3357367c5f6e00c984b4bbbfc5cdd4fd2/sdk/go/x/auto/local_workspace.go#L404-L414

As well as `Stack.runPulumiCmdSync`: https://github.com/pulumi/pulumi/blob/741485d7d07735d457c38c330d60a7e49e9d94af/sdk/go/x/auto/stack.go#L629-L649

Let's try to start with this and see if we can get by without Stack scoped environment variables as well. If it turns out we need to go down that route let's open another is.

metral:
Closes #5227

### DIFF
--- a/sdk/go/x/auto/local_workspace.go
+++ b/sdk/go/x/auto/local_workspace.go
@@ -42,6 +42,7 @@ type LocalWorkspace struct {
 	workDir    string
 	pulumiHome string
 	program    pulumi.RunFunc
+	envvars    map[string]string
 }
 
 var settingsExtensions = []string{".yaml", ".yml", ".json"}
@@ -244,6 +245,47 @@ func (l *LocalWorkspace) RefreshConfig(ctx context.Context, fqsn string) (Config
 	return cfg, nil
 }
 
+// GetEnvVars returns the environment values scoped to the current workspace.
+func (l *LocalWorkspace) GetEnvVars() map[string]string {
+	if l.envvars == nil {
+		return nil
+	}
+	return l.envvars
+}
+
+// SetEnvVars sets the specified map of environment values scoped to the current workspace.
+// These values will be passed to all Workspace and Stack level commands.
+func (l *LocalWorkspace) SetEnvVars(envvars map[string]string) error {
+	if envvars == nil {
+		return errors.New("unable to set nil environment values")
+	}
+	if l.envvars == nil {
+		l.envvars = map[string]string{}
+	}
+	for k, v := range envvars {
+		l.envvars[k] = v
+	}
+	return nil
+}
+
+// SetEnvVar sets the specified environment value scoped to the current workspace.
+// This value will be passed to all Workspace and Stack level commands.
+func (l *LocalWorkspace) SetEnvVar(key, value string) {
+	if l.envvars == nil {
+		l.envvars = map[string]string{}
+	}
+	l.envvars[key] = value
+}
+
+// UnsetEnvVar unsets the specified environment value scoped to the current workspace.
+// This value will be removed from all Workspace and Stack level commands.
+func (l *LocalWorkspace) UnsetEnvVar(key string) {
+	if l.envvars == nil {
+		return
+	}
+	delete(l.envvars, key)
+}
+
 // WorkDir returns the working directory to run Pulumi CLI commands.
 // LocalWorkspace expects that this directory contains a Pulumi.yaml file.
 // For "Inline" Pulumi programs created from NewStackInlineSource, a Pulumi.yaml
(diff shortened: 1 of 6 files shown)


sdk/go/x/auto/local_workspace.go
```diff
@@ -244,6 +245,37 @@ func (l *LocalWorkspace) RefreshConfig(ctx context.Context, fqsn string) (Config
...
 
+// GetEnvVars returns the environment values scoped to the current workspace.
+func (l *LocalWorkspace) GetEnvVars() ([]string, error) {
+	if l.envvars == nil {
+		return nil, errors.New("unable to get environment values")
+	}
+	return l.envvars, nil
+}
+
+// SetEnvVars sets the specified map of environment values scoped to the current workspace.
```
EvanBoyle, line 256: The comment here mentions a map, but the function parameter is an array. I wonder if accepting a map would actually be a good idea? If we use a map we can offer an `Unsetenv` API which would be consistent with the os package: https://golang.org/pkg/os/#Unsetenv

A map might also be easier for users to manipulate as it doesn't require parsing out key-value pairs from the array.

sdk/go/x/auto/local_workspace.go
```diff
@@ -244,6 +245,37 @@ func (l *LocalWorkspace) RefreshConfig(ctx context.Context, fqsn string) (Config
 	return cfg, nil
 }
 
+// GetEnvVars returns the environment values scoped to the current workspace.
+func (l *LocalWorkspace) GetEnvVars() ([]string, error) {
+	if l.envvars == nil {
+		return nil, errors.New("unable to get environment values")
```
EvanBoyle, line 251: I wonder if we could avoid having error return values for any of these functions? If we have an empty environment variable perhaps we can just return `[]string{}`

EvanBoyle (commented):
Two suggestions to consider but overall looks great!

sdk/go/x/auto/local_workspace.go
EvanBoyle: A note on the doc comments. It might be worth mentioning that these values will be passed to all `Workspace` and `Stack` level commands like `Up`, `ListStacks`, `SetConfig` etc.

EvanBoyle:
Would also be great to see some examples for the documentation side of things: https://github.com/pulumi/pulumi/blob/c23f26b3357367c5f6e00c984b4bbbfc5cdd4fd2/sdk/go/x/auto/example_test.go#L797

sdk/go/x/auto/local_workspace.go
metral: >The comment here mentions a map, but the function parameter is an array. 
>The second comment refers to using a map[string]string as the underlying data structure to store env vars as well.

Originally `map[string]string` was used, hence the left over comments :).

We switched to a `[]string` part way through since this is the type used in the `cmd.Env` and what the issue originally called out for the API, but happy to switch it to a map.

> If we use a map we can offer an Unsetenv API which would be consistent with the os package: https://golang.org/pkg/os/#Unsetenv

Fair point. Though I wonder how often it will be used in these scenarios. Thoughts?

> A note on the doc comments. It might be worth mentioning that these values will be passed to all Workspace and Stack level commands like Up, ListStacks, SetConfig etc.

👍

sdk/go/x/auto/local_workspace.go
metral: Dropping the error works with me

sdk/go/x/auto/local_workspace.go
EvanBoyle: Ah, Indeed I did put that in the original issue. Sounds like you had the right intuition from the get-go, but I needed a minute to think about it :) 

> Fair point. Though I wonder how often it will be used in these scenarios. Thoughts?

I think that if we want to allow workspaces to be stateful and reusable, mutating the environment variables in some way would probably be helpful. A single Unsetenv method is probably enough to satisfy this, but I'll leave it up to you.

EvanBoyle (approved):
Implementation and examples look great! 

If I had to guess, tests are failing due to the uninitialized map. Probably just need to initialize it here in `NewLocalWorkspace`:

https://github.com/pulumi/pulumi/blob/dec6a6a49a24e5f015a2f9b7a06bfa43a6a12ac4/sdk/go/x/auto/local_workspace.go#L455

PR #5276, opened 2020-09-02
add initial pull-request workflow

jaxxstorm:
This pull request switches the pull request validation process from Travis to GitHub actions.

Some notable changes here:

- the makefile target for `travis_pull_request` has been emptied, so the travis run still executes but effectively does nothing
- some tests for the auto api have been disabled after conversation with @EvanBoyle see https://github.com/pulumi/pulumi/issues/5303
- one test that detects the travis branch has been disabled (for now) see https://github.com/pulumi/pulumi/issues/5303
- some golangci-lint stuff was ignored: https://github.com/pulumi/pulumi/issues/5304

Other than that, this is funtionally equivalent to the `travis_pull_request` make target and runs the same test suite.

The intention here is to implement something that an be iterated on. Some things that have not been implemented in this first run:

- output to slack (will come shortly) in a subsequent PR
- git tags/releases
- master branches
- matrix builds


EvanBoyle:
Where do the SDK tests run? I'm having trouble finding the output.

jaxxstorm:
@EvanBoyle good catch, looks like the SDK tests aren't being captured by the `test_all` make target, I'll fix that up

jaxxstorm:
changed this back to a draft

pkg/cmd/pulumi/util_test.go
```diff
@@ -177,16 +177,19 @@ func TestReadingGitRepo(t *testing.T) {
...
 	{
 		test := &backend.UpdateMetadata{
 			Environment: make(map[string]string),
 		}
 		assert.NoError(t, addGitMetadata(e.RootPath, test))
 		name, ok := test.Environment[backend.GitHeadName]
+		t.Log(name)
 		assert.True(t, ok, "Expected 'git.headName' key, from CI util.")
-		assert.Equal(t, "branch-from-ci", name)
+		// assert.Equal(t, "branch-from-ci", name)
```
EvanBoyle, line 190: Is this left from dev, or should it be removed?

sdk/dotnet/Makefile
```diff
@@ -49,8 +49,6 @@ install:: build install_plugin
 dotnet_test:: install
 	# include the version prefix/suffix to avoid generating a separate nupkg file
 	dotnet test /p:VersionPrefix=${VERSION_PREFIX} /p:VersionSuffix=${VERSION_SUFFIX}
-	# work around a bug where running `dotnet test` disables console stdinput display
-	stty echo
```
EvanBoyle, line 53: Any reason why this was removed? This was added to workaround a bug where stdinput is all messed up after running the dotnet tests.

sdk/go/x/auto/errors_test.go
```diff
@@ -27,6 +27,7 @@ import (
 )
 
 func TestConcurrentUpdateError(t *testing.T) {
+	t.Skip("disabled")
```
EvanBoyle, line 30: Why was this disabled?

If this is really necessary can we open an issue and link it here?

sdk/nodejs/tests/unwrap.spec.ts
```diff
@@ -51,7 +51,7 @@ function testResources(val: any, expected: any, resources: TestResource[], allRe
         assert.deepStrictEqual(asyncResources, new Set(allResources));
 
         for (const res of syncResources) {
-            if (!asyncResources.has(<TestResource>res)) {
+            if (!asyncResources.has(<Resource><any>res)) {
```
EvanBoyle, line 54: What is the intention with this type assertion? With two type assertions here isn't this just a cast to `any`?

EvanBoyle (commented):
A few clarifications, but structurally looks good to me.

sdk/dotnet/Makefile
jaxxstorm: yeah, it doesn't actually work. It's related to https://github.com/actions/runner/issues/241
I can wrap it in an if statement if needed, but I don't actually see this bug anymore so I think it's fixed in later versions of the dotnet sdk.

sdk/nodejs/tests/unwrap.spec.ts
jaxxstorm: This is related to https://github.com/pulumi/pulumi/pull/5311

sdk/dotnet/Makefile
EvanBoyle: Sounds good, we can always come back if we notice the issue after this gets merged.

stack72 (approved):
I think if we have a green build here we should proceed and then keep iterating on it as we see any potential issues - this is only for PRs so we will see if there are problems pretty fast - there is a huge surface area here so we may miss something minor on the first pass but it's something we will continually iterate on

EvanBoyle (approved):

Issue #5184, opened 2020-08-18
generatePythonPropertyCaseMaps slows down doc generation


mikhailshilkov:
For a new large-ish provider API reference docs generation takes ~25 minutes. Apparently, if I remove [this line](https://github.com/pulumi/pulumi/blob/f8d929d197b3cc5b4a8dbac5bf207587e93fcb2c/pkg/codegen/docs/gen.go#L1781) with Python casing mapping, the time drops to 1 minute.

stack72:
FWIW, I am greater than 2 hours now to regenerate the GCP provider...

mikhailshilkov:
@pgavlin Do you remember why we call `pkg.ImportLanguages` for every property in https://github.com/pulumi/pulumi/blob/f8d929d197b3cc5b4a8dbac5bf207587e93fcb2c/pkg/codegen/python/doc.go#L135 ?

Isn't this single import enough https://github.com/pulumi/pulumi/blob/master/pkg/codegen/docs/gen.go#L1757 ?

mikhailshilkov:
Yes, that's the problem:
![image](https://user-images.githubusercontent.com/1454008/90526145-53fe8f80-e170-11ea-89a8-edc697112ca0.png)

pgavlin:
> @pgavlin Do you remember why we call `pkg.ImportLanguages` for every property in https://github.com/pulumi/pulumi/blob/f8d929d197b3cc5b4a8dbac5bf207587e93fcb2c/pkg/codegen/python/doc.go#L135 ?
> 
> Isn't this single import enough https://github.com/pulumi/pulumi/blob/master/pkg/codegen/docs/gen.go#L1757 ?

The code that's in there is the equivalent of the original code prior to the `ImportLanguages` API. Because the first function you're referencing is exported, I didn't want to make a behavioral change at that time. I think it's fine to do so now.

Do we know why this slowed down so much recently? Or has this been slow since that port?

mikhailshilkov:
> Do we know why this slowed down so much recently? Or has this been slow since that port?

I don't know. @stack72 do you remember a moment when generation started being slow, if there was one?

PR #5277, opened 2020-09-02
Speed up doc generation by avoiding repeated python language import

viveklak:
Fixes https://github.com/pulumi/pulumi/issues/5184

Repeated imports seem to really slow down doc generation as discovered by @mikhailshilkov in associated ticket.

### DIFF
--- a/pkg/codegen/python/doc.go
+++ b/pkg/codegen/python/doc.go
@@ -160,9 +160,11 @@ func (d DocLanguageHelper) GetResourceFunctionResultName(modName string, f *sche
 
 // GenPropertyCaseMap generates the case maps for a property.
 func (d DocLanguageHelper) GenPropertyCaseMap(pkg *schema.Package, modName, tool string, prop *schema.Property, snakeCaseToCamelCase, camelCaseToSnakeCase map[string]string, seenTypes codegen.Set) {
-	if err := pkg.ImportLanguages(map[string]schema.Language{"python": Importer}); err != nil {
-		fmt.Printf("error building case map for %q in module %q", prop.Name, modName)
-		return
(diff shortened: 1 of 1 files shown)


pkg/codegen/python/doc.go
```diff
@@ -160,9 +160,11 @@ func (d DocLanguageHelper) GetResourceFunctionResultName(modName string, f *sche
 
 // GenPropertyCaseMap generates the case maps for a property.
 func (d DocLanguageHelper) GenPropertyCaseMap(pkg *schema.Package, modName, tool string, prop *schema.Property, snakeCaseToCamelCase, camelCaseToSnakeCase map[string]string, seenTypes codegen.Set) {
-	if err := pkg.ImportLanguages(map[string]schema.Language{"python": Importer}); err != nil {
-		fmt.Printf("error building case map for %q in module %q", prop.Name, modName)
-		return
+	if _, imported := pkg.Language["python"]; !imported {
+		if err := pkg.ImportLanguages(map[string]schema.Language{"python": Importer}); err != nil {
```
mikhailshilkov, line 164: I believe in practice this will never hit because we import those [in advance](https://github.com/pulumi/pulumi/blob/dec6a6a49a24e5f015a2f9b7a06bfa43a6a12ac4/pkg/codegen/docs/gen.go#L1774).

Curious if that's true and if you keep this block because `GenPropertyCaseMap` is exported and we avoid some theoretic breaking change.

viveklak: Yeah I was being conservative because of it being exported.

mikhailshilkov: @pgavlin Do you think we should remove it?

PR #5279, opened 2020-09-02
refactor(autoapi-gitrepo): use Workspace in SetupFn callback

### LINKED ISSUE #5266: Refactor GitRepo.Setup callback to pass in entire Workspace
Currently, this callback passes in just the `WorkDir`: https://github.com/pulumi/pulumi/blob/c23f26b3357367c5f6e00c984b4bbbfc5cdd4fd2/sdk/go/x/auto/local_workspace.go#L523
This isn't enough information for some use cases. The kubernetes operator needs ProjectSettings from the workspace to know what kind of runtime it's working with and how to install the source. 

We should refactor the SetupFn type to pass in a Workspace instead of a string. 

Currently the repo setup is done in a utility function: https://github.com/pulumi/pulumi/blob/cf247c1245886ce62afff2ec9e4c80fc0ab40773/sdk/go/x/auto/git.go#L65

it will probably need to be pulled out of that method and into the last step of `NewLocalWorkspace`, to include the workspace object that has otherwise been setup: https://github.com/pulumi/pulumi/blob/c23f26b3357367c5f6e00c984b4bbbfc5cdd4fd2/sdk/go/x/auto/local_workspace.go#L473

metral:
Closes #5266

### DIFF
--- a/sdk/go/x/auto/local_workspace.go
+++ b/sdk/go/x/auto/local_workspace.go
@@ -470,6 +470,14 @@ func NewLocalWorkspace(ctx context.Context, opts ...LocalWorkspaceOption) (Works
 		}
 	}
 
+	// setup
+	if lwOpts.Repo != nil && lwOpts.Repo.Setup != nil {
+		err := lwOpts.Repo.Setup(ctx, l)
+		if err != nil {
+			return nil, errors.Wrap(err, "error while running setup function")
+		}
+	}
+
 	return l, nil
 }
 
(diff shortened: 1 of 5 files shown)


sdk/go/x/auto/local_workspace.go
```diff
@@ -520,7 +528,7 @@ type GitRepo struct {
 
 // SetupFn is a function to execute after enlisting in a git repo.
 // It is called with a PATH containing the pulumi program post-enlistment.
```
EvanBoyle, line 530: Can we update the docstring to indicate that this receives the full workspace after all other options have been processed (rather than the path).

EvanBoyle (approved):
Looks great!

Looks like the only thing left is to update the examples: https://github.com/pulumi/pulumi/blob/c23f26b3357367c5f6e00c984b4bbbfc5cdd4fd2/sdk/go/x/auto/example_test.go#L170

EvanBoyle (approved):
🚢

Issue #5278, opened 2020-09-02
Python type hints in providers use List[T] for arguments, not Sequence[T]


csssuf:
The generated Python type hints for providers use `typing.List[T]` to represent lists of items in constructors/arguments, rather than `typing.Sequence[T]`. In particular, this is a problem when `T` is an Input, since `typing.List` is invariant over T, and Input is just a union under the hood. This means that many calls which should pass typechecking do not.

## Expected Behavior
Running `mypy` against Pulumi programs which pass lists of inputs into Pulumi providers passes (given that those inputs are the correct type, of course.)

## Current Behavior
mypy reports that the lists are of the incorrect type.

## Steps to Reproduce
Here's a simple Python program which reproduces the behavior:
```python
from typing import List, Sequence

import pulumi


def accepts_list(lst: List[pulumi.Input[str]]):
    pass


def accepts_sequence(lst: Sequence[pulumi.Input[str]]):
    pass


lst = ["string1", "string2"]
accepts_list(lst)
accepts_sequence(lst)
```

Running mypy against this produces the following:
```
list_test.py:15: error: Argument 1 to "accepts_list" has incompatible type "List[str]"; expected "List[Union[str, Awaitable[str], Output[str]]]"
list_test.py:15: note: "List" is invariant -- see http://mypy.readthedocs.io/en/latest/common_issues.html#variance
list_test.py:15: note: Consider using "Sequence" instead, which is covariant
Found 1 error in 1 file (checked 1 source file)
```

## Context (Environment)
We have Pulumi programs which we have already been typechecking via mypy without type hints in the provider libraries. Now that the provider libraries have these incorrect type hints, mypy flags many new type errors in our Pulumi programs. This can be worked around by explicitly annotating the lists are `List[pulumi.Input[...]]`, but this is quite tedious, especially for types with deeper nesting (via a real example from our pulumi_kubernetes codebase): 
```python
role_arg_rules = [rbac_rule_to_pulumi_arg(rule) for rule in rules]  # type: List[pulumi.Input[pulumi.InputType[PolicyRuleArgs]]]
```

justinvp:
Thanks for the suggestion, @csssuf! Makes sense, we should definitely move to `Sequence`. We use `Mapping` rather than `Dict` for similar reasons.

justinvp:
The changes for this have been merged and will roll out as the providers are released.

csssuf:
Thanks for the quick fix!

PR #5282, opened 2020-09-03
[sdk/python] Support Sequence[T] for array types rather than List[T]

justinvp:
We currently emit array types as `List[T]` for Python, but `List[T]` is invariant, which causes type checkers like mypy to produce errors when values like `["foo", "bar"]` are passed as args typed as `List[pulumi.Input[str]]` (since `Input[str]` is an alias for `Union[T, Awaitable[T], Output[T]]`. To address this, we should move to using [`Sequence[T]`](https://docs.python.org/3/library/typing.html#typing.Sequence) which is covariant, and does not have this problem.

We actually already do this for `Dict` vs. `Mapping`, emitting map types as `Mapping[str, T]` rather than `Dict[str, T]` because `Mapping[str, T]` is covariant for the value. This change makes us consistent for array types.

These are the SDK changes necessary to support `Sequence[T]`.

Codegen PR: https://github.com/pulumi/pulumi/pull/5283

Part of https://github.com/pulumi/pulumi/issues/5278

### DIFF
--- a/sdk/python/lib/pulumi/runtime/rpc.py
+++ b/sdk/python/lib/pulumi/runtime/rpc.py
@@ -17,10 +17,10 @@ out of RPC calls.
 """
 import sys
 import asyncio
-import collections
+from collections import abc
 import functools
 import inspect
-from typing import List, Any, Callable, Dict, Mapping, Optional, Tuple, Union, TYPE_CHECKING, cast, get_type_hints
+from typing import List, Any, Callable, Dict, Mapping, Optional, Sequence, TYPE_CHECKING, cast
 
 from google.protobuf import struct_pb2
 import six
@@ -92,7 +92,9 @@ async def serialize_property(value: 'Input[Any]',
     Serializes a single Input into a form suitable for remoting to the engine, awaiting
     any futures required to do so.
     """
-    if isinstance(value, list):
+    # Exclude some built-in types that are instances of Sequence that we don't want to treat as sequences here.
+    # From: https://github.com/python/cpython/blob/master/Lib/_collections_abc.py
+    if isinstance(value, abc.Sequence) and not isinstance(value, (tuple, str, range, memoryview, bytes, bytearray)):
         props = []
         for elem in value:
             props.append(await serialize_property(elem, deps, input_transformer))
@@ -125,7 +127,7 @@ async def serialize_property(value: 'Input[Any]',
             remote_asset = cast('RemoteAsset', value)
             obj["uri"] = await serialize_property(remote_asset.uri, deps, input_transformer)
         else:
-            raise AssertionError(f"unknown asset type: {value}")
+            raise AssertionError(f"unknown asset type: {value!r}")
 
         return obj
 
@@ -147,7 +149,7 @@ async def serialize_property(value: 'Input[Any]',
             remote_archive = cast('RemoteArchive', value)
             obj["uri"] = await serialize_property(remote_archive.uri, deps, input_transformer)
         else:
-            raise AssertionError(f"unknown archive type: {value}")
+            raise AssertionError(f"unknown archive type: {value!r}")
 
         return obj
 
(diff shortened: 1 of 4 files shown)


sdk/python/lib/pulumi/runtime/rpc.py
```diff
@@ -125,7 +127,7 @@ async def serialize_property(value: 'Input[Any]',
             remote_asset = cast('RemoteAsset', value)
             obj["uri"] = await serialize_property(remote_asset.uri, deps, input_transformer)
         else:
-            raise AssertionError(f"unknown asset type: {value}")
+            raise AssertionError(f"unknown asset type: {value!r}")
```
justinvp, line 130: Otherwise mypy errors:

```
error: On Python 3 '{}'.format(b'abc') produces "b'abc'"; use !r if this is a desired behavior
```

due to the `not isinstance(value, bytes)` above.

sdk/python/lib/pulumi/runtime/rpc.py
```diff
@@ -194,7 +196,7 @@ async def serialize_property(value: 'Input[Any]',
         value = _types.input_type_to_dict(value)
         transform_keys = False
 
-    if isinstance(value, Mapping):  # pylint: disable=bad-option-value,isinstance-second-argument-not-valid-type
+    if isinstance(value, abc.Mapping):
```
justinvp, line 199: While the previous code worked, we should be checking against the abstract base class from `collections.abc` rather than the type from `typing`.

sdk/python/lib/test/test_next_serialize.py
```diff
@@ -530,7 +550,7 @@ async def test_apply_does_not_propagate_secret_on_unknown_unknown_output_during_
         settings.SETTINGS.dry_run = True
 
         out = self.create_output(0, is_known=False)
-        r = out.apply(lambda v: self.create_output("inner", is_known=false, is_secret=True))
+        r = out.apply(lambda v: self.create_output("inner", is_known=False, is_secret=True))
```
justinvp, line 553: Noticed this when I had the file open in VS Code with Pylance installed.

sdk/python/lib/test/test_next_serialize.py
```diff
@@ -951,7 +988,18 @@ class InputTypeSerializationTests(unittest.TestCase):
     async def test_simple_input_type(self):
         it = FooArgs(first_arg="hello", second_arg=42)
         prop = await rpc.serialize_property(it, [])
-        self.assertDictEqual(prop, {"firstArg": "hello", "secondArg": 42})
+        self.assertEqual({"firstArg": "hello", "secondArg": 42}, prop)
```
justinvp, line 991: `assertEqual` will use `assertDictEqual` internally per https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertDictEqual

> This method will be used by default to compare dictionaries in calls to assertEqual().

justinvp:
This is a non-breaking change. Python doesn’t have types at runtime, so nothing would be hard broken by the change.

It is possible someone using `mypy` (or some other type checker) will see new warnings, depending on what they’re doing. In particular, if someone happened to be mutating a resulting `list`, they'll likely see new warnings from their type checker because `Sequence` is a readonly interface whereas `List` is mutable (this is a non-issue at runtime as the instance will be a mutable `list`).

We already did this for maps, though, and I haven't heard any user feedback from users about it. Previously maps were annotated as `dict` (mutable), and the new type annotations recently released use `Mapping[str, T]` (readonly). So this change would just be doing the same for lists.

I do wonder if we should consider annotating outputs (return types) as `MutableSequence[T]` or `List[T]`, and doing the same for maps, annotating outputs as `MutableMapping[str, T]` or `Dict[str, T]`. However, I'm inclined to keep these annotated as the readonly `Sequence[T]` and `Mapping[str, T]` types for both inputs and outputs for now. This aligns with what we've done in other languages (e.g. .NET uses `ImmutableArray<T>` and `ImmutableDictionary<string, TValue>`). If we hear feedback otherwise from users, we can always consider changing the annotations for results to be the mutable variants.

komalali (approved):

sdk/python/lib/test/test_next_serialize.py
justinvp: The lambda passed to `apply` is not called because `settings.SETTINGS.dry_run = True`.  If `settings.SETTINGS.dry_run` was set to `False`, then it fails as you'd expect with:

```
NameError: name 'false' is not defined
```

Issue #5281, opened 2020-09-03
[Go] Automatic plugin acquisition not working with vendoring


elsesiy:
The go language runtime is supposed to automatically acquire plugins needed during execution. The list of plugins is generated based off the `go.mod` file using `go list -m -json all`. There's a case where this fails when users decided to use vendoring via `go mod vendor` since in this case `go list` needs to bypass the vendor dir:
```
go list -m: can't compute 'all' using the vendor directory
        (Use -mod=mod or -mod=readonly to bypass.)
```

As this prevents the language runtime from determining the necessary plugin versions, users need to manually acquire plugins via `pulumi plugin install resource <plugin-name>`.

## Expected Behavior
Plugin acquisition succeeds regardless of vendoring.

## Current Behavior
Any pulumi command such as `preview` or `up` fails since the needed resource plugin is not available:

```
Diagnostics:
  pulumi:providers:azure (default):
    error: no resource plugin 'azure' found in the workspace or on your $PATH
```

## Steps to Reproduce
1. Create a new go project e.g. `pulumi new azure-go`
2. `pulumi preview` succeeds
3. `go mod vendor`
4. Remove existing plugin via `pulumi plugin rm resource azure`
4. `pulumi preview` fails

## Context (Environment)
This is a pretty nasty bug that plagued me for a while. The current implementation makes the usability of go really tedious as users have to manually keep track of plugins and update them as `go.mod` is being updated. Only after a long conversation with @EvanBoyle and @clstokes we were able to figure it out.
I'd also like to propose to not indefinitely cache acquired plugins since this leads to a lot of outdated versions taking up space on disk.

PR #5286, opened 2020-09-03
Update Go plugin acquisition to work for users who vendor dependencies

EvanBoyle:
Fixes https://github.com/pulumi/pulumi/issues/5281

Tested plugin acquisition manually on vendored and non-vendored projects and it works as expected on both.


elsesiy (approved):
Lgtm, thanks for the quick fix!

lblackstone (approved):

komalali (approved):

Issue #5212, opened 2020-08-24
[codegen/python] Error when specifying nested Provider inputs using classes


justinvp:
Provider inputs are handled a bit differently than standard resource inputs in that they are first serialized to JSON. We're not correctly handling this for the new input classes.

Attempting to do something like:

```python
provider = aws.Provider(
    'privileged',
    assume_role=aws.ProviderAssumeRoleArgs(
        role_arn=role_to_assume_arn,
        session_name='PulumiSession',
        external_id='PulumiApplication',
    ),
    region=aws_config.require('region')
)
```

Will result in the following `TypeError: Object of type ProviderAssumeRoleArgs is not JSON serializable`

<details>
  <summary>Full stack trace</summary>
  
```
    error: Program failed with an unhandled exception:
    error: Traceback (most recent call last):
      File "/opt/pulumi/bin/pulumi-language-python-exec", line 85, in <module>
        loop.run_until_complete(coro)
      File "/usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8/lib/python3.8/asyncio/base_events.py", line 616, in run_until_complete
        return future.result()
      File "/Users/justin/go/src/github.com/pulumi/examples/aws-py-assume-role/assume-role/venv/lib/python3.8/site-packages/pulumi/runtime/stack.py", line 83, in run_in_stack
        await run_pulumi_func(lambda: Stack(func))
      File "/Users/justin/go/src/github.com/pulumi/examples/aws-py-assume-role/assume-role/venv/lib/python3.8/site-packages/pulumi/runtime/stack.py", line 51, in run_pulumi_func
        await RPC_MANAGER.rpcs.pop()
      File "/Users/justin/go/src/github.com/pulumi/examples/aws-py-assume-role/assume-role/venv/lib/python3.8/site-packages/pulumi/runtime/rpc_manager.py", line 67, in rpc_wrapper
        result = await rpc
      File "/Users/justin/go/src/github.com/pulumi/examples/aws-py-assume-role/assume-role/venv/lib/python3.8/site-packages/pulumi/runtime/resource.py", line 435, in do_register_resource_outputs
        serialized_props = await rpc.serialize_properties(outputs, {})
      File "/Users/justin/go/src/github.com/pulumi/examples/aws-py-assume-role/assume-role/venv/lib/python3.8/site-packages/pulumi/runtime/rpc.py", line 71, in serialize_properties
[...]

PR #5307, opened 2020-09-09
[sdk/python] Add pulumi.runtime.to_json utility

justinvp:
Non-string provider inputs must be projected as JSON formatted strings. The current codegen simply calls `json.dumps` for such properties, but this does not work for the new input types, which aren't JSON serializable.

This commit adds a new `pulumi.runtime.to_json` utility function to the core SDK, which is capable of serializing both raw dicts and input types as JSON. The codegen will be updated to make use of this new function rather than `json.dumps`.

It's exposed from `pulumi.runtime` rather than `pulumi` because it isn't really meant to be used directly by users, but rather from our generated code. Being under `pulumi.runtime` seems sufficiently "hidden" from common use. I decided to add this utility to the SDK, as I didn't want to expose `is_input_type` and `input_type_to_dict` just yet (though, it may be useful to expose them eventually).

Codegen PR: https://github.com/pulumi/pulumi/pull/5308

Part of https://github.com/pulumi/pulumi/issues/5212

### DIFF
--- a/sdk/python/lib/pulumi/runtime/_json.py
+++ b/sdk/python/lib/pulumi/runtime/_json.py
@@ -1,0 +1,45 @@
+# Copyright 2016-2020, Pulumi Corporation.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from collections import abc
+import json
+from typing import Any
+
+from .. import _types
+
+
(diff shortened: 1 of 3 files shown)


sdk/python/lib/pulumi/runtime/_json.py
```diff
@@ -0,0 +1,45 @@
...
+# limitations under the License.
+
+from collections import abc
+import json
+from typing import Any
+
+from .. import _types
+
+
+def _to_json_serializable_obj(obj: Any) -> Any:
```
komalali, line 22: I'm curious if we considered using pickle here, I believe it's capable of serializing class instances. https://docs.python.org/3/library/pickle.html#pickling-class-instances

komalali (approved):
This looks fine, but I'm curious whether we could use `pickle.dumps()` instead of `json.dumps()` in the codegen and avoid having to add this utility function.

EDIT: Pickling does not work because it serializes to byte streams, not JSON. IGNORE ME!

justinvp:
Even if we could make use of something built-in to serialize classes to JSON, we'd still need to expose something in the SDK to convert the class instance's variable names from snake_case to Pulumi camelCase names (which `_types.input_type_to_dict` does inside `_to_json_serializable_obj`).

PR #5308, opened 2020-09-09
[codegen/python] Use pulumi.runtime.to_json

justinvp:
https://github.com/pulumi/pulumi/pull/5307 needs to be released before any providers adopting this change are released.

Non-string provider inputs must be projected as JSON formatted strings. The current codegen simply calls `json.dumps` for such properties, but this does not work for the new input types, which aren't JSON serializable.

To address this, make use of the new `pulumi.runtime.to_json` utility function, which is capable of serializing raw dicts and input types as JSON.

Part of https://github.com/pulumi/pulumi/issues/5212


justinvp:
The downstream test failures are due to https://github.com/pulumi/pulumi-terraform-bridge/issues/253

pgavlin (approved):

komalali (approved):

Issue #4514, opened 2020-04-28
Python - numbers get typed as floats, but should be integers


clstokes:
In Python, some resource arguments that are integers (ports, counts, etc) are typed as float when they should be integers. 

Examples:
- `port` at https://www.pulumi.com/docs/reference/pkg/aws/elasticache/replicationgroup/#inputs
- `max_size`, `min_size `, `default_cooldown `, `desired_capacity `, etc. at https://www.pulumi.com/docs/reference/pkg/aws/autoscaling/group/#properties
- `priority` at https://www.pulumi.com/docs/reference/pkg/gcp/compute/firewall/
- `partitions`, `replication_factor` at https://www.pulumi.com/docs/reference/pkg/kafka/topic/#inputs

justinvp:
I'm actually not sure why we haven't typed these as `int`.
https://github.com/pulumi/pulumi/blob/2585b86aa4252b1ea430dbfd2324c47ea57763b1/pkg/codegen/python/gen.go#L1546-L1547

.NET codegen types them as `int`:

https://github.com/pulumi/pulumi/blob/2585b86aa4252b1ea430dbfd2324c47ea57763b1/pkg/codegen/dotnet/gen.go#L287-L290

I don't see any reason why we couldn't do the same for Python. Would need to confirm it doesn't cause any problems roundtripping through serialization/deserialization.

justinvp:
It's worth considering this now as we're refining the type annotations.

PR #5309, opened 2020-09-09
[codegen/python] Use `int` as the type annotation for integers

justinvp:
Fixes https://github.com/pulumi/pulumi/issues/4514


justinvp:
The downstream test failures are due to https://github.com/pulumi/pulumi-terraform-bridge/issues/253

komalali (approved):

pgavlin (approved):

PR #5311, opened 2020-09-09
Simplify type assertion in unwrap.spec.ts

justinvp:
Reverts change from https://github.com/pulumi/pulumi/pull/5280

### DIFF
--- a/sdk/nodejs/tests/unwrap.spec.ts
+++ b/sdk/nodejs/tests/unwrap.spec.ts
@@ -15,7 +15,7 @@
 // tslint:disable
 
 import * as assert from "assert";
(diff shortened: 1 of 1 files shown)


EvanBoyle:
Just out of curiosity, what does the double type assertion here mean? Is `<Resource><any>` equivalent to `<any>`?

justinvp:
> Is `<Resource><any>` equivalent to `<any>`?

Yes, we should be able to simplify to just `<any>`. I'll update this PR to do that after https://github.com/pulumi/pulumi/pull/5276 is merged.

jaxxstorm (approved):

pgavlin (approved):

Issue #4003, opened 2020-02-27
Add enum support to the schema and code generators


praneetloke:
This issue tracks enum support in the schema, codegen, sdk and docs.

* [x] Support enums in the pulumi schema.
* [x] [codegen/nodejs] Add typescript support for enums. #5433
* [x] [codegen/nodejs] Translate enum output properties. #5598 
* [x] [sdk/nodejs] Add support for enums.  #5600
* [x] [codegen/dotnet] Add C# support for enums. #5434
* [x] [sdk/dotnet] Add support for enums. #5601
* [x] [codegen/python] Add python support for enums. #5435
* [x] [sdk/python] Add support for enums. #5602
* [x] [codegen/go] Add go support for enums. #5436
* [x] [sdk/go] Add support for enums. #5603
* [x] Add enums to the docs. #5437
* [x] Schematize aws enum overlays. https://github.com/pulumi/pulumi-aws/issues/1130

praneetloke:
I removed the size label and reassigned it to @tasiah. We will have a size estimate before this work begins.

praneetloke:
@pgavlin I have split out the work to add support for typed-JSON objects in the schema. I am assigning this specific work item related to adding support for enums, to you.

mikhailshilkov:
Azure API definitions have two types of enums:

- A "strong" enum with a fixed set of possible values, e.g. `LogLevel` with values of `Off, Verbose, Information, Warning, Error`
- A more relaxed enum that still has a list of possible values but instructs the clients to model this value as a `string`. This relaxes some restrictions about what constitutes a breaking change and adds flexibility. In practice, the actual values that end users would pass would still be from the enumerated set in most cases

We may want to have this distinction of closed/open enums in the schema and codegen too.

komalali:
@mikhailshilkov can you explain a bit more about what you mean by "relaxed enum"? Or maybe an example from the Azure API?

Specifically, I'm trying to understand this:
> This relaxes some restrictions about what constitutes a breaking change and adds flexibility.

mikhailshilkov:
@komalali See Azure docs here: http://azure.github.io/autorest/extensions/#x-ms-enum

They have a `modelAsString` property which instructs the client SDKs to give the type `string` to these fields instead of a proper enum. If the service defines possible values `A` and `B` for such a field, the SDK should still accept `C` as a value and send it to the service without a validation failure.

[...]

PR #5313, opened 2020-09-09
[codegen/schema] - Add enums to the schema

komalali:
Adds enum types to the schema.

Related conversation: #4003

### DIFF
--- a/pkg/codegen/schema/schema.go
+++ b/pkg/codegen/schema/schema.go
@@ -128,6 +128,18 @@ func (t *ArrayType) String() string {
 
 func (*ArrayType) isType() {}
 
+// EnumType represents an enum.
+type EnumType struct {
+	// ElementType is the element type of the enum.
+	ElementType Type
+}
+
+func (t *EnumType) String() string {
+	return fmt.Sprintf("Enum<%v>", t.ElementType)
+}
+
+func (*EnumType) isType() {}
+
 // UnionType represents values that may be any one of a specified set of types.
 type UnionType struct {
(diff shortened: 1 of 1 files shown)


pkg/codegen/schema/schema.go
```diff
@@ -128,6 +128,34 @@ func (t *ArrayType) String() string {
...
+	// Value is the value of the enum.
+	Value interface{}
+	// Description for the enum value.
+	Description string
+	// Name for the enum.
+	Name string
+}
+
+func (t *EnumType) String() string {
+	return fmt.Sprintf("Enum<%v>", t.ElementType)
```
komalali, line 154: Should this just be the type name?

pkg/codegen/schema/schema.go
```diff
@@ -128,6 +128,34 @@ func (t *ArrayType) String() string {
...
 func (*ArrayType) isType() {}
 
+// EnumType represents an enum.
+type EnumType struct {
+	// ElementType is the element type of the enum.
+	ElementType Type
+	// Elements are the predefined enum values.
+	Elements []*Enum
+	// ModelAsString indicates whether the enum should be strictly enforced.
+	ModelAsString bool
```
komalali, line 138: `ModelAsString` locks us in with string enums and perhaps isn't totally clear. Maybe something like `Enforced` or `Strict` or `Validated` makes more sense here?

komalali:
Closing this to address feedback from the design doc. Will re-open when ready.

PR #5314, opened 2020-09-09
Add IsSelectStack404Error and IsCreateStack409Error

### LINKED ISSUE #5255: Add IsSelectStack404Error and IsCreateStack409Error
@metral had great feedback that it would be useful to identify some of the failure cases with creating/selecting stacks to understand how to respond rather than inferring:

https://github.com/pulumi/pulumi-kubernetes-operator/pull/86/files#diff-e31cd52d2ed37a9f588a46e7ca7e52b6R551-R556

We should add support to `autoError` to be able to detect this from automation API. The implementation will look very similar to that of update conflict error: https://github.com/pulumi/pulumi/blob/b0d91478bfc4af68f93d75d74abb1ae8a2b89443/sdk/go/x/auto/errors.go#L50

A sample of error output when selecting a stack that doesn't exist:

```
error: no stack named 'aasdwfgsadga' found
```

A sample of error output when trying to create a stack that already exists:

```
error: stack 'EvanBoyle/aws-go-fargate/ddd' already exists
```

metral:
Closes #5255

### DIFF
--- a/sdk/go/x/auto/errors_test.go
+++ b/sdk/go/x/auto/errors_test.go
@@ -165,6 +165,59 @@ func TestCompilationErrorGo(t *testing.T) {
 	}
 }
 
+func TestSelectStack404Error(t *testing.T) {
+	ctx := context.Background()
+	sName := fmt.Sprintf("int_test%d", rangeIn(10000000, 99999999))
+	fqsn := FullyQualifiedStackName(pulumiOrg, "testproj", sName)
+
+	// initialize
+	pDir := filepath.Join(".", "test", "testproj")
+	opts := []LocalWorkspaceOption{WorkDir(pDir)}
+	w, err := NewLocalWorkspace(ctx, opts...)
+	if err != nil {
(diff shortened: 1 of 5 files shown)


sdk/go/x/auto/errors_test.go
```diff
@@ -165,6 +165,53 @@ func TestCompilationErrorGo(t *testing.T) {
...
+	fqsn := FullyQualifiedStackName(pulumiOrg, "testproj", sName)
+
+	// initialize first stack
+	pDir := filepath.Join(".", "test", "testproj")
+	_, err := NewStackLocalSource(ctx, fqsn, pDir)
+	if err != nil {
+		t.Errorf("failed to initialize stack, err: %v", err)
+		t.FailNow()
+	}
+
```
EvanBoyle, line 200: Would be good to add a defer here that removes the stack for cleanup.

EvanBoyle (approved):
🚀  Looks great!

Issue #5297, opened 2020-09-08
[Automation API] create stack with non-default secrets provider


EvanBoyle:
The APIs don't currently support this. We need to think about how to do it cleanly, and what scope this should be (stack level, workspace level, etc). 

Related https://github.com/pulumi/pulumi/issues/5254

EvanBoyle:
CC @metral who noted that this functionality was missing and needed for the kubernetes provider. Perhaps he can provide some input @stack72

PR #5320, opened 2020-09-10
Allow passing a non-default secrets provider to AutomationAPI

stack72:
Fixes: #5297

### DIFF
--- a/sdk/go/x/auto/local_workspace.go
+++ b/sdk/go/x/auto/local_workspace.go
@@ -39,10 +39,11 @@ import (
 // alter the Workspace Pulumi.yaml file, and setting config on a Stack will modify the Pulumi.<stack>.yaml file.
 // This is identical to the behavior of Pulumi CLI driven workspaces.
 type LocalWorkspace struct {
-	workDir    string
-	pulumiHome string
-	program    pulumi.RunFunc
-	envvars    map[string]string
+	workDir         string
+	pulumiHome      string
+	program         pulumi.RunFunc
+	envvars         map[string]string
+	secretsProvider string
 }
 
 var settingsExtensions = []string{".yaml", ".yml", ".json"}
@@ -330,7 +331,11 @@ func (l *LocalWorkspace) CreateStack(ctx context.Context, fqsn string) error {
 		return errors.Wrap(err, "failed to create stack")
 	}
 
-	stdout, stderr, errCode, err := l.runPulumiCmdSync(ctx, "stack", "init", fqsn)
+	args := []string{"stack", "init", fqsn}
+	if l.secretsProvider != "" {
+		args = append(args, fmt.Sprintf("--secrets-provider=%s", l.secretsProvider))
+	}
+	stdout, stderr, errCode, err := l.runPulumiCmdSync(ctx, args...)
 	if err != nil {
 		return newAutoError(errors.Wrap(err, "failed to create stack"), stdout, stderr, errCode)
 	}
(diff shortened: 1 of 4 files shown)


sdk/go/x/auto/local_workspace.go
```diff
@@ -136,7 +137,7 @@ func (l *LocalWorkspace) GetConfig(ctx context.Context, fqsn string, key string)
 	if err != nil {
 		return val, errors.Wrapf(err, "could not get config, unable to select stack %s", fqsn)
 	}
-	stdout, stderr, errCode, err := l.runPulumiCmdSync(ctx, "config", "get", key, "--show-secrets", "--json")
+	stdout, stderr, errCode, err := l.runPulumiCmdSync(ctx, "config", "get", key, "--json")
```
EvanBoyle, line 140: `--show-secrets` was an intentional decision here (and below). The return type of `GetConfig` is a `ConfigValue` which has:

- `Value`: the plaintext context of the config
- `Secret`: a boolean flag indicating secretness

This allows users (who are presumably privileged) to handle secrets with the maximum flexibility possible without overcomplicating the API surface.

EvanBoyle (commented):
This is looking great. Other than the comment, it would be nice to add some godoc examples:

Something similar to this, but that demonstrates creating a workspace and a stack with a secrets provider: https://github.com/pulumi/pulumi/blob/46da25492e38607967a83c0fa4bfa3f9772cb409/sdk/go/x/auto/example_test.go#L188

I believe that you can get an example entry if you do something like `ExampleLocalWorkspace_secretsProvider`:

https://godoc.org/github.com/fluhus/godoc-tricks#Examples

A test would be nice as well if possible.

sdk/go/x/auto/local_workspace.go
stack72: There is no such flag on config get - it *always* returns the unencrypted value by default

EvanBoyle: LOL good catch :)

sdk/go/x/auto/local_workspace_test.go
```diff
@@ -32,6 +33,71 @@ import (
...
 
+func TestWorkspaceSecretsProvider(t *testing.T) {
+	ctx := context.Background()
+	pName := "go_secrets_provider"
+	sName := fmt.Sprintf("int_test%d", rangeIn(10000000, 99999999))
+	fqsn := FullyQualifiedStackName(pulumiOrg, pName, sName)
+
+	// We can't use Workspace EnvVars as the Workspace uses the secrets provider to
+	// create the Stack
+	os.Setenv("PULUMI_CONFIG_PASSPHRASE", "password")
```
EvanBoyle, line 44: Note that you can do something along the lines of: 

```go
w, err := NewLocalWorkspace( Program(...), SecretsProvider(...), Project(...))
w.SetEnv()
NewStack(ctx, fqsn, ...)
```

This is fine for now. CC @metral what do you think about adding environment variables as workspace options to make this a little more ergonomic? That way both SecretsProvider and Envs could be specified at workspace creation time and enable using the `NewStackInlineSource` helper?

EvanBoyle (approved):
Looks good. Just a heads up you can use LocalWorkspace directly to avoid having to set OS envs, but this is fine for now as this is just for testing.

stack72:
> Looks good. Just a heads up you can use LocalWorkspace directly to avoid having to set OS envs, but this is fine for now as this is just for testing.

yeah my testing used exactly that - i was using this for an automated test only

sdk/go/x/auto/local_workspace_test.go
metral: > what do you think about adding environment variables as workspace options to make this a little more ergonomic? That way both SecretsProvider and Envs could be specified at workspace creation time

SGTM. Feel free to open an issue and assign it to me

PR #5321, opened 2020-09-10
[codegen/dotnet] Don't append Result to mixed output type names

mikhailshilkov:
Problem: every output class in AzureRM .NET SDK gets "Result" appended to its name. That's because the same types are used for both resources and corresponding invokes.

This PR stops appending "Result" for types that service both resources and functions. The input type change is for symmetry and shouldn't affect any existing provider.

AFAIK, this change does not affect TF-based providers (tried with Azure locally).

### DIFF
--- a/pkg/codegen/dotnet/gen.go
+++ b/pkg/codegen/dotnet/gen.go
@@ -1473,14 +1473,20 @@ func generateModuleContextMap(tool string, pkg *schema.Package, info CSharpPacka
 		mod.functions = append(mod.functions, f)
 		if f.Inputs != nil {
 			visitObjectTypes(f.Inputs, func(t *schema.ObjectType) {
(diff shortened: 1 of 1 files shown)


justinvp (approved):
LGTM

pgavlin:
> That's because the same types are used for both resources and corresponding invokes.

Can you elaborate on this or give a specific example? Invokes shouldn't accept input-typed arguments.

mikhailshilkov:
> Can you elaborate on this or give a specific example?

The same GET in ARM API is used for both a resource and a get-invoke. The corresponding Pulumi schema types are defined once and are referred to from both the resource and the invoke. The generated classes are the same too (they don't use `Output<T>` because they are inside an `Output<T>` already). This PR removes `Result` from the class name in this case.

Issue #5305, opened 2020-09-09
Support passing passphrase via a file


clstokes:
Pulumi should allow `PULUMI_CONFIG_PASSPHRASE_FILE` to be set in a mutually exclusive manner with `PULUMI_CONFIG_PASSPHRASE`, and if it’s set, read the passphrase from the path contained in the variable. There are bunch of CI systems which expose secrets in the filesystem but don’t mask secret values in logs, so it would be nice to avoid a stray `set -xtrace` logging the passphrase when the variable is set.

stack72:
@clstokes do you think the file needs to have the plain text secret in it only or needs to have a different structure?

clstokes:
> @clstokes do you think the file needs to have the plain text secret in it only or needs to have a different structure?

plaintext is sufficient IMO.

stack72:
ok so we can do this that it will just read the contents of the file

PR #5327, opened 2020-09-11
Add the ability to set the passphrase secrets provider to read from a passphrase file

stack72:
Fixes: #5305

https://asciinema.org/a/INd7fs6nALeuKgxusIxr73VHk

### DIFF
--- a/pkg/cmd/pulumi/crypto_local.go
+++ b/pkg/cmd/pulumi/crypto_local.go
@@ -17,12 +17,13 @@ package main
 import (
 	cryptorand "crypto/rand"
 	"encoding/base64"
-	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 
+	"github.com/pkg/errors"
 	"github.com/pulumi/pulumi/pkg/v2/secrets"
 	"github.com/pulumi/pulumi/pkg/v2/secrets/passphrase"
 	"github.com/pulumi/pulumi/sdk/v2/go/common/diag"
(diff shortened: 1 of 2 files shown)


pkg/cmd/pulumi/crypto_local.go
```diff
@@ -35,8 +37,20 @@ func readPassphrase(prompt string) (string, error) {
 	if phrase, ok := os.LookupEnv("PULUMI_CONFIG_PASSPHRASE"); ok {
 		return phrase, nil
 	}
+	if phraseFile, ok := os.LookupEnv("PULUMI_CONFIG_PASSPHRASE_FILE"); ok {
+		phraseFilePath, err := filepath.Abs(phraseFile)
+		if err != nil {
+			return "", errors.New("unable to construct a path the PULUMI_CONFIG_PASSPHRASE_FILE")
+		}
+		phraseDetails, err := ioutil.ReadFile(phraseFilePath)
```
justinvp, line 45: What if the file has trailing newlines? I assume we do not want to treat that as part of the passphrase.

Maybe do it like how we read from the console elsewhere (which can be from a file via redirected std input)?

https://github.com/pulumi/pulumi/blob/848a98d5112ec93239501fd66b51c9f1f30353ac/sdk/go/common/util/cmdutil/console.go#L72-L78

justinvp (approved):
LGTM otherwise.

Also, mind opening a PR in the docs repo to add the new environment variable to the docs? https://github.com/pulumi/docs/blob/master/content/docs/reference/cli/environment-variables.md

pkg/cmd/pulumi/crypto_local.go
```diff
@@ -35,8 +37,27 @@ func readPassphrase(prompt string) (string, error) {
...
+			return "", errors.New("unable to open PULUMI_CONFIG_PASSPHRASE_FILE")
+		}
+
+		reader := bufio.NewReader(f)
+		passphraseDetails, err := reader.ReadString('\n')
+		if err != nil {
+			return "", errors.New("unable to read PULUMI_CONFIG_PASSPHRASE_FILE")
+		}
+
+		return cmdutil.RemoveTrailingNewline(passphraseDetails), nil
```
justinvp, line 56: Sorry, looking this over now, prob easiest to read it as before, but just trim the string before returning it as Lee suggested.

pkg/cmd/pulumi/crypto_local.go
```diff
@@ -35,8 +37,27 @@ func readPassphrase(prompt string) (string, error) {
 	if phrase, ok := os.LookupEnv("PULUMI_CONFIG_PASSPHRASE"); ok {
 		return phrase, nil
 	}
+	if phraseFile, ok := os.LookupEnv("PULUMI_CONFIG_PASSPHRASE_FILE"); ok {
+		phraseFilePath, err := filepath.Abs(phraseFile)
+		if err != nil {
+			return "", errors.New("unable to construct a path the PULUMI_CONFIG_PASSPHRASE_FILE")
```
justinvp, line 43:
```suggestion
			return "", errors.Wrap(err, "unable to construct a path the PULUMI_CONFIG_PASSPHRASE_FILE")
```

(Suggestion applies to all errors)

justinvp: This requires changing the import to `"github.com/pkg/errors"`

pkg/cmd/pulumi/crypto_local.go
stack72: fixed!

Issue #5224, opened 2020-08-25
Add streaming Stack lifecycle APIs


EvanBoyle:
Automation API currently only has synchronous APIs for up/preview/refresh/destroy. This isn't ideal as larger deployments could take many minutes, and we'd like to enable users to build applications that can provide incremental updates to their users. 

This work depends on defining structured output for all lifecycle methods https://github.com/pulumi/pulumi/issues/5218 https://github.com/pulumi/pulumi/issues/5219 https://github.com/pulumi/pulumi/issues/5220

EvanBoyle:
Initially, we're just adding streaming unstructured output for Up, Refresh, and Destroy. The intent here (and the ask from users) is to be able to redirect that output during development. For example, redirect command output into a file, or to os.stdout and be able to tail that output as the deployment is running (think long-running eks/gke deployments and the like).

PR #5329, opened 2020-09-11
Automation API - add the ability to stream progress updates during up/refresh/destroy

EvanBoyle:
Fixes https://github.com/pulumi/pulumi/issues/5224 for now until we address structured output. 

After talking with some users in slack, it seems that the real desire (structured output aside) is to be able to tail a file or redirect output to stdout to have some notion of progress for long-running deployments (EKS/GKE). This solution was pretty simple both from a consumption and implementation perspective. We give the user the ability to specify a list of `io.Writer`, and we append command output to that writer incrementally in addition to the complete output that we return when the command completes. 

I left this out for preview for the time being, as this API already has structured output. We may want to revisit that, and make preview unstructured by default, with a functional option to force structured output but I'd rather design that separately. 

I'll rebase on main and add a changelog entry after @pgavlin's PR lands https://github.com/pulumi/pulumi/pull/5317


metral (approved):
LGTM 🚀 🎉

EvanBoyle:
Will link a new rebased PR and link back to here.

Issue #5319, opened 2020-09-10
Automation API - add support for private git repositories (SSH key authentication)


omerlh:
I am trying to use the automation with a private git repo using SSH: 
```
repo := auto.GitRepo{
			URL:         "git@github.com:<>",
			ProjectPath: "pulumi-gcp-infra",
		}
```
But I am getting the following error:
```
Failed to create workspace from git repo: failed to create workspace, unable to enlist in git repo: unable to clone repo: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain
```

Please add support for cloning private repos!

justinvp:
Related:
- https://github.com/pulumi/pulumi/issues/4872
- https://github.com/pulumi/pulumi-kubernetes-operator/issues/84

PR #5333, opened 2020-09-11
Add support for cloning private repos as part of automation API

stack72:
Fixes: #5319

### DIFF
--- a/sdk/go/x/auto/example_test.go
+++ b/sdk/go/x/auto/example_test.go
@@ -214,30 +214,16 @@ func ExampleGitRepo_personalAccessToken() {
 	// Get the Sourcecode Repository PERSONAL_ACCESS_TOKEN
 	token, _ := os.LookupEnv("PERSONAL_ACCESS_TOKEN")
 
-	// we'll compile a the program into an executable with the name "examplesBinary"
-	binName := "examplesBinary"
 	repo := GitRepo{
 		URL:         "https://github.com/pulumi/test-repo.git",
 		ProjectPath: "goproj",
-		// This will use a personal access token for the chosen source code repository
-		PersonalAccessToken: token,
-		// this call back will get executed post-clone to allow for additional program setup
-		Setup: func(ctx context.Context, workspace Workspace) error {
-			cmd := exec.Command("go", "build", "-o", binName, "main.go")
-			cmd.Dir = workspace.WorkDir()
-			return cmd.Run()
+		Auth: &GitAuth{
+			PersonalAccessToken: token,
 		},
-	}
-	// an override to the project file in the git repo, specifying our pre-built executable
-	project := workspace.Project{
-		Name: tokens.PackageName(pName),
-		Runtime: workspace.NewProjectRuntimeInfo("go", map[string]interface{}{
-			"binary": binName,
-		}),
 	}
 
 	// initialize a stack from the git repo, specifying our project override
-	NewStackRemoteSource(ctx, fqsn, repo, Project(project))
+	NewStackRemoteSource(ctx, fqsn, repo)
 }
 
 func ExampleGitRepo_privateKeyPath() {
@@ -245,31 +231,17 @@ func ExampleGitRepo_privateKeyPath() {
 	pName := "go_remote_proj"
 	fqsn := FullyQualifiedStackName("myOrg", pName, "myStack")
 
-	// we'll compile a the program into an executable with the name "examplesBinary"
-	binName := "examplesBinary"
 	repo := GitRepo{
 		URL:         "https://github.com/pulumi/test-repo.git",
 		ProjectPath: "goproj",
-		// This will use a private SSH Key on a local filesystem.
-		SSHPrivateKeyPath: "/Users/myuser/.ssh/id_rsa",
-		Password:          "PrivateKeyPassword",
-		// this call back will get executed post-clone to allow for additional program setup
-		Setup: func(ctx context.Context, workspace Workspace) error {
-			cmd := exec.Command("go", "build", "-o", binName, "main.go")
-			cmd.Dir = workspace.WorkDir()
-			return cmd.Run()
+		Auth: &GitAuth{
+			SSHPrivateKeyPath: "/Users/myuser/.ssh/id_rsa",
+			Password:          "PrivateKeyPassword",
 		},
-	}
-	// an override to the project file in the git repo, specifying our pre-built executable
-	project := workspace.Project{
-		Name: tokens.PackageName(pName),
-		Runtime: workspace.NewProjectRuntimeInfo("go", map[string]interface{}{
-			"binary": binName,
-		}),
 	}
 
 	// initialize a stack from the git repo, specifying our project override
-	NewStackRemoteSource(ctx, fqsn, repo, Project(project))
+	NewStackRemoteSource(ctx, fqsn, repo)
 }
 
 func ExampleGitRepo_usernameAndPassword() {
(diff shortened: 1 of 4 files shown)


sdk/go/x/auto/local_workspace.go
```diff
@@ -584,6 +584,14 @@ type GitRepo struct {
 	CommitHash string
 	// Optional function to execute after enlisting in the specified repo.
 	Setup SetupFn
+	// The absolute path to a private key for access to the github repo
```
EvanBoyle, line 587: From a user-facing perspective, I could imagine it being confusing that there are four different top-level fields here that can be used in various combinations for auth. It's very clear from reading the examples in the godoc, but not necessarily from the IDE. I have two ideas on how we could make this more clear, and I'm curious to hear what you think:

1. We could enhance the comments on these fields to more clearly indicate the acceptable permutations (PAT, ssh + pass, user + pass).  It might also be helpful to include this in the top level comment for `GitRepo`
2. Another idea would be to hide this under an Auth field (in addition to adding some of the comments described in number 1):

```go
type GitRepo struct{
...
Auth GitAuth,
...
}

// GitAuth top-level comment indicating valid permutations and precedence.
type GitAuth struct {
        // The absolute path to a private key for access to the github repo
	SSHPrivateKeyPath string
	// The password that pairs with a username or as part of an SSH Private Key
	Password string
	// PersonalAccessToken is a GitHub personal access token is in replacment of your password
	PersonalAccessToken string
	// Username is the username to use when authenticating to a git repository
	Username string
}
```

sdk/go/x/auto/local_workspace.go
```diff
@@ -584,6 +584,14 @@ type GitRepo struct {
...
+	SSHPrivateKeyPath string
+	// The password that pairs with a username or as part of an SSH Private Key
+	Password string
+	// PersonalAccessToken is a GitHub personal access token is in replacment of your password
```
EvanBoyle, line 591: nit: is there any reason to use github over the more generic `git` (here and comments above)? I believe gitlab and others support similar auth mechanisms.

sdk/go/x/auto/example_test.go
```diff
@@ -205,6 +206,104 @@ func ExampleGitRepo() {
...
+			return cmd.Run()
+		},
+	}
+	// an override to the project file in the git repo, specifying our pre-built executable
+	project := workspace.Project{
+		Name: tokens.PackageName(pName),
+		Runtime: workspace.NewProjectRuntimeInfo("go", map[string]interface{}{
+			"binary": binName,
+		}),
+	}
```
EvanBoyle, line 237: To simplify this example and the other auth related examples, we have the option of removing the highlighted portion. Just a little less cognitive overhead for the user.

EvanBoyle (commented):
This looks great! Just a few ideas I'm curious to hear your thoughts on.

sdk/go/x/auto/git.go
```diff
@@ -21,11 +21,45 @@ import (
...
+
+	// Firstly we will try to check that an SSH Private Key Path has been specified
+	if repoArgs.SSHPrivateKeyPath != "" {
+		publicKeys, err := ssh.NewPublicKeysFromFile("git", repoArgs.SSHPrivateKeyPath, repoArgs.Password)
+		if err != nil {
+			return "", errors.Wrap(err, "unable to use SSH Private Key")
+		}
+
+		cloneOptions.Auth = publicKeys
+	}
```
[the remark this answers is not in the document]
metral: I'd add to this that we consider a switch instead to not fall through to each of the `if` statements and override `cloneOptions.Auth` if a user were to specify all auth options. Doing so would help the user have a clearer expectation of what the state of `cloneOptions.Auth` would be during use.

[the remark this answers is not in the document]
metral: Additionally, it may be worth adding checks at the top to enforce mutual exclusivity of a single auth option before configuring `cloneOptions.Auth`

sdk/go/x/auto/local_workspace.go
metral: Agreed to help clear this up in the code and comments. Here's an example of how we handle similar logic in p/eks

checks: 

https://github.com/pulumi/pulumi-eks/blob/7998754242c87d2217b8a087494c83c74f451662/nodejs/eks/nodegroup.ts#L338-L351

---

type comments:

https://github.com/pulumi/pulumi-eks/blob/7998754242c87d2217b8a087494c83c74f451662/nodejs/eks/nodegroup.ts#L695-L711

metral (changes requested):

sdk/go/x/auto/git.go
```diff
@@ -21,11 +21,57 @@ import (
...
+	cloneOptions := &git.CloneOptions{
+		URL: repoArgs.URL,
+	}
+
+	if repoArgs.Auth != nil {
+
+		authDetails := repoArgs.Auth
+		// Each of the authentication options are mutually exclusive so let's check that only 1 is specified
+		if (authDetails.SSHPrivateKeyPath != "" && authDetails.PersonalAccessToken != "") ||
+			(authDetails.SSHPrivateKeyPath != "" && authDetails.Password != "") ||
```
metral, line 38: This condition currently causes an err for a valid case if providing a passphrase for the SSH key when used below in L46.

I believe we'd like this line to error if the username is also set.

sdk/go/x/auto/git.go
```diff
@@ -21,11 +21,57 @@ import (
...
 )
 
 func setupGitRepo(ctx context.Context, workDir string, repoArgs *GitRepo) (string, error) {
+	cloneOptions := &git.CloneOptions{
+		URL: repoArgs.URL,
+	}
+
+	if repoArgs.Auth != nil {
+
+		authDetails := repoArgs.Auth
```
metral, line 35: nit: we set `authDetails` here but only use it in the mutual exclusive checks. we should make use of this var in the rest of the `if` statements e.g. L45, L47, L65

metral (changes requested):

sdk/go/x/auto/git.go
stack72: You are correct - it should be using username in this condition not the password

sdk/go/x/auto/git.go
stack72: Should be taken care of now!

sdk/go/x/auto/git.go
```diff
@@ -21,11 +21,57 @@ import (
...
+		// Each of the authentication options are mutually exclusive so let's check that only 1 is specified
+		if (authDetails.SSHPrivateKeyPath != "" && authDetails.PersonalAccessToken != "") ||
+			(authDetails.SSHPrivateKeyPath != "" && authDetails.Username != "") ||
+			(authDetails.PersonalAccessToken != "" && authDetails.Username != "") {
+			return "", errors.New("please specify one authentication option of `Personal Access Token`," +
+				"`Username\\Password` or `SSH Private Key Path`")
```
EvanBoyle, line 41: Looks like there's a space missing before "`Username..."

EvanBoyle (approved):
🚢

metral (approved):
:shipit: 🎉

sdk/go/x/auto/git.go
stack72: Fixed!

sdk/go/x/auto/example_test.go
```diff
@@ -205,6 +206,63 @@ func ExampleGitRepo() {
...
+	NewStackRemoteSource(ctx, fqsn, repo)
+}
+
+func ExampleGitRepo_privateKeyPath() {
+	ctx := context.Background()
+	pName := "go_remote_proj"
+	fqsn := FullyQualifiedStackName("myOrg", pName, "myStack")
+
+	repo := GitRepo{
+		URL:         "https://github.com/pulumi/test-repo.git",
```
justinvp, line 235: Interesting, I thought you'd have to specify the URL like `git@github.com:pulumi/test-repo.git` when using SSH for auth, but it works when doing HTTPS?. Do the `git@...` style URLs work as well?

stack72: they do AFAICT

metral: @stack72 Just ran into this for the operator, and the URL does in fact need to be `git@github.com...` as Justin points out.

Using the HTTPS url results in `unable to clone repo: invalid auth method`

metral: Opened https://github.com/pulumi/pulumi/issues/5382

Issue #5330, opened 2020-09-11
[Automation API] relax stack name requirements


EvanBoyle:
Since OSS stack identity will take a bit longer to flesh out, we're going to relax the constraints and remove any of the magic around "fully qualified stack names". This means removing the validation and making docs clear that fqsn is optional. There are also some methods like `pulumi stack` and `pulumi stack ls` that make an effort to massage names into the fully qualified form before returning to the user. We'll remove this in its entirety for the time being.

PR #5337, opened 2020-09-12
[Automation API] remove stack name validation and fqsn enforcement

EvanBoyle:
Fixes https://github.com/pulumi/pulumi/issues/5330

This change removes the requirement and validation around using fully qualified stack names (org/proj/name). It removes all of the places where we transform stack names into fully qualified stack names such as listing stacks or getting the current stack. This has one notable impact, which is requiring an additional parameter for `NewStackInlineSource` (project name) and related methods. 

I did keep the `FullyQualifiedStackName` method and it's usage in the examples as I believe this to be a best practice to get users to think about where they want a stack to be parented (in an org or under a personal user account). I'm open to feedback on this, but it seems like a net positive.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ CHANGELOG
 
 ## HEAD (Unreleased)
 
+- Relax stack name validations for Automation API [#5337](https://github.com/pulumi/pulumi/pull/5337)
+
 - Allow Pulumi to read a passphrase file, via `PULUMI_CONFIG_PASSPHRASE_FILE` to interact
   with the passphrase secrets provider. Pulumi will first try and use the `PULUMI_CONFIG_PASSPHRASE`
   to get the passphrase then will check `PULUMI_CONFIG_PASSPHRASE_FILE` and then all through to 
--- a/sdk/go/x/auto/local_workspace.go
+++ b/sdk/go/x/auto/local_workspace.go
@@ -620,7 +620,12 @@ func NewStackLocalSource(ctx context.Context, stackName, workDir string, opts ..
 // from the specified WorkDir. If the Stack already exists, it will not error
 // and proceed to selecting the Stack.This Workspace will pick up any available
 // Settings files (Pulumi.yaml, Pulumi.<stack>.yaml).
-func UpsertStackLocalSource(ctx context.Context, stackName, workDir string, opts ...LocalWorkspaceOption) (Stack, error) {
+func UpsertStackLocalSource(
+	ctx context.Context,
+	stackName,
+	workDir string,
+	opts ...LocalWorkspaceOption,
+) (Stack, error) {
 	opts = append(opts, WorkDir(workDir))
 	w, err := NewLocalWorkspace(ctx, opts...)
 	var stack Stack
@@ -634,7 +639,12 @@ func UpsertStackLocalSource(ctx context.Context, stackName, workDir string, opts
 // SelectStackLocalSource selects an existing Stack backed by a LocalWorkspace created on behalf of the user,
 // from the specified WorkDir. This Workspace will pick up
 // any available Settings files (Pulumi.yaml, Pulumi.<stack>.yaml).
-func SelectStackLocalSource(ctx context.Context, stackName, workDir string, opts ...LocalWorkspaceOption) (Stack, error) {
+func SelectStackLocalSource(
+	ctx context.Context,
+	stackName,
+	workDir string,
+	opts ...LocalWorkspaceOption,
+) (Stack, error) {
 	opts = append(opts, WorkDir(workDir))
 	w, err := NewLocalWorkspace(ctx, opts...)
 	var stack Stack
@@ -649,7 +659,12 @@ func SelectStackLocalSource(ctx context.Context, stackName, workDir string, opts
 // with source code cloned from the specified GitRepo. This Workspace will pick up
 // any available Settings files (Pulumi.yaml, Pulumi.<stack>.yaml) that are cloned into the Workspace.
 // Unless a WorkDir option is specified, the GitRepo will be clone into a new temporary directory provided by the OS.
-func NewStackRemoteSource(ctx context.Context, stackName string, repo GitRepo, opts ...LocalWorkspaceOption) (Stack, error) {
+func NewStackRemoteSource(
+	ctx context.Context,
+	stackName string,
+	repo GitRepo,
+	opts ...LocalWorkspaceOption,
+) (Stack, error) {
 	opts = append(opts, Repo(repo))
 	w, err := NewLocalWorkspace(ctx, opts...)
 	var stack Stack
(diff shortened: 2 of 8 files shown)


sdk/go/x/auto/local_workspace.go
```diff
@@ -375,16 +354,6 @@ func (l *LocalWorkspace) RemoveStack(ctx context.Context, fqsn string) error {
 // ListStacks returns all Stacks created under the current Project.
 // This queries underlying backend and may return stacks not present in the Workspace (as Pulumi.<stack>.yaml files).
 func (l *LocalWorkspace) ListStacks(ctx context.Context) ([]StackSummary, error) {
-	user, err := l.WhoAmI(ctx)
```
metral, line 378: What's the thinking behind removing the whoami and stack list checks?

sdk/go/x/auto/local_workspace.go
```diff
@@ -824,29 +791,25 @@ func SelectStackInlineSource(
...
-	if err := ValidateFullyQualifiedStackName(fqsn); err != nil {
-		return "", err
-	}
-	return strings.Split(fqsn, "/")[2], nil
+// stack names come in many forms:
+// s, o/p/s, u/p/s o/s
+// so just return the last chunk which is what will be used in pulumi.<stack>.yaml
+func getStackSettingsName(stackName string) string {
+	parts := strings.Split(stackName, "/")
+	return parts[len(parts)-1]
```
metral, line 812: We'll need to check for an empty slice here

sdk/go/x/auto/local_workspace_test.go
```diff
@@ -31,21 +31,21 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-const pulumiOrg = "moolumi"
+const pulumiOrg = "pulumi-test"
```
metral, line 34: we should use whoami or make this an envvar versus hard coding it

sdk/go/x/auto/stack.go
```diff
@@ -111,77 +111,80 @@ import (
...
 	workspace Workspace
-	fqsn      string
+	stackName string
 }
 
-// FullyQualifiedStackName returns an appropriately formatted name to be used for Stack creation/selection.
+// FullyQualifiedStackName returns a stack name formatted with the greatest possible specificity:
+// org/project/stack or user/project/stack
+// Using this format avoids ambiguity in stack identity guards creating or selecting the wrong stack.
+// Note that filestate backends (local file, S3, Azure Blob) do not support stack names in this format.
```
metral, line 124:
```suggestion
// Note that filestate backends (local file, S3, Azure Blob) do not support stack names in this format, and instead only use the stack name without an org/user or project.
```

metral (changes requested):
Overall LGTM 👍  🎉 

Just a couple of questions and suggestions.

nit: for future consideration, the rename of `fqsn` -> `stackName` would've been easier to review in its own commit from the changes that drop the name validation, but this PR was straight forward to follow 🙂

sdk/go/x/auto/local_workspace_test.go
EvanBoyle: this is the permanent engineering test account, it shouldn't change any more. Or are there other reasons for turning it into an env?

sdk/go/x/auto/local_workspace.go
EvanBoyle: Previously this API made an effort to normalize the returned stack name into either `user/proj/stack` or `org/proj/stack`. We've dropped that logic as a part of this change given that the local backend behavior is problematic and it's safer to just trust what the CLI gives us back.

As a result, the info from whomai/project are no longer needed here.

sdk/go/x/auto/local_workspace_test.go
metral: The primary reasons are to decouple the code from the CI env e.g. we've had projects start out in `moolumi` and move around since that became unmanageable for all repos.

metral: cc @stack72 for input ^

EvanBoyle: Ah, this is definitely a good idea for anyone who wants to build the project and run tests locally. I've gone ahead and updated to use `PULUMI_TEST_ORG` and default to "pulumi-test" if the env isn't set.

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ CHANGELOG
 
 ## HEAD (Unreleased)
 
+- Relax stack name validations for Automation API [#5337](https://github.com/pulumi/pulumi/pull/5337)
```
stack72, line 6: I don't think this is CHANGELOG worthy?

EvanBoyle: technically it's a bugfix for OSS backend users, but happy to remove if you think it shouldn't be here.

EvanBoyle:
@metral Apologies for that! I'll make sure to use separate commits for noisy refactors like that next time. I've made the suggested changes. PTAL.

CHANGELOG.md
stack72: nah you are right - it's a good bugfix to note - I wonder if it's better to phrase something like "support OSS backends for automation-api"? 

Trying to make changelogs useful... according to an issue that was opened before ;)

We can change it later though - we don't need to gate the PR

metral (approved):
🚀  LGTM

metral:
> I'll make sure to use separate commits for noisy refactors like that next time

All good! 😃

chrsmith:
Sorry to insert myself here. I just came across the issue around removing stack name validation and FQSN enforcement, and definitely want to take a close look here. Since there are a lot of bad things that can happen if we aren't very careful here.

For example, you should be blocked from trying to perform a stack update after changing the `project` field in `Pulumi.yaml` between stack updates. (Since otherwise you might be surprised to see every resource needing to be recreated.)

Anyways, if you are blocked needing to get this PR in, I can take a look after the fact. But if not, would you mind giving me a day to take a close look?

EvanBoyle:
@chrsmith Just to make sure we're both on the same page, this doesn't alter any behavior in the CLI just the layer of validation that the Automation API was adding ontop. Any stack names being used still have to go through the CLI validation layer when we shell out to `pulumi stack init` or `pulumi stack create`.

EvanBoyle:
> For example, you should be blocked from trying to perform a stack update after changing the project field in Pulumi.yaml between stack updates

For the SaaS backend, this would fail as there would be no stacks under the new project. (this would manifest as an error like: the update failing with no stack is currently selected in the workspace).

EvanBoyle:
@chrsmith I have other changes backed up behind this, so I'm going to go ahead and merge. Given that we're in alpha, this decision is easily reversible. Happy to open up a separate PR to address any concerns that you may have and update our strategy here accordingly.

PR #5342, opened 2020-09-12
Update CHANGELOG.md

cowwoc:
Fixed typo


cowwoc:
Sounds like your automated build is broken. I only fixed a typo in the readme. There is no way this could break the build.

stack72 (approved):

stack72:
hi @cowwoc 

The CI system currently has a bug where non-org members don't have a way of being able to use the secrets

We are just about to roll out a fix for this as we are finalizing a move to a new CI environment

Thanks so much for picking up the bug here

Paul

PR #5345, opened 2020-09-13
Do not generate unused input/output types in Go

mikhailshilkov:


lblackstone (approved):

EvanBoyle (approved):
Looks good. It might be good idea to build one or two of the resulting branches (aws and azure perhaps) just to verify that everything is OK. I'm not sure that the comparison job does this.

mikhailshilkov:
@EvanBoyle I built aws, azure, and k8s locally. I'll also work with @stack72 to make sure the upgrades are smooth and chagelogs are updated.

mikhailshilkov:
This change caused some trouble in `azure-nextgen` that I didn't catch while trying it out locally. I reverted the change in master and will give it another try later.

Issue #5233, opened 2020-08-26
Add `pulumi console` command


clstokes:
I frequently want to open the Pulumi Console when working with a Pulumi project or stack. It would be great to have a `pulumi console` command that could read my current stack details and open a browser to the stack's page - e.g. `https://app.pulumi.com/org/project/stack`. Bonus points (maybe?) if it had a flag to take the user to the latest stack update page - `https://app.pulumi.com/org/project/stack/updates/2`.

I currently use the shell alias below to provide this, but it would be great to have this built-in.

```
alias pulumi-console='pulumi stack | grep "More information at:" | cut -d" " -f4 | xargs open'
```

PR #5368, opened 2020-09-15
Add `pulumi console` command

sean1588:
Adding the `pulumi console` command to open console url for the current stack.

I am currently looking into the "nice to have" feature called out in the issue to open the most recent update, but for now here is the base functionality. If it is a lot more work, I'll break it off to a separate issue just to keep things moving.

Resolves: #5233

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,9 @@ CHANGELOG
 - Automation API - add recovery APIs (cancel/export/import)
   [#5369](https://github.com/pulumi/pulumi/pull/5369)
 
+- Add `pulumi console` command which  opens the currently selected stack in the Pulumi console.
+  [#5368](https://github.com/pulumi/pulumi/pull/5368)
+
 ## 2.10.0 (2020-09-10)
 
 - feat(autoapi): add Upsert methods for stacks
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ CHANGELOG
 - [sdk/go] Add missing Version field to invokeOptions
   [#5401](https://github.com/pulumi/pulumi/pull/5401)
 
+- Add `pulumi console` command which opens the currently selected stack in the Pulumi console.
+  [#5368](https://github.com/pulumi/pulumi/pull/5368)
+
 ## 2.10.1 (2020-09-16)
 
 - feat(autoapi): add GetPermalink for operation result
@@ -34,9 +37,6 @@ CHANGELOG
 
 - Automation API - add recovery APIs (cancel/export/import)
   [#5369](https://github.com/pulumi/pulumi/pull/5369)
-
-- Add `pulumi console` command which opens the currently selected stack in the Pulumi console.
-  [#5368](https://github.com/pulumi/pulumi/pull/5368)
 
 ## 2.10.0 (2020-09-10)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,7 @@ CHANGELOG
 - Automation API - add recovery APIs (cancel/export/import)
   [#5369](https://github.com/pulumi/pulumi/pull/5369)
 
-- Add `pulumi console` command which  opens the currently selected stack in the Pulumi console.
+- Add `pulumi console` command which opens the currently selected stack in the Pulumi console.
   [#5368](https://github.com/pulumi/pulumi/pull/5368)
 
 ## 2.10.0 (2020-09-10)
--- a/pkg/cmd/pulumi/console.go
+++ b/pkg/cmd/pulumi/console.go
@@ -1,0 +1,81 @@
+// Copyright 2016-2020, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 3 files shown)


pkg/cmd/pulumi/console.go
```diff
@@ -0,0 +1,81 @@
...
+						// console URL fails.
+						launchConsole(cloudBackend.URL())
+					}
+				} else {
+					launchConsole(cloudBackend.URL())
+				}
+				return nil
+			}
+			fmt.Println("This command is currently only supported when using a Pulumi managed backend" +
+				"For more information please visit, https://www.pulumi.com/pricing")
```
sean1588, line 68: Not sure if this is the right thing to put here, but I figure if they are not a saas user then might as well point them to a place where they can become one if they are interested.

praneetloke: I think this is a really good idea! As far as the messaging goes, what do you think of this?

> This command is not available for your backend. To migrate to the Pulumi Managed backend, please see https://www.pulumi.com/docs/intro/concepts/state/#adopting-the-pulumi-service-backend

CHANGELOG.md
```diff
@@ -30,6 +30,9 @@ CHANGELOG
 - Automation API - add recovery APIs (cancel/export/import)
   [#5369](https://github.com/pulumi/pulumi/pull/5369)
 
+- Add `pulumi console` command which  opens the currently selected stack in the Pulumi console.
```
praneetloke, line 33:
```suggestion
- Add `pulumi console` command which opens the currently selected stack in the Pulumi Console.
```

pkg/cmd/pulumi/console.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/pulumi/pulumi/pkg/v2/backend/display"
+	"github.com/pulumi/pulumi/pkg/v2/backend/httpstate"
+	"github.com/pulumi/pulumi/pkg/v2/backend/state"
+	"github.com/pulumi/pulumi/sdk/v2/go/common/util/cmdutil"
+)
+
+func newConsoleCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "console",
+		Short: "Opens the current stack in the Pulumi console",
```
praneetloke, line 32:
```suggestion
		Short: "Opens the current stack in the Pulumi Console",
```

pkg/cmd/pulumi/console.go
```diff
@@ -0,0 +1,81 @@
...
+
+			// Do a type assertion in order to determine if this is a cloud backend based on whether the assertion
+			// succeeds or not.
+			cloudBackend, isCloud := backend.(httpstate.Backend)
+			if isCloud {
+				// Open the stack specific URL (e.g. app.pulumi.com/{org}/{project}/{stack}) for this
+				// stack if a stack is selected and is a cloud stack, else open the cloud backend URL
+				// home page, e.g. app.pulumi.com.
+				if s, ok := stack.(httpstate.Stack); ok {
+					if consoleURL, err := s.ConsoleURL(); err == nil {
```
praneetloke, line 55: I assume this works with the self-hosted Pulumi Service as well?

praneetloke (commented):
Overall looks great to me. I'll defer the approval to @justinvp so that he has had a chance to look at it as well.

pkg/cmd/pulumi/console.go
sean1588: yes it does work with self-hosted as well.

CHANGELOG.md
```diff
@@ -35,6 +35,9 @@ CHANGELOG
 - Automation API - add recovery APIs (cancel/export/import)
   [#5369](https://github.com/pulumi/pulumi/pull/5369)
 
+- Add `pulumi console` command which opens the currently selected stack in the Pulumi console.
```
justinvp, line 38: This will need to be moved up to HEAD since 2.10.1 has already been released.

pkg/cmd/pulumi/console.go
```diff
@@ -0,0 +1,81 @@
...
+						// Open the cloud backend home page if retrieving the stack
+						// console URL fails.
+						launchConsole(cloudBackend.URL())
+					}
+				} else {
+					launchConsole(cloudBackend.URL())
+				}
+				return nil
+			}
+			fmt.Println("This command is not available for your backend." +
```
justinvp:
```suggestion
			fmt.Println("This command is not available for your backend. " +
```

pkg/cmd/pulumi/console.go
```diff
@@ -0,0 +1,82 @@
...
+						// console URL fails.
+						launchConsole(cloudBackend.URL())
+					}
+				} else {
+					launchConsole(cloudBackend.URL())
+				}
+				return nil
+			}
+			fmt.Println("This command is not available for your backend. " +
+				"To migrate to the Pulumi managed backend, " +
```
justinvp, line 68: Should this be "Pulumi Service backend"?

```suggestion
				"To migrate to the Pulumi Service backend, " +
```

pkg/cmd/pulumi/console.go
```diff
@@ -0,0 +1,82 @@
...
+				"please see https://www.pulumi.com/docs/intro/concepts/state/#adopting-the-pulumi-service-backend")
+			return nil
+		}),
+	}
+	return cmd
+}
+
+// launchConsole attmepts to open the console in the browser using the specified URL.
```
justinvp, line 76: Nit: Typo:

```suggestion
// launchConsole attempts to open the console in the browser using the specified URL.
```

justinvp (approved):
LGTM

CHANGELOG.md
sean1588: Ahh yes, thank you.

pkg/cmd/pulumi/console.go
sean1588: Yeah that would make more sense since that is what it is called in the doc we are linking to. thanks.

Issue #5362, opened 2020-09-14
`pulumi cancel` ignores `--non-interactive`


EvanBoyle:
I'd expect it to fail without passing `--yes`, but it still prompts in `--non-interactive` mode:

```shell
$ pulumi cancel --non-interactive
This will irreversibly cancel the currently running update for 'prod'!
Please confirm that this is what you'd like to do by typing ("prod"): ^C
```

Issue #5296, opened 2020-09-08
[Automation API] add pulumi cancel


EvanBoyle:
A user pointed out that this was missing over the weekend. Most common reason I've seen for needing this with automation API is ctrl+c when running an inline program as this can leave the host hanging. 

Along with this we should consider adding stack export and stack import commands, as canceling an update could require stack surgery to get things back in a good place. 

We should also decide where these methods live (stack or workspace). My intuition is that cancel should be on stack and the other two on workspace, but we should design this intentionally.

PR #5369, opened 2020-09-15
Automation API - add recovery APIs (cancel/export/import)

EvanBoyle:
This PR provides coverage over the Pulumi "recovery" APIs. `pulumi cancel` `pulumi stack export` `pulumi stack import`. We'll improve our story over time, but for now, we'll start with exposing the existing tools. 

Note that I exposed import/export as workspace level operations as their implementation requires intermediate on-disk files (this is the characteristic that defines local workspace). We can revise this in the future by extending the command to accept JSON input from a flag. 

Closes: https://github.com/pulumi/pulumi/issues/5362
Closes: https://github.com/pulumi/pulumi/issues/5296

### DIFF
--- a/sdk/go/x/auto/example_test.go
+++ b/sdk/go/x/auto/example_test.go
@@ -17,6 +17,7 @@ package auto
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -25,6 +26,7 @@ import (
 	"path/filepath"
 
 	"github.com/pkg/errors"
+	"github.com/pulumi/pulumi/sdk/v2/go/common/apitype"
 	"github.com/pulumi/pulumi/sdk/v2/go/common/tokens"
 	"github.com/pulumi/pulumi/sdk/v2/go/common/workspace"
 	"github.com/pulumi/pulumi/sdk/v2/go/pulumi"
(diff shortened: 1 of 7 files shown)


sdk/go/x/auto/example_test.go
```diff
@@ -1073,3 +1075,98 @@ func ExampleStack_SetAllConfig() {
...
+	_ = stack.Cancel(ctx)
+}
+
+func ExampleStack_Export() {
+	ctx := context.Background()
+	stackName := FullyQualifiedStackName("org", "project", "stack")
+	stack, _ := SelectStackLocalSource(ctx, stackName, filepath.Join(".", "program"))
+	dep, _ := stack.Export(ctx)
+	// import/export is backwards compatible, and we must write code specific to the verison we're dealing with.
+	if dep.Version != 3 {
```
metral, line 1094: It feels like this check should happen in `stack.{Export,Import}` instead of putting the onus on the user

sdk/go/x/auto/example_test.go
```diff
@@ -1073,3 +1075,98 @@ func ExampleStack_SetAllConfig() {
 	// set all config in map
 	_ = stack.SetAllConfig(ctx, cfg)
 }
+
+func ExampleStack_Cancel() {
+	ctx := context.Background()
+	stackName := FullyQualifiedStackName("org", "project", "stack")
```
metral, line 1081: we should show an example for local backends that don't use the FQSN

sdk/go/x/auto/workspace.go
```diff
@@ -96,6 +97,12 @@ type Workspace interface {
 	Program() pulumi.RunFunc
 	// SetProgram sets the program associated with the Workspace to the specified `pulumi.RunFunc`.
 	SetProgram(pulumi.RunFunc)
+	// ExportStack exports the deployment state of the stack matching the given name.
+	// This can be combined with ImportStack to edit a stack's state (such as recovery from failed deployments).
+	ExportStack(context.Context, string) (apitype.UntypedDeployment, error)
+	// ImportStack imports the specified deployment state into a pre-existing stack.
+	// This can be combined with ExportStack to edit a stack's state (such as recovery from failed deployments).
+	ImportStack(context.Context, string, apitype.UntypedDeployment) error
```
metral, line 105: `Cancel` is missing from the interface

sdk/go/x/auto/local_workspace.go
```diff
@@ -409,6 +411,62 @@ func (l *LocalWorkspace) SetProgram(fn pulumi.RunFunc) {
...
+	bytes, err := json.Marshal(state)
+	if err != nil {
+		return errors.Wrap(err, "could not import stack, failed to marshal stack state.")
+	}
+
+	_, err = f.Write(bytes)
+	if err != nil {
+		return errors.Wrap(err, "could not import stack. failed to write out stack intermediate.")
+	}
+
```
metral, line 461:
```suggestion
// Close the file
if err := f.Close(); err != nil {
  return errors.Wrap(err, "could not import stack. failed to close out stack intermediate")
}
```

sdk/go/x/auto/example_test.go
```diff
@@ -1073,3 +1075,98 @@ func ExampleStack_SetAllConfig() {
...
+
+	// ... perform edits on the state ...
+
+	// marshal out updated deployment state
+	bytes, _ := json.Marshal(state)
+	dep.Deployment = bytes
+	// import our edited deployment state back to our stack
+	_ = stack.Import(ctx, dep)
+}
+
```
metral, line 1129: We should consider adding an example of working _through_ `pending_operations` e.g.

- export state
- delete pending resources from the state
- import updated state object
- instruct the user to manually clean up any leaked provider resources

metral (changes requested):
These additions are great!

Added some points to think through 👍

sdk/go/x/auto/workspace.go
EvanBoyle: Cancel is exposed on stack and not workspace similar to Up/Preview/Refresh/Destroy

sdk/go/x/auto/example_test.go
EvanBoyle: Without generics, I think this is the best we get if we want this API to be backwards and forwards compatible. We could potentially add helper methods `ImportV1/ImportV2/ImportV3`, but I'd prefer to do the minimum here for now.

sdk/go/x/auto/example_test.go
EvanBoyle: Good point. I'd like to invert the solution a little though. Rather than having full coverage of examples for the caveats of SaaS vs non-saas backends. Let's keep all examples as is assuming usage of SaaS backend, and invest in a few end to end high polish examples for local backends (Stack and LocalWorkspace level examples that show the end to end workflow). 

I've captured this here: https://github.com/pulumi/pulumi/issues/5381
Perhaps @stack72 might be able to take a look as a part of the OSS backend audit.

sdk/go/x/auto/local_workspace.go
EvanBoyle: I believe this is handled by the defer a few lines up https://github.com/pulumi/pulumi/pull/5369/files#diff-226d1106206f1bf744142667c0ade384R450

sdk/go/x/auto/example_test.go
EvanBoyle: A simple export/import does clear pending operations according to our docs: https://www.pulumi.com/docs/troubleshooting/#interrupted-update-recovery

I do agree that our story here is weak. My goal here is to just expose the building blocks. We know we'll be investing in the product to make this experience better (and the current cancel/import/export workflow is not the answer).

EvanBoyle:
Followed up with Mike offline and we're shipping this as-is for now. We have planned work on failure recovery post-launch.

PR #5373, opened 2020-09-15
[codegen] Adjust Go/Python package naming for Azure NextGen

mikhailshilkov:
No other providers use dashes in their names except the new `azure-nextgen` one. This PR normalizes the folder structure 

- In Go, the package `azure` gets generated (the module is still `azure-nextgen`
- In Python, the module is `pulumi_azure_nextgen`

### DIFF
--- a/pkg/codegen/go/gen.go
+++ b/pkg/codegen/go/gen.go
@@ -992,7 +992,7 @@ func (pkg *pkgContext) genHeader(w io.Writer, goImports []string, importedPackag
 
 	var pkgName string
 	if pkg.mod == "" {
-		pkgName = pkg.pkg.Name
+		pkgName = goPackage(pkg.pkg.Name)
 	} else {
 		pkgName = path.Base(pkg.mod)
 	}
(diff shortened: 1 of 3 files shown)


pkg/codegen/go/gen.go
```diff
@@ -1379,6 +1379,11 @@ func GeneratePackage(tool string, pkg *schema.Package) (map[string][]byte, error
 	return files, nil
 }
 
+// goPackage returns the suggested package name for the given string.
+func goPackage(name string) string {
+	return strings.Split(name, "-")[0]
```
stack72, line 1384: should we be defensive here and check to see if there are any items after the split?

mikhailshilkov: do you mean when `name` is `""`?

mikhailshilkov: Added a test case for `""`. I think `Split` would always return one item if a separator is specified.

stack72: yeah

stack72 (approved):

PR #5379, opened 2020-09-15
[codegen/schema] - Expand ObjectTypeSpec to ComplexTypeSpec

komalali:
Relax the constraints around ObjectTypeSpec and rename to ComplexTypeSpec to allow for other complex types (i.e. enums) to be included in the types section of the schema. Note: downstream codegen is failing because tfbridge needs changes as a follow-up from this PR (see linked PRs below).

### Downstream changes
- [pulumi-kubernetes](https://github.com/pulumi/pulumi-kubernetes/pull/1322/files)
- [pulumi-terraform-bridge](https://github.com/pulumi/pulumi-terraform-bridge/pull/266)

### Updated schema
Here is an example of what the updated types section will look like with both object and enum types represented.
```jsonc
{
    "types": {
		// A traditional object type
        "azurerm:labservices/v20181015:ResourceSettings": {
            "description": "Represents resource specific settings",
            "properties": {
                "galleryImageResourceId": {
                    "type": "string",
                    "description": "The resource id of the gallery image used for creating the virtual machine"
                },
                "referenceVm": {
                    "$ref": "#/types/azurerm:labservices/v20181015:ReferenceVm",
                    "description": "Details specific to Reference Vm"
                },
                "size": {
                    // Representation of a “relaxed” enum will use “oneOf”, while a strict enum will simply use “$ref”
                    "oneOf": [
                        {"$ref": "azurerm:compute/v20190701:ManagedLabVmSize"},
                        {"type": "string"}
                    ],
                    "description": "The size of the virtual machine"
                }
            },
            "type": "object",
            "required": [
                "referenceVm"
            ]
        },
        // Representation of the enum type itself
        "azurerm:labservices/v20181015:ManagedLabVmSize": {
            "type": "string",
            "enum": [
                {
                    "value": "Basic",
                    "description":"The base VM size"
                },
                {
                    "value": "Standard",
                    "description": "The standard or default VM size"
                },
                {
                    "value": "Performance",
                    "description": "The most performant VM size"
                }
            ]
        }
    }
}
```

### DIFF
--- a/pkg/codegen/docs/gen_test.go
+++ b/pkg/codegen/docs/gen_test.go
@@ -66,7 +66,7 @@ func initTestPackageSpec(t *testing.T) {
 		Meta: &schema.MetadataSpec{
 			ModuleFormat: "(.*)(?:/[^/]*)",
 		},
-		Types: map[string]schema.ObjectTypeSpec{
+		Types: map[string]schema.ComplexTypeSpec{
 			// Package-level types.
 			"prov:/getPackageResourceOptions:getPackageResourceOptions": {
 				Description: "Options object for the package-level function getPackageResource.",
@@ -123,7 +123,7 @@ func initTestPackageSpec(t *testing.T) {
 		},
 		Resources: map[string]schema.ResourceSpec{
 			"prov:module/resource:Resource": {
-				ObjectTypeSpec: schema.ObjectTypeSpec{
+				ComplexTypeSpec: schema.ComplexTypeSpec{
 					Description: `This is a module-level resource called Resource.
 {{% examples %}}
 ## Example Usage
@@ -189,7 +189,7 @@ func initTestPackageSpec(t *testing.T) {
 				},
 			},
 			"prov:/packageLevelResource:PackageLevelResource": {
-				ObjectTypeSpec: schema.ObjectTypeSpec{
+				ComplexTypeSpec: schema.ComplexTypeSpec{
 					Description: "This is a package-level resource.",
 				},
 				InputProperties: map[string]schema.PropertySpec{
@@ -206,7 +206,7 @@ func initTestPackageSpec(t *testing.T) {
 			// Package-level Functions.
 			"prov:/getPackageResource:getPackageResource": {
 				Description: "A package-level function.",
-				Inputs: &schema.ObjectTypeSpec{
+				Inputs: &schema.ComplexTypeSpec{
 					Description: "Inputs for getPackageResource.",
 					Type:        "object",
 					Properties: map[string]schema.PropertySpec{
@@ -217,7 +217,7 @@ func initTestPackageSpec(t *testing.T) {
 						},
 					},
 				},
-				Outputs: &schema.ObjectTypeSpec{
+				Outputs: &schema.ComplexTypeSpec{
 					Description: "Outputs for getPackageResource.",
 					Properties:  simpleProperties,
 					Type:        "object",
@@ -227,7 +227,7 @@ func initTestPackageSpec(t *testing.T) {
 			// Module-level Functions.
 			"prov:module/getModuleResource:getModuleResource": {
 				Description: "A module-level function.",
-				Inputs: &schema.ObjectTypeSpec{
+				Inputs: &schema.ComplexTypeSpec{
 					Description: "Inputs for getModuleResource.",
 					Type:        "object",
 					Properties: map[string]schema.PropertySpec{
@@ -238,7 +238,7 @@ func initTestPackageSpec(t *testing.T) {
 						},
 					},
 				},
-				Outputs: &schema.ObjectTypeSpec{
+				Outputs: &schema.ComplexTypeSpec{
 					Description: "Outputs for getModuleResource.",
 					Properties:  simpleProperties,
 					Type:        "object",
--- a/pkg/codegen/docs/gen_test.go
+++ b/pkg/codegen/docs/gen_test.go
@@ -69,53 +69,61 @@ func initTestPackageSpec(t *testing.T) {
 		Types: map[string]schema.ComplexTypeSpec{
 			// Package-level types.
 			"prov:/getPackageResourceOptions:getPackageResourceOptions": {
-				Description: "Options object for the package-level function getPackageResource.",
-				Type:        "object",
-				Properties:  simpleProperties,
+				ObjectTypeSpec: schema.ObjectTypeSpec{
+					Description: "Options object for the package-level function getPackageResource.",
+					Type:        "object",
+					Properties:  simpleProperties,
+				},
 			},
 
 			// Module-level types.
 			"prov:module/getModuleResourceOptions:getModuleResourceOptions": {
-				Description: "Options object for the module-level function getModuleResource.",
-				Type:        "object",
-				Properties:  simpleProperties,
+				ObjectTypeSpec: schema.ObjectTypeSpec{
+					Description: "Options object for the module-level function getModuleResource.",
+					Type:        "object",
+					Properties:  simpleProperties,
+				},
 			},
 			"prov:module/ResourceOptions:ResourceOptions": {
-				Description: "The resource options object.",
-				Type:        "object",
-				Properties: map[string]schema.PropertySpec{
-					"stringProp": {
-						Description: "A string prop.",
-						Language:    pythonMapCase,
-						TypeSpec: schema.TypeSpec{
-							Type: "string",
+				ObjectTypeSpec: schema.ObjectTypeSpec{
+					Description: "The resource options object.",
+					Type:        "object",
+					Properties: map[string]schema.PropertySpec{
+						"stringProp": {
+							Description: "A string prop.",
+							Language:    pythonMapCase,
+							TypeSpec: schema.TypeSpec{
+								Type: "string",
+							},
 						},
-					},
-					"boolProp": {
-						Description: "A bool prop.",
-						Language:    pythonMapCase,
-						TypeSpec: schema.TypeSpec{
-							Type: "boolean",
+						"boolProp": {
+							Description: "A bool prop.",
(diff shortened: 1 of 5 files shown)


pkg/codegen/schema/schema.go
```diff
@@ -607,21 +637,33 @@ type PropertySpec struct {
 	Secret bool `json:"secret,omitempty"`
 }
 
-// ObjectTypeSpec is the serializable form of an object type.
-type ObjectTypeSpec struct {
+// ComplexTypeSpec is the serializable form of an object or enum type.
+type ComplexTypeSpec struct {
```
pgavlin, line 641: I don't think that we should rename this type. Instead, I think that we should define a new type that embeds an `ObjectTypeSpec` and adds the enum-related fields.

pgavlin: For example:

```go
type ComplexTypeSpec struct {
    ObjectTypeSpec

    Enum []EnumValueSpec `json:"enum,omitempty"`
}
```

Then we can define the schema with stronger typing.

pkg/codegen/schema/schema.go
```diff
@@ -128,6 +130,34 @@ func (t *ArrayType) String() string {
...
+// EnumType represents an enum.
+type EnumType struct {
+	// Token is the type's Pulumi type token.
+	Token string
+	// Comment is the description of the type, if any.
+	Comment string
+	// Elements are the predefined enum values.
+	Elements []*Enum
+	// ElementType is the underlying type for the enum.
+	ElementType string
```
pgavlin, line 142: It feels like this should be a `Type` rather than a `string`.

pkg/codegen/schema/schema.go
```diff
@@ -128,6 +130,34 @@ func (t *ArrayType) String() string {
...
+}
+
+// Enum contains information about an enum.
+type Enum struct {
+	// Value is the value of the enum.
+	Value interface{}
+	// Description for the enum value.
+	Description string
```
pgavlin, line 150: Nit: let's use `Comment` rather than `Description` for consistency.

pkg/codegen/schema/schema.go
```diff
@@ -611,17 +641,35 @@ type PropertySpec struct {
...
 	// Language specifies additional language-specific data about the type.
 	Language map[string]json.RawMessage `json:"language,omitempty"`
 }
 
+// ComplexTypeSpec is the serializable form of an object or enum type.
+type ComplexTypeSpec struct {
+	ObjectTypeSpec
+
+	// Enum, if present, if the list of possible values for an enum type.
+	Enum []*EnumValueSpec `json:"enum,omitempty"`
```
pgavlin, line 660: In this case, we should also require that the `Type` field is `"enum"`.

pkg/codegen/schema/schema.go
```diff
@@ -611,17 +641,35 @@ type PropertySpec struct {
...
+}
+
+// EnumValuesSpec is the serializable form of the values metadata associated with an enum type.
+type EnumValueSpec struct {
+	// Name, if present, overrides the name of the enum value that would usually be derived from the value.
+	Name string `json:"name,omitempty"`
+	// Description of the enum value
+	Description string `json:"description,omitempty"`
+	// Value is the enum value itself and must match an element in the Enum[] list.
```
pgavlin, line 669: > must match an element in the Enum[] list.

Can you clarify? These values compose the `Enum[]` list, right?

pkg/codegen/schema/schema.go
```diff
@@ -1129,33 +1182,69 @@ func (t *types) bindObjectType(token string, spec ObjectTypeSpec) (*ObjectType,
...
+	enum.ElementType = spec.Type
+	enum.Comment = spec.Description
+
+	return nil
+}
+
+func (t *types) bindEnumValues(values []*EnumValueSpec) (enums []*Enum) {
+	for _, spec := range values {
+		enum := &Enum{
+			Value:       spec.Value,
```
pgavlin, line 1197: It would be good to check that the type of `Value` is assignable to the element type of the enum. You can do that in a follow-up change, though.

pkg/codegen/schema/schema.go
```diff
@@ -52,6 +52,8 @@ const (
 	jsonType    primitiveType = 8
 )
 
+const object = "object"
```
pgavlin, line 55: IMO this sort of constant is not particularly helpful. Was `goconst` complaining?

komalali: yeah I didn't like it either, I can just turn off `goconst` instead.

pkg/codegen/schema/schema.go
komalali: this is a leftover comment from when the enumMetadata was a separate object, will fix up.

pkg/codegen/schema/schema.go
komalali: The Type field for enums is usually used to capture the underlying primitive type, i.e. `string`. I was trying to keep this consistent with the [spec](https://swagger.io/docs/specification/data-models/enums/#ctxM:~:text=schema%3A,%20%20%20%20%20%20%20%20%20enum%3A%20%5Basc%2C%20desc%5D,-In). I flip-flopped on this a bunch, are you suggesting there should be an `ElementType` and a `Type` in the json schema?

pkg/codegen/schema/schema.go
```diff
@@ -128,6 +128,34 @@ func (t *ArrayType) String() string {
 
 func (*ArrayType) isType() {}
 
+// EnumType represents an enum.
+type EnumType struct {
+	// Token is the type's Pulumi type token.
+	Token string
+	// Comment is the description of the type, if any.
+	Comment string
```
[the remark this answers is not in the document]
komalali: Lol https://github.com/pulumi/pulumi/pull/5379#discussion_r489084524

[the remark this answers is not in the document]
mikhailshilkov: 🤣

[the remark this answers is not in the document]
lblackstone: @pgavlin suggested we change this to Comment
https://github.com/pulumi/pulumi/pull/5379#discussion_r489084524

pkg/codegen/schema/schema.go
```diff
@@ -1129,33 +1181,75 @@ func (t *types) bindObjectType(token string, spec ObjectTypeSpec) (*ObjectType,
...
-		// object type is its token. While this doesn't affect object types directly, it breaks the interning of types
-		// that reference object types (e.g. arrays, maps, unions)
-		typs.objects[token] = &ObjectType{Token: token}
 	}
 
 	// Process properties.
-	for token, spec := range objects {
-		if err := typs.bindObjectTypeDetails(typs.objects[token], token, spec); err != nil {
-			return nil, errors.Wrapf(err, "failed to bind type %s", token)
+	for token, spec := range complexTypes {
```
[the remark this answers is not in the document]
komalali: They are pre-existing but my impression from the comments is that we want to declare all the object tokens first before processing their properties since the properties might otherwise reference object types that haven't been declared yet? I guess this isn't really relevant for enums so those could be processed fully in the first loop. @pgavlin am I understanding this correctly?

pkg/codegen/docs/gen_test.go
```diff
@@ -66,56 +66,64 @@ func initTestPackageSpec(t *testing.T) {
 		Meta: &schema.MetadataSpec{
 			ModuleFormat: "(.*)(?:/[^/]*)",
 		},
-		Types: map[string]schema.ObjectTypeSpec{
+		Types: map[string]schema.ComplexTypeSpec{
```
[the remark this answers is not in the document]
lblackstone: Just chatted with @komalali offline on this, and I misunderstood the change. This looks correct. 👍

pkg/codegen/schema/schema.go
```diff
@@ -710,8 +756,9 @@ type PackageSpec struct {
 
 	// Config describes the set of configuration variables defined by this package.
 	Config ConfigSpec `json:"config"`
-	// Types is a map from type token to ObjectTypeSpec that describes the set of object types defined by this package.
-	Types map[string]ObjectTypeSpec `json:"types,omitempty"`
+	// Types is a map from type token to ComplexTypeSpec that describes the set of complex types (ie. object, enum)
+	// defined by this package.
+	Types map[string]ComplexTypeSpec `json:"types,omitempty"`
```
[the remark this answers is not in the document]
komalali: This has to change to `ComplexTypeSpec` to allow enums to also fit into the types section. The `ObjectTypeSpec` is embedded in `ComplexTypeSpec`, so the json representation doesn't change but we do need to update how they are defined. See the [related k8s PR](https://github.com/pulumi/pulumi-kubernetes/pull/1322/files#diff-3573d0e00d52aadffe503ad3d646a022) for what needs to change.

pkg/codegen/schema/schema.go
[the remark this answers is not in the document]
pgavlin: That's correct, yes.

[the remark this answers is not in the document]
pgavlin: (future changes will move this to a depth-first approach to accommodate #5385)

pkg/codegen/schema/schema.go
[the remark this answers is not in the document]
pgavlin: @komalali has it right--the wire format is backwards-compatible, but consumers of the API will need to update their code a little bit when constructing these values.

pkg/codegen/schema/schema.go
pgavlin: > I was trying to keep this consistent with the spec. 

Ah, perfect! That works, then.

pgavlin (approved):

mikhailshilkov (approved):

PR #5386, opened 2020-09-15
Switch master builds to github actions

jaxxstorm:
This switches master builds to using github actions rather than travis


stack72 (approved):
Let's merge this in and iterate now that today's release is out

PR #5399, opened 2020-09-17
[codegen/nodejs] Format dashed import for invokes

mikhailshilkov:
Invokes from `azure-nextgen` are currently imported as `import * as azure_nextgen from "@pulumi/azure_nextgen";`.

This change fixes them to `import * as azure_nextgen from "@pulumi/azure-nextgen";`.


mikhailshilkov:
Downstream Codegen failures are not related

viveklak (approved):

Issue #5228, opened 2020-08-26
[codegen/hcl] Stack overflow on recursive types


mikhailshilkov:
Recursive types in the schema cause the program codegen to fail with a stack overflow panic:

```
runtime: goroutine stack exceeds 1000000000-byte limit
runtime: sp=0xc0239e0358 stack=[0xc0239e0000, 0xc0439e0000]
fatal error: stack overflow

runtime stack:
runtime.throw(0x1d3533c, 0xe)
	/usr/local/Cellar/go/1.14.3/libexec/src/runtime/panic.go:1116 +0x72
runtime.newstack()
	/usr/local/Cellar/go/1.14.3/libexec/src/runtime/stack.go:1034 +0x6ce
runtime.morestack()
	/usr/local/Cellar/go/1.14.3/libexec/src/runtime/asm_amd64.s:449 +0x8f

goroutine 54 [running]:
runtime.mallocgc(0x30, 0x1cd5cc0, 0x1, 0x0)
	/usr/local/Cellar/go/1.14.3/libexec/src/runtime/malloc.go:891 +0x9bb fp=0xc0239e0368 sp=0xc0239e0360 pc=0x100ee1b
runtime.newobject(0x1cd5cc0, 0x0)
	/usr/local/Cellar/go/1.14.3/libexec/src/runtime/malloc.go:1165 +0x38 fp=0xc0239e0398 sp=0xc0239e0368 pc=0x100efe8
runtime.makemap_small(0x0)
	/usr/local/Cellar/go/1.14.3/libexec/src/runtime/map.go:293 +0x2d fp=0xc0239e03b8 sp=0xc0239e0398 pc=0x100fc6d
github.com/pulumi/pulumi/pkg/v2/codegen/hcl2.(*binder).schemaTypeToType(0xc0022c4960, 0x1f444e0, 0xc0022f24b0, 0x0, 0x0)
	/Users/mikhailshilkov/go/src/github.com/pulumi/pulumi/pkg/codegen/hcl2/binder_schema.go:158 +0x107 fp=0xc0239e0500 sp=0xc0239e03b8 pc=0x1a89eb7
```

The repro is available in [this branch](https://github.com/pulumi/pulumi/compare/mikhailshilkov/programgen-recursivetypes?expand=1)

mikhailshilkov:
I believe this is not fully fixed yet: I got the following stack overflow when trying to generate Go examples in azure-nextgen:

```
fatal error: stack overflow

[...]

PR #5404, opened 2020-09-18
Support recursive object types in pkg/codegen model

lukehoban:
Fixes https://github.com/pulumi/pulumi/issues/5228.

I am not positive this addresses every place in `pkg/codegen/hcl2/model` where we assume types will be non-recursive, but it addresses all the cases that appear to currently be blocking the new `arm2pulumi` tool.

### DIFF
--- a/pkg/codegen/hcl2/model/type.go
+++ b/pkg/codegen/hcl2/model/type.go
@@ -35,7 +35,7 @@ func (k ConversionKind) Exists() bool {
 type Type interface {
 	Definition
 
-	Equals(other Type) bool
+	Equals(other Type, seen map[Type]struct{}) bool
 	AssignableFrom(src Type) bool
 	ConversionFrom(src Type) ConversionKind
 	String() string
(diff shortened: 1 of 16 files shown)


mikhailshilkov (approved):
Looks great, that's very helpful!

viveklak (approved):
Can confirm it unblocks arm2pulumi on a bunch of resources.

pkg/codegen/hcl2/model/type.go
```diff
@@ -35,7 +35,7 @@ func (k ConversionKind) Exists() bool {
 type Type interface {
 	Definition
 
-	Equals(other Type) bool
+	Equals(other Type, seen map[Type]struct{}) bool
```
pgavlin, line 38: Do you think that we can replace this with an unexported version of the same s.t. the public API remains identical?

```suggestion
	Equals(other Type) bool
	equals(other Type, seen map[Type]struct{}) bool
```

Also, is referential identity okay for the set, given that types with different references can have the same identity?

lukehoban: > Do you think that we can replace this with an unexported version of the same s.t. the public API remains identical?

We can.  It unfortunately means a lot of boilerplate in every implementation (since they all need identical implementations of `Equals`.  But that's fine I suppose.  I wasn't sure what the real "public contract" was here as there are very few external callers of this (I only found one).

I'll make the change though to avoid any public API churn.

> Also, is referential identity okay for the set, given that types with different references can have the same identity?

The set watches the visited receiver types, but doesn't pay attention to the "other" types.  So the only risk here would be that somehow there was recursion in the receiver type graph that didn't have referential identity.  I honestly don't *think* that should be possible.   Unless I am misunderstanding the question?

PR #5406, opened 2020-09-18
Fixed issue #5014 (Missing Pulumi console tags on Windows)

### LINKED ISSUE #5014: Pulumi cli on Windows does not include all vcs and Github tags to Pulumi console
When running Pulumi commands locally there is a difference between doing so on Windows and Linux. In Linux more tag information is included to Pulumi console. This results in commit information and repo information not being included with the different Pulumi commands losing some information for the different stacks.

These are the tags included on Linux: 
image

These are the tags included on Windows:
image

jetvova:
The existing `path.Join(gitRoot, "..")` function appears to have unexpected behavior on Windows. Instead of appending `..` and returning gitRoot's parent directory, it replaces the whole string with a single dot. The fix changes the code to instead use `filepath.Dir(gitRoot)`, which correctly returns the parent directory.

This seems to be the only instance in which `path.Join()` is used with `..` in the entire repository. There are some usages of a function called `filepath.Join()` with dots, but they work correctly.

Fixes #5014

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ CHANGELOG
 - [sdk/go] Add missing Version field to invokeOptions
   [#5401](https://github.com/pulumi/pulumi/pull/5401)
 
+- Pulumi Windows CLI now uploads all VCS information to console
+  (fixes [#5014](https://github.com/pulumi/pulumi/issues/5014))
+  [#5406](https://github.com/pulumi/pulumi/pull/5406)
+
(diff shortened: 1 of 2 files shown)


justinvp (approved):
LGTM. Want to add a note to CHANGELOG.md?

Also, for future reference, if you had a `Fixes #<insert issue number>` anywhere in the PR description, the issue will automatically be closed when the PR is merged. (I went ahead and edited it to do that.)

PR #5419, opened 2020-09-19
[sdk/python] Cast numbers intended to be integers to int

### LINKED ISSUE #5418: Python: Functions that return numbers result in `TypeError`
From Luke:

> When I try to call `network.latest.get_public_ip_address` in Python I get an error about `TypeError: Expected argument 'idle_timeout_in_minutes' to be a int`.  This error comes from the construction of `GetPublicIPAddressResult` which does a type test on the data [...] that comes back that it is an `int` - but it is actually a `float`.

This is a regression caused by https://github.com/pulumi/pulumi/pull/5309. It's due to protobuf structs treating number values as floats (like JSON). We'll need to cast number values intended to be integers to `int` in such cases.

Note: This technically affects all number values we unmarshal from protobuf structs, but we only have the strict `isinstance` type checks for function call results, which leads to this runtime error.

justinvp:
We recently made a [change](https://github.com/pulumi/pulumi/pull/5309) to the Python codegen to emit `int` type annotations, instead of `float`, for properties that are typed as `schema.IntType`.

But the number values that come back from protobuf structs are always floats (like JSON), so we need to cast the values intended to be integers to `int`.

Fixes #5418

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ CHANGELOG
 - Add `pulumi console` command which opens the currently selected stack in the Pulumi console.
   [#5368](https://github.com/pulumi/pulumi/pull/5368)
 
+- Python SDK: Cast numbers intended to be integers to `int`.
+  [#5419](https://github.com/pulumi/pulumi/pull/5419)
+
 ## 2.10.1 (2020-09-16)
 
 - feat(autoapi): add GetPermalink for operation result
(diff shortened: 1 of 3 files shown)


justinvp:
I tried out the Python example being added in https://github.com/pulumi/examples/pull/799. Before these changes, I observed the same error mentioned in #5418. Then I installed a local build of the Python SDK with the changes from this PR and the `pulumi up` worked without error.

komalali (approved):

PR #5421, opened 2020-09-20
[codegen/docs] Don't emit API links in Azure NextGen resource docs

justinvp:
We're not going to generate language-specific API docs for the Azure NextGen provider, only resource docs. This change makes it so the resource docs do not emit any links to nonexistent API docs.

Part of https://github.com/pulumi/docs/issues/4071


mikhailshilkov (approved):
Looks great, thanks for doing that.

PR #5443, opened 2020-09-22
Add support for provider-side preview.

### LINKED ISSUE #4992: Confusing transitive resource updates when dependencies have changes
If a resource is updated it can cause confusing and invalid updates in the preview for downstream resources that depend on any of its outputs. As a concrete example, if I provision an AWS KMS Key and an RDS Cluster that uses the KMS Key ID and then update tags on the KMS Key, Pulumi will indicate the RDS Cluster will be replaced, but this doesn't actually happen.

Note: This is a general issue, not specific to AWS KMS or RDS.

#### Suggestions for a fix

Only indicate changes to downstream resources that will actually occur.

#### Code to reproduce

```ts

const kmsKey = new aws.kms.Key("main", {
    tags: {
        /**
         * Change any of these tags and the RDS Cluster will be marked as `replace`, but won't actually happen.
         */
        // k1: "v1",
        k2: "v2",
        // k3: "v3",
    }
});

const defaultCluster = new aws.rds.Cluster("default", {

    storageEncrypted: true,
    kmsKeyId: kmsKey.arn,

    availabilityZones: [
        "us-west-2a",
        "us-west-2b",
        "us-west-2c",
    ],
    backupRetentionPeriod: 5,
    clusterIdentifier: "aurora-cluster-demo",
    databaseName: "mydb",
    engine: "aurora-mysql",
    engineVersion: "5.7.mysql_aurora.2.03.2",
    masterPassword: "o209dsjvnlasdkfjasldf;.oad",
    masterUsername: "foo",
    preferredBackupWindow: "07:00-09:00",
    skipFinalSnapshot: true,
}, {
    protect: false,
    // ignoreChanges: ["kmsKeyId"],
});
```

#### Steps to reproduce

1. `pulumi stack init`
1. `pulumi config set aws:region us-west-2`
1. `up` the code above
1. change tags on the `aws.kms.Key` resource
1. `up` again and observe the `replace` operation on the `aws:rds:Cluster` resource in the output

```
p-rds-kms-diff % pulumi up
Previewing update (cameorn):
     Type                 Name                    Plan        Info
     pulumi:pulumi:Stack  p-rds-kms-diff-cameorn              
 ~   ├─ aws:kms:Key       main                    update      [diff: ~tags]
 +-  └─ aws:rds:Cluster   default                 replace     [diff: ~kmsKeyId]
 
Resources:
    ~ 1 to update
    +-1 to replace
    2 changes. 1 unchanged

Updating (cameorn):
     Type                 Name                    Status      Info
     pulumi:pulumi:Stack  p-rds-kms-diff-cameorn              
 ~   └─ aws:kms:Key       main                    updated     [diff: ~tags]
 
Resources:
    ~ 1 updated
    2 unchanged

Duration: 4s

Permalink: https://app.pulumi.com/clstokes/p-rds-kms-diff/cameorn/updates/2
```

#### Workaround

A workaround for this is to use [`ignoreChanges`](https://www.pulumi.com/docs/intro/concepts/programming-model/#ignorechanges) on the downstream resource to ignore changes to the `kmsKeyId` in this case. Also, can use [`protect: true`](https://www.pulumi.com/docs/intro/concepts/programming-model/#protect) to protect against potential resource deletions as a safety mechanism.

pgavlin:
These changes add support for provider-side previews of create and
update operations, which allows resource providers to supply output
property values for resources that are being created or updated during a
preview.

If a plugin supports provider-side preview, its create/update methods
will be invoked during previews with the `preview` property set to true.
It is the responsibility of the provider to fill in any output
properties that are known before returning. It is a best practice for
providers to only fill in property values that are guaranteed to be
identical if the preview were instead an update (i.e. only those output
properties whose values can be conclusively determined without
actually performing the create/update operation should be populated).
Providers that support previews must accept unknown values in their
create and update methods.

If a plugin does not support provider-side preview, the inputs to a
create or update operation will be propagated to the outputs as they are
today.

Fixes #4992.


pgavlin:
Note that these changes are not quite complete: I'm in the process of adding tests and a configuration knob to allow users to disable provider previews.

pgavlin:
tfbridge changes: https://github.com/pulumi/pulumi-terraform-bridge/pull/276

mikhailshilkov:
I don't think I understand how to implement the provider-side preview in Azure NextGen. I have no TF to rely on, so what are the semantics of this operation, exactly?

#4992 is very much visible there, probably even more so than in TF providers. updating a tag on a resource group causes its `name` to go unknown which results in `replace` previews for all resources that referenced `resourceGroup.name` in their inputs.

Do I need to implement the provider-side preview by auto-populating all outputs that we consider "stable" (starting with names)?

pgavlin:
> I don't think I understand how to implement the provider-side preview in Azure NextGen. I have no TF to rely on, so what are the semantics of this operation, exactly?

When previewing an operation, a provider should fill in any property whose value can be determined without actually running the operation itself. For TF-based providers, we use TF's semantics; for k8s, we'll use server-side dry run. 

> Do I need to implement the provider-side preview by auto-populating all outputs that we consider "stable" (starting with names)?

That's the approach I would take, yes. @lukehoban suggested that ARM's `what-if` functionality might be able to help with completeness.

lukehoban:
> @lukehoban suggested that ARM's what-if functionality might be able to help with completeness.

It looks like unfortunately that functionality may be limited to templates - so might not be an option here. But I would assume we should be able to answer this conservatively from the schema?

mikhailshilkov:
> we should be able to answer this conservatively from the schema?

What do you mean? The schema has no notion of output stability (we double-checked with Azure folks). Conservatively, all properties can change during an update. We can start adding our heuristics on top of that: e.g. we know that `name` will never change.

CHANGELOG.md
```diff
@@ -3,6 +3,16 @@ CHANGELOG
 
 ## HEAD (Unreleased)
 
+- Support provider-side previews. These changes will admit more known property values during
+  `pulumi preview`, which should improve the accuracy of the proposed changes. Because the
+  Pulumi engine does not lock the operations performed in an update to those proposed during a
+  preview, it is possible for a misbehaving provider--i.e. a provider that produces one set of
+  output property values during a preview and a different set during an update--to cause
+  mismatches between a preview and an update. In these cases, provider-side previews can be
+  disabled entirely by setting the `PULUMI_DISABLE_PROVIDER_PREVIEW` environment variable to
```
lukehoban, line 12: I do not see any code that implements this.

CHANGELOG.md
```diff
@@ -3,6 +3,16 @@ CHANGELOG
 
 ## HEAD (Unreleased)
 
+- Support provider-side previews. These changes will admit more known property values during
```
lukehoban, line 6: I would perhaps change this whole thing to something more like:

> Improve the accuracy of previews by allowing providers to participate in determining what the impact of a change will be on output properties.  Previously, Pulumi previews conservatively assumed that any output-only properties changed their values when an update occurred.  For many properties, this was guaranteed to not be the case (because those properties are immutable for example), and by suggesting the value might change, this could lead to the preview suggesting additional transitive updates of even replaces that would not actually happen during an update.  Pulumi now allows the provider to specify the details of what properties will change during a preview, allowing them to expose more accurate provider-specific knowledge.  This change is less conservative than the previous behavior, and so in case it causes preview results which are not deemed correct in some case - the `PULUMI_DISABLE_PROVIDER_PREVIEW` flag can be set to a truthy value to go back to the previous and more conservative Pulumi preview behavior.

CHANGELOG.md
pgavlin: Yeah, this is still pending:

> I'm in the process of adding tests and a configuration knob to allow users to disable provider previews.

pkg/resource/deploy/providers/registry.go
```diff
@@ -304,14 +294,9 @@ func (r *Registry) Diff(urn resource.URN, id resource.ID, olds, news resource.Pr
...
 
 	// If the diff requires replacement, unload the provider: the engine will reload it during its replacememnt Check.
-	//
-	// If the diff does not require replacement but does have changes, and we are running a preview, register it under
-	// its current ID so that references to the provider from other resources will resolve properly.
 	if diff.Replace() {
 		closeErr := r.host.CloseProvider(provider)
 		contract.IgnoreError(closeErr)
-	} else if r.isPreview && diff.Changes == plugin.DiffSome {
-		r.setProvider(mustNewReference(urn, id), provider)
```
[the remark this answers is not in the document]
pgavlin: > What impact do the changes here have in practice on when providers are configured? I recall this was added originally to address a significant performance issue around configuring providers multiple times. But perhaps the removal of both this and the section above address that?

These changes should only push provider configuration later in the program, and only during previews. Instead of being configured in `Check` and `Diff` when previewing, the provider will be configured when the `Create` or `Update` step is performed. This should result in the same number of calls to `Configure`.

CHANGELOG.md
pgavlin: Added.

pgavlin:
Downstream codegen tests are failing due to a breaking API change; `pulumi-terraform-bridge` will need to be updated once these changes land.

pgavlin:
@lukehoban I am going to go ahead and merge this. I'll address any further feedback after the fact.

PR #5472, opened 2020-09-29
[codegen/nodejs] - Enums

komalali:
Resolves: https://github.com/pulumi/pulumi/issues/5433

WIP pulumi-aws PR: https://github.com/pulumi/pulumi-aws/pull/1151


pkg/codegen/internal/test/testdata/schema/simple-enum-schema.json
```diff
@@ -0,0 +1,93 @@
+{
+  "version": "0.0.1",
+  "name": "plant-provider",
```
pgavlin, line 3: 🌳 💯

pkg/codegen/nodejs/gen.go
```diff
@@ -112,9 +113,12 @@ func (mod *modContext) tokenToModName(tok string) string {
 	return modName
 }
 
-func (mod *modContext) tokenToType(tok string, input bool) string {
+func (mod *modContext) tokenToType(tok string, input, enum bool) string {
 	modName, name := mod.tokenToModName(tok), tokenToName(tok)
 
+	if enum {
+		return "enums." + modName + title(name)
+	}
```
pgavlin, line 121: Do we need to emit enums into their own namespace, or can they live in the same namespace as resources?

pkg/codegen/nodejs/gen_test.go
```diff
@@ -59,6 +60,51 @@ func TestGeneratePackage(t *testing.T) {
...
+			filepath.Join("schema", "simple-enum-schema.json"),
+			false,
+			func(files map[string][]byte) {
+				assert.Contains(t, files, "rubberTree.ts")
+				assert.Contains(t, files, "types/input.ts")
+				assert.Contains(t, files, "types/enums.ts")
+				assert.Contains(t, files, "types/index.ts")
+
+				for fileName, file := range files {
+					if fileName == "rubberTree.ts" {
```
pgavlin, line 74: Can we instead keep the expected contents around as either files or literal strings?

pgavlin (commented):
The core looks great to me; just a couple questions about namespacing and tests

pkg/codegen/nodejs/gen.go
```diff
@@ -987,6 +993,7 @@ func (mod *modContext) sdkImports(nested, utilities bool) []string {
 	if nested {
 		imports = append(imports, fmt.Sprintf("import * as inputs from \"%s/types/input\";", relRoot))
 		imports = append(imports, fmt.Sprintf("import * as outputs from \"%s/types/output\";", relRoot))
+		imports = append(imports, fmt.Sprintf("import * as enums from \"%s/types/enums\";", relRoot))
```
lblackstone, line 996: Should this import be conditional based on whether or not any enums are present? Probably not a big deal.

pkg/codegen/nodejs/gen.go
```diff
@@ -1047,45 +1057,103 @@ func (mod *modContext) getNamespaces() map[string]*namespace {
...
+		mod.genNamespace(w, child, input, enum, level+1)
 		fmt.Fprintf(w, "%s}\n", indent)
 		if i != len(ns.children)-1 {
 			fmt.Fprintf(w, "\n")
 		}
 	}
 }
 
+func (mod *modContext) genEnum(w io.Writer, enum *schema.EnumType, level int) {
+	indent := strings.Repeat("    ", level)
```
lblackstone, line 1122: TIL about `strings.Repeat`

pkg/codegen/nodejs/gen.go
```diff
@@ -1047,45 +1057,103 @@ func (mod *modContext) getNamespaces() map[string]*namespace {
...
+	for _, e := range enum.Elements {
+		if e.Name == "" {
+			e.Name = makeValidIdentifier(fmt.Sprintf("%v", e.Value))
+		}
+		if e.Comment != "" {
+			fmt.Fprintf(w, "%s/** %s */\n", indent, e.Comment)
+		}
+		fmt.Fprintf(w, "%[1]sexport const %[2]s%[3]s: %[3]s = ", indent, e.Name, enumName)
+		if val, ok := e.Value.(string); ok {
+			fmt.Fprintf(w, "\"%v\";\n", val)
```
lblackstone, line 1134: I think you could do `"%q;\n"` instead
```suggestion
			fmt.Fprintf(w, "%q;\n", val)
```

lblackstone (commented):
LGTM overall

pkg/codegen/nodejs/gen.go
komalali: Yeah, I was asking myself the same question. Why do inputs and outputs live in their own namespaces? Naming conflicts?

pkg/codegen/nodejs/gen.go
komalali: ooh TIL

pkg/codegen/nodejs/gen.go
komalali: There are a lot of references to enums in the azure code. Object types and resources have enums as their properties, meaning there's a lot of importing enums and then referencing the types. So I think they do need to be namespaced, otherwise I'm not sure how those imports would work.

komalali: It's not great that the import path is now, for example, `aws.types.enums.lambda.NodeJS10dXRuntime` instead of `aws.lambda.NodeJS10dXRuntime`. 

I'm trying to figure out if there's some way to export as an alias from within the specific module, but I can't quite work out the semantics or whether that's possible. For example, I can `export {lambda} from "../types/enums";` in `lambda/index.ts`, but that would make the enum path `aws.lambda.lambda.NodeJS10dXRuntime`. What I really need is something like `export {* from lambda} from "../types/enums";` but that's not a real thing. Also not clear how this would work for provider-level enums.

I'm wondering if it makes sense to have `types/enums` follow a nested file module structure, so we can `export * from "../types/enums/lambda";` in `lambda/index.ts`, which would allow us to maintain the `aws.lambda.NodeJS10dXRuntime` enum path.

lukehoban: I haven't looked at all the details here - but my sense is that `aws.lambda.NodeJS10dXRuntime` is *much* better than `aws.types.enums.lambda.NodeJS10dXRuntime`.  

> I'm wondering if it makes sense to have types/enums follow a nested file module structure, so we can export * from "../types/enums/lambda"; in lambda/index.ts, which would allow us to maintain the aws.lambda.NodeJS10dXRuntime enum path.

This sounds right to me.  I think this just requires writing separate files for each module inside the `enums` folder.

pkg/codegen/internal/test/testdata/simple-enum-schema/types/enums.ts
```diff
@@ -0,0 +1,25 @@
+// *** WARNING: this file was generated by test. ***
+// *** Do not edit by hand unless you're certain you know what you are doing! ***
+
+import * as pulumi from "@pulumi/pulumi";
+
+export const redContainerColor: ContainerColor = "red";
+export const blueContainerColor: ContainerColor = "blue";
+export const yellowContainerColor: ContainerColor = "yellow";
+
+export type ContainerColor = "red" | "blue" | "yellow";
```
justinvp, line 10: If there's a description for the enum, should it be emitted as a JSDoc comment?

```typescript
/** plant container sizes */
export type ContainerColor = "red" | "blue" | "yellow";
```

pkg/codegen/internal/test/testdata/simple-enum-schema/types/enums.ts
```diff
@@ -0,0 +1,25 @@
...
+export type ContainerSize = 4 | 6 | 8;
+
+/** A burgundy rubber tree. */
+export const BurgundyRubberTreeVariety: RubberTreeVariety = "Burgundy";
+/** A ruby rubber tree. */
+export const RubyRubberTreeVariety: RubberTreeVariety = "Ruby";
+/** A tineke rubber tree. */
+export const TinekeRubberTreeVariety: RubberTreeVariety = "Tineke";
+
+export type RubberTreeVariety = "Burgundy" | "Ruby" | "Tineke";
```
justinvp, line 25: Sorry for not commenting on this in the design doc earlier, but it only occurred to me just now...

I know we've typically emitted these in the past like what's currently being done here, but I was wondering if we should use [`const enums`](https://www.typescriptlang.org/docs/handbook/enums.html#const-enums) instead:

```typescript
export const enum ContainerColor {
    Red = "red",
    Blue = "blue",
    Yellow = "yellow",
}

export const enum ContainerSize {
    FourInch = 4,
    SixInch = 6,
    EightInch = 8,
}

export const enum RubberTreeVariety {
    /** A burgundy rubber tree. */
    Burgundy = "Burgundy",
    /** A ruby rubber tree. */
    Ruby = "Ruby",
    /** A tineke rubber tree. */
    Tineke = "Tineke",
}
```

Unfortunately, there appears to be a downside: the value would have to be specified as `RubberTreeVariety.Burgundy` rather than `"Burgundy"`.

```typescript
export interface RubberTreeArgs {
    readonly type?: RubberTreeVariety;
}

const a: RubberTreeArgs = {
    type: RubberTreeVariety.Burgundy, // Works
};

const b: RubberTreeArgs = {
    type: "Burgundy", // Doesn't work
};
```

To solve this, I was wondering if there was a built-in way in TypeScript to lookup the type of the enum property values, but there doesn't look like there's a way to do that, so we'd have to do something like:

```typescript
export const enum RubberTreeVariety {
    /** A burgundy rubber tree. */
    Burgundy = "Burgundy",
    /** A ruby rubber tree. */
    Ruby = "Ruby",
    /** A tineke rubber tree. */
    Tineke = "Tineke",
}

export type RubberTreeVarietyValues = "Burgundy" | "Ruby" | "Tineke";

export interface RubberTreeArgs {
    readonly type?: RubberTreeVariety | RubberTreeVarietyValues;
}
```

But having to have different named types (`RubberTreeVariety` and `RubberTreeVarietyValues`) is not ideal. So not really loving that.

However, this exploration led me to the following blog post: https://www.designcise.com/web/tutorial/is-it-possible-to-allow-literal-string-values-with-typeScripts-enum-type

I think we should consider doing something along the lines of that post:

```typescript
export const ContainerColor = {
    Red: "red",
    Blue: "blue",
    Yellow: "yellow",
} as const;

export type ContainerColor = (typeof ContainerColor)[keyof typeof ContainerColor]; // "red" | "blue" | "yellow"

export const ContainerSize = {
    FourInch: 4,
    SixInch: 6,
    EightInch: 8,
} as const;

export type ContainerSize = (typeof ContainerSize)[keyof typeof ContainerSize]; // 4 | 6 | 8

export const RubberTreeVariety = {
    /** A burgundy rubber tree. */
    Burgundy: "Burgundy",
    /** A ruby rubber tree. */
    Ruby: "Ruby",
    /** A tineke rubber tree. */
    Tineke: "Tineke",
} as const;

export type RubberTreeVariety = (typeof RubberTreeVariety)[keyof typeof RubberTreeVariety]; // "Burgundy" | "Ruby" | "Tineke"
```

It allows both of these uses:

```typescript
// Both work.
const a: RubberTreeArgs = {
    type: RubberTreeVariety.Burgundy, // Works
};
const b: RubberTreeArgs = {
    type: "Burgundy", // Works
};
```

And doesn't require two different type names.

And provides a cleaner way of grouping the constants together, improving discoverability, etc.

komalali: Interesting find! I agree that it's cleaner / more discoverable. It's not immediately clear to me if this would play nice with the existing enums in pulumi-aws where the constants follow a `BurgundyRubberTreeVariety` pattern and are already in use, so this would be a breaking change. I'll spend some time looking into whether/how we could support both.

komalali: Okay, I've created an issue for this https://github.com/pulumi/pulumi/issues/5494 and definitely would like to see us make this improvement in the future. For now, I'm going to continue with the existing implementation as it matches the enums we already have in the wild (pulumi-aws), and would like to focus on schematizing what we have and moving on to the other languages.

pkg/codegen/internal/test/testdata/simple-enum-schema/types/enums.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export const redContainerColor: ContainerColor = "red";
+export const blueContainerColor: ContainerColor = "blue";
+export const yellowContainerColor: ContainerColor = "yellow";
+
+export type ContainerColor = "red" | "blue" | "yellow";
+
+export const FourInchContainerSize: ContainerSize = 4;
+export const SixInchContainerSize: ContainerSize = 6;
+export const EightInchContainerSize: ContainerSize = 8;
```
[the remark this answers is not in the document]
komalali: I'll add some tests for this. I am using makeValidIdentifier here, so it would actually be:
```typescript
export const _4ContainerSize: ContainerSize = 4;
export const _6ContainerSize: ContainerSize = 6;
export const _8ContainerSize: ContainerSize = 8;
```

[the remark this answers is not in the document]
komalali: Opened an issue to add validation for the identifiers #5502

komalali:
This is almost ready, I'd just like to get https://github.com/pulumi/pulumi/issues/5476 in first.

pkg/codegen/nodejs/gen.go
```diff
@@ -1086,6 +1105,48 @@ func (mod *modContext) genNamespace(w io.Writer, ns *namespace, input bool, leve
...
+	enumName := tokenToName(enum.Token)
+	for _, e := range enum.Elements {
+		if e.Name == "" {
+			e.Name = fmt.Sprintf("%v", e.Value)
+		}
+		e.Name = makeSafeEnumName(e.Name)
+		if e.Comment != "" || e.DeprecationMessage != "" {
+			printComment(w, e.Comment, e.DeprecationMessage, "")
+		}
+		fmt.Fprintf(w, "export const %[1]s%[2]s: %[2]s = ", e.Name, enumName)
```
komalali, line 1122: This naming is not working out with the existing AWS enums. Changing it.

pkg/codegen/internal/test/helpers.go
```diff
@@ -68,6 +68,6 @@ func LoadFiles(dir string, files []string) (map[string][]byte, error) {
 func ValidateFileEquality(t *testing.T, actual, expected map[string][]byte) {
 	for name, file := range expected {
 		assert.Contains(t, actual, name)
-		assert.Equal(t, file, actual[name])
+		assert.Equal(t, string(file), string(actual[name]))
```
pgavlin, line 71: Tiny nit: I think that this will work with byte slices.

pgavlin (approved):

lblackstone (approved):

pkg/codegen/internal/test/helpers.go
komalali: It works but the output isn't particularly helpful when there's an error. This makes the error output human-readable.

pgavlin: Ah, cool! I was under the mistaken impression that `testify` would do that itself ¯\_(ツ)_/¯

pkg/codegen/nodejs/gen.go
justinvp: RE: Introduce nodejs-specific enum name overrides

If we do https://github.com/pulumi/pulumi/issues/5494, instead of name overrides, can we have way that would allow us to keep the old names? Whether it's specified in the schema, or we just have a some overlay files for AWS that have the existing constants?

justinvp (approved):

pkg/codegen/nodejs/gen.go
komalali: @justinvp fair point.

> we just have a some overlay files for AWS that have the existing constants?
I think this might work. We could also add a deprecation message to these constants and redirect folks to the "new" enums. 

It would mean that we would have to update both the schema and the overlay files if there are new enum values added until we can remove the older constants with a major version bump.

I'll give this a go in a separate PR.

PR #5484, opened 2020-09-30
Fix gzip compression

birdypme:
I'm experimenting with a python implementation of a backend for pulumi, and noticed the gzip bodies failed to decompress.
After playing with it and looking at the docs, it is explicitly stated that `gzip.Writer.Close()` must be called to write trailing bytes into the compressed stream (`Flush()` is not enough). Adding an explicit call to `Close()` when we're done successfully fixes the issue - the resulting bytes stream can be read by both golang and python - and doesn't cause any issue with the deferred `Close()` because we are ignoring its error anyway (with `contract.IgnoreClose`)

I've also fixed a case where `gzip.NewReader()` was returning `nil`, and the pulumi client was crashing in the deferred `Close()` call. This was happening only when my python backend was returning garbage, but I still believe the client should return the expected error instead of a null dereference crash when that happens.

Let me know if you need anything more.
I've tried to run the `make travis_pull_request` command as indicated in the contributing guidelines, but I only got a message `TRAVIS PULL REQUEST: moved to GitHub Actions`. If there's another command to run I can try it out and update the contribution guidelines.


pkg/backend/httpstate/client/api.go
```diff
@@ -141,6 +141,11 @@ func pulumiAPICall(ctx context.Context, d diag.Sink, cloudAPI, method, path stri
 		if err := writer.Flush(); err != nil {
 			return "", nil, errors.Wrapf(err, "flushing compressed payload")
 		}
+		// gzip.Writer will not actually write everything unless it is closed.
+		// without this, the compressed bytes do not decompress properly e.g. in python.
+		if err := writer.Close(); err != nil {
+			return "", nil, errors.Wrapf(err, "closing compressed payload")
+		}
```
lukehoban, line 148: I believe it is the case that the Flush above is no longer needed, as the Close will automatically flush?

pkg/backend/httpstate/client/api.go
```diff
@@ -331,7 +336,9 @@ func readBody(resp *http.Response) ([]byte, error) {
 	case "gzip":
 		logging.V(apiRequestDetailLogLevel).Infoln("decompressing gzipped response from service")
 		reader, err := gzip.NewReader(resp.Body)
-		defer contract.IgnoreClose(reader)
+		if reader != nil {
+			defer contract.IgnoreClose(reader)
+		}
```
lukehoban, line 341: Is it the case that NewReader can return a nil error but also a nil reader?  I thought no. In which case - I think the typical way to write this would be to move the defer after the err check.

lukehoban (approved):
Thanks!

pkg/backend/httpstate/client/api.go
birdypme: That's what I assumed initially ([my initial commit](https://github.com/pulumi/pulumi/commit/bcac6cf1d772c2a0bfbba45df821d0a1a7e6c331) moved the defer IgnoreClose call to after the err check), but I didn't find any formal confirmation about this behaviour ([gzip.NewReader documentation](https://golang.org/pkg/compress/gzip/#NewReader)). I then decided to keep it here to reduce the potential side-effect that I don't know about.
If your experience confirms that in case of error the other pointer is always nil, I will move the defer call to after the error check.

pkg/backend/httpstate/client/api.go
birdypme: Yes, after checking the [documentation for gzip.NewWriter](Close closes the Writer by flushing any unwritten data to the underlying io.Writer and writing the GZIP footer), I confirm that Close() calls Flush(). I have implemented the change.

pkg/backend/httpstate/client/api.go
birdypme: @lukehoban can you have a look here? I'd like to know if I you think I should change this part, or if it's ready to go, thanks!

lukehoban:
Kicked off a test run in https://github.com/pulumi/pulumi/pull/5512 to do a full validation.

pkg/backend/httpstate/client/api.go
lukehoban: Looks fine as is - thanks!.

PR #5499, opened 2020-10-02
feat(autoapi): expose EnvVars LocalWorkspaceOption to set in ctor

### LINKED ISSUE #5370: [Automation API] Expose LocalWorkspaceOption `Envs` to allow specifying environment variables on creation
It would be convenient to be able to specify environment variables at workspace creation time (or when calling `NewStackXSource`). Then you can create a stack with a passphrase secrets provider in a single step. Further context discussed here:

https://github.com/pulumi/pulumi/pull/5320#discussion_r486482339

This work should include updating any examples/tests in #5320 that use `os.Setenv` or have to avoid the workspace syntactic sugar methods for this reason.

metral:
Closes #5370

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ CHANGELOG
 
 ## HEAD (Unreleased)
 
+- feat(autoapi): expose EnvVars LocalWorkspaceOption to set in ctor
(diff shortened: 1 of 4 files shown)


EvanBoyle (approved):
🚢

Issue #3255, opened 2019-09-23
Bad error if running pulumi new but stack name already exists


chrsmith:
Run the following command twice. `pulumi new` will use the empty string the second time, leading to an odd error message.

```
$ mkdir repro-name && cd ./repro-name
$ TERM=dumb pulumi new azure-go -f

error: failed to load Pulumi project located at "/Users/chris/wasteland/env828445713/env828445713/env828445713/Pulumi.yaml": project is missing a 'name' attribute
```

PR #5504, opened 2020-10-03
Correctly validate project names during 'pulumi new'

joeduffy:
The logic for validating prompted values in 'new' wasn't quite right,
leading to the possibility of creating Pulumi.yaml files with blank
project names.

This manifests in various ways and I've hit it a number of times
over the past few months because of the way we handle project/stack
name conflicts in 'new' -- which itself is a bit annoying too:

https://github.com/pulumi/pulumi/blob/master/pkg/cmd/pulumi/new.go#L206-L207

Because we substitue a default value of "", and because the prompting
logic assumed default values are always valid, we would skip validation
and therefore accept a blank project name.

This generates an invalid Pulumi.yaml which causes errors elsewhere, such as

    error: failed to load Pulumi project located at ".../Pulumi.yaml":
        project is missing a 'name' attribute

I hit this all the time with our getting started guide because I've
gone through it so many times and have leftover stacks from prior
run-throughs. I wouldn't be surprised if a lot of people hit this.

The solution here validates all values, including the default.

Note also that we failed to validate the value used by 'new --yes'
which meant you could bypass all validation by passing --yes, leading
to similar outcomes.

I've added a couple new tests for these cases. There is a risk we
depend on illegal default values somewhere which will now be rejected,
but that would seem strange, and assuming the tests pass, I would
assume that's not true. Let me know if that's wrong.

Fixes pulumi/pulumi#3255.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@ CHANGELOG
 =========
 
 ## HEAD (Unreleased)
-_(none)_
+
+- Correctly validate project names during 'pulumi new'
+  [#5504](https://github.com/pulumi/pulumi/pull/5504)
 
 ## 2.11.2 (2020-10-01)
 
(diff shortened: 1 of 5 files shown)


justinvp (approved):
LGTM. Thanks for fixing this!

Issue #5223, opened 2020-08-25
pulumi preview with incorrect passphrase prints error message endlessly


mdaniel:
## Expected Behavior

It should print `incorrect passphrase` and `exit 1`

## Current Behavior

```console
$ PULUMI_CONFIG_PASSPHRASE=nope pulumi preview -v 1000 --diff --non-interactive --color never
error: incorrect passphrase
error: incorrect passphrase
error: incorrect passphrase
error: incorrect passphrase
error: incorrect passphrase
^C
```

## Steps to Reproduce

```sh
docker run --rm -it pulumi/pulumi-go:2.9.0 bash -exc '
export PULUMI_CONFIG_PASSPHRASE=password
PATH=/pulumi/bin:$PATH
pulumi login file://$PWD
pulumi new go --name sample --yes
PULUMI_CONFIG_PASSPHRASE=bogus
pulumi preview --diff
'
```

## Context (Environment)

We had a CI job run for over an hour without any indication what it was trying to do, so we wondered if a wrong password in the environment would cause it

clstokes:
This happens with `PULUMI_CONFIG_PASSPHRASE_FILE` too.

PR #5507, opened 2020-10-05
Quit if a static passphrase is invalid

mikhailshilkov:
Fix https://github.com/pulumi/pulumi/issues/5223


stack72 (approved):
thank you so much!! :)

Issue #4885, opened 2020-06-24
[Program Gen] Binary bool and expressions seem to be non-deterministic


EvanBoyle:
We see this test case occasionally fail in CI:

https://github.com/pulumi/pulumi/blob/master/pkg/codegen/go/gen_program_expression_test.go#L73

Where we expect `a && true` we instead get that expression enclosed in parenthesis `(a && true)`

https://travis-ci.com/github/pulumi/pulumi/jobs/353249467#L6908

PR #5508, opened 2020-10-05
Ensure top-level precedence to avoid flaky tests

mikhailshilkov:
Fix https://github.com/pulumi/pulumi/issues/4885

I'm not able to reproduce the original failure. However, the parenthesis is likely added due to messed-up precedence [here](https://github.com/pulumi/pulumi/blob/3be1ea9592b305df8aeb83c3810aee9bb8adeb7a/pkg/codegen/hcl2/model/format/gen.go#L113-L119).

This PR makes sure to set `parentPrecedence` to `0` if `f.Precision()` returns `false` as the second result. Most commonly, the first result is `0`, but not always! 

With old code, if I run `go test -run TestBinaryOpExpression -count=10 -parallel 10` and log `parentPrecedence` if `ok` is `false`, I occasionally get `2`, `6`, and `14`. A `14` would activate the parenthesis adding logic.

### DIFF
--- a/pkg/codegen/go/gen_program_expression_test.go
+++ b/pkg/codegen/go/gen_program_expression_test.go
@@ -45,7 +45,6 @@ func TestLiteralExpression(t *testing.T) {
 }
 
 func TestBinaryOpExpression(t *testing.T) {
-	t.Skip("Skipping test due to https://github.com/pulumi/pulumi/issues/4885")
 	env := environment(map[string]interface{}{
 		"a": model.BoolType,
 		"b": model.BoolType,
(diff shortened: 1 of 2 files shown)


EvanBoyle (approved):
Wow, nice find!

pgavlin:
> This PR makes sure to set parentPrecedence to 0 if f.Precision() returns false as the second result. Most commonly, the first result is 0, but not always!

Yikes! That's so far out of typical behavior for a Go API--great work finding this.

pgavlin (approved):

PR #5514, opened 2020-10-06
[codegen] Add support for config generation in go programgen

viveklak:
- Also adds test scaffolding for azure provider and adds a basic config generation test

### DIFF
--- a/pkg/codegen/go/gen_program.go
+++ b/pkg/codegen/go/gen_program.go
@@ -591,20 +591,21 @@ func (g *generator) genLocalVariable(w io.Writer, v *hcl2.LocalVariable) {
 
 func (g *generator) genConfigVariable(w io.Writer, v *hcl2.ConfigVariable) {
 	if !g.configCreated {
-		g.Fprintf(w, "%sconfig := config.New(ctx, \"\")\n", g.Indent)
+		g.Fprint(w, "cfg := config.New(ctx, \"\")\n")
 		g.configCreated = true
 	}
 
-	getType := "Object"
+	getType := ""
 	switch v.Type() {
-	case model.StringType:
-		getType = ""
+	case model.StringType: // Already default
 	case model.NumberType:
 		getType = "Float"
 	case model.IntType:
 		getType = "Int"
 	case model.BoolType:
 		getType = "Boolean"
+	case model.DynamicType:
+		getType = "Object"
 	}
 
 	getOrRequire := "Get"
@@ -613,7 +614,7 @@ func (g *generator) genConfigVariable(w io.Writer, v *hcl2.ConfigVariable) {
 	}
 
 	if v.DefaultValue == nil {
-		g.Fgenf(w, "%[1]s%[2]s := config.%[3]s%[4]s(\"%[2]s\")\n", g.Indent, v.Name(), getOrRequire, getType)
+		g.Fgenf(w, "%[1]s := cfg.%[2]s%[3]s(\"%[1]s\")\n", v.Name(), getOrRequire, getType)
 	} else {
 		expr, temps := g.lowerExpression(v.DefaultValue, v.DefaultValue.Type(), false)
 		g.genTemps(w, temps)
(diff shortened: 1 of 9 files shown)


pkg/codegen/go/gen_program.go
```diff
@@ -598,19 +658,21 @@ func (g *generator) useLookupInvokeForm(token string) bool {
...
-	if mod == "index" {
-		mod = ""
-	}
 	fn := Title(member)
-	if len(modSplit) >= 2 {
+	if mod == "index" && len(modSplit) >= 2 {
+		mod = ""
 		fn = Title(modSplit[1])
+	} else {
+		// e.g. for type ec2/getVpc:getVpcArgs
```
viveklak, line 666: This whole function is a little hard to reason about and could do with rewriting? The change here is necessary for invokes in the azure-nextgen provider which I tested manually as well.

pkg/codegen/go/gen_program.go
```diff
@@ -584,7 +587,64 @@ func (g *generator) genLocalVariable(w io.Writer, v *hcl2.LocalVariable) {
 		g.Fgenf(w, "%s := %.3v;\n", name, expr)
 
 	}
+}
+
+func (g *generator) genConfigVariable(w io.Writer, v *hcl2.ConfigVariable) {
+	if !g.configCreated {
+		g.Fprintf(w, "%sconfig := config.New(ctx, \"\")\n", g.Indent)
```
EvanBoyle, line 594: I think we need to use a different name here (can't reassign `config`). Also, given that the generator runs `go fmt` at the end, we don't need to worry about indentation (applies elsewhere as well).

pkg/codegen/go/gen_program.go
```diff
@@ -584,7 +587,64 @@ func (g *generator) genLocalVariable(w io.Writer, v *hcl2.LocalVariable) {
...
+			case hcl2.Invoke:
+				g.Fgenf(w, "%s, err := %.3v;\n", v.Name(), expr)
+				g.isErrAssigned = true
+				g.Fgenf(w, "if err != nil {\n")
+				g.Fgenf(w, "return err\n")
+				g.Fgenf(w, "}\n")
+			}
+		default:
+			g.Fgenf(w, "%s := %.3v;\n", v.Name(), expr)
+		}
```
EvanBoyle, line 632: This looks pretty similar to `genLocalVariable`, perhaps we should replace this with a call to that function?

pkg/codegen/go/gen_program.go
EvanBoyle: Yeah, I think this has accrued a lot of incremental complexity. For the time being, perhaps you could add a comment or two to document the case we're accommodating here. Every time I look back on this code it's like archeology to figure out what all of the special cases are for. Some context will be helpful if we do decide to do some refactoring and architectural improvements as a part of q4 debt collection.

pkg/codegen/internal/test/testdata/aws-eks.pp.go
```diff
@@ -71,7 +71,7 @@ func main() {
 			vpcSubnet = append(vpcSubnet, __res)
 		}
 		var rta []*ec2.RouteTableAssociation
-		for key0, _ := range zones.Names {
+		for key0 := range zones.Names {
```
EvanBoyle, line 74: Did you happen to run an auto-formatter or linter on these from your IDE? These changes to the go test files look accidental.

pkg/codegen/go/gen_program.go
```diff
@@ -584,7 +587,64 @@ func (g *generator) genLocalVariable(w io.Writer, v *hcl2.LocalVariable) {
...
 	}
+}
+
+func (g *generator) genConfigVariable(w io.Writer, v *hcl2.ConfigVariable) {
+	if !g.configCreated {
+		g.Fprintf(w, "%sconfig := config.New(ctx, \"\")\n", g.Indent)
+		g.configCreated = true
+	}
+
+	getType := "Object"
```
EvanBoyle, line 598: I don't think this is the right default. I would start with string, and only default to object if the type is `model.Dynamic`. Take a look through the generated aws examples. It looks like every instance of `RequireObject` really just wants strings (or IDs which are a subset of strings):

https://github.com/pulumi-bot/pulumi-aws/compare/ba96ff026..05673f914

EvanBoyle (commented):
A few things to look into, but generally looks like it's in the right direction. 

Have you spot-checked some of the examples across providers to see if the results compile? It would be good to take a half dozen from each of the diffs above for AWS, Azure, and GCP to see if we can identify any obvious defects.

pkg/codegen/go/gen_program.go
viveklak: Yeah I wasn't unsure of this, thanks for your suggestion. I have made the change you suggested and will audit the downstream diffs more closely.

pkg/codegen/internal/test/testdata/aws-eks.pp.go
viveklak: I did in response to a lint failure. Reverted.

pkg/codegen/go/gen_program.go
viveklak: There are some subtle differences. Chose for the moment to allow a bit of duplication instead of refactoring both functions to share code.

pkg/codegen/go/gen_program.go
viveklak: Done

pkg/codegen/go/gen_program.go
viveklak: > I don't think this is the right default. I would start with string, and only default to object if the type is `model.Dynamic`. Take a look through the generated aws examples. It looks like every instance of `RequireObject` really just wants strings (or IDs which are a subset of strings):
> 
> https://github.com/pulumi-bot/pulumi-aws/compare/ba96ff026..05673f914

Hmm seems like checking for dynamic type is not sufficient - a bunch of the ID types are still being emitted as objects. Wonder if there is a way to do an assignability check. Will update with what I find.

viveklak: > > I don't think this is the right default. I would start with string, and only default to object if the type is `model.Dynamic`. Take a look through the generated aws examples. It looks like every instance of `RequireObject` really just wants strings (or IDs which are a subset of strings):
> > https://github.com/pulumi-bot/pulumi-aws/compare/ba96ff026..05673f914
> 
> Hmm seems like checking for dynamic type is not sufficient - a bunch of the ID types are still being emitted as objects. Wonder if there is a way to do an assignability check. Will update with what I find.

Actually seems like these are showing up as object in the other language generators too. We should probably just cut a separate ticket to track and fix these.

EvanBoyle (approved):

PR #5519, opened 2020-10-06
Execute self-contained .NET binary

### LINKED ISSUE #5334: dotnet binary option for self-contained executables
I noticed the binary option only works with a `dll`, despite the build generating an `exe`:
```yml
name: aws
runtime:
  name: dotnet
  options:
    binary: Aws.exe
description: AWS infrastructure
config: ..
```
Fails with the following error:
```sh
user@b8b9b2241208:/app/Aws$ p up
Previewing update (pharos/alpha)

View Live: https://app.pulumi.com/pharos/aws/alpha/previews/e85b6b48-b096-4857-9700-30d8a29da561

     Type                 Name       Plan     Info
     pulumi:pulumi:Stack  aws-alpha           1 error; 5 messages

Diagnostics:
  pulumi:pulumi:Stack (aws-alpha):
    Error:
      An assembly specified in the application dependencies manifest (Aws.deps.json) has already been found but with a different file extension:
        package: 'Aws', version: '0.0.0.0'
        path: 'Aws.dll'
        previously found assembly: '/app/Aws/Aws.exe'

    error: an unhandled error occurred: Program exited with non-zero exit code: 140
```
I’ll probably want to switch to a trimmed single file `exe` when dotnet 5.0 lands! 😉 

Another problem is that pulumi plugins are not automatically acquired, so given the following project dependencies:
```xml
<ItemGroup>
  <PackageReference Include="Pulumi" Version="2.10.0" />
  <PackageReference Include="Pulumi.Aws" Version="3.2.1" />
  <PackageReference Include="Pulumi.Kubernetes" Version="2.6.0" />
  <PackageReference Include="Pulumi.Tls" Version="2.3.0" />
</ItemGroup>
```
I have to run the following inside the container for the binary to work:
```sh
pulumi plugin install resource aws v3.2.1
pulumi plugin install resource kubernetes v2.6.0
pulumi plugin install resource tls v2.3.0
```

mikhailshilkov:
Resolves #5334 (other than a binary CLI option, which will move to a new issue)

When the `binary` option is specified, this PR runs it as `dotnet foo.dll` for dll files and directly otherwise.

This change looks legit to me but is slightly scary in case I'm missing some other non-dll scenario, so reviews are appreciated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ CHANGELOG
 - Add internal scaffolding for using cross-language components from .NET.
   [#5485](https://github.com/pulumi/pulumi/pull/5485)
 
+- Support self-contained executables as binary option for .NET programs.
+  [#5519](https://github.com/pulumi/pulumi/pull/5519)
+
 ## 2.11.2 (2020-10-01)
 
 - feat(autoapi): expose EnvVars LocalWorkspaceOption to set in ctor
--- a/sdk/dotnet/cmd/pulumi-language-dotnet/main.go
+++ b/sdk/dotnet/cmd/pulumi-language-dotnet/main.go
@@ -477,11 +477,18 @@ func (host *dotnetLanguageHost) Run(ctx context.Context, req *pulumirpc.RunReque
 		return nil, err
 	}
 
+	executable := host.exec
 	args := []string{}
 
-	if host.binary != "" {
+	switch {
+	case host.binary != "" && strings.HasSuffix(host.binary, ".dll"):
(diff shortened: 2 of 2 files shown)


EvanBoyle (approved):
Looks good to me. The additional functionality you've added here is basically what we do for Go today: https://github.com/pulumi/pulumi/blob/9ef5c72f7284c646e88ee9fc75d142c42987d6d6/sdk/go/pulumi-language-go/main.go#L45-L75

justinvp (approved):
Looks reasonable to me.

pgavlin (approved):

PR #5520, opened 2020-10-07
Fix config params and default handling for .NET programgen

viveklak:
Fixes https://github.com/pulumi/pulumi-azure-nextgen/issues/42

### DIFF
--- a/pkg/codegen/dotnet/gen_program.go
+++ b/pkg/codegen/dotnet/gen_program.go
@@ -515,11 +515,16 @@ func (g *generator) genConfigVariable(w io.Writer, v *hcl2.ConfigVariable) {
 	}
 
 	if v.DefaultValue != nil {
-		g.Fgenf(w, "%[1]svar %[2]s = Output.Create(config.%[3]s%[4]s(\"%[2]s\"))", g.Indent, v.Name(), getOrRequire, getType)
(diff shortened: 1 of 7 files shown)


pkg/codegen/dotnet/gen_program.go
```diff
@@ -514,9 +514,12 @@ func (g *generator) genConfigVariable(w io.Writer, v *hcl2.ConfigVariable) {
 		getOrRequire = "Require"
 	}
 
-	g.Fgenf(w, "%[1]svar %[2]s = config.%[3]s%[4]s(\"%[2]s\")", g.Indent, v.Name(), getOrRequire, getType)
 	if v.DefaultValue != nil {
+		g.Fgenf(w, "%[1]svar %[2]s = Output.Create(config.%[3]s%[4]s(\"%[2]s\"))", g.Indent, v.Name(), getOrRequire, getType)
```
mikhailshilkov, line 518: I think the right part of the `??` expression could have a type of than `Output<T>` too? Could you extend the test program so that both possibilities are tested (e.g. have a config with a fixed default value)?

viveklak: Made the conversion only apply if the default value is a promise type. Also updated the test case to be realistic.

mikhailshilkov (approved):
LGTM

PR #5523, opened 2020-10-07
Reorder resource docs templates to make deprecated more prominent

stack72:
Fixes: https://github.com/pulumi/docs/issues/4340

The deprecated message sits below the examples so it is not clear
to the user that the resource / datasource is deprecated


cnunciato (approved):
LGTM. cc @praneetloke in case this might cause any issues.

praneetloke:
Thank you for the heads-up, @cnunciato! This looks good to me.

@stack72 Is the error in the [PR build](https://github.com/pulumi/pulumi/pull/5523/checks?check_run_id=1221821031) something to be concerned about? Looks like it failed due to not being able to install a plugin for a k8s version. That workflow generates the resource docs for AWS and K8s to preview docs related changes.

stack72:
@praneetloke this was actually due to the fact that a K8s deployment was underway at the same time

PR #5525, opened 2020-10-08
[codegen/nodejs] - Update enum generation

### LINKED ISSUE #5494: [codegen/nodejs] - Improve enums in typescript SDKs
I was wondering if we should use [`const enums`](https://www.typescriptlang.org/docs/handbook/enums.html#const-enums) instead:

```typescript
export const enum ContainerColor {
    Red = "red",
    Blue = "blue",
    Yellow = "yellow",
}

export const enum ContainerSize {
    FourInch = 4,
    SixInch = 6,
    EightInch = 8,
}

export const enum RubberTreeVariety {
    /** A burgundy rubber tree. */
    Burgundy = "Burgundy",
    /** A ruby rubber tree. */
    Ruby = "Ruby",
    /** A tineke rubber tree. */
    Tineke = "Tineke",
}
```

Unfortunately, there appears to be a downside: the value would have to be specified as `RubberTreeVariety.Burgundy` rather than `"Burgundy"`.

```typescript
export interface RubberTreeArgs {
    readonly type?: RubberTreeVariety;
}

const a: RubberTreeArgs = {
    type: RubberTreeVariety.Burgundy, // Works
};

const b: RubberTreeArgs = {
    type: "Burgundy", // Doesn't work
};
```

To solve this, I was wondering if there was a built-in way in TypeScript to lookup the type of the enum property values, but there doesn't look like there's a way to do that, so we'd have to do something like:

```typescript
export const enum RubberTreeVariety {
    /** A burgundy rubber tree. */
    Burgundy = "Burgundy",
    /** A ruby rubber tree. */
    Ruby = "Ruby",
    /** A tineke rubber tree. */
    Tineke = "Tineke",
}

export type RubberTreeVarietyValues = "Burgundy" | "Ruby" | "Tineke";

export interface RubberTreeArgs {
    readonly type?: RubberTreeVariety | RubberTreeVarietyValues;
}
```

But having to have different named types (`RubberTreeVariety` and `RubberTreeVarietyValues`) is not ideal. So not really loving that.

However, this exploration led me to the following blog post: https://www.designcise.com/web/tutorial/is-it-possible-to-allow-literal-string-values-with-typeScripts-enum-type

I think we should consider doing something along the lines of that post:

```typescript
export const ContainerColor = {
    Red: "red",
    Blue: "blue",
    Yellow: "yellow",
} as const;

export type ContainerColor = (typeof ContainerColor)[keyof typeof ContainerColor]; // "red" | "blue" | "yellow"

export const ContainerSize = {
    FourInch: 4,
    SixInch: 6,
    EightInch: 8,
} as const;

export type ContainerSize = (typeof ContainerSize)[keyof typeof ContainerSize]; // 4 | 6 | 8

export const RubberTreeVariety = {
    /** A burgundy rubber tree. */
    Burgundy: "Burgundy",
    /** A ruby rubber tree. */
    Ruby: "Ruby",
    /** A tineke rubber tree. */
    Tineke: "Tineke",
} as const;

export type RubberTreeVariety = (typeof RubberTreeVariety)[keyof typeof RubberTreeVariety]; // "Burgundy" | "Ruby" | "Tineke"
```

It allows both of these uses:

```typescript
// Both work.
const a: RubberTreeArgs = {
    type: RubberTreeVariety.Burgundy, // Works
};
const b: RubberTreeArgs = {
    type: "Burgundy", // Works
};
```

And doesn't require two different type names.

And provides a cleaner way of grouping the constants together, improving discoverability, etc.

_Originally posted by @justinvp in https://github.com/pulumi/pulumi/pull/5472#discussion_r497844694_

komalali:
I went with @justinvp's suggestion of updating how our TS enums are emitted (more details in #5494) instead of adding nodejs-specific name overrides. This allows us to keep standardized naming for the newly generated enums, while also allowing us to support the existing constants as extra overlays. We also get the added benefit of a cleaner/more idiomatic enum interface while still allowing us to use string literals. Thanks for the suggestion @justinvp!

Resolves #5494 

WIP pulumi-aws PR: https://github.com/pulumi/pulumi-aws/pull/1151

### DIFF
--- a/pkg/codegen/internal/test/testdata/simple-enum-schema/types/enums/index.ts
+++ b/pkg/codegen/internal/test/testdata/simple-enum-schema/types/enums/index.ts
@@ -5,23 +5,27 @@
 import * as tree from "./tree";
 export {tree};
 
-export const RedContainerColor: ContainerColor = "red";
-export const BlueContainerColor: ContainerColor = "blue";
-export const YellowContainerColor: ContainerColor = "yellow";
+export const ContainerColor = {
+    Red: "red",
+    Blue: "blue",
+    Yellow: "yellow",
+} as const;
 
 /**
  * plant container colors
  */
-export type ContainerColor = "red" | "blue" | "yellow";
+export type ContainerColor = (typeof ContainerColor)[keyof typeof ContainerColor];
 
-export const FourInchContainerSize: ContainerSize = 4;
-export const SixInchContainerSize: ContainerSize = 6;
-/**
- * @deprecated Eight inch pots are no longer supported.
- */
-export const EightInchContainerSize: ContainerSize = 8;
+export const ContainerSize = {
+    FourInch: 4,
+    SixInch: 6,
+    /**
+     * @deprecated Eight inch pots are no longer supported.
+     */
+    EightInch: 8,
+} as const;
 
 /**
  * plant container sizes
  */
-export type ContainerSize = 4 | 6 | 8;
+export type ContainerSize = (typeof ContainerSize)[keyof typeof ContainerSize];
--- a/pkg/codegen/internal/test/testdata/simple-enum-schema/types/enums/tree/v1/index.ts
+++ b/pkg/codegen/internal/test/testdata/simple-enum-schema/types/enums/tree/v1/index.ts
@@ -2,20 +2,22 @@
 // *** Do not edit by hand unless you're certain you know what you are doing! ***
 
 
-/**
- * A burgundy rubber tree.
- */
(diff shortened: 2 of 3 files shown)


komalali:
Hmm, not sure why the tests in pulumi-aws are [failing](https://github.com/pulumi/pulumi-aws/runs/1224102569#step:20:245). It looks like it's not picking up the new changes.

lblackstone (approved):
Nice! 🎉

justinvp (approved):
Looks great! Thanks!

PR #5551, opened 2020-10-10
[codegen/dotnet] - Enums

### LINKED ISSUE #5434: [codegen/dotnet] Add C# support for enums.

komalali:
### TODO
- [x] Write example test
- [x] Initial design review
- [x] Implement codegen
- [x] Final review

Resolves #5434


pkg/codegen/internal/test/testdata/simple-enum-schema/dotnet/Enums.cs
```diff
@@ -0,0 +1,38 @@
+// *** WARNING: this file was generated by test. ***
+// *** Do not edit by hand unless you're certain you know what you are doing! ***
+
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Threading.Tasks;
+using Pulumi.Serialization;
+
+namespace Pulumi.PlantProvider.Enums
```
mikhailshilkov, line 10: Do you know of any issues if put enums to the main namespace? (`Pulumi.PlantProvider`) If a resource uses enums for inputs, every user's program will have to import an extra namespace, which would be great to avoid.

pkg/codegen/internal/test/testdata/simple-enum-schema/dotnet/Enums.cs
```diff
@@ -0,0 +1,38 @@
...
+        Yellow
+    }
+
+    /// <summary>
+    /// plant container sizes
+    /// </summary>    
+    public enum ContainerSize
+    {
+        [EnumMember(Value = 4)]
+        FourInch,
```
mikhailshilkov, line 31: C# enums are integers by default, so we should assign int values to options if our underlying type is int.
```suggestion
        FourInch = 4,
```

This way users can cast between enum and int without extra ceremony.

mikhailshilkov (commented):
Looks great! Two suggestions below

pkg/codegen/internal/test/testdata/simple-enum-schema/dotnet/Enums.cs
komalali: Good point. I don't foresee any issues adding it to the main namespace, will proceed with that 👍

justinvp (commented):
Added a comment about compatibility.

Also, are you planning to add support for "relaxed" enums as mentioned in the design doc?

We'll need to implement support in `sdk/dotnet` for serializing/deserializing these. See https://github.com/pulumi/pulumi/tree/master/sdk/dotnet/Pulumi/Serialization

komalali:
@justinvp Since we redesigned the schema representation of enums, relaxed enums are now just the union of the enum type and the underlying element type (there's an example in [ContainerArgs](https://github.com/pulumi/pulumi/pull/5551/files#diff-57208866363ea5bbcab7be4f1091a1d482bddd0ecf71ef52cc397648471a4659R16). 

This should mean that we only have to implement the enum type itself, and don't need two separate implementations for the strict vs relaxed. This works pretty well for typescript and python, but I'm not familiar enough with C# to know if there are any hurdles presented here. Do you foresee any?

pkg/codegen/internal/test/testdata/simple-enum-schema/dotnet/Enums.cs
```diff
@@ -0,0 +1,31 @@
...
+    /// plant container colors
+    /// </summary>
+    public enum ContainerColor
+    {
+        [EnumMember(Value = "red")]
+        Red,
+        [EnumMember(Value = "blue")]
+        Blue,
+        [EnumMember(Value = "yellow")]
+        Yellow,
```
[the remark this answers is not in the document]
komalali: I'm going to dig into this more deeply, but my naive understanding is that using C# enums gets us the added goodness of Intellisense that we don't get with this kind of struct. Thanks for raising this though, I'll spend some time looking through the options here.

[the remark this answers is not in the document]
justinvp: > my naive understanding is that using C# enums gets us the added goodness of Intellisense that we don't get with this kind of struct

Would be curious to know the differences. From my initial check (VS Code, would be curious to know the difference on full VS on Windows).

Enum:

Screen Shot 2020-10-13 at 3 32 00 PM

Struct:

Screen Shot 2020-10-13 at 3 33 11 PM

justinvp:
> relaxed enums are now just the union of the enum type and the underlying element type (there's an example in [ContainerArgs](https://github.com/pulumi/pulumi/pull/5551/files#diff-57208866363ea5bbcab7be4f1091a1d482bddd0ecf71ef52cc397648471a4659R16).

Ah, missed that. This should work.

pkg/codegen/internal/test/testdata/simple-enum-schema/dotnet/Enums.cs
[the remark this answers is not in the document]
komalali: Oh neat! Yeah, will check it out in Visual Studio and also [Rider](https://www.jetbrains.com/rider/).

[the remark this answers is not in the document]
mikhailshilkov: @justinvp The background here is that we designed C# representation around Azure API. Two facts that make it a non-issue there:
- Adding enum values is [considered a breaking change in the API](https://github.com/Azure/azure-rest-api-specs/blob/master/documentation/Breaking%20changes%20guidelines.md#allowed-values-for-an-enum-have-changed)
- Azure .NET SDK uses enums ([example](https://github.com/Azure/azure-sdk-for-net/blob/78cb37ff1ba9fb628ce8f053eaf6cec5320008a2/sdk/search/Microsoft.Azure.Search.Service/src/Generated/Models/StemmerTokenFilterLanguage.cs#L22))

However, your point makes sense in the global scope of different providers. I agree that we need to formalize this. Either add a number as hashcode indeed, or move to the struct-based approach (here is [another example](https://github.com/Azure/azure-sdk-for-net/blob/f775e1f8b8d33904015800c4d0894115a6d22515/sdk/storage/Azure.ResourceManager.Storage/src/Generated/Models/Kind.cs#L14) from the Azure SDK).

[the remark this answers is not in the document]
komalali: @mikhailshilkov In fact, this is still an issue even for the azure provider/API. Adding enum values is only considered a breaking change in the API if they are "strict" enums, not so for "relaxed" enums. However, since we've decided to represent "relaxed" enums as just the union of the underlying and enum type, we are representing all enums using this same representation.

I'll move forward with the struct-based approach once I confirm IDE support. This would apply to enums of both string and number types, while int and bool can continue to be emitted as enums since they are naturally integers.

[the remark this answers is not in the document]
mikhailshilkov: Yeah, makes sense. If we use the structs, I wonder if we should simplify the relaxed enums from `Union<ContainerColor, string>` to a variant of `ContainerColor` with a public constructor and/or an implicit conversion from `string`.

pkg/codegen/internal/test/testdata/simple-enum-schema/dotnet/Enums.cs
```diff
@@ -0,0 +1,47 @@
+// *** WARNING: this file was generated by test. ***
+// *** Do not edit by hand unless you're certain you know what you are doing! ***
+
+using System.Runtime.Serialization;
```
justinvp, line 4: Nit: Doesn't look like this namespace is needed in this file.

pkg/codegen/internal/test/testdata/simple-enum-schema/dotnet/Enums.cs
```diff
@@ -0,0 +1,47 @@
...
+
+namespace Pulumi.PlantProvider
+{
+    /// <summary>
+    /// plant container colors
+    /// </summary>
+    public readonly struct ContainerColor : IEquatable<ContainerColor>
```
justinvp, line 11: Have you had a chance to think much about how the serialization/deserialization is going to work?

We likely need to add a marker attribute [in the SDK](https://github.com/pulumi/pulumi/blob/master/sdk/dotnet/Pulumi/Serialization/Attributes.cs), something like:

```c#
namespace Pulumi
{
    [AttributeUsage(AttributeTargets.Class)]
    public sealed class EnumTypeAttribute : Attribute
    {
    }
}
```

That we can then annotate on these like:

```c#
[EnumType]
public readonly struct ContainerColor : IEquatable<ContainerColor>
{
    ...
}
```

So we can detect these during serialization/deserialization and do the proper thing to convert to/from the underlying string.

We'd only have to annotate these structs... the ones that are actual enums (e.g. `public enum ContainerSize`), we can already detect via [`Type.IsEnum`](https://docs.microsoft.com/en-us/dotnet/api/system.type.isenum?view=netcore-3.1).

Let me know if you need any help here, I can throw together a rough draft of what the serialization/deserialization looks like for .NET.

pkg/codegen/internal/test/testdata/simple-enum-schema/dotnet/Enums.cs
```diff
@@ -0,0 +1,47 @@
...
+        }
+
+        public static ContainerColor Red { get; } = new ContainerColor("red");
+        public static ContainerColor Blue { get; } = new ContainerColor("blue");
+        public static ContainerColor Yellow { get; } = new ContainerColor("yellow");
+
+        public static bool operator ==(ContainerColor left, ContainerColor right) => left.Equals(right);
+        public static bool operator !=(ContainerColor left, ContainerColor right) => !left.Equals(right);
+
+        [EditorBrowsable(EditorBrowsableState.Never)]
```
justinvp, line 27: Note: `EditorBrowsableAttribute` and `EditorBrowsableState` are in the `System.ComponentModel` namespace, so we'll need to emit a `using System.ComponentModel;` at the top of the file (or, alternatively, fully qualify in the emitted code).

pkg/codegen/internal/test/testdata/simple-enum-schema/dotnet/Enums.cs
```diff
@@ -0,0 +1,47 @@
...
+        public override bool Equals(object obj) => obj is ContainerColor other && Equals(other);
+        public bool Equals(ContainerColor other) => string.Equals(_value, other._value, StringComparison.Ordinal);
```
justinvp, line 29: `StringComparison` is in the `System` namespace, so you'll need a `using System;` at the top of the file.

pkg/codegen/internal/test/testdata/simple-enum-schema/dotnet/Enums.cs
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    /// <summary>
+    /// plant container sizes
+    /// </summary>
+    public enum ContainerSize
+    {
+        FourInch = 4,
+        SixInch = 6,
+        [Obsolete(@"Eight inch pots are no longer supported.")]
```
justinvp, line 44: `ObsoleteAttribute` is also in the `System` namespace and needs a `using System;` at the top of the file.

pkg/codegen/internal/test/testdata/simple-enum-schema/dotnet/Enums.cs
komalali: Yeah, I started working on that yesterday. And yeah detecting the actual enums is pretty straightforward, thanks for the tip on the structs! I'm gonna work through the SDK work first before coming back to this.

pkg/codegen/internal/test/testdata/simple-enum-schema/dotnet/Enums.cs
komalali: Heh, yeah I realized I had some imports missing once I started working through the SDK changes.

komalali:
Note: pulumi-aws codegen tests are failing because it needs the new SDK changes to support the EnumType attribute.

pkg/codegen/internal/test/testdata/simple-enum-schema/dotnet/Enums.cs
```diff
@@ -0,0 +1,76 @@
...
+            _value = value ?? throw new ArgumentNullException(nameof(value));
+        }
+
+        public static ContainerColor Red { get; } = new ContainerColor("red");
+        public static ContainerColor Blue { get; } = new ContainerColor("blue");
+        public static ContainerColor Yellow { get; } = new ContainerColor("yellow");
+
+        public static bool operator ==(ContainerColor left, ContainerColor right) => left.Equals(right);
+        public static bool operator !=(ContainerColor left, ContainerColor right) => !left.Equals(right);
+
```
komalali, line 55: Just realized I'm missing the explicit conversion operator. Fixing.

pkg/codegen/internal/test/testdata/simple-enum-schema/dotnet/Enums.cs
```diff
@@ -0,0 +1,76 @@
...
+        public static bool operator ==(ContainerBrightness left, ContainerBrightness right) => left.Equals(right);
+        public static bool operator !=(ContainerBrightness left, ContainerBrightness right) => !left.Equals(right);
+
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public override bool Equals(object? obj) => obj is ContainerBrightness other && Equals(other);
+        public bool Equals(ContainerBrightness other) => _value == other._value;
+
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public override int GetHashCode() => _value?.GetHashCode();
```
justinvp, line 31: For `double`-based enums, the `?` isn't needed here (it is for `string`).

```suggestion
        public override int GetHashCode() => _value.GetHashCode();
```

PR #5553, opened 2020-10-10
[codegen/python] - Enums

### LINKED ISSUE #5435: [codegen/python] Add python support for enums.

komalali:
### TODO
 - [x] Write example test
 - [x] Example design review
 - [x] Implement codegen
 - [x] Final review

Resolves #5435


pkg/codegen/internal/test/testdata/simple-enum-schema/python/pulumi_plant_provider/_inputs.py
```diff
@@ -0,0 +1,53 @@
+# coding=utf-8
+# *** WARNING: this file was generated by test. ***
+# *** Do not edit by hand unless you're certain you know what you are doing! ***
+
+import warnings
+import pulumi
+import pulumi.runtime
+from typing import Any, Mapping, Optional, Sequence, Union
+from . import _utilities, _tables
+from . import _enums as enums
```
justinvp, line 10: Despite use of `__all__` in this module, should we keep this named `_enums` (with the underscore) as a deterrent to keep it out of intellisense?

```suggestion
from . import _enums
```

pkg/codegen/internal/test/testdata/simple-enum-schema/python/pulumi_plant_provider/_enums.py
```diff
@@ -0,0 +1,25 @@
+# coding=utf-8
+# *** WARNING: this file was generated by test. ***
+# *** Do not edit by hand unless you're certain you know what you are doing! ***
+
+from enum import Enum
+
+__all__ = ["ContainerColor", "ContainerSize"]
```
justinvp, line 7: Should this match the similar codegen of `__all__` for `_inputs.py` (with each on its own line?)

pkg/codegen/internal/test/testdata/simple-enum-schema/python/pulumi_plant_provider/_inputs.py
```diff
@@ -16,31 +16,31 @@
 @pulumi.input_type
 class ContainerArgs:
     def __init__(__self__, *,
-                 size: pulumi.Input[enums.ContainerSize],
-                 color: Optional[pulumi.Input[Union[enums.ContainerColor, str]]] = None,
+                 size: pulumi.Input[_enums._ContainerSize],
+                 color: Optional[pulumi.Input[Union[_enums._ContainerColor, str]]] = None,
                  material: Optional[pulumi.Input[str]] = None):
-        pulumi.set(__self__, "size", size)
+        pulumi.set(__self__, "size", size.value if isinstance(size, Enum) else size)
```
komalali, line 22: Going to move this typecheck into `pulumi.set()` itself.

justinvp: I suspect this isn't going to work in all cases. This is an `Input[T]`, so it _may_ be `T` (where this would work), but it could also be `Awaitable[T]` or `Output[T]`. I think we're going to have to just pass the value as-is and add this logic inside [`rpc.serialize_property`](https://github.com/pulumi/pulumi/blob/eaac9280f5484f94213dbf54c9fd6f009d87db12/sdk/python/lib/pulumi/runtime/rpc.py#L88).

pkg/codegen/internal/test/testdata/simple-enum-schema/python/pulumi_plant_provider/tree/v1/rubber_tree.py
```diff
@@ -18,7 +19,7 @@ def __init__(__self__,
                  resource_name: str,
                  opts: Optional[pulumi.ResourceOptions] = None,
                  container: Optional[pulumi.Input[pulumi.InputType['_root_inputs.ContainerArgs']]] = None,
-                 type: Optional[pulumi.Input[enums.RubberTreeVariety]] = None,
+                 type: Optional[pulumi.Input[_enums._RubberTreeVariety]] = None,
```
justinvp, line 22: One thing I've been wondering: The test schema only has `inputProperties` for the `RubberTree` resource. What do output properties look like for enums?

 - If it's a strict enum, will it be typed as the enum type? (And therefore needs to be converted to the enum type during deserialization?)

- If it's a relaxed enum (e.g. `Union[EnumClass, str]`), rather than the output being typed as `Union[EnumClass, str]`, should it be just `str` to make it easier for consumers (i.e. it will always be a string)?

pkg/codegen/python/gen.go
```diff
@@ -1190,6 +1258,89 @@ func (mod *modContext) genFunction(fun *schema.Function) (string, error) {
...
+
+			fmt.Fprintf(w, "%s%s = ", indent, e.Name)
+			if val, ok := e.Value.(string); ok {
+				fmt.Fprintf(w, "%q\n", val)
+			} else {
+				fmt.Fprintf(w, "%v\n", e.Value)
+			}
+		}
+	default:
+		return fmt.Errorf("enums of type %s are not yet implemented for this language", enum.ElementType.String())
```
justinvp, line 1338: Nit:

```suggestion
		return errors.Errorf("enums of type %s are not yet implemented for this language", enum.ElementType.String())
```

justinvp (approved):
LGTM

Issue #2321, opened 2018-12-23
Accept `-f` as a shorthand for `--skip-preview`


lukehoban:
In the dev inner loop, it is very common to want to skip previews on every `pulumi up` operation.  Currently, this requires `--skip-preview`, which is cumbersome to type.  

It would be nicer to be able to have this common case be just `pulumi up -f`.

We do have `-y` flags for these commands, but that still spends time doing the preview, which is frequently not desired during active development.

None of the operations which accept `--skip-preview` yet have a `-f`, and this feels like the mosts natural interpretation of a "force" flag for these.  There's a question of whether `--skip-preview` should be aliased to `--force` as well, but I don't feel strongly about that.

Operations which should adopt this:
- `up`
- `refresh`
- `destroy`

clstokes:
`-f/--force` is nice for consistency's sake. 

What if we removed `-y/--skip-preview` in favor of `-f/--force`?

mikhailshilkov:
I actually like `skip-preview` as the full name. `force` is usually for accepting overwrites (like we use it in `new` or in `stack rm`), so it *could* be close to `yes` but doesn't sound like skipping the preview.

For the same reason, I wouldn't use `f` as the short name for `skip-preview`, as it might be confusing. `s` is already used for `stack`.

Maybe we could use `S`? Otherwise, `x` (execute), `g` (go), `k` (sKip), `o` (omit). None of them looks great though.

Issue #4569, opened 2020-05-05
`--yes` shouldn't be required for commands that won't have any `yes/no` prompts


lukehoban:
Using `--skip-preview` is an alternative to using `--yes`.  It should be allowed to pass that without needing to also pass `--yes` as there will be no `yes/no` prompt.

```
➜  foo pulumi up --skip-preview --non-interactive
error: --yes must be passed in to proceed when running in non-interactive mode
```

joeduffy:
Just hit this myself -- agree this seems unnecessary!

PR #5556, opened 2020-10-12
Accept `-f` as a shorthand for `--skip-preview` on `up`, `refresh` and `destroy` operations

stack72:
Fixes: #2321


jaxxstorm (approved):

pkg/cmd/pulumi/destroy.go
```diff
@@ -193,8 +193,8 @@ func newDestroyCmd() *cobra.Command {
 	cmd.PersistentFlags().BoolVar(
 		&showSames, "show-sames", false,
 		"Show resources that don't need to be updated because they haven't changed, alongside those that do")
-	cmd.PersistentFlags().BoolVar(
-		&skipPreview, "skip-preview", false,
+	cmd.PersistentFlags().BoolVarP(
+		&skipPreview, "skip-preview", "f", false,
```
lukehoban, line 197: Just to make sure - I don't have to also pass `--yes` do I?

(found the issue - I guess I do but only if I include `--non-interactive` - which is weird? -- https://github.com/pulumi/pulumi/issues/4569).

lukehoban (approved):

pkg/cmd/pulumi/destroy.go
stack72: @lukehoban so skip-preview and yes have 2 separate purposes here - what I can do is to have a new parameter called force (-f shorthand) that means we don't even care what skip-preview and yes are doing here

What do you think?

lukehoban: If I'm right on what #4569 is about - then I think we just should also go fix that issue - though it's mostly orthogonal to this.  I don't think we need to introduce another flag - I believe `--skip-preview`/`-f` is intended to skip the preview and not prompt the user (there is nothing to prompt the user about).  So #4569 is just a bug.

justinvp (approved):

PR #5558, opened 2020-10-12
Support remote components in Go

justinvp:
Fixes https://github.com/pulumi/pulumi/issues/5355


sdk/go/pulumi/types.go
```diff
@@ -435,74 +450,6 @@ func AllWithContext(ctx context.Context, inputs ...interface{}) ArrayOutput {
...
-	deps := make([]Resource, 0, len(depSet))
-	for d := range depSet {
-		deps = append(deps, d)
-	}
-	return deps
-}
-
-var resourceType = reflect.TypeOf((*Resource)(nil)).Elem()
-
-func gatherDependencySet(v reflect.Value, deps map[Resource]struct{}) {
```
justinvp, line 459: Still need to double check we're not missing any deps due to removing this code.

sdk/go/pulumi/context.go
```diff
@@ -505,7 +520,13 @@ func (ctx *Context) RegisterResource(
 func (ctx *Context) RegisterComponentResource(
 	t, name string, resource ComponentResource, opts ...ResourceOption) error {
 
-	return ctx.RegisterResource(t, name, nil, resource, opts...)
+	return ctx.RegisterResource(t, name, nil /*props*/, resource, opts...)
+}
+
+func (ctx *Context) RegisterRemoteComponentResource(
```
justinvp, line 526: @pgavlin, what do you think of this new function? Or did you envision exposing this some other way?

lblackstone (approved):
LGTM overall, but should get @pgavlin's feedback prior to merging

justinvp:
The one remaining thing I do want to get back to here is adding some more tests related to the changes to `Output`.

sdk/go/pulumi/context.go
pgavlin: I think that what you have here is fine--this should really only be called from generated code.

sdk/go/pulumi/types.go
pgavlin: Yeah, at a glance it seems like this would lose nested dependencies in an `Any`... why was it removed?

justinvp: I initially removed this to avoid gathering dependencies up-front in favor of forcing all dependencies to be awaited. I was trying to make it so that when creating an `Output`, you'd only pass a single up-front Resource as a dependency, and the only way to get the final set of all dependencies would be to await it.

`gatherDependencySet` was being called to gather up-front dependencies and it looked like the implementation of `awaitInputs` would have picked up most of the nested dependencies anyway, so that's why I initially removed it. However, `gatherDependencySet` does at least one thing that `awaitInputs` doesn't: if a nested value is a `Resource` the resource value is added to the set of dependencies.

In any case, I decided to add back `gatherDependencySet` to avoid missing the resource case. Now, you can continue to pass an array of up-front dependencies you know about when creating an `Output` (as is currently the case), and when you await it, you'll get a merged set of all dependencies (up-front + any other dependencies that came back as part of resolving the output).

justinvp:
@pgavlin, this is ready for a more detailed review. I added back `gatherDependencySet` and added some tests.

pgavlin (approved):

PR #5562, opened 2020-10-13
Pass the preview flag to Update

mikhailshilkov:
I believe the `Update` invocation is missing a value for `Preview`


pgavlin (approved):

pgavlin:
Good find.

This does reinforce an irritating hole in our test coverage: while we have great coverage of the high-level plugin interfaces, we have worse coverage of the plugin -> gRPC bridges. We should think about how we can improve this coverage (e.g. by adding support for previews to the dynamic provider, perhaps as a config option).

PR #5567, opened 2020-10-14
Remove author name and committer name. Fixes #5430

### LINKED ISSUE #5430: Too much information is leaked about the git repository
Git information uploaded to the pulumi website currently contains

```
git.author          Me
git.author.email    my@email
git.committer       Me
git.committer.email my@email
git.head            a6f1609eda40019363bffbd5384927ba40ca7519
git.headName        refs/heads/master
```

The author and committer is completely unnecessary there, since all of that can
already be found from the repository using the commit hash. I think there is
no reason to leak the information to the website.

cvhariharan:
This PR fixes #5430 where the author and committer names were being sent from the CLI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,9 @@ CHANGELOG
 - [sdk/dotnet] Fix HashSet concurrency issue.
   [#5563](https://github.com/pulumi/pulumi/pull/5563)
 
+- [cli] Remove author name and committer name
+  [#5567](https://github.com/pulumi/pulumi/pull/5567)
+
 ## 2.11.2 (2020-10-01)
 
 - feat(autoapi): expose EnvVars LocalWorkspaceOption to set in ctor
--- a/pkg/cmd/pulumi/util.go
+++ b/pkg/cmd/pulumi/util.go
@@ -624,9 +624,7 @@ func addGitCommitMetadata(repo *git.Repository, repoRoot string, m *backend.Upda
 	}
 
 	// Store committer and author information.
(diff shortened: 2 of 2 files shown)


chrsmith (changes requested):
Sorry for not coming across this sooner. I'll leave some more detailed notes on the root issue, but we should keep the author/committer names in the update metadata.

While it is true the author information could be derived from the commit hash, that also means that the Pulumi backend would need to be given access to the repo too. And that, IMHO, seems like the wrong tradeoff. Since the backend -- whether it is app.pulumi.com, or some other implementation -- would benefit from being able to say _who_ committed the code. (But does not need to know the contents of the code itself.)

So rather than just removing the metadata values entirely (since they are useful), it might be more prudent to simply allow an opt-out. For example, omitting all of the `GitCommiter` and `GitAuthor` metadata values if `PULUMI_OMIT_COMMIT_AUTHOR_METADATA` is truthy.

leezen:
@cvhariharan Any chance you've had time to take a look?

cvhariharan:
@leezen Sorry, I have been caught up with work and haven't been able to look into this. I'll try to look into this over the weekend and give an update.

leezen:
@cvhariharan Just checking in to see if you're planning to respond to the PR feedback?

cvhariharan:
Hi @leezen, I am not able to find time to look into this. Kindly assign this to someone else. Sorry, I forgot inform earlier.

leezen:
Thanks for the update. I'll close this PR for now.

PR #5599, opened 2020-10-20
[codegen/go] - Add support for enums

### LINKED ISSUE #5436: [codegen/go] Add go support for enums.

komalali:
Resolves #5436


pkg/codegen/internal/test/testdata/simple-enum-schema/schema.json
```diff
@@ -1,38 +1,38 @@
 {
   "version": "0.0.1",
-  "name": "plant-provider",
```
komalali, line 3: I couldn't get the go codegen to handle this hyphen, but it's clearly possible since azure-nextgen is hyphenated. I didn't spend too much time looking into it though.

PR #5605, opened 2020-10-20
Add tests for ts enums

### LINKED ISSUE #5598: [codegen/nodejs] - Translate enum output properties
One thing I've been wondering: The test schema only has `inputProperties` for the `RubberTree` resource. What do output properties look like for enums?

 - If it's a strict enum, will it be typed as the enum type? (And therefore needs to be converted to the enum type during deserialization?)

- If it's a relaxed enum (e.g. `Union[EnumClass, str]`), rather than the output being typed as `Union[EnumClass, str]`, should it be just `str` to make it easier for consumers (i.e. it will always be a string)?

_Originally posted by @justinvp in https://github.com/pulumi/pulumi/pull/5553#discussion_r507964874_

komalali:
Since our implementation of enums in TS is really just an alias for specific values of the underlying type, we don't need any special handling for serialization/deserialization and we can also continue to emit the outputs as the union of the Enum and underlying type since it doesn't require any additional handling by the consumer. This PR adds tests to confirm the expected behavior.

Resolves #5598


justinvp (approved):
LGTM

PR #5608, opened 2020-10-21
switch uuid packages

jaxxstorm:
The uuid package we're currently using to generate UUIDs is both
unmaintained and has a critical flaw that compromises its randomness

A forked, patched version is available but it changes the API. This is
an attempt to fix it


pkg/resource/deploy/builtins.go
```diff
@@ -114,7 +114,12 @@ func (p *builtinProvider) Create(urn resource.URN, inputs resource.PropertyMap,
 
 	var id resource.ID
 	if !preview {
-		id = resource.ID(uuid.NewV4().String())
+		// generate a new uuid
+		uuid, err := uuid.NewV4()
+		if err != nil {
+			return "", nil, resource.StatusPartialFailure, err
```
pgavlin, line 120:
```suggestion
			return "", nil, resource.StatusOK, err
```

pkg/resource/deploy/deploytest/provider.go
```diff
@@ -129,7 +129,12 @@ func (prov *Provider) Create(urn resource.URN, props resource.PropertyMap, timeo
 	preview bool) (resource.ID, resource.PropertyMap, resource.Status, error) {
 
 	if prov.CreateF == nil {
-		return resource.ID(uuid.NewV4().String()), resource.PropertyMap{}, resource.StatusOK, nil
+		// generate a new uuid
+		uuid, err := uuid.NewV4()
+		if err != nil {
+			return "", nil, resource.StatusPartialFailure, err
```
pgavlin, line 135:
```suggestion
			return "", nil, resource.StatusOK, err
```

pkg/resource/deploy/providers/registry.go
```diff
@@ -322,7 +322,12 @@ func (r *Registry) Create(urn resource.URN, news resource.PropertyMap, timeout f
 
 	var id resource.ID
 	if !preview {
-		id = resource.ID(uuid.NewV4().String())
+		// generate a new uuid
+		uuid, err := uuid.NewV4()
+		if err != nil {
+			return "", nil, resource.StatusPartialFailure, err
```
pgavlin, line 328:
```suggestion
			return "", nil, resource.StatusOK, err
```

pgavlin (approved):
`StatusPartialFailure` means specifically that the resource was partially created, and implies that the returned error is specially-formed s.t. it contains the partial state. These should just return `StatusOK`.

FWIW, I intend to go scrub these statuses from the codebase some day soon.

PR #5614, opened 2020-10-22
[sdk/dotnet] - Support enums

komalali:


justinvp:
I added a commit that implements serialization/deserialization.

sdk/dotnet/Pulumi.Tests/Serialization/EnumConverterTests.cs
```diff
@@ -0,0 +1,214 @@
...
+using System.Threading.Tasks;
+using Pulumi.Serialization;
+using Xunit;
+
+namespace Pulumi.Tests.Serialization
+{
+    public class EnumConverterTests : ConverterTests
+    {
+        [EnumType]
+        public readonly struct ContainerColor : IEquatable<ContainerColor>
```
justinvp, line 15: Here are the requirements for a struct-based enum to be able to be serialized/deserialized. It must:

 - Be a value type (`struct`) decorated with `EnumTypeAttribute`
 - Have a constructor that takes a single parameter of the underlying type (the constructor can be `private`)
 - Have an explicit conversion operator that converts the enum type to the underlying type
 - Have an underlying type of `String` or `Double`

Implementing `IEquatable`, adding `==`/`!=` operators, and overriding `ToString` isn't required, but is recommended and is what our codegen will do.

sdk/dotnet/Pulumi.Tests/Serialization/EnumConverterTests.cs
```diff
@@ -0,0 +1,214 @@
...
+            }
+
+            public static ContainerColor Red { get; } = new ContainerColor("red");
+            public static ContainerColor Blue { get; } = new ContainerColor("blue");
+            public static ContainerColor Yellow { get; } = new ContainerColor("yellow");
+
+            public static bool operator ==(ContainerColor left, ContainerColor right) => left.Equals(right);
+            public static bool operator !=(ContainerColor left, ContainerColor right) => !left.Equals(right);
+
+            public static explicit operator string(ContainerColor value) => value._value;
```
justinvp, line 31: To get the value out of the struct, I decided to opt for an explicit conversion operator that converts the enum type to its underlying type (in this case `string`), rather than have a `Value` property (or something else).

This makes getting the underlying value consistent with normal enums that are built-in to the language, e.g.:

```c#
int four = (int)ContainerSize.FourInch;
string red = (string)ContainerColor.Red;
```

sdk/dotnet/Pulumi.Tests/Serialization/EnumConverterTests.cs
```diff
@@ -0,0 +1,214 @@
...
+
+            public override string ToString() => _value;
+        }
+
+        [EnumType]
+        public readonly struct ContainerBrightness : IEquatable<ContainerBrightness>
+        {
+            private readonly double _value;
+
+            private ContainerBrightness(double value)
```
justinvp, line 48: I assume we want to support float-based enums (given the discussions we've had about the other languages).

Note that if that's the case, we'll need to specifically support `double` in the codegen. The emitted code is subtly different than for `string` (e.g. the constructor doesn't throw, `Equals` doesn't use `string.Equals`, `GetHashCode` doesn't have any null checks, and `ToString` calls `ToString()` on the `_value`).

mikhailshilkov: Have we discussed/seen a need for enums based on a boolean?

sdk/dotnet/Pulumi/Serialization/Attributes.cs
```diff
@@ -76,4 +76,12 @@ public sealed class OutputTypeAttribute : Attribute
     public sealed class OutputConstructorAttribute : Attribute
     {
     }
+
+    /// <summary>
+    /// Attribute used by a Pulumi Cloud Provider Package to mark enum types.
```
mikhailshilkov, line 81: Should we copy your nice comment about the expected structure of the enum type in here?

mikhailshilkov (approved):
Looks fabulous!

sdk/dotnet/Pulumi.Tests/Serialization/EnumConverterTests.cs
komalali: Boolean-based enums are supported by the schema (and nodejs) but I haven't seen them in practice anywhere yet. Are you aware of such enums in azure @mikhailshilkov? On the codegen side, C# will throw an unimplemented error if bool-based enums are present in the schema. I can create an issue and add it as a comment in the codegen unless you feel strongly that it should be implemented immediately.

mikhailshilkov: No, I haven't seen them yet. An issue sounds good.

PR #5615, opened 2020-10-23
[sdk/python] - Support enums

komalali:
Resolves: https://github.com/pulumi/pulumi/issues/5602


sdk/python/lib/test/test_next_serialize.py
```diff
@@ -1020,3 +1021,39 @@ def transformer(prop: str) -> str:
...
+        one = IntEnum.ONE
+        prop = await rpc.serialize_property(one, [])
+        self.assertEqual(IntEnum.ONE, prop)
+
+    @async_test
+    async def test_float_enum(self):
+        one = FloatEnum.ZERO_POINT_ONE
+        prop = await rpc.serialize_property(one, [])
+        self.assertEqual(FloatEnum.ZERO_POINT_ONE, prop)
+
```
komalali, line 1059: In a somewhat surprising turn of events, it turns out that these serialize just fine.

sdk/python/lib/pulumi/runtime/rpc.py
```diff
@@ -693,6 +709,10 @@ def register_resource_package(pkg: str, package: ResourcePackage):
     resource_packages.append(package)
 
 
+def _package_key(typ: str, version: str) -> str:
+    return f"{typ}@{version}"
+
+
```
justinvp, line 715: Bad merge?

justinvp (approved):
Looks like there may be one thing that was included after a rebase, but otherwise LGTM!

sdk/python/lib/pulumi/runtime/rpc.py
komalali: Ah thanks, good catch.

sdk/python/lib/pulumi/runtime/rpc.py
```diff
@@ -540,6 +551,10 @@ def translate_output_properties(output: Any,
                 raise AssertionError(f"Unexpected type. Expected 'list' got '{typ}'")
         return [translate_output_properties(v, output_transformer, element_type) for v in output]
 
+    if isinstance(output, (int, float, str)) and typ:
+        if issubclass(typ, Enum):
+            return typ(output)
```
pgavlin, line 556:
```suggestion
    if typ and isinstance(output, (int, float, str)) and issubclass(typ, Enum):
        return typ(output)
```

pgavlin (approved):
One nit; LGTM otherwise.

PR #5645, opened 2020-10-29
Add resource modules.

pgavlin:
This is necessary due to the way we've factored the libraries imported
by users into modules. The primary alternative is to ensure that each
child module imports the root module for a package and registers itself
with that package where necessary to prevent circular dependencies. This
simplifies the core SDKs slightly at the cost of greater complications
in the generated SDKs; the approach taken by these changes seems like a
more maintainable option.

The `ResourcePackage` types remain for the use of the root module for
a package, which will register itself in order to construct providers.

Contributes to #2430.

### DIFF
--- a/sdk/dotnet/Pulumi/Deployment/Deployment_ResourceModules.cs
+++ b/sdk/dotnet/Pulumi/Deployment/Deployment_ResourceModules.cs
@@ -9,7 +9,7 @@ namespace Pulumi
     {
         public static void RegisterResourceModule(string name, string version, IResourceModule module)
         {
-            ResourceModules.RegisterResourceModule(name, version, package);
+            ResourceModules.RegisterResourceModule(name, version, module);
         }
     }
 }
--- a/sdk/dotnet/Pulumi/Deployment/Deployment_ResourceModules.cs
+++ b/sdk/dotnet/Pulumi/Deployment/Deployment_ResourceModules.cs
@@ -7,7 +7,7 @@ namespace Pulumi
 {
     public partial class Deployment
     {
-        public static void RegisterResourceModule(string name, string version, IResourceModule package)
+        public static void RegisterResourceModule(string name, string version, IResourceModule module)
         {
             ResourceModules.RegisterResourceModule(name, version, package);
         }
--- a/sdk/dotnet/Pulumi/Deployment/Deployment_ResourceModules.cs
+++ b/sdk/dotnet/Pulumi/Deployment/Deployment_ResourceModules.cs
@@ -1,0 +1,15 @@
+// Copyright 2016-2020, Pulumi Corporation
+
+using System;
+using Pulumi.Serialization;
+
+namespace Pulumi
+{
+    public partial class Deployment
+    {
+        public static void RegisterResourceModule(string name, string version, IResourceModule package)
+        {
+            ResourceModules.RegisterResourceModule(name, version, package);
+        }
+    }
+}
--- a/sdk/dotnet/Pulumi/PublicAPI.Shipped.txt
+++ b/sdk/dotnet/Pulumi/PublicAPI.Shipped.txt
@@ -232,7 +232,7 @@ override Pulumi.Union<T0, T1>.ToString() -> string
 static Pulumi.ComponentResourceOptions.Merge(Pulumi.ComponentResourceOptions options1, Pulumi.ComponentResourceOptions options2) -> Pulumi.ComponentResourceOptions
 static Pulumi.CustomResourceOptions.Merge(Pulumi.CustomResourceOptions options1, Pulumi.CustomResourceOptions options2) -> Pulumi.CustomResourceOptions
 static Pulumi.Deployment.Instance.get -> Pulumi.DeploymentInstance
-static Pulumi.Deployment.RegisterResourceModule(string name, string version, Pulumi.IResourceModule package) -> void
+static Pulumi.Deployment.RegisterResourceModule(string name, string version, Pulumi.IResourceModule module) -> void
 static Pulumi.Deployment.RegisterResourcePackage(string name, string version, Pulumi.IResourcePackage package) -> void
 static Pulumi.Deployment.RunAsync(System.Action action) -> System.Threading.Tasks.Task<int>
 static Pulumi.Deployment.RunAsync(System.Func<System.Collections.Generic.IDictionary<string, object>> func) -> System.Threading.Tasks.Task<int>
--- a/sdk/dotnet/Pulumi/PublicAPI.Shipped.txt
+++ b/sdk/dotnet/Pulumi/PublicAPI.Shipped.txt
@@ -175,8 +175,9 @@ Pulumi.ResourceTransformationResult
 Pulumi.ResourceTransformationResult.Args.get -> Pulumi.ResourceArgs
 Pulumi.ResourceTransformationResult.Options.get -> Pulumi.ResourceOptions
 Pulumi.ResourceTransformationResult.ResourceTransformationResult(Pulumi.ResourceArgs args, Pulumi.ResourceOptions options) -> void
+Pulumi.IResourceModule
+Pulumi.IResourceModule.Construct(string name, string type, System.Collections.Generic.IDictionary<string, object> args, string urn) -> Pulumi.Resource
 Pulumi.IResourcePackage
-Pulumi.IResourcePackage.Construct(string name, string type, System.Collections.Generic.IDictionary<string, object> args, string urn) -> Pulumi.Resource
 Pulumi.IResourcePackage.ConstructProvider(string name, string type, System.Collections.Generic.IDictionary<string, object> args, string urn) -> Pulumi.ProviderResource
 Pulumi.Stack
 Pulumi.Stack.Stack(Pulumi.StackOptions options = null) -> void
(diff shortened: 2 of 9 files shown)


lblackstone (approved):

sdk/nodejs/runtime/rpc.ts
```diff
@@ -601,3 +603,29 @@ export function registerResourcePackage(pkgName: string, version: string, pkg: R
     }
     resourcePackages.set(key, pkg);
 }
+
+/**
+ * A ResourceModule is a type that understands how to construct resources given a name, type, args, and URN.
+ */
+export type ResourceModule = {
```
justinvp, line 610: Nit: Any reason this is a type alias rather than an interface? Just wondering if it could be an interface (same for `ResourcePackage`). https://www.typescriptlang.org/docs/handbook/advanced-types.html#interfaces-vs-type-aliases

sdk/dotnet/Pulumi/Deployment/Deployment_ResourceModules.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright 2016-2020, Pulumi Corporation
+
+using System;
+using Pulumi.Serialization;
+
+namespace Pulumi
+{
+    public partial class Deployment
+    {
+        public static void RegisterResourceModule(string name, string version, IResourceModule package)
```
justinvp, line 10:
```suggestion
        public static void RegisterResourceModule(string name, string version, IResourceModule module)
```

?

sdk/dotnet/Pulumi/PublicAPI.Shipped.txt
```diff
@@ -229,6 +230,7 @@ override Pulumi.Union<T0, T1>.ToString() -> string
 static Pulumi.ComponentResourceOptions.Merge(Pulumi.ComponentResourceOptions options1, Pulumi.ComponentResourceOptions options2) -> Pulumi.ComponentResourceOptions
 static Pulumi.CustomResourceOptions.Merge(Pulumi.CustomResourceOptions options1, Pulumi.CustomResourceOptions options2) -> Pulumi.CustomResourceOptions
 static Pulumi.Deployment.Instance.get -> Pulumi.DeploymentInstance
+static Pulumi.Deployment.RegisterResourceModule(string name, string version, Pulumi.IResourceModule package) -> void
```
justinvp, line 233:
```suggestion
static Pulumi.Deployment.RegisterResourceModule(string name, string version, Pulumi.IResourceModule module) -> void
```

?

sdk/dotnet/Pulumi/Serialization/ResourceModules.cs
```diff
@@ -0,0 +1,40 @@
+// Copyright 2016-2019, Pulumi Corporation
```
justinvp, line 1: Nit:

```suggestion
// Copyright 2016-2020, Pulumi Corporation
```

sdk/dotnet/Pulumi/Serialization/ResourceModules.cs
```diff
@@ -0,0 +1,40 @@
...
+        }
+
+        public static void RegisterResourceModule(string name, string version, IResourceModule package)
+        {
+            var key = ModuleKey(name, version);
+            if (!_resourceModules.TryAdd(key, package))
+            {
+                throw new InvalidOperationException($"Cannot re-register package {key}.");
+            }
+        }
```
justinvp, line 38:
```suggestion
        public static void RegisterResourceModule(string name, string version, IResourceModule module)
        {
            var key = ModuleKey(name, version);
            if (!_resourceModules.TryAdd(key, module))
            {
                throw new InvalidOperationException($"Cannot re-register module {key}.");
            }
        }
```

?

sdk/go/pulumi/rpc.go
```diff
@@ -697,3 +699,22 @@ func RegisterResourcePackage(name, version string, pkg ResourcePackage) {
...
+}
+
+// RegisterResourceModule register a resource module with the Pulumi runtime.
+func RegisterResourceModule(name, version string, pkg ResourceModule) {
+	key := moduleKey(name, version)
+	existing, hasExisting := resourceModules.LoadOrStore(key, pkg)
+	if hasExisting {
+		panic(fmt.Errorf("a resource module for %v is already registered: %v", key, existing))
+	}
+}
```
justinvp, line 720:
```suggestion
func RegisterResourceModule(name, version string, module ResourceModule) {
	key := moduleKey(name, version)
	existing, hasExisting := resourceModules.LoadOrStore(key, module)
	if hasExisting {
		panic(fmt.Errorf("a resource module for %v is already registered: %v", key, existing))
	}
}
```

sdk/nodejs/runtime/rpc.ts
```diff
@@ -601,3 +603,29 @@ export function registerResourcePackage(pkgName: string, version: string, pkg: R
...
+ * the module name and version that are deserialized by the current instance of the Pulumi JavaScript SDK.
+ */
+export function registerResourceModule(pkgName: string, version: string, pkg: ResourceModule) {
+    const key = moduleKey(pkgName, version);
+    const existing = resourceModules.get(key);
+    if (existing) {
+        throw new Error(`Cannot re-register module ${key}. Previous registration was ${existing}, new registration was ${pkg}.`);
+    }
+    resourceModules.set(key, pkg);
+}
```
justinvp, line 631:
```suggestion
export function registerResourceModule(name: string, version: string, module: ResourceModule) {
    const key = moduleKey(name, version);
    const existing = resourceModules.get(key);
    if (existing) {
        throw new Error(`Cannot re-register module ${key}. Previous registration was ${existing}, new registration was ${pkg}.`);
    }
    resourceModules.set(key, module);
}
```

?

sdk/nodejs/runtime/rpc.ts
```diff
@@ -590,7 +592,7 @@ function packageKey(name: string, version: string): string {
 }
 
 /**
- * registerResourcePackage registers a resource package that will be used to construct resources for any URNs matching
+ * registerResourcePackage registers a resource package that will be used to construct providers for any URNs matching
  * the package name and version that are deserialized by the current instance of the Pulumi JavaScript SDK.
  */
 export function registerResourcePackage(pkgName: string, version: string, pkg: ResourcePackage) {
```
justinvp, line 598:
```suggestion
export function registerResourcePackage(name: string, version: string, pkg: ResourcePackage) {
```

?

sdk/python/lib/pulumi/runtime/rpc.py
```diff
@@ -654,3 +660,15 @@ def register_resource_package(typ: str, version: str, package):
     if existing is not None:
         raise ValueError(f"Cannot re-register package {key}. Previous registration was {existing}, new registration was {package}.")
     RESOURCE_PACKAGES[key] = package
+
+RESOURCE_MODULES: Dict[str, Any] = dict()
```
justinvp, line 664: Instead of the values being typed as `Any` and assuming they are objects with `construct` function (or `construct_provider` for the case of `RESOURCE_PACKAGES`), should these just be functions typed as `Callable`?

I was kind of wondering the same for the other languages, especially now that `IResourcePackage`/`IResourceModule` are single function interfaces (e.g. C# should these just be delegates or even `Func<>`? Same for TS and Go).

justinvp, line 664:
```suggestion
_RESOURCE_MODULES: Dict[str, Any] = dict()
```

Also `RESOURCE_PACKAGES` => `_RESOURCE_PACKAGES`

sdk/python/lib/pulumi/runtime/rpc.py
```diff
@@ -654,3 +660,15 @@ def register_resource_package(typ: str, version: str, package):
...
+
+def module_key(typ: str, version: str) -> str:
```
justinvp, line 666:
```suggestion
def _module_key(typ: str, version: str) -> str:
```

Also `package_key` => `_package_key`

sdk/dotnet/Pulumi/Serialization/ResourceModules.cs
```diff
@@ -0,0 +1,40 @@
...
+namespace Pulumi
+{
+    public interface IResourceModule
+    {
+        Resource Construct(string name, string type, IDictionary<string, object?>? args, string urn);
+    }
+
+    internal static class ResourceModules
+    {
+        internal static ConcurrentDictionary<string, IResourceModule> _resourceModules = new ConcurrentDictionary<string, IResourceModule>();
```
justinvp, line 19: Nit:

```suggestion
        internal static readonly ConcurrentDictionary<string, IResourceModule> _resourceModules = new ConcurrentDictionary<string, IResourceModule>();
```

sdk/nodejs/runtime/rpc.ts
```diff
@@ -516,25 +516,28 @@ export function deserializeProperty(prop: any): any {
...
-                        throw new Error(`Unable to deserialize resource URN ${urn}, no resource package is registered for type ${type}.`);
+                    const resourceModule = resourceModules.get(moduleKey(modName, version || ""));
+                    if (!resourceModule) {
+                        throw new Error(`Unable to deserialize resource ${urn}, no module is registered for ${modName}.`);
                     }
-                    const urnName = urnParts[3];
-                    return !isProvider ?
-                        resourcePackage.construct(urnName, type, {}, { urn }) :
-                        resourcePackage.constructProvider(urnName, type, {}, { urn });
+                    return resourceModule.construct(urnName, type, {}, { urn });
```
justinvp, line 540: Passing the empty `{}` for the args is just point-in-time until `getResource` from your other PR is exposed to get the actual state?

justinvp (approved):
LGTM otherwise.

sdk/python/lib/pulumi/runtime/rpc.py
justinvp: Or perhaps we should use abstract classes in Python? Something like:

```python
from abc import ABCMeta

class ResourcePackage(metaclass=ABCMeta):
    @abstractmethod
    def construct_provider(name: str, typ: str, args: Mapping[str, Any], opts: Mapping[str, Any]) -> ProviderResource:
        ...

class ResourceModule(metaclass=ABCMeta):
    @abstractmethod
    def construct(name: str, typ: str, args: Mapping[str, Any], opts: Mapping[str, Any]) -> Resource:
        ...
```

sdk/nodejs/runtime/rpc.ts
pgavlin: Yes, we should probably make these interfaces. Not sure why I used a type alias, tbh.

sdk/python/lib/pulumi/runtime/rpc.py
pgavlin: I could have sworn that I had added a `ResourcePackage` abstract class, but I haven't. Time to dig through some old changes.

sdk/nodejs/runtime/rpc.ts
pgavlin: Yes, that's right.

sdk/python/lib/pulumi/runtime/rpc.py
pgavlin: Done.

Issue #5648, opened 2020-10-29
Windows build break


justinvp:
https://github.com/pulumi/pulumi/runs/1329251775

Error installing grpcio:

```
         [pipenv.exceptions.InstallError]:    command: 'C:\Users\runneradmin\.virtualenvs\python-FORTJMCV\Scripts\python.exe' -u -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pip-install-9bubz6dw\\grpcio\\setup.py'"'"'; __file__='"'"'C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pip-install-9bubz6dw\\grpcio\\setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' bdist_wheel -d 'C:\Users\runneradmin\AppData\Local\Temp\pip-wheel-cz0n5f0s'
542
         [pipenv.exceptions.InstallError]:        cwd: C:\Users\runneradmin\AppData\Local\Temp\pip-install-9bubz6dw\grpcio\
543
         [pipenv.exceptions.InstallError]:   Complete output (79 lines):
544
         [pipenv.exceptions.InstallError]:   C:\Users\runneradmin\AppData\Local\Temp\pip-install-9bubz6dw\grpcio\src\python\grpcio\commands.py:102: SyntaxWarning: "is not" with a literal. Did you mean "!="?
```

Possibly due to running on Python 3.9 instead of 3.7?

More logs:

```
     1>Target "PythonDependencies" in project "D:\a\pulumi\pulumi\src\github.com\pulumi\pulumi\build.proj" (target "BuildPythonSDK" depends on it):
500
       Task "MakeDir"
501
         Creating directory "D:\a\pulumi\pulumi\src\github.com\pulumi\pulumi\\sdk\\python\\env\src".
502
       Done executing task "MakeDir".
503
       Task "Exec"
504
         pipenv install
505
         Creating a virtualenv for this projectΓÇª
506
         
507
         Pipfile: D:\a\pulumi\pulumi\src\github.com\pulumi\pulumi\sdk\python\Pipfile
508
         
509
[...]

PR #5649, opened 2020-10-30
Trying to solve Python build on windows

stack72:
Fixes: #5648

Recently, GHA windows image was updated with Python 3.9. We didn't specific a specific version of Python when creating a Pipenv so it was using the latest version installed - marking this as a specific version ignores the 3.9 installed as default

### DIFF
--- a/.github/workflows/windows-pr.yml
+++ b/.github/workflows/windows-pr.yml
@@ -18,8 +18,8 @@ jobs:
         platform: [windows-latest]
         go-version: [1.15.x]
         node-version: [10.x]
-        python-version: [3.7]
-        dotnet: ['3.1.301']
+        python-version: [3.7.x]
+        dotnet: [3.1.x]
     runs-on: ${{ matrix.platform }}
     env:
       GOPATH: ${{ github.workspace }}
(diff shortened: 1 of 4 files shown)


.github/workflows/windows-pr.yml
```diff
@@ -37,13 +37,17 @@ jobs:
...
       with:
         python-version: ${{ matrix.python-version }}
+    - name: Display Python version
+      run: python -c "import sys; print(sys.version)"
     - name: Install Python Deps
       run: |
         pip3 install pyenv-win
         pip3 install pipenv
+    - name: Display Python version
+      run: python -c "import sys; print(sys.version)"
```
mikhailshilkov, line 50: Is this for debugging or you want both of these two steps to stay?

mikhailshilkov (approved):
LGTM (although I can't say I understand version management in Python)

.github/workflows/windows-pr.yml
stack72: Sorry, yes, I should remove that!! That was indeed for debugging

Issue #5662, opened 2020-11-02
Support Python 3.9


leezen:
We need to make sure that pulumi works correctly on Python 3.9

komalali:
Looks like grpcio is still working through Windows support https://github.com/grpc/grpc/issues/24344

komalali:
`protobuf` doesn't have python 3.9 support yet, so it’s missing the equivalent of the `_message.cpython-38-darwin.so` file that is needed for the `SetAllowOversizeProtos` method. However, it appears we only use that method to override a default that is only really applicable for a k8s bug, more context in [this issue](https://github.com/pulumi/pulumi/issues/3981). 

However, I think we can wrap this in a try/except since it's not always needed, and not having the try/except is causing `Getting Started` workflows for folks using python 3.9 to fail.

Issue #5664, opened 2020-11-02
Check the language matrix for builds


leezen:

stack72:
Now that we are consistent for our builds I am going to close this out. We have the following matrix as of January 2021:

* Python 3.9.x
* Go 1.15.x
* NodeJS 14.x

Any other changes will be tested and coordinated

PR #5669, opened 2020-11-03
Support python 3.9

komalali:
Fixes: #5662


sdk/go/common/util/archive/archive.go
```diff
@@ -76,6 +76,7 @@ func UnTGZ(tarball []byte, dir string) error {
 			return errors.Wrapf(err, "untarring")
 		}
 
+		// nolint:gosec
```
komalali, line 79: Had to add this because gosec was complaining (only in CI, not locally)

stack72 (approved):
LGTM! Subject to adding the appropriate TO-DO items with regards to supporting Python 3.9 on Windows when the package is available

lblackstone (approved):

.github/workflows/master.yml
```diff
@@ -25,7 +25,7 @@ jobs:
     strategy:
       matrix:
         go-version: [ 1.15.x ]
-        python-version: [ 3.7.x ]
+        python-version: [ 3.9.x ]
```
justinvp, line 28: Not really a comment for this PR, but hopefully as part of https://github.com/pulumi/pulumi/issues/5664 we can find some way to have coverage of all the versions of Python we support (e.g. 3.6.x, 3.7.x, 3.8.x, and 3.9.x). Maybe as some cron run, and keep the matrix smaller for PRs? cc: @stack72

sdk/python/cmd/pulumi-language-python-exec
```diff
@@ -9,6 +9,17 @@ import sys
 import traceback
 import runpy
 
+# python 3.9 does not have grpcio support on Windows yet [1]. Return an error suggesting a python downgrade in this
+# case. Remove this check once Windows is supported [2].
+# [1] https://github.com/grpc/grpc/issues/24344
+# [2] https://github.com/pulumi/pulumi/issues/5676
```
justinvp, line 15: Nit: minor rearrangement to make the TODO more clear.

```suggestion
# python 3.9 does not have grpcio support on Windows yet (https://github.com/grpc/grpc/issues/24344).
# Return an error suggesting a python downgrade in this case.
# TODO[pulumi/pulumi#5676]: Remove this check once Windows is supported.
```

justinvp (approved):
LGTM

.github/workflows/master.yml
komalali: Yeah I agree, I had initially run this as a matrix using `[3.7.x, 3.9.x]` - I think it makes sense to do that as a cron.

PR #5670, opened 2020-11-03
Adding support for go enum generation

### LINKED ISSUE #5436: [codegen/go] Add go support for enums.

viveklak:
Resolves #5436 

After a lot of back and forth on what the input/output types should be for enum types I have settled on the following:
1. Enum type will be an alias for one of the pulumi wrapped `pulumi` Apply-able primitives, e.g. `pulumi.String`, `pulumi.Bool`, `pulumi.Int` etc.
2. The enum type will itself implement the underlying wrapped type's input interface
3. Strict enums will refer to the enum type alias while relaxed enums would refer to the aliased input type

This results in usage in pulumi programs as follows:
```
                group, err := ec2.NewSecurityGroup(ctx, "web-secgrp", &ec2.SecurityGroupArgs{
                        Ingress: ec2.SecurityGroupIngressArray{
                                ec2.SecurityGroupIngressArgs{
-                                       Protocol:   pulumi.String("tcp"),
+                                       Protocol:   ec2.TCP,
                                        FromPort:   pulumi.Int(80),
                                        ToPort:     pulumi.Int(80),
                                        CidrBlocks: pulumi.StringArray{pulumi.String("0.0.0.0/0")},
@@ -41,8 +41,8 @@
 
                // Create a simple web server using the startup script for the instance.
                srv, err := ec2.NewInstance(ctx, "web-server-www", &ec2.InstanceArgs{
-                       Tags:                pulumi.Map{"Name": pulumi.String("web-server-www")},
-                       InstanceType:        pulumi.String("t2.micro"), // t2.micro is available in the AWS free tier.
+                       Tags:                pulumi.StringMap{"Name": pulumi.String("web-server-www")},
+                       InstanceType:        ec2.T2_Micro, // t2.micro is available in the AWS free tier.
                        VpcSecurityGroupIds: pulumi.StringArray{group.ID()},
                        Ami:                 pulumi.String(ami.Id),
                        UserData: pulumi.String(`#!/bin/bash
```


EvanBoyle (commented):
Just capturing a few comments after our discussion. I know this is a draft but I thought it might be helpful to capture them regardless. Sorry if this is needless noise. 

1. We need type assertions wherever we call the `Validate()` function. 
2. Not sure if this is just handled by the underlying sdk gen machinery, but it would be worth verifying that naming collisions are appropriately handled (enum type collision with a pre-existing type in the same package). 
3. I don't see any updates to program gen. Given the implementation strategy, all examples for Go will continue to be valid so this isn't strictly necessary.

komalali:
@EvanBoyle yeah program gen isn't in scope for the first round of changes but definitely something I'd like to take a look at once this is wrapped up. As you said, not strictly necessary immediately but certainly relevant for the azure nextgen SDK where we will be using strict enums.

EvanBoyle:
@viveklak @komalali your comment brings a potential issue to mind. Strict enum callsite will need a call to `Validate` but a member function is not enough as the plain type won't have the validate member. Does `Validate` need to be a static function or equivalent? Would it be better to either:

a. call `Validate` if you're working with an enum type
b. construct a new enum type from the plain type and have the constructor delegate the validation?

viveklak:
> @viveklak @komalali your comment brings a potential issue to mind. Strict enum callsite will need a call to `Validate` but a member function is not enough as the plain type won't have the validate member. Does `Validate` need to be a static function or equivalent? Would it be better to either:
> 
> a. call `Validate` if you're working with an enum type
> b. construct a new enum type from the plain type and have the constructor delegate the validation?

The validate member function is only on the enum type and will only be called on strict enum references. For instance,
taking the test schema as example, the Args parameter type for the RubberTree resource is generated as follows:
```
// The set of arguments for constructing a RubberTree resource.
type RubberTreeArgs struct {
	Container plant.ContainerPtrInput
	Farm      pulumi.StringPtrInput
	Type      RubberTreeVariety
}
```

Here `Farm` is a relaxed (and optional) enum reference, while `Type` is a strict enum. We only call validate in the resource's constructor for the RubberTreeVariety enum and not for the relaxed types (where as you mentioned, its not accessible):
```
func NewRubberTree(ctx *pulumi.Context,
	name string, args *RubberTreeArgs, opts ...pulumi.ResourceOption) (*RubberTree, error) {
	if args == nil {
		return nil, errors.New("missing one or more required arguments")
	}
	if err := args.Type.Validate(); err != nil {
		return nil, fmt.Errorf("invalid value for required argument 'Type': %w", err)
	}
...
}
```

See the tests I just added here: https://github.com/pulumi/pulumi/pull/5670/files#diff-75be79a9073076cb4553e27d2521d3d1693bce90cba86a1c797757b388255279R123

viveklak:
> Just capturing a few comments after our discussion. I know this is a draft but I thought it might be helpful to capture them regardless. Sorry if this is needless noise.
> 
> 1. We need type assertions wherever we call the `Validate()` function.

As discussed, this function should only be on references to the enum type so I think we don't need to type assert here.

> 2. Not sure if this is just handled by the underlying sdk gen machinery, but it would be worth verifying that naming collisions are appropriately handled (enum type collision with a pre-existing type in the same package).

I discussed this with @komalali and it seemed like the onus here should be with the provider/schema rather than go gen defending against it. That said, detecting collisions early seems like a good idea. Do you have recommendations on how to achieve that? Do we do something like this already?

EvanBoyle:
@viveklak there's a top-level `pkg.names` map that contains all symbol names. You can use it to check for collisions as we do with function names (new/create): https://github.com/pulumi/pulumi/blob/da1f25928d6fa4d9b9645f005f80654586f14d90/pkg/codegen/go/gen.go#L1184. 

We also do this for types already, but I haven't checked to see if this is covered by the enum code path: 
https://github.com/pulumi/pulumi/blob/da1f25928d6fa4d9b9645f005f80654586f14d90/pkg/codegen/go/gen.go#L156

pkg/codegen/go/gen.go
```diff
@@ -1365,6 +1573,25 @@ func GeneratePackage(tool string, pkg *schema.Package) (map[string][]byte, error
...
+				pkg.getImports(e, imports)
+			}
+
+			buffer := &bytes.Buffer{}
+			pkg.genHeader(buffer, []string{"context", "reflect", "fmt"}, imports)
+
+			for _, e := range pkg.enums {
+				pkg.genEnum(buffer, e)
+			}
+			// pkg.genEnumTypeRegistrations(buffer, pkg.enums)
```
komalali, line 1589: Commented code, delete if not needed.

pkg/codegen/internal/test/testdata/simple-enum-schema/go/plant/pulumiEnums.go
```diff
@@ -0,0 +1,101 @@
...
+	"reflect"
+
+	"github.com/pulumi/pulumi/sdk/v2/go/pulumi"
+)
+
+// plant container colors
+type ContainerColor pulumi.String
+
+const (
+	Red    = ContainerColor("red")
```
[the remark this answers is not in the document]
viveklak: Good catch. I updated to use <typename><enum val> instead since it seems a bit more idiomatic and easier to discover options through IDE intellisense. I also added a change to use a `_` separator if the enum value contains `_`. See https://github.com/pulumi/pulumi/pull/5670/files#diff-0b3967193235a3bfca3614bcccb84337da74174690c263aecdcc2216c42239d2R17 for an example.

viveklak:
As discussed with @komalali and @leezen, no longer performing validation on strict enums.

pkg/codegen/go/gen.go
```diff
@@ -468,6 +508,100 @@ func genInputMethods(w io.Writer, name, receiverType, elementType string, ptrMet
...
+	case schema.StringType:
+		return "pulumi.String" + suffix
+	default:
+		// We only expect to support the above element types for enums
+		panic(t)
+	}
+}
+
+func makeSafeEnumName(name string) string {
+	return makeValidIdentifier(Title(name))
```
komalali, line 573: We had discussed bringing in the [common enum replacements](https://github.com/pulumi/pulumi/blob/master/pkg/codegen/utilities.go#L124-L129) here... is that still the plan?

pkg/codegen/go/gen.go
```diff
@@ -468,6 +508,100 @@ func genInputMethods(w io.Writer, name, receiverType, elementType string, ptrMet
...
+		return "pulumi.Bool" + suffix
+	case schema.IntType:
+		return "pulumi.Int" + suffix
+	case schema.NumberType:
+		return "pulumi.Float64" + suffix
+	case schema.StringType:
+		return "pulumi.String" + suffix
+	default:
+		// We only expect to support the above element types for enums
+		panic(t)
```
komalali, line 568:
```suggestion
		panic(fmt.Sprintf("Invalid enum type: %s", t))
```

pkg/codegen/go/gen.go
```diff
@@ -468,6 +508,100 @@ func genInputMethods(w io.Writer, name, receiverType, elementType string, ptrMet
...
+
+func (pkg *pkgContext) genEnumInputFuncs(w io.Writer, typeName string, enum *schema.EnumType, elementType, inputType string) {
+	fmt.Fprintln(w)
+	asFuncName := Title(strings.Replace(elementType, "pulumi.", "", -1))
+	fmt.Fprintf(w, "func (%s) ElementType() reflect.Type {\n", typeName)
+	fmt.Fprintf(w, "return reflect.TypeOf((*%s)(nil)).Elem()\n", elementType)
+	fmt.Fprintln(w, "}")
+	fmt.Fprintln(w)
+
+	fmt.Fprintf(w, "func (e %[1]s) To%[2]sOutput() %[3]sOutput {\n", typeName, asFuncName, elementType)
```
komalali, line 584: Nit: No need for `[1]`, `[2]`, etc. if we're not repeating the same element. 
```suggestion
	fmt.Fprintf(w, "func (e %s) To%sOutput() %sOutput {\n", typeName, asFuncName, elementType)
```

pkg/codegen/go/gen.go
```diff
@@ -1352,7 +1541,7 @@ func GeneratePackage(tool string, pkg *schema.Package) (map[string][]byte, error
 		// Run Go formatter on the code before saving to disk
 		formattedSource, err := format.Source([]byte(contents))
 		if err != nil {
-			panic(errors.Errorf("invalid Go source code:\n\n%s", contents))
+			panic(fmt.Errorf("invalid Go source code:\n\n%s: %s: %w", relPath, contents, err))
```
komalali, line 1544: Prefer `errors.Errorf`

```suggestion
			panic(errors.Errorf("invalid Go source code:\n\n%s: %s: %w", relPath, contents, err))
```

pkg/codegen/go/gen.go
```diff
@@ -468,6 +508,100 @@ func genInputMethods(w io.Writer, name, receiverType, elementType string, ptrMet
...
+		enumName := makeSafeEnumName(elementName)
+		if strings.Contains(enumName, "_") {
+			enumName = fmt.Sprintf("_%s", enumName)
+		}
+		e.Name = name + enumName
+		contract.Assertf(!modPkg.names.has(e.Name), "Name collision for enum constant: %s for %s",
+			e.Name, enumType.Token)
+		switch reflect.TypeOf(e.Value).Kind() {
+		case reflect.String:
+			fmt.Fprintf(w, "%s = %s(\"%v\")\n", e.Name, name, e.Value)
```
komalali, line 540:
```suggestion
			fmt.Fprintf(w, "%s = %s(%q)\n", e.Name, name, e.Value)
```

komalali (commented):
This is looking really good! I left a few comments about the tests and single char enum name replacements, after that we should be good to go.

pkg/codegen/go/gen.go
viveklak: Yes, sorry missed that. Updated.

pkg/codegen/go/gen_test.go
```diff
@@ -42,41 +48,163 @@ func TestInputUsage(t *testing.T) {
...
 func TestGeneratePackage(t *testing.T) {
 	tests := []struct {
 		name          string
 		schemaDir     string
 		expectedFiles []string
 	}{
 		{
-			"Simple schema with local resource properties",
-			"simple-resource-schema",
+			"Simple schema with enum types",
```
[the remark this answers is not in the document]
viveklak: Fixed.

pkg/codegen/go/gen.go
komalali: Might also be worth adding a [check](https://github.com/pulumi/pulumi/blob/41a881a37418291b3278ba32e62b5fbd4e3a8af0/pkg/codegen/nodejs/utilities.go#L110-L113) for names that are one illegal character.

komalali (approved):
🎉 🎉 Looks great!

komalali (approved):
🚀 🚀 🚀

Issue #4595, opened 2020-05-08
Node 14 compatibility


jaxxstorm:
We need to determine if node 14 works with Pulumi. There are related issues indicated that there may be some breaking grpc changes that might affect us:

https://github.com/nodejs/node/issues/32463
https://github.com/grpc/grpc-node/issues/1407

danfhernandez:
Ran into a node 14 issue this morning that may be related.  I downgraded and things started working properly.  See error log below: 

```
node-pre-gyp WARN Using needle for node-pre-gyp https download 
node-pre-gyp WARN Tried to download(404): https://node-precompiled-binaries.grpc.io/grpc/v1.24.2/node-v83-darwin-x64-unknown.tar.gz 
node-pre-gyp WARN Pre-built binaries not found for grpc@1.24.2 and node@14.2.0 (node-v83 ABI, unknown) (falling back to source compile with node-gyp) 
No receipt for 'com.apple.pkg.CLTools_Executables' found at '/'.

No receipt for 'com.apple.pkg.DeveloperToolsCLILeo' found at '/'.

No receipt for 'com.apple.pkg.DeveloperToolsCLI' found at '/'.

gyp: No Xcode or CLT version detected!
gyp ERR! configure error 
gyp ERR! stack Error: `gyp` failed with exit code: 1
gyp ERR! stack     at ChildProcess.onCpExit (/usr/local/lib/node_modules/npm/node_modules/node-gyp/lib/configure.js:351:16)
gyp ERR! stack     at ChildProcess.emit (events.js:315:20)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:276:12)
gyp ERR! System Darwin 19.4.0
gyp ERR! command "/usr/local/Cellar/node/14.2.0/bin/node" "/usr/local/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "configure" "--fallback-to-build" "--library=static_library" "--module=/Users/dan/Documents/playground/support/s3-zendesk-import/node_modules/grpc/src/node/extension_binary/node-v83-darwin-x64-unknown/grpc_node.node" "--module_name=grpc_node" "--module_path=/Users/dan/Documents/playground/support/s3-zendesk-import/node_modules/grpc/src/node/extension_binary/node-v83-darwin-x64-unknown" "--napi_version=6" "--node_abi_napi=napi" "--napi_build_version=0" "--node_napi_label=node-v83"
[...]

PR #5673, opened 2020-11-03
Upgrade the build to target Node14 now that it is LTS

stack72:
Fixes: #4595

### DIFF
--- a/.github/workflows/codegen-test.yml
+++ b/.github/workflows/codegen-test.yml
@@ -9,9 +9,12 @@ env:
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
 jobs:
-  downstream-aws:
-    name: Test AWS Downstream
+  downstream-test:
+    name: Test ${{ matrix.provider }} Downstream
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        provider: ["aws", "gcp", "azure", "azuread", "random", "kubernetes"]
     steps:
       - name: Install Go
         uses: actions/setup-go@v1
@@ -20,7 +23,7 @@ jobs:
       - name: Install Node.js
         uses: actions/setup-node@v1
         with:
-          node-version: '10.x'
+          node-version: '14.x'
       - name: Install Python
         uses: actions/setup-python@v1
         with:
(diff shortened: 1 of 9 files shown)


.github/workflows/codegen-test.yml
```diff
@@ -9,9 +9,12 @@ env:
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
 jobs:
-  downstream-aws:
-    name: Test AWS Downstream
+  downstream-test:
```
stack72, line 12: This is refactored to just require new providers to be added to the list rather than entire codeblocks to be added

.github/workflows/master.yml
```diff
@@ -75,6 +75,9 @@ jobs:
     name: Publish Binaries
     runs-on: macos-latest
     needs: build-and-test
+    strategy:
```
stack72, line 78: We were missing the strategy block

.github/workflows/prerelease.yml
```diff
@@ -74,6 +74,9 @@ jobs:
     name: Publish Binaries
     runs-on: macos-latest
     needs: build-and-test
+    strategy:
```
stack72, line 77: We were missing the strategy block

.github/workflows/release.yml
```diff
@@ -147,6 +147,9 @@ jobs:
     name: Publish Binaries
     runs-on: macos-latest
     needs: build-and-test
+    strategy:
```
stack72, line 150: We were missing the strategy block

.github/workflows/res-docs-test.yml
```diff
@@ -21,20 +21,23 @@ jobs:
     env:
       GOPATH: ${{ github.workspace }}
     runs-on: ubuntu-latest
-
+    strategy:
```
stack72, line 24: We were missing the strategy block

.github/workflows/master.yml
komalali: Do we need it if there's no matrix? i.e. there is only 1 version.

The [documentation](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix) seems to imply it's only necessary if there is more than 1 version/language/whatever that we are testing.

komalali (approved):
Just a comment on whether or not we actually need `strategy.matrix` if we're not testing in a number of configurations, but otherwise LGTM!

.github/workflows/master.yml
stack72: I am being cunning here! this means that we have a single place to update the versions of our builds - this will allow me to add multiple go versions / multiple dotnet / python etc soon! So I am kinda future proofing the scripts

komalali: Makes sense 👍

PR #5694, opened 2020-11-05
Add support for getResource to the Python SDK.

pgavlin:
Just what it says on the tin.

These changes take advantage of the fact that Python invokes are
synchronous to fetch the resource state inside of its constructor. The
SDK code generator will be updated to use the new `urn` parameter inside
of each module's implementation of `ResourceModule.construct`.

Part of #2430.

### DIFF
--- a/pkg/resource/deploy/builtins.go
+++ b/pkg/resource/deploy/builtins.go
@@ -272,7 +272,7 @@ func (p *builtinProvider) getResource(inputs resource.PropertyMap) (resource.Pro
 	contract.Assert(ok)
 	contract.Assert(urn.IsString())
 
-	// TODO(pdg): track secret outputs in the resource map. This is necessary to ensure that the
+	// #5803: track secret outputs in the resource map. This is necessary to ensure that the
 	// `additionalSecretOutputs` option that was provided when the resource was registered is properly respected by
 	// `getResource`.
 
--- a/pkg/resource/deploy/builtins.go
+++ b/pkg/resource/deploy/builtins.go
@@ -272,6 +272,10 @@ func (p *builtinProvider) getResource(inputs resource.PropertyMap) (resource.Pro
 	contract.Assert(ok)
 	contract.Assert(urn.IsString())
 
+	// TODO(pdg): track secret outputs in the resource map. This is necessary to ensure that the
+	// `additionalSecretOutputs` option that was provided when the resource was registered is properly respected by
+	// `getResource`.
+
 	state, ok := p.resources.get(resource.URN(urn.StringValue()))
 	if !ok {
 		return nil, errors.Errorf("unknown resource %v", urn.StringValue())
--- a/sdk/python/lib/pulumi/resource.py
+++ b/sdk/python/lib/pulumi/resource.py
@@ -723,7 +723,7 @@ class Resource:
 
         if urn is not None:
             # This is a resource that already exists. Read its state from the engine.
-            get_resource(self, custom, urn)
+            get_resource(self, props, custom, urn)
         elif opts.id is not None:
             # If this is a custom resource that already exists, read its state from the provider.
             if not custom:
--- a/sdk/python/lib/pulumi/resource.py
+++ b/sdk/python/lib/pulumi/resource.py
@@ -21,7 +21,7 @@ from typing import Optional, List, Any, Mapping, Union, Callable, TYPE_CHECKING,
 import copy
 
 from .runtime import known_types
-from .runtime.resource import register_resource, register_resource_outputs, read_resource
+from .runtime.resource import get_resource, register_resource, register_resource_outputs, read_resource
 from .runtime.settings import get_root_resource
 
 from .metadata import get_project, get_stack
@@ -608,7 +608,8 @@ class Resource:
                  props: Optional['Inputs'] = None,
                  opts: Optional[ResourceOptions] = None,
                  remote: bool = False,
-                 dependency: bool = False) -> None:
+                 dependency: bool = False,
+                 urn: Optional[str] = None) -> None:
         """
         :param str t: The type of this resource.
         :param str name: The name of this resource.
@@ -720,7 +721,10 @@ class Resource:
                 self._aliases.append(collapse_alias_to_urn(
                     alias, name, t, opts.parent))
 
-        if opts.id is not None:
+        if urn is not None:
+            # This is a resource that already exists. Read its state from the engine.
+            get_resource(self, custom, urn)
+        elif opts.id is not None:
             # If this is a custom resource that already exists, read its state from the provider.
             if not custom:
                 raise Exception(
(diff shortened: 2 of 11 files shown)


lblackstone (approved):

sdk/python/lib/pulumi/runtime/resource.py
```diff
@@ -138,8 +139,62 @@ async def prepare_resource(res: 'Resource',
...
+    # Create resolved outputs for each property on res.
+    for key, value in result.value["state"].items():
+        # Skip well-known properties.
+        if key == "urn" or key == "id":
+            continue
+
+        is_known = not rpc.contains_unknowns(value)
+        is_secret = rpc.is_rpc_secret(value)
+
+        res.__dict__[key] = resolved_resource_output(res, rpc.unwrap_rpc_secret(value), is_known, is_secret)
```
justinvp, line 194: Don't we need to translate the key into the Python name and also specify the intended type? I think we need to do something like:

https://github.com/pulumi/pulumi/blob/f22b47e54f39312b9b13ea359d3d523ee42e7a82/sdk/python/lib/pulumi/runtime/rpc.py#L571-L578

sdk/python/lib/pulumi/runtime/resource.py
```diff
@@ -138,8 +139,62 @@ async def prepare_resource(res: 'Resource',
...
+
+    # If this is a custom resource, resolve the ID.
+    if custom:
+        id = result.value["id"]
+        id_known = bool(id)
+
+        res.__dict__["id"] = resolved_resource_output(res, id, id_known, False)
+
+    # Create resolved outputs for each property on res.
+    for key, value in result.value["state"].items():
```
justinvp, line 186: Do we also need to handle the case where there may be optional outputs that aren't present in the state, but do exist as properties on the resource?

https://github.com/pulumi/pulumi/blob/f22b47e54f39312b9b13ea359d3d523ee42e7a82/sdk/python/lib/pulumi/runtime/rpc.py#L633-L638

sdk/python/lib/pulumi/runtime/resource.py
pgavlin: I think that's already been done by the deserializer, hasn't it?

sdk/python/lib/pulumi/runtime/resource.py
pgavlin: Hmmm, we do. That's a very good point.

sdk/python/lib/pulumi/runtime/resource.py
justinvp: No, I don't think so. When I implemented support for dataclasses, it was much simpler to implement the type conversion inside `translate_output_properties` rather than inside `deserialize_properties` (where you'd expect that code to live). So unfortunately, `translate_output_properties` is currently doing two things: 1. Translating key names (as it did originally), and 2. Translating the value to the target type, if specified. (Kinda gross, I know).

You can see inside the `invoke` call, all it's doing this:

https://github.com/pulumi/pulumi/blob/f22b47e54f39312b9b13ea359d3d523ee42e7a82/sdk/python/lib/pulumi/runtime/invoke.py#L125-L130

The `typ` parameter is an optional parameter to `invoke`. Our codegen for functions specifies `typ` when calling `invoke`, which takes care of translating the keys and instantiating the result class.

For example, in `pulumi_aws`, `get_bucket` does [this](https://github.com/pulumi/pulumi-aws/blob/2825d208fcf76bc890c5c2f5bb45921f896dc338/sdk/python/pulumi_aws/s3/get_bucket.py#L186):

```python
__ret__ = pulumi.runtime.invoke('aws:s3/getBucket:getBucket', __args__, opts=opts, typ=GetBucketResult).
```

`GetBucketResult` is decorated with `@pulumi.output_type`, so `translate_output_properties` will end up instantiating it and translating the keys based on the decorators on the class's properties.

In this particular case, for `getResource`, passing a `typ` to this invoke isn't going to work, because the target type is a resource, not an output type. So instead, I think we're going to need to do something along the lines of that code snippet in my earlier comment.

Basically call `types = _types.resource_types(type(res)) ` to get a dict of types for the resource, and then loop over the items in `state` translating the key with `res.translate_output_property(key)` and value with `translate_output_properties(value, res.translate_output_property, types.get(key)) `.

pgavlin: Ah, no, you're right--the `invoke` deserializer won't do the translation. I may need to resort to a lower-level invoke s.t. I can call the translator.

justinvp, line 194: I think something like this would work (untested):

```suggestion
    # Create resolved outputs for each property on res.
    types = _types.resource_types(type(res))
    for key, value in result.value["state"].items():
        # Skip well-known properties.
        if key == "urn" or key == "id":
            continue
            
        # Outputs coming from the provider are NOT translated. Do so here. 
        translated_key = res.translate_output_property(key) 
        translated_value = translate_output_properties(value, res.translate_output_property, types.get(key))

        is_known = not rpc.contains_unknowns(translated_value)
        is_secret = rpc.is_rpc_secret(translated_value)

        res.__dict__[translated_key] = resolved_resource_output(res,  rpc.unwrap_rpc_secret(translated_value), is_known, is_secret)
```

pgavlin:
@justinvp I reimplemented this on top of the usual resource functions; PTAL

sdk/python/lib/pulumi/resource.py
```diff
@@ -608,7 +608,8 @@ def __init__(self,
                  props: Optional['Inputs'] = None,
                  opts: Optional[ResourceOptions] = None,
                  remote: bool = False,
-                 dependency: bool = False) -> None:
+                 dependency: bool = False,
+                 urn: Optional[str] = None) -> None:
```
pgavlin, line 612: I think I'll need to move this into `ResourceOptions`.

pgavlin: (otherwise we'll need to add another parameter to all generated constructors)

justinvp (approved):
LGTM

sdk/python/lib/pulumi/runtime/resource.py
```diff
@@ -138,8 +139,109 @@ async def prepare_resource(res: 'Resource',
...
+            secret_future.set_exception(exn)
+        else:
+            value_future.set_result(value)
+            known_future.set_result(known)
+            secret_future.set_result(secret)
+
+    return (resolve, Output({res}, value_future, known_future, secret_future))
+
+
+def resolved_resource_output(res: 'Resource', value: Any, known: bool, secret: bool) -> 'Output':
```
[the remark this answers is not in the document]
pgavlin: Not anymore. I'll remove it.

pkg/resource/deploy/builtins.go
```diff
@@ -272,6 +272,10 @@ func (p *builtinProvider) getResource(inputs resource.PropertyMap) (resource.Pro
 	contract.Assert(ok)
 	contract.Assert(urn.IsString())
 
+	// TODO(pdg): track secret outputs in the resource map. This is necessary to ensure that the
+	// `additionalSecretOutputs` option that was provided when the resource was registered is properly respected by
+	// `getResource`.
```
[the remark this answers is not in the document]
pgavlin: Will do.

PR #5711, opened 2020-11-07
[codegen/dotnet,nodejs] - Improve makeSafeEnumName

### LINKED ISSUE #5706: [codegen/dotnet] Special cases for .NET enum values
I gave the .NET enum generation a run on NextGen. Overall, looks great.

However, it exposed two special cases that don't compile as-is:

1. If an enum value is `Equals`, that conflicts with the `Equals()` method.

```cs
public static Operator Equals { get; } = new Operator("Equals");
public static Operator NotEquals { get; } = new Operator("NotEquals");
public static Operator Contains { get; } = new Operator("Contains");
public static Operator DoesNotContain { get; } = new Operator("DoesNotContain");
...
[EditorBrowsable(EditorBrowsableState.Never)]
public override bool Equals(object? obj) => obj is Operator other && Equals(other);
public bool Equals(Operator other) => string.Equals(_value, other._value, StringComparison.Ordinal);
```

Suggestion: rename `Equals` to `EqualsValue`?

2. Number values aren't handled properly because C# identifiers can't start with a number:

```cs
public static AlertSeverity _ { get; } = new AlertSeverity("0");
```

Suggestion: rename to `Value0`?

komalali:
Fixes #5706

### DIFF
--- a/pkg/codegen/dotnet/utilities.go
+++ b/pkg/codegen/dotnet/utilities.go
@@ -15,6 +15,8 @@
 package dotnet
 
 import (
+	"fmt"
+	"regexp"
 	"strings"
 	"unicode"
 )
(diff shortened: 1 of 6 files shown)


pkg/codegen/dotnet/utilities.go
```diff
@@ -78,3 +80,45 @@ func makeValidIdentifier(name string) string {
...
+			"8", "Eight",
+			"9", "Nine",
+			"*", "Asterisk",
+		)
+
+		safeName = enumReplacer.Replace(safeName)
+
+		// If it's still an illegal character (we weren't able to find a replacement), return an error.
+		if !isLegalIdentifierStart(rune(safeName[0])) {
+			return "", fmt.Errorf("enum name %s is not a valid identifier", safeName)
```
justinvp, line 107: Nit: Prefer `errors.Errorf`

pkg/codegen/dotnet/utilities.go
```diff
@@ -78,3 +80,45 @@ func makeValidIdentifier(name string) string {
...
+	// Capitalize and make a valid identifier.
+	safeName = strings.Title(makeValidIdentifier(safeName))
+
+	// If there are multiple underscores in a row, replace with one.
+	regex := regexp.MustCompile(`_+`)
+	safeName = regex.ReplaceAllString(safeName, "_")
+
+	// "Equals" conflicts with a method on the EnumType struct, change it to EqualsValue.
+	if safeName == "Equals" {
+		safeName = fmt.Sprintf("%sValue", safeName)
```
justinvp, line 120: Nit: Any reason not to do this?

```suggestion
		safeName = "EqualsValue"
```

pkg/codegen/nodejs/utilities.go
```diff
@@ -98,3 +100,40 @@ func makeValidIdentifier(name string) string {
...
+			"8", "Eight",
+			"9", "Nine",
+			"*", "Asterisk",
+		)
+
+		safeName = enumReplacer.Replace(safeName)
+
+		// If it's still an illegal character (we weren't able to find a replacement), return an error.
+		if !isLegalIdentifierStart(rune(safeName[0])) {
+			return "", fmt.Errorf("enum name %s is not a valid identifier", safeName)
```
justinvp, line 127: Nit: Prefer `errors.Errorf`

pkg/codegen/dotnet/utilities.go
```diff
@@ -78,3 +80,45 @@ func makeValidIdentifier(name string) string {
 func propertyName(name string) string {
 	return makeValidIdentifier(Title(name))
 }
+
+func makeSafeEnumName(name string) (string, error) {
+	safeName := name
+
+	// If the name is one illegal character, replace it.
+	if len(safeName) == 1 && !isLegalIdentifierStart(rune(safeName[0])) {
+		enumReplacer := strings.NewReplacer(
```
justinvp, line 89: Nit: This could also be implemented using a `map[string]string` (or even `map[rune]string`) lookup, but what you have is fine with me.

justinvp (approved):
LGTM

pkg/codegen/dotnet/utilities.go
komalali: I left it flexible in case there are other values like "Equals" that would cause issues but no reason not to simplify since there aren't at the moment.

PR #5718, opened 2020-11-09
Ensure multi-line deprecation warnings are escaped correctly in Python

stack72:
Fixes: https://github.com/pulumi/pulumi-vsphere/issues/143


justinvp (approved):
LGTM

PR #5721, opened 2020-11-09
[codegen/*] - Extract common enum name replacer

komalali:
We will use these replacements for all languages, so extracting them into a common function.

### DIFF
--- a/pkg/codegen/dotnet/gen_test.go
+++ b/pkg/codegen/dotnet/gen_test.go
@@ -65,6 +65,7 @@ func TestMakeSafeEnumName(t *testing.T) {
 		{"SystemAssigned, UserAssigned", "SystemAssigned_UserAssigned", false},
 		{"Dev(NoSLA)_Standard_D11_v2", "Dev_NoSLA_Standard_D11_v2", false},
 		{"Standard_E8as_v4+1TB_PS", "Standard_E8as_v4_1TB_PS", false},
+		{"Equals", "EqualsValue", false},
 	}
 	for _, tt := range tests {
 		t.Run(tt.input, func(t *testing.T) {
--- a/pkg/codegen/nodejs/utilities.go
+++ b/pkg/codegen/nodejs/utilities.go
@@ -104,15 +104,12 @@ func makeValidIdentifier(name string) string {
 }
 
 func makeSafeEnumName(name string) (string, error) {
-	var err error
-	safeName := name
+	// Replace common single character enum names.
+	safeName := codegen.ExpandShortEnumName(name)
 
-	// If the name is one illegal character, replace it.
+	// If the name is one illegal character, return an error.
 	if len(safeName) == 1 && !isLegalIdentifierStart(rune(safeName[0])) {
-		safeName, err = codegen.ReplaceInvalidEnumName(safeName)
-		if err != nil {
-			return "", errors.Errorf("enum name %s is not a valid identifier", safeName)
-		}
+		return "", errors.Errorf("enum name %s is not a valid identifier", safeName)
 	}
 
 	// Capitalize and make a valid identifier.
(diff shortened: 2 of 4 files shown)


pkg/codegen/utilities.go
```diff
@@ -106,3 +107,25 @@ func CleanDir(dirPath string, exclusions StringSet) error {
...
+	"3": "Three",
+	"4": "Four",
+	"5": "Five",
+	"6": "Six",
+	"7": "Seven",
+	"8": "Eight",
+	"9": "Nine",
+}
+
+func ReplaceInvalidEnumName(name string) (string, error) {
```
mikhailshilkov, line 125: It could be me, but this function feels slightly odd on its own. Would it be possible to extract the full `makeSafeEnumName` function, which seems more self-sufficient?

pkg/codegen/dotnet/gen_test.go
```diff
@@ -65,6 +65,7 @@ func TestMakeSafeEnumName(t *testing.T) {
 		{"SystemAssigned, UserAssigned", "SystemAssigned_UserAssigned", false},
 		{"Dev(NoSLA)_Standard_D11_v2", "Dev_NoSLA_Standard_D11_v2", false},
 		{"Standard_E8as_v4+1TB_PS", "Standard_E8as_v4_1TB_PS", false},
+		{"Equals", "EqualsValue", false},
```
mikhailshilkov, line 68: Is this test case unrelated to the change? (no need to remove it, just for my understanding)

pkg/codegen/utilities.go
komalali: That's not quite as simple. Each language has its own idea of what is a valid identifier. And each language has its own representation of what an enum name should look like (like all caps for python, `Equals` having special meaning in C#, etc.). I tried to extract what was reasonably extractable.

pkg/codegen/dotnet/gen_test.go
komalali: Yeah just making sure `Equals` works as expected.

pkg/codegen/utilities.go
mikhailshilkov: Yes, I was too optimistic. The following would be slightly more apealing to me but it's up to your judgment:
1. Call it something like `ExpandShorEnumName`.
2. Never return an error, just the original string or the expanded string.
3. Call it from `makeSafeEnumName` before the `if`. 
4. `if len(safeName) == 1 && !isLegalIdentifierStart(rune(safeName[0]))` always returns an error if `true`.

pkg/codegen/nodejs/utilities.go
```diff
@@ -103,28 +104,13 @@ func makeValidIdentifier(name string) string {
...
-			"*", "Asterisk",
-		)
-
-		safeName = enumReplacer.Replace(safeName)
-
-		// If it's still an illegal character (we weren't able to find a replacement), return an error.
-		if !isLegalIdentifierStart(rune(safeName[0])) {
+		safeName, err = codegen.ReplaceInvalidEnumName(safeName)
+		if err != nil {
 			return "", errors.Errorf("enum name %s is not a valid identifier", safeName)
```
mikhailshilkov, line 114: Are we losing the `err` details here (which symbol couldn't be replaced)?

mikhailshilkov (approved):
LGTM with two nits in the comments

pkg/codegen/utilities.go
komalali: I like that, will change it up.

pkg/codegen/utilities.go
```diff
@@ -106,3 +106,25 @@ func CleanDir(dirPath string, exclusions StringSet) error {
...
+	"7": "Seven",
+	"8": "Eight",
+	"9": "Nine",
+}
+
+func ExpandShortEnumName(name string) string {
+	replacement, ok := commonEnumNameReplacements[name]
+	if ok {
+		return replacement
+	}
```
justinvp, line 128: Nit:

```suggestion
	if replacement, ok := commonEnumNameReplacements[name]; ok {
		return replacement
	}
```

PR #5723, opened 2020-11-10
Enable creation of TStack using IServiceProvider

nathanwinder-perficient:
Enabled use of `IServiceProvider` for TStack instance creation. This supports the use of dependency injection when creating a stack.

- Added an overload for RunAsync that accepts an `IServiceProvider`. 
- Updated existing RunAsync methods to use the factory pattern to create instance of TStack
- Added a new method `TestWithServiceProviderAsync` for testing a stack that is generated using and `IServiceProvider`. I would like to use an overload of `TestAsync` but the backcompat rules would not allow it.


mikhailshilkov:
Hi @nathanwinder-perficient thanks for this PR!

One thing that we'd need to add is a test program similar to [this one](https://github.com/pulumi/pulumi/tree/master/tests/integration/stack_component/dotnet) but with the new capability plus a corresponding Go test like [here](https://github.com/pulumi/pulumi/blob/a6f8b9b205cdf975421520030a713b9ea489f625/tests/integration/integration_dotnet_test.go#L50)

nathanwinder-perficient:
@mikhailshilkov I added the integration test. I am using Windows and am not able to test that locally. I see that the test suite is failing and was failing before I added my test case. I tried to confirm that my test was passing but was unable to find it in the output of the test run.

mikhailshilkov (approved):
Sorry it took me so long but I reviewed the code, made some tweaks, ran the tests in #5836
LGTM

Issue #4868, opened 2020-06-22
Create an ARM based build for the pulumi cli (and plugins)


jaxxstorm:
The prerequisite work is to get our projects to go1.16 so that macOS arm builds are supported

The work required here is as follows:

- [x] Update tf providers with arm builds
- [x] Update azure-native with arm build
- [x] Update pulumi-kubernetes with arm build
- [x] Update pulumi/pulumi with arm build
- [x] Update pulumi/awsx with arm build
- [x] Update pulumi/eks with arm build
- [x] Update plugin acquisition to detect arm in the downloads
- [x] Update get.pulumi.com to allow arm downloads of the CLI
- [x] Add some nightly tests to ensure that we can install and use the arm builds

justinvp:
https://www.apple.com/newsroom/2020/06/apple-announces-mac-transition-to-apple-silicon/

lukehoban:
Just a note - we will want ARM builds across all supported OS’s. Linux builds will also be valuable to support using Pulumi on cloud-based ARM servers such as AWS Graviton.

lukehoban:
Also - this will require not just CLI, but also all plugins (both resource provider plugins but also language host plugins).

jaxxstorm:
related: #5244

lukehoban:
Per #5466 - this should also include ARM Docker images.

justinvp:
ARM-based MacBook Pro 13", MacBook Air, and Mac mini available starting 11/17: https://www.apple.com/apple-events/november-2020/

jaxxstorm:
I had a try at building the ARM binaries yesterday, the PR is here:

https://github.com/pulumi/pulumi/pull/5729

The binary worked without issue. I also built an AWS provider ARM binary which worked without any issues either.

I tried out a simple Pulumi program on an AWS graviton machine. It worked successfully with the Go SDK

What I ran into was:

- the language SDKs need to have GRPC binaries available
  - Python didn't work (even though it's allegedly meant to be available, pip install failed)
  - Go did work
- plugin acquisition doesn't recognise the machine's chipset yet, so plugin acquisition failed on Python

lukehoban:
Great! 

> plugin acquisition doesn't recognise the machine's chipset yet, so plugin acquisition failed on Python

What exactly was the error?  Was this due to https://github.com/pulumi/pulumi/blob/186e2f059ec633c34eb414b3a9391a850cda2941/sdk/go/common/workspace/plugins.go#L190?

[...]

PR #5729, opened 2020-11-10
try an arm64 build

jaxxstorm:
Related: #4868

### DIFF
--- a/.goreleaser.prerelease.yml
+++ b/.goreleaser.prerelease.yml
@@ -24,6 +24,7 @@ builds:
   dir: pkg
   goarch:
     - amd64
+    - arm64
   goos:
     - linux
     - darwin
@@ -35,6 +36,7 @@ builds:
   dir: sdk
   goarch:
     - amd64
+    - arm64
   goos:
     - linux
     - darwin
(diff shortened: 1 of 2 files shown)


lukehoban (commented):
I suspect we can also fixup plugin publishing in https://github.com/pulumi/pulumi/blob/c8b7e1fc715557f1016cf4b290caf61e979f8d5a/scripts/publish_tgz.sh#L16 and plugin download in https://github.com/pulumi/pulumi/blob/186e2f059ec633c34eb414b3a9391a850cda2941/sdk/go/common/workspace/plugins.go?

stack72:
We are going to publish these binaries so that we can start to develop the mechanism around get.pulumi.com and plugin acquisition

stack72:
> I suspect we can also fixup plugin publishing in
> 
> https://github.com/pulumi/pulumi/blob/c8b7e1fc715557f1016cf4b290caf61e979f8d5a/scripts/publish_tgz.sh#L16
> 
> and plugin download in
> https://github.com/pulumi/pulumi/blob/186e2f059ec633c34eb414b3a9391a850cda2941/sdk/go/common/workspace/plugins.go#L190

We are going to publish the package so we can test this out and ensure that it's all ok as we are fixing up plugin acquisition

PR #5730, opened 2020-11-11
Create a builtin provider during import.

pgavlin:
`pulumi import` may otherwise fail if there are providers in the
statefile.

### DIFF
--- a/pkg/resource/deploy/import.go
+++ b/pkg/resource/deploy/import.go
@@ -73,8 +73,10 @@ func NewImportPlan(ctx *plugin.Context, target *Target, projectName tokens.Packa
 		return nil, err
 	}
 
+	builtins := newBuiltinProvider(nil, nil)
+
 	// Create a new provider registry.
-	reg, err := providers.NewRegistry(ctx.Host, oldResources, preview, nil)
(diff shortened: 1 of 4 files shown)


pkg/resource/deploy/import.go
```diff
@@ -73,8 +73,10 @@ func NewImportPlan(ctx *plugin.Context, target *Target, projectName tokens.Packa
 		return nil, err
 	}
 
+	builtins := newBuiltinProvider(nil, nil)
```
lukehoban, line 76: Is a `nil` backendClient legal?  Are we sure that can't trigger panics in any cases `import` might hit?

lukehoban (approved):
LGTM with one question.

Would be great to add tests that cover the 1st class providers scenario - but possibly as a follow-up if a quick fix is needed here.

pkg/resource/deploy/import.go
pgavlin: I'm not 100% certain about that, no. I'm exploring the downstream code now, but I wanted to get this start out ASAP.

pgavlin: I've verified that all uses of the backend client are nil-checked, so this won't hit any panics. Working on tests now.

PR #5742, opened 2020-11-12
Respect provider config secretness.

pgavlin:
Just what it says on the tin. This is implemented by changing the
`GetPackageConfig` method of `ConfigSource` to return a `PropertyMap`
and ensuring that any secret config is represented by a `Secret`.


pkg/resource/deploy/import.go
```diff
@@ -219,16 +219,12 @@ func (i *importer) registerProviders(ctx context.Context) (map[resource.URN]stri
...
 
 		// Fetch, prepare, and check the configuration for this provider.
-		cfg, err := i.plan.target.GetPackageConfig(req.Package())
+		inputs, err := i.plan.target.GetPackageConfig(req.Package())
 		if err != nil {
 			return nil, result.Errorf("failed to fetch provider config: %v", err), false
 		}
 
 		// Calculate the inputs for the provider using the ambient config.
-		inputs := resource.PropertyMap{}
```
stack72, line 228: oh great find! this is ace - thank you!!

stack72 (approved):

EvanBoyle (approved):
Looks good. Does this cause a provider diff if an existing stack has providers that fall into this path?

pgavlin:
> Does this cause a provider diff if an existing stack has providers that fall into this path?

It shouldn't, no. Any diffs should be independent of secretness.

PR #5744, opened 2020-11-12
Parse string arrays from env in Go first-class provider

mikhailshilkov:
Fix generation for https://github.com/pulumi/pulumi-azure-nextgen/issues/126

Other languages to be tracked in https://github.com/pulumi/pulumi/issues/5743

### DIFF
--- a/pkg/codegen/go/gen.go
+++ b/pkg/codegen/go/gen.go
@@ -682,7 +682,7 @@ func (pkg *pkgContext) getDefaultValue(dv *schema.DefaultValue, t schema.Type) (
 		parser, typDefault, typ := "nil", "\"\"", "string"
 		switch t.(type) {
 		case *schema.ArrayType:
-			parser, typDefault, typ = "parseEnvStringArray", "[]pulumi.StringInput{}", "[]pulumi.StringInput"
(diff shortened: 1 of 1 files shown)


pkg/codegen/go/gen.go
```diff
@@ -1477,6 +1482,14 @@ func parseEnvFloat(v string) interface{} {
 	return f
 }
 
+func parseEnvStringArray(v string) interface{} {
+	var result []pulumi.StringInput
+	for _, item := range strings.Split(v, ";") {
+		result = append(result, pulumi.StringInput(pulumi.String(item)))
```
EvanBoyle, line 1488: `pulumi.String` implements `pulumi.StringInput`: https://github.com/pulumi/pulumi/blob/00f8966a9a5e8b693cf41c1e78457956a8b8ec33/sdk/go/pulumi/types_builtins.go#L6515

You should be able to simplify all of this:

```go
var result pulumi.StringArray
...
result = append(result, pulumi.String(item))
```

Several different incantations of this will work FWIW.

EvanBoyle (approved):

pkg/codegen/go/gen.go
mikhailshilkov: That's a good one. I tried a couple of combos but I didn't try returning a `StringArray` directly.

PR #5751, opened 2020-11-12
Reduce verbosity on tests

### LINKED ISSUE #5727: Reduce verbosity of pulumi/pulumi test runs

viveklak:
Fixes #5727.

### DIFF
--- a/pkg/testing/integration/program.go
+++ b/pkg/testing/integration/program.go
@@ -619,16 +619,6 @@ func ProgramTestManualLifeCycle(t *testing.T, opts *ProgramTestOptions) *Program
 	return pt
 }
 
-// fprintf works like fmt.FPrintf, except it explicitly drops the return values. This keeps the linters happy, since
-// they don't like to see errors dropped on the floor. It is possible that our call to fmt.Fprintf will fail, even
-// for "standard" streams like `stdout` and `stderr`, if they have been set to non-blocking by an external process.
-// In that case, we just drop the error on the floor and continue. We see this behavior in Travis when we try to write
-// a lot of messages quickly (as we do when logging test failures)
-func fprintf(w io.Writer, format string, a ...interface{}) {
-	_, err := fmt.Fprintf(w, format, a...)
-	contract.IgnoreError(err)
-}
-
 // ProgramTester contains state associated with running a single test pass.
 type ProgramTester struct {
 	t            *testing.T          // the Go tester for this run.
@@ -880,8 +870,8 @@ func (pt *ProgramTester) runVirtualEnvCommand(name string, args []string, wd str
 		}
 
 		if pt.opts.Verbose {
-			fprintf(pt.opts.Stdout, "acquired pip install lock\n")
-			defer fprintf(pt.opts.Stdout, "released pip install lock\n")
+			pt.t.Log("acquired pip install lock")
+			defer pt.t.Log("released pip install lock")
 		}
 		defer func() {
 			if err := pipMutex.Unlock(); err != nil {
@@ -931,8 +921,8 @@ func (pt *ProgramTester) runPipenvCommand(name string, args []string, wd string)
 		}
 
 		if pt.opts.Verbose {
-			fprintf(pt.opts.Stdout, "acquired pip install lock\n")
-			defer fprintf(pt.opts.Stdout, "released pip install lock\n")
+			pt.t.Log("acquired pip install lock")
+			defer pt.t.Log("released pip install lock")
 		}
 		defer func() {
 			if err := pipMutex.Unlock(); err != nil {
@@ -1062,7 +1052,7 @@ func (pt *ProgramTester) TestLifeCycleInitialize() error {
 	}
 
 	// Ensure all links are present, the stack is created, and all configs are applied.
-	fprintf(pt.opts.Stdout, "Initializing project (dir %s; stack %s)\n", dir, stackName)
+	pt.t.Logf("Initializing project (dir %s; stack %s)", dir, stackName)
 
 	// Login as needed.
 	stackInitName := string(pt.opts.GetStackNameWithOwner())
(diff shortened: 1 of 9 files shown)


komalali:
@viveklak this is looking so much better already! If we could also get the stacktraces to not dump in the window for cases where we *want* the error, like [this](https://github.com/pulumi/pulumi/pull/5751/checks?check_run_id=1393170980#step:15:1897) or [this](https://github.com/pulumi/pulumi/pull/5751/checks?check_run_id=1393170980#step:15:668) that would really drive it home. Those are coming from the lang-specific tests for [python](https://github.com/pulumi/pulumi/tree/master/sdk/python/lib/test/langhost) and [nodejs](https://github.com/pulumi/pulumi/tree/master/sdk/nodejs/tests)

komalali:
I think unsetting `-v` in the [python makefile](https://github.com/pulumi/pulumi/blob/master/sdk/python/Makefile#L43-L44) and doing something (unclear what exactly) in the [nodejs makefile](https://github.com/pulumi/pulumi/blob/master/sdk/nodejs/Makefile) would help.

viveklak:
> I think unsetting `-v` in the [python makefile](https://github.com/pulumi/pulumi/blob/master/sdk/python/Makefile#L43-L44) and doing something (unclear what exactly) in the [nodejs makefile](https://github.com/pulumi/pulumi/blob/master/sdk/nodejs/Makefile) would help.

Thanks I will give it a shot!

komalali:
Did a little comparison, a regular test run for the ubuntu build & test is ~5300 lines of output. With just the changes you've made so far it's down to ~2300 lines. 🎉

viveklak:
> > I think unsetting `-v` in the [python makefile](https://github.com/pulumi/pulumi/blob/master/sdk/python/Makefile#L43-L44) and doing something (unclear what exactly) in the [nodejs makefile](https://github.com/pulumi/pulumi/blob/master/sdk/nodejs/Makefile) would help.

I was hoping to make the mocha nodejs tests disable console logging (unless failed) but looks like this is a longstanding limitation: https://github.com/mochajs/mocha/issues/2107. The approaches proposed as workarounds seem less than ideal in that they exclude all the relevant mocha output too which sucks. We might have to live with nodejs test output...

viveklak:
> This is looking great! Are you planning on addressing this comment from #5751?
> 
> > We should also consider dropping [the prefix we attach to test output](https://github.com/pulumi/pulumi/blob/master/pkg/testing/integration/program.go#L1516-L1533). That prefix is only really useful if all of the tests are simultaneously streaming output

Yes - done in https://github.com/pulumi/pulumi/pull/5751/commits/fb0d21030cfb15f5e771932693b282d8675f7597

viveklak:
> > This is looking great! Are you planning on addressing this comment from #5751?
> > > We should also consider dropping [the prefix we attach to test output](https://github.com/pulumi/pulumi/blob/master/pkg/testing/integration/program.go#L1516-L1533). That prefix is only really useful if all of the tests are simultaneously streaming output
> 
> Yes - done in [fb0d210](https://github.com/pulumi/pulumi/commit/fb0d21030cfb15f5e771932693b282d8675f7597)

Looks like I need to clean out some sprayed references to the embedded stdout/stderr. I think I can mostly replace them `t.Logf` calls instead. Will update when those changes are good to review. The rest should be straightforward to review in the meantime.

pgavlin (approved):

viveklak:
Sorry for the noise but I think I got this to a state I am comfortable with. @pgavlin - quick once over before I merge?

pkg/testing/integration/program.go
```diff
@@ -931,8 +921,8 @@ func (pt *ProgramTester) runPipenvCommand(name string, args []string, wd string)
 		}
 
 		if pt.opts.Verbose {
-			fprintf(pt.opts.Stdout, "acquired pip install lock\n")
-			defer fprintf(pt.opts.Stdout, "released pip install lock\n")
+			pt.t.Log("acquired pip install lock")
+			defer pt.t.Log("released pip install lock")
```
komalali, line 925: Just checking, we're dropping the newlines because `t.Log` already includes a newline?

sdk/python/Makefile
```diff
@@ -40,8 +40,11 @@ install:: install_package install_plugin
 
 test_fast:: build
 	pipenv run pip install ./env/src
-	pipenv run python -m unittest discover -s lib/test -v
-	pipenv run python -m unittest discover -s lib/test_with_mocks -v
+	# TODO the ignored test seems to fail in pytest but not unittest. Need to trackdown why
+	pipenv run pytest lib/test --ignore lib/test/langhost/resource_thens/test_resource_thens.py
+	pipenv run python -m unittest lib/test/langhost/resource_thens/test_resource_thens.py
+	# Using python -m also adds lib/test_with_mocks to sys.path which avoids issues package resolution issues.
```
komalali, line 46:
```suggestion
	# Using python -m also adds lib/test_with_mocks to sys.path which avoids package resolution issues.
```

komalali (approved):

viveklak:
@justinvp a bunch of the output now seem to be deprecation notices related to some python protobuf calls in the sdk. Do you know if there is something tracking that?

pkg/testing/integration/program.go
viveklak: Correct - logf will handle the newlines

pgavlin (approved):
This looks awesome.

justinvp:
> @justinvp a bunch of the output now seem to be deprecation notices related to some python protobuf calls in the sdk. Do you know if there is something tracking that?

I don't believe this is being tracked anywhere. Can you file an issue? It's likely we just need to rerun `sdk/proto/generate.sh` after more recent updates of protobuf/grpcio for Python 3.9.

Issue #5218, opened 2020-08-25
Add structured output to Stack.Up


EvanBoyle:
Right now `Stack.Up` returns limited output (stdout, stderr, Outputs, Summary). This is mainly due to the fact that `pulumi up` doesn't support a `--json` flag. We should add a `--json` flag to pulumi up, and then consume this from the Automation API. 

Ideally, we should be able to consume this output in both sync and streaming contexts.

lukehoban:
Shared one approach we could take to this in #5754 which builds on the `--event-log` flag that already exists on the CLI.

EvanBoyle:
I had imagined streaming JSON events incrementally as many of the `go` tools do. Good find, `--event-log` offers a code path to follow and get things started.

komalali:
Closing command-based issues for language-based issues 
* #6441 
* #6442 
* #6443 
* #6444

PR #5754, opened 2020-11-13
Expose structured event stream to JavaScript Automation API

lukehoban:
Use the existing `--event-log` flag to write the event log and expose this to an event handler on update/destroy/refresh operations.

The approach of writing to a file and tailing that from automation API feels like it is quite likely too fragile, and we may need to take another approach to make this robust - both to how we expose the event log from the CLI, and to how we consume that from the Automation API.  Marking as Draft for now as I'm not yet comfortable with this approach as is. @pgavlin @EvanBoyle curious if you have thoughts on a more robust way to support this in the near term.  An alternative would be to add the `--json` flag to `up`/`destroy`/`refresh` and make that takeover `stdout` - but to drive that outputs from the same event logging infrastructure that the engine/CLI implements.

Also need to add Go support and tests before merging either way.

The JavaScript part of #5218.

### DIFF
--- a/pkg/cmd/pulumi/destroy.go
+++ b/pkg/cmd/pulumi/destroy.go
@@ -228,9 +228,12 @@ func newDestroyCmd() *cobra.Command {
 	cmd.PersistentFlags().BoolVarP(
 		&yes, "yes", "y", false,
 		"Automatically approve and perform the destroy after previewing it")
-	cmd.PersistentFlags().StringVar(
-		&eventLogPath, "event-log", "",
-		"Log events to a file at this path")
+
+	if hasDebugCommands() {
+		cmd.PersistentFlags().StringVar(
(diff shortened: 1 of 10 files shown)


pgavlin:
> An alternative would be to add the --json flag to up/destroy/refresh and make that takeover stdout - but to drive that outputs from the same event logging infrastructure that the engine/CLI implements.

This would be my strong preference, even if it requires a little bit more work in the near term. As I see it, the only piece of work that this requires that is not necessarily well understood is defining the compatibility promise for engine events and documenting their structure. Though even that we could punt a bit if we stuck `--json` under a feature flag; the aspects of the `--json` approach I find most appealing are a) a cleaner architecture and b) the lack of a temporary file.

komalali:
I'm going to close this PR and re-open once I've worked out the kinks.

Issue #2695, opened 2019-05-02
Installing a version of plugin sometimes fails on Windows


mikhailshilkov:
```
pulumi plugin install resource azure v0.18.2-dev.1556768307+g78c053
[resource plugin azure-0.18.2-dev.1556768307+g78c0537] installing
Downloading plugin:  36.55 MiB / 36.55 MiB [=======================] 100.00% 22s
error: installing [resource plugin azure-0.18.2-dev.1556768307+g78c0537] from https://api.pulumi.com:
moving plugin: rename C:\Users\name\.pulumi\plugins\resource-azure-v0.18.2-dev.1556768307+g78c0537.tmp804464559 C:\Users\name\.pulumi\plugins\resource-azure-v0.18.2-dev.1556768307+g78c0537: 
Access is denied.
```

The temporary folder does get created and the manual rename does work, so I suspect some API which fails for this kind of path?

## Edit: Summary of the underlying issue and proposed solution

This exact same problem was happening with golang's module cache on Windows: module zips were downloaded/extracted to a temp dir and then renamed with `os.Rename` to the final dir, just like how we handle plugin installs (https://github.com/golang/go/issues/36568). The problem on Windows is that aggressive file scanners are checking the files in the temp dir, preventing the rename operation from succeeding (errors with "Access is denied.").

The golang team's solution was to extract the zip in-place in the final destination, rather than extracting to a temp dir and renaming it. Note that they were leveraging some file locking mechanism they already had in place, and we'll likely need to do something similar to prevent multiple concurrent installs from cobbling over each other.

A simpler initial "fix" may be just to introduce some retries. If I'm remembering correctly, the golang team initially tried this, but still ran into the problem. However, the amount of plugin files is typically very small (~1-2 files), so we may have better luck with retries since we're not dealing with that many files. Though, as more multi-lang components are introduced, the number of files in plugin directories will also increase as those will be running `npm install` (and equivalents for the other languages).

Reference:
https://github.com/golang/go/commit/093049b3709eda7537ece92a2991918cf53782d6
https://github.com/golang/go/commit/576fa692774137633b09dd244e1de36993dd2803

lukehoban:
Does this only happen for dev versions?  That is particularly odd.

mikhailshilkov:
This week is the first time I saw this issue, and I already got it twice for two dev versions...

mikhailshilkov:
[...]

PR #5759, opened 2020-11-13
Fix plugin install failures on Windows

justinvp:
When installing a plugin, previous versions of Pulumi extracted the plugin tarball to a temp directory and then renamed the temp directory to the final plugin directory. This was done to prevent concurrent installs: if a process fails to rename the temp dir because the final dir already exists, it means another process already installed the plugin. Unfortunately, on Windows the rename operation often fails due to aggressive virus scanners opening files in the temp dir.

In order to provide reliable plugin installs on Windows, we now extract the tarball directly into the final directory, and use file locks to prevent concurrent installs from toppling over one another.

During install, a lock file is created in the plugin cache directory with the same name as the plugin's final directory but suffixed with `.lock`. The process that obtains the lock is responsible for extracting the tarball. Before it does that, it cleans up any previous temp directories of failed installs of previous versions of Pulumi. Then it creates an empty `.partial` file next to the `.lock` file. The `.partial` file indicates an installation is in-progress. The `.partial` file is deleted when installation is complete, indicating the plugin was successfully installed. If a failure occurs during installation, the`.partial` file will remain indicating the plugin wasn't fully installed. The next time the plugin is installed, the old installation directory will be removed and replaced with a fresh install.

This is the same approach Go uses for installing modules in its module cache.

Fixes https://github.com/pulumi/pulumi/issues/2695

### DIFF
--- a/sdk/go/common/workspace/plugins_install_test.go
+++ b/sdk/go/common/workspace/plugins_install_test.go
@@ -25,6 +25,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"sync"
 	"testing"
 
 	"github.com/blang/semver"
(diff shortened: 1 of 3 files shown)


sdk/go/common/workspace/plugins_install_test.go
```diff
@@ -58,48 +61,163 @@ func createTGZ(files map[string][]byte) ([]byte, error) {
...
+	assert.NoError(t, err)
+
+	assert.False(t, HasPlugin(plugin))
+
+	has, err := HasPluginGTE(plugin)
+	assert.Error(t, err)
+	assert.False(t, has)
+
+	plugins, err := getPlugins(dir)
+	assert.Equal(t, 0, len(plugins))
```
mikhailshilkov, line 222: Consider adding a test that tries to run multiple installations in parallel and validates that the result is consistent.

mikhailshilkov (approved):
Looks great! A bit complicated, but this seems warranted. Do you think our existing test infrastructure would catch most race conditions if they creeped in here? Or do we need any extra testing?

sdk/go/common/workspace/plugins_install_test.go
justinvp: Done.

Issue #5685, opened 2020-11-04
Plugin installation can fail silently during `pulumi new` on Windows


MisinformedDNA:
When running `pulumi new azure-csharp`, the azure plugin seems to download fine, but it actually didn't.

## Expected Behavior
When pulumi cannot install a plugin, an error (or warning) message is displayed.

## Current Behavior
It says "Your new project is ready to go!"

## Steps to Reproduce
1. Run `pulumi new azure-csharp` (I'm using Powershell Core on Windows)

## Context (Environment)
I thought the plugin had installed because it was successful.

## Console output:

> C:\repos\ArmVSPulumi\function-app-consumption\PulumiPortLocal3 [master +7 ~1 -1 !]> pulumi new
> Please choose a template: azure-csharp                A minimal Azure C# Pulumi program
> This command will walk you through creating a new Pulumi project.
> 
> Enter a value or leave blank to accept the (default), and press <ENTER>.
> Press ^C at any time to quit.
> 
> project name: (PulumiPortLocal3)
> project description: (A minimal Azure C# Pulumi program)
> Created project 'PulumiPortLocal3'
> 
> stack name: (dev)
> Created stack 'dev'
> Enter your passphrase to protect config/secrets:
> Re-enter your passphrase to confirm:
> 
> Enter your passphrase to unlock config/secrets
>     (set PULUMI_CONFIG_PASSPHRASE or PULUMI_CONFIG_PASSPHRASE_FILE to remember):
> azure:location: The Azure location to use: (WestUS)
> Saved config
> 
> Installing dependencies...
> 
> running 'dotnet build -nologo .'
>   Determining projects to restore...
> 
>   Restored C:\repos\ArmVSPulumi\function-app-consumption\PulumiPortLocal3\PulumiPortLocal3.csproj (in 5.38 sec).
> 
>   You are using a preview version of .NET. See: https://aka.ms/dotnet-core-preview
> 
>   PulumiPortLocal3 -> C:\repos\ArmVSPulumi\function-app-consumption\PulumiPortLocal3\bin\Debug\netcoreapp3.1\PulumiPortLocal3.dll
> 
> 
> 
> Build succeeded.
>     0 Warning(s)
>     0 Error(s)
> 
> Time Elapsed 00:00:11.04
> 
> 'dotnet build -nologo .' completed successfully
> [resource plugin azure-3.29.0] installing
> Downloading plugin: 56.96 MiB / 56.96 MiB [========================] 100.00% 11s
[...]

PR #5760, opened 2020-11-13
.NET: Report plugin install errors during `pulumi new`

justinvp:
The way `pulumi new` installs dependencies for .NET projects is slightly different from other languages. For Node.js, Python, and Go, `pulumi new` runs the appropriate command to install project dependencies (e.g. `npm install`, `pip install`, or `go mod download`). For .NET, it calls the same routine used during `preview|up` to ensure required plugins are installed. For .NET, this ends up running `dotnet build` which implicitly installs Nuget packages, builds the project, and also attempts to determine and install the needed Pulumi plugins. When this operation runs during `preview|up`, and there are failures installing a plugin, the error is logged, but deliberately not returned, because an error will be shown for missing plugins later on during the `preview|up` operation. However, during `pulumi new`, we should show any plugin install errors.

Note: Plugin install failure during `pulumi new` is now much less likely now on Windows with: https://github.com/pulumi/pulumi/pull/5759

Fixes https://github.com/pulumi/pulumi/issues/5685

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,9 @@ CHANGELOG
 - Fix plugin install failures on Windows.
   [#5759](https://github.com/pulumi/pulumi/pull/5759)
 
+- .NET: Report plugin install errors during `pulumi new`.
+  [#5760](https://github.com/pulumi/pulumi/pull/5760)
+
 ## 2.13.2 (2020-11-06)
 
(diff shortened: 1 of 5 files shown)


mikhailshilkov (approved):
LGTM. This would also help find https://github.com/pulumi/pulumi-azure-nextgen/issues/70 sooner.

Issue #5596, opened 2020-10-20
[Epic] Automation API: C#


EvanBoyle:
The current plan is to pick the work back up on November 16th with the goal of delivering it (along with [Python](https://github.com/pulumi/pulumi/issues/5299)) by the end of the year. We've scoped the effort and it looks like the following:

- [ ] Create `Workspace` Interface [TypeScript reference](https://github.com/pulumi/pulumi/blob/master/sdk/nodejs/x/automation/workspace.ts). Note there are some places where the [Go interface](https://github.com/pulumi/pulumi/blob/master/sdk/go/x/auto/workspace.go) is more complete (cancel, export, import). Ideally, we should shoot for the same level of completeness as the Go impl. 
- [ ] Implement `RunPulumiCmd` [TypeScript reference](https://github.com/pulumi/pulumi/blob/master/sdk/nodejs/x/automation/cmd.ts)
- [ ] `LocalWorkspace` Implementation [TypeScript reference](https://github.com/pulumi/pulumi/blob/master/sdk/nodejs/x/automation/localWorkspace.ts)
- [ ] `Stack` implementation supporting local programs [TypeScript reference](https://github.com/pulumi/pulumi/blob/master/sdk/nodejs/x/automation/stack.ts)
- [ ] Implement idiomatic structured error handling [TypeScript reference](https://github.com/pulumi/pulumi/blob/master/sdk/nodejs/x/automation/errors.ts)
- [ ] Add `InlineProgram` support. This involves creating a C# implementation of the gRPC LanguageServer. TypeScript references: [LanguageServer](https://github.com/pulumi/pulumi/blob/master/sdk/nodejs/x/automation/server.ts), [Preview consumption](https://github.com/pulumi/pulumi/blob/master/sdk/nodejs/x/automation/stack.ts#L228-L250), and [Up consumption](https://github.com/pulumi/pulumi/blob/master/sdk/nodejs/x/automation/stack.ts#L147-L169)
- [ ] Add `RemoteProgram` support (clone and setup programs from git). Go references: [LocalWorkspace](https://github.com/pulumi/pulumi/blob/master/sdk/go/x/auto/local_workspace.go#L514-L521), and [checkout impl](https://github.com/pulumi/pulumi/blob/master/sdk/go/x/auto/git.go), note that this is not yet implemented in TypeScript. 

There are also many tests that should be replicated. These should make the implementation and acceptance criteria fairly straight forward:

[...]

PR #5761, opened 2020-11-13
[Automation API] - C# Implementation

orionstudt:
For: #5596

Here is what I've got so far.

Got kind of busy so wasn't able to get as much done as I liked this week - but will be trying to start the `RunPulumiCmd` portion as well as project settings deserialization early next week.

I will be placing some thoughts / questions in the applicable files...


sdk/dotnet/Pulumi/X/Automation/IWorkspace.cs
```diff
@@ -0,0 +1,198 @@
...
+
+        /// <summary>
+        /// Environment values scoped to the current workspace. These will be supplied to every Pulumi command.
+        /// </summary>
+        IDictionary<string, string>? EnvironmentVariables { get; set;}
+
+        /// <summary>
+        /// Returns project settings for the current project if any.
+        /// </summary>
+        Task<ProjectSettings?> GetProjectSettingsAsync();
```
orionstudt, line 49: Should we be passing a `CancellationToken` around for the asynchronous methods on `IWorkspace`? *Are* these methods cancel-able?

sdk/dotnet/Pulumi/X/Automation/StackSettings.cs
```diff
@@ -0,0 +1,29 @@
...
+        /// <summary>
+        /// This is this stack's base64 encoded encryption salt. Only used for
+        /// passphrase-based secrets providers.
+        /// </summary>
+        public string? EncryptionSalt { get; set; }
+
+        /// <summary>
+        /// This is an optional configuration bag.
+        /// </summary>
+        public IDictionary<string, string>? Config { get; set; } // TODO: figure out what to do for value type argument here
```
orionstudt, line 27: Any thoughts on how to represent the value type argument of this dictionary here, since there is some union type representation going on in the TS implementation? Should we use the `Union<Tx, Ty>` implementation I saw in the repo or can you think of a better way?

sdk/dotnet/Pulumi/X/Automation/IWorkspace.cs
```diff
@@ -0,0 +1,198 @@
...
+        /// <param name="stackName">The name of the stack.</param>
+        Task PostCommandCallbackAsync(string stackName);
+
+        /// <summary>
+        /// Returns the value associated with the specified stack name and key, scoped
+        /// to the Workspace.
+        /// </summary>
+        /// <param name="stackName">The name of the stack to read config from.</param>
+        /// <param name="key">The key to use for the config lookup.</param>
+        Task<ConfigValue> GetConfigAsync(string stackName, string key);
```
orionstudt, line 98: I have some thoughts about the naming of the `Config` related methods on `IWorkspace`. Currently it is:

- `GetConfigAsync` for a single `ConfigValue`.
- `GetAllConfigAsync` for the config map.

The other methods (set, remove) are named similarly. I suggest doing this:

- `GetConfigValueAsync` for a single `ConfigValue`.
- `GetConfigAsync` for the config map.

Just because it will keep the config-related methods near each other for intellisense. Thoughts?

sdk/dotnet/Pulumi/X/Automation/LocalWorkspace.cs
```diff
@@ -0,0 +1,184 @@
...
+                if (options.StackSettings != null && options.StackSettings.Any())
+                {
+                    foreach (var pair in options.StackSettings)
+                        readyTasks.Add(this.SaveStackSettingsAsync(pair.Key, pair.Value));
+                }
+            }
+
+            if (string.IsNullOrWhiteSpace(dir))
+            {
+                dir = Path.Combine(Path.GetTempPath(), "automation-"); // TODO: should this not be randomized in some way?
```
orionstudt, line 66: Should the `automation-` portion of this file path have some randomized part of it or do we intend to always be in the same directory?

sdk/dotnet/Pulumi/X/Automation/PluginInfo.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        public string Path { get; }
+
+        public PluginKind Kind { get; }
+
+        public string? Version { get; }
+
+        public long Size { get; } // TODO: or double? will know once get to implementation
+
+        public DateTime InstallTime { get; } // TODO: is this UTC or local?
```
orionstudt, line 17: Are these `DateTime` values returned from the CLI as UTC or local?

sdk/dotnet/Pulumi/X/Automation/StackSettings.cs
orionstudt: [Here](https://github.com/pulumi/pulumi/blob/master/sdk/nodejs/x/automation/stackSettings.ts) is the TS implementation.

sdk/dotnet/Pulumi/X/Automation/PluginInfo.cs
justinvp: I don't recall off hand if the values are UTC or local (will need to check), but regardless, these properties should be typed as `DateTimeOffset` rather than `DateTime` (and then no need to suffix `Utc`).

sdk/dotnet/Pulumi/X/Automation/IWorkspace.cs
```diff
@@ -0,0 +1,198 @@
...
+
+namespace Pulumi.X.Automation 
+{
+    /// <summary>
+    /// Workspace is the execution context containing a single Pulumi project, a program, and multiple stacks.
+    /// <para/>
+    /// Workspaces are used to manage the execution environment, providing various utilities such as plugin
+    /// installation, environment configuration ($PULUMI_HOME), and creation, deletion, and listing of Stacks.
+    /// </summary>
+    public interface IWorkspace
```
justinvp, line 13: For .NET, we may want to model this as an abstract class (e.g. `public abstract class Workspace`) rather than as an interface to give us more flexibility to add members in the future.

sdk/dotnet/Pulumi/X/Automation/IWorkspace.cs
justinvp: I like this suggestion. If we do make a naming change here, we should consider updating the other languages as well to be consistent.

Another option:

- `GetConfigValueAsync` (singular) for a single `ConfigValue`.
- `GetConfigValuesAsync` (plural) for the config map.

Which is similar to the built-in [`Environment.GetEnvironmentVariable`](https://docs.microsoft.com/en-us/dotnet/api/system.environment.getenvironmentvariable?view=net-5.0) and [`Environment.GetEnvironmentVariables`](https://docs.microsoft.com/en-us/dotnet/api/system.environment.getenvironmentvariables?view=net-5.0).

Another option: for languages that support overloads (C#, TypeScript, Python), we could have two overloads with the same name (e.g. `GetConfigAsync`). The overload that accepts just the `stackName` returns all config, the overload that accepts both the `stackName` and `key` returns just the value:

```c#
Task<ImmutableDictionary<string, ConfigValue>> GetConfigAsync(string stackName);
Task<ConfigValue> GetConfigAsync(string stackName, string key);
```

Go doesn't support overloads, so if we move to this approach, we'd need to either keep Go's naming as-is, or rename based on the other naming suggestions above. Or another possible option for Go is `GetConfig` (single value) and `GetConfigMap` (all config).

@EvanBoyle, thoughts?

justinvp: Another idea I've been kicking around, but not sure how I feel about it just yet... grouping functionality together more. Something like:

```c#
public abstract class Workspace
{
    // ...

    public abstract WorkspaceConfig Config { get; }
    public abstract WorkspaceStacks Stacks { get; }
    public abstract WorkspacePlugins Plugins { get; }

    // ...
}

public abstract class WorkspaceConfig
{
    public abstract Task<ConfigValue> GetAsync(string stackName, string key);
    public abstract Task<ImmutableDictionary<string, ConfigValue>> GetAllAsync(string stackName);

    public abstract Task SetAsync(string stackName, string key, ConfigValue value);
    public abstract Task SetAllAsync(string stackName, IDictionary<string, ConfigValue> configMap);

    public abstract Task RemoveAsync(string stackName, string key);
    public abstract Task RemoveAllAsync(string stackName, IEnumerable<string> keys);

    public abstract Task<ImmutableDictionary<string, ConfigValue>> RefreshAsync(string stackName);
}

public abstract class WorkspaceStacks
{
    public abstract Task<StackInfo?> GetSelectedAsync();
    public abstract Task CreateAsync(string stackName);
    public abstract Task SelectAsync(string stackName);
    public abstract Task RemoveAsync(string stackName);
    public abstract Task<ImmutableList<StackInfo>> ListAsync();
}

public abstract class WorkspacePlugins
{
    public abstract Task InstallAsync(string name, string version, string? kind = null);
    public abstract Task RemoveAsync(string? name = null, string? versionRange = null, string? kind = null);
    public abstract Task<ImmutableList<PluginInfo>> ListAsync();
}
```

With usage like:

```c#
Workspace ws = ...;
var config = await ws.Config.GetAllAsync();
await ws.Stacks.SelectAsync("foo");
await ws.Plugins.InstallAsync("aws", "2.0.0");
```

sdk/dotnet/Pulumi/X/Automation/IWorkspace.cs
orionstudt: We could do that. Is there some functionality that is currently is the LocalWorkspace implementations that would belong more in an abstract Workspace class?

orionstudt: There's also that new C# feature for default method implementations on interfaces :) don't know if that would be useful here though if you're just thinking about future dev

sdk/dotnet/Pulumi/X/Automation/IWorkspace.cs
EvanBoyle: @justinvp Interesting idea. One thing I have a hard time with is the relationship between `Workspace`, `Stack`, and `Config`. The example is missing a param for config:

```cs
var config = await ws.Config.GetAllAsync("bar");
await ws.Stacks.SelectAsync("foo");
await ws.Plugins.InstallAsync("aws", "2.0.0");
```

Both config and stack are keyed by stack name, which makes the grouping feel a bit odd. The config belongs to a stack and the relationship with workspace is only really for persistence. Not sure if this subgrouping over-emphasizes the relationship between config and workspace.

I am a fan of using better names for single value vs group operations on config.

sdk/dotnet/Pulumi/X/Automation/LocalWorkspace.cs
EvanBoyle: The API should support creating multiple of these "anonymous" working directories. For instance, two inline programs for project A and B could be created in separate working directories, but have stacks that are named the same and would collide if backed by the same workdir. In other words:

- projA + stackFoo = `pulumi.foo.yaml`
- projB + stackFoo = `pulumi.foo.yaml`

sdk/dotnet/Pulumi/X/Automation/StackInfo.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    public class StackInfo
+    {
+        public string Name { get; }
+
+        public bool IsCurrent { get; }
+
+        public string? LastUpdate { get; }
+
+        public bool IsUpdateInProgress { get; }
```
EvanBoyle, line 11: I'm concerned this may be a little misleading. We don't have a way of knowing if an update is in progress (short of trying to start one) since another machine or user could be running one independently of this instance of the Stack.

sdk/dotnet/Pulumi/X/Automation/StackSettings.cs
EvanBoyle: For a little more context, here's the reference to the original format: https://github.com/pulumi/pulumi/blob/6d09fe32df2ac7eca3db08c3b4a61dddcb7d756e/sdk/go/common/resource/config/value.go#L25

Serde is a little complicated on this type. In practice, I think it will be rare for users to manipulate config via this interface (any secret values retrieved from this will be encrypted).

EvanBoyle (commented):
This is shaping up nicely!

sdk/dotnet/Pulumi/X/Automation/IWorkspace.cs
orionstudt: So if we went with that then it would probably make more sense for `Config` manipulation to be on the `Stacks` object?
```csharp
var config = await ws.Stacks.GetConfigAsync("stack");
var value = await ws.Stacks.GetConfigValueAsync("stack", "key");
```

so at that point is there still value in keeping `Stacks` & `Plugins` separated?

sdk/dotnet/Pulumi/X/Automation/LocalWorkspace.cs
orionstudt: Ok, that makes sense. So we should probably have `LocalWorkspace` implement `IDisposable` to cleanup that directory it created?

sdk/dotnet/Pulumi/X/Automation/StackInfo.cs
orionstudt: I think that is from [this.](https://github.com/pulumi/pulumi/blob/66b7794fbceb1db2c4d8d492b308d973bc167598/sdk/nodejs/x/automation/workspace.ts#L200) Should we come up with a more accurate name for that?

StefanXhunga (commented):
Thank you for your assistance 18/19 viewed

sdk/dotnet/Pulumi/X/Automation/LocalWorkspace.cs
EvanBoyle: This is a good idea. So far we've relied on the operating system to clean up any temp directories, but this would be good hygiene.

sdk/dotnet/Pulumi/X/Automation/IWorkspace.cs
justinvp: > Both config and stack are keyed by stack name, which makes the grouping feel a bit odd. The config belongs to a stack and the relationship with workspace is only really for persistence. Not sure if this subgrouping over-emphasizes the relationship between config and workspace.

Agreed. Yeah, my initial factoring isn't quite right: if we were to go down this path, the config-related things would probably be better on `Stacks`, but even then it doesn't really make it clear that this is config for the current stack.

In any case, apologies for taking us on this tangent. Let's ignore my grouping idea for now and go back to discussing the previous naming suggestions on the top-level workspace:

### Current naming:

- `GetConfigAsync` for a single `ConfigValue`.
- `GetAllConfigAsync` for the config map.

### Option 1 (@orionstudt's original suggestion):

- `GetConfigValueAsync` for a single `ConfigValue`.
- `GetConfigAsync` for the config map.

### Option 2:

- `GetConfigValueAsync` (singular) for a single `ConfigValue`.
- `GetConfigValuesAsync` (plural) for the config map.

Which is similar to the built-in [`Environment.GetEnvironmentVariable`](https://docs.microsoft.com/en-us/dotnet/api/system.environment.getenvironmentvariable?view=net-5.0) and [`Environment.GetEnvironmentVariables`](https://docs.microsoft.com/en-us/dotnet/api/system.environment.getenvironmentvariables?view=net-5.0).

### Option 3:

For languages that support overloads (C#, TypeScript, Python), we could have two overloads with the same name (e.g. `GetConfigAsync`). The overload that accepts just the `stackName` returns all config, the overload that accepts both the `stackName` and `key` returns just the value:

```c#
Task<ImmutableDictionary<string, ConfigValue>> GetConfigAsync(string stackName);
Task<ConfigValue> GetConfigAsync(string stackName, string key);
```

Go doesn't support overloads, so if we move to this approach, we'd need to either keep Go's naming as-is, or rename based on the other naming suggestions above. Or another possible option for Go is `GetConfig` (single value) and `GetConfigMap` (all config).

---

Personally, I'm in favor of Option 1 (@orionstudt's original suggestion). If we do this, we should be prepared to change the naming for TS and Go for consistency.

sdk/dotnet/Pulumi/X/Automation/IWorkspace.cs
justinvp: > Is there some functionality that is currently in the LocalWorkspace implementations that would belong more in an abstract Workspace class?

It's possible. Might be easiest to start with implementing on `LocalWorkspace` and then see if there's anything general enough that could be provided as a default implementation on the base `Workspace`.

sdk/dotnet/Pulumi/X/Automation/IWorkspace.cs
orionstudt: I went with Option 1 for now, can decide more later.

orionstudt: Also IDK if you want any naming changes made to the other automation APIs to be done in this PR or a different PR?

EvanBoyle: I think it would be best to keep this change set targeted to C#. Before merging we can open some tracking issues for updates to be made to other SDKs. These will involve updating the samples repo as well, so it is a little less contained.

orionstudt:
@EvanBoyle @justinvp 

Alright so latest commit has:
- `RunPulumiCmd` implementation. For .NET I made this into an internal interface / instance so that it could be dropped in and out for testing (which IDK how we're doing yet)
- Fleshed out all of the `LocalWorkspace` methods
- Started on `Stack` implementation which is currently called `XStack` pending naming decision
- Started to separate out `LocalWorkspace`'s serialization into an internal `LocalSerializer` service because I am expecting that once I start testing I will need to write some custom converters and this means it can be tested independently + serialization settings can be configured in 1 place
- I believe I have the "idiomatic structured error handling" completed in here as well

Big next step for me right now is getting any custom serialization converters written and then testing each of the methods on `LocalWorkspace`, at which point I will move to the up, preview, & destroy methods on `XStack`.

Current questions:

1. Original issue mentions the Import, Export, & Cancel methods on `Workspace` in GO. I found code for import & export but couldn't find cancel?
2. Currently the TypeScript implementation of `static LocalWorkspace.Create : LocalWorkspace` doesn't set default project settings if they are not provided, but the `static LocalWorkspace.(Create or Select)Stack : Stack` methods do set default project settings if not provided. Should we make sure the static method that just creates a workspace does so as well?
3. The TypeScript implementation of `GetProjectSettings` and `GetStackSettings` both throw an exception if it is unable to find a settings file, while the XML comment describing the methods seem to indicate that it could or should return `null` if settings are not found. Should we be throwing if there is no settings file or just returning `null`?

orionstudt:
I will return to this after the holiday.

EvanBoyle:
I made my way through the recent changes, functionally it looks great. I'll leave it to @justinvp to comment on dotnet-isms.

To answer your questions:

> I found code for import & export but couldn't find cancel?

cancel is an operation on stack, similar to up/preview/destroy/refresh: https://github.com/pulumi/pulumi/blob/master/sdk/go/x/auto/stack.go#L579-L595

> Should we make sure the static method that just creates a workspace does so as well?

A stack has to have a project settings file, but a workspace does not (it could be empty). 

> static LocalWorkspace.(Create or Select)Stack : Stack methods do set default project settings if not provided.

We only set a default project settings if none is provided in the inline program case:
https://github.com/pulumi/pulumi/blob/master/sdk/nodejs/x/automation/localWorkspace.ts#L171-L187

For stacks using local programs, we assume an existing `pulumi.yaml` already exists in the directory. 

> Should we be throwing if there is no settings file or just returning null?

Yes, this sounds reasonable. The existing design of throwing in go/typescript isn't ideal and we should probably change it.

orionstudt:
Alright some things to note for the state of this PR now that it is ready for review.

From the checklist on #5596 this PR includes everything except:
- Cancel, Export, Import APIs on `Workspace`
- `RemoteProgram` support.

This PR also has 1 major limitation in that due to global state on the `Deployment` object (specifically `Deployment.Instance`) only 1 call to `UpAsync` or `PreviewAsync` may run at a time when running with `InlineProgram`. It does not currently support parallel execution of `InlineProgram` such as might be useful in a Web API use-case. Future development will need to detach existing types in `Pulumi.Core` and `Pulumi.Resources` namespaces from referencing the static singleton `Deployment.Instance`.

orionstudt:
Scratch the aforementioned parallel execution of inline program limitation. It should be working as per the latest commit. There is a new test written for it.

orionstudt:
Alright I am at an impass and cannot continue without assistance. I was working on getting this merged with current master so that I could have this PR into master and we could have the workflows passing for you guys. I have local pending outgoing commits that are resolving all remaining merge conflicts and test are passing but I am encountering this error when I try to `git push` due to changes made to workflows:

```
Enumerating objects: 45, done.
Counting objects: 100% (45/45), done.
Delta compression using up to 12 threads
Compressing objects: 100% (21/21), done.
Writing objects: 100% (21/21), 2.49 KiB | 196.00 KiB/s, done.
Total 21 (delta 19), reused 0 (delta 0)
remote: Resolving deltas: 100% (19/19), completed with 12 local objects.
To https://github.com/orionstudt/pulumi
 ! [remote rejected]     auto/dotnet -> auto/dotnet (refusing to allow a Personal Access Token to create or update workflow `.github/workflows/command-dispatch.yml` without `workflow` scope)
error: failed to push some refs to 'https://github.com/orionstudt/pulumi'
```

sdk/dotnet/Pulumi/Serialization/ResourcePackages.cs
```diff
@@ -10,14 +10,35 @@
...
+                () => DiscoverResourceTypes(LoadCurrentDomainReferencedAssemblies()));
+
+        private readonly ImmutableList<Assembly>? _assemblies;
+        private readonly Lazy<ImmutableDictionary<string, ImmutableList<(string?, Type)>>> _resourceTypes;
+
+        /// <param name="assemblies">
+        /// Assemblies that should be used to discover resource types. If <c>null</c>, than <see cref="AppDomain.CurrentDomain"/>
+        /// will be used to find referenced assemblies and discover resource types.
+        /// </param>
+        public ResourcePackages(IEnumerable<Assembly>? assemblies)
```
orionstudt, line 26: I want to explain this ahead of time.

Our use-case for Pulumi Automation API at my company requires this.. hoping you're OK with it being part of this PR. Essentially our AspNetCore WebAPI will be dynamically loading plugins in order to get our `PulumiFn` equivalent. This means our `PulumiFn` is not part of the same solution as our WebAPI, which means that it is possible for our WebAPI to not have the references that contain those `CustomResource` implementations - as only the plugin loaded at runtime will have those references.

So this allows us to pass through any of the assemblies that the plugin references to get `CustomResource` implementations, without the WebAPI itself having those references, since `AppDomain.CurrentDomain.GetReferencedAssemblies()` would only return the WebAPI's references.

sdk/dotnet/Pulumi/Serialization/ResourcePackages.cs
```diff
@@ -106,7 +122,10 @@ private static IEnumerable<Assembly> LoadReferencedAssemblies()
                 }
             }
 
-            static bool PossibleMatch(AssemblyName? assembly) => assembly != null && !assembly.FullName.StartsWith("System", StringComparison.Ordinal);
+            static bool PossibleMatch(AssemblyName? assembly)
+                => assembly != null
+                && !assembly.FullName.StartsWith("System", StringComparison.Ordinal)
+                && assembly.ContentType != AssemblyContentType.WindowsRuntime;
```
orionstudt, line 128: I believe I needed this condition because of the AspNetCore Framework reference. It was throwing trying to load `Windows.Storage` which is a WindowsRuntime assembly.

orionstudt:
By the way, enabling test collection parallelization in our XUnit test suite works now.

I also noticed that the GO implementation has some additional tests for the command exceptions (update concurrency, stack already exists) so I could add those to this PR.

sdk/dotnet/Pulumi/X/Automation/PulumiFn.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System.Collections.Generic;
+
+namespace Pulumi.X.Automation
+{
+    /// <summary>
+    /// A Pulumi program as an inline function (in process).
+    /// </summary>
+    public delegate IDictionary<string, object?> PulumiFn();
```
orionstudt, line 8: I do wonder if we should have this return `Task<IDictionary<string, object?>` instead? It would allow consumers to make their `PulumiFn` asynchronous, but it would also make synchronous functions a bit uglier because they would need to return `Task.FromResult(output)`. 

We could have a `PulumiFnAsync` and a `PulumiFn` at the same time, and just have an overload for `PulumiFn` anywhere that we take in `PulumiFnAsync`.. that might be a bit easier to swallow. But all of our options objects would have `PulumiFnAsync` and you wouldn't be able to set that with a `PulumiFn` directly.

I also think that if we do have a `PulumiFnAsync`, that it should take in a `CancellationToken` - specifically the one we get from the CLI call so that if the CLI cancels our inline function is made aware.

orionstudt:
Another thing to note: the TS implementation of import/export [here](https://github.com/orionstudt/pulumi/blob/31e82fcee0f8feb50dd43fa06e6ac1fcb46573fd/sdk/nodejs/x/automation/workspace.ts#L196) returns an object called `Deployment` so we have another naming collision issue when adding the import/export functionality..

mikhailshilkov:
Hi @orionstudt thank you for this awesome work! 🎉 

On a high level:

1. We should separate all changes that aren't directly part of the automation API to a separate PR. Particularly, the `AsyncLocal` changes in a minor version make me nervous. Let's open a focused PR just for that, test and review it in detail. I'm interested to know whether there are any undesired externally-visible changes in behavior related to this.

2. Consider moving all the automation API to a separate `Pulumi.Automation` assembly. This way, we can mark it as a preview, iterate on it separately, avoid the unconventional `X` namespace, avoid bloating the references and the API surface of the core library. I guess we'll have to open internals of `Pulumi` to `Pulumi.Automation` which sounds fine.

Ideally, (2) could exist without (1), with some obvious limitations. This gives us the opportunity to consider shipping (1) as part of 3.0 if we identify it as breaking.

orionstudt:
That makes sense - I was thinking something along those lines as well. It will be easier to reason about them separately. I will work on splitting those PRs for you this morning.

I also agree that it makes way more sense to have Automation namespace in a separate assembly, if only because of the necessary AspNetCore framework reference.

orionstudt:
Alright, there is now a separate PR for only the changes required to allow concurrent execution with out global state. That PR is #6139. Now that that can be reviewed separately, I will need to work on back-tracking some of the changes made in this PR. That will take me a little bit longer.

orionstudt:
Alright this PR now only includes automation API without any changes necessary for execution concurrency. Meaning it has 2 major limitations:
1. The obvious one, you cannot execute 2 calls to `Deployment.Run` simultaneously because of the global state interference.
2. The executing assembly _must_ itself reference all necessary assemblies needed by the `ResourcePackages` implementation (specifically at the time of the first call to `Deployment.Run`), because our current `ResourcePackages` implementation finds those assemblies via `AppDomain.CurrentDomain`, caches the result statically, and does not allow passthrough of external assemblies.

sdk/dotnet/Pulumi.Automation/Runtime/LanguageRuntimeService.cs
```diff
@@ -0,0 +1,69 @@
...
+            var settings = new InlineDeploymentSettings(
+                engineAddr,
+                request.MonitorAddress,
+                request.Config,
+                request.Project,
+                request.Stack,
+                request.Parallel,
+                request.DryRun);
+
+            await Deployment.RunInlineAsync(settings, () => Task.FromResult(this._program())).ConfigureAwait(false);
```
EvanBoyle, line 49: Have we thought through how errors propagate when user code (the program being run) throws exceptions or does unsafe things (dereferencing null values)? Providing helpful error messages is important here.

sdk/dotnet/Pulumi.Automation/XStack.cs
```diff
@@ -0,0 +1,631 @@
...
+                    upResult.StandardOutput,
+                    upResult.StandardError,
+                    summary!,
+                    output);
+            }
+            finally
+            {
+                if (inlineHost != null)
+                {
+                    await inlineHost.DisposeAsync().ConfigureAwait(false);
```
EvanBoyle, line 286: One thing to consider is that the language runtime server is running arbitrary user code that could throw, dereference null values, etc. 

Have we done any manual testing to ensure that the resulting error is caught and returned to the user?

EvanBoyle (commented):
Wow! Overall this looks amazing. The level of quality (including docs), test coverage, and attention to detail is outstanding. It is obvious that a lot of hard work went into this. Thank you!

Structurally, this is exactly what I would expect. The examples (unit tests for inline program) feel natural and are concise. 

I have two questions about error handling noted in the PR. I'm not familiar with how these values are propagated in our dotnet runtime, but this is an area that warrants a little bit of consideration and manual testing. 

Overall this looks great, and my sense is that this is very close to being ready to merge. I would like @mikhailshilkov to take another look from more of a C# specific point of view as this is not my area of expertise.

Another item that doesn't need to block merge, but that we need to take care of before official release, blogs, etc is porting examples from: https://github.com/pulumi/automation-api-examples

I'm not sure if you're interested in participating in that effort, you definitely don't have to unless you're interested! It was a helpful exercise with the python implementation that helped identify a few usability corners and docs issues (intellisense, IDE formatting etc). https://github.com/pulumi/automation-api-examples

One thing that I would be curious to know iwhat a program structure looks like for a "hybrid" setup where the same program is used both with the CLI and with automation API. Similar to this example in go: 

https://github.com/pulumi/automation-api-examples/tree/main/go/inline_local_hybrid

I remember that a part of the initial design was ensuring that existing programs would be compatible with Automation API drivers, so it would be nice to confirm.

sdk/dotnet/Pulumi.Automation/XStack.cs
```diff
@@ -0,0 +1,631 @@
...
+                this._cancelToken = cancellationToken;
+                this._host = Host.CreateDefaultBuilder()
+                    .ConfigureWebHostDefaults(webBuilder =>
+                    {
+                        webBuilder
+                            .ConfigureKestrel(kestrelOptions =>
+                            {
+                                kestrelOptions.Listen(IPAddress.Any, 0, listenOptions =>
+                                {
+                                    // TODO: HTTPS?
```
mikhailshilkov, line 569: What's the plan for TODOs? Will we open follow-up issues?

mikhailshilkov (commented):
This looks great indeed! I took a quick pass, see my questions below

sdk/dotnet/Pulumi.Automation/InlineProgramArgs.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace Pulumi.Automation
+{
+    public class InlineProgramArgs : LocalWorkspaceOptions
+    {
+        public string StackName { get; }
+
+        public InlineProgramArgs(
+            string projectName,
+            string stackName,
+            PulumiFn program)
```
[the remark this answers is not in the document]
orionstudt: Yea I made this remark further up about possibly adding async support to `PulumiFn`. There would be trade-offs and in some scenarios the delegate code may be ugly unless we can provide synchronous delegate overloads.

We could make some generic `TStack` overloads in key entrypoints the way the `Deployment.Run` methods do and just pipe the `new TStack();` into our `PulumiFn` delegate to allow for existing `Pulumi.Stack` implementations to be sent in?

sdk/dotnet/Pulumi.Automation/UpdateKind.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace Pulumi.Automation
+{
+    public enum UpdateKind
```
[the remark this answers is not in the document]
orionstudt: yea: https://github.com/pulumi/pulumi/blob/0f06e03931fc4d6b3d765671e5f3a97ead795b48/sdk/nodejs/x/automation/stack.ts#L510

sdk/dotnet/Pulumi.Automation/Pulumi.Automation.csproj
```diff
@@ -0,0 +1,63 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <Nullable>enable</Nullable>
+    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
+    <Authors>Pulumi</Authors>
+    <Company>Pulumi Corp.</Company>
+    <Description>The Pulumi .NET SDK lets you write cloud programs in C#, F#, and VB.NET.</Description>
```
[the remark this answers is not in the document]
orionstudt: Sure, any suggestions?

sdk/dotnet/Pulumi.Automation/XStack.cs
```diff
@@ -0,0 +1,631 @@
...
+    /// Pulumi program. <see cref="XStack"/> exposes methods for the full pulumi lifecycle
+    /// (up/preview/refresh/destroy), as well as managing configuration.
+    /// <para/>
+    /// Multiple stacks are commonly used to denote different phases of development
+    /// (such as development, staging, and production) or feature branches (such as
+    /// feature-x-dev, jane-feature-x-dev).
+    /// <para/>
+    /// Will dispose the <see cref="Workspace"/> on <see cref="Dispose"/>.
+    /// </summary>
+    public sealed class XStack : IDisposable // TODO: come up with a name for this
```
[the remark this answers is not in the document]
orionstudt: I can't think of anything else that would make sense given that we are committed to `Stack` in the other SDKs. I think this could be `Pulumi.Automation.Stack` but yea, since `Stack` implements `Resource` it's possible we have collisions if they are used in the same program.

sdk/dotnet/Pulumi.Automation.Tests/LocalWorkspaceTests.cs
```diff
@@ -0,0 +1,490 @@
...
+            Assert.True(secureValue.IsSecure);
+        }
+
+        [Fact]
+        public async Task AddRemoveListPlugins()
+        {
+            using var workspace = await LocalWorkspace.CreateAsync();
+            await workspace.InstallPluginAsync("aws", "v3.0.0");
+            await workspace.RemovePluginAsync("aws", "3.0.0");
+            await workspace.ListPluginsAsync();
```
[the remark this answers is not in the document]
orionstudt: yea definitely. I was just mimicking this: https://github.com/pulumi/pulumi/blob/0f06e03931fc4d6b3d765671e5f3a97ead795b48/sdk/nodejs/tests/automation/localWorkspace.spec.ts#L44

but I can update it.

sdk/dotnet/Pulumi.Automation.Tests/LocalWorkspaceTests.cs
```diff
@@ -0,0 +1,490 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Pulumi.Automation.Commands.Exceptions;
+using Xunit;
+
+namespace Pulumi.Automation.Tests
+{
+    public class LocalWorkspaceTests
+    {
+        private static readonly string DataDirectory =
```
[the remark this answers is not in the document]
orionstudt: I guess I've always named statics capitalized. But you're right you guys have a set naming convention. I'll fix it.

sdk/dotnet/Pulumi/Deployment/Deployment_Inline.cs
```diff
@@ -0,0 +1,46 @@
...
+            //var queryMode = Environment.GetEnvironmentVariable("PULUMI_QUERY_MODE");
+            //var parallel = Environment.GetEnvironmentVariable("PULUMI_PARALLEL");
+            //var tracing = Environment.GetEnvironmentVariable("PULUMI_TRACING");
+
+            if (string.IsNullOrEmpty(settings.MonitorAddr)
+                || string.IsNullOrEmpty(settings.EngineAddr)
+                || string.IsNullOrEmpty(_projectName)
+                || string.IsNullOrEmpty(_stackName))
+            {
+                throw new InvalidOperationException("Program run without the Pulumi engine available; re-run using the `pulumi` CLI");
```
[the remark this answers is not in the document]
orionstudt: No that error doesn't really make sense lol.  I'll see if I can put something better. Open to suggestions

sdk/dotnet/Pulumi.Automation/XStack.cs
orionstudt: This is an excellent point and I don't think we are handling it properly. I will write a test for it and see what I can do.

sdk/dotnet/Pulumi.Automation.Tests/LocalWorkspaceTests.cs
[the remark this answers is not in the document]
orionstudt: These were updated in the other tests as well.

sdk/dotnet/Pulumi.Automation.Tests/LocalWorkspaceTests.cs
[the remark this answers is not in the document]
orionstudt: This test as well as the create/remove/list stack has been updated.

sdk/dotnet/Pulumi/Deployment/Deployment_Inline.cs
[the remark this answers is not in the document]
orionstudt: I have updated the error let me know if the new text is OK

sdk/dotnet/Pulumi.Automation/XStack.cs
orionstudt: I have written a test for this. Exceptions thrown in the user provided code are now propagated up to the caller as I would expect as a consumer when calling `UpAsync` or `PreviewAsync`. I think that captures what you had in mind?

orionstudt:
> Another item that doesn't need to block merge, but that we need to take care of before official release, blogs, etc is porting examples from: https://github.com/pulumi/automation-api-examples
> 
> I'm not sure if you're interested in participating in that effort, you definitely don't have to unless you're interested! It was a helpful exercise with the python implementation that helped identify a few usability corners and docs issues (intellisense, IDE formatting etc). https://github.com/pulumi/automation-api-examples
> 
> One thing that I would be curious to know iwhat a program structure looks like for a "hybrid" setup where the same program is used both with the CLI and with automation API. Similar to this example in go:
> 
> https://github.com/pulumi/automation-api-examples/tree/main/go/inline_local_hybrid

I can definitely work on some of this at some point. Either as part of this PR or another.

sdk/dotnet/Pulumi.Automation/InlineProgramArgs.cs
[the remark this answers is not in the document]
orionstudt: @mikhailshilkov one thing to think about if we allow the delegate to return a `Task` and as such allow `async` execution is all of the `.InvokeAsync` APIs that return `Task<T>` and are meant to be piped into `Output.Create(Func<Task<T>>)`.

Like If I, instead of sending the `.InvokeAsync` to `Output.Create(...)`, just did `await XXX.InvokeAsync(...);` would that cause any issues or would that work out of the box?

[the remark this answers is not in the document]
mikhailshilkov: That would work if you are in an `async` method. Piping to `Output.Create` is just a way to avoid calling `.Result` inside a sync constructor.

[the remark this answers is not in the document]
orionstudt: OK cool I didn't know if you guys were doing anything special and just hadn't implemented your own TaskAwaiter or something. Well we can go whatever route you guys choose I just chose synchronous to begin with because that is essentially mimicking what we had before doing it in the `Pulumi.Stack` constructor.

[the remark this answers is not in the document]
mikhailshilkov: IMO, it's important to support both `Task`-based functions and `Stack`'s for consistency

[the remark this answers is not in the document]
orionstudt: So, should we have two delegates - a `PulumiFn` and a `PulumiFnAsync`, and just fallback to the async one with overloads for the synchronous one? Or should we just make `PulumiFn` return a task?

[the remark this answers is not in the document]
mikhailshilkov: Two delegates make sense to me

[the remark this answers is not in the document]
orionstudt: OK please checkout the most recent commit and let me know what you think.

I tried to do 2 delegates but it was messy. Because delegates can't do any implicit conversions i required us to have overloads everywhere for sync & async delegates, as well as overloads for `TStack` generic being provided. This was tolerable but still verbose on things like `InlineProgramArgs` where an inline program is required, but it was downright messy on something like `UpOptions` where an inline program was optional - as that required either: 1. sticking with a single async property and requiring the consumer to return a task or 2. having 2 conflicting properties, sync & async, and making sure the consumer only used 1.

Instead I tried changing the `PulumiFn` to a class. This can take in any number of delegates, including a `TStack`. So anywhere we take in `PulumiFn` we now support `TStack`. And it is very easy for us to change the shapes of any delegates that we want to allow as those changes will be encapsulated within `PulumiFn` and won't affect our other APIs since the overloads are no longer necessary.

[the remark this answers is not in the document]
orionstudt: Disregard there is an issue with this that I need to resolve.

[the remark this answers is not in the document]
orionstudt: Alright it is good now and there is a test for `TStack` usage. Some changes to internals had to be made to that `TStack` was passed down to `IRunner` correctly so that the output registration continues to occur as it should in `IRunner`. Let me know how you feel about these changes to `PulumiFn`.

sdk/dotnet/Pulumi.Automation/Serialization/Json/MapToModelJsonConverter.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright 2016-2021, Pulumi Corporation
+
+using System;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace Pulumi.Automation.Serialization.Json
+{
+    // necessary because this version of System.Text.Json
+    // can't deserialize a type that doesn't have a parameterless constructor
```
Hawxy, line 10: Would there be any issues in the context of this package by taking a direct dependency on the latest version of System.Text.Json? The version that shipped with .NET Core 3.1 is pretty well known as being very undercooked.

orionstudt: Might need input from the pulumi guys on this since they have a better understanding of how they are using `System.Text.Json` in the core library.

sdk/dotnet/Pulumi.Automation/InlineProgramArgs.cs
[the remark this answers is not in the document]
mikhailshilkov: I gave this a try and it looks great. I'm happy with what you came up with here!

sdk/dotnet/Pulumi.Automation/Pulumi.Automation.csproj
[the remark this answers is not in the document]
mikhailshilkov: "Pulumi Automation API, the programmatic interface for driving Pulumi programs without the CLI." - copied from the Go SDK

sdk/dotnet/Pulumi.Automation/InlineProgramArgs.cs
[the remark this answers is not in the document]
orionstudt: Awesome! I was wondering if you guys like `PulumiFn.Create(...)` or `PulumiFn.New(...)`, I waffled between the two and went with Create but I keep going back and forth.

sdk/dotnet/Pulumi.Automation/XStack.cs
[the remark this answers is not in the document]
mikhailshilkov: `AutomationStack`? `WorkspaceStack`? `PulumiStack`? cc @EvanBoyle

sdk/dotnet/Pulumi.Automation/Serialization/Json/MapToModelJsonConverter.cs
mikhailshilkov: We rely on `System.Text.Json` in our core Pulumi package. Why would it be a bigger problem in automation specifically? I think we'll go ahead with this.

sdk/dotnet/Pulumi.Automation/InlineProgramArgs.cs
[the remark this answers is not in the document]
mikhailshilkov: `Create` sounds natural to me 👍

sdk/dotnet/Pulumi.Automation/XStack.cs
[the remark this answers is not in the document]
orionstudt: ah something like that could work. Keeping in mind that there is static methods on it so like: `AutomationStack.SelectAsync(..)`, `WorkspaceStack.SelectAsync(...)`, or `PulumiStack.SelectAsync(...)`

sdk/dotnet/Pulumi.Automation/Serialization/Json/MapToModelJsonConverter.cs
orionstudt: For clarity's sake, you want to go ahead with referencing a more recent version in the automation API package?

mikhailshilkov: Sorry, I actually mean that I'm fine with what you have now

sdk/dotnet/Pulumi.Automation/XStack.cs
orionstudt: I think that's up to you guys. I put a question mark after this one and the logging one because I didn't know if you guys would think those were actually a goal or not. Do we want our language host traffic to go over HTTPS? Do we want to allow consumers to hook into the logging on the language host? If not we can get rid of those TODOs.

sdk/dotnet/Pulumi.Automation/XStack.cs
EvanBoyle: Yes, I believe it does. Just wanted to make sure that errors are handled gracefully and propagated to the user, and that the gRPC server doesn't leak in the error case.

sdk/dotnet/Pulumi.Automation/Pulumi.Automation.csproj
```diff
@@ -0,0 +1,63 @@
...
+    <PackageProjectUrl>https://www.pulumi.com</PackageProjectUrl>
+    <RepositoryUrl>https://github.com/pulumi/pulumi</RepositoryUrl>
+    <PackageLicenseExpression>Apache-2.0</PackageLicenseExpression>
+    <PackageIcon>pulumi_logo_64x64.png</PackageIcon>
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+  </PropertyGroup>
```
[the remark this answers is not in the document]
EvanBoyle: I don't want to overcomplicate versioning. Ideally, we just version it in unison with the SDK. Preview label is fine, but this was the intention of keeping it in an `x` namespace originally. Do we need/want to maintain extra machinery of keeping this in a different assembly? 

While unconventional, we're doing the same thing in other languages and communicating risk of API churn to users via the announcements.

sdk/dotnet/Pulumi.Automation/XStack.cs
EvanBoyle: We can get rid of the todos. Other languages run gRPC services over HTTP on localhost, no https yet.

sdk/dotnet/Pulumi.Automation/Pulumi.Automation.csproj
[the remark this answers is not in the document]
mikhailshilkov: I feel pretty strong about the separate assembly here. Versioning in line with Pulumi core is fine.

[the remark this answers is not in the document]
orionstudt: While it's ultimately up to you guys, as a primarily dotnet dev I agree that the separate assembly is the way to go here.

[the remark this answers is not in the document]
mikhailshilkov: Here is my proposal: https://github.com/pulumi/pulumi/issues/6312
WDYT @orionstudt?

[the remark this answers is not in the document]
orionstudt: I hadn't considered the difficulty of having a single solution that is only partially pre-release - but splitting it into a separate solution sounds like it would resolve that. Sounds good to me. How do you want me to handle that in this PR? Maybe just put an `<IsPackable>false` in the automation .csproj until you guys are ready to do that in a separate PR?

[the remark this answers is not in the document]
orionstudt: I was going to suggest maybe just putting `<VersionSuffix>preview` in the automation .csproj until it is out of pre-release but I think the fact that you guys are calling `dotnet build dotnet.sln /p:VersionPrefix=${VERSION_PREFIX} /p:VersionSuffix=${VERSION_SUFFIX}` everytime would take priority over the the attribute in the .csproj.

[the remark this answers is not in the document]
mikhailshilkov: Disabling `IsPackable` in this PR makes sense to me. I'll follow-up with the solution split and a Makefile amendment.

orionstudt:
alright a final recap: @EvanBoyle @mikhailshilkov 

I believe the only thing we're waiting for here (that I'm aware of) is deciding on final naming of the `Pulumi.Automation.XStack`.

I realize now that the automation API examples are in a different repo, so I can make a PR to complete some of those once this has merged.

And finally, I noticed #6257 and went to add the `--limit` support to this implementation but then realized that obviously locally I can only test against released CLI versions, and that has not been released yet. Idk if your build server tests against unreleased CLI? If so we can include that here too, it's very minimal.

sdk/dotnet/Pulumi.Automation/XStack.cs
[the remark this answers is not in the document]
orionstudt: Having returned to this I think of the 3 suggested here that I prefer `WorkspaceStack`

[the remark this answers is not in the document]
mikhailshilkov: I agree

sdk/dotnet/Pulumi.Automation/WorkspaceStack.cs
```diff
@@ -517,17 +517,34 @@ public Task RemoveConfigAsync(IEnumerable<string> keys, CancellationToken cancel
...
+            var args = new List<string>()
+            {
+                "history",
+                "--json",
+                "--show-secrets",
+            };
+
+            if (limit.HasValue)
+            {
+                args.Add("--limit");
```
EvanBoyle, line 535: Nice catch! I went back before we released this change and exposed the full set of pagination options via https://github.com/pulumi/pulumi/pull/6292

This should now be `--page-size` and `--page`

orionstudt: missed that, should be good now!

mikhailshilkov (approved):
LGTM! Thank you!

PR #5762, opened 2020-11-13
Ensure resource docs have a newline after description but before examples to render correctly

stack72:


komalali (approved):

stack72:
Still a known issue with the Kubernetes resource docs until there's a new tag made - this is due to a missing makefile target in the tag that was fixed *after* the release

PR #5767, opened 2020-11-14
add support for installing plugins from servers

jaxxstorm:
I've chosen to implement this as a new function, but I'm open to
feedback on a better way to support this

### DIFF
--- a/sdk/go/x/auto/local_workspace.go
+++ b/sdk/go/x/auto/local_workspace.go
@@ -382,8 +382,10 @@ func (l *LocalWorkspace) InstallPlugin(ctx context.Context, name string, version
 }
 
 // InstallPluginWithServer acquires the plugin matching the specified name and version with a path to a plugin server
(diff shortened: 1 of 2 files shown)


EvanBoyle (commented):
Interesting approach, but I'm concerned that this strategy may not scale in terms of API design. Perhaps we should create an optional args bag as the last parameter? I have a feeling we will continue to add more knobs here as we flesh out gallery and 3p multi-lang components. 

for typescript:

```ts
    async installPlugin(name: string, version: string, kind = "resource", args?: PluginArgs ): Promise<void> {
        let cliArgs = ["plugin", "install", kind, name, version];
        if(args) {
        cliArgs.push("--server");
        cliArgs.push(server);
        }
        await this.runPulumiCmd(cliArgs);
    }

export interface PluginArgs {
    server?: string;
}
```

Go will require a more involved incantation to use functional options. You can use `LocalWorkspaceOptions` as an example.

emiliza:
Closing out this stale PR. Will revisit this as part of https://github.com/pulumi/pulumi/issues/8263

Issue #5626, opened 2020-10-26
Change order of resource type and name in policy output


clstokes:
The display of policy violations includes the resource information `<resource_name>: <resource_type>`. The order of these is opposite how these are displayed in the preview and update output as well as in how they are written in the application itself. 

#### Proposal

Change the order to `<resource_type>: <resource_name>`.

#### Screenshot

![image](https://user-images.githubusercontent.com/293763/97210599-0b201400-177b-11eb-94d6-14930f6c56ec.png)

PR #5773, opened 2020-11-17
Small UX change to the output of policy violations

stack72:
Fixes: #5626

It used to be:

```
Policy Violations:
    [advisory]  aws v0.1.20200912  allowed-image-owner (demo-aws-ts-webserver-server-0: aws:ec2/instance:Instance)
    Check machine image is from an approved publisher.
    Publisher [137112412989] is not one of [self,099720109477].

```

Notice that it was name: type

We would rather this was type: name

```
Policy Violations:
    [advisory]  aws v0.1.20200912  allowed-image-owner (aws:ec2/instance:Instance: demo-aws-ts-webserver-server-0)
    Check machine image is from an approved publisher.
    Publisher [137112412989] is not one of [self,099720109477].

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,9 @@ CHANGELOG
 - Correct error message on KeyNotFoundException against StackReference.
   [#5740](https://github.com/pulumi/pulumi/pull/5740)
 
+- [cli] Small UX change on the policy violations output to render as `type: name`
+  [#5773](https://github.com/pulumi/pulumi/pull/5773)
+
 ## 2.13.2 (2020-11-06)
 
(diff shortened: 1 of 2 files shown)


clstokes:
The new output looks great, thanks!

lukehoban (approved):

justinvp (approved):
LGTM

We're likely going to need to update tests in https://github.com/pulumi/pulumi-policy/blob/master/tests/integration/integration_test.go and https://github.com/pulumi/pulumi-policy-aws/tree/master/integration-tests that check the CLI output.

stack72:
Thanks for that @justinvp will open an issue now and track getting those changes made!

Issue #5388, opened 2020-09-16
Windows 10: Failed to locate any of ["python3" "python"] on your PATH


trevorkennedy:
When I try to run `pulumi new aws-python` on Windows 10, I get the error `Failed to locate any of ["python3" "python"] on your PATH.  Have you installed Python 3.6 or greater?` 

I can type `python --version` into the command prompt which returns `Python 3.7.9` so this do the trick...

## Expected Behavior
It should create the project scaffolding

## Current Behavior
Displays the `Failed to locate any of ["python3" "python"] on your PATH.  Have you installed Python 3.6 or greater` error message

## Steps to Reproduce

1. python --version
2. mkdir quickstart
3. cd quickstart
4. pulumi new aws-python

## Context (Environment)
Windows 10 with Python installed (from the Microsoft Store) to C:\Users\USERNAME\AppData\Local\Microsoft\WindowsApps

lukehoban:
Heard another report of this - definitely seems like something is off here - possibly something changed on the windows/Python side?

> I did have one "interesting" hiccup getting started that maybe could be considered a bug, but really is just some strange Windows behavior. I have (well had) Microsoft's store version of Python on my machine. It seems that the Windows store has some funky protection with its installed files. During the initial project creation, I was getting this error. Turns out that the exec.LookPath above that is failing during the chkStat call due to the binary not being readable. (If you want to try to replicate, just run in powershell `Get-Content python.exe` on the store delivered binary.)

moerketh:
Found some more information on [this SO post](https://superuser.com/questions/1437590/typing-python-on-windows-10-version-1903-command-prompt-opens-microsoft-stor) and [this blog](https://www.tiraniddo.dev/2019/09/overview-of-windows-execution-aliases.html)

My understanding is that path %LOCALAPPDATA%\Microsoft\WindowsApps contains Windows Execution Aliases appearing as 0 byte files. This is on the PATH by default, so that when a user types the command such as `python` the windows store app is launched allowing the user to install this app. When the app is installed, the alias is changed to run the app instead. Magic.

lukehoban:
We are working on a fix for this in #5775.  It appears there are some fundamental bugs in Go's handling of these special 0-length reparse-point files, which prevents them from being `exec`d via Go's core libraries.  We are looking into possible workarounds for this to unblock this in the meantime.   

For now though, we recommend using a python.org installation of Python on Windows if possible.

viveklak:
[...]

PR #5775, opened 2020-11-18
Check for python among execution aliases on Windows

viveklak:
Fixes https://github.com/pulumi/pulumi/issues/5388

### DIFF
--- a/sdk/python/python.go
+++ b/sdk/python/python.go
@@ -73,23 +73,44 @@ func Command(arg ...string) (*exec.Cmd, error) {
 // called an execution alias. This directory is also added to the PATH.
 // See https://www.tiraniddo.dev/2019/09/overview-of-windows-execution-aliases.html
 // for an overview.
+// Most of this code is a replacement of the windows version of exec.LookPath
+// but uses os.Lstat instead of an os.Stat which fails with a
+// "CreateFile <path>: The file cannot be accessed by the system".
 func resolveWindowsExecutionAlias(pythonCmds []string) (string, error) {
+	exts := []string{""}
+	x := os.Getenv(`PATHEXT`)
+	if x != "" {
+		for _, e := range strings.Split(strings.ToLower(x), `;`) {
+			if e == "" {
+				continue
+			}
+			if e[0] != '.' {
+				e = "." + e
+			}
+			exts = append(exts, e)
+		}
+	} else {
+		exts = append(exts, ".com", ".exe", ".bat", ".cmd")
+	}
+
 	path := os.Getenv("PATH")
 	for _, dir := range filepath.SplitList(path) {
 		if !strings.Contains(strings.ToLower(dir), filepath.Join("microsoft", "windowsapps")) {
 			continue
 		}
 		for _, pythonCmd := range pythonCmds {
-			path := filepath.Join(dir, pythonCmd)
-			f, err := os.Stat(path)
-			if err != nil && !os.IsNotExist(err) {
-				return "", errors.Wrap(err, "evaluating python execution alias")
-			}
-			if os.IsNotExist(err) {
-				continue
-			}
-			if f.Size() == 0 {
-				return path, nil
+			for _, ext := range exts {
+				path := filepath.Join(dir, pythonCmd+ext)
+				f, err := os.Lstat(path)
+				if err != nil && !os.IsNotExist(err) {
+					return "", errors.Wrap(err, "evaluating python execution alias")
+				}
+				if os.IsNotExist(err) {
+					continue
+				}
+				if f.Size() == 0 {
+					return path, nil
+				}
 			}
 		}
 	}
--- a/sdk/python/python.go
+++ b/sdk/python/python.go
@@ -51,12 +51,50 @@ func Command(arg ...string) (*exec.Cmd, error) {
 		}
 	}
 	if err != nil {
-		return nil, fmt.Errorf(
-			"Failed to locate any of %q on your PATH.  Have you installed Python 3.6 or greater?",
-			pythonCmds)
+		// second-chance on windows for python being installed through the Windows app store.
+		if runtime.GOOS == windows {
+			pythonPath, err = resolveWindowsExecutionAlias(pythonCmds)
+		}
+		if err != nil {
+			// TODO: Wrap error? The resulting user facing message isn't very pretty though.
+			return nil, errors.Errorf(
+				"locating any of %q on your PATH.  Have you installed Python 3.6 or greater?",
+				pythonCmds)
+		}
+	}
+	return exec.Command(pythonPath, arg...), nil
+}
+
+// resolveWindowsExecutionAlias performs a lookup for python among UWP
+// application execution aliases which exec.LookPath() can't handle.
+// Windows 10 supports execution aliases for UWP applications. If python
+// is installed using the Windows store app, the installer will drop an alias
+// in %LOCALAPPDATA%\Microsoft\WindowsApps which is a zero-length file - also
+// called an execution alias. This directory is also added to the PATH.
+// See https://www.tiraniddo.dev/2019/09/overview-of-windows-execution-aliases.html
+// for an overview.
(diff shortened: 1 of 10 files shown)


sdk/python/python.go
```diff
@@ -51,14 +51,73 @@ func Command(arg ...string) (*exec.Cmd, error) {
...
 	if err != nil {
-		return nil, fmt.Errorf(
-			"Failed to locate any of %q on your PATH.  Have you installed Python 3.6 or greater?",
-			pythonCmds)
+		// second-chance on windows for python being installed through the Windows app store.
+		if runtime.GOOS == windows {
+			pythonPath, err = resolveWindowsExecutionAlias(pythonCmds)
+		}
+		if err != nil {
+			// TODO: Wrap error? The resulting user facing message isn't very pretty though.
```
justinvp, line 59: I'm fine not wrapping to keep the user facing message nice. Please remove the TODO.

justinvp (approved):
Aside from removing the TODO comment, LGTM. Were you able to validate the change works on Windows 10?

viveklak:
Failed to update this - sorry. Essentially while the above code gets past the resolution issue, we are unable to launch the windows app store's version of python because Golang's exec implementation performs a `os.Stat` which doesn't seem to support `syscall.FILE_FLAG_OPEN_REPARSE_POINT` - which execution aliases use. `os.Lstat` does support it so I was able to bypass it in my change here but can't find a reasonable way to bypass `os.exec`'s use of `os.Stat`. I have a repro which I need to file with the Go team. Will file and update this PR.

lukehoban:
Is there a way to at least "fail better" for now in this case?  Can we identify when the target file is one of these reparse points and give a more helpful error?

justinvp:
> but can't find a reasonable way to bypass `os.exec`'s use of `os.Stat`

If copying our own implementation of `os.exec` with workaround is unpalatable in the meantime, I'm wondering if we could use some kind of shim as a workaround, e.g. ship an extra batch script on Windows that we can call from `os.exec`, which then calls the app store version of Python we found?

lukehoban:
FWIW - A few more notes from trying this out on a Windows machine.

The following program is a repro of the core issue:

```go
package main

import (
	"fmt"
	"os"
	"os/exec"
)

func do() error {
	res, err := os.Lstat("C:\\Users\\lukeh\\AppData\\Local\\Microsoft\\WindowsApps\\python3.exe")
	if err != nil {
		return err
	}
	fmt.Printf("%v\n", res)
	cmd := exec.Command("C:\\Users\\lukeh\\AppData\\Local\\Microsoft\\WindowsApps\\python3.exe", "-c", "print('hello')")
	btes, err := cmd.CombinedOutput()
	if err != nil {
		return err
	}
	fmt.Printf("%v\n", btes)
	return nil
}

func main() {
	err := do()
	if err != nil {
		panic(err)
	}
}
```

This results in:

```
&{python3.exe 1056 {2654840238 30774702} {2957402258 30850392} {2957402258 30850392} 0 0 2147483675 0 {0 0}  2853655769 458752 49311 false}
panic: exec: "C:\\Users\\lukeh\\AppData\\Local\\Microsoft\\WindowsApps\\python3.exe": file does not exist

goroutine 1 [running]:
main.main()
        C:/Users/lukeh/goexecalias/main.go:27 +0x4b
exit status 2
```

Even though it definitely does exist:

```
PS C:\Users\lukeh\goexecalias> C:\\Users\\lukeh\\AppData\\Local\\Microsoft\\WindowsApps\\python3.exe -c "print('hello')"
hello
```

The `1056` file attribute indicates this is an archive and a reparse point.  The `2147483675` in `Reserved0` apparently indicates this is (see https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-fscc/c8e77b37-3909-4fe6-a4ea-2b9d423b1ee4):

IO_REPARSE_TAG_APPEXECLINK0x8000001B | Used by Universal Windows Platform (UWP) packages to encode information that allows the application to be launched by CreateProcess. Server-side interpretation only, not meaningful over the wire.
-- | --

viveklak:
> Is there a way to at least "fail better" for now in this case? Can we identify when the target file is one of these reparse points and give a more helpful error?

I am definitely leaning towards this option! Thanks for the dive into the flags @lukehoban. I will update this PR to be a guard and look to file the Go bug.

joeduffy:
Can we construct the `*exec.Cmd` struct directly, bypassing the `LookPath` in the `exec.Command` constructor? For instance, `cmd := &exec.Cmd{Path: "manually populated", Args: ...}`. It may require some Windows specific code to resolve the Windows Execution Alias metadata, but heck, we can even just hardcode the `python` case for now since it's the only one causing us troubles right now (until the Go libraries themselves are fixed to handle this).

Is there a bug in the https://github.com/golang/go repo, by the way? I didn't see one.

lukehoban:
> Is there a bug in the https://github.com/golang/go repo, by the way? I didn't see one.

@viveklak was planning to open one.  We have been unable to find anything tracking this issue.

> Can we construct the *exec.Cmd struct directly, bypassing the LookPath in the exec.Command constructor?

Unfortunately no - that's not the one that fails - the problem is ultimately here: https://github.com/golang/go/blob/master/src/os/exec/exec.go#L352 (and we could possibly avoid that one, but it doesn't look like we can avoid at least one of the `LookPath` calls there).  AFAICT - there is no way to use the existing `exec` package without hitting this.  And it looks like we would have to port over a copy of the majority of that package (including a lot of internals) to avoid this without a fix to the packge.

lukehoban:
Actually - it does look like we could use ` os.StartProcess` directly instead of the `exec` package to workaround this.

viveklak:
I cut https://github.com/golang/go/issues/42919 with the repro and details. Investigating some of the suggestions above next.

viveklak:
> Actually - it does look like we could use ` os.StartProcess` directly instead of the `exec` package to workaround this.

Confirmed that `os.RunCommand` does indeed work at least as far as executing the interpreter. I will try and fit it in tomorrow and update this PR accordingly.

viveklak:
@justinvp @lukehoban I wrapped the `exec.Cmd` usage in an adapter that can fall back to `os.StartProcess` in the specific situation affected by https://github.com/golang/go/issues/42919. This turned into a bit of a yak shave but the changes in https://github.com/pulumi/pulumi/pull/5775/commits/47525eabfa882e42fea673e376acff25b7f9c4a1 are as minimal as I could get it to make `os.StartProcess` actually work for our various use-cases.

I tested this on a windows machine and could confirm that it works.

sdk/python/python.go
```diff
@@ -27,9 +27,51 @@ import (
...
+		cmd.cmd.Stdout = stdout
+	})
+}
+
+func WithStdErr(stderr *os.File) CommandOpt {
+	return commandOptFunc(func(cmd *wrappedCmd) {
+		cmd.stderr = stderr
+		cmd.cmd.Stderr = stderr
+	})
+}
```
lukehoban, line 70: I didn't exactly follow why we needed to add `CommandOpt` here - vs. just relying on string varargs as before?

sdk/python/wrappedcmd.go
```diff
@@ -0,0 +1,290 @@
...
+	"io"
+	"os"
+	"os/exec"
+	"runtime"
+	"strings"
+	"syscall"
+)
+
+// WrappedCmd wraps the key portions of exec.Cmd used
+// to launch the python interpreter. Large portions of this
```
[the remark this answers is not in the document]
lukehoban: Also - perhaps a `TODO` to remove this all once https://github.com/golang/go/issues/42919 is fixed just so we record that this is all an ugly workaround for that bug?

lukehoban (approved):
Would love @justinvp's review as well - but LGTM.

justinvp (approved):
LGTM

justinvp:
Ah, sorry, @viveklak, my other change introduced some conflicts here. Let me know if you have any questions.

viveklak:
Closing in favor of #5874.

Issue #5593, opened 2020-10-19
`pulumi plugin install --reinstall` doesn't work for Python projects


clstokes:
#### Problem

`pulumi plugin install --reinstall` doesn't work for Python projects. Running with verbose output seems to indicate zero required plugins were found.

Looking through the code I suspect this `TODO` is the issue.

https://github.com/pulumi/pulumi/blob/89c956d18942c1fcbf687da3052dd26089d8f486/sdk/python/cmd/pulumi-language-python/main.go#L144-L149

#### Verbose Output

```
% pulumi plugin install --reinstall -v9 --logtostderr
I1019 16:15:34.901943   11777 pulumi.go:129] skipping update check
I1019 16:15:34.906450   11777 plugins.go:470] GetPluginPath(language, python, <nil>): found on $PATH /usr/local/bin/pulumi-language-python
I1019 16:15:34.906923   11777 plugin.go:83] Launching plugin 'python' from '/usr/local/bin/pulumi-language-python' with args: 127.0.0.1:61623
I1019 16:15:34.932973   11777 langruntime_plugin.go:178] langhost[python].GetPluginInfo() executing
I1019 16:15:34.934407   11777 langruntime_plugin.go:91] langhost[python].GetRequiredPlugins(proj=aws-py-fargate,pwd=/Users/clstokes/go/src/github.com/pulumi/examples/aws-py-fargate,program=.) executing
I1019 16:15:34.935436   11777 langruntime_plugin.go:133] langhost[python].GetRequiredPlugins(proj=aws-py-fargate,pwd=/Users/clstokes/go/src/github.com/pulumi/examples/aws-py-fargate,program=.) success: #versions=0
I1019 16:15:34.935522   11777 plugins.go:470] GetPluginPath(language, python, <nil>): found on $PATH /usr/local/bin/pulumi-language-python
```

justinvp:
Likely related to https://github.com/pulumi/pulumi/issues/4848

Issue #5164, opened 2020-08-11
Python: Automatically create virtualenv when specified in Pulumi.yaml and the dir is missing


justinvp:
When running `pulumi preview` or `pulumi up` on Python programs, if the `virtualenv` option is specified in `Pulumi.yaml` and the specified dir does not exist, we currently show an error with the commands to run to create the virtual environment manually:

```
error: The 'virtualenv' option in Pulumi.yaml is set to "venv", but "/path/to/project/venv" doesn't exist; run the following commands to create the virtual environment and install dependencies into it:

1. python3 -m venv /path/to/project/venv
2. /path/to/project/venv/bin/python -m pip install --upgrade pip setuptools wheel
3. /path/to/project/venv/bin/python -m pip install -r requirements.txt


For more information see: https://www.pulumi.com/docs/intro/languages/python/#virtual-environments
```

Instead, we should automatically/transparently create the virtual environment and install dependencies into it, so users don't have to run any additional commands.

This would make Python projects behave closer to Go/.NET, which automatically take care of acquiring deps. And it will make it easier for folks to get started using existing Pulumi programs that have this option specified in `Pulumi.yaml` (e.g. clone example, run `pulumi up`, and the CLI automatically takes care of creating the virtual env and installing deps). And it would allow us to get rid of [this](https://github.com/pulumi/pulumi/blob/c9d8549ae657ac882e6ec34fae0cea16944d9f2b/dist/actions/entrypoint.sh#L120-L131) and [this](https://github.com/pulumi/pulumi-kubernetes-operator/blob/29ab7bf0133f2b9e02087033a4dccee710d737bf/pkg/controller/stack/stack_controller.go#L602-L636).

Issue #4848, opened 2020-06-17
Consider implementing `GetRequiredPlugins` for Python


justinvp:
As of #2560, the engine will attempt to install any missing plugins during `pulumi up` by scanning the snapshot and asking the language host for required plugins. This is a best effort as not all language hosts implement `GetRequiredPlugins`, in particular Python:

https://github.com/pulumi/pulumi/blob/2d61852e4f5577f2b42e87d9ce2dcfc85da452b6/sdk/python/cmd/pulumi-language-python/main.go#L144-L149

Some folks are still experiencing errors due to missing plugins when Python dependencies are installed before Pulumi is installed (e.g. in CI scenarios like https://github.com/pulumi/pulumi-azure/issues/601).

The missing plugin errors wouldn't happen for Python programs if `GetRequiredPlugins` was implemented. I don't see any particular reason why we couldn't implement it for Python.

The way it's implemented for .NET is by running `dotnet list package --include-transitive` and parsing the output.

For Python, we could do something similar. If we assume that pip is being used (seems like a reasonable assumption) we could run `python -m pip list` and parse the output:

```sh
$ venv/bin/pip list
Package    Version
---------- -------
Arpeggio   1.9.2  
attrs      19.3.0 
dill       0.3.2  
grpcio     1.29.0 
parver     0.3.0  
pip        19.2.3 
protobuf   3.12.2 
pulumi     2.4.0  
pulumi-aws 2.9.1  
semver     2.10.2 
setuptools 41.2.0 
six        1.15.0
```

Maybe we'd only support it when the `virtualenv` runtime option is set in `Pulumi.yaml`, as then we'd have stronger idea that `pip` is being used? But that may limit its usefulness as it will be some time before `virtualenv` is adopted broadly.

CortalUX:
Yeah I'm not using the virtualenv option - mainly because as mentioned, I'm using cloud hosted VMs which are 'clean' on setup, then destroyed after. Although - I do use virtual environments on my local machine, and I would use the virtualenv option if it was required for this to work :)

justinvp:
`python -m pip list --format json` outputs JSON

PR #5787, opened 2020-11-19
Implement GetRequiredPlugins for Python

justinvp:
This PR implements `GetRequiredPlugins` for Python, which is called by the engine at the beginning of `pulumi preview|up` to determine/install any missing plugins needed by the program.

This is implemented by running `python -m pip list --format json` to determine the installed Pulumi packages (packages with names that start with `pulumi-`). If the `virtualenv` runtime option is set in `Pulumi.yaml`, it runs the `python` inside the virtual environment. Otherwise, it calls `python` in the usual ways, which works with other tools that manage their own virtual environment (such as `poetry run pulumi up`).

The required plugins are determined based on the Pulumi packages. If a package contains an optional `pulumiplugin.json` file, the information in this file is used to determine the associated plugin: the `name`, `version`, and `server`. There's also a `resource` option that must be set to `true` to indicate this package has an associated plugin, or `false` to indicate the package does not have an associated plugin (e.g. I'd imagine we'd set this to `false` for `pulumi-policy`, in the meantime we hardcode this package as a known package that does not have a plugin). This file is similar to information included in `package.json` for Node.js and `version.txt` in .NET NuGet packages (though, `version.txt` does not currently support specifying a `server`). If the `pulumiplugin.json` file doesn't exist, we attempt to determine the plugin name and version based on the package name and version. This works well for all released packages we have today (which don't have a `pulumiplugin.json` yet), but does not work for alpha/beta/rc/dev builds of these packages because the package version does not include enough information to determine the plugin version for these cases. A commit is included with this PR, which adds support for generating the `pulumiplugin.json` file for packages (via a new schema option). I suspect we'll want to roll this out for all our packages, so `GetRequiredPlugins` will work for dev builds.

This PR also implements support for automatically creating the virtual environment and installing dependencies into it at the beginning of `pulumi preview|up` if the `virtualenv` option is set in `Pulumi.yaml` and the specified directory is missing or empty.

I'd love to add more automated testing around this, but may have to do that as a follow-up item (I don't believe we have much automated testing around similar functionality for the other languages).

Fixes https://github.com/pulumi/pulumi/issues/4848
Fixes https://github.com/pulumi/pulumi/issues/5164
Fixes https://github.com/pulumi/pulumi/issues/5593


pkg/backend/httpstate/policypack.go
```diff
@@ -321,16 +321,16 @@ func completeNodeJSInstall(finalDir string) error {
...
-		}
-		return nil
-	}); err != nil {
+	const venvDir = "venv"
+	if err := python.InstallDependencies(finalDir, venvDir, false /*showOutput*/); err != nil {
 		return err
 	}
 
+	// Save project with venv info.
+	proj.Runtime.SetOption("virtualenv", venvDir)
```
stack72, line 330: now that this is a requirement - does this mean that the user must specify this in their Pulumi.yaml?

justinvp: It’s not a requirement. There isn’t any user-facing change with this modification.

pkg/codegen/python/importer.go
```diff
@@ -42,6 +42,8 @@ type PackageInfo struct {
 	Compatibility string `json:"compatibility,omitempty"`
 	// Indicates whether the package generates input/output classes.
 	UsesIOClasses bool `json:"usesIOClasses,omitempty"`
+	// Indicates whether the pulumiplugin.json file should be generated.
+	EmitPulumiPluginFile bool `json:"emitPulumiPluginFile,omitempty"`
```
lukehoban, line 46: Any particular reason we don't want to do this by default? (Vs manually updating all providers to include this?)

sdk/python/cmd/pulumi-language-python/main.go
```diff
@@ -144,8 +148,365 @@ func newLanguageHost(exec, engineAddress, tracing, virtualenv string) pulumirpc.
...
+		Severity:  w.severity,
+	}); err != nil {
+		return 0, err
+	}
+	return len(val), nil
+}
+
+// These packages are known not to have any plugins.
+var packagesWithoutPlugins = map[string]struct{}{
+	"pulumi-policy": {},
```
lukehoban, line 291: Should we TODO this to undo the hard coding?

pkg/codegen/python/gen.go
```diff
@@ -1215,8 +1216,34 @@ func sanitizePackageDescription(description string) string {
 	return ""
 }
 
+func genPulumiPluginFile(pkg *schema.Package) (string, error) {
+	type pulumiPlugin struct {
```
lukehoban, line 1220: Can we define and version this serialization format in a single place?  I see this is re-defined in the Python language host implementation.

sdk/python/cmd/pulumi-language-python/main.go
```diff
@@ -144,8 +148,365 @@ func newLanguageHost(exec, engineAddress, tracing, virtualenv string) pulumirpc.
...
+		return nil, errors.Wrapf(err, "parsing `python %s` output", strings.Join(args, " "))
+	}
+
+	// Only return Pulumi packages.
+	var pulumiPackages []pythonPackage
+	for _, pkg := range packages {
+		// We're only interested in packages that start with "pulumi-".
+		if !strings.HasPrefix(pkg.Name, "pulumi-") {
+			continue
+		}
```
lukehoban, line 321: This is temporary - right?  In principle, we will want to support 3rd party packages with plugins?

lukehoban (approved):
> This file is similar to information included in package.json for Node.js and version.txt in .NET NuGet packages (though, version.txt does not currently support specifying a server).

Should we be considering this as a standard format to use across these different language ecosystems?

sdk/python/cmd/pulumi-language-python/main.go
justinvp: Yes, temporary. Go and .NET have this same limitation:

https://github.com/pulumi/pulumi/blob/002a228000bdcb10cc9851cc8bce09858065058b/sdk/go/pulumi-language-go/main.go#L140

https://github.com/pulumi/pulumi/blob/002a228000bdcb10cc9851cc8bce09858065058b/sdk/dotnet/cmd/pulumi-language-dotnet/main.go#L251-L252

As we work on making it possible/easier to publish 3rd party plugins, we'll need to address this across all the languages.

justinvp:
> Should we be considering this as a standard format to use across these different language ecosystems?

We certainly could consider it.

- Node.js probably doesn't need it as there's already a `pulumi` section in `package.json` where this information lives.
- .NET has `version.txt` which indicates the NuGet package has a plugin. The file contains the plugin version ([and also now the name of the plugin](https://github.com/pulumi/pulumi/pull/5629)). We could consider further extending `version.txt` to include additional information like the plugin `server`, or consider moving .NET to `pulumiplugin.json`.
- Go doesn't currently have anything like this. It does a best effort of determining the plugin information based on the go modules. There isn't currently a way to specify a custom `server` for the plugin or explicit name or version. I suspect we could adopt `pulumiplugin.json` for Go.

lukehoban:
> Node.js probably doesn't need it as there's already a pulumi section in package.json where this information lives.

I guess I wonder if we should at least standardize on the JSON payload that we include inside the package.json, so that we have a consistent data model for this, even if it lives in different places for different package managers.  Probably not a huge deal - but it feels like consistency here will help avoid capability drift between language ecosystems.

pkg/codegen/python/importer.go
justinvp: Ah, I recall now why I had this as an option. We're going to need to update the `Makefile` of each provider to replace `${PLUGIN_VERSION}` in `pulumiplugin.json` with the plugin version, just like we do with `setup.py`.

https://github.com/pulumi/pulumi-aws/blob/ed9cd60d861bc2d930150c272fa00fac2f774518/Makefile#L51

justinvp:
> standardize on the JSON payload that we include inside the package.json

I went ahead and did this. I defined a common `PulumiPluginJSON` struct that's used by both Python and now Node.js.

PR #5788, opened 2020-11-20
Allow schemas to reference to provider types.

pgavlin:
These changes extend the type reference parser in the schema package to
accept references of the form "(package/version/schema.json)?#/provider".
These references refer to the package's provider type, which is
otherwise not referenceable, as it is not present in the "resources"
array.


justinvp (approved):
LGTM

PR #5805, opened 2020-11-20
Update Python resource ref deserialization.

pgavlin:
There are two significant changes in this commit: one to the way
resource packages/modules are stored and retrieved, and one to resource
ref deserialization in the face of missing resource packages/modules.

Resource packages and modules no longer require an exact version match
during deserialization. Instead, the newest compatible version of the
package or module is selected. If no version was specified, the newest
version of the package or module will be chosen. As a special case, a
package or module that has no version will always be treated as the best
version for that package or module.

If a resource package or module is not found when attempting to
deserialize a resource reference, the SDK no longer emits an error, and
instead deserializes the reference as its URN or ID (if present). This
accommodates providers that have not yet been updated to include the
appropriate factory registrations.

### DIFF
--- a/sdk/python/Pipfile
+++ b/sdk/python/Pipfile
@@ -8,6 +8,7 @@ protobuf = ">=3.6.0"
 grpcio = ">=1.9.1,!=1.30.0"
 dill = ">=0.3.0"
 six = ">=1.12.0"
+semver = ">=2.8.1"
 
 [dev-packages]
 pylint = ">=2.1"
--- a/sdk/python/Pipfile.lock
+++ b/sdk/python/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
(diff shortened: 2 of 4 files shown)


pgavlin:
cc @mikhailshilkov @viveklak for context. We'll want to make similar updates to the other language SDKs.

lblackstone (approved):

justinvp (approved):
LGTM

pgavlin:
@justinvp any idea why the test don't seem to be pulling down `semver`?

justinvp:
Haven't verified this is the root cause, but looks like our `install_requires` in `setup.py` isn't up-to-date with the `Pipfile`.

https://github.com/pulumi/pulumi/blob/5302a5749c0f8f39a5193c37ea26d1d1a3a4fee9/sdk/python/lib/setup.py#L36-L40

Issue #5807, opened 2020-11-21
Unpleasant output when installing dependencies of a multi-lang plugin


justinvp:
When installing a multi-lang plugin's dependencies (e.g. when running `npm install`), some output of the `npm install` command is displayed, which doesn't look great combined with the progress bar:

```
$ pulumi plugin install resource eks v0.20.1-alpha.1605897480+2d045387.dirty
[resource plugin eks-0.20.1-alpha.1605897480+2d045387.dirty] installing
[...]

Issue #5747, opened 2020-11-12
`pulumi preview --json` shows plugin installation output


EvanBoyle:
A JSON preview still displays plugin acquisition output. This breaks the ability to parse the JSON output as you have a bunch of unstructured text at the beginning of the output. The `--json` flag should disable all other calls to stdout. To repro: 

`pulumi plugin rm --all && pulumi preview --json`

EvanBoyle:
To workaround this for Automation API, you can use  `workspace.InstallPlugin` to install plugins ahead of time:

- go: https://pkg.go.dev/github.com/pulumi/pulumi/sdk/v2/go/x/auto#example-LocalWorkspace.InstallPlugin
- node: https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/pulumi/x/automation/#LocalWorkspace-installPlugin

joeduffy:
I should note that it's also just bad stdout/stderr hygiene to be emitting things like plugin download status to stdout.

lukehoban:
I believe there are two options here:
1. Render plugin installation status only to debug logs (`logging.V(preparePluginVerboseLog).Infof`)
2. Render plugin installation status to diag sink (`sink.Infof`)

With either option - we likely have to change the style of the rendering - as it's not clear we can use the ProgressBar approach in either of those approaches.

I'm not entirely clear on what the user experience would be for (2).  It will in part depend on what event stream renderer is being applied - and I suspect it would be fairly natural in all existing renderers (not shown at all for `pulumi preview --json`, and shown as an ephemeral status on the stack resource for the default update status table rendering).  I am not sure though whether common cases of plugin installation happen too early for this rendering to work well. I expect we'll need to try this out and see what works well from a user experience perspective.

(1) is simpler, but hides a potential source of significantly delays.  Users could frequently (but semi-randomly) sit waiting for 2-10s with no indication of progress.  I'm inclined to try to do (2) if possible.

EvanBoyle:
It might be a helpful reminder that the only reason I raised this issue is because it breaks parsing `pulumi preview --json` which came up from the Automation API Go SDK. Given that we are planning on removing this from the Automation API via https://github.com/pulumi/pulumi/issues/6308, the priority of fixing this might be lower for the time being. Still good to tackle if it is straightforward.

komalali:
[...]

PR #5808, opened 2020-11-21
Improve the output when installing plugins

justinvp:
Two improvements:

1. Don't display "[resource plugin <foo>] installing" if the plugin is already installed.

2. Close the plugin download progress bar before displaying any subsequent output, and only show output of `npm install` when there is an error.

Fixes https://github.com/pulumi/pulumi/issues/5807

### DIFF
--- a/sdk/go/common/workspace/plugins.go
+++ b/sdk/go/common/workspace/plugins.go
@@ -15,6 +15,7 @@
 package workspace
 
 import (
+	"bytes"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


sdk/go/common/workspace/plugins.go
```diff
@@ -380,7 +386,9 @@ func (info PluginInfo) Install(tarball io.ReadCloser) error {
 		// TODO[pulumi/pulumi#1334]: move to the language plugins so we don't have to hard code here.
 		switch runtime {
 		case "nodejs":
-			if _, err := npm.Install(finalDir, nil, os.Stderr); err != nil {
+			var b bytes.Buffer
+			if _, err := npm.Install(finalDir, &b, &b); err != nil {
+				os.Stderr.Write(b.Bytes())
```
lukehoban, line 391: Any chance we can easily fix https://github.com/pulumi/pulumi/issues/5747 while we are here?

lukehoban (approved):

sdk/go/common/workspace/plugins.go
justinvp: I could make some additional targeted changes here to work towards fixing #5747, but I think there are some other places that would need to be fixed (not showing the progress bar, other places we output "installing" messages, etc.), so would rather hold off and tackle that altogether to ensure it's truly addressed.

Issue #4463, opened 2020-04-21
Pulumi.yaml file renamed to wrong value after 'stack rename'


chrsmith:
When you rename a stack managed within a Pulumi organization, you need to qualify it with the organization name.

For example, the following would fail because the rename step would assume the target organization would be the currently logged in user and not the `pulumi` organization.

```
pulumi stack init pulumi/hosted-stack
pulumi stack rename hosted-stack-renamed
```

So you need to provide a qualified stack name. The following will work as expected:

```
pulumi stack rename pulumi/hosted-stack-renamed
```

The problem comes in that as part of the rename operation we also rename the `Pulumi.<stack-name>.yaml` file. When you use `pulumi stack rename` with a qualified stack name, we generate a file `Pulumi.pulumi-hosted-stack-renamed.yaml`. (That is, including the "organization" in the new `.yaml` file name.) However, that won't work and essentially means that we erase all the configuration data for the stack. (Since `pulumi` will look for `Pulumi.hosted-stack-renamed.yaml` and fail to find the file.)

Found this out the hard way with https://github.com/pulumi/get.pulumi.com/pull/58/commits/3937926e5b824052dcec883eb5285a4d112a0757.

Issue #5731, opened 2020-11-11
Breaking changes to consider for Pulumi 3.0


leezen:
#### Confirmed for 3.0

- [x] https://github.com/pulumi/pulumi/issues/6168
- [x] https://github.com/pulumi/pulumi/issues/6592
- [x] https://github.com/pulumi/pulumi/issues/5257
- [x] https://github.com/pulumi/pulumi/issues/3151

#### Under discussion for 3.0

- [ ] https://github.com/pulumi/pulumi-terraform-bridge/issues/315

#### Deferred

- https://github.com/pulumi/pulumi/issues/2484
- https://github.com/pulumi/pulumi/issues/5941

joeduffy:
I was hoping we would have tackled https://github.com/pulumi/pulumi-terraform-bridge/issues/10 by now. If we're not doing so because it's a breaking change (my vote is still to take it since it's security related), we should absolutely consider that as part of the 3.0 milestone.

lukehoban:
> I was hoping we would have tackled pulumi/pulumi-terraform-bridge#10 by now

It is still currently planned to deliver in the next few weeks as a non-breaking change ahead of 3.0. If we discover we cannot keep the breaking change nature of this constrained enough to feel comfortable shipping it now (or require shipping it under a flag), we will definitely do it no later than 3.0.

leezen:
Closing out this tracking issue from 3.0.

Issue #4605, opened 2020-05-11
Improve stability and functionality of self-managed state backends


leezen:
This epic tracks overall improvements to non-Pulumi Console [backends](https://www.pulumi.com/docs/intro/concepts/state/) to make them more reliable and feature-complete.

- [ ] #2055 Fix code duplication between httpstate and filestate backends
- [x] #2522 Project name in stack identity inconsistent between backends
- [ ] #2801 Failed to load checkpoint after write failure to S3 backend
- [x] #2814 Using S3 backend with multiple projects is unpleasant
- [ ] #2996 Feature request: persist the filestate backend URL in Pulumi.<stackname>.yaml
- [x] #3194 Specifying an incorrect API service backend for pulumi login results in a confusing error message
- [x] #3395 StackReference doesn't support multiple backends
- [ ] #3683 Be able to move/transfer a stack from a backend to another
- [ ] #3714 [bug] s3 backend deletes state early and then tries to JSON parse html
- [ ] #3760 Add test coverage for all supported filestate backends
- [x] #3120 Simple project support for local backends
- [ ] #2058 Feature: Store release history in Kubernetes using CRDs
- [ ] #4665 Support configuration for pluggable backends and secrets providers
- [ ] #4679 S3 state provider will remove stack data on connection lost while `pulumi up`
- [ ] #5036 Concurrent Stack Updates Results in Reference Error with s3 Backend
- [ ] https://github.com/pulumi/pulumi-az-pipelines-task/issues/47 Read backend config from Pulumi.yaml
- [ ] #6074 Pulumi eats bandwidth (when using S3 backend)
- [ ] #8571 Don't touch local in stack rm
- [x] #9031 Use Pulumi.yaml template as schema to validate Pulumi.$stack.yaml

ig0rsky:
@leezen so what's the proposed way of moving forward here? Is there draft somewhere that is being worked on for us to read?

leezen:
There's a planned item to write a proposal tracked in https://github.com/pulumi/pulumi/issues/4710

joeduffy:
@leezen @stack72 I see that https://github.com/pulumi/pulumi/issues/4710 was closed, but I couldn't find where we posted the design document for community feedback/visibility. Did we do that somewhere?

stack72:
@joeduffy apologies for not posting this - this is the proposal we are going with

# Backend Enhancement Design Proposal

## Goal
[...]

Issue #2522, opened 2019-03-06
Project name in stack identity inconsistent between backends


nesl247:
In https://github.com/pulumi/pulumi/pull/2386 it was changed so that http backends use the project in the stack identity. Why was this only done for http instead of for local backends? This is very inconsistent.

ellismg:
It was just an oversight, but we will be addressing it shortly.  There was already some skew here (the Pulumi service further scopes things by username, where as the local backend does not) but we should keep these in sync as possible.

nesl247:
Curious, does this type of skew not imply that the architecture could be improved? I would think this type of logic would not need to be implemented in each backend, but in a higher layer. The backend would only be responsible for the execution of whatever command it is receiving.

rhyek:
Any workarounds for this? I was planning to use multiple pulumi projects and during local development (using local pulumi state) I hit this road-block.

JakeGinnivan:
We fully qualify our stack names, ie project.env as the stack name. Bit clunky but works fine
________________________________

nguyenxndaidev:
For local development with local pulumi state, I organize my projects follow [Inter-Stack Dependencies](https://www.pulumi.com/docs/intro/concepts/organizing-stacks-projects/#inter-stack-dependencies).

For all related projects, I configure `Pulumi.yaml` to point to the same `.pulumi` directory.

For example:
```
backend:
  url: file://../
```
This will use the `.pulumi` directory in the parent directory.

Then for each stack, I also add environment separated with a dot. For example: `Pulumi.network.dev.yaml`, `Pulumi.services.dev.yaml`, `Pulumi.network.prod.yaml`, `Pulumi.services.prod.yaml`

rafaribe:
I can also hope to see some development into addressing this problem. @nguyenxndaidev 's solution is somewhat of a workaround but feels a bit clunky :)

adriantorrie:
Any word on when this will be resolved?

Want to know how much effort to factor in for establishing our local environments, and our CI/CD workflows.

Workarounds so far are clunky, as they've said, and not sure it would help if we chose to migrate to the pulumi cloud at some point. Seems like it would be a right PITA trying to map to [inter-stack-dependencies](https://www.pulumi.com/docs/intro/concepts/organizing-stacks-projects/#inter-stack-dependencies) which are correctly namespaced when/if a decision to migrate was made.

confiq:
[...]

PR #5812, opened 2020-11-22
Correctly rename stack files during a rename

joeduffy:
This fixes pulumi/pulumi#4463, by renaming a stack's configuration file based on its stack-part, and ignoring the owner-part. Our workspace system doesn't recognize configuration files with fully qualified names. That, by the way, causes problems if we have multiple stacks in different organizations that share a stack-part.

The fix here is simple: propagate the new StackReference from the Rename operation and rely on the backend's normalization to a simple name, and then use that the same way we are using a StackReference to determine the path for the origin stack.

An alternative fix is to recognize fully qualified config files, however, there's a fair bit of cleanup we will be doing as part of https://github.com/pulumi/pulumi/issues/2522 and https://github.com/pulumi/pulumi/issues/4605, so figured it is best to make this work the way the system expects first, and revisit it as part of those overall workstreams. I also suspect we may want to consider changing the default behavior here as part of https://github.com/pulumi/pulumi/issues/5731.

Tests TBD; need some advice on how best to test this since it only happens with our HTTP state backend -- all integration tests appear to use the local filestate backend at the moment.


joeduffy:
@chrsmith thoughts on testing strategy for this? I thought we had SaaS-backend tests in this repo, but I've been away for awhile, and can't seem to find them? Am I imagining this?

lukehoban (commented):
> @chrsmith thoughts on testing strategy for this? I thought we had SaaS-backend tests in this repo, but I've been away for awhile, and can't seem to find them? Am I imagining this?

The tests in https://github.com/pulumi/pulumi/blob/master/tests/stack_test.go should be able to target either backend, and in particular `TestStackRenameAfterCreate`.

> That, by the way, causes problems if we have multiple stacks in different organizations that share a stack-part.

FWIW - this is tracked in https://github.com/pulumi/pulumi/issues/5511.

joeduffy:
> The tests in https://github.com/pulumi/pulumi/blob/master/tests/stack_test.go should be able to target either backend, and in particular TestStackRenameAfterCreate.

I see, so if the test doesn't include an explicit `pulumi login`, it will default to the service? If yes, great, will add some new tests to the stack rename test suite. (All explicit logins seemed to target the local backend, so wasn't sure.)

joeduffy:
Ok, @chrsmith or @lukehoban , PTAL, should be ready to go.

pkg/backend/httpstate/backend.go
```diff
@@ -762,38 +762,51 @@ func (b *cloudBackend) RemoveStack(ctx context.Context, stack backend.Stack, for
...
 	if err != nil {
-		return err
+		return nil, err
 	}
 
 	if stackID.Owner != newIdentity.Owner {
-		url := "."
+		errMsg := "New stack owner, %s, does not match existing owner, %s.\n\n"
+
+		qn, _ := b.parseStackName(string(newName))
```
chrsmith, line 787: It seems like we should delete lines 787-793.

Unless I'm missing something, `newName` doesn't get mutated after line 774, so there is no difference between `newRef` introduced earlier and `qn` defined here. Right?

Similarly, `cloudBackend::ParseStackReference` will always return the fully-qualified stack name (owner/project/name), so it should never be the case that `qn.Owner == ""`. (`cloudBackend::parseStackName` on the other hand will leave the value as "", since it doesn't infer missing values to fully-qualify the name.)

pkg/backend/httpstate/backend.go
```diff
@@ -762,38 +762,51 @@ func (b *cloudBackend) RemoveStack(ctx context.Context, stack backend.Stack, for
...
 		}
-		errMsg := "You cannot transfer stack ownership via a rename. If you wish to transfer ownership\n" +
-			"of a stack to another organization, you can do so in the Pulumi Console by going to the\n" +
-			"\"Settings\" page of the stack and then clicking the \"Transfer Stack\" button" + url
+		errMsg += "       You cannot transfer stack ownership via a rename. If you wish to transfer ownership\n" +
+			"       of a stack to another organization, you can do so in the Pulumi Console by going to the\n" +
+			"       \"Settings\" page of the stack and then clicking the \"Transfer Stack\" button" + url
 
-		return errors.Errorf(errMsg)
+		return nil, errors.Errorf(errMsg, stackID.Owner, newIdentity.Owner)
```
chrsmith, line 803: I'd suggest we just use `errors.New` here, and use `errMsg := fmt.Spritnf(..., stackID.Owner, newIdentity.Owner)` earlier. That way we don't need to "remember" there are 2x values to be filled into the error string later.

pkg/backend/httpstate/backend.go
```diff
@@ -762,38 +762,51 @@ func (b *cloudBackend) RemoveStack(ctx context.Context, stack backend.Stack, for
...
+
+		qn, _ := b.parseStackName(string(newName))
+		if qn.Owner == "" {
+			errMsg += fmt.Sprintf(
+				"       Did you forget to include the owner name? If yes, rerun the command as follows:\n\n"+
+					"           $ pulumi stack rename %s/%s\n\n",
+				stackID.Owner, newName)
+		}
 
+		url := "."
```
chrsmith, line 795: Nit: Rather than `url` being either "." or ":\n\n <console URL>", why don't we rename this to `errMsgSuffix` instead?

That makes it clearer that it isn't actually a URL, but rather some text that gets added to the end of the error message, and so it seems a little less confusing that it gets initialized to ".".

chrsmith (commented):
The change seems simple and straight forward enough, though either I'm missing something or there are a few ways we can simplify/clean this up a bit.

chrsmith (approved):
Anyways, these changes look good assuming you take a look and apply whatever cleanups as appropriate.

Just so I understand, the next step towards cleaning up the "stack owner x project name x Pulumi.project-name.yaml" is to update callers of `RenameStack(...)` to refer to the newly returned `StackReference`, and perhaps be smarter about updating the `Pulumi.*.yaml` file and/or persisting more information inside of that?

pkg/backend/httpstate/backend.go
joeduffy: Happy to do that and leave the code better than it was before -- this was just how it was written before.

pkg/backend/httpstate/backend.go
joeduffy: > cloudBackend::parseStackName on the other hand will leave the value as "", since it doesn't infer missing values to fully-qualify the name.

Yes, that's exactly why we're using `parseStackName` here, so that we can check whether the owner was left off and give a better error message in that case ("did you forget?") -- because `ParseStackReference` populates it.

joeduffy: BTW, the reason I made this change is that I found the current error message confusing. It seems more likely that I've simply forgotten to include the fully qualified name than it does I'm trying to transfer the stack between organizations. In the event I _am_ trying to transfer, the current error is great, but wanted to also cater to the likely common case too.

chrsmith (approved):
Thanks for making the changes. All is well, with some tiny nitpicks if you are interested.

> Yes, that's exactly why we're using parseStackName here, so that we can check whether the owner was left off...

Ah, I get it now. Thanks. Having a more specific error message like "did you mean..." sounds great.

Two very minor things though:

```go
qn, _ := b.parseStackName(string(newName))
```

Is `qn` short for qualified name? If so, part of what confused me was that we call `parseStackName` explicitly because it is _not_ qualified. So maybe leaving a comment to call out this non-obvious expectation?

Also, I thought the Golang linter would complain if you ever swallowed an `error` return value from a function. So I'm surprised it isn't complaining about `qn, _ := `. That having been said, I won't harp on you for not adding tedious and likely unnecessarily excessive error handling.

joeduffy:
> So maybe leaving a comment to call out this non-obvious expectation?

Fair feedback! This _is_ actually a qualified name (it includes the `Owner` property), it's just not auto-populated with an owner in the event that it's missing. I agree it's not super clear, so I've renamed the variable, added a comment, and also explicitly ignored the error for better hygienic purposes. Hope this is better!

PR #5822, opened 2020-11-23
Populate package for doc args types

mikhailshilkov:
This fixes a panic during docs generation which occurs in the recent .NET codegen block which evaluates the package of an object type

https://github.com/pulumi/docs/runs/1443615556?check_suite_focus=true#step:15:83


stack72 (approved):

justinvp (approved):

stack72:
This looks like it might still panic - https://github.com/pulumi/pulumi/pull/5822/checks?check_run_id=1444298689#step:11:472

mikhailshilkov:
@stack72  Random had no invokes - fixed that path too. Do you understand that k8s error now?

PR #5817, opened 2020-11-23
Add support for getResource to the .NET SDK

mikhailshilkov:
Resolve https://github.com/pulumi/pulumi/issues/5796

PR #5829, opened 2020-11-25
[dotnet] Resource package registration in .NET

### LINKED ISSUE #5800: [codegen/dotnet] Register package/module resources
.NET side of https://github.com/pulumi/pulumi/pull/5806

mikhailshilkov:
Introduce package and module discovery to the .NET SDK. Resolve #5800.

The implementation relies heavily on reflection:

1. Most public package/module related APIs are removed. Technically, it's breaking, but it does not affect anyone.
2. Every provider gets a single class `ResourcePackage` which implements both provider and custom resource creation.
3. At the time of the first module lookup, we scan the list of referenced assemblies and search for all `ResourcePackage` types. This dictionary is built once and then reused.

All downstream providers fail to build because they haven't been updated to use #5817 yet. I did test Random and Azure locally.


mikhailshilkov:
Okay, I found https://github.com/pulumi/pulumi/pull/5805 so I need to add version comparisons.

Though - how could two versions of the same package be referenced at the same time? What is the exact scenario of version resolution?

sdk/dotnet/Pulumi/Serialization/ResourcePackages.cs
```diff
@@ -1,40 +1,98 @@
...
+                {
+                    if (PossibleMatch(reference))
+                    {
+                        var assembly = Assembly.Load(reference);
+                        assembliesToCheck.Enqueue(assembly);
+                    }
+                }
+            }
+
+            static bool PossibleMatch(AssemblyName? assembly) => assembly != null && !assembly.FullName.StartsWith("System");
```
justinvp, line 95: Nit: `StartsWith` does a culture-sensitive comparison by default. Best to specify an explicit `StringComparison`.

```suggestion
            static bool PossibleMatch(AssemblyName? assembly) => assembly != null && !assembly.FullName.StartsWith("System", StringComparison.Ordinal);
```

sdk/dotnet/Pulumi.Tests/Serialization/ResourcePackagesTests.cs
```diff
@@ -0,0 +1,117 @@
...
+                throw new System.NotImplementedException();
+            }
+
+            public Resource Construct(string name, string type, string urn)
+            {
+                throw new System.NotImplementedException();
+            }
+        }
+        
+        private class VersionBlankTestPackage : BaseTestPackage
```
justinvp, line 81: Why does this have "Blank" in the name when there is a version specified?

pkg/codegen/dotnet/gen.go
```diff
@@ -1021,6 +1021,72 @@ func (mod *modContext) genEnums(w io.Writer, enums []*schema.EnumType) error {
...
+	fmt.Fprintf(w, "namespace %s\n", assemblyName)
+	fmt.Fprintf(w, "{\n")
+
+	fmt.Fprintf(w, "    internal class ResourcePackage : IResourcePackage\n")
+	fmt.Fprintf(w, "    {\n")
+	fmt.Fprintf(w, "        public string Name => \"%s\";\n\n", pkg.Name)
+	fmt.Fprintf(w, "        public string? Version => null;\n\n")
+	fmt.Fprintf(w, "        [Obsolete(\"The overload with 4 parameters is deprecated and will be removed\")]\n")
+	fmt.Fprintf(w, "        public ProviderResource ConstructProvider(string name, string type, System.Collections.Generic.IDictionary<string, object?>? args, string urn)\n")
+	fmt.Fprintf(w, "            => ConstructProvider(name, type, urn);\n\n")
```
justinvp, line 1058: I don't quite follow why we need to keep this.

pkg/codegen/dotnet/gen.go
```diff
@@ -1021,6 +1021,72 @@ func (mod *modContext) genEnums(w io.Writer, enums []*schema.EnumType) error {
...
+	}
+
+	// Open the namespace.
+	fmt.Fprintf(w, "namespace %s\n", assemblyName)
+	fmt.Fprintf(w, "{\n")
+
+	fmt.Fprintf(w, "    internal class ResourcePackage : IResourcePackage\n")
+	fmt.Fprintf(w, "    {\n")
+	fmt.Fprintf(w, "        public string Name => \"%s\";\n\n", pkg.Name)
+	fmt.Fprintf(w, "        public string? Version => null;\n\n")
```
justinvp, line 1055: I added version info to Node.js and Python in https://github.com/pulumi/pulumi/pull/5839. Want to do the same here?

sdk/dotnet/Pulumi/Serialization/ResourcePackages.cs
```diff
@@ -1,40 +1,98 @@
...
+                .Select(c => c.Invoke(new object[0]))
+                .Cast<IResourcePackage>()
+                .ToImmutableList();
         }
-
-        public static void RegisterResourcePackage(string name, string version, IResourcePackage package)
+        
+        // Assemblies are loaded on demand, so it could be that some assemblies aren't yet loaded to the current
+        // app domain at the time of discovery. This method iterates through the list of referenced assemblies
+        // recursively.
```
justinvp, line 60: Will we be able to address this in the future, if needed (e.g. via `AssemblyLoad` event)?

sdk/dotnet/Pulumi/Serialization/ResourcePackages.cs
```diff
@@ -1,40 +1,98 @@
...
+        // Note: If an assembly is referenced but its types are never used anywhere in the program, that reference
+        // will be optimized out and won't appear in the result of the enumeration.
```
justinvp, line 62: How would someone mitigate this in their program if they're running into it?

justinvp (approved):
Thanks for putting this together, @mikhailshilkov, and apologies for the delay reviewing it! Some comments, but LGTM otherwise.

@pgavlin, any other feedback?

sdk/dotnet/Pulumi/Serialization/ResourcePackages.cs
mikhailshilkov: This whole loop through `GetReferencedAssemblies` and call `Assembly.Load` below is to work around this problem. I'll extend the comment.

sdk/dotnet/Pulumi/Serialization/ResourcePackages.cs
mikhailshilkov: They would have to use something from that assembly in their code. This won't happen often in practice, I suspect.

sdk/dotnet/Pulumi.Tests/Serialization/ResourcePackagesTests.cs
mikhailshilkov: This looks like a left-over, I'll delete this class, sorry for that

pkg/codegen/dotnet/gen.go
mikhailshilkov: If a provider updates to the latest `pulumi/pulumi` Go package before updating to a published version of NuGet, the generated code won't compile without this shim. We can remove it if we think it's not a problem.

justinvp: Ah, ok. Leaving it sounds fine. Let's open an issue to track removing it.

sdk/dotnet/Pulumi/Serialization/ResourcePackages.cs
```diff
@@ -1,40 +1,98 @@
...
 
 namespace Pulumi
 {
     public interface IResourcePackage
     {
-        ProviderResource ConstructProvider(string name, string type, IDictionary<string, object?>? args, string urn);
+        string Name { get; }
+        string? Version { get; }
+        ProviderResource ConstructProvider(string name, string type, string urn);
+        Resource Construct(string name, string type, string urn);
```
pgavlin, line 18: Just a note that coupling these operations--provider and resource construction--constrains Pulumi SDKs s.t. they are delivered in the same assembly. I _think_ that's fine, but I want to call it out explicitly.

pgavlin: One alternative that we might want to consider is marking resource classes with an attribute that records their type token and then mapping from type token to class. I think that approach is additive, and we could switch to it in the future without any breaks.

pgavlin (approved):
LGTM! Just a couple notes re: the `IResourcePackage` approach.

sdk/dotnet/Pulumi/Serialization/ResourcePackages.cs
mikhailshilkov: These are brilliant points. I went ahead and drafted an alternative implementation based on attributes. I fired a separate PR https://github.com/pulumi/pulumi/pull/5875 to be able to discuss both. I find it simpler, more flexible, and more idiomatic. PTAL cc @justinvp

justinvp: I also like the idea of using an attribute to record the type token. I'll take a look at the other PR.

Issue #5798, opened 2020-11-20
[sdk/nodejs] Add support for getResource


justinvp:
TypeScript implementation of https://github.com/pulumi/pulumi/pull/5694

PR #5837, opened 2020-11-30
Add support for getResource to Node.js SDK

justinvp:
And update Node's resource ref deserialization to match Python.

Also, fixed a bug in Python resource ref deserialization that I noticed.

Fixes https://github.com/pulumi/pulumi/issues/5798

### DIFF
--- a/sdk/nodejs/runtime/rpc.ts
+++ b/sdk/nodejs/runtime/rpc.ts
@@ -20,6 +20,8 @@ import { ComponentResource, CustomResource, ProviderResource, Resource, URN } fr
 import { debuggablePromise, errorString, promiseDebugString } from "./debuggable";
 import { excessiveDebugOutput, isDryRun, monitorSupportsResourceReferences, monitorSupportsSecrets } from "./settings";
 
+import * as semver from "semver";
+
 const gstruct = require("google-protobuf/google/protobuf/struct_pb.js");
 
 export type OutputResolvers = Record<string, (value: any, isStable: boolean, isSecret: boolean, deps?: Resource[], err?: Error) => void>;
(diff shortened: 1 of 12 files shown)


sdk/python/lib/pulumi/runtime/rpc.py
```diff
@@ -748,7 +748,7 @@ def register_resource_module(pkg: str, mod: str, module: ResourceModule):
                 raise ValueError(f"Cannot re-register module {key}@{module.version()}. Previous registration was {existing}, new registration was {module}.")
     else:
         resource_modules = []
-        _RESOURCE_MODULES[pkg] = resource_modules
+        _RESOURCE_MODULES[key] = resource_modules
```
justinvp, line 751: Noticed this bug, so went ahead and fixed as part of this change.

sdk/nodejs/runtime/rpc.ts
```diff
@@ -578,54 +586,113 @@ export function suppressUnhandledGrpcRejections<T>(p: Promise<T>): Promise<T> {
...
+    }
+    return have.major === want.major && have.minor >= want.minor && have.patch >= want.patch;
+}
+
 /**
  * A ResourcePackage is a type that understands how to construct resource providers given a name, type, args, and URN.
  */
 export interface ResourcePackage {
-    constructProvider(name: string, type: string, args: any, opts: { urn: string }): ProviderResource;
+    readonly version: semver.SemVer | undefined;
```
justinvp, line 605: Should we keep the return type `string` to avoid the need for packages to depend on the `semver` package?

sdk/nodejs/runtime/rpc.ts
```diff
@@ -578,54 +586,113 @@ export function suppressUnhandledGrpcRejections<T>(p: Promise<T>): Promise<T> {
...
+export function getResourcePackage(pkg: string, version: string): ResourcePackage | undefined {
+    const ver = semver.parse(version) ?? undefined;
+
+    let bestPackage: ResourcePackage | undefined = undefined;
+    for (const resourcePackage of resourcePackages.get(pkg) ?? []) {
+        if (!checkVersion(ver, resourcePackage.version)) {
+            continue;
+        }
+        if (bestPackage === undefined ||
+            (resourcePackage.version && bestPackage.version && semver.gt(resourcePackage.version, bestPackage.version))) {
```
justinvp, line 641: Is this correct?

sdk/nodejs/runtime/rpc.ts
justinvp: Yeah, I'm going to change this to be a string -- otherwise, we can run into TypeScript errors if we don't include `@types/semver` as a runtime dependency.

sdk/nodejs/runtime/rpc.ts
pgavlin: I think that this should use `checkVersion`, which accommodates the `undefined` wildcard behavior.

pgavlin: Likewise for modules.

lblackstone (approved):

PR #5839, opened 2020-11-30
[codegen/nodejs] package/module registration

justinvp:
Generate ResourcePackage and ResourceModule implementations and
registrations. A ResourcePackage is generated for any module that
includes a provider resource (which should be the root module only), and
a ResourceModule is generated for any module that includes a resource.

Update: Added a commit that adds version info for Python and Node.

Fixes https://github.com/pulumi/pulumi/issues/5802
Fixes https://github.com/pulumi/pulumi/issues/5840


justinvp:
@pgavlin, in https://github.com/pulumi/pulumi/pull/5806 you note:

> Note that version information is currently omitted. We should fix this up before enabling resource reference deserialization end-to-end.

What do we need to do for this? I've omitted it here as well, so we'll have to do it for both Python and TypeScript (and other langs).

lblackstone (approved):
I think the changes look reasonable, but it would be nice to see a test with the expected output.

pgavlin (approved):

justinvp:
@pgavlin, @lblackstone, I've added a commit to this PR that adds the version info to Node and Python, along with updates to existing test data. PTAL.

lblackstone (approved):

PR #5857, opened 2020-12-03
Make credentials.json file writes concurrency safe

### LINKED ISSUE #3877: Malformed credentials file
Running Pulumi CLI commands in CI is failing with the following error:

```
error: could not get cloud url: unmarshalling credentials file: unexpected end of JSON input
```

The following scripts are representative of what was running when the error happened:

```bash
# prepare.sh

curl https://sdk.cloud.google.com | bash > /dev/null
export PATH=$PATH:/root/google-cloud-sdk/bin
KEY_FILE="$(mktemp)"
echo "$GCP_SERVICE_ACCOUNT" > "$KEY_FILE"
gcloud auth activate-service-account --key-file "$KEY_FILE"

curl -fsSL https://get.pulumi.com/ | bash
export PATH=$PATH:$HOME/.pulumi/bin
pulumi login
```

```bash
# deploy.sh

pulumi stack init "$PULUMI_STACK" || true

if [[ "$PULUMI_STACK" != "production" ]] || [[ "$PULUMI_STACK" != "staging" ]]; then
  pulumi config refresh --stack development || true
  json="$(pulumi config --stack development --show-secrets --json)"
  keys="$(echo $json | jq -r 'keys[]')"

  for key in $(echo $keys | tr '\n' ' ')
  do
    value="$(echo $json | jq -r '.["'"$key"'"].value')"
    is_secret="$(echo $json | jq -r '.["'"$key"'"].secret')"
    pulumi config --stack "$PULUMI_STACK" set "$key" "$value" $(if [ $is_secret = 'true' ]; then echo '--secret'; fi)
  done
fi
```

Some things to keep in mind:
- In some instances, the error doesn't happen at all.
- The `prepare.sh` bash script is run once before updating each stack.
- The `deploy.sh` bash script above executes in parallel for multiple Pulumi projects. In our case there are around 9 projects.
- The scripts were modified in some runs to output the Google key file and well as the Pulumi credentials file (`$HOME/.pulumi/credentials.json`) before running the scripts and they seemed to always be valid JSON.
- The scripts were modified to trace each command and the error seemed to happen during the `pulumi config set` call in `deploy.sh`.
- There is a conversation in the Pulumi Slack in the channel `#gcp` that precludes this issue.

Evaluating all the information throughout the conversation and presented above a potential cause for the issue could be concurrent use of Pulumi commands in the CLI. If the commands are writing to the credentials file while others read it, it's possible that if the writes are not atomic, the reads would be in a transient state that is malformed.

lukehoban:
Running `pulumi` operations in parallel could occasionally result in truncating the `~/.pulumi/credentials.json` file and reading that truncated file from another process before the content could be written.

Instead, use `os.Rename` to atomically replace the file contents.

Concurrent `pulumi` operations could still compete for who gets to write the file first, and could lead to surprising results in some extreme cases.   But we should not see the corrupted file contents any longer.

Fixes #3877.

### DIFF
--- a/sdk/go/common/workspace/creds.go
+++ b/sdk/go/common/workspace/creds.go
@@ -18,11 +18,13 @@ import (
 	"encoding/json"
 	"io/ioutil"
 	"os"
+	"path"
 	"path/filepath"
 	"time"
 
 	"github.com/pkg/errors"
 
+	"github.com/pulumi/pulumi/sdk/v2/go/common/util/contract"
 	"github.com/pulumi/pulumi/sdk/v2/go/common/util/logging"
 )
 
@@ -214,7 +216,9 @@ func StoreCredentials(creds Credentials) error {
 		return errors.Wrapf(err, "marshalling credentials object")
 	}
 
-	tempCredsFile, err := ioutil.TempFile("", "credentials-*.json")
+	// Use a temporary file and atomic os.Rename to ensure the file contents are
+	// updated atomically to ensure concurrent `pulumi` CLI operations are safe.
+	tempCredsFile, err := ioutil.TempFile(path.Dir(credsFile), "credentials-*.json")
 	if err != nil {
 		return err
 	}
@@ -228,6 +232,7 @@ func StoreCredentials(creds Credentials) error {
 	}
 	err = os.Rename(tempCredsFile.Name(), credsFile)
 	if err != nil {
+		contract.IgnoreError(os.Remove(tempCredsFile.Name()))
 		return err
 	}
 
--- a/sdk/go/common/workspace/creds.go
+++ b/sdk/go/common/workspace/creds.go
@@ -213,5 +213,23 @@ func StoreCredentials(creds Credentials) error {
 	if err != nil {
 		return errors.Wrapf(err, "marshalling credentials object")
 	}
-	return ioutil.WriteFile(credsFile, raw, 0600)
+
+	tempCredsFile, err := ioutil.TempFile("", "credentials-*.json")
+	if err != nil {
+		return err
+	}
+	_, err = tempCredsFile.Write(raw)
+	if err != nil {
+		return err
+	}
+	err = tempCredsFile.Close()
+	if err != nil {
+		return err
+	}
+	err = os.Rename(tempCredsFile.Name(), credsFile)
+	if err != nil {
+		return err
+	}
+
+	return nil
 }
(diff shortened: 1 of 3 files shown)


sdk/go/common/workspace/creds.go
```diff
@@ -213,5 +213,23 @@ func StoreCredentials(creds Credentials) error {
 	if err != nil {
 		return errors.Wrapf(err, "marshalling credentials object")
 	}
-	return ioutil.WriteFile(credsFile, raw, 0600)
+
+	tempCredsFile, err := ioutil.TempFile("", "credentials-*.json")
```
justinvp, line 217: Nit: Add a brief comment explaining why we write to a temporary file first and then rename (for concurrency safe writes) vs. just writing directly?

justinvp, line 217: Looks like Windows is failing with:

> problem logging in: rename C:\Users\RUNNER~1\AppData\Local\Temp\credentials-502826475.json D:\a\pulumi\pulumi\src\stackName-1606979904380190300\credentials.json: The system cannot move the file to a different disk drive.

One workaround would be to create the temp file in the same dir as the final file. The downside is leftover temp files may be left around in error cases, but that should hopefully be rare. If we do this, should we try to remove the temp file if the rename fails?

justinvp (approved):
LGTM otherwise.

EvanBoyle (approved):

orteme:
I got errors now from windows agents in jenkins after this change:
```
E:\workspace\PM - Dev@2>pulumi logout && pulumi login -c azblob://some-azure-blob-here
error: problem logging in: rename credentials-139624339.json C:\Users\username-here\.pulumi\credentials.json: The system cannot move the file to a different disk drive.
```
pulumi version is 2.15.1

lukehoban:
@orteme @mmmmmarco @xs4free Sorry about the error you are seeing here.  This was introduced in `2.15.1` and we'll work on getting a patch out with a fix as soon as possible.  We'll use https://github.com/pulumi/pulumi-az-pipelines-task/issues/52 to track this.

gitfool:
Could the fix in `2.15.3` please be published to Chocolatey too (https://chocolatey.org/packages/pulumi#versionhistory)?

stack72:
Hey @gitfool 

We are in a queue for chocolatey publishing right now - it's working their way through their automated system

Sorry for the delay in this publishing

Paul

Issue #5452, opened 2020-09-24
`change-secrets-provider` is unable to change from one passphrase to another passphrase


clstokes:
#### Problem

If I try to use `change-secrets-provider` to change the passphrase my stack is currently using, I get prompted for my current passphrase several times, but I'm never prompted for a _new_ passphrase and then the command exits.

```
% pulumi stack change-secrets-provider passphrase
Enter your passphrase to unlock config/secrets
    (set PULUMI_CONFIG_PASSPHRASE or PULUMI_CONFIG_PASSPHRASE_FILE to remember):
Enter your passphrase to unlock config/secrets
    (set PULUMI_CONFIG_PASSPHRASE or PULUMI_CONFIG_PASSPHRASE_FILE to remember):
Enter your passphrase to unlock config/secrets
    (set PULUMI_CONFIG_PASSPHRASE or PULUMI_CONFIG_PASSPHRASE_FILE to remember):
Enter your passphrase to unlock config/secrets
    (set PULUMI_CONFIG_PASSPHRASE or PULUMI_CONFIG_PASSPHRASE_FILE to remember):
%
```

#### Suggestions for a fix

Prompt for a new passphrase.

Also, why do I get prompted for the current passphrase so many times? Could this be improved too?

stack72:
So this is when you are trying to rotate the passphrase used - so staying within the passphrase provider right?

clstokes:
Yes, staying within the passphrase provider, but trying to change the passphrase that is used.

PR #5865, opened 2020-12-04
Ensure we can rotate pulumi passphrase secrets providers

stack72:
Fixes: #5452

When the user is requesting to change the secrets provider to a
passphrase provider, we now calculate that has been requested.

This means, we will prompt for a new passphrase for use in encrypting
the stack.

```
pulumi stack change-secrets-provider passphrase
Enter your passphrase to unlock config/secrets
    (set PULUMI_CONFIG_PASSPHRASE or PULUMI_CONFIG_PASSPHRASE_FILE to remember):
Enter your new passphrase to protect config/secrets:
Re-enter your new passphrase to confirm:
Migrating old configuration and state to new secrets provider
Enter your passphrase to unlock config/secrets
    (set PULUMI_CONFIG_PASSPHRASE or PULUMI_CONFIG_PASSPHRASE_FILE to remember):
```

Thanks to @pierskarsenbarg for helping to drive through the workflow here!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ CHANGELOG
 - Automatically install missing Python dependencies.
   [#5787](https://github.com/pulumi/pulumi/pull/5787)
 
+- [cli] Ensure `pulumi stack change-secrets-provider` allows rotating the key for a passphrase provider
+  [#5865](https://github.com/pulumi/pulumi/pull/5865/)
+
 ## 2.15.0 (2020-12-02)
 
 - [sdk/python] Add deserialization support for enums.
--- a/pkg/cmd/pulumi/crypto.go
+++ b/pkg/cmd/pulumi/crypto.go
@@ -58,12 +58,14 @@ func getStackSecretsManager(s backend.Stack) (secrets.Manager, error) {
 		}
 
 		if ps.EncryptionSalt != "" {
-			return newPassphraseSecretsManager(s.Ref().Name(), stackConfigFile)
+			return newPassphraseSecretsManager(s.Ref().Name(), stackConfigFile,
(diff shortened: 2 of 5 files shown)


justinvp:
After we print `Migrating old configuration and state to new secrets provider`, why does the user have to re-enter their new passphrase again? Is there any way we could avoid that?

Ideally, the flow would be:

```
pulumi stack change-secrets-provider passphrase
Enter your passphrase to unlock config/secrets
    (set PULUMI_CONFIG_PASSPHRASE or PULUMI_CONFIG_PASSPHRASE_FILE to remember):
Enter your new passphrase to protect config/secrets:
Re-enter your new passphrase to confirm:
Migrating old configuration and state to new secrets provider
```

stack72:
> After we print `Migrating old configuration and state to new secrets provider`, why does the user have to re-enter their new passphrase again? Is there any way we could avoid that?
> 
> Ideally, the flow would be:
> 
> ```
> pulumi stack change-secrets-provider passphrase
> Enter your passphrase to unlock config/secrets
>     (set PULUMI_CONFIG_PASSPHRASE or PULUMI_CONFIG_PASSPHRASE_FILE to remember):
> Enter your new passphrase to protect config/secrets:
> Re-enter your new passphrase to confirm:
> Migrating old configuration and state to new secrets provider
> ```

So the reason the user needs to do that is that they need to unlock the new secrets manager again - we don't (currently) carry the secrets manager out of `createSecretsManager` to make that change would be a larger diff and would potentially break some areas - we can defo look at that as a follow up 

This only happens as part of the passphrase secrets manager. It used to ask twice after the new secrets manager was created (once for config and once for state) but I was able to merge those funcs

justinvp (approved):
LGTM

Issue #5871, opened 2020-12-06
Panics during imports: index out of range [X] with length Y


joeduffy:
We have seen multiple instances where a `pulumi import` fails with an error `panic: `. For example, see https://github.com/pulumi/pulumi-aws/issues/1207#issuecomment-726784347 and https://github.com/pulumi/pulumi-aws/issues/1245. This is a bug in the implementation of import, not the resource providers, so opening up an issue here.

PR #5872, opened 2020-12-06
Properly resize arrays when adding

joeduffy:
The current logic attempts to update the array but, because
append may need to allocate a new array with adequate space,
the code can currently leave dest referring to the old,
under-sized array. The solution is to use the set(dest)
logic that already exists and is used for the IsNull case.

Added a test case that would fail before this fix and now passes.

This fixes pulumi/pulumi#5871.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@ CHANGELOG
 =========
 
 ## HEAD (Unreleased)
-_(none)_
+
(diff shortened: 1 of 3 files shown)


EvanBoyle (approved):

stack72 (approved):

stack72:
Thank you so much!!

PR #5875, opened 2020-12-07
[dotnet] Resource package registration in .NET (attributes)

### LINKED ISSUE #5800: [codegen/dotnet] Register package/module resources

mikhailshilkov:
Introduce package and module discovery to the .NET SDK. Resolve #5800.

This is a revamp of #5829 to the attribute-based model instead of the interface-based packages.


pkg/codegen/dotnet/gen.go
```diff
@@ -684,6 +684,7 @@ func (mod *modContext) genResource(w io.Writer, r *schema.Resource) error {
 	if r.DeprecationMessage != "" {
 		fmt.Fprintf(w, "    [Obsolete(@\"%s\")]\n", strings.Replace(r.DeprecationMessage, `"`, `""`, -1))
 	}
+	fmt.Fprintf(w, "    [ResourceIdentifier(\"%s\", \"%s\")]\n", r.Token, mod.pkg.Version.String())
```
justinvp, line 687: Unfortunately, I don't think we can emit the version like this as we would have already generated the code and committed it before the version has been tagged in Git.

For Node and Python, we've looked it up at runtime, just like how we do inside the resource constructors.

https://github.com/pulumi/pulumi/blob/a5c30f1547c9865d55f6aaab4bb4c434d29b4202/pkg/codegen/nodejs/gen.go#L1408

https://github.com/pulumi/pulumi/blob/a5c30f1547c9865d55f6aaab4bb4c434d29b4202/pkg/codegen/python/gen.go#L534

Should we do the same here, using the same `Utilities.Version` used inside the resource constructors?

https://github.com/pulumi/pulumi/blob/a5c30f1547c9865d55f6aaab4bb4c434d29b4202/pkg/codegen/dotnet/gen.go#L820

If we still want the version to come from the attribute, would we unseal it and make the `Version` property `virtual`, and then generate a subclass attribute in the package that overrides the `Version` to return `Utilities.Version`, and then apply the subclass attribute throughout the package's resources?

I guess an alternative would be to use a string replacement, and at build time first go through all .cs files for resources and replace a placeholder value with the actual version, but that'd involve changing a bunch of files.

sdk/dotnet/Pulumi/Serialization/Attributes.cs
```diff
@@ -94,4 +94,17 @@ public sealed class OutputConstructorAttribute : Attribute
     public sealed class EnumTypeAttribute : Attribute
     {
     }
+    
+    [AttributeUsage(AttributeTargets.Class)]
+    public sealed class ResourceIdentifierAttribute : Attribute
```
justinvp, line 99: Nit: Any reason why we wouldn't name it simply `ResourceAttribute`? Or `ResourceTypeAttribute`?

sdk/dotnet/Pulumi/Serialization/Attributes.cs
```diff
@@ -94,4 +94,17 @@ public sealed class OutputConstructorAttribute : Attribute
...
+    {
+        public string Token { get; }
```
justinvp, line 101: Nit: Should we name this `Type` to match the naming we've used elsewhere (same for the .ctor arg)?

https://github.com/pulumi/pulumi/blob/a5c30f1547c9865d55f6aaab4bb4c434d29b4202/sdk/dotnet/Pulumi/Resources/CustomResource.cs#L38

sdk/dotnet/Pulumi/Serialization/Attributes.cs
mikhailshilkov: I believe `[Resource()]` will confuse the compiler vs. `Pulumi.Resource` (and maybe confusing for humans?) `ResourceTypeAttribute` is probably a better name indeed.

pkg/codegen/dotnet/gen.go
mikhailshilkov: Right, thanks for that. I got confused by this line https://github.com/pulumi/pulumi/blob/223f6501eb826f6020e7209b205e3e61359eccd6/pkg/codegen/dotnet/gen.go#L1442 (I guess I should get rid of it)

mikhailshilkov: I think a runtime lookup is fine. Are we okay requiring every resource-containing assembly to require a class called `Utilities` with a property `Version`? That's true for all our providers, for now.

sdk/dotnet/Pulumi/Serialization/ResourcePackages.cs
```diff
@@ -61,10 +61,14 @@ internal static bool TryGetResourceType(string name, string? version, [NotNullWh
                 from a in LoadReferencedAssemblies()
                 from t in a.GetTypes()
                 where typeof(CustomResource).IsAssignableFrom(t)
-                let attr = t.GetCustomAttribute<ResourceIdentifierAttribute>()
+                let attr = t.GetCustomAttribute<ResourceTypeAttribute>()
                 where attr != null
-                let versionType = (attr.Version, t)
-                group versionType by attr.Token into g
+                let ut = a.GetType($"{a.GetName().Name}.Utilities")
+                let versionProp = ut?.GetProperty("Version", BindingFlags.Static | BindingFlags.Public | BindingFlags.GetProperty)
```
justinvp, line 67: Do we want to rely on the internal `Utilities` class being present?

An alternative would be to generate an internal subclass attribute in each provider, and keep this encapsulated like:

In `Pulumi.dll` SDK:

```c#
[AttributeUsage(AttributeTargets.Class)]
public class ResourceTypeAttribute : Attribute
{
    public string Type { get; }
    public string? Version { get; set; }

    public ResourceTypeAttribute(string type) {
        Type = type;
    }
}
```

Generated code:

```c#
internal sealed class AwsResourceTypeAttribute : Pulumi.ResourceTypeAttribute
{
    public AwsResourceTypeAttribute(string type)
        : base(type)
    {
        Version = Utilities.Version;
    }
}

[AwsResourceType("aws:s3/bucket:Bucket")]
public class Bucket : Pulumi.CustomResource
{
}
```

sdk/dotnet/Pulumi/Serialization/Attributes.cs
```diff
@@ -96,14 +96,15 @@ public sealed class EnumTypeAttribute : Attribute
...
-    public sealed class ResourceIdentifierAttribute : Attribute
+    public sealed class ResourceTypeAttribute : Attribute
     {
-        public string Token { get; }
+        public string Type { get; }
+        
         public string? Version { get; }
 
-        public ResourceIdentifierAttribute(string token, string? version)
+        public ResourceTypeAttribute(string type, string? version = null)
```
justinvp, line 105: Nit: Should we remove the `version` parameter and change the `Version` property to have a setter, to align with the Framework Design Guidelines?

https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/attributes

> ❌ AVOID providing constructor parameters to initialize properties corresponding to the optional arguments.
> 
> In other words, do not have properties that can be set with both a constructor and a setter. This guideline makes very explicit which arguments are optional and which are required, and avoids having two ways of doing the same thing.

Though, not sure if this guideline has evolved at all since the C# optional arguments feature was introduced.

I don't feel strongly about this one way or the other, just wanted to point it out. I'm good with keeping it as-is.

sdk/dotnet/Pulumi/Serialization/ResourcePackages.cs
mikhailshilkov: I didn't want to create class hierarchies at first but I must admit reliance on Utilities looks worse than I anticipated. I'll explore subclasses tomorrow.

mikhailshilkov: Implemented

sdk/dotnet/Pulumi/Serialization/Attributes.cs
mikhailshilkov: I actually made `version` a required (but nullable) parameter

justinvp (approved):
LGTM!

mikhailshilkov:
Do we want to break this down into two parts: (1) just attribute (2) everything else, make a release + upgrade the major provider between (1) and (2) to avoid breaking downstream changes? Or do we take this pill?

@pgavlin Do you want to take a look before we go ahead?

justinvp:
> Do we want to break this down into two parts: (1) just attribute (2) everything else, make a release + upgrade the major provider between (1) and (2) to avoid breaking downstream changes? Or do we take this pill?

I don't think it's worth all that just to be able to see the downstream codegen diffs and that they build (especially since I can already see what the output looks like from the updated codegen tests you've included in this PR). As long as you've validated the generated code in a couple providers, I'd be ok merging.

pgavlin (approved):
LGTM! This approach looks more reliable and more .NETty :)

PR #5887, opened 2020-12-08
Allow display output to be explicitly redirected.

pgavlin:
This is intended to make it easier to write tests for components (e.g.
the backends) that consume these packages.


justinvp (approved):
LGTM

Issue #5879, opened 2020-12-07
pulumi 2.15.1 generates file access error upon login on hosted Windows Agents


Aaronontheweb:
**Describe the bug**
When attempting to run our standard Pulumi build script on Windows, which has been working without issue for months now, we get the following error as of Pulumi v2.15.1:

```
Logging in using access token from PULUMI_ACCESS_TOKEN
error: problem logging in: rename credentials-117197647.json C:\Users\VssAdministrator\.pulumi\credentials.json: The system cannot move the file to a different disk drive.
```

**To Reproduce**
Our Pulumi task that fails the issue is below:

```
- task: AzureCLI@2
          displayName: Azure CLI
          inputs:
            azureSubscription: "SdkbinDeployment"
            scriptType: bash
            scriptLocation: inlineScript
            addSpnToEnvironment: true # emits ServicePrincipal data into environment
            inlineScript: |
              ARM_SUBSCRIPTION_ID=$(az account show --query 'id' --output tsv)
              echo "##vso[task.setvariable variable=ARM_SUBSCRIPTION_ID]${ARM_SUBSCRIPTION_ID}"
              echo "##vso[task.setvariable variable=AZURE_CLIENT_ID]${servicePrincipalId}"
              echo "##vso[task.setvariable variable=AZURE_CLIENT_SECRET]${servicePrincipalKey}"
              echo "##vso[task.setvariable variable=AZURE_TENANT_ID]${tenantId}"

        - task: Pulumi@1
          displayName: "Pulumi Confg Set dockerTagVersion (QA)"
          inputs:
            azureSubscription: "SdkbinDeployment"
            args: set dockerTagVersion $(Build.SourceBranchName)
            command: "config" # set the Docker tag version equal to the current release tag
            cwd: "infrastructure/cluster-config/"
            stack: "dev" 
          env:
            AZURE_KEYVAULT_AUTH_VIA_CLI: false
```

Nothing to it - this task runs on a hosted Azure VM with a `vs2017-win2016` configuration and the task fails upon login to Pulumi. This same task works fine when run on a Linux VM but for reasons unrelated to Pulumi we're stuck with Windows for this particular task.

**Expected behavior**
Successfully logging in and running the command, which we were able to do prior to 2.15.1

**Additional context**
Going to attempt a work-around now with the `versionSpec` property and reverting back to 2.14

dustinbrooks:
[...]

PR #5893, opened 2020-12-08
Re-apply #5857

### LINKED ISSUE #3877: Malformed credentials file

lukehoban:
This re-applies the fix in #5857 to make credentials.json writes concurrency safe.

The original fix used `path.Dir` instead of `filepath.Dir` - which led to not placing the temp file in the same folder (and drive) as the renamed file target on Windows.  This led to errors on Windows environments where the working directory was on a different drive than the `~/.pulumi` directory (https://github.com/pulumi/pulumi/issues/5879).  

The change to use `filepath.Dir` instead ensures that even on Windows, the true directory containing the credentials file is used for the temp file as well.

Fixes #3877.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@ CHANGELOG
 =========
 
 ## HEAD (Unreleased)
-_(none)_
+
+- Re-apply fix for running multiple `pulumi` processes concurrently.
+  [#5893](https://github.com/pulumi/pulumi/issues/5893)
 
 ## 2.15.4 (2020-12-08)
 
--- a/sdk/go/common/workspace/creds.go
+++ b/sdk/go/common/workspace/creds.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/pkg/errors"
 
+	"github.com/pulumi/pulumi/sdk/v2/go/common/util/contract"
 	"github.com/pulumi/pulumi/sdk/v2/go/common/util/logging"
 )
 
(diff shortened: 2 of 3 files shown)


justinvp (approved):
LGTM

PR #5894, opened 2020-12-08
[codegen/docs] - Improve resource doc titles

komalali:
Fixes: https://github.com/pulumi/docs/issues/4629

### DIFF
--- a/pkg/codegen/docs/gen.go
+++ b/pkg/codegen/docs/gen.go
@@ -1305,7 +1305,7 @@ func (mod *modContext) genResourceHeader(r *schema.Resource) header {
 		baseDescription = fmt.Sprintf("Explore the %s resource of the %s module, "+
 			"including examples, input properties, output properties, "+
 			"lookup functions, and supporting types.", resourceName, mod.mod)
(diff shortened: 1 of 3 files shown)


komalali:
Looking into the downstream codegen issues.

pgavlin:
That was fast :D

pgavlin (approved):

lukehoban:
cc @davidwrede and @zchase for potential SEO impact (both positive and potentially negative?)

zchase:
There shouldn't be a negative impact from this change. There can be some positive movement in our rankings for specific keywords but since we are just removing more "general" words from the title, I wouldn't expect a major shift. For example, we rank 8th for the term `aws_lambda_permission` and this change could help elevate the change, but probably by only a spot or two.

From an SEO UX perspective, I think cleaning up the titles is a net positive change. The title is fairly noisy in its current state so this should help clear up exactly what the pages are for.

davidwrede:
+1 to everything @zchase said. :)

davidwrede (approved):

PR #5897, opened 2020-12-08
Fix new resource recording.

### LINKED ISSUE #5803: `getResource`: track `additionalSecretOutputs`
Today, the result of `getResource` does not respect the `additionalSecretOutputs` setting for the fetched resource. This can cause secret outputs to be returned as plaintext. We should fix this by having the resource map track `additionalSecretOutputs` and mark state values as secret appropriately.

pgavlin:
Record new resources after their operations have been run rather than
before in order to ensure that all state is available and appropriately
marked as secret.

Fixes #5803.


pgavlin:
Working on adding test coverage now.

lblackstone (commented):
Change LGTM

PR #5911, opened 2020-12-10
[codegen/docs] - Remove primitive type links

komalali:
The primitive type links go to external sites and aren't consistent or particularly helpful. This PR removes primitive type links from the resource docs.


komalali:
Looks like merging https://github.com/pulumi/pulumi/pull/5875/files has broken downstream codegen tests until the next release of the .NET SDK, but that's unrelated to anything in this PR.

cnunciato (approved):
I'm not the best reference for whether there might be additional bits to clear out along with these, but what's here looks good to me, and I'm +1 on removing these links in general.

davidwrede (approved):
Spot checked through the AWS and K8s ref content and I think the experience is better (i.e. only having links for non-primitive types).

PR #5914, opened 2020-12-10
[sdk/nodejs] Implement getResource in the mock monitor

justinvp:
Otherwise, unit tests for programs that reference resources that have been registered with `registerResourceModule` fail with unhandled exceptions.

We didn't have any existing tests for mocks for Node.js, so I added some based on the [ones we have for Python](https://github.com/pulumi/pulumi/blob/master/sdk/python/lib/test_with_mocks/test_testing_with_mocks.py), and augmented it to cover this case. The new test fails before the change to the SDK, and passes afterwards.

Part of https://github.com/pulumi/pulumi/issues/5832

### DIFF
--- a/sdk/nodejs/runtime/mocks.ts
+++ b/sdk/nodejs/runtime/mocks.ts
@@ -26,7 +26,7 @@ const structproto = require("google-protobuf/google/protobuf/struct_pb.js");
  */
 export interface Mocks {
     /**
-     * call mocks provider-implemented function calls (e.g. aws.get_availability_zones).
+     * Mocks provider-implemented function calls (e.g. aws.get_availability_zones).
      *
      * @param token: The token that indicates which function is being called. This token is of the form "package:module:function".
      * @param args: The arguments provided to the function call.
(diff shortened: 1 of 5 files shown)


justinvp:
@pgavlin, if the approach in this PR looks good, I'll make similar changes for Python and the other languages.

sdk/nodejs/runtime/mocks.ts
```diff
@@ -65,7 +64,21 @@ export class MockMonitor {
 
     public async invoke(req: any, callback: (err: any, innerResponse: any) => void) {
         try {
-            const result = this.mocks.call(req.getTok(), deserializeProperties(req.getArgs()), req.getProvider());
+            const tok = req.getTok();
+            const inputs = deserializeProperties(req.getArgs());
+
+            if (tok === "pulumi:pulumi:getResource") {
```
justinvp, line 70: Presumably we are OK handling this on behalf of the user, so they don't have to implement it themselves.

If there ever was a scenario where it's interesting to allow mocking calls for `"pulumi:pulumi:getResource"` in the future, I could imagine updating the `Mocks` interface to support an optional `getResource` function, that we could call if it was set instead of using the implementation here.

pgavlin: We should be implementing this, as it is a builtin that deals only in state that is internal to the engine.

pgavlin (approved):
LGTM

PR #5944, opened 2020-12-15
[codegen/*] - Align makeValidIdentifier between languages

komalali:
Currently, in [python](https://github.com/pulumi/pulumi/blob/master/pkg/codegen/python/utilities.go#L45) and [nodejs](https://github.com/pulumi/pulumi/blob/master/pkg/codegen/nodejs/utilities.go#L87), `makeValidIdentifier` adds an underscore **before** the starting character (i.e., `8` -> `_8`), but in C# and Go it **replaces** the starting character (i.e., `8` -> `_`). This PR updates the latter to match the former.

### DIFF
--- a/pkg/codegen/dotnet/gen_test.go
+++ b/pkg/codegen/dotnet/gen_test.go
@@ -70,6 +70,7 @@ func TestMakeSafeEnumName(t *testing.T) {
 		{"+", "", true},
 		{"*", "Asterisk", false},
 		{"0", "Zero", false},
+		{"8.3", "_8_3", false},
 		{"Microsoft-Windows-Shell-Startup", "Microsoft_Windows_Shell_Startup", false},
 		{"Microsoft.Batch", "Microsoft_Batch", false},
 		{"readonly", "@Readonly", false},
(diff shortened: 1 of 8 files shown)


komalali:
Whoops, [`dd2d47d..86809fc`#diff-88ca4e782a](https://github.com/pulumi-bot/pulumi-azuread/compare/dd2d47d..86809fc#diff-88ca4e782a48a96c557295aab1444032dc18661c257645ddbdf5578659b94609R28) is definitely not what we want. Fixing.

mikhailshilkov:
Were you able to check that none of the non-enum identifiers are affected, including Azure NextGen resources?

komalali:
@mikhailshilkov yup I checked in the 5 major providers and azure nextgen. No changes in the others, and the only changes in azure nextgen are related to enums with values/names that are similar to what we found breaking the update. For example, [SQL Server Version](https://github.com/pulumi/pulumi-azure-nextgen/blob/master/sdk/dotnet/Sql/V20140401/Enums.cs#L285-L286)

pgavlin (approved):

Issue #5939, opened 2020-12-14
Preview fails when resource refs are enabled


justinvp:
From https://github.com/pulumi/pulumi/issues/5933:

Previews fail in very incorrect ways with CLI `2.15.5`.

```ts
import * as aws from "@pulumi/aws";
const bucket = new aws.s3.Bucket("mybucket");
bucket.onObjectCreated("foofunction", async (ev, ctx) => {
    console.log(JSON.stringify(ev));
});
```

```
$ pulumi preview
...
Diagnostics:
  aws:lambda:Permission (foofunction):
    error: aws:lambda/permission:Permission resource 'foofunction' has a problem: "function_name" doesn't comply with restrictions ("^(arn:[\\w-]+:lambda:)?([a-z]{2}-(?:[a-z]+-){1,2}\\d{1}:)?(\\d{12}:)?(function:)?([a-zA-Z0-9-_]+)(:(\\$LATEST|[a-zA-Z0-9-_]+))?$"): "urn:pulumi:dev::promisecapture::aws:s3/bucket:Bucket$aws:s3:BucketEventSubscription$aws:lambda/function:Function::foofunction"
```

This happens when resource refs are enabled.

justinvp:
The problem is how we marshal resource references when we're not keeping resources:

https://github.com/pulumi/pulumi/blob/cae129c7fa73df2c88115aae71071c7e040e9802/sdk/go/common/resource/plugin/rpc.go#L166-L174

If we see the ID is an empty string, we return the URN, which isn't taking into account that the ID may be unknown.

If the ID is unknown (`ref.HasID == true && ref.ID == ""`) we should return the ID as an unknown string, instead of returning the URN.

leezen:
I'm not sure that this is actually fixed. With pulumi 2.17.0 and enabling via `PULUMI_ENABLE_RESOURCE_REFERENCES=1` I still see the same error.

PR #5952, opened 2020-12-15
Fix marshalling unknown resource IDs

justinvp:
Fixes https://github.com/pulumi/pulumi/issues/5939

### DIFF
--- a/sdk/go/common/resource/plugin/rpc.go
+++ b/sdk/go/common/resource/plugin/rpc.go
@@ -166,8 +166,11 @@ func MarshalPropertyValue(v resource.PropertyValue, opts MarshalOptions) (*struc
 		ref := v.ResourceReferenceValue()
 		if !opts.KeepResources {
 			val := string(ref.URN)
-			if ref.ID != "" {
+			if ref.HasID {
 				val = string(ref.ID)
+				if val == "" {
+					val = UnknownStringValue
+				}
 			}
 			logging.V(5).Infof("marshalling resource value as raw URN or ID as opts.KeepResources is false")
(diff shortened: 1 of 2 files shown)


sdk/go/common/resource/plugin/rpc.go
```diff
@@ -166,8 +166,11 @@ func MarshalPropertyValue(v resource.PropertyValue, opts MarshalOptions) (*struc
 		ref := v.ResourceReferenceValue()
 		if !opts.KeepResources {
 			val := string(ref.URN)
-			if ref.ID != "" {
+			if ref.HasID {
 				val = string(ref.ID)
+				if val == "" {
+					val = UnknownStringValue
```
pgavlin, line 172: I'm not sure that this is correct--don't the SDKs assume that an empty ID is an unknown ID? Or does this "just work" in the various unmarshallers b/c they call into lower-level unmarshalling code to unmarshal the ID itself?

justinvp: Ah, I see. I'll have to check (but you're probably right).

pgavlin: I do think that we'll either need to change all of the marshalers to understand the unknown property value in the ID field of a resource ref or add a `HasID` bit to the wire format. I'm slightly in favor of the former, in which case we should consider revising the in-memory representation similarly. I chose the `HasID` bit there because I wanted to retain the strongly-typed ID field (instead of a `resource.PropertyValue` that could hold a computed property).

justinvp:
Closing in favor of https://github.com/pulumi/pulumi/pull/5960

Issue #5957, opened 2020-12-16
panic when attempting to use pulumi from terminal with 0 width


leezen:
Upon `pulumi up` (for any new project), and receive the following stack trace:

```
panic: fatal: An assertion has failed: action must be non empty

goroutine 54 [running]:
github.com/pulumi/pulumi/sdk/v2/go/common/util/contract.failfast(...)
        /Users/runner/work/pulumi/pulumi/sdk/go/common/util/contract/failfast.go:23
github.com/pulumi/pulumi/sdk/v2/go/common/util/contract.Assertf(0x0, 0x1cdded5, 0x18, 0x0, 0x0, 0x0)
        /Users/runner/work/pulumi/pulumi/sdk/go/common/util/contract/assert.go:33 +0x1a5
github.com/pulumi/pulumi/pkg/v2/backend/display.makeActionProgress(0x2b16a20, 0x1, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
        /Users/runner/work/pulumi/pulumi/pkg/backend/display/progress.go:60 +0xe5
github.com/pulumi/pulumi/pkg/v2/backend/display.(*ProgressDisplay).refreshColumns(0xc0016e2480, 0x2b16a20, 0x1, 0xc0016c4370, 0x5, 0x5, 0xc0010bd5f0, 0x5, 0x5)
        /Users/runner/work/pulumi/pulumi/pkg/backend/display/progress.go:431 +0x10e
github.com/pulumi/pulumi/pkg/v2/backend/display.(*ProgressDisplay).refreshAllRowsIfInTerminal(0xc0016e2480)
        /Users/runner/work/pulumi/pulumi/pkg/backend/display/progress.go:649 +0x373
github.com/pulumi/pulumi/pkg/v2/backend/display.(*ProgressDisplay).processNormalEvent(0xc0016e2480, 0x1cc7688, 0xc, 0x1b168c0, 0xc0013a9c20)
        /Users/runner/work/pulumi/pulumi/pkg/backend/display/progress.go:1128 +0x376
github.com/pulumi/pulumi/pkg/v2/backend/display.(*ProgressDisplay).processEvents(0xc0016e2480, 0xc00154a460, 0xc00007d260)
        /Users/runner/work/pulumi/pulumi/pkg/backend/display/progress.go:1196 +0x145
[...]

Issue #5643, opened 2020-10-28
Redirects and pipes cause a golang exception


tenwit:
When piping or redirecting the CLI's stdout, I see an exception on stdout (and nothing on stderr).
Note that I am running through Docker (pulumi/pulumi-nodejs:2.12.1).

## Expected Behavior
stdout and stderr redirection should work "as expected". Colour off, stderr and stdout handled separately, no clever cursor movement, etc.

## Current Behavior
Both piping and redirecting stdout causes this exception:
```
panic: fatal: An assertion has failed: action must be non empty

goroutine 42 [running]:
github.com/pulumi/pulumi/sdk/v2/go/common/util/contract.failfast(...)
        /Users/runner/work/pulumi/pulumi/sdk/go/common/util/contract/failfast.go:23
github.com/pulumi/pulumi/sdk/v2/go/common/util/contract.Assertf(0x0, 0x1cce43e, 0x18, 0x0, 0x0, 0x0)
        /Users/runner/work/pulumi/pulumi/sdk/go/common/util/contract/assert.go:33 +0x1a5
github.com/pulumi/pulumi/pkg/v2/backend/display.makeActionProgress(0x2b00940, 0x1, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
        /Users/runner/work/pulumi/pulumi/pkg/backend/display/progress.go:59 +0xe5
github.com/pulumi/pulumi/pkg/v2/backend/display.(*ProgressDisplay).refreshColumns(0xc000ebe000, 0x2b00940, 0x1, 0xc001b1f770, 0x5, 0x5, 0xc001564e40, 0x5, 0x5)
        /Users/runner/work/pulumi/pulumi/pkg/backend/display/progress.go:410 +0x10e
github.com/pulumi/pulumi/pkg/v2/backend/display.(*ProgressDisplay).refreshAllRowsIfInTerminal(0xc000ebe000)
        /Users/runner/work/pulumi/pulumi/pkg/backend/display/progress.go:628 +0x373
github.com/pulumi/pulumi/pkg/v2/backend/display.(*ProgressDisplay).processNormalEvent(0xc000ebe000, 0x1cb7cca, 0xc, 0x1b07820, 0xc000a30f00)
        /Users/runner/work/pulumi/pulumi/pkg/backend/display/progress.go:1107 +0x396
[...]

Issue #4904, opened 2020-06-25
Panic in display logic on Windows+AWSSessionManager+CentOS7+bash


lukehoban:
A user reported hitting the crash below when running `pulumi up` from Windows using AWS Session Manager to connect to Cent OS 7 running bash using Pulumi 1.14.1.

This is the same sympton as https://github.com/pulumi/pulumi/issues/3935, but that was fixed before in 1.11.0, so this may be another way to trigger the underlying issue.

```
panic: fatal: An assertion has failed: action must be non empty
goroutine 32 [running]:
github.com/pulumi/pulumi/pkg/util/contract.failfast(...)
	/home/travis/gopath/src/github.com/pulumi/pulumi/pkg/util/contract/failfast.go:23
github.com/pulumi/pulumi/pkg/util/contract.Assertf(0x0, 0x16d97be, 0x18, 0x0, 0x0, 0x0)
	/home/travis/gopath/src/github.com/pulumi/pulumi/pkg/util/contract/assert.go:33 +0x197
github.com/pulumi/pulumi/pkg/backend/display.makeActionProgress(0x2659b50, 0x1, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
	/home/travis/gopath/src/github.com/pulumi/pulumi/pkg/backend/display/progress.go:60 +0xe8
github.com/pulumi/pulumi/pkg/backend/display.(*ProgressDisplay).refreshColumns(0xc001238000, 0x2659b50, 0x1, 0xc0012cc230, 0x5, 0x5, 0xc0012c2d80, 0x5, 0x5)
	/home/travis/gopath/src/github.com/pulumi/pulumi/pkg/backend/display/progress.go:373 +0x10d
github.com/pulumi/pulumi/pkg/backend/display.(*ProgressDisplay).refreshAllRowsIfInTerminal(0xc001238000)
	/home/travis/gopath/src/github.com/pulumi/pulumi/pkg/backend/display/progress.go:591 +0x36e
github.com/pulumi/pulumi/pkg/backend/display.(*ProgressDisplay).processNormalEvent(0xc001238000, 0x16c72e0, 0xc, 0x1559ec0, 0xc000aba640)
	/home/travis/gopath/src/github.com/pulumi/pulumi/pkg/backend/display/progress.go:966 +0x45a
[...]

PR #5959, opened 2020-12-16
Assume no terminal if the display width/height is 0

leezen:
I'm not super sure about the warning going to stderr, but that seems like the right thing. Otherwise, the user is left wondering why they aren't getting the normal interactive terminal experience when the terminal is not reporting width/height correctly.

Fixes https://github.com/pulumi/pulumi/issues/5957


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ CHANGELOG
 - [sdk/dotnet] Add collection initializers for smooth support of Union<T, U> as element type
   [#5938](https://github.com/pulumi/pulumi/pull/5938)
 
+- Fix a bug in the Go SDK that could result in dropped resource dependencies.
```
pgavlin, line 9: Was this intentional?

leezen: No... I must have done something odd when merging from master. Let me fix that up.

lukehoban:
Does this also fix https://github.com/pulumi/pulumi/issues/5643??

leezen:
@lukehoban I haven't had a chance to try to repro #5643. Only #4904, which I wasn't able to repro. I'll try to find some time today to repro #5643

pgavlin (approved):

Issue #5943, opened 2020-12-14
Document scenarios and create test coverage for resource references


leezen:
Resource references introduces a number of changes that affects the engine's behavior. We need to document what the various scenarios are that it affects and add test coverage for those scenarios.

pgavlin:
Here's an initial cut:

## Resource Scenarios

1. Custom resources with known IDs
2. Custom resources with unknown IDs
3. Component resources

Tests for all of these can be easily written. We should ensure that each SDK and the CLI
covers marshalling and unmarshaling a resource reference that fits each scenario.

## Marshaling Boundaries

1. SDK <-> CLI
2. CLI <-> Statefile
3. CLI <-> Resource Providers
4. Statefile <-> Pulumi Console

- SDK <-> CLI and CLI <-> Resource Providers are runtime compatibility boundaries
	- Downlevel SDK/Resource Provider <-> Current CLI
	- Downlevel CLI with current SDK/Resource Provider
	- Support for resource references must be negotiated
	- It is the responsibility of the entity that does support resource references to
      declare its support and to respect the absence of support in other entities
		- For example, a CLI that does support resource references must serialize them as
          URNs (for references to component resources) or IDs (for references to custom
          resources) when communicating with an SDK or resource provider that does not
          support resource references.
	- We should be able to emulate these scenarios in unit tests (i.e. in `pkg/engine/lifecycletest/pulumi_tests`)
		- We can emulate a downlevel CLI by selectively disabling support for resource
          references in the resource monitor
		- We can emulate a downlevel SDK by selectively disabling support for resource
		  references in the test harness
		- We can emulate a downlevel resource provider by selectively disabling support
		  for resource references in the test harness
	- We may also be able to emulate these scenarios in integration tests, though doing so
      in a way that persists once the feature is enabled by default involves configuration
      knobs (e.g. envvars) that selectively disable support for resource refs in *only* the
      SDK or *only* the engine. We may not want those knobs long-term.
	- We may be also able to emulate these scenarios in SDK unit tests. If we cannot, we
	  should strongly consider enabling this emulation s.t. we can validate the behavior
	  in a more fine-grained fashion.
	- The examples test passes do construct the SDK <-> CLI scenarios, but will cease to
	  do so once both the latest release and the dev build support resource references
- CLI <-> Statefile and Statefile <-> Pulumi Console are static compatibility boundaries
	- Downlevel CLI <-> current statefile
	- Downlevel Console <-> current statefile
[...]

PR #5960, opened 2020-12-16
Improve resource ref unit tests.

### LINKED ISSUE #5939: Preview fails when resource refs are enabled

pgavlin:
- Add component ref coverage to the existing test
- Add coverage for a downlevel SDK communicating with an engine that
  supports resource refs
- Add coverage for a downlevel engine communicating with an SDK that
  supports resource refs

As part of improving coverage, these changes add a knob to explicitly
disable resource refs in the engine without the use of the environment
variable. The environment variable is now only read by the CLI.

I believe that these changes also fix #5939, which I needed to
address in order for the tests to pass.

Contributes to #5943.

### DIFF
--- a/pkg/cmd/pulumi/util.go
+++ b/pkg/cmd/pulumi/util.go
@@ -72,13 +72,10 @@ func disableProviderPreview() bool {
 }
 
 func disableResourceReferences() bool {
-	// TODO: Temporarily disabling resource ref support (https://github.com/pulumi/pulumi-kubernetes/issues/1405)
-
-	// Allow the resource reference feature to be disabled by explicitly setting an env var.
-	//	if v, ok := os.LookupEnv("PULUMI_DISABLE_RESOURCE_REFERENCES"); ok && cmdutil.IsTruthy(v) {
-	//		hasSupport = false
-	//	}
-
+	// Allow the resource reference feature to be enabled by explicitly setting an env var.
+	if v, ok := os.LookupEnv("PULUMI_ENABLE_RESOURCE_REFERENCES"); ok && cmdutil.IsTruthy(v) {
+		return false
+	}
 	return true
 }
 
--- a/pkg/cmd/pulumi/util.go
+++ b/pkg/cmd/pulumi/util.go
@@ -71,6 +71,17 @@ func disableProviderPreview() bool {
 	return cmdutil.IsTruthy(os.Getenv("PULUMI_DISABLE_PROVIDER_PREVIEW"))
 }
 
+func disableResourceReferences() bool {
+	// TODO: Temporarily disabling resource ref support (https://github.com/pulumi/pulumi-kubernetes/issues/1405)
+
+	// Allow the resource reference feature to be disabled by explicitly setting an env var.
+	//	if v, ok := os.LookupEnv("PULUMI_DISABLE_RESOURCE_REFERENCES"); ok && cmdutil.IsTruthy(v) {
+	//		hasSupport = false
+	//	}
+
+	return true
+}
+
 // skipConfirmations returns whether or not confirmation prompts should
 // be skipped. This should be used by pass any requirement that a --yes
 // parameter has been set for non-interactive scenarios.
--- a/pkg/engine/lifeycletest/resource_reference_test.go
+++ b/pkg/engine/lifeycletest/resource_reference_test.go
@@ -1,0 +1,221 @@
+//nolint:goconst
+package lifecycletest
+
+import (
+	"testing"
+
+	"github.com/blang/semver"
+	"github.com/stretchr/testify/assert"
+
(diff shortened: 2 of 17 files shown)


pgavlin:
@justinvp , I think that these changes obviate https://github.com/pulumi/pulumi/pull/5952. I'll add additional SDK-specific tests as a follow-up.

justinvp:
> I think that these changes obviate #5952. I'll add additional SDK-specific tests as a follow-up.

Thanks! Since I'm OOF until Jan, I'm going to close out #5952 and assign #5939 to you.

pgavlin:
Ping. I'd love to get these changes in ASAP to unblock other testing.

pkg/cmd/pulumi/util.go
```diff
@@ -71,6 +71,17 @@ func disableProviderPreview() bool {
...
 }
 
+func disableResourceReferences() bool {
+	// TODO: Temporarily disabling resource ref support (https://github.com/pulumi/pulumi-kubernetes/issues/1405)
+
+	// Allow the resource reference feature to be disabled by explicitly setting an env var.
+	//	if v, ok := os.LookupEnv("PULUMI_DISABLE_RESOURCE_REFERENCES"); ok && cmdutil.IsTruthy(v) {
+	//		hasSupport = false
+	//	}
+
```
leezen, line 81: We can remove the comment, close pulumi-kubernetes#1405, and uncomment the code as well, right?

pkg/engine/lifeycletest/resource_reference_test.go
```diff
@@ -0,0 +1,221 @@
...
+
+					id := "created-id"
+					if preview {
+						id = ""
+					}
+
+					// If we have resource references here, the engine has not properly disabled them.
+					if urn.Name() == "resC" {
+						assert.Equal(t, resource.NewStringProperty(string(urnA)), news["resA"])
+						assert.Equal(t, resource.NewStringProperty(string(idB)), news["resB"])
```
leezen, line 174: I don't think I quite understand this. The resource can be referenced by either the id or the urn? I would have expected this to always be via the urn? Sorry if this is simple, but I'm just not familiar with this part of the code very much.

pkg/cmd/pulumi/util.go
pgavlin: > We can remove the comment, close pulumi-kubernetes#1405, and uncomment the code as well, right?

I wasn't certain that was the plan. I don't think that we want to enable resource refs by default until we've landed all of the planned test coverage, do we?

pkg/engine/lifeycletest/resource_reference_test.go
pgavlin: > // TestResourceReferences_DownlevelEngine tests an SDK that supports resource references communicating with an engine
 // that does not.

This test simulates an SDK that supports resource references communicating with an engine that does not support resource references. In that case, the marshaller should transform a resource reference into either a plain URN (for references to component resources) or a plan ID (for references to custom resources). These checks validate that the transformation occurred as expected.

pkg/cmd/pulumi/util.go
leezen: Sorry, I didn't mean enable by default, but don't we need some way to do the reverse (enable so we can test)? I assumed that logic would live in this block.

pgavlin: > don't we need some way to do the reverse (enable so we can test)? I assumed that logic would live in this block.

Not for the purposes of these tests, which enable and disable support for resource refs by hitting the engine directly. For E2E/integration tests, though, we do need to restore the `PULUMI_ENABLE_RESOURCE_REFERENCES` envvar. I'll make that change.

leezen (approved):

PR #5965, opened 2020-12-16
make import command help a little clearer

jaxxstorm:
fixes #5965


lukehoban (approved):
I do wonder if we should also add more details on using these options at https://www.pulumi.com/docs/guides/adopting/import/#pulumi-import-command.

PR #5968, opened 2020-12-17
Improve meta tags on documentation

komalali:
@zchase I went with the suggestion you had in the issue but happy to adjust as needed. Note the difference between the Package level and Resource level tags... both for the title tag and the meta description. Let me know if we want to bring those in alignment.

Fixes: https://github.com/pulumi/docs/issues/4799


zchase (approved):
LGTM --- Thanks for the quick turnaround! 🎉

davidwrede (approved):

PR #5975, opened 2020-12-17
Use `beforeEach` in NodeJS tests.

pgavlin:
Call `runtime._reset` prior to each test to ensure that the runtime is
in a consistent state.

### DIFF
--- a/sdk/nodejs/tests/runtime/props.spec.ts
+++ b/sdk/nodejs/tests/runtime/props.spec.ts
@@ -112,6 +112,8 @@ interface TestInputs {
 }
 
 describe("runtime", () => {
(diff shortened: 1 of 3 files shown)


lblackstone (approved):

sdk/nodejs/tests/runtime/props.spec.ts
```diff
@@ -112,6 +112,8 @@ interface TestInputs {
 }
 
 describe("runtime", () => {
+    beforeEach(runtime._reset);
```
leezen, line 115:
```suggestion
    beforeEach(runtime._reset());
```

Issue #5299, opened 2020-09-08
[Automation API] Python Implementation


EvanBoyle:
The current plan is to pick the work back up on November 16th with the goal of delivering it (along with [C#](https://github.com/pulumi/pulumi/issues/5596)) by the end of the year. We've scoped the effort and it looks like the following:

- [x] Create `Workspace` Interface [TypeScript reference](https://github.com/pulumi/pulumi/blob/master/sdk/nodejs/x/automation/workspace.ts). Note there are some places where the [Go interface](https://github.com/pulumi/pulumi/blob/master/sdk/go/x/auto/workspace.go) is more complete (cancel, export, import). Ideally, we should shoot for the same level of completeness as the Go impl. 
- [x] Implement `RunPulumiCmd` [TypeScript reference](https://github.com/pulumi/pulumi/blob/master/sdk/nodejs/x/automation/cmd.ts)
- [x] `LocalWorkspace` Implementation [TypeScript reference](https://github.com/pulumi/pulumi/blob/master/sdk/nodejs/x/automation/localWorkspace.ts)
- [x] `Stack` implementation supporting local programs [TypeScript reference](https://github.com/pulumi/pulumi/blob/master/sdk/nodejs/x/automation/stack.ts)
- [x] Implement idiomatic structured error handling [TypeScript reference](https://github.com/pulumi/pulumi/blob/master/sdk/nodejs/x/automation/errors.ts)
- [x] Add `InlineProgram` support. This involves creating a Python implementation of the gRPC LanguageServer. TypeScript references: [LanguageServer](https://github.com/pulumi/pulumi/blob/master/sdk/nodejs/x/automation/server.ts), [Preview consumption](https://github.com/pulumi/pulumi/blob/master/sdk/nodejs/x/automation/stack.ts#L228-L250), and [Up consumption](https://github.com/pulumi/pulumi/blob/master/sdk/nodejs/x/automation/stack.ts#L147-L169)
- [ ] Port over a few examples from https://github.com/pulumi/automation-api-examples

There are also many tests that should be replicated. These should make the implementation and acceptance criteria fairly straight forward:

https://github.com/pulumi/pulumi/blob/master/sdk/nodejs/tests/automation/localWorkspace.spec.ts
https://github.com/pulumi/pulumi/blob/master/sdk/go/x/auto/local_workspace_test.go
https://github.com/pulumi/pulumi/blob/master/sdk/go/x/auto/errors_test.go

[...]

Issue #5449, opened 2020-09-23
[Automation API] Support parallel execution of nodejs inline programs


EvanBoyle:
The Pulumi nodejs runtime implementation uses quite a bit of global state: https://github.com/pulumi/pulumi/blob/master/sdk/nodejs/runtime/settings.ts#L188

We'll need to refactor all usage of global state out of the node SDK to allow Automation API to execute concurrent programs on different stacks that use the inline/hosted execution model.

One consequence of this is that multiple versions of pulumi can cause problems due to conflicting module level global state. Automation API users should pin to a single version of pulumi/pulumi via yarn `resolutions`: https://classic.yarnpkg.com/en/docs/selective-version-resolutions/

EvanBoyle:
To be clear for nodejs users, the Automation API allows one concurrent update per node process for the time being. If you want to run multiple updates in parallel, they will need to be in separate node processes.

georgeportillo:
> pulumi/pulumi

pulumi/pulumi or @ pulumi/pulumi ?

AMKamel:
I think the issue still exists, can you tell me what is wrong with follow cod to trigger the issue when using `yarn test`
errors
```
    Program run without the Pulumi engine available; re-run using the `pulumi` CLI

      100 |     const utility = new Utility()
      101 |     const virtualminDomainCommand =
    > 102 |       pulumi.all([args])
          |                        ^
      103 |         .apply(([args]) => {
      104 |             const playbookCreateFile = path.resolve(__dirname, `playbooks/create.yaml`)
      105 |             const playbookDeleteFile = path.resolve(__dirname, `playbooks/delete.yaml`)

      at requireTestModeEnabled (../../../node_modules/@pulumi/command/node_modules/@pulumi/runtime/settings.ts:153:15)
      at Object.getMonitor (../../../node_modules/@pulumi/command/node_modules/@pulumi/runtime/settings.ts:255:13)
      at Object.registerResource (../../../node_modules/@pulumi/command/node_modules/@pulumi/runtime/resource.ts:288:21)
      at new Resource (../../../node_modules/@pulumi/command/node_modules/@pulumi/resource.ts:393:13)
      at new CustomResource (../../../node_modules/@pulumi/command/node_modules/@pulumi/resource.ts:772:9)
      at new Command (../../../node_modules/@pulumi/local/command.ts:143:9)
      at new VirtualminDomain (index.ts:102:24)
[...]

PR #5979, opened 2020-12-18
[Automation API] Python Implementation

komalali:
This PR implements the Automation API at parity with the `Typescript` implementation.

Notably missing from full parity to the `Go` implementation are `export`, `import` and `RemoteProgram` which will be implemented in a separate PR.

Resolves: https://github.com/pulumi/pulumi/issues/5299


sdk/python/lib/pulumi/x/automation/local_workspace.py
```diff
@@ -0,0 +1,399 @@
...
+        self.select_stack(stack_name)
+        secret_arg = "--secret" if value.secret else "--plaintext"
+        self._run_pulumi_cmd_sync(["config", "set", key, value.value, secret_arg])
+
+    def set_all_config(self, stack_name: str, config: ConfigMap) -> None:
+        config_adds = []
+        with futures.ThreadPoolExecutor() as executor:
+            for key in config:
+                config_adds.append(executor.submit(self.set_config, stack_name, key, config[key]))
+        futures.wait(config_adds)
```
justinvp, line 274: Why is the use of `ThreadPoolExecutor` and `futures.wait` necessary? `set_config` is currently synchronous, so couldn't it just be called in a loop?

```python
def set_all_config(self, stack_name: str, config: ConfigMap) -> None:
    for key in config:
        self.set_config(stack_name, key, config[key])
```

sdk/python/lib/pulumi/x/automation/stack.py
```diff
@@ -0,0 +1,430 @@
...
+    # post-update info
+    result: UpdateResult
+    end_time: datetime
+    version: Optional[int]
+    Deployment: Optional[str]
+    resource_changes: Optional[OpMap]
+
+    def __init__(self,
+                 kind: UpdateKind,
+                 startTime: str,
```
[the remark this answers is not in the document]
komalali: The input is camelCase here. Like, the JSON that we are deserializing is camelCase, but the [resulting object](https://github.com/pulumi/pulumi/pull/5979/files/dd8df192616f029cb2fdead2da291758d0efc848#diff-09b628447d9d04fcab0ccc8ecd76c428765bfb3ae26701de96b520d9eb6941a8R65) is snake_case. 

Do you have an alternate suggestion for how you would like me to do this?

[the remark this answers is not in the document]
justinvp: > Do you have an alternate suggestion for how you would like me to do this?

Make the parameters snake_case, and handle converting camelCase => snake_case as part of deserializing from JSON.

sdk/python/lib/pulumi/x/automation/local_workspace.py
komalali: `set_config` is synchronous but `set_all` should execute in bulk. Using the `ThreadPoolExecutor` lets each set_config run in parallel without making the user deal with the async/await interface.

It improved the test for [bulk_config_ops](https://github.com/pulumi/pulumi/pull/5979/files/dd8df192616f029cb2fdead2da291758d0efc848#diff-9fab3179db2be36998c2b9bd1a1fb1744888a8e4b99316d5872d5dafb5a00642R209-R231) from 50s to 12s. This has been a point of contention in our [nodejs implementation](https://github.com/pulumi/pulumi/issues/6042) as well.

sdk/python/lib/pulumi/x/automation/local_workspace.py
```diff
@@ -0,0 +1,399 @@
...
+    This is identical to the behavior of Pulumi CLI driven workspaces.
+    """
+
+    @classmethod
+    def new_stack(cls,
+                  stack_name: str,
+                  project_name: Optional[str] = None,
+                  program: Optional[PulumiFn] = None,
+                  work_dir: Optional[str] = None,
+                  opts: Optional[LocalWorkspaceOptions] = None) -> Stack:
```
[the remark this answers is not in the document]
komalali: The [Go API](https://github.com/pulumi/pulumi/blob/master/sdk/go/x/auto/local_workspace.go) is largely synchronous. IMO asyncio is not very common in python quite like it is in `nodejs` and `.NET`. It adds a layer of complexity that seems unnecessary and so I modeled this after the Go implementation. I discussed this with @EvanBoyle when I was first thinking through this. I'd need a strong argument for why this **should** be async.

sdk/python/lib/pulumi/x/automation/local_workspace.py
```diff
@@ -0,0 +1,399 @@
...
+        args = locals()
+        if _is_inline_program(**args):
+            # Type checks are ignored because we have already asserted that the correct args are present.
+            return _inline_source_stack_helper(stack_name, program, project_name, Stack.create, opts)  # type: ignore
+        elif _is_local_program(**args):
+            return _local_source_stack_helper(stack_name, work_dir, Stack.create, opts)  # type: ignore
+        raise ValueError(f"unexpected args: {' '.join(args)}")
+
+    @classmethod
+    def select_existing_stack(cls,
```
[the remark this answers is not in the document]
komalali: Because of https://github.com/pulumi/pulumi/pull/5979#discussion_r550814241

sdk/python/lib/pulumi/x/automation/local_workspace.py
```diff
@@ -0,0 +1,399 @@
...
+from typing import Optional, List, Mapping, Callable
+
+from .config import ConfigMap, ConfigValue
+from .project_settings import ProjectSettings
+from .stack_settings import StackSettings
+from .workspace import Workspace, PluginInfo, StackSummary, WhoAmIResult, PulumiFn
+from .stack import Stack
+from .cmd import _run_pulumi_cmd, CommandResult
+
+setting_extensions = [".yaml", ".yml", ".json"]
```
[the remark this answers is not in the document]
komalali: Yeah, looks like I missed this one.

sdk/python/lib/pulumi/x/automation/workspace.py
```diff
@@ -0,0 +1,350 @@
...
+        Returns all Stacks created under the current Project.
+        This queries underlying backend and may return stacks not present in the Workspace
+        (as Pulumi.<stack>.yaml files).
+
+        :returns: List[StackSummary]
+        """
+        pass
+
+    @abstractmethod
+    def install_plugin(self, plugin_name: str, version: str, kind: str) -> None:
```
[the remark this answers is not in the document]
komalali: Yes, but in [python](https://docs.python.org/3/library/typing.html#typing.Optional), `Optional` means that `None` is a valid value for that parameter, which is not the case here. 

> Note that this is not the same concept as an optional argument, which is one that has a default. An optional argument with a default does not require the Optional qualifier on its type annotation just because it is optional.

That being said, I wasn't sure if the abstract method needed to specify the defaults (it is specified in the [LocalWorkspace](https://github.com/pulumi/pulumi/pull/5979/files#diff-ec008bd37a4735c2e8200ae287b2614e6d6586bddbe38c5b596817aba90ab8b4R324) implementation).

sdk/python/lib/pulumi/x/automation/workspace.py
```diff
@@ -0,0 +1,350 @@
...
+        Installs a plugin in the Workspace, for example to use cloud providers like AWS or GCP.
+
+        :param plugin_name: The name of the plugin to install.
+        :param version: The version range to install.
+        :param kind: The kind of plugin.
+        """
+        pass
+
+    @abstractmethod
+    def remove_plugin(self, plugin_name: Optional[str], version_range: Optional[str], kind: str) -> None:
```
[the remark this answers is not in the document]
komalali: Should the `ABC` specify default values?

sdk/python/lib/pulumi/x/automation/local_workspace.py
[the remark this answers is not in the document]
justinvp: The inconsistency across languages feels odd to me. Should we change Node.js and .NET to be synchronous? @leezen, @lukehoban, what do you think?

EvanBoyle:
Working through a full review today, but one initial question I have is about python documentation. For got we have [godoc](https://pkg.go.dev/github.com/pulumi/pulumi/sdk/v2/go/x/auto), node has the generated [language sdk docs](https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/pulumi/x/automation/), but it looks like our [python docs](https://www.pulumi.com/docs/reference/pkg/python/pulumi/) are [hand-curated to some degree](https://github.com/pulumi/docs/blob/ec2de7efdeb753728d72d2fe7c9f9511d44fc222/content/docs/reference/pkg/python/pulumi/_index.md#L1141). What's our plan for this? Do we have any way to preview/render the current state of the docs? 

It would be nice to do a visual inspection of the docs. I notice in the review that there are some class-level comments missing here or there (stack.py).

EvanBoyle:
Brought up [elsewhere](https://github.com/pulumi/pulumi/issues/5299#issuecomment-754099707), but resurfacing here as it relates to supported python version, dependencies, and tooling. It might be nice to see one or two of the full [automation-api-examples](https://github.com/pulumi/automation-api-examples) in the context of this review. Might help us identify if there are any idiosyncracies WRT dependencies or setup for a standalone project that we might not notice from unit tests alone.

This might also be a good way to kick the tires on documentation/autocomplete from within an IDE.

sdk/python/lib/pulumi/runtime/settings.py
```diff
@@ -232,8 +235,31 @@ def do_rpc_call():
...
+
 async def monitor_supports_secrets() -> bool:
     return await monitor_supports_feature("secrets")
 
+
 async def monitor_supports_resource_references() -> bool:
     return await monitor_supports_feature("resourceReferences")
+
+
+def reset_options(project: Optional[str] = None,
```
EvanBoyle, line 247: Do we have an issue tracking the global state issues in the python SDK? 

We have an issue open for Node to track this. Might be worth creating a new issue for python or updating the node issue. https://github.com/pulumi/pulumi/issues/5449

sdk/python/lib/pulumi/x/automation/local_workspace.py
komalali: Never mind, going to do this in a loop until we sort out https://github.com/pulumi/pulumi/issues/6050

sdk/python/lib/pulumi/x/automation/local_workspace.py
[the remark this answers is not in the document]
EvanBoyle: IMO inconsistency across languages is appropriate in the name of embracing idiomatic _language culture_. Ultimately we want to create experiences that are pleasing to use in the context of expectations of the broader language community.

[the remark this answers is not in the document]
komalali: I vehemently agree with @EvanBoyle

sdk/python/lib/pulumi/x/automation/workspace.py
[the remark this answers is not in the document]
komalali: This feels very nit-picky, but I'll change it to `name`.

sdk/python/lib/pulumi/x/automation/local_workspace.py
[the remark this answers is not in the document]
justinvp: > IMO inconsistency across languages is appropriate in the name of embracing idiomatic _language culture_.

I totally agree! But it seems like if Go and Python can be synchronous, why are we making Node.js and .NET async, when they could also just as easily be synchronous?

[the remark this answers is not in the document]
EvanBoyle: Again, I think it has to do with expectations. Take a look at the nodejs file APIs for instance. You have `read` and `readSync` where async/callback driven is the default expectation: https://nodejs.org/api/fs.html#fs_dir_readsync

For Go, I expect in the fullness of time we'll support both sync and async operations similar to the APIs offered by the `exec.Cmd` API: https://golang.org/pkg/os/exec/#Cmd (mainly `cmd.Run()` and cmd.Start() style invocation). 

Picking async/sync should be based on what feels natural and results in the fewest surprises for each given language. Ultimately we are the only ones who really switch between language APIs frequently, so any burden created via inconsistency falls primarily on us alone.

sdk/python/lib/pulumi/x/automation/server.py
```diff
@@ -0,0 +1,80 @@
...
+
+        # Reset globals
+        reset_options()
+        if not self.running:
+            raise RunError(self.get_exit_error(preview))
+
+    def GetRequiredPlugins(self, request, context):
+        return language_pb2.GetRequiredPluginsResponse()
+
+    def Run(self, request, context):
```
EvanBoyle, line 52: I have some questions about the error handling strategy here. I'm not overly familiar with the python runtime, so please excuse any silly questions I may be asking. I think it's worth talking through given some of the issues that have popped up with node recently.

In typescript, we add a set of uncaught exception handlers in addition to an outer try/catch block: 

https://github.com/pulumi/pulumi/blob/39297cb9ddb4e2499db72dc64175b6352e9ab7a8/sdk/nodejs/x/automation/server.ts#L96-L117

It doesn't seem like we have anything equivalent. It appears on the surface that an unhandled exception in user code could potentially bubble up unhandled, and short circuit our calls to `on_exit`. `run_pulumi_func` does have a try/catch block, but it can rethrow: 

https://github.com/pulumi/pulumi/blob/f36431106443fc431076b43df68b03538eae0855/sdk/python/lib/pulumi/runtime/stack.py#L74

I guess this isn't so much a question, as it is a desire to better understand the error handling strategy. Do we know what happens if we write a `program` that throws an exception or fails for some reason? 

If there's uncertainty here, we may be able to work through some of this together. I need to dive into the typescript error handling approach tomorrow anyways. I don't have a great deal of familiarity with gRPC so I'm not sure if the original implementations really are handling errors in the server appropriately. I did find that guide that has some code snippets for multi-language server error handling, but I'm not yet sure if these are applicable to our use case. https://avi.im/grpc-errors/#python

sdk/python/lib/pulumi/runtime/settings.py
komalali: We do now! https://github.com/pulumi/pulumi/issues/6052

EvanBoyle (approved):
Overall this looks great! In terms of correctness and surface area, this is what I would expect. The test coverage seems to handle most of the important stuff. 🎉 🚀 

There are a few things we probably want to take a look at.  but I don't think they need to block merging if the plan is to iterate on those before we make a public announcement:
- Validate and try to manually fuzz some of the error handling for bad user programs, gRPC server, etc.
- Bigger end to end examples. This helped draw focus on rough edges for other languages.
- Docs. I'm unsure what goes into this, but we can probably cut a separate ticket to take a look.

sdk/python/lib/pulumi/x/automation/server.py
komalali: Yeah, I kinda rushed through this bit - will spend some time looking into error handling and add some bad tests.

komalali:
@EvanBoyle I think you encapsulated the immediate issues pretty well. 

Here's my plan before I merge this PR:
1. Address some of the PR feedback especially re: error handling in inline programs.
2. Port over some of the examples in https://github.com/pulumi/automation-api-examples

RE: docs, I opened #6053 and will look into it once this PR is merged.

sdk/python/lib/pulumi/x/automation/local_workspace.py
```diff
@@ -0,0 +1,399 @@
...
+    providing various utilities such as plugin installation, environment configuration
+    ($PULUMI_HOME), and creation, deletion, and listing of Stacks.
+    LocalWorkspace relies on Pulumi.yaml and Pulumi.<stack>.yaml as the intermediate format
+    for Project and Stack settings. Modifying ProjectSettings will
+    alter the Workspace Pulumi.yaml file, and setting config on a Stack will modify the Pulumi.<stack>.yaml file.
+    This is identical to the behavior of Pulumi CLI driven workspaces.
+    """
+
+    @classmethod
+    def new_stack(cls,
```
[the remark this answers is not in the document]
komalali: I changed it to `create_new_stack` and `select_existing_stack` and kept `create_or_select_stack`.

sdk/python/lib/pulumi/x/automation/stack.py
```diff
@@ -0,0 +1,442 @@
...
+        args.extend(["--exec-kind", kind])
+
+        try:
+            up_result = self._run_pulumi_cmd_sync(args, on_output)
+            outputs = self.outputs()
+            summary = self.info()
+            assert (summary is not None)
+            return UpResult(stdout=up_result.stdout, stderr=up_result.stderr, summary=summary, outputs=outputs)
+        except CommandError as exn:
+            if on_exit is not None:
```
EvanBoyle, line 237: I think the cleanup routine in on_exit is something we want to make sure to call in both success and failure scenarios. Since we're dealing with funky global state, always resetting when we're done can help avoid hard to diagnose bugs in the future. 

For example, if the update finishes and the user for some reason decided to call `register_resource` outside of the inline program. In the success case, I believe there would still be a monitor configured. The user would be able to make the request, but the engine and monitor servers would be shut down and they would get an obscure rcp dial error.

sdk/python/lib/pulumi/x/automation/stack.py
```diff
@@ -0,0 +1,442 @@
...
+
+        args.extend(["--exec-kind", kind])
+
+        try:
+            up_result = self._run_pulumi_cmd_sync(args, on_output)
+            outputs = self.outputs()
+            summary = self.info()
+            assert (summary is not None)
+            return UpResult(stdout=up_result.stdout, stderr=up_result.stderr, summary=summary, outputs=outputs)
+        except CommandError as exn:
```
EvanBoyle, line 236: Are there other categories of error that could occur or are we guaranteed to get a command error? Maybe a finally block is the answer to this and my other questions about always running `on_exit` but not sure what the pythonic answer is.

EvanBoyle (approved):
Looks great to me! Two questions about the cleanup gRPC cleanup routine. 

I'd recommend trying to manually validate (probably don't need to add a full test) one additional error case. The failure mode for gRPC requests and the provider chain is a little more complicated to reason about. A program like this will fail without the GCP project and zone set. Might be work running the equivalent in python just to make sure the error is caught and the automation script terminates gracefully (no hangs);

```ts
const run = async () => {
    // This is our pulumi program in "inline function" form
    const pulumiProgram = async () => {
        // this call fails if gcp:project and gcp:zone are not set. 
        const engineVersion = gcp.container.getEngineVersions().then((v) => v.latestMasterVersion);
        return {
            engv: engineVersion
        };
    };

    // Create our stack 
    const args: InlineProgramArgs = {
        stackName: "dev1",
        projectName: "inlineNode",
        program: pulumiProgram
    };

    // create (or select if one already exists) a stack that uses our inline program
    const stack = await LocalWorkspace.createOrSelectStack(args);

    console.info("updating stack...");
    try {
        const upRes = await stack.up({ onOutput: console.info });
        console.log(`update summary: \n${JSON.stringify(upRes.summary.resourceChanges, null, 4)}`);
    }
    catch (e){
        console.log(e)
        process.exit(1)
    }
};

run().catch(err => {
    console.log("there was an error")
    console.log(err)
    console.log("error logged")
});
```

komalali:
@EvanBoyle Yeah, I did some more manual testing last night and have been testing what happens when there are failures related to incorrect config. We always result in a `CommandError` and the test fails (I haven't experienced any hanging), but there is some RPC cleanup issues that I'm trying to track down. For example, if the AWS credentials are incorrect, this is the output:

```
Updating (dev)

View Live: https://app.pulumi.com/komalali/inline_s3_project/dev/updates/51


+  pulumi:pulumi:Stack inline_s3_project-dev creating
aws:s3:Bucket s3-website-bucket  error: 1 error occurred:
+  pulumi:pulumi:Stack inline_s3_project-dev created
aws:s3:Bucket s3-website-bucket **failed** 1 error

Diagnostics:
aws:s3:Bucket (s3-website-bucket):
error: 1 error occurred:
* error configuring Terraform AWS Provider: error validating provider credentials: error calling sts:GetCallerIdentity: InvalidClientTokenId: The security token included in the request is invalid
status code: 403, request id: d06b2ae2-d83e-467e-8ea0-8ad6ed415c8e

Resources:
+ 1 created

Duration: 2s

Task was destroyed but it is pending!
task: <Task pending name='Task-30' coro=<RPCManager.do_rpc.<locals>.rpc_wrapper() running at /Users/komal/go/src/github.com/pulumi/pulumi/sdk/python/env/src/pulumi/runtime/rpc_manager.py:67> wait_for=<Task finished name='Task-83' coro=<register_resource.<locals>.do_register() done, defined at /Users/komal/go/src/github.com/pulumi/pulumi/sdk/python/env/src/pulumi/runtime/resource.py:375> exception=SystemExit(0)>>
---------------------------------------------------------------------------

CommandError                              Traceback (most recent call last)

<ipython-input-7-e90392d9f360> in <module>
----> 1 up_output = stack.up(on_output=print)
      2 

~/go/src/github.com/pulumi/pulumi/sdk/python/env/src/pulumi/x/automation/stack.py in up(self, parallel, message, target, expect_no_changes, target_dependents, replace, on_output, program)
    240 
    241         try:
--> 242             up_result = self._run_pulumi_cmd_sync(args, on_output)
    243             outputs = self.outputs()
    244             summary = self.info()

~/go/src/github.com/pulumi/pulumi/sdk/python/env/src/pulumi/x/automation/stack.py in _run_pulumi_cmd_sync(self, args, on_output)
    417         args.extend(additional_args)
    418 
--> 419         result = _run_pulumi_cmd(args, self.workspace.work_dir, envs, on_output)
    420         self.workspace.post_command_callback(self.name)
    421         return result

~/go/src/github.com/pulumi/pulumi/sdk/python/env/src/pulumi/x/automation/cmd.py in _run_pulumi_cmd(args, cwd, additional_env, on_output)
     74     result = CommandResult(stderr=stderr_contents, stdout='\n'.join(stdout_chunks), code=code)
     75     if code != 0:
---> 76         raise create_command_error(result)
     77 
     78     return result

CommandError: 
 code: 255
 stdout: Updating (dev)

View Live: https://app.pulumi.com/komalali/inline_s3_project/dev/updates/51


+  pulumi:pulumi:Stack inline_s3_project-dev creating
aws:s3:Bucket s3-website-bucket  error: 1 error occurred:
+  pulumi:pulumi:Stack inline_s3_project-dev created
aws:s3:Bucket s3-website-bucket **failed** 1 error

Diagnostics:
aws:s3:Bucket (s3-website-bucket):
error: 1 error occurred:
* error configuring Terraform AWS Provider: error validating provider credentials: error calling sts:GetCallerIdentity: InvalidClientTokenId: The security token included in the request is invalid
status code: 403, request id: d06b2ae2-d83e-467e-8ea0-8ad6ed415c8e

Resources:
+ 1 created

Duration: 2s

 stderr: 
```

Notice, specifically: 
```Task was destroyed but it is pending!
task: <Task pending name='Task-30' coro=<RPCManager.do_rpc.<locals>.rpc_wrapper() running at /Users/komal/go/src/github.com/pulumi/pulumi/sdk/python/env/src/pulumi/runtime/rpc_manager.py:67> wait_for=<Task finished name='Task-83' coro=<register_resource.<locals>.do_register() done, defined at /Users/komal/go/src/github.com/pulumi/pulumi/sdk/python/env/src/pulumi/runtime/resource.py:375> exception=SystemExit(0)>>```

I'm not exactly sure what's going on but will spend some time spelunking. 

I've also been using jupyter notebooks in PyCharm to vet the outputs and whatnot and have some dev-experience-related improvements incoming.

**Comment by EvanBoyle** [comment]
@komalali have you spent time looking at the python exec shim? It might be worth comparing our handling of errors here to that of the shim.

Specifically, it looks like the leaked futures may be expected based on this comment. That being said, I'm not sure what that means for the output we show to the user, and whether or not it makes sense to suppress any of it in the automation API context. Probably needs some careful thought and consideration. 

https://github.com/pulumi/pulumi/blob/e3725a6e409e5b6f1be732084eb8265698c6b7a9/sdk/python/cmd/pulumi-language-python-exec#L66-L81

File: sdk/python/lib/pulumi/x/automation/local_workspace.py (hunk 10 above)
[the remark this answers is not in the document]
Comment (komalali) [reply sdk/python/lib/pulumi/x/automation/local_workspace.py]: After getting some use with the API, I decided to extract the class methods on `LocalWorkspace` into standalone functions `create_stack`, `select_stack` and `create_or_select_stack`. The slightly differing names for the methods on LocalWorkspace felt clunky in the IDE with autocomplete.

I noticed that the [Go API](https://github.com/pulumi/pulumi/blob/master/sdk/go/x/auto/local_workspace.go#L712-L902) uses factory functions itself, so the discrepancy between language implementations to align with language culture is already present.

We also have the option of overwriting the class method with an instance-specific version as suggested in [this post](https://stackoverflow.com/a/28766809) but I wasn't really loving that.

I think the standalone functions are pythonic, easy to use, are aligned with the method names in `LocalWorkspace` and `Stack` and `auto.create_stack()` reads better than `auto.LocalWorkspace.create[_new]_stack()`.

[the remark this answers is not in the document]
Comment (komalali) [reply sdk/python/lib/pulumi/x/automation/local_workspace.py]: Notice though, the Go SDK uses `upsert` instead of `create_or_select`. Do we want to do that in python? Or is `create_or_select_stack` the way to go? Another possible option would be to eliminate the 3rd function and have a bool argument to `create_stack`, something like `error_if_exists` that is default `True` and setting to `False` is effectively `create_or_select`.

File: sdk/python/lib/pulumi/x/automation/server.py [hunk 13]
```diff
@@ -0,0 +1,81 @@
...
+            return result
+        except Exception as exn:
+            result.error = str(f"python inline source runtime error: {exn}\n{traceback.format_exc()}")
+            return result
+        finally:
+            loop.close()
+            sys.stdout.flush()
+            sys.stderr.flush()
+            # Reset the asyncio logger to the original level
+            logging.getLogger("asyncio").setLevel(asyncio_log_level)
```
komalali, line 75: @EvanBoyle this is my proposed solution for suppressing the asyncio output for the duration of the `run_in_stack` function. I figure we don't want to completely suppress it as we do in https://github.com/pulumi/pulumi/blob/master/sdk/python/cmd/pulumi-language-python-exec#L66, so I turn it off only for the duration of the `Run()` function and then reset to the original level during cleanup. Do you foresee any issues with this strategy? 

It means concurrent executions in the same process would run into issues since the `asyncio` logger is global. Concurrent executions in the same process aren't possible currently anyway due to https://github.com/pulumi/pulumi/issues/6052, but they should still be possible using [multiprocessing](https://docs.python.org/3/library/multiprocessing.html)

komalali: Actually, thinking about this some more, I suppose we could just eat the `asyncio` logs when we [process `stdout`](https://github.com/pulumi/pulumi/pull/5979/files#diff-729dbfc166002236b2941fa41696b4e269e15abbca76b057408b3dee5ca5c708R59-R67) rather than changing the logging level. That way it is unaffected by concurrency issues.

sdk/python/lib/pulumi/x/automation/stack.py
```diff
@@ -0,0 +1,532 @@
...
+        self.result = result
+        self.Deployment = deployment
+        self.resource_changes = resource_changes
+        self.version = version
+        self.config: ConfigMap = {}
+        for key in config:
+            self.config[key] = ConfigValue(**config[key])
+
+    def __repr__(self):
+        return f"UpdateSummary(result={self.result!r}, version={self.version!r}, start_time='{self.start_time!s} UTC', end_time='{self.end_time!s} UTC')"
```
justinvp, line 103: Now that the times are `datetime` should these args be shown as strings here?

sdk/python/lib/pulumi/x/automation/stack.py
```diff
@@ -0,0 +1,532 @@
...
+    stderr: str
+    summary: UpdateSummary
+
+    def __init__(self, stdout: str, stderr: str, summary: UpdateSummary):
+        self.stdout = stdout
+        self.stderr = stderr
+        self.summary = summary
+
+    def __repr__(self):
+        return f"{self.__class__.__name__}(result={self.summary.result!r}, start_time='{self.summary.start_time!s} UTC', end_time='{self.summary.end_time!s} UTC')"
```
justinvp, line 117: This doesn't look right (this class doesn't have `result`, `start_time`, etc. args).

sdk/python/lib/pulumi/x/automation/stack.py
```diff
@@ -0,0 +1,532 @@
...
+
+    def __init__(self, name: str, workspace: Workspace, mode: StackInitMode) -> None:
+        self.name = name
+        self.workspace = workspace
+
+        if mode is StackInitMode.CREATE:
+            workspace.create_stack(name)
+        elif mode is StackInitMode.SELECT:
+            workspace.select_stack(name)
+        elif mode is StackInitMode.CREATE_OR_SELECT:
```
justinvp, line 193: Raise an error if `mode` isn't one of the known values?

sdk/python/lib/pulumi/x/automation/stack.py
```diff
@@ -0,0 +1,532 @@
...
+            try:
+                workspace.create_stack(name)
+            except StackAlreadyExistsError:
+                workspace.select_stack(name)
+
+    def __repr__(self):
+        return f"Stack(stack_name={self.name!r})"
```
justinvp, line 200: What about the other args?

sdk/python/lib/pulumi/x/automation/stack.py
```diff
@@ -0,0 +1,532 @@
...
+
+        :param stack_name: The name identifying the Stack
+        :param workspace: The Workspace the Stack was created from.
+        :return: Stack
+        """
+        return Stack(stack_name, workspace, StackInitMode.CREATE_OR_SELECT)
+
+    def __init__(self, name: str, workspace: Workspace, mode: StackInitMode) -> None:
+        self.name = name
+        self.workspace = workspace
```
justinvp, line 187: Check these args and raise errors if name isn't a string? Workspace isn't a `Workspace`?

justinvp (approved):
Minor nits, but LGTM! Nice work!

pgavlin (approved):
This looks great--well done!

sdk/python/lib/pulumi/x/automation/stack.py
komalali: The `repr()` for datetime objects isn't particularly helpful... and even pycharm uses the str representation for debug. I'll go ahead and use it as it is represented elsewhere (without the quotes around it and the extra UTC - but in that case the timezone info is not obvious from the `repr()` but maybe that's not a big deal)

Screen Shot 2021-01-12 at 1 40 45 PM

justinvp: Shouldn't we show whatever values you'd have to pass to create the same `datetime` instance?

```
UpdateSummary(result='succeeded', version=32, start_time=datetime(<todo>), end_time=datetime(<todo>))
```

?

komalali: Yeah, I hear what you're saying. I'll implement `__str__` functions for each of these too because that's really the experience I'm trying to improve here.

PR #5983, opened 2020-12-19
Do not read TGZs into memory.

pgavlin:
This runs a serious risk of exhausting the memory on lower-end machines
(e.g. certain CI VMs), especially given the potential size of some
plugins.


pgavlin:
@jkisk I suspect that the service CI failures may be related to this issue.

lukehoban (approved):

PR #5992, opened 2020-12-22
Fix virtual env path

komalali:
There was a [regression](https://github.com/pulumi/pulumi/pull/5787/files#diff-4a0aed9481c8b2728775fb1fd8bf00087e85c56b10002168c675e4c4c32f2325L147-L148) introduced in #5787 that causes the `venv` for the policy pack to not be created correctly. This PR fixes the regression.

Fixes: https://github.com/pulumi/pulumi-policy/issues/262

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,13 @@ CHANGELOG
 =========
 
 ## HEAD (Unreleased)
-_(none)_
+
+- Fix regression in `venv` creation for python policy packs.
+  [#5992](https://github.com/pulumi/pulumi/pull/5992)
 
 ## 2.16.0 (2020-12-21)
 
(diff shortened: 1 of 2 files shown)


stack72 (approved):

lukehoban (approved):
Fix looks good. Is there a test case we could add that would ensure this is working?

komalali:
@lukehoban We could test this by publishing a python policy pack and applying it to one or more of our test stacks.

lukehoban:
> @lukehoban We could test this by publishing a python policy pack and applying it to one or more of our test stacks.

FWIW - looks like we should address these to ensure we have coverage here (which would likely have caught this regression):
* https://github.com/pulumi/pulumi-policy/issues/247
* https://github.com/pulumi/pulumi/issues/4149

PR #3005, opened 2019-07-30
Pass ignoreChanges to providers.

pgavlin:
These changes add support for passing `ignoreChanges` paths to resource
providers. This is intended to accommodate providers that perform diffs
between resource inputs and resource state (e.g. all Terraform-based
providers, the k8s provider when using API server dry-runs). These paths
are specified using the same syntax as the paths used in detailed diffs.

In addition to passing these paths to providers, the existing support
for `ignoreChanges` in inputs has been extended to accept paths rather
than top-level keys. It is an error to specify a path that is missing
one or more component in the old or new inputs.

Fixes #2936, #2663.

Issue #2663, opened 2019-04-22
Support paths in ignoreChanges


lukehoban:
In #2657 we added the ability to ignore change to top-level properties of a resource for the purposes of updates and deletes via the `ignoreChanges` resource option.

We should expand this to also allow specifying nested resources. This will likely require a custom path syntax to allow paths through JSON-like Pulumi resource objects/arrays.

pgavlin:
This should use the path syntax introduced by detailed diffs.

pgavlin:
This is complete.

Issue #4861, opened 2020-06-21
Feature request: Support ACME provider for provisioning certificates


UnoSD:
Can you please consider supporting the orchestration of certificate provisioning and renewal in idiomatic Pulumi resources?

Terraform has the following provider that can be used as a base https://www.terraform.io/docs/providers/acme/index.html

Thank you

fa-bwilliams:
I second this. For now, I am using the TF ACME provider and exporting the certs to pull into other Pulumi projects.

PR #6011, opened 2020-12-25
Illustrate ignoreChanges diffing issue

MythicManiac:
I'm not well-versed enough in the codebase to be able to fix the issue, but at least I should be able to illustrate it by modifying the existing unit tests a bit.

Currently it's either impossible to properly match deep structures in `ignoreChanges`, or the syntax for doing so is undocumented or at the very least hard to find. I was able to ignore properties just fine to the point where some key contains the same character as pulumi uses for separating nested props.

I would propose adding some syntax to escape this behavior for scenarios where the path separator is supposed to be a part of the key identifier. Alternatively the `ignoreChanges` syntax could be a nested structure in itself, rather than a string.

### This is pull request does not attempt to fix the issue, just illustrate it via tests

Related issue: #2663
Related PR: #3005

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ CHANGELOG
 ## HEAD (Unreleased)
 _(none)_
 
+- Fix an issue in the `ignoreChanges` diffing logic which prevented deep matching properties
+  which used a period in their path keys.
+
 ## 2.16.2 (2020-12-23)
 
 - Fix a bug in the core engine that could cause previews to fail if a resource with changes had
--- a/pkg/engine/lifeycletest/pulumi_test.go
+++ b/pkg/engine/lifeycletest/pulumi_test.go
@@ -902,23 +902,23 @@ func TestSingleResourceIgnoreChanges(t *testing.T) {
 	snap := updateProgramWithProps(nil, resource.NewPropertyMapFromMap(map[string]interface{}{
 		"a": 1,
 		"b": map[string]interface{}{
-			"c": "foo",
+			"c.d": "foo",
 		},
-	}), []string{"a", "b.c"}, []deploy.StepOp{deploy.OpCreate})
+	}), []string{"a", "b.c.d"}, []deploy.StepOp{deploy.OpCreate})
 
 	// Ensure that a change to an ignored property results in an OpSame
 	snap = updateProgramWithProps(snap, resource.NewPropertyMapFromMap(map[string]interface{}{
 		"a": 2,
 		"b": map[string]interface{}{
-			"c": "bar",
+			"c.d": "bar",
 		},
-	}), []string{"a", "b.c"}, []deploy.StepOp{deploy.OpSame})
+	}), []string{"a", "b.c.d"}, []deploy.StepOp{deploy.OpSame})
 
 	// Ensure that a change to an un-ignored property results in an OpUpdate
 	snap = updateProgramWithProps(snap, resource.NewPropertyMapFromMap(map[string]interface{}{
 		"a": 3,
 		"b": map[string]interface{}{
-			"c": "qux",
+			"c.d": "qux",
 		},
 	}), nil, []deploy.StepOp{deploy.OpUpdate})
 


MythicManiac:
I was hoping your CI runners would run tests but apparently that does not happen. There also isn't very much documentation on how to set up the development environment for this project to be able to run the tests locally, and it does not work out of the box with make, so looks like I won't be verifying if the tests actually fail (as would be the intention).

MythicManiac:
A practical example where you can hit this issue:

When deploying the [kubed](https://appscode.com/products/kubed/v0.12.0/setup/install/) helm chart, it will automatically generate a TLS certificate for you. This means that whenever pulumi is run, a new certificate will be generated when rendering the chart, and pulumi will think the state is differing, attempting to re-deploy the certificate.

To get around this issue, I've set up the following transformer on the helm chart:
```typescript
const ignoreCertChanges = (args: ResourceTransformationArgs): ResourceTransformationResult | undefined => {
    if (args.type === "kubernetes:core/v1:Secret" && args.name === "kube-system/default-kubed-apiserver-cert") {
        return {
            props: args.props,
            opts: _.merge(args.opts, { ignoreChanges: ["data.tls.crt"] }),
        }
    }
    if ("kubernetes:apps/v1:Deployment" && args.name === "kube-system/default-kubed") {
        return {
            props: args.props,
            opts: _.merge(args.opts, {
                ignoreChanges: ["spec.template.metadata.annotations.checksum/apiregistration.yaml"]}),
        }
    }
    return undefined;
}
```

The issue is, pulumi now thinks the structure of the ignores is as follows:
```yaml
# Secret
data:
  tls:
    crt

# Deployment
spec:
  template:
    metadata:
      annotations:
        checksum/apiregistration:
          yaml
```
when in reality it's as follows:
```yaml
# Secret
data:
  tls.cert

# Deployment
spec:
  template:
    metadata:
      annotations:
          template:
    metadata:
      annotations:
        checksum/apiregistration.yaml
```

The only workaround I could come up with right now is to ignore the closest key that pulumi is able to match, which unfortunately in the deployment's case for example ignores all annotations.

lukehoban:
> ```ts
>  ignoreChanges: ["spec.template.metadata.annotations.checksum/apiregistration.yaml"]}),
> ```

I expect this should work with:

```ts
ignoreChanges: ['spec.template.metadata.annotations["checksum/apiregistration.yaml"]']}),
```

> Currently it's either impossible to properly match deep structures in ignoreChanges, or the syntax for doing so is undocumented or at the very least hard to find.

You are right that this is un(der)-documented. We should definitely add this to official documentation, but for now, note that the the allowed formats for `ignoreChanges` paths are the ones documented at https://github.com/pulumi/pulumi/blob/master/sdk/go/common/resource/properties_path.go#L14-L43.

lukehoban:
> We should definitely add this to official documentation

See https://github.com/pulumi/docs/pull/4861.

MythicManiac:
Looks like the syntax you suggested works, thanks!

Closing this PR as the functionality does exist, and the actual issue was a lack of documentation that should be resolved by #4861

Sytten:
It doesn't seem to work on gcp cloud run service when do something like:
```typescript
ignoreChanges: [
          'metadata.annotations["client.knative.dev/user-image"]',
          'metadata.annotations["run.googleapis.com/client-name"]',
          'metadata.annotations["run.googleapis.com/client-version"]',
          'metadata.annotations["run.googleapis.com/ingress-status"]',
          'metadata.labels',
          'template.spec.containers[0].image',
          'template.metadata.annotations["client.knative.dev/user-image"]',
          'template.metadata.annotations["run.googleapis.com/client-name"]',
          'template.metadata.annotations["run.googleapis.com/client-version"]',
        ],
```

MythicManiac:
@Sytten are you sure those are the actual paths of the resource? If you're deploying a helm chart, you might have to dig into the specifc resource with a transformer.

Here's what I use for the kubed chart for example:
```ts
const ignoreCertChanges = (args: ResourceTransformationArgs): ResourceTransformationResult | undefined => {
    if (args.type === "kubernetes:core/v1:Secret" && args.name === "kube-system/default-kubed-apiserver-cert") {
        return {
            props: args.props,
            opts: _.merge({}, args.opts, { ignoreChanges: ['data["tls.crt"]', 'data["tls.key"]'] }),
        }
    }
    if ("kubernetes:apps/v1:Deployment" && args.name === "kube-system/default-kubed") {
        const ignore = 'spec.template.metadata.annotations["checksum/apiregistration.yaml"]';
        return {
            props: args.props,
            opts: _.merge({}, args.opts, { ignoreChanges: [ignore] }),
        }
    }
    return undefined;
}


export class Kubed extends Chart {
    constructor(name: string, opts?: ComponentResourceOptions) {
        const kubedOpts = _.merge({}, opts, {
            transformations: [ignoreCertChanges],
        });
        super(name, {
            path: "./kubernetes/helm/kubed-v0.12.0.tgz",
            namespace: "kube-system",
        }, kubedOpts)
    }
}```

Sytten:
Yes, I am not building a k8s cluster I am building Cloud Run services. And it "works" if I ignore all the annotations but then I cant modify the one I care about.

PR #6013, opened 2020-12-25
Allow `serializeFunction` to capture secrets

### LINKED ISSUE #2718: SerializeFunction should communicate if a secret output was captured
With #2678 we now have the ability to mark `Output<T>` as secrets, such that when they flow into resources they are stored in an encrypted form in the checkpoint.  We also flow this information across `all`s and `apply`'s, to track secrets across your program.

However, the one way a secret can leak today is if it is captured by a lambda, which is then serialized.  For now, the serialized function text has to contain the raw value (in order to build an `Output<T>` at Runtime that has the correct value) but there is no way for SerializeFunction to communicate that it captured a secret value and the program text is typed as `string` so we can not make it an `Output<T>`.  This means if you create a cloud function using `pulumi` and secret output, the value will exist in the contents of the text asset for the function source that is stored as part of the deployment.

We should augment SerializeFunction such that it is able to notice one or more secrets were captured, and communicate that as part of the SeralizedFunction return value.  Higher level code which them passes that to a resource would then be able to wrap that in an secret output.  Alternately, we can deprecate this function and create a new version which types text as `Output<string>` such that the right thing happens.

When we make this change, we should move the dynamic provider and the code in `@pulumi/{aws,azure,gcp}` over to using the new form.

lukehoban:
Adds an opt-in `allowSecrets` flag to `serializeFunction` to allow it to capture secrets.  If passed, `serializeFunction` will now report back whether it captured any secrets.  This information can be used by callers to wrap the resulting text in a Secret value.

Fixes #2718.


stack72 (approved):
my TypeScript foo isn't brilliant - but this works as it says on the tin so the only other thing I could offer would be opinionated comments that wouldn't help the functionality 😃

PR #6014, opened 2020-12-27
[sdk/go] Support maps in Invoke outputs and Read inputs

lukehoban:
These are already supported by the implementation, but were prevented by overzealous input validation in Invoke and ReadResource.

Follow-up to #4522 and #4521.


EvanBoyle (approved):

pgavlin (commented):
We should check the map's key type as well: https://github.com/pulumi/pulumi/blob/master/sdk/go/pulumi/context.go#L479-L488

pgavlin (approved):

PR #6016, opened 2020-12-28
Allow VERSION to be set explicitly

Frassle:
When building for alpine aports we download the pulumi source from a
github archive (e.g.
https://github.com/pulumi/pulumi/archive/v2.16.0.tar.gz) but this
doesn't include the .git information to allow scripts/get-version to
work.

This patch, which is currently applied on top of the base pulumi
code as part of the alpine build (see [1]), simply allows the alpine
build script to explicitly set the version used. In the alpine build
script this always matches the source code archive version to download.

[1] https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/16281

### DIFF
--- a/scripts/get-version
+++ b/scripts/get-version
@@ -7,7 +7,7 @@ if [ $# -eq 0 ]; then
 fi
 
 # Allow version to be set manually when building outside of git repo
-if [ ! -z ${PULUMI_VERSION+x} ]; then
+if [ -n ${PULUMI_VERSION+x} ]; then
     echo "${PULUMI_VERSION}"
     exit 0
 fi
--- a/scripts/get-version
+++ b/scripts/get-version
@@ -7,7 +7,7 @@ if [ $# -eq 0 ]; then
 fi
 
 # Allow version to be set manually when building outside of git repo
-if [ -n ${PULUMI_VERSION+x} ]; then
+if [ -n "${PULUMI_VERSION+x}" ]; then
     echo "${PULUMI_VERSION}"
     exit 0
 fi
--- a/scripts/get-version
+++ b/scripts/get-version
@@ -6,6 +6,12 @@ if [ $# -eq 0 ]; then
     exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


Frassle:
I've been doing some work on the alpine build script today and this patch isn't quite right. I'll do some more testing and get back to this once I've sorted the alpine build out.

Frassle:
I've now updated this for the latest version of what I'm using in the alpine pipeline (Still waiting for the merge request to be merged in aports).

Frassle:
This patch is now merged on the alpine side. I've got one more large change to the alpine pipeline to get it working properly and the only patch needed to the base pulumi code for that pipeline is what's in this PR. 
That is if this PR is merged the alpine pipeline won't have to patch the pulumi source at all to build, which would be nice.

stack72:
Thanks so much for this @Frassle 

I was able to get the correct behaviour here:

```
▶ make dist
DIST:
go install -ldflags "-X github.com/pulumi/pulumi/sdk/v2/go/common/version.Version=v2.19.0-alpha.1611576811+g95c03c623" github.com/pulumi/pulumi/sdk/v2/dotnet/cmd/pulumi-language-dotnet
^Cmake[1]: *** [dist] Interrupt: 2
make: *** [sdk/dotnet_dist] Interrupt: 2


github.com/pulumi/pulumi  Frassle-explictversion ✔                                                                                                                                     0m  ⍉
▶ PULUMI_VERSION=v100.x make dist
DIST:
go install -ldflags "-X github.com/pulumi/pulumi/sdk/v2/go/common/version.Version=v100.x" github.com/pulumi/pulumi/sdk/v2/dotnet/cmd/pulumi-language-dotnet
^Cmake[1]: *** [dist] Interrupt: 2
make: *** [sdk/dotnet_dist] Interrupt: 2
```

I just kicked off a set of acceptance tests to ensure our build process works as expected and will merge on a green build!

Thanks again

Paul

PR #6017, opened 2020-12-28
[codegen] Add support for default values in enum properties

mikhailshilkov:
While trying to implement https://github.com/pulumi/pulumi-azure-nextgen/issues/183 I noticed that schema binding fails for enum properties if they have a default value.

This PR binds the default value to the underlying type of the enum, both strict and "relaxed" (union-based).

It also adjusts the .NET codegen to set the default value to a strict enum in constructors.

### DIFF
--- a/pkg/codegen/schema/schema.go
+++ b/pkg/codegen/schema/schema.go
@@ -1281,6 +1281,21 @@ func bindDefaultValue(value interface{}, spec *DefaultSpec, typ Type) (*DefaultV
 	}
 
 	if value != nil {
+		switch typ := typ.(type) {
+		case *UnionType:
+			if typ.DefaultType != nil {
+				return bindDefaultValue(value, spec, typ.DefaultType)
+			}
+			for _, elementType := range typ.ElementTypes {
+				v, err := bindDefaultValue(value, spec, elementType)
+				if err == nil {
+					return v, nil
+				}
+			}
+		case *EnumType:
+			return bindDefaultValue(value, spec, typ.ElementType)
+		}
+
 		switch typ {
 		case BoolType:
 			if _, ok := value.(bool); !ok {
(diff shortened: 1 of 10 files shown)


pkg/codegen/schema/schema.go
```diff
@@ -1527,14 +1542,19 @@ func bindTypes(pkg *Package, complexTypes map[string]ComplexTypeSpec, loader Loa
 		typs.resources[r.Token] = &ResourceType{Token: r.Token}
 	}
 
-	// Process properties.
+	// Process enums.
```
mikhailshilkov, line 1545: I resolve enums before processing object types because the default value resolution expects `ElementTypes` and `DefaultType` to be populated. This looks safe but also seems not ideal: I'm not a fan of this implicit dependency in processing order. Is there a better way to resolve enums on demand?

komalali: There is already a processing order dependency in this function right above this section (https://github.com/pulumi/pulumi/blob/master/pkg/codegen/schema/schema.go#L1508-L1523). Seems reasonable to resolve the enums within that first loop rather than adding in an extra loop. It doesn't help with the order dependency but that is a pre-existing condition so you're not introducing it. I think clearly documenting the reasoning in a comment as we've done for declaring tokens above is sufficient.

komalali:
It would be great to add a test case in https://github.com/pulumi/pulumi/tree/master/pkg/codegen/internal/test/testdata/simple-enum-schema by setting a default value for one of the properties in the schema.json

mikhailshilkov:
I added two default values to the test. Apparently, our node and Go SDKs ignore default values for type properties. I will open an issue to track this once this PR is merged.

komalali (approved):
LGTM!

mikhailshilkov:
Opened https://github.com/pulumi/pulumi/issues/6025 to follow-up

PR #6019, opened 2020-12-29
Add suport to use pipenv in pulumi/actions

### LINKED ISSUE #6018: Support pipenv on the pulumi/actions container
Right now pulumi is highly coupled with the standard requirements.txt file for python dependencies.
But there may be cases in which this file is not used, like when using pipenv.

For these cases you need to maintain both the Pipfile and the requirements.txt

We could change the https://github.com/pulumi/pulumi/blob/master/docker/actions/entrypoint.sh file, to also check if there is a Pipfile.lock file and use pipenv to install the dependencies.

I was going to open the PR straight away, but wanted to get the feedback opinion from maintainers first.

This also could be a first step to make pulumi less coupled with venv and pip, and allow to use other methods like pipfile. i.e, when creating a new python project, allow the user to use pipenv, instead of installing the .venv straight away.

GaboFDC:
Fixes #6018 

About the PULUMI_PREFIX, we could also define PULUMI_COMMAND earlier, and overwrite it in the if.

I was not sure which approach was better. But open to suggestion, and can change it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@ CHANGELOG
   give the user some information on how to resolve the problem
   [#6044](https://github.com/pulumi/pulumi/pull/6044)
 
+- Add suport for pipenv on the github action
+  [6019](https://github.com/pulumi/pulumi/pull/6019)
+
 ## 2.16.2 (2020-12-23)
 
 - Fix a bug in the core engine that could cause previews to fail if a resource with changes had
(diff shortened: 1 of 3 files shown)


stack72:
Hey @GaboFDC 

this is closed via https://github.com/pulumi/actions/commit/ba58c77ae0471c1f154bd36645e61d9657222565

We just moved the actions work to this location

Paul

Issue #5453, opened 2020-09-24
Excessive chattiness to pulumi service


barclayadam:
The number of HTTP calls to api.pulumi.com seems overly excessive.

## Expected Behavior

I'd have expected a few things:

 - Caching of calls that are identical
 - Getting a small payload when requiring output from a previous stack, not an export with all resources
 - Batch requests. Not sure how viable, but is it possible to batch decrypt / encrypt, or possible even an endpoint that could grab the decrypted outputs from the service, meaning we reduce what is currently 62 calls to a single one that is reduced in size.

## Current Behavior

I have one project/stack that makes 337 calls to `api.pulumi.com`, consisting of:

 - Multiple calls to `/api/stacks/[user]/[project]/[stack]/decrypt`. 61 calls per stack (a few are required as it's multi-region and I need all). Multiple calls are exact duplicates for the same thing to decrypt.

 - Multiple calls to `/api/stacks/[user]/[project]/[stack]/encrypt`. Some of these are duplicate text values being encrypted.

 - Multiple calls to  `/api/stacks/[user]/[project]/[stack]/export`. Some of these are pretty large as they contain many resources (the largest is 6,944,217 bytes and takes over a second). The app only requires the _output_ values from the stack, not all of the resources too.

 - Multiple calls to `/api/stacks/[user]/[project]/[stack]/update`. As this is I'm assuming checkpointing progress assuming these are par for course

## Steps to Reproduce

1. Create a project/stack that references multiple other stacks, each one having a number of resources and secrets
2. Run the final stack and watch Fiddler or equivalent

## Context (Environment)

This massively decreases the speed of deployment. A recent run only managed to try the update of a single resource and failed (testing of a custom resource), but still took >6m.

Pulumi version: 2.10.0
OS: Windows
Language: TypeScript

barclayadam:
Following up on this as I have been progressing our use of Pulumi, and in particular, looking at using the new Automation APIs for managing our deployment scripts.

One feature I am making use of is setting the configuration of projects via. `stack.setAllConfig`. I have a central config project / stack that I load config from and then push to all stacks (which at the moment is 19).

Timing the call to `setAllConfig` shows that one call of that method takes >60s every time. With the around 3s to load the config from the shared stack I'm looking at nearly 20 minutes just moving some config values around. I have 32 config keys, 7 of which are encrypted.

This is a massive time sink at the moment and I'd love to know if this is something I could workaround in some way

barclayadam:
[...]

PR #6022, opened 2020-12-29
[auto/nodejs] - Parallelize config operations

komalali:
I noticed that #3877 was fixed so I took the opportunity to parallelize some config operations that have been mentioned in #5453. Also did some miscellaneous cleanup that the IDE was complaining about.


sdk/nodejs/x/automation/localWorkspace.ts
```diff
@@ -384,10 +382,11 @@ export class LocalWorkspace implements Workspace {
      * @param config The `ConfigMap` to upsert against the existing config.
      */
     async setAllConfig(stackName: string, config: ConfigMap): Promise<void> {
-        // TODO: do this in parallel after this is fixed https://github.com/pulumi/pulumi/issues/3877
+        const configAdds: Promise<any>[] = [];
```
lukehoban, line 385: Minor - but can this be?

```suggestion
        const configAdds: Promise<void>[] = [];
```

lukehoban (approved):

lukehoban:
Unfortunately, I'm fairly certain this change will actually introduce concurrency issues, even though #3877 was fixed.  Opened https://github.com/pulumi/pulumi/issues/6050 to track the remaining issue.  Sorry for missing that in my initial review of the PR - quite likely need to revert this for now and explore a slightly broader fix here.

komalali:
Okay, reverting.

PR #6027, opened 2020-12-30
Ensure only latest nuget packages are locally installed

stack72:
Fixes: #3882

This will clean the old packages as well as only "installing" the
latest built package

This ensures old packages are not "installed"

This will only affect local builds / installs


mikhailshilkov (approved):
LGTM

komalali (approved):

Issue #5854, opened 2020-12-02
Feature request: open the Pulumi Console with a single command


praneetloke:
Today, the `pulumi stack ls` command shows the corresponding Pulumi Console for each stack in a project. The links are clickable too. It would be nice to be able to launch the Pulumi Console in the browser by simply running `pulumi stack console` for the currently selected stack, as well as support the `-s` flag for passing a stack name. I tried searching for an existing issue but couldn't find one. Please feel free to close this as a dupe if one already exists.

stack72:
This is covered by:

```
▶ pulumi console --help
Opens the current stack in the Pulumi Console

Usage:
  pulumi console [flags]
```

It doesn't support adding `-s` so we can mark this as an enhancement

PR #6031, opened 2020-12-30
[CLI] Allow pulumi console to accept a stack name

stack72:
Fixes: #5854

If a stackname is not passed as an argument then it will use the
currently selected stack

### DIFF
--- a/pkg/cmd/pulumi/console.go
+++ b/pkg/cmd/pulumi/console.go
@@ -20,6 +20,7 @@ import (
 	"github.com/skratchdot/open-golang/open"
 	"github.com/spf13/cobra"
 
+	"github.com/pulumi/pulumi/pkg/v2/backend"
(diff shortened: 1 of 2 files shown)


pkg/cmd/pulumi/console.go
```diff
@@ -70,6 +88,8 @@ func newConsoleCmd() *cobra.Command {
 			return nil
 		}),
 	}
+	cmd.PersistentFlags().StringVarP(
+		&stackName, "stack", "s", "", "The name of the stack to create")
```
komalali, line 92:
```suggestion
		&stackName, "stack", "s", "", "The name of the stack to view")
```

komalali (approved):
Just a correction in the flag description, otherwise LGTM!

pkg/cmd/pulumi/console.go
stack72: Oh nice catch! I will update - thanks!

Issue #5531, opened 2020-10-08
[Automation API] Support recovery workflow for nodejs


EvanBoyle:
We should expose export/import/cancel similar to what we added for Go https://github.com/pulumi/pulumi/pull/5369

roderik:
I wanted to add some more context from some Slack threads because from my observations the issue is a bit more complex.

- https://pulumi-community.slack.com/archives/C019YSXN04B/p1609161459201200
- https://pulumi-community.slack.com/archives/C019YSXN04B/p1609409517206400

Some background first:

We use the Automation API in Typescript to deploy K8S clusters and services on top of them. This is not a manually run script nor a CI/CD system, but in a typescript based (NestJS) server that uses a job queue (Bull) to schedule all kinds of work, including pulumi up and destroy jobs.

While developing, this server restart on code changes (nodemon) killing jobs in the queue while they are in progress. When in the middle of a GKE cluster creation, this presents some obvious issues, how to continue/recover after the restart. The job queue will retry this job as it neither completed nor failed.

We noticed the following behavior (synthesis of the slack threads):

1. in the pulumi backend, the 'up' run would remain 'running'
2. in the computers process list, a zombie pulumi process was leftover
3. when retrying the `pulumi.up`, the backend reports a 409 (correct because of 1)
4. no code exists to manually handle recovery, cancel, import, export.

5. A secondary issue is that some stack configuration issues, which cause unhandled exceptions in the pulumi binary, do not bubble upwards through the Automation API so they can be caught and in my case, mark the job as failed. The `await up` call would just stall and never complete, locking up the job forever, requiring a restart, and then we have the same effect as the paragraph before.

An example of this is doing `up` to GKE to deploy a GKE cluster with a location set in the GKE cluster config, but no zone in `gcp:zone` configuration. This might be the 6th issue since it is listed as optional in the docs, but it is not.

Now I see there are PR's open for some of the recovery options in 4, but I feel like 2 and 5 are actually closer to the root cause that would make 4 (while useful) not needed.

As to what I (from the outside) think needs to be solved is the following:

1. Exceptions in the pulumi stack configuration/up should bubble up to the Automation API and the calling code. That would solve hanging await up's and allow calling code to deal with it as it sees fit.
[...]

PR #6038, opened 2020-12-31
[automation-api/nodejs] - Support recovery workflow

komalali:
Resolves: https://github.com/pulumi/pulumi/issues/5531

### DIFF
--- a/sdk/nodejs/x/automation/localWorkspace.ts
+++ b/sdk/nodejs/x/automation/localWorkspace.ts
@@ -494,6 +494,8 @@ export class LocalWorkspace implements Workspace {
     /**
      * exportStack exports the deployment state of the stack.
      * This can be combined with Workspace.importStack to edit a stack's state (such as recovery from failed deployments).
+     *
+     * @param stackName the name of the stack.
      */
     async exportStack(stackName: string): Promise<Deployment> {
         await this.selectStack(stackName);
@@ -503,6 +505,9 @@ export class LocalWorkspace implements Workspace {
     /**
      * importStack imports the specified deployment state into a pre-existing stack.
      * This can be combined with Workspace.exportStack to edit a stack's state (such as recovery from failed deployments).
+     *
+     * @param stackName the name of the stack.
+     * @param state the stack state to import.
      */
     async importStack(stackName: string, state: Deployment): Promise<void> {
         await this.selectStack(stackName);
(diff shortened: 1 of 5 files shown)


sdk/nodejs/x/automation/workspace.ts
```diff
@@ -202,6 +211,14 @@ export interface StackSummary {
     url?: string;
 }
 
+/**
+ * Deployment encapsulates the state of a stack deployment.
+ */
+export interface Deployment {
+    version: number;
+    deployment: object;
```
komalali, line 219: Is there a better type for this? In [Go](https://github.com/pulumi/pulumi/blob/cedab0003bda3b4cbb767d1a8875a620c34ba9da/sdk/go/common/apitype/core.go#L159-L167) we've left this unmarshalled but it's not possible to partially parse a json object in Typescript as far as I'm aware.

EvanBoyle: The type of deployment ends up being one of `DeploymentV1/2/3`: https://github.com/pulumi/pulumi/blob/48f906e0263335ad5fc922f6836b75c5f27da5eb/sdk/go/common/apitype/core.go#L108

That API in go is designed for backwards/forwards compat. SerDe and typing in typescript are far more flexible than in Go, so we may be able to improve the experience here with union types. If you're going to stick with the untyped route that requires the user to do type assertions, `any` may be more idiomatic.

EvanBoyle (approved):

sdk/nodejs/x/automation/workspace.ts
komalali: Ah perfect, thanks for the pointers. I'll fix this up with a union of the deployment versions.

sdk/nodejs/x/automation/localWorkspace.ts
```diff
@@ -493,6 +493,28 @@ export class LocalWorkspace implements Workspace {
             return value;
         });
     }
+    /**
+     * exportStack exports the deployment state of the stack.
+     * This can be combined with Workspace.importStack to edit a stack's state (such as recovery from failed deployments).
```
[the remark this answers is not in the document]
komalali: Can you clarify what you mean by `interfaces` when you refer to other areas where this needs to be fixed?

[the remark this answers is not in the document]
EvanBoyle: I was referring to methods in the `workspace` interface. Documenting params of the methods that were added similar to how the existing methods are documented: https://github.com/pulumi/pulumi/pull/6038/files/d434da0774e7ced3a6ca69907a84779ab7e5f248#diff-5d90703c04c98bf207aef5a82697fa29f4655353f9f02a9ed83902abce9c6b2aL180-L185

sdk/nodejs/x/automation/workspace.ts
komalali: Okay I'm realizing this is a bottomless hole so I'm gonna stick with `any` and leave a note to make it better in the future.

Issue #6039, opened 2021-01-02
Testing custom resources throws an unhandled exception after success


roderik:
I’m trying to add tests to my pulumi code, in this case, a CustomResource I’ve created. I have to use jest so the code is a bit different than the docs. My tests work and I can verify what I want to verify, but at the end, it tries to log an unhandled rejection: unhandled rejection: CONTEXT(221): monitor.registerResourceOutputs(...)-initial 

Full test code and error here: https://gist.github.com/roderik/1ff8d4df9c76f13f7441f7e5c34cc7dd

## Expected Behavior

When I run the test, the test can verify that a gcloud kubeconfig is being created, and no after the fact unhandled exceptions appear.

## Current Behavior

The test itself works, so the expect verifies what it needs to (checked this with some advanced console logging 🤣)

But after the test is done, this gets logged.

```
  ●  Cannot log after tests are done. Did you forget to wait for something async in your test?
    Attempted to log "unhandled rejection: CONTEXT(221): monitor.registerResourceOutputs(...)-initial
    STACK_TRACE:
    Error
        at Object.debuggablePromise (/Users/roderik/Development/bpaas-launchpad/packages/launchpad-api/node_modules/@pulumi/pulumi/runtime/debuggable.js:69:75)
        at runAsyncResourceOp (/Users/roderik/Development/bpaas-launchpad/packages/launchpad-api/node_modules/@pulumi/pulumi/runtime/resource.js:634:75)
        at Object.registerResourceOutputs (/Users/roderik/Development/bpaas-launchpad/packages/launchpad-api/node_modules/@pulumi/pulumi/runtime/resource.js:553:5)
        at ClusterResource.registerOutputs (/Users/roderik/Development/bpaas-launchpad/packages/launchpad-api/node_modules/@pulumi/pulumi/resource.js:444:20)
        at ClusterResource.<anonymous> (/Users/roderik/Development/bpaas-launchpad/packages/launchpad-api/node_modules/@pulumi/pulumi/resource.js:410:18)
        at Generator.next (<anonymous>)
        at fulfilled (/Users/roderik/Development/bpaas-launchpad/packages/launchpad-api/node_modules/@pulumi/pulumi/resource.js:18:58)
        at processTicksAndRejections (node:internal/process/task_queues:93:5)".

      at Object.debuggablePromise (node_modules/@pulumi/pulumi/runtime/debuggable.js:69:75)
[...]

PR #6040, opened 2021-01-02
fix: typo to get the outputs while registering them in a mock

roderik:
Resolves: #6039

Sidenote, to actually find the issue, I had to console log deep inside resource.js here https://github.com/pulumi/pulumi/blob/master/sdk/nodejs/runtime/resource.ts#L702

```
      RegisterResourceOutputs RPC finished: urn=urn:pulumi:stack::project::bpaas:cluster::test1; err: TypeError: req.getOuptuts is not a function, resp: undefined
```

Now I'm totally not deep enough in the way it all works under the hood, but if 'err' has an error, wouldn't reject() be better than resolve()?


pgavlin:
> Now I'm totally not deep enough in the way it all works under the hood, but if 'err' has an error, wouldn't reject() be better than resolve()?

Looks like there are two instances of that call to log. The other immediately precedes the code that calls `reject`:
https://github.com/pulumi/pulumi/blob/master/sdk/nodejs/runtime/resource.ts#L689-L699

pgavlin (approved):

Issue #2390, opened 2019-01-24
Support `--json` on `pulumi {update|destroy|preview}`


ellismg:
During an update operation, the engine emits a bunch of "engine events" about how the update is progressing.  Our display layers consume these events and print them in different ways (if the standard "progress" view, a more bare-bones view when stdout is a dumb terminal or we are in CI and yet another way if `--diff` is passed).

It would be nice to expose these events in a structured JSON form, for folks that were invoking `pulumi` and wanted to present its output in a different form.

Note that the set of events the engine emits was not so much "designed" as "evolved" and so it's possible that as part of exposing this, we may want to change the shape of how these events, so they make sense for folks who don't have a deep understanding of the engine.

joeduffy:
The preview `--json` part is done, but keeping open for any cleanup, and for the rest of the JSON-ification of updates, etc.

tmehlinger:
Could we also get this on `pulumi stack output`?

pgavlin:
The greatest complication here may be https://github.com/pulumi/pulumi/issues/5747.

Issue #2318, opened 2018-12-21
Enable constraining an update or refresh to a preview result


lukehoban:
Today we run a `preview` operation prior to an `update` by default.  However, there is not firm guarantee that the update will do only what was presented in the `preview`.  The `preview` output itself is only guaranteed to be a conservative estimate - it may include changes that *might* happen (depending on how updates change other resources), but also may not need to happen.

It would be nice to offer an option to be stricter here, and to constrain an `update` to only be allowed to make changes that were reported during the `preview`.  The `update` would fail if any operation the update needs to preform is not a subset of what the preview reported.

There are several pieces to this:
1. How to serialize out preview results in a way that could be used as a constraint (and potentially as an input to analyzers and other tools that want to be able to report feedback on a preview).
2. Whether there is a sufficiently robust interpretation for "subset of" between an update operation and corresponding element(s) in a serialized preview output.
3. Whether this can be applied to `update`, `refresh`, `import` in a consistent way

casey-robertson:
We have what we think is a bug related to this.  Getting our info together and will open a new issue and/or reach out via Slack about it.

clstokes:
Rather than constraining updates to operations that were present in the preview, the subsequent `update`, `refresh`, `import`, etc. should run what was shown in the preview or nothing at all. A partially executed set of operations will be a poor user experience and lead to more issues trying to reconcile them.

Another behavior related to this is Pulumi picks up code changes on the filesystem in between a preview completing and selecting `yes` at the `Do you want to perform this update?` prompt. This isn't a common scenario but should still be protected against.

lukehoban:
> A partially executed set of operations will be a poor user experience and lead to more issues trying to reconcile them.

It's in general not possible to guarantee this - things can fail for a multitude of reasons during the update - cloud provider outages, invalid user requirements that are only checked by the provider, changes that have been applied to the cloud environment between the preview and the update, etc.

So in general I believe the best that can be done is to say that we will not try to do anything which wasn't already approved by the user after seeing the preview. 

[...]

PR #6045, opened 2021-01-04
[CLI] Support passing --json to destroy, refresh and update commands

stack72:
Fixes: #2390

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ CHANGELOG
 - [CLI] Allow `pulumi console` to accept a stack name
   [#6031](https://github.com/pulumi/pulumi/pull/6031)
 
+- [CLI] Support JSON outputs (`--json`) for `update`, 
+  `destroy` and `refresh` operations
+  [#6045](https://github.com/pulumi/pulumi/pull/6045)
+
 ## 2.16.2 (2020-12-23)
 
 - Fix a bug in the core engine that could cause previews to fail if a resource with changes had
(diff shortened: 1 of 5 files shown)


stack72:
After talking with @pgavlin and @lukehoban we are going to enhance this a little more. For `up`, `destroy` and `refresh` we are going to make `--json` stream to stdout by default and then have the ability to take and `--event-log` path for writing to file

techmouse84:
I have been following this for some time. 
Would this item be released? 
If not , what would be a workaround ? 
Currently I'm using pulumi preview --json but that's not a realistic representation especially when there's failure during pulumi up. 
My use case would be to be able to do some post processing and send an email on the affected resources.  However looking at pulumi output can be cryptic to some users. post processing free form text is not ideal as well.

lukehoban:
> Currently I'm using pulumi preview --json but that's not a realistic representation especially when there's failure during pulumi up. My use case would be to be able to do some post processing and send an email on the affected resources. However looking at pulumi output can be cryptic to some users. post processing free form text is not ideal as well.

We are working on support for saving the results of a `preview`, and using those to constrain a future update (or to present preview results in other contexts) as part of https://github.com/pulumi/pulumi/issues/2318.  I expect that is likely the core feature that will most directly help with your scenario?

techmouse84:
Definitely it would restrict the upper limit of the affected resources. 
However in the event of failures , there might be less resources being spin up successfully. 
Hence using the result from preview to report to other stack holders would be incorrect. Hence it would still be ideal if I can have an exact json or some form of structured output during pulumi up.

Issue #4328, opened 2020-04-08
Add update ID to stack history output


clstokes:
#### Problem

`pulumi stack history` and `pulumi stack history --json` do not provide the update ID so it's difficult to correlate to an update in the console and difficult to automate against.

Pulumi version - `v1.14.0`

#### Suggestion

Add `id` as a distinct field in the normal and json output.

##### Normal 

```
% pulumi history
UpdateKind: destroy
Status: succeeded
Message: Update Azure policy demo
+0-9~0 0 Updated 1 month ago took 6m11s
    <snip>

UpdateKind: update
Status: succeeded
Message: ~encryption-providers: hard code KMS alias name
+9-0~0 0 Updated 1 month ago took 3m42s
    <snip>
```

##### JSON

```
% pulumi history --json
[
  {
    "kind": "destroy",
    "startTime": "2020-02-14T19:54:57.000Z",
    "message": "Update Azure policy demo",
    "environment": {
        // snip
    },
    "config": {
        // snip
    },
    "result": "succeeded",
    "endTime": "2020-02-14T20:01:08.000Z",
    "resourceChanges": {
      "delete": 9
    }
  },
  {
    "kind": "update",
    "startTime": "2020-02-13T05:04:28.000Z",
    "message": "~encryption-providers: hard code KMS alias name",
    "environment": {
        // snip
    },
    "config": {
        // snip
    },
    "result": "succeeded",
    "endTime": "2020-02-13T05:08:10.000Z",
    "resourceChanges": {
      "create": 9
    }
  },
  ...
```

PR #6049, opened 2021-01-04
Export UpdateID as part of the UpdateInfo event that comes from the SaaS

stack72:
Part of #4328

### DIFF
--- a/sdk/go/common/apitype/history.go
+++ b/sdk/go/common/apitype/history.go
@@ -84,6 +84,7 @@ const (
 // Should generally mirror backend.UpdateInfo, but we clone it in this package to add
 // flexibility in case there is a breaking change in the backend-type.
 type UpdateInfo struct {
+	UpdateVersion int `json:"updateVersion"`
 	// Information known before an update is started.
 	Kind        UpdateKind             `json:"kind"`
 	StartTime   int64                  `json:"startTime"`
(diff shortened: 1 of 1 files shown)


sdk/go/common/apitype/history.go
```diff
@@ -84,6 +84,7 @@ const (
 // Should generally mirror backend.UpdateInfo, but we clone it in this package to add
 // flexibility in case there is a breaking change in the backend-type.
 type UpdateInfo struct {
+	UpdateVersion int `json:"updateVersion"`
```
chrsmith, line 87: I'd suggest we just keep this as version to be consistent with the other places in the code we refer to this value. 

```suggestion
	Version int `json:"version"`
```

chrsmith (approved):
Aside from suggesting we rename the field to just `Version`, this looks fine. Though it might be useful to get a better sense for the scenario(s) you are trying to enable here.

For example, each update operation (`update`, `destroy`, `import`, `refresh`, etc.) has a unique `Version` integer. But if you want to get a list of _previews_ that have been ran against the current stack, we would need to include the `UpdateID string` instead. (Since all of them would have the same `Version` value -- they are previews of the update to version XX.

Anyways, once this is in we just need to start setting the new field from the Pulumi API service. I believe it's just the `.../updates` endpoint where we return the stack's update history.

Issue #6032, opened 2020-12-31
Importing from file shows error if resources have same name but different types.


rchamorro:
When importing resources using a file with the command `pulumi import -f resources.json` and C# as target language.

It works importing the same resources without the file.

Example imported file:

```json
{
  "resources": [      
    {
      "type": "azure-nextgen:resources/v20200601:ResourceGroup",
      "name": "resource-name",
      "id": "/subscriptions/subscription-id/resourceGroups/resource-group-name"
    },    
    {
      "type": "azure-nextgen:web/v20200601:AppServicePlan",
      "name": "resource-name",
      "id": "/subscriptions/subscription-id/resourceGroups/resource-group-name/providers/Microsoft.Web/serverfarms/app-service-plan-name"
    }
  ]
}
```

## Expected Behavior

Do not show error importing resources from file when they have same names but different types, and show the code that should be included in the Pulumi Stack.

## Current Behavior

Pulumi imports the resource but shows and error and does not show the code required to handle the resource.

This is the output:

```
     Type                                                Name           Status
 +   pulumi:pulumi:Stack                                 stack-name     created
 =   ├─ azure-nextgen:web/v20200601:AppServicePlan       resource-name  imported
 =   └─ azure-nextgen:resources/v20200601:ResourceGroup  resource-name  imported

Resources:
    + 1 created
    = 2 imported
    3 changes

Duration: 7s

error: internal error: Error: "internal-apis" already declared

  on anonymous.pp line 0:
   1: resource internal-apis "azure-nextgen:resources/v20200601:ResourceGroup" {
   2:     location = "westeurope"
   3:     resourceGroupName = "resource-name"
   4: options {
   5: protect =true
   7: }
   9: }

"resource-name" already declared
```

## Steps to Reproduce

1. Create a file with two resources two import with different types and same name 
2. pulumi import -f resources.json

## Context (Environment)

- Windows 10
- C# Target language
- No other resources in the stack

HagaiBarel:
Hitting this as well using `go` as the target language, `pulumi v2.16.2`

mikhailshilkov:
Is this not "by design"? Name is translated to variable names in the target programming language - is it not a reasonable requirement to keep them unique? Does the requirement to have them unique block any scenario?

EvanBoyle:
This is by design. There cannot be two resources within the same lexical scope in a program.

clstokes:
I believe this should be re-opened. 

[...]

PR #6056, opened 2021-01-05
[codegen/pcl] Add a urnName resource option.

pgavlin:
This resource option controls the "name" portion of the resource's URN.
It is intended for use by the import code generator, which may need to
generate code for multiple resources that share the "name" portion of
the URN, but must still generate unique names for the PCL resource
declarations.

This is the backend half of the fix for #6032. The frontend half of the
fix will ensure that names for PCL resource declarations are unique.


mikhailshilkov:
I asked https://github.com/pulumi/pulumi/issues/6032#issuecomment-756079919 on the issue - trying to understand why we want to fix this

mikhailshilkov:
Closing this since we discussed it's superseded by newer work

PR #6069, opened 2021-01-08
Allow optional strict enum types to be omitted

viveklak:
Fixes https://github.com/pulumi/pulumi/issues/6068

### DIFF
--- a/pkg/codegen/go/gen.go
+++ b/pkg/codegen/go/gen.go
@@ -276,8 +276,12 @@ func (pkg *pkgContext) inputType(t schema.Type, optional bool) string {
 	var typ string
 	switch t := t.(type) {
 	case *schema.EnumType:
+		var enumType string
+		if optional {
+			enumType = "*"
+		}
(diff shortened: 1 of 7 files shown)


pkg/codegen/go/gen.go
```diff
@@ -276,8 +276,12 @@ func (pkg *pkgContext) inputType(t schema.Type, optional bool) string {
 	var typ string
 	switch t := t.(type) {
 	case *schema.EnumType:
+		var enumType string
```
mikhailshilkov, line 279: Nit: `enumType` doesn't feel like an obvious name. Is it more of a `modifier` thing, or `prefix`, or `nullable`?

mikhailshilkov (approved):
LGTM!

komalali (approved):

pkg/codegen/go/gen.go
viveklak: Fixed.

PR #6070, opened 2021-01-08
Fixes to support non-pulumi backends in kubernetes operator

viveklak:
Couple of quick but necessary fixes for Kubernetes operator to support backends other than app.pulumi.com.

### DIFF
--- a/sdk/go/common/workspace/creds.go
+++ b/sdk/go/common/workspace/creds.go
@@ -131,7 +131,8 @@ func getCredsFilePath() (string, error) {
 }
 
 // GetCurrentCloudURL returns the URL of the cloud we are currently connected to. This may be empty if we
-// have not logged in.
+// have not logged in. Note if PULUMI_BACKEND_URL is set, the corresponding value is returned
+// instead irrespective of the backend for current project or stored credentials.
 func GetCurrentCloudURL() (string, error) {
 	// Allow PULUMI_BACKEND_URL to override the current cloud URL selection
 	if backend := os.Getenv(PulumiBackendURLEnvVar); backend != "" {
--- a/sdk/go/common/workspace/creds.go
+++ b/sdk/go/common/workspace/creds.go
@@ -133,13 +133,12 @@ func getCredsFilePath() (string, error) {
 // GetCurrentCloudURL returns the URL of the cloud we are currently connected to. This may be empty if we
 // have not logged in.
 func GetCurrentCloudURL() (string, error) {
-	var url string
-
 	// Allow PULUMI_BACKEND_URL to override the current cloud URL selection
 	if backend := os.Getenv(PulumiBackendURLEnvVar); backend != "" {
-		url = backend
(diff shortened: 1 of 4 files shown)


sdk/go/common/workspace/creds.go
```diff
@@ -133,13 +133,12 @@ func getCredsFilePath() (string, error) {
 // GetCurrentCloudURL returns the URL of the cloud we are currently connected to. This may be empty if we
 // have not logged in.
 func GetCurrentCloudURL() (string, error) {
-	var url string
-
 	// Allow PULUMI_BACKEND_URL to override the current cloud URL selection
 	if backend := os.Getenv(PulumiBackendURLEnvVar); backend != "" {
-		url = backend
+		return backend, nil
```
viveklak, line 138: This was introduced in https://github.com/pulumi/pulumi/pull/5789 however, this change was needed to use a local filesystem backend for the kubernetes operator.

sdk/go/x/auto/stack.go
```diff
@@ -644,8 +644,8 @@ func (ur *UpResult) GetPermalink() (string, error) {
 // GetPermalink returns the permalink URL in the Pulumi Console for the update
 // or refresh operation. This will error for alternate, local backends.
 func GetPermalink(stdout string) (string, error) {
-	const permalinkSearchStr = "View Live: "
-	var startRegex = regexp.MustCompile(permalinkSearchStr)
+	const permalinkSearchStr = "View Live: |Permalink: "
```
viveklak, line 647: `View Live:` is only relevant to app.pulumi.com backend. Other backends fail to get the permalink.

sdk/go/common/workspace/creds.go
lukehoban: I thought we had tests covering this. If we did, was there a case missing?  If we didn't, worth adding them as part of this PR?

sdk/go/x/auto/stack.go
```diff
@@ -644,8 +644,8 @@ func (ur *UpResult) GetPermalink() (string, error) {
...
+	startRegex := regexp.MustCompile(permalinkSearchStr)
 	var endRegex = regexp.MustCompile("\n")
```
lukehoban, line 649: Should style also be fixed here?

lukehoban (approved):

sdk/go/common/workspace/creds.go
viveklak: > I thought we had tests covering this. If we did, was there a case missing? If we didn't, worth adding them as part of this PR?

I didn't see coverage for this particular code path. Will add now.

viveklak: Added test case.

sdk/go/x/auto/stack.go
viveklak: Were you referring to use of `var`s? Updated. Let me know if you meant something else.

PR #6075, opened 2021-01-09
Fix #5577

### LINKED ISSUE #5577: Assets and Archives fail silently when they point to a non-existent path
When using FileArchive, it appears that passing in a non-existent path will cause it to fail silently and the related resource won't be created.

## Expected Behavior
Error message indicating an invalid path.

## Current Behavior
Silent failure and the relevant resource is not created.

## Steps to Reproduce
```
const fn = new aws.lambda.Function("webapp-api-callback", {
        handler: "handler",
        role: role.arn,
        runtime: aws.lambda.NodeJS12dXRuntime,

        code: new pulumi.asset.AssetArchive({
            ".": new pulumi.asset.FileArchive("./lambda"),
        }),
    });
```

where in the above example, `./lambda` does not exist.

There are no errors and no attempt to create the function (preview doesn't show it).

joeduffy:
This change fixes the problem where a missing asset or archive
will silently fail. This is because the engine returns an RPC error
which is ignored by (some) language SDKs. 

I ended up starting by fixing this in the engine, but then reverted to
doing so in the client SDK. It really ought to be the responsibility of
the SDK to deal with a failed RPC correctly, and the server shouldn't
be prone to failing just due to bad client requests.

I've verified that this addresses the bug, and included a regression test.

Instead of a silent resource going missing, I now see an error:

     ├─ aws:lambda:Function    mylambdaDoesNotExist              1 error

    Diagnostics:
      aws:lambda:Function (mylambdaDoesNotExist):
        error: problem registering resource: failed to compute archive
            hash: couldn't read archive path './DoesNotExist':
                stat ./DoesNotExist: no such file or directory

Assuming folks are on board with this, I believe we'll also need to
fix other RPC methods (e.g., read resource) as well as other SDKs
(e.g., it appears the Python SDK also exhibits the buggy behavior).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ CHANGELOG
   correlate events back to the Pulumi SaaS
   [#6063](https://github.com/pulumi/pulumi/pull/6063)
 
+- Fix silent failures when assets are missing.
+  [#6075](https://github.com/pulumi/pulumi/pull/6075)
+
 ## 2.17.0 (2021-01-06)
 
 - Respect the `version` resource option for provider resources.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ CHANGELOG
 - Fix silent failures when assets are missing.
   [#6075](https://github.com/pulumi/pulumi/pull/6075)
 
+- [sdk/dotnet] Moved urn value retrieval into if statement for MockMonitor
+  [#6081](https://github.com/pulumi/pulumi/pull/6081)
+
 ## 2.17.0 (2021-01-06)
 
 - Respect the `version` resource option for provider resources.
--- a/pkg/engine/update.go
+++ b/pkg/engine/update.go
@@ -403,7 +403,7 @@ func newUpdateSource(
 
 	// If that succeeded, create a new source that will perform interpretation of the compiled program.
 	// TODO[pulumi/pulumi#88]: we are passing `nil` as the arguments map; we need to allow a way to pass these.
-	return deploy.NewEvalSource(plugctx, opts.Diag, &deploy.EvalRunInfo{
+	return deploy.NewEvalSource(plugctx, &deploy.EvalRunInfo{
 		Proj:    proj,
 		Pwd:     pwd,
 		Program: main,
--- a/pkg/engine/update.go
+++ b/pkg/engine/update.go
@@ -403,7 +403,7 @@ func newUpdateSource(
 
 	// If that succeeded, create a new source that will perform interpretation of the compiled program.
 	// TODO[pulumi/pulumi#88]: we are passing `nil` as the arguments map; we need to allow a way to pass these.
-	return deploy.NewEvalSource(plugctx, &deploy.EvalRunInfo{
+	return deploy.NewEvalSource(plugctx, opts.Diag, &deploy.EvalRunInfo{
 		Proj:    proj,
 		Pwd:     pwd,
 		Program: main,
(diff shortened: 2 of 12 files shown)


stack72 (approved):

joeduffy:
@stack72 @EvanBoyle @pgavlin One thing that surprised me: If errors get issued, the CLI still returns an exit code of `0`. I am 29% certain it used to return a non-`0` exit code for this .... but am probably mistaken.

Question is: What do we want to happen here? (Might need to be a 3.0 breaking change if we decide to change?)

EvanBoyle:
> If errors get issued, the CLI still returns an exit code of 0
Zero for this particular error, or for all errors? From what I've seen the CLI returns 255 on error fairly consistently.

stack72:
> This change fixes the problem where a missing asset or archive
> 
> will silently fail. This is because the engine returns an RPC error
> 
> which is ignored by (some) language SDKs. 
> 
> 
> 
> I ended up starting by fixing this in the engine, but then reverted to
> 
> doing so in the client SDK. It really ought to be the responsibility of
> 
> the SDK to deal with a failed RPC correctly, and the server shouldn't
> 
> be prone to failing just due to bad client requests.
> 
> 
> 
> I've verified that this addresses the bug, and included a regression test.
> 
> 
> 
> Instead of a silent resource going missing, I now see an error:
> 
> 
> 
>      ├─ aws:lambda:Function    mylambdaDoesNotExist              1 error
> 
> 
> 
>     Diagnostics:
> 
>       aws:lambda:Function (mylambdaDoesNotExist):
> 
>         error: problem registering resource: failed to compute archive
> 
>             hash: couldn't read archive path './DoesNotExist':
> 
>                 stat ./DoesNotExist: no such file or directory
> 
> 
> 
> Assuming folks are on board with this, I believe we'll also need to
> 
> fix other RPC methods (e.g., read resource) as well as other SDKs
> 
> (e.g., it appears the Python SDK also exhibits the buggy behavior).

Personally I don't believe this is a breaking change but is a bug fix

It's an error that users don't get an error on so I'd ship it as a normal change

$0.02

joeduffy:
> Zero for this particular error, or for all errors? From what I've seen the CLI returns 255 on error fairly consistently.

There are certainly many cases that issue an error. I was surprised to find that we don't return non-0 if there are _any_ error messages printed, however. For example:

```
pulumi.log.error("boom");
```

This yields the following output:

```
$ pulumi preview
Previewing update (dev)

     Type                 Name      Plan       Info
 +   pulumi:pulumi:Stack  errs-dev  create     1 error

Diagnostics:
  pulumi:pulumi:Stack (errs-dev):
    error: boom
```

And yet, a zero exit code:

```
$ echo $?
0
```

That's surprising. There are error-level diagnostics, which I assumed would yield a non-zero exit code. Given this, I'm actually not quite sure the way the SDK client is supposed to incite a failure other than letting an unhandled exception fly.

lukehoban:
>  I was surprised to find that we don't return non-0 if there are any error messages printed, however.

Note that it's not just about returning 0 vs. non-0 - the program above doesn't fail at all - it will continue to run to completion.  All it does is print an "error" diagnostic.  I've opened https://github.com/pulumi/pulumi/issues/6090 to track this.  I believe it's always been this way.

> Given this, I'm actually not quite sure the way the SDK client is supposed to incite a failure other than letting an unhandled exception fly.

Yes - the only way today to fail the deployment from a language host is to cause the host language process to die - either through an unhandled error or a process exit with a failing error code.  Unhandled exceptions are I believe preferable, as they allow the most information to be bubbled up to users.

PR #6081, opened 2021-01-11
[sdk/dotnet] Moved urn value retrieval into if statement

### LINKED ISSUE #6080: Dotnet pulumi unit tests crashes with "the key 'urn' is not present in the dictionary"
Pulumi Dotnet unit tests crashes with "The given key 'urn' was not present in the dictionary." after updating to version pulumi 2.15.5 and later.

It looks like the crash was introduced with this code line: https://github.com/pulumi/pulumi/blob/master/sdk/dotnet/Pulumi/Testing/MockMonitor.cs#L36

This line will also result in the urn is not present in the dictionary https://github.com/pulumi/pulumi/blob/master/sdk/dotnet/Pulumi/Deployment/Deployment_Invoke.cs#L39 

A solution to this problem would be to move the urn retrieval into the if check, since it looks like the urn is only used when the token type is "pulumi:pulumi:getResource".

This error was discovered while using the pulumi azure library with the following code in a test scenario
```csharp
var clientConfig = Output.Create(GetClientConfig.InvokeAsync());
```
```csharp
await Deployment.TestAsync<Stack>(new Mock());
```

## Expected Behavior
No crashing

## Current Behavior
Crashing

## Steps to Reproduce
1. create a dotnet pulumi project for azure
2. create unit tests which creates a stack which uses "Output.Create(GetClientConfig.InvokeAsync())"
3. The test should crash

## Context (Environment)

We are trying to test our common infrastructure code, and pulumi broke almost all of our unit tests with this change. We can not upgrade to a newer version of pulumi before this is resolved.

nicro950:
Fixes #6080

For more details check #6080 
The problem is that if the urn is not present in the given args, then the unit tests will fail.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ CHANGELOG
   correlate events back to the Pulumi SaaS
   [#6063](https://github.com/pulumi/pulumi/pull/6063)
 
+- [sdk/dotnet] Moved urn value retrieval into if statement for MockMonitor
+  [#6081](https://github.com/pulumi/pulumi/pull/6081)
+
 ## 2.17.0 (2021-01-06)
 
 - Respect the `version` resource option for provider resources.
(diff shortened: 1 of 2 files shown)


justinvp (approved):
LGTM. Thanks, @nicro950!

PR #6084, opened 2021-01-11
Fix looking up empty version in ResourcePackages.TryGetResourceType

justinvp:
When a resource reference is deserialized, it may not have a version in which case `version` will be an empty string:

https://github.com/pulumi/pulumi/blob/9fbdc51fcd38cd6e805652928583eba4ce6a5b5b/sdk/dotnet/Pulumi/Serialization/Deserializer.cs#L227-L230

This change fixes `TryGetResourceType` to work correctly when an empty version is passed.

Fixes https://github.com/pulumi/pulumi/issues/6088


mikhailshilkov (approved):
LGTM

Issue #5653, opened 2020-10-30
Support for Understanding Output Secrets


stack72:
This is an issue to bring together the design required to implement #3795 and #2886 and is a requirement for https://github.com/pulumi/pulumi-terraform-bridge/issues/10 to proceed

This issue will house the "design" we choose for the API

Effectively, we need the 2 following-type functions:

pulumi.IsSecretOutput
pulumi.Secret

We need to choose if these are instance methods or static methods.

stack72:
So last week I spoke with @justinvp @mikhailshilkov and @EvanBoyle about this work in all of the respective language SDKs and @justinvp came up with the following scenarios which we all agree would work

1. Static functions
2. Instance functions

These would manifest themselves as following:

### Unsecret

```csharp
var exposedOutput = someOutput.UnsafeUnsecret();
var exposedOutput = Output.Unsecret(someOutput);
```

```typescript
const exposedOutput = pulumi.Unsecret(someOutput);
const exposedOutput = someOutput.Unsecret();
```

```python
exposed_output = Output.unsecret(some_output)
exposed_output = some_output.unsecret()
```

In Go, @EvanBoyle and I discussed that it would be easier to have a Static function that is the opposite of ToSecret and ToSecretwithContext and is more in keeping with the Go SDK

These would be:

```go
exposedOutput := Unsecret(someOutput)
exposedOutput := UnsecretWithContext(ctx, someOutput)
```

With regards to naming, @justinvp suggested we should really point out that Unsecret is actually an operation we don't suggest is good practice by naming the func: `UnsafeUnsecret`

mikhailshilkov:
Do we want to pick one style (static vs instance)?

If we want to hide it from casual use, in C# we could put it to a separate namespace as we did here: https://github.com/pulumi/pulumi/blob/master/sdk/dotnet/Pulumi/Core/OutputUtilities.cs

What about the proposal on `IsSecret`?

justinvp:
> TypeScript
> 
> ```ts
> const exposedOutput = pulumi.Unsecret(someOutput);
> const exposedOutput = someOutput.Unsecret();
> ```

Note: For TypeScript the function name should be lower-case.

> With regards to naming, @justinvp suggested we should really point out that Unsecret is actually an operation we don't suggest is good practice by naming the func: `UnsafeUnsecret`

[...]

PR #6086, opened 2021-01-11
Add the ability to pulumi.unsecret an existing output

stack72:
Related: #5653

This will take an existing output and then unwrap the secret, and
return a new output

```

const x = pulumi.secret("test")
export const xVal = x;

const y = pulumi.unsecret(x);
export const yVal = y;
```

```
▶ pulumi stack output
Current stack outputs (3):
    OUTPUT         VALUE
    xVal           [secret]
    yVal           test
```

Also adds the ability to check if an output is as secret:

```

const x = pulumi.secret("test")
const isSecret = pulumi.isSecret(x);

export const isSecretDeets = isSecret;
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ CHANGELOG
 - [automation/go] Set DryRun on previews so unknowns are identified correctly.
   [#6099](https://github.com/pulumi/pulumi/pull/6099)
 
+- [sdk/nodejs] Added `pulumi.unsecret` which will take an existing secret output and
+  create a non-secret variant with an unwrapped secret value. Also adds,
+  `pulumi.isSecret` which will take an existing output and
+  determine if an output has a secret within the output.
+  [#6086](https://github.com/pulumi/pulumi/pull/6086)
+
 ## 2.17.1 (2021-01-13)
 
 - Fix an issue with go sdk generation where optional strict enum values
@@ -30,7 +36,7 @@ CHANGELOG
 - [sdk/dotnet] Moved urn value retrieval into if statement
   for MockMonitor
   [#6081](https://github.com/pulumi/pulumi/pull/6081)
-
+  
 - [sdk/dotnet] Added `Pulumi.Output.Unsecret` which will
   take an existing secret output and
   create a non-secret variant with an unwrapped secret value.
--- a/sdk/nodejs/output.ts
+++ b/sdk/nodejs/output.ts
@@ -388,9 +388,10 @@ async function applyHelperAsync<T, U>(
 // Returns an promise denoting if the output is a secret or not. This is not the same as just calling `.isSecret`
 // because in cases where the output does not have a `isSecret` property and it is a Proxy, we need to ignore
 // the isSecret member that the proxy reports back.
+// This calls the public implementation so that we only make any calculations in a single place.
 /** @internal */
 export function isSecretOutput<T>(o: Output<T>): Promise<boolean> {
-    return Output.isInstance(o.isSecret) ? Promise.resolve(false) : o.isSecret;
+    return isSecret(o);
 }
 
 // Helper function for `output`.  This function trivially recurses through an object, copying it,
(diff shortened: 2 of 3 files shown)


stack72:
After speaking with @lukehoban we are going to revert the changes to IsSecret and have a top level func 'issecret' alongside 'unsecret' - this will ensure we don't break the backwards compatibility with old versions

sdk/nodejs/output.ts
```diff
@@ -38,7 +38,7 @@ class OutputImpl<T> implements OutputInstance<T> {
      * encrypted format when they are persisted as part of a state file. When`true` this "taints" any
      * additional resources created from it via an [all] or [apply], such that they are also treated as
      * secrets.
-     * @internal
+     * * @internal
```
justinvp:
```suggestion
     * @internal
```

komalali (approved):
Couple of comment/changelog things to fix up but otherwise LGTM

sdk/nodejs/output.ts
```diff
@@ -534,6 +534,20 @@ export function secret<T>(val: Input<T | undefined>): Output<Unwrap<T | undefine
         o.isKnown, Promise.resolve(true), o.allResources!());
 }
 
+/**
+ * [unsecret] behaves the same as [output] except the returned output takes the existing output and unwraps the secret
+ */
+export function unsecret<T>(val: Output<T>): Output<T> {
```
justinvp, line 540: I could go either way, but I am wondering if we should consider making this accept `Input<T>` like the existing `output` and `secret` functions (in the implementation, if `val` isn't an `Output`, then it'd be a no-op). @lukehoban, do you have an opinion?

sdk/nodejs/output.ts
```diff
@@ -534,6 +534,20 @@ export function secret<T>(val: Input<T | undefined>): Output<Unwrap<T | undefine
...
+   return new Output(
```
justinvp, line 541: I think we want to first do `const o = output(val);` to ensure we're operating on an instance of `Output`, which also makes it safe to call `.allResources` per the comment in the implementation of `secret`:

https://github.com/pulumi/pulumi/blob/bd1554d33e45f73c8543bc79252f50bb83a03a1d/sdk/nodejs/output.ts#L526-L535

sdk/nodejs/output.ts
```diff
@@ -534,6 +534,20 @@ export function secret<T>(val: Input<T | undefined>): Output<Unwrap<T | undefine
...
+        val.resources(), val.promise(/*withUnknowns*/ true),
+        val.isKnown, Promise.resolve(false), val.allResources!());
+}
+
+export function isSecret<T>(val: Output<T>): Promise<boolean> {
```
justinvp, line 546: Similarly, should we consider making this accept `Input<T>` (and if it's not an `Output`, it'd return `false`)?

sdk/nodejs/output.ts
```diff
@@ -534,6 +534,20 @@ export function secret<T>(val: Input<T | undefined>): Output<Unwrap<T | undefine
...
+    return val.isSecret.then(x => x);
```
justinvp, line 547: We likely need to implement this like the existing internal `isSecretOutput` function, which properly handles Proxies:

https://github.com/pulumi/pulumi/blob/bd1554d33e45f73c8543bc79252f50bb83a03a1d/sdk/nodejs/output.ts#L388-L394

Then, can we get rid of the existing internal func in favor of the new public func?

sdk/nodejs/output.ts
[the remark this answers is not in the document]
stack72: it was not! I will fix this up now :) Good spot! thanks!!

CHANGELOG.md
```diff
@@ -18,6 +18,14 @@ CHANGELOG
...
   [#6081](https://github.com/pulumi/pulumi/pull/6081)
 
+- [sdk/nodejs] Added `pulumi.unsecret` which will take an existing secret output and
+  create a non-secret variant with an unwrapped secret value.
+  [#6086](https://github.com/pulumi/pulumi/pull/6086)
+
+- [sdk/nodejs] Added `pulumi.isSecret` which will take an existing output and
+  determine if an output has a secret within the output.
+  [#6086](https://github.com/pulumi/pulumi/pull/6086)
+
```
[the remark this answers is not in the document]
stack72: these are different entries relating to the same feature - I will fix this up :)

stack72:
So @justinvp and I had a call about this - we are going to leave these funcs as taking the Output itself for now - if we decide that needs to change in the future then we can do so.

PR #6092, opened 2021-01-12
[dotnet] Unsecret and IsSecret implementation for .NET

mikhailshilkov:
Related: #5653

This will take an existing output and then unwrap the secret, and return a new output

```
var x = Output.CreateSecret("test");
var y = Output.Unsecret(x); // new output that has "test" but is not a secret
```

Also adds the ability to check if an output is a secret:

```
var x = Output.CreateSecret("test");
var isSecret = await Output.IsSecretAsync(x); // true
```


stack72 (approved):
Thank you so much for doing this! LGTM!

Issue #2993, opened 2019-07-27
Support decrypting secrets in integration test framework


lukehoban:
In https://github.com/pulumi/pulumi-awsx/pull/377#issuecomment-515647437, we saw that some tests which read stack outputs that have recently become secrets are now failing, as the values are no longer strings but instead sentinel structs representing the secrets.

Unfortunately, it is not easy to decrypt these.  The biggest challenge is that the test framework does local `pulumi login` in each test's working directory, overriding credentials to be stored locally to that folder via an env var passed to each command.  

However, we cannot use that env var for the code running in the `ExtraRuntimeValidation`.  

We will need to provide a way to get back the decrypter for the stack (or even a pre-deserialized State) as part of the runtime info.

lukehoban:
> One thing we don't support today is something like `pulumi stack export --show-secrets`.

This is tracked in https://github.com/pulumi/pulumi/issues/2918.  And indeed I think that is the fundamental thing needed to enable this.

lukehoban:
> This is tracked in #2918. And indeed I think that is the fundamental thing needed to enable this.

#2918 is fixed now - so this should be fairly easy to add support for.

PR #6094, opened 2021-01-12
Add an option to be able to stack export showing secrets for use in ExtraRuntimeValidation

stack72:
Fixes: #2993

### DIFF
--- a/pkg/testing/integration/program.go
+++ b/pkg/testing/integration/program.go
@@ -204,6 +204,9 @@ type ProgramTestOptions struct {
 	RunBuild bool
 	// RunUpdateTest will ensure that updates to the package version can test for spurious diffs
 	RunUpdateTest bool
+	// DecryptSecretsInOutput will ensure that stack output is passed `--show-secrets` parameter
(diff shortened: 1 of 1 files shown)


pkg/testing/integration/program.go
```diff
@@ -1197,7 +1197,7 @@ func (pt *ProgramTester) TestPreviewUpdateAndEdits() error {
 }
 
 func (pt *ProgramTester) exportImport(dir string) error {
-	exportCmd := []string{"stack", "export", "--file", "stack.json"}
+	exportCmd := []string{"stack", "export", "--show-secrets", "--file", "stack.json"}
```
lukehoban, line 1200: Note that I think the important place we need to update is on line 1430 in performExtraRuntimeValidation.

I'm not positive we do have to update this location in `exportImport` - I think in this case we're directly roundtripping - so shouldn't matter whether we use `--show-secrets`.

stack72: We won't need to update this one - you are correct!

justinvp (approved):
LGTM

lukehoban (approved):

PR #6095, opened 2021-01-12
[Go] Add debug logging settings for common automation API operations

viveklak:


sdk/go/x/auto/example_test.go
```diff
@@ -36,7 +35,7 @@ import (
 	"github.com/pulumi/pulumi/sdk/v2/go/x/auto/optup"
 )
 
-func Example() error {
+func Example() {
```
viveklak, line 38: TIL that Godoc examples don't return errors.. My IDE was complaining about them but seems linter doesn't care? I can kill these changes if you prefer.

komalali: I think this makes sense, handling errors will really be up to the user anyway 👍

komalali (commented):
Is there any way we could test this?

viveklak:
> Is there any way we could test this?

I have tested it with an automation api example myself but can't think of a nice way to add as an automated test. An e2e example would be hard to assert that debug logging is actually occurring. Could try and intercept the args passed to [`runPulumiCmdSync`](https://github.com/pulumi/pulumi/blob/9616bde745fbeaa032108bad7816be5b9a317c95/sdk/go/x/auto/stack.go#L734) but that also seems awkward.  Open to ideas on this...

komalali (approved):
@viveklak yeah that's fair, I can't think of a succinct automated test off the top of my head either.

EvanBoyle:
Looks great! Sorry for being late on the review!

PR #6104, opened 2021-01-14
Add .NET resource ref unit tests.

pgavlin:
- Add tests that serialize custom and component resources for targets
  that support resource references
- Add tests that serialize custom and component resources for downlevel
  targets
- Add tests that deserialize known custom and component resources
- Add tests that deserialize missing custom and component resources

These changes also fix a few bugs that were encountered during testing:
- Component resource construction was not supported
- Resources with missing packages could not be deserialized

In the latter case, a missing resource is deserialized as a generic
DependencyResource.

These changes also update the signature of IMocks.NewResourceAsync to
allow the returned ID to be null. This is technically a C# breaking change
with respect to nullability.

Contributes to #5943.

### DIFF
--- a/sdk/dotnet/Pulumi.Tests/Serialization/ResourceRefPropertyTests.cs
+++ b/sdk/dotnet/Pulumi.Tests/Serialization/ResourceRefPropertyTests.cs
@@ -1,4 +1,4 @@
-// Copyright 2016-2020, Pulumi Corporation
+// Copyright 2016-2021, Pulumi Corporation
 
 using System;
 using System.Collections.Generic;
--- a/sdk/dotnet/Pulumi.Tests/Serialization/ResourceRefPropertyTests.cs
+++ b/sdk/dotnet/Pulumi.Tests/Serialization/ResourceRefPropertyTests.cs
@@ -1,0 +1,324 @@
+// Copyright 2016-2020, Pulumi Corporation
+
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using System.Threading.Tasks;
+using Google.Protobuf.WellKnownTypes;
+using Pulumi.Serialization;
+using Pulumi.Testing;
+using Xunit;
+
+namespace Pulumi.Tests.Serialization
+{
+    public class ResourceRefPropertyTests : ConverterTests
+    {
+        public sealed class MyArgs : Pulumi.ResourceArgs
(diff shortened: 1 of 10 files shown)


sdk/dotnet/Pulumi/Testing/MockMonitor.cs
```diff
@@ -65,7 +65,9 @@ public async Task<ReadResourceResponse> ReadResourceAsync(Resource resource, Rea
             {
                 var builder = ImmutableDictionary.CreateBuilder<string, object>();
                 builder.Add("urn", urn);
-                builder.Add("id", id);
+                if (id != null) {
```
mikhailshilkov, line 68: It would be nice if you could put `{` on a new line everywhere (or set your editor to do so).

```suggestion
                if (id != null)
                {
```

sdk/dotnet/Pulumi.Tests/Serialization/ResourceRefPropertyTests.cs
```diff
@@ -0,0 +1,324 @@
...
+        public class DeserializeMissingComponentResourceStack : Stack
+        {
+            [Output("values")]
+            public Output<ImmutableDictionary<string, string>> Values { get; private set; } = null!;
+
+            public DeserializeMissingComponentResourceStack()
+            {
+                var res = new MissingComponentResource("test", null, null);
+
+                var urn = res.Urn.DataTask.Result.Value;
```
mikhailshilkov, line 299: I guess it's a nit for a test but you could assign `Values` with `Output.Tuple` machinery and avoid using `.Result`'s.

sdk/dotnet/Pulumi.Tests/Serialization/ResourceRefPropertyTests.cs
```diff
@@ -0,0 +1,324 @@
+// Copyright 2016-2020, Pulumi Corporation
```
mikhailshilkov, line 1: Nit
```suggestion
// Copyright 2016-2021, Pulumi Corporation
```

mikhailshilkov (approved):
LGTM

sdk/dotnet/Pulumi/Testing/MockMonitor.cs
pgavlin: Ah, yeah--will do.

PR #6128, opened 2021-01-16
[codegen/go] @viveklak Add a rename mapping to package info to control collision renames

viveklak:
Fix for https://github.com/pulumi/pulumi-azure-nextgen/issues/137.

This is a targeted addition to allow an escape hatch to handle fallout from our poor name collision handling. The proper long term fix is tracked in https://github.com/pulumi/pulumi/issues/6170

### DIFF
--- a/pkg/codegen/go/gen.go
+++ b/pkg/codegen/go/gen.go
@@ -145,8 +145,8 @@ func (pkg *pkgContext) tokenToType(tok string) string {
 
 	if ok {
 		renamed := false
-		// First determine if there is an existing rename required for the pkg+type
-		if renames, ok := modPkg.typeRenameMappings[pkg.pkg.Name]; ok {
+		// First determine if there is an existing rename required for the module+type
+		if renames, ok := modPkg.typeRenameMappings[mod]; ok {
 			if newName, ok := renames[name]; ok {
 				name = newName
 				renamed = true
@@ -1669,7 +1669,7 @@ func generatePackageContextMap(tool string, pkg *schema.Package, goInfo GoPackag
 				modToPkg:           goInfo.ModuleToPackage,
 				pkgImportAliases:   goInfo.PackageImportAliases,
 				packages:           packages,
-				typeRenameMappings: goInfo.PackageToTypeRenameMapping,
+				typeRenameMappings: goInfo.ModuleToTypeRenameMapping,
 			}
 			packages[mod] = pack
 		}
--- a/pkg/codegen/go/gen.go
+++ b/pkg/codegen/go/gen.go
@@ -98,8 +98,9 @@ type pkgContext struct {
 	packages       map[string]*pkgContext
 
 	// Name overrides set in GoPackageInfo
-	modToPkg         map[string]string // Module name -> package name
-	pkgImportAliases map[string]string // Package name -> import alias
+	modToPkg           map[string]string            // Module name -> package name
+	pkgImportAliases   map[string]string            // Package name -> import alias
+	typeRenameMappings map[string]TypeRenameMapping // package name -> map[type name] -> renamed name
 }
 
 func (pkg *pkgContext) detailsForType(t *schema.ObjectType) *typeDetails {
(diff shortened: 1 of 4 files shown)


mikhailshilkov (approved):
LGTM

mikhailshilkov (approved):

EvanBoyle (commented):
Have we carefully gone through the SDK diffs? It looks like there's at least one instance of a break in AWS where there wasn't a compilation error previously:

https://github.com/pulumi-bot/pulumi-aws/blob/1abe2452facb0f9c301980fcd7c53db09f2d6df1/sdk/go/aws/kinesis/pulumiTypes.go

https://github.com/pulumi/pulumi-aws/search?q=%22type+AnalyticsApplicationOutput%22

We were generating an OutputOutput. Now we're generating OutputTypeOutput. I could be convinced it's OK to do this

EvanBoyle:
I think this change makes those names unique globally across the entire go module? I suppose we only really need to worry about this at a "package" level (in go terms).

viveklak:
> Have we carefully gone through the SDK diffs? It looks like there's at least one instance of a break in AWS where there wasn't a compilation error previously:
> 
> https://github.com/pulumi-bot/pulumi-aws/blob/1abe2452facb0f9c301980fcd7c53db09f2d6df1/sdk/go/aws/kinesis/pulumiTypes.go
> 
> https://github.com/pulumi/pulumi-aws/search?q=%22type+AnalyticsApplicationOutput%22
> 
> We were generating an OutputOutput. Now we're generating OutputTypeOutput. I could be convinced it's OK to do this

Thanks for pointing out @EvanBoyle ! There is in fact a compilation error because we have conflicting types for  [`AnalyticsApplicationOutput` type](https://github.com/pulumi/pulumi-aws/blob/master/sdk/go/aws/kinesis/pulumiTypes.go#L2204) and the [output type for `AnalyticsApplication`](https://github.com/pulumi/pulumi-aws/blob/master/sdk/go/aws/kinesis/analyticsApplication.go#L277). The above change would "fix" it but I have been worried about potentially inadvertently impacting existing users as you were pointing out. 

I see the following options:
1. Go ahead with the blanket renaming strategy as implemented here. Quantify and evaluate the "break" to existing packages against the breaking-change guidelines.
2. Manually curate the renames for such conflicts and inject them through language specific metadata as proposed here: https://github.com/pulumi/pulumi-azure-nextgen/issues/137#issuecomment-763776126. This allows for more nuanced handling of renames. e.g. penalize new types and leave old types unchanged etc.However,  this can be quite painstaking for large packages, especially azure-nextgen where these are very common.

Keen to hear your thoughts on the above. All ears for other ideas as well!

CC @mikhailshilkov

mikhailshilkov:
Can we do (1) and apply (2) to places where we want to avoid breaks?

> There is in fact a compilation error 

Does this mean the existing code is unusable? If so, I don't think your change is a break at all?

viveklak:
> Can we do (1) and apply (2) to places where we want to avoid breaks?
> 

Yes I believe we can do that.

> 
> Does this mean the existing code is unusable? If so, I don't think your change is a break at all?

Yes - I believe that package is not importable as it stands. However I haven't conducted a thorough enough audit to make sure we don't have instances of a break in other SDKs. I will take a better look through in the morning.

viveklak:
@EvanBoyle @mikhailshilkov as discussed offline, this adds the targeted flag to allow one-off renames. Subsequently I will add it to the azure-nextgen schema to workaround the issue in https://github.com/pulumi/pulumi-azure-nextgen/issues/137. The longer term fix is tracked in https://github.com/pulumi/pulumi/issues/6170.

EvanBoyle (commented):
Two questions:

1. Are we skipping updates to the program generators? I imagine they need to be aware of this feature, but not sure if/how that impacts azure for this targeted fix. 
2. I am less familiar with the SDK. I assume `tokenForType` is the correct layer to do this such that all references to this renamed type get updated (for instance other types that consume this type, other parts of the schema that use this type as an input or output type). I don't see test coverage that makes this obvious, so just want to ask to confirm.

viveklak:
> Two questions:
> 
> 1. Are we skipping updates to the program generators? I imagine they need to be aware of this feature, but not sure if/how that impacts azure for this targeted fix.

Good point. I am looking through the program gen code to see how we handle renamed types and I don't really see any such handling there so far. I will try it out with a test case and report back. If this is not a regression, I will cut a separate ticket to track programgen handling of renamed types.

> 2. I am less familiar with the SDK. I assume `tokenForType` is the correct layer to do this such that all references to this renamed type get updated (for instance other types that consume this type, other parts of the schema that use this type as an input or output type). I don't see test coverage that makes this obvious, so just want to ask to confirm.

That is my understanding so far. I am currently verifying this works for azure-nextgen. Will provide a link to the SDK changes resulting from this.

mikhailshilkov (approved):
This looks fine as a temporary workaround if we think it brings peace to Azure NextGen. I hope we can tear this down soon with a more generic solution that doesn't require manual intervention.

EvanBoyle:
related to https://github.com/pulumi/pulumi/issues/6170 and https://github.com/pulumi/pulumi/pull/6128

This is a schematype colliding with generated types (codegen error). @viveklak can you take a look?

stack72:
@viveklak is this still needed or can we close this out?

mikhailshilkov:
I think this can be closed

viveklak:
Yes good to close. Superseded by some of the other codegen improvements.

PR #6139, opened 2021-01-19
[sdk/dotnet] - Thread-safe concurrency-friendly global state

orionstudt:
This PR contains the changes necessary to allow `Deployment.Run` methods to run concurrently without the independent deployments interfering with the global state of the others - primarily by taking advantage of the functionality of `AsyncLocal<T>`.

Here are some articles that can help to explain how `AsyncLocal<T>` scope functions:
- https://riptutorial.com/dot-net/example/14971/flowing-execution-context-with-asynclocal
- https://blog.stephencleary.com/2013/04/implicit-async-context-asynclocal.html
- https://blog.stephencleary.com/2016/12/eliding-async-await.html (specifically the AsyncLocal section)

The `AsyncLocal<T>` encapsulation works for us because of the following assumptions:
1. Consumers are not accessing `Deployment.Instance` _before_ any of the `Deployment.Run` methods are called. This is a safe assumption because our `Deployment.Instance` has an internal setter and is only set in `CreateRunner` which is an internal method used by the `Deployment.Run` methods to create an `IRunner` instance.
2. Within the scope of a `Deployment.Run` call, `Deployment.Instance` is only set once at the very top of execution. Again, this is a safe assumption because we are only setting `Deployment.Instance` in `CreateRunner`. Since `AsyncLocal<T>` value flows downwards in async execution, if `Deployment.Instance` were set again later in downstream execution than it would not flow upwards into the calling scope (which would be bad). This is why our setting it once at the very top works for us.


sdk/dotnet/Pulumi/Serialization/ResourcePackages.cs
```diff
@@ -10,14 +10,35 @@
 
 namespace Pulumi
 {
-    internal static class ResourcePackages
+    internal class ResourcePackages
```
mikhailshilkov, line 13: I'm somewhat confused why you needed to make all the changes related to `ResourcePackages` in this PR. Why can't it be static? Are we assuming that multiple deployments may run from different app domains? Why do you need to introduce an external resource test project? Why add the `resourcePackageAssemblies` parameter to `RunAsync`?

orionstudt: The external assembly support is not relevant for console app execution, because with console app execution the executing assembly will always reference the necessary resource package assemblies so it will always be OK to fall back on the on `AppDomain.CurrentDomain`. So strictly speaking, it didn't *need* to be in this PR.

But, in that vein, concurrent execution isn't necessary for console execution either.

In any scenario where you actually might *need* to make use of concurrent execution, it is important that the executing assembly be able to explicitly pass in assemblies to be used as resource package assemblies. Because:
1. Individual deployments might use different resource package assemblies
2. The executing assembly (like a theoretical Web API) might not explicitly reference those resource package assemblies itself so `AppDomain.CurrentDomain` wouldn't find them
3. Those assemblies might be loaded at runtime

At my company we will be using an internal Web API to automate deployment, and our pulumi deployments are abstracted as plugins. There's a plugin for an "application type" i.e. a plugin to deploy a static website, or one of our larger proprietary application types. So, the Web API is going to load the plugin to get the pulumi deployment code. It will not itself have the resource package assemblies in its `AppDomain.CurrentDomain` necessary for the deployment to function as needed.

And since it is possible for those assemblies to be loaded at runtime, Pulumi should be sure not to cache any of them too. Because if a reference is held to an `Assembly` or `Type` than that assembly can never be un-loaded.

mikhailshilkov: Thank you for the great comment! I'm surprised you have put much thought into this functionality which is brand-new and not officially released yet (it's only needed for multi-language components).

Yet, without judging it as right or wrong, could you please split this change out of this PR?

orionstudt: I'm confused- how would splitting this change out look? Is the point of this PR not to move away from static global state so pulumi can support concurrency? If this PR were to leave some static global state un-accounted for than it would be incomplete, no?

mikhailshilkov: > it would be incomplete, no?

IMO, it doesn't have to be complete - it should move a needle and fix a specific problem. For instance, the problem of not being able to run tests in parallel.

As of now, is there an observable behavior for this (ResourcePackages) global state, or a limitation that it brings?

orionstudt: > For instance, the problem of not being able to run tests in parallel.

For what it's worth, I still was not able to enable test collection parallelization after the `AsyncLocal<T>` change to `Deployment.Instance`. After doing so I was encountering intermittent issues when running tests in parallel from the pulumi CLI, unrelated to anything in the dotnet SDK. It was trying to grab multiple file locks on my local pulumi `credentials.json` at once.

Back to topic though - Is the fact that it would be unable to function with runtime assemblies or assemblies not referenced by the executing assembly not a limitation?

mikhailshilkov: > After doing so I was encountering intermittent issues when running tests in parallel from the pulumi CLI, unrelated to anything in the dotnet SDK. 

This sounds to be specific to the automation API. Are you able to enable parallel tests in this PR?

> assemblies not referenced by the executing assembly not a limitation?

I'm fine with this for now, TBH. My goal is to have the smallest PR possible that still delivers value. If we need more PRs later, that's okay.

orionstudt: OK, I have backtracked the resource package changes. Yes, you can enable test parallelization in this PR.

EvanBoyle (commented):
My only question here is around config, given that it is read from environment variables. It looks like you've already taken care of this by adding a new method manage config directly for inline programs: https://github.com/pulumi/pulumi/pull/5761/files#diff-21f2270d821d960ebfe77162b8b0013d03bb5daae6a9d053328b806fed3c0110

Nice! 

@mikhailshilkov any additional feedback or are we ready to merge this?

orionstudt:
That's a good thing to note. This PR doesn't expose concurrency (in the sense that consumers would use pulumi concurrently) because it doesn't also allow you to provide all of those parameters to the Deployment instance on instantiation and still relies on environment variables.

But if the scope is just to run our tests in parallel and make the necessary step to allow concurrency for the Automation API, than I think it suffices.

orionstudt:
This should now be updated to reflect the changes with merging in the automation API preview.

orionstudt:
@mikhailshilkov anything holding us back here that I can help with?

mikhailshilkov:
@orionstudt Nothing in particular - I haven't had a chance to look into it thoroughly.

t0yv0:
Changes look good to me. Sorry for delay reviewing.

Taking it into https://github.com/pulumi/pulumi/pull/6671 to do the trivial master merge and testing.

I am slightly suspicious of "magic" dynamically scoped AsyncLocal variables but it makes sense, if refactoring away global static .Instance in favor of explicitly propagated parameter is probably a combination of tedious and breaking.

orionstudt:
@t0yv0 OK this PR should be back up to snuff. I tried to find the lock exception in the previously failed run on the other PR but couldn't find the specific message that you saw?

t0yv0:
The failure I had on Mac OS said this:
```
--- FAIL: TestLocalStateLocking (8.75s)
2594
Error:     environment.go:87: Created new test environment:  /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/test-env653913954
2595
Error:     pulumi.go:32: Running command git init
2596
Error:     stack_test.go:419: Running command pulumi stack init foo
2597
Error:     stack_test.go:420: Running command yarn install
2598
Error:     stack_test.go:421: Running command yarn link @pulumi/pulumi
2599
Error:     stack_test.go:431: Running command pulumi up --non-interactive --skip-preview --yes
2600
Error:     stack_test.go:431: Running command pulumi up --non-interactive --skip-preview --yes
2601
Error:     stack_test.go:431: Running command pulumi up --non-interactive --skip-preview --yes
2602
Error:     stack_test.go:431: Running command pulumi up --non-interactive --skip-preview --yes
2603
Error:     stack_test.go:431: Running command pulumi up --non-interactive --skip-preview --yes
2604
Error:     stack_test.go:431: Running command pulumi up --non-interactive --skip-preview --yes
2605
Error:     stack_test.go:431: Running command pulumi up --non-interactive --skip-preview --yes
2606
Error:     stack_test.go:431: Running command pulumi up --non-interactive --skip-preview --yes
2607
Error:     stack_test.go:431: Running command pulumi up --non-interactive --skip-preview --yes
2608
Error:     stack_test.go:431: Running command pulumi up --non-interactive --skip-preview --yes
2609
Error:     stack_test.go:446: error: the stack is currently locked by 1 lock(s). Either wait for the other process(es) to end or manually delete the lock file(s).
2610
          file:///var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/test-env653913954/.pulumi/locks/foo/5904a646-2abc-470e-a03f-6df55ed5590d.json: created by runner@Mac-1617894141134.local (pid 37401) at 2021-04-08T15:58:29Z
2611
        
2612
    stack_test.go:445: 
2613
        	Error Trace:	stack_test.go:445
2614
        	Error:      	"error: blob (key \".pulumi/locks/foo/39ec34fd-6c15-426b-a7b2-5f9f217f014b.json\") (code=NotFound): stat /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/test-env653913954/.pulumi/locks/foo/39ec34fd-6c15-426b-a7b2-5f9f217f014b.json: no such file or directory\n" does not contain "the stack is currently locked by 1 lock(s)"
2615
        	Test:       	TestLocalStateLocking
2616
Error:     stack_test.go:446: error: blob (key ".pulumi/locks/foo/39ec34fd-6c15-426b-a7b2-5f9f217f014b.json") (code=NotFound): stat /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/test-env653913954/.pulumi/locks/foo/39ec34fd-6c15-426b-a7b2-5f9f217f014b.json: no such file or directory
2617
        
2618
    stack_test.go:445: 
2619
        	Error Trace:	stack_test.go:445
2620
        	Error:      	"error: blob (key \".pulumi/locks/foo/39ec34fd-6c15-426b-a7b2-5f9f217f014b.json\") (code=NotFound): stat /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/test-env653913954/.pulumi/locks/foo/39ec34fd-6c15-426b-a7b2-5f9f217f014b.json: no such file or directory\n" does not contain "the stack is currently locked by 1 lock(s)"
2621
        	Test:       	TestLocalStateLocking
2622
Error:     stack_test.go:446: error: blob (key ".pulumi/locks/foo/39ec34fd-6c15-426b-a7b2-5f9f217f014b.json") (code=NotFound): stat /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/test-env653913954/.pulumi/locks/foo/39ec34fd-6c15-426b-a7b2-5f9f217f014b.json: no such file or directory
2623
        
2624
    stack_test.go:445: 
2625
        	Error Trace:	stack_test.go:445
2626
        	Error:      	"error: blob (key \".pulumi/locks/foo/39ec34fd-6c15-426b-a7b2-5f9f217f014b.json\") (code=NotFound): stat /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/test-env653913954/.pulumi/locks/foo/39ec34fd-6c15-426b-a7b2-5f9f217f014b.json: no such file or directory\n" does not contain "the stack is currently locked by 1 lock(s)"
2627
        	Test:       	TestLocalStateLocking
2628
Error:     stack_test.go:446: error: blob (key ".pulumi/locks/foo/39ec34fd-6c15-426b-a7b2-5f9f217f014b.json") (code=NotFound): stat /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/test-env653913954/.pulumi/locks/foo/39ec34fd-6c15-426b-a7b2-5f9f217f014b.json: no such file or directory
2629
        
2630
Error:     stack_test.go:446: error: the stack is currently locked by 1 lock(s). Either wait for the other process(es) to end or manually delete the lock file(s).
2631
          file:///var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/test-env653913954/.pulumi/locks/foo/1e50d88c-44d0-4c3a-a44c-df647e06b864.json: created by runner@Mac-1617894141134.local (pid 37405) at 2021-04-08T15:58:29Z
2632
        
2633
Error:     stack_test.go:446: error: the stack is currently locked by 1 lock(s). Either wait for the other process(es) to end or manually delete the lock file(s).
2634
          file:///var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/test-env653913954/.pulumi/locks/foo/1e50d88c-44d0-4c3a-a44c-df647e06b864.json: created by runner@Mac-1617894141134.local (pid 37405) at 2021-04-08T15:58:29Z
2635
        
2636
Error:     stack_test.go:446: error: the stack is currently locked by 1 lock(s). Either wait for the other process(es) to end or manually delete the lock file(s).
2637
          file:///var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/test-env653913954/.pulumi/locks/foo/1e50d88c-44d0-4c3a-a44c-df647e06b864.json: created by runner@Mac-1617894141134.local (pid 37405) at 2021-04-08T15:58:29Z
2638
        
2639
Error:     stack_test.go:446: error: the stack is currently locked by 1 lock(s). Either wait for the other process(es) to end or manually delete the lock file(s).
2640
          file:///var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/test-env653913954/.pulumi/locks/foo/1e50d88c-44d0-4c3a-a44c-df647e06b864.json: created by runner@Mac-1617894141134.local (pid 37405) at 2021-04-08T15:58:29Z
2641
        
2642
Error:     stack_test.go:446: error: the stack is currently locked by 1 lock(s). Either wait for the other process(es) to end or manually delete the lock file(s).
2643
          file:///var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/test-env653913954/.pulumi/locks/foo/1e50d88c-44d0-4c3a-a44c-df647e06b864.json: created by runner@Mac-1617894141134.local (pid 37405) at 2021-04-08T15:58:29Z
2644
        
2645
FAIL
```

Maybe that individual test is flaky also. Hm..

t0yv0:
Yeah let me dig in.

t0yv0:
It clearly is unrelated to the PR so will not block.

t0yv0 (approved):

PR #6142, opened 2021-01-19
Add resource ref unit tests for the Go SDK.

pgavlin:
- Add tests that deserialize known custom and component resources
- Add tests that deserialize missing custom and component resources

These changes also add support for deserializing resources with missing
modules/packages. Such resources are deserialized as generic component,
custom, or provider resources as appropriate.

Contributes to #5943.


stack72 (approved):
This fixes up the issue with the previous PR - that was already approved - thanks @pgavlin

Issue #6135, opened 2021-01-18
Engine throws with panic error when using Tags as below


WillooWisp:
Creating this issue here as suggested from original issue, https://github.com/pulumi/pulumi-azure-nextgen/issues/216.

Adding this to Pulumi.AzureNextGen.Web.V20200601.Inputs.WebAppArgs for a Pulumi.AzureNextGen.Web.V20200601.WebApp, where the appServicePlanName variable is a simple string.

```
Tags = new InputMap<string>
{
    { $"hidden-related:/subscriptions/<subscriptionId>/resourceGroups/{resourceGroup.Name}/providers/Microsoft.Web/serverfarms/{appServicePlanName}", "empty" }
}
```

Error during preview when web app resource and app service plan resource have been imported already.

```
panic: fatal: An assertion has failed

goroutine 25 [running]:
github.com/pulumi/pulumi/sdk/v2/go/common/util/contract.failfast(...)
        /Users/runner/work/pulumi/pulumi/sdk/go/common/util/contract/failfast.go:23
github.com/pulumi/pulumi/sdk/v2/go/common/util/contract.Assert(...)
        /Users/runner/work/pulumi/pulumi/sdk/go/common/util/contract/assert.go:26
github.com/pulumi/pulumi/pkg/v2/backend/display.translateDetailedDiff(0x1cbd6a4, 0x6, 0xc0012ac720, 0x5c, 0xc0019d7680, 0x22, 0xc001554a20, 0xc001554ab0, 0xc001554b40, 0x0, ...)
        /Users/runner/work/pulumi/pulumi/pkg/backend/display/detailedDiff.go:145 +0x237
github.com/pulumi/pulumi/pkg/v2/backend/display.renderDiff(0x1fd16c0, 0xc001509620, 0x1cbd6a4, 0x6, 0xc0012ac720, 0x5c, 0xc0019d7680, 0x22, 0xc001554a20, 0xc001554ab0, ...)
        /Users/runner/work/pulumi/pulumi/pkg/backend/display/diff.go:322 +0x145
github.com/pulumi/pulumi/pkg/v2/backend/display.renderDiffResourcePreEvent(0x1cbd6a4, 0x6, 0xc0012ac720, 0x5c, 0xc0019d7680, 0x22, 0xc001554a20, 0xc001554ab0, 0xc001554b40, 0x0, ...)
        /Users/runner/work/pulumi/pulumi/pkg/backend/display/diff.go:351 +0x298
[...]

PR #6159, opened 2021-01-21
Relax the diff path check to avoid panic

mikhailshilkov:
Fixes https://github.com/pulumi/pulumi/issues/6135

The cause is probably the Azure NextGen provider that returns the diff in tags as if tags are a typed object rather than an untyped map with arbitrary keys. In the case of panic, there is a tag that contains `[Something]` which fails to parse as a valid path.

However, my point in this PR is that the diff comes from a semi-trusted source (an arbitrary provider, could be 3rd party, eventually), so it's not worth panicking. Rather, we can assume a reasonable default.  In this case, the diff shows as 
image
which isn't perfect but probably way better than panic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ CHANGELOG
   only when the monitor reports that resource references are supported.
   [#6172](https://github.com/pulumi/pulumi/pull/6172)
 
+- [CLI] Avoid panic for diffs with invalid property paths.
+  [#6159](https://github.com/pulumi/pulumi/pull/6159)
+
 ## 2.18.1 (2021-01-21)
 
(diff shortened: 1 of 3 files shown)


pgavlin:
It looks like this essentially treats detailed diff elements with bad paths as leaf properties in the final object diff. Is that right?

mikhailshilkov:
> It looks like this essentially treats detailed diff elements with bad paths as leaf properties in the final object diff. Is that right?

Yes, this sounds correct.

pgavlin (approved):
FWIW, the provider can avoid this by escaping the path. We should probably add some helpers to make the creation of properly-escaped property paths easy.

PR #6165, opened 2021-01-21
[Feat] Upgrade 3.1  To .NET 5 And Upgrade Serilog To 2.10.0

MeladKamari:
Upgrade Project To Stable .NET 5 !

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -26,7 +26,7 @@ jobs:
       matrix:
         go-version: [ 1.15.x ]
         python-version: [ 3.9.x ]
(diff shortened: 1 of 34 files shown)


mikhailshilkov:
.NET Core 3.1 is the current LTS release, while .NET 5.0 is not. We likely want to keep 3.1 as the default until .NET 6.0.

mikhailshilkov:
To add, it's totally possible to use .NET 5 in user's projects right now ([example](https://github.com/pulumi/examples/tree/master/azure-nextgen-cs-net5-aks-webapp)), so I'm not sure what an upgrade would give us

PR #6179, opened 2021-01-22
[codegen/go] Fix default value handling for enums

viveklak:
Fixes a couple of bugs in go codegen for enums - specifically around setting defaults correctly.
Updated tests with improved coverage for the above.

The PR is split in 2 commits, the first implements the core Go changes and adds the relevant go test data. The latter updates testdata for other languages due to the test schema update.

### DIFF
--- a/pkg/codegen/internal/test/testdata/simple-enum-schema/go/plant/tree/v1/rubberTree.go
+++ b/pkg/codegen/internal/test/testdata/simple-enum-schema/go/plant/tree/v1/rubberTree.go
@@ -16,7 +16,9 @@ type RubberTree struct {
 	pulumi.CustomResourceState
 
 	Container plant.ContainerPtrOutput `pulumi:"container"`
+	Diameter  pulumi.Float64Output     `pulumi:"diameter"`
 	Farm      pulumi.StringPtrOutput   `pulumi:"farm"`
+	Size      pulumi.StringPtrOutput   `pulumi:"size"`
 	Type      pulumi.StringOutput      `pulumi:"type"`
 }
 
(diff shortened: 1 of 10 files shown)


pkg/codegen/internal/test/testdata/simple-enum-schema/go/plant/tree/v1/rubberTree.go
```diff
@@ -30,6 +31,13 @@ func NewRubberTree(ctx *pulumi.Context,
 	if args.Farm == nil {
 		args.Farm = pulumi.StringPtr("(unknown)")
 	}
+	if args.Size == nil {
+		e := TreeSize("medium")
+		args.Size = &e
+	}
+	if !args.Type.Valid() {
+		args.Type = RubberTreeVariety("Burgundy")
+	}
```
komalali, line 40: I'm wondering if this would be confusing for the user. If they enter an invalid `Type`, should we tell them it's invalid rather than setting a default that they wouldn't be aware of? 

Since these are just string types underneath it all, in the situation where the valid options are "Hot" or "Cool" and the default is "Cool" but the user enters "hot" on accident when they meant "Hot" and instead of telling them there's an error we just automatically set "Cool" it could be confusing from a user perspective.

viveklak: Fair point. The primary goal in this specific call site is just to determine if the required enum is not set and if so set the default. I didn't have a great use for Valid as yet so decided to throw it in but I can see the confusion here. Will update to check for "zero"-ness instead.

viveklak: Updated.

komalali (approved):

Issue #6110, opened 2021-01-14
Test our full language runtime version matrix


lukehoban:
Our test coverage should at least nightly include validation against all major supported language runtime versions - or at least lowest supported and highest supported.
- [ ] Python 3.7+ (currently 3.7❌, 3.8❌, 3.9✅ and 3.10❌)
- [ ] Node.js Current, Active and Maintenance versions (currently Node.js 20❌,19❌,18❌, 16❌, 14✅)
- [ ] .NET 3.1+ (currently 3.1✅, 6.0❌)
- [ ] Go 1.18+ (currently 1.18❌, 1.19❌)
- [ ] Java 11+ (currently 11✅, 17❌,19❌, 20❌, 21❌)

stack72:
The work required here includes the ability to decouple our builds from our tests - when we currently set a language, we immediately use that same language version for testing - we need to be able to build a binary and SDK and then use that on systems with the other language runtimes involved

t0yv0:
.NET 5 & 6  requested here also

lukehoban:
Updated to include Node.js 20, 19 and 18, and Go 1.18.  For Node.js in particular, we are now only testing a "Maintenance" version.

kpitzen:
Do we expect to continue supporting Python 3.6.x? It was EOL Dec 2021

Frassle:
No. We need to update the table above, and the language pages as well.
I think we'll take a similar stance for all the languages as our nodejs stance (any of the Current, Active and Maintenance LTS versions).
Under that we would for example stop supporting python 3.6, dotnet 5.0, and in December also dotnet 3.1.

Frassle:
I've updated the table at the top to reflect our currently supported versions (as declared on the language pages, e.g. https://www.pulumi.com/docs/intro/languages/python/).
Some of these will change again soon. E.g. dotnet 7.0 is being released in November and dotnet3.1 will no longer be supported in December.

Frassle:
Note that Java 18 is technically currently supported but falls out of support _this_ month so probably isn't worth testing.

mikhailshilkov:
> cartesian product 

Could you please help me why we need a product here? Can't we run separate jobs for each version of each runtime, so that  the result is a sum of all versions, not a product?

justinvp:
[...]

PR #6193, opened 2021-01-25
Split the windows build + testing to be separate jobs to allow for future language matrix tests

stack72:
Related: #6110

This is the first part of the work to allow us to be able to test multiple versions of supported languages

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -75,7 +75,7 @@ jobs:
   publish-binaries:
     name: Publish Binaries
     runs-on: macos-latest
-    needs: [build-and-test, windows-build]
+    needs: [build-and-test, windows-test]
     strategy:
       matrix:
         go-version: [ 1.15.x ]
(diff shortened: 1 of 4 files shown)


.github/workflows/master.yml
```diff
@@ -269,14 +257,256 @@ jobs:
...
+        with:
+          name: pulumi-windows-python-sdk.tar.gz
+          path: ${{ github.workspace}}\src\github.com\${{ github.repository }}
+      - name: Uncompress SDK folder
+        run: |
+          cd src\github.com\${{ github.repository }}
+          tar -zxf ./nodejs-sdk.tar.gz -C sdk\nodejs
+          tar -zxf ./python-sdk.tar.gz -C sdk\python
+      - name: Uninstall Choco Pulumi
+        run: choco uninstall pulumi -y
```
komalali, line 371: Why do we do this? It looks like this step took 3 minutes in the action, wondering if there's a way to avoid it.

komalali (approved):
Just a question about one of the steps but otherwise LGTM!

.github/workflows/master.yml
stack72: This is a temp step that I am trying to work through a way to remove - this is ensuring that the version of Pulumi we have is the *exact* version we want

PR #6196, opened 2021-01-25
Adding try-catch to avoid complicated output register crash directly in python sdk

wuqunfei:


wuqunfei:
https://github.com/pulumi/pulumi/pull/6196 this is the code example

justinvp:
@wuqunfei, can you show an example Pulumi program where this is hit, and the output before/after? Thanks!

leezen:
@wuqunfei Could you please take a look at Justin's question above?

emiliza:
Closing due to inactivity. @wuqunfei if this is still needed - please open a new PR with the requested info from Justin

PR #6209, opened 2021-01-27
Various fixes to go codegen to address SDK compilation problems

viveklak:
Multiple fixes to compilation problems in various generated Go SDKs.
Fixes https://github.com/pulumi/pulumi-aws/issues/1321 among other issues.

This follows up from https://github.com/pulumi/pulumi/pull/6179 

Having run on some major SDKs, the last set of compilation issues are due to:
1. Lack of Array and Map input/output variants generated for enum types
2. Name collisions among input/output variants of types and other synonymous types.
These are being addressed in a separate PR.

### DIFF
--- a/pkg/codegen/go/gen.go
+++ b/pkg/codegen/go/gen.go
@@ -92,6 +92,7 @@ type pkgContext struct {
 	resources      []*schema.Resource
 	functions      []*schema.Function
 	names          codegen.StringSet
+	renamed        map[string]string
 	functionNames  map[*schema.Function]string
 	needsUtils     bool
 	tool           string
@@ -139,14 +140,20 @@ func (pkg *pkgContext) tokenToType(tok string) string {
 
 	mod, name := pkg.tokenToPackage(tok), components[2]
 
-	// If the package containing the type's token already has a resource with the
-	// same name, add a `Type` suffix.
 	modPkg, ok := pkg.packages[mod]
-
 	name = Title(name)
+
 	if ok {
-		if modPkg.names.Has(name) {
-			name += "Type"
+		newName, renamed := modPkg.renamed[name]
+		if renamed {
+			name = newName
+		} else if modPkg.names.Has(name) {
+			// If the package containing the type's token already has a resource with the
+			// same name, add a `Type` suffix.
+			newName = name + "Type"
+			modPkg.renamed[name] = newName
+			modPkg.names.Add(newName)
+			name = newName
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/codegen/go/gen.go
```diff
@@ -1757,6 +1786,8 @@ func generatePackageContextMap(tool string, pkg *schema.Package, goInfo GoPackag
 		pkg.resources = append(pkg.resources, r)
 
 		pkg.names.Add(resourceName(r))
+		pkg.names.Add(resourceName(r) + "Input")
+		pkg.names.Add(resourceName(r) + "Output")
```
viveklak, line 1790: These two lines are probably the most controversial since these will rename types that have already been emitted. However, in all such cases the code is unable to be compiled due to a redeclaration, e.g. for AWS:
```
# github.com/pulumi/pulumi-aws/sdk/v3/go/aws/kinesis
go/aws/kinesis/analyticsApplication.go:274:43: impossible type assertion:
	AnalyticsApplicationOutput does not implement pulumi.Output (missing Apply method)
go/aws/kinesis/pulumiTypes.go:2204:6: AnalyticsApplicationOutput redeclared in this block
	previous declaration at go/aws/kinesis/analyticsApplication.go:356:6
```

As a result, these don't violate our breaking change policy.

pkg/codegen/go/gen.go
```diff
@@ -784,7 +791,12 @@ func (pkg *pkgContext) genOutputTypes(w io.Writer, t *schema.ObjectType, details
 		printCommentWithDeprecationMessage(w, p.Comment, p.DeprecationMessage, false)
 		outputType, applyType := pkg.outputType(p.Type, !p.IsRequired), pkg.plainType(p.Type, !p.IsRequired)
 
-		fmt.Fprintf(w, "func (o %sOutput) %s() %s {\n", name, Title(p.Name), outputType)
+		propName := Title(p.Name)
```
viveklak, line 794: This prevents an error such as:
```
../../pulumi-azure-nextgen-provider/sdk/go/azure/containerregistry/latest/pulumiTypes.go:455:43: impossible type assertion:
    	ArgumentResponseOutput does not implement pulumi.Output (wrong type for IsSecret method)
    		have IsSecret() pulumi.BoolPtrOutput
    		want IsSecret() bool
    ../../pulumi-azure-nextgen-provider/sdk/go/azure/containerregistry/latest/pulumiTypes.go:530:4: impossible type assertion:
    	ArgumentResponseOutput does not implement pulumi.Output (wrong type for IsSecret method)
    		have IsSecret() pulumi.BoolPtrOutput
    		want IsSecret() bool
```

pkg/codegen/go/gen.go
```diff
@@ -784,7 +791,12 @@ func (pkg *pkgContext) genOutputTypes(w io.Writer, t *schema.ObjectType, details
...
+		switch strings.ToLower(p.Name) {
+		case "elementtype", "issecret":
```
viveklak, line 796: This prevents an error such as:
```
../../pulumi-azure-nextgen-provider/sdk/go/azure/containerregistry/latest/pulumiTypes.go:455:43: impossible type assertion:
    	ArgumentResponseOutput does not implement pulumi.Output (wrong type for IsSecret method)
    		have IsSecret() pulumi.BoolPtrOutput
    		want IsSecret() bool
    ../../pulumi-azure-nextgen-provider/sdk/go/azure/containerregistry/latest/pulumiTypes.go:530:4: impossible type assertion:
    	ArgumentResponseOutput does not implement pulumi.Output (wrong type for IsSecret method)
    		have IsSecret() pulumi.BoolPtrOutput
    		want IsSecret() bool
```

lblackstone (approved):

pkg/codegen/go/gen.go
EvanBoyle: Do you think it's problematic that we rely on the ordering of resources in the schema? Would it be better to sort (probably in multiple places here) to get more predictability?

viveklak: > Do you think it's problematic that we rely on the ordering of resources in the schema? Would it be better to sort (probably in multiple places here) to get more predictability?

Thanks for the pointer. In general I think we want to be as deterministic as possible. I will follow up shortly with sorted version (I have to do something similar for collisions with input/output variants among types themselves.

Issue #6195, opened 2021-01-25
Potential Panic when changing secrets provider


stack72:
```
Enter your new passphrase to protect config/secrets:
Re-enter your new passphrase to confirm:
Migrating old configuration and state to new secrets provider
Enter your passphrase to unlock config/secrets
    (set PULUMI_CONFIG_PASSPHRASE or PULUMI_CONFIG_PASSPHRASE_FILE to remember):
I0125 09:56:09.622710   87766 sink.go:154] defaultSink::Error(error: could not deserialize deployment: decrypting secret value: failed to decrypt: incorrect passphrase, please set PULUMI_CONFIG_PASSPHRASE to the correct passphrase
main.checkDeploymentVersionError
    /Users/stack72/code/go/src/github.com/pulumi/pulumi/pkg/cmd/pulumi/util.go:799
main.migrateOldConfigAndCheckpointToNewSecretsProvider
    /Users/stack72/code/go/src/github.com/pulumi/pulumi/pkg/cmd/pulumi/stack_change_secrets_provider.go:154
main.newStackChangeSecretsProviderCmd.func1
    /Users/stack72/code/go/src/github.com/pulumi/pulumi/pkg/cmd/pulumi/stack_change_secrets_provider.go:103
github.com/pulumi/pulumi/sdk/v2/go/common/util/cmdutil.RunFunc.func1
    /Users/stack72/code/go/src/github.com/pulumi/pulumi/sdk/go/common/util/cmdutil/exit.go:96
github.com/pulumi/pulumi/sdk/v2/go/common/util/cmdutil.RunResultFunc.func1
    /Users/stack72/code/go/src/github.com/pulumi/pulumi/sdk/go/common/util/cmdutil/exit.go:112
github.com/spf13/cobra.(*Command).execute
    /private/tmp/hc-pulumi-20210122-82888-144l7b/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:846
github.com/spf13/cobra.(*Command).ExecuteC
    /private/tmp/hc-pulumi-20210122-82888-144l7b/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:950
github.com/spf13/cobra.(*Command).Execute
    /private/tmp/hc-pulumi-20210122-82888-144l7b/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:887
main.main
[...]

PR #6215, opened 2021-01-28
Ensure a user has access to the secret manager before trying to decrypt checkpoint

stack72:
Fixes: #6195

Before:

```
▶ pulumi stack export --show-secrets
error: could not deserialize deployment: decrypting secret value: failed to decrypt: incorrect passphrase, please set PULUMI_CONFIG_PASSPHRASE to the correct passphrase
```

Notice that it doesn't even ask for the passphrase

After adding a simple check to get the secrets manager we can now
ensure that the user has access to the correct credentials before
we even use the secrets manager

```
pulumi stack export --show-secrets
Enter your passphrase to unlock config/secrets
    (set PULUMI_CONFIG_PASSPHRASE or PULUMI_CONFIG_PASSPHRASE_FILE to remember):
{
    "version": 3,
    "deployment": {
        "manifest": {
            "time": "2021-01-27T14:21:23.197301Z",
            "magic": "72fc43d05bda7c857308a563548faa323639f314a8d27c0ce550589753311fad",
            "version": "v2.19.0-alpha.1611602537+g0f06e0393.dirty"
        },
        "secrets_providers": {
            "type": "passphrase",
            "state": {
                "salt": "v1:BW++rEnZJd4=:v1:KS0AGrKyxgf6cpWW:Lw27W/I/sGUV1mvuywjZY/LG3xdEEA=="
            }
        },
        "resources": [
            {
                "urn": "urn:pulumi:password::passphrase-err::pulumi:pulumi:Stack::passphrase-err-password",
                "custom": false,
                "type": "pulumi:pulumi:Stack",
                "outputs": {
                    "x": {
                        "plaintext": "\"Password1234!\""
                    }
                }
            }
        ]
    }
}
```

### DIFF
--- a/pkg/cmd/pulumi/stack_export.go
+++ b/pkg/cmd/pulumi/stack_export.go
@@ -90,6 +90,13 @@ func newStackExportCmd() *cobra.Command {
 			}
 
 			if showSecrets {
+				// Currently, the stack.DefaultSecretsProvider is cached so adding a call to getStackSecretsManager
+				// will ensure that the user has the correct credentials to decrypt the stack deployment
+				_, err := getStackSecretsManager(s)
+				if err != nil {
+					return errors.Wrap(err, "getting secrets manager")
+				}
+
 				snap, err := stack.DeserializeUntypedDeployment(deployment, stack.DefaultSecretsProvider)
 				if err != nil {
 					return checkDeploymentVersionError(err, stackName)
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -58,3 +58,7 @@ type PreviewResult struct {
 
 - [automation/dotnet] Add ReadDiscard OperationType
   [#6493](https://github.com/pulumi/pulumi/pull/6493)
+
+- [cli] Ensure the user has the correct access to the secrets manager before using it as part of
+  `pulumi stack export --show-secrets`.
+  [#6215](https://github.com/pulumi/pulumi/pull/6210)


pkg/cmd/pulumi/stack_export.go
```diff
@@ -90,6 +90,11 @@ func newStackExportCmd() *cobra.Command {
 			}
 
 			if showSecrets {
+				_, err := getStackSecretsManager(s)
+				if err != nil {
+					return errors.Wrap(err, "getting secrets manager")
+				}
```
justinvp, line 96: I would have never guessed that we have to add this call to avoid the error. Is this necessary anytime `stack.DefaultSecretsProvider` is used? Are there any other code paths where this is missing? Is there any way we can make this more explicit?

justinvp (approved):
LGTM but it isn't obvious to me that that call was needed to avoid such an error.

pkg/cmd/pulumi/stack_export.go
stack72: So the issue is only regarding passphrase backend - if the environment variable isn't set, then we don't prompt for the passphrase as part of `stack.DefaultSecretsProvider`. This ultimately breaks the `stack change-secrets-provider` if there is encrypted state. For all the other backends, we usually have Env Vars set to interact with the backend - the passphrase provider doesn't need it.

pgavlin: It would be great to add a comment to this explaining that we need to make this call in order to trigger the passphrase prompt.

Issue #6176, opened 2021-01-22
Ensure master and nightly builds update the npm dev tag


stack72:
Right now, we are only updating the beta tag which isn't correct

We also need to reverse this https://github.com/pulumi/scripts/pull/130 when we are happy it has been updated

PR #6231, opened 2021-02-01
Ensure nightly and master builds with an alpha build version are tagged dev

stack72:
Fixes: #6176


jaxxstorm (approved):

jaxxstorm:
I'm not sure why this is better, but it'll achieve the desired result, so approved

Issue #4132, opened 2020-03-20
Create a canary to ensure that all methods of downloading the CLI work


leezen:
Should cover direct downloads, homebrew, chocolately, etc.

PR #6246, opened 2021-02-03
Adding a nightly cron job to ensure we are testing Pulumi install

stack72:
Fixes: #4132

This will test across all available download styles that are Pulumi
managed. I will not install anything like nixos etc

- [x] Windows
  - [x] chocolatey
  - [x] install script
  - [x] direct download
  - [x] GitHub Action
- [x] macOS
  - [x] homebrew
  - [x] install script
  - [x] direct download
  - [x] GitHub Action
- [x] Linux
  - [x] install script
  - [x] direct download 
  - [x] GitHub Action

### DIFF
--- a/.github/workflows/download-pulumi-cron.yml
+++ b/.github/workflows/download-pulumi-cron.yml
@@ -1,0 +1,141 @@
+name: Download Pulumi Cron
+"on":
+  schedule:
+    - cron: 0 8 * * *
+
+jobs:
+  macos-homebrew-install:
+    name: Install Pulumi with Homebrew on macOS
+    runs-on: macos-latest
+    steps:
+      - name: homebrew install
+        run: brew install pulumi
+      - name: Pulumi Version Details
+        id: vars
+        run: |
+          echo ::set-output name=installed-version::$(pulumi version)
+          echo ::set-output name=expected-version::v$(curl -sS https://www.pulumi.com/latest-version)
+      - name: Error if incorrect version found
+        if: steps.vars.outputs.expected-version != steps.vars.outputs.installed-version
+        run: |
+          echo "Expected version ${{ steps.vars.outputs.expected-version }} but found ${{ steps.vars.outputs.installed-version }}"
+          exit 1
+  macOS-direct-install:
+    name: Install Pulumi via script on macOS
+    runs-on: macos-latest
+    steps:
+      - run: curl -fsSL https://get.pulumi.com | sh
+      - run: echo "/Users/runner/.pulumi/bin" >> $GITHUB_PATH
+      - name: Pulumi Version Details
+        id: vars
+        run: |
+          echo ::set-output name=installed-version::$(pulumi version)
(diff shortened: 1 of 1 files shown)


.github/workflows/download-pulumi-cron.yml
```diff
@@ -0,0 +1,141 @@
...
+        id: vars
+        run: |
+          echo ::set-output name=installed-version::$(pulumi version)
+          echo ::set-output name=expected-version::v$(curl -sS https://www.pulumi.com/latest-version)
+      - run: which pulumi
+      - name: Error if incorrect version found
+        if: steps.vars.outputs.expected-version != steps.vars.outputs.installed-version
+        run: |
+          echo "Expected version ${{ steps.vars.outputs.expected-version }} but found ${{ steps.vars.outputs.installed-version }}"
+          exit 1
```
leezen, line 39: Being unfamiliar w/ GHA, is there a way to have this block be re-used across all the jobs?

.github/workflows/download-pulumi-cron.yml
```diff
@@ -0,0 +1,141 @@
...
+    steps:
+      - name: Direct Download
+        run: curl -L -o pulumi.tar.gz https://get.pulumi.com/releases/sdk/pulumi-v$(curl -sS https://www.pulumi.com/latest-version)-darwin-x64.tar.gz
+      - run: ls -la
+  linux-direct-install:
+    name: Install Pulumi via script on Ubuntu
+    runs-on: ubuntu-latest
+    steps:
+      - name: Remove existing version
+        run: sudo rm /usr/local/bin/pulumi
```
leezen, line 52: Reading the documentation, `echo ... >> $GITHUB_PATH` prepends, in which case, won't the downloaded pulumi supersede this and we don't need to remove?

.github/workflows/download-pulumi-cron.yml
```diff
@@ -0,0 +1,141 @@
+name: Download Pulumi Cron
+"on":
+  schedule:
+    - cron: 0 8 * * *
```
leezen, line 4: I actually think it's worthwhile to run this more frequently (say hourly) and be okay with the fact that it might occasionally fail during a release where the versions will mismatch.

leezen (approved):

.github/workflows/download-pulumi-cron.yml
stack72: I was going to run this once a day first then crank it up when we know it's actually doing the correct thing

.github/workflows/download-pulumi-cron.yml
stack72: I am trying to ensure there isn't anything else clouding the test :) This is a first iteration of it that I will ensure is cleaner when I know it's working well

.github/workflows/download-pulumi-cron.yml
stack72: I wish I could do that :/

stack72:
@leezen this functionality is going to expand and I will ensure to test the GITHUB_PATH to make it cleaner in that next set of work

Issue #5578, opened 2020-10-15
[Automation API] Unintuitive behavior with relative paths (FileAsset, closure serialization) and inline programs


EvanBoyle:
Automation API inline programs create a temporary working directory (unless directly specified) to store pulumi.yaml and to invoke the CLI. This breaks any relative path references the user may define in their inline program:

```ts
const pulumiProgram = () async => {
// at runtime this actually resolves to $AUTOMATION_API_TEMP/build
const relative = "./build"
}
```

We have a few options:

1. Change Automation API to use `.` as the current working dir. This preserves relative paths, but has the downside of leaking files like `pulumi.yaml` and `pulumi.stack.yaml` which the user might not care about or might find confusing to see generated out of nowhere. 
2. Document this behavior and leave it as is. encourage users to only use absolute paths with inline programs. 
3. Do some sort of deeper fix where we specify the temp directory to be used only for settings (yaml files), and use `.` as the CWD for program execution.

EvanBoyle:
Note that this likely breaks closure serialization as `node_modules` won't be in the expected location.

EvanBoyle:
To be clear the current work around it to set an explicit workdir when creating the stack or workspace:

```ts
const args = {
stack: "",
project: "",
program ()=> {},
workDir: ".", // set work dir to current directory
}
```

barclayadam:
To add to this the breakage goes further, as dynamic resources cannot be used without the workaround as dynamic module cannot be found:

`Error: Cannot find module '@pulumi/pulumi/cmd/dynamic-provider'` as well `could not read plugin [C:\Users\barcl\.pulumi\bin\pulumi-resource-pulumi-nodejs.cmd] stdout: EOF.`

Forcing the current workDir makes this work, but can cause further problems due to the files that end up on disk. I think related to #5679 the files on disk can cause issues if you have multiple inline programs with same named stacks but different projects / configs / programs (I've not tracked down exactly the issue).

EvanBoyle:
@barclayadam thanks for reporting. I need some time to think through a resolution here.

roderik:
I had the same issue using the eks package in the automation API.
However, now it writes yaml files in the root of my application, introducing another concurrency issue.

EvanBoyle:
[...]

PR #6247, opened 2021-02-03
prototype fix for auto api relative paths

EvanBoyle:
Prototypes a fix for https://github.com/pulumi/pulumi/issues/5578 by wiring up the `main` field for the generated default `pulumi.yaml` for inline programs in nodejs.

A few things that need to be completed before this can ship:

1. update some of the engine tests that block main being an absolute path, and enforcing main is subdir of root
2. For cases where the user is authoring an `index.ts` that gets compiled into a `/bin/index.js` or other location, do we need convenience util to allow the user to configure this as it impacts relative paths? Like an `auto API build root`? Otherwise they have to specify the entire `ProjectSettings` object, and know to specify the `main` folder. Not making any suggestions here, but something we need to think through documenting at the least.
3. Update other languages. Compiled languages will have to deal with (2) as well. 

Regarding (1), I believe this is safe to do and is a holdover of PPC constraints that are no longer relevant. Would be good to check with @chrsmith to verify. I've summarized my findings in a comment on the relevant code:

https://github.com/pulumi/pulumi/blob/c8c3db0197845fed9bfdde27ad165fc5563e8983/pkg/engine/project.go#L52-L60

I tested this with an S3 bucket object:

```ts
        // write our index.html into the site bucket
        let object = new s3.BucketObject("index", {
            bucket: siteBucket,
            source: new pulumi.asset.FileAsset("./README.md"),
            contentType: "text/html; charset=utf-8",
            key: "index.html"
        });
```
|
Before this change, no object gets uploaded. After the change, it works as expected.

### DIFF
--- a/pkg/engine/project.go
+++ b/pkg/engine/project.go
@@ -18,7 +18,6 @@ import (
 	"os"
 	"path"
 	"path/filepath"
-	"strings"
 
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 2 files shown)


pkg/engine/project.go
```diff
@@ -50,16 +49,20 @@ func getPwdMain(root, main string) (string, string, error) {
...
-		}
+		// Note: I had to relax constraints on main being a subfolder of root. I tried computing a relative
+		// path from the auto api tmp directory. But the tmp directory on osx is `/private/var/tmp` and
+		// computing a relative path doesn't work as `/private/User/...` (where the program would be) isn't
+		// a valid file path.
+		//
+		// The reason we had this check in the first place is due to PPC constraints for hosted execution. Look through the original PR
+		// https://github.com/pulumi/pulumi/pull/615 https://github.com/pulumi/pulumi-ppc/pull/102
+		// It would appear that it is now safe to allow root and main to be in disjoint file trees.
+		// Empyrically from my testing it seems to work, but would be worth checking with @chrsmith
```
lukehoban, line 60: FWIW - this matches my understanding as well.  The original motivation for the constraint was that we needed to be able to zip up the folder containing the `Pulumi.yaml` file and have that zip define the full source context for the deployment.  This is no longer a formal scenario.

That said - it is certainly possible that we developed other implicit dependencies on this requirement elsewhere because this has been a guarantee.  I cannot think of any likely places to be concerned about though (closure serialization came to mind, but it keys off of `package.json` location, not `Pulumi.yaml` location).

chrsmith: Yes, we had the "need to zip everything we'll need up for the PPC" scenario. So any PPC-specific requirements can be ignored.

I don't recall any other invariants we needed to maintain in this area. The only thing I do recall, is support for adding the `main` field to `Pulumi.yaml` to support "up and over" references. (See https://github.com/pulumi/pulumi/pull/582.) ... but that seems like it is essentially what you are trying to enable here.

EvanBoyle:
closed in favor of: https://github.com/pulumi/pulumi/pull/6734

Issue #6228, opened 2021-01-30
[Automation API] - asyncio errors in python inline programs 


komalali:
There is an issue with error-handling in python inline programs.

## Expected Behavior
Error is returned gracefully without asyncio error messages.

## Current Behavior
asyncio errors with `Task exception was never retrieved`.

Full output:
```
pulumi:pulumi:Stack bigquery-dev running
+  gcp:organizations:Project project creating
+  gcp:organizations:Project project created
+  gcp:serviceAccount:Account sa creating
+  gcp:serviceAccount:Account sa created
+  gcp:serviceAccount:Key key creating
+  gcp:projects:IAMBinding permissions creating
+  gcp:projects:IAMBinding permissions creating error: 1 error occurred:
+  gcp:projects:IAMBinding permissions **creating failed** error: 1 error occurred:
+  gcp:serviceAccount:Key key created
Task exception was never retrieved
future: <Task finished name='Task-92' coro=<RPCManager.do_rpc.<locals>.rpc_wrapper() done, defined at /Users/komal/go/src/github.com/pulumi/automation-api-examples/python/pulumi_via_jupyter/venv/lib/python3.8/site-packages/pulumi/runtime/rpc_manager.py:61> exception=<_InactiveRpcError of RPC that terminated with:
	status = StatusCode.UNAVAILABLE
	details = "failed to connect to all addresses"
	debug_error_string = "{"created":"@1612033464.773224000","description":"Failed to pick subchannel","file":"src/core/ext/filters/client_channel/client_channel.cc","file_line":4143,"referenced_errors":[{"created":"@1612033464.772631000","description":"failed to connect to all addresses","file":"src/core/ext/filters/client_channel/lb_policy/pick_first/pick_first.cc","file_line":398,"grpc_status":14}]}"
>>
Traceback (most recent call last):
  File "/Users/komal/go/src/github.com/pulumi/automation-api-examples/python/pulumi_via_jupyter/venv/lib/python3.8/site-packages/pulumi/runtime/rpc_manager.py", line 67, in rpc_wrapper
    result = await rpc
  File "/Users/komal/go/src/github.com/pulumi/automation-api-examples/python/pulumi_via_jupyter/venv/lib/python3.8/site-packages/pulumi/runtime/resource.py", line 495, in do_register
    rpc.resolve_outputs(res, resolver.serialized_props, resp.object, deps, resolvers)
[...]

PR #6249, opened 2021-02-04
[sdk/python] - Gracefully handle monitor shutdown

komalali:
The main goal of this PR is to fix all the areas where we were previously exiting the process when the monitor is unavailable and instead just absorb the grpc `UNAVAILABLE` message. There are more details on the exact behavior this was causing and the motivation behind the change in issue #6228. 

I've tested this manually against my initial program that was logging the asyncio warnings, but I'm  not really sure how to write automated tests for this - open to suggestions if there are any. The way to know that it works is that we don't see any asyncio warnings in the console when running a program that fails.

There are also a few misc. formatting corrections.

Fixes: #6228

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@ CHANGELOG
 =========
 
 ## HEAD (Unreleased)
-_(none)_
+
+- [sdk/python] Gracefully handle monitor shutdown in the python runtime without exiting the process. 
+  [#6249](https://github.com/pulumi/pulumi/pull/6249)
 
 ## 2.20.0 (2021-02-03)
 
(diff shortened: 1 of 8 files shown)


pgavlin (approved):
LGTM. I don't have any great suggestions for testing this, unfortunately.

Is there anywhere else in the Python SDK that we call `sys.exit`? I know I found several in the Node SDK a while back.

komalali:
@pgavlin The only places where it still exists is in [cmd/pulumi-language-python-exec](https://github.com/pulumi/pulumi/blob/master/sdk/python/cmd/pulumi-language-python-exec#L97-L98) and [policy/__main__.py](https://github.com/pulumi/pulumi/blob/master/sdk/python/lib/pulumi/policy/__main__.py#L61-L62) where it makes sense to leave them in since those are entry points into the execution. I got the rest of them in this PR.

EvanBoyle (commented):
👍

PR #6250, opened 2021-02-04
Avoid stack overflow in Azure NextGen program examples

mikhailshilkov:
Extends https://github.com/pulumi/pulumi/pull/5404 to fix https://github.com/pulumi/pulumi/issues/5228 for all Azure NextGen program examples

### DIFF
--- a/pkg/codegen/hcl2/model/type_object.go
+++ b/pkg/codegen/hcl2/model/type_object.go
@@ -179,13 +179,22 @@ func (u *objectTypeUnifier) unify(t *ObjectType) {
 // This conversion is always unsafe, and may fail if the map does not contain an appropriate set of keys for the
 // destination type.
 func (t *ObjectType) ConversionFrom(src Type) ConversionKind {
-	return t.conversionFrom(src, false)
+	return t.conversionFrom(src, false, nil)
 }
 
-func (t *ObjectType) conversionFrom(src Type, unifying bool) ConversionKind {
-	return conversionFrom(t, src, unifying, func() ConversionKind {
+func (t *ObjectType) conversionFrom(src Type, unifying bool, seen map[Type]struct{}) ConversionKind {
+	return conversionFrom(t, src, unifying, seen, func() ConversionKind {
 		switch src := src.(type) {
 		case *ObjectType:
+			if seen != nil {
+				if _, ok := seen[t]; ok {
+					return NoConversion
+				}
+			} else {
+				seen = map[Type]struct{}{}
+			}
+			seen[t] = struct{}{}
+
 			if unifying {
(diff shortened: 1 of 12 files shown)


pkg/codegen/hcl2/model/type_test.go
```diff
@@ -668,20 +668,32 @@ func TestUnifyType(t *testing.T) {
 }
 
 func TestRecursiveObjectType(t *testing.T) {
-	props := map[string]Type{
-		"data": NewOutputType(IntType),
+	makeType := func(prop string) Type {
```
viveklak, line 671: Nit - rename to `makeLinkedListType`?

pkg/codegen/hcl2/model/type_object.go
```diff
@@ -218,20 +227,35 @@ func (t *ObjectType) conversionFrom(src Type, unifying bool) ConversionKind {
...
+	if t.s != "" {
+		return t.s
+	}
 
-		annotations := ""
-		if len(t.Annotations) != 0 {
-			annotations = fmt.Sprintf(", annotated(%p)", t)
+	if seen != nil {
+		if _, ok := seen[t]; ok {
+			return "..."
```
viveklak, line 240: Can it be the case that t.s is not set but the object type exists in `seen`? Is that a panic worthy event? What does `...` actually mean in hcl2?

pkg/codegen/hcl2/model/type_test.go
viveklak: Should we consider adding some more tests for unions etc.?

viveklak (commented):
Generally looks good apart from a couple of things pointed out in comments.

pkg/codegen/hcl2/model/type_object.go
mikhailshilkov: > Can it be the case that t.s is not set but the object type exists in seen?

Yes - that's when the same type is the "child" of itself - as seen in the test https://github.com/pulumi/pulumi/pull/6250/files#diff-eff2ea294a07fa6b0e4ca9b4e2766d60c1e923cad0771d3deb9514d9b3a4036dR693

> What does ... actually mean in hcl2?

It means nothing in HCL. We aren't producing HCL here, we are producing a text representation of a type.

pkg/codegen/hcl2/model/type_test.go
mikhailshilkov: We could... I mostly extended the existing set of recursive tests to the new function. I don't think it makes the coverage lower but it's true it doesn't improve it either.

viveklak (approved):

Issue #4029, opened 2020-03-06
skip signed url creation


geekflyer:
This request may be somewhat specific to using GCS as state backend.

Currently when running any pulumi command locally with normal user / gcloud credentials, every operation starts and ends with this message:
```
warning: Pulumi will not be able to print a statefile permalink using these credentials. Neither a GoogleAccessID or PrivateKey are available. Try using a GCP Service Account.
```

This is an inherent limitation in GCS due to GCP only being able to create signed urls using a service account, but not using personal credentials.
However using a service account locally would violate a bunch of best practices / security practices and also would is rather inconvenient setup for a team.
In other words while this warning is informative, there's imo no reasonable way to fix this from a users perspective / it's not very actionable.

On the other hand to get this warning over and over again is annoying and noisy.
Most of my team members actually first think this is an error.

**Therefore i'd like to request the option to skip signed url creation altogether or be able to suppress this warning**.

This could be for example an argument like `pulumi up --skip-signed-url` or environment variable like `PULUMI_BACKEND_DISABLE_SIGNED_URL=1`.
Another, perhaps preferable alternative variant would be to introduce an option to specify the link type / format. In case of GCS I could imagine these 2 variants:
1. `signedURL` (the default)
2. `objectURI` , e.g.: `gs://my-state-bucket/.pulumi/stacks/prod.json`
3. `off`

A user than may invoke the CLI with `pulumi up --state-link-format objectURI` and would get `gs://my-state-bucket/.pulumi/stacks/prod.json` instead.

Hope this could be considered as it would solve some of the UX issues when using blob storage as backend. Thanks!

HagaiBarel:
I second this. Being hitting this recently and it's really annoying / confusing

stack72:
Hi @geekflyer / @HagaiBarel 

I am sorry for the lack of response on this - a while back there was a new cmd flag added

```
pulumi up --suppress-permalink
```

This will suppress the output that is causing the issue here. This is supported on up / refresh / destroy / preview and import

I hope this helps

Paul

HagaiBarel:
Hi Paul,

Thanks for the reply. Yes, using `--suppress-permalink` does suppresses the output of the permalink which resolves one of errors, however the above warning also happens at the start of the operation not only at the end. It's also printed with other actions, e.g. `pulumi whoami`, `pulumi stack ls` etc.

stack72:
Hey @geekflyer / @HagaiBarel 

[...]

Issue #3537, opened 2019-11-20
Pulumi up with Azure Storage SAS token returns SignedURL error


ndonze:
Repro steps:

1. Create a new Pulumi project. For example:

```
> mkdir ReproAzureSasTokenSignedUrlError
> cd ReproAzureSasTokenSignedUrlError
> pulumi new -g azure-typescript
> npm i

Set a location on the storage account in index.ts
```

2. Set up a blob container in Azure Storage for Pulumi stack state
2a. Create an Azure Storage account
2b. Within your Azure Storage account, create a blob container with private access
2c. Within your blob container, add a new Access policy (Stored access policies) with:
--- All Permissions selected
--- An Expiry time well in the future
2d. Within your Azure Storage account, navigate to Storage Explorer (preview) > BLOB CONTAINERS
2e. Right-click on [your blob container]
2f. Get Shared Access Signature
2g. Select your Access policy
2h. Create
2i. Copy the query string value in full (including the leading question mark)

3. Log into Pulumi with stack state in Azure Storage
3a. Set env var AZURE_STORAGE_ACCOUNT to your Azure Storage account name
3b. Set env var AZURE_STORAGE_SAS_TOKEN to the query string copied when generating your Shared Access Signature
3c. `pulumi login azblob://[your blob container name]`

4. Initialize stack and provision
4a. `pulumi stack init repro`
4b. `pulumi up`

Expected: Provisioning completes and updates stack state in Azure Storage without error

Actual: Provisioning completes and stack state appears to be updated, but the following error is returned:

> error: Could not get signed url for stack location: blob (code=Unknown): to use SignedURL, you must call OpenBucket with a non-nil Options.Credential

WillooWisp:
Same thing with preview:

warning: Could not get signed url for stack location: blob (key ".pulumi/stacks/mytestdev.json") (code=Unknown): azureblob: to use SignedURL, you must call OpenBucket with a non-nil Options.Credential

stack72:
Linked to #4029

stack72:
Hi @WillooWisp / @ndonze 

I just updated the linked issue to say we will be suppressing permalinks (by default) on non-SaaS backends so this error will no longer be surfaced

I am going to be starting work on that now so it will ship in Pulumi v2.21.0

Paul

PR #6251, opened 2021-02-04
[cli] Disable permalinks to the update details page when using self-managed backends (S3, Azure, GCS)

stack72:
Fixes: #4029
Fixes: #3537

Should the user want to get permalinks when using a self-managed backend, they can pass a flag:

```
$ pulumi up --suppress-permalink false
```

Permalinks for these self-managed backends will be suppressed on `update`, `preview`, `destroy`,
`import` and `refresh` operations.

A view of the functionality:

### self-managed backend

```
$ pulumi whoami -v
User: stack72
Backend URL: file://~


$ PULUMI_CONFIG_PASSPHRASE=password pulumi up
Previewing update (pphrase-2):
     pulumi:pulumi:Stack  testing-pphrase-2
Resources:
    2 unchanged


$ PULUMI_CONFIG_PASSPHRASE=password pulumi up --suppress-permalink false
Previewing update (pphrase-2):
     pulumi:pulumi:Stack  testing-pphrase-2
Resources:
    2 unchanged

Permalink: file:///Users/stack72/.pulumi/stacks/pphrase-2.json
```

### Using the SaaS

```
$ pulumi whoami -v
User: stack72
Backend URL: https://app.pulumi.com/stack72


$ pulumi up
Previewing update (dev4)
View Live: https://app.pulumi.com/stack72/testing/dev4/previews/11d3539a-93be-4272-8070-190875b8697f
     Type                          Name                  Plan
 +   pulumi:pulumi:Stack           testing-dev4          create
 +   └─ random:index:RandomString  test                  create
Resources:
    + 2 to create


$ pulumi up --suppress-permalink false
Previewing update (dev4)
View Live: https://app.pulumi.com/stack72/testing/dev4/previews/416a8023-ddcd-4ed1-8d48-5744831b836d
     Type                          Name                  Plan
 +   pulumi:pulumi:Stack           testing-dev4          create
 +   └─ random:index:RandomString  test                  create
Resources:
    + 2 to create


$ pulumi up --suppress-permalink true
Previewing update (dev4)
     Type                          Name                  Plan
 +   pulumi:pulumi:Stack           testing-dev4          create
 +   └─ random:index:RandomString  test                  create
Resources:
    + 2 to create
```

### DIFF
--- a/pkg/backend/filestate/backend.go
+++ b/pkg/backend/filestate/backend.go
@@ -599,18 +599,25 @@ func (b *localBackend) apply(
 		} else {
 			link, err = b.bucket.SignedURL(context.TODO(), b.stackPath(stackName), nil)
 			if err != nil {
+				// set link to be empty to when there is an error to hide use of Permalinks
+				link = ""
+
 				// we log a warning here rather then returning an error to avoid exiting
 				// pulumi with an error code.
 				// printing a statefile perma link happens after all the providers have finished
 				// deploying the infrastructure, failing the pulumi update because there was a
 				// problem printing a statefile perma link can be missleading in automated CI environments.
-				cmdutil.Diag().Warningf(diag.Message("", "Could not get signed url for stack location: %v"), err)
+				cmdutil.Diag().Warningf(diag.Message("", "Unable to create signed url for current backend to "+
+					"create a Permalink. Please visit https://www.pulumi.com/docs/troubleshooting/ "+
+					"for more information\n"))
 			}
 		}
 
-		fmt.Printf(op.Opts.Display.Color.Colorize(
-			colors.SpecHeadline+"Permalink: "+
-				colors.Underline+colors.BrightBlue+"%s"+colors.Reset+"\n"), link)
+		if link != "" {
+			fmt.Printf(op.Opts.Display.Color.Colorize(
+				colors.SpecHeadline+"Permalink: "+
+					colors.Underline+colors.BrightBlue+"%s"+colors.Reset+"\n"), link)
+		}
 	}
 
 	return changes, nil
--- a/pkg/cmd/pulumi/destroy.go
+++ b/pkg/cmd/pulumi/destroy.go
@@ -46,7 +46,7 @@ func newDestroyCmd() *cobra.Command {
 	var showSames bool
 	var skipPreview bool
 	var suppressOutputs bool
-	var suppressPermaLink bool
+	var suppressPermaLink string
 	var yes bool
 	var targets *[]string
 	var targetDependents bool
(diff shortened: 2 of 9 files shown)


infin8x (approved):
Approved from a "does what I expected it to do" perspective. Full code review left to others more qualified :)

pkg/cmd/pulumi/destroy.go
```diff
@@ -203,6 +214,9 @@ func newDestroyCmd() *cobra.Command {
 	cmd.PersistentFlags().BoolVar(
 		&suppressPermaLink, "suppress-permalink", false,
 		"Suppress display of the state permalink")
+	cmd.PersistentFlags().BoolVar(
+		&optInNonSaasPermaLink, "non-saas-permalink-enabled", false,
```
lukehoban, line 218: Two things:
1. We don't use the team "non-saas" anywhere else.  We use "self-managed" in the backends docs, and "filestate" in the implementation.  I'm not sure exactly what's right here - but we should likely base around one of these names instead of introducing "non-saas"
2. We already have `suppress-permalink`.  It feels like this new option should be `suppress-permalink=false`.  We are basically making the defalult for `suppress-permalink` true for the filestate backend with these changes. Not sure exactly how best to represent that with Cobra.

stack72: so `suppress-permalink` still works - as we allow suppressing even for hosted backend. I can't make it work with both. The default for this is `false`. Changing this to true gives us chicken and egg based on the hosted backend. I can't figure out a way of being able to use the existing flag without breaking those who use the hosted backend so I was trying to create an explicit flag to make it opt in for self-managed backends

(FWIW, I like selfManagedBackends rather than non-SaaS so thanks for that(

infin8x: Ah, the naming is a good point. We're starting to circle around a final set of names for the various backends here: https://docs.google.com/document/d/10cLvyua-xJzVEuZjY0InfLPYeF2RQtG_BLdqUun35D4/edit#heading=h.q32ufnmp2er2

Lemme see if I can draw us to a close there so we can start using those new names.

infin8x: Let's stick with `selfManagedBackends` for now - we should have a new name sorted here in the next few weeks, but I don't want to block this PR in the meantime.

komalali (approved):
LGTM but curious what happens if you just pass in `pulumi up --suppress-permalink`

lukehoban (approved):

pkg/backend/filestate/backend.go
```diff
@@ -599,18 +599,25 @@ func (b *localBackend) apply(
...
+
 				// we log a warning here rather then returning an error to avoid exiting
 				// pulumi with an error code.
 				// printing a statefile perma link happens after all the providers have finished
 				// deploying the infrastructure, failing the pulumi update because there was a
 				// problem printing a statefile perma link can be missleading in automated CI environments.
-				cmdutil.Diag().Warningf(diag.Message("", "Could not get signed url for stack location: %v"), err)
+				cmdutil.Diag().Warningf(diag.Message("", "Unable to create signed url for current backend to "+
+					"create a Permalink. Please visit https://www.pulumi.com/docs/troubleshooting/ "+
+					"for more information\n"))
```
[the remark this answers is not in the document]
stack72: There are 2 places the docs will be updated

The troubleshooting section will have details as to why they get an open bucket url error - it's usually regarding using an Azure and GCp service principal to log into the self managed backends

I will also be updating the state and backends page to talk about the new defaults here of not showing a permalink

Both will be ready early Monday morning PST to make sure they are ready to go

fcavaco:
when trying this in WLS2 , the syntax is slightly different from above if you actually want to add the argument:
--suppress-permalink=false , rather than --suppress-permalink false.

Issue #6259, opened 2021-02-05
[codegen/*] - Generate accessors for provider resource properties


komalali:
Provider configuration was not originally intended to be reflected into outputs. Somewhere along the way, we intentionally decided that provider configuration *should* be reflected into outputs and made changes to the engine to accommodate functionality. However, the code generators were never updated to account for this. The work here is twofold:
* The properties are currently being returned as serialized JSON. We should see if we can refactor this s.t. the Go -> gRPC bridge is responsible for serialization and deserialization.
* Update the Provider resource in the schema to reflect the input properties as output properties.

PR #6260, opened 2021-02-05
[codegen/python] - Generate accessors for provider resource properties

komalali:
@justinvp I'm not sure if something different needs to happen here since these are provider properties. Should the resource also have a `get` method? Is there anything that needs to happen with respect to the camelCase <-> snake_case conversion in the getter functions?

Fixes: https://github.com/pulumi/pulumi/issues/6259

### DIFF
--- a/pkg/codegen/python/gen.go
+++ b/pkg/codegen/python/gen.go
@@ -1117,6 +1117,14 @@ func (mod *modContext) genResource(res *schema.Resource) (string, error) {
 		return fmt.Sprintf("pulumi.Output[%s]", ty)
 	})
 
+	// Write out Python property getters for all inputs if this is a provider resource because all inputs are implicitly outputs.
+	if res.IsProvider {
+		mod.genProperties(w, res.InputProperties, false /*setters*/, func(prop *schema.Property) string {
+			ty := mod.typeString(prop.Type, false /*input*/, false /*wrapInput*/, !prop.IsRequired, false /*acceptMapping*/)
+			return fmt.Sprintf("pulumi.Output[%s]", ty)
+		})
+	}
+
 	// Override translate_{input|output}_property on each resource to translate between snake case and
 	// camel case when interacting with tfbridge.
 	fmt.Fprintf(w,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ CHANGELOG
 - [sdk/python] Gracefully handle monitor shutdown in the python runtime without exiting the process. 
   [#6249](https://github.com/pulumi/pulumi/pull/6249)
 
+- [codegen/python] Generate accessors for provider resource properties.
+  [#6249](https://github.com/pulumi/pulumi/pull/6260)
+
 ## 2.20.0 (2021-02-03)
 
 - [sdk/python] Fix `Output.from_input` to unwrap nested output values in input types (args classes), which addresses


pkg/codegen/python/gen.go
```diff
@@ -1117,6 +1117,14 @@ func (mod *modContext) genResource(res *schema.Resource) (string, error) {
...
 	})
 
+	// Write out Python property getters for all inputs if this is a provider resource because all inputs are implicitly outputs.
+	if res.IsProvider {
+		mod.genProperties(w, res.InputProperties, false /*setters*/, func(prop *schema.Property) string {
+			ty := mod.typeString(prop.Type, false /*input*/, false /*wrapInput*/, !prop.IsRequired, false /*acceptMapping*/)
+			return fmt.Sprintf("pulumi.Output[%s]", ty)
+		})
+	}
+
```
justinvp, line 1127: I'm actually not sure we want to handle this here. It doesn't look like we generate equivalent getters for the Provider resource in any of the other languages. If we do want to add these, I wonder if this should be something that gets defined in the schema, basically mirroring the InputProperties in Properties. Then this would likely be handled automatically for all the languages, without having to add this special case.

Though:

1. Are we generating nested output types for these? (Looks like we are for Python and Node.js, but I haven't checked the other languages)
2. Provider inputs are serialized as JSON. Are the outputs coming back as JSON that would need to be deserialized?

@pgavlin, any thoughts on this?

komalali: It's worth explaining the motivation for this change because I'm really not sure this is the right solution. What we're missing in python right now is auto-complete in the IDE for the properties. You can access, for example, the region in aws with `provider.region`. It works but it doesn't auto-complete in the IDE. What are we missing?

justinvp: Understood. I'm just wondering if we didn't originally intend for these to be output properties, and that's why we don't expose these as proper outputs (with getters, doc strings, etc.) in any of the languages. I'm wondering if the reason the values are getting set on the provider resource for Python is just an accident of how inputs are propagated to outputs and how those values will always be set as fields on the resource instance. Are these actual output values coming back from the engine (even during an update, not a preview), or are they the [inputs being returned as outputs](https://github.com/pulumi/pulumi/blob/3a3b96de72337517ec5ce7664c2d2d336ac42d42/sdk/python/lib/pulumi/runtime/rpc.py#L584-L586) from the Python SDK? If coming back from the engine during an update, are they coming back as JSON strings?

If we do want to expose these as proper outputs, wondering if we'd do it more holistically for all languages (likely by making these output properties for the Provider resource in the schema), but my only question then is whether the output values are being returned as serialized JSON. Would we deserialize the JSON in the engine before being returned as outputs? Or return the values as JSON that would then need to be deserialized before being returned (which would involve updating the codegen of the Provider resource for each language, ugh)?

komalali: So there are a couple of things that make me think that these are intended to be exposed as outputs.

1. [This statement](https://www.pulumi.com/docs/reference/pkg/aws/provider/#outputs:~:text=All%20input%20properties%20are%20implicitly%20available%20as%20output%20properties.) in the resource docs that says `All input properties are implicitly available as output properties.`
2. We use them in [awsx](https://github.com/pulumi/pulumi-awsx/blob/master/nodejs/awsx/utils.ts#L137-L140) for nodejs, so it's not just a python thing.

But yeah, it would be good to get some input from @pgavlin about what the intended functionality is here.

pgavlin: > I'm just wondering if we didn't originally intend for these to be output properties

This is correct: we did not originally intend for provider configuration to be reflected into outputs. However, we did make an intentional choice to do so in the engine--I think we simply never updated the code generators to account for this.

pgavlin: > If we do want to expose these as proper outputs, wondering if we'd do it more holistically for all languages (likely by making these output properties for the Provider resource in the schema), but my only question then is whether the output values are being returned as serialized JSON

They are almost certainly being returned as serialized JSON. We should see if we can refactor this s.t. the Go -> gRPC bridge is responsible for serialization and deserialization.

komalali: Thank you! I've updated the issue https://github.com/pulumi/pulumi/issues/6259 with details discussed here and will close this PR until we can address this fully.

Issue #6000, opened 2020-12-22
contains_unknowns fails if val contains a "values" key


clstokes:
The `contains_unknowns` function fails if the input `val` contains a `"values"` key.

https://github.com/pulumi/pulumi/blob/7e3d69a53ad34134f57d9bb3ad62d4e5d95f5723/sdk/python/lib/pulumi/runtime/rpc.py#L562-L575

Changing the return statement to the code below _seems_ to fix this:

```py
return impl(val.values(), stack)
```

komalali:
I can't seem to repro this, are there any more details on what the failure looks like or a case where it fails?

PR #6264, opened 2021-02-05
[python/sdk] - Correctly handle outputs with properties named "values"

komalali:
Fixes: https://github.com/pulumi/pulumi/issues/6000

The test case errors before the change and succeeds after.

### DIFF
--- a/sdk/python/lib/test/test_next_serialize.py
+++ b/sdk/python/lib/test/test_next_serialize.py
@@ -36,17 +36,21 @@ class FakeCustomResource(CustomResource):
         self.__dict__["urn"] = Output.from_input(urn)
         self.__dict__["id"] = Output.from_input("id")
 
+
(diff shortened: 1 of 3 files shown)


sdk/python/lib/test/test_next_serialize.py
```diff
@@ -72,8 +77,24 @@ def new_resource(self, typ, name, inputs, provider, id):
...
+
+@pulumi.output_type
+class MyOutputTypeDict(dict):
+
+    def __init__(self, values: list):
+        pulumi.set(self, "values", values)
+
+    # Property with empty body.
+    @property
+    @pulumi.getter(name="firstValue")
```
justinvp, line 89: Nit: It doesn't matter at all here, but this should really be:

```suggestion
    @pulumi.getter
```

justinvp (approved):
LGTM

Issue #6265, opened 2021-02-05
[programgen/dotnet] Template expressions should be wrapped in Output.Format


viveklak:
Currently template expressions default to using string templates but that doesn't work with output types.
See example here: https://github.com/pulumi/arm2pulumi/issues/20

PR #6266, opened 2021-02-05
Fix template expressions for dotnet

viveklak:
Fix for #6265

### DIFF
--- a/pkg/codegen/dotnet/gen_program_expressions.go
+++ b/pkg/codegen/dotnet/gen_program_expressions.go
@@ -519,16 +519,21 @@ func (g *generator) GenSplatExpression(w io.Writer, expr *model.SplatExpression)
 func (g *generator) GenTemplateExpression(w io.Writer, expr *model.TemplateExpression) {
 	multiLine := false
 	expressions := false
+	hasOutput := false
 	for _, expr := range expr.Parts {
 		if lit, ok := expr.(*model.LiteralValueExpression); ok && lit.Type() == model.StringType {
 			if strings.Contains(lit.Value.AsString(), "\n") {
 				multiLine = true
 			}
 		} else {
+			hasOutput = true
 			expressions = true
 		}
 	}
 
+	if hasOutput {
+		g.Fgen(w, "Output.Format(")
+	}
 	if multiLine {
 		g.Fgen(w, "@")
 	}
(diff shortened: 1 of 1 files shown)


pkg/codegen/dotnet/gen_program_expressions.go
```diff
@@ -519,16 +519,21 @@ func (g *generator) GenSplatExpression(w io.Writer, expr *model.SplatExpression)
...
 	multiLine := false
 	expressions := false
+	hasOutput := false
 	for _, expr := range expr.Parts {
 		if lit, ok := expr.(*model.LiteralValueExpression); ok && lit.Type() == model.StringType {
 			if strings.Contains(lit.Value.AsString(), "\n") {
 				multiLine = true
 			}
 		} else {
+			hasOutput = true
```
viveklak, line 529: Not sure if we need to be more selective around when to wrap in outputs here?

mikhailshilkov: There is a test failure - the new code is `Source = new FileAsset(Output.Format($"{siteDir}/{range.Value}")),` which won't work. It needs to be something like `Source = Output.Format($"{siteDir}/{range.Value}").Apply(v => (AssetOrArchive)new FileAsset(v)),`.

I guess we need to change the return type of this expression to an Output, not just add the `Output.Format` mechanically.

However, your code may be "less" broken - so maybe it's okay as the first step? How do other SDKs solve that?

mikhailshilkov: On the second look, those tests aren't outputs, so they shouldn't actually change - they seem fine as they were before the change. So yes, I think we should also be more selective.

viveklak: Yes I was looking at the test failure as well. Will try to modify the expression as you suggested and am making the `Output.Format` more selective.

pgavlin: In other code generators we drive this off of pattern-matching on applies. That's probably what we should do here as well. @viveklak do you want to take that on? If not, I can take it.

viveklak: @pgavlin yes sorry I have obviously lost track of this PR :(. I'd appreciate it if you could take over!

PR #6268, opened 2021-02-07
[sdk/go] Support multiple folders in GOPATH

### LINKED ISSUE #6223: Cant find go binary when GOPATH has multiple folders
According to https://golang.org/cmd/go/#hdr-GOPATH_environment_variable, GOPATH can be a list of folders like the PATH environment variable.

In the code seems that you're not taking that into account.
https://github.com/pulumi/pulumi/blob/d3187d630f9555d7fae2c76ce33caa18b6de00b4/sdk/go/common/util/executable/executable.go#L35-L38

Quick test:
image

ynden:
FindExecutable function now supports multiple paths in the GOPATH environment variable. 

Fixes #6223


sdk/go/common/util/executable/executable.go
```diff
@@ -55,3 +64,15 @@ func FindExecutable(program string) (string, error) {
 
 	return "", errors.Errorf(unableToFindProgramTemplate, program)
 }
+
+func getPotentialPaths(goPath string, os string) []string {
```
ynden, line 68: I believe this function could return a (err, string[]), because what happens if the os is not linux/darwin or windows? But at the same time, I checked in the README, and windows and linux/darwin are the only supported platforms.

sdk/go/common/util/executable/executable.go
```diff
@@ -32,18 +32,27 @@ func FindExecutable(program string) (string, error) {
...
+		// getPotentialPaths will return paths where to look.
+		// Because the GOPATH can take the form of multiple paths (e.g: GOPATH="home/user/go:/usr/local/go")
+		// we need to split the GOPATH, and look into each of the paths.
+		// If the GOPATH hold only one path, there will only be one element in the slice.
+		potentialPaths := getPotentialPaths(goPath, runtime.GOOS)
+		for i, pp := range potentialPaths {
+			goPathProgram := filepath.Join(pp, "bin", program)
+			fileInfo, err := os.Stat(goPathProgram)
+
+			if err != nil && i+1 == len(potentialPaths) {
```
ynden, line 46: This code does the job, but is there a more elegant way to do this check?

sdk/go/common/util/executable/executable.go
```diff
@@ -32,18 +32,27 @@ func FindExecutable(program string) (string, error) {
...
-		goPathProgram := filepath.Join(goPath, "bin", program)
-		fileInfo, err := os.Stat(goPathProgram)
-		if err != nil {
-			if !os.IsNotExist(err) {
-				return "", errors.Wrapf(err, "unable to find program in %q", goPathProgram)
+		// getPotentialPaths will return paths where to look.
+		// Because the GOPATH can take the form of multiple paths (e.g: GOPATH="home/user/go:/usr/local/go")
+		// we need to split the GOPATH, and look into each of the paths.
+		// If the GOPATH hold only one path, there will only be one element in the slice.
+		potentialPaths := getPotentialPaths(goPath, runtime.GOOS)
```
ynden, line 41: Here instead of having a function that takes no parameters, I preferred to add these two, so it's more testable.

sdk/go/common/util/executable/executable.go
stack72, line 68: I think this should be called `splitGoPath` - it would make it more clear to me about what it does

sdk/go/common/util/executable/executable.go
```diff
@@ -32,18 +32,27 @@ func FindExecutable(program string) (string, error) {
...
 
-	// look in $GOPATH/bin
+	// look in potentials $GOPATH/bin
 	if goPath := os.Getenv("GOPATH"); len(goPath) > 0 {
-		goPathProgram := filepath.Join(goPath, "bin", program)
-		fileInfo, err := os.Stat(goPathProgram)
-		if err != nil {
-			if !os.IsNotExist(err) {
-				return "", errors.Wrapf(err, "unable to find program in %q", goPathProgram)
+		// getPotentialPaths will return paths where to look.
```
stack72, line 37: "will return a list of paths in which to look for the binary"

sdk/go/common/util/executable/executable.go
```diff
@@ -32,18 +32,27 @@ func FindExecutable(program string) (string, error) {
...
+		// Because the GOPATH can take the form of multiple paths (e.g: GOPATH="home/user/go:/usr/local/go")
```
stack72, line 38: Would it just be worth linking to the docs on GOPATH having multiple paths?

sdk/go/common/util/executable/executable.go
stack72: How does something like:

```
if goPath := os.Getenv("GOPATH"); len(goPath) > 0 {
  // splitGoPath will return paths where to look.
  // Because the GOPATH can take the form of multiple paths (e.g: GOPATH="home/user/go:/usr/local/go")
  // we need to split the GOPATH, and look into each of the paths.
  // If the GOPATH hold only one path, there will only be one element in the slice.
  goPathParts := splitGoPath(goPath, runtime.GOOS)
  for i, pp := range goPathParts {
    goPathProgram := filepath.Join(pp, "bin", program)
    if fileInfo, err := os.Stat(filename); !os.IsNotExist(err) {
      return "", err
    }

    if fileInfo != nil && !fileInfo.Mode().IsDir() {
      logging.V(5).Infof("program %s found in %s/bin", program, pp)
      return goPathProgram, nil
    }
  }

  return "", errors.Wrapf(err, "unable to find program in path: %s", goPath)
}
```

feel to you?

stack72 (commented):
Hi @ynden 

Thanks so much for this work! this is great to have in - I've left a comment on a dfferent way of handling it that I'd love your thoughts on?

sdk/go/common/util/executable/executable.go
ynden: I wrote it like this the first time. Let's say GOPATH contains three paths. 
The drawback of doing this: 
```if fileInfo, err := os.Stat(filename); !os.IsNotExist(err)``` is that it will return immediately for the first path that has an error.

I think what we want to do here, is give each path a chance, and return an error (if there is one) only for the last path.

sdk/go/common/util/executable/executable.go
ynden: Yes I agree

sdk/go/common/util/executable/executable.go
stack72: Hi @ynden

So the func would only return if the error is NOT IsNotExist - that would mean it is a legit error that we would want to return right?

ynden: Right! I thought that, even if there was a more serious error (NOT isNotExist), we could try the other paths as well. 
But now I understand, I'll change it, thank you for your guidance!

stack72: 👌

stack72:
Hi @ynden 

I just wanted to let you know that I have not forgotten about this issue - I plan to pull it into main line after we release v2.22.0 tomorrow

Sorry I went quiet here

Paul

ynden:
Hey @stack72,
Thank you for the feedback!

stack72:
Hey @ynden 

So I pulled this PR and rebased it and sorted the changelog out to the new entry style - I pushed the code as #6506 - notice it's squashed for cleaner commit history and you still keep your attribution!

I am just going to let the tests pass and then we can merge it

Thanks again so much and sorry this took so long

Paul

Issue #6003, opened 2020-12-23
Create Output.all alternative as an object / dict


colearendt:
`Output.all()` (in Python) creates a list. It would be great have some variant of `Output.all()` that creates an object or dict so keys / objects are easier to reference by name!

Slack thread: https://pulumi-community.slack.com/archives/C84L4E3N1/p1608681663484600

## Expected Behavior
```
# something like
Output.all(name=output1, name2=output2).apply(lambda x: x['name'])
```

## Current Behavior
```
# instead of
Output.all(output1, output2).apply(lambda x: x[0])
```

## Steps to Reproduce
N/A

## Context (Environment)

I have helper functions that do things like **create_a_vpc**. There are many outputs here, so it is nice returning `Output.all(output1, output2, output3)`. However, this means utilizing those resources later need to be referenced by index rather than by name.

I'm not sure whether IDEs would be able to pick up the keys for the `Output` type of objects, but if so, autocomplete would be amazing here as well!!

nbyl:
This would greatly help us. We have a pretty neat use case for that. We have an application that needs a yaml config file that must be written to a Kubernetes secret. The yaml structure contains values that are based on outputs from other resources (database, buckets, ...). This get pretty confusing if you have more than just a handful parameters.

komalali:
This is included in [v2.21.0](https://pypi.org/project/pulumi/) of the SDK.

It is now possible to do the following:
```python
Output.all(name=output1, name2=output2).apply(lambda x: x['name'])
```

PR #6269, opened 2021-02-07
[sdk/python] - Allow kwargs in Output.all() to create a dict

komalali:
Fixes: #6003

### DIFF
--- a/sdk/python/lib/pulumi/output.py
+++ b/sdk/python/lib/pulumi/output.py
@@ -331,10 +331,10 @@ class Output(Generic[T]):
         Output which can then be used as the target of `apply`. Resource dependencies
         are preserved in the returned Output.
 
-        Examples:
+        Examples::
 
-        Output.all(foo, bar) -> Output[[foo, bar]]
-        Output.all(foo=foo, bar=bar) -> Output[{"foo": foo, "bar": bar}]
+            Output.all(foo, bar) -> Output[[foo, bar]]
+            Output.all(foo=foo, bar=bar) -> Output[{"foo": foo, "bar": bar}]
 
         :param Input[T] args: A list of Inputs to convert.
         :param Input[T] kwargs: A list of named Inputs to convert.
--- a/sdk/python/lib/pulumi/output.py
+++ b/sdk/python/lib/pulumi/output.py
@@ -27,7 +27,8 @@ from typing import (
     List,
     Dict,
     Optional,
-    TYPE_CHECKING
+    TYPE_CHECKING,
+    overload
 )
 
 from . import _types
@@ -263,7 +264,7 @@ class Output(Generic[T]):
             # Since Output.all works on lists early, serialize this dictionary into a list of lists first.
             # Once we have a output of the list of properties, we can use an apply to re-hydrate it back into a dict.
             dict_items = [[k, Output.from_input(v)] for k, v in val.items()]
-            # type checker doesn't like returing a Dict in the apply callback
+            # type checker doesn't like returning a Dict in the apply callback
             fn = cast(Callable[[List[Any]], T], lambda props: {k: v for k, v in props}) # pylint: disable=unnecessary-comprehension
             return Output.all(*dict_items).apply(fn, True)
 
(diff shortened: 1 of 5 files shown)


sdk/python/lib/test/test_next_serialize.py
```diff
@@ -295,10 +295,33 @@ async def test_output_all(self):
...
+        known_fut = asyncio.Future()
+        known_fut.set_result(True)
+        out = Output({res}, fut, known_fut)
+
+        other = Output.from_input(99)
+        combined = Output.all(out, other=other)
+        deps = []
+        prop = await rpc.serialize_property(combined, deps)
+        self.assertSetEqual(set(deps), {res})
+        self.assertEqual([42], prop)
```
komalali, line 324: I've written it this way because this is how the current `Output.all()` function behaves. However, I would prefer that we error in the case of mixed inputs and clarify that either args or kwargs should be used but not both. That would technically be a breaking change though.

sdk/python/lib/pulumi/output.py
```diff
@@ -321,49 +322,71 @@ def secret(val: Input[T]) -> 'Output[T]':
         return Output(o._resources, o._future, o._is_known, is_secret)
 
     @staticmethod
-    def all(*args: Input[T]) -> 'Output[List[T]]':
+    def all(*args: Input[T], **kwargs: Input[T]) -> 'Union[Output[List[T]], Output[Dict[T]]]':
```
justinvp, line 325: I was thinking we'd use [`@typing.overload`](https://docs.python.org/3/library/typing.html#typing.overload) for this, something like:

```python
@overload
@staticmethod
def all(*args: Input[T]) -> 'Output[List[T]]':
    ...

@overload
@staticmethod
def all(**kwargs: Input[T]) -> 'Output[Dict[T]]':
    ...

@staticmethod
def all(*args: Input[T], **kwargs: Input[T]):
    # implementation
```

komalali: TIL! Thanks, will fix up.

sdk/python/lib/pulumi/output.py
```diff
@@ -320,50 +322,83 @@ def secret(val: Input[T]) -> 'Output[T]':
...
+            return await _gather_from_dict(value_futures_dict)
         from_input = cast(Callable[[Union[T, Awaitable[T], Output[T]]], Output[T]], Output.from_input)
+
         # First, map all inputs to outputs using `from_input`.
-        all_outputs = list(map(from_input, args))
+        if len(args) > 0:
+            all_outputs: Union[list, dict] = list(map(from_input, args))
+        elif len(kwargs) > 0:
+            all_outputs = {k: from_input(v) for k, v in kwargs.items()}
+        else:
```
lukehoban, line 398: Should this also error if both args and kwargs are passed?

sdk/python/lib/pulumi/output.py
```diff
@@ -320,50 +322,83 @@ def secret(val: Input[T]) -> 'Output[T]':
...
+            is_secret_futures = [o._is_secret for o in outputs]
             each_is_secret = await asyncio.gather(*is_secret_futures)
             return any(each_is_secret)
 
-        async def get_resources(outputs):
-            resources_futures = list(map(lambda o: o._resources, outputs))
+        async def get_resources(outputs: Union[dict, list]):
+            if isinstance(outputs, dict):
+                outputs = [outputs[k] for k in outputs]
+            resources_futures = [o._resources for o in outputs]
```
lukehoban, line 379: Kind of a nit - but it does feel like we could avoid forking the implementation of these helpers, and instead pass a list into all of these.  Their behaviour doesn't depend on the dict structure at all - so just pre-computing the list of values once and then calling the existing functions on those would lead to simpler logic throughout I think?

sdk/python/lib/test/test_next_serialize.py
lukehoban: > However, I would prefer that we error in the case of mixed inputs and clarify that either args or kwargs should be used but not both. That would technically be a breaking change though.

I'm inclined to take this very minor breaking change as part of this change.  It's not clear that code that was taking advantage of this before could possibly have been doing what the user expected it to do.  (And more likely, there is no code out there taking advantage of this).

lukehoban (approved):

justinvp (approved):
LGTM

PR #6270, opened 2021-02-07
Remove language SDK blocks from resource templates

cnunciato:
These sections link to the language-specific provider docs, which we removed with https://github.com/pulumi/docs/pull/5133 and are currently hiding with CSS. This PR removes them from the resource-doc generation process.


stack72 (approved):

praneetloke:
The changes are fine but is there a separate issue to also remove generating those links in the generator code? This PR basically switches off rendering them.

praneetloke:
Looking into this more and I think this just removes the table at the end of the module doc page that lists links to the various SDKs. I see that the types in the construction of a new resource are redirected using the redirect rules in the docs. But I can see 3 issues:

* Resource construction shows a link to the actual resource being constructed (`Bucket` in this case`) and the args property (`BucketArgs` in this case) which all lead to the module page when clicked. Moreover, below the snippet to show how to create the resource, the table also shows a link to the `BucketArgs` that also goes nowhere. I think we should just point those to the `Inputs` section right on the same page since that is what the `BucketArgs` is (and input args in any other resource will be.) The link to `CustomResourceOptions` is fine, though.
image

* Clicking on the "state" type in the "Looking up an existing resource" leads to the module page. `BucketState` in the below screenshot. Maybe we should just not link that anymore since the type's properties are literally just a few lines below.
image

* Remove the `input` and `output` type links from each supporting type. Basically, removing the lines: https://github.com/pulumi/pulumi/blob/master/pkg/codegen/docs/templates/resource.tmpl#L166-L182

image

cnunciato:
Thanks @praneetloke -- I'll follow up with these changes in subsequent PR today.

PR #6277, opened 2021-02-08
Move linting check into command-dispatch job

stack72:
This fixes not picking up linting issues when sending a PR as
the PR job uses `pull_request_target` which in turn only checks out
from master which is a security measure


komalali (approved):

komalali:
This is obviously not going to pass CI so we should just merge it @stack72

stack72:
> This is obviously not going to pass CI so we should just merge it @stack72

agreed!

PR #6279, opened 2021-02-08
Fix up resource-docs links

cnunciato:


praneetloke:
@cnunciato alright I pushed a few commits to this PR that should take care of the issues from the linked issue. Let's check it out once the draft docs PR has a new preview link ready for us to evaluate.

pkg/codegen/docs/gen.go
```diff
@@ -812,82 +748,26 @@ func (mod *modContext) genNestedTypes(member interface{}, resourceType bool) []d
...
 					continue
 				}
 
-				// Create maps to hold the per-language properties of this object and links to
-				// the API doc for each language.
+				// Create a map to hold the per-language properties of this object.
 				props := make(map[string][]property)
-				apiDocLinks := make(map[string]apiTypeDocLinks)
 				for _, lang := range supportedLanguages {
-					// The nested type may be under a different package in a language.
```
praneetloke, line 831: All or most of this deletion was related to generating the per-language API doc link for each object type.

pkg/codegen/docs/gen.go
```diff
@@ -329,13 +322,12 @@ func (ss nestedTypeUsageInfo) contains(token string, input bool) bool {
 }
 
 type modContext struct {
-	pkg          *schema.Package
-	mod          string
-	resources    []*schema.Resource
-	functions    []*schema.Function
-	children     []*modContext
-	tool         string
-	emitAPILinks bool
```
praneetloke, line 338: I debated keeping this property around and hence the code that uses it and simply default to passing `false` so that the links wouldn't be emitted. But I felt it was unnecessary to do so. If we ever do go back to generating the doc links I think a fresh start would be good given the state of things at _that_ time with regards to how the links should be generated. Anyway, deleting this property caused me to delete a lot of conditional code and also code that we just won't need anymore.

pkg/codegen/docs/gen.go
```diff
@@ -182,19 +182,12 @@ type enum struct {
...
-type apiTypeDocLinks struct {
-	InputType  string
-	OutputType string
-}
-
 // docNestedType represents a complex type.
 type docNestedType struct {
-	Name        string
-	AnchorID    string
-	APIDocLinks map[string]apiTypeDocLinks
```
praneetloke, line 195: This is the map that used to contain the `input` and `output` links for each object type, aka Supporting Types.

pkg/codegen/docs/gen.go
```diff
@@ -1799,7 +1598,6 @@ func (mod *modContext) genIndex() indexData {
 		Functions:          functions,
 		Modules:            modules,
 		PackageDetails:     packageDetails,
-		LanguageLinks:      mod.getLanguageLinks(),
```
praneetloke, line 1802: The use of this in the template was removed in https://github.com/pulumi/pulumi/pull/6270.

pkg/codegen/docs/gen.go
```diff
@@ -606,7 +588,7 @@ func (mod *modContext) genConstructorTS(r *schema.Resource, argsOptional bool) [
 			OptionalFlag: argsFlag,
 			Type: propertyType{
 				Name: argsType,
-				Link: argsDocLink,
+				Link: "#input",
```
praneetloke, line 591: For TS, C#, and Go, the "args" type will now just link to the "Input Properties"  section on the same page which details the properties of the args type.

praneetloke, line 591:
```suggestion
				Link: "#inputs",
```

pkg/codegen/docs/gen.go
```diff
@@ -661,7 +636,7 @@ func (mod *modContext) genConstructorGo(r *schema.Resource, argsOptional bool) [
 			OptionalFlag: argsFlag,
 			Type: propertyType{
 				Name: argsType,
-				Link: argsTypeLink,
+				Link: "#input",
```
praneetloke, line 639:
```suggestion
				Link: "#inputs",
```

pkg/codegen/docs/gen.go
```diff
@@ -748,7 +684,7 @@ func (mod *modContext) genConstructorCS(r *schema.Resource, argsOptional bool) [
 			DefaultValue: argsDefault,
 			Type: propertyType{
 				Name: name + "Args",
-				Link: argsTypeLink,
+				Link: "#input",
```
praneetloke, line 687:
```suggestion
				Link: "#inputs",
```

cnunciato:
This looks great. Thanks for the help @praneetloke!

justinvp (approved):
LGTM

PR #6290, opened 2021-02-09
Pin version of pip on Windows for Python tests

stack72:
Fixes: #6268

On Feb 1st 2021, GHA upgraded the windows build runner to set the
pip version to be 21.0.1 which broke our python tests on windows

This PR sets the version back to 20.3.3 so we can have time to
understand why we were broken by that upgrade

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -227,8 +227,11 @@ jobs:
         run: dotnet nuget locals all --clear
       - name: Install Python Deps
         run: |
-          pip3 install pyenv-win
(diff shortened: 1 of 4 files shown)


leezen:
@stack72 I haven't looked at the build failures, but depending on where it's failing, if it's because we're using Python 2 anywhere (e.g. for building a container), that could be the reason since pip 21 discontinued support for Python 2 (see https://github.com/pulumi/examples/pull/889 as an example where this affected us elsewhere)

Issue #6282, opened 2021-02-08
Fix issues with pulumi/pulumi windows build for Python


stack72:
The python tests are failing on windows and blocking our pipeline

stack72:
We are tracking the issue here as it is manifesting itself with the following error:

                 import pulumi
695
             ModuleNotFoundError: No module named 'pulumi'
696
             It looks like the Pulumi SDK has not been installed. Have you run pip install?
697
             If you are running in a virtualenv, you must run pip install -r requirements.txt from inside the virtualenv.

This is due to a change in how pipenv is acting in creating a virtual environment

Code changes were merged to master and subsequently reverted so we could try and get back to a normal state but the build continued to fail

PR #6296, opened 2021-02-09
Fix Python Windows test failures

justinvp:
The Python installation on the Windows GHA runner includes a `python3.exe` which is a symlink to `python.exe`.

Our integration tests look for the Python binary to use to initialize the test's virtual environment by first looking for `python3` on `PATH`, falling back to `python`. In this case, it finds `python3` and uses it to run `python3 -m venv venv`. However, this call ends up failing because `venv` will try to run `ensurepip` to ensure pip is available inside the virtual environment.

https://github.com/python/cpython/blob/ea46579067fd2d4e164d6605719ffec690c4d621/Lib/venv/__init__.py#L291-L298:

```python
    def _setup_pip(self, context):
        """Installs or upgrades pip in a virtual environment"""
        # We run ensurepip in isolated mode to avoid side effects from
        # environment vars, the current directory and anything else
        # intended for the global Python environment
        cmd = [context.env_exe, '-Im', 'ensurepip', '--upgrade',
                                                    '--default-pip']
        subprocess.check_output(cmd, stderr=subprocess.STDOUT)
```

`context.env_exe` is initialized as (pseudocode):

```python
executable = sys._base_executable
dirname, exename = os.path.split(os.path.abspath(executable))
if sys.platform == 'win32':
    binname = 'Scripts'
else:
    binname = 'bin'
binpath = os.path.join(env_dir, binname)
context.env_exe = os.path.join(binpath, exename)
```

So `context.env_exe` is based on `sys._base_executable`.

Here’s what `sys._base_executable` returns on GHA:

```
python -c "import sys;print(sys._base_executable)"
C:\hostedtoolcache\windows\Python\3.9.1\x64\python.exe

python3 -c "import sys;print(sys._base_executable)"
C:\hostedtoolcache\windows\Python\3.9.1\x64\python3.exe
```

So when we run `python3 -m venv /path/to/venv`, `context.env_exe` will be `/path/to/venv/Scripts/python3.exe`
But looking inside the `/path/to/venv/Scripts` dir, `python3.exe` doesn’t exist. Only `python.exe`.

To workaround, we set `PythonBin` explicitly to `python` for all the Python integration tests, so the tests will run `python -m venv venv` rather than `python3 -m venv venv`.

Fixes: #6282

### DIFF
--- a/tests/integration/integration_python_test.go
+++ b/tests/integration/integration_python_test.go
@@ -226,9 +226,7 @@ func TestResourceWithSecretSerializationPython(t *testing.T) {
 // Tests that we issue an error if we fail to locate the Python command when running
 // a Python example.
 func TestPython3NotInstalled(t *testing.T) {
-	if runtime.GOOS == WindowsOS {
-		t.Skip("Temporarily skipping test on Windows - pulumi/pulumi#6304")
-	}
+	t.Skip("Temporarily skipping failing test - pulumi/pulumi#6304")
 	stderr := &bytes.Buffer{}
 	badPython := "python3000"
 	expectedError := fmt.Sprintf(
--- a/tests/integration/integration_python_test.go
+++ b/tests/integration/integration_python_test.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"runtime"
 	"testing"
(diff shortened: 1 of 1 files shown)


pkg/testing/integration/program.go
```diff
@@ -691,6 +691,18 @@ func (pt *ProgramTester) getPythonBin() (string, error) {
 			if err != nil {
 				return "", errors.Wrapf(err, "Expected to find one of %q on $PATH", pythonCmds)
 			}
+
+			// If the path is a symlink, follow it.
+			info, err := os.Lstat(pt.pythonBin)
+			if err != nil {
+				return "", err
+			}
+			if info.Mode()&os.ModeSymlink == os.ModeSymlink {
```
stack72, line 700: this is the magic!!

stack72 (approved):

komalali:
Looks like ubuntu tests are all failing now :( 

https://github.com/pulumi/pulumi/pull/6296/checks?check_run_id=1866581976#step:15:3367

.github/workflows/run-build-and-acceptance-tests.yml
```diff
@@ -71,6 +71,11 @@ jobs:
         uses: actions/setup-node@v1
         with:
           node-version: ${{ matrix.node-version }}
+      - name: Install pip
+        run: |
+          pip3 install --upgrade pip==20.3.3
```
stack72, line 76: do we need pip3 or pip on non-windows?

Issue #6242, opened 2021-02-02
Add 'pulumi state protect' command


gpduck:
We currently occasionally have to unprotect resources for troubleshooting purposes due to #5749 and it would be nice if there was a way to re-protect them from the command line. I can't really think of any use cases outside of this scenario, but it still seems appropriate to have a way to "undo" the unprotect operation from the command line as well.

clstokes:
@gpduck Would the exact same semantics as https://www.pulumi.com/docs/reference/cli/pulumi_state_unprotect/ work here?

e.g.
`pulumi state protect <resource URN>`
and 
`pulumi state protect --all`

gpduck:
@clstokes yeah I think that seems fair. It would let people undo the unprotect operations easily by just changing the command name.

stack72:
Hi @gpduck 

So I was just looking at this issue - so the reason state unprotect works is that it would mean a delete operation will work without first needing to run an update

With a state protect command, the resources themselves would need to be updated with the 'protect: true' resource option or the next Pulumi update would effectively remove the protection again

This seems like it would negate the usefulness of the state protect command

I'd love your thoughts here

Paul

gpduck:
In the case mentioned in #5749 we're unprotecting so that we can get a diff, we don't actually want to delete the resources at all, so running a pulumi up after we do unprotect to get the protection re-enabled is extra scary.

komalali:
I'm curious if we've considered fixing this by addressing https://github.com/pulumi/pulumi/issues/5749 to generate a diff for protected resources rather than adding a new command for a very niche use-case?

leezen:
> In the case mentioned in #5749 we're unprotecting so that we can get a diff, we don't actually want to delete the resources at all, so running a pulumi up after we do unprotect to get the protection re-enabled is extra scary.

I have a slightly different take on this and am curious if the following would work for you @gpduck. If I understand the use case correctly, is the following true?

1. You are at commit A with a corresponding state (S) that matches the code at this commit.
2. You write some code (commit B) and want to see the diff, which results in hitting #5749.
3. You unprotect the resources (generating state S') in order to see the diff.
4. At this point, you'd like to re-protect the resources.

I understand the hesitancy to run `pulumi up` with commit B since you don't actually want to replace the resources. But is there a reason you cannot checkout commit A and run `pulumi up`? I would expect the diff at that point to only be adding the `protect` flag?

gpduck:
[...]

PR #6313, opened 2021-02-11
[cli] Add the ability to protect resources via pulumi stack protect

stack72:
Fixes: #6242

[asciicast](https://asciinema.org/a/jCigBkgTNY5qDV3zTuiClRgzW)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ CHANGELOG
 
 ## HEAD (Unreleased)
 
+- [cli] Add the ability to protect resources via `pulumi state protect`
+  [#6313](https://github.com/pulumi/pulumi/pull/6313)
+
 - [CLI] Add pagination options to `pulumi stack history` (`--page`, `--page-size`). These replace the `--limit` flag.
   [#6292](https://github.com/pulumi/pulumi/pull/6292)
 
--- a/pkg/cmd/pulumi/state.go
+++ b/pkg/cmd/pulumi/state.go
@@ -49,6 +49,7 @@ troubleshooting a stack or when performing specific edits that otherwise would r
 
 	cmd.AddCommand(newStateDeleteCommand())
 	cmd.AddCommand(newStateUnprotectCommand())
+	cmd.AddCommand(newStateProtectCommand())
 	return cmd
 }
 
--- a/pkg/cmd/pulumi/state_protect.go
+++ b/pkg/cmd/pulumi/state_protect.go
@@ -1,0 +1,105 @@
+// Copyright 2016-2018, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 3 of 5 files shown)


infin8x (approved):

komalali (commented):
I have a general question. Am I understanding correctly that this command doesn't do anything if the resource isn't already protected? Then what is this command doing? And should it return an error if the resource isn't already protected with the resource option?

stack72:
@komalali this follows *exactly* the implementation of the unprotect - it uses the same func and just passed the opposite

It has a super specific use case...

The customer who has been using this hasn't complained about the output so I wasn't sure of pulling at that thread but happy to it you think we should check it out? (Note: it's an easy change)

komalali:
@stack72 since this is now a public command that other folks are likely to see, I think it's worth doing our due diligence to make sure we don't confuse people. I realize that this is more of a `undo unprotect` rather than an actual `protect` but it is a little odd.

If it's not entirely straightforward I'm okay to leave it as is, but if it's easy that'd be nice.

stack72:
Great point!

So one thought on the experience here

When a user passes '--all' - I will keep a list of the resources it can't protect (and unprotect by fixing the only one) and then I'll print an error saying unable to protect (or unprotect) the following urns
- *urn1
- *urn2
Etc

Sound ok?

komalali:
Yup sounds great! Maybe we just want to print a warning about the resources we couldn't protect/unprotect rather than an error in the `--all` case, but we should definitely error when there's a `pulumi state protect <resource>` and that resource doesn't have the correct resource option set.

stack72:
It'll be with you in the morning ;)

stack72:
So @komalali / @infin8x I spent the morning looking at making the change to error if the resource wasn't actually able to be protected / unprotected (due to the existing state). The way the function runs is that we don't save the state on a resource by resource basis so if we error out then we are not changing any resources. This is due to how `runTotalStateEdit` works - I think we may be pulling at a thread here to refactor how all of this would work...

My thoughts for this would be as follows:

1. We accept the current workflow (where we don't error) and we open an issue to follow up in the area. Technically it's not a breaking change - it's introducing better error handling and would allow us to deliver the value to the user now
2. We update the text on the command help to point out that we will only change eligible resources (ie. if we run protect then the resource won't be updated if already updated). We can also update the user message to suggest all eligible resources have been updated
3. We can add some diagnostics messages to suggest which resources have been changed

Thoughts?  

Paul

komalali:
Sure, that's fine with me. Looks like @infin8x already approved the PR too so perhaps my usability trepidations are unwarranted 🤷

stack72:
> Sure, that's fine with me. Looks like @infin8x already approved the PR too so perhaps my usability trepidations are unwarranted 🤷

I defo don't think that's the case at all! You made some great points on it - things we need to follow up and fix IMO

infin8x:
100% agree. Should've left a note with my approval - y'all are still much more familiar with the CLI ergonomics than I am, and I'll defer to you on making sure we ship the best experience.

stack72:
closing this out

PR #6318, opened 2021-02-11
Default to UseAutomaticVirtualEnv in integration tests

justinvp:
This way, the tests use the built-in virtual environment support by default, which is what most customers will be using. A new `UsePipenv` option is available to go back to using pipenv for tests.

### DIFF
--- a/pkg/testing/integration/program.go
+++ b/pkg/testing/integration/program.go
@@ -259,8 +259,11 @@ type ProgramTestOptions struct {
 	// Additional environment variables to pass for each command we run.
 	Env []string
 
-	// Automatically create and use a virtual environment, rather than using the Pipenv tool.
+	// Automatically create and use a virtual environment, rather than using the Pipenv tool. This is now the default
+	// behavior, so this option no longer has any affect. To go back to the old behavior use the `UsePipenv` option.
 	UseAutomaticVirtualEnv bool
+	// Use the Pipenv tool to manage the virtual environment.
+	UsePipenv bool
 }
 
 func (opts *ProgramTestOptions) GetDebugLogLevel() int {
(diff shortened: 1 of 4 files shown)


pkg/testing/integration/program.go
```diff
@@ -1748,11 +1758,7 @@ func (pt *ProgramTester) preparePythonProject(projinfo *engine.Projinfo) error {
 		}
 
 		if err := pt.runVirtualEnvCommand("virtualenv-pip-install",
-			[]string{"pip", "install", "-r", "requirements.txt"}, cwd); err != nil {
-			return err
-		}
-	} else {
-		if err = pt.preparePythonProjectWithPipenv(cwd); err != nil {
+			[]string{"python", "-m", "pip", "install", "-r", "requirements.txt"}, cwd); err != nil {
```
justinvp, line 1761: I changed this here to match what we do inside the Python SDK, running it as `python -m pip install` rather than `pip install`.

https://github.com/pulumi/pulumi/blob/066507cd009ae188db7a2930ef35c1f8b02fade6/sdk/python/python.go#L231

stack72:
Do we need the Pulumi.yaml file to be updated before the venv will be created or am I conflating 2 separate things here?

justinvp:
> Do we need the Pulumi.yaml file to be updated before the venv will be created or am I conflating 2 separate things here?

No. The integration test code will handle that automatically.

https://github.com/pulumi/pulumi/blob/066507cd009ae188db7a2930ef35c1f8b02fade6/pkg/testing/integration/program.go#L1744-L1748

It doesn't hurt for the `Pulumi.yaml` to already have it. And for the examples repo, we will want to go ahead and add it to the `Pulumi.yaml` for the examples ([this PR](https://github.com/pulumi/examples/pull/885) that I will finish up), just to make it easier for folks who are acquiring the examples (either via cloning the repo, `pulumi new <url>`, or the Pulumi button). When it is specified in `Pulumi.yaml` and the dir doesn't exist, the CLI will create it and install deps automatically.

stack72 (approved):
Awesome! The set of PR tests and master should help us get some mileage around this :)

PR #6322, opened 2021-02-12
WIP: provider implementer's guide

pgavlin:


sdk/go/common/resource/plugin/provider-implementers-guide.md
```diff
@@ -0,0 +1,609 @@
...
+To update a resource, the engine calls the provider's [`Update`](#update) method with the inputs
+returned from [`Check`](#check), the resource's ID, and its last refreshed state. [`Update`](#update) returns
+the new state of the resource. The resource's ID may not be changed by a call to [`Update`](#update).
+
+To replace a resource, the engine first calls [`Check`](#check) with an empty set of prior inputs
+and the inputs supplied with the resource's registration. If [`Check`](#check) fails, the resource
+is not replaced. Otherwise, the inputs returned by this call to [`Check`](#check) will be used to
+create the replacement resource. Next, the engine inspects the resource options supplied
+with the resource's registration and result of the call to [`Diff`](#diff) to determine whether
+the replacement can be created before the original resource is deleted. This order of
```
lukehoban, line 321: Perhaps explicitly note - "if either the `deleteBeforeReplace` resource option or the `deleteBeforeReplace` property of the `Diff` output is true, the resource will be deleted before it's replacement is created, else its replacement will be created before it is deleted."

sdk/go/common/resource/plugin/provider-implementers-guide.md
```diff
@@ -0,0 +1,609 @@
...
+first deletes the transitive closure of resource that depend on the resource being
+replaced. Once these deletes have completed, the engine deletes the original resource by
+calling the provider's [`Delete`](#delete) method with the resource's ID and original state. Finally,
+the engine creates the replacement resource by calling [`Create`](#create) with the re-checked
+inputs.
+
+If a managed resource registered by a Pulumi program is not re-registered by the next
+successful execution of a Pulumi progam in the resource's stack, the engine deletes the
+resource by calling the resource's provider's [`Delete`](#delete) method with the resource's ID and
+last refereshed state.
```
lukehoban, line 336: There's also another case where pending deletes can get processed at the beginning of an update - right?

sdk/go/common/resource/plugin/provider-implementers-guide.md
```diff
@@ -0,0 +1,609 @@
...
+- `failures`: any validation failures present in the inputs. These failures should be
+              constrained to type and range mismatches. A failure is a tuple of a
+              [property path](#property-paths) and a failure reason.
+
+#### Diff
+
+The `Diff` method is responsible for calculating the differences between the actual and
+desired state of a resource as represented by its last recorded state and new input
+`Object` as returned from [`Check`](#check) or [`Read`](#read) and the logical
+operation necessary to reconcile the two (i.e. no operation, an `Update, or a `Replace`).
```
lukehoban, line 377:
```suggestion
operation necessary to reconcile the two (i.e. no operation, an `Update`, or a `Replace`).
```

sdk/go/common/resource/plugin/provider-implementers-guide.md
```diff
@@ -0,0 +1,609 @@
...
+In addition, the following properties should be returned for compatibility with older
+clients:
+
+- `replaceKeys`: the list of top-level input property names with changes that require that the
+                 resource be replaced.
+- `stableKeys`: the list of top-level input property names that did not change and
+                top-level output properties that are guaranteed not to change.
+- `changedKeys`: the list of top-level input property names that changed.
+
+If a provider is unble to compute a diff because its configuration contained
```
lukehoban, line 414:
```suggestion
If a provider is unable to compute a diff because its configuration contained
```

sdk/go/common/resource/plugin/provider-implementers-guide.md
```diff
@@ -0,0 +1,609 @@
...
+             by comparing the resource's last recorded _inputs_ with its current inputs.
+
+In addition, the following properties should be returned for compatibility with older
+clients:
+
+- `replaceKeys`: the list of top-level input property names with changes that require that the
+                 resource be replaced.
+- `stableKeys`: the list of top-level input property names that did not change and
+                top-level output properties that are guaranteed not to change.
+- `changedKeys`: the list of top-level input property names that changed.
```
lukehoban, line 412: Some of these have different names from the RPC names - which names are we using here?  

(I see `Diffs`, `Stables` and `Replaces` in the RPC proto (and therefore in all our current implementations that implement the raw Go RPC interface))

sdk/go/common/resource/plugin/provider-implementers-guide.md
```diff
@@ -0,0 +1,609 @@
...
+- `olds`: the last recorded state `Object` for the resource. This `Object` must have been
+          generated by a call to `Create`, `Read`, or `Update`, and will never contain
+          [`Unknown`s](#unknowns).
+- `news`: the current input `Object` for the resource as returned by [`Check`](#check) or
+          [`Read`](#read). This value may contain [`Unknown`s](#unknowns).
+- `ignoreChanges`: the set of [property paths](#property-paths) to treat as unchanged.
+
+##### Results
+
+- `detailedDiff`: the [detailed diff](#detailed-diffs) between the resource's actual and
```
lukehoban, line 392: This is still optional - right?

Worth noting what happens if you don't return it?  Or are you intentionally trying to lean toward insisting this is returned?

sdk/go/common/resource/plugin/provider-implementers-guide.md
```diff
@@ -0,0 +1,609 @@
...
+clients:
+
+- `replaceKeys`: the list of top-level input property names with changes that require that the
+                 resource be replaced.
+- `stableKeys`: the list of top-level input property names that did not change and
+                top-level output properties that are guaranteed not to change.
+- `changedKeys`: the list of top-level input property names that changed.
+
+If a provider is unble to compute a diff because its configuration contained
+[`Unknown`s](#unknowns), it can return an error that indicates as such. The client should
```
lukehoban, line 415: What exactly does it mean to "return an error that indicates as such"?

sdk/go/common/resource/plugin/provider-implementers-guide.md
```diff
@@ -0,0 +1,609 @@
...
+chosen over an approach that used old/new values due in order to remove the possibility
+of a provider accidentally revealing a secret value as part of a diff. We should
+reconsider this approach if we can find an easy way to maintain secretness.
+
+#### Create
+
+The `Create` method is responsible for creating a new instance of a resource from an
+input `Object` and returning the resource's state `Object`. `Create` may be called during
+a [preview](#preview) in order to compute a hypothetical state `Object` without actually
+creating the resource, in which case the `preview` argument will be `true`.
```
lukehoban, line 447: Huh - I did know about this.  Do any of our existing providers support this?  What happens if a provider doesn't support this (as I assume it's nontrivial to support?)

sdk/go/common/resource/plugin/provider-implementers-guide.md
```diff
@@ -0,0 +1,609 @@
...
+
+##### Parameters
+
+- `urn`: the [URN](#urns) of the resource.
+- `news`: the input `Object` for the resource. This value must have been generated by a
+          prior call to `Check`. If `preview` is true, this value may contain
+          [`Unknown`](#unknowns) value; otherwise, it is guaranteed to be fully-known.
+- `timeout`: the timeout for the create operation. If this value is `0`, the provider
+             should apply the default creation timeout for the resource.
```
lukehoban, line 456: Perhaps also note "if the resource does not support timeouts, this can be ignored"?

sdk/go/common/resource/plugin/provider-implementers-guide.md
```diff
@@ -0,0 +1,609 @@
...
+             Nested properties with values that are not computable must be explicitly set
+             to [`Unknown`](#unknowns). If it is not possible to guarantee that the value
+             produced by a preview will match the value that would be produced by actually
+             creating the resource, the value should be left unknown.
+
+##### Results
+
+- `id`: the ID for the created resource. If `preview` is true, this value will be ignored.
+- `state`: the new state `Object` for the resource. If `preview` is true, this value may
+           contain [`Unknown`s](#unknowns).
```
lukehoban, line 470: It's not part of the core RPC contract itself - but it seems like in principle all native Pulumi providers will want to use the "partial error" support to ensure they return `id` and `state` along with any error that occurs after a backing physical resource has been allocated - like we do in the Kubernetes provider, and likely should ~anywhere else to avoid leaking resources while waiting for them to initialize (and potentially fail to initialize).  Should this be called out here somehow?

lukehoban (approved):
🎉

pgavlin:
I'm going to go ahead and merge this. I've marked the incomplete sections with `TODO` and will return to them later.

PR #6323, opened 2021-02-12
Allow optional inputs to be Input<T | undefined>

### LINKED ISSUE #6175: Input in Typescript should accept undefined
## Expected Behavior
Given an input like `readonly website?: Input<boolean>;` and a `gcp.storage.Bucket` resource, I want to be able to the following:
```typescript
new gcp.storage.Bucket(
      `some-name`,
      {
        name: 'my_url,
        location: 'US',
        storageClass: 'MULTI_REGIONAL',
        uniformBucketLevelAccess: true,
        website: output(website).apply((w) =>
          w === false
            ? undefined
            : {
                mainPageSuffix: 'index.html',
                notFoundPage: 'index.html',
              }
        ),
        forceDestroy: false,
      },
    );
```

## Current Behavior

Since the `undefined` is inside the `Output`, the compiler complains that `Type 'undefined' is not assignable to type 'BucketWebsite'`.

## Steps to Reproduce

Code above is pretty straight forward.

## Context (Environment)

Provider: GCP
Version: 2.17.0

lukehoban:
The Node.js runtime accepts `Output<undefined>` values as inputs to optional parameters, but the TypeScript typing currently does not allow these.  This extends the TypeScript typings to allow `Output<undefined>` values as inputs to optional input properties.

I cannot think of any way in which this is breaking or would regress any aspect of the TypeScript experience, other than making the `.d.ts` files a little "noisier".

Fixes #6175

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,5 +1,9 @@
 ### Improvements
 
+- [sdk/nodejs] - Type optional input properties as `Input<T | undefined> | undefined` instead of
+  `Input<T> | undefined`. This allows users to pass `Output<T | undefined>` for optional input
+  properties.
+  [#6323](https://github.com/pulumi/pulumi/pull/6323)
 
 ### Bug Fixes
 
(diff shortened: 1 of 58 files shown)


komalali:
@lukehoban looks like this is causing some badness for azure-nextgen

pgavlin:
@lukehoban is this still something you think we should pursue?

leezen:
@pgavlin I've been linking occurrences of this to #6175 when I see them and there've been two in the last two months, so I suspect this comes up 'often enough' for TypeScript users.

lukehoban:
This change is probably still generally a good idea, and #6175 is worth fixing, but this PR has gotten a bit stale, and we likely need to think a bit more carefully about potential breaking change experience (I don't *think* this is a breaking change, but can't convince myself immediately that I should be sure of that).  For now, closing this out.

PR #6325, opened 2021-02-12
[codegen/docs] - Standardize function and resource titles.

komalali:
Fixes: https://github.com/pulumi/docs/issues/5209


davidwrede (approved):

lukehoban:
One thing that is maybe outside the scope of this PR - but that we probably want to also address here - these two pages have titles that are still aligned with the "old" style. It's possible/likely they should match the new style as well?

https://www.pulumi.com/docs/reference/pkg/aws/

https://www.pulumi.com/docs/reference/pkg/aws/acm/

komalali:
@lukehoban I'll follow up with a PR for those.

PR #6337, opened 2021-02-16
[sdk/go] Add To methods to convert array and map values to Inputs

### LINKED ISSUE #3861: Add primitive helper methods to the Go SDK
There are a few repetitive motions with the Go SDK especially with converting prompt arrays and maps into pulumi equivalents. 

Joe also ran into a similar issue with `StringArrayInput` that required a helper when working on an [ECS example](https://github.com/pulumi/examples/pull/540):

```go
func toPulumiStringArray(a []string) pulumi.StringArrayInput {
	var res []pulumi.StringInput
	for _, s := range a {
		res = append(res, pulumi.String(s))
	}
	return pulumi.StringArray(res)
}
```

We should add some helpers to make this a little more ergonomic.

lukehoban:
Adds two new methods for every Array and Map type:

```go
func ToStringArray(in []string) StringArray 
func ToStringArrayOutput(in []StringOutput) StringArrayOutput
```

These make it easier to take raw arrays of underlying values and project them into `*Array` instances of the lifted Pulumi type.

This addresses most of the cases in #3861, but does not handle cross-casting between things like `[]ID` and `[]string`.  Conversions like that always require writing a `for` loop in Go, and I believe it's reasonable to say that they require a for loop in an Apply in the Pulumi Go programming model.

It could be argued that we don't even need the helpers added here along similar grounds - but in this case, the Pulumi Go programming model itself is the one forcing a type projection that wouldn't have been necessary in plain Go, so it feels more justified to try and make this possible without writing a for loop manually.  Though notably, these conversion functions are all quite simple (much simpler than the implementations suggested in #3861), and it could be argued that Go idiom would be to just write these conversions inline in code that needed it. But to avoid the difficulty discovering the simple implementations of these functions as encountered in #3861 and elsewhere, this feels like a reasonable addition.

This does *not* introduce all the various intermediate combinations for the `ArrayArray`/`ArrayMap`/`MapArray`/`MapMap` types.  For example, we generate `func ToStringArrayArray(in [][]string) StringArrayArray` and `func ToStringArrayArrayOutput(in []StringArrayOutput) StringArrayArrayOutput` but do not generate any method that takes `[][]StringOutput` as input.  We could in theory add this later - but it's unclear what name to give it.  

Fixes #3861.

### DIFF
--- a/sdk/go/pulumi/generate.go
+++ b/sdk/go/pulumi/generate.go
@@ -104,21 +104,21 @@ func (b builtin) MapIndexReturnType() string {
 }
 
 func (b builtin) IndexConversion() string {
-	if b.item.inputType != "" {
+	toType := b.item.Name
+	switch {
+	case b.item.inputType != "":
 		// No conversion necessary for types which are their own input type.
 		return ""
-	}
-	toType := b.item.Name
-	if toType == "Input" {
+	case toType == "Input":
 		// There is no direct conversion into `Input`, so special case to use `ToOutput`
 		return "ToOutput"
-	}
-	if b.item.item == nil {
+	case b.item.item == nil:
 		// The item type is a primitive - so use a direct conversion into the defined type (e.g. `String(...)`)
 		return toType
+	default:
+		// The item type is itself a container - so use one of the other code-generated `To<Type>` methods.
+		return "To" + toType
 	}
-	// The item type is itself a container - so use one of the other code-generated `To<Type>` methods.
-	return "To" + toType
 }
 
 func (b builtin) ElementType() string {
--- a/sdk/go/pulumi/generate.go
+++ b/sdk/go/pulumi/generate.go
@@ -142,6 +142,10 @@ func (b builtin) DefineToFunction() bool {
 	return b.DefineInputMethods()
 }
 
+func (b builtin) ItemExample() string {
+	return b.item.Example
+}
+
 func (b builtin) ElemExample() string {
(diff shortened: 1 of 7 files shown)


sdk/go/pulumi/generate.go
```diff
@@ -100,6 +103,24 @@ func (b builtin) MapIndexReturnType() string {
...
+		return "ToOutput"
+	}
+	if b.item.item == nil {
+		// The item type is a primitive - so use a direct conversion into the defined type (e.g. `String(...)`)
+		return toType
+	}
+	// The item type is itself a container - so use one of the other code-generated `To<Type>` methods.
+	return "To" + toType
+}
+
```
pgavlin, line 123: Nit: I slightly prefer the `switch` version of this sort of logic

```suggestion
func (b builtin) IndexConversion() string {
	toType := b.item.Name
	switch {
	case b.item.inputType != "":
		// No conversion necessary for types which are their own input type.
		return ""
	case toType == "Input":
		// There is no direct conversion into `Input`, so special case to use `ToOutput`
		return "ToOutput"
	case b.item.item == nil:
		// The item type is a primitive - so use a direct conversion into the defined type (e.g. `String(...)`)
		return toType
	default:
		// The item type is itself a container - so use one of the other code-generated `To<Type>` methods.
		return "To" + toType
	}
}

```

pgavlin (approved):
One nit. LGTM otherwise.

PR #6339, opened 2021-02-16
Exposing the version in the UpdateSummary struct.

ringods:
TL;DR After running `stack.Up(...)` via the Automation API, I want access to the update `version` number for the stack.

A resource type integration in [Concourse](https://concourse-ci.org) expects that after an "update" (semantic specific to the resource type implementation) that an identifier for this new version is returned to Concourse. 

For the [Pulumi Concourse CI](https://github.com/ringods/pulumi-resource) integration, I am using the `version` field to model the series of Pulumi deployments. After running `stack.Up(...)` via the Automation API, I would like access to the `version` field from the result JSON so I can return it to Concourse:

https://github.com/ringods/pulumi-resource/blob/c122a20c802fc200f69c303d525c6f83e28ed83b/pkg/out/out.go#L45-L49

/cc @EvanBoyle

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ CHANGELOG
 
 ## HEAD (Unreleased)
 
+- [automation/go] Exposing the version in the UpdateSummary struct.
+  [#6339](https://github.com/pulumi/pulumi/pull/6339)
+
 - [CLI] Add pagination options to `pulumi stack history` (`--page`, `--page-size`). These replace the `--limit` flag.
   [#6292](https://github.com/pulumi/pulumi/pull/6292)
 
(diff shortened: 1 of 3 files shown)


ringods:
A quick chat with @stack72 on parity of this identifier between the different language SDKs brought us the info that this `version` field is already exposed in the other SDKs:

* NodeJS: https://github.com/pulumi/pulumi/blob/master/sdk/nodejs/x/automation/stack.ts#L520
* Python: https://github.com/pulumi/pulumi/blob/master/sdk/python/lib/pulumi/x/automation/stack.py#L72

which means my PR brings parity again amongst the languages. 😄

EvanBoyle (approved):
@ringods thanks for the contribution!

PR #6345, opened 2021-02-16
[programgen] Match union types based on the discriminator value

mikhailshilkov:
Resolve https://github.com/pulumi/pulumi-azure-nextgen/issues/210


pkg/codegen/hcl2/binder_schema.go
```diff
@@ -171,6 +171,9 @@ func (b *binder) schemaTypeToTypeImpl(src schema.Type, seen map[schema.Type]mode
 			if !prop.IsRequired {
 				t = model.NewOptionalType(t)
 			}
+			if prop.ConstValue != nil {
+				t = model.NewConstType(t, prop.ConstValue)
+			}
```
pgavlin, line 176: This seems reasonable to me, as it will improve our ability to type check PCL programs against a schema. that contains const-valued props.

pkg/codegen/hcl2/rewrite_convert.go
```diff
@@ -140,6 +143,59 @@ func rewriteConversions(x model.Expression, to model.Type) (model.Expression, bo
...
+// A given object expression would only match a single type of the union.
+func matchDiscriminatedObjectTypes(modelType model.Type, x *model.ObjectConsExpression) model.Type {
+	switch typ := modelType.(type) {
+	case *model.ObjectType:
+		if els, ok := GetSchemaForType(typ); ok {
+			if v, ok := els.(*schema.ObjectType); ok {
+				for _, prop := range v.Properties {
+					if prop.ConstValue != nil {
+						// This is the essence of the function: we found the discriminator property, so we can
+						// exclude the current type if its const value doesn't match the actual property value.
```
pgavlin, line 158: This is the only thing that really concerns me: there's no guarantee that an object type will have a single const property and that this property (if any) will serve as a discriminant. I think that we need more metadata to make that guarantee.

mikhailshilkov: > there's no guarantee

True.

> have a single const property

In theory, this should work for multiple const properties too, I.e. two discriminators (that doesn't happen in Azure NG).

A worse thing could be two or more types that would match a given PCL code. FWIW, in this case, we will get the behavior that we have currently in master.

We could introduce a first-class "base type" concept with a discriminator property but that sound like a bigger effort. Or do you have other ideas for the guarantee?

pgavlin: > Or do you have other ideas for the guarantee?

I was thinking that we could add the [OpenAPI 
`discriminator` field](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#discriminatorObject) to `schema.TypeSpec`, then use that in the conversion rewriter.

mikhailshilkov: I agree it's a good step. This should also help generate class hierarchies in C#, for example.

Do you want me to try this in this PR or should we take those separately?

pgavlin: I think that we should add the `discriminator` field in a separate PR before landing this PR. I don't think that adding the `discriminator` field is too much work.

mikhailshilkov: Opened https://github.com/pulumi/pulumi/issues/6392

PR #6349, opened 2021-02-16
[Automation] Fix python nested config parsing

komalali:
Also adds a test that parses nested configuration in each language.

To be clear, automation API reads lists and maps as json strings. For now, it is up to the user to de/reserialize the json. This PR is merely fixing the python implementation to match what is already happening in nodejs and go.


justinvp (approved):
LGTM

PR #6351, opened 2021-02-16
[sdk/python] - Define __all__ for better autocomplete.

komalali:
The Python SDK has sub-par autocomplete in VS Code if `__all__` is not defined. This PR makes it better. 

## Before

Screen Shot 2021-02-16 at 3 08 54 PM
Screen Shot 2021-02-16 at 3 09 03 PM

## After

Screen Shot 2021-02-16 at 3 09 33 PM
Screen Shot 2021-02-16 at 3 09 47 PM


viveklak (approved):
LGTM. One thought though - is the universe of these things going to change over time? Can we enforce some means to signal that this list is out-of-date through some sort of a test failure?

sdk/python/lib/pulumi/__init__.py
```diff
@@ -97,3 +97,86 @@
...
+    "invoke",
+    "metadata",
+    "resource",
+    "output",
+    "log",
+    "stack_reference",
+    "runtime",
+    "dynamic",
+    "policy",
+    "x"
```
justinvp, line 181: Should we prune some of these submodules that aren't really intended to be used as modules -- the ones where we're already importing from them and exporting at the top-level?

```suggestion
    "runtime",
    "dynamic",
    "policy",
    "x"
```

Might require having to fix-up some imports internally or in tests.

sdk/python/lib/pulumi/runtime/__init__.py
```diff
@@ -57,3 +57,48 @@
...
+    "ResourcePackage",
+    "register_resource_module",
+    "register_resource_package",
+
+    # submodules
+    "config",
+    "mocks",
+    "settings",
+    "stack",
+    "invoke",
```
justinvp, line 102: Not that it matters, but `"invoke"` is already listed above.

sdk/python/lib/pulumi/runtime/__init__.py
```diff
@@ -57,3 +57,48 @@
...
+
+    # settings
+    "Settings",
+    "configure",
+    "is_dry_run",
+    "reset_options",
+
+    # stack
+    "run_in_stack",
+    "get_root_resource",
```
justinvp, line 82: This is actually defined in `settings.py`. The `__init__.py` file is wrong, it should get it from `.settings` rather than `.stack`.

komalali:
> is the universe of these things going to change over time?

Definitely. As new features are added we'll have to add them to these lists depending on whether we want to expose them publicly and/or via that module.

> Can we enforce some means to signal that this list is out-of-date through some sort of a test failure?

This is tough for a couple of reasons, the biggest one being that this is somewhat of a "curated list" and so there's no cut and dry logical reasoning for whether or not something should be included in this list. It is python convention to add an underscore before "private" identifiers, but it's not one we've consistently applied throughout our SDK. Perhaps this is worth a sweep-through as part of 3.0 though.

sdk/python/lib/pulumi/runtime/__init__.py
```diff
@@ -57,3 +57,48 @@
...
+    "register_resource_module",
+    "register_resource_package",
+
+    # submodules
+    "config",
+    "mocks",
+    "settings",
+    "stack",
+    "invoke",
+    "rpc"
```
justinvp, line 103: Similarly, wondering if we can prune any of these (though, less of a concern here since these are already more "hidden" under `runtime`).

```suggestion
```

The only one I'm not sure about is `"rpc"` which may be worth still exporting because it has some utilities that aren't imported and then exported from `runtime`.

justinvp (approved):
Some comments, but LGTM. Thanks for cleaning this up!

sdk/python/lib/pulumi/__init__.py
komalali: Yeah, I was wondering this myself and hoping you'd confirm this for me 😄👍

komalali: Although, it looks like VSCode still has file names show up in autocomplete (as evidenced by the screenshot of the `After` automation autocomplete. Seems like if it doesn't start with an `_` it will show up in autocomplete. Wondering if 3.0 might be an opportunity to rename the files that we don't intend on exporting.

justinvp: For `automation` can we go ahead and rename such files to have the preceding `_`? (Definitely before we promote from `x`.)

komalali: Yup, doable for `automation` even now since it's in preview and all of the documentation/examples import directly from `automation` anyway.

PR #6375, opened 2021-02-18
Fix serialization tests of .NET Automation

mikhailshilkov:


mikhailshilkov:
Self-merging to fix the master

Issue #3919, opened 2020-02-12
GOOGLE_CREDENTIALS environment var doesn't work


rawkode:
Weirdly, the `pulumi login` works - but the `pulumi update` fails.

```
#!/bin/bash -eo pipefail
pulumi login gs://influxdata-devrel-operations/devrel-metrics
pulumi update --stack production --cwd .

Logged into ebab012ec61f as root (gs://influxdata-devrel-operations/devrel-metrics)

error: constructing secrets manager of type "cloud": open keeper gcpkms://<redacted>: google: could not find default credentials.

See https://developers.google.com/accounts/docs/application-default-credentials for more information.

Exited with code exit status 255
CircleCI received exit code 255
```

rawkode:
@Place1 Are you seeing this behaviour too, when using gcpkms secrets provider?

clstokes:
Unfortunately this seems to be a case where Pulumi is inconsistent in the environment variables that are being used. Using `GOOGLE_CREDENTIALS` _does_ work for the Pulumi application itself, but does _not_ work for the `gpckms` encryption provider (I was able to reproduce this).

I was able to use the `gcpkms` provider using the `GOOGLE_APPLICATION_CREDENTIALS` environment variable (per https://cloud.google.com/docs/authentication/production). 

```
$ echo $GOOGLE_APPLICATION_CREDENTIALS
/home/vagrant/gcreds.json

$ pulumi stack init feb20 --secrets-provider="gcpkms://projects/REDACTED/locations/us-east1/keyRings/pulumi-feb20-test/cryptoKeys/one"
Created stack 'feb20'
```

The Pulumi application also works with `GOOGLE_APPLICATION_CREDENTIALS` so you can use that instead of `GOOGLE_CREDENTIALS`.

rawkode:
@clstokes  Argh, so frustrating. What about this PR? There was a lot of effort and back n' forth to make this work

https://github.com/pulumi/pulumi/pull/2906

Frassle:
This came up in community slack, prompted me to look at it quickly. Looks like out envvar workaround we use for GCP state providers (found in pkg/backend/filestate/gcpauth.go) isn't also applied to GCP secret providers.

PR #6379, opened 2021-02-18
Add support for authentication via GOOGLE_CREDENTIALS

farcaller:
This PR also propagates the auth pulumi used for the state storage into the `gcpkms://` secrets provider if one is used.

Fixes https://github.com/pulumi/pulumi/issues/3919


farcaller:
I can split the PULUMI_GOOGLE_CREDENTIALS_HELPER and gcpkms:// parts into two separate PRs if it helps to make a decision. The latter is a much significant pain point personally (although the former is annoying too).

farcaller:
@stack72 tentative ping?

farcaller:
Another ping?

farcaller:
I'd still appreciate some feedback on this 😄

farcaller:
> Why optionally? We don't make this optional for the file state backend, feels like it would be ok to just copy the same behavior across for the secret backend.

Mostly the backwards compatibility. It allows one to migrate to `PULUMI_GOOGLE_CREDENTIALS_HELPER` without breaking their secrets provider (if e.g. different credentials are used)

> What workflow is this solving? We don't have any other authentication systems that would shell out like this and it feels a bit weird adding one just for gcp.

The unique name solves the part where you might have a valid `GOOGLE_APPLICATION_CREDENTIALS` in your shell pointing to a different account, to be picked up by the default GCP provider. By making this explicit there's no room for such confusion. You have one set of credentials for state and/or secret provider and anything else you want for the gcp provider.

Using the helper instead of a path to a secret account file is just more flexible and is generally in line with how credential helpers work in e.g. git or kubectl. It allows one to store the secret account key encrypted or in a remote storage (1P or vault), without having the secret data spilled to the local disk.

Frassle:
> Mostly the backwards compatibility.

If we dropped the PULUMI_GOOGLE_CREDENTIALS_HELPER addition from this PR then the only change here is if someone had GOOGLE_APPLICATION_CREDENTIALS and GOOGLE_CREDENTIALS set to different settings then the secret provider would of just been looking at the former, and would now prefer the later. But it would bring the secret provider auth inline with the state backend and resource provider auth. I think we'd be ok with that change.

> Using the helper instead of a path to a secret account file is just more flexible and is generally in line with how credential helpers work in e.g. git or kubectl.

Sounds like we should be supporting this for other authenticated systems as well though? PULUMI_AWS_CREDENTIALS_HELPER, PULUMI_AZURE_CREDENTIALS etc. While that sounds like it could be useful I think we'd need to think some more about what the appropriate way to plumb that change through everything would look like.

If you want to merge in master, make the gcpkms:// change non-optional and remove the PULUMI_GOOGLE_CREDENTIALS_HELPER I'd be happy to approve that change. 

Also please raise an issue about PULUMI_GOOGLE_CREDENTIALS_HELPER I think we need a bigger discussion about how we should support helpers for everything. Might be it makes sense to just take your idea here, but want some wider input on that before approving.

Frassle:
> Currently both the state and the secret provider use GOOGLE_APPLICATION_CREDENTIALS.

No they're actually inconsistent here. The state will use GOOGLE_CREDENTIALS and then fall back to GOOGLE_APPLICATION_CREDENTIALS. The gcp resource provider is the same. But the secret provider only looks at GOOGLE_APPLICATION_CREDENTIALS. 

The changes in this PR to unify state and secret to both use the ResolveGoogleCredentials fixes that and is worth having.

farcaller:
Right, I see your point.

Frassle:
I've raised https://github.com/pulumi/pulumi/issues/9246 to discuss the idea of adding support for auth helpers.

Frassle (approved):

Issue #6378, opened 2021-02-18
CreateSelectRemoveStack test is flaky


mikhailshilkov:
I had to [skip](https://github.com/pulumi/pulumi/blob/1ccfd7413e58bf4706b544bf8a014cf3bff60089/sdk/dotnet/Pulumi.Automation.Tests/LocalWorkspaceTests.cs#L104) the `CreateSelectRemoveStack` test because it failed on a subsequent run: https://github.com/pulumi/pulumi/runs/1926458771?check_suite_focus=true#step:15:508

```
X Pulumi.Automation.Tests.LocalWorkspaceTests.CreateSelectRemoveStack [1s 64ms]
  Error Message:
   Assert.Empty() Failure
Collection: [StackSummary { IsCurrent = False, IsUpdateInProgress = False, LastUpdate = null, Name = "int_test205572", ResourceCount = null, ... }, StackSummary { IsCurrent = False, IsUpdateInProgress = False, LastUpdate = null, Name = "int_test358983", ResourceCount = null, ... }, StackSummary { IsCurrent = False, IsUpdateInProgress = False, LastUpdate = null, Name = "int_test438318", ResourceCount = null, ... }, StackSummary { IsCurrent = False, IsUpdateInProgress = False, LastUpdate = null, Name = "int_test805274", ResourceCount = null, ... }, StackSummary { IsCurrent = False, IsUpdateInProgress = False, LastUpdate = null, Name = "int_test814210", ResourceCount = null, ... }]
  Stack Trace:
     at Pulumi.Automation.Tests.LocalWorkspaceTests.CreateSelectRemoveStack() in /home/runner/work/pulumi/pulumi/sdk/dotnet/Pulumi.Automation.Tests/LocalWorkspaceTests.cs:line 120
--- End of stack trace from previous location where exception was thrown ---
```

@orionstudt Could you take a look?

orionstudt:
I think #6382 should resolve that. Just need to not make assertions about the entire collection so that it doesn't break on existing stacks, like the plugin manipulation test.

PR #6382, opened 2021-02-18
[Automation API] - resolve stack manipulation test assertions

orionstudt:
By just not making assertions about the entire collection, as that will break if there are existing stacks.

Fix https://github.com/pulumi/pulumi/issues/6378


mikhailshilkov:
I added a label to not require a changelog - this is not a user-facing change

mikhailshilkov (approved):

PR #6387, opened 2021-02-19
[sdk/nodejs] Skip re-registrations of same package+version

### LINKED ISSUE #6258: Cannot re-register module aws:accessanalyzer/analyzer@3.27
When I run pulumi preview command I receive the follow message 
`Error: Cannot re-register module aws:accessanalyzer/analyzer/analayzer@3.27. Previous registration was [Object Object], new registration was [Object Object]
            at register (~/myproject/node_module/@pulumi/pulumi/runtime/rpc.js:549:23)
            at Object.registerResourceModule (~/myproject/node_module/@pulumi/pulumi/runtime/rpc.js:613:5)
            at Object.<anonymous> ~/myproject/node_modules/node_modules/@pulumi/accesanalyzer/index.ts:24:15)
           at Module._compile(internal/modules/cjs/loader.js:999:30)`

## Expected Behavior
Pulumi previous shuold be succesful

## Current Behavior
Error  Cannot re-register module aws:accessanalyzer/analyzer@3.27

lukehoban:
It is possible for the same version of the same provider SDK to be loaded multiple times in Node.js.  In this case, we might legitimately get mutliple registrations of the same resource.  It should not matter which we use, so we can just skip re-registering.  De-serialized resources will always be instances of classes from the first registered package.

Example layout this addresses.  Registrations of resources in `package3` at the same verrsion.

* `node_modules`
  *  `@pulumi/pulumi`
  * `package1`
    *  `node_modules`
       *  `package3`
  * `package2`
    * `node_modules`
      *  `package3`

Fixes #6258.

### DIFF
--- a/sdk/nodejs/tests/runtime/registrations.spec.ts
+++ b/sdk/nodejs/tests/runtime/registrations.spec.ts
@@ -28,15 +28,14 @@ describe("runtime", () => {
         describe("register", () => {
             const tests = [
                 { name: "wildcard version", version: undefined },
-                { name: "blank version", version: "" },
                 { name: "version", version: "1.2.3" },
                 { name: "alpha version", version: "1.0.0-alpha1" },
             ];
             for (const { name, version } of tests) {
-                it(`throws on same ${name}`, () => {
+                it(`ignores registration on same ${name}`, () => {
                     const source = new Map<string, runtime.ResourceModule[]>();
-                    runtime.register(source, rt, "test", { version, construct });
-                    assert.throws(() => runtime.register(source, rt, "test", { version, construct }));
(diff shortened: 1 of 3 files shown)


sdk/nodejs/tests/runtime/registrations.spec.ts
```diff
@@ -28,15 +28,14 @@ describe("runtime", () => {
         describe("register", () => {
             const tests = [
                 { name: "wildcard version", version: undefined },
-                { name: "blank version", version: "" },
```
lukehoban, line 31: This isn't a legal value at all - it happened to throw for an unrelated reason inside `sameVersion` - but that wasn't really what this test was trying to test.

justinvp (approved):
LGTM

pgavlin (approved):

PR #6397, opened 2021-02-22
serialize default parameters

jakzo:
```ts
const defaultValue = 123;
const func = (arg = defaultValue) => {};
const result = await require('@pulumi/pulumi').runtime.serializeFunction(func);
console.log(result.text);
```

Currently this will not serialise `defaultValue` and result in:

```js
function __f0() {
  return (function() {
    with({  }) {

return (arg = defaultValue) => {};

    }
  }).apply(undefined, undefined).apply(this, arguments);
}
```

This PR fixes this by visiting the default parameter AST nodes when visiting a function declaration node.


komalali:
@jakzo thanks for the PR! I ran the tests and everything looks good - would you mind fixing up the CHANGELOG? Note: we are now using the [CHANGELOG_PENDING](https://github.com/pulumi/pulumi/blob/master/CHANGELOG_PENDING.md) file for new entries.

komalali (approved):

PR #6398, opened 2021-02-22
[codegen/dotnet] Fixes to version.txt handling

justinvp:
We currently include `version.txt` in the package via `<Content Include="version.txt">`. This places the file in the NuGet package in two locations: the `content` directory and `contentFiles`.

We want the file to be in `content` because this is where the Pulumi .NET Language host looks for the file when determining a program's required plugins.

However, having the file in `contentFiles` is problematic. For packages that reference other Pulumi resource packages (e.g. for multi-lang components), referenced `contentFiles` are included in the package by default. This means another package's `version.txt` will be included and used over the current project's `version.txt`. For example, if a multi-lang component package `Pulumi.Xyz` references `Pulumi.Aws`, the `version.txt` from `Pulumi.Aws` will be included in `Pulumi.Xyz` package rather than the intended `version.txt` for `Pulumi.Xyz`.

To address this, stop including `version.txt` in `contentFiles`. We do this by changing the `<Content Include="version.txt" />` to `<None Include="version.txt" Pack="True" PackagePath="content" />` in the project file. This ensures the file will still be included in the package in the `content` dir (where the Pulumi .NET language host expects to find it), but doesn't include the file in `contentFiles` which would cause it to be used in other packages that reference the package.

Further, when generating `<PackageReference>`s for packages that start with "Pulumi.", include an additional `ExcludeAssets="contentFiles"` attribute to prevent the package's `contentFiles` from being included. This prevents the `version.txt` in `contentFiles` of referenced existing/older packages from being used by the package.

Part of https://github.com/pulumi/pulumi-eks/issues/529

### DIFF
--- a/pkg/codegen/dotnet/templates.go
+++ b/pkg/codegen/dotnet/templates.go
@@ -16,6 +16,7 @@
 package dotnet
 
 import (
+	"strings"
 	"text/template"
 
 	"github.com/pulumi/pulumi/pkg/v2/codegen/schema"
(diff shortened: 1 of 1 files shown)


mikhailshilkov (approved):
This is awesome! I've been meaning to figure out how to avoid exposing `version.txt` to downstream packages for months. Thank you for fixing that!

PR #6404, opened 2021-02-22
Upgrading Go to 1.16.x

stack72:
This unlocks the ability to start creating arm64 builds


CHANGELOG.md
```diff
@@ -2,7 +2,9 @@ CHANGELOG
 =========
 
 ## HEAD (Unreleased)
-_(none)_
+
+- [chore] Updating Pulumi to use Go 1.16
```
justinvp, line 6: Nit:

```suggestion
- Updated Pulumi to use Go 1.16
```

justinvp (approved):
https://golang.org/doc/go1.16#go-command

> the GO111MODULE environment variable now defaults to on.

Would it be worth cleaning up any uses of `GO111MODULE` now that we're on 1.16?


https://golang.org/doc/go1.16#ioutil

> The io/ioutil package has turned out to be a poorly defined and hard to understand collection of things. All functionality provided by the package has been moved to other packages. The io/ioutil package remains and will continue to work as before, but we encourage new code to use the new definitions in the io and os packages.

Not necessary for this PR, but we should consider moving off `io/ioutil` at some point.

Issue #5630, opened 2020-10-27
`PreviewResult.summary` has wrong type in ts Pulumi Automation API


DSoko2:
Using the automation API with typescript, the type of `summary` in `PreviewResult` is `UpdateSummary`. However, when running `preview()` on a new stack the returned `PreviewResult` object's `summary` property is `undefined`.

## Expected Behavior

With this type signature, the `summary` is expected to be an `UpdateSummary` object ; it should never be `undefined`.

## Current Behavior

`summary` is undefined.

## Steps to Reproduce
1. Take any basic Pulumi API ts project
2. Run as first action on a new stack `preview()`
3. Await and observe the returned `PreviewResult` object; its summary property is `undefined`

## Context (Environment)

I want to use Pulumi Automation API's `stack.preview()` to evaluate whether any update of the stack would be performed on the next `stack.up()`. Seeing the type signature of PreviewReslut and the lack of documentation, I would assume `summary` describes the changes that would be performed to update the stack to the programmed state. However, it is `undefined`, which is an invalid value according to the `PreviewResult` type signature.

Looking into the code, I figured that pulumi automation API is using `pulumi history` of the CLI to get the information for `summary`. However, pulumi history does not correspond to previews, as it only shows performed updates. Accordingly, it does not return anything before the stack was ever updated (e.g. through up) and afterwards it also reflects the changes performed through the last update and *not* of the last preview action. With this implementation, I assume a correct fix would be to remove the current `summary` completely from `PreviewResult`. However, a properly working and correct mechanism to find out whether (and what?) changes would be performed wiht the next `stack.up()` would be great.

Issue #6308, opened 2021-02-10
[Automation API] remove --json preview output from Go SDK


EvanBoyle:
Go is the only SDK that uses structured preview output. This prevents streaming preview output from STDOUT as you can do with other languages. 

We're going to make the SDKs consistent by remove the usage of `--json` for `stack.preview()` from the Go SDK Automation API. This will be a breaking change. We will follow up with support for structured output in a more official capacity a la plans via:

https://github.com/pulumi/pulumi/issues/5218
https://github.com/pulumi/pulumi/issues/5532

PR #6405, opened 2021-02-23
[automation] - Remove summary from PreviewResult

komalali:
The current implementation in nodejs, python and dotnet attempts to get an action summary for `preview` by running `pulumi history --page-size 1`. However, `preview` summaries are not pushed to `history` because (you guessed it!) it's just a preview and not an actually executed command. As a result, the `summary` ends up being either nothing (before an `up` has ever been run) or the summary of the last executed command of `up/refresh/destroy`. In either case, it is an incorrect result.

This PR removes the `summary` property from `PreviewResult`. 

Notes:
* It might be preferable to wait until structured output is implemented and we can access the preview summary and use that here rather than removing the summary property.
* Go will be handled in a separate PR as it has a different implementation and a [separate issue](https://github.com/pulumi/pulumi/issues/6308) relating to it.

Fixes: https://github.com/pulumi/pulumi/issues/5630

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -5,3 +5,7 @@
 
 
 ### Bug Fixes
+
+- [automation/python,nodejs,dotnet] - BREAKING - Remove `summary` property from `PreviewResult`.
+  The `summary` property on `PreviewResult` returns a result that is always incorrect and is being removed.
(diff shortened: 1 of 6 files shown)


EvanBoyle (approved):
I'm fine with removing this for now and updating the API again after we have structured output.

Issue #5257, opened 2020-08-29
Consider standardizing stack selection behavior


EvanBoyle:
Pulumi commands accept a `--stack` parameter to use a stack other than the currently selected one, but there is a mix of behavior with this command. Sometimes we set the stack in the workspace when `-s` is passed:

https://github.com/pulumi/pulumi/search?q=%22true+%2F*setCurrent*%2F%22&unscoped_q=%22true+%2F*setCurrent*%2F%22

Sometimes we don't: 

https://github.com/pulumi/pulumi/search?q=%22false+%2F*setCurrent*%2F%22&unscoped_q=%22false+%2F*setCurrent*%2F%22

I think there are two options to rectify this:

1. Take a breaking change in 3.0 and standardize all commands to setStack=false.
2. Add a new flag to be used in combination with --stack: --set-stack=false

komalali:
What's the preferred solution here @EvanBoyle? Do we want the extra flag or just standardize to false?

lukehoban:
From an offline conversation:

> I see... super strange... the entire point of `-s` was to not need to set the global stack for the project, so feels like that's probably a bug that was copied far and wide :disappointed:

Let's fix the meaning of the existing `-s` to not set the stack.  Users can easily set the stack via `pulumi stack select` if needed.  

This was marked as `impact/breaking` because it is (technically - though probably not significantly in practice) a breaking change.  So we will likely need to fix this in a `3.0` branch instead of in `master`.

Issue #6062, opened 2021-01-06
[Automation API] all operations should use `-s` to specify stack rather than modifying global state (select)


EvanBoyle:
Automation API operations always select a stack before performing the operation. This leads to problems when doing things like concurrently settings config values on multiple stacks in the same local workspace. To avoid this category of error altogether, we should be able to avoid selecting stacks and instead use the `-s` global flag. The only problem is that the `-s` flag has inconsistent behavior which we need to think about addressing first: https://github.com/pulumi/pulumi/issues/5257

Will need to be fixed across SDKs:
- [x] Node
- [x] Go
- [x] Python
- [x] C#

PR #6412, opened 2021-02-23
[BREAKING] - Standardize stack select behavior

stack72:
This PR has 2 purposes:

Standardize the --stack flag in the CLI to setCurrent=False (this is a breaking change and we will keep it in a 3.0 branch)
Change the automation API implementation in each language to use the --stack flag for stack operations rather than first running selectStack, followed by the operation in question.
Note: I would like to get this PR reviewed but will hold off on merging this until we have our 3.0 branch figured out.

Fixes: #5257 and #6062

sorry @komalali - I merged and reverted the old version to make sure our builds were working! 

No functionality has changed at all here

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -4,6 +4,7 @@ on:
       [ "master", "feature/**", "feature-**" ]
     paths-ignore:
       - 'CHANGELOG.md'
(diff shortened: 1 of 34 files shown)


komalali:
My pushes to this branch are no longer working so I'll close this and re-open

PR #6413, opened 2021-02-24
[automation] - Add diff to up and preview in automation API

bob-bins:


CONTRIBUTING.md
```diff
@@ -22,7 +22,9 @@ You can easily get all required dependencies with brew and npm
 
 ```bash
 brew install node pipenv python@3 typescript yarn go@1.13 golangci/tap/golangci-lint pulumi/tap/pulumictl
-brew cask install dotnet dotnet-sdk
+curl https://raw.githubusercontent.com/Homebrew/homebrew-cask/0272f0d33f/Casks/dotnet-sdk.rb > dotnet-sdk.rb  # v3.1.0
+brew install --HEAD -s dotnet-sdk.rb
+rm dotnet-sdk.rb
```
bob-bins, line 27: `brew cask install dotnet dotnet-sdk` is not valid for two reasons: the packages conflict with each other and the latest version of dotnet-sdk is not compatible with the tests.

brew is picky and requires that the filename be the same as the package name, which is why i couldn't just pipe the output from `curl` into `brew install`.

komalali:
Thanks so much for the PR @bob-bins, this is awesome! Extra gratitude for implementing it across languages 🙏

komalali (approved):

PR #6415, opened 2021-02-24
[BREAKING] - Standardize stack select behavior

komalali:
This PR has 2 purposes:

* Standardize the `--stack` flag in the CLI to `setCurrent=False` (this is a breaking change and we will keep it in a 3.0 branch)
* Change the automation API implementation in each language to use the `--stack` flag for stack operations rather than first running `selectStack`, followed by the operation in question.

Also added a test that previously failed to correctly set config values across stacks asyncronously in nodejs.

Fixes: #5257 and #6062

### DIFF
--- a/sdk/go/x/auto/stack.go
+++ b/sdk/go/x/auto/stack.go
@@ -236,7 +236,7 @@ func (s *Stack) Preview(ctx context.Context, opts ...optpreview.Option) (Preview
 		sharedArgs = append(sharedArgs, fmt.Sprintf("--parallel=%d", preOpts.Parallel))
 	}
 
-	kind, args := constant.ExecKindAutoLocal, []string{"preview", "--json", "--stack", s.Name()}
+	kind, args := constant.ExecKindAutoLocal, []string{"preview", "--json"}
 	if program := s.Workspace().Program(); program != nil {
 		server, err := startLanguageRuntimeServer(program)
 		if err != nil {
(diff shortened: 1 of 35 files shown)


sdk/go/x/auto/stack.go
```diff
@@ -471,22 +451,17 @@ func (s *Stack) Destroy(ctx context.Context, opts ...optdestroy.Option) (Destroy
...
 func (s *Stack) Outputs(ctx context.Context) (OutputMap, error) {
-	err := s.Workspace().SelectStack(ctx, s.Name())
-	if err != nil {
-		return nil, errors.Wrap(err, "failed to get stack outputs")
-	}
-
 	// standard outputs
 	outStdout, outStderr, code, err := s.runPulumiCmdSync(ctx, nil, /* additionalOutputs */
-		"stack", "output", "--json",
+		"stack", "output", "--json", "--stack", s.Name(),
```
EvanBoyle, line 456: I think this applies in all languages, but we could probably set `--stack` consistently from within `stack.runPulumiCmdSync`. Then it would be enforced for every command run on the stack including future functionality that gets added. 

https://github.com/pulumi/pulumi/blob/bcd14d806046a562218e074d1f7872ec6ecf2409/sdk/go/x/auto/stack.go#L742

I believe that this should work. Every stack command should require this flag. Those commands that don't likely belong on workspace instead of stack (pulumi version for instance).

EvanBoyle (approved):
🎉 

One suggestion, but looks good to me!

sdk/go/x/auto/stack.go
komalali: Ah yes, good idea :)

PR #6416, opened 2021-02-24
[automation/go] - Remove --json flag from preview

komalali:
The `--json` flag prevents us from being able to stream incremental updates via `ProgressStreams` for Stack.Preview(), so we are removing it and changing the shape of the `PreviewResult` to match the other language implementations and the other commands.

This is a breaking change because we are changing the shape of `PreviewResult`, but this is to be expected while automation api is in preview.

Fixes: #6308


komalali:
Not going to do this until we have structured output.

Issue #6402, opened 2021-02-22
win10 PermissionError on create_or_select_stack


ob-server83:
pulumi version
v2.21.1
pulumi	2.21.2	2.21.2
pulumi-aws	3.30.0	3.30.0

When running example code from https://github.com/pulumi/automation-api-examples/blob/main/python/inline_program/main.py on Windows10

## Expected behavior
not getting this error.

## Current behavior
im getting this error:
```
D:\GIT\GIT\git_lab\pre-prod_system_test\pulumi\PyClient>python myPulumiClient.py
Traceback (most recent call last):
  File "D:\GIT\GIT\git_lab\pre-prod_system_test\pulumi\PyClient\myPulumiClient.py", line 61, in <module>
    stack = auto.create_or_select_stack(stack_name=stack_name,
  File "C:\Python\Python391\lib\site-packages\pulumi\x\automation\local_workspace.py", line 398, in create_or_select_stack
    return _inline_source_stack_helper(stack_name, program, project_name, Stack.create_or_select, opts)  # type: ignore
  File "C:\Python\Python391\lib\site-packages\pulumi\x\automation\local_workspace.py", line 416, in _inline_source_stack_helper
    return init_fn(stack_name, ws)
  File "C:\Python\Python391\lib\site-packages\pulumi\x\automation\stack.py", line 188, in create_or_select
    return Stack(stack_name, workspace, StackInitMode.CREATE_OR_SELECT)
  File "C:\Python\Python391\lib\site-packages\pulumi\x\automation\stack.py", line 214, in __init__
    workspace.create_stack(name)
  File "C:\Python\Python391\lib\site-packages\pulumi\x\automation\local_workspace.py", line 205, in create_stack
    self._run_pulumi_cmd_sync(args)
  File "C:\Python\Python391\lib\site-packages\pulumi\x\automation\local_workspace.py", line 276, in _run_pulumi_cmd_sync
    return _run_pulumi_cmd(args, self.work_dir, envs, on_output)
  File "C:\Python\Python391\lib\site-packages\pulumi\x\automation\cmd.py", line 75, in _run_pulumi_cmd
    os.remove(stderr_file.name) <<<<
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\romlaz\\AppData\\Local\\Temp\\tmp8o_qs82u'
```

[...]

Issue #6417, opened 2021-02-24
Run SDK unit tests in the Windows build/test pipeline


komalali:
We don't currently run our language-specific SDK unit tests in our Windows build+test pipeline. We should be doing this as we may be missing Windows bugs in our SDKs.

mikhailshilkov:
I think this does not hold anymore, so I'll close the issue

PR #6421, opened 2021-02-24
[automation/python] - Use TemporaryFile and seek()

komalali:
This PR fixes the usage of `TemporaryFile` in python automation API. Particularly in Windows, `NamedTemporaryFile` [cannot be reopened](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile:~:text=Whether%20the%20name%20can%20be%20used,cannot%20on%20Windows%20NT%20or%20later) while it is already open, which was leading to errors. We don't need a `NamedTemporaryFile` nor did we need to reopen it, we just needed `seek` to get back to the beginning of the file before we read it.

We don't run SDK unit tests in our Windows build which is why this wasn't caught, but I've opened an [issue](https://github.com/pulumi/pulumi/issues/6417) to track that work.

https://github.com/pulumi/pulumi/pull/6424 has an example of the [repro](https://github.com/pulumi/pulumi/pull/6424/checks?check_run_id=1974871868#step:14:215) and a rerun with this change included that [passes](https://github.com/pulumi/pulumi/pull/6424/checks?check_run_id=1975009417#step:14:489).

Fixes: #6402

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -10,3 +10,6 @@
 - [automation/python,nodejs,dotnet] - BREAKING - Remove `summary` property from `PreviewResult`.
   The `summary` property on `PreviewResult` returns a result that is always incorrect and is being removed.
   [#6405](https://github.com/pulumi/pulumi/pull/6405)
+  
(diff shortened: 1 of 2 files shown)


EvanBoyle (approved):
🎉 🐍 🚢

PR #6424, opened 2021-02-25
Enable python automation API tests on Windows

komalali:
Realized while fixing https://github.com/pulumi/pulumi/pull/6421 that we don't run SDK unit tests on Windows, so automation API tests were not running.


stack72 (approved):
This is awesome :) we can talk about getting the other tests in piece by piece

PR #6430, opened 2021-02-26
[codegen] Discriminated unions in programgen

### LINKED ISSUE #6392: [schema] Introduce discriminators
Add the [OpenAPI discriminator](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#discriminatorObject) field to `schema.TypeSpec`.

They will be used by Azure NextGen for discriminated object types, and then in the program's conversion rewriter as discussed in https://github.com/pulumi/pulumi/pull/6345#discussion_r577824327

mikhailshilkov:
Resolves #6392
Needed for https://github.com/pulumi/pulumi-azure-native/issues/549

The scenario that this PR tries to address is when a resource property is a union of several object types. Each object type has a discriminator: a property `Foo` with a const value. Each type has its own const value for that property. See the `azure-native.json` for real-life examples (it's a small subset of the full Azure Native spec).

Two large-ish changes here:

1. Adjust to `RewriteConversions` to exclude object types that don't match the discriminator value of particular input in this program. The program gen can then use the only remaining union case as the schema type for property initialization.
2. Introduce a new HCL type to hold a const value. This is needed to prevent collapsing multiple types into one when the union is calculated. Those multiple types have exactly the same shape but are only different in the discriminator value.

I have a draft in Azure-Native that fills the discriminator properties and it seems to be working nicely: https://github.com/pulumi/pulumi-azure-native/pull/802

### DIFF
--- a/pkg/codegen/go/gen_program_test.go
+++ b/pkg/codegen/go/gen_program_test.go
@@ -28,6 +28,11 @@ func TestGenProgram(t *testing.T) {
 			continue
 		}
 
+		if filepath.Base(f.Name()) == "azure-native.pp" {
+			// The generated code fails to compile
+			continue
+		}
+
 		t.Run(f.Name(), func(t *testing.T) {
 			path := filepath.Join(testdataPath, f.Name())
 			contents, err := ioutil.ReadFile(path)
--- a/pkg/codegen/hcl2/binder_schema.go
+++ b/pkg/codegen/hcl2/binder_schema.go
@@ -171,6 +171,9 @@ func (b *binder) schemaTypeToTypeImpl(src schema.Type, seen map[schema.Type]mode
 			if !prop.IsRequired {
 				t = model.NewOptionalType(t)
 			}
+			if prop.ConstValue != nil {
+				t = model.NewConstType(t, prop.ConstValue)
+			}
 			properties[prop.Name] = t
 		}
 		return objType
@@ -191,6 +194,9 @@ func (b *binder) schemaTypeToTypeImpl(src schema.Type, seen map[schema.Type]mode
 		types := make([]model.Type, len(src.ElementTypes))
 		for i, src := range src.ElementTypes {
 			types[i] = b.schemaTypeToTypeImpl(src, seen)
+		}
+		if src.Discriminator != "" {
+			return model.NewUnionTypeAnnotated(types, src)
 		}
 		return model.NewUnionType(types...)
 	default:
@@ -249,6 +255,11 @@ func GetSchemaForType(t model.Type) (schema.Type, bool) {
 	case *model.PromiseType:
 		return GetSchemaForType(t.ElementType)
 	case *model.UnionType:
+		for _, a := range t.Annotations {
+			if t, ok := a.(*schema.UnionType); ok {
+				return t, true
+			}
+		}
 		schemas := codegen.Set{}
 		for _, t := range t.ElementTypes {
 			if s, ok := GetSchemaForType(t); ok {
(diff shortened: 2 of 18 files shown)


pkg/codegen/hcl2/rewrite_convert.go
```diff
@@ -90,6 +90,9 @@ func rewriteConversions(x model.Expression, to model.Type) (model.Expression, bo
 	case *model.IndexExpression:
 		x.Key, _ = rewriteConversions(x.Key, x.KeyType())
 	case *model.ObjectConsExpression:
+		if v := resolveDiscriminatedUnions(to, x); v != nil {
```
mikhailshilkov, line 93: Hmm... I think this is the wrong spot to do that. E.g., I see that I miss conversions for tuples containing objects. I'll try to think of a better one - let me know if you know the answer!

pgavlin:
The current changes look like a good direction to me. The only thing I'd want to triple-check is that annotations on unions don't affect convertibility and are propagated as necessary.

pkg/codegen/go/gen_program_test.go
```diff
@@ -28,6 +28,11 @@ func TestGenProgram(t *testing.T) {
 			continue
 		}
 
+		if filepath.Base(f.Name()) == "azure-native.pp" {
```
mikhailshilkov, line 31: This Go program generation already fails in master. This PR doesn't fix it but it doesn't make it worse either, so I suggest filing a separate issue for later.

pkg/codegen/hcl2/model/type_object.go
```diff
@@ -110,6 +111,19 @@ func (t *ObjectType) equals(other Type, seen map[Type]struct{}) bool {
 			return false
 		}
 	}
+	if len(t.Annotations) != len(otherObject.Annotations) {
```
mikhailshilkov, line 114: Some discriminated union types have a shape `A | B` where both `A` and `B` have exactly the same property names. Currently, these types are smashed into one when a union is built, which is wrong.

If we don't want to account for annotations in the equality check, we'd probably need to re-introduce the `Const` type from my other PR, because that's the only difference between two types.

mikhailshilkov:
@pgavlin I believe this is ready for review.

No downstream changes, as far as I can see, which is expected. It should only affect Azure-Native once it starts generating discriminated union mappings.

lblackstone (approved):

pkg/codegen/hcl2/binder_schema.go
```diff
@@ -276,3 +287,32 @@ func GetSchemaForType(t model.Type) (schema.Type, bool) {
...
+
+func getDiscriminatedUnionObjectItem(t model.Type) (string, model.Type) {
+	switch t := t.(type) {
+	case *model.ListType:
+		return getDiscriminatedUnionObjectItem(t.ElementType)
+	case *model.ObjectType:
+		if bla, ok := GetSchemaForType(t); ok {
+			if foo, ok := bla.(*schema.ObjectType); ok {
+				return foo.Token, t
+			}
```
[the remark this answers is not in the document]
mikhailshilkov: Shame on me :) I was going through a refactoring and forgot to clean up. Will do!

pkg/codegen/internal/test/host.go
```diff
@@ -14,6 +14,9 @@ func NewHost(schemaDirectoryPath string) plugin.Host {
 		deploytest.NewProviderLoader("azure", semver.MustParse("3.24.0"), func() (plugin.Provider, error) {
 			return Azure(schemaDirectoryPath)
 		}),
+		deploytest.NewProviderLoader("azure-native", semver.MustParse("0.9.0"), func() (plugin.Provider, error) {
```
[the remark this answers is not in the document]
mikhailshilkov: It doesn't matter for tests right now but I will update

pkg/codegen/go/gen_program_test.go
pgavlin: SGTM

pkg/codegen/hcl2/model/type_const.go
```diff
@@ -0,0 +1,97 @@
...
+func (t *ConstType) equals(other Type, seen map[Type]struct{}) bool {
+	if t == other {
+		return true
+	}
+
+	otherConst, ok := other.(*ConstType)
+	return ok && t.Value == otherConst.Value && t.Type.equals(otherConst.Type, seen)
+}
+
+// AssignableFrom returns true if this type is assignable from the indicated source type. A token(name) is assignable
```
pgavlin, line 63: Could you file an issue for follow-up to expand convertibility and assignability for these types? E.g. a constant string should be convertible to a string (which requires changes in the core convertibility rules).

Also, `s/token(name)/const(value)/g` :)

pgavlin (approved):

pkg/codegen/hcl2/model/type_const.go
mikhailshilkov: https://github.com/pulumi/pulumi/issues/6967

PR #6436, opened 2021-03-01
[automation/go] - Expose structured logging

komalali:
See also: https://github.com/pulumi/pulumi/pull/6454 for the nodejs implementation.

### Changes

Adds the ability to pass in channels to receive raw events from the Pulumi event stream via an option passed to the `Preview`, `Up`, `Refresh` and `Destroy` commands.

This change is marked **breaking** because it changes the shape of the `PreviewResult` struct.

Before:
```go
type PreviewResult struct {
	Steps         []PreviewStep  `json:"steps"`
	ChangeSummary map[string]int `json:"changeSummary"`
}
```

After:
```go
type PreviewResult struct {
	StdOut        string
	StdErr        string
	ChangeSummary map[apitype.OpType]int
}
```

Fixes: #6441, #6308

### DIFF
--- a/sdk/go/x/auto/stack.go
+++ b/sdk/go/x/auto/stack.go
@@ -263,7 +263,11 @@ func (s *Stack) Preview(ctx context.Context, opts ...optpreview.Option) (Preview
 	eventChannel := make(chan apitype.EngineEvent)
 	go collectEvents(eventChannel, &events)
 
-	t, err := tailLogs("preview", preOpts.EventStreams, eventChannel)
+	eventChannels := []chan<- apitype.EngineEvent{eventChannel}
+	if len(preOpts.EventStreams) > 0 {
+		eventChannels = append(eventChannels, preOpts.EventStreams...)
+	}
+	t, err := tailLogs("preview", eventChannels)
 	if err != nil {
 		return res, errors.Wrap(err, "failed to tail logs")
 	}
@@ -288,6 +292,7 @@ func (s *Stack) Preview(ctx context.Context, opts ...optpreview.Option) (Preview
 	res.StdOut = stdout
 	res.StdErr = stderr
 	res.ChangeSummary = previewSummary.ResourceChanges
+	res.EventLog = events
 
 	return res, nil
 }
@@ -347,7 +352,11 @@ func (s *Stack) Up(ctx context.Context, opts ...optup.Option) (UpResult, error) 
 	eventChannel := make(chan apitype.EngineEvent)
 	go collectEvents(eventChannel, &events)
 
-	t, err := tailLogs("up", upOpts.EventStreams, eventChannel)
+	eventChannels := []chan<- apitype.EngineEvent{eventChannel}
+	if len(upOpts.EventStreams) > 0 {
+		eventChannels = append(eventChannels, upOpts.EventStreams...)
+	}
+	t, err := tailLogs("up", eventChannels)
 	if err != nil {
 		return res, errors.Wrap(err, "failed to tail logs")
 	}
@@ -425,7 +434,11 @@ func (s *Stack) Refresh(ctx context.Context, opts ...optrefresh.Option) (Refresh
 	eventChannel := make(chan apitype.EngineEvent)
 	go collectEvents(eventChannel, &events)
 
-	t, err := tailLogs("refresh", refreshOpts.EventStreams, eventChannel)
+	eventChannels := []chan<- apitype.EngineEvent{eventChannel}
+	if len(refreshOpts.EventStreams) > 0 {
+		eventChannels = append(eventChannels, refreshOpts.EventStreams...)
+	}
+	t, err := tailLogs("refresh", eventChannels)
 	if err != nil {
 		return res, errors.Wrap(err, "failed to tail logs")
 	}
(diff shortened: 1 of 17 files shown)


sdk/go/x/auto/watcher.go
```diff
@@ -0,0 +1,41 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package auto
+
+import (
+	"encoding/json"
+	"github.com/nxadm/tail"
```
viveklak, line 18: Nice find!

sdk/go/x/auto/stack.go
```diff
@@ -751,6 +830,8 @@ func (s *Stack) runPulumiCmdSync(
 	args ...string,
 ) (string, string, int, error) {
 	var env []string
+	debugEnv := fmt.Sprintf("%s=%s", "PULUMI_DEBUG_COMMANDS", "true")
```
viveklak, line 833: Was this intended?

sdk/go/x/auto/stack.go
```diff
@@ -722,24 +799,26 @@ type PropertyDiff struct {
 
 // PreviewResult is the output of Stack.Preview() describing the expected set of changes from the next Stack.Up()
 type PreviewResult struct {
-	Steps         []PreviewStep  `json:"steps"`
-	ChangeSummary map[string]int `json:"changeSummary"`
+	StdOut        string
+	StdErr        string
+	ChangeSummary map[string]int
+	EventLog      []apitype.EngineEvent
```
viveklak, line 805: I think you probably have the tradeoff considered appropriately here but since you can pass event channels in the options anyway, do you think getting a full copy of the events in return as well seems a bit heavyweight?

sdk/go/x/auto/watcher.go
```diff
@@ -0,0 +1,41 @@
...
+	t, err := tail.TailFile(path, tail.Config{
+		Follow: true,
+		Logger: tail.DiscardingLogger,
+	})
+	if err != nil {
+		return nil, err
+	}
+	go func(tailedLog *tail.Tail) {
+		for line := range tailedLog.Lines {
+			for _, r := range receivers {
```
viveklak, line 33: Not too familiar with the api but it seems the line could have an `Err` as well? Do we want to handle that?

sdk/go/x/auto/stack.go
komalali: Yeah, the `--event-log` flag is only accessible if the debug env var is present. We discussed removing that requirement but have decided to keep it for now because we will (probably at some point) implement similar functionality via a `--json` flag and didn't want to derail this work by conflating the two.

sdk/go/x/auto/stack.go
komalali: Yeah, this is the thing that I am most on the fence about. The [nodejs PR description](https://github.com/pulumi/pulumi/pull/6454#issue-583398443) best encapsulates my hesitation here. I agree that the full copy is overkill, what I really want is just the resource events and only really in the `Preview` case when there isn't an `UpdateSummary`. But it's probably "good enough" that you can get that from the event channel should you want it.

komalali: Actually, thinking about it more I'm inclined to leave it in since we do include stdout and stderr on the result object (even though users could have the same access via `ProgressStreams`). Including `EventLog` on the `Result` object would be most consistent.

sdk/go/x/auto/watcher.go
komalali: Done in latest commit :)

sdk/go/x/auto/stack.go
komalali: Nope, changed my mind again. `EventLog` is too noisy, gonna exclude it from the `Result` and leave it as an opt-in via the channels.

viveklak (approved):

viveklak:
LGTM but probably want to get @EvanBoyle to sign off.

sdk/go/x/auto/stack.go
```diff
@@ -924,3 +989,25 @@ func (s *languageRuntimeServer) GetPluginInfo(ctx context.Context, req *pbempty.
 		Version: "1.0.0",
 	}, nil
 }
+
+func tailLogs(command string, receivers []chan<- events.EngineEvent) (*tail.Tail, error) {
+	logDir, err := ioutil.TempDir(os.TempDir(), fmt.Sprintf("automation-logs-%s-", command))
```
EvanBoyle, line 994: Do you think we should clean up the temp file on behalf of the user? Looks like we could make that addition to the cleanup function before without too much trouble. 

We are opting them into this behavior on preview since we rely on it for even summary counts. Just in terms of limiting surprises.

EvanBoyle (approved):
Excellent work here 🎉 🚢  Really nice to see the test coverage here as well. 

One small question, but looks good to ship.

sdk/go/x/auto/stack.go
komalali: Yeah, good call. Should do for nodejs as well. Also learned that if you just pass in an empty string as the first arg to `ioutil.TempDir` it automatically uses `os.TempDir()`

EvanBoyle (approved):
🔥

PR #2697, opened 2019-05-02
Support locking on filestate logins

bigkraig:
When using the filestate backend (local files and cloud buckets) there is no protection to prevent two processes from managing the same stack simultaneously.

This PR creates a `locks` directory in the management directory that stores lock files for a stack. Each backend implementation gets its own UUID that is joined with the stack name.  The feature is currently available behind the `PULUMI_SELF_MANAGED_STATE_LOCKING=1` environment variable flag.

@Place1 This follows the distributed locking idea you had in #2455.

PR #6437, opened 2021-03-01
Locking for self-managed backends

lukehoban:
Adds opt-in state locking for self-managed backends.  Builds on the work from @bigkraig in https://github.com/pulumi/pulumi/pull/2697. 

Deltas in this PR:
* Rebased that PR onto current Pulumi backend support
* Refactored to make locking part of the core filestate backend interface
* Added tests for local filesystem backend 
* Moved locking support behind an environment variable so that it is opt-in initially, since it has non-trivial performance implications


lukehoban:
Merged this work via @2697.

PR #6454, opened 2021-03-02
[automation/nodejs] - Expose structured logging

komalali:
See also: #6436 for the go implementation.

### Changes

* Adds an `onEvent` option to `Stack.up()`, `Stack.preview()`, `Stack.refresh()` and `Stack.destroy()` for acting on structured json events.
* Updates the `PreviewResult` interface to include a `changeSummary` property that summarizes changes to resources.

Fixes: #6442

### DIFF
--- a/sdk/nodejs/x/automation/events.ts
+++ b/sdk/nodejs/x/automation/events.ts
@@ -1,0 +1,196 @@
+// Copyright 2016-2020, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// NOTE: The interfaces in this file are intended to align with the serialized
+// JSON types defined and versioned in sdk/go/common/apitype/events.go
+
+// CancelEvent is emitted when the user initiates a cancellation of the update in progress, or
+// the update successfully completes.
+export interface CancelEvent {
+}
+
+// StdoutEngineEvent is emitted whenever a generic message is written, for example warnings
+// from the pulumi CLI itself. Less common that DiagnosticEvent
+export interface StdoutEngineEvent {
+    message: string;
+    color: string;
+}
+
+// DiagnosticEvent is emitted whenever a diagnostic message is provided, for example errors from
+// a cloud resource provider while trying to create or update a resource.
+export interface DiagnosticEvent {
+    urn?: string;
+    prefix?: string;
+    message: string;
+    color: string;
+    severity: string;
+    streamID?: number;
+    ephemeral?: boolean;
+}
+
+// PolicyEvent is emitted whenever there is Policy violation.
+export interface PolicyEvent {
+    resourceUrn?: string;
+    message: string;
+    color: string;
+    policyName: string;
+    policyPackName: string;
+    policyPackVersion: string;
+    policyPackVersionTag: string;
+    enforcementLevel: string;
+}
+
+// PreludeEvent is emitted at the start of an update.
+export interface PreludeEvent {
+    config: Record<string, string>;
+}
+
+// SummaryEvent is emitted at the end of an update, with a summary of the changes made.
+export interface SummaryEvent {
+    // MaybeCorrupt is set if one or more of the resources is in an invalid state.
+    maybeCorrupt: boolean;
+    // Duration is the number of seconds the update was executing.
+    durationSeconds: number;
+    // ResourceChanges contains the count for resource change by type. The keys are deploy.StepOp,
+    // which is not exported in this package.
+    resourceChanges: Record<string, number>;
+    // PolicyPacks run during update. Maps PolicyPackName -> version.
+    // Note: When this field was initially added, we forgot to add the JSON tag
+    // and are now locked into to using PascalCase for this field to maintain backwards
+    // compatibility. For older clients this will map to the version, while for newer ones
+    // it will be the version tag prepended with "v".
+    policyPacks: Record<string, string>;
+}
+
+export enum DiffKind {
+    // DiffAdd indicates that the property was added.
+    "add",
+    // DiffAddReplace indicates that the property was added and requires that the resource be replaced.
+    "add-replace",
+    // DiffDelete indicates that the property was deleted.
+    "delete",
+    // DiffDeleteReplace indicates that the property was deleted and requires that the resource be replaced.
+    "delete-replace",
+    // DiffUpdate indicates that the property was updated.
+    "update",
+    // DiffUpdateReplace indicates that the property was updated and requires that the resource be replaced.
+    "update-replace",
+}
+
+// PropertyDiff describes the difference between a single property's old and new values.
+export interface PropertyDiff  {
+    // Kind is the kind of difference.
+    diffKind: DiffKind;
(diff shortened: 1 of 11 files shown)


sdk/nodejs/x/automation/stack.ts
```diff
@@ -104,6 +111,39 @@ export class Stack {
...
+            })
+            .start()
+            .catch(err => {
+                throw new Error(`Cannot start, does the file exist? ${err}`);
+            });
+
+        eventLogTail
+            .pipe(split2())
+            .on("data", line => {
+                const event = JSON.parse(line);
```
lukehoban, line 141: Very minor - but might be worth a type annotation here to make clear what type we are expecting to get back from the JSON.parse.

sdk/nodejs/x/automation/stack.ts
```diff
@@ -104,6 +111,39 @@ export class Stack {
...
+            fileExists = fs.existsSync(path);
+            n++;
+        }
+        const eventLogTail = new TailFile(path, { startPos: 0 });
+        eventLogTail
+            .on("tail_error", (err) => {
+                console.error("TailFile had an error!", err);
+                throw err;
+            })
+            .start()
```
lukehoban, line 133: It may not be a big win - but looking at the docs for `start()`, it appears it (a) returns a promise and (b) rejects that promise only if the file doesn't exist yet.  That would allow simplifying this function a fair bit - and using the call to `start` as the thing that is retried in a loop instead of `fs.existsSync`.  It would also allow avoiding the need for the `catch` here, and just `await`ing the `start` call so that any error from that naturally bubbles up (after the timeout if you do the first suggestion above) into the `Promise<TailFile>`  return value.

sdk/nodejs/x/automation/stack.ts
```diff
@@ -174,16 +214,36 @@ export class Stack {
         }
 
         args.push("--exec-kind", kind);
+
+        let logPromise;
```
lukehoban, line 218: Probably should put a type annotation on this?

sdk/nodejs/x/automation/stack.ts
```diff
@@ -174,16 +214,36 @@ export class Stack {
...
+            const logFile = createLogFile("up");
+            args.push("--event-log", logFile);
+
+            logPromise = this.readLines(logFile, (event) => {
+                onEvent(event);
+            });
+        }
+
+        const upPromise = this.runPulumiCmd(args, opts?.onOutput);
+        const promises: Promise<any>[] = logPromise ? [upPromise, logPromise] : [upPromise];
```
lukehoban, line 231: Since `undefined` can be passed to `Promise.all` directly and just flows through as a no-op, could probably avoid this all - and just do `Promise.all([upPromise, logPromise])` below.

Related - but since you are kicking off both calls already ahead of this - it isn't strictly necessary to `Promise.all` at all - you could just `await` each sequentially.  Since they are already running concurrently, this will have the same effect.  But `Promise.all` is fine if you think it is clearer.

sdk/nodejs/x/automation/stack.ts
```diff
@@ -262,15 +321,46 @@ export class Stack {
...
+        });
+        const prePromise = this.runPulumiCmd(args);
+
         let preResult;
+        let tail;
         try {
-            preResult = await this.runPulumiCmd(args);
+            [preResult, tail] = await Promise.all([prePromise, logPromise]);
         } finally {
             onExit();
```
lukehoban, line 343: I don't know either way - but do you have an intuition for whether/why `onExit` should be called before or after quitting the tailing?

sdk/nodejs/x/automation/stack.ts
```diff
@@ -262,15 +321,46 @@ export class Stack {
         }
 
         args.push("--exec-kind", kind);
+
+        // Set up event log tailing
+        const logFile = createLogFile("preview");
+        args.push("--event-log", logFile);
+        const eventLog: EngineEvent[] = [];
+        const logPromise = this.readLines(logFile, (event) => {
+            eventLog.push(event);
```
lukehoban, line 330: Do we want/need to fill this array up in memory?  If I'm following the code below, we are going to ignore almost all of these later?  Can we just only store the ones we need here - or even just store a `summaryEvent` if/when we see it?

sdk/nodejs/x/automation/stack.ts
```diff
@@ -104,6 +111,39 @@ export class Stack {
                 throw new Error(`unexpected Stack creation mode: ${mode}`);
         }
     }
+    // Try for up to 10s to start tailing the file, invoking the callback once per line. Returns
+    // a promise for a callback to invoke to stop tailing the file.
```
lukehoban, line 115: I think the second sentence is no longer true?

lukehoban (approved):
Looks great!

sdk/nodejs/x/automation/stack.ts
komalali: Hm, I don't *think* it matters either way. They are both cleanup tasks, independent of each other - this wasn't written in this order for any other reason than I had to pick one out of two possible options :)

sdk/nodejs/x/automation/stack.ts
komalali: Ah yeah good catch, I did clean that up in the Go implementation but missed it here.

sdk/nodejs/x/automation/stack.ts
komalali: I learned so many things in those 5 lines of text. I think I will keep the `Promise.all()` for clarity but I did actually think they didn't kick off until that line so 💡 🧠

sdk/nodejs/x/automation/stack.ts
komalali: Huh, my IDE usually calls me out on this stuff 🤔

sdk/nodejs/x/automation/stack.ts
```diff
@@ -104,6 +111,39 @@ export class Stack {
...
+                throw new Error("Log file never found.");
+            }
+            await delay(100);
+            fileExists = fs.existsSync(path);
+            n++;
+        }
+        const eventLogTail = new TailFile(path, { startPos: 0 });
+        eventLogTail
+            .on("tail_error", (err) => {
+                console.error("TailFile had an error!", err);
```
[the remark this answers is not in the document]
komalali: Yup, leftover from the copy-pasta from the example :)

sdk/nodejs/x/automation/events.ts
```diff
@@ -0,0 +1,196 @@
...
+    inputDiff: boolean;
+}
+
+// StepEventMetadata describes a "step" within the Pulumi engine, which is any concrete action
+// to migrate a set of cloud resources from one state to another.
+export interface StepEventMetadata {
+    // Op is the operation being performed, a deploy.StepOp.
+    op: string;
+    urn: string;
+    type: string;
```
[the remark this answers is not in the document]
komalali: I went through and did a second pass, the ones still missing don't have a description in `common/apitype/events`.

sdk/nodejs/x/automation/events.ts
```diff
@@ -0,0 +1,198 @@
...
+    diffKind: DiffKind;
+    // inputDiff is true if this is a difference between old and new inputs rather than old state and new inputs.
+    inputDiff: boolean;
+}
+
+// StepEventMetadata describes a "step" within the Pulumi engine, which is any concrete action
+// to migrate a set of cloud resources from one state to another.
+export interface StepEventMetadata {
+    // Op is the operation being performed, a deploy.StepOp.
+    op: string;
```
EvanBoyle, line 105: This sorta goes for Go as well, but this field is an enum in the engine. It would be convenient to have enums in our SDKs as well unless there's a good reason not to throw a type on Op. 

Through my usage, I repeatedly had to reference the `deploy.StepOp` type in the engine. Might be a little more annoying for node developer (and anyone not familiar with the layout of pulumi/pulumi).

EvanBoyle (approved):
👍  🔥

PR #6460, opened 2021-03-03
Migrate the version calculation to use pulumictl

stack72:
This will allow us to take advantage of setting a VERSION_PREFIX
in our scripts which will tag new major releases easily

This is only a change to the mechanism - no version changes are included as part of this work!! So things will continue as normal


leezen:
Can we remove `get-version` and `get-version.cmd` as part of this PR? That would make it clear we don't use it anymore (and also act as a safeguard to make sure we didn't miss anywhere where other parts of the build rely on those scripts still).

stack72:
> Can we remove `get-version` and `get-version.cmd` as part of this PR? That would make it clear we don't use it anymore (and also act as a safeguard to make sure we didn't miss anywhere where other parts of the build rely on those scripts still).

yeah that's my plan! I had to stop the work to get today's release out but will finish it asap :)

jaxxstorm (approved):

jaxxstorm (approved):

Issue #6383, opened 2021-02-18
`github.com/pulumi/pulumi/pkg/v2`'s dependency on `github.com/pulumi/pulumi/sdk/v2` is broken


pgavlin:
Multi-module repositories do not work well if the modules in the repository are interdependent and simultaneous edits must be made:

https://github.com/golang/go/issues/27056#issuecomment-414028435
https://github.com/golang/go/issues/27542

In our case, this means that the version of `github.com/pulumi/pulumi/pkg/v2` at a particular commit does not have a stated dependency on the version of `github.com/pulumi/pulumi/sdk/v2` at the same commit. Instead, the dependency is stuck at its last update (today, this is [2.2.1](https://github.com/pulumi/pulumi/blob/master/pkg/go.mod#L37) despite the latest version being v2.21.0).

I think that we can address this without losing our ability to make simultaneous changes with a slight adjustment to our release process. For a particular Pulumi version vA.B.C, we could release like so:
1. Tag the `github.com/pulumi/pulumi/sdk/v2` release
2. Remove the `replace` from the go.mod for `github.com/pulumi/pulumi/pkg/v2` and update its dependency on `github.com/pulumi/pulumi/sdk/v2` to the latest version
3. Tag the `github.com/pulumi/pulumi/pkg/v2`and CLI releases

EvanBoyle:
How does this concretely manifest? Is the issue that a consumer of `pkg` needs to specify a replace of `sdk` in its go.mod?

pgavlin:
> How does this concretely manifest? Is the issue that a consumer of pkg needs to specify a replace of sdk in its go.mod?

The issue is that a consumer of `pkg` might need to either replace or explicitly update `sdk`.

stack72:
@pgavlin I don't believe it's wise to change this before the release in the morning - so from tomorrow, we can get this in mainline for a while and ensure we run some tests on it

stack72:
Ok, so we can close this out - there is no engineering work required here - we are going to handle this as part of the deployments. So the steps will now be:

* Tag sdk
* Push sdk tag
* Use SDK tag in pkg mod
* Tag pkg
* Push pkg

It has 1 manual in the middle for now but is better for our users

pgavlin:
@emiliza Are these actually the steps we have currently documented for the release process? I have a feeling that they aren't.

komalali:
@pgavlin They are not.

pgavlin:
[...]

PR #6461, opened 2021-03-03
Removing the replacements of the internal packages

stack72:
Fixes: #6383


pgavlin:
Do we have appropriate changes to the release process to make this workable?

FWIW, I was not suggesting that we remove these replacements, as they are convenient for making parallel changes. Removing them will require that we break such changes up. That's not necessarily a problem--it will make it impossible to release modules with broken dependencies--but it will be a speed bump for development.

PR #6470, opened 2021-03-04
Upgrading Go to 1.16.x

stack72:


stack72:
Newer version of #6404

pgavlin:
What's up with the changes to `build.proj`?

stack72:
@pgavlin for some reason, using the .Identity style syntax from ItemGroup was causing the go installs to fail 

I was being super careful here to ensure this built on windows so local debugging (in a windows machine) got this working.

FWIW, this makes zero sense but 🤷

stack72:
I'm more than happy to spend more time trying to not need these changes if you think it's worthwhile

pgavlin:
> I'm more than happy to spend more time trying to not need these changes if you think it's worthwhile

If the changes to `build.proj` definitely don't affect the functionality of the CLI, then they're okay. The tests in this repo are passing, so that's a good sign.

stack72:
Yeah it defo doesn't - when I had a typo all the windows tests failed due to ldflags :)

justinvp (approved):

PR #6478, opened 2021-03-07
[sdk/python] mypy was not found

mabels:
I had actually some problems, if i try to build pulumi from the source.
Than mypi 0.78 was not found, which is https://pypi.org/project/mypy/#history somehow correct.

Thx in advance

meno


komalali:
Thanks for fixing this @mabels!

komalali (approved):

Issue #3611, opened 2019-12-05
Python: Can't catch exceptions from invokes ("data sources")


geekofalltrades:
This looks like #2908 again, but that issue is closed, so I'm opening a new one.

I'm using pulumi 1.6.0 with pulumi-aws version 1.11.0.

I'm trying to use `get_volume` for disaster recovery to see if a previously dynamically-allocated volume exists so I can reattach it. Here's some sample code:

```python
try:
    backups_volume = ebs.get_volume(
        filters=[
            {
                "Name": "tag:KubernetesCluster",
                "Values": [config.require("cluster.name")],
            },
            {
                "Name": "tag:kubernetes.io/created-for/pvc/namespace",
                "Values": [base_namespace.metadata["name"]],
            },
            {
                "Name": "tag:kubernetes.io/created-for/pvc/name",
                "Values": ["backups"],
            },
        ]
    )
except:
    pass
```

This is resulting in this:

```
Diagnostics:
  pulumi:pulumi:Stack (mine):
    error: Program failed with an unhandled exception:
    error: Traceback (most recent call last):
      File "/home/me/.pulumi/bin/pulumi-language-python-exec", line 85, in <module>
        loop.run_until_complete(coro)
      File "/usr/lib/python3.6/asyncio/base_events.py", line 484, in run_until_complete
        return future.result()
      File "/usr/lib/python3.6/asyncio/tasks.py", line 180, in _step
        result = coro.send(None)
      File "/home/me/projects/pulumi/pulumi-project/venv/lib/python3.6/site-packages/pulumi/runtime/stack.py", line 72, in run_in_stack
        raise RPC_MANAGER.unhandled_exception.with_traceback(RPC_MANAGER.exception_traceback)
      File "/home/me/projects/pulumi/pulumi-project/venv/lib/python3.6/site-packages/pulumi/runtime/rpc_manager.py", line 67, in rpc_wrapper
        result = await rpc_function(*args, **kwargs)
      File "/home/me/projects/pulumi/pulumi-project/venv/lib/python3.6/site-packages/pulumi/runtime/invoke.py", line 157, in do_invoke
        resp = await asyncio.get_event_loop().run_in_executor(None, do_invoke)
      File "/usr/lib/python3.6/asyncio/tasks.py", line 250, in _wakeup
        future.result()
      File "/usr/lib/python3.6/concurrent/futures/thread.py", line 56, in run
        result = self.fn(*self.args, **self.kwargs)
[...]

PR #6504, opened 2021-03-11
Make data sources catchable in Python

pgavlin:
The current logic lets unhandled errors in the RPC invocation
unahandled in the async loop, which crashes the process due to
the way we await completion of RPCs before exiting the process
in Python. Instead of doing that, we can just marshal them back
to the synchronous awaiter as part of the calling convention,
and have that awaiter (which is called by the invoke methods)
re-raise the exception. This should fix pulumi/pulumi#3611.

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -6,4 +6,5 @@
 
 ### Bug Fixes
 
-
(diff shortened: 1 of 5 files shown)


justinvp (approved):
The change LGTM. Mind adding a test?

komalali (approved):
Would also love to see a test here 🙏

Issue #6511, opened 2021-03-12
[automation/*] - Consider adding the ability to capture incremental `stderr` output.


komalali:
### Description

Currently, automation api provides an option to capture incremental stdout output via an `onOutput` event handler / `ProgressStreams` writer array, but there is no way to capture `stderr` incrementally. We should add the ability to do so. 

### Proposal

* For nodejs, python, dotnet: Add a `onStderr`/`on_stderr` event handler.
* For go, add a `StderrStreams` option.

### Context

We would like to explicitly avoid naming this handler `OnError` because it is a misnomer for the actual functionality. We are piping all standard error output here, `OnError` implies that this will only be triggered on actual errors, but conventionally standard error is not only used for errors and Pulumi also logs debug output to standard error.

We should have originally named OnOutput to be OnStandardOutput, that was a miss on our part. I would, however, like to avoid continuing down the misnaming path just because we have done so previously.

### Languages Completed

- [x] [Nodejs](https://github.com/pulumi/pulumi/pull/19372)
- [x] [Go](https://github.com/pulumi/pulumi/pull/10179)
- [x] Dotnet
- [x] [Python](https://github.com/pulumi/pulumi/pull/19402)

teto:
I dont know if it's related but I can't find a way to retreive my logging statements with
```
def create_relay():
    print("toto")
    pulumi.info("tata")
```

stack = auto.create_or_select_stack(stack_name="core-redis-relay-5" ,  program=create_relay)
```
# deploy the stack, tailing the logs to stdout
stack.refresh()
res = stack.up(on_output=print)
```
The `on_output=print` works for native pulumi messages but none of my own logging implemented in create_relay

lunaris:
Reopening this as we had to revert due to Windows issues.

PR #6513, opened 2021-03-12
[automation/dotnet] Add ability to capture stderr

### LINKED ISSUE #6483: [Automation API] Consider adding OnOutput-callback to preview
Currently the dotnet Automation API allows providing a callback `Action<string>? OnOutput` to Up-, Refresh- and DestroyOptions.
Is there a reason no such option is available for PreviewOptions?

From what I checked, this also applies to the python and nodejs automation APIs.

vipentti:
Dotnet support for #6511 and also fixes #6483 

NOTE: I did move the `OnOutput` and added the new `OnStdErr` action to the options base class `UpdateOptions` instead of duplicating them for each option type.

I tried to figure out a way to test that the StdErr action would be called, but I did not seem to find a way to get data written to the standard error stream. Even logging with `Log.Warn` or `Log.Error` would end up being written to the standard output stream.

### DIFF
--- a/sdk/dotnet/Pulumi.Automation/UpdateOptions.cs
+++ b/sdk/dotnet/Pulumi.Automation/UpdateOptions.cs
@@ -17,8 +17,14 @@ namespace Pulumi.Automation
 
         public List<string>? Target { get; set; }
 
+        /// <summary>
+        /// Optional callback which is invoked whenever StandardOutput is written into
+        /// </summary>
         public Action<string>? OnOutput { get; set; }
 
-        public Action<string>? OnStdErr { get; set; }
+        /// <summary>
+        /// Optional callback which is invoked whenever StandardError is written into
+        /// </summary>
+        public Action<string>? OnError { get; set; }
     }
 }
--- a/sdk/dotnet/Pulumi.Automation/UpdateOptions.cs
+++ b/sdk/dotnet/Pulumi.Automation/UpdateOptions.cs
@@ -1,5 +1,6 @@
 ﻿// Copyright 2016-2021, Pulumi Corporation
 
+using System;
 using System.Collections.Generic;
 
 namespace Pulumi.Automation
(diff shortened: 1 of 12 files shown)


orionstudt:
FYI OnOutput was not on the base options class because it is only used for Up and Preview. Looks good otherwise

sdk/dotnet/Pulumi.Automation/UpdateOptions.cs
```diff
@@ -15,5 +16,9 @@ public class UpdateOptions
         public string? Message { get; set; }
 
         public List<string>? Target { get; set; }
+
+        public Action<string>? OnOutput { get; set; }
+
+        public Action<string>? OnStdErr { get; set; }
```
MisinformedDNA, line 22: For consistency, we should go with `OnOutput` and `OnError` or OnStandardOutput` and `OnStandardError`. Also, we should not use abbreviations in .NET.

vipentti: I think `OnError` should be sufficient, maybe with an added doc comment 

```
/// <summary>
/// Optional callback which is invoked whenever StandardError is written into
/// </summary>
public Action<string>? OnError { get; set; }
```

komalali: Hey! I would explicitly like to avoid calling this `OnError` because it is a misnomer for the actual functionality. We are piping Standard Error output here, `onError` implies that this will only be triggered on actual errors, but conventionally standard error is not only used for errors and Pulumi also logs debug output to standard error. 

We should have originally named `OnOutput` to be `OnStandardOutput`, that was a miss on our part. I would, however, like to avoid continuing down the misnaming path just because we have done so previously.

vipentti: Fair enough, I'll rename both of these

vipentti:
Renamed `OnOutput` -> `OnStandardOutput` and the new stderr is `OnStandardError`. I will update the dotnet automation examples next to utilize the new naming.

komalali:
Thanks @vipentti. Can you please rebase this on master and add an entry to CHANGELOG-PENDING in the "Breaking" section?

vipentti:
Rebased and added an entry to CHANGELOG_PENDING

komalali (approved):

PR #6518, opened 2021-03-12
Removing x namespace from go/python/nodejs automation packages

stack72:
All of these were moved with `git mv` so that we preserve the history!


komalali (commented):
This looks 99% of the way there. I think we're just missing exporting `automation` from the top-level `index.ts` for nodejs.

stack72:
> This looks 99% of the way there. I think we're just missing exporting `automation` from the top-level `index.ts` for nodejs.

Interesting - the original module didn't have the import of X so I didn't remove anything there - if this was needed wouldn't the tests fail?

komalali:
It's not "needed" like it would make the tests fail (since the tests are using relative imports), but it is needed so folks can do `import {automation} from "@pulumi/pulumi"` - it was probably an oversight that this was left out initially. If you look at the top-level `index.ts` it includes all of the other submodules. 

Right now the only way to import from the automation module is to do something like `import { InlineProgramArgs, LocalWorkspace } from "@pulumi/pulumi/automation";` which is inconsistent with the rest of our modules.

stack72:
> It's not "needed" like it would make the tests fail (since the tests are using relative imports), but it is needed so folks can do `import {automation} from "@pulumi/pulumi"` - it was probably an oversight that this was left out initially. If you look at the top-level [index.ts](https://github.com/pulumi/pulumi/blob/master/sdk/nodejs/index.ts?rgh-link-date=2021-03-16T16%3A38%3A07Z#L35) it includes all of the other submodules.
> 
> Right now the only way to import from the automation module is to do something like `import { InlineProgramArgs, LocalWorkspace } from "@pulumi/pulumi/automation";` which is inconsistent with the rest of our modules.

gotcha!! Then it's a good thing to add :)

komalali (approved):
:shipit:

Issue #6425, opened 2021-02-25
Support prompt values for multi-lang component inputs


justinvp:
Components often need to conditionally create child resources based on input arguments. But multi-lang components currently project all inputs as `Input` (each will be `Output`) meaning there's really no good way to do this since we don't recommend creating resources inside an `apply`.

During a recent planning discussion, we decided the reasonable path forward here is to allow prompt values for multi-lang component inputs. This improves authorability at the expense of increasing some of the complexity on the consumer side, but not by much as current `ComponentResource` consumers already need to think about prompt values.

This requires changes in the schema, codegen/docgen, and `construct` provider helpers (currently TS and Go).

- The schema will be augmented to support indicating that an input property is a prompt value.
- The code/doc generators will emit such properties typed as the prompt value type itself, rather than being wrapped as `Input<T>`.
- The provider implementations of `construct` will be updated to pass prompt values as-is rather than wrapping the values in `Output<T>`.

PR #6522, opened 2021-03-12
[sdk/nodejs] Keep prompt values prompt in construct

justinvp:
In order to support prompt values in multi-lang components, if an input value is prompt, keep it as-is instead of wrapping it in an Output.

Part of #6425


lblackstone (approved):

sdk/nodejs/provider/server.ts
```diff
@@ -299,8 +299,15 @@ class Server implements grpc.UntypedServiceImplementation {
                 const deps = (inputDeps ? <resource.URN[]>inputDeps.getUrnsList() : [])
                     .map(depUrn => new resource.DependencyResource(depUrn));
                 const input = deserializedInputs[k];
-                inputs[k] = new Output(deps, Promise.resolve(runtime.unwrapRpcSecret(input)), Promise.resolve(true),
-                    Promise.resolve(runtime.isRpcSecret(input)), Promise.resolve([]));
+                const isSecret = runtime.isRpcSecret(input);
+                if (!isSecret && deps.length === 0) {
```
t0yv0, line 303: This made me think a little bit, trying to type check this in my head. Somehow from the test it's evident that `children: Optional[Int]` travels through this code just fine, but how is that value represented here is not obvious to me. I also wonder about the condition `(!isSecret && deps.length === 0)`. I don't know yet what inputDependencies are very well. 

Would this code confuse `T` and `Input<T>` or it does not matter?  I imagine there is a way to lift `pure :: T -> Input<T>` and that `pure(x)` should not have any dependencies so presumably it would satisfy `deps.length == 0? `

tests/integration/integration_go_test.go
```diff
@@ -211,6 +211,37 @@ func TestConstructSlowGo(t *testing.T) {
...
 }
 
+// Test remote component construction with prompt inputs.
+func TestConstructPlainGo(t *testing.T) {
+	pathEnv, err := testComponentPlainPathEnv()
+	if err != nil {
+		t.Fatalf("failed to build test component PATH: %v", err)
+	}
+
+	// TODO[pulumi/pulumi#5455]: Dynamic providers fail to load when used from multi-lang components.
```
t0yv0, line 221: I didn't realize there's a problem here want to take a look now.

t0yv0 (approved):

sdk/nodejs/provider/server.ts
justinvp: `inputDependencies` is documented in the .proto:

https://github.com/pulumi/pulumi/blob/3dde9d02bf0b3fd84ab9630ef33cb7aecc99cf67/sdk/proto/provider.proto#L267

https://github.com/pulumi/pulumi/blob/3dde9d02bf0b3fd84ab9630ef33cb7aecc99cf67/sdk/proto/provider.proto#L251-L254

If the property was assigned an output of another resource, or the result of an `apply` from one or more outputs, then the property will have an associated list of resource URNs that the property depends on. If there aren't any dependencies associated with the property, then we don't need to project it as an `Output` because we don't have any dependencies to track. Unless it's a secret: secrets in the type system are typed as `Output<T>` (we did this rather than having a separate `Secret<T>` type to keep the programming model simpler), so in that case, we need to wrap the value as `Output<T>`.

We don't have a specific flag that says "Property 'foo' was set a plain value", so instead, we just see whether the raw gRPC value can remain a plain value, otherwise we wrap it as an output. We're sort of relying on the type system of the generated SDK to indicate that the type of the value should be `T` and not `Input<T>`, but in some languages that don't have a real type system (e.g. JavaScript/Python), someone could force passing an `Output`, so right now we're sort of relying on type checking to prevent that. We should probably add some runtime type checking for JS/Python for plain values, to further help prevent that from happening.

> Would this code confuse `T` and `Input<T>` or it does not matter?

`Input<T>` can be `T` or `Output<T>`:

https://github.com/pulumi/pulumi/blob/3dde9d02bf0b3fd84ab9630ef33cb7aecc99cf67/sdk/nodejs/output.ts#L727

If an input property is typed as accepting an `Input<T>`, it can be given a `T`, and it'll _just work_.

Issue #5146, opened 2020-08-10
Python: Consider supporting `<Resource>Args` classes for creating resources


justinvp:
Python resources are created by passing inputs directly to a resource's `__init__` method, e.g.:

```python
bucket = aws.s3.Bucket("bucket",
    acl="private",
    tags={
        "Environment": "Dev",
        "Name": "My bucket",
    })
```

This approach is different from other Pulumi languages, which have a separate type (e.g. `BucketArgs`) to pass the inputs to the resource's constructor.

A customer asked if we'd consider supporting `<Resource>Args` for Python. The primary reason is to have an object "bag" that can be used to hold/manipulate/validate/pass-around inputs _before_ creating the resource. It'd also provide consistency with all the other Pulumi languages.

We could add optional support for this in a compatible way via an `__init__` [overload](https://docs.python.org/3/library/typing.html#typing.overload):

```python
class Bucket(CustomResource):
    @overload
    def __init__(__self__, args: Optional[BucketArgs] = None, opts: Optional[pulumi.ResourceOptions] = None)
        ...
    @overload
    def __init__(__self__,
                 resource_name,
                 opts: Optional[pulumi.ResourceOptions] = None,
                 acceleration_status: Optional[pulumi.Input[str]] = None,
                 acl: Optional[pulumi.Input[str]] = None,
                 ...)
```

And usage:

```python
args = aws.s3.BucketArgs(
    acl="private",
    tags={
        "Environment": "Dev",
        "Name": "My bucket",
    })
bucket = aws.s3.Bucket("bucket", args)
```

KentShikama:
Python doesn't support having multiple `__init__` functions in a class. If you provide multiple, CPython will just ignore the earlier ones. You could support a `fromArgs` factory like follows:

```
@classmethod
def fromArgs(cls, resource_name, args, opts):
    return cls(resource_name, opts, args.foo, args.bar, ...)

def __init__(__self__, resource_name, ...
```

Usage:

```
args = aws.s3.BucketArgs(
    acl="private",
    tags={
        "Environment": "Dev",
        "Name": "My bucket",
    })
bucket = aws.s3.Bucket.fromArgs("bucket", args)
```

justinvp:
> Python doesn't support having multiple `__init__` functions in a class. If you provide multiple, CPython will just ignore the earlier ones.

[...]

PR #6525, opened 2021-03-13
[codegen/python] Support `<Resource>Args` classes

justinvp:
Add support for creating instances of resources in Python using a `<Resource>Args` class. This capability aligns with how args are passed to resources in all the other language SDKs and the separate object bag allows the properties to be manipulated/validated/passed-around before creating the resource.

Part of #5146


pkg/codegen/python/gen.go
```diff
@@ -965,34 +982,75 @@ func (mod *modContext) genResource(res *schema.Resource) (string, error) {
...
 
-	// Old versions of TFGen emitted parameters named __name__ and __opts__. In order to preserve backwards
-	// compatibility, we still emit them, but we don't emit documentation for them.
-	fmt.Fprintf(w, ",\n                 __props__=None")
-	fmt.Fprintf(w, ",\n                 __name__=None")
-	fmt.Fprintf(w, ",\n                 __opts__=None):\n")
-	mod.genInitDocstring(w, res)
+	// Now generate __init__ overloads along with an implementation...
+
+	// First, generate an __init__ overload that accepts the resource's inputs from the args class.
```
justinvp, line 1003: Do we have a preference for which overload shows up first in code completion lists? VS Code shows the overloads in the order they're declared (at least in this case).

Screen Shot 2021-03-15 at 12 59 03 PM

leezen: I'd propose having the keyword-based style first ahead of the args-based given that's what people are currently using. We could choose to flip this order for 3.0 to encourage the args-based as the 'default'

justinvp: SGTM. I've pushed a commit that does that.

Screen Shot 2021-03-16 at 6 22 18 AM

Screen Shot 2021-03-16 at 6 22 41 AM

komalali (approved):
LGTM. Do we know what impact this will have on docs and how we will handle presenting the overloads there?

komalali:
Oh, I realized that all the tests here are of the codegen itself. Could we add some tests of this in use?

PR #6527, opened 2021-03-14
[automation/python - Expose structured logging

komalali:
See also: #6436 (go), #6454 (nodejs), #6572 (dotnet).

### Changes

* Adds an `on_event` option to `Stack.up()`, `Stack.preview()`, `Stack.refresh()` and `Stack.destroy()` for acting on structured json events.
* Updates the `PreviewResult` interface to include a `change_summary` property that summarizes changes to resources.

Fixes: #6443


sdk/python/lib/pulumi/x/automation/_stack.py
```diff
@@ -352,19 +339,39 @@ def on_exit_fn():
             args.append(f"--client=127.0.0.1:{port}")
         args.extend(["--exec-kind", kind])
 
+        log_file, temp_dir = _create_log_file("preview")
+        args.extend(["--event-log", log_file])
+        summary_events: List[SummaryEvent] = []
+
+        def on_event_callback(event: EngineEvent):
```
t0yv0, line 346: Since it does not have `-> <return-type>` annotation, mypy skips checking the body of this def. Just FYI.

sdk/python/lib/pulumi/x/automation/_stack.py
```diff
@@ -620,3 +670,45 @@ def fully_qualified_stack_name(org: str, project: str, stack: str) -> str:
...
     """
     return f"{org}/{project}/{stack}"
+
+
+def _create_log_file(command: str) -> Tuple[str, tempfile.TemporaryDirectory]:
+    log_dir = tempfile.TemporaryDirectory(prefix=f"automation-logs-{command}-")
+    filepath = os.path.join(log_dir.name, "eventlog.txt")
+
+    # Open and close the file to ensure it exists before we start polling for logs
+    f = open(filepath, "w+")
```
t0yv0, line 680: This works, I think, another thing to think about what does it do when the file exists already? Guess that's not possible given TemporaryDirectory randomization.

There is another API achieves a similar effect potentially using different sys calls:

```
from pathlib import Path
Path('path/to/file.txt').touch()
```

sdk/python/lib/pulumi/x/automation/_stack.py
```diff
@@ -620,3 +670,45 @@ def fully_qualified_stack_name(org: str, project: str, stack: str) -> str:
...
+    f.close()
+    return filepath, log_dir
+
+
+def _watch_logs(filename: str, callback: OnEvent):
```
t0yv0, line 685: I tested this a bit it seems to work pretty well. sleep() does not hold GIL so other threads can make progress. The `f.readline()` keeps returning empty str even when it reached eof. 

Something you might want to think about is what bad things happen when we run user callbacks on the logging thread. It does not seem to be the user's expectation. These callbacks cannot run simultaneously with user Py code because Py code always takes the GIL, but if they utilize sys calls concurrency is possible. I can't construct a confusing example but I sense there might be one here. Let me look up python threading modules to see what else we can do to push the events back to the parent thread and execute the callbacks there.

sdk/python/lib/pulumi/x/automation/_stack.py
```diff
@@ -620,3 +670,45 @@ def fully_qualified_stack_name(org: str, project: str, stack: str) -> str:
...
+    with open(filename) as f:
+        while True:
+            line = f.readline()
+
+            # sleep if file hasn't been updated
+            if not line:
+                time.sleep(0.1)
+                continue
+
+            event = EngineEvent.from_json(json.loads(line))
```
t0yv0, line 695: Another thing to think about is what happens when an exception gets thrown here, either we are encountering corrupt non-JSON garbage on the input, or the user callback throws an exception.  Does this surface to the user in an intelligible way?

sdk/python/lib/pulumi/x/automation/_stack.py
```diff
@@ -620,3 +670,45 @@ def fully_qualified_stack_name(org: str, project: str, stack: str) -> str:
...
+
+            # if this is the cancel event, stop watching logs.
+            if event.cancel_event:
+                break
+
+
+def _cleanup(temp_dir: Optional[tempfile.TemporaryDirectory],
+             thread: Optional[threading.Thread],
+             on_exit_fn: Optional[Callable[[], None]] = None) -> None:
+    # If there's an on_exit function, execute it (used in preview/up to shut down server)
```
t0yv0, line 706: Sequencing seems to matter here if any of these throw exception. Let me try thread.join and see if that throws.

sdk/python/lib/pulumi/x/automation/events.py
```diff
@@ -0,0 +1,537 @@
...
+        self.diagnostic_event = diagnostic_event
+        self.prelude_event = prelude_event
+        self.summary_event = summary_event
+        self.resource_pre_event = resource_pre_event
+        self.res_outputs_event = res_outputs_event
+        self.res_op_failed_event = res_op_failed_event
+        self.policy_event = policy_event
+
+    @classmethod
+    def from_json(cls, data: dict) -> 'EngineEvent':
```
t0yv0, line 517: This is works and gives precise control. I wonder if in the future we'll want matching to_json. There seems to be something out there that can attempt metaprogramming based name conversion so we don't have to spell it out, though ofc we lose some control:

https://pypi.org/project/dataclasses-json/

sdk/python/lib/pulumi/x/automation/_stack.py
```diff
@@ -620,3 +670,45 @@ def fully_qualified_stack_name(org: str, project: str, stack: str) -> str:
...
+
+            # sleep if file hasn't been updated
+            if not line:
+                time.sleep(0.1)
+                continue
+
+            event = EngineEvent.from_json(json.loads(line))
+            callback(event)
+
+            # if this is the cancel event, stop watching logs.
```
t0yv0, line 698: I'd really like to know more about this one, what kind of cancel event? I do not think I have matching logic in dotnet SDK, perhaps a miss.

sdk/python/lib/pulumi/x/automation/events.py
komalali: We have thus far avoided using dataclasses because they were introduced in python 3.7 and even though there is a [backport](https://pypi.org/project/dataclasses/), @justinvp hasn't been too into the idea of taking the dependency.

It's also not always as straightforward as pure snake case to camel case, as in the case of [PolicyPacks](https://github.com/pulumi/pulumi/pull/6527/files#diff-1c8278368b2a8ab8d889dc9ffa979f0abfdf1c646782cefbe0b22f8443d4fa55R244)

t0yv0: Right, sounds like that one is out of reach. OK let's move on.

sdk/python/lib/pulumi/x/automation/_stack.py
komalali: `CancelEvent` is the last event transmitted by the engine as a signal that logging is over. I've used it here to signal that the thread can terminate, so when `thread.join()` waits it doesn't always time out because otherwise there would be no end to this `_watch_logs` function. We can also use an `Event` object to [signal completion](https://docs.python.org/3/library/threading.html#id1:~:text=A%20thread%20can%20be%20flagged%20as,signalling%20mechanism%20such%20as%20an%20Event.) though, if there is hesitation to go this route.

sdk/python/lib/pulumi/x/automation/_stack.py
komalali: > Guess that's not possible given TemporaryDirectory randomization.

Yup, it's not possible. Thanks for the reference, good to know but I think I'll leave this as-is.

sdk/python/lib/pulumi/x/automation/_stack.py
t0yv0:
```// CancelEvent is emitted when the user initiates a cancellation of the update in progress, or
// the update successfully completes.
export interface CancelEvent {
}
```

So it is not the above Cancel event? And nothing about stack.Cancel() right?

Yeah this is awesome that the engine signals completion of logging. I wanted something like his for .NET to avoid race conditions, but I did not realize there is an event we can count on to be the last event. Calling it "cancel event" seems a little misleading but it does seem like we need something like this.

If we use built-in concurrency primitives to try to cancel the bg thread when we think the work is done, we're always open to a subtle race condition where we guess wrong (we think the work is done but IO has not propagated all the logs yet).

sdk/python/lib/pulumi/x/automation/_stack.py
t0yv0: Yeah seems so. If there's an exception in the background thread, it gets nicely printed out "Excpetion in thread Thread-1" with a stack trace. So far so good.

t0yv0: Yes sounds like none of these finalizers will throw so we might be good here! The way to avoid having to reason/worry through this is nesting `with` statements and `from contextlib import contextmanager` - the with statements basically a condensed form of try..finally, that kind of code is easy to see at a glance that finalizers don't interfere and all try to run even if some fail.

t0yv0:
OK update on threading issue. We can't really run callbacks from the parent thread in this design because of `_run_pulumi_cmd`. Our main thread is busy being hogged by that function `while True:` loop. So if we are to run user's callbacks concurrently we need another thread.

To get out of this situation would be to async-io-ify the SDK from _run_pulumi_cmd, in which case we could be polling the log file in AIO loop as well without threading. There seems to be support in 3.6 but it's gnarly:

https://docs.python.org/3.6/library/asyncio-subprocess.html

Def out of scope fo this PR.

t0yv0 (approved):

sdk/python/lib/pulumi/x/automation/_stack.py
komalali: It is indeed the same `CancelEvent` that you referenced. It is emitted on `stack.cancel()` but also whenever an update completes, in both the error and success case. It's not so much a "completion of logging" but a "completion of update" which I am equating to the same thing here. I admit I'm not 100% confident that this will always be the case, but we do have the timeout on `thread.join()` to fall back on in any case (maybe that should be shorter?).

> Calling it "cancel event" seems a little misleading but it does seem like we need something like this.

Heh, agreed 😄 

I'm tempted to go ahead with this and we can make adjustments if we run into issues. All of our error tests complete successfully so I'm not too worried.

t0yv0: Yes please, don't block on this. I will follow up to understand this better. Perhaps we can rename things later to make more sense.

Issue #6297, opened 2021-02-09
[codegen/go] Handle collisions between types and their container type variants


viveklak:
While we handle collisions between resources and types by renaming types ( #6105 ) we also run into collisions within types and their variants, e.g.`<TypeName>State` (or input/output etc.) and another type of name `TypeNameState`.

pgavlin:
Are we hitting this in Azure nextgen?

viveklak:
> Are we hitting this in Azure nextgen?

Sorry late to reply. Yes we are. e.g.:
```
github.com/pulumi/pulumi-azure-nextgen/sdk/go/azure/documentdb/latest
    /Users/levi/go/pkg/mod/github.com/pulumi/pulumi-azure-nextgen/sdk@v0.6.1/go/azure/documentdb/latest/pulumiTypes.go:6608:19: undefined: CompositePathArrayArrayInput
    /Users/levi/go/pkg/mod/github.com/pulumi/pulumi-azure-nextgen/sdk@v0.6.1/go/azure/documentdb/latest/pulumiTypes.go:6703:50: undefined: CompositePathArrayArrayOutput
    /Users/levi/go/pkg/mod/github.com/pulumi/pulumi-azure-nextgen/sdk@v0.6.1/go/azure/documentdb/latest/pulumiTypes.go:6704:91: undefined: CompositePathArrayArrayOutput
    /Users/levi/go/pkg/mod/github.com/pulumi/pulumi-azure-nextgen/sdk@v0.6.1/go/azure/documentdb/latest/pulumiTypes.go:6756:53: undefined: CompositePathArrayArrayOutput
    /Users/levi/go/pkg/mod/github.com/pulumi/pulumi-azure-nextgen/sdk@v0.6.1/go/azure/documentdb/latest/pulumiTypes.go:6762:6: undefined: CompositePathArrayArrayOutput
    /Users/levi/go/pkg/mod/github.com/pulumi/pulumi-azure-nextgen/sdk@v0.6.1/go/azure/documentdb/latest/pulumiTypes.go:6837:19: undefined: CompositePathResponseArrayArrayInput
    /Users/levi/go/pkg/mod/github.com/pulumi/pulumi-azure-nextgen/sdk@v0.6.1/go/azure/documentdb/latest/pulumiTypes.go:6932:58: undefined: CompositePathResponseArrayArrayOutput
[...]

PR #6528, opened 2021-03-15
[codegen/go] Support nested collection types

viveklak:
Fixes https://github.com/pulumi/pulumi-azure-native/issues/353
Fixes https://github.com/pulumi/pulumi/issues/6297

With this we are able to cover all constructs required by azure-native go SDK be compiled. I was able to confirm by compiling all the top-level namespaces locally.

*Update* - this PR is ready for review

### DIFF
--- a/pkg/codegen/go/gen.go
+++ b/pkg/codegen/go/gen.go
@@ -497,7 +497,7 @@ func printCommentWithDeprecationMessage(w io.Writer, comment, deprecationMessage
 	}
 }
 
-func genInputInterface(w io.Writer, name, argsName string) {
+func genInputInterface(w io.Writer, name string) {
 	printComment(w, getInputUsage(name), false)
 	fmt.Fprintf(w, "type %sInput interface {\n", name)
 	fmt.Fprintf(w, "\tpulumi.Input\n\n")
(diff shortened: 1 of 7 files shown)


mikhailshilkov:
I'm not fluent enough with our Go SDK structure to confidently review this. I'd love others to take a look - it would be great to land the fix soon!

pkg/codegen/go/gen.go
```diff
@@ -507,7 +497,7 @@ func printCommentWithDeprecationMessage(w io.Writer, comment, deprecationMessage
 	}
 }
 
-func genInputInterface(w io.Writer, name string) {
+func genInputInterface(w io.Writer, name, argsName string) {
```
pgavlin, line 500: Looks like the new parameter here is unused?

viveklak: Sorry it took a while to return to this. Yes - fixed now. I also fixed a niggling bug with some usage documentation generation which was causing unnecessary changes in some providers.

viveklak:
The PR is rebased and updated/ready for another look. Verified that the only provider with changes is [pulumi-azure-native](https://github.com/pulumi-bot/pulumi-azure-native/compare/875bfa4884..7ab8bfc82e). I also verified that the generated SDK now compiles.

EvanBoyle (approved):
I appreciate the refactoring that you did in gen.go! Made this change fairly easy to follow along as far as go codegen changes go 😄  and the tests/stable diffs seem to speak for themselves.

PR #6530, opened 2021-03-15
[automation/dotnet,nodejs,python] Specify exec-kind for refresh & destroy

vipentti:
dotnet, nodejs and python automation APIs did not specify exec-kind for
refresh or destroy operations. This is now added following the same
logic from the go automation API.


komalali:
🥇

komalali (approved):

PR #6544, opened 2021-03-16
[dotnet] Fix InputList implicit conversion with default ImmutableArray

vipentti:
I discovered an issue with `Pulumi.Kubernetes.Helm.V3.Charts` pulumi/pulumi-kubernetes#1496
when trying to write unit tests in C# for a stack using Helm charts.

After looking at the problem, I think the problem can be solved inside the core Pulumi SDK.

The problem occurs when an InputList is initialized from a default `ImmutableArray` using the
implicit conversion operator. The conversion operator eventually calls `ImmutableArray.CreateRange`,
which ends up failing due to accessing an uninitialized array.

This PR now adds some extra safeguards to both `InputList` and `Output` types against
`ImmutableArray` values which have `IsDefault` true.

The `InputList` fix is necessary, as implicit conversion operations should not fail.
So when given an array with `IsDefault` `true`, we return `ImmutableArray<T>.Empty` instead.

I also applied same logic to `Output.All` overloads.

### DIFF
--- a/sdk/dotnet/Pulumi/Core/InputList.cs
+++ b/sdk/dotnet/Pulumi/Core/InputList.cs
@@ -120,7 +120,7 @@ namespace Pulumi
             => values.EnsureNotDefault().SelectAsArray(v => (Input<T>)v);
 
         public static implicit operator InputList<T>(ImmutableArray<Input<T>> values)
-            => Output.All(values);
+            => Output.All(values.EnsureNotDefault());
 
         #endregion
 
@@ -136,7 +136,7 @@ namespace Pulumi
             => values.Apply(a => ImmutableArray.CreateRange(a));
 
         public static implicit operator InputList<T>(Output<ImmutableArray<T>> values)
-            => new InputList<T>(values);
+            => new InputList<T>(values.Apply(v => v.EnsureNotDefault()));
 
         #endregion
 
--- a/sdk/dotnet/Pulumi/Core/InputList.cs
+++ b/sdk/dotnet/Pulumi/Core/InputList.cs
@@ -114,10 +114,10 @@ namespace Pulumi
         #region construct from immutable array
 
         public static implicit operator InputList<T>(ImmutableArray<T> values)
-            => values.SelectAsArray(v => (Input<T>)v);
+            => values.EnsureNotDefault().SelectAsArray(v => (Input<T>)v);
 
         public static implicit operator InputList<T>(ImmutableArray<Output<T>> values)
-            => values.SelectAsArray(v => (Input<T>)v);
+            => values.EnsureNotDefault().SelectAsArray(v => (Input<T>)v);
 
         public static implicit operator InputList<T>(ImmutableArray<Input<T>> values)
             => Output.All(values);
--- a/sdk/dotnet/Pulumi/Core/Output.cs
+++ b/sdk/dotnet/Pulumi/Core/Output.cs
@@ -58,7 +58,13 @@ namespace Pulumi
         /// </summary>
         public static Output<ImmutableArray<T>> All<T>(IEnumerable<Input<T>> inputs)
             => Output<T>.All(ImmutableArray.CreateRange(inputs));
-        
+
+        /// <summary>
+        /// <see cref="All{T}(Input{T}[])"/> for more details.
+        /// </summary>
+        public static Output<ImmutableArray<T>> All<T>(ImmutableArray<Input<T>> inputs)
+            => Output<T>.All(ImmutableArray.CreateRange(inputs.EnsureNotDefault()));
+
         /// <summary>
         /// Combines all the <see cref="Output{T}"/> values in <paramref name="outputs"/>
         /// into a single <see cref="Output{T}"/> with an <see cref="ImmutableArray{T}"/>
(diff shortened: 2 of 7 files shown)


sdk/dotnet/Pulumi/Core/Output.cs
```diff
@@ -58,7 +58,13 @@ public static Output<ImmutableArray<T>> All<T>(params Input<T>[] inputs)
...
         public static Output<ImmutableArray<T>> All<T>(IEnumerable<Input<T>> inputs)
             => Output<T>.All(ImmutableArray.CreateRange(inputs));
-        
+
+        /// <summary>
+        /// <see cref="All{T}(Input{T}[])"/> for more details.
+        /// </summary>
+        public static Output<ImmutableArray<T>> All<T>(ImmutableArray<Input<T>> inputs)
+            => Output<T>.All(ImmutableArray.CreateRange(inputs.EnsureNotDefault()));
+
```
vipentti, line 67: Actually maybe it might be better to remove the `ImmutableArray` overloads and update the `IEnumerable` ones such as:

```csharp
public static Output<ImmutableArray<T>> All<T>(IEnumerable<Input<T>> inputs)
    => Output<T>.All(ImmutableArray.CreateRange(inputs switch {
            ImmutableArray<Input<T>> immutable => immutable.EnsureNotDefault(),
            _ => inputs,
        }));

public static Output<ImmutableArray<T>> All<T>(IEnumerable<Output<T>> outputs)
    => Output<T>.All(ImmutableArray.CreateRange((outputs switch {
            ImmutableArray<Output<T>> immutable => immutable.EnsureNotDefault(),
            _ => outputs,
        }).Select(o => (Input<T>)o)));
```
Doing this would also ensure the following tests work

```csharp
await AssertEmpty(Output.All((IEnumerable<Input<bool>>)(ImmutableArray<Input<bool>>) default));
await AssertEmpty(Output.All((IEnumerable<Output<bool>>)(ImmutableArray<Output<bool>>) default));
```

sdk/dotnet/Pulumi/Core/Output.cs
```diff
@@ -73,7 +76,10 @@ public static Output<ImmutableArray<T>> All<T>(params Output<T>[] outputs)
         /// <see cref="All{T}(Output{T}[])"/> for more details.
         /// </summary>
         public static Output<ImmutableArray<T>> All<T>(IEnumerable<Output<T>> outputs)
-            => Output<T>.All(ImmutableArray.CreateRange(outputs.Select(o => (Input<T>)o)));
+            => Output<T>.All(ImmutableArray.CreateRange((outputs switch {
```
t0yv0, line 79: Hmm, this is interesting.  

Something here strikes me as very unfortunate. The function expects IEnumerable and it so happens that one of the many subtypes of that, ImmutableArray admits a default value that breaks the IEnumerable contracts, so the PR suggest to detect and patch that. However, the problem is that there's a potentially infinite variety of invalid IEnumerable implementations that break the contract, and it would not be possible to detect them all here (the switch will get too large). What I am saying perhaps the fault belongs to the caller of this function and should be fixed there, not in this function?

Do we know where the default ImmutableArray originated, where was it constructed?

If we were to edit this function, perhaps a null check plus meaningful failure would do? Still I feel it might be a bit too defensive given that IEnumerable is a very common thing to use and typically there are no null checks around it.

sdk/dotnet/Pulumi/Core/InputList.cs
```diff
@@ -114,10 +114,10 @@ internal InputList<T> Clone()
         #region construct from immutable array
 
         public static implicit operator InputList<T>(ImmutableArray<T> values)
-            => values.SelectAsArray(v => (Input<T>)v);
+            => values.EnsureNotDefault().SelectAsArray(v => (Input<T>)v);
 
         public static implicit operator InputList<T>(ImmutableArray<Output<T>> values)
-            => values.SelectAsArray(v => (Input<T>)v);
+            => values.EnsureNotDefault().SelectAsArray(v => (Input<T>)v);
```
t0yv0, line 120: This is more reasonable but why these functions and not others? There's many functions in this file accepting ImmutableArray<T> and some of them actually accept it via implicit cast to IEnumerable<T>. 

For example:

```
 public static implicit operator InputList<T>(Output<ImmutableArray<T>> values)
            => new InputList<T>(values);
```

t0yv0:
Let me review the linked PR now sounds like it has more clues to where the default value originates.

sdk/dotnet/Pulumi/Core/Output.cs
vipentti: After thinking about it more carefully, I think the changes to `Output.All` should be reverted, as indeed it is the caller of the function who is responsible for providing a valid `IEnumerable`.

sdk/dotnet/Pulumi/Core/InputList.cs
vipentti: I think adding safeguard for the ` public static implicit operator InputList<T>(Output<ImmutableArray<T>> values)` is probably reasonable as well?

This would allow the following conversion to work

```csharp
await AssertCanConvert(Output.Create((ImmutableArray<bool>) default));
```

Not sure what to do about the `IEnumerable` overload, since unfortunately you may indeed have `ImmutableArray` via `IEnumerable`

```csharp
  public static implicit operator InputList<T>(Output<IEnumerable<T>> values)
      => values.Apply(a => ImmutableArray.CreateRange(a));
```

t0yv0:
Sorry for a bit of delay here, on balance I think we should decline this, the reasoning being that the "default" or null value was unintentional and this change makes the SDK hide it further and replace it with empty collection. It think we'd want instead to fail harder closer to the code where the unexpected null value originates; or absent of that do nothing.

Closing for the moment, please feel free to reopen and then I can get a second opinion here.

PR #6545, opened 2021-03-16
Make template repository URL and ref modifiable

jen20:
This commit changes the repository URLs and default branch names for both program and policy pack template repositories from constants to variables, in order that they are customizable by linker flags.

For example, compiling the Pulumi CLI with the following linker flags:

- `-X [module path]/workspace.pulumiTemplateGitRepository=https://github.com/jen20/pulumi-templates.git`
- `-X [module path]/workspace.pulumiTemplateBranch=main`

Would instruct the `pulumi new` command to pull the list of available templates from the `main` branch of the specified repository instead of the upstream default. The defaults are left unchanged.

### DIFF
--- a/sdk/go/common/workspace/templates.go
+++ b/sdk/go/common/workspace/templates.go
@@ -35,9 +35,6 @@ import (
 const (
 	defaultProjectName = "project"
 
-	pulumiTemplateGitRepository       = "https://github.com/pulumi/templates.git"
-	pulumiPolicyTemplateGitRepository = "https://github.com/pulumi/templates-policy.git"
-
 	// This file will be ignored when copying from the template cache to
 	// a project directory.
 	legacyPulumiTemplateManifestFile = ".pulumi.template.yaml"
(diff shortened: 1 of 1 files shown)


jen20:
I believe this does not need a CHANGELOG entry, since it is not user-facing.

sdk/go/common/workspace/templates.go
```diff
@@ -51,6 +48,17 @@ const (
 	pulumiLocalPolicyTemplatePathEnvVar = "PULUMI_POLICY_TEMPLATE_PATH"
 )
 
+var (
```
justinvp, line 51: Would it be worth adding a comment explaining _why_ these are variables rather than constants so we don't unknowingly revert in the future?

justinvp (approved):
LGTM. Thank you!

sdk/go/common/workspace/templates.go
jen20: Yes, that's probably a good idea.

jen20: This is addressed in `5b71f81`.

PR #6551, opened 2021-03-17
Make `custom` available to `Mocks::newResource`

jen20:
This commit adds a new optional parameter to the `newResource` function of the `Mocks` interface for TypeScript. This can be useful when writing tests which assert differing behavior between Custom and Component resources.

Although the new parameter will always be set, the paramteter is marked as optional in order to maintain backwards compatibility with existing implementations of `Mocks`.

The tests are updated to verify that `custom` is set appropriately.


lukehoban:
> This commit adds a new optional parameter to the newResource function of the Mocks interface for TypeScript. 

Note that ideally we would add this in all languages to maintain consistency.  It appears this is consistently not available in the other languages either - and should be added in the same way.  If we can do that as part of this PR, great!  Else perhaps we can open an issue to track closing this language inconsistency gap again as follow-up.

jen20:
@lukehoban Yes, I can add it to the other runtimes which support mocks too.

I'm starting to wonder if it might be better to replace `provider`, `id` and `custom` with a `metadata` class to which fields can easily be added without proliferating optional arguments as more gets added here - I can imagine that it might be desirable in future to get access to things such as `deleteBeforeReplace` also. Perhaps that is something to file as "future work" though.

jen20:
I think I prefer the idea on the whole of a change to a metadata bag after looking at this in other languages:

- In Python, we'd need to add the parameter to the `new_resource` method on the `Mocks` ABC [here](https://github.com/pulumi/pulumi/blob/master/sdk/python/lib/pulumi/runtime/mocks.py#L58). I'm not 100% certain whether this would break existing implementations or not yet.
- In Go, we'd need to change the signature of `NewResource` on `MockResourceMonitor` ([here](https://github.com/pulumi/pulumi/blob/master/sdk/go/pulumi/mocks.go#L20-L21)), which would break existing implementations of the interface.
- For .NET, we'd need to add to the `NewResourceAsync` method on the `IMocks` interface  ([here](https://github.com/pulumi/pulumi/blob/master/sdk/dotnet/Pulumi/Testing/IMocks.cs#L23-L24))

The lax approach of JavaScript to function parameters seems to be a win here for backwards compatibility! Perhaps the Node.js version could be taken for V2, and a breaking change could be adopted for V3 that moves to a metadata bag approach across all languages?

lukehoban:
> Perhaps the Node.js version could be taken for V2, and a breaking change could be adopted for V3 that moves to a metadata bag approach across all languages?

That sounds great to me.  We can merge this PR as-is. Do you want to follow-up with the breaking change PR for 3.0?

lukehoban:
Opened https://github.com/pulumi/pulumi/issues/6617 on the work to add this functionality to the other languages.

jen20:
Yes, I'll open a new PR for the 3.0 branch - is there a particular target branch that's accepting breaking changes there right now?

lukehoban:
> is there a particular target branch that's accepting breaking changes there right now?

Yes - the `feature-3.0` branch is currently accepting breaking changes for 3.0.

https://github.com/pulumi/pulumi/tree/feature-3.0

PR #6554, opened 2021-03-17
[breaking] Changing the version of go.mod in sdk / pkg to be v3

stack72:
This means imports now need to be:

```
github.com/pulumi/pulumi/sdk/v3/
github.com/pulumi/pulumi/pkg/v3/
```

There were a few chicken and egg moments in this PR due to the Go tests and needing to get some version downloaded before replacing it.

The plan for this is to pin to a specific commit for now and when this gets to a beta release, we will update all of the commits

We had to do the same with pulumi-random as we rely on that for a test - we pull in a feature-3.0 based commit and then we will update that commit when we go to beta

One last thing is that we will be cleaning up all of the extra go.mod files we added when we have a version of v3 pkg and sdk modules upon release

**NOTE**: I expect the docs generation to fail here as the docs repo hasn't been updated to deal with v3 libraries


pkg/codegen/go/gen.go
```diff
@@ -1575,7 +1575,7 @@ func (pkg *pkgContext) getImports(member interface{}, importsAndAliases map[stri
 		return
 	}
 
-	importsAndAliases["github.com/pulumi/pulumi/sdk/v2/go/pulumi"] = ""
+	importsAndAliases["github.com/pulumi/pulumi/sdk/v3/go/pulumi"] = ""
```
justinvp, line 1578: I assume these changes to the codegen are intended.

justinvp (approved):
I did a quick pass over the changes. There seems to be some unrelated commits in this PR (e.g. file renames like `sdk/nodejs/x/automation/stack.ts → sdk/nodejs/automation/stack.ts`), but aside from that and a question about the codegen changes, LGTM.

pkg/codegen/go/gen.go
stack72: yeah this is a update to generate the new code based on the new sdk version

Issue #6565, opened 2021-03-18
Homebrew install of Pulumi doesn't show the version


stack72:
```
▶ which pulumi
/usr/local/bin/pulumi
```

```
▶ pulumi version

warning: A new version of Pulumi is available. To upgrade from version '0.0.0' to '2.22.0', run
   $ brew upgrade pulumi
or visit https://pulumi.com/docs/reference/install/ for manual instructions and release notes.
```

Suspected that ldflag isn't working as expected during the build

stack72:
The bump of the homebrew package is as follows - https://github.com/Homebrew/homebrew-core/commit/02acd866e2b4ec1018896e10591777abe23dde91

stack72:
The issue here is that homebrew installs run "make brew" to install Pulumi. Since the 3.0.0 release machinery changes, Pulumi now uses "pulumictl" to determine the version of the software - this isn't available as part of the homebrew setup on a customer machine so therefore the version passed to the ldflag is empty and thus an empty response from pulumi verison command

leshibily:
is this fixed?

stack72:
Hi @leshibily 

I am preparing the release right now

leshibily:
Hi @stack72 

Thanks for working hard to get the right away

PR #6566, opened 2021-03-18
Ensure that make brew works as expected rather than passing empty version

stack72:
Fixes: #6565

As part of #6460, the logic for determing the version of the build was
moved to be a dependency on pulumictl.

Unfortunately, the homebrew installs use the "make dist" command to
build + install Pulumi to the user maching and as that would have a
dependency on pulumictl and it not existing on the user machine, it
would pass an empty version to the ldflag

This then manifested to the user as:

```
▶ pulumi version
warning: A new version of Pulumi is available. To upgrade from version '0.0.0' to '2.22.0', run
   $ brew upgrade pulumi
or visit https://pulumi.com/docs/reference/install/ for manual instructions and release notes.
```

We are able to mitigate this behaviour by bringing back the get-version
script and using that script as part of the make brew installation

We can see that the versions are the same between the 2 different
installation techniques

```
make dist <------- uses pulumict
DIST:
go install -ldflags "-X github.com/pulumi/pulumi/sdk/v2/go/common/version.Version=2.24.0-alpha.1616029310+787eb70a" github.com/pulumi/pulumi/sdk/v2/dotnet/cmd/pulumi-language-dotnet
DIST:
BUILD:
```

```
make brew <----- uses the legacy script
▶ make brew
BREW:
go install -ldflags "-X github.com/pulumi/pulumi/sdk/v2/go/common/version.Version=v2.24.0-alpha.1616029310+g787eb70a2" github.com/pulumi/pulumi/sdk/v2/dotnet/cmd/pulumi-language-dotnet
BREW:
```

A full post mortem will be carried out to ensure we mitigate these
types of errors going forward and that we are able to better test
these types of situations

### DIFF
--- a/sdk/dotnet/Makefile
+++ b/sdk/dotnet/Makefile
@@ -70,8 +70,9 @@ test_all:: dotnet_test
 dist::
 	go install -ldflags "-X github.com/pulumi/pulumi/sdk/v2/go/common/version.Version=${DOTNET_VERSION}" ${LANGHOST_PKG}
 
-brew:: dist
-	go install -ldflags "-X github.com/pulumi/pulumi/sdk/v2/go/common/version.Version=${DOTNET_VERSION}" ${LANGHOST_PKG}
+brew:: BREW_VERSION := $(shell ../../scripts/get-version HEAD)
+brew::
+	go install -ldflags "-X github.com/pulumi/pulumi/sdk/v2/go/common/version.Version=${BREW_VERSION}" ${LANGHOST_PKG}
 
 publish:: build install
 	echo "Publishing .nupkgs to nuget.org:"
--- a/Makefile
+++ b/Makefile
@@ -40,8 +40,9 @@ dist:: build
 	cd pkg && go install -ldflags "-X github.com/pulumi/pulumi/pkg/v2/version.Version=${VERSION}" ${PROJECT}
 
 # NOTE: the brew target intentionally avoids the dependency on `build`, as it does not require the language SDKs.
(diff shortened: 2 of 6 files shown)


sdk/dotnet/Makefile
```diff
@@ -70,8 +70,9 @@ test_all:: dotnet_test
 dist::
 	go install -ldflags "-X github.com/pulumi/pulumi/sdk/v2/go/common/version.Version=${DOTNET_VERSION}" ${LANGHOST_PKG}
 
-brew:: dist
-	go install -ldflags "-X github.com/pulumi/pulumi/sdk/v2/go/common/version.Version=${DOTNET_VERSION}" ${LANGHOST_PKG}
```
lukehoban, line 74: Not sure - but this used `DOTNET_VERSION` which was derived from `pulumictl get version --language dotnet`.  Does that do anything different than the vanilla `scripts/get-version HEAD`?

lukehoban: Also - weird that this was building things twice previously (once in dist and once in brew)?

lukehoban (approved):

sdk/dotnet/Makefile
stack72: DOTNET_VERSION keeps the *new* functionality - the pulumictl based one so we can continue to build the feature-3.0 alphas etc

The BREW_VERSION is a variable specifically using the get-version script - that way, homebrew is the only thing using that script

Yes, I did find that we were building the DotNet binary twice :D

PR #6572, opened 2021-03-18
[automation/dotnet] Expose structured logging

### LINKED ISSUE #6444: [automation/dotnet] - Add structured output to preview/up/refresh/destroy

vipentti:
I wanted to see how to go about implementing the structured event log support for dotnet automation API.

~~This is still WIP and contains TODOs, but I figured it would be a good idea to get some feedback on the implementation in general before going further.~~

~~This implements support for watching the event log via `FileSystemWatcher` which is utilized in a new `EventLogWatcher` class.
The FileSystemWatcher is notified with LastWrite or Size changes which calls the `HandleEvent` method.
The `HandleEvent` currently uses a semaphore to ensure only one event handler reads the file at a time, in case multiple events are triggered in succession.~~

~~Currently events and their json models are in files `Pulumi.Automation/Events/Events.cs` and `Pulumi.Automation/Serialization/EventModels.cs`, once all events are implemented they can be split into separate files.
Not all events are implemented yet.~~

~~**NOTE** All events in `Events.cs` are marked `internal` currently while working on this, since this avoids having to specify and update them continuously in `PublicAPI.Unshipped.txt`. The events **will** be made `public` once all events have been created.~~

Fixes #6444

See also: #6436 (go), #6454 (nodejs), #6527 (python)


sdk/dotnet/Pulumi.Automation/Events/Events.cs
```diff
@@ -0,0 +1,495 @@
...
+
+using System.Collections.Generic;
+using System.Collections.Immutable;
+
+// NOTE: The classes in this file are intended to align with the serialized
+// JSON types defined and versioned in sdk/go/common/apitype/events.go
+namespace Pulumi.Automation.Events
+{
+    // TODO(vipentti): Split to separate files?
+
```
vipentti, line 11: Should the Events be split into separate files per typical C# standards?

sdk/dotnet/Pulumi.Automation/Events/Events.cs
```diff
@@ -0,0 +1,495 @@
...
+        /// <summary>
+        /// Old is the state of the resource before performing the step.
+        /// </summary>
+        public StepEventStateMetadata? Old { get; }
+
+        /// <summary>
+        /// New is the state of the resource after performing the step.
+        /// </summary>
+        // TODO: can this actually ever be null?
+        public StepEventStateMetadata? New { get; }
```
vipentti, line 251: Can `StepEventMetadata.New` ever be null?

sdk/dotnet/Pulumi.Automation/Events/Events.cs
```diff
@@ -0,0 +1,495 @@
...
+        public string? Urn { get; }
+
+        public string? Prefix { get; }
+
+        public string Message { get; }
+
+        public string Color { get; }
+
+        // TODO: Make this into enum?
+        public string Severity { get; }
```
vipentti, line 53: Should Severity be its own `enum`?

sdk/dotnet/Pulumi.Automation/Events/Events.cs
komalali: Yes, on `Destroy` operations. Just like `Old` is `null` on `Create` operations.

sdk/dotnet/Pulumi.Automation/Events/Events.cs
komalali: I'll let C# people speak to this @orionstudt @t0yv0 @justinvp @mikhailshilkov - but in general, I would say yes, follow language standards where appropriate.

sdk/dotnet/Pulumi.Automation/Events/Events.cs
komalali: I agree that it probably could be but I haven't made it one in any of the other languages, I would leave it as-is for now.

sdk/dotnet/Pulumi.Automation/Commands/LocalPulumiCmd.cs
```diff
@@ -25,7 +25,12 @@ internal class LocalPulumiCmd : IPulumiCmd
             // this causes commands to fail rather than prompting for input (and thus hanging indefinitely)
             var completeArgs = args.Concat(new[] { "--non-interactive" });
 
-            var env = new Dictionary<string, string>();
+            var env = new Dictionary<string, string>()
+            {
+                // Required for event log
+                ["PULUMI_DEBUG_COMMANDS"] = "true",
```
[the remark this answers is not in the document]
vipentti: I moved setting of the variable to `Workspace` and behind a conditional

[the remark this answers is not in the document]
orionstudt: That works. But I think the conditional can just check if the delegate is null rather than iterating over the Args. And I think we should find a way to not always instantiate the log watcher even when we don't need it. It seems weird to instantiate a type that may or may not do anything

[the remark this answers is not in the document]
orionstudt: That may be simpler if you make it lower level. For instance if you pass the nullable OnEvent action all the way down to LocalPulumiCmd and have that be the single place that your event log watcher is conditionally instantiated

sdk/dotnet/Pulumi.Automation/Serialization/CancelEventModel.cs
```diff
@@ -0,0 +1,15 @@
...
+using Pulumi.Automation.Events;
+
+// NOTE: The classes in this file are intended to align with the serialized
+// JSON types defined and versioned in sdk/go/common/apitype/events.go
+namespace Pulumi.Automation.Serialization
+{
+    internal class CancelEventModel : IJsonModel<CancelEvent>
+    {
+        public CancelEvent Convert() =>
+            new CancelEvent();
```
orionstudt, line 13: You don't need this model, you can just deserialize directly to `CancelEvent` if it has a parameterless constructor. These models are only necessary to get around System.Text.Json not supporting deserialization of parameterized constructors at this version

sdk/dotnet/Pulumi.Automation/Events/StdoutEngineEvent.cs
```diff
@@ -0,0 +1,23 @@
...
+// NOTE: The class in this file are intended to align with the serialized
+// JSON types defined and versioned in sdk/go/common/apitype/events.go
+
+namespace Pulumi.Automation.Events
+{
+    /// <summary>
+    /// StdoutEngineEvent is emitted whenever a generic message is written, for example warnings
+    /// from the pulumi CLI itself. Less common than DiagnosticEvent.
+    /// </summary>
+    public class StdoutEngineEvent
```
orionstudt, line 11: I think for event naming in dotnet we should prefer full names rather than any abbreviations. StandardOutputEngineEvent or ResourceOutputsEvent

vipentti: How about the property names in `EngineEvent`? Currently they use the abbreviations since that's how the properties are named in the JSON. We could use full names for the properties in `EngineEvent` and the abbreviations in `EngineEventModel`?

orionstudt: That's how I would do it!

sdk/dotnet/Pulumi.Automation/WorkspaceStack.cs
```diff
@@ -376,9 +394,15 @@ public Task RemoveConfigAsync(IEnumerable<string> keys, CancellationToken cancel
                 if (inlineHost != null && inlineHost.TryGetExceptionInfo(out var exceptionInfo))
                     exceptionInfo.Throw();
 
+                if (summaryEvent is null)
+                {
+                    throw new InvalidOperationException("No summary of changes.");
```
orionstudt, line 399: I think maybe this should be a different exception, since this will bubble to the caller and `InvalidOperationException` kind of implies that the caller provided invalid inputs or that they are otherwise doing something they shouldn't - when really this is a pulumi problem.

vipentti: Since we have the `result` available, we could send `CommandException`  or create a subclass of `CommandException` for this particular case?

orionstudt: I think a 1st class exception is the right idea but I don't think it should derive from `CommandException` since those are specific to the CLI writing to the StandardError stream and this case doesn't have that. Could just make a `NoSummaryEventException` or maybe `MissingExpectedEventException` or something

orionstudt:
Alright I still have thoughts on this but I'll try to keep them brief -

I'm not personally a fan of the `SetupEventLogWatcher` method. It seems an anti-pattern to pass in a collection of arguments to a static method and have that static method conditionally mutate that collection.

I also think the `EventLogFileWatcher` class is a bit of an anti-pattern - it exists purely as a wrapper so you have something to `using` every time - even though the `EventLogWatcher` may or may not be instantiated every time. I feel like we should be consistent and just do what was done with `InlineHost` - where it is conditionally instantiated and therefore conditionally disposed. It seems weird to instantiate something that may or may not do anything and then dispose it everytime.

And finally an architectural thought - I really think that the `EventLogWatcher` should be lower level than `WorkspaceStack`. The functionality of the `EventLogWatcher` depends heavily on CLI implementation details (the fact that we are reading a physical log file that the CLI is writing to). Therefore I think we should pass the nullable `Action<EngineEvent>` all the way down into the `LocalPulumiCmd` implementation, since `IPulumiCmd` is supposed to be the abstraction that encapsulates CLI implementation details. This has the added benefit of being the single place that an `EventLogWatcher` is conditionally instantiated, and also we can tie the lifetime of the file watcher to the pulumi child process - which I think is more correct conceptually.

Theoretically in the future if we have an `EnginePulumiCmd` implementation of `IPulumiCmd` that was plugging the arguments directly into the pulumi engine, it may have it's own way of listening to engine events (as opposed to watching a physical log file) - so I think doing it outside of that service may be breaking that abstraction.

sdk/dotnet/Pulumi.Automation/Commands/LocalPulumiCmd.cs
```diff
@@ -36,6 +40,21 @@ internal class LocalPulumiCmd : IPulumiCmd
...
+                // Required for event log
+                // We add it after the provided env vars to ensure it is set to true
+                env["PULUMI_DEBUG_COMMANDS"] = "true";
+
+                eventLogFile = CreateEventLogFile(completeArgs.FirstOrDefault() ?? "event-log");
+                eventLogWatcher = new EventLogWatcher(eventLogFile, onEngineEvent);
+
+                completeArgs = completeArgs.Concat(new[] { "--event-log", eventLogFile });
+            }
+
```
vipentti, line 57: event-log setup now occurs in one place

vipentti:
> And finally an architectural thought - I really think that the `EventLogWatcher` should be lower level than `WorkspaceStack`. The functionality of the `EventLogWatcher` depends heavily on CLI implementation details (the fact that we are reading a physical log file that the CLI is writing to). Therefore I think we should pass the nullable `Action<EngineEvent>` all the way down into the `LocalPulumiCmd` implementation, since `IPulumiCmd` is supposed to be the abstraction that encapsulates CLI implementation details. This has the added benefit of being the single place that an `EventLogWatcher` is conditionally instantiated, and also we can tie the lifetime of the file watcher to the pulumi child process - which I think is more correct conceptually.
>

Per your suggestion @orionstudt  I went with this, I think it turned out simpler over all since the logic for setting up and handling the event-log is only in one place.

orionstudt:
@vipentti I think so too - thank you for considering the suggestion!

sdk/dotnet/Pulumi.Automation/Commands/LocalPulumiCmd.cs
```diff
@@ -131,7 +149,36 @@ internal class LocalPulumiCmd : IPulumiCmd
             proc.Start();
             proc.BeginOutputReadLine();
             proc.BeginErrorReadLine();
-            return await tcs.Task.ConfigureAwait(false);
+            try
+            {
+                return await tcs.Task.ConfigureAwait(false);
```
orionstudt, line 154: @vipentti Do you think this may be an issue in the event that the cancellation token is triggered? Consider this scenario:

1. Process is running with event log watcher
2. Cancellation token is triggered
3. TaskCompletionSource is cancelled, triggering cleanup of event log file because `await tcs.Task` will return
4. Process is killed, but this may not happen instantaneously - it might still be trying to write to the log file

vipentti: That might indeed be the case, is the `Exited` event triggered also when the Process is killed? If so I think we could move the cleanup logic there after the `WaitForExit`

orionstudt: yea I think the bottom of the `Exited` event might be the place for cleanup - I believe it is _always_ triggered but I can't find verification of that

t0yv0:
Starting to digest and review this, lots of great work here. I'm new to Pulumi so it will take me a little bit to get fully up to speed here.

sdk/dotnet/Pulumi.Automation/Commands/LocalPulumiCmd.cs
vipentti: I did some testing trying to figure out in which cases the `Exited` event is called and in which cases it is not.
So far it looks like the `Exited` event **may not**  always be called, at least it was not in my contrived example.
For example when the `CancellationToken` is cancelled and a `System.Threading.Tasks.TaskCanceledException` is thrown,
the `Exited` event may not be called.

Basically what I did was add some extra `Console.Error.Writeline` calls before the `proc.Start()`, inside the `finally`-block,
inside `cancellationToken.Register` before `proc.Kill()` and inside `Exited` eventhandler after `proc.WaitForExit()`.

I then ran the following test code in Debug-mode to produce some debug output

```csharp
[Fact]
public async Task ProcessExitedEventOnCancellation()
{
    var program = PulumiFn.Create(() =>
    {
        {
            ["exp_static"] = "foo",
        };
    });
    var projectName = "process_exited_test";
    var stackName = $"exited_on_cancel{GetTestSuffix()}";
    using var stack = await LocalWorkspace.CreateStackAsync(new InlineProgramArgs(projectName, stackName, program)
    {
        EnvironmentVariables = new Dictionary<string, string>()
        {
            ["PULUMI_CONFIG_PASSPHRASE"] = "test",
        }
    });

    try
    {
        var cts = new CancellationTokenSource();
        var previewTask = stack.PreviewAsync(cancellationToken: cts.Token);
        // Using cts.Cancel() produces similiar end results, with slightly different
        // results in the output
        cts.CancelAfter(0);
        await previewTask;
    }
    catch(Exception ex)
    {
        Console.Error.WriteLine("Uncaught exception: {0}", ex);
        throw;
    }
    finally
    {
        await stack.Workspace.RemoveStackAsync(stackName);
    }
}
```

Running this produces the following debug output (formatted for readability)

```yaml
# First two commands succeed and the order of the messages is the same
Start: Process 'version --non-interactive'
ExitedEvent: Process 'version --non-interactive' ExitCode=0
Finally: Process 'version --non-interactive' HasExited=True

Start: Process 'stack init exited_on_cancel191679 --non-interactive'
ExitedEvent: Process 'stack init exited_on_cancel191679 --non-interactive' ExitCode=0
Finally: Process 'stack init exited_on_cancel191679 --non-interactive' HasExited=True

# This call will fail due to the cancellation
Start: Process 'stack select exited_on_cancel191679 --non-interactive'
# NOTE 1: proc.HasExited is false here
# NOTE 2: No ExitedEvent for this command
# NOTE 3: The Register callback is called further below
Finally: Process 'stack select exited_on_cancel191679 --non-interactive' HasExited=False
# This is printed inside the test code above
Uncaught exception: System.Threading.Tasks.TaskCanceledException: A task was canceled.
   at Pulumi.Automation.Commands.LocalPulumiCmd.RunAsync(IEnumerable`1 args, String workingDir, IDictionary`2 additionalEnv, Action`1 onStandardOutput, Action`1 onStandardError, Action`1 onEngineEvent, CancellationToken cancellationToken) in <path>/sdk/dotnet/Pulumi.Automation/Commands/LocalPulumiCmd.cs:line 161
   at Pulumi.Automation.WorkspaceStack.PreviewAsync(PreviewOptions options, CancellationToken cancellationToken) in <path>/sdk/dotnet/Pulumi.Automation/WorkspaceStack.cs:line 314
   <snip>

# This is the RemoveStackAsync call inside the test
Start: Process 'stack rm --yes exited_on_cancel191679 --non-interactive'

# Register callback gets called
Register: Killing process 'stack select exited_on_cancel191679 --non-interactive'
# proc.Kill() throws
Register: Killing process 'stack select exited_on_cancel191679 --non-interactive' error System.InvalidOperationException No process is associated with this object.
   at System.Diagnostics.Process.EnsureState(State state)
   at System.Diagnostics.Process.Kill()
   at Pulumi.Automation.Commands.LocalPulumiCmd.<>c__DisplayClass0_0.<RunAsync>b__2() in <path>/sdk/dotnet/Pulumi.Automation/Commands/LocalPulumiCmd.cs:line 114

# RemoveStackAsync succeeds
ExitedEvent: Process 'stack rm --yes exited_on_cancel191679 --non-interactive' ExitCode=0
Finally: Process 'stack rm --yes exited_on_cancel191679 --non-interactive' HasExited=True
```

At least based on this contrived example, running the cleanup in the `Exited` event handler may not always end up disposing the `EventLogWatcher`.

To ensure the process gets killed before removing the watcher and the file, when cancellation is requested, we could add the following inside the `finally`-block

```csharp
// The process may not have exited because cancellation was requested
// Ensure the process gets killed before cleaning up resources
if (!proc.HasExited)
{
    try { proc.Kill(); } catch { /* ignored */ }
    try { proc.WaitForExit(); } catch { /* ignored */ }
}
```

t0yv0:
Can we consider polling instead of FileSystemWatcher? I have two issues with FSW:

1. does not detect intermediate file changes until the writer closes the file (at least on Mac OS X)

Here is an example writer:

```python
import time

i = 0

with open('a.log', 'a') as fp:
    while True:
        fp.write('Gotcha: {}\n'.format(i))
        print('Gotcha: {}\n'.format(i))
        time.sleep(1)
        i = i + 1
        fp.flush()
```

In my experimentation FileSystemWatcher does not get notified of these writes (either via events interface or via synchronous WaitForChanged) until I kill the writer process. In contrast, a polling loop detects the changes immediately.

```csharp
        private async Task ObserveFile() {
            while (true) {
                await Task.Delay(1000);
                using var fs = new FileStream(_fullPath, FileMode.Open, FileAccess.Read);
                Console.WriteLine("Observing {0}", fs.Length);
            }
        }

        private async Task ObserveFSW() {
            var logFile = _fullPath;
            using var fsw = new FileSystemWatcher 
            {
                Path = Path.GetDirectoryName(logFile),
                Filter = Path.GetFileName(logFile),
            };
            while (true) {
                var chg = fsw.WaitForChanged(WatcherChangeTypes.All, 1000);
                Console.WriteLine("FSW Detecting change: {0}", chg.ChangeType);
                using var fs = new FileStream(_fullPath, FileMode.Open, FileAccess.Read);
                Console.WriteLine("FSW Observing {0}", fs.Length);
            }
        }
```

2. General irrational FUD. I have not been working with .NET recently (years), but a long time ago I spent a great deal of time debugging memory issues with FileSystemWatcher. It might be a lot more reliable now in .NET core days but it still is a complicated piece of machinery (notify? on linux, some there OS API on Mac and Windows), and not a straightforward interface either - which thread does the callback run on, can the callback block that thread, what happens if a lot of events are triggered at once, etc. So I am thinking of something like, long-running user callback holds the lock too long, FSW generates multiple threads that wait on the lock, how does this degrade?

Now the concrete proposal here is to use a polling loop. It would waste cycles in the case of long-running commands, but is "simpler" and automatically detects intermediate writes of long-running commands.

Counter-proposal: if we wanted to continue with FSW, it would work well if the log writer periodically check-pointed the log file, that is, closed the file descriptor and reopened it. I'd like to poke around and find out how it does it at the moment, and whether we could make it checkpoint.

sdk/dotnet/Pulumi.Automation/Commands/LocalPulumiCmd.cs
orionstudt: It makes sense that if the process hasn't started yet than the `Exited` event wouldn't be triggered on cancellation.

Does it really need to be that complex though? Couldn't your finally block just be:
```csharp
if (!proc.HasExited)
    proc.WaitForExit();

// do cleanup
```

t0yv0: Can we consider pro/con of taking a dependency here for example:

https://github.com/Tyrrrz/CliWrap 

This seems actively maintained and is a package trying to address the deficiencies of System.Diagnostics.Process API, in particular ability to safely async-wait and cancel.

orionstudt: I think that is definitely worth considering and there is a good argument for that. But I think if we go that route it should be a separate PR after we get the log file watching working

t0yv0: That sounds great! I can try drafting the PR if it sounds like a good idea. I'm also checking internally what is Pulumi's policy on dependencies, e.g. how conservative we want to be here.

orionstudt: Or I guess it doesn't need to be after we get this working - if you think it would make managing process lifetime simpler. But I do think it should be a separate PR.

I think that I made a mistake not going with a library like that to manage the child process to begin with. But we could ask Mikhail or Justin their thoughts on taking the dependency - I think if it's fairly lightweight, actively maintained, and we're not exposing it publicly they won't mind at all.

sdk/dotnet/Pulumi.Automation/Events/PolicyEvent.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        public string PolicyName { get; }
+
+        public string PolicyPackName { get; }
+
+        public string PolicyPackVersion { get; }
+
+        public string PolicyPackVersionTag { get; }
+
+        public string EnforcementLevel { get; }
```
t0yv0, line 26: Perhaps a TODO note here to consider an enum later:
 
```	// EnforcementLevel is one of "warning" or "mandatory".
	EnforcementLevel string `json:"enforcementLevel"`
```

sdk/dotnet/Pulumi.Automation/Commands/LocalPulumiCmd.cs
t0yv0: Yes that sounds wise, @mikhailshilkov any thoughts on adding CliWrap dependency. Justin is temporarily unavailable.

I am just looking at this code here and I remember struggling with something very similar circa 2013. The concurrency bugs around subprocess management times multiple Ones can take a toll. I do not know if the CliWrap ironed it all out but I like the idea of leaving it to OSS and solving this problem once and for all in there, if we can.

mikhailshilkov: Off the top of my head, I don't have any concerns about it.

sdk/dotnet/Pulumi.Automation/Events/DiagnosticEvent.cs
```diff
@@ -0,0 +1,46 @@
...
+    {
+        public string? Urn { get; }
+
+        public string? Prefix { get; }
+
+        public string Message { get; }
+
+        public string Color { get; }
+
+        public string Severity { get; }
```
t0yv0, line 21: Perhaps a comment here to match go, or consider an enum:

```
	// Severity is one of "info", "info#err", "warning", or "error".
```

komalali: I think a comment here should suffice for now. I don't want to make this an enum since it isn't one in any of the other languages and I don't want to back us into a corner in case there's a change. We can consider an enum in the future when we are ready to commit to this set of values.

vipentti:
Per your suggestion @t0yv0  I went with the polling EventLogWatcher. It should be consistent on all platforms. I don't think the overhead of polling should be a problem.

sdk/dotnet/Pulumi.Automation/Events/EventLogWatcher.cs
```diff
@@ -0,0 +1,123 @@
...
+            _internalCancellationTokenSource.Cancel();
+            await _pollingTask.ConfigureAwait(false);
+        }
+
+        private async Task PollForEvents()
+        {
+            var internalToken = _internalCancellationTokenSource.Token;
+            var externalToken = _externalCancellationToken;
+
+            using var linkedSource = CancellationTokenSource.CreateLinkedTokenSource(internalToken, externalToken);
```
orionstudt, line 48: can you help me understand the function of the internal CTS?

vipentti: The internal CTS allows us to stop polling when the watcher is being disposed by calling its `Cancel` method. Using `CreateLinkedTokenSource` ensures that we respect both our internal CTS as well as the user provided token, so if the 
user token ends up requesting cancellation we also stop polling.

orionstudt: hmmm OK well that makes sense functionally. I wonder if it could just be another boolean. The internal CTS needs to be disposed as well

vipentti: A boolean flag could work, but it would mean that we'd end up waiting for the `Task.Delay` to finish before stopping, with cancelling the CTS the Task.Delay will throw `TaskCanceledException` which we catch to `break` from the polling. If the polling interval is sufficiently small, I don't think this makes much of a practical difference, but if the polling interval was sufficiently large, for example a second or more we would end up waiting longer before stopping.

orionstudt: no you're right this is fine - For some reason I was thinking that disposing the internal CTS so soon after cancelling would cause issues but since you are awaiting the polling task before disposing it is perfectly fine

t0yv0:
Just FYI from some joint digging with @komalali:

1. Go SDK uses fsnotify - https://github.com/fsnotify/fsnotify - it is interesting to see the homepage of that project as it shows what OS APIs are utilized under the hood

2. Node SDK uses logdna/tail-file https://www.npmjs.com/package/@logdna/tail-file - it avoids file watcher APIs and uses polling for similar concerns as we have here

So given there's at least a precedent for polling I think it's good to start with that and upgrade to FileSystemWatcher when/if polling is insufficient and we can test it out.

komalali:
Yup, I will also be using polling for the python implementation so 👍🏽

sdk/dotnet/Pulumi.Automation/Events/EventLogWatcher.cs
```diff
@@ -0,0 +1,123 @@
...
+
+using System;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using Pulumi.Automation.Serialization;
+
+namespace Pulumi.Automation.Events
+{
+    internal class EventLogWatcher : IAsyncDisposable
```
t0yv0, line 11: OK this is pretty cool. I did not quite understand cancellation so I tried to write my own version and test it out and eventually ended up with something extremely similar to this so it may not be worth to change much here.

The principal difference where I landed was handling exceptions, might be worth a consideration. I err on the side of never ignoring exceptions either from polling code or from user callbacks, and instead terminate the polling task with the first exception and propagate it to the code that is calling EventLogWatcher, so it would fail also. It appears from IAsyncDisposable interface that it might not be appropriate to throw from DisposeAsync, so I ended up with a simple IDisposable but the cleanup is a separate call:

```
async Task Stop();
``` 

And then we basically propagate any exception that might have been raised as a result of this Stop task.

sdk/dotnet/Pulumi.Automation/Events/EventLogWatcher.cs
```diff
@@ -0,0 +1,123 @@
...
+                    await Task.Delay(_pollingIntervalMilliseconds, linkedToken).ConfigureAwait(false);
+                    ReadEvents();
+                }
+                catch (TaskCanceledException)
+                {
+                    break;
+                }
+                catch
+                {
+                    // We do our best effort to keep reading until the polling is cancelled
```
t0yv0, line 68: First location of ignored exceptions.

sdk/dotnet/Pulumi.Automation/Events/EventLogWatcher.cs
```diff
@@ -0,0 +1,123 @@
...
+            }
+
+            // Try reading events one final time
+            try
+            {
+                ReadEvents();
+            }
+            catch
+            {
+                // ignored
```
t0yv0, line 79: Ditto.

sdk/dotnet/Pulumi.Automation/Events/EventLogWatcher.cs
```diff
@@ -0,0 +1,123 @@
...
+            {
+                var @event = _localSerializer.DeserializeJson<EngineEvent>(line);
+
+                try
+                {
+                    _onEvent.Invoke(@event);
+                }
+                catch
+                {
+                    // Don't let the provided event handler cause reading of events to fail
```
t0yv0, line 116: Ditto.

sdk/dotnet/Pulumi.Automation/Events/EventLogWatcher.cs
```diff
@@ -0,0 +1,123 @@
...
+            while (!linkedToken.IsCancellationRequested)
+            {
+                try
+                {
+                    // NOTE: Waiting before reading so that if ReadEvents throws we will
+                    // wait before attempting it again
+                    await Task.Delay(_pollingIntervalMilliseconds, linkedToken).ConfigureAwait(false);
+                    ReadEvents();
+                }
+                catch (TaskCanceledException)
```
t0yv0, line 62:
```
            catch (OperationCanceledException error) when (error.CancellationToken == linkedToken)
```

To make it precise to the token pair that we control.

sdk/dotnet/Pulumi.Automation/Events/EventLogWatcher.cs
```diff
@@ -0,0 +1,123 @@
...
+            if (newLength == _previousLength)
+            {
+                return;
+            }
+
+            fs.Seek(_previousLength, SeekOrigin.Begin);
+
+            using var reader = new StreamReader(fs);
+
+            var lines = reader
```
t0yv0, line 102: This bit me on a badly behaved test, not sure if it's super relevant but worth considering. A fast running chatty command will be generating lots of data here, which will require O(N) ram as we load it all up, moreover, we do not process the initial events until we finish ReadToEnd. 

An alternative would be to loop over the StreamReader with ReadLine or ReadLineAsync.

sdk/dotnet/Pulumi.Automation/Events/EventLogWatcher.cs
orionstudt: I don't think we need to propagate exceptions from cleanup - since it is us attempting to delete a temp folder as a courtesy, that the user's OS should be cleaning up on its own anyways. We have precedence for ignoring an error here in LocalWorkspace disposal as well.

But I agree we should propagating the exceptions up everywhere else.

sdk/dotnet/Pulumi.Automation/Events/EventLogWatcher.cs
t0yv0: I've enjoyed the CreateLinkedTokenSource, seems like a great find that fits the problem well.

sdk/dotnet/Pulumi.Automation/Events/EventLogWatcher.cs
t0yv0: https://github.com/vipentti/pulumi/pull/2/files the version I ended up with for reference.

orionstudt: gotcha. Yea it does make sense to separate out the `Stop` function from the dispose if we want to propagate errors from it. Where directory cleanup would remain in the `Dispose` action

t0yv0:
I feel like this is getting very close, a few questions I had:

- is there some unfinished work that is not obvious, like "This is still WIP and contains TODOs, but I figured it would be a good idea to get some feedback on the implementation in general before going further." - are there still any TODOs before merging?

- Would we like to backlog some JSON serialization tests (that'd be separate PR), perhaps something I could help out with. Since the JSON forms of these events form a contract so to say between Pulumi CLI and .NET SDK but the contract is not auto-generated from something like gRPC, we in the long term might want these kind of test to remind us if we accidentally change this contract. Or would that be a bit overkill?

vipentti:
> * is there some unfinished work that is not obvious, like "This is still WIP and contains TODOs, but I figured it would be a good idea to get some feedback on the implementation in general before going further." - are there still any TODOs before merging?

I think once https://github.com/vipentti/pulumi/pull/2 is included and I've done one final pass on the doc-comments on Events this should be good to go

> * Would we like to backlog some JSON serialization tests (that'd be separate PR), perhaps something I could help out with. Since the JSON forms of these events form a contract so to say between Pulumi CLI and .NET SDK but the contract is not auto-generated from something like gRPC, we in the long term might want these kind of test to remind us if we accidentally change this contract. Or would that be a bit overkill?

I think this sounds like a good idea, considering that if the engine events fail to deserialize it will cause exceptions to be thrown by the `EventLogWatcher` when it is being utilized. And since the `LocalWorkspaceTests` do not necessarily generate all of the possible events some problems may only surface during an actual update.

t0yv0 (approved):
LGTM

Issue #6090, opened 2021-01-12
`log.error` does not fail the program


lukehoban:
The documentation for `log.error` (and its equivalents in other languages) indicates that:

> error logs a **fatal error to indicate that the tool should stop processing resource operations immediately**.

Currently though, `log.error` does not cause the program to terminate.  It *does* report an "error: ..." message, but neither the client SDK nor the engine triggers a failure of any kind, so the program will continue to run. And of course, ultimately the  preview/update operation will return a success error code as there was no failure.

```
import * as pulumi from "@pulumi/pulumi";
pulumi.log.error("boom");
```

It appears this has always been the case, so it is not immediately clear whether the documentation is incorrect, or the implementation.

There are three options;
1. Treat this as by-design and fix the documentation.  Programs would need to raise unhandled exceptions to terminate the language host to indicate failure.
2. Treat as a bug, and fix by having the client SDKs throw an unhandled exception in their `log.error` implementations after sending the `Log` RPC to the engine, so that the client fails in a "normal" way immediately after sending the log message.
3. Treat as a bug, and fix by having the engines `Log` RPC server handle severity of ERROR as a failure.

(3) is non-trivial, as the logging machinery in the engine does not participate in the deployment process at all currently. And making it participate in the deployment process may have quite a lot of subtle interactions with reliability of timely logging.  However, as long as (3) is not fixed, it will be possible for "error:..." diagnostics to be rendered without failing the program.

My inclination is that (2) is the right solution, though it is a breaking change we would want to consider in a major release.

EvanBoyle:
If you trace through the following code, it looks like we thought we were catching logged errors: https://github.com/pulumi/pulumi/blob/2585b86aa4252b1ea430dbfd2324c47ea57763b1/sdk/nodejs/cmd/run/index.ts#L157

This bottoms out on some uncaught handlers attached to the process that are supposed to be triggered by `console.error`.

My experimentation shows that `console.error` does not trigger `process.on("uncaughtException"...`:

```js
process.on('uncaughtException', function(err) { console.log('Caught exception: ' + err); });
console.error("foo") // not caught
var x = null;
x[1] // caught and logged
```

`console.error` just prints to stderr according to nodejs docs: https://nodejs.org/api/console.html#console_console_error_data_args

EvanBoyle:
[...]

PR #6573, opened 2021-03-18
Fix docstring on node log.error and other SDK equivalents

t0yv0:
Resolves https://github.com/pulumi/pulumi/issues/6090 by updating docs to reflect current non-terminating behavior of log.error and equivalents.

- [x] node SDK
- [x] golang SDK
- [x] dotnet SDK
- [x] python SDK

### DIFF
--- a/sdk/dotnet/Pulumi/Log.cs
+++ b/sdk/dotnet/Pulumi/Log.cs
@@ -31,17 +31,15 @@ namespace Pulumi
             => Deployment.InternalInstance.Logger.WarnAsync(message, resource, streamId, ephemeral);
 
         /// <summary>
-        /// Logs a fatal condition. Calling Error does not by itself
-        /// stop processing resource operations. To stop programs
-        /// should raise unhandled exceptions after calling Error.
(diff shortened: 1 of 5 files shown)


sdk/dotnet/Pulumi/Log.cs
```diff
@@ -31,15 +31,17 @@ public static void Warn(string message, Resource? resource = null, int? streamId
             => Deployment.InternalInstance.Logger.WarnAsync(message, resource, streamId, ephemeral);
 
         /// <summary>
-        /// Logs a fatal error to indicate that the tool should stop processing resource
-        /// operations immediately.
+        /// Logs a fatal condition. Calling Error does not by itself
+        /// stop processing resource operations. To stop programs
+        /// should raise unhandled exceptions after calling Error.
```
mikhailshilkov, line 36: Nit: something like "Raise an unhandled exception after calling Error to stop the program" seems more natural to me. Or "You should raise...".

mikhailshilkov (approved):
LGTM

EvanBoyle (approved):

Issue #6348, opened 2021-02-16
Addition of a function in the automation API to expose the underlying CLI version


stack72:
This would be something that exposes `pulumi version`

## Languages
- [x] Go - https://github.com/pulumi/pulumi/pull/6577
- [x] Python - https://github.com/pulumi/pulumi/pull/6589
- [x] Nodejs - https://github.com/pulumi/pulumi/pull/6580
- [x] Dotnet - https://github.com/pulumi/pulumi/pull/6590

PR #6575, opened 2021-03-18
[automation/*] - Add version command

komalali:
Fixes: https://github.com/pulumi/pulumi/issues/6348


komalali:
Favoring https://github.com/pulumi/pulumi/pull/6577 style for each language

Issue #6419, opened 2021-02-24
[automation] - Automation API fails with CLI error if CLI version is out of date


komalali:
Automation API should have a mechanism for alerting the user if the minimum required version of the CLI isn't installed.

## Expected behavior

Some sort of "This version of automation API requires minimum CLI version [v.x.x.x]"

## Current behavior

Error raised by the CLI when a command is executed.
```
Error: code: 1
 stdout: 
 stderr: Error: unknown flag: --page-size
```

## Languages
- [x] Go - https://github.com/pulumi/pulumi/pull/6577
- [x] Python - https://github.com/pulumi/pulumi/pull/6589
- [x] Nodejs - https://github.com/pulumi/pulumi/pull/6580
- [x] Dotnet - https://github.com/pulumi/pulumi/pull/6590

komalali:
So, I think we should actually handle this on a case-by-case basis rather than imposing a minimum version requirement on all of Automation SDK. Moving forward, if we add features to the Automation SDK that have a minimum version requirement from the CLI outside of a major version bump this should be considered as part of the error handling for that feature (i.e. handle the error returned by the CLI and provide a more user-friendly/actionable "This feature requires v[x.x.x] of the Pulumi CLI" error).

PR #6577, opened 2021-03-18
[automation/go] - Implement min version checking

komalali:
## Changes

* Added a `LocalWorkspace.PulumiVersion()` method that returns the underlying Pulumi CLI version.
* `NewLocalWorkspace()` now checks the underlying pulumi version during creation and stores it on the LocalWorkspace struct. 
* `NewLocalWorkspace()` checks that the `PulumiVersion` >= the minimum version set in `minimum_version.go` and returns an error to update the CLI if the check fails.

Go part of https://github.com/pulumi/pulumi/issues/6348 and https://github.com/pulumi/pulumi/issues/6419


sdk/go/x/auto/local_workspace.go
```diff
@@ -46,6 +47,55 @@ type LocalWorkspace struct {
...
+}
+
+func (v Version) String() string {
+	if v.Suffix != "" {
+		return fmt.Sprintf("v%v.%v.%v-%s", v.Major, v.Minor, v.Patch, v.Suffix)
+	}
+	return fmt.Sprintf("v%v.%v.%v", v.Major, v.Minor, v.Patch)
+}
+
+func parseVersion(versionString string) (Version, error) {
```
justinvp, line 67: Any reason not to use https://github.com/blang/semver and one of it's `Parse` functions (`ParseTolerant`?) like we do elsewhere?

Also, wonder if we should just return `semver.Version` rather than defining our own `Version` struct. We already return it publicly elsewhere in the Go SDK, e.g.:

https://github.com/pulumi/pulumi/blob/b82b850af0d62461d47ce8948f62ef011c8fdd7b/sdk/go/pulumi/rpc.go#L703-L705

komalali: Only so I would have the opportunity to do it all over again 😄

sdk/go/x/auto/minimum_version.go
```diff
@@ -0,0 +1,23 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package auto
+
+import "github.com/blang/semver"
+
+var minimumVersion = semver.Version{
```
stack72, line 19: IS there anything we are going to need to do in our 3.0 work relating to this? Is this something we need to bump on each release?

stack72 (approved):
@komalali this is great - a nice non-intrusive way to calculate this that only hits on the workspace creation. I am really worried about the version being embedded in this - it seems we need to do that for all of our code which means we need to remember to bump it in 4 places if we need to change it

I am not sure how much effort we want to put into this feature at this point but this feels like something we need to give some thought about

EvanBoyle:
> which means we need to remember to bump it in 4 places if we need to change it

This problem is kinda inherent to any feature and/or bug fix made to automation API. Just part of the game, unfortunately.

stack72:
> > which means we need to remember to bump it in 4 places if we need to change it
> 
> This problem is kinda inherent to any feature and/or bug fix made to automation API. Just part of the game, unfortunately.

Totally understand, it's definitely why I still hit approve as it's something we just need to be aware of rather than blocking on it

But you are 100% correct here

sdk/go/x/auto/minimum_version.go
komalali: > Is there anything we are going to need to do in our 3.0 work relating to this?

Yes, for the 3.0 release we will need to bump this to 3.0.0. 

> Is this something we need to bump on each release?

No, only when there is a change to the automation SDK that is dependent on a CLI change.

sdk/go/x/auto/local_workspace.go
```diff
@@ -479,6 +487,32 @@ func (l *LocalWorkspace) ImportStack(ctx context.Context, stackName string, stat
 	return nil
 }
 
+func (l *LocalWorkspace) getPulumiVersion(ctx context.Context) (semver.Version, error) {
+	stdout, stderr, errCode, err := l.runPulumiCmdSync(ctx, "version")
```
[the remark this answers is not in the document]
lukehoban: (Same idea of course applies in all languages)

sdk/go/x/auto/minimum_version.go
lukehoban: > Yes, for the 3.0 release we will need to bump this to 3.0.0.

Out of curiosity - is this true?  Will 3.0 of automation API require that a 3.0 CLI is installed?  It's not immediately obvious it would have to?

Related - is it a breaking change to increase this?  I assume so - which will make us want to do this very infrequently.

sdk/go/x/auto/local_workspace.go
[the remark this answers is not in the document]
komalali: > Should we make this call asynchronously, and only block when a user (directly or indirectly via some other api call that internally checks the version) requests the version?

@lukehoban the design choice here for ensuring that we have the minimum version of the CLI installed (as a means for addressing https://github.com/pulumi/pulumi/issues/6419) is that we will validate the `pulumi version` against the minimum valid version at `Workspace` creation. We chose to only make this call once so we could then store it on the `Workspace` object and avoid redoing that work later. 

Typical performance impact would vary by language but a benchmark test of Workspace creation in Go result in `0.07434 ns` so I feel pretty comfortable taking that hit :) 

Screen Shot 2021-03-22 at 8 05 45 PM

sdk/go/x/auto/minimum_version.go
komalali: Yes. We made breaking changes to the CLI in 3.0 (standardizing select behavior) that we use under the hood in Automation API, so we will need to bump the minimum version.

sdk/go/x/auto/local_workspace.go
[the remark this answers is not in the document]
lukehoban: Great to see the perf impact is so small!

PR #6584, opened 2021-03-19
[automation/dotnet] - Isolate project and stack names, defer stack removal for tests

komalali:


t0yv0 (approved):
Basically adding finally RemoveStackAsync right? Looking good.

PR #6590, opened 2021-03-20
[automation/dotnet] - Implement min version checking

komalali:
## Changes
* Added a `LocalWorkspace.PulumiVersion` field that returns the underlying Pulumi CLI version.
* `LocalWorkspace` constructor checks that the `PulumiVersion` >= `LocalWorkspace._minimumVersion` and raises an exception to update the CLI if the check fails.

Please let me know if there's a more idiomatic way to do this!

C# part of #6348 and #6419

Also see: https://github.com/pulumi/pulumi/pull/6577 for Go, https://github.com/pulumi/pulumi/pull/6580 for nodejs and https://github.com/pulumi/pulumi/pull/6589 for python

### DIFF
--- a/sdk/dotnet/Pulumi.Automation.Tests/LocalWorkspaceTests.cs
+++ b/sdk/dotnet/Pulumi.Automation.Tests/LocalWorkspaceTests.cs
@@ -848,12 +848,12 @@ namespace Pulumi.Automation.Tests
             var testCurrentVersion = SemVersion.Parse("2.21.1");
             if (errorExpected)
             {
-                Action act = () => LocalWorkspace.CheckVersionIsValid(minVersion, testCurrentVersion);
+                Action act = () => LocalWorkspace.ValidatePulumiVersion(minVersion, testCurrentVersion);
                 Assert.Throws<InvalidOperationException>(act);
             }
             else
             {
-                LocalWorkspace.CheckVersionIsValid(minVersion, testCurrentVersion);
+                LocalWorkspace.ValidatePulumiVersion(minVersion, testCurrentVersion);
             }
         }
     }
--- a/sdk/dotnet/Pulumi.Automation.Tests/LocalWorkspaceTests.cs
+++ b/sdk/dotnet/Pulumi.Automation.Tests/LocalWorkspaceTests.cs
@@ -826,5 +826,12 @@ namespace Pulumi.Automation.Tests
                 Assert.True(expSecretValue.IsSecret);
             }
         }
+    
+        [Fact]
+        public async Task PulumiVersionTest()
+        {
+            using var workspace = await LocalWorkspace.CreateAsync();
+            Assert.Matches("(\\d+\\.)(\\d+\\.)(\\d+)(-.*)?", workspace.PulumiVersion.ToString());
+        }
     }
 }
--- a/sdk/dotnet/Pulumi.Automation.Tests/LocalWorkspaceTests.cs
+++ b/sdk/dotnet/Pulumi.Automation.Tests/LocalWorkspaceTests.cs
@@ -832,7 +832,7 @@ namespace Pulumi.Automation.Tests
         public async Task PulumiVersionTest()
         {
             using var workspace = await LocalWorkspace.CreateAsync();
-            Assert.Matches("(\\d+\\.)(\\d+\\.)(\\d+)(-.*)?", workspace.PulumiVersion.ToString());
+            Assert.Matches("(\\d+\\.)(\\d+\\.)(\\d+)(-.*)?", workspace.PulumiVersion);
         }
 
         [Theory]
--- a/sdk/dotnet/Pulumi.Automation.Tests/LocalWorkspaceTests.cs
+++ b/sdk/dotnet/Pulumi.Automation.Tests/LocalWorkspaceTests.cs
@@ -838,22 +838,24 @@ namespace Pulumi.Automation.Tests
         [Theory]
         [InlineData("100.0.0", true)]
         [InlineData("1.0.0", true)]
-        [InlineData("2.22.0", true)]
-        [InlineData("2.1.0", false)]
-        [InlineData("2.21.2", true)]
+        [InlineData("2.22.0", false)]
+        [InlineData("2.1.0", true)]
+        [InlineData("2.21.2", false)]
         [InlineData("2.21.1", false)]
-        [InlineData("2.21.0", false)]
-        public void ValidVersionTheory(string minVersion, bool errorExpected)
+        [InlineData("2.21.0", true)]
+        // Note that prerelease < release so this case should error
+        [InlineData("2.21.1-alpha.1234", true)]
+        public void ValidVersionTheory(string currentVersion, bool errorExpected)
         {
-            var testCurrentVersion = SemVersion.Parse("2.21.1");
+            var testMinVersion = SemVersion.Parse("2.21.1");
             if (errorExpected)
             {
-                Action act = () => LocalWorkspace.ValidatePulumiVersion(minVersion, testCurrentVersion);
+                Action act = () => LocalWorkspace.ValidatePulumiVersion(testMinVersion, currentVersion);
                 Assert.Throws<InvalidOperationException>(act);
             }
             else
             {
-                LocalWorkspace.ValidatePulumiVersion(minVersion, testCurrentVersion);
+                LocalWorkspace.ValidatePulumiVersion(testMinVersion, currentVersion);
             }
         }
     }
--- a/sdk/dotnet/Pulumi.Automation.Tests/LocalWorkspaceTests.cs
+++ b/sdk/dotnet/Pulumi.Automation.Tests/LocalWorkspaceTests.cs
@@ -1,5 +1,6 @@
 // Copyright 2016-2021, Pulumi Corporation
 
+using Semver;
 using System;
 using System.Collections.Generic;
 using System.IO;
@@ -832,6 +833,28 @@ namespace Pulumi.Automation.Tests
         {
             using var workspace = await LocalWorkspace.CreateAsync();
             Assert.Matches("(\\d+\\.)(\\d+\\.)(\\d+)(-.*)?", workspace.PulumiVersion.ToString());
+        }
+
+        [Theory]
+        [InlineData("100.0.0", true)]
+        [InlineData("1.0.0", true)]
+        [InlineData("2.22.0", true)]
+        [InlineData("2.1.0", false)]
+        [InlineData("2.21.2", true)]
+        [InlineData("2.21.1", false)]
+        [InlineData("2.21.0", false)]
+        public void ValidVersionTheory(string minVersion, bool errorExpected)
+        {
+            var testCurrentVersion = SemVersion.Parse("2.21.1");
+            if (errorExpected)
+            {
+                Action act = () => LocalWorkspace.CheckVersionIsValid(minVersion, testCurrentVersion);
+                Assert.Throws<InvalidOperationException>(act);
+            }
+            else
+            {
+                LocalWorkspace.CheckVersionIsValid(minVersion, testCurrentVersion);
+            }
         }
     }
 }
--- a/sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
+++ b/sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
@@ -1,5 +1,6 @@
 ﻿// Copyright 2016-2021, Pulumi Corporation
 
+using Semver;
 using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
@@ -34,12 +35,27 @@ namespace Pulumi.Automation
         private readonly LocalSerializer _serializer = new LocalSerializer();
         private readonly bool _ownsWorkingDir;
         private readonly Task _readyTask;
+        private static readonly SemVersion _minimumVersion = SemVersion.Parse("2.21.0");
 
         /// <inheritdoc/>
         public override string WorkDir { get; }
 
         /// <inheritdoc/>
         public override string? PulumiHome { get; }
+
+        private SemVersion? pulumiVersion;
+        /// <inheritdoc/>
+        public override SemVersion PulumiVersion 
+        { 
+            get 
+            { 
+                if (pulumiVersion == null)
+                {
+                    throw new ArgumentNullException("pulumiVersion");
+                }
+                return pulumiVersion; 
+            }
+        }
 
         /// <inheritdoc/>
         public override string? SecretsProvider { get; }
@@ -320,6 +336,8 @@ namespace Pulumi.Automation
 
             this.WorkDir = dir;
 
+            readyTasks.Add(this.GetPulumiVersionAsync(cancellationToken));
+
             // these are after working dir is set because they start immediately
             if (options?.ProjectSettings != null)
                 readyTasks.Add(this.SaveProjectSettingsAsync(options.ProjectSettings, cancellationToken));
@@ -334,6 +352,39 @@ namespace Pulumi.Automation
         }
 
         private static readonly string[] SettingsExtensions = new string[] { ".yaml", ".yml", ".json" };
+
+        private async Task GetPulumiVersionAsync(CancellationToken cancellationToken = default)
+        {
+            var result = await this.RunCommandAsync(new[] { "version" }, cancellationToken).ConfigureAwait(false);
+            var versionString = result.StandardOutput.Trim();
+            if (versionString[0] == 'v') {
+                versionString = versionString.TrimStart('v');
+            }
+            var version = SemVersion.Parse(versionString);
+            if (version == null)
+            {
+                throw new ArgumentNullException("version");
+            }
+            LocalWorkspace.CheckVersionIsValid(LocalWorkspace._minimumVersion, version);
+            this.pulumiVersion = version;
+        }
+
+        private static void CheckVersionIsValid(SemVersion minVersion, SemVersion currentVersion)
+        {
+            var err = new Exception($"Minimum version requirement failed. The minimum CLI version requirement is {minVersion}, your current CLI version is ${currentVersion}. Please update the Pulumi CLI.");
+            if (minVersion.Major != currentVersion.Major) {
+                throw err;
+            }
+            if (minVersion.Minor < currentVersion.Minor) {
+                return;
+            }
+            if (minVersion.Minor == currentVersion.Minor) {
+                if (minVersion.Patch <= currentVersion.Patch) {
+                    return;
+                }
+            }
+            throw err;
+        }
 
         /// <inheritdoc/>
         public override async Task<ProjectSettings?> GetProjectSettingsAsync(CancellationToken cancellationToken = default)
--- a/sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
+++ b/sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
@@ -369,9 +369,9 @@ namespace Pulumi.Automation
             this.pulumiVersion = version;
         }
 
-        private static void CheckVersionIsValid(SemVersion minVersion, SemVersion currentVersion)
+        public static void CheckVersionIsValid(SemVersion minVersion, SemVersion currentVersion)
         {
-            var err = new Exception($"Minimum version requirement failed. The minimum CLI version requirement is {minVersion}, your current CLI version is ${currentVersion}. Please update the Pulumi CLI.");
+            var err = new InvalidOperationException($"Minimum version requirement failed. The minimum CLI version requirement is {minVersion}, your current CLI version is ${currentVersion}. Please update the Pulumi CLI.");
             if (minVersion.Major != currentVersion.Major) {
                 throw err;
             }
--- a/sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
+++ b/sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
@@ -45,13 +45,7 @@ namespace Pulumi.Automation
 
         private SemVersion? pulumiVersion;
         /// <inheritdoc/>
-        public override SemVersion PulumiVersion 
-        { 
-            get 
-            { 
-                return pulumiVersion!; 
-            }
-        }
+        public override SemVersion PulumiVersion => pulumiVersion ?? throw new InvalidOperationException("Failed to get Pulumi version.");
 
         /// <inheritdoc/>
         public override string? SecretsProvider { get; }
@@ -349,7 +343,7 @@ namespace Pulumi.Automation
 
         private static readonly string[] SettingsExtensions = new string[] { ".yaml", ".yml", ".json" };
 
-        private async Task PopulatePulumiVersionAsync(CancellationToken cancellationToken = default)
+        private async Task PopulatePulumiVersionAsync(CancellationToken cancellationToken)
         {
             var result = await this.RunCommandAsync(new[] { "version" }, cancellationToken).ConfigureAwait(false);
             var versionString = result.StandardOutput.Trim();
@@ -367,8 +361,8 @@ namespace Pulumi.Automation
             if (minVersion.Major < currentVersion.Major) {
                 throw new InvalidOperationException($"Major version mismatch. You are using Pulumi CLI version {currentVersion} with Automation SDK v{minVersion.Major}. Please update the SDK.");
             }
-            if (minVersion.CompareTo(currentVersion) == 1) {
-                throw new InvalidOperationException($"Minimum version requirement failed. The minimum CLI version requirement is {minVersion}, your current CLI version is ${currentVersion}. Please update the Pulumi CLI.");
+            if (minVersion > currentVersion) {
+                throw new InvalidOperationException($"Minimum version requirement failed. The minimum CLI version requirement is {minVersion}, your current CLI version is {currentVersion}. Please update the Pulumi CLI.");
             }
         }
 
--- a/sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
+++ b/sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
@@ -45,7 +45,7 @@ namespace Pulumi.Automation
 
         private SemVersion? pulumiVersion;
         /// <inheritdoc/>
-        public override SemVersion PulumiVersion => pulumiVersion ?? throw new InvalidOperationException("Failed to get Pulumi version.");
+        public override string PulumiVersion => pulumiVersion is null ? throw new InvalidOperationException("Failed to get Pulumi version.") : pulumiVersion.ToString();
 
         /// <inheritdoc/>
         public override string? SecretsProvider { get; }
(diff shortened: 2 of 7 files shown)


sdk/dotnet/Pulumi.Automation/PublicAPI.Unshipped.txt
```diff
@@ -312,6 +314,7 @@ override Pulumi.Automation.LocalWorkspace.SetConfigAsync(string stackName, Syste
 override Pulumi.Automation.LocalWorkspace.SetConfigValueAsync(string stackName, string key, Pulumi.Automation.ConfigValue value, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) -> System.Threading.Tasks.Task
 override Pulumi.Automation.LocalWorkspace.WhoAmIAsync(System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) -> System.Threading.Tasks.Task<Pulumi.Automation.WhoAmIResult>
 override Pulumi.Automation.LocalWorkspace.WorkDir.get -> string
+static Pulumi.Automation.LocalWorkspace.CheckVersionIsValid(Semver.SemVersion minVersion, Semver.SemVersion currentVersion) -> void
```
komalali, line 317: Not sure how to test this functionality without making this a public method.

sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
```diff
@@ -34,13 +35,28 @@ public sealed class LocalWorkspace : Workspace
         private readonly LocalSerializer _serializer = new LocalSerializer();
         private readonly bool _ownsWorkingDir;
         private readonly Task _readyTask;
+        private static readonly SemVersion _minimumVersion = SemVersion.Parse("2.21.0");
```
orionstudt, line 38: Is this something that you guys are intending will be updated regularly?

sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
```diff
@@ -34,13 +35,28 @@ public sealed class LocalWorkspace : Workspace
...
 
+        private SemVersion? pulumiVersion;
+        /// <inheritdoc/>
+        public override SemVersion PulumiVersion 
+        { 
+            get 
+            { 
+                if (pulumiVersion == null)
+                {
+                    throw new ArgumentNullException("pulumiVersion");
```
orionstudt, line 54: I would throw a different exception here - maybe `InvalidOperationException`.. the exceptions that are prefixed with `Argument...` are kind of reserved for like parameter validation at the beginning of a method

sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
```diff
@@ -335,6 +353,39 @@ public static Task<WorkspaceStack> CreateOrSelectStackAsync(LocalProgramArgs arg
...
+        {
+            var result = await this.RunCommandAsync(new[] { "version" }, cancellationToken).ConfigureAwait(false);
+            var versionString = result.StandardOutput.Trim();
+            if (versionString[0] == 'v') {
+                versionString = versionString.TrimStart('v');
+            }
+            var version = SemVersion.Parse(versionString);
+            if (version == null)
+            {
+                throw new ArgumentNullException("version");
```
orionstudt, line 366: Throwing an `Argument...` exception here is the same issue as above, since `version` wasn't an argument to this method it seems weird

sdk/dotnet/Pulumi.Automation/PublicAPI.Unshipped.txt
orionstudt: Since we aren't intending for this method to be accessible publicly, the "correct" way  would be to mark it `internal` and then to expose internal methods to the test library. Which is what the assembly attribute `InternalsVisibleTo(assemblyName)` does, which I think we're already doing - if not it's an easy add

orionstudt (commented):
Left some remarks, otherwise looks good :)

sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
```diff
@@ -335,6 +353,39 @@ public static Task<WorkspaceStack> CreateOrSelectStackAsync(LocalProgramArgs arg
 
         private static readonly string[] SettingsExtensions = new string[] { ".yaml", ".yml", ".json" };
 
+        private async Task GetPulumiVersionAsync(CancellationToken cancellationToken = default)
+        {
+            var result = await this.RunCommandAsync(new[] { "version" }, cancellationToken).ConfigureAwait(false);
+            var versionString = result.StandardOutput.Trim();
+            if (versionString[0] == 'v') {
+                versionString = versionString.TrimStart('v');
```
mikhailshilkov, line 361: `TrimStart` is safe to call even if the string doesn't start with this character, so you don't need the `if` above.

sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
```diff
@@ -335,6 +353,39 @@ public static Task<WorkspaceStack> CreateOrSelectStackAsync(LocalProgramArgs arg
...
+                throw new ArgumentNullException("version");
+            }
+            LocalWorkspace.CheckVersionIsValid(LocalWorkspace._minimumVersion, version);
+            this.pulumiVersion = version;
+        }
+
+        public static void CheckVersionIsValid(SemVersion minVersion, SemVersion currentVersion)
+        {
+            var err = new InvalidOperationException($"Minimum version requirement failed. The minimum CLI version requirement is {minVersion}, your current CLI version is ${currentVersion}. Please update the Pulumi CLI.");
+            if (minVersion.Major != currentVersion.Major) {
```
mikhailshilkov, line 375: Have you tried doing `currentVersion < minVersion` instead of this whole function?

sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
```diff
@@ -335,6 +353,39 @@ public static Task<WorkspaceStack> CreateOrSelectStackAsync(LocalProgramArgs arg
...
+            {
+                throw new ArgumentNullException("version");
+            }
+            LocalWorkspace.CheckVersionIsValid(LocalWorkspace._minimumVersion, version);
+            this.pulumiVersion = version;
+        }
+
+        public static void CheckVersionIsValid(SemVersion minVersion, SemVersion currentVersion)
+        {
+            var err = new InvalidOperationException($"Minimum version requirement failed. The minimum CLI version requirement is {minVersion}, your current CLI version is ${currentVersion}. Please update the Pulumi CLI.");
```
mikhailshilkov, line 374: It might be a preference but I would return `bool` and throw the exception at the caller side.

sdk/dotnet/Pulumi.Automation/PublicAPI.Unshipped.txt
```diff
@@ -269,6 +269,7 @@ abstract Pulumi.Automation.Workspace.PostCommandCallbackAsync(string stackName,
 abstract Pulumi.Automation.Workspace.Program.get -> Pulumi.Automation.PulumiFn
 abstract Pulumi.Automation.Workspace.Program.set -> void
 abstract Pulumi.Automation.Workspace.PulumiHome.get -> string
+abstract Pulumi.Automation.Workspace.PulumiVersion.get -> Semver.SemVersion
```
mikhailshilkov, line 272: How important is it to return an instance of `SemVersion` vs. just a string? By doing this, we take the hard dependency on the `Semver` library which is better to avoid if possible. I'd probably go as far as creating our own thin property bag to represent a version rather than exposing the underlying library.

sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
```diff
@@ -34,13 +35,28 @@ public sealed class LocalWorkspace : Workspace
...
+        { 
+            get 
+            { 
+                if (pulumiVersion == null)
+                {
+                    throw new ArgumentNullException("pulumiVersion");
+                }
+                return pulumiVersion; 
+            }
+        }
```
mikhailshilkov, line 58: This feels overcomplicated. Would it work if we define the property `PulumiVersion` with a getter and a private setter, initialize it from `GetPulumiVersionAsync`, and get rid of the backing field? We already checked that the version is there in `GetPulumiVersionAsync`, why check it here again?

sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
```diff
@@ -335,6 +353,39 @@ public static Task<WorkspaceStack> CreateOrSelectStackAsync(LocalProgramArgs arg
 
         private static readonly string[] SettingsExtensions = new string[] { ".yaml", ".yml", ".json" };
 
+        private async Task GetPulumiVersionAsync(CancellationToken cancellationToken = default)
```
mikhailshilkov, line 356: Nit: `GetX` should return a value without side-effects. If we assign the version inside this method, I would call it `PopulatePulumiVersionAsync` or something along that line.

mikhailshilkov, line 356: Why is `= default` needed?

sdk/dotnet/Pulumi.Automation/PublicAPI.Unshipped.txt
orionstudt: Could we use the "Version" type? It's in the System namespace.

komalali: Every other SDK already has a Semver dependency (as does the `Pulumi` [SDK](https://github.com/pulumi/pulumi/blame/master/sdk/dotnet/Pulumi/Serialization/ResourcePackages.cs#L9) - although I do realize we've split them out for dotnet). So I feel comfortable taking the dependency here. 

I could, as @orionstudt suggested, use `Version` from System instead, but that begs the question why we took the `Semver` dependency in the Pulumi SDK in the first place. And I think you might be the best person to answer that question @mikhailshilkov since you're the one that [added it](https://github.com/pulumi/pulumi/commit/cab39dcfa0f97a35fa526f801b0d0f7d48de40a9#diff-a68acad3699d028399a5705febb7673a0db1f5c664bbdf73e72f5f89629c5187R35)! 🙂 

Additionally, we've exposed them as the SemVer object in every other language as well, so for consistency (and because `SemVersion` is a much more accurate type for this property) I would prefer not to use a string here.

sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
komalali: Only when there is a change to the automation SDK that is dependent on a change in the CLI (e.g. the `--page-size` flag being added to history). When we release 3.0 this will be bumped to `3.0.0` (more context: https://github.com/pulumi/pulumi/pull/6577#discussion_r599238753). Recognizing that this can be disruptive we will most likely avoid this outside of major version bumps.

sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
komalali: It's like that for every single other `Async` function in this file 🤷🏽

sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
komalali: I had it your way when I started, but if I were to do it your way I would have to make `PulumiVersion` nullable, when in reality it will never be null if the `LocalWorkspace` has made it past the constructor. That is the reason for the additional "complication". Is there a less complicated way to do this?

sdk/dotnet/Pulumi.Automation/PublicAPI.Unshipped.txt
komalali: Aha, thank you! I did not know of `internal`, only `public` and `private`.

sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
komalali: Clarified the error messages a bit. I'm going to keep this as-is and just use it as a validator since it's not quite as straightforward as a true/false now.

sdk/dotnet/Pulumi.Automation/PublicAPI.Unshipped.txt
mikhailshilkov: I think the built-in `Version` class doesn't support semver v2, so it will fail on our dev/alpha builds.

Taking `SemVer` as an internal dependency is not the same as exposing it in our public API. Right now, we can remove it if we want and nobody notices.

sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
mikhailshilkov: I see. You could also do

```cs
public override SemVersion PulumiVersion => pulumiVersion!;
```

sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
```diff
@@ -335,6 +349,29 @@ public static Task<WorkspaceStack> CreateOrSelectStackAsync(LocalProgramArgs arg
...
+                throw new InvalidOperationException("Failed to get Pulumi version.");
+            }
+            LocalWorkspace.ValidatePulumiVersion(LocalWorkspace._minimumVersion, version);
+            this.pulumiVersion = version;
+        }
+
+        internal static void ValidatePulumiVersion(SemVersion minVersion, SemVersion currentVersion)
+        {
+            if (minVersion.Major < currentVersion.Major) {
+                throw new InvalidOperationException($"Major version mismatch. You are using Pulumi CLI version {currentVersion} with Automation SDK v{minVersion.Major}. Please update the SDK.");
```
mikhailshilkov, line 368: So, if somebody has the Automation SDK 3.0, will they get this message suggesting to update their SDK? Or am I reading this wrong?

sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
```diff
@@ -335,6 +349,29 @@ public static Task<WorkspaceStack> CreateOrSelectStackAsync(LocalProgramArgs arg
...
+            }
+            if (minVersion.CompareTo(currentVersion) == 1) {
```
mikhailshilkov, line 370: Curious - why use CompareTo instead of operators (`<`)? Off the top of my head, you shouldn't check the result of `CompareTo` to be equal to 1 - rather being higher than zero. But operators are much easier to read.

sdk/dotnet/Pulumi.Automation.Tests/LocalWorkspaceTests.cs
```diff
@@ -826,5 +827,34 @@ static void AssertUpResult(UpResult upResult, string value)
...
+        [InlineData("100.0.0", true)]
+        [InlineData("1.0.0", true)]
+        [InlineData("2.22.0", true)]
+        [InlineData("2.1.0", false)]
+        [InlineData("2.21.2", true)]
+        [InlineData("2.21.1", false)]
+        [InlineData("2.21.0", false)]
+        public void ValidVersionTheory(string minVersion, bool errorExpected)
+        {
+            var testCurrentVersion = SemVersion.Parse("2.21.1");
```
mikhailshilkov, line 848: Curious - why do you have a fixed current version and varying min version? It would feel more natural to have it similar to the actual situation: fix the min version and test with many current versions.

sdk/dotnet/Pulumi.Automation.Tests/LocalWorkspaceTests.cs
```diff
@@ -826,5 +827,34 @@ static void AssertUpResult(UpResult upResult, string value)
...
+        [Fact]
+        public async Task PulumiVersionTest()
+        {
+            using var workspace = await LocalWorkspace.CreateAsync();
+            Assert.Matches("(\\d+\\.)(\\d+\\.)(\\d+)(-.*)?", workspace.PulumiVersion.ToString());
+        }
+
+        [Theory]
+        [InlineData("100.0.0", true)]
+        [InlineData("1.0.0", true)]
```
mikhailshilkov, line 840: Isn't 2.x good enough when the minimal requirement is 1.0.0?

sdk/dotnet/Pulumi.Automation/PublicAPI.Unshipped.txt
justinvp: > Taking `SemVer` as an internal dependency is not the same as exposing it in our public API.

I'm very sympathetic to this. My general inclination is to avoid exposing another library's types in public APIs, because then our API depends on some external library that we have no control over.

On the other hand, for better or worse, some of our language SDKs have already shipped with such types exposed in their public API, so adding another API that returns the same type in the same SDK makes it no worse than it already is, which was the reasoning behind my comment here: https://github.com/pulumi/pulumi/pull/6577#discussion_r597819438

- Go: https://github.com/pulumi/pulumi/blob/b82b850af0d62461d47ce8948f62ef011c8fdd7b/sdk/go/pulumi/rpc.go#L703-L705
- Python: https://github.com/pulumi/pulumi/blob/768db3e0672225340d2d5e5ea9b9467fa36d6bda/sdk/python/lib/pulumi/runtime/rpc.py#L27 https://github.com/pulumi/pulumi/blob/768db3e0672225340d2d5e5ea9b9467fa36d6bda/sdk/python/lib/pulumi/runtime/rpc.py#L722-L725

What are the scenarios for users calling this API? Having users check versions is generally an anti-pattern, because user's often do the check incorrectly (i.e. they check for specific versions and don't include an "is greater than" clause, causing them to not work correctly on newer versions). The [Windows API for getting the OS's current version](https://docs.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getversionexw) famously lies in more recent versions of Window (unless the app is manifested) to make existing programs with the problematic version checking work on newer OS versions. Instead, users are encouraged to do feature checking rather than version checking, and if you really have to check the OS version, use [less problematic helpers](https://docs.microsoft.com/en-us/windows/win32/sysinfo/version-helper-apis) instead. Similarly, I don't think we should be encouraging users to check this version. Could we just not expose this API publicly at all (i.e. keep it private and only use it internally)?

orionstudt: I think keeping it private sounds like a good way to go, it could just serve internally as a means of verifying that the dependent CLI is an adequate version

sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
orionstudt: We're doing it on the public APIs so that consumers can choose to provide or not provide a cancellation token - not everybody does and generally it is commonplace to have overloads for both. We may have issues in the future if we need to make changes to the parameters on those APIs, but cancellation token being the final optional parameter (in ordering) makes it relatively safe.

In this scenario with a private method that will only be called further down the method chain we can make the assumption that we will always have a cancellation token - because the method further up the chain that is calling this method already took in the optional. So here it is not necessary.

sdk/dotnet/Pulumi.Automation/PublicAPI.Unshipped.txt
komalali: This was something that was requested by a customer. @justinvp to make sure I'm on the same page, you're now suggesting that I remove the public availability of PulumiVersion from all languages and don't do https://github.com/pulumi/pulumi/issues/6348?

sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
komalali: Got it! FWIW it looks like [CreateStackAsync](https://github.com/pulumi/pulumi/blob/master/sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs#L517) is also a public method that does not have ` = default`

sdk/dotnet/Pulumi.Automation.Tests/LocalWorkspaceTests.cs
komalali: No. Because a major version bump can include a breaking change to the CLI we will require major version parity for Automation SDK.

sdk/dotnet/Pulumi.Automation.Tests/LocalWorkspaceTests.cs
komalali: I'm going to leave this as is because I've tested it thoroughly and no one else had a problem with it in any of the other languages :)

sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
orionstudt: On first glance that looks like an oversight, but if you look at the [abstract Workspace](https://github.com/pulumi/pulumi/blob/b187ce18c24b2229531f1e9100334e672976f451/sdk/dotnet/Pulumi.Automation/Workspace.cs#L185) that `LocalWorkspace` extends then you'll see the overloads that allow you to exclude a cancellation token.

orionstudt: I believe that for `CreateStackAsync` and `SelectStackAsync` I needed to use concrete parameters and provide physical overloads (instead of optional parameters) because if the parameters were optional it was causing method signature collisions with the [static methods](https://github.com/pulumi/pulumi/blob/b187ce18c24b2229531f1e9100334e672976f451/sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs#L84).

sdk/dotnet/Pulumi.Automation/PublicAPI.Unshipped.txt
komalali: Just so everyone's on the same page, in the Go implementation I had initially created "our own thin property bag to represent a version" specifically to avoid taking the dependency. @justinvp asked me to change it to use the SemVer library. So I used that pattern across SDKs.
So I'm a little perplexed at the conversation here, especially since @justinvp has commented on/approved the other PRs.

justinvp: @komal, sorry if my original feedback [here](https://github.com/pulumi/pulumi/pull/6577#discussion_r597819438) led you down a potentially wrong path. I suggested using the existing semver library to do the parsing rather than manual parsing, and for Go, since we already exposed the semver library type elsewhere in public APIs, exposing that rather than creating a custom struct. It wasn't until I saw @mikhailshilkov's questions about returning a string in this PR, that it triggered in my brain that we should be deliberate about whether we want to expose anything (whether it's a custom struct or some external lib's version type) that encourages folks to compare versions, given the problems that it can cause.

> This was something that was requested by a customer.

Do we know what the customer wants to do with the version? If it's just for logging purposes, a string is likely sufficient (and is the same thing the CLI returns), and we wouldn't be encouraging anyone to do comparisons using the API we expose (if a user really wanted to, they could still parse and compare at their own peril).

If the customer wants it to do version checking, are they doing it to see if a specific feature is supported? If so, can we consider alternatives, like exposing feature checking APIs (e.g. `public bool SupportsFoo { get; }`)?

If the custom really is requesting something they can do version checking against, can we consider APIs that encourage more correct behavior, something like (I know it's a mouthful, open to better naming):

```c#
public bool IsPulumiVersionEqualToOrGreaterThan(int major, int? minor = null, int? patch = null, string? prerelease = null);
```

> @justinvp to make sure I'm on the same page, you're now suggesting that I remove the public availability of PulumiVersion from all languages and don't do #6348?

In general, we should be consistent across languages as much as possible (as long as doing so is idiomatic for each language). If we decide not to make it public, it should be private in all languages. If we decide to make it a string, make it a string in all languages.

leezen: Given we've already shipped this in the other languages via using an external library and we've already exposed the API publicly, it feels like we should be consistent here and just use the library to expose the version.

From a "one-way door" perspective, we've already walked through the door to expose the version. And we've already chosen the way to expose it (via a Semver type in the other languages). Let's not be halfway in the doorway.

mikhailshilkov: Sure, but why not use our own Semver type? It will take ~10 lines of code if we are worried about extra work.

komalali: https://github.com/pulumi/pulumi/pull/6590#discussion_r599726196

sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
mikhailshilkov: Thank you!

sdk/dotnet/Pulumi.Automation/PublicAPI.Unshipped.txt
mikhailshilkov: > since we already exposed the semver library type elsewhere in public APIs, exposing that rather than creating a custom struct.

This does not apply to .NET

komalali: > It will take ~10 lines of code if we are worried about extra work.

This is a pretty patronizing comment, TBH.

mikhailshilkov: Oh, I apologize for inappropriate wording

komalali: To summarize the feedback, it appears that my options are:
1. Implement a separate `Version` object for C# - because taking the hard dependency on Semver is "better to avoid if possible"
2. Not expose the `PulumiVersion` publicly, and therefore not do #6348 (and change this for the other 3 languages)
3. Expose `PulumiVersion` as a string (and change this for the other 3 languages)

After chatting with the customer in question here and their use-case, it appears 3 is sufficient and what they really want is version validation so we could maybe even get away with doing 2 (assuming that no other customers share the desire to have this accessible). As a happy medium, I will move forward with option 3.

justinvp: Option 3 sounds good to me. Thank you, @komalali!

komalali: Option 3 is done in this PR for dotnet and in https://github.com/pulumi/pulumi/pull/6604 for the other languages.

sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
```diff
@@ -34,13 +35,18 @@ public sealed class LocalWorkspace : Workspace
...
 
         /// <inheritdoc/>
         public override string WorkDir { get; }
 
         /// <inheritdoc/>
         public override string? PulumiHome { get; }
 
+        private SemVersion? pulumiVersion;
+        /// <inheritdoc/>
+        public override string PulumiVersion => pulumiVersion is null ? throw new InvalidOperationException("Failed to get Pulumi version.") : pulumiVersion.ToString();
```
mikhailshilkov, line 48:
```suggestion
        public override string PulumiVersion => pulumiVersion?.ToString() ?? throw new InvalidOperationException("Failed to get Pulumi version.");
```

mikhailshilkov (approved):
LGTM

Issue #6471, opened 2021-03-04
Automatic virtual environment creation uses wrong path and fails


clstokes:
If I `pulumi new python -y`, virtual environment creation works fine. 

If I `pulumi up` and do not already have a virtual environment, pulumi attempts to create a virtual environment uses the wrong path and fails. In this scenario, `pulumi` appears to create a `venv` directory using the full absolute path but from the relative path - e.g `/Users/clstokes/cc/scratch/p-python-venv/p-python1/Users/clstokes/cc/scratch/p-python-venv/p-python1/venv` when running from `/Users/clstokes/cc/scratch/p-python-venv/p-python1/`.

## Expected behavior

`pulumi up` should create a `venv` within my current project directory - e.g. `/Users/clstokes/cc/scratch/p-python-venv/p-python1/venv`.

## Current behavior

`pulumi up` creates a `venv` using the full absolute path, but within my current directory - e.g. `/Users/clstokes/cc/scratch/p-python-venv/p-python1/Users/clstokes/cc/scratch/p-python-venv/p-python1/venv` when running from `/Users/clstokes/cc/scratch/p-python-venv/p-python1/`

### Error output

```
p-python1 % pulumi up
Previewing update (dev)

View Live: https://app.pulumi.com/clstokes/p-python1/dev/previews/e952a0b6-b101-41b9-a82b-a9a1dcca0ce2

     Type                 Name           Plan
     pulumi:pulumi:Stack  p-python1-dev

error: failed to discover plugin requirements: The 'virtualenv' option in Pulumi.yaml is set to "venv", but "/Users/clstokes/cc/scratch/p-python-venv/p-python1/venv" doesn't exist; run the following commands to create the virtual environment and install dependencies into it:

    1. python3 -m venv /Users/clstokes/cc/scratch/p-python-venv/p-python1/venv
    2. /Users/clstokes/cc/scratch/p-python-venv/p-python1/venv/bin/python -m pip install --upgrade pip setuptools wheel
    3. /Users/clstokes/cc/scratch/p-python-venv/p-python1/venv/bin/python -m pip install -r requirements.txt


For more information see: https://www.pulumi.com/docs/intro/languages/python/#virtual-environments
```

## Steps to reproduce

1. mkdir dir1 && cd dir1
2. `pulumi new python -y`
3. `rm -rf venv`
4. `pulumi up`
    - *observe error*

## Context (Environment)

```
p-python1 % python3 --version
Python 3.9.1
[...]

PR #6599, opened 2021-03-22
Avoid double-quailfying venv folder path

t0yv0:
Fixes bug in pulumi-managed venv https://github.com/pulumi/pulumi/issues/6471 that is double-qualified paths such as "/Users/me/my-project/Users/me/my-project/venv"

- Do we want a Changelog entry for this?

- Do we have or want a test?

I discovered this on building a Python example in pulumi/examples so perhaps CI testing of examples could have some coverage, but we might want a more direct test that pulumi cli can successfully init a venv.


komalali (approved):
LGTM ⭐

Yes please do add this to the CHANGELOG. We most likely don't have a test that brings its own venv but would be ideal to add one for each of the absolute and the relative path case.

justinvp (approved):
One nit, but LGTM otherwise.

A note about testing... Our integration test package has some [automatic handling of venv](https://github.com/pulumi/pulumi/blob/768db3e0672225340d2d5e5ea9b9467fa36d6bda/pkg/testing/integration/program.go#L1755-L1768), so you'll want to avoid using `integration.ProgramTest` like most of the existing tests in https://github.com/pulumi/pulumi/blob/768db3e0672225340d2d5e5ea9b9467fa36d6bda/tests/integration/integration_python_test.go.

I'd probably just add a couple tests to https://github.com/pulumi/pulumi/blob/768db3e0672225340d2d5e5ea9b9467fa36d6bda/tests/integration/integration_python_test.go that look more like the tests in https://github.com/pulumi/pulumi/blob/768db3e0672225340d2d5e5ea9b9467fa36d6bda/tests/stack_test.go (for example), using `e := ptesting.NewEnvironment(t)` to create a new test environment and then having tests with relative and absolute paths for virtualenv in `Pulumi.yaml`, and then running a `pulumi up` command, and ensuring the venv directory is created as expected.

t0yv0:
I have added a test and confirmed it goes red->green with the fix. It actually does not test local/global path it tests more of a high-level user flow. Check out a project, run Pulumi preview (same thing happens in pulumi up). Before the fix the test fails right at "pulumi preview" step as that errors out. But I also put an assert that venv got created.

The test is a bit slow (16s) but it feels like a good scenario to have covered.

Shall I merge as is or change how we test here? Thanks.

tests/integration/integration_python_test.go
```diff
@@ -400,3 +403,30 @@ func TestGetResourcePython(t *testing.T) {
 		AllowEmptyPreviewChanges: true,
 	})
 }
+
+// Regresses https://github.com/pulumi/pulumi/issues/6471
+func TestAutomaticVenvCreation(t *testing.T) {
+	// Do not use itnegration.ProgramTest to avoid automatic venv
+	// handling by test harness; we actually are testing venv
+	// handilng by the pulumi CLI itself.
```
justinvp, line 411: Nit

```suggestion
	// Do not use integration.ProgramTest to avoid automatic venv
	// handling by test harness; we actually are testing venv
	// handling by the pulumi CLI itself.
```

justinvp (approved):
The test LGTM. Should we add a case for when `Pulumi.yaml` has an absolute path specified for `virtualenv`? One way to do it for that case would be to emit (or modify) the `Pulumi.yaml` in the test environment to insert an absolute path.

t0yv0:
Oh I did not realize that's an option one can set in Pulumi.yaml. Yes that sounds good to test.

t0yv0:
Absolute path test fails on windows. I'll need to dig in.

t0yv0:
Seriously this is confusing

```
func TestAbsPathAnomaly(t *testing.T) {
	p := "C:\\Users\\anton\\AppData\\Local\\Temp\\2\\test-env598977482\\absvenv"
	t.Errorf("Path=%s filepath.IsAbs=%v path.IsAbs=%v", p, filepath.IsAbs(p), path.IsAbs(p))
}

PS C:\Users\anton\pulumi\tests\integration> go test -tags python -run AbsPathAnomaly
--- FAIL: TestAbsPathAnomaly (0.00s)
    integration_python_test.go:412: Path=C:\Users\anton\AppData\Local\Temp\2\test-env598977482\absvenv filepath.IsAbs=tr
ue path.IsAbs=false
FAIL
exit status 1
```

t0yv0:
https://stackoverflow.com/questions/39181790/whats-the-difference-between-path-path-filepath-packages-in-go

Sounds like that is the root cause here. Trying.

PR #6601, opened 2021-03-23
Demo AllowNull in Output.Create - do not merge

t0yv0:


mikhailshilkov:
Thank you for this example. So, what's your conclusion?

MisinformedDNA:
Here's a [good example/explanation](https://endjin.com/blog/2020/05/dotnet-csharp-8-nullable-references-transcending-the-type-system-with-allownull):

```csharp
[AllowNull]
public string DisplayName
{
   get => this.displayName ?? this.Id;
   set => this.displayName = value;
}
private string? displayName;

public string Id { get; set; }
```

> Here we have a DisplayName property which will fall back to returning the Id property if no display name has been set, so it can be counted on never to return a null. That means its type, string correctly describes the get behaviour. But its set accepts null to unset the display name, reverting to the fallback behaviour. The [AllowNull] attribute tells the compiler that this asymmetric behaviour is our intention.

t0yv0:
Ah sorry @mikhailshilkov  it was not clear from cross-linking. I intended this PR here to supplement discussion at: https://github.com/pulumi/pulumi/pull/6600 - specifically you asked me to investigate if by any chance we meant to use AllowNull where the SDK code was using MaybeNull.

TLDR is that I do not think we should be trying to use AllowNull attribute in the SDK Output/Input.Create, as there is no reasonable way that I see for these methods to do something reasonable with the null value in such a way that the rest of the program does not have to deal with the null value. @MisinformedDNA has a nice example of a correct use of AllowNull above - that example actually would never generate a NullPointer exception in consuming code. I do no think we can do the same with Output.Create.

Conclusion: I'd recommend we move forward with merged https://github.com/pulumi/pulumi/pull/6600 that simply removes MaybeNull and leave things like this. Please feel free to reopen the issue if we need to do more work here. Thanks!

PR #6613, opened 2021-03-24
Ensure we include alpha releases as part of the min version checking in automation api

stack72:


komalali:
This isn't really relevant for 2.21.0 since it's an older version and we are no longer doing `2.21.0-alpha` dev releases.

We'll need to be conscious of this going forward (3.0.0+) but I think we can leave `2.21.0` as-is for now.

Issue #6608, opened 2021-03-24
Permissions error when running `pulumi new <path_to_repo_folder>


lifeofzero:
I am trying to maintain a configuration repo that will have specific project settings for my three stacks (dev, stage, prod).  
From this configuration repo I would like to use it as a base to seed new repos/services.  I get two failures with 'permission denied'

- We need to create a new repo that maps to a service
- developer creates a new folder locally for their service
- inside this folder they create a folder 'infrastructure/dev'
- from this dir they run pulumi new <path_to_local_git_pulumi_config_repo>
- this runs but errors out with  
## Expected behavior

I would think that I could run` pulumi new <path_to_github_config_repo> `and it would populate the folder with the pulumi base yaml and python from the config repo that I have.  Then I can modify the __main__.py with my new code and deploy into the given stack (dev, stage, or prod) using `pulumi up`

## Current behavior

I get the following error during `pulumi new`
reated stack 'dev'
Creating virtual environment...
Finished creating virtual environment
Updating pip, setuptools, and wheel in virtual environment...
error: updating pip, setuptools, and wheel via '/Users/tyler/projects/propel-api/infrastructure/dev/venv/bin/python -m pip install --upgrade pip setuptools wheel': fork/exec /Users/tyler/projects/propel-api/infrastructure/dev/venv/bin/python: permission denied

From this directory I can run `pip install -r requirements.txt` and it does work, but if I then run `pulumi up` I get the following error.
error: failed to discover plugin requirements: fork/exec /Users/tyler/projects/propel-api/infrastructure/dev/venv/bin/python: permission denied

## Steps to reproduce

1. create a new repo on git for the new service
2. git clone locally
3. cd to repo folder
4. git branch
5. pulumi new <path_to_local_git_pulumi_config_repo>
6. Get the above permissions denied error
7. pip install -r requirements.txt
8. pulumi up
9. Get the above failed to discover plugin requirements

## Context (Environment)

I am trying to create a developer workflow that has a baseline set of pulumi configurations for each of the stacks that we will operate from.  This configuration will be managed in a git repo and will only have dev, stage, prod configurations for pulumi and right now GCP

lifeofzero:
[...]

PR #6623, opened 2021-03-25
Fixes 6608: unusable venv after template copy, python not executable

t0yv0:
Toward fixing #6608.

To reproduce 6608:

```
#!/bin/sh

mkdir template
pushd template
pulumi new --yes aws-python --name "template-6608-1"
popd

mkdir user-project
pushd user-project
pulumi new --yes ../template --name "user-project-6608-1"
popd

```

What happens is that the second `pulumi new` copies `venv` but forgets to mark files like `python` executable. It also ignores any symlink structure, if any... Symlinks become regular files.

*Option 1*: as coded in the PR, propagate more permission bits from source to dest, so executable files remain executable after template expansion, but everything is at least 0600.

*Option 2*: drop 0600, use exact permissions from source file. This did NOT fix the bug when using os.Stat - because of symlink handling.. begs the question of what should template expansion do with symlinks anyway? Copying symlinks as symlinks its not possible when the contents are being modified by the logic in here.

*Option 3*: skip virtual environment from the template copying process. Where could we do this cleanly? If we can do it cleanly it might be the best as then we will let Python manage virtual environment folder including deciding which one is a symlink and which one is not; we do not mess with that. It seems that upon copying the template, Pulumi up does re-initialization of the virtual env (if configured to manage it in the yaml). So that'd be ideal perhaps.


justinvp:
Thanks, @t0yv0! I want to take a look over this before merging. Hopefully I’ll be able to get to it tomorrow, but if not, Monday.

t0yv0:
Oh right - the user has found a workaround so there is no rush on this one.

justinvp:
Our stance on this has generally been that `venv` should not be committed in the template repo. If your template has specific dependencies, include them in `requirements.txt` and Pulumi will automatically create the virtual environment and install the `requirements.txt` dependencies into it.

> _Option 1_: as coded in the PR, propagate more permission bits from source to dest, so executable files remain executable after template expansion, but everything is at least 0600.

That said, this still seems like a reasonable change to make

t0yv0:
Can still repro on 3.x. Accepting the fix I think is better than leaving as-is, although there's better way to fix it.

t0yv0:
> Our stance on this has generally been that venv should not be committed in the template repo. If your template has specific dependencies, include them in requirements.txt and Pulumi will automatically create the virtual environment and install the requirements.txt dependencies into it.

Hm that sounds good but should we then teach Pulumi CLI to exclude venv from consideration when it scans template folders and venv is accidentally present? I think this happens when an existing project is used as a template before the user commits it cleanly to a template repo.

lblackstone (approved):
This seems like a reasonable improvement to me.

PR #6625, opened 2021-03-26
Make .old and .new nullable on StateEventMetadata

t0yv0:
Go model seems to have these optional:

```
// StepEventMetadata describes a "step" within the Pulumi engine, which is any concrete action
// to migrate a set of cloud resources from one state to another.
type StepEventMetadata struct {
	// Op is the operation being performed.
	Op   OpType `json:"op"`
	URN  string `json:"urn"`
	Type string `json:"type"`

	// Old is the state of the resource before performing the step.
	Old *StepEventStateMetadata `json:"old"`
	// New is the state of the resource after performing the step.
	New *StepEventStateMetadata `json:"new"`
	// Omitted from the type sent to the Pulumi Service is "Res", which may be either Old or New.
```

The emerging C# model treats them optional also. 

Should they be optional/nullable in TypeScript?

### DIFF
--- a/sdk/nodejs/x/automation/events.ts
+++ b/sdk/nodejs/x/automation/events.ts
@@ -109,9 +109,9 @@ export interface StepEventMetadata {
     type: string;
 
     // Old is the state of the resource before performing the step.
-    old: StepEventStateMetadata;
+    old?: StepEventStateMetadata;
     // New is the state of the resource after performing the step.
-    new: StepEventStateMetadata;
+    new?: StepEventStateMetadata;
 
     // Keys causing a replacement (only applicable for "create" and "replace" Ops).
     keys?: string[];


komalali (approved):
Yes, thanks for catching this! `old` will be `null` on `create` operations, and `new` will be `null` on `delete` operations.

t0yv0:
I guess I can add a little changelog entry here.

komalali:
@t0yv0 I think it's okay to just add the `no-changelog-needed` label here.

t0yv0:
Ah good to know. I thought it's technically breaking, but perhaps it's an API that has no stability guarantees yet.

t0yv0:
I'll merge later when tests pass then. Thank you.

PR #2599, opened 2019-03-27
Be resilient to 3rd party libraries placing an __awaiter function on the global object.

CyrusNajmabadi:
Fixes https://github.com/pulumi/pulumi/issues/2530

Issue #2530, opened 2019-03-07
ts + node + dynamic provider: __awaiter is not always captured


johlrich:
There are times when typescript's `__awaiter` is not being serialized with the provider, causing errors when applying the pulumi app.

I have created a minimal repro that demonstrates the same provider being serialized both [with](https://github.com/johlrich/pulumi-repros/blob/5499f3247f1bdf1c9e938e41fa22bd3e4ecf43a3/awaiter-exists/captured.js) and [without](https://github.com/johlrich/pulumi-repros/blob/5499f3247f1bdf1c9e938e41fa22bd3e4ecf43a3/awaiter-exists/missing.js) the `__awaiter` included. https://github.com/johlrich/pulumi-repros/tree/5499f3247f1bdf1c9e938e41fa22bd3e4ecf43a3/awaiter-exists 

The only difference between the runs is referencing something from `@azure/storage-blob` [index.ts](https://github.com/johlrich/pulumi-repros/blob/5499f3247f1bdf1c9e938e41fa22bd3e4ecf43a3/awaiter-exists/index.ts) vs [index-ok.ts](https://github.com/johlrich/pulumi-repros/blob/5499f3247f1bdf1c9e938e41fa22bd3e4ecf43a3/awaiter-exists/index-ok.ts). What's interesting is it doesn't even need to be used in the provider being serialized to cause the difference.

My wild guess for the difference is the azure js sdks use of tslib is throwing off serialization.

As a workaround, I tried to make my library also use tslib but then the provider serialized with a `this`, which fails.

CyrusNajmabadi:
thanks for the repro!  i can take a look tomorrow :)

CyrusNajmabadi:
Thanks for the report @johlrich .  This has been fixed and will be in an upcoming `@pulumi/pulumi` release!

PR #6630, opened 2021-03-27
Be resilient to 3rd party libraries placing a __rest function on the global object.

mweathers:
This fixes a similar issue with `__rest` as was documented for `__awaiter` in #2530 and fixed in #2599


jeffrichie:
@viveklak @tusharshahrs hey guys - can we get this guy merged?

pgavlin (approved):

PR #6639, opened 2021-03-29
[automation/python] - Fix passing of extra environment variables

komalali:
The previous implementation did not correctly pass forward the environment variables and `env` ended up as `None`. 

This isn't entirely straightforward to test in isolation but it is used by https://github.com/pulumi/pulumi/pull/6527 and I was able to test that it works as expected because the `PULUMI_DEBUG_COMMANDS` env var is successfully passed through and the tests don't fail with `unknown flag --event-log`.

### DIFF
--- a/sdk/python/lib/pulumi/x/automation/_cmd.py
+++ b/sdk/python/lib/pulumi/x/automation/_cmd.py
@@ -46,7 +46,7 @@ def _run_pulumi_cmd(args: List[str],
     # All commands should be run in non-interactive mode.
     # This causes commands to fail rather than prompting for input (and thus hanging indefinitely).
     args.append("--non-interactive")
-    env = os.environ.copy().update(additional_env)
+    env = {**os.environ, **additional_env}
     cmd = ["pulumi"]
     cmd.extend(args)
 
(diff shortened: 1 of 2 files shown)


sdk/python/lib/pulumi/x/automation/_cmd.py
```diff
@@ -46,7 +46,7 @@ def _run_pulumi_cmd(args: List[str],
     # All commands should be run in non-interactive mode.
     # This causes commands to fail rather than prompting for input (and thus hanging indefinitely).
     args.append("--non-interactive")
-    env = os.environ.copy().update(additional_env)
+    env = {**os.environ.copy(), **additional_env}
```
justinvp, line 49: Is the `copy()` necessary now? Wondering if it could be:

```suggestion
    env = {**os.environ, **additional_env}
```

justinvp (approved):
LGTM

sdk/python/lib/pulumi/x/automation/_cmd.py
komalali: Good call!

EvanBoyle (approved):

PR #6640, opened 2021-03-29
Use CliWrap for subprocess management

t0yv0:
Use CliWrap to avoid custom async code around System.Diagnostics.Process.

On top of https://github.com/pulumi/pulumi/pull/6572

There is perhaps something obvious I am missing but my latest pulumi CLI does not support the --log-xx parameter we are giving it. Is that still being built out?


t0yv0:
@vipentti @orionstudt @mikhailshilkov - as we discussed separate PR for the subprocess handling. What do you think?

t0yv0:
Hm, GitHub does not let me base the PR against the source. I probably need to open a fork. Let me tinker.

t0yv0:
This fork https://github.com/vipentti/pulumi/pull/1 lets GitHub show just the diff I added.

sdk/dotnet/Pulumi.Automation/Commands/LocalPulumiCmd.cs
```diff
@@ -34,103 +99,86 @@ internal class LocalPulumiCmd : IPulumiCmd
...
-                }
-                catch
-                {
-                    // in case the process hasn't started yet
-                    // or has already terminated
-                }
-            });
+        private class EventLogFile : IDisposable
+        {
+            private string _eventLogFile;
```
vipentti, line 135: Since we don't modify the file path, we can use get-only property e.g

```csharp
public string FilePath { get; }
```

t0yv0: Most .nET i've written was F# so please keep the style comments coming, very helpful.

sdk/dotnet/Pulumi.Automation.Tests/LocalPulumiCmdTests.cs
```diff
@@ -0,0 +1,30 @@
...
+            var localCmd = new LocalPulumiCmd();
+            IDictionary<string,string> extraEnv = new Dictionary<string,string>();
+            IEnumerable<string> args = new string[]{ "version" };
+            var result = await localCmd.RunAsync(args, ".", extraEnv
+                // Uncommenting this currently fails the test on v2.23.1:
+                //
+                // Which versions of Pulumi CLI support --event-log param?
+                //
+                // , onEngineEvent: _ev => {}
+            );
```
komalali, line 24: `--event-log` is supported only when using the `PULUMI_DEBUG_COMMANDS` environment variable.

orionstudt:
I assume you only really touched `LocalPulumiCmd` and the rest is still @vipentti ? Changes in `LocalPulumiCmd` look good to me 👍

sdk/dotnet/Pulumi.Automation.Tests/LocalPulumiCmdTests.cs
t0yv0: Thank you for the pointer! That means I have a bug in env var setting, or something like this. I will investigate further.

komalali: Although it seems like it is being [passed in](https://github.com/pulumi/pulumi/pull/6640/files#diff-ca2533e085f4365e71a44e2dd166bb28a79ea3610b41aff54b122b1764c5b12eR108-R111) so I'm not sure what the error is here.

vipentti:
The changes look good to me! `CliWrap`  is really nice for this kind of stuff

sdk/dotnet/Pulumi.Automation.Tests/LocalPulumiCmdTests.cs
orionstudt: well he's doing a `version` command in that test and I'd wager that the `--event-log` argument is only supported on the few update commands, right? It's at the very least only relevant on those commands. Unless the other commands are streaming the `StandardOutput` events, I guess

komalali: Oh yes you're totally right, `--event-log` only works on `up/preview/destroy/refresh`

t0yv0: Ah that makes sense! Let me see if we where a test of one of those would belong.

orionstudt: maybe in `LocalWorkspaceTests` on the APIs where `UpdateOptions` accepts an `Action<EngineEvent>` property?

orionstudt: we could test directly on `LocalPulumiCmd` as well but there is a lot of workspace / project setup that precedes being able to test the engine event listener, setup that `LocalWorkspace` takes care of

t0yv0: Let me start grokking that file. That's probably the right place..

t0yv0: Right. If this is indirectly covered it might  not separate component tests.

orionstudt: There might be value in testing the log file watcher independently, though

t0yv0: Leaving "pulumi version" test and added a note which other test covers onEngineEvent, looks like there is decent smoke-test coverage at least already, I was happy to find that.

PR #6641, opened 2021-03-29
[codegen/go] Fix Go resource registrations

justinvp:
We've been emitting calls to `New<Resource>` for resource registrations in Go, passing `nil` for args. However, some of those `New<Resource>` functions actually check for `nil` args and return an error if the resource has required arguments.

At first, I was looking for a way to check inside `New<Resource>` if the `URN` option was specified and in that case not error on `nil` args (like we do in other languages), but we don't provide a way to access the resource option values outside the Go SDK), so I don't think there is a way to do it this way for Go.

So instead, this change updates the registration code to call `ctx.RegisterResource` directly instead of `New<Resource>`, where we can pass a `nil` args.

Fixes https://github.com/pulumi/pulumi-eks/issues/537

### DIFF
--- a/pkg/codegen/go/gen.go
+++ b/pkg/codegen/go/gen.go
@@ -1724,11 +1724,12 @@ func (pkg *pkgContext) genResourceModule(w io.Writer) {
 
 			registrations.Add(tokenToModule(r.Token))
 			fmt.Fprintf(w, "\tcase %q:\n", r.Token)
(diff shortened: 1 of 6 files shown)


lblackstone (approved):

justinvp:
Thanks for the help, @lblackstone!

bothra90:
Thanks for the fix, @justinvp!

Issue #6643, opened 2021-03-29
[nodejs] registerResource errors failing to propagate


EvanBoyle:
As a part of https://github.com/pulumi/pulumi/pull/5347/files#diff-56e5453d5130bf4ada128fb338b9fb7c7abd9df00abb7090844b88c03ed9c2dc we made some modifications to error propagation on registerResource, readResource, etc.

This introduced a bug where some operations were run in a try catch, and the resulting errors were not appropriately propagated up the promise chain via resolvers for resource IDs and URNs. 

This can lead to masked errors. For instance, if registerResource fails without rejecting the URN promise, then we can see errors along the following lines in registerResourceOutputs:

```
Error: 2 UNKNOWN: missing required URN
```

This is due to the URN resolving as undefined, instead of rejecting with an error. RegisterResourceOutputs, and other dependent operations then try to proceed operation in this error state instead of halting. 

The fix is to appropriately propagate errors via the promise chain.

EvanBoyle:
Some related error reports where relevant info is likely getting masked: 
https://github.com/pulumi/pulumi/issues/5832
https://github.com/pulumi/pulumi/issues/5093

EvanBoyle:
Also masked: https://github.com/pulumi/pulumi/issues/6645

EvanBoyle:
closed via https://github.com/pulumi/pulumi/pull/6644

EvanBoyle:
closed via https://github.com/pulumi/pulumi/pull/6644

PR #6644, opened 2021-03-29
fix nodejs resource functions (ie registerResource) to properly propagate errors

EvanBoyle:
Fixes https://github.com/pulumi/pulumi/issues/6643

More detail in the issues above, but TLDR is that we were swallowing errors in registerResource. This change fixes `prepareResource` to set up reject handlers on ID and URN promises to appropriately propagate the errors.

### DIFF
--- a/sdk/nodejs/tests/runtime/props.spec.ts
+++ b/sdk/nodejs/tests/runtime/props.spec.ts
@@ -67,6 +67,7 @@ class TestMocks implements runtime.Mocks {
                     state: {},
                 };
             case "error":
+                throw new Error("this is an intentional error");
             default:
                 throw new Error(`unknown resource type ${type}`);
         }
@@ -366,7 +367,7 @@ describe("runtime", () => {
                 const customURN = await errResource.urn.promise();
                 const customID = await errResource.id.promise();
             }, (err: Error) => {
-                const containsMessage = err.stack!.indexOf("unknown resource type error") >= 0;
+                const containsMessage = err.stack!.indexOf("this is an intentional error") >= 0;
                 const containsRegisterResource = err.stack!.indexOf("registerResource") >= 0;
                 return containsMessage && containsRegisterResource;
             });
--- a/sdk/nodejs/tests/runtime/props.spec.ts
+++ b/sdk/nodejs/tests/runtime/props.spec.ts
@@ -32,6 +32,12 @@ class TestCustomResource extends CustomResource {
     }
 }
 
+class TestErrorResource extends CustomResource {
+    constructor(name: string) {
+        super("error", name, {});
+    }
+}
+
 class TestResourceModule implements runtime.ResourceModule {
     construct(name: string, type: string, urn: string): Resource {
         switch (type) {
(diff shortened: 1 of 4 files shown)


EvanBoyle:
The original repro that lead me to discover this. Unforunately it takes ~5 mins to run and only fails ~80% of the time: https://github.com/pulumi/pulumi/issues/6645

pgavlin (commented):
Can we set up a test for this using mocks? Throwing from within `newResource` should work, I think.

EvanBoyle:
@pgavlin wouldn't that just be testing the mocked [registerResource](https://github.com/pulumi/pulumi/blob/c1ce3da3e681cd603be270f219f14b3dba2d8a34/sdk/nodejs/runtime/mocks.ts#L115) implementation? If so that wouldn't really validate this fix. 

I had a hard time coming up with a way to test this. The only repro I currently have is the one linked here #6645 and it takes 5 minutes to fail, and the failure is non-deterministic as it seems related to resource exhaustion at some layer.

EvanBoyle:
(the mocked registerResource impl does't call prepareResource which is the layer at which the fix was made)

justinvp:
> (the mocked registerResource impl does't call prepareResource which is the layer at which the fix was made)

When using a mock, `prepareResource` should still be called...

The `Resource` constructor calls `registerResource` in `runtime/resource.ts`:

https://github.com/pulumi/pulumi/blob/ff2cf701a70fb1319fb7beabcf88ad34d55b6217/sdk/nodejs/resource.ts#L342

`registerResource` in `runtime/resource.ts` calls `prepareResource`:

https://github.com/pulumi/pulumi/blob/ff2cf701a70fb1319fb7beabcf88ad34d55b6217/sdk/nodejs/runtime/resource.ts#L284

And then call's the monitor's `registerResource` (whether `monitor` is an actual gRPC monitor or a mock monitor):

https://github.com/pulumi/pulumi/blob/ff2cf701a70fb1319fb7beabcf88ad34d55b6217/sdk/nodejs/runtime/resource.ts#L340

EvanBoyle:
@justinvp ah, thanks for the pointer. I did forgot that this was mocking the monitor and not the exported calls in runtime. I'll work on getting these tests added.

EvanBoyle:
@pgavlin @justinvp added the test. please take another look.

sdk/nodejs/tests/runtime/props.spec.ts
```diff
@@ -60,6 +66,7 @@ class TestMocks implements runtime.Mocks {
                     id: runtime.isDryRun() ? undefined : "test-id",
                     state: {},
                 };
+            case "error":
```
justinvp, line 69: Nit: Can we throw a specific error for this case, rather than relying on the fall through for "unknown resource"?

justinvp (commented):
Do we need to do something similar for the resolvers for all the other outputs in `transferProperties`?

EvanBoyle:
> Do we need to do something similar for the resolvers for all the other outputs in transferProperties?

This is a very good point. These are actually the most common outstanding promises that we see leaked, and that would make sense as we aren't properly rejecting them...

Fix incoming.

komalali (approved):

komalali:
Huh, looks like the transferProperties changes are causing unhandled rejections in the tests now

pgavlin (approved):

Issue #6645, opened 2021-03-29
dynamic provider resource exhaustion


EvanBoyle:
Gist with repro: https://gist.github.com/EvanBoyle/ac4d50a5cf9c66ce5302e0c325a18e8b

It appears that creating lots of dynamic resources creates resource exhaustion at some layer (investigation ongoing). 

This was getting masked by https://github.com/pulumi/pulumi/issues/6643, but with a fix for that in, I can now see the following output:

```
error: Running program '/Users/evanboyle/go/src/dynprov' failed with an unhandled exception:
Error: failed to register new resource foo [lotsofdynamics]: 8 RESOURCE_EXHAUSTED: Bandwidth exhausted
    at Object.registerResource (/opt/pulumi/node_modules/@pulumi/pulumi/runtime/resource.js:218:27)
    at new Resource (/opt/pulumi/node_modules/@pulumi/pulumi/resource.js:215:24)
    at new ComponentResource (/opt/pulumi/node_modules/@pulumi/pulumi/resource.js:386:9)
    at new Component (/Users/evanboyle/go/src/dynprov/index.ts:36:9)
    at /Users/evanboyle/go/src/dynprov/index.ts:48:9
    at Array.forEach (<anonymous>)
    at /Users/evanboyle/go/src/dynprov/index.ts:47:20
    at Generator.next (<anonymous>)
    at /Users/evanboyle/go/src/dynprov/index.ts:8:71
    at new Promise (<anonymous>)

```

I need to do another pass to figure out at exactly what layer the exhaustion is occurring. This error is coming from the resource monitor, but it is unclear if it originates there or from the dynamic provider. 

It does seem to be related to dynamic providers specifically, as I was not able to get the same behavior when creating a similar number of custom resources (`randomId`). 

Update:

I've made a few improvements in this area to both dynamic provider performance and error logging: 

- Made a perf improvement provider side via #6657 
- Program side perf improvement via #6673 
- Surfacing swallowed errors via #6644

I opened https://github.com/pulumi/pulumi/issues/6675 to track porting these improvements to python. 

There is still somewhat of a logging hole. The error message returned, and the debug logs in the engine do not make it clear that this message originates in the dynamic provider. Theoretically, the resource monitor or some other gRPC service in the call chain could be returning with this error. We should consider adding better instrumentation here. What we currently get back:

```
    error: Running program '/Users/evanboyle/go/src/dynprov' failed with an unhandled exception:
[...]

PR #6646, opened 2021-03-29
cache serialized/deserialized dynamic providers in nodejs

EvanBoyle:
This fix was done as a part of investigating https://github.com/pulumi/pulumi/issues/6645

On the user program side, every instance of a dynamic resource individually serializes its provider. On the dynamic provider side, every call to check, diff, etc, deserialized the dynamic provider in the payload. This change does two things:

1. Add program level caching of serialized dynamic providers
2. Add dynamic provider level caching of deserialized dynamic providers

(1) alone cuts memory usage in the linked repro by 2/3 from 1.5Gb to 500Mb. The repro program also runs in 1/5th the time (100s -> 20s) after this change. 

Unfortunately, the failures in the linked issue still happen although with seemingly lower frequency. 

As for the implementation, both of these rely on globals caches. This is OK, as the caching mechanism is tolerant to multiple provider processes or instances of the SDK. 

We should consider porting this change to python as well.

### DIFF
--- a/sdk/nodejs/cmd/dynamic-provider/index.ts
+++ b/sdk/nodejs/cmd/dynamic-provider/index.ts
@@ -59,28 +59,14 @@ process.on("exit", (code: number) => {
 
 const providerCache: { [key: string]: dynamic.ResourceProvider} = {};
 
-const hashCode = (str: string) => {
-    let hash = 0;
-    let i: number;
-    let chr: number;
-    if (str.length === 0) return hash;
-    for (i = 0; i < str.length; i++) {
-        chr = str.charCodeAt(i);
-        hash = ((hash << 5) - hash) + chr;
-        hash |= 0; // Convert to 32bit integer
-    }
-    return hash;
-};
-
 function getProvider(props: any): dynamic.ResourceProvider {
     const providerString = props[providerKey];
-    const providerHash = hashCode(providerString);
     let provider: any;
-    if(providerCache[providerHash]) {
-        provider = providerCache[providerHash];
+    if(providerCache[providerString]) {
+        provider = providerCache[providerString];
     } else {
         provider = requireFromString(props[providerKey]).handler();
-        providerCache[providerHash] = provider;
+        providerCache[providerString] = provider;
     }
     
     // TODO[pulumi/pulumi#414]: investigate replacing requireFromString with eval
--- a/sdk/nodejs/cmd/dynamic-provider/index.ts
+++ b/sdk/nodejs/cmd/dynamic-provider/index.ts
@@ -57,9 +57,34 @@ process.on("exit", (code: number) => {
     }
 });
 
+const providerCache: { [key: string]: dynamic.ResourceProvider} = {};
+
+const hashCode = (str: string) => {
+    let hash = 0;
+    let i: number;
+    let chr: number;
+    if (str.length === 0) return hash;
+    for (i = 0; i < str.length; i++) {
+        chr = str.charCodeAt(i);
+        hash = ((hash << 5) - hash) + chr;
+        hash |= 0; // Convert to 32bit integer
+    }
+    return hash;
+};
+
 function getProvider(props: any): dynamic.ResourceProvider {
+    const providerString = props[providerKey];
+    const providerHash = hashCode(providerString);
+    let provider: any;
+    if(providerCache[providerHash]) {
+        provider = providerCache[providerHash];
+    } else {
+        provider = requireFromString(props[providerKey]).handler();
+        providerCache[providerHash] = provider;
+    }
+    
     // TODO[pulumi/pulumi#414]: investigate replacing requireFromString with eval
-    return requireFromString(props[providerKey]).handler();
+    return provider
 }
 
 // Each of the *RPC functions below implements a single method of the resource provider gRPC interface. The CRUD
--- a/sdk/nodejs/dynamic/index.ts
+++ b/sdk/nodejs/dynamic/index.ts
@@ -165,6 +165,8 @@ function serializeProvider(provider: ResourceProvider): Promise<string> {
     return runtime.serializeFunction(() => provider).then(sf => sf.text);
 }
 
+const providerCache = new Map<ResourceProvider, Promise<string>>();
+
 /**
  * Resource represents a Pulumi Resource that incorporates an inline implementation of the Resource's CRUD operations.
  */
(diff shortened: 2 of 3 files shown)


sdk/nodejs/cmd/dynamic-provider/index.ts
```diff
@@ -57,9 +57,34 @@ process.on("exit", (code: number) => {
...
+    let i: number;
+    let chr: number;
+    if (str.length === 0) return hash;
+    for (i = 0; i < str.length; i++) {
+        chr = str.charCodeAt(i);
+        hash = ((hash << 5) - hash) + chr;
+        hash |= 0; // Convert to 32bit integer
+    }
+    return hash;
+};
```
pgavlin, line 73: I think we should just use the provider string itself as the cache key. Using the hash code without keeping the original string around could lead to incorrect behavior if two different strings hash to the same value.

sdk/nodejs/dynamic/index.ts
```diff
@@ -184,7 +186,15 @@ export abstract class Resource extends resource.CustomResource {
         if (props[providerKey]) {
             throw new Error("A dynamic resource must not define the __provider key");
         }
-        props[providerKey] = serializeProvider(provider);
+        let serializedProvider: Promise<string>;
+        if (providerCache.has(provider)) {
```
pgavlin, line 190: Will this be valid if the implementations of the various functions that implement the resource provider have changed? I'm thinking in particular of the functions being reassigned to new implementations.

pgavlin (changes requested):

sdk/nodejs/cmd/dynamic-provider/index.ts
EvanBoyle: Good point, at 1k providers the chance of collision becomes 1/10000. Storing the entire provider is probably not a big deal as the common case is multiple instances of a few providers over thousands of single instances of unique providers.

sdk/nodejs/dynamic/index.ts
EvanBoyle: Unfortunately, there are endless ways that this could be invalid. For instance, even if you don't mutate the functions you could increment some shared global variable. Through that lens, this is technically a breaking change although one that seems less likely for users to hit. Through that lense I see a few options:

1. accept this change, but extend the diffing logic to check equality of each of the provider functions. This still has issues as noted, but probably covers the 99.99%
2. Make this behavior opt-in (maybe via `dynamic.CacheableResourceProvider`). I don't like this option as this behavior should really be the default. 

Thoughts? I suppose we could do (2), and cut over to (1) as a part of 3.0.

pgavlin: > Unfortunately, there are endless ways that this could be invalid. For instance, even if you don't mutate the functions you could increment some shared global variable. Through that lens, this is technically a breaking change although one that seems less likely for users to hit.

This is a good point. FWIW, we probably haven't documented that closing over mutable global state in a dynamic provider is generally a bad idea, since the provider won't observe any mutations once it is serialized.

> Thoughts? I suppose we could do (2), and cut over to (1) as a part of 3.0.

This seems good to me, though I would use an envvar for the opt-in (despite the big hammer, it doesn't require code changes).

Honestly, I'm probably being a little too careful with this particular aspect of the change. @lukehoban wdyt?

sdk/nodejs/cmd/dynamic-provider/index.ts
EvanBoyle: Done.

sdk/nodejs/dynamic/index.ts
lukehoban: FWIW - I'm actually not sure this cache will help in many/most cases.  It's pretty common style to do something like `super(new MyResourceProvider(), ...)` or equivalent.  In fact, of the 5 examples of TypeScript dynamic providers in pulumi/examples, I'm pretty sure only one of them would see any benefit from this cache.  

As for the breaking change aspect of this - I think given we are so close to 3.0, and that this isn't a particularly practically severe issue (4k dynamic resources is meaningfully outside of current "normal"), we could just choose to only deliver this side of the change into 3.0, and take it as a documented breaking change.

> FWIW, we probably haven't documented that closing over mutable global state in a dynamic provider is generally a bad idea, since the provider won't observe any mutations once it is serialized.

We should probably add a paragraph to the dynamic providers docs on this point, both in terms of the impact that closure serialization has here, and that this caching has.

sdk/nodejs/cmd/dynamic-provider/index.ts
```diff
@@ -57,9 +57,20 @@ process.on("exit", (code: number) => {
     }
 });
 
+const providerCache: { [key: string]: dynamic.ResourceProvider} = {};
```
justinvp, line 60: Super nitpick: 

```suggestion
const providerCache: { [key: string]: dynamic.ResourceProvider } = {};
```

sdk/nodejs/cmd/dynamic-provider/index.ts
```diff
@@ -57,9 +57,20 @@ process.on("exit", (code: number) => {
...
+
 function getProvider(props: any): dynamic.ResourceProvider {
+    const providerString = props[providerKey];
+    let provider: any;
+    if(providerCache[providerString]) {
+        provider = providerCache[providerString];
+    } else {
+        provider = requireFromString(props[providerKey]).handler();
+        providerCache[providerString] = provider;
+    }
```
justinvp, line 70: `providerString` could be quite large. I'm not sure how much V8 will optimize how it's currently written, but looking it up in the cache once instead of twice may have a measurable improvement for large provider strings, and it's less code:

```suggestion
    const providerString = props[providerKey];
    let provider: any = providerCache[providerString];
    if (!provider) {
        provider = requireFromString(providerString).handler();
        providerCache[providerString] = provider;
    }
```

sdk/nodejs/cmd/dynamic-provider/index.ts
```diff
@@ -57,9 +57,20 @@ process.on("exit", (code: number) => {
...
+    
     // TODO[pulumi/pulumi#414]: investigate replacing requireFromString with eval
-    return requireFromString(props[providerKey]).handler();
+    return provider
```
justinvp, line 73: Nit:

```suggestion
    return provider;
```

sdk/nodejs/dynamic/index.ts
EvanBoyle: Another thing to note here is that as long as serialization is deterministic, we'll get cache hits on the provider side even if they use the pattern you describe @lukehoban.

EvanBoyle:
closing in favor of https://github.com/pulumi/pulumi/pull/6657 and https://github.com/pulumi/pulumi/pull/6673

PR #6654, opened 2021-03-30
Use CliWrap for subprocess management

t0yv0:
Moves from System.Diagnostics.Process to CliWrap for managing pulumi sub-process.

As discussed in https://github.com/pulumi/pulumi/pull/6572 - this is the rebased version of https://github.com/pulumi/pulumi/pull/6640 that was reviewed already.


vipentti:
Looks good to me!

Issue #6632, opened 2021-03-29
Automation: overwriting existing Pulumi.yaml


headconnect:
When running an automation script with the name of the project specified, as well as workdir specified as `"."`, automation overwrites the existing `Pulumi.yaml` file instead of reading its contents.

This also means that it will not find the selected stack as, in my case, the Pulumi.yaml contains the "config" parameter to indicate that my stacks are stored in the `stacks/` folder. Therefore it creates a new stack in `"."` and complains that it cannot read config.requireObject("---") parameters.

## Expected behavior

I would expect that automation attempts to read the existing `Pulumi.yaml` file before overwriting it, and thus using the correct stack.

## Current behavior

Currently automation will overwrite an existing `Pulumi.yaml` with just `name:` and `runtime:` fields, where name is a required parameter set to LocalWorkspace.selectStack().

Workaround is to manually read the `Pulumi.yaml` and pass its contents to the `projectSettings` key of the second parameter to LocalWorkspace.selectStack(). This will still overwrite `Pulumi.yaml` but it will be overwritten with the same contents (just slightly different formating). 

## Steps to reproduce

Example of reproduction here:
https://github.com/headconnect/pulumi-bug-reproducer/tree/issue/automation-6632-overwriting-pulumi-yaml

## Context (Environment)

This occurs with both tsc & ts-node execution on windows 10. 

This makes it scary to use automation on any form of important resource stacks, as the behavior might be unpredictable.

EvanBoyle:
Yes, this looks like a bug. The implementation checks for the presence of a projectSettings object. It should also check if a workdir containing a project is specified. Probably it should work as:

1. Use project settings if specified
1. If workdir is specified, try to call `getProjectSettings`
2. Finally, create the default project if this is an inline program.

PR #6655, opened 2021-03-30
[automation/go,nodejs,python] - Respect existing project settings

komalali:
Go, nodejs and python part of #6632

### DIFF
--- a/sdk/go/x/auto/local_workspace.go
+++ b/sdk/go/x/auto/local_workspace.go
@@ -57,17 +57,7 @@ var settingsExtensions = []string{".yaml", ".yml", ".json"}
 // LocalWorkspace reads settings from the Pulumi.yaml in the workspace.
 // A workspace can contain only a single project at a time.
 func (l *LocalWorkspace) ProjectSettings(ctx context.Context) (*workspace.Project, error) {
-	for _, ext := range settingsExtensions {
-		projectPath := filepath.Join(l.WorkDir(), fmt.Sprintf("Pulumi%s", ext))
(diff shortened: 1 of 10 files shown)


sdk/go/x/auto/local_workspace.go
```diff
@@ -972,3 +965,59 @@ func getStackSettingsName(stackName string) string {
...
+	if optsBag.Project != nil {
+		return optsBag.Project, nil
+	}
+
+	// If WorkDir is specified, try to read any existing project settings before resorting to
+	// creating a default project.
+	if optsBag.WorkDir != "" {
+		proj, err := readProjectSettingsFromDir(ctx, optsBag.WorkDir)
+		if err == nil {
+			return proj, nil
```
EvanBoyle, line 1003: Minor, but I don't think we should round trip the YAML in the noop case as it can cause comments to be stripped.

If the user specifies a workDir with a pulumi.yaml, we will read in the contents and then write them back out onto disk when we create the workspace.

EvanBoyle: Looks like this applies nodejs and elsewhere. We likely don't want to specify a projectSettings in this case, just avoid setting the default.

EvanBoyle (approved):
Looks good! One small behavior change that would be nice to make. If not as a part of this change, as a fast follow tomorrow that can go out in the next release.

sdk/go/x/auto/local_workspace.go
komalali: I'm gonna merge this and make another PR with those changes.

PR #6657, opened 2021-03-31
Add provider side caching for dynamic provider deserialization

EvanBoyle:
Contributes to https://github.com/pulumi/pulumi/issues/6645
Closes https://github.com/pulumi/customer-support/issues/360

I went ahead and broke out the non-breaking portion of https://github.com/pulumi/pulumi/pull/6646 into a separate PR. 

This change caches deserialized dynamic providers, so we don't have to repeatedly call `requireFromString` for each diff, check, etc call when many of the same dynamic resource are created. That library ultimately bottoms out on `vm.createContext` https://nodejs.org/api/vm.html#vm_what_does_it_mean_to_contextify_an_object

There was a perf regression in node v12 on this method that exacerbates the issue for our users on that version: https://github.com/nodejs/node/issues/29842

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -3,6 +3,9 @@
 
 ### Improvements
 
+- [sdk/nodejs] Add provider side caching for dynamic provider deserialization
+  [#6657](https://github.com/pulumi/pulumi/pull/6657)
+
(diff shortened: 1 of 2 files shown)


lukehoban (approved):

justinvp (approved):
LGTM

PR #6658, opened 2021-03-31
[automation/dotnet] - Respect existing project settings for inline projects.

komalali:
Missing the fix

Dotnet part of #6632


t0yv0:
Giving it a shot, hold on.

PR #6659, opened 2021-03-31
Ensure that v is passed to the ldflags as part of windows binary creation

stack72:


stack72:
Going to self merge this - this is actually only testable when we make a release (ie. master or a tagged release) so I want to get this tested before v2.24.0

Issue #545, opened 2017-11-09
Support tracing in the JavaScript langhost implementation


lukehoban:
With #521 we have added support for OpenTracing through the engine, providers and language hosts.  However, we do not yet generate any traces from the JavaScript side of the langhost RPC boundary.  In practice, this doesn't obscure too much, but we may be able to get additional insight by instrumenting the langhost with key spans - such as resource creation.

joeduffy:
I think we've decided to move away from Jaeger tracing overall, so closing out. Please reopen if this is still something we will do.

PR #6668, opened 2021-03-31
Fix to verbose log stack traces: report origin not logutil

t0yv0:
I want diagnostic traces to be more helpful in locating code that originated them. Currently seeing things like this:

```
I0330 17:03:36.872190   86713 eventsink.go:88] eventSink::Error(<{%reset%}>unexpected unknown property value<{%reset%}>)
```

Now eventsink.go:88 is a helper logging function does not tell me where this is from.

It turns out `glog` provides a `depth` stack frame offset to see how far to climb the stack to report the origination of the error. By this logic, log helper functions should call another logging function with `depth+1` to avoid being reported as the origin.

I have not gotten to the bottom of this but already having errors a little better, like this:

```
I0330 17:55:23.657542   10870 deployment_executor.go:112] eventSink::Error(<{%reset%}>unexpected unknown property value<{%reset%}>)
```

Now deployment_executor.go:112 is a log helper also so digging a bit deeper here.

### DIFF
--- a/pkg/resource/deploy/step_executor.go
+++ b/pkg/resource/deploy/step_executor.go
@@ -324,11 +324,13 @@ func (se *stepExecutor) executeStep(workerID int, step Step) error {
 func (se *stepExecutor) log(workerID int, msg string, args ...interface{}) {
 	if logging.V(stepExecutorLogLevel) {
 		message := fmt.Sprintf(msg, args...)
-		logging.InfofDepth(
-			1,
-			"StepExecutor worker(%d): %s",
-			workerID,
-			message)
+		if logging.V(stepExecutorLogLevel) {
+			logging.InfofDepth(
+				1,
+				"StepExecutor worker(%d): %s",
+				workerID,
+				message)
+		}
 	}
 }
 


pkg/engine/eventsink.go
```diff
@@ -37,65 +37,96 @@ type eventSink struct {
...
 	case diag.Error:
-		s.Errorf(d, args...)
+		s.ErrorfDepth(depth+1, d, args...)
 	default:
 		contract.Failf("Unrecognized severity: %v", sev)
 	}
 }
 
-func (s *eventSink) Debugf(d *diag.Diag, args ...interface{}) {
+func (s *eventSink) DebugfDepth(depth int, d *diag.Diag, args ...interface{}) {
```
pgavlin, line 62: Just curious: do you expect that we may call these methods with `depth > 1` in the future?

t0yv0: Eh, nothing thoughtful here, if it's unused I'd like to delete. Let me try deleting.

pkg/resource/deploy/step_executor.go
```diff
@@ -324,7 +324,11 @@ func (se *stepExecutor) executeStep(workerID int, step Step) error {
 func (se *stepExecutor) log(workerID int, msg string, args ...interface{}) {
 	if logging.V(stepExecutorLogLevel) {
 		message := fmt.Sprintf(msg, args...)
-		logging.V(stepExecutorLogLevel).Infof("StepExecutor worker(%d): %s", workerID, message)
```
t0yv0, line 327: This edit is not good, need `if logging.V(stepExecutorLogLevel)`.

t0yv0:
This turned out not as useful as I hoped, leaving it on the back burner to circle back later. What I really want is visibility across gRPC boundary and logs with stack traces when gRPC servers malfunction. I played a little bit with gRPC interceptors - something to check out later.

Frassle:
This would actually be really nice to fix up. All the logs showing up as `eventsink.go` is a real annoyance when debugging logs from users.

mikhailshilkov:
Let's close the PR but make sure there is an issue tracking this, and point it to the PR? FWIW, we have error improvements on the roadmap for Q3.

squaremo:
> What I really want is visibility across gRPC boundary

Tracing would help with this. I see tracing flags in `Pulumi` and that `pulumi-language-nodejs` is instrumented, and extrapolating from that, following an invocation through gRPC to see its fate is possible in principle. Although: https://github.com/pulumi/pulumi/issues/545#issuecomment-434112323.

> Let's close the PR but make sure there is an issue tracking this, and point it to the PR

--> https://github.com/pulumi/pulumi/issues/9849. @t0yv0 if you feel this PR can be continued, please just reopen it.

PR #6670, opened 2021-03-31
Avoid overriding dotnet proj settings accidentally

t0yv0:
It's been a while since I did .NET (mostly F# background) so appreciate comments on what's C#-idiomatic, is the IEqualityComparer thing over the top? IN F# we take for granted auto-gen value equality on maps etc.

### DIFF
--- a/sdk/dotnet/Pulumi.Automation.Tests/LocalWorkspaceTests.cs
+++ b/sdk/dotnet/Pulumi.Automation.Tests/LocalWorkspaceTests.cs
@@ -908,7 +908,7 @@ namespace Pulumi.Automation.Tests
                 Assert.True(expSecretValue.IsSecret);
             }
         }
-    
+
         [Fact]
         public async Task PulumiVersionTest()
         {
@@ -938,6 +938,22 @@ namespace Pulumi.Automation.Tests
             {
                 LocalWorkspace.ValidatePulumiVersion(testMinVersion, currentVersion);
             }
+        }
+
+        [Fact]
+        public async Task RespectsProjectSettingsTest()
+        {
+            var program = PulumiFn.Create<ValidStack>();
+
+            var stackName = $"{RandomStackName()}";
+            var projectName = "project_was_overwritten";
+            var stack = await LocalWorkspace.CreateStackAsync(new InlineProgramArgs(projectName, stackName, program)
+            {
+                WorkDir = Path.Combine(_dataDirectory, "correct_project")
+            });
+            var settings = await stack.Workspace.GetProjectSettingsAsync();
+            Assert.Equal("correct_project", settings!.Name);
+            Assert.Equal("This is a description", settings.Description);
         }
     }
 }
--- a/sdk/dotnet/Pulumi.Automation.Tests/LocalWorkspaceTests.cs
+++ b/sdk/dotnet/Pulumi.Automation.Tests/LocalWorkspaceTests.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Reflection;
+using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
 using Pulumi.Automation.Commands.Exceptions;
@@ -16,9 +17,6 @@ namespace Pulumi.Automation.Tests
 {
     public class LocalWorkspaceTests
     {
-        private static readonly string _dataDirectory =
-            Path.Combine(new FileInfo(Assembly.GetExecutingAssembly().Location).DirectoryName, "Data");
-
         private static readonly string _pulumiOrg = GetTestOrg();
 
         private static string GetTestSuffix()
@@ -55,7 +53,7 @@ namespace Pulumi.Automation.Tests
         [InlineData("json")]
         public async Task GetProjectSettings(string extension)
         {
-            var workingDir = Path.Combine(_dataDirectory, extension);
+            var workingDir = ResourcePath(Path.Combine("Data", extension));
             using var workspace = await LocalWorkspace.CreateAsync(new LocalWorkspaceOptions
             {
                 WorkDir = workingDir,
@@ -74,7 +72,7 @@ namespace Pulumi.Automation.Tests
         [InlineData("json")]
         public async Task GetStackSettings(string extension)
         {
-            var workingDir = Path.Combine(_dataDirectory, extension);
+            var workingDir = ResourcePath(Path.Combine("Data", extension));
             using var workspace = await LocalWorkspace.CreateAsync(new LocalWorkspaceOptions
             {
                 WorkDir = workingDir,
@@ -283,7 +281,7 @@ namespace Pulumi.Automation.Tests
         public async Task StackLifecycleLocalProgram()
         {
             var stackName = $"{RandomStackName()}";
-            var workingDir = Path.Combine(_dataDirectory, "testproj");
+            var workingDir = ResourcePath(Path.Combine("Data", "testproj"));
             using var stack = await LocalWorkspace.CreateStackAsync(new LocalProgramArgs(stackName, workingDir)
             {
                 EnvironmentVariables = new Dictionary<string, string>()
@@ -622,7 +620,7 @@ namespace Pulumi.Automation.Tests
             }
 
             static async Task<T> RunCommand<T, TOptions>(Func<TOptions, CancellationToken, Task<T>> func, string command)
-                where TOptions: UpdateOptions, new()
+                where TOptions : UpdateOptions, new()
             {
                 var events = new List<EngineEvent>();
 
@@ -947,13 +945,24 @@ namespace Pulumi.Automation.Tests
 
             var stackName = $"{RandomStackName()}";
             var projectName = "project_was_overwritten";
-            var stack = await LocalWorkspace.CreateStackAsync(new InlineProgramArgs(projectName, stackName, program)
-            {
-                WorkDir = Path.Combine(_dataDirectory, "correct_project")
-            });
+
+            var workdir = ResourcePath(Path.Combine("Data", "correct_project"));
+
+            var stack = await LocalWorkspace.CreateStackAsync(
+                new InlineProgramArgs(projectName, stackName, program)
+                {
+                    WorkDir = workdir
+                });
+
             var settings = await stack.Workspace.GetProjectSettingsAsync();
             Assert.Equal("correct_project", settings!.Name);
             Assert.Equal("This is a description", settings.Description);
+        }
+
+        private string ResourcePath(string path, [CallerFilePath] string pathBase = "LocalWorkspaceTests.cs")
+        {
+            var dir = Path.GetDirectoryName(pathBase) ?? ".";
+            return Path.Combine(dir, path);
         }
     }
 }
--- a/sdk/dotnet/Pulumi.Automation.Tests/LocalWorkspaceTests.cs
+++ b/sdk/dotnet/Pulumi.Automation.Tests/LocalWorkspaceTests.cs
@@ -9,6 +9,7 @@ using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
+using Pulumi.Automation.Exceptions;
 using Pulumi.Automation.Commands.Exceptions;
 using Pulumi.Automation.Events;
 using Xunit;
@@ -957,6 +958,29 @@ namespace Pulumi.Automation.Tests
             var settings = await stack.Workspace.GetProjectSettingsAsync();
             Assert.Equal("correct_project", settings!.Name);
             Assert.Equal("This is a description", settings.Description);
+        }
+
+        [Fact]
+        public async Task DetectsProjectSettingConflictTest()
+        {
+            var program = PulumiFn.Create<ValidStack>();
+
+            var stackName = $"{RandomStackName()}";
+            var projectName = "project_was_overwritten";
+
+            var workdir = ResourcePath(Path.Combine("Data", "correct_project"));
+
+            var projectSettings = ProjectSettings.Default(projectName);
+            projectSettings.Description = "non-standard description";
+
+            await Assert.ThrowsAsync<ProjectSettingsConflictException>(() =>
+                LocalWorkspace.CreateStackAsync(
+                    new InlineProgramArgs(projectName, stackName, program)
+                    {
+                        WorkDir = workdir,
+                        ProjectSettings = projectSettings
+                    })
+            );
         }
 
         private string ResourcePath(string path, [CallerFilePath] string pathBase = "LocalWorkspaceTests.cs")
--- a/sdk/dotnet/Pulumi.Automation/DictionaryContentsComparer.cs
+++ b/sdk/dotnet/Pulumi.Automation/DictionaryContentsComparer.cs
@@ -26,6 +26,10 @@ namespace Pulumi.Automation
             {
                 return x == null;
             }
+            if (ReferenceEquals(x, y))
+            {
+                return true;
+            }
             if (x.Count != y.Count)
             {
                 return false;
--- a/sdk/dotnet/Pulumi.Automation/DictionaryContentsComparer.cs
+++ b/sdk/dotnet/Pulumi.Automation/DictionaryContentsComparer.cs
@@ -1,0 +1,54 @@
+// Copyright 2016-2019, Pulumi Corporation
+
+using System.Collections.Generic;
+
+namespace Pulumi.Automation
+{
+    /// Compares two dictionaries for equality by content, as F# maps would.
+    internal sealed class DictionaryContentsComparer<K, V> : IEqualityComparer<IDictionary<K, V>> where K : notnull
+    {
+        private IEqualityComparer<K> _keyComparer;
+        private IEqualityComparer<V> _valueComparer;
+
+        public DictionaryContentsComparer(IEqualityComparer<K> keyComparer, IEqualityComparer<V> valueComparer)
+        {
+            this._keyComparer = keyComparer;
+            this._valueComparer = valueComparer;
+        }
+
+        bool IEqualityComparer<IDictionary<K, V>>.Equals(IDictionary<K, V>? x, IDictionary<K, V>? y)
+        {
+            if (x == null)
+            {
+                return y == null;
+            }
+            if (y == null)
+            {
+                return x == null;
+            }
+            if (x.Count != y.Count)
+            {
+                return false;
+            }
+            var y2 = new Dictionary<K, V>(y, this._keyComparer);
+            foreach (var pair in x)
+            {
+                if (!y2.ContainsKey(pair.Key))
+                {
+                    return false;
+                }
+
+                if (!this._valueComparer.Equals(pair.Value, y2[pair.Key]))
+                {
+                    return false;
+                }
+            }
+            return true;
+        }
+
+        int IEqualityComparer<IDictionary<K, V>>.GetHashCode(IDictionary<K, V> obj)
+        {
+            return 0; // inefficient but correct
+        }
+    }
+}
--- a/sdk/dotnet/Pulumi.Automation/DictionaryContentsComparer.cs
+++ b/sdk/dotnet/Pulumi.Automation/DictionaryContentsComparer.cs
@@ -7,8 +7,8 @@ namespace Pulumi.Automation
     /// Compares two dictionaries for equality by content, as F# maps would.
     internal sealed class DictionaryContentsComparer<K, V> : IEqualityComparer<IDictionary<K, V>> where K : notnull
     {
-        private IEqualityComparer<K> _keyComparer;
-        private IEqualityComparer<V> _valueComparer;
+        private readonly IEqualityComparer<K> _keyComparer;
+        private readonly IEqualityComparer<V> _valueComparer;
 
         public DictionaryContentsComparer(IEqualityComparer<K> keyComparer, IEqualityComparer<V> valueComparer)
         {
--- a/sdk/dotnet/Pulumi.Automation/Exceptions/ProjectSettingsConflictException.cs
+++ b/sdk/dotnet/Pulumi.Automation/Exceptions/ProjectSettingsConflictException.cs
@@ -4,7 +4,22 @@ using System;
 
 namespace Pulumi.Automation.Exceptions
 {
-    internal class ProjectSettingsConflictException : Exception
+    /// <summary>
+    ///
+    /// Thrown when creating a Workspace detects a conflict between
+    /// project settings found on disk (such as Pulumi.yaml) and a
+    /// ProjectSettings object passed to the Create API.
+    ///
+    /// There are two resolutions:
+    ///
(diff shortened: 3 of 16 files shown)


sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
```diff
@@ -341,6 +342,31 @@ public static Task<WorkspaceStack> CreateOrSelectStackAsync(LocalProgramArgs arg
             this._readyTask = Task.WhenAll(readyTasks);
         }
 
+        private async Task InitializeProjectSettingsAsync(ProjectSettings projectSettings,
```
t0yv0, line 345: This is the meat of the change right here. The Comparer* boilerplate noise is to implement `projectSettings.IsDefault`. The assumption here its that if the user passes non-default settings, we want to overwrite what is on disk already, and put out a warning (though not clear where warnings go?).

Alternatively we could throw and hard-fail when both non-default settings are passed in and there is something on disk already.

Now, avoiding IsDefault check would be nice but it seems sometimes our API creates default ProjectSettings invisibly so it's hard to be sure it was the user's intention. In the test it seems the user's intention is to use the settings from disk, not from the default ProjectSettings object constructed by the framework.

sdk/dotnet/Pulumi.Automation.Tests/LocalWorkspaceTests.cs
```diff
@@ -939,5 +937,32 @@ public void ValidVersionTheory(string currentVersion, bool errorExpected)
...
+
+            var stackName = $"{RandomStackName()}";
+            var projectName = "project_was_overwritten";
+
+            var workdir = ResourcePath(Path.Combine("Data", "correct_project"));
+
+            var stack = await LocalWorkspace.CreateStackAsync(
+                new InlineProgramArgs(projectName, stackName, program)
+                {
+                    WorkDir = workdir
```
t0yv0, line 954: The constructor of InlineProgramArgs crates default ProjectSettings implicitly here, without user intending to customize them.

sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
```diff
@@ -341,6 +342,31 @@ public static Task<WorkspaceStack> CreateOrSelectStackAsync(LocalProgramArgs arg
...
+            }
+            else
+            {
+                if (!projectSettings.IsDefault)
+                {
+                    // TODO we are overwriting settings found on disk
+                    // with settings passed in through code. Seems
+                    // like a good place to emit a warning, but not
+                    // clear which logging facility to use.
+                    await this.SaveProjectSettingsAsync(projectSettings, cancellationToken);
```
komalali, line 365: Can you explain this a bit? Why are we overwriting settings found on disk with settings passed in through code? 

Settings already on disk should always take precedence. There are 3 possible scenarios:
1. Create an inline program stack that passes in a `WorkDir` LocalWorkspaceOption - in this case we should check if the workdir has a Pulumi.yaml and read it. If that succeeds, do nothing (i.e do not set projectSettings)
2. Create an inline program stack that passes in a `ProjectSettings` LocalWorkspaceOption - in this case, save the passed in `ProjectSettings`.
3. Create an inline program with neither of the above. In this case, create the default project settings and save those.

Although I guess this begs the question of what if both workdir and project settings are passed in. Is that what you're addressing here? 

In my implementation for nodejs/python/go, the existing `Pulumi.yaml` always takes precedence - but I could see how that might also be confusing for the user if they have passed in `ProjectSettings` and they are not respected.

t0yv0: Ok let me try that! To clarify, one ya non-default ProjectSettings. If Default ProjectSettings is passed in, then a lot of code that did not bother customizing the settings, including our tests, will start failing, simply because the framework creates a default ProjectSettings.

komalali: >Are WorkDir and ProjectSettings mutually exclusive? Hm I guess not we might have a user that wants to set both.

Yeah those two settings are not mutually exclusive. Because a `WorkDir` may not have a pre-existing `Pulumi.yaml`. We only run into trouble if the workdir already has settings.

komalali: >Ok let me try that! To clarify, only non-default ProjectSettings. If Default ProjectSettings is passed in, then a lot of code that did not bother customizing the settings, including our tests, will start failing, simply because the framework creates a default ProjectSettings.

🤔 We should not create default project settings if there is already a `Pulumi.yaml` in the `WorkDir` or if a user passes in `ProjectSettings`.

t0yv0: I think the last version gets to that behavior.

It's a bit awkward because of the comparer changes.

I'm poking around to see if there is anything simpler; perhaps we could remove current uses of ProjectSettings.IsDefault and use `null` value as "default", and only coerce null to default when needed.

t0yv0: I tried that and it's not pretty, I prefer what's in here; the ProjectSettings currently carries 'required' arguments like ProjectName and runtime (defaulting to nodejs), we'd have to remove null checks and add ProjectName back to something like LocalWorkspaceOptions, it's a bit confusing. I will leave as is for now unless there's more pushback.

komalali: How it is now is fine with me. 🙏🏽

komalali (approved):
Approving from a behavior perspective, I'll let the C# folks speak to the language-specific changes.

I thought some more about throwing the exception in the case of conflicting project settings, and I feel that's the right way to go to make behavior explicit. 

If folks are intentionally trying to overwrite the existing file, they can do so in a follow-up step with `SaveProjectSettingsAsync`. If this appears to be a common-enough workflow, we can consider adding a flag like `overwriteExistingSettings` for the user to explicitly signal that they intended the overwrite. But we can cross that bridge if/when we get to it.

sdk/dotnet/Pulumi.Automation/ProjectRuntimeOptions.cs
```diff
@@ -29,5 +32,34 @@ public class ProjectRuntimeOptions
...
+                return x.TypeScript == y.TypeScript && x.Binary == y.Binary && x.VirtualEnv == y.VirtualEnv;
+            }
+
+            int IEqualityComparer<ProjectRuntimeOptions>.GetHashCode(ProjectRuntimeOptions obj)
+            {
+                var hash = new HashCode();
+                hash.Add(obj.TypeScript);
+                hash.Add(obj.Binary);
+                hash.Add(obj.VirtualEnv);
+                return hash.ToHashCode();
```
vipentti, line 61: Can consider using [HashCode.Combine](https://docs.microsoft.com/en-us/dotnet/api/system.hashcode.combine?view=net-5.0#System_HashCode_Combine__3___0___1___2_)

```suggestion
                return HashCode.Combine(obj.TypeScript, obj.Binary, obj.VirtualEnv);
```

sdk/dotnet/Pulumi.Automation/ProjectBackend.cs
```diff
@@ -8,5 +11,32 @@ namespace Pulumi.Automation
...
+                }
+
+                return x.Url == y.Url;
+            }
+
+            int IEqualityComparer<ProjectBackend>.GetHashCode(ProjectBackend obj)
+            {
+                var hash = new HashCode();
+                hash.Add(obj.Url);
+                return hash.ToHashCode();
```
vipentti, line 38: Can consider using [HashCode.Combine](https://docs.microsoft.com/en-us/dotnet/api/system.hashcode.combine?view=net-5.0)

```suggestion
                return HashCode.Combine(obj.Url);
```

sdk/dotnet/Pulumi.Automation/ProjectTemplate.cs
```diff
@@ -16,5 +17,44 @@ public class ProjectTemplate
...
+            }
+
+            int IEqualityComparer<ProjectTemplate>.GetHashCode(ProjectTemplate obj)
+            {
+                var hash = new HashCode();
+                hash.Add(obj.Description);
+                hash.Add(obj.QuickStart);
+                hash.Add(obj.Important);
+                // omit hashing Config dict for efficiency
+                return hash.ToHashCode();
```
vipentti, line 56: Can consider using [HashCode.Combine](https://docs.microsoft.com/en-us/dotnet/api/system.hashcode.combine?view=net-5.0)

```suggestion
                // omit hashing Config dict for efficiency
                return HashCode.Combine(obj.Description, obj.QuickStart, obj.Important);
```

sdk/dotnet/Pulumi.Automation/ProjectTemplateConfigValue.cs
```diff
@@ -12,5 +15,34 @@ public class ProjectTemplateConfigValue
...
+                return x.Description == y.Description && x.Default == y.Default && x.Secret == y.Secret;
+            }
+
+            int IEqualityComparer<ProjectTemplateConfigValue>.GetHashCode(ProjectTemplateConfigValue obj)
+            {
+                var hash = new HashCode();
+                hash.Add(obj.Description);
+                hash.Add(obj.Default);
+                hash.Add(obj.Secret);
+                return hash.ToHashCode();
```
vipentti, line 44: Can consider using [HashCode.Combine](https://docs.microsoft.com/en-us/dotnet/api/system.hashcode.combine?view=net-5.0)

```suggestion
                return HashCode.Combine(obj.Description, obj.Default, obj.Secret);
```

sdk/dotnet/Pulumi.Automation/ProjectRuntime.cs
```diff
@@ -15,5 +18,37 @@ public ProjectRuntime(ProjectRuntimeName name)
...
+
+            int IEqualityComparer<ProjectRuntime>.GetHashCode(ProjectRuntime obj)
+            {
+                var hash = new HashCode();
+                hash.Add(obj.Name);                
+                if (obj.Options != null)
+                {
+                    hash.Add(ProjectRuntimeOptions.Comparer.GetHashCode(obj.Options));
+                }
+                return hash.ToHashCode();
```
vipentti, line 49: Can consider using [HashCode.Combine](https://docs.microsoft.com/en-us/dotnet/api/system.hashcode.combine?view=net-5.0)

In this case we default to 0 if Options are null, which HashCode.Combine also does

```suggestion
                return HashCode.Combine(
                    obj.Name,
                    obj.Options != null ? ProjectRuntimeOptions.Comparer.GetHashCode(obj.Options) : 0
                );
```

sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
```diff
@@ -358,52 +381,62 @@ private async Task PopulatePulumiVersionAsync(CancellationToken cancellationToke
...
         public override async Task<ProjectSettings?> GetProjectSettingsAsync(CancellationToken cancellationToken = default)
         {
-            foreach (var ext in SettingsExtensions)
+            var path = this.FindSettingsFile();
+            var isJson = Path.GetExtension(path) == ".json";
+            if (!File.Exists(path))
+            {
+                return null;
+            }
+            else
```
vipentti, line 403: I think this `else` can be removed to reduce nesting, since we already return if the file does not exist

sdk/dotnet/Pulumi.Automation/DictionaryContentsComparer.cs
```diff
@@ -0,0 +1,54 @@
...
+
+using System.Collections.Generic;
+
+namespace Pulumi.Automation
+{
+    /// Compares two dictionaries for equality by content, as F# maps would.
+    internal sealed class DictionaryContentsComparer<K, V> : IEqualityComparer<IDictionary<K, V>> where K : notnull
+    {
+        private IEqualityComparer<K> _keyComparer;
+        private IEqualityComparer<V> _valueComparer;
```
vipentti, line 11: These can be made `readonly` since they are only assigned once in the constructor

```suggestion
        private readonly IEqualityComparer<K> _keyComparer;
        private readonly IEqualityComparer<V> _valueComparer;
```

sdk/dotnet/Pulumi.Automation/Exceptions/ProjectSettingsConflictException.cs
```diff
@@ -0,0 +1,24 @@
+// Copyright 2016-2021, Pulumi Corporation
+
+using System;
+
+namespace Pulumi.Automation.Exceptions
+{
+    internal class ProjectSettingsConflictException : Exception
```
vipentti, line 7: Does Pulumi have a policy whether exceptions should be public? 

Can consider making exceptions inheriting from `System.Exception` public:
> An internal exception is only visible inside its own internal scope. After the exception falls outside the internal scope, only the base exception can be used to catch the exception. If the internal exception is inherited from Exception, SystemException, or ApplicationException, the external code will not have sufficient information to know what to do with the exception.

Source: https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1064

t0yv0:
Awesome thanks for review @vipentti

sdk/dotnet/Pulumi.Automation/ProjectSettings.cs
```diff
@@ -44,5 +47,57 @@ public class ProjectSettings
...
+                hash.Add(obj.Name);
+                hash.Add(obj.Main);
+                hash.Add(obj.Description);
+                hash.Add(obj.Author);
+                hash.Add(obj.Website);
+                hash.Add(obj.License);
+                hash.Add(obj.Config);
+                hash.Add(obj.Backend);
+                // fields with custom Comparer skipped for efficiency
+                return hash.ToHashCode();
```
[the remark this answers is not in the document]
t0yv0: That's fine.. omitting values from hashing is fine. I don't have concrete use cases for GetHashCode here it's just for correctness sake so GetHashCode corresponds to the same notion of equality as Equals. Just in case.. Even `return 0` is an inefficient but valid implementation.

sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
t0yv0: Ah right this is one of those style things i'm unsure about. I'll take your suggestion in this case. In F# there is high tolerance for nesting as indeed the language does not even have an early return construct :)

sdk/dotnet/Pulumi.Automation/Exceptions/ProjectSettingsConflictException.cs
t0yv0: Right. I was stumped about RS0016 I got when trying to make this public but looks like I found what it's about, adding some notes to README.md and a comment on the new exception.

t0yv0:
CC @orionstudt - is this a reasonable way to add a new public exception? Ty.

sdk/dotnet/Pulumi.Automation/Exceptions/ProjectSettingsConflictException.cs
orionstudt: I think it should be public with an internal constructor, That way it can be caught explicitly if for any reason a consumer wishes to do so.

sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
```diff
@@ -341,6 +342,28 @@ public static Task<WorkspaceStack> CreateOrSelectStackAsync(LocalProgramArgs arg
...
+            // If given project settings, we want to write them out to
+            // the working dir. We do not want to override existing
+            // settings with default settings though.
+
+            var existingSettings = await this.GetProjectSettingsAsync(cancellationToken);
+            if (existingSettings == null)
+            {
+                await this.SaveProjectSettingsAsync(projectSettings, cancellationToken);
+            }
+            else
```
orionstudt, line 357: Is there any reason that in the `else` scenario we don't want to do a more robust comparison? I mean you could write an equality comparer for the `ProjectSettings` class and you could straight up check if the `providedSettings != existingSettings` and only throw if it doesn't.

orionstudt: lol ignore me. I need to scroll more before commenting. I see you already wrote a comparer. We could use it here though

sdk/dotnet/Pulumi.Automation/DictionaryContentsComparer.cs
```diff
@@ -0,0 +1,54 @@
+// Copyright 2016-2019, Pulumi Corporation
+
+using System.Collections.Generic;
+
+namespace Pulumi.Automation
+{
+    /// Compares two dictionaries for equality by content, as F# maps would.
+    internal sealed class DictionaryContentsComparer<K, V> : IEqualityComparer<IDictionary<K, V>> where K : notnull
```
orionstudt, line 8: Style nitpick - I don't think this should be in the root directory of the project if it isn't public / exported

sdk/dotnet/Pulumi.Automation/DictionaryContentsComparer.cs
```diff
@@ -0,0 +1,54 @@
...
+                return y == null;
+            }
+            if (y == null)
+            {
+                return x == null;
+            }
+            if (x.Count != y.Count)
+            {
+                return false;
+            }
```
orionstudt, line 32: After you have determined that they are both not null you can do:
```csharp
if (ReferenceEquals(x, y))
    return true;
```

Which will ensure they're not the same ref, since it is a reference type. And then instead of a foreach loop you can knock it out with LINQ / set-comparison:
```csharp
// first ensure keys are the same
if (!x.Keys.ToHashSet().SetEquals(y.Keys))
    return false;
    
// then ensure values are the same
if (x.All(xkvp => this._valueComparer.Equals(xkvp.Value, y[xkvp.Key])))
    return true;
    
return false;

File: sdk/dotnet/Pulumi.Automation/Exceptions/ProjectSettingsConflictException.cs [hunk 16]
```diff
@@ -0,0 +1,39 @@
...
+    /// (A) to use the ProjectSettings, delete the Pulumi.yaml file
+    ///     from WorkDir or use a different WorkDir
+    ///
+    /// (B) to use the exiting Pulumi.yaml from WorkDir, avoid
+    ///     customizing the ProjectSettings
+    ///
+    /// </summary>
+    public class ProjectSettingsConflictException : Exception
+    {
+        internal string SettingsFileLocation { get; }
```
orionstudt, line 24: I would make this public, no reason for anything to be `internal` other than the constructor I think. If a consumer catches it I don't think there is a reason for us to hide this path.

sdk/dotnet/Pulumi.Automation/Exceptions/ProjectSettingsConflictException.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        internal ProjectSettingsConflictException(string settingsFileLocation)
+            : base(ErrMsg(settingsFileLocation))
```
orionstudt, line 27: You can do the string interpolation right here instead of having a single-use method:
```csharp
internal ProjectSettingsConflictException(string settingsFileLocation)
    : base($"Custom {nameof(ProjectSettings)} passed in code conflict with settings found on disk: {settingsFileLocation}")
{
    SettingsFileLocation = settingsFileLocation;
}
```

I like using the `nameof(...)` whenever I can when referring to a class in the library, which helps to enforce that string literals get updated whenever a class name changes. I try to do the same thing in xml comments using the `<see ...>` tag whenever referring to a class.

sdk/dotnet/Pulumi.Automation/ProjectBackend.cs
```diff
@@ -8,5 +11,30 @@ namespace Pulumi.Automation
     public class ProjectBackend
     {
         public string? Url { get; set; }
+
+        internal static IEqualityComparer<ProjectBackend> Comparer { get; } = new ProjectBackendComparer();
```
orionstudt, line 15: style nitpick - Putting private class definitions at the bottom is good practice I think, but I think `static/const` properties should be defined at the top of a class, before any instance properties

sdk/dotnet/Pulumi.Automation/DictionaryContentsComparer.cs
orionstudt: I guess you're using a comparer for the keys too so the key comparison would need that comparer in the hashset. So like:
```csharp
var xkeys = x.Keys.ToHashSet(this._keyComparer);
var yKeys = y.Keys.ToHashSet(this._keyComparer);
if (!xKeys.SetEquals(yKeys))
    return false;
```

sdk/dotnet/Pulumi.Automation/ProjectSettings.cs
```diff
@@ -44,5 +47,57 @@ public class ProjectSettings
...
+                {
+                    return y == null;
+                }
+
+                if (y == null)
+                {
+                    return x == null;
+                }
+
+                return x.Name == y.Name &&
```
orionstudt, line 75: I'm wondering if for string comparison we may want to use `string.Equals(...)` and decide on a `StringComparison` enum value that works for our use-case so we support culture-specific string comparison

sdk/dotnet/Pulumi.Automation/DictionaryContentsComparer.cs
t0yv0: The suggested replacement may not use _keyComparer when deciding SetEquals or when looking things up from y[xkvp.Key]. It will use KeyComparer from x and y. This is subtle but potentially very annoying. I'll leave the original code I think.. there are minor stylistic gains to be had by changing it but at least I'm confident in it. It would be tempting if there was a library method already to do this all thing though..

orionstudt: I do think that at the very least you want the `ReferenceEquals(...)` check in all of our comparers for reference types.

t0yv0: Yeah that suggestion is good, adding it.

sdk/dotnet/Pulumi.Automation/ProjectSettings.cs
t0yv0: I'd pass on that for the moment since it's hard to know how deep this should go.

sdk/dotnet/Pulumi.Automation/Exceptions/ProjectSettingsConflictException.cs
t0yv0: I think that's what we have now! Thank you.

sdk/dotnet/Pulumi.Automation/ProjectSettings.cs
orionstudt: Sounds good. Something to keep in mind, just because we're comparing strings from a consumer-controlled file so it might not always be US english or whatever

PR #6671, opened 2021-03-31
dotnet concurrent global state

t0yv0:
This is essentially the PR from @orionstudt merged with master, to pass acceptance tests and squash-merge.

https://github.com/pulumi/pulumi/pull/6139


orionstudt:
The current error in the MacOS test here https://github.com/pulumi/pulumi/pull/6671/checks?check_run_id=2239815670 is caused by the `ResourcePackages` class no being strict enough when looking for Pulumi resource assemblies, and finding a non-Pulumi assembly that cannot be loaded at runtime.

t0yv0:
That is a bit strange, and it only started failing on this change-set? Or the test is unrelated but flaky. I'm leaving this unmerged to have a close look tomorrow.

orionstudt:
@t0yv0 agreed it is strange. It seems unrelated but flaky. If you examine the commit history of `ResourcePackages.cs` you'll see I think that I added the `assembly.ContentType != AssemblyContentType.WindowsRuntime;` clause - that was in an attempt to resolve this same issue I was having when testing locally while writing the Automation API implementation. And I thought that had solved it.

Castle Proxies is an assembly that generates runtime types for proxy purposes so there must be an assembly that depends on that somewhere downstream - I believe it passes the `Assembly.Load(...)` call and the `assembly.GetTypes()` call is what finally throws it. Maybe we could benefit from switching to `assembly.GetExportedTypes()`, but idk if that would solve this specific issue.

It's weird that it is flaky though - because it's not like `Application.CurrentDomain` ever changes, right? I think you can also checkout the commits here because I refactored `ResourcePackages` at some point to no longer be static as part of the concurrency support change and I ended up reverting it after discussion with Mikhail. You can read that [here](https://github.com/pulumi/pulumi/pull/6139#discussion_r561073236).

orionstudt:
We could straight-up do the `!assemblyName.StartsWith("Castle.Whatever")` clause like we do with "System" assemblies but that seems like a nasty band-aid. Our best solution I think would be if we could verify that whatever assembly we are trying to load does in fact reference `Pulumi` or have a dependency that references `Pulumi`, recursively - before loading it.

t0yv0:
It seems that https://github.com/pulumi/pulumi/pull/6688 helps get this to green (assuming this is deterministic and I'm not just getting lucky). Let's see if I can get that merged first.

orionstudt:
That looks like exactly what I had in mind, nice.

t0yv0:
Fascinating looks like macOS issue is persisting.

orionstudt:
@t0yv0 very odd. Is it consistent now or still flaky? Weird that it is still platform-specific. Wonder if we can figure out where in our dependency tree castle proxies is

t0yv0:
Eh, I can try rerunning. Interestingly enough I can't repro locally.

t0yv0:
I think your changes introduce the bug because they enable tests to run in parallel, this mixes up the order of operation and uncovers the race.

My fix is not helping because DynamicProxyGenAssembly2 passes the MayContainResourceTypes test.

orionstudt:
Ok - that makes sense. So what is generating a runtime assembly, and why? I didn't realize we were doing that.

t0yv0:
Moq does, it seems.

orionstudt:
Oh! That makes a lot of sense. So why the heck are we only throwing on MacOS?

t0yv0:
Well it's a race right? It might occasionally throw.

```
IsDynamic: True
Ref System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e
Ref Pulumi, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null
Ref Moq, Version=4.13.0.0, Culture=neutral, PublicKeyToken=69f491c39445e920
Ref Castle.Core, Version=4.0.0.0, Culture=neutral, PublicKeyToken=407dd0808d44fbdc
Ref System.Reflection, Version=4.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a
Ref System.Runtime, Version=4.2.2.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a
Ref System.Runtime, Version=4.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a
```

I wonder how to discriminate against assemblies like this one. Perhaps we can just downgrade exception to a warning if GetTypes() fails on a dynamic assembly.

orionstudt:
Does our search need to support dynamic assemblies?

t0yv0:
Or we could exclude assemblies that are dynamic and reference Moq and Castle. It's a strange rule ..

It probably doesn't but a day will come when someone tries something legit with Moq and will be surprised.

t0yv0:
Finally we could consider some hook to exclude assemblies, and make just our test suite opt into this hook somehow.

t0yv0:
I'm calling it a night, TBC next week.

orionstudt:
I think `ResourcePackages` will need to be more robust. For instance, our default assembly discovery could exclude dynamic assemblies. But in the future we could allow a collection of types to be passed in, in lieu of a collection of a assembly - skirting that issue and allowing runtime types.

I mentioned in that other commit that `ResourcePackages` would need to be refactored to allow the consumer to pass in assemblies, so that multi-language resources can be supported in scenarios where the necessary assembly isn't part of the `CurrentDomain` - such as when it was loaded at runtime via a plugin.

t0yv0:
Yes I was discussing the same with @pgavlin internally today. We'll log something into the backlog. For now I wonder if it is indeed acceptable just to skip dynamic assemblies. The current mechanism is not robust over those anyway since they can be added after AppDomain init and therefore always subject to race condition wrt the ResourcePackages scan.

orionstudt:
So short-term we just want to figure out if we can exclude dynamic assemblies, I guess.

t0yv0:
Sorry for delay here, picking this up. Let's try to exclude dynamic assemblies to get some confidence it actually fixes the test suite, then make the call.

orionstudt:
Sounds good. Would you like me to bring the original PR up to speed? Could do that quickly.

t0yv0:
Sure, either way we can do that or I can just merge this one.

Looks like we have an unrelated problem. The  current version failed on some lock conflict. I have just glanced at it does anything ring a bell?

Also I am thinking rather then fixing it we have the option of reverting the change that makes the tests run in parallel. At least for now. Unfortunate but an option.

Issue #6617, opened 2021-03-25
[python, go, dotnet] Make custom available to Mocks::newResource


lukehoban:
In https://github.com/pulumi/pulumi/pull/6551 we added support for a `custom` argument to `newResource` mocks for Node.js.  We need to do this in the other languages as well.

Per discussion on that PR, we will likely address this with a breaking change to accept a property bag instead of discreet arguments, to ensure we do not have to make further breaking changes here in the future.

stack72:
Closed via #6672

PR #6672, opened 2021-03-31
Refactor Mock newResource and call to accept property bag rather than individual args

stack72:
Fixes: #6617

@justinvp / @komalali / @lukehoban tagging you fine folks on this PR - this is the follow up to the work of https://github.com/pulumi/pulumi/pull/6551 where we want to allow extensibility in the future without any need to break things

### DIFF
--- a/sdk/nodejs/runtime/mocks.ts
+++ b/sdk/nodejs/runtime/mocks.ts
@@ -20,6 +20,61 @@ const resproto = require("../proto/resource_pb.js");
 const structproto = require("google-protobuf/google/protobuf/struct_pb.js");
 
 /**
+ * MockResourceArgs is a used to construct a newResource Mock
+ */
+export interface MockResourceArgs {
+    /**
+     * The token that indicates which resource type is being constructed. This token is of the form "package:module:type".
+     */
+    type: string;
+
+    /**
+     * The logical name of the resource instance.
+     */
+    name: string;
+
+    /**
+     * The inputs for the resource.
+     */
+    inputs: any;
+
+    /**
+     * If provided, the identifier of the provider instance being used to manage this resource.
+     */
+    provider?: string;
+
+    /**
+     * Specifies whether or not the resource is Custom (i.e. managed by a resource provider).
+     */
+    custom?: boolean;
+
+    /**
+     * If provided, the physical identifier of an existing resource to read or import.
+     */
+    id?: string;
+}
+
+/**
+ * MockResourceArgs is used to construct call Mock
+ */
+export interface MockCallArgs {
+    /**
+     * The token that indicates which function is being called. This token is of the form "package:module:function".
+     */
+    token: string;
+
+    /**
+     * The arguments provided to the function call.
+     */
+    inputs: any;
+
+    /**
+     * If provided, the identifier of the provider instance being used to make the call.
+     */
+    provider?: string;
+}
+
+/**
  * Mocks is an abstract class that allows subclasses to replace operations normally implemented by the Pulumi engine with
  * their own implementations. This can be used during testing to ensure that calls to provider functions and resource constructors
  * return predictable values.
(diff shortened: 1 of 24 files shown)


stack72:
Need to do some more work for this to be ready for merging!

sdk/nodejs/runtime/mocks.ts
```diff
@@ -114,13 +167,14 @@ export class MockMonitor {
...
-                req.getType(),
-                req.getName(),
-                deserializeProperties(req.getObject()),
-                req.getProvider(),
-                req.getImportid(),
-                req.getCustom());
+            const result = this.mocks.newResource({
+                type: req.getType(),
+                name: req.getName(),
+                inputs: deserializeProperties(req.getProperties()),
```
EvanBoyle, line 173: Shoudl this be `deserializeProperties(req.getObject()),`? Does `req.getProperties` exist for register resource?

EvanBoyle: There might be a similar issue with `getId` vs `getImportid`, but not sure.

justinvp (commented):
Have we considered doing this in a non-breaking way? Keep the existing interface as-is, introduce a new interface, and support passing both the old and new interfaces when setting mocks? Existing user code then wouldn't need to change, and we wouldn't have to get this in for 3.0.

The existing interface name is `Mocks`/`IMocks`, I could imagine coming up with a similarly decent enough name for the new interface (e.g. `TestMocks`/`ITestMocks`).

For TS/JS, we could make no change and continue to add additional parameters to the function (like we did for `custom`). Or introduce the new interface with differing function names. Or introduce an [abstract class](https://www.typescriptlang.org/docs/handbook/2/classes.html#abstract-classes-and-members) (instead of an interface) that we could runtime type-check for when the mock is set.

stack72:
> Have we considered doing this in a non-breaking way? Keep the existing interface as-is, introduce a new interface, and support passing both the old and new interfaces when setting mocks? Existing user code then wouldn't need to change, and we wouldn't have to get this in for 3.0.
> 
> The existing interface name is `Mocks`/`IMocks`, I could imagine coming up with a similarly decent enough name for the new interface (e.g. `TestMocks`/`ITestMocks`).
> 
> For TS/JS, we could make no change and continue to add additional parameters to the function (like we did for `custom`). Or introduce the new interface with differing function names. Or introduce an [abstract class](https://www.typescriptlang.org/docs/handbook/2/classes.html#abstract-classes-and-members) (instead of an interface) that we could runtime type-check for when the mock is set.

That approach would work fine for TS but it would be much more messy for Go - this new way of thinking uses the 3.0 approach to make the break so that we have a consistent feel to the API here across all of the language SDKS

justinvp:
> it would be much more messy for Go

Is it much more messy? This way, users don't have to modify their existing code, and we could decide to mark the existing interface and function as deprecated (and potentially remove it in a future release after giving users fair warning).

```go
// Existing (keep)...

type MockResourceMonitor interface {
	Call(token string, args resource.PropertyMap, provider string) (resource.PropertyMap, error)
	NewResource(typeToken, name string, inputs resource.PropertyMap,
		provider, id string) (string, resource.PropertyMap, error)
}

func WithMocks(project, stack string, mocks MockResourceMonitor) RunOption {
	return func(info *RunInfo) {
		info.Project, info.Stack, info.Mocks = project, stack, mocks
	}
}

// New...

type TestMockResourceMonitor interface {
	Call(args MockCallArgs) (resource.PropertyMap, error)
	NewResource(args MockResourceArgs) (string, resource.PropertyMap, error)
}

func WithTestMocks(project, stack string, mocks TestMockResourceMonitor) RunOption {
	return func(info *RunInfo) {
		info.Project, info.Stack, info.TestMocks = project, stack, mocks
	}
}
```

sdk/nodejs/tests_with_mocks/mocks.spec.ts
```diff
@@ -37,12 +38,12 @@ pulumi.runtime.setMocks({
                     publicDns: "ec2-203-0-113-12.compute-1.amazonaws.com",
                     publicIP: "203.0.113.12",
                 };
-                return { id: "i-1234567890abcdef0", state: { ...inputs, ...state } };
+                return { id: "i-1234567890abcdef0", state: { ...args.inputs, ...state } };
             case "pkg:index:MyCustom":
-                assert.strictEqual(custom, true);
-                return { id: name + "_id", state: inputs };
+                assert.strictEqual(args.custom, true);
+                return { id: name + "_id", state: args.inputs };
```
komalali, line 44: This should fix the promise leak
```suggestion
                return { id: args.name + "_id", state: args.inputs };
```

stack72: 🥇

komalali (approved):

komalali (commented):
Hmm looks like the python tests are failing from changes here.

stack72:
> Hmm looks like the python tests are failing from changes here.

Yeah I will fix these up - this is an issue where Pulumi.runtime.Mocks is being used in another place but I didn’t update that in the tests

will get it fixed!

stack72:
Ok, I spoke to @leezen about this - we are going to proceed with the planned approach in this PR - we have the opportunity to make a breaking change - we are going to take this opportunity here. We could indeed have gone with the suggestion above regarding keeping it non-breaking and introduce a new interface and deprecate the old one but we decided lets make the clean break here and not manage 2 code paths in the future

adriantorrie:
Just leaving this here incase anyone else needs pointing to the new mocks for python tests if your CI/CD unit testing has failed.

https://www.pulumi.com/docs/guides/testing/unit/#add-mocks

**Old** way:

```python
class Mock(pulumi.runtime.Mocks):

    def call(self, token, args, provider):
        {}

    def new_resource(self, type_, name, inputs, provider, id_):
        # Generic return if no mocks-by-type detected
        return (name, inputs)
```

**New** way:

```python
class Mock(pulumi.runtime.Mocks):

    def call(self, pulumi.runtime.MockCallArgs):
        {}

    def new_resource(self, pulumi.runtime.MockCallArgs):
        # Then use `args.<arg_name>` when referencing, e,g.
        return [args.name + '_id', args.inputs]
```

PR #6673, opened 2021-03-31
Add program side dynamic provider caching behind env var

EvanBoyle:
An outcome of investigating https://github.com/pulumi/pulumi/issues/6645
Doing this behind an env var as an outcome of discussion in https://github.com/pulumi/pulumi/pull/6646 as this is technically a breaking change. 

We repeatedly serialize dynamic provider code for every instance of a dynamic resource, even if it shares the same dynamic provider with other resources. This change adds caching behind an environment variable that users can opt into. It stores the result of serialization in a cache keyed by provider object equality. In the test case outline in #6645 this change reduces runtime by 80% (120s -> 20s) and memory usage by a 75% (1.5gb -> 400mb). Users with programs that just use a few hundred dynamic resources have also confirmed drastic perf improvements. 

This is behind an env var, `PULUMI_NODEJS_CACHE_DYNAMIC_PROVIDERS`, as providers can technically capture global variables, swap out functions, and generally make changes that won't be detected by the equality check. I'll make a PR to the 3.0 branch to change the default value of this env var from false to true once this is merged.


sdk/nodejs/runtime/settings.ts
```diff
@@ -63,6 +64,7 @@ const nodeEnvKeys = {
     monitorAddr: "PULUMI_NODEJS_MONITOR",
     engineAddr: "PULUMI_NODEJS_ENGINE",
     syncDir: "PULUMI_NODEJS_SYNC",
+    cacheDynamicProviders: "PULUMI_NODEJS_CACHE_DYNAMIC_PROVIDERS",
```
lukehoban, line 67: I think the rest of the things in the list are automatically passed from the CLI.  This one seems different, as I don't think we expect the CLI to pass it.  Not sure that "matters" - but maybe at least a comment that this is different than the others and that users must set it themselves if they want to enable this?

sdk/nodejs/dynamic/index.ts
```diff
@@ -161,8 +161,22 @@ export interface ResourceProvider {
     delete?: (id: resource.ID, props: any) => Promise<void>;
 }
 
+const providerCache = new Map<ResourceProvider, Promise<string>>();
```
lukehoban, line 164: Total nit - as I don't think it matters at all in practice - but this may be a case where [`WeakMap`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap) is actually appropriate.

lukehoban (approved):

sdk/nodejs/runtime/settings.ts
EvanBoyle: I don't think it matters, other keys such as `pulumiEnvs` below are controlled via individual environment variables. I'll add a comment to clarify

sdk/nodejs/dynamic/index.ts
EvanBoyle: Ah, TIL. Will update.

Issue #6499, opened 2021-03-10
LocalWorkspaceOptions.EnvironmentVariables should allow null values


MisinformedDNA:
I'm seeing nullability warnings since `LocalWorkspaceOptions.EnvironmentVariables` is of type `IDictionary<string, string>` and not `IDictionary<string, string?>`.

I would expect it to be the latter since it maintains consistency with CLR types:

[Environment.SetEnvironementVariable](https://docs.microsoft.com/en-us/dotnet/api/system.environment.setenvironmentvariable?view=net-5.0#System_Environment_SetEnvironmentVariable_System_String_System_String_) accepts a nullable string as its value and [Environment.GetEnvironmentVariable](https://docs.microsoft.com/en-us/dotnet/api/system.environment.getenvironmentvariable?view=net-5.0#System_Environment_GetEnvironmentVariable_System_String_) returns a nullable string.

## Steps to reproduce

```csharp
programArgs.EnvironmentVariables = new Dictionary<string, string>
{
    // possible null reference assignment
    ["ARM_CLIENT_ID"] = Environment.GetEnvironmentVariable("AzureServicePrincipalId"),
};
```

If you agree, I can file a PR.

## Affected feature

 area/automation-api

justinvp:
Sounds reasonable.

> If you agree, I can file a PR.

That'd be great! Thanks!

mikhailshilkov:
@MisinformedDNA We had to revert #6520 because it conflicted with #6654 that was merged several hours before it. In particular, `CliWrap` does not allow nullable environment values [here](https://github.com/pulumi/pulumi/blob/7f8985ab0fb870f0cd577f474255b348d6ebeefe/sdk/dotnet/Pulumi.Automation/Commands/LocalPulumiCmd.cs#L75). And this is likely because it [assigns](https://github.com/Tyrrrz/CliWrap/blob/0938c4dfb5d829181e387067ec62e74a2a72231f/CliWrap/Command.cs#L287-L288) those values to `System.Diagnostics.ProcessStartInfo.Environment` of type `IDictionary<string, string>`. We'd need to reconcile those somehow.

cc @t0yv0 who is the author of the CliWrap-based implementation.

t0yv0:
Ouch! What is the desired meaning of null environment variable when staring as sub-process? AFAIK the OS does not admit nulls so we have a choice to either omit or set to default empty "". I can add a patch if we know what we want to do here. CC @MisinformedDNA

[...]

PR #6680, opened 2021-04-01
Allow null env vars and treat null/empty as deletions

t0yv0:
Continuing with https://github.com/pulumi/pulumi/issues/6499 which is the reverted work of @orionstudt reconciled with CliWrap, plus we are adding a new meaning to empty/null strings as "deleting" environment variables, similar to how dotnet BCL does it.

### DIFF
--- a/sdk/dotnet/Pulumi.Automation/Commands/LocalPulumiCmd.cs
+++ b/sdk/dotnet/Pulumi.Automation/Commands/LocalPulumiCmd.cs
@@ -22,7 +22,7 @@ namespace Pulumi.Automation.Commands
         public async Task<CommandResult> RunAsync(
             IEnumerable<string> args,
             string workingDir,
-            IDictionary<string, string> additionalEnv,
+            IDictionary<string, string?> additionalEnv,
             Action<string>? onStandardOutput = null,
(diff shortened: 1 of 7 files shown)


sdk/dotnet/Pulumi.Automation/Commands/LocalPulumiCmd.cs
```diff
@@ -107,7 +107,20 @@ internal class LocalPulumiCmd : IPulumiCmd
...
             foreach (var pair in additionalEnv)
             {
-                env[pair.Key] = pair.Value;
+                // For consistency with System.SetEnvironmentVaraible,
+                // we treat null and empty strings as instructions to
+                // delete the env var from the sub-process
+                // environment.
+                //
+                // https://docs.microsoft.com/en-us/dotnet/api/system.environment.setenvironmentvariable
+                if (String.IsNullOrEmpty(pair.Value))
```
vipentti, line 116:
```suggestion
                if (string.IsNullOrEmpty(pair.Value))
```

t0yv0:
Something is not right.. still testing. The CliWrap library does not do what I think it does.

PR #6682, opened 2021-04-01
Simplify env var propagation

t0yv0:
This change should not result in visible behavior difference but simplifies the code. It turns out CliWrap treats the envVar arg as modifications on top of existing system environment, which it propagates already. We do not need to do it therefore.


t0yv0:
@vipentti @orionstudt

orionstudt:
Oh well that is good to know. I opened #6683 but if CliWrap handles this I will close.

vipentti:
Looks good to me!

komalali (approved):

Issue #5758, opened 2020-11-13
Functions/datasources don't compose well with Pulumi model - should accept Outputs


lukehoban:
One of the most common usability issues I've seen new users hit with Pulumi is that datasources/functions take and return raw values instead of Inputs/Outputs, meaning they are very difficult to compose with the rest of the Pulumi resource model.  We encourage folks to get used to writing code like this in the normal Pulumi usage:

```ts
const a = new ResourceA("name");
const b = new ResourceB("name", { prop: a.id }); // pass an `Output` as an `Input`
```

But then when you want to do the same with a Pulumi function - it doesn't work.

```ts
const a = new ResourceA("name");
const b = someFunction({ prop: a.id }); // cannot pass `Output` where `string` is expected
```

Instead - you have to write something like this:

```ts
const a = new ResourceA("name");
const b = a.id.apply(v => someFunction({ prop: v }));
```

This is not obvious, and not pleasant.  

The original motivation was to provide the additional flexibility of being able to use these functions in cases where you don't need to get caught up in `Outputs`.  And it may be we feel we have to continue to support those use cases.

A non-breaking way to improve this would be to project two copies of every function `somFunc` and `someFuncOutput` for example - where the latter accepts Inputs and returns Outputs.  

----

### M64
- [x] #7946 (In review)
- [x] #7948 (In progress)
   - [x]  pulumi/examples#1093
   - [x]  pulumi/examples#1094
   - [x]  pulumi/examples#1095
   - [x]  pulumi/examples#1096
- [x] #7949 (In review)
- [x] #7947
- [x] https://github.com/pulumi/pulumi/issues/8172
- [x] https://github.com/pulumi/pulumi/issues/8322 (released in 3.17.1)

### M63
- [x] #7945 
PR: https://github.com/pulumi/pulumi/pull/7899

### M62
- [x] #7944
PR: https://github.com/pulumi/pulumi/pull/8047
- [x] https://github.com/pulumi/pulumi/pull/7989
- [x] https://github.com/pulumi/pulumi/pull/8039

### M61
- [x] #7784
- [x] #7825
- [x] https://github.com/pulumi/pulumi/pull/7934

### Future Iterations
- [ ] https://github.com/pulumi/examples/issues/1115 (blocked)
- [ ] https://github.com/pulumi/pulumi-azure-native/issues/1279

justinvp:
[...]

PR #6686, opened 2021-04-01
Fix inconsistent object type naming.

### LINKED ISSUE #6200: [codegen] Inconsistent code generated for schema types that are used both for resources and invokes
### Context

A schema defines:

- Types `A` and `B`
- Resource `R`
- Invoke `I`
- Both `R` and `I` use the type `A` in to define an input property and type `B` to define an output property

### Current behavior

Problem 1: .NET, Node.js, and Python SDKs (not sure about Go) would only generate a single `AArgs` type with every property being an `Input<T>` which doesn't match the common shape of invoke's input types (with no `Input`'s)

Problem 2: .NET and Python SDKs (Node.js is fine, not sure about Go) would name the output type `B` or `BResult` depending on whether it's used for both resource and invokes, or just for invokes.

Both problems lead to breaking changes if a new resource is introduced that points to a type currently used by an invoke (version `X` has `A`, `B`, and `I` and version `X+1` introduces `R`) . Examples in https://github.com/pulumi/pulumi-azure-nextgen-provider/pull/455/

pgavlin:
See #6200 for a complete description of the issue. In short, we generate
inconsistent names for object types depending on whether or not they are
transitively reachable from resources or functions, which risks
unintentional breaking changes due to schema updates.

1. Name "input" types differently: `TArgs` for a type that is used in
   resource inputs, having `Input<T>` properties, and `T` for a type
   that is used in invoke inputs. The same schema type can produce both.

2. Always keep the name `T` for output types, avoid appending `Result` to
   the name.

3. As needed, introduce a flag in the existing providers' schemas to avoid
   breaking changes. Consider removing it on a major version bump.

Fixes #6200.

### DIFF
--- a/pkg/codegen/nodejs/gen.go
+++ b/pkg/codegen/nodejs/gen.go
@@ -40,6 +40,7 @@ import (
 type typeDetails struct {
 	outputType   bool
 	inputType    bool
+	argsType     bool
 	functionType bool
 }
 
@@ -137,7 +138,7 @@ func (mod *modContext) namingContext(pkg *schema.Package) (namingCtx *modContext
 	return
 }
 
-func (mod *modContext) objectType(pkg *schema.Package, tok string, input, enum bool) string {
+func (mod *modContext) objectType(pkg *schema.Package, tok string, input, args, enum bool) string {
 	root := "outputs."
 	if input {
 		root = "inputs."
@@ -155,6 +156,10 @@ func (mod *modContext) objectType(pkg *schema.Package, tok string, input, enum b
 
 	if enum {
 		return "enums." + modName + title(name)
+	}
+
+	if args && mod.compatibility != tfbridge20 {
+		name += "Args"
 	}
 	return pkgName + root + modName + title(name)
 }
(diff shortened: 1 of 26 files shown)


pgavlin:
Note that the test failures are expected, as they demonstrate the effects of this change w/o updating downstream compat flags. I'll fix up the tests before committing.

komalali:
@pgavlin curious if there's any interplay between this fix and the issue described in https://github.com/pulumi/pulumi/issues/5758

pkg/codegen/nodejs/gen.go
```diff
@@ -845,7 +845,13 @@ func (mod *modContext) genType(w io.Writer, obj *schema.ObjectType, input bool,
 	}
 
 	wrapInput := input && !mod.details(obj).functionType
-	mod.genPlainType(w, tokenToName(obj.Token), obj.Comment, properties, input, wrapInput, false, level)
+
+	name := tokenToName(obj.Token)
+	if wrapInput && mod.compatibility != tfbridge20 {
+		name += "Args"
```
mikhailshilkov, line 851: I ran this on Azure-Native - it looks like this changes the type name from `Foo` to `FooArgs`, but the parent type still tries referencing `Foo`. Example of an error:

```
web/webAppSourceControlSlot.ts:148:66 - error TS2694: Namespace '"/Users/mikhailshilkov/go/src/github.com/pulumi/pulumi-azure-native/sdk/nodejs/types/input".web' has no exported member 'GitHubActionConfiguration'.

148     readonly gitHubActionConfiguration?: pulumi.Input<inputs.web.GitHubActionConfiguration>;
```

mikhailshilkov:
Azure-Native preview is here: https://github.com/pulumi/pulumi-azure-native/pull/716

pkg/codegen/nodejs/gen.go
mikhailshilkov: See https://github.com/pulumi/pulumi-azure-native/actions/runs/711650762

pgavlin:
@stack72 something seems to be wrong with the diffs. If I run these locally, I see plenty of diffs, but the jobs that are posting to the PR show no diffs. Any ideas?

stack72:
> @stack72 something seems to be wrong with the diffs. If I run these locally, I see plenty of diffs, but the jobs that are posting to the PR show no diffs. Any ideas?

@pgavlin so the issue is that pulumi@master uses v3 of the pkg & sdk modules and the downstream repos use v2 of the pkg & sdk modules so those diffs won't work until we merge feature-3.0 to master in those downstream repos

that will happen today!

pgavlin:
Okay, I believe that this is now good to go: I've verified that the combat flag works for `pulumi-aws`, and that the new codegen works for `pulumi-azure-ng`. 

After we merge this, we will need to update `pulumi-terraform-bridge` to pass the combat flag prior to rollout.

mikhailshilkov (approved):
LGTM. I admit it's hard to validate the changes in my head, so I rely on the generation results. Azure changes are looking good and we have a flag to prevent any changes in other providers, so it feels like this is ready to merge.

pgavlin:
Pretty certain that the downstream docs tests are only failing due to module differences.

stack72:
> Pretty certain that the downstream docs tests are only failing due to module differences.

Yeah resource docs will fail in PR gen until feature-3.0 goes to master tomorrow on each of these repos :/

PR #6688, opened 2021-04-01
Shortcut resource packages scan

t0yv0:
It seems reasonable that if an assembly does not reference Pulumi.dll that defines Resource type, then it cannot contain types that inherit from Resource, and therefore we can skip GetTypes() for that assembly. This saves some System.Type traversals (4000 -> 200 in our test). More importantly it seems to help with an issue @orionstudt had in their PR where this code blows up on trying to GetType() from Castle dynamic assembly.

### DIFF
--- a/sdk/dotnet/Pulumi/Serialization/ResourcePackages.cs
+++ b/sdk/dotnet/Pulumi/Serialization/ResourcePackages.cs
@@ -14,7 +14,7 @@ namespace Pulumi
     {
         private static ImmutableDictionary<string, ImmutableList<(string?, Type)>>? _resourceTypes;
         private static readonly object _resourceTypesLock = new object();
-        
+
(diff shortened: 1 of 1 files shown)


mikhailshilkov (approved):
LGTM overall. Do we have enough coverage to be sure we don't miss some edge cases?

t0yv0:
I'm wondering about that also. The happy path is definitely covered. I failed to get a big picture though, how do users use this thing?

orionstudt:
I believe the `ResourcePackages` class is part of bringing this epic to .NET: https://github.com/pulumi/pulumi/issues/2430

t0yv0:
Something funky with Mac OS tests timing out. I'm going to rerun them. There were some GitHub Actions availability issues yesterday.

justinvp (approved):

Issue #3151, opened 2019-08-27
Issues with snake_casing in Python


lukehoban:
There are several intersecting issues with the current approach to automatic snake_casing for Python SDKs, which in total lead to:
1. Us not implementing what we're trying to do correctly
2. What we are trying to do being fundamentally "wrong"
3. Significant inconsistencies within and among providers

Our code for name translation in the Python SDK intends to implement the following: "For inputs, all properties, recursively, can be passed in either snake_case or camelCase.  For outputs, all properties, recursively, are snake_cased."

The act of "snake_casing" is done not based on an algorithm but based on a fixed table provided by each provider.  This table is "correctly" generated for Kubernetes, but is incorrectly (relative the the intention above) generated for TFGen.  In particular, for TFGen, only top level property names are added to the fixed table (along with keys of resources nested in `TypeMap`s which nearly never occur in Terraform - there are 5 cases total of this across AWS, Azure and GCP).

In practice, this means that for TFGen providers, the "actual" experience is effectively: "For inputs, top-level property names (projected as Python named constructor arguments) can be passed in either snake_case or camelCase, nested properties (projected as dicts) must be passed camelCase.  For outputs, top-level property names (projected as Python class properties) are available snake_cased,  and nested properties (projected as dicts) are available camelCased."  The Kubernetes provider implements the intention as described earlier.

The "intended design" is fundamentally broken though.  If implemented correctly, it causes user-defined dictionary keys passed to inputs like `Tags` to be transformed.  As it so happens, because we mostly don't implement it correctly for TFGen, we've avoided issues related to this.  (also because snake_case tag keys are likely relatively less common).

We believe that the "right" thing is actually this: "For inputs, top-level property names (projected as Python named constructor arguments) must be passed in snake_case, nested properties (projected as dicts) must be passed camelCase.  For outputs, top-level property names (projected as Python class properties) are available snake_cased,  and nested properties (projected as dicts) are available camelCased."  This matches a reasonable Python intuition that Python property and argument names are snake_case, but dict keys passed with quoted strings are not.

Properly "fixing" this would be a significant breaking change.  Even in our own code samples we appear to have almost every combination of things the above set of rules allows for.

But we believe we should take the following actions ASAP to get into a reasonable state where it is at least possible to recommend a consistent approach here (and apply it across all our examples and documentation), without a breaking change.

Specifically:
[...]

PR #6695, opened 2021-04-02
[sdk/python] Improved dict key translation support

justinvp:
This change addresses Python dictionary key translation issues. When the
type of `props` passed to the resource is decorated with `@input_type`,
the type's and resource's property name metadata will be used for dict
key translations instead of the resource's `translate_input_property`
and `translate_output_property` methods.

The generated provider SDKs will be updated to opt-in to this new
behavior:

- FIX: Keys in user-defined dicts will no longer be unintentionally
  translated/modified.

- BREAKING: Dictionary keys in nested output classes are now
  consistently snake_case. If accessing camelCase keys from such output
  classes, move to accessing the values via the snake_case property
  getters (or snake_case keys). Generated SDKs will log a warning
  when accessing camelCase keys.

When serializing inputs:

  - If a value is a dict and the associated type is an input type, the
    dict's keys will be translated based on the input type's property
    name metadata.

  - If a value is a dict and the associated type is a dict (or Mapping),
    the dict's keys will _not_ be translated.

When resolving outputs:

  - If a value is a dict and the associated type is an output type, the
    dict's keys will be translated based on the output type's property
    name metadata.

  - If a value is a dict and the associated type is a dict (or Mapping),
    the dict's keys will _not_ be translated.

---

Example of opting-in to the new behavior:

```python
# A <Resource>Args class decorated with `@input_type` with type/name metadata for the resource's
# top-level inputs.
@pulumi.input_type
class FooArgs:
    def __init__(self, some_value: pulumi.Input[str]):
        pulumi.set(self, "some_value", some_value)

    @property
    @pulumi.getter(name="someValue")
    def some_value(self) -> pulumi.Input[str]:
        return pulumi.get(self, "some_value")

class Foo(pulumi.CustomResource):
    def __init__(__self__, resource_name: str, args: FooArgs, opts: pulumi.ResourceOptions):
        # An instance of <Resource>Args is passed for `props`. Since it is decorated with
        # `@input_type` this opts-in to using the input type and the resource's
        # type/name metadata for dict key translations, rather than calling
        # `translate_input_property` and `translate_output_property`.
        super().__init__(__self__, "example::Foo", args, opts)
```

Part of https://github.com/pulumi/pulumi/issues/3151
Codegen change: https://github.com/pulumi/pulumi/pull/6696

### DIFF
--- a/sdk/python/lib/pulumi/resource.py
+++ b/sdk/python/lib/pulumi/resource.py
@@ -26,6 +26,8 @@ from .runtime.settings import get_root_resource
 
 from .metadata import get_project, get_stack
 
+from . import _types
+
 if TYPE_CHECKING:
     from .output import Input, Inputs, Output
     from .runtime.stack import Stack
@@ -622,7 +624,10 @@ class Resource:
         :param str t: The type of this resource.
         :param str name: The name of this resource.
         :param bool custom: True if this resource is a custom resource.
-        :param Optional[dict] props: An optional list of input properties to use as inputs for the resource.
+        :param Optional[Inputs] props: An optional list of input properties to use as inputs for the resource.
+               If props is an input type (decorated with `@input_type`), dict keys will be translated using
+               the type's and resource's type/name metadata rather than using the `translate_input_property`
+               and `translate_output_property` methods.
         :param Optional[ResourceOptions] opts: Optional set of :class:`pulumi.ResourceOptions` to use for this
                resource.
         :param bool remote: True if this is a remote component resource.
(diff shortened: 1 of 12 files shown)


sdk/python/lib/pulumi/resource.py
```diff
@@ -26,6 +26,8 @@
 
 from .metadata import get_project, get_stack
 
+from . import _types
```
komalali, line 29: Why all the line breaks in the imports here?

sdk/python/lib/pulumi/runtime/rpc.py
```diff
@@ -208,23 +257,50 @@ async def serialize_property(value: 'Input[Any]',
...
+    # If value is an input type, convert it to a dict.
     value_cls = type(value)
     if _types.is_input_type(value_cls):
         value = _types.input_type_to_dict(value)
-        transform_keys = False
+        types = _types.input_type_types(value_cls)
+
+        return {
+            k: await serialize_property(v, deps, input_transformer, types.get(k))
+            for k, v in value.items()
```
komalali, line 268: Is this likely to run into issues if the input type has an `items` property?

I think no because `value` is already a `dict`, but just making sure.

komalali (approved):
LGTM, just a question about the use of `.items()` and whether we're certain we're not passing in any `InputType` or `OutputType`s there.

sdk/python/lib/pulumi/resource.py
justinvp: I added a commit that cleans this up.

sdk/python/lib/pulumi/runtime/rpc.py
justinvp: > I think no because `value` is already a `dict`, but just making sure.

Right. It's guaranteed to be a `dict` here, so we don't have to worry about that.

PR #6696, opened 2021-04-02
[codegen/python] Adopt improved key translation

justinvp:
This change updates the Python SDK codegen to opt-in to the new casing
translation behavior, which will use the passed-in props type's property
name metadata for translations, rather than calling the resource's
`translate_input_property` and `translate_output_property` methods.

- FIX: Keys in user-defined dicts will no longer be unintentionally
  translated/modified.

- BREAKING: Dictionary keys in nested output classes are now
  consistently snake_case. If accessing camelCase keys from such output
  classes, move to accessing the values via the snake_case property
  getters (or snake_case keys). A warning will be logged when accessing
  camelCase keys.

When serializing inputs:

- If a value is a dict and the associated type is an input type, the
dict's keys will be translated based on the input type's property
name metadata.

- If a value is a dict and the associated type is a dict (or Mapping),
the dict's keys will _not_ be translated.

When resolving outputs:

- If a value is a dict and the associated type is an output type, the
dict's keys will be translated based on the output type's property
name metadata.

- If a value is a dict and the associated type is a dict (or Mapping),
the dict's keys will _not_ be translated.

Note: There is some additional codegen/docgen code related to generating the casing tables that can be deleted, but it can be removed in a follow-up change.

Part of https://github.com/pulumi/pulumi/issues/3151
Depends on SDK change: https://github.com/pulumi/pulumi/pull/6695

### DIFF
--- a/pkg/codegen/internal/test/testdata/external-resource-schema/python/pulumi_example/arg_function.py
+++ b/pkg/codegen/internal/test/testdata/external-resource-schema/python/pulumi_example/arg_function.py
@@ -6,7 +6,7 @@ import warnings
 import pulumi
 import pulumi.runtime
 from typing import Any, Mapping, Optional, Sequence, Union, overload
-from . import _utilities, _tables
+from . import _utilities
 import pulumi_random
 
 __all__ = [
--- a/pkg/codegen/internal/test/testdata/external-resource-schema/python/pulumi_example/component.py
+++ b/pkg/codegen/internal/test/testdata/external-resource-schema/python/pulumi_example/component.py
@@ -6,7 +6,7 @@ import warnings
 import pulumi
 import pulumi.runtime
 from typing import Any, Mapping, Optional, Sequence, Union, overload
-from . import _utilities, _tables
+from . import _utilities
 import pulumi_aws
 import pulumi_kubernetes
 
@@ -88,12 +88,12 @@ class Component(pulumi.CustomResource):
         if opts.id is None:
             if __props__ is not None:
                 raise TypeError('__props__ is only valid when passed in combination with a valid opts.id to get an existing resource')
-            __props__ = dict()
+            __props__ = ComponentArgs.__new__(ComponentArgs)
 
-            __props__['metadata'] = metadata
-            __props__['provider'] = None
-            __props__['security_group'] = None
-            __props__['storage_classes'] = None
+            __props__.__dict__['metadata'] = metadata
+            __props__.__dict__['provider'] = None
+            __props__.__dict__['security_group'] = None
+            __props__.__dict__['storage_classes'] = None
         super(Component, __self__).__init__(
             'example::Component',
             resource_name,
@@ -114,11 +114,11 @@ class Component(pulumi.CustomResource):
         """
         opts = pulumi.ResourceOptions.merge(opts, pulumi.ResourceOptions(id=id))
 
-        __props__ = dict()
+        __props__ = ComponentArgs.__new__(ComponentArgs)
 
-        __props__["provider"] = None
-        __props__["security_group"] = None
-        __props__["storage_classes"] = None
+        __props__.__dict__['provider'] = None
+        __props__.__dict__['security_group'] = None
+        __props__.__dict__['storage_classes'] = None
         return Component(resource_name, opts=opts, __props__=__props__)
 
     @property
@@ -135,10 +135,4 @@ class Component(pulumi.CustomResource):
     @pulumi.getter(name="storageClasses")
     def storage_classes(self) -> pulumi.Output[Optional[Mapping[str, 'pulumi_kubernetes.storage.v1.StorageClass']]]:
         return pulumi.get(self, "storage_classes")
-
-    def translate_output_property(self, prop):
-        return _tables.CAMEL_TO_SNAKE_CASE_TABLE.get(prop) or prop
-
-    def translate_input_property(self, prop):
-        return _tables.SNAKE_TO_CAMEL_CASE_TABLE.get(prop) or prop
 
--- a/pkg/codegen/internal/test/testdata/external-resource-schema/python/pulumi_example/component.py
+++ b/pkg/codegen/internal/test/testdata/external-resource-schema/python/pulumi_example/component.py
@@ -90,10 +90,10 @@ class Component(pulumi.CustomResource):
                 raise TypeError('__props__ is only valid when passed in combination with a valid opts.id to get an existing resource')
             __props__ = ComponentArgs.__new__(ComponentArgs)
 
-            __props__.__dict__['metadata'] = metadata
-            __props__.__dict__['provider'] = None
-            __props__.__dict__['security_group'] = None
-            __props__.__dict__['storage_classes'] = None
+            __props__.__dict__["metadata"] = metadata
+            __props__.__dict__["provider"] = None
+            __props__.__dict__["security_group"] = None
+            __props__.__dict__["storage_classes"] = None
         super(Component, __self__).__init__(
             'example::Component',
             resource_name,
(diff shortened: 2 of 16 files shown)


pkg/codegen/internal/test/testdata/external-resource-schema/python/pulumi_example/arg_function.py
```diff
@@ -6,7 +6,7 @@
 import pulumi
 import pulumi.runtime
 from typing import Any, Mapping, Optional, Sequence, Union, overload
-from . import _utilities, _tables
```
komalali, line 9: 😍

pkg/codegen/internal/test/testdata/simple-enum-schema/python/pulumi_plant/tree/v1/rubber_tree.py
```diff
@@ -88,6 +88,28 @@ def size(self, value: Optional[pulumi.Input['TreeSize']]):
...
+class _RubberTreeState:
+    def __init__(__self__, *,
+                 farm: Optional[pulumi.Input[Union['Farm', str]]] = None):
+        """
+        Input properties used for looking up and filtering RubberTree resources.
+        """
+        if farm is None:
+            farm = '(unknown)'
+        if farm is not None:
+            pulumi.set(__self__, "farm", farm)
```
komalali, line 101: Very nitty but the inconsistent use of single and double quotes is a personal pet peeve. We seem to generally opt for double quotes in our python code, but I'm guessing because of the generated code and the burden of having to escape the single quotes is why we've ended up in this situation. An alternative we can use in code generation is to use `%q` instead of `\"%s\"` or `'%s'`. [[Docs]](https://golang.org/pkg/fmt/#hdr-Printing:~:text=%25q%09a%20double%2Dquoted%20string%20safely%20escaped%20with%20Go%20syntax)

I've noted some spots in gen.go below where we could utilize this.

pkg/codegen/internal/test/testdata/external-resource-schema/python/pulumi_example/component.py
```diff
@@ -136,9 +136,3 @@ def security_group(self) -> pulumi.Output[Optional['pulumi_aws.ec2.SecurityGroup
     def storage_classes(self) -> pulumi.Output[Optional[Mapping[str, 'pulumi_kubernetes.storage.v1.StorageClass']]]:
         return pulumi.get(self, "storage_classes")
 
-    def translate_output_property(self, prop):
-        return _tables.CAMEL_TO_SNAKE_CASE_TABLE.get(prop) or prop
-
-    def translate_input_property(self, prop):
-        return _tables.SNAKE_TO_CAMEL_CASE_TABLE.get(prop) or prop
-
```
komalali, line 144: 😍

pkg/codegen/python/gen.go
```diff
@@ -1124,7 +1142,7 @@ func (mod *modContext) genResource(res *schema.Resource) (string, error) {
 		if res.IsProvider && !isStringType(prop.Type) {
 			arg = fmt.Sprintf("pulumi.Output.from_input(%s).apply(pulumi.runtime.to_json) if %s is not None else None", arg, arg)
 		}
-		fmt.Fprintf(w, "            __props__['%s'] = %s\n", PyName(prop.Name), arg)
+		fmt.Fprintf(w, "            __props__.__dict__['%s'] = %s\n", PyName(prop.Name), arg)
```
komalali, line 1145:
```suggestion
		fmt.Fprintf(w, "            __props__.__dict__[%q] = %s\n", PyName(prop.Name), arg)
```

pkg/codegen/python/gen.go
```diff
@@ -1134,7 +1152,7 @@ func (mod *modContext) genResource(res *schema.Resource) (string, error) {
 		// Default any pure output properties to None.  This ensures they are available as properties, even if
 		// they don't ever get assigned a real value, and get documentation if available.
 		if !ins.Has(prop.Name) {
-			fmt.Fprintf(w, "            __props__['%s'] = None\n", PyName(prop.Name))
+			fmt.Fprintf(w, "            __props__.__dict__['%s'] = None\n", PyName(prop.Name))
```
komalali, line 1155:
```suggestion
			fmt.Fprintf(w, "            __props__.__dict__[%q] = None\n", PyName(prop.Name))
```

pkg/codegen/python/gen.go
```diff
@@ -1205,18 +1223,24 @@ func (mod *modContext) genResource(res *schema.Resource) (string, error) {
...
+			// to opt-in to the improved translation behavior.
+			fmt.Fprintf(w, "        __props__ = %[1]sArgs.__new__(%[1]sArgs)\n\n", name)
+		}
 
 		stateInputs := codegen.NewStringSet()
 		if res.StateInputs != nil {
 			for _, prop := range res.StateInputs.Properties {
 				stateInputs.Add(prop.Name)
-				fmt.Fprintf(w, "        __props__[%[1]q] = %[1]s\n", PyName(prop.Name))
+				fmt.Fprintf(w, "        __props__.__dict__['%[1]s'] = %[1]s\n", PyName(prop.Name))
```
komalali, line 1238: Oh interesting, you actually changed the `%q` to `'%s'` - are you trying to standardize to single quotes?
```suggestion
				fmt.Fprintf(w, "        __props__.__dict__[%[1]q] = %[1]s\n", PyName(prop.Name))
```

pkg/codegen/python/gen.go
```diff
@@ -1205,18 +1223,24 @@ func (mod *modContext) genResource(res *schema.Resource) (string, error) {
...
 			}
 		}
 		for _, prop := range res.Properties {
 			if !stateInputs.Has(prop.Name) {
-				fmt.Fprintf(w, "        __props__[%[1]q] = None\n", PyName(prop.Name))
+				fmt.Fprintf(w, "        __props__.__dict__['%s'] = None\n", PyName(prop.Name))
```
komalali, line 1243:
```suggestion
				fmt.Fprintf(w, "        __props__.__dict__[%q] = None\n", PyName(prop.Name))
```

pkg/codegen/python/gen.go
```diff
@@ -2119,6 +2101,46 @@ func (mod *modContext) genType(w io.Writer, name, comment string, properties []*
...
+			fmt.Fprintf(w, "    def __key_warning(key: str):\n")
+			fmt.Fprintf(w, "        suggest = None\n")
+			prefix := "if"
+			for _, prop := range props {
+				pname := PyName(prop.Name)
+				if pname == prop.Name {
+					continue
+				}
+				fmt.Fprintf(w, "        %s key == \"%s\":\n", prefix, prop.Name)
+				fmt.Fprintf(w, "            suggest = \"%s\"\n", pname)
```
komalali, line 2126: Depending on whether we want to stick to single or double quotes, you might take the suggestion or change it to `'%s'`

```suggestion
				fmt.Fprintf(w, "        %s key == %q:\n", prefix, prop.Name)
				fmt.Fprintf(w, "            suggest = %q\n", pname)
```

komalali (approved):
LGTM but would be nice to align on quote styles

pkg/codegen/internal/test/testdata/simple-enum-schema/python/pulumi_plant/tree/v1/rubber_tree.py
justinvp: It's a pet peeve of mine as well! But, I've just shrugged it off in this generated code since we're already inconsistent throughout. In any case, I fixed all the lines that I modified in this PR to use double quotes. We can follow-up later as we have time to improve the rest of the generated code so it's consistent. My initial thought on a style guideline would be to use double quotes everywhere, but perhaps stick to single quotes for forward reference type hints?

pkg/codegen/python/gen.go
justinvp: I was just changing `get` to match `__init__` so that they were both setting `__props__` using the same quote style. I've now changed both to use double quotes.

Issue #6697, opened 2021-04-03
[sdk/python] Exception raised when serializing an output type that has "items" as one of the properties


praneetloke:
I was trying to export the `trusted_signers` output property of the `Distribution` resource when I hit this error. Note that this same problem exists when registering outputs in component resources via [`self.register_outputs()`](https://github.com/pulumi/pulumi/blob/master/sdk/python/lib/pulumi/runtime/resource.py#L482) as well when we [serialize properties](https://github.com/pulumi/pulumi/blob/master/sdk/python/lib/pulumi/runtime/rpc.py#L222-L227) there.

Although I hit this problem while exporting a CFN `Distribution` resource property, I believe this affects any resource that uses an output type with a property called `items`. To test the theory I made a local patch to the output serialization for [stack outputs](For https://github.com/pulumi/pulumi/blob/master/sdk/python/lib/pulumi/runtime/stack.py#L172-L175) using this:

```python
for key in attr:
    # ignore private keys
    if not key.startswith("_"):
        result[key] = massage(attr[key], seen)
```

which then correctly produces:

```
Outputs:
    trusted_signers: [
        [0]: {
            enabled: true
            items  : [
                [0]: {
                    awsAccountNumber: "self"
                }
            ]
        }
    ]
```

## Expected behavior

Output is exported without any errors.

## Current behavior

```
Diagnostics:
  pulumi:pulumi:Stack (aws-cloudfront-dev):
    error: Program failed with an unhandled exception:
    error: Traceback (most recent call last):
      File "/Users/praneetloke/.pulumi/bin/pulumi-language-python-exec", line 85, in <module>
        loop.run_until_complete(coro)
      File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/asyncio/base_events.py", line 608, in run_until_complete
        return future.result()
      File "/Users/praneetloke/temp-pulumi-stacks/aws-cloudfront/venv/lib/python3.8/site-packages/pulumi/runtime/stack.py", line 104, in run_in_stack
        await run_pulumi_func(lambda: Stack(func))
      File "/Users/praneetloke/temp-pulumi-stacks/aws-cloudfront/venv/lib/python3.8/site-packages/pulumi/runtime/stack.py", line 85, in run_pulumi_func
        raise exception
      File "/Users/praneetloke/temp-pulumi-stacks/aws-cloudfront/venv/lib/python3.8/site-packages/pulumi/output.py", line 178, in run
[...]

PR #6701, opened 2021-04-05
[python/sdk] - Fix serialization bug if output contains 'items' property

komalali:
Fixes: https://github.com/pulumi/pulumi/issues/6697

### DIFF
--- a/sdk/python/lib/pulumi/runtime/stack.py
+++ b/sdk/python/lib/pulumi/runtime/stack.py
@@ -169,10 +169,11 @@ def massage(attr: Any, seen: List[Any]):
     # make sure this is a popo.
     if isinstance(attr, dict):
         result = {}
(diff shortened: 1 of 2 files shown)


sdk/python/lib/pulumi/runtime/stack.py
```diff
@@ -169,10 +169,10 @@ def massage(attr: Any, seen: List[Any]):
     # make sure this is a popo.
     if isinstance(attr, dict):
         result = {}
-        for key, value in attr.items():
+        for key in attr:
```
justinvp, line 172: Add a comment on why we're not using `items()` here, so we don't accidentally regress in the future?

justinvp (approved):
LGTM

komalali:
Lol just adding that comment has added 4 hours and counting to be able to merge this

PR #6704, opened 2021-04-05
Enable nodejs dynamic provider caching by default on program side

EvanBoyle:
The full description and motivation of this change is outlined here: https://github.com/pulumi/pulumi/pull/6673

TLDR; this change enables caching by default for dynamic providers serialized as a part of executing a program via nodejs. This decreases memory usage and increases runtime significantly for programs making heavy use of dynamic providers. 

This is technically a breaking change as we are now caching dynamic providers based on object equality. It is technically true (but unlikely in practice) that a dynamic provider could rely on some global state, or that in-between resource creation operations that individual methods of the provider (diff/check/create/etc) could be mutated. 

Due to this, users should treat dynamic resource providers as immutable and opt to create singleton instances to get access to these performance benefits. This behavior can be disabled by setting `PULUMI_NODEJS_CACHE_DYNAMIC_PROVIDERS=false`.

### DIFF
--- a/sdk/nodejs/tests/automation/localWorkspace.spec.ts
+++ b/sdk/nodejs/tests/automation/localWorkspace.spec.ts
@@ -16,7 +16,6 @@ import * as assert from "assert";
 import * as semver from "semver";
 import * as upath from "upath";
 
-import { Config } from "../../index";
 import {
     ConfigMap,
     EngineEvent,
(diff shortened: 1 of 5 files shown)


sdk/nodejs/tests/automation/localWorkspace.spec.ts
```diff
@@ -26,6 +25,7 @@ import {
     Stack,
     validatePulumiVersion,
 } from "../../automation";
+import { Config } from "../../index";
```
EvanBoyle, line 28: Unrelated change, however linting was failing both locally and in CI.

EvanBoyle:
I'm seeing automation API tests fail due to the CLI version check in CI, however these passed locally after I built the CLI via after setting `VERSION_PREFIX=3.0.0`

stack72 (approved):

Issue #6561, opened 2021-03-17
Flakey python test errors with "Task attached to a different loop"


komalali:
The TestingWithMocks.test_component test intermittently errors with a `Task attached to a different loop` error: https://github.com/pulumi/pulumi/runs/2098866118#step:18:1754

## Expected behavior
The test should pass every time.

## Current behavior

```
=================================== FAILURES ===================================
_______________________ TestingWithMocks.test_component ________________________

args = (<test_with_mocks.test_testing_with_mocks.TestingWithMocks testMethod=test_component>,)
kwargs = {}

    def wrapper(*args, **kwargs):
        from .. import Output  # pylint: disable=import-outside-toplevel
>       _sync_await(run_pulumi_func(lambda: _sync_await(Output.from_input(fn(*args, **kwargs)).future())))

../pulumi/runtime/mocks.py:37: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
Error: ../pulumi/runtime/sync_await.py:54: in _sync_await
    return loop.run_until_complete(fut)
Error: /opt/hostedtoolcache/Python/3.9.2/x64/lib/python3.9/asyncio/base_events.py:642: in run_until_complete
    return future.result()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

func = <function test.<locals>.wrapper.<locals>.<lambda> at 0x7f68c09c2940>

    async def run_pulumi_func(func: Callable):
        try:
            func()
        finally:
            log.debug("Waiting for outstanding RPCs to complete")
    
            # Pump the event loop, giving all of the RPCs that we just queued up time to fully execute.
            # The asyncio scheduler does not expose a "yield" primitive, so this will have to do.
            #
            # Note that "asyncio.sleep(0)" is the blessed way to do this:
            # python/asyncio/issues/284#issuecomment-154180935
            #
            # We await each RPC in turn so that this loop will actually block rather than busy-wait.
            while True:
                await asyncio.sleep(0)
                rpcs_remaining = len(RPC_MANAGER.rpcs)
                if rpcs_remaining == 0:
                    break
                log.debug(f"waiting for quiescence; {rpcs_remaining} RPCs outstanding")
[...]

Issue #6108, opened 2021-01-14
Python 3.6 support broken under 2.17.1


csssuf:
With the release of #5979, Pulumi no longer runs under Python 3.6. Specifically, [this `__futures__` import](https://github.com/pulumi/pulumi/blob/master/sdk/python/lib/pulumi/x/automation/stack.py#L15) is only available on 3.7+. After a quick look, it looks like that file already uses quoted annotations for forward references [(e.g. here)](https://github.com/pulumi/pulumi/blob/master/sdk/python/lib/pulumi/x/automation/stack.py#L155), so there should be no changes necessary beyond removing the `__futures__` import.

Assuming this is unintended, I'd be happy to cut a PR to fix it.

## Expected Behavior
Per [the docs](https://www.pulumi.com/docs/reference/pkg/python/pulumi/#pulumi-python-sdk%C2%B6), Pulumi should support Python 3.6.

## Current Behavior
It does not.

## Steps to Reproduce
```jforcier@C02XR3YXJG5M ~/repos/pulumi/sdk/python/lib
(pulumi-36) > $ pip freeze | grep pulumi
pulumi==2.17.1
jforcier@C02XR3YXJG5M ~/repos/pulumi/sdk/python/lib
(pulumi-36) > $ python
Python 3.6.6 (default, Feb 27 2020, 11:00:07)
[GCC 4.2.1 Compatible Apple LLVM 11.0.0 (clang-1100.0.20.17)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import pulumi
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/jforcier/repos/pulumi/sdk/python/lib/pulumi/__init__.py", line 99, in <module>
    from . import runtime, dynamic, policy, x
  File "/Users/jforcier/repos/pulumi/sdk/python/lib/pulumi/x/__init__.py", line 15, in <module>
    from . import automation
  File "/Users/jforcier/repos/pulumi/sdk/python/lib/pulumi/x/automation/__init__.py", line 30, in <module>
    from .local_workspace import (
  File "/Users/jforcier/repos/pulumi/sdk/python/lib/pulumi/x/automation/local_workspace.py", line 26, in <module>
    from .stack import _DATETIME_FORMAT, Stack
  File "/Users/jforcier/repos/pulumi/sdk/python/lib/pulumi/x/automation/stack.py", line 15
    from __future__ import annotations
    ^
[...]

PR #6715, opened 2021-04-06
Remote component py SDK

t0yv0:
Toward pulumi#5487

- [x] finish build shenanigans so CI passes the test on all platforms
- [x] port isSecret && noDeps check that special-cases prompt value propagation from Node SDK
- [x] resolve question on locking Construct calls for serializable execution
- [x] resolve question on using runtime.disconnect in Construct calls
- [x] test on py 3.6
- [x] integration test prompt value propagation


sdk/python/lib/pulumi/provider/server.py
```diff
@@ -0,0 +1,189 @@
...
+
+    engine_address: str
+    provider: Provider
+    args: List[str]
+
+    async def Construct(self, request: ConstructRequest, context):  # pylint: disable=invalid-overridden-method
+        assert isinstance(request, ConstructRequest), \
+            f'request is not ConstructRequest but is {type(request)} instead'
+
+        # Should we take a lock so Construct calls are never executing
```
t0yv0, line 49: Unresolved Q1

sdk/python/lib/pulumi/provider/server.py
```diff
@@ -0,0 +1,189 @@
...
+
+        result = self.provider.construct(name=request.name,
+                                         resource_type=request.type,
+                                         inputs=inputs,
+                                         options=self._construct_options(request))
+
+
+        response = await self._construct_response(result)
+
+        ## should we do runtime.disconnect()? where is this
```
t0yv0, line 73: Unresolved Q2

t0yv0:
Looks like  I need to do some work to pass the mypy checks.

t0yv0:
OK I'd like to ask for some build help file. It's been too interesting, I think I am spending more effort on getting the CI run the test that I know works locally than getting the code right. Here's my problem. The example runs a process tree approx as follows:

```
go test (from tests/integration)
  pulumi
    ...
    pulumi-resource-pulumi-python
    pulumi-resource-testcomponent (py)
```

So integration framework initializes pulumi CLI which runs plugins; pulumi up is driven by node program, but a python program called pulumi-resource-testcomponent that I'm adding here is invoked as it's a remote resource provider; in addition pulumi-resource-pulumi-python is invoked because dynamic python resources are asked for in the test.

I need both of these Python subprocesses be able to reference Pulumi python SDK from the repo. I've tried many things at this point. 

The latest attempt goes as follows:

- `sdk/python` stores a reference to its "good to go" virtual environment in a file `$(PULUMI_BIN)/pulumi-buildtime-venv.conf`

- go test framework sets `PULUMI_RUNTIME_VIRTUALENV` env var to that venv; the env var is set in the env received by parent pulumi process; but it is inherited into pulumi-resource-pulumi-python so that works. Without the env var that subprocess crashes as it cannot locate pulumi SDK

- similarly, pulumi-resource-testcomponent receives  PULUMI_RUNTIME_VIRTUALENV and starts it's python code in the same venv; that helps it resolve pulumi SDK from the repo

However, the code does not work on Windows. Apparently windows build does not set PULUMI_BIN and seemingly does not utilize the same Makefile? Is that a totally parallel build process?

I would really appreciate some advice on what's the most reasonable way forward here.

CC @stack72

t0yv0:
After chatting with Paul have a few decent options going forward wrt builds, trying some of them here.

sdk/python/lib/pulumi/provider/server.py
justinvp: We did this in Node.js as a short-term "solution" since the `Construct` call modifies global state. The downside is that it'll result in deadlock if you make a nested `construct` call within the same provider.

I believe the Python SDK has the same problem as Node.js (modifying global state in `Construct`).

We’re going to need to find a better long-term solution. For Node.js, we’ll likely end up making use of [`vm.createContext`](https://nodejs.org/api/vm.html#vm_vm_createcontext_contextobject_options) to run the user's `construct` in its own context (with it's own globals). I'm planning to try this out this week.

Go doesn't have this problem as each `Construct` call will get its own [`pulumi.Context`](https://github.com/pulumi/pulumi/blob/ee2f65510bd8129adffb18d7e198bf12c1b84fee/sdk/go/pulumi/context.go#L44).

The short-term thing for Python would be to do the same thing we did with Node.js (use locks to serialize `Construct` calls). That would at least ensure correctness when there are multiple concurrent `Construct` calls happening within the same process.

But I'm not sure what the right long-term solution will be for Python. Launch a new `python` process for each `Construct` call? Is there an equivalent of `vm.createContext` for Python?

sdk/python/lib/pulumi/provider/server.py
justinvp: The main thing we're using `runtime.disconnect()` for in Node.js is to wait for in-flight RPC calls to finish before returning from `Construct`, to ensure any pending `registerResource` calls (for a component's child resources) have completed before returning.

We do the same for `Go` by calling `waitForRPCs`: https://github.com/pulumi/pulumi/blob/34eae7e7e3ddae475c52f2e248b1eabb23d62cf7/sdk/go/pulumi/provider.go#L687-L690

I'm less familiar with this part of the codebase, but I think we may need to run the user's `construct` within `run_pulumi_func`, which looks like it'll handle this:

https://github.com/pulumi/pulumi/blob/ee2f65510bd8129adffb18d7e198bf12c1b84fee/sdk/python/lib/pulumi/runtime/stack.py#L41

sdk/python/lib/pulumi/provider/server.py
t0yv0: Ack on the short-term will implement that straight away. I can do some research on what we can do in Python for this long term and get back to you. At a first glance there is `threading.local()` so multithreaded programs can have per-thread state, which is a bit less overhead than a process; there's gotchas with Python threading though. I can see what async Python offers.

Would a "clean" solution of refactoring the SDK to not have to set the globals but pass parameters explicitly be too prohibitive?

sdk/python/lib/pulumi/provider/server.py
t0yv0: Ack will make it happen in Python.

Curious, what kind of user code would lead to a situation where there are outstanding RPCs after provider.construct has returned, and especially after we're waited on the constructed component's URN Output. Something along the lines of the code creating child components and not awaiting their URNs?

justinvp: > Something along the lines of the code creating child components and not awaiting their URNs?

Yeah

sdk/python/lib/pulumi/provider/server.py
komalali: Maybe worth noting here is that currently, each RPC in python [runs in its own thread](https://github.com/pulumi/pulumi/blob/master/sdk/python/lib/pulumi/runtime/resource.py#L304) as we have not yet adopted grpc-python's [async API](https://grpc.github.io/grpc/python/grpc_asyncio.html). This has led to some [odd behavior](https://github.com/pulumi/pulumi/issues/6561) in the interaction of asyncio event loops and threads.

I'm not familiar enough with the code here to know if it's relevant, but just wanted to flag it in case it is.

sdk/python/lib/pulumi/provider/server.py
```diff
@@ -0,0 +1,189 @@
...
+        servicer = ProviderServicer(provider, args, engine_address=engine_address)
+        provider_pb2_grpc.add_ResourceProviderServicer_to_server(servicer, server)
+        port = server.add_insecure_port(address='0.0.0.0:0')
+        await server.start()
+        sys.stdout.buffer.write(f'{port}\n'.encode())
+        sys.stdout.buffer.flush()
+        await server.wait_for_termination()
+
+    try:
+        asyncio.run(serve())
```
komalali, line 170: We can't use `asyncio.run` as it was only introduced in python 3.7 and we support python 3.6. You'll have to use `asyncio.get_event_loop().run_until_complete(serve())` or something along those lines.

sdk/python/lib/pulumi/provider/server.py
komalali: Follow-up - We'll need to [bump the version](https://github.com/grpc/grpc/releases/tag/v1.33.2#tag-select-menu-0e7b813a-9bc3-11eb-8090-b5026ca30356:~:text=Allow%20asyncio%20API%20to%20be%20imported%20as%20grpc.aio.) of grpc to use the async API in our [Pipfile](https://github.com/pulumi/pulumi/blob/master/sdk/python/Pipfile#L9)

t0yv0: I'm adopting gRPC async API in this PR. Are there problems with it?

sdk/python/stubs/grpc/aio.pyi
```diff
@@ -0,0 +1,12 @@
...
+
+
+class Server:
+    def add_insecure_port(self, address: str) -> int: ...
+    async def start(self) -> None: ...
+    async def stop(self, grace: Optional[float]) -> None: ...
+    async def wait_for_termination(self, timeout: Optional[float]=...) -> bool: ...
+
+
+def server(options: Any) -> Server: ...
```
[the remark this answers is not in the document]
t0yv0: Hm does it? I couldn't quite locate them, and mypy was complaining.

sdk/python/lib/pulumi/provider/server.py
t0yv0: Hmm. I can def make that fix but I'm not confident the entire thing runs on 3.6. Do we have a way to run automated tests against 3.6?

tests/integration/integration_nodejs_test.go
```diff
@@ -734,3 +746,30 @@ func TestGetResourceNode(t *testing.T) {
...
+	return venv, nil
+}
+
+func pulumiRuntimeVirtualEnv(pulumiRepoRootDir string) (string, error) {
+	venvFolder, err := venvFromPipenv(filepath.Join(pulumiRepoRootDir, "sdk", "python"))
+	if err != nil {
+		return "", fmt.Errorf("PULUMI_RUNTIME_VIRTUALENV guess failed: %w", err)
+	}
+	return fmt.Sprintf("PULUMI_RUNTIME_VIRTUALENV=%s", venvFolder), nil
+}
```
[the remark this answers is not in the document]
t0yv0: Not necessarily. They test nodejs client program working with python component in multi-lang. So they test the integration of two languages. I can move then to integration_multilang_test.go if desired.

sdk/python/lib/pulumi/provider/server.py
komalali: > I'm adopting gRPC async API in this PR. Are there problems with it?

Not necessarily! I'm of the opinion that we should be adopting it for all of our usage of grpc across the codebase, so I'm glad to see it here. I just don't know if/how this will interact with the existing code in the rest of the python SDK. This isn't a "don't do it" but more like "I'm curious to see what happens".

t0yv0: This new code would run inside a subprocess dedicated to wrapping user code that implements the new provider interface. I'm optimistic this will therefore not impact any existing code, but we should be comfortable maintaining it in the future.

sdk/python/lib/pulumi/provider/server.py
komalali: We definitely have users that are on 3.6 and they certainly [bring it up](https://github.com/pulumi/pulumi/issues/6108) when there's a break. We also go to [great pains](https://github.com/pulumi/pulumi/blob/master/sdk/python/lib/pulumi/runtime/sync_await.py#L20-L38) to make sure we remain backward compatible with 3.6. 

We don't have automated testing with 3.6, we've tried implementing it but have run into issues with the test runners themselves, and not the actual SDK. [There's an open issue](https://github.com/pulumi/pulumi/issues/6110).

tests/integration/integration_nodejs_test.go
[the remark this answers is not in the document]
komalali: Ah got it. I don't have an opinion in that case.

sdk/python/lib/pulumi/dynamic/__main__.py
```diff
@@ -107,7 +107,7 @@ def Create(self, request, context):
         props = rpc.deserialize_properties(request.properties)
         provider = get_provider(props)
         result = provider.create(props)
-        outs = result.outs
+        outs = result.outs if result.outs is not None else {}
```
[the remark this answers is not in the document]
t0yv0: Hm, that's less strict. "if result.outs is not None" is stricter than "if result" for example `{}` value evaluates differently. Here it probably does not matter too much just explaining where my habit comes from..

sdk/python/lib/pulumi/provider/server.py
t0yv0: Thanks for the context. I'll implement your suggestion and watch out for any complaints.

komalali: At the very least, if we are choosing to break with our existing [supported languages](https://www.pulumi.com/docs/intro/languages/#languages) the decision should be explicit and agreed upon by the team.

sdk/python/lib/pulumi/dynamic/__main__.py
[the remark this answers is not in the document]
komalali: Yeah, fair enough.

sdk/python/lib/pulumi/provider/server.py
komalali: Yeah, I've been playing around with the async API in a local branch and would certainly like to adopt it elsewhere due to the aforementioned oddities with how our async code interacts with the RPCs. Thanks for taking a stab at it!

t0yv0:
Phew, we finally have a passing Windows test.

sdk/python/lib/pulumi/provider/server.py
t0yv0: Not good, adding `await wait_for_rpcs()` makes the example test not terminate. I'm trying to get detailed logs out to understand what is going on.

t0yv0: OK so, as it turns out, wait_for_rpcs() waits for     _all_tasks = asyncio.Task.all_tasks

```
all_tasks(loop=None) method of builtins.type instance
    Return a set of all tasks for an event loop.
    
    By default all tasks for the current event loop are returned.
```

That I think is a bug - this time it picked up AIO tasks related to running the gRPC server and started blocking on those.

I think it is a bug to ever poke into all_tasks as we want user code to be able to schedule their own tasks without interference on the shared event loop.

Do we have a dedicated event loop to segregate?

 However it might not be possible to refactor swiftly. Please advice what's a good way forward. @komalali

komalali: ...where/what is `wait_for_rpcs()`?

In any case, these are the dragons I was hinting at [up here](https://github.com/pulumi/pulumi/pull/6715#discussion_r611876606). `run_pulumi_func` has some very odd behavior that I haven't been able to tease apart which is why I've largely left the whole [RPC_MANAGER stuff](https://github.com/pulumi/pulumi/blob/master/sdk/python/lib/pulumi/runtime/stack.py#L48-L58) alone because I also ran into infinite hangs when trying to address it. 

> Do we have a dedicated event loop to segregate?
Unfortunately, we tend to just [grab the existing event loop](https://github.com/pulumi/pulumi/blob/master/sdk/python/lib/pulumi/runtime/sync_await.py#L90-L98) so we don't have a dedicated one.

You have found yourself in some of the hairiest python code in our codebase. Maybe we can do a quick call to talk through options?

t0yv0: Yeah that sounds good. I am tempted to burn it with FIRE so talking is good.

t0yv0: https://github.com/pulumi/pulumi/issues/6762 we logged this, I will proceed to try to isolate only the code that waits on RPCs, that might work for the purposes of this PR. Revisit the 6761 later.

t0yv0: Waiting just for RPCs does not lock, great news.

sdk/python/lib/pulumi/provider/server.py
t0yv0: Regarding Pipfile, as written it resolves to 1.37.0 so I think we're good? Do we have Pipfile.lock anywhere?

komalali: > Regarding Pipfile, as written it resolves to 1.37.0

Right, but it is incorrect because you actually need `>=1.33.2`. For folks who have an existing version on their machines with a `Pipfile.lock` on a version prior to 1.33.2 wouldn't know to update. 

> Do we have Pipfile.lock anywhere?

We don't put the `Pipfile.lock` in source control because we target multiple versions of python and multiple platforms and the `Pipfile.lock` will contain version-specific/platform-specific dependencies.

t0yv0: Thanks so much for explaining makes sense. I will add.

t0yv0:
Should be ready to go now @justinvp @komalali

sdk/python/lib/pulumi/provider/server.py
```diff
@@ -0,0 +1,231 @@
...
+        }
+
+    @staticmethod
+    def _construct_output(the_input: Any, deps: Set[Resource]) -> Any:
+        is_secret = rpc.is_rpc_secret(the_input)
+
+        # If it's a prompt value, return it directly without wrapping
+        # it as an output.
+        if not is_secret and len(deps) == 0:
+            return the_input
```
justinvp, line 120: We have a basic integration test for prompt values at https://github.com/pulumi/pulumi/tree/master/tests/integration/construct_component_plain.

I could imagine implementing a `testcomponent-python` similar to the integration test in this PR.

justinvp (approved):
🎉 LGTM otherwise.

sdk/python/lib/pulumi/provider/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+# limitations under the License.
+
+from pulumi.provider.provider import Provider, ConstructResult
+from pulumi.provider.server import main
+
+__all__ = [getattr(x, '__name__') for x in [
+    Provider,
+    ConstructResult,
+    main
+]]
```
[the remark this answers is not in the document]
komalali: Yes, it seems that at least in VS Code, Pylance can't follow this and fails to autocomplete. I would suggest that we use the string literals.

We should also add "provider" to the [list of submodules](https://github.com/pulumi/pulumi/blob/master/sdk/python/lib/pulumi/__init__.py#L168-L172) 

```suggestion
__all__ = [
    "Provider",
    "ConstructResult",
    "main"
]
```

getattr:
Screen Shot 2021-04-14 at 12 46 52 PM

string literals:
Screen Shot 2021-04-14 at 12 49 03 PM

sdk/python/lib/pulumi/provider/provider.py
```diff
@@ -0,0 +1,58 @@
...
+
+    """
+
+    version: str
+
+    def __init__(self, version: str) -> None:
+        self.version = version
+
+    def construct(self, name: str, resource_type: str, inputs: Inputs,
+                  options: Optional[ResourceOptions]=None) -> ConstructResult:
```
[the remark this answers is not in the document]
t0yv0: https://github.com/pulumi/pulumi/issues/6773 - applied pycodestyle manually for now, it picked it up. It enforces PEP 8.

sdk/python/lib/pulumi/runtime/proto/provider_pb2_grpc.py
```diff
@@ -103,44 +103,44 @@ def GetSchema(self, request, context):
     """GetSchema fetches the schema for this resource provider.
     """
     context.set_code(grpc.StatusCode.UNIMPLEMENTED)
-    context.set_details('Method not implemented!')
-    raise NotImplementedError('Method not implemented!')
+    context.set_details('Method not implemented: GetSchema!')
```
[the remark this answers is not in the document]
t0yv0: Sure, will revert. This mostly helped me in the initial phase trying to trace the gRPC. I now have some better ways to trace gRPC anyway so this can be reverted.

sdk/python/lib/pulumi/provider/server.py
```diff
@@ -0,0 +1,231 @@
...
+
+        # Otherwise, wrap it as an output so we can handle secrets
+        # and/or track dependencies.
+        return pulumi.Output(
+            resources=deps,
+            future=_as_future(rpc.unwrap_rpc_secret(the_input)),
+            is_known=_as_future(True),
+            is_secret=_as_future(is_secret))
```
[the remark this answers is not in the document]
t0yv0: https://github.com/pulumi/pulumi/issues/6774 tabled in there

build.proj
```diff
@@ -217,9 +217,12 @@
           DestinationFolder="$(PythonSdkDirectory)\env\src" />
     <Exec Command="pipenv run python setup.py build bdist_wheel --universal"
           WorkingDirectory="$(PythonSdkDirectory)\env\src" />
+    <!-- Install pulumi SDK into the venv managed by pipenv. -->
+    <Exec Command="pipenv run pip install -e ."
+          WorkingDirectory="$(PythonSdkDirectory)\env\src" />
```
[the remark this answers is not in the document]
t0yv0: I will try it. If there are any adverse side-effects it definitely is better to confine them to test phase.

PR #6727, opened 2021-04-08
[automation/dotnet] Import & Export support

vipentti:
This is currently WIP, but since the basic functionality is working, at least based on the simple test (which is based on the one in the nodejs automation API), I'd like to get some feedback on what kind of API we want to have for managing the deployment resources.

Some questions I have:

- What kind of problems or other reasons usually require editing the exported json?
- Should we support older versions of the types (e.g. ResourceV1, ResourceV2 etc)?

Some notes:

- Currently all the import/export types have public getters and setters, which I'm not a huge fan of, it makes the implementation simple, but do we want users to be able to create deployments manually via constructing resources with all of their properties?

- Currently the types for import/export are in a separate namespace `Pulumi.Automation.Types` and the namings for the resources include the version e.g. `ResourceV3` to match the types in apitype/core.go. Maybe the types should be in the top-level `Pulumi.Automation` namespace? or some more descriptive namespace.

- By default System.Text.Json encodes certain characters when serializing such as `+` gets encoded as `\u002B`, this affects the serialized output which is used when importing. I believe according to JSON-specification these should be treated as identical, but something worth nothing. This can be changed by specifying an encoder for the [JsonSerializerOptions](https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-character-encoding#serialize-all-characters).

  > The default encoder escapes non-ASCII characters, HTML-sensitive characters within the ASCII-range, and characters that must be escaped according to the RFC 8259 JSON spec.

  Source: https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-how-to?pivots=dotnet-core-3-1#serialization-behavior

- Currently everything is intentionally marked as `internal` while working on this to avoid having to keep `PublicAPI.Unshipped.txt` in sync

- Currently all the types used by import/export are in a single file for simplicity during development, the types will be split into separate files when they are ready

TODO:

- [ ] Serialization tests
- [ ] Additional import & export tests with edits
- [ ] Ensure all the types, methods and properties which should be public are public and updated in `PublicAPI.Unshipped.txt`

### DIFF
--- a/sdk/dotnet/Pulumi.Automation/Types/ExportTypes.cs
+++ b/sdk/dotnet/Pulumi.Automation/Types/ExportTypes.cs
@@ -1,0 +1,252 @@
+// Copyright 2016-2021, Pulumi Corporation
+// NOTE: The classes in this file are intended to align with the serialized
+// JSON types defined and versioned in sdk/go/common/apitype/core.go
+
+using System;
+using System.Collections.Generic;
+using System.Text.Json.Serialization;
+
+// TODO(vipentti): Select proper namespace
+namespace Pulumi.Automation.Types
+{
+    // TODO(vipentti): internal -> public
+    // TODO(vipentti): Split into file per class
+    // TODO(vipentti): Separate JSON models & public types?
+
+    internal class VersionedDeployment
+    {
+        public int? Version { get; set; }
+
+        public DeploymentV3 Deployment { get; set; } = null!;
+    }
+
+    /// <summary>
+    /// <see cref="EngineOperationType"/> is the type of an operation initiated by the engine. Its value indicates the type of operation
+    /// that the engine initiated.
+    /// </summary>
+    internal enum EngineOperationType
+    {
+        /// <summary>
+        /// Creating is the state of resources that are being created.
+        /// </summary>
+        Creating,
+
+        /// <summary>
+	    /// Updating is the state of resources that are being updated.
+        /// </summary>
+        Updating,
+
+        /// <summary>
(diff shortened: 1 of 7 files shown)


sdk/dotnet/Pulumi.Automation/Types/ExportTypes.cs
```diff
@@ -0,0 +1,252 @@
...
+    // TODO(vipentti): internal -> public
+    // TODO(vipentti): Split into file per class
+    // TODO(vipentti): Separate JSON models & public types?
+
+    internal class VersionedDeployment
+    {
+        public int? Version { get; set; }
+
+        public DeploymentV3 Deployment { get; set; } = null!;
+    }
```
vipentti, line 21: This type is basically `UntypedDeployment` with a single specified deployment type

t0yv0:
Oh no, just seeing this @vipentti - I'm really sorry I merged some bare-bones support for this before looking at / triaging this PR. Lesson learned I guess I will be looking at existing PRs before trying to implement something.

Even with what I merged, I think your PR here goes deeper in providing a typed model for the data returned by `stack export`. This can be valuable? Consider resurrecting the PR and we can work with you to carry it to completion?

vipentti:
No problem at all @t0yv0 , I've rebased this on top of the latest changes and now this is only limited to implementing typed import/export support for dotnet. The new PR is in #6770 

Ideally I think it would be nice if we were able to provide somewhat safe ways of modifying parts of the deployment state to fix common problems.

PR #6730, opened 2021-04-08
explicitly create event log file for nodejs automation api

EvanBoyle:
An attempt at fixing https://github.com/pulumi/actions/issues/163

The current implementation just creates a temp dir and waits for the CLI to start writing to that file. If that process is delayed by more than 10s we fail. This change tries to be resilient to that case.

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -17,6 +17,9 @@
 
 ### Bug Fixes
 
+- [sdk/nodejs] Explicitly create event log file for NodeJS Automation API.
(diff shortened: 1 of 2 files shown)


komalali (approved):
LGTM, the existing tests should be signal enough that it works as expected.

komalali:
Should we update the [readLines function](https://github.com/pulumi/pulumi/blob/master/sdk/nodejs/x/automation/stack.ts#L115-L141) to avoid the 10s poll for the file to exist since we're just creating it up front now?

Issue #6459, opened 2021-03-03
Automation API C# Updates for GA


leezen:
I can refresh that #5761 did everything except:

- `RemoteProgram` support
- Stack `import/export/cancel` support
- Concurrency (#6139)

Also during #5761 we talked about re-naming the following methods:

1. `GetConfig` -> `GetConfigValue`
2. `GetAllConfig` -> `GetConfig`
3. `SetConfig` -> `SetConfigValue`
4. `SetAllConfig` -> `SetConfig`

And those changes haven't been reflected in the other SDKs yet.

_Originally posted by @orionstudt in https://github.com/pulumi/pulumi/issues/5596#issuecomment-789742923_

leezen:
Based on discussion 3/16, we think the only things that need to happen are:
* Stack import/export/cancel support
* Concurrency (#6139)

komalali:
@leezen about renaming the methods, what we discussed during the meeting today is actually the opposite of the current state. All of the values listed first (i.e. GetConfig) are what is used in the other SDKs (nodejs/python/go), the ones listed second `GetConfigValue` is what is used only in dotnet. So if we want to bring them into alignment there is still work to do here for dotnet.

orionstudt:
I'm sorry if I made that confusing. I meant for what was before the arrow to be what it is currently, and after the arrow to be what was agreed it should be changed too

leezen:
@komalali Thanks for clarifying. In which case, I'll update the comment above to indicate we should bring that in as well.

t0yv0:
This is done.

PR #6731, opened 2021-04-08
Method renames for dotnet automation API GA

t0yv0:
Takes care of the part of https://github.com/pulumi/pulumi/issues/6459 as relates to method renames. In addition renames RemoveConfigAsync and RemoveAllConfigAsync.


t0yv0:
CC @orionstudt

orionstudt:
I will say that I'm not a fan of these names... I had thought that in the original Automation API PR we had agreed to go with names that unify related methods alphabetically for intellisense and usage purposes [here](https://github.com/pulumi/pulumi/pull/5761#discussion_r523156519)

t0yv0:
Thanks for that pointer @orionstudt and the feedback. I think the current priority is cross-SDK consistency of the operation names so the decision was made to got his way. I will take a note to bring this up again with the team so we may reconsider.

komalali:
@orionstudt Ah, I definitely missed that entire conversation. 

The issues for the other languages were never opened, and I wasn't aware that we had decided that we would align the other languages with C# rather than the other way around.

At this point, it wouldn't be entirely painless to take this breaking change in the 3 other SDKs, not to mention the related changes in the code that consumes these APIs. If we were to do it, we need to somehow fit that into the next 12 days which I'm not sure we have the capacity for.

orionstudt:
@komalali it is what it is! I don't think it is that big of an issue, but I do think there is value in related methods being near each other alphabetically for intellisense. I think I tried to clarify the miscommunication [here](https://github.com/pulumi/pulumi/issues/6459#issuecomment-800731883) but I failed to be specific enough so its on me 😅

PR #6734, opened 2021-04-09
Enable absolute and relative parent paths for pulumi main

EvanBoyle:
Required to fix https://github.com/pulumi/pulumi/issues/5578

See previous discussion in prototype PR here: https://github.com/pulumi/pulumi/pull/6247

TLDR; this change relaxes constraints on the `main` field in a pulumi project (pulumi.yaml) to accept absolute paths and relative paths that point to a parent directory. These are old constraints from PPC that are no longer relevant. Relaxing these constraints will allow us to fix bugs around file assets, closure serialization, and dynamic providers for automation API inline programs.

### DIFF
--- a/tests/integration/integration_nodejs_test.go
+++ b/tests/integration/integration_nodejs_test.go
@@ -102,7 +102,7 @@ func TestProjectMain(t *testing.T) {
 		e.ImportDirectory("project_main_abs")
 
 		// write a new Pulumi.yaml using the absolute path of the environment as "main"
-		yamlPath := filepath.Join(e.RootPath, "Pulumi.Yaml")
+		yamlPath := filepath.Join(e.RootPath, "Pulumi.yaml")
 		absYamlContents := fmt.Sprintf(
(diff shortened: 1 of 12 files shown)


tests/integration/integration_nodejs_test.go
```diff
@@ -92,35 +92,50 @@ func TestProjectMain(t *testing.T) {
...
 		e := ptesting.NewEnvironment(t)
 		defer func() {
 			if !t.Failed() {
 				e.DeleteEnvironment()
 			}
 		}()
 		e.ImportDirectory("project_main_abs")
+
+		// write a new Pulumi.yaml using the absolute path of the environment as "main"
+		yamlPath := filepath.Join(e.RootPath, "Pulumi.Yaml")
```
lukehoban, line 105: Total nit - but I'm surprised this works without the file bring `Pulumi.yaml`?

lukehoban (approved):

tests/integration/integration_nodejs_test.go
EvanBoyle: Yup, that is a typo. I must have a case insensitive file system. Updated.

Issue #6737, opened 2021-04-09
limit `pulumi stack history` to a default of 10


EvanBoyle:
This command can be extremely slow for stacks with many updates. We should change the default behavior for 3.0.

PR #6739, opened 2021-04-09
Set defaults for `pulumi stack history`

EvanBoyle:
fixes https://github.com/pulumi/pulumi/issues/6737

Set a default limit of 10 entries for `pulumi stack history`. It isn't uncommon for a stack to have hundreds of history entries, in which case accidentally running this command can cause hangs for several seconds.


stack72 (approved):

chrsmith (approved):
LGTM!

PR #6741, opened 2021-04-09
Don't return an error if DeleteAllAccounts failed because the creds file doesn't exist

svangordon-fruit:
Currently, running `pulumi logout --all` will error out if the `credentials.json` file doesn't exist. With this change, the client won't complain if it fails to delete the creds file because it doesn't exist.


stack72:
Thanks so much @svangordon-fruit!!

PR #6743, opened 2021-04-09
Set main for default projects in Automation API

EvanBoyle:
Fixes https://github.com/pulumi/pulumi/issues/5578

This sets the `main` property when create default projects for inline programs so that relative paths properly resolve. This allows many utilities to function as expected such as assets, archives, closure serialization, magic functions, and dynamic providers. 

Writing tests for this proved to be challenging. Assets were the first thing I thought to reach for, but they don't wail when pointing at an invalid path so that was effectively useless: https://github.com/pulumi/pulumi/issues/5577

I did try to write a dynamic provider test, and couldn't get it to pass inside of our existing test infrastructure (the defaults are designed for situations where you have /index.ts and /node_modules as siblings). 

I did manually verify this with an inline program that uses a dynamic provider and everything worked as expected (and failed without the change applied as expected):

```ts
import { InlineProgramArgs, LocalWorkspace } from "@pulumi/pulumi/x/automation";

const process = require('process');

const args = process.argv.slice(2);
let destroy = false;
if (args.length > 0 && args[0]) {
    destroy = args[0] === "destroy";
}

const run = async () => {
    // This is our pulumi program in "inline function" form
    const pulumiProgram = async () => {
        
        const myProvider: pulumi.dynamic.ResourceProvider = {
            async create(inputs: any) {
        
                return { id: "foo", outs: { "abc": "def"} };
            },
        }
        class MyResource extends pulumi.dynamic.Resource {
            constructor(name: string, props: {}, opts?: any) {
                super(myProvider, name, props, opts);
            }
        }


       const res = new MyResource("foo", {});

        return {
            res,
        };
    };

    // Create our stack 
    const args: InlineProgramArgs = {
        stackName: "dev22",
        projectName: "inlineNode",
        program: pulumiProgram
    };

    // create (or select if one already exists) a stack that uses our inline program
    const stack = await LocalWorkspace.createOrSelectStack(args);

    console.info("successfully initialized stack");
    console.info("installing plugins...");
    await stack.workspace.installPlugin("aws", "v3.6.1");
    console.info("plugins installed");
    console.info("setting up config");
    await stack.setConfig("aws:region", { value: "us-west-2" });
    console.info("config set");
    console.info("refreshing stack...");
    await stack.refresh({ onOutput: console.info });
    console.info("refresh complete");

    if (destroy) {
        console.info("destroying stack...");
        await stack.destroy({ onOutput: console.info });
        console.info("stack destroy complete");
        process.exit(0);
    }

    console.info("updating stack...");
    const upRes = await stack.up({ onOutput: console.info });
    console.log(`update summary: \n${JSON.stringify(upRes.summary.resourceChanges, null, 4)}`);
    console.log(`res : ${upRes.outputs.res.value}`);
};

run().catch(err => console.log(err));
```

### DIFF
--- a/sdk/go/x/auto/minimum_version.go
+++ b/sdk/go/x/auto/minimum_version.go
@@ -20,4 +20,10 @@ var minimumVersion = semver.Version{
 	Major: 2,
 	Minor: 25,
 	Patch: 0,
+	Pre: []semver.PRVersion{
+		{
+			VersionStr: "alpha",
+			IsNum:      false,
+		},
+	},
 }
--- a/sdk/go/x/auto/minimum_version.go
+++ b/sdk/go/x/auto/minimum_version.go
@@ -18,6 +18,6 @@ import "github.com/blang/semver"
 
 var minimumVersion = semver.Version{
 	Major: 2,
-	Minor: 21,
+	Minor: 25,
 	Patch: 0,
 }
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -28,6 +28,9 @@
 
 ### Bug Fixes
 
+- [automation] Set default value for 'main' for inline programs to support relative paths, assets, and closure serialization.
+  [#6743](https://github.com/pulumi/pulumi/pull/6743)
+
 - [sdk/nodejs] Explicitly create event log file for NodeJS Automation API.
   [#6730](https://github.com/pulumi/pulumi/pull/6730)
 
--- a/sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
+++ b/sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
@@ -35,7 +35,7 @@ namespace Pulumi.Automation
         private readonly LocalSerializer _serializer = new LocalSerializer();
         private readonly bool _ownsWorkingDir;
         private readonly Task _readyTask;
-        private static readonly SemVersion _minimumVersion = SemVersion.Parse("2.25.0");
+        private static readonly SemVersion _minimumVersion = SemVersion.Parse("2.25.0-alpha");
 
         /// <inheritdoc/>
         public override string WorkDir { get; }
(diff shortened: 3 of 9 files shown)


sdk/go/x/auto/minimum_version.go
```diff
@@ -18,6 +18,6 @@ import "github.com/blang/semver"
 
 var minimumVersion = semver.Version{
 	Major: 2,
-	Minor: 21,
+	Minor: 25,
 	Patch: 0,
 }
```
komalali, line 23: We'll need this to include "-alpha.0" otherwise our dev builds break.

komalali:
> I did manually verify this with an inline program that uses a dynamic provider and everything worked as expected

Am I understanding correctly that this exact test as an inline program in `localWorkspace.spec.ts` doesn't work?

sdk/go/x/auto/minimum_version.go
komalali: Or if we want to be specific it should be `-alpha.[whatever-build-this-will-be-once-merged]` but the latest build on npmsjs is `alpha.1618004182`

EvanBoyle:
> Am I understanding correctly that this exact test as an inline program in localWorkspace.spec.ts doesn't work?

You are understanding correctly, I did try to write this as a test and it didn't work. Dynamic providers rely on doing a require of `@pulumi/pulumi` at program runtime which resolves the module via `node_modules` and since this is `@pulumi/pulumi` that require doesn't exist in its node modules. We would need something that mimicked a true integration test, `yarn install` and run `yarn start` a standalone automation script and observe the results.

sdk/go/x/auto/minimum_version.go
EvanBoyle: Applied and tested this locally on go and node with my current dev build.

komalali (approved):
We can add a test once we get around to fixing #5577

EvanBoyle:
opened https://github.com/pulumi/pulumi/issues/6744 for follow up

PR #6746, opened 2021-04-12
Fix CHANGELOG_PENDING to mark 6731 as breaking, expand note

t0yv0:


komalali (approved):
Awesome, thanks for fixing this up!

stack72 (approved):
Thank you so much!!

PR #6751, opened 2021-04-12
Update serialization keys for StackSettings

leezen:


sdk/dotnet/Pulumi.Automation/StackSettings.cs
```diff
@@ -9,18 +9,24 @@ public class StackSettings
         /// <summary>
         /// This stack's secrets provider.
         /// </summary>
+        [YamlDotNet.Serialization.YamlMember(Alias="secretsprovider")]
+        [System.Text.Json.Serialization.JsonPropertyName("secretsprovider")]
```
t0yv0, line 13: The code might be using a custom serializer, see StackSettingsConfigValueJsonConverter it looks very suspect. A test would make sure this is taking effect.

leezen:
Oops, should be against `master` instead of 3.0

PR #6752, opened 2021-04-12
[dotnet] Fix serialization/deserialization for StackSettings

leezen:


leezen:
@t0yv0 You mentioned in the https://github.com/pulumi/pulumi/pull/6749 worry about the annotations. The custom serializer/deserializer you mentioned is specifically for the config portion (anything under the `config` key in the settings) which these attributes don't affect. There are tests for deserializing `secretsprovider` which pass.

t0yv0:
Confirmed, this passes also:

```
    public class StackSettingsConfigValueJsonConverterTests
    {
        private static LocalSerializer _serializer = new LocalSerializer();

        [Fact]
        public void SerializesKeysCorrectly()
        {
            var settings = new StackSettings()
            {
                SecretsProvider = "SP",
            };
            var json = _serializer.SerializeJson(settings);
            Assert.Equal(@"{""secretsprovider"":""SP""}", json);
        }

```

CHANGELOG_PENDING.md
```diff
@@ -44,6 +44,9 @@
 - [sdk/nodejs] Allow prompt values in `construct` for multi-lang components.
   [#6522](https://github.com/pulumi/pulumi/pull/6522)
 
+- [automation/dotnet] Fix (de)serialization of StackSettings.
```
t0yv0, line 47: Would this be considered breaking? It will not automatically read users's files in the previous serialization? 

Ah I recall automation/dotnet is pre-GA so this may be ok.

t0yv0 (approved):

CHANGELOG_PENDING.md
komalali: I this this was just broken prior to this fix because pulumi wouldn't be able to use the incorrectly cased keys.

This should be under Bug Fixes though. Might be worth just lumping all of nodejs/python/dotnet under one changelog entry.

leezen: I'll merge this PR, rebase the other PR I have open, and consolidate all the entries into one.

PR #6765, opened 2021-04-13
[codegen/python] Remove deprecated __name__ and __opts__ args

justinvp:
These have been deprecated for a very long time and it's a trivial change to remove them from the generated code. Let's clean this up for the 3.0-based providers.


leezen (approved):
I would be surprised if anyone is using these, so I'm happy to see these go.

komalali (approved):

PR #6772, opened 2021-04-14
Merging feature-3.0 branch back to master

stack72:
This branch includes all the breaking changes for the next major release. These include:

* Changing the Go codegen to remove Apply<T>
* Bump sdk and pkg to v3
* Moving the automation api out of the "x" namespace
* Dropping support for < NodeJS 11.x
* Refactoring Mocks to use MocksRuntimeArgs & MocksCallArgs


stack72:
I expect downstream docs generation to fail as the downstream repos don't use the new v3 sdk endpoints

stack72:
Ok PR all green! Merging so we can move forward!

Issue #5506, opened 2020-10-05
Structured configuration not supported in Automation API


ringods:
## Expected Behavior

I need a programmatic counterpart of structured configuraiton.

## Current Behavior

I can only set a single value (`SetConfig`) or all values at once `SetAllConfig`), but structured config doesn't seem to be supported at the moment

## Context (Environment)

* https://www.pulumi.com/docs/intro/concepts/config/#structured-configuration
* https://github.com/pulumi/pulumi/blob/3be1ea9592b305df8aeb83c3810aee9bb8adeb7a/sdk/go/x/auto/local_workspace.go#L166-L184

## Support

- [x] [Go](https://github.com/pulumi/pulumi/pull/12265)
- [x] https://github.com/pulumi/pulumi/pull/13052
- [x] https://github.com/pulumi/pulumi/pull/14305
- [x] https://github.com/pulumi/pulumi-dotnet/pull/191

EvanBoyle:
As a workaround, is it possible to read the entire top-level config entry, change the appropriate property, and then write it back?

ringods:
@EvanBoyle I don't understand what you hint here.

There is no config yet in my case: I developped my stack with structured config and now using the automation api rolling out customer specific versions of that stack. This means I have a Git workspace, upserting the stack after configuring it.

EvanBoyle:
So you want to set a leaf value: 

```
SetConfig("a.b.c.d", "foo");
```

This is equivalent to:

```
val = {
b: {
c: {
d: "foo"
}
}
}

SetConfig("a", toJSON(val))
```

alexeyzimarev:
If the config object has one of the leaf as a secret, I would assume it won't work.

alexeyzimarev:
Oh, I see, the whole object is set as secret if one value is a secret value.

tmccombs:
> So you want to set a leaf value:
> 
> SetConfig("a.b.c.d", "foo");
> 
> This is equivalent to:
> 
> val = {
> b: {
> c: {
> d: "foo"
> }
> }
> }
> 
> SetConfig("a", toJSON(val))
> 

It doesn't look like that works.   With the equivalent (I think) code:
``` 
await stack.setConfig("a", {
      value: JSON.stringify({
        b: {
          c: {
            d: "foo",
          },
        },
      }),
    });

```
I get config that looks like:

```  
  my-stack:a: '{"b":{"c":{"d":"foo"}}}'
```

[...]

PR #6776, opened 2021-04-15
[automation/python] - Fix stack settings serialization

komalali:
Marking this no-changelog-required because we're in a weird in-between state and this will go out as a patch fix so the changelog will be added during release.

Changelog entry:

Bugfix:

- [automation/python] - Fix serialization bug in `StackSettings`

### DIFF
--- a/sdk/python/lib/pulumi/automation/_stack_settings.py
+++ b/sdk/python/lib/pulumi/automation/_stack_settings.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Optional, Mapping, Any, Union, Dict
+from typing import Optional, Any, Dict
 
 
 class StackSettingsSecureConfigValue:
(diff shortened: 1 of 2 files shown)


stack72:
This is going into master and then being cherry picked to a patch of v2.25.0 so we don’t need the changelog specifically in master right now

stack72 (approved):

sdk/python/lib/pulumi/automation/_stack_settings.py
```diff
@@ -28,7 +28,7 @@ class StackSettings:
     secrets_provider: Optional[str]
     encrypted_key: Optional[str]
     encryption_salt: Optional[str]
-    config: Optional[Mapping[str, Any]]
+    config: Optional[Dict[str, Union[str, StackSettingsSecureConfigValue]]]
```
leezen, line 31: Shouldn't this technically be `Any` though since we can have nested objects in the config?

komalali: For now, in automation API, we don't unmarshal objects and arrays from config, and they are read in as strings. There's an [open issue](https://github.com/pulumi/pulumi/issues/5506) tracking this. So, this matches the current reality and the other SDKs.

leezen: Ah, right. Thanks for clarifying.

leezen (approved):

sdk/python/lib/pulumi/automation/_stack_settings.py
komalali: Actually, you're right, kinda. I was thinking of `pulumi config` commands in my earlier answer. 

When the settings are parsed from json/yaml, in python the nested objects will parse to dicts and lists. So we should be recursing through if it's an object or a list to find the "secure" and regular config values. That means it's not just the type that's wrong, but also the serde for nested stack settings config in general 😞 

Gonna have to look at this with fresh eyes in the morning

komalali: Okay, I updated the types. The deserialization is still not perfect but the issue falls under the umbrella covered in #5506.

PR #6784, opened 2021-04-15
Clean the template dir if the remote has changed

svangordon-fruit:
If the remote of the `~/.pulumi/templates` repo is different than the remote provided through the `workspace.pulumiTemplateGitRepository`, remove that directory. Without this change, if the user has templates cached locally (extremely likely), the client will attempt to clone the directory in `~/.pulumi`; when that fails because the dir already exists, it will drop into the `templates` dir and do a pull. This results in the client continuing to use the previous templates source, instead of the new one set by `pulumiTemplateGitRepository`.

This change makes https://github.com/pulumi/pulumi/pull/6545 usable.

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -53,3 +53,6 @@
   [#6771](https://github.com/pulumi/pulumi/pull/6771)
 
 ### Bug Fixes
+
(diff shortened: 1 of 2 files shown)


justinvp (approved):
LGTM, thanks!

PR #6790, opened 2021-04-16
[sdk/go] Support prompt values in `Construct`

### LINKED ISSUE #6425: Support prompt values for multi-lang component inputs

justinvp:
- Adds support for prompt values.
- Renames `ConstructInputs.SetArgs` => `ConstructInputs.CopyTo` (improved name, IMO)

Fixes #6425

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,6 +57,7 @@ test_build:: $(SUB_PROJECTS:%=%_install)
 	cd tests/integration/construct_component/testcomponent-go && go build -o pulumi-resource-testcomponent
 	cd tests/integration/construct_component_slow/testcomponent && yarn install && yarn link @pulumi/pulumi && yarn run tsc
 	cd tests/integration/construct_component_plain/testcomponent && yarn install && yarn link @pulumi/pulumi && yarn run tsc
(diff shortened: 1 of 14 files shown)


justinvp:
I forgot to include the integration test in the PR, so I've pushed a commit that does that.

@pgavlin, I'm planning to merge this later tonight or Sunday morning to ensure it's included in the release. I'm happy to address any review feedback subsequently.

PR #6793, opened 2021-04-16
[automation/dotnet] Use Grpc.AspNetCore.Server package - without Grpc.Tools dependency

### LINKED ISSUE #6792: Cannot use Pulumi.Automation from F# & Paket
The Pulumi Automation api can't be referenced from F# using Paket because the Pulumi.Automation package references Grpc.AspNetCore, which has a dependency on Grpc.Tools, which refuses to build under anything other than a .csproj.

If the package reference in Pulumi.Automation.csproj was changed to Grpc.AspNetCore.Server, then Pulumi.Automation would still build but without Grpc.Tools as a dependency so the F# project can build.

It looks like the reason it's sepecific to paket is that it automatically includes all the targets files for the packages and their dependencies.

## Expected behavior
```
dotnet build
```
should work

## Current behavior
```
C:\Users\Mark\.nuget\packages\grpc.tools\2.37.0\build\_protobuf\Google.Protobuf.Tools.targets(51,5): error : Google.Protobuf.Tools proto compilation is only supported by default in a C# project (extension .csproj) 
```

## Steps to reproduce
```
mkdir \temp\pulumiaut
cd \temp\pulumiaut
pulumi new fsharp
dotnet new tool-manifest
dotnet tool install paket
dotnet paket convert-from-nuget
dotnet paket add Pulumi.Automation --project pulumiaut
dotnet restore
dotnet build
```
## Context (Environment)
Trying to experiment with the Pulumi.Automation package

## Affected feature

.net automation sdk

marklam:
Fixes #6792

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -59,4 +59,7 @@
 - [sdk/go] Support defining remote components in Go.
   [#6403](https://github.com/pulumi/pulumi/pull/6403)
 
+- [automation/dotnet] Remove dependency on Gprc.Tools for F# / Paket compatibility
+  [#6793](https://github.com/pulumi/pulumi/pull/6793)  
+
 ### Bug Fixes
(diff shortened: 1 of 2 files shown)


t0yv0 (approved):
Thanks for this, looking good.

PR #6801, opened 2021-04-17
[docgen] Fix horizontal scroll in python resource docs

justinvp:
Add line breaks and whitespace to avoid long horizontal scrolls for Python constructor/function arguments. Also, include the new ResourceArgs constructor overload.

## Before

Screen Shot 2021-04-17 at 12 07 17 PM

## After

Screen Shot 2021-04-17 at 12 07 33 PM

Fixes https://github.com/pulumi/docs/issues/5365

### DIFF
--- a/pkg/codegen/docs/gen.go
+++ b/pkg/codegen/docs/gen.go
@@ -192,11 +192,17 @@ type docNestedType struct {
 
 // propertyType represents the type of a property.
 type propertyType struct {
-	DisplayName string
-	Name        string
+	DisplayName     string
+	DescriptionName string // Name used in description list.
+	Name            string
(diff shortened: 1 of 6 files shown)


stack72:
Thank you so much!! This is an amazing improvement

stack72 (approved):
Do we need to make any changes to the bridge for this to take effect?

justinvp:
> Do we need to make any changes to the bridge for this to take effect?

I don't think so. I was able to regenerate the docs locally from the docs repo for some TF-based providers with `$ ./scripts/gen_all_resource_docs.sh` without any changes to the bridge.

stack72:
Even better :)

PR #6811, opened 2021-04-19
Distinguish between inputty and plain args types.

### LINKED ISSUE #6796: [codegen] Regression referencing plain types
Regression introduced by https://github.com/pulumi/pulumi/pull/6686

The following schema:

```json
{
    "name": "apigateway",
    "types": {
        "apigateway:index:EventHandlerRoute": {
            "type": "object",
            "properties": {
                "path": {
                    "type": "string"
                },
                "method": {
                    "type": "string"
                },
                "function": {
                    "$ref": "/aws/v3.30.0/schema.json#/resources/aws:lambda%2Ffunction:Function"
                }
            }
        }
    },
    "resources": {
        "apigateway:index:RestAPI": {
            "isComponent": true,
            "inputProperties": {
                "routes": {
                    "type": "array",
                    "items": {
                        "$ref": "#/types/apigateway:index:EventHandlerRoute"
                    }
                }
            },
            "requiredInputs": [
                "routes"
            ],
            "plainInputs": [
                "routes"
            ],
            "properties": {
                "url": {
                    "type": "string"
                },
                "api": {
                    "$ref": "/aws/v3.30.0/schema.json#/resources/aws:apigateway%2FrestApi:RestApi"
                }
            },
            "required": [
                "url",
                "api"
            ]
        }
    },
    "language": {
        "csharp": {
            "packageReferences": {
                "Pulumi": "2.*",
                "Pulumi.Aws": "3.*"
            }
        },
        "go": {
            "importBasePath": "github.com/pulumi/pulumi-apigateway/sdk/go/apigateway"
        },
        "nodejs": {
            "dependencies": {
                "@pulumi/aws": "^3.30.0"
            },
            "devDependencies": {
                "typescript": "^3.7.0"
            }
        },
        "python": {
            "requires": {
                "pulumi": ">=2.21.2,<3.0.0",
                "pulumi-aws": ">=3.30.0,<4.0.0"
            }
        }
    }
}

```

Results in the following generate code with compilation errors:

`sdk/nodejs/types/input.ts`:

```typescript
export interface EventHandlerRouteArgs {
    function?: pulumi.Input<aws.lambda.Function>;
    method?: pulumi.Input<string>;
    path?: pulumi.Input<string>;
}
```

`sdk/nodejs/restAPI.ts`:

```typescript
export interface RestAPIArgs {
    readonly routes: inputs.EventHandlerRoute[];
}
```

The `RestAPIArgs` is referring to `inputs.EventHandlerRoute`, but the type is `EventHandlerRouteArgs`.

If I add the `"compatibility": "tfbridge20",` flag to the schema, then the type in `input.ts` is `EventHandlerRoute`.

But we should generate compilable code without the compat flag.

Note: `EventHandlerRoute` is a type in the schema. It is referenced by the `routes` input property of the `RestAPI` resource in the schema. The `routes` input property is a plain input.

pgavlin:
These changes fix a regression introduced by #6686 that caused the SDK
code generators for .NET, Python, and Typescript to omit definitions for
plain object types. This regression occurred because #6686 drew a
clearer line between types used as resource arguments and types used
as function arguments, but conflated "resource arguments" with "inputty
types". This caused the code generators to generate inputty types for
any types used as resource arguments, even those that are used for
plainly-typed properties.

Fixes #6796.

### DIFF
--- a/pkg/codegen/dotnet/gen.go
+++ b/pkg/codegen/dotnet/gen.go
@@ -54,6 +54,7 @@ type typeDetails struct {
 	inputType    bool
 	stateType    bool
 	argsType     bool
+	plainType    bool
 	functionType bool
 }
 
@@ -223,7 +224,7 @@ func (mod *modContext) typeName(t *schema.ObjectType, state, input, args bool) s
 	switch {
 	case input:
 		return name + "Args"
-	case mod.details(t).functionType:
+	case mod.details(t).plainType:
 		return name + "Result"
 	}
 	return name
@@ -425,7 +426,7 @@ type plainType struct {
 func (pt *plainType) genInputProperty(w io.Writer, prop *schema.Property, indent string) {
 	wireName := prop.Name
 	propertyName := pt.mod.propertyName(prop)
-	propertyType := pt.mod.typeString(prop.Type, pt.propertyTypeQualifier, true, pt.state, pt.args && !prop.IsPlain, pt.args, false, !prop.IsRequired)
+	propertyType := pt.mod.typeString(prop.Type, pt.propertyTypeQualifier, true, pt.state, pt.args && !prop.IsPlain, pt.args && !prop.IsPlain, false, !prop.IsRequired)
 
 	// First generate the input attribute.
 	attributeArgs := ""
@@ -1207,33 +1208,12 @@ func (mod *modContext) genEnum(w io.Writer, enum *schema.EnumType) error {
 	return nil
 }
 
-func visitObjectTypesAcc(t schema.Type, visitor func(*schema.ObjectType), visited codegen.Set) {
-	if visited.Has(t) {
-		return
-	}
-	visited.Add(t)
-
-	switch t := t.(type) {
-	case *schema.ArrayType:
-		visitObjectTypesAcc(t.ElementType, visitor, visited)
-	case *schema.MapType:
-		visitObjectTypesAcc(t.ElementType, visitor, visited)
-	case *schema.ObjectType:
-		for _, p := range t.Properties {
-			visitObjectTypesAcc(p.Type, visitor, visited)
+func visitObjectTypes(properties []*schema.Property, visitor func(*schema.ObjectType, bool)) {
+	codegen.VisitTypeClosure(properties, func(t codegen.Type) {
+		if o, ok := t.Type.(*schema.ObjectType); ok {
+			visitor(o, t.Plain)
 		}
-		visitor(t)
-	case *schema.UnionType:
-		for _, e := range t.ElementTypes {
-			visitObjectTypesAcc(e, visitor, visited)
-		}
-	}
-}
-
-func visitObjectTypes(t schema.Type, visitor func(*schema.ObjectType)) {
-	// Accumulator to avoid visiting the same node twice in case of recursive types.
-	visited := codegen.Set{}
-	visitObjectTypesAcc(t, visitor, visited)
+	})
 }
 
 func (mod *modContext) genType(w io.Writer, obj *schema.ObjectType, propertyTypeQualifier string, input, state, args bool, level int) error {
@@ -1249,7 +1229,7 @@ func (mod *modContext) genType(w io.Writer, obj *schema.ObjectType, propertyType
 
 	if input {
 		pt.baseClass = "ResourceArgs"
-		if mod.details(obj).functionType {
+		if !args && mod.details(obj).plainType {
 			pt.baseClass = "InvokeArgs"
 		}
 		return pt.genInputType(w, level)
@@ -1614,7 +1594,7 @@ func (mod *modContext) gen(fs fs) error {
 					return err
 				}
 			}
-			if mod.details(t).functionType {
+			if mod.details(t).plainType {
 				if err := mod.genType(buffer, t, "Inputs", true, false, false, 1); err != nil {
 					return err
 				}
@@ -1648,7 +1628,7 @@ func (mod *modContext) gen(fs fs) error {
 			fmt.Fprintf(buffer, "}\n")
 
 			suffix := ""
-			if (mod.isTFCompatMode() || mod.isK8sCompatMode()) && mod.details(t).functionType {
+			if (mod.isTFCompatMode() || mod.isK8sCompatMode()) && mod.details(t).plainType {
 				suffix = "Result"
 			}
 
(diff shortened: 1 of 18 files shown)


pkg/codegen/dotnet/gen.go
```diff
@@ -1249,7 +1229,7 @@ func (mod *modContext) genType(w io.Writer, obj *schema.ObjectType, propertyType
 
 	if input {
 		pt.baseClass = "ResourceArgs"
-		if mod.details(obj).functionType {
+		if !args && mod.details(obj).plainType {
```
justinvp, line 1232: Something still isn't quite right with the .NET codegen...

This schema:

```json
{
    "name": "apigateway",
    "types": {
        "apigateway:index:EventHandlerRoute": {
            "type": "object",
            "properties": {
                "path": {
                    "type": "string"
                },
                "method": {
                    "type": "string"
                },
                "function": {
                    "$ref": "/aws/v3.30.0/schema.json#/resources/aws:lambda%2Ffunction:Function"
                }
            }
        }
    },
    "resources": {
        "apigateway:index:RestAPI": {
            "isComponent": true,
            "inputProperties": {
                "routes": {
                    "type": "array",
                    "items": {
                        "$ref": "#/types/apigateway:index:EventHandlerRoute"
                    }
                }
            },
            "requiredInputs": [
                "routes"
            ],
            "plainInputs": [
                "routes"
            ],
            "properties": {
                "url": {
                    "type": "string"
                },
                "api": {
                    "$ref": "/aws/v3.30.0/schema.json#/resources/aws:apigateway%2FrestApi:RestApi"
                }
            },
            "required": [
                "url",
                "api"
            ]
        }
    },
    "language": {
        "csharp": {
            "packageReferences": {
                "Pulumi": "2.*",
                "Pulumi.Aws": "3.*"
            }
        },
        "go": {
            "importBasePath": "github.com/pulumi/pulumi-apigateway/sdk/go/apigateway"
        },
        "nodejs": {
            "dependencies": {
                "@pulumi/aws": "^3.30.0"
            },
            "devDependencies": {
                "typescript": "^3.7.0"
            }
        },
        "python": {
            "requires": {
                "pulumi": ">=2.21.2,<3.0.0",
                "pulumi-aws": ">=3.30.0,<4.0.0"
            }
        }
    }
}
```

Is producing...

`sdk/dotnet/Inputs/EventHandlerRouteArgs.cs`:

```c#
namespace Pulumi.Apigateway.Inputs
{
    public sealed class EventHandlerRoute : Pulumi.InvokeArgs
    {
        [Input("function")]
        public Pulumi.Aws.Lambda.Function? Function { get; set; }

        [Input("method")]
        public string? Method { get; set; }

        [Input("path")]
        public string? Path { get; set; }

        public EventHandlerRoute()
        {
        }
    }
}
```

`sdk/dotnet/RestAPI.cs`:

```c#
namespace Pulumi.Apigateway
{
    public sealed class RestAPIArgs : Pulumi.ResourceArgs
    {
        [Input("routes", required: true)]
        private ImmutableArray<Inputs.EventHandlerRouteArgs>? _routes;
        public ImmutableArray<Inputs.EventHandlerRouteArgs> Routes
        {
            get => _routes ?? (_routes = new ImmutableArray<Inputs.EventHandlerRouteArgs>());
            set => _routes = value;
        }

        public RestAPIArgs()
        {
        }
    }
}
```

- Why is `EventHandlerRoute` inheriting from `InvokeArgs` and not `ResourceArgs`?
- `RestAPIArgs` is referring to `Inputs.EventHandlerRouteArgs` which doesn't exist.

justinvp: Also, nit: If we're generating a `EventHandlerRoute` type, I would expect it to be generated in `EventHandlerRoute.cs` not `EventHandlerRouteArgs.cs`.

pgavlin:
> Why is EventHandlerRoute inheriting from InvokeArgs and not ResourceArgs?

This is b/c the codegen for plain types is shared between function args and resource args. I was hoping that you or @mikhailshilkov could give me some guidance on this--is it acceptable for a resource argument type to derive from `InvokeArgs`? If not, we're going to end up with even more type duplication, as we'll have to generate two versions of each plain type that is used by both a function and a resource: the former would derive from `InvokeArgs` while the latter would derive from `ResourceArgs`.

> Also, nit: If we're generating a EventHandlerRoute type, I would expect it to be generated in EventHandlerRoute.cs not EventHandlerRouteArgs.cs.

Yeah, this is a bit of expediency on my part. I _think_ it should be relatively easy to emit the more idiomatically-named file.

pkg/codegen/dotnet/gen.go
pgavlin: > RestAPIArgs is referring to Inputs.EventHandlerRouteArgs which doesn't exist.

This should be fixed with the latest changes.

mikhailshilkov:
> is it acceptable for a resource argument type to derive from InvokeArgs?

I think it will work, since we don't have specific checks for `InvokeArgs` or `ResourceArgs` in our serialization code. But it does look confusing. Another option could be deriving from `InputArgs` directly - not sure what other effects of this would be yet.

mikhailshilkov:
The [diff](https://github.com/pulumi-bot/pulumi-azure-native/compare/72d9771cfa..835ce5ed92) for `azure-native` is disturbing... Is it intended that way?

mikhailshilkov:
> The diff for azure-native is disturbing... Is it intended that way?

The .NET part actually looks like a valid fix... What about Python?

pkg/codegen/internal/test/testdata/simple-plain-schema/dotnet/Component.cs
```diff
@@ -67,6 +73,17 @@ public sealed class ComponentArgs : Pulumi.ResourceArgs
...
         public bool? B { get; set; }
 
+        [Input("bar")]
+        public Inputs.Foo? Bar { get; set; }
+
+        [Input("baz")]
+        private ImmutableArray<Inputs.Foo>? _baz;
+        public ImmutableArray<Inputs.Foo> Baz
+        {
+            get => _baz ?? (_baz = new ImmutableArray<Inputs.Foo>());
```
justinvp, line 83: This has uncovered an orthogonal bug. (I'll open a separate issue and work on a fix). Compiler error:

```
Cannot implicitly convert type 'System.Collections.Immutable.ImmutableArray<Pulumi.Example.Inputs.Foo>?' to 'System.Collections.Immutable.ImmutableArray<Pulumi.Example.Inputs.Foo>'. An explicit conversion exists (are you missing a cast?)
```

I think because `System.Collections.Immutable.ImmutableArray<T>` is a struct.

justinvp: https://github.com/pulumi/pulumi/issues/6813

justinvp (approved):
LGTM -- it addresses all the known issues I reported. I have not verified the diffs of any of the core providers, though.

PR #6812, opened 2021-04-19
Do not wrap types in Input<> when generating docs.

### LINKED ISSUE #6797: [docgen] Regression in args types
Regression introduced by #6686

When regenerating resource docs using the latest commit in `master`, I'm seeing diffs on args types. Previously, we were emitting unwrapped types (by design, as to not clutter the table of args), but now they're being emitted wrapped as `Input`.

Screen Shot 2021-04-16 at 1 10 40 PM

pgavlin:
Just what it says on the tin. These changes also add more expansive
tests for documentation generation using our existing example schemas.

Fixes #6797.


pkg/codegen/internal/test/testdata/external-resource-schema/docs/cat.md
```diff
@@ -0,0 +1,449 @@
...
+{{% /choosable %}}
+
+{{% choosable language python %}}
+<div class="highlight"><pre class="chroma"><code class="language-python" data-lang="python"><span class=nd>@overload</span>
+<span class="k">def </span><span class="nx">Cat</span><span class="p">(</span><span class="nx">resource_name</span><span class="p">:</span> <span class="nx">str</span><span class="p">,</span>
+        <span class="nx">opts</span><span class="p">:</span> <span class="nx"><a href="/docs/reference/pkg/python/pulumi/#pulumi.ResourceOptions">Optional[ResourceOptions]</a></span> = None<span class="p">,</span>
+        <span class="nx">age</span><span class="p">:</span> <span class="nx">Optional[int]</span> = None<span class="p">,</span>
```
justinvp, line 28: The types in the signature here should be wrapped as `Optional[pulumi.Input[int]]`.

In the table of properties below, they should not be wrapped (this looks good now).

pkg/codegen/internal/test/testdata/external-resource-schema/docs/cat.md
```diff
@@ -0,0 +1,449 @@
...
+{{% /choosable %}}
+
+{{% choosable language nodejs %}}
+<dl class="resources-properties"><dt class="property-optional"
+            title="Optional">
+        <span id="age_nodejs">
+<a href="#age_nodejs" style="color: inherit; text-decoration: inherit;">age</a>
+</span>
+        <span class="property-indicator"></span>
+        <span class="property-type">number</span>
```
justinvp, line 229: 👍 This looks correct now (should not be wrapped here).

pkg/codegen/internal/test/testdata/external-resource-schema/docs/cat.md
```diff
@@ -0,0 +1,449 @@
...
+{{% /choosable %}}
+
+{{% choosable language python %}}
+<dl class="resources-properties"><dt class="property-optional"
+            title="Optional">
+        <span id="age_python">
+<a href="#age_python" style="color: inherit; text-decoration: inherit;">age</a>
+</span>
+        <span class="property-indicator"></span>
+        <span class="property-type">int</span>
```
justinvp, line 249: 👍 Also correct here (should not be wrapped in this table).

pkg/codegen/internal/test/testdata/simple-plain-schema/docs/provider.md
```diff
@@ -0,0 +1,265 @@
...
+<a href="#id_python" style="color: inherit; text-decoration: inherit;">id</a>
+</span>
+        <span class="property-indicator"></span>
+        <span class="property-type">str</span>
+    </dt>
+    <dd>{{% md %}}The provider-assigned unique ID for this managed resource.{{% /md %}}</dd></dl>
+{{% /choosable %}}
+
+
+
```
viveklak, line 252: Minor nit: seems like a bunch of additional empty lines are being rendered. Probably just an existing template rendering bug in which case we can fix that separately.

viveklak (approved):

pkg/codegen/internal/test/testdata/external-resource-schema/docs/cat.md
pgavlin: > The types in the signature here should be wrapped as Optional[pulumi.Input[int]].

Were they wrapped before?

PR #6821, opened 2021-04-20
[automation/dotnet] Fix EventLogWatcher failing to read events after an exception was thrown

vipentti:
Discovered this while trying to preview some code which threw an exception. Instead of propagating the actual exception to the caller, the following exception was thrown

```
System.IO.IOException : The process cannot access the file '<temp>\automation-logs-preview-bc4fcvtm.ayq\eventlog.txt' 
because it is being used by another process.
```

This error seems to only occur on Windows.


t0yv0 (approved):
Awesome thanks for this!

I wonder what's up with this tests not failing on CI or maybe it's not running on CI? Checking.

vipentti:
@t0yv0 The problem only occurs when event-log is used, which by default only happens during preview, which was not previously used in the error propagation test cases so I think that's why it was not happening in CI.

t0yv0:
Gotcha so we needed the test change to repro. Awesome. Thanks again for this.

vipentti:
The problem should be reproducible by removing the `FileShare.ReadWrite` and running the tests on a Windows machine.

PR #6827, opened 2021-04-21
Import subpackages lazily

t0yv0:
This work is motivated by https://github.com/pulumi/pulumi-azure-native/issues/738 slow loading of Python programs referencing pulumi_azure_native SDK.

Somewhat naive measurements of a simple python program `speed_test.py`:

```python
import base64
from pulumi import Config, Output, export
import pulumi_azure_native.compute as compute
import pulumi_azure_native.network as network
import pulumi_azure_native.resources as resources
```

Speed of import loader:

```
Before:

time python speed_test.py
python speed_test.py  23.66s user 1.41s system 98% cpu 25.417 total

After:

time python speed_test.py

real	0m0.816s
user	0m0.722s
sys	0m0.091s
```

So we go from 23s eager loading to 0.8s lazy-loading. I will need to measure actual `pulumi up` before we can analyze these numbers, but there is hope that typical Pulumi programs will not load all the sub-models eagerly and will retain some of the win.

Marking this as draft because there is an unresolved problem with module registrations that used to be eager. At least this kind of issue is possible with double-init of modules:

```
>>> import pulumi_azure_native.network
>>> dir(pulumi_azure_native.network.v20200701)
  File "<stdin>", line 1, in <module>
  File "/nix/store/cp9nlaz4a24n8sp9f738ngdnrd19s4rp-python3-3.8.5/lib/python3.8/importlib/util.py", line 245, in __getattribute__
    self.__spec__.loader.exec_module(self)
  File "<frozen importlib._bootstrap_external>", line 783, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/Users/anton/pulumi-azure-native/sdk/python/pulumi_azure_native/network/v20200701/__init__.py", line 324, in <module>
    _register_module()
  File "/Users/anton/pulumi-azure-native/sdk/python/pulumi_azure_native/network/v20200701/__init__.py", line 322, in _register_module
    pulumi.runtime.register_resource_module("azure-native", "network/v20200701", _module_instance)
  File "/Users/anton/pulumi-azure-native/sdk/python/venv/lib/python3.8/site-packages/pulumi/runtime/rpc.py", line 847, in register_resource_module
    raise ValueError(f"Cannot re-register module {key}@{module.version()}. Previous registration was {existing}, new registration was {module}.")
ValueError: Cannot re-register module azure-native:network/v20200701@0.0.1. Previous registration was <pulumi_azure_native.network.v20200701._register_module.<locals>.Module object at 0x10e5d81f0>, new registration was <pulumi_azure_native.network.v20200701._register_module.<locals>.Module object at 0x10e2621c0>.
```

Also, perhaps something will break unless we eagerly register all modules?

### DIFF
--- a/pkg/codegen/python/gen.go
+++ b/pkg/codegen/python/gen.go
@@ -97,6 +97,7 @@ type modContext struct {
 	functions            []*schema.Function
 	typeDetails          map[*schema.ObjectType]*typeDetails
 	children             []*modContext
+	parent               *modContext
 	snakeCaseToCamelCase map[string]string
 	camelCaseToSnakeCase map[string]string
 	tool                 string
@@ -106,6 +107,11 @@ type modContext struct {
 	// Name overrides set in PackageInfo
 	modNameOverrides map[string]string // Optional overrides for Pulumi module names
 	compatibility    string            // Toggle compatibility mode for a specified target.
+}
+
+func (mod *modContext) addChild(child *modContext) {
+	mod.children = append(mod.children, child)
+	child.parent = mod
 }
 
 func (mod *modContext) details(t *schema.ObjectType) *typeDetails {
@@ -498,10 +504,37 @@ func (mod *modContext) submodulesExist() bool {
 	return len(mod.children) > 0
 }
 
+func (mod *modContext) unqualifiedImportName() string {
+	name := mod.mod
+
+	// Extract version suffix from child modules. Nested versions will have their own __init__.py file.
+	// Example: apps/v1beta1 -> v1beta1
+	parts := strings.SplitN(name, "/", 2)
+	if len(parts) == 2 {
+		name = parts[1]
+	}
+
+	return PyName(name)
+}
+
+func (mod *modContext) fullyQualifiedImportName() string {
+	name := mod.unqualifiedImportName()
+	if mod.parent == nil {
+		if name == "" {
+			return pyPack(mod.pkg.Name)
+		} else {
+			return fmt.Sprintf("%s.%s", pyPack(mod.pkg.Name), name)
+		}
+	} else {
+		return fmt.Sprintf("%s.%s", mod.parent.fullyQualifiedImportName(), name)
+	}
+}
+
 // genInit emits an __init__.py module, optionally re-exporting other members or submodules.
 func (mod *modContext) genInit(exports []string) string {
 	w := &bytes.Buffer{}
 	mod.genHeader(w, false /*needsSDK*/, nil)
+	fmt.Fprintf(w, "%s\n", mod.genUtilitiesImport())
 
 	// Import anything to export flatly that is a direct export rather than sub-module.
 	if len(exports) > 0 {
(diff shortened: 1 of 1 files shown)


t0yv0:
Interesting after 1590648 which uses official Python example of lazy_import that's subtly different from the code before, I can't repro the double-import issue. Still question remains if we need to eagerly register the child modules.

viveklak:
Adding @justinvp and @komalali for python expertise.

mikhailshilkov:
This sounds awesome! Does it have a similar effect for `import pulumi_azure_native as azure_native` that we show in all our docs? Or do folks need to import per module?

t0yv0:
Yeah it does! It affects everything. I will work on getting some more measurements out here.

justinvp:
> Still question remains if we need to eagerly register the child modules.

We are going to need these registrations to support deserializing resource references. However, I have wondered if we could do these registrations lazily as well. Register some kind of registration callback instead which wouldn't be run until an actual resource reference needs to be deserialized for a particular package & module. This would make these registrations more pay for play.

t0yv0:
I'm working on a patch to do registrations semi-lazily, stay tuned.

t0yv0:
This passes tests (I think - there was one SKIP, rerunning) and ready for review now.

komalali (commented):
@t0yv0 I'm curious if these changes have any effect on autocomplete in PyCharm/IntelliJ and VSCode - do you have any data points there?

t0yv0:
> @t0yv0 I'm curious if these changes have any effect on autocomplete in PyCharm/IntelliJ and VSCode - do you have any data points there?

Not in great position to test unfortunately. I would hope there is no effect if they respect the `if typing.*` directive that flags some of the code for typecheck-only context.

justinvp:
> > @t0yv0 I'm curious if these changes have any effect on autocomplete in PyCharm/IntelliJ and VSCode - do you have any data points there?
> 
> Not in great position to test unfortunately at the moment. I would hope there is no effect if they respect the `if typing.*` directive that flags some of the code for typecheck-only context.

I am curious about this as well. Planning to try this out later today and I'll report back.

justinvp:
I tried this out with VS Code (w/ Pylance). Unfortunately, the submodules don't show up in the code completion list when doing `import pulumi_aws` and `pulumi_aws.`...

## Before this change

Screen Shot 2021-04-27 at 4 55 48 PM

## After this change

Screen Shot 2021-04-27 at 4 54 53 PM

Maybe it isn't the end of the world since accessing the submodules this way is a little unusual for Python. When doing the more regular thing of `import pulumi_aws.` it does show the modules as one would expect:

Screen Shot 2021-04-28 at 8 42 55 AM

Trying PyCharm now.

justinvp:
PyCharm looks good. These are with the changes:

Screen Shot 2021-04-28 at 8 47 34 AM

Screen Shot 2021-04-28 at 8 47 58 AM

t0yv0:
Thanks for checking that, that's good to know! I can play with Pylance. It seems it won't support `if typing.TYPE_CHECKING` directive but maybe it has some other way that it can learn about our exports.

t0yv0:
This is really weird. So it turns out pylance is supported by pyright, and these things respect the `if typing.TYPE_CHECKING` declaration fine. On a small example I could not reproduce the issue. On a full AWS example, I can reproduce the issue, however note that if you type:

`pulumi_aws.s3.Bu` - it starts completing Bucket again, and it can correctly type-check files that reference Bucket and other resources that way.

justinvp (approved):
LGTM. The minor code completion regression is unfortunate, but I think the performance improvement outweighs the downside there. Nice work!

t0yv0:
Checking with @justinvp over Zoom today, Justin feels that benefits outweigh costs here. @komalali really would appreciate your opinion 🙏  Let me know if you consider this ready to go, or would like some more information or tests.

komalali (approved):
@t0yv0 that works for me, go for it 🚀 !

PR #6831, opened 2021-04-21
add a package type to dotnet libraries

jaxxstorm:


loic-sharma:
In hindsight, it looks I gave bad advice on Twitter. Custom package types currently breaks package installation on Visual Studio and nuget.exe (see https://github.com/NuGet/Home/issues/10468). I would wait until we fix that before moving forward with this. Sorry about that!

PR #6833, opened 2021-04-21
Fix bug in semver usage by enabling typechecking

t0yv0:
This bug actually prevents resources from hydrating that is:

```
get_resource_module('azure-native', 'authorization', '1.0.0')
```

Is broken..


justinvp (approved):
LGTM

Looks like we're missing some test coverage here.

komalali (approved):

t0yv0:
Taking a note, I'll follow up on where to add coverage.

PR #6834, opened 2021-04-21
Changing the logo reference in the DotNet Codegen to be the new Pulumi logo

stack72:


cnunciato (approved):

justinvp (approved):
LGTM

I did notice the new logo is 175x175 vs 64x64. Does it matter for nuget logos?

justinvp:
> I did notice the new logo is 175x175 vs 64x64. Does it matter for nuget logos?

To answer my own question:

https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#packing-an-icon-image-file

> We recommend an image resolution of 128x128.

So I'm guessing 175x175 would work just fine.

leezen (approved):

Issue #6809, opened 2021-04-19
Pulumi automation can not handle objectValue in Config


ncd:
I made pulumi config as an object.
```yaml
ram:
    roles:
    - assume_policy: autoscaler_assume
      name: cluster-autoscaler
      policy_attachments:
        custom_policies:
        - cluster_autoscaler
```
When I run stack refresh/up using pulumi.x.automation module --> I always failed due to an error: unexpected keyword argument 'objectValue'.
Recheck with pulumi stack history command:
```
"devops-ram:ram": {
        "value": "{\"roles\":[{\"assume_policy\":\"autoscaler_assume\",\"name\":\"cluster-autoscaler\",\"policy_attachments\":{\"custom_policies\":[\"cluster_autoscaler\"]}}]}",
        "objectValue": {
          "roles": [
            {
              "assume_policy": "autoscaler_assume",
              "name": "cluster-autoscaler",
              "policy_attachments": {
                "custom_policies": [
                  "cluster_autoscaler"
                ]
              }
            }
          ]
        },
        "secret": false
      }
    }
```

## Expected behavior

The code will run normally

## Current behavior

The code has error

## Steps to reproduce

1. Run pulumi automation with configuration is in object type.

leezen:
I believe this is a duplicate of https://github.com/pulumi/pulumi/issues/5506, which has some suggestions for potential workarounds.

komalali:
Hi, okay looking at this again I think I see what's going on here and it's not a dupe of #5506.

@ncd it looks like you're using automation API with python. Is that right? What version are you using?

komalali:
Okay, I just rechecked that we did indeed fix this issue in Pulumi 2.21.0 and we have tests for it, so I'm going to close this out. If you are using the latest version and are somehow still hitting this error, please reopen.

followben:
I'm striking the same issue.  I have a similar code & setup as [the local python example](https://github.com/pulumi/automation-api-examples/blob/main/python/local_program/automation/main.py) generates the following stacktrace on trying perform either the `refresh` or `up` operation on a stack:

```
Traceback (most recent call last):
  File "main.py", line 90, in <module>
    up_res = stack.up(on_output=print)
  File "/Users/bstovold/Documents/venv/lib/python3.8/site-packages/pulumi/automation/_stack.py", line 279, in up
    summary = self.info()
[...]

PR #6838, opened 2021-04-21
[automation/python] - Fix deserialization of UpdateSummary

komalali:
Deserialization of config values in UpdateSummary was erroring when the config value is a map or array. This fixes the issue.

Fixes: #6809

### DIFF
--- a/sdk/python/lib/pulumi/automation/_stack.py
+++ b/sdk/python/lib/pulumi/automation/_stack.py
@@ -88,7 +88,8 @@ class UpdateSummary:
         self.version = version
         self.config: ConfigMap = {}
         for key in config:
-            self.config[key] = ConfigValue(**config[key])
+            config_value = config[key]
+            self.config[key] = ConfigValue(value=config_value["value"], secret=config_value["secret"])
 
     def __repr__(self):
         return f"UpdateSummary(result={self.result!r}, version={self.version!r}, " \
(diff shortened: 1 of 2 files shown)


sdk/python/lib/pulumi/automation/_stack.py
```diff
@@ -88,7 +88,8 @@ def __init__(self,
         self.version = version
         self.config: ConfigMap = {}
         for key in config:
-            self.config[key] = ConfigValue(**config[key])
+            config_value = config[key]
+            self.config[key] = ConfigValue(value=config_value["value"], secret=config_value["secret"])
```
t0yv0, line 92: What was the error message before, did config[key] pass some unexpected parameters? Are "value" and "secret" fields required and always present? Especially "secret"? If it's optional consider `config_value.get("secret", None)`.

t0yv0 (approved):

sdk/python/lib/pulumi/automation/_stack.py
komalali: Yeah, both `secret` and `value` are always present and required. Merged so we could get this in 3.1. 

The error was `TypeError: __init__() got an unexpected keyword argument 'objectValue'` because when the config is a map or an array it includes `objectValue` which is the deserialized object. I think we probably ideally want to include that as a way of resolving #5506 - but for now for consistency with other SDKs, I've removed it.

t0yv0: That's the right change then, thank you!!

PR #6859, opened 2021-04-23
[automation/*] Add support for getting stack outputs using Workspace

vipentti:
This adds the ability to get the stack outputs using Workspace without having to first create or select a stack.

This was inspired by @komalali's awesome [self-service-platyform](https://github.com/komalali/self-service-platyform), specifically [these lines](https://github.com/komalali/self-service-platyform/blob/37244ca9bb3cffcba75aff077b87f2c3ced99db7/app/sites.py#L109-L117).

Regarding the Python implementation, since I'm less familiar with Python:

I'm not sure what is the correct way to avoid circular dependencies, specifically regarding `OutputMap`, so I figured the simplest way would be to move `OutputValue` and `OutputMap` to a separate file `_output.py` and then import them where necessary.

I'm not sure if this considered a breaking change, which I would like to avoid. Based on my understanding the `OutputValue` and `OutputMap` are not part of the public interface since they are not mentioned in `automation/__init__.py` variable `__all__`.


komalali:
Hey, @vipentti thanks for the PR!

Re: python - you're on the right track. 
Specifically, we have named the files starting with an underscore as this is considered "private" in python. We've done this so we can change the file structure without breaking the public API. The public API is exposed through `__all__` in `automation/__init__.py`. As for the fact that `OutputValue` and `OutputMap` are not included in `__all__` in `automation/__init__.py`, that is an oversight and they should have been included.

All of that to say:
* Moving `OutputMap` and `OutputValue` into `_output.py` is totally fine and non-breaking.
* We should also include them in `__all__` in `automation/__init__.py` (if you don't mind adding that on).

komalali:
Pylint errors in the test run are fixed in https://github.com/pulumi/pulumi/pull/6872

komalali:
LGTM in general, but would prefer naming it `StackOutputs` rather than `Outputs` when on `Workspace`

vipentti:
> We should also include them in __all__ in automation/__init__.py (if you don't mind adding that on).

I'll add `OutputMap` and `OutputValue` to `__all__`

> LGTM in general, but I would prefer naming it `StackOutputs` rather than `Outputs` when on `Workspace`

This sounds reasonable. So for go, python and nodejs I'll rename `Workspace.Outputs` as `Workspace.StackOutputs` and for the dotnet version `Workspace.GetStackOutputsAsync`

komalali (approved):

komalali (approved):

Issue #6861, opened 2021-04-23
Unable to mock StackReference in tests


clstokes:
It seems something changed in 3.0 that breaks when attempting to mock `StackReference` in a test.

## Current behavior

Tests fail with this error:

```
DEBUG    root:mocks.py:227 exception when preparing or executing rpc: Traceback (most recent call last):
  File "/home/user1/.local/share/virtualenvs/my-app/lib/python3.9/site-packages/pulumi/runtime/resource.py", line 376, in do_read
    resp = await asyncio.get_event_loop().run_in_executor(None, do_rpc_call)
  File "/usr/lib/python3.9/concurrent/futures/thread.py", line 52, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/user1/.local/share/virtualenvs/my-app/python3.9/site-packages/pulumi/runtime/resource.py", line 371, in do_rpc_call
    return monitor.ReadResource(req)
  File "/home/user1/.local/share/virtualenvs/my-app/python3.9/site-packages/pulumi/runtime/mocks.py", line 174, in ReadResource
    custom=request.custom or False)
AttributeError: custom
```

## Steps to reproduce

TBD

stack72:
Do you have any code that worked before that we can trace here?

komalali:
@stack72 the user provided an example in slack which I have used as a test case in the resolving PR: https://github.com/pulumi/pulumi/pull/6863

PR #6863, opened 2021-04-24
Fix python mock args

komalali:
The [refactor](https://github.com/pulumi/pulumi/pull/6672) to `MockResourceArgs` introduced a bug for `ReadResource` as the request does not include a `custom` attribute, leading to an error when mocking `StackReference`. 

This PR corrects the issue and adds a test for `StackReference`.

Fixes: #6861

### DIFF
--- a/sdk/python/lib/pulumi/runtime/mocks.py
+++ b/sdk/python/lib/pulumi/runtime/mocks.py
@@ -15,7 +15,6 @@
 """
 Mocks for testing.
 """
-import asyncio
 import logging
 from abc import ABC, abstractmethod
 from typing import Dict, List, NamedTuple, Optional, Tuple, TYPE_CHECKING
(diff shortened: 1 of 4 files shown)


stack72 (approved):

sdk/python/lib/pulumi/runtime/mocks.py
```diff
@@ -167,8 +173,7 @@ def ReadResource(self, request):
                                          name=request.name,
                                          inputs=state,
                                          provider=request.provider,
-                                         resource_id=request.id,
-                                         custom=request.custom or False)
```
t0yv0, line 171: request.custom is not propagated here - that is on purpose? Thanks.

komalali: Yeah, that's the fix actually. The [ReadResource](https://github.com/pulumi/pulumi/blob/master/sdk/python/lib/pulumi/runtime/resource.py#L347-L359) request does not contain a `custom` attribute.

t0yv0 (approved):

Issue #6711, opened 2021-04-06
[automation/*] - Introduce env var to bypass min version checking


komalali:
We need to add an environment variable to bypass the CLI minimum version checking used by automation API.

PR #6882, opened 2021-04-27
[automation/*] - Optionally skip Automation API version check

komalali:
This PR adds an option to skip the minimal CLI version check for Automation API . If `PULUMI_AUTOMATION_API_SKIP_VERSION_CHECK` has a non-empty value, the check is skipped.

Docs PR: https://github.com/pulumi/pulumi-hugo/pull/154

Fixes: #6711

### DIFF
--- a/sdk/go/auto/local_workspace.go
+++ b/sdk/go/auto/local_workspace.go
@@ -53,6 +53,8 @@ type LocalWorkspace struct {
 
 var settingsExtensions = []string{".yaml", ".yml", ".json"}
 
+var skipVersionCheckVar = "PULUMI_AUTOMATION_API_SKIP_VERSION_CHECK"
+
 // ProjectSettings returns the settings object for the current project if any
 // LocalWorkspace reads settings from the Pulumi.yaml in the workspace.
 // A workspace can contain only a single project at a time.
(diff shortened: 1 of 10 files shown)


stack72 (approved):
Implementation looks good! Nice work on adding this so fast

sdk/go/auto/local_workspace.go
```diff
@@ -499,6 +499,9 @@ func (l *LocalWorkspace) getPulumiVersion(ctx context.Context) (semver.Version,
 
 //nolint:lll
 func validatePulumiVersion(minVersion semver.Version, currentVersion semver.Version) error {
+	if os.Getenv("PULUMI_AUTOMATION_API_SKIP_VERSION_CHECK") != "" {
```
EvanBoyle, line 502: General question for all implementations, do we need to check workspace environment variables? I believe these can be specified on creation, not sure if this method should be aware of those but that is likely how a user would specify this, via code.

EvanBoyle (approved):
Overall looks good. One case with workspace envs that should probably be considered first. 

Also, I'm not sure if we've been keeping up with this (we should be) but would be worth adding a godoc example https://github.com/pulumi/pulumi/blob/master/sdk/go/auto/example_test.go

sdk/go/auto/local_workspace.go
komalali: Hmm... interesting point. The env vars that are passed to workspace just get passed on to the subprocesses for now - should we also be setting them on the current process? Wondering if that's what a user's expectation would be from those env vars. I know I could also pass that flag to this method but my question is more generally about all env vars passed to workspace.

Also, re: adding an example. I don't think this is something that I want to advertise excessively. I did add it to the [list of env vars](pulumi/pulumi-hugo#154) that we support and that seems sufficient for all languages.

EvanBoyle: I don't think we should set environment variables/mutate the current process. The current design is optimized for running pulumi commands, and this workspace check seems to be the exceptional case. I'd just also check to see if a a workspace level environment variable exists in addition to the process level one. 

Not advertising makes sense here as well.

PR #6890, opened 2021-04-27
[sdk/python] Address issues when using resource subclasses

### LINKED ISSUE #6850: Inputs not available as implicit Outputs on subclassed resources in Pulumi 3.0
Subclassed Resources appear to behave differently between Pulumi 2.x and 3.x. In particular, input properties on subclassed resources are not available as implicit output properties.

## Expected behavior
Input properties should be accessible as output properties on subclassed resources, as they are on the native Pulumi resources.

## Current behavior
Input values do not resolve to a usable value, causing failures in the creation of resources that depend on them. They do not show up as stack outputs, either.

## Steps to reproduce
I've created a reproduction at https://github.com/christophermaier/pulumi-3.0-subclass-bug, with further background information.

## Context (Environment)
I have several subclassed resources in my Pulumi code for restriction/specialization purposes (e.g., ensure all DynamoDB tables generated from a subclass have the same billing mode, etc.). While I could model this as a simple function call (as opposed to object creation), it is also useful to attach helper methods to the subclassed objects to make it easier to perform additional operations, like granting permissions to the resource from other resources (as with AWS Role Policies, for instance).

Such code worked without issue with Pulumi 2.x, but fails with 3.x, preventing us from upgrading (at least without rewriting our infrastructure code)

justinvp:
We were only looking at the current resource class's type/name metadata for camelCase <=> snake_case property name translations which prevented it from working correctly when using a subclass of a resource. This change addresses this by looking at metadata of the current class and any base classes.

Additionally, to help resolve forward references when getting type hints, we'd pass along the current resource class's globals, which doesn't work correctly when using a subclass of a resource. This change also addresses this, by using the globals of the current class and any base classes.

Fixes #6850

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -21,6 +21,9 @@
 - [automation/dotnet] Use stackName in ImportStack
   [#6858](https://github.com/pulumi/pulumi/pull/6858)
 
+- [sdk/python] Address issues when using resource subclasses.
+  [#6890](https://github.com/pulumi/pulumi/pull/6890)
+
 ### Misc.
 
 - [auto/*] - Bump minimum version to v3.1.0.
--- a/sdk/python/lib/pulumi/_types.py
+++ b/sdk/python/lib/pulumi/_types.py
@@ -407,13 +407,13 @@ def _create_py_property(a_name: str, pulumi_name: str, typ: Any, setter: bool = 
 
 
 def _py_properties(cls: type) -> Iterator[Tuple[str, str, builtins.property]]:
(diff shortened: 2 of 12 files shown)


komalali (approved):
LGTM 🚀

t0yv0:
Read up here a little bit, this is pretty scary stuff. My takeaway here is that `get_type_hints` from stdlib is underpowered and requires to do some non-trivial magic to resolve forward references. If improved get_type_hints was somewhere on PyPy that'd be ideal (instead of maintaining here) but there's nothing I could find.

PR #6892, opened 2021-04-27
[sdk/nodejs|python] Add GetSchema support to providers

### LINKED ISSUE #6777: Add getSchema support to multilang providers
Go is currently the only language that supports this. This is required for SDK codegen when a provider's schema references types in another multi-lang provider's schema.

justinvp:
Fixes #6777


sdk/nodejs/provider/server.ts
```diff
@@ -62,10 +62,13 @@ class Server implements grpc.UntypedServiceImplementation {
     }
 
     public getSchema(call: any, callback: any): void {
-        callback({
-            code: grpc.status.UNIMPLEMENTED,
-            details: "Not yet implemented: GetSchema",
-        }, undefined);
+        const req: any = call.request;
+        if (req.getVersion() !== 0) {
+            callback(new Error(`unsupported schema version ${req.getVersion()}`), undefined);
```
lukehoban, line 67: Do `Error`s bubble up correctly here?  Or is returning `UNIMPLEMENTED` in this case going to behave better?

sdk/nodejs/provider/server.ts
```diff
@@ -62,10 +62,13 @@ class Server implements grpc.UntypedServiceImplementation {
...
+        }
+        const resp: any = new provproto.GetSchemaResponse();
+        resp.setSchema(this.provider.schema || "{}");
+        callback(undefined, resp);
```
lukehoban, line 71: Is there any test coverage we can provide for this in this repo?

sdk/nodejs/provider/server.ts
justinvp: > Do `Error`s bubble up correctly here?

Yes, from a client's perspective this returns the same error as providers written in other languages and is the same error our providers return (e.g. [k8s](https://github.com/pulumi/pulumi-kubernetes/blob/0072954b2cdf088fc2e336ca4c289929f75ec1a5/provider/pkg/provider/provider.go#L188-L190), [tfbridge](https://github.com/pulumi/pulumi-terraform-bridge/blob/30ae8ae5cc2d2f5987854984243c74ace4ccc952/pkg/tfbridge/provider.go#L294-L296)).

justinvp:
Not sure what's going on with the test failure on Windows. Will look into it.

sdk/nodejs/provider/provider.ts
```diff
@@ -147,6 +147,11 @@ export interface Provider {
      */
     version: string;
 
+    /**
+     * The JSON-encoded schema for this provider's package.
```
t0yv0, line 151: This would be a great place to link to the meta-schema (the schema of the schema), if we have it.

t0yv0:
🚢 love having schema available. We can build a lot of useful things on top.

t0yv0 (approved):

Issue #6286, opened 2021-02-09
Not prompted for passphrase after first `up` when using local backend


clstokes:
## Expected Behavior

Pulumi should prompt users for their passphrase if `PULUMI_CONFIG_PASSPHRASE` and `PULUMI_CONFIG_PASSPHRASE_FILE` are not set when using any state backend.

## Current Behavior

If I initialize and `up` a stack when using the `--local` state backend, I'm not prompted for the passphrase after the first `up` and receive an error:

```
(venv) p-py-config-passphrase % pulumi up -y -f                   
error: decrypting secret value: failed to decrypt: incorrect passphrase, please set PULUMI_CONFIG_PASSPHRASE to the correct passphrase
```

### Important Notes

- this behavior does not occur if I don't export a secret from the Pulumi app.
- this behavior does not occur if I use the Pulumi Console backend.
- setting `PULUMI_CONFIG_PASSPHRASE_FILE` does not resolve this

## Steps to Reproduce

1. `pulumi login --local`
1. `export PULUMI_CONFIG_PASSPHRASE=1234567890`
1. `pulumi new python -y`
2. `pulumi stack init pphrase-2`
4. `pulumi config set --secret foo bar`
3. `echo 'import pulumi\nfoo = pulumi.config.Config().require_secret("foo")\npulumi.export("foo", foo)' > __main__.py`
4. `pulumi up -y -f`
5. `unset PULUMI_CONFIG_PASSPHRASE`
4. `pulumi up -y -f`
    - observe error `error: decrypting secret value: failed to decrypt: incorrect passphrase, please set PULUMI_CONFIG_PASSPHRASE to the correct passphrase`
1. `export PULUMI_CONFIG_PASSPHRASE=1234567890`
4. `pulumi up -y -f`
    - observe: completes successfully
5. `echo $PULUMI_CONFIG_PASSPHRASE > ./my-password.txt`
5. `export PULUMI_CONFIG_PASSPHRASE_FILE=./my-password.txt`
5. `unset PULUMI_CONFIG_PASSPHRASE`
4. `pulumi up -y -f`
    - observe: still fails even though `PULUMI_CONFIG_PASSPHRASE_FILE` is set

## Full output

```
(venv) p-py-config-passphrase % export PULUMI_CONFIG_PASSPHRASE=1234567890
(venv) p-py-config-passphrase % pulumi stack init pphrase-2
Created stack 'pphrase-2'
[...]

PR #6893, opened 2021-04-27
[cli] Ensure the user has set PULUMI_CONFIG_PASSPHRASE or PULUMI_CONFIG_PASSPHRASE_FILE

stack72:
Fixes: #6286

When a user is using the passphrase provider and unsets the environment
variables that let them interact with the secrets provider, then would
get an error as follows:

```
▶ pulumi up -y -f
error: decrypting secret value: failed to decrypt: incorrect passphrase, please set PULUMI_CONFIG_PASSPHRASE to the correct passphrase
```

We are oging to change this error message to make it more obvious
to a user what the error is and how they need to fix it

```
▶ pulumi up -y -f
error: constructing secrets manager of type "passphrase": unable to find either `PULUMI_CONFIG_PASSPHRASE` nor `PULUMI_CONFIG_PASSPHRASE_FILE` when trying to access the Passphrase Secrets Manager. Please ensure one of these values are set to allow the operation to continue
```

Ideally, we would like to prompt the user for the passphrase at this
point rather than error, but the CLI could be in the middle of an
update operation as the same codepath is used for reading stackreference
values

### DIFF
--- a/pkg/cmd/pulumi/stack_export.go
+++ b/pkg/cmd/pulumi/stack_export.go
@@ -90,13 +90,6 @@ func newStackExportCmd() *cobra.Command {
 			}
 
 			if showSecrets {
-				// Currently, the stack.DefaultSecretsProvider is cached so adding a call to getStackSecretsManager
-				// will ensure that the user has the correct credentials to decrypt the stack deployment
-				_, err := getStackSecretsManager(s)
-				if err != nil {
-					return errors.Wrap(err, "getting secrets manager")
-				}
-
 				snap, err := stack.DeserializeUntypedDeployment(deployment, stack.DefaultSecretsProvider)
 				if err != nil {
 					return checkDeploymentVersionError(err, stackName)
--- a/pkg/secrets/passphrase/manager.go
+++ b/pkg/secrets/passphrase/manager.go
@@ -18,7 +18,9 @@ package passphrase
 import (
 	"encoding/base64"
 	"encoding/json"
+	"io/ioutil"
 	"os"
+	"path/filepath"
 	"strings"
 	"sync"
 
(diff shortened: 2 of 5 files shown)


pkg/cmd/pulumi/stack_export.go
```diff
@@ -90,13 +90,6 @@ func newStackExportCmd() *cobra.Command {
 			}
 
 			if showSecrets {
-				// Currently, the stack.DefaultSecretsProvider is cached so adding a call to getStackSecretsManager
```
stack72, line 93: Since this is now going to error properly, we can remove this workaround we added in this specific usecase

justinvp (approved):

pkg/secrets/passphrase/manager.go
```diff
@@ -151,7 +153,35 @@ func NewPassphaseSecretsManagerFromState(state json.RawMessage) (secrets.Manager
...
+	}
+
+	// At this point, we don't know if it's an incorrect passphrase. We only know if there is a passphrase or there is
+	// not. Ideally, we would prompt the user for the passphrase at this point but we can't do that in the CLI is in an
+	// update operation, so we should at least error out with an appropriate message here to ensure that the user
+	// understands why the operation fails unexpectedly.
+	if phrase == "" {
+		return nil, errors.New("unable to find either `PULUMI_CONFIG_PASSPHRASE` nor " +
+			"`PULUMI_CONFIG_PASSPHRASE_FILE` when trying to access the Passphrase Secrets Manager. Please ensure one " +
+			"of these values are set to allow the operation to continue")
```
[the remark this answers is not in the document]
stack72: Nice change here - reads nicer! thanks @justinvp

pkg/secrets/passphrase/manager.go
```diff
@@ -151,7 +153,35 @@ func NewPassphaseSecretsManagerFromState(state json.RawMessage) (secrets.Manager
...
+		phrase = strings.TrimSpace(string(phraseDetails))
+	}
+
+	// At this point, we don't know if it's an incorrect passphrase. We only know if there is a passphrase or there is
+	// not. Ideally, we would prompt the user for the passphrase at this point but we can't do that in the CLI is in an
+	// update operation, so we should at least error out with an appropriate message here to ensure that the user
+	// understands why the operation fails unexpectedly.
+	if phrase == "" {
+		return nil, errors.New("unable to find either `PULUMI_CONFIG_PASSPHRASE` nor " +
+			"`PULUMI_CONFIG_PASSPHRASE_FILE` when trying to access the Passphrase Secrets Manager; please ensure one " +
```
lukehoban, line 182: The feature is called "secrets provider" right?  https://www.pulumi.com/docs/intro/concepts/secrets/#initializing-a-stack-with-alternative-encryption

pkg/secrets/passphrase/manager.go
```diff
@@ -151,7 +153,35 @@ func NewPassphaseSecretsManagerFromState(state json.RawMessage) (secrets.Manager
...
 	// called during an update when trying to read stack outputs as part servicing a StackReference request
 	// (since we need to decrypt the deployment)
-	phrase := os.Getenv("PULUMI_CONFIG_PASSPHRASE")
+	phrase := ""
+	passphrase, isOk := os.LookupEnv("PULUMI_CONFIG_PASSPHRASE")
+	if isOk {
+		phrase = passphrase
+	}
+	phraseFile, isOk := os.LookupEnv("PULUMI_CONFIG_PASSPHRASE_FILE")
+	if isOk {
```
lukehoban, line 162: Is it intentional that this branch runs even if the previous one found a passphrase?  This is slightly unusual structure for the logic here - and feels maybe not like the intended behaviour?

pkg/secrets/passphrase/manager.go
```diff
@@ -151,7 +153,35 @@ func NewPassphaseSecretsManagerFromState(state json.RawMessage) (secrets.Manager
...
+
+		phrase = strings.TrimSpace(string(phraseDetails))
+	}
+
+	// At this point, we don't know if it's an incorrect passphrase. We only know if there is a passphrase or there is
+	// not. Ideally, we would prompt the user for the passphrase at this point but we can't do that in the CLI is in an
+	// update operation, so we should at least error out with an appropriate message here to ensure that the user
+	// understands why the operation fails unexpectedly.
+	if phrase == "" {
+		return nil, errors.New("unable to find either `PULUMI_CONFIG_PASSPHRASE` nor " +
```
lukehoban, line 181: Either...or?  Not "nor" right?

stack72:
I've opened a further PR to address the comments by @lukehoban 

https://github.com/pulumi/pulumi/pull/6909

simonpinn:
Hi @justinvp - we've just upgraded to this version and despite setting ```export PULUMI_CONFIG_PASSPHRASE=""``` we face this error ```error: constructing secrets manager of type "passphrase": unable to find either `PULUMI_CONFIG_PASSPHRASE` or `PULUMI_CONFIG_PASSPHRASE_FILE` when trying to access the Passphrase Secrets Provider; please ensure one of these environment variables is set to allow the operation to continue```
Is an empty passphrase no longer supported?

lyenliang:
I got the same problem for not able to use pulumi because my `PULUMI_CONFIG_PASSPHRASE` was an empty string `""`.

stack72:
Hey @lyenliang / @simonpinn 

I just opened a Pull request that will fix this up - this was not intended to remove support for ""

You will still need to set PULUMI_CONFIG_PASSPHRASE as an "" but then it will work

Please stay tuned for v3.2.1 release

Paul

PR #6896, opened 2021-04-27
[sdk/nodejs] Warn when a secret config is read as a non-secret

### LINKED ISSUE #3139: Warn when secret config read as a non-secret
We currently let you silently read a secret configuration value -- that is, one set with `--secret` -- back as a non-secret -- that is, using plain old `Config.get` or `Config.require`. This is probably never what you actually wanted to do and makes it easy to accidentally leak a secret.

We actually need to explain this to people in our docs. For example:

image

I would expect most people would be surprised that we let them do this so easily.

I propose we consider either issuing a warning or, even better, rejecting this outright by throwing an exception (and maybe offering an `--enable-insecure-config-access` flag for back-compat).

Assigning to @lukehoban for triage.

---

- [ ] Node.js
- [ ] Python
- [ ] Go
- [ ] .NET

justinvp:
For example, if I have a config secret set with `pulumi config set foo --secret bar`, the following program:

```typescript

const config = new pulumi.Config();
const foo = config.require("foo");
```

... results in:

```
$ pulumi up
Previewing update (dev)

View Live: https://app.pulumi.com/<...>

     Type                 Name               Plan       Info
 +   pulumi:pulumi:Stack  configsecrets-dev  create     1 warning
 
Diagnostics:
  pulumi:pulumi:Stack (configsecrets-dev):
    warning: Configuration 'configsecrets:foo' value is a secret; use `requireSecret` instead of `require`
```

This is the Node.js change. Once we're happy with this, I'll follow-up with support in the other languages.

Part of #3139

---

Implementation Notes:

- For Pulumi programs, the engine passes the config to the language host over gRPC and then the language host passes the config to the Pulumi program as a serialized JSON object in the `PULUMI_CONFIG` environment variable. This change passes a list of secret config keys over gRPC and a new `PULUMI_CONFIG_SECRET_KEYS` environment variable containing a JSON serialized array of strings representing the config keys that contain secret values.

- When reading a config as a non-secret, if the key is within the set of secret keys, a warning is logged.


EvanBoyle (commented):
Will try to take a deeper look today, at a glance seems to make sense. One question: does this work for structured config? My understanding is that portions of a config value (certain key paths) can be marked as secret. Does that information come back from the engine in `configSecretKeys`?

justinvp:
> One question: does this work for structured config?

Yes. A nested secret means the entire object should be treated as a secret.

EvanBoyle:
I don't think that is completely true. A secret taints the entire object, but not non-secret leaf nodes within the object.

```sh
$ pulumi config set --path foo.bar abc --secret
$ pulumi config set --path foo.buzz def
$ pulumi config get foo --json
{
  "value": "{\"bar\":\"abc\",\"buzz\":\"def\"}",
  "objectValue": {
    "bar": "abc",
    "buzz": "def"
  },
  "secret": true
}
$ pulumi config get --path foo.bar --json
{
  "value": "abc",
  "secret": true
}
$ pulumi config get --path foo.buzz --json
{
  "value": "def",
  "secret": false
}

**Comment by justinvp** [comment]
> A secret taints the entire object, but not non-secret leaf nodes within the object.

Correct. But you can't access individual leaf nodes in a Pulumi program other than accessing the whole object.

File: sdk/nodejs/runtime/config.ts [hunk 1]
```diff
@@ -53,6 +59,21 @@ export function getConfig(k: string): string | undefined {
...
+/**
+ * isConfigSecret returns true if the key contains a secret value.
+ * @internal
+ */
+export function isConfigSecret(k: string): boolean {
+    const envConfigSecretKeys = process.env[configSecretKeysEnvKey];
+    if (envConfigSecretKeys) {
+        const envConfigSecretArray = JSON.parse(envConfigSecretKeys);
+        if (Array.isArray(envConfigSecretArray)) {
+            return envConfigSecretArray.includes(k);
```
EvanBoyle, line 71: Does this need to handle path keys like `foo.bar` or `foo[0]`?

EvanBoyle (approved):

sdk/nodejs/runtime/config.ts
justinvp: No, we don't have any APIs in the SDKs for accessing leaf nodes by path. Such values must be accessed via the top-level key with `getObject`, `requireObject`, `getSecretObject`, or `requireSecretObject`.

PR #6899, opened 2021-04-27
[automation/dotnet] Add PulumiFn.Create with service type overload

### LINKED ISSUE #6898: PulumiFn.Create should support `serviceType` overload
I'm trying to use the dotnet automation api to instantiate a stack via a service provider given the service type. i.e. like:
https://github.com/pulumi/pulumi/blob/eb2dcf24884bab173af175cd41d8279e07797a75/sdk/dotnet/Pulumi.Automation/PulumiFn.cs#L105-L106
but with a `serviceType` parameter:
```cs
public static PulumiFn Create<TStack>(IServiceProvider serviceProvider, Type serviceType)
    where TStack : Pulumi.Stack
```
This is fairly straightforward so I'll follow up with a PR.

gitfool:
Fixes #6898.

### DIFF
--- a/sdk/dotnet/Pulumi.Automation/PulumiFn.cs
+++ b/sdk/dotnet/Pulumi.Automation/PulumiFn.cs
@@ -104,6 +104,18 @@ namespace Pulumi.Automation
         /// <typeparam name="TStack">The <see cref="Pulumi.Stack"/> type.</typeparam>
         public static PulumiFn Create<TStack>(IServiceProvider serviceProvider)
             where TStack : Pulumi.Stack
+            => Create<TStack>(serviceProvider, typeof(TStack));
+
+        /// <summary>
+        /// Creates an inline (in process) pulumi program via a traditional <see cref="Pulumi.Stack"/> implementation.
+        /// <para/>
+        ///     When invoked, a new stack instance will be resolved based
+        ///     on the provided <typeparamref name="TStack"/> type parameter
+        ///     using the <paramref name="serviceProvider"/> and <paramref name="serviceType"/>.
+        /// </summary>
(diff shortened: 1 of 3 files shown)


sdk/dotnet/Pulumi.Automation/PulumiFn.cs
```diff
@@ -104,6 +104,18 @@ public static PulumiFn Create<TStack>()
...
+
+        /// <summary>
+        /// Creates an inline (in process) pulumi program via a traditional <see cref="Pulumi.Stack"/> implementation.
+        /// <para/>
+        ///     When invoked, a new stack instance will be resolved based
+        ///     on the provided <typeparamref name="TStack"/> type parameter
+        ///     using the <paramref name="serviceProvider"/> and <paramref name="serviceType"/>.
+        /// </summary>
+        /// <typeparam name="TStack">The <see cref="Pulumi.Stack"/> type.</typeparam>
+        public static PulumiFn Create<TStack>(IServiceProvider serviceProvider, Type serviceType)
```
orionstudt, line 117: Doesn't this still having a type argument give you the same issue when you want to dynamically pass in your stack type? Like you don't know the `TStack` at compile-time which is why you want `serviceType`?

I think in order to support this we need to drop the type argument, and re-name `serviceType` to `stackType` and document it as needing to be a type that is assignable to `Pulumi.Stack`. And then right after we verify that `serviceProvider is not null` we should ensure that `typeof(Pulumi.Stack).IsAssignableFrom(stackType)`. Then we would use reflection to call to `PulumiFn.Create<TStack>` using `stackType`.

orionstudt: Or to avoid the reflection we could make a non-generic implementation of `PulumiFn<TStack>` that takes in a type instance.

Issue #5432, opened 2020-09-21
Figure out story for secrets provider state in pulumi.stack.yaml and Automation API


EvanBoyle:
Creating a stack with a secrets provider involves creating state in the `Pulumi.stack.yaml` file that sometimes can't be known in advance (encryption salts for password providers). This is typically fine for traditional Pulumi programs where users check these values in, but with inline programs, we want to enable a world where a user is able to work entirely through code. Some options:

1. Make sure every secrets provider (or new variants) supports state that can be fully specified ahead of time
2. Support for storing `pulumi.stack.yaml` in the various backends. 

For GA, we should document these in Automation API examples to show a program that works not only on first update but on subsequent rehydration as well.

EvanBoyle:
Here's an example where a user got tripped up by this: https://github.com/pulumi/automation-api-examples/issues/4

viveklak:
FWIW this affects kubernetes operator as well: https://github.com/pulumi/pulumi-kubernetes-operator/issues/135

viveklak:
> 
> I didnt see a `StackSettings` equivalent in the Go sdk for instance. 
I stand corrected - https://github.com/pulumi/pulumi/blob/daa6045381ab4c80ca4801d4a4b9f706ecc70bed/sdk/go/auto/local_workspace.go#L90

viveklak:
> FWIW this affects kubernetes operator as well: [pulumi/pulumi-kubernetes-operator#135](https://github.com/pulumi/pulumi-kubernetes-operator/issues/135)
> It seems fairly insidious and leaves the stack in an unexpected/surprising state. Also - this affects remote or local programs as well - if config is not explicitly checked-in.
> 

One suggestion from @EvanBoyle was to perhaps allow config to be stored/queried directly from the service instead of relying on the workspace as is currently done.

leezen:
@viveklak I purposely mixed the examples up a bit so that there would be a mix of a different backend + secrets provider along with passphrase vs. non-passphrase secrets provider. But you're right that it's probably better to just show all the combinations in all the languages.

viveklak:
> @viveklak I purposely mixed the examples up a bit so that there would be a mix of a different backend + secrets provider along with passphrase vs. non-passphrase secrets provider. But you're right that it's probably better to just show all the combinations in all the languages.

Agreed. I think short term certainly makes sense to add examples for all combinations.

PR #6901, opened 2021-04-28
[auto/go] Fix stack settings save/load typo and add tests

viveklak:
Helps with #5432 

Test demonstrates a panic without the fixes. Also - on case-sensitive filesystems the use of `pulumi.xxx.yaml` vs. `Pulumi.xxx.yaml` matters.

### DIFF
--- a/sdk/go/auto/local_workspace.go
+++ b/sdk/go/auto/local_workspace.go
@@ -73,7 +73,7 @@ func (l *LocalWorkspace) SaveProjectSettings(ctx context.Context, settings *work
 func (l *LocalWorkspace) StackSettings(ctx context.Context, stackName string) (*workspace.ProjectStack, error) {
 	name := getStackSettingsName(stackName)
 	for _, ext := range settingsExtensions {
-		stackPath := filepath.Join(l.WorkDir(), fmt.Sprintf("pulumi.%s%s", name, ext))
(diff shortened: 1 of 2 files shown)


sdk/go/auto/local_workspace.go
```diff
@@ -73,7 +73,7 @@ func (l *LocalWorkspace) SaveProjectSettings(ctx context.Context, settings *work
 func (l *LocalWorkspace) StackSettings(ctx context.Context, stackName string) (*workspace.ProjectStack, error) {
 	name := getStackSettingsName(stackName)
 	for _, ext := range settingsExtensions {
-		stackPath := filepath.Join(l.WorkDir(), fmt.Sprintf("pulumi.%s%s", name, ext))
+		stackPath := filepath.Join(l.WorkDir(), fmt.Sprintf("Pulumi.%s%s", name, ext))
```
komalali, line 76: Is this something that we need to address across languages?

komalali (approved):

sdk/go/auto/local_workspace.go
viveklak: Sorry I didn't check the other SDKs. But if we are assuming case insensitivity then definitely.

EvanBoyle (approved):

PR #6902, opened 2021-04-28
Assertion fix in config key generation

viveklak:
Turns out the assertion check would only fail if the namespace was ":" or "".


komalali:
@viveklak is this still relevant?

komalali (approved):

viveklak:
> @viveklak is this still relevant?

Yes I believe so. Sorry I forgot about this.

PR #6910, opened 2021-04-28
[auto/dotnet] - Add PulumiFn implementation for runtime stack type

### LINKED ISSUE #6898: PulumiFn.Create should support `serviceType` overload

orionstudt:
Fixes #6898


orionstudt:
@t0yv0 @komalali this is ready for review whenever either of you have a moment. no hurry & ty!

orionstudt:
Made a couple minor changes to make sure tests were passing. Should be good now.

sdk/dotnet/Pulumi.Automation/PulumiFn.ServiceProvider.cs
```diff
@@ -0,0 +1,69 @@
...
+            Type stackType)
+        {
+            if (serviceProvider is null)
+                throw new ArgumentNullException(nameof(serviceProvider));
+
+            if (stackType is null)
+                throw new ArgumentNullException(nameof(stackType));
+
+            var pulumiStackType = typeof(Pulumi.Stack);
+            if (!pulumiStackType.IsAssignableFrom(stackType) || pulumiStackType == stackType)
```
t0yv0, line 27: So what happens if `pulumiStackType == stackType`, that is, `new PulumiFnServiceProvider(serviceProvider, typeof(Pulumi.Stack))` - we currently would throw here. Is that as intended? Sorry, I have not used inline programs before so building up my intuition here.

t0yv0:
Looking great! Just had this one question above.

t0yv0 (approved):

sdk/dotnet/Pulumi.Automation/PulumiFn.ServiceProvider.cs
orionstudt: I decided to throw for that because in practice if you provide `Pulumi.Stack` than your `up` operation would be a no-op and wouldn't declare any resources, so it seemed like the right thing to do. I figured we'd want to enforce that the provided type be a derivation of `Pulumi.Stack` and not just assignable to `Pulumi.Stack`, but we could change that

orionstudt: but I guess if we wanted parity with the generic API we wouldn't throw for using `Pulumi.Stack`, since you could do `PulumiFn.Create<Pulumi.Stack>()` without throwing.. so we can revert that if you want

t0yv0: That's reasonable, perfect so this line is as intended.

orionstudt:
@t0yv0 FYI this is still waiting on merge, in case you missed it

PR #6921, opened 2021-04-29
Improve default formatting of auto.autoError

jen20:
This commit changes the `error` interface for `auto.autoError` such that the formatting does not contain punctuation at the start of lines. Previously the formatting for an error was similar to this:

```
code: 255
, stdout: Created stack 'ateststack-90f37904'
, stderr: error: no Pulumi project found in the current working directory
: failed to create stack: exit status 255
```

After this change, the same error would be formatted:

```
failed to create stack: exit status 255
code: 255
stdout: Created stack 'ateststack-90f37904'
stderr: error: no Pulumi project found in the current working directory
```

The implementation no longer relies on `github.com/pkg/errors`, instead just using `fmt.Sprintf` since the error contract returns a plain string.

### DIFF
--- a/sdk/go/auto/errors.go
+++ b/sdk/go/auto/errors.go
@@ -15,10 +15,9 @@
 package auto
 
 import (
+	"fmt"
 	"regexp"
 	"strings"
-
-	"github.com/pkg/errors"
 )
 
 type autoError struct {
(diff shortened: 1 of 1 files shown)


komalali:
Closed in favor of https://github.com/pulumi/pulumi/pull/6924 which includes the changelog entry

Issue #6503, opened 2021-03-11
Add user-agent


EvanBoyle:
Automation API exposes execution kind `auto` or `inline`, but doesn't offer any way to distinguish software built on top of it. We should consider adding some sort of `user-agent` field so that users can see (and set) what deployed their stack. Whether that be the kubernetes operator, github actions, or any of the higher-level tools that get built on top of automation API. 

Eventually, this can be surfaced in the pulumi console, CLI history ux, etc. 

- [x] cli
- [x] go sdk
- [x] node sdk
- [ ] dotnet sdk
- [ ] python sdk

EvanBoyle:
Concretely, in automation API I think we would expose this as an `updateOption` (preview, refresh, and destroy as well).

```ts
stack.up({
program ...,
userAgent: "foo",
})
```

afro-coder:
@EvanBoyle Hi, was just going through the issues, maybe we can close this one out since the PR is already merged? Thanks

tgummerer:
@afro-coder did you check if this was implemented in the dotnet and python SDK?  Those are the checkboxes still missing above.  If you did, please link the merged PRs here, the PR linked above is only for Go and NodeJS.

Frassle:
afaik this isn't in place for python, dotnet, or java.

afro-coder:
Oh you're correct I didn't see that

PR #6935, opened 2021-04-30
Add user agent to the CLI, Go and Nodejs Automation API SDKs

EvanBoyle:
# Description

Adds CLI support, and automation API support for the nodejs and Go SDKs for specifying a user agent at update time. 

This user agent is stored with the update and can be displayed in various backends so that users can distinguish between human driven activity, CLI activity, and activity driven by automation like the Kubernetes Operator or the Pulumi GHA. 

Leaving this as a hidden flag in the CLI for now. We can eventually open and up, and add documentation so users can use this via the CLI but I will leave that as a later exercise. For now the current implementation with a hidden flag suits the desired use case. 

Contributes to https://github.com/pulumi/pulumi/issues/6503 for CLI, Go, and Node. Required for https://github.com/pulumi/home/issues/1375

## Checklist

- [x] I have labelled my PR with the relevant change type

### DIFF
--- a/sdk/go/auto/minimum_version.go
+++ b/sdk/go/auto/minimum_version.go
@@ -19,6 +19,6 @@ import "github.com/blang/semver"
 var minimumVersion = calculateMinVersion()
 
 func calculateMinVersion() semver.Version {
-	v, _ := semver.ParseTolerant("3.1.0-alpha")
+	v, _ := semver.ParseTolerant("3.2.0-alpha")
 	return v
 }
--- a/sdk/go/auto/minimum_version.go
+++ b/sdk/go/auto/minimum_version.go
@@ -19,6 +19,6 @@ import "github.com/blang/semver"
 var minimumVersion = calculateMinVersion()
 
 func calculateMinVersion() semver.Version {
-	v, _ := semver.ParseTolerant("3.1.0")
+	v, _ := semver.ParseTolerant("3.1.0-alpha")
 	return v
 }
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -18,6 +18,8 @@
 - [codegen/python] Lazy module import to improve CLI startup performance
   [#6827](https://github.com/pulumi/pulumi/pull/6827)
 
+- [auto/go,nodejs] Add UserAgent to update/pre/refresh/destroy options.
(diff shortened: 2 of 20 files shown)


sdk/go/auto/minimum_version.go
```diff
@@ -19,6 +19,6 @@ import "github.com/blang/semver"
 var minimumVersion = calculateMinVersion()
 
 func calculateMinVersion() semver.Version {
-	v, _ := semver.ParseTolerant("3.1.0")
+	v, _ := semver.ParseTolerant("3.1.0-alpha")
```
stack72, line 22: I believe this should be 3.2.0-alpha

stack72 (approved):
I still feel the introduction of backwards incompatible changes between the automation api and the cli can be frustrating for users who upgrade 1 and not the other but I understand it's a necessary evil

I see a mix of UserAgent and --exec-agent in this functionality - my only thoughts would be if we can make these actually line up

But apart from that LGTM

sdk/go/auto/minimum_version.go
lukehoban: It's kind of painful to bump the minimum required CLI version - that will break some existing users - are we sure we want to do that here - vs. handling older CLIs and just doing "better" on newer CLIs?

EvanBoyle:
`exec` is just the namespace of the value of which we have keys "kind" and "agent" that can be set. I picked UserAgent as a friendly term to expose via the SDK. 

In the service users will already see `exec.kind`. Now they will see `exec.agent` as well if automation api authors choose to specify the `UserAgent` option.

sdk/go/auto/minimum_version.go
komalali: Honestly, even after everything I said about why it's fine to bump min versions, I really do struggle with justifying the minimum version bump for such a small specific change. It would take some very minor refactoring to be able to do this check per feature rather than an overall version bump. 

I think it's fine to defer to if/until we hear more feedback but can we please open an issue to track making that improvement and hold ourselves to doing that work if bumping this version becomes a regular occurrence?

EvanBoyle: I'm fine with designing a new policy, but I would ask that we do it outside of the context of a code review if we really feel this to be an important topic.

I've opened an issue here: https://github.com/pulumi/pulumi/issues/6937

Let's continue the discussion there.

PR #6942, opened 2021-04-30
[sdk/python] Fix type-related regression on Python 3.6

### LINKED ISSUE #6886: Tests using mocks are failing on Python 3.6 with Pulumi v3
## Expected behavior
The tests would pass without an error.

## Current behavior
Tests run using pytest fail on Python 3.6 with:
```
args = (), kwargs = {}

    def wrapper(*args, **kwargs):
>       _sync_await(run_pulumi_func(lambda: _sync_await(Output.from_input(fn(*args, **kwargs)).future())))

/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/site-packages/pulumi/runtime/mocks.py:37: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
... shortened...
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

arg = ~T, tvars = (~T, ~T), args = [<class 'str'>]

    def _replace_arg(arg, tvars, args):
        """An internal helper function: replace arg if it is a type variable
        found in tvars with corresponding substitution from args or
        with corresponding substitution sub-tree if arg is a generic type.
        """
    
        if tvars is None:
            tvars = []
        if hasattr(arg, '_subs_tree') and isinstance(arg, (GenericMeta, _TypingBase)):
            return arg._subs_tree(tvars, args)
        if isinstance(arg, TypeVar):
            for i, tvar in enumerate(tvars):
                if arg == tvar:
>                   return args[i]
E                   IndexError: list index out of range

/opt/hostedtoolcache/Python/3.6.13/x64/lib/python3.6/typing.py:566: IndexError
```

The tests pass on Python 3.7 and 3.8.

## Steps to reproduce
Here's a link to the GitHub Actions workflow run with failing tests where the full output of pytest can be seen:
https://github.com/tlinhart/pulumi-aws-tags/runs/2445335544

## Context (Environment)
I wrote a simple [Pulumi package](https://github.com/tlinhart/pulumi-aws-tags) that provides auto-tagging for AWS resources and which is based on these two resources:
- https://www.pulumi.com/blog/automatically-enforcing-aws-resource-tagging-policies/
- https://github.com/joeduffy/aws-tags-example

I've upgraded the package to Pulumi v3 (following https://www.pulumi.com/docs/get-started/install/migrating-3.0/) and Pulumi AWS provider package v4.

The actual versions used in the failing tests are:
Python 3.6.13
pulumi-3.1.0
pulumi-aws-4.1.0

justinvp:
Pulumi 3.0 uses type annotations for input values to determine whether dict keys should be translated from snake_case to camelCase or not. This additional inspection of types did not work correctly on Python 3.6 due to some missing functionality on that version of Python which we need to provide an implementation for. Specifically, when inspecting `Union` args to determine whether or not a value is intended to be an input class or user-defined dict. To address the issue, this change improves how we get the args for `Union` types when running on Python 3.6 to behave the same way as later versions of Python (for our purposes). Existing tests fail on Python 3.6 before this change, and pass after.

Updating CI to include Python 3.6 in the test matrix is tracked by #6110

Fixes #6886


sdk/python/lib/pulumi/_types.py
```diff
@@ -657,6 +657,13 @@ def get_origin(tp):
         return None
 
     def get_args(tp):
+        # Emulate the behavior of get_args for Union on Python 3.6.
```
justinvp, line 660: Basically, what was happening was that `tp.__args__` for `Union` wasn't returning the flattened list of args on Python 3.6 like later versions. So when we were [inspecting those args](https://github.com/pulumi/pulumi/blob/69d50ced27920c30d519313603cb247654ea7341/sdk/python/lib/pulumi/_types.py#L906-L933) to determine if the type annotation was intended to be an input class vs. user-defined dict, we weren't able to because the args we were getting weren't what was expected.

We have tests for this https://github.com/pulumi/pulumi/blob/master/sdk/python/lib/test/test_types_input_type_types.py that were failing on Python 3.6 before the change in this PR, but we weren't running the tests against Python 3.6 in CI.

t0yv0: Yup. Reproduced fail/fix on Py 3.6. This is the substitution that _eval is doing (on `pulumi.Input[str]` context):

```
                assert _eval((str, (typing.Awaitable, str), (pulumi.output.Output, None))) == \
                    (str, typing.Awaitable[str], pulumi.output.Output[None])
```

t0yv0 (approved):

komalali (approved):
🙏🏽

PR #6944, opened 2021-04-30
[plugins] Allow opt out of loading plugins on PATH

jen20:
# Description

This commit makes it possible to opt out of loading plugins from `PATH` by setting `PULUMI_IGNORE_AMBIENT_PLUGINS` to any non-empty value. This is useful when automatic IDE tooling may build remote component plugins into `GOBIN` unbeknownst to the user, and a resulting stale version of the plugin is loaded in place of newer versions - even those , explicitly installed.

## Checklist

No existing test coverage of plugin loading as far as I can tell.


stack72 (approved):
This is great - thanks so much for this! I am going to add a further enhancement here to add a warning when a provider is found on the PATH to make sure that users don't get hit by this

PR #6949, opened 2021-04-30
[sdk/python] Don't error when dict input value has a mismatched type

### LINKED ISSUE #6818: [sdk/python] - AssertionError: Unexpected type. Expected 'dict' got '<class 'str'>' for Policy
Before 3.0.0 I could pass in a dict as an `aws.BucketPolicy.policy`. Now the same code fails with: `AssertionError: Unexpected type. Expected 'dict' got '<class 'str'>'`

## Expected behavior

The code should succeed as it did before

## Current behavior

It fails with an assertion error

## Steps to reproduce

The following code reproduces the issue if you try to run `pulumi up`

```python
import pulumi
from pulumi_aws import s3

bucket = s3.Bucket("my-bucket")

policy = s3.BucketPolicy("my-policy", bucket=bucket.id, policy={
    "Version": "2012-10-17",
    "Statement": {
        "Effect": "Allow",
        "Principal": "*",
        "Action": ["s3:GetObject"],
        # Policy refers to bucket explicitly
        "Resource": [pulumi.Output.concat("arn:aws:s3:::", bucket.id, "/*")]
    }
})
```

justinvp:
Pulumi 3.0 raises an error when a dict value is passed as an input but the type annotation does not accept a dict. Unfortunately, this prevents historical cases where a dict value is allowed but the type annotation doesn't match. We need to fix the type annotations, but in the meantime, we should not raise an error in the SDK for such cases as it breaks existing programs.

Fixes #6818
Fixes https://github.com/pulumi/pulumi-kubernetes/issues/1541

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -45,6 +45,9 @@
 - [sdk/python] Fix type-related regression on Python 3.6.
   [#6942](https://github.com/pulumi/pulumi/pull/6942)
 
+- [sdk/python] Don't error when a dict input value has a mismatched type annotation.
+  [#6949](https://github.com/pulumi/pulumi/pull/6949)
+
 ### Misc.
 
 - [auto/dotnet] Bump YamlDotNet to 11.1.1
--- a/sdk/python/lib/pulumi/runtime/rpc.py
+++ b/sdk/python/lib/pulumi/runtime/rpc.py
@@ -300,7 +300,15 @@ async def serialize_property(value: 'Input[Any]',
                         get_type = lambda k: args[1]
                         translate = None
                 else:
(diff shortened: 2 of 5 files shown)


t0yv0 (approved):
Sounds good, this was the place where it failed that I was looking at last week 👍 

Just an idea you don't want the loophole to be stricter that is `typ == str`? Or the issue does only arise for `str`? Thanks.

justinvp:
> Just an idea you don't want the loophole to be stricter that is `typ == str`? Or the issue does only arise for `str`? Thanks.

There are other cases that are currently hitting this that isn't `str`. Like https://github.com/pulumi/pulumi-kubernetes/issues/1541 (the actual bug for that is https://github.com/pulumi/pulumi/issues/6950 -- we need to handle `Union`).

Issue #5448, opened 2020-09-23
[Automation API] implement plugin acquisition for inline programs


EvanBoyle:
We migrated inline programs (host mode) to use a [language-native language service RPC implementation](https://github.com/pulumi/pulumi/pull/5317). This should now make it possible to implement `GetRequiredPlugins` for inline programs. We'll need to translate the current go implementations into the various languages.

yarinm:
Is there any ETA about this? Maybe just for Go?

Issue #5991, opened 2020-12-21
Support configuring `--server` flag to plugin installation in all languages


lukehoban:
Currently we have support for metadata in NodeJS `package.json` to tell plugin installation to use a customer server URL (via passing `--server` to `pulumi plugin install`).  We do not have equivalent support in other languages yet - and will need to do so to support custom plugin hosting for 3rd party components.

Related to https://github.com/pulumi/pulumi-tf-provider-boilerplate/issues/13 and https://github.com/pulumi/pulumi/issues/5937.

There are several aspects of this tracked as individual issues:

- [x] [#7243] Python
  - Need to handle non `pulumi-` prefixed packages when searching for packages that need plugins
- [x] [#7244] .NET
  - Need to adopt `pulumiplugin.json` to support custom server URLs
  - Allow non-`Pulumi.` packages when searching for packages that need plugins
- [x] [#7245] Go
  - Need to adopt `pulumiplugin.json` to be able to specify custom server URL for Go
  - Need to handle non-Pulumi prefixed modules
- [ ] [#7242] When determining required plugins for an existing stack by looking for providers in the state, we need a way to determine if any of those providers came from a custom server URL so we can properly install the required plugins.

justinvp:
There are several aspects of this:

- When determining required plugins for an existing stack by looking for providers in the state, we need a way to determine if any of those providers came from a custom server URL so we can properly install the required plugins.
- .NET
  - Need to adopt `pulumiplugin.json` to support custom server URLs
  - Allow non-`Pulumi.` packages when searching for packages that need plugins
- Go
  - Need to adopt `pulumiplugin.json` to be able to specify custom server URL for Go
  - Need to handle non-Pulumi prefixed modules
- Python
  - Need to handle non `pulumi-` prefixed packages
- Node.js
  - Confirm there is nothing else needed here

infin8x:
@justinvp @iwahbe another one we might be able to close as duplicate?

iwahbe:
>@justinvp @iwahbe another one we might be able to close as duplicate?

@infin8x I'm not sure what you mean?

infin8x:
ah sorry, was moving quickly. I think this is a duplicate of pulumi/home#1655, and figured it might be good to close this one as duplicate of that one to reduce issue debt. but, happy to leave this open if it represents something different.

iwahbe:
[...]

PR #6951, opened 2021-05-01
[auto/dotnet] - plugin acquisition for inline programs

orionstudt:
For the .NET portion of #5448

Looking for thoughts/criticism on this approach. I will do a review and leave some questions that I am looking for specific thoughts on.

Any suggestions on how we want to unit test this? Since it means manipulating the referenced assemblies for whatever we pass in to `DiscoveryAssembly` - so not sure how to best approach that. I was testing in a console app (lol) where I could reference and de-reference resource providers as I worked.

@EvanBoyle @komalali @t0yv0 any thoughts welcome :)

### DIFF
--- a/sdk/dotnet/Pulumi.Automation/PulumiFn.Inline.cs
+++ b/sdk/dotnet/Pulumi.Automation/PulumiFn.Inline.cs
@@ -1,8 +1,8 @@
 ﻿// Copyright 2016-2021, Pulumi Corporation
 
-
 using System;
 using System.Collections.Generic;
+using System.Reflection;
 using System.Runtime.ExceptionServices;
 using System.Threading;
 using System.Threading.Tasks;
@@ -16,6 +16,7 @@ namespace Pulumi.Automation
         public PulumiFnInline(Func<CancellationToken, Task<IDictionary<string, object?>>> program)
         {
             this._program = program;
+            this.DiscoveryAssembly = Assembly.GetEntryAssembly();
         }
 
         internal override async Task<ExceptionDispatchInfo?> InvokeAsync(IRunner runner, CancellationToken cancellationToken)
--- a/sdk/dotnet/Pulumi.Automation/PulumiFn.cs
+++ b/sdk/dotnet/Pulumi.Automation/PulumiFn.cs
@@ -3,6 +3,7 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
+using System.Reflection;
 using System.Runtime.ExceptionServices;
 using System.Threading;
 using System.Threading.Tasks;
@@ -14,6 +15,11 @@ namespace Pulumi.Automation
     /// </summary>
     public abstract class PulumiFn
     {
+        /// <summary>
+        /// Get or set the assembly that will be used for required plugin discovery.
+        /// </summary>
+        public Assembly? DiscoveryAssembly { get; set; }
+
         internal PulumiFn()
         {
         }
--- a/sdk/dotnet/Pulumi.Automation/Runtime/LanguageRuntimeService.cs
+++ b/sdk/dotnet/Pulumi.Automation/Runtime/LanguageRuntimeService.cs
@@ -33,7 +33,7 @@ namespace Pulumi.Automation
             if (assembly is null)
                 return response;
 
-            var pulumiAssemblyNames = assembly.GetReferencedAssemblies().Where(x => x.Name != null && x.Name.StartsWith("Pulumi.")).ToArray();
+            var pulumiAssemblyNames = assembly.GetReferencedAssemblies().Where(x => x.Name != null && x.Name.StartsWith("Pulumi.", StringComparison.Ordinal)).ToArray();
             foreach (var assemblyName in pulumiAssemblyNames)
             {
                 var pulumiAssembly = Assembly.Load(assemblyName);
@@ -43,7 +43,7 @@ namespace Pulumi.Automation
                 // instead of using the embedded version.txt resource
                 // pulumiAssembly.GetCustomAttribute<ResourcePluginAttribute>();
                 var resources = pulumiAssembly.GetManifestResourceNames();
-                var versionResource = resources.FirstOrDefault(x => x.EndsWith("version.txt"));
+                var versionResource = resources.FirstOrDefault(x => x.EndsWith("version.txt", StringComparison.Ordinal));
                 if (versionResource is null)
                 {
                     // no version.txt so is not resource plugin
@@ -61,7 +61,7 @@ namespace Pulumi.Automation
                 var versionText = await versionReader.ReadToEndAsync().ConfigureAwait(false);
 
                 // ToLower() everything after "Pulumi." to determine plugin name
-                var pluginName = assemblyName.Name!.Substring("Pulumi.".Length).ToLower();
+                var pluginName = assemblyName.Name!.Substring("Pulumi.".Length).ToLowerInvariant();
                 var version = versionText.Trim();
                 var parts = versionText.Split('\n', StringSplitOptions.RemoveEmptyEntries)
                     .Select(x => x.Trim())
--- a/sdk/dotnet/Pulumi.Automation/Runtime/LanguageRuntimeService.cs
+++ b/sdk/dotnet/Pulumi.Automation/Runtime/LanguageRuntimeService.cs
@@ -1,8 +1,9 @@
 ﻿// Copyright 2016-2021, Pulumi Corporation
 
 using System;
-using System.Collections.Generic;
+using System.IO;
 using System.Linq;
+using System.Reflection;
 using System.Runtime.ExceptionServices;
 using System.Threading;
 using System.Threading.Tasks;
@@ -23,10 +24,69 @@ namespace Pulumi.Automation
             this._callerContext = callerContext;
         }
 
-        public override Task<GetRequiredPluginsResponse> GetRequiredPlugins(GetRequiredPluginsRequest request, ServerCallContext context)
+        public override async Task<GetRequiredPluginsResponse> GetRequiredPlugins(GetRequiredPluginsRequest request, ServerCallContext context)
         {
             var response = new GetRequiredPluginsResponse();
-            return Task.FromResult(response);
+            var assembly = this._callerContext.Program.DiscoveryAssembly;
+
+            // if discovery assembly is null, assuming consumer installed plugins manually
+            if (assembly is null)
+                return response;
+
+            var pulumiAssemblyNames = assembly.GetReferencedAssemblies().Where(x => x.Name != null && x.Name.StartsWith("Pulumi.")).ToArray();
+            foreach (var assemblyName in pulumiAssemblyNames)
+            {
+                var pulumiAssembly = Assembly.Load(assemblyName);
+
+                // get version.txt
+                // this part would probably be less error prone if there was an assembly attribute available
+                // instead of using the embedded version.txt resource
+                // pulumiAssembly.GetCustomAttribute<ResourcePluginAttribute>();
+                var resources = pulumiAssembly.GetManifestResourceNames();
+                var versionResource = resources.FirstOrDefault(x => x.EndsWith("version.txt"));
+                if (versionResource is null)
+                {
+                    // no version.txt so is not resource plugin
+                    continue;
+                }
+
+                using var versionStream = pulumiAssembly.GetManifestResourceStream(versionResource);
+                if (versionStream is null)
+                {
+                    // we've already verified that version.txt should be present so throw here?
+                    throw new ApplicationException($"Unable to load version.txt from {assemblyName.FullName}");
+                }
+
+                var versionReader = new StreamReader(versionStream);
+                var versionText = await versionReader.ReadToEndAsync().ConfigureAwait(false);
+
+                // ToLower() everything after "Pulumi." to determine plugin name
+                var pluginName = assemblyName.Name!.Substring("Pulumi.".Length).ToLower();
+                var version = versionText.Trim();
+                var parts = versionText.Split('\n', StringSplitOptions.RemoveEmptyEntries)
+                    .Select(x => x.Trim())
+                    .ToArray();
+
+                // if version.txt is 2 lines than it is specifying plugin name explicitly
+                if (parts.Length is 2)
+                {
+                    pluginName = parts[0];
+                    version = parts[1];
+                }
+
+                // pre-pend 'v' if it isn't present on version string
+                if (!version.StartsWith('v'))
+                    version = $"v{version}";
+
+                response.Plugins.Add(new PluginDependency
+                {
+                    Name = pluginName,
+                    Version = version,
+                    Kind = "resource",
+                });
+            }
+
+            return response;
         }
 
         public override async Task<RunResponse> Run(RunRequest request, ServerCallContext context)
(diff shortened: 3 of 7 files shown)


sdk/dotnet/Pulumi.Automation/PulumiFn.Inline.cs
```diff
@@ -16,6 +16,7 @@ internal sealed class PulumiFnInline : PulumiFn
         public PulumiFnInline(Func<CancellationToken, Task<IDictionary<string, object?>>> program)
         {
             this._program = program;
+            this.DiscoveryAssembly = Assembly.GetEntryAssembly();
```
orionstudt, line 19: I waffled a bit on what the delegate-driven inline program should use as the default assembly, if anything. It can be hard to nail down a single assembly since delegates can be chained and added to from different assemblies and we don't know where it was started. I wanted to do `GetCallingAssembly()` in the `PulumiFn.Create(...)` factory methods but was a bit nervous about compiler optimization making that unpredictable. This is at least predictable, so do we think it is a reasonable default?

sdk/dotnet/Pulumi.Automation/PulumiFn.cs
```diff
@@ -14,6 +15,11 @@ namespace Pulumi.Automation
     /// </summary>
     public abstract class PulumiFn
     {
+        /// <summary>
+        /// Get or set the assembly that will be used for required plugin discovery.
+        /// </summary>
+        public Assembly? DiscoveryAssembly { get; set; }
```
orionstudt, line 21: Thoughts on naming, or if this should even be exposed? I was thinking we should allow it to be overridden so the consumer can be explicit, which I think is especially important for the delegate-driven inline program.

Do we think that this should be an optional parameter on the `PulumiFn.Create(..)` factory methods, or is the fact that it can be overriden on this property sufficient?

sdk/dotnet/Pulumi.Automation/Runtime/LanguageRuntimeService.cs
```diff
@@ -23,10 +24,69 @@ public LanguageRuntimeService(CallerContext callerContext)
...
 
-        public override Task<GetRequiredPluginsResponse> GetRequiredPlugins(GetRequiredPluginsRequest request, ServerCallContext context)
+        public override async Task<GetRequiredPluginsResponse> GetRequiredPlugins(GetRequiredPluginsRequest request, ServerCallContext context)
         {
             var response = new GetRequiredPluginsResponse();
-            return Task.FromResult(response);
+            var assembly = this._callerContext.Program.DiscoveryAssembly;
+
+            // if discovery assembly is null, assuming consumer installed plugins manually
+            if (assembly is null)
```
orionstudt, line 33: Should we make the `DiscoveryAssembly` required, and throw if it isn't present? I was thinking making it nullable would allow the consumer to disable the automatic plugin acquisition if they were installing plugins manually for whatever reason.

sdk/dotnet/Pulumi.Automation/Runtime/LanguageRuntimeService.cs
```diff
@@ -23,10 +24,69 @@ public LanguageRuntimeService(CallerContext callerContext)
...
+                return response;
+
+            var pulumiAssemblyNames = assembly.GetReferencedAssemblies().Where(x => x.Name != null && x.Name.StartsWith("Pulumi.")).ToArray();
```
orionstudt, line 36: Should we throw if the `DiscoveryAssembly` doesn't reference `Pulumi` and also doesn't reference any assemblies that start with `Pulumi.`, like what is happening [here](https://github.com/pulumi/pulumi/blob/73a66f48ead301bacf1d7d2833c016cf803b57c1/sdk/dotnet/cmd/pulumi-language-dotnet/main.go#L259)?

sdk/dotnet/Pulumi.Automation/Runtime/LanguageRuntimeService.cs
```diff
@@ -23,10 +24,69 @@ public LanguageRuntimeService(CallerContext callerContext)
...
+            foreach (var assemblyName in pulumiAssemblyNames)
+            {
+                var pulumiAssembly = Assembly.Load(assemblyName);
```
orionstudt, line 39: For both this `Assembly.Load(...)` call and the `assembly.GetManifestResourceStream(...)` call below, should we be catching `FileLoadException` or `FileNotFoundException` and doing something silently, or should we let them throw?

sdk/dotnet/Pulumi.Automation/Runtime/LanguageRuntimeService.cs
```diff
@@ -23,10 +24,69 @@ public LanguageRuntimeService(CallerContext callerContext)
...
+
+                // get version.txt
+                // this part would probably be less error prone if there was an assembly attribute available
+                // instead of using the embedded version.txt resource
+                // pulumiAssembly.GetCustomAttribute<ResourcePluginAttribute>();
```
orionstudt, line 44: Looking for your thoughts on this, @mikhailshilkov 

I think an assembly attribute would be a lot less error prone than trying to retrieve the embedded `version.txt` resource. We could put it in the `.csproj` for the resource provider packages like:
```
<ItemGroup>
  <AssemblyAttribute Include="Pulumi.ResourcePluginAttribute">
    <_Parameter1>$(Version)</_Parameter1>
  </AssemblyAttribute>
</ItemGroup>
```

And then the version could be inserted on build the same way that you are creating the `version.txt` file on build, I believe. Would probably still keep `version.txt` around since I think the GO dotnet language host would still need it, but just a thought. I know it may be a big lift to add that to everything.

sdk/dotnet/Pulumi.Automation/Runtime/LanguageRuntimeService.cs
```diff
@@ -23,10 +24,69 @@ public LanguageRuntimeService(CallerContext callerContext)
...
+                {
+                    pluginName = parts[0];
+                    version = parts[1];
+                }
+
+                // pre-pend 'v' if it isn't present on version string
+                if (!version.StartsWith('v'))
+                    version = $"v{version}";
+
+                response.Plugins.Add(new PluginDependency
```
orionstudt, line 81: And finally, should we consider any caching of `DiscoveryAssembly -> RequiredPlugins` to cut down on how often we might be loading the same assemblies, or would that be pre-optimization?

justinvp (commented):
@EvanBoyle, @komalali, are we planning to re-implement `GetRequiredPlugins` for each language for automation API? Is there any reason we couldn't leverage the existing implementations in each language provider? For example, possibly by adding a new CLI command for use by automation API only (it could be a hidden command), that launches the language plugin for the specified runtime, calls the language plugin's `GetRequiredPlugins` and returns the results as JSON over standard output, that the automation API implementation of `GetRequiredPlugins` could make use of? Is there any reason that wouldn't work for inline programs?

As part of https://github.com/pulumi/pulumi/issues/5991 we are going to be making some changes to the .NET language provider's implementation of `GetRequiredPlugins` to support custom server URLs. It'd be unfortunate to have to implement and maintain this logic twice. We have to make some similar changes for some of the other languages as well.

sdk/dotnet/Pulumi.Automation/Runtime/LanguageRuntimeService.cs
justinvp, line 36: `StartsWith` does a culture-sensitive comparison by default. If we do keep this implementation, we should use an ordinal comparison here:

```suggestion
            var pulumiAssemblyNames = assembly.GetReferencedAssemblies().Where(x => x.Name != null && x.Name.StartsWith("Pulumi.", StringComparison.Ordinal)).ToArray();
```

sdk/dotnet/Pulumi.Automation/Runtime/LanguageRuntimeService.cs
```diff
@@ -23,10 +24,69 @@ public LanguageRuntimeService(CallerContext callerContext)
...
+            foreach (var assemblyName in pulumiAssemblyNames)
+            {
+                var pulumiAssembly = Assembly.Load(assemblyName);
+
+                // get version.txt
+                // this part would probably be less error prone if there was an assembly attribute available
+                // instead of using the embedded version.txt resource
+                // pulumiAssembly.GetCustomAttribute<ResourcePluginAttribute>();
+                var resources = pulumiAssembly.GetManifestResourceNames();
+                var versionResource = resources.FirstOrDefault(x => x.EndsWith("version.txt"));
```
justinvp, line 46: `EndsWith` does a culture-sensitive comparison by default. If we do keep this implementation, we should use an ordinal comparison:

```suggestion
                var versionResource = resources.FirstOrDefault(x => x.EndsWith("version.txt", StringComparison.Ordinal));
```

sdk/dotnet/Pulumi.Automation/Runtime/LanguageRuntimeService.cs
```diff
@@ -23,10 +24,69 @@ public LanguageRuntimeService(CallerContext callerContext)
...
+                {
+                    // we've already verified that version.txt should be present so throw here?
+                    throw new ApplicationException($"Unable to load version.txt from {assemblyName.FullName}");
+                }
+
+                var versionReader = new StreamReader(versionStream);
+                var versionText = await versionReader.ReadToEndAsync().ConfigureAwait(false);
+
+                // ToLower() everything after "Pulumi." to determine plugin name
+                var pluginName = assemblyName.Name!.Substring("Pulumi.".Length).ToLower();
```
justinvp, line 64: `ToLower` uses the casing conventions of the current culture. If we keep this implementation, we should use `ToLowerInvariant` instead:

```suggestion
                var pluginName = assemblyName.Name!.Substring("Pulumi.".Length).ToLowerInvariant();
```

orionstudt:
@justinvp I'm very curious how a CLI implementation of `GetRequiredPlugins` would be able to resolve required plugins for an assembly loaded at runtime, or for a pulumi program built from a delegate? Maybe I'm not understanding how this would function:
> For example, possibly by adding a new CLI command for use by automation API only (it could be a hidden command), that launches the language plugin for the specified runtime, calls the language plugin's GetRequiredPlugins and returns the results as JSON over standard output,

orionstudt:
What if there was a CLI command similar to `preview`, but lower level in that it doesn't depend on the plugin being installed? Like what if you could run `pulumi discover-plugins` on your pulumi program, and it would gather all of the resources that you declared and use those to determine what plugins you need? Then the single `GetRequiredPlugins` implementation for each SDK could reside in the SDK itself (such as the .NET SDK using reflection to determine necessary plugins), and there wouldn't be the code duplication you're concerned about - as well as the fact that `GetRequiredPlugins` would no longer rely on the CLI having done the `dotnet build` for the nuget packages to be present. Since you could just run this against any pulumi program, including inline. 

Anyways just a thought. Currently I'm not seeing how you can support `GetRequiredPlugins` for inline program without a little code-dupe, given the fact that every variation of `GetRequiredPlugins` for an inline-program involved a pre-built binary and even the CLI is currently unable to determine required plugins if the binary is already built.

justinvp:
> What if there was a CLI command similar to `preview`, but lower level in that it doesn't depend on the plugin being installed? Like what if you could run `pulumi discover-plugins` on your pulumi program, and it would gather all of the resources that you declared and use those to determine what plugins you need? Then the single `GetRequiredPlugins` implementation for each SDK could reside in the SDK itself (such as the .NET SDK using reflection to determine necessary plugins), and there wouldn't be the code duplication you're concerned about - as well as the fact that `GetRequiredPlugins` would no longer rely on the CLI having done the `dotnet build` for the nuget packages to be present. Since you could just run this against any pulumi program, including inline.

This is an interesting idea! I'm not sure we'd bite this off now, but we may want to consider something along these lines in the future, especially if there's more duplication with the other languages. It is nice that something like this would be able to handle the case for both regular and inline programs as well as prebuilt binaries.

> Currently I'm not seeing how you can support `GetRequiredPlugins` for inline programs without a little code-dupe

I think we'll just live with some amount of code-dupe here for now.

t0yv0:
I'm a little behind on understanding inline programs. My 2c here is that i'm afraid of assembly traversal through System.Reflection. Its failure modes are too interesting, and it has potential for slowdowns and/or high memory use. Given this, if we cannot avoid this work (as in, through explicit manifests) then isolating it to a dedicated process and perhaps caching the result sounds like a good idea to me to future-proof our other processes from these problems.

orionstudt:
When you say "isolating it to a dedicated process" do you mean that you would like to avoid the code-dupe and keep the implementation in Go for now, or that you would be OK with the System.Reflection route but encapsulated elsewhere somehow?

t0yv0:
My concern is not with code duplication - though that is a good concern too!

Having a look now at what we have in this function:

```
func (host *dotnetLanguageHost) DeterminePulumiPackages(
	ctx context.Context, engineClient pulumirpc.EngineClient) ([][]string, error) {
```

So it sounds like the Go implementation shells out to `dotnet` CLI for example `dotnet list package --include-transitive`. 

So this implementation checks off my boxes - there is probably a .NET process started under the hood there, but it's terminated when the information is retrieved, and it will not contaminate any app processes or interfere with dynamic assemblies like ones created by Moq.  This is what I meant by process isolation - the plugin discovery runs in a separate process from mainline Pulumi things.

Justin's comment suggestion also checks my boxes:

> For example, possibly by adding a new CLI command for use by automation API only (it could be a hidden command), that launches the language plugin for the specified runtime, calls the language plugin's GetRequiredPlugins and returns the results as JSON over standard output

There is overhead of course but IMO the isolation we get is worth the overhead here.

Or perhaps there's other reasons it just wouldn't work for inline programs? Thank you.

t0yv0:
You hint at one problem with assemblies loaded at runtime:

> I'm very curious how a CLI implementation of GetRequiredPlugins would be able to resolve required plugins for an assembly loaded at runtime, or for a pulumi program built from a delegate?

That is indeed difficult, presumably since to determine correct assemblies that are loaded at runtime, one has to execute the code.

Are there scenarios where this is really useful? I imagine it would be much much easier to maintain a system that did not have this capability; where the set of assemblies that are needed (or a superset of this set) is available statically before running the code. By easier to maintain I mean this: plugin discovery either fails deterministically, or it works, there is no randomness that is introduced by depending on program execution paths and races.

orionstudt:
> Are there scenarios where this is really useful?

I believe so! Being able to use Automation API to make a Web API that can run Pulumi code triggered by web request is really cool, and allows for some useful deployment scenarios. But the limitation here is that your Web API needs to have the Pulumi code to deploy what you want it to deploy. If you want a static website in an S3 bucket with a cloud front distribution in front of it, the Web API needs to have the Pulumi program that describes that stack.

So in my mind the logical next step is a Web API that can orchestrate Pulumi deployments with any Pulumi code you give it. That's what I've been trying to build.. so in practice before you did any deployments you would define a "StackType" or something. You would upload the Pulumi program for "S3CloudfrontWebsite" to the Web API. And then a consumer would request to deploy an "S3CloudfrontWebsite" given some inputs, and the Web API would look up & load that previously uploaded program and execute the deployment. So this would require the Web API loading the assembly that contains the Pulumi code at runtime.

All that said, I don't think it would be too big of a deal if such a use-case couldn't do automatic discovery and needed to manually define which plugins it requires - I just think it would be cool :)

orionstudt:
Ultimately what we go with is up to you guys - so let me know how/if you want to go forward with this PR & issue!

orionstudt:
Gonna abandon this just because it's been sitting forever and I believe other directions have been taken.

gitfool:
@orionstudt I'm curious what's changed. Do we still need to explicitly call `Workspace.InstallPluginAsync` for all required plugins?

orionstudt:
@gitfool I believe @t0yv0 had plans to accomplish this going a different direction, or there was some secondary goal on the roadmap that would resolve this anyways. But if I am wrong we can re-open this but it is likely due some changes since it hasn't been touched in over a year. #5448 still stands.

PR #6959, opened 2021-05-03
Remove leading and trailing whitespace in resource properties

cnunciato:
Resources with properties without `.Comment` values are rendered with empty lines, which breaks Markdown rendering, so this change just removes any leading and trailing whitespace.

Fixes https://github.com/pulumi/docs/issues/5904.


lukehoban (approved):

komalali (approved):
✨ Thank you @cnunciato!

Issue #6945, opened 2021-04-30
Flaky test: "correctly sets config on multiple stacks concurrently"


t0yv0:
There's an automation API test that appears correct but has failed.

"correctly sets config on multiple stacks concurrently"

```
  1 failing

  1) LocalWorkspace correctly sets config on multiple stacks concurrently:
     StackNotFoundError: code: 255
 stdout: 
 stderr: error: no stack named 'dev' found
 err?: 
      at Object.createCommandError (bin/automation/errors.js:68:42)
      at ChildProcess.<anonymous> (bin/automation/cmd.js:63:40)
      at Process.ChildProcess._handle.onexit (internal/child_process.js:277:12)
```

https://github.com/pulumi/pulumi/pull/6941/checks?check_run_id=2477458536

I cannot repro but it also is extremely surprising how could this in principle happen?

There is either:

- (A) incomplete isolation; `LocalWorkspace.create({ workDir });` is using some concurrently modified state, from other CI runners? from other tests running concurrently in this same CI runner?

- (B) eventual consistency in Pulumi state; say `pulumi stack init` cli returns OK but the stack is not reporting as "created" until a few ms after that; if that is true we need a retry somewhere

# Environment

CI on Ubuntu

komalali:
Yeah I have noticed this in CI but I am equally confused by how this is happening. The strangest thing to me is the `no stack named 'dev' found` because we use `Stack.createOrSelect`, which is supposed to create the stack if it gets a `no stack named 'dev' found` error.

t0yv0:
Yes I saw that implementation but I think that is still correct. It should work and not throw "no stack named `dev` found". I still think (A) and (B) are our possibilities here, with (A) being benign test issue, and (B) potentially being a rare but bad issue in Pulumi itself that might affect our customers.

komalali:
So, a couple of things. I think this is actually failing on [this line](https://github.com/pulumi/pulumi/blob/98ae3adf605b3d8faaf40d325a04f60d2f7db6ad/sdk/nodejs/tests/automation/localWorkspace.spec.ts#L485) which makes sense according to the `no stack named dev found` message. Secondly, I imagine it's more likely (A) that is the culprit here, caused by other CI runners since we aren't randomly naming stacks and the project has the same name (since its `workDir` based). 

I've proposed a solution that uses the local backend so we can avoid the interop with the service and therefore avoid any concurrent modified state in the service https://github.com/pulumi/pulumi/pull/6960

[...]

PR #6960, opened 2021-05-04
Use local backend for concurrency test

komalali:
# Description

Fixes a flakey test by using a local file backend to avoid interop in the service.

Fixes https://github.com/pulumi/pulumi/issues/6945

## Checklist


t0yv0 (approved):
Awesome, quick win here.

PR #6963, opened 2021-05-04
Small change to the logic around ambient plugins on the PATH

stack72:
As a followup to #6944, we needed to invert the check for the
Env var that skips the ambient plugins on the PATH

### DIFF
--- a/sdk/go/common/workspace/plugins.go
+++ b/sdk/go/common/workspace/plugins.go
@@ -587,7 +587,7 @@ func GetPluginPath(kind PluginKind, name string, version *semver.Version) (strin
 
 	// If we have a version of the plugin on its $PATH, use it, unless we have opted out of this behavior explicitly.
 	// This supports development scenarios.
-	if os.Getenv("PULUMI_IGNORE_AMBIENT_PLUGINS") != "" {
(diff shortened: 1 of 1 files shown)


sdk/go/common/workspace/plugins.go
```diff
@@ -587,7 +587,7 @@ func GetPluginPath(kind PluginKind, name string, version *semver.Version) (strin
 
 	// If we have a version of the plugin on its $PATH, use it, unless we have opted out of this behavior explicitly.
 	// This supports development scenarios.
-	if os.Getenv("PULUMI_IGNORE_AMBIENT_PLUGINS") != "" {
+	if _, isFound := os.LookupEnv("PULUMI_IGNORE_AMBIENT_PLUGINS"); !isFound {
```
t0yv0, line 590: It seems you're inverting the logic. Seems to be the same as:

```
os.Getenv("PULUMI_IGNORE_AMBIENT_PLUGINS") == ""
```

But more emphatically readable !isFound.

Judging by the comment, the correction is needed here yes. The IGNORE env var should skip the logic.

t0yv0 (approved):

sdk/go/common/workspace/plugins.go
stack72: yeah that's exactly it - we should skip loading from PATH if that env var is set

PR #6966, opened 2021-05-04
Make virtualenv paths relative to root when main points elsewhere

### LINKED ISSUE #6844: Unexpected path resolution of virtualenv
First, thanks for pulumi, it's quite nice!

When using both `runtime:options:virtualenv` and `main` option, the virtualenv path is resolved relative to `main` and should be relative to the `Pulumi.yaml` file.

This is unexpected and is enough of an inconsistent UX that IMO is a bug (though you may disagree)

## Expected behavior

When defining relative path from a configuration file (here `Pulumi.yaml`), it is expected that they'll be resolved as relative to the parent folder of this file, not relative to each other.

## Current behavior

```
      File "/home/me/myproject/venv/bin/pulumi-language-python-exec", line 14, in <module>
        import pulumi
    ModuleNotFoundError: No module named 'pulumi'
    It looks like the Pulumi SDK has not been installed. Have you run pip install?
    If you are running in a virtualenv, you must run pip install -r requirements.txt from inside the virtualenv.
 
```

## Steps to reproduce

Given a project structure

```
├── infra/
│   └── __main__.py
├── venv/
└── Pulumi.yaml
```

With Pulumi.yaml

```yaml
name: myproject
description: Nice project
runtime:
  name: python
  options:
    virtualenv: venv
main: infra
```

make sure pulumi is installed

```shell
source venv/bin/activate
pip install pulumi
```

run pulumi

```shell
pulumi up
```

pulumi looks for a virualenv at `/home/me/myproject/infra/venv` but should find it in `/home/me/myproject/venv`

## Workaround

Given the project defined above,

```yaml
name: myproject
description: Nice project
runtime:
  name: python
  options:
    virtualenv: ../venv  # this shoudl do the trick
main: infra
```

---

**EDIT**
-  added workaround

t0yv0:
# Description

Fixes #6844 

## Checklist


t0yv0:
It's taken a surprisingly disruptive and hackish surgery to propagate the project location information to the language plugin (and I'm not showing my many failed attempts here that were even more invasive). Is this normal or am I missing some obvious solution?

sdk/go/common/resource/plugin/langruntime_plugin.go
```diff
@@ -59,6 +59,11 @@ func NewLanguageRuntime(host Host, ctx *Context, runtime string,
 	for k, v := range options {
 		args = append(args, fmt.Sprintf("-%s=%v", k, v))
 	}
+
+	if project.ConfigFileLocation != "" && runtime == "python" {
```
lukehoban, line 63: It's a bit unfortunate to have to do anything language specific at this level.  Should we always pass this through to language hots?

Also - should we pull this up to add as `options` to callers instead of handling it separately here?  It feels like this could be injected into language host options a couple levels higher than this?

t0yv0: Hmm, yes need another pass tomorrow.

I'll have a closer look at the callers. 

Notes to self: looking again, it seems that `options` are copied into `plugin.NewContext` from `projinfo.Proj.Runtime.Options()`. And `projinfo.Proj` comes straight out the YAML parser.  `NewContext` has call sites in the following places, some of the call sites set nil `options`:

```
pkg/cmd/pulumi/import.go
pkg/cmd/pulumi/policy_publish.go
pkg/codegen/hcl2/binder.go
pkg/codegen/schema/schema.go
pkg/engine/deployment.go
sdk/go/common/resource/plugin/context.go
```

Need to try to simplify this.

t0yv0: I've now removed `runtime == "python"` check at the cost of modifying all `pulumi-langage-*` main funcs to accept a new parameter. These programs fail if the parameter is not declared when invoked with the parameter. Only the python runtime actually uses the parameter value at the moment. 

I've opted not to send the new parameter to all options upstream, because it feels like it might require similar edits to resource plugins which becomes a bigger change.

t0yv0:
This attempt seems a little cleaner to me, there is an existing concept of `Projinfo.Root` and propagating it via plugins.Context, delivers it to the right place in the code with minimal surgery.

justinvp (approved):
LGTM

Unfortunately, this makes me think we have bugs elsewhere that will also require fixing up...

## 1. Policy Packs

For example, we have analyzer plugins for Policy Packs that do similar things:

This ends up passing a similar `-virtualenv` flag to the `pulumi-analyzer-policy-python` script:

https://github.com/pulumi/pulumi/blob/f378e6836788ab3e7b0fa910236781c8d8c0ebbc/sdk/go/common/resource/plugin/analyzer_plugin.go#L125-L130

- [Linux/Mac script](https://github.com/pulumi/pulumi/blob/f378e6836788ab3e7b0fa910236781c8d8c0ebbc/sdk/python/dist/pulumi-analyzer-policy-python)
- [Windows script](https://github.com/pulumi/pulumi/blob/f378e6836788ab3e7b0fa910236781c8d8c0ebbc/sdk/python/dist/pulumi-analyzer-policy-python.cmd)

Those scripts make use of the `-virtualenv` flag, and probably also need to be fixed to have a `-root` to find the proper directory. Maybe we should just open an issue to track following up to see if we need to do anything here.

## 2. Python Dynamic Providers

Dynamic providers are plugins that are just scripts that we distribute with the CLI:

- [Linux/Mac script](https://github.com/pulumi/pulumi/blob/f378e6836788ab3e7b0fa910236781c8d8c0ebbc/sdk/python/dist/pulumi-resource-pulumi-python)
- [Windows script](https://github.com/pulumi/pulumi/blob/f378e6836788ab3e7b0fa910236781c8d8c0ebbc/sdk/python/dist/pulumi-resource-pulumi-python.cmd)

These also get the virtualenv from the Project, but for resource plugins these are passed as `PULUMI_RUNTIME_<OPTION>` environment variables:

https://github.com/pulumi/pulumi/blob/f378e6836788ab3e7b0fa910236781c8d8c0ebbc/sdk/go/common/resource/plugin/provider_plugin.go#L87-L91

It seems like those scripts also need the root (probably passed as a new env variable), to be able to resolve the virtual environment correctly. Also probably worth filing an issue to follow-up separately.

t0yv0:
These are now

https://github.com/pulumi/pulumi/issues/7060
https://github.com/pulumi/pulumi/issues/7061

PR #6971, opened 2021-05-05
Permit override of Python package naming

jen20:
# Description

This commit adds a new language option for Python generation to specify the package name instead of using `pulumi_x` where x is the name defined in the schema.

This does not extend to docs, since I can't find a clean way to populate the doc helper struct from the context of a given language options set, and the docs bits are highly specific to Pulumi internals in any case (given the URL fragments generated).

## Checklist

This is not a user facing change, so likely does not need a CHANGELOG entry.


justinvp (approved):
LGTM. Thank you!

PR #6975, opened 2021-05-06
Using goreleaser as the publishing mechanism for Windows binaries

stack72:
This will allow us to ensure that our windows binaries are built
in the same way our linux and macOS binaries are - and more importantly
they will be included in our checksums

This PR also cleans up all of the old scripts that are no longer
required due to the removal of the custom windows publishing mechanism

This is the contents of the current v3.2.0 release

Screenshot 2021-05-06 at 13 02 21

This is the contents of a newly created goreleaser windows release

Screenshot 2021-05-06 at 13 03 57

AS you can see, the contents are the same - I will going to ensure that we have a nightly cron testing the templates using these new dev builds to make sure that things work as expected


leezen (approved):

stack72:
So I am not going to merge this until I have a set of nightly tests for templates that use the nightly SDK build

Issue #6974, opened 2021-05-06
Strange errors when using blank stack passphrase


Oliver-Sellwood:
Using Pulumi with a blank stack passphrase in GCP results in strange errors.

I am not 100% certain this is just for GCP. Another project I have worked on with `"@pulumi/pulumi": "^2.24.1"` in AWS does not have this issue.

## Expected behavior
Blank passphrase works the same as a non blank passphrase

## Current behavior
Running an initial `pulumi up` works as expected but any further interactions with the stack result in the folllowing error

```
error: constructing secrets manager of type "passphrase": unable to find either `PULUMI_CONFIG_PASSPHRASE` or `PULUMI_CONFIG_PASSPHRASE_FILE` when trying to access the Passphrase Secrets Provider; please ensure one of these environment variables is set to allow the operation to continue
```

## Steps to reproduce

1. echo 'export PULUMI_CONFIG_PASSPHRASE=""' >> ~/.bashrc
2. pulumi up
3. pulumi stack output kubeconfig

## Workaround
Set a stack password
```
echo 'export PULUMI_CONFIG_PASSPHRASE=secret' >> ~/.bashrc
```

## Context (Environment)

package.json
```
{
    "name": "project",
    "devDependencies": {
        "@types/node": "^10.0.0"
    },
    "dependencies": {
        "@pulumi/gcp": "^5.2.0",
        "@pulumi/kubernetes": "^3.1.1",
        "@pulumi/pulumi": "^3.2.0"
    }
}

```

index.ts
```
import * as k8s from "@pulumi/kubernetes";
import * as pulumi from "@pulumi/pulumi";
import * as gcp from "@pulumi/gcp";

const name = "helloworld";

// Create a GKE cluster
const engineVersion = gcp.container.getEngineVersions().then(v => v.latestMasterVersion);
const cluster = new gcp.container.Cluster(name, {
    initialNodeCount: 2,
    minMasterVersion: engineVersion,
    nodeVersion: engineVersion,
    nodeConfig: {
        machineType: "n1-standard-1",
        oauthScopes: [
            "https://www.googleapis.com/auth/compute",
            "https://www.googleapis.com/auth/devstorage.read_only",
            "https://www.googleapis.com/auth/logging.write",
            "https://www.googleapis.com/auth/monitoring"
        ],
    },
});

// Export the Cluster name
export const clusterName = cluster.name;

// Manufacture a GKE-style kubeconfig. Note that this is slightly "different"
[...]

PR #6976, opened 2021-05-06
[cli] Prevent regression in loading passphrase secrets provider from state

stack72:
Fixes: #6974

Passphrase Environment variables were set before loading the
secrets provider from state

Unfortunately, it seems that some users are using empty passphrases
and thus this newly introduced logic has broken their usecases

We now check that the environment variables are set - it doesn't
matter if they are set as empty, but the existance of an empty
environment variabe still suggests that it is an intentional
empty passphrase

### DIFF
--- a/pkg/secrets/passphrase/manager.go
+++ b/pkg/secrets/passphrase/manager.go
@@ -144,26 +144,26 @@ func NewPassphaseSecretsManager(phrase string, state string) (secrets.Manager, e
 
 // Tries to find the Passphrase first using `PULUMI_CONFIG_PASSPHRASE` then
 // `PULUMI_CONFIG_PASSPHRASE_FILE` if it is not found and defaulting to an empty string
-func getConfigPassphrase() (string, error) {
+func getConfigPassphrase() (string, bool, error) {
 	if passphrase, isOk := os.LookupEnv("PULUMI_CONFIG_PASSPHRASE"); isOk {
-		return passphrase, nil
+		return passphrase, true, nil
 	}
 
 	if phraseFile, isOk := os.LookupEnv("PULUMI_CONFIG_PASSPHRASE_FILE"); isOk {
 		phraseFilePath, err := filepath.Abs(phraseFile)
 		if err != nil {
-			return "", errors.Wrap(err, "unable to detect passphrase path")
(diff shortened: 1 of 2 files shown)


mikhailshilkov:
Do you have a link to the PR which introduced the regression?

stack72:
> Do you have a link to the PR which introduced the regression?

https://github.com/pulumi/pulumi/pull/6893

pkg/secrets/passphrase/manager.go
```diff
@@ -186,7 +186,7 @@ func NewPassphaseSecretsManagerFromState(state json.RawMessage) (secrets.Manager
 	// not. Ideally, we would prompt the user for the passphrase at this point but we can't do that in the CLI is in an
 	// update operation, so we should at least error out with an appropriate message here to ensure that the user
 	// understands why the operation fails unexpectedly.
-	if phrase == "" {
+	if !isFound && phrase == "" {
```
mikhailshilkov, line 189: Would `!isFound` be enough?

stack72: It *could* be but I am super keen to unblock these users - I can test that out and make it simpler once this P1 is fixed up

stack72: Ok, tested and you are correct that on;y the check for !isFound is necessary here - we don't care about the "" part

mikhailshilkov: Looking at the code, you never return `false` with a non-empty phrase, it seems. So, it looks like no risk to me. I'll leave it to your judgment.

mikhailshilkov (approved):

thiagosalvatore:
When is it going to be released? It is impacting our production pipeline

joeduffy:
Should we have a test to ensure empty passphrases continue working in the future?

stack72:
> Should we have a test to ensure empty passphrases continue working in the future?

I can follow up with that - our entire test framework sets one by default so I'll need to pull at that thread to make sure we can do this

I'll open an issue to make sure this is inplace

hakoerber:
Hey @stack72, thank you very much for the super fast fix!

Just pulled `3.2.1`, works like a charm.

stack72:
great news @hakoerber - thanks so much for verifying this for me!

Issue #6106, opened 2021-01-14
Relax usage of `readonly` in nodejs SDKs


EvanBoyle:
For instance CustomResourceArgs are all generated with readonly: https://github.com/pulumi/pulumi-aws/blob/0e6b9fc2f81e565df027eee7afb499d7a9617c7f/sdk/nodejs/dynamodb/table.ts#L373-L443

This makes it impossible to assign properties after initialization:

```ts
const dynamoArgs: TableArgs = args.props || {};
        dynamoArgs.billingMode = "PAY_PER_REQUEST"; // this will not compile
```

To get the above to compile you have to do some annoying type massaging:

```ts
type Writeable<T> = { -readonly [P in keyof T]: T[P] };

export type DynamoTableArgs = Writeable<TableArgs>;
const dynamoArgs: DynamoTableArgs = args.props || {};
 dynamoArgs.billingMode = "PAY_PER_REQUEST"; // this will not compile
```

EvanBoyle:
closed via #6980

PR #6980, opened 2021-05-06
Relax readonly requirement for generated nodejs resource args

EvanBoyle:
# Description

This change removes the `readonly` requirement for nodejs resource and function input args. The readonly requirement makes it unnecessarily difficult to do simple things like conditional argument construction:

```ts
let args: ResourceArgs = { someProp: "someValue" }
if (something) {
  args.someOtherProp = "someOtherValue"
}
```

I can't think of any way that this would impact or break users. Just seems like we're being overly-conservative here.

Fixes https://github.com/pulumi/pulumi/issues/6106

## Checklist

(no changelog as this shouldn't impact users at all)

### DIFF
--- a/pkg/codegen/nodejs/gen.go
+++ b/pkg/codegen/nodejs/gen.go
@@ -716,7 +716,7 @@ func (mod *modContext) genResource(w io.Writer, r *schema.Resource) error {
 	// Emit the state type for get methods.
 	if r.StateInputs != nil {
 		fmt.Fprintf(w, "\n")
-		mod.genPlainType(w, stateType, r.StateInputs.Comment, r.StateInputs.Properties, true, true, true, 0)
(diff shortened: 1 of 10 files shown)


pkg/codegen/nodejs/gen.go
```diff
@@ -722,7 +722,7 @@ func (mod *modContext) genResource(w io.Writer, r *schema.Resource) error {
 	// Emit the argument type for construction.
```
lukehoban, line 722: Should we make the same change to the state inputs?  (a couple lines above this).

lukehoban (approved):
LGTM.  I believe this is a safe (and good) change to make.

pkg/codegen/nodejs/gen.go
EvanBoyle: Good catch! Done.

PR #6983, opened 2021-05-07
Await outstanding async work in Go.

pgavlin:
The Pulumi Go SDK does not currently await all outstanding asynchronous
work associated with a Pulumi program. Because all relevant asynchronous
work is created via the Pulumi SDK, we can track this asynchronous work
and ensure that it has all completed prior to returning from
`Context.Run`.

This is complicated by the fact that many of the existing APIs that are
able to create `Output`s--`NewOutput`, `ToOutput`, `Any`,  `ToSecret`,
and `All`--do not have a `*Context` parameter, and so have no
straightforward way to associate themselves with a `*Context`. To address
this, these changes add new versions of each of these APIs as methods on
`*Context`.

Despite these new methods, most Pulumi programs should work without
changes: the bulk of `Output`s are created by the SDK itself as part of
resource registration, and for `Any` and `All`, we can pick up the
context from any `Output`s present in the arguments. The only programs
that should require changes are those that create outputs from whole
cloth using `NewOutput`, `ToOutput`, or `ToSecret` and create unawaited
async work rooted at those outputs.

On an implementation level, these changes track asynchronous work using
a `sync.WaitGroup` associated with each `*Context`. This `WaitGroup` is
passed to each output associated with the context. The SDK increments
this `WaitGroup`'s count prior to starting any asynchronous work and
decrements it once the work (including any callbacks triggered by the
work) is complete.

This fixes the Go portion of #3991.


t0yv0:
Speaking of horrifying implementation alternatives there is also:

https://github.com/jtolio/gls

It hacks goroutine-local storage on top of Go. Very much inferior to using explicit Context propagation but is an option for cases where refactoring to use Context propagation is prohibitive. Wanted to list it here for completeness of our options..

sdk/go/pulumi/provider.go
```diff
@@ -113,10 +113,9 @@ func construct(ctx context.Context, req *pulumirpc.ConstructRequest, engineConn
 		return nil, err
 	}
 
-	// Ensure all outstanding RPCs have completed before proceeding. Also, prevent any new RPCs from happening.
-	pulumiCtx.waitForRPCs()
-	if pulumiCtx.rpcError != nil {
-		return nil, errors.Wrap(pulumiCtx.rpcError, "waiting for RPCs")
+	// Wait for async work to finish.
+	if err = pulumiCtx.wait(); err != nil {
```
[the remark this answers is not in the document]
pgavlin: > Why should we wait for (presumably locally initiated) work to finish before returning from construct? There might be interesting work waiting to be scheduled until construct returns, that might have an opportunity to start in parallel if we do not wait here.

As designed, `Construct` builds a scope whose lifetime is bounded by the call to `Construct`. This design allows component authors and consumers to make some simplifying assumptions at the cost of parallelism (e.g. all resources associated with the component have been registered by the time `Construct` returns). I don't think that we're locked in to that approach long-term, but it does make things easier for the MVP.

sdk/go/pulumi/log.go
```diff
@@ -36,6 +37,7 @@ type Log interface {
 type logState struct {
 	engine pulumirpc.EngineClient
 	ctx    context.Context
+	join   *sync.WaitGroup
```
justinvp, line 40: Maybe I missed it, but where is this used?

sdk/go/pulumi/context.go
```diff
@@ -1177,6 +1188,67 @@ func (ctx *Context) Export(name string, value Input) {
...
+
+func (ctx *Context) newOutputState(elementType reflect.Type, deps ...Resource) *OutputState {
+	return newOutputState(&ctx.join, elementType, deps...)
+}
+
+func (ctx *Context) newOutput(typ reflect.Type, deps ...Resource) Output {
+	return newOutput(&ctx.join, typ, deps...)
+}
+
+// NewOutput creates a new output associated with this context.
```
justinvp, line 1202: Include the documentation about the returned funcs?

sdk/go/pulumi/types.go
```diff
@@ -240,6 +262,13 @@ func NewOutput() (Output, func(interface{}), func(error)) {
 	return AnyOutput{out}, resolve, reject
 }
 
+// NewOutput returns an output value that can be used to rendezvous with the production of a value or error.  The
+// function returns the output itself, plus two functions: one for resolving a value, and another for rejecting with an
+// error; exactly one function must be called. This acts like a promise.
+func NewOutput() (Output, func(interface{}), func(error)) {
```
justinvp, line 268: Do we want to augment the doc comments for these to mention the new methods on Context? Are the Context methods now preferable, and if so, are we eventually going to mark these as deprecated? Should we just go ahead and do that now?

pgavlin: > if so, are we eventually going to mark these as deprecated? Should we just go ahead and do that now?

I think we should, yes (meant to leave a review comment to that effect, d'oh). @lukehoban wdyt?

sdk/go/pulumi/log.go
pgavlin: Good catch. I added this to ensure that we don't drop logs, but forgot to write the code to use it. Done.

sdk/go/pulumi/provider.go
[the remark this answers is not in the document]
t0yv0: Interesting, thanks for that context. When that design naturally comes up for discussion next time I'd be interested to probe into why.

sdk/go/pulumi/types.go
lukehoban: Which methods exactly are moving?

Is it all the global methods here including `All`?  Some of these are very heavily used in typical Pulumi code, and I'm not sure it's "better" to ask users to write the alternative using context?  What does some real code look like after this change?  What exactly goes wrong if users continue to use the existing functions?

sdk/go/pulumi/run_test.go
```diff
@@ -385,3 +387,378 @@ func TestRegisterResourceWithResourceReferences(t *testing.T) {
...
+	err := RunErr(func(ctx *Context) error {
+		newResource(ctx, 10)
+		return nil
+	}, WithMocks("project", "stack", mocks))
+	assert.NoError(t, err)
+
+	assert.Equal(t, 10, resources)
+}
+
+func TestWaitOrphanedManualOutput(t *testing.T) {
```
t0yv0, line 712: This one is fun.

t0yv0 (approved):

sdk/go/pulumi/types.go
```diff
@@ -64,6 +78,8 @@ type OutputState struct {
 	mutex sync.Mutex
 	cond  *sync.Cond
 
+	join *sync.WaitGroup // the wait group associated with this output, if any.
```
t0yv0, line 81: OK so this is the meat. Since Output.getState() *OutputState, we now have `Output -> WaitGroup`, so presumably Apply chains propagate the wait group. 

I will use my beginner status to ask stupid questions freely.

# Zipping outputs across wait groups

What happens when zip join two outputs with different waitgroups?

```
All(a, b) // now what?

Given a.getState().join != b.getState().join
```

# Monadic bind across wait groups

Similar scenario, suppose you have `o Output<A>` and `f func(x A) Output<B>`, and we're computing `o2 = o.ApplyT(f)` to get `o2 Output<B>`.

```
func TestMonadicBind(t *testing.T) {
	o, resolveO, _ := NewOutput()
	p, resolveP, _ := NewOutput()

	// assume o.getState().join != p.getState().join

	resolveO(1)
	resolveP("OK")

	f := func(xval interface{}) Output {
		return p
	}

	o2 := o.ApplyT(f)

	// what is o2.getState().join ?

	v, known, secret, deps, err := await(o2)

	t.Errorf("v=%v known=%v secret=%v deps=%v err=%v", v, known, secret, deps, err)
}
```

If we have the answer to joining two wait groups we can apply the same answer here.


#  Which state to propagate

Is waitGroup the right (minimal) thing to propagate? Better than propagating Context?

sdk/go/pulumi/types.go
leezen: I think moving all the global methods would make sense, or at least the combinators. From a developer perspective, having them be context-less introduces the potential to wonder about if it's possible to pass in multiple outputs with different contexts and what underlying context the combinator then uses. When, in reality, I can't think of any scenario where one would do something like that and would expect all outputs to share the same context as the combinator. Moving the combinators to the context-level would make it fairly unambiguous that the combinator operates purely within that context. One could make a similar argument for the other methods (e.g. does `ToSecret` have some context that's different from the underlying `Output`'s?) so I would argue for moving them all to hang off context as opposed to leaving them at the top-level.

pgavlin: One interesting data point I just encountered: we use `pulumi.All` inside the implementation of the `Index`/`MapIndex` methods that we generate for array and map outputs. These methods do not have direct access to a `pulumi.Context`, though they do have indirect access via the receiver (since the receiver is an output). This pushes me a bit towards the side of only moving `NewOutput` onto `pulumi.Context`. From a technical standpoint, we could leave the combinators (`All`, `Any`, `ToOutput`, `ToSecret` and others we might eventually add) at the top-level and have them continue to take the strategy implemented in this PR of digging the appropriate wait group out of any output arguments.

pgavlin: That would scope the deprecations to only `NewOutput`.

Note that in general users should not be using combinators on outputs associated with different contexts. The only way to do that would be to escape outputs from multiple contexts via shared state. These changes do handle that situation as correctly as possible by bumping the outstanding work counter on all contexts, but it is very much an edge case.

sdk/go/pulumi/types.go
pgavlin: > Given a.getState().join != b.getState().join
> 
> Assume c = All(a, b) 
> 
> Now what do we have?
> 
> c.getState().join == ?

In this case we bump the counts on both a.join and b.join, create a new wait group for c.join, and fire off a goroutine that decrements the counts for a.join and b.join once c.join completes. This keeps a.join and b.join alive until all work associated with c (incl. any applies that are hung off of it) finishes. It's a little crude, but it gets the job done.

pgavlin: > If we have the answer to joining two wait groups we can apply the same answer here.

Interesting, yeah. I don't think that these changes handle that case--I'll have to think this scenario over a bit more closely. I do think that we _can_ handle it, though, with some care.

pgavlin: That said, I think that combining outputs from multiple contexts will be very uncommon, so I'm content to ship these changes without handling this case and follow up with a fix later.

t0yv0: Your solution on `All` sounds clever. So it implies some method to combine wait groups `combineWG(a WG, b WG): WG` that produces c.join out of a.join and b.join. 

For the monad case, presumably you can use the same method, but the big caveat is lifecycle. `f(x).join` is not going to be there util the computation resolves, essentially until the Output resolves. 

If you had a magic method that did time travel:

```
createDeferredWaitgroup(wg: Output<WaitGroup>): WG
```

Then you could hook things up with:

```
combineWG(x.join, createDeferredWaitgroup(f(x).Apply(\y-> y.join)))
```

pgavlin: I think that we could update the `fulfillValue` method to use a solution similar to `All`. Instead of deferring `o.join.Done()` [here](https://github.com/pulumi/pulumi/pull/6983/files#diff-192ea26149bdd1fec39b6476e5896ad8247830cfac2b0dbc47cf1f9f73313cb0R127-R130), we would fire off a grouting iff the fulfilled value is an `output` with a different wait group.

t0yv0: Tried a few test variations involving sleeping in apply; current implementation is holding up.

PR #6989, opened 2021-05-07
Cleanly exit when nodejs multi-lang provider encounters exception

EvanBoyle:
Draft of a fix for hangs when errors are thrown within `.apply` in provider component implementations.

### DIFF
--- a/sdk/nodejs/provider/server.ts
+++ b/sdk/nodejs/provider/server.ts
@@ -442,9 +442,12 @@ export async function main(provider: Provider, args: string[]) {
     const uncaughtHandler = (err: Error) => {
         if (!uncaughtErrors.has(err)) {
             uncaughtErrors.add(err);
-            // Use `pulumi.log.error` here to tell the engine there was a fatal error, which should
-            // stop processing subsequent resource operations.
-            log.error(err.stack || err.message || ("" + err));
+
+            // log the first error we see, kill any outstanding work, and exit.
+            log.error(err.stack || err.message || ("" + err)).then(() => {
+                runtime.disconnectSync();
+                process.exit(-1);
+            });
         }
     };
 


sdk/nodejs/provider/server.ts
```diff
@@ -442,9 +442,12 @@ export async function main(provider: Provider, args: string[]) {
     const uncaughtHandler = (err: Error) => {
         if (!uncaughtErrors.has(err)) {
             uncaughtErrors.add(err);
-            // Use `pulumi.log.error` here to tell the engine there was a fatal error, which should
-            // stop processing subsequent resource operations.
-            log.error(err.stack || err.message || ("" + err));
+
+            // log the first error we see, kill any outstanding work, and exit.
+            log.error(err.stack || err.message || ("" + err)).then(() => {
```
t0yv0, line 447: How does this interact with `process.on("exit")` at L455?

EvanBoyle: More specifically, the assumptions of the implementation without this change is that an uncaught exception/unhandled promise rejection would eventually cause a program failure after all outstanding work completes. I'm observing that this assumption is not totally correct, and that the outstanding work never completes.

t0yv0: Which sounds like a key correctness property is violated somewhere we have not found yet, and remains violated after this fix, which is unfortunate, though of course finding the right place can be hard.

EvanBoyle:
Just copying notes from slack:

> I did not intend on shepherding this fix through. Just intended for the draft PR to be informative for the team when someone picks up the investigation. In general, I think our strategy of trying to continue work after uncaught exceptions to be fraught with peril. I believe provider lifecycle is always scoped to an update, so in theory we have the option of halting outstanding work and exiting.

EvanBoyle:
closing this in favor of https://github.com/pulumi/pulumi/pull/7365

Issue #6377, opened 2021-02-18
[sdk/dotnet] Concurrency issue in WhileRunningAsync HandleCompletion


mikhailshilkov:
We received the following error in one of our [test runs](https://github.com/pulumi/pulumi-azure-nextgen-provider/pull/522/checks?check_run_id=1921033867#step:19:183) which means our task awaiting logic isn't entirely thread-safe. It doesn't happen often but that's how concurrency issues manifest... We should fix this.

```
Diagnostics:
pulumi:pulumi:Stack (cs-simple-p-it-fv-az59-79-cs-simple-c27e31cf):
   error: Running program '/tmp/p-it-fv-az59-79-cs-simple-c27e31cf-036146210/bin/Debug/netcoreapp3.1/cs-simple.dll' failed with an unhandled exception:
  System.Collections.Generic.KeyNotFoundException: The given key 'System.Runtime.CompilerServices.AsyncTaskMethodBuilder`1+AsyncStateMachineBox`1[System.Threading.Tasks.VoidTaskResult,Pulumi.Deployment+<RegisterResourceOutputsAsync>d__74]' was not present in the dictionary.
   at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
     at Pulumi.Deployment.Runner.<>c__DisplayClass9_0.<<WhileRunningAsync>g__HandleCompletion|0>d.MoveNext()
   --- End of stack trace from previous location where exception was thrown ---
  at Pulumi.Deployment.Runner.WhileRunningAsync()
```

Badabum:
Just got the same error

orionstudt:
could change the task dictionary to a `ConcurrentDictionary`

andrewdmoreno:
@mikhailshilkov apologize if you're not the right person to answer this but I think I am running into same issue. Seems like it was resolved, but my question is if there a specific version of a package that we need to be using that has this fixed?

t0yv0:
Could you open a new issue @andrewdmoreno with your dependency versions? Much appreciated.

PR #6993, opened 2021-05-07
Await outstanding async work in .NET.

pgavlin:
The Pulumi .NET SDK does not currently await all outstanding asynchronous
work associated with a Pulumi program. Because all relevant asynchronous
work is created via the Pulumi SDK, we can track this asynchronous work
and ensure that it has all completed prior to returning from
`Deployment.RunAsync`.

The implementation here is simpler than that in #6983, and re-uses the
existing support for tracking outstanding RPCs. If this proves to
negatively impact performance (which is a very real possibility for
programs that create many `Output` instances), we can simplify this
using a semaphore and a counter (essentially Go's `sync.WaitGroup`).

This fixes the .NET portion of #3991.


sdk/dotnet/Pulumi/Deployment/Deployment.cs
```diff
@@ -57,6 +58,18 @@ internal set
...
+            if (_instance.Value != null)
+            {
+                instance = _instance.Value.Internal;
+                return true;
+            }
+
+            instance = null;
+            return false;
+        }
+
```
vipentti, line 72: Since this is only used by the one place currently, this could be replaced with a more specific method for registering a task. E.g. something like

```csharp
// If we care about reporting whether we are able to register the task or not
internal static bool TryRegisterTask(string description, Task task)
{
    if (_instance.Value is null)
    {
        return false;
    }

    _instance.Value.Internal.Runner.RegisterTask(description, task);
    return true;
}

// or simply
internal static void RegisterTask(string description, Task task)
{
    if (_instance.Value != null)
    {
        _instance.Value.Internal.Runner.RegisterTask(description, task);
    }
}
```

sdk/dotnet/Pulumi/Deployment/Deployment.cs
```diff
@@ -57,6 +58,18 @@ internal set
             }
         }
 
+        internal static bool TryGetInternalInstance([NotNullWhen(true)] out IDeploymentInternal? instance)
+        {
+            if (_instance.Value != null)
```
t0yv0, line 63: Hmm, I think it's okay. If _instance was shared across threads here, we'd need to avoid races by reading it only once:

```
value = _instance.Value
```

However maybe if it's AsyncLocal I think it's race-proof as it is. Anyway there is super low contention here.

t0yv0 (approved):
LGTM - cleanly fits into existing machinery.

I would bet perf does not matter since we'd have 1E3's not 1E6's of outputs but if that's interesting we can add an output object counter as a metric?

mikhailshilkov:
> I would bet perf does not matter since we'd have 1E3's not 1E6's of outputs but if that's interesting we can add an output object counter as a metric?

This does make me worried. I remember hunting down slow updates in stacks with 1000+ resources in https://github.com/pulumi/pulumi/pull/5015

Now, we'll add tasks for each resource output which makes it an order of magnitude (?) slower. Worth a test?

Also, we should probably prioritize https://github.com/pulumi/pulumi/issues/6377 as it may become more likely to occur now.

Issue #6952, opened 2021-05-01
Tests fail to run from source due to external integration dependencies


gitfool:
I'm working through https://github.com/pulumi/pulumi/issues/4963#issuecomment-827312020 and at the point where I can build everything but some tests fail to run due to external integration dependencies.

## Expected behavior
All tests should pass out of the box, given environment specific configuration if required.

## Current behavior
```
make ensure
...
make only_build
...
make
...
  480 passing (4m)
  1 failing
  1) LocalWorkspace nested_config:
     CommandError: code: 255
 stdout: 
 stderr: error: could not create stack: [404] Not Found: Organization 'pulumi-test' not found
 err?: 
      at Object.createCommandError (bin/automation/errors.js:71:17)
      at ChildProcess.<anonymous> (bin/automation/cmd.js:63:40)
      at Process.ChildProcess._handle.onexit (internal/child_process.js:276:12)
Makefile:45: recipe for target 'istanbul_tests' failed
make[1]: *** [istanbul_tests] Error 1
make[1]: Leaving directory '/workspaces/pulumi/sdk/nodejs'
build/common.mk:206: recipe for target 'sdk/nodejs_default' failed
make: *** [sdk/nodejs_default] Error 2
```

## Steps to reproduce

1. from linux distro (ubuntu 20.04) in wsl2 via window terminal
2. `git clone https://github.com/pulumi/pulumi.git`
2. `cd pulumi`
3. `code .`
4. add `.devcontainer` folder from issue comment above
5. re-open in remote container
6. `make ensure` ... succeeds
7. `make only_build` ... succeeds
8. `make` ... fails as above

EvanBoyle:
The test org should be controlled via the environment variable `PULUMI_TEST_ORG`. Looks like this change was missed in the python and nodejs automation api implementations. 

https://github.com/pulumi/pulumi/blob/c3dc2d54abf85da811de0fbaa5e98f22b0f054c8/sdk/go/auto/local_workspace_test.go#L1772

EvanBoyle:
Also, the test org appears to be used inconsistently in node for some tests.

gitfool:
@EvanBoyle I pulled your latest changes and added the `PULUMI_TEST_ORG` env var to `.devcontainer/devcontainer.json` but the nodejs tests now fail with:
```
...
480 passing (4m)
  1 failing

  1) LocalWorkspace nested_config:
     CommandError: code: 255
 stdout: 
[...]

PR #6996, opened 2021-05-07
Set nodejs automation api test org from env

EvanBoyle:
# Description

We should be setting a test org for automation api tests. This updates nodejs auto tests to read from `PULUMI_TEST_ORG` env, or fall back to the default `pulumi-test`

Contributes to https://github.com/pulumi/pulumi/issues/6952 for nodejs

## Checklist


EvanBoyle:
Test failure due to unrelated dotnet multi-lang test.

komalali (approved):

PR #6999, opened 2021-05-08
[sdk/nodejs] Fix a bug in closure serialization.

### LINKED ISSUE #6908: Function serialization seems to execute code unexpectedly
Hey,

I'm trying to serialize a function that refers to a TypeORM entity that uses the TypeORM Active Record pattern.
For some reason, Pulumi just fails to serialize such an entity. It looks like Pulumi attempts to run some code behind the scenes for that entity which triggers an attempt to retrieve a connection to the database.

The code I have is very simple and easy to reproduce:

 ```ts
import * as pulumi from '@pulumi/pulumi';
import { BaseEntity } from 'typeorm';

(async () => {
  const val = await pulumi.runtime.serializeFunction(foo);
  console.log(val);
})();

class FooEntity extends BaseEntity {
  readonly id: string;
}

function foo() {
  const a = FooEntity;
}
```

When trying to execute this code the returned error message is:

>    error: Running program '/Users/kfir/code/integrations/features/xxx' failed with an unhandled exception:
    ConnectionNotFoundError: Connection "default" was not found.
        at new ConnectionNotFoundError (/Users/kfir/code/integrations/src/error/ConnectionNotFoundError.ts:8:9)
        at ConnectionManager.get (/Users/kfir/code/integrations/src/connection/ConnectionManager.ts:40:19)
        at Object.getConnection (/Users/kfir/code/integrations/src/index.ts:252:35)
        at Function.BaseEntity.getRepository (/Users/kfir/code/integrations/src/repository/BaseEntity.ts:99:72)
        at Function.get (/Users/kfir/code/integrations/src/repository/BaseEntity.ts:109:21)
        at /Users/kfir/code/integrations/features/xxx/node_modules/@pulumi/pulumi/runtime/closure/createClosure.js:1086:19
        at Generator.next (<anonymous>)
        at /Users/kfir/code/integrations/features/xxx/node_modules/@pulumi/pulumi/runtime/closure/createClosure.js:21:71
        at new Promise (<anonymous>)
        at __awaiter (/Users/kfir/code/integrations/features/xxx/node_modules/@pulumi/pulumi/runtime/closure/createClosure.js:17:12)

## Expected behavior

Pulumi should just “copy-paste” the code without trying to execute any of it.

## Current behavior

Pulumi somehow triggers some code under the hood that attempts to retrieve a connection and fails.

## Steps to reproduce

1. Create a new aws-typescript project (or really any typescript project)
2. Install NPM dependencies: `npm install typeorm`
3. Place the following code inside your `index.ts` file and try to execute it with `pulumi up`:
```ts
import * as pulumi from '@pulumi/pulumi';
import { BaseEntity } from 'typeorm';

(async () => {
  const val = await pulumi.runtime.serializeFunction(foo);
  console.log(val);
})();

class FooEntity extends BaseEntity {
  readonly id: string;
}

function foo() {
  const a = FooEntity;
}
```
## Affected Feature
https://www.pulumi.com/docs/intro/concepts/function-serialization/#function-serialization

swen128:
# Description

When a closure captures an object which has complex properties (e.g. getters / setters), Pulumi serializes the properties using the `defineProperty` function.

This does not work properly if a function object has complex properties.
For example, when a class has a static getter, the getter is evaluated at deploy time and the returned value is used to serialize the class.

```typescript
import {serializeFunction} from "@pulumi/pulumi/runtime";

class Class {
  static get foo() {
    return Math.random();
  }
}

serializeFunction(Class).then(f => {
  console.log(f.text);
});
```

The above script prints out the following serialized code (Notice the random value injected):

```javascript
exports.handler = __f0;

__f0.foo = 0.7598867621294356;

function __f0() {
  return (function() {
    with({  }) {

return function /*constructor*/() { };

    }
  }).apply(undefined, undefined).apply(this, arguments);
}
```

Fixes #6908 
The reported error occurs when Pulumi tries to serialize a getter [BaseEntity.target](https://github.com/typeorm/typeorm/blob/70938017ff17dc21e4a98512479b3645a78665e5/src/repository/BaseEntity.ts#L108)

## Checklist


kfirba:
Bravo!
Thanks for that PR! 👏🏻

kfirba:
@lukehoban Any plans on merging this? I'm stumbling across many issues with this, unfortunately

sdk/nodejs/tests/runtime/tsClosureCases.ts
```diff
@@ -1309,7 +1309,7 @@ return function () { console.log(v); };
 
 var __v = {};
 var __v_d1_proto = {};
-__f1.prototype = __v_d1_proto;
+Object.defineProperty(__f1, "prototype", { value: __v_d1_proto });
```
lukehoban, line 1312: The one thing I'm slightly nervous about is that this change will cause a lot of usage of closure serialization to change - which will lead to e.g. redeploying all Lambdas when this version of Pulumi is adopted. That's not necessarily a blocker - though it is a downside we'll need to message. 

I was initially wondering if this particular part of the change could be suppressed - but I do see that it is also technically incorrect to do what we were doing previously. 

How would you characterize the cases where serialization would change after this PR?  Is it "if you serialize a class definition, the serialization will be different"?

swen128: > How would you characterize the cases where serialization would change after this PR? Is it "if you serialize a class definition, the serialization will be different"?

That characterization sounds good to me.

This PR could also affect any function whose property is defined via `Object.defineProperty`, but I don't think there are many use cases other than class definition.

pgavlin:
@lukehoban @swen128 is this ready to go?

lukehoban (approved):

lukehoban:
> @lukehoban @swen128 is this ready to go?

Yes - I think we should merge this.  It will lead to some magic functions being redeployed even with no user changes, but we've generally been okay with that in minor versions in the past.

PR #7027, opened 2021-05-11
Update Dockerfile

komalali:
# Description

Update the pulumi base image to python 3.9. We updated our python and node-specific images but not this one.


stack72 (approved):

komalali:
@stack72 this won't work. It appears that there is no `python-3.9-slim-stretch` image and `python-3.9-slim-buster` doesn't have the required packages.

Issue #7026, opened 2021-05-11
Flaky LocalWorkspace correctly sets config on multiple stacks concurrently


t0yv0:
## Expected behavior

Tests are deterministic.

## Current behavior

```
  480 passing (3m)
  1 failing

  1) LocalWorkspace correctly sets config on multiple stacks concurrently:
     CommandError: code: 255
 stdout: Created stack 'dev2'

 stderr: error: unexpected end of JSON input

 err?: 

      at Object.createCommandError (bin/automation/errors.js:71:17)
      at ChildProcess.<anonymous> (bin/automation/cmd.js:63:40)
      at Process.ChildProcess._handle.onexit (internal/child_process.js:277:12)
make[1]: *** [istanbul_tests] Error 1
```

## Steps to reproduce

Sporadically on CI.

## Context (Environment)

Unrelated PR.

PR #7031, opened 2021-05-12
Create stacks serially before setting config concurrently

komalali:
# Description

The test in question was running `Stack.createOrSelect` concurrently which modifies global state by setting each stack [as current](https://github.com/pulumi/pulumi/blob/73a66f48ead301bacf1d7d2833c016cf803b57c1/pkg/cmd/pulumi/stack_init.go#L120) which can cause issues. This PR moves the stack creation out of the promise and sets config via the `ws` to avoid the global state corruption cause by `createStack` and `selectStack`.

Fixes https://github.com/pulumi/pulumi/issues/7026

If the test stops being flakey this PR has been successful. 

## Checklist

### DIFF
--- a/sdk/nodejs/tests/automation/localWorkspace.spec.ts
+++ b/sdk/nodejs/tests/automation/localWorkspace.spec.ts
@@ -478,12 +478,14 @@ describe("LocalWorkspace", () => {
             },
         });
         for (let i = 0; i < stacks.length; i++) {
+            await Stack.create(stacks[i], ws);
+        }
+        for (let i = 0; i < stacks.length; i++) {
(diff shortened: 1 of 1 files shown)


EvanBoyle (approved):

t0yv0:
Huh great find. This implicit state worries me. Could we rip it out of the API? The FP person in me thinks that'd be a win, though it is breaking at this point.

komalali:
Yeah it's definitely not great, but we can't take a breaking change right now. We could probably introduce a hidden flag in the CLI that doesn't set the stack to current on init and then use it in automation api to circumvent this issue.

PR #7049, opened 2021-05-13
Fix hanging deployments and improve errmsgs for programs with incorrect typings for output values

### LINKED ISSUE #6981: [sdk/python] Deployment hangs when output value's type doesn't match type annotation
When resolving outputs, if there is a type mismatch between the output value and the expected type from the type annotation, the deployment appears to hang, and canceling yields an unhelpful error message, e.g. something like:

```
  File "/Users/user/temp/proj/venv/lib/python3.8/site-packages/pulumi/runtime/rpc.py", line 634, in translate_output_properties
    raise AssertionError(f"Unexpected type; expected 'dict' got '{typ}'")
AssertionError: Unexpected type; expected 'dict' got 'typing.Sequence[pulumi_kubernetes.core.v1.outputs.Container]'
```

This can happen if inputs aren't typed correct (when the inputs are propagated as outputs) during preview, or when there is a bug in the provider itself (e.g. dynamic providers).

We should ensure such error messages bubble up and fail the deployment rather than hanging.

We should also thread through the property path and include it in the error message, and consider also including the output value.

Note that this issue has been around for a while since when we introduced typed input/output classes for nested data.

t0yv0:
# Description

Before the fix: test hanging.

With the fix:

```
Diagnostics:
  pulumi:pulumi:Stack (dynamic_broken_py-p-it-antons-mac-python-bro-a196afa0):
      File "/Users/anton/.nix-profile/bin/pulumi-language-python-exec", line 92, in <module>
        loop.run_until_complete(coro)
        return future.result()
      File "/Users/anton/pulumi/sdk/python/env/src/pulumi/runtime/stack.py", line 110, in run_in_stack
      File "/Users/anton/pulumi/sdk/python/env/src/pulumi/runtime/stack.py", line 45, in run_pulumi_func
        await wait_for_rpcs()
      File "/Users/anton/pulumi/sdk/python/env/src/pulumi/runtime/stack.py", line 94, in wait_for_rpcs
        raise exception
      File "/Users/anton/pulumi/sdk/python/env/src/pulumi/output.py", line 155, in run
        is_known = await self._is_known
      File "/Users/anton/pulumi/sdk/python/env/src/pulumi/runtime/rpc_manager.py", line 67, in rpc_wrapper
        result = await rpc
      File "/Users/anton/pulumi/sdk/python/env/src/pulumi/output.py", line 91, in is_value_known
        return await is_known and not contains_unknowns(await future)
      File "/Users/anton/pulumi/sdk/python/env/src/pulumi/output.py", line 91, in is_value_known
        return await is_known and not contains_unknowns(await future)
      File "/Users/anton/pulumi/sdk/python/env/src/pulumi/runtime/resource.py", line 537, in do_register
        rpc.resolve_outputs(res, resolver.serialized_props, resp.object, deps, resolvers, transform_using_type_metadata)
      File "/Users/anton/pulumi/sdk/python/env/src/pulumi/runtime/rpc.py", line 746, in resolve_outputs
        translated_value = translate_output_properties(value, translate_to_pass, types.get(key),
      File "/Users/anton/pulumi/sdk/python/env/src/pulumi/runtime/rpc.py", line 676, in translate_output_properties
        return {
      File "/Users/anton/pulumi/sdk/python/env/src/pulumi/runtime/rpc.py", line 678, in <dictcomp>
        translate_output_properties(v,
      File "/Users/anton/pulumi/sdk/python/env/src/pulumi/runtime/rpc.py", line 672, in translate_output_properties
        raise AssertionError((f"Unexpected type; expected a value of type `{typ}`"
    AssertionError: Unexpected type; expected a value of type `<class 'str'>` but got a value of type `<class 'dict'>` at resource `my_x`, property `x.my_key_1`: {'extra_buggy_key': 'my_value_1!'}
```

Fixes #6981 

## Checklist


sdk/python/lib/pulumi/runtime/resource.py
```diff
@@ -511,36 +511,41 @@ def do_rpc_call():
                     return None
 
             resp = await asyncio.get_event_loop().run_in_executor(None, do_rpc_call)
+
+            if resp is None:
```
t0yv0, line 515: The code block is simply moved under the try..except handler from below.

sdk/python/lib/pulumi/runtime/resource.py
```diff
@@ -511,36 +511,41 @@ def do_rpc_call():
...
+
+            rpc.resolve_outputs(res, resolver.serialized_props, resp.object, deps, resolvers, transform_using_type_metadata)
+
         except Exception as exn:
             log.debug(
                 f"exception when preparing or executing rpc: {traceback.format_exc()}")
             rpc.resolve_outputs_due_to_exception(resolvers, exn)
-            resolve_urn(None, True, False, exn)
-            if resolve_id is not None:
+            if not resolved_urn:
```
t0yv0, line 543: The handler is extended with resolved_urn, resolved_id flags so that in the happy case where we do resolve some of those before blowing up, we do not resolve them twice.

t0yv0:
I'll need to work out a test here. I've been using this intentionally broken example to repro: https://github.com/pulumi/examples/pull/1003/files

`pulumi destroy --yes && pulumi up --yes``

sdk/python/lib/pulumi/runtime/rpc.py
```diff
@@ -582,8 +583,41 @@ def translate_output_properties(output: Any,
     :param Optional[type] typ: The output's target type.
     :param bool transform_using_type_metadata: Set to True to use the metadata from `typ` to do name translation instead
                                                of using `output_transformer`.
+
+    :param Optional[Any]: Used internally to track recursive descent and enhance error messages.
```
justinvp, line 587:
```suggestion
    :param Optional[Any] path: Used internally to track recursive descent and enhance error messages.
```

sdk/python/lib/pulumi/runtime/rpc.py
```diff
@@ -582,8 +583,41 @@ def translate_output_properties(output: Any,
...
     """
 
+
+    def sub_path(prop_name: str) -> Any:
```
justinvp, line 591: Should we use a stronger type than `Any` and dict? How about a [`NamedTuple`](https://docs.python.org/3/library/typing.html#typing.NamedTuple)? Or is it because we don't want to "expose" a new type?

justinvp (approved):
Some comments, but LGTM otherwise. Thanks!

sdk/python/lib/pulumi/runtime/rpc.py
t0yv0: You know, I think this is my 'FP hacker' bad habits showing. I wanted an immutable linked list (to avoid quadratic `((([..] + [..]) + [..]) + [..]) + [..]` in case we ever have deep paths here, but Python does not have that in sdlib. I can clean this out a little and take it to private top-level _members.

t0yv0:
Hmm there is a failure on Windows in the newly introduced test,  panic: Test timed out after 60 seconds

t0yv0:
Back to investigating this as it definitely seems I've introduced a hang in one of the tests.

t0yv0:
@justinvp can you give me another look-over for me? IT turns out integration tests timing out was a real regression here so I reproduced locally, started the fix over, and was extra careful this time to keep both tests working.

I'd think we should check this in however something is iffy here. 

First is iffy thing is that symmetry is broken in this solution. That is `rpc.resolve_outpus` and `rpc.resolve_outputs_due_to_exception` started of as symmetric (use either or) but ended up having the `_due_to_exception` bugs catch exc from `rpc.resolve_outputs`.

Second the exception *was* being thrown *somewhere* out of this func. It feels to me that when some internal service (RpcManager?) blows up with an exception, we should cascade failure to all Pulumi components and wrap things up notifying the user. Sort of like in Erlang all process trees that used `spawn_link` fail together when something somewhere is busted. If we do want that behavior than perhaps the fix is in the wrong place?

If this resonates please give me a few pointers and I'll try to backlog this as something to come back to and fix better.

t0yv0:
At least we have good news that the checks passed in here 😅

justinvp (commented):
This only just occurred to me now and I'm really sorry for not mentioning it sooner... This PR addresses the problem in `register_resource`, but don't we have similar problems in `get_resource` (used when resolving a resource reference) and `read_resource` (used when looking up a resource, [for example](https://github.com/pulumi/pulumi-aws/blob/a9e64a7e354df3e23306b0e17b1093e1c9460f07/sdk/python/pulumi_aws/s3/bucket.py#L1508-L1509))?

> Iffy thing 1

The broken symmetry is unfortunate, but `rpc.resolve_outputs` is doing much more than `_due_to_exception`. I wonder if we need to refactor `resolve_outputs` to return the `all_properties` in its implementation rather than resolving the values (and possibly rename the function something like `prepare_outputs`?), and call this from inside the original try. Then have `resolve_properties` (possibly renamed to `resolve_outputs`) only resolve the values like it does currently, to be more like `_due_to_exception` and keep the symmetry. I'm assuming we don't currently raise from `resolve_properties`, but haven't looked closely enough to see if a raise is possible.

> Iffy thing 2

We may want to do something like this, but it sounds like a potentially even larger change that we probably don't want to bite off right now without planning for it.

t0yv0:
I'll backlog the refactor but I'll investigate for these other failures right away. Let's merge this as-is though so further fixes are next PRs.

justinvp (approved):

t0yv0:
Flaky test logged in https://github.com/pulumi/pulumi/issues/7355

PR #7051, opened 2021-05-14
[auto/dotnet] Fix deserialization of CancelEvent in .NET 5

### LINKED ISSUE #7047: [auto/dotnet] - System.Text.Json behavior change in .NET 5 breaks OnEngineEvent deserialization
.NET Pulumi libs are currently targetting `netcoreapp3.1` and depend on `System.Text.Json` for JSON serialization. In the 3.1 version of STJ it allows deserialization of classes that have non-public parameterless constructors, but in the 5.0 version of STJ it does not. This means that if a consumer is developing an application that targets `net5.0` and they are using STJ, than their version of STJ will supercede Pulumi's and we will fail deserialization when handling the `OnEngineEvent` delegate. The affected class is [Pulumi.Automation.Events.CancelEvent](https://github.com/pulumi/pulumi/blob/f378e6836788ab3e7b0fa910236781c8d8c0ebbc/sdk/dotnet/Pulumi.Automation/Events/CancelEvent.cs#L11) which is deserialized as part of [Pulumi.Automation.Serialization.EngineEventModel](https://github.com/pulumi/pulumi/blob/f378e6836788ab3e7b0fa910236781c8d8c0ebbc/sdk/dotnet/Pulumi.Automation/Serialization/EngineEventModel.cs#L10).

The following exception is thrown:
```
System.NotSupportedException: Deserialization of types without a parameterless constructor, a singular parameterized constructor, or a parameterized constructor annotated with 'JsonConstructorAttribute' is not supported. Type
'Pulumi.Automation.Events.CancelEvent'. Path: $.cancelEvent | LineNumber: 0 | BytePositionInLine: 54.
     System.NotSupportedException: Deserialization of types without a parameterless constructor, a singular parameterized constructor, or a parameterized constructor annotated with 'JsonConstructorAttribute' is not supported. Type
     'Pulumi.Automation.Events.CancelEvent'.
```

Here are links to relevant STJ issues:
- https://github.com/dotnet/runtime/pull/32213
- https://github.com/dotnet/docs/issues/20887

## Expected behavior
Successful `CancelEvent` deserialization regardless of targetted platform

## Current behavior
`NotSupportedException` if consuming application targets `net5.0`

## Steps to reproduce
1. Create an app that takes a dependency on dotnet `Pulumi.Automation`
2. Target `net5.0`
3. Do any pulumi update action with a provided `OnEngineEvent` delegate
4. Trigger a `CancelEvent` on your action so that it will be deserialized into the delegate

vipentti:
# Description

Fixes #7047 

## Checklist

Unfortunately the problem only happens when using .NET 5 so writing a suitable test currently is not possible afaik. I verified the change locally by using a minimal .NET 5 program and then calling `PreviewAsync` with the provided `OnEvent`-callback. Without the change the same error occurs as described in #7047. With the change the deserialization works as expected.

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -15,6 +15,9 @@
 - [auto/dotnet] - Disable Language Server Host logging and checking appsettings.json config
   [#7023](https://github.com/pulumi/pulumi/pull/7023)
 
+- [auto/dotnet] Fix deserialization of CancelEvent in .NET 5
+  [#7051](https://github.com/pulumi/pulumi/pull/7051)
+
 - [auto/python] - Export missing `ProjectBackend` type
   [#6984](https://github.com/pulumi/pulumi/pull/6984)
 
--- a/sdk/dotnet/Pulumi.Automation/Serialization/CancelEventModel.cs
+++ b/sdk/dotnet/Pulumi.Automation/Serialization/CancelEventModel.cs
@@ -1,0 +1,15 @@
+// Copyright 2016-2021, Pulumi Corporation
+
+using Pulumi.Automation.Events;
+using Pulumi.Automation.Serialization.Json;
+
+// NOTE: The classes in this file are intended to align with the serialized
+// JSON types defined and versioned in sdk/go/common/apitype/events.go
+namespace Pulumi.Automation.Serialization
+{
+    internal class CancelEventModel : IJsonModel<CancelEvent>
+    {
+        public CancelEvent Convert()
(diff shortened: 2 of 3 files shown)


t0yv0 (approved):
Looks like it's making it more consistent. LGTM

vipentti:
> /run-accceptance-tests

@t0yv0  Small typo in the command so the acceptance tests were not actually run

t0yv0:
Thanks a ton for this 🚢

PR #7059, opened 2021-05-14
Reify `Input` and `Optional` types in the schema type system.

### LINKED ISSUE #6957: Plain inputs typed as a type in the schema should not be typed as a plain type in the generated SDKs
It used to be possible to have a plain input `foo` typed as a type `xyz:index:Foo` (or as an array or map of `xyz:index:Foo`) where the generated code would type the plain property `foo` as an unwrapped `T` rather than `Input<T>`.

However, after #6686 and #6811, the codegen now emits a plain type for `xyz:index:Foo` and types the plain property `foo` as the plain type. So there's no way to accept a plain property as a type that itself has properties that are not plain.

pgavlin:
These changes support arbitrary combinations of input + plain types
within a schema. Handling plain types at the property level was not
sufficient to support such combinations. Reifying these types
 required updating quite a bit of code. This is likely to have caused
some temporary complications, but should eventually lead to
substantial simplification in the SDK and program code generators.

With the new design, input and optional types are explicit in the schema
type system. Optionals will only appear at the outermost level of a type
(i.e. Input<Optional<>>, Array<Optional<>>, etc. will not occur). In
addition to explicit input types, each object type now has a "plain"
shape and an "input" shape. The former uses only plain types; the latter
uses input shapes wherever a plain type is not specified. Plain types
are indicated in the schema by setting the "plain" property of a type spec
to true.


pgavlin:
@cnunciato it looks like a bunch of docs tests changed here--any idea why? I didn't touch anything in the docs generator directly...

cnunciato:
@pgavlin Not immediately sure either. I don't see this on master, but I do see it when I check out your branch and run the tests with `PULUMI_ASSERT`. Anything in your branch that would affect template rendering?

pgavlin:
> Anything in your branch that would affect template rendering?

There shouldn't be, no. The changes I made should be localized to SDK generation.

t0yv0:
Where do we define shallow/deep unwrapping? Looking at this PR, Foo and FooArgs in the example seem structurally identical. Would a difference appear if they referenced nested Output/Input props?

jen20:
This does not address the issue for Go, where `plain` container types still generate plain args classes.

justinvp (commented):
Something isn't right with these changes... I tried generating SDKs with these changes with the following schema:

```json
{
    "name": "xyz",
    "types": {
        "example:index:Foo": {
          "properties": {
            "a": {
              "type": "boolean"
            }
          },
          "type": "object"
        }
      },
    "resources": {
        "xyz:index:StaticPage": {
            "isComponent": true,
            "inputProperties": {
                "indexContent": {
                    "type": "string",
                    "description": "The HTML content for index.html."
                },
                "foo": {
                    "$ref": "#/types/example:index:Foo"
                }
            },
            "requiredInputs": [
                "indexContent"
            ],
            "properties": {
                "bucket": {
                    "$ref": "/aws/v4.0.0/schema.json#/resources/aws:s3%2Fbucket:Bucket",
                    "description": "The bucket resource."
                },
                "websiteUrl": {
                    "type": "string",
                    "description": "The website URL."
                }
            },
            "required": [
                "bucket",
                "websiteUrl"
            ],
            "plainInputs": ["foo"]
        }
    },
    "language": {
        "csharp": {
            "packageReferences": {
                "Pulumi": "3.*",
                "Pulumi.Aws": "4.*"
            }
        },
        "go": {
            "importBasePath": "github.com/pulumi/pulumi-xyz/sdk/go/xyz"
        },
        "nodejs": {
            "dependencies": {
                "@pulumi/aws": "^4.0.0"
            },
            "devDependencies": {
                "typescript": "^3.7.0"
            }
        },
        "python": {
            "requires": {
                "pulumi": ">=3.0.0,<4.0.0",
                "pulumi-aws": ">=4.0.0,<5.0.0"
            }
        }
    }
}
```

### .NET

The resource args refers to `FooArgs`:

```c#
public sealed class StaticPageArgs : Pulumi.ResourceArgs
{
    [Input("foo")]
    public Pulumi.Example.Inputs.FooArgs? Foo { get; set; }

    /// <summary>
    /// The HTML content for index.html.
    /// </summary>
    [Input("indexContent", required: true)]
    public Input<string> IndexContent { get; set; } = null!;

    public StaticPageArgs()
    {
    }
}
```

And `sdk/dotnet/Inputs/FooArgs.cs` is emitted, but it's empty (missing definition for `FooArgs`):

```c#
// *** WARNING: this file was generated by Pulumi SDK Generator. ***

using System;
using System.Collections.Immutable;
using System.Threading.Tasks;
using Pulumi.Serialization;

namespace Pulumi.Example.Inputs
{
}
```

### Node.js

Similarly, the resource's args refers to the `FooArgs` correctly:

```typescript
export interface StaticPageArgs {
    foo?: inputs.FooArgs;
    /**
     * The HTML content for index.html.
     */
    indexContent: pulumi.Input<string>;
}
```

And `sdk/nodejs/types/input.ts` is emitted, but it's empty (missing definition for `FooArgs`):

```typescript
// *** WARNING: this file was generated by Pulumi SDK Generator. ***

import { input as inputs, output as outputs } from "../types";
```

### Python

Similarly, the resource args refers to `FooArgs`:

```python
class StaticPage(pulumi.ComponentResource):
    @overload
    def __init__(__self__,
                 resource_name: str,
                 foo: Optional[pulumi.InputType['FooArgs']] = None,
                 index_content: Optional[pulumi.Input[str]] = None,
                 __props__=None):
```

And `sdk/python/pulumi_xyz/_inputs.py` is emitted, but it's empty (missing definition of `FooArgs`):

```python
# coding=utf-8
# *** WARNING: this file was generated by Pulumi SDK Generator. ***
# *** Do not edit by hand unless you're certain you know what you are doing! ***

import warnings
import pulumi
import pulumi.runtime
from typing import Any, Mapping, Optional, Sequence, Union, overload
from . import _utilities

__all__ = [
]
```

pgavlin:
>  This does not address the issue for Go, where plain container types still generate plain args classes.

This is now done.

> Something isn't right with these changes... I tried generating SDKs with these changes with the following schema:

Excellent catch, @justinvp. Fixed.

justinvp:
One more oddity. For this schema (basically the same as what I showed prior, but added `fooArray` and `fooMap` plain properties and removed the `indexContent` property):

```json
{
    "name": "xyz",
    "types": {
        "example:index:Foo": {
          "properties": {
            "a": {
              "type": "boolean"
            }
          },
          "type": "object"
        }
      },
    "resources": {
        "xyz:index:StaticPage": {
            "isComponent": true,
            "inputProperties": {
                "foo": {
                    "$ref": "#/types/example:index:Foo"
                },
                "fooArray": {
                    "type": "array",
                    "items": {
                        "$ref": "#/types/example:index:Foo"
                    }
                },
                "fooMap": {
                    "type": "object",
                    "additionalProperties": {
                        "$ref": "#/types/example:index:Foo"
                    }
                }
            },
            "properties": {
                "bucket": {
                    "$ref": "/aws/v4.0.0/schema.json#/resources/aws:s3%2Fbucket:Bucket",
                    "description": "The bucket resource."
                },
                "websiteUrl": {
                    "type": "string",
                    "description": "The website URL."
                }
            },
            "required": [
                "bucket",
                "websiteUrl"
            ],
            "plainInputs": ["foo", "fooArray", "fooMap"]
        }
    },
    "language": {
        "csharp": {
            "packageReferences": {
                "Pulumi": "3.*",
                "Pulumi.Aws": "4.*"
            }
        },
        "go": {
            "importBasePath": "github.com/pulumi/pulumi-xyz/sdk/go/xyz"
        },
        "nodejs": {
            "dependencies": {
                "@pulumi/aws": "^4.0.0"
            },
            "devDependencies": {
                "typescript": "^3.7.0"
            }
        },
        "python": {
            "requires": {
                "pulumi": ">=3.0.0,<4.0.0",
                "pulumi-aws": ">=4.0.0,<5.0.0"
            }
        }
    }
}
```

The Go output looks different from the other languages...

.NET:

```c#
    public sealed class StaticPageArgs : Pulumi.ResourceArgs
    {
        [Input("foo")]
        public Pulumi.Example.Inputs.FooArgs? Foo { get; set; }

        [Input("fooArray")]
        private ImmutableArray<Pulumi.Example.Inputs.FooArgs>? _fooArray;
        public ImmutableArray<Pulumi.Example.Inputs.FooArgs> FooArray
        {
            get => _fooArray ?? (_fooArray = new ImmutableArray<Pulumi.Example.Inputs.FooArgs>());
            set => _fooArray = value;
        }

        [Input("fooMap")]
        private ImmutableDictionary<string, Pulumi.Example.Inputs.FooArgs>? _fooMap;
        public ImmutableDictionary<string, Pulumi.Example.Inputs.FooArgs> FooMap
        {
            get => _fooMap ?? (_fooMap = new ImmutableDictionary<string, Pulumi.Example.Inputs.FooArgs>());
            set => _fooMap = value;
        }
}
```

Node.js:

```typescript
export interface StaticPageArgs {
    foo?: inputs.FooArgs;
    fooArray?: inputs.FooArgs[];
    fooMap?: {[key: string]: inputs.FooArgs};
}
```

Python:

```python
class StaticPageArgs:
    def __init__(__self__, *,
                 foo: Optional['FooArgs'] = None,
                 foo_array: Optional[Sequence['FooArgs']] = None,
                 foo_map: Optional[Mapping[str, 'FooArgs']] = None):
```

But here's Go:

```go
type StaticPageArgs struct {
	Foo      *FooArgs
	FooArray []FooInput
	FooMap   map[string]FooInput
}
```

To be consistent with the other languages, wouldn't it be the following?

```go
type StaticPageArgs struct {
	Foo      *FooArgs
	FooArray []FooArgs
	FooMap   map[string]FooArgs
}
```

pgavlin:
> To be consistent with the other languages, wouldn't it be the following?

Yes, you're correct. That's a straightforward fix, but it begs the question of which approach gives us what we want. The original behavior generated what we see in the non-Go languages, right?

justinvp:
> but it begs the question of which approach gives us what we want

This is a good question.

If a property is specified as plain and it's an array or map, my initial take is to make the collection value plain as well (i.e. the value isn't wrapped as `Input<T>` just like the overall collection isn't wrapped as `Input<T>`):

<table>
<tr>
<td> Schema </td> <td> TypeScript </td>
</tr>
<tr>
<td>

```json
{
    "inputProperties": {
        "fooArray": {
            "type": "array",
            "items": {
                "$ref": "#/types/example:index:Foo"
            }
        },
        "fooMap": {
            "type": "object",
            "additionalProperties": {
                "$ref": "#/types/example:index:Foo"
            }
        }
    },
    "plainInputs": ["fooArray", "fooMap"]
}
```

</td>
<td>

```typescript
export interface StaticPageArgs {
    fooArray?: inputs.FooArgs[];
    fooMap?: {[key: string]: inputs.FooArgs};
}
````

</td>
</tr>
</table>

However, if someone _did_ want the value wrapped as `Input<T>` (and the collection itself not wrapped), and we wanted to support this case, how would we enable it?

If we went the other way by default, and had the collection values wrapped:

<table>
<tr>
<td> Schema </td> <td> TypeScript </td>
</tr>
<tr>
<td>

```json
{
    "inputProperties": {
        "fooArray": {
            "type": "array",
            "items": {
                "$ref": "#/types/example:index:Foo"
            }
        },
        "fooMap": {
            "type": "object",
            "additionalProperties": {
                "$ref": "#/types/example:index:Foo"
            }
        }
    },
    "plainInputs": ["fooArray", "fooMap"]
}
```

</td>
<td>

```typescript
export interface StaticPageArgs {
    fooArray?: Input<inputs.FooArgs[]>;
    fooMap?: {[key: string]: Input<inputs.FooArgs>};
}
````

</td>
</tr>
</table>

With that, I could imagine introducing new `plainItems`/`plainInputItems` schema properties to control whether the value is wrapped or not:

<table>
<tr>
<td> Schema </td> <td> TypeScript </td>
</tr>
<tr>
<td>

```json
{
    "inputProperties": {
        "fooArray": {
            "type": "array",
            "items": {
                "$ref": "#/types/example:index:Foo"
            }
        },
        "fooMap": {
            "type": "object",
            "additionalProperties": {
                "$ref": "#/types/example:index:Foo"
            }
        }
    },
    "plainInputs": ["fooArray", "fooMap"],
    "plainInputItems": ["fooArray", "fooMap"]
}
```

</td>
<td>

```typescript
export interface StaticPageArgs {
    fooArray?: inputs.FooArgs[];
    fooMap?: {[key: string]: inputs.FooArgs};
}
````

</td>
</tr>
</table>

But is this too complicated?

**Comment by jen20** [comment]
@pgavlin Yes, will do :+1:

**Comment by pgavlin** [comment]
@stack72 looks like the azure-native job is hung setting up. Any way to re-run just that job?

**Comment by pgavlin** [comment]
So I _think_ that this is ready to go. As far as I can tell, any diffs that remain are due to legitimate fixes, though there _might_ be one more provider-related change I need to make. I'm going to try rearranging the commits for better clarity.

**Comment by pgavlin** [comment]
@justinvp PTAL; these changes are ready to go.

**Comment by pgavlin** [comment]
(cc @jen20 to give these another go)

File: pkg/codegen/internal/test/testdata/simple-enum-schema/go/plant/tree/v1/rubberTree.go [hunk 1]
```diff
@@ -64,19 +64,11 @@ func GetRubberTree(ctx *pulumi.Context,
 
 // Input properties used for looking up and filtering RubberTree resources.
 type rubberTreeState struct {
-	Container *plant.Container `pulumi:"container"`
-	Diameter  *float64         `pulumi:"diameter"`
-	Farm      *string          `pulumi:"farm"`
-	Size      *string          `pulumi:"size"`
-	Type      *string          `pulumi:"type"`
+	Farm *string `pulumi:"farm"`
```
justinvp, line 67: I know this is an expected change, but can you explain it for my own edification? Thanks!

pkg/codegen/internal/test/testdata/resource-options.pp.go
```diff
@@ -9,7 +9,7 @@ import (
 func main() {
 	pulumi.Run(func(ctx *pulumi.Context) error {
 		provider, err := providers.Newaws(ctx, "provider", &providers.awsArgs{
-			Region: pulumi.String("us-west-2"),
+			Region: "us-west-2",
 		})
```
justinvp, line 13: These lines seem suspect, but appears to be an existing issue (i.e. does `providers`, `Newaws` exist?And  I wouldn't expect `awsArgs` to be accessible).

Though, is the `pulumi.String` -> `string` intentional?

pkg/codegen/internal/test/testdata/simple-enum-schema/docs/tree/v1/rubbertree.md
```diff
@@ -239,7 +239,7 @@ The RubberTree resource accepts the following [input]({{< relref "/docs/intro/co
 <a href="#size_go" style="color: inherit; text-decoration: inherit;">Size</a>
 </span>
         <span class="property-indicator"></span>
-        <span class="property-type"><a href="#treesize">Tree<wbr>Size</a></span>
+        <span class="property-type"><a href="#treesize">string</a></span>
```
justinvp, line 242: Shouldn't this remain `TreeSize`?

pkg/codegen/internal/test/testdata/simple-enum-schema/docs/tree/v1/rubbertree.md
```diff
@@ -623,7 +623,7 @@ The following state arguments are supported:
 <a href="#brightness_go" style="color: inherit; text-decoration: inherit;">Brightness</a>
 </span>
         <span class="property-indicator"></span>
-        <span class="property-type"><a href="#containerbrightness">Container<wbr>Brightness</a></span>
+        <span class="property-type"><a href="#containerbrightness">float64</a></span>
```
justinvp, line 626: Shouldn't this remain `ContainerBrightness`?

pkg/codegen/internal/test/testdata/kubernetes-pod.pp.go
```diff
@@ -16,7 +16,7 @@ func main() {
 				Name:      pulumi.String("bar"),
 			},
 			Spec: &corev1.PodSpecArgs{
-				Containers: corev1.ContainerArray{
+				Containers: []corev1.ContainerArgs{
```
justinvp, line 19: Will this work?

justinvp (approved):
Some questions and comments, but LGTM otherwise.

justinvp:
This is the most recent automatic AWS diff link on this PR. Not sure if it's old, but looks suspicious.

https://github.com/pulumi-bot/pulumi-aws/compare/a9da55b2d..54305d8e1#diff-5888fbeb51d8c230f7aa02e47fefec595743f006e50a54517815ce224d19c06c

Screen Shot 2021-06-22 at 7 31 28 AM

Is it still happening with the latest commits, and if so, is it expected?

justinvp:
Same most recent AWS diff:

https://github.com/pulumi-bot/pulumi-aws/compare/a9da55b2d..54305d8e1#diff-149be5c8d8dee6b14fccac22a8d12aaa43e1efe8fdc0cd0208fb289d88a7423f

Screen Shot 2021-06-22 at 7 34 43 AM

The output type shouldn't have any `Input[T]` types.

t0yv0:
Hrm.. Not a PR review but adjacent comment: I'd like to understand this area better, if we have any docs around it... Like grammars for the embedded Type system. I have not yet put in a true effort to try to understand it from code to be honest so maybe I should go try that some more.

The idea of "reify Option[T] in the type system" resonates with me as a good idea but then we for some reason reject "Array[Option[T]]" forms even when some of the languages can represent them? Is that because we don't use those in practice?

On the practical side, something I ran into, regarding "plain" and "input" shapes. If we have `T` the plain type and `IT` an input-ty form of the type - then we may need a transformation `IT -> Input[T]` and `Input[T] -> IT`. Fully tested and supported ability to generate object code for these transformations in all languages would be a minor super-power. I'm ending up with code for this that I am testing it against exemplars in providers without confidence it works in full generality. Part of it is that I don't have a good grasp on the grammar of possible plain and input shapes.

pkg/codegen/schema/schema.go
```diff
@@ -219,6 +240,9 @@ func (t *ObjectType) Property(name string) (*Property, bool) {
 }
 
 func (t *ObjectType) String() string {
+	if t.PlainShape != nil {
+		return t.Token + "•Input"
```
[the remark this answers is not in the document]
pgavlin: Yes. This token should never be used. The name for an input-shaped-type should be derived from the token for its plain shape. The bullet is intended to help catch cases where this isn't happening, as it's generally an invalid identifier character in our target languages.

pkg/codegen/schema/schema.go
```diff
@@ -765,7 +817,8 @@ type ResourceSpec struct {
 	InputProperties map[string]PropertySpec `json:"inputProperties,omitempty"`
 	// RequiredInputs is a list of the names of the resource's required input properties.
 	RequiredInputs []string `json:"requiredInputs,omitempty"`
-	// PlainInputs is a list of the names of the resource's plain input properties that only accept prompt values.
+	// PlainInputs was a list of the names of the resource's plain input properties. This property is ignored:
+	// instead, property types should be marked as plain where necessary.
```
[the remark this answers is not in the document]
pgavlin: Ah, yes--I'll add one.

pkg/codegen/schema/schema.go
```diff
@@ -928,6 +981,7 @@ func importSpec(spec PackageSpec, languages map[string]Language, loader Loader)
 	}
 	for _, t := range types.objects {
 		typeList = append(typeList, t)
+		typeList = append(typeList, t.InputShape)
```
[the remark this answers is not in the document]
pgavlin: `types.objects` already contains all the plain shapes, so `append(typeList, t)` is already appending the plain shape. Would a comment be helpful here?

pkg/codegen/schema/schema.go
```diff
@@ -1664,16 +1795,16 @@ func bindResource(token string, spec ResourceSpec, types *types,
 	if len(spec.Plain) > 0 {
 		return nil, errors.New("plain cannot be specified on resources")
 	}
```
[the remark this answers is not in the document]
pgavlin: Ah, good idea. I'll change the error message.

pkg/codegen/go/gen_program_expressions.go
```diff
@@ -564,66 +563,87 @@ func (g *generator) GenUnaryOpExpression(w io.Writer, expr *model.UnaryOpExpress
...
+	//			tokenRange = expr.SyntaxNode().Range()
+	//		}
+	//	}
+
+	if schemaType, ok := hcl2.GetSchemaForType(destType); ok {
+		pkg := &pkgContext{pkg: &schema.Package{Name: "main"}}
+		return pkg.argsType(schemaType)
+	}
+
+	//	if _, isInput := schemaType.(*schema.InputType); ok && !isInput {
```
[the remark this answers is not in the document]
pgavlin: Yes, I will delete this. I wanted to keep it easily accessible while dealing with codegen diffs.

pkg/codegen/internal/test/testdata/aws-s3-folder.pp.go
```diff
@@ -66,7 +67,7 @@ func main() {
 					return _zero, err
 				}
 				json0 := string(tmpJSON0)
-				return pulumi.String(json0), nil
+				return json0, nil
```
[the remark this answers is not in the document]
pgavlin: Hah, the return type shouldn't be `pulumi.String`. The type-related bits of the Go code generator need some serious reexamination once these changes are landed: the reification changes + additional annotations should clean them up.

pkg/codegen/internal/test/testdata/aws-fargate.pp.go
```diff
@@ -176,10 +177,3 @@ func main() {
 		return nil
 	})
 }
-func toPulumiStringArray(arr []string) pulumi.StringArray {
```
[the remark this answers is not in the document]
pgavlin: Hm, no, that conversion should be present in this case. We should take care of this in follow-up.

pkg/codegen/internal/test/testdata/kubernetes-template.pp.go
```diff
@@ -16,13 +16,26 @@ func main() {
...
+					},
 					Spec: &corev1.PodSpecArgs{
 						Containers: corev1.ContainerArray{
 							&corev1.ContainerArgs{
+								Name:  pulumi.String("nginx"),
+								Image: pulumi.String("nginx"),
 								ReadinessProbe: &corev1.ProbeArgs{
 									HttpGet: &corev1.HTTPGetActionArgs{
-										Port: pulumi.Int(8080),
+										Port: pulumi.Any(8080),
```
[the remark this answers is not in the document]
pgavlin: `Port` is a `string | number` union, so this type is correct for Go.

pkg/codegen/internal/test/testdata/simple-enum-schema/go/plant/tree/v1/rubberTree.go
pgavlin: Go was incorrectly populating the contents of resource state objects from the resource's properties. The generator has been fixed s.t. it uses the schema's `StateInputs`, which is consistent with all other SDKs (and the intent of the schema).

pkg/codegen/internal/test/testdata/resource-options.pp.go
pgavlin: Yeah, this is definitely suspect. Another thing to follow up on w.r.t. Go codegen.

pkg/codegen/internal/test/testdata/simple-enum-schema/docs/tree/v1/rubbertree.md
pgavlin: That's a good question. For Go--which is the affected language here--I'm pretty sure that the "correct" type is `string`, but that might not be what we want to render for docs.

pkg/codegen/internal/test/testdata/simple-enum-schema/docs/tree/v1/rubbertree.md
pgavlin: Same answer as above.

pkg/codegen/internal/test/testdata/kubernetes-pod.pp.go
pgavlin: Nope, not sure why this changed.

pgavlin:
> The output type shouldn't have any Input[T] types.

I recall looking into this after posting the last set of changes, and IIRC the output type isn't actually used. I'll double-check to be sure.

pgavlin:
> Hrm.. Not a PR review but adjacent comment: I'd like to understand this area better, if we have any docs around it... Like grammars for the embedded Type system. I have not yet put in a true effort to try to understand it from code to be honest so maybe I should go try that some more.

Yes, better docs would be great here. In essence, the only restrictions are:
- `Input<Input<Input<...>>>` should simply be `Input<>`
- `Optional` only appears at the top level

The API is not correct-by-construction--i.e. nothing will stop you from manually creating nested optionals or inputs--but they will never appear in types derived from a schema (except for bugs!).

> If we have T the plain type and IT an input-ty form of the type - then we may need a transformation IT -> Input[T] and Input[T] -> IT

Can you provide a scenario for this? Shouldn't be difficult, but an example of the need would be good.

pgavlin:
> Is it still happening with the latest commits, and if so, is it expected?

@justinvp I believe that this is still happening, and IIRC it is expected b/c the prior codegen was incorrect.

justinvp:
> IIRC it is expected b/c the prior codegen was incorrect.

Can you explain why it's expected? I would expect this still to be wrapped as an `Input<string>` like all the other input properties. The type isn't marked as `plain`.

Screen Shot 2021-06-22 at 12 39 38 PM

The schema for it looks like:

```json
        "aws:apigateway/basePathMapping:BasePathMapping": {
            "description": "Connects a custom domain name registered via `aws.apigateway.DomainName`\nwith a deployed API so that its methods can be called via the\ncustom domain name.\n\n{{% examples %}}\n## Example Usage\n{{% example %}}\n\n```typescript\nimport * as pulumi from \"@pulumi/pulumi\";\nimport * as aws from \"@pulumi/aws\";\nimport * from \"fs\";\n\nconst exampleStage = new aws.apigateway.Stage(\"exampleStage\", {\n    deployment: aws_api_gateway_deployment.example.id,\n    restApi: aws_api_gateway_rest_api.example.id,\n    stageName: \"example\",\n});\nconst exampleDomainName = new aws.apigateway.DomainName(\"exampleDomainName\", {\n    domainName: \"example.com\",\n    certificateName: \"example-api\",\n    certificateBody: fs.readFileSync(`${path.module}/example.com/example.crt`),\n    certificateChain: fs.readFileSync(`${path.module}/example.com/ca.crt`),\n    certificatePrivateKey: fs.readFileSync(`${path.module}/example.com/example.key`),\n});\nconst exampleBasePathMapping = new aws.apigateway.BasePathMapping(\"exampleBasePathMapping\", {\n    restApi: aws_api_gateway_rest_api.example.id,\n    stageName: exampleStage.stageName,\n    domainName: exampleDomainName.domainName,\n});\n```\n```python\nimport pulumi\nimport pulumi_aws as aws\n\nexample_stage = aws.apigateway.Stage(\"exampleStage\",\n    deployment=aws_api_gateway_deployment[\"example\"][\"id\"],\n    rest_api=aws_api_gateway_rest_api[\"example\"][\"id\"],\n    stage_name=\"example\")\nexample_domain_name = aws.apigateway.DomainName(\"exampleDomainName\",\n    domain_name=\"example.com\",\n    certificate_name=\"example-api\",\n    certificate_body=(lambda path: open(path).read())(f\"{path['module']}/example.com/example.crt\"),\n    certificate_chain=(lambda path: open(path).read())(f\"{path['module']}/example.com/ca.crt\"),\n    certificate_private_key=(lambda path: open(path).read())(f\"{path['module']}/example.com/example.key\"))\nexample_base_path_mapping = aws.apigateway.BasePathMapping(\"exampleBasePathMapping\",\n    rest_api=aws_api_gateway_rest_api[\"example\"][\"id\"],\n    stage_name=example_stage.stage_name,\n    domain_name=example_domain_name.domain_name)\n```\n```csharp\nusing System.IO;\nusing Pulumi;\nusing Aws = Pulumi.Aws;\n\nclass MyStack : Stack\n{\n    public MyStack()\n    {\n        var exampleStage = new Aws.ApiGateway.Stage(\"exampleStage\", new Aws.ApiGateway.StageArgs\n        {\n            Deployment = aws_api_gateway_deployment.Example.Id,\n            RestApi = aws_api_gateway_rest_api.Example.Id,\n            StageName = \"example\",\n        });\n        var exampleDomainName = new Aws.ApiGateway.DomainName(\"exampleDomainName\", new Aws.ApiGateway.DomainNameArgs\n        {\n            DomainName = \"example.com\",\n            CertificateName = \"example-api\",\n            CertificateBody = File.ReadAllText($\"{path.Module}/example.com/example.crt\"),\n            CertificateChain = File.ReadAllText($\"{path.Module}/example.com/ca.crt\"),\n            CertificatePrivateKey = File.ReadAllText($\"{path.Module}/example.com/example.key\"),\n        });\n        var exampleBasePathMapping = new Aws.ApiGateway.BasePathMapping(\"exampleBasePathMapping\", new Aws.ApiGateway.BasePathMappingArgs\n        {\n            RestApi = aws_api_gateway_rest_api.Example.Id,\n            StageName = exampleStage.StageName,\n            DomainName = exampleDomainName.Domain,\n        });\n    }\n\n}\n```\n{{% /example %}}\n{{% /examples %}}\n\n## Import\n\n`aws_api_gateway_base_path_mapping` can be imported by using the domain name and base path, e.g. For empty `base_path` (e.g. root path (`/`))\n\n```sh\n $ pulumi import aws:apigateway/basePathMapping:BasePathMapping example example.com/\n```\n\n Otherwise\n\n```sh\n $ pulumi import aws:apigateway/basePathMapping:BasePathMapping example example.com/base-path\n```\n\n ",
            "properties": {
                "basePath": {
                    "type": "string",
                    "description": "Path segment that must be prepended to the path when accessing the API via this mapping. If omitted, the API is exposed at the root of the given domain.\n"
                },
                "domainName": {
                    "type": "string",
                    "description": "The already-registered domain name to connect the API to.\n"
                },
                "restApi": {
                    "type": "string",
                    "description": "The id of the API to connect.\n"
                },
                "stageName": {
                    "type": "string",
                    "description": "The name of a specific deployment stage to expose at the given path. If omitted, callers may select any stage by including its name as a path element after the base path.\n"
                }
            },
            "required": [
                "restApi",
                "domainName"
            ],
            "inputProperties": {
                "basePath": {
                    "type": "string",
                    "description": "Path segment that must be prepended to the path when accessing the API via this mapping. If omitted, the API is exposed at the root of the given domain.\n"
                },
                "domainName": {
                    "type": "string",
                    "description": "The already-registered domain name to connect the API to.\n"
                },
                "restApi": {
                    "type": "string",
                    "oneOf": [
                        {
                            "type": "string"
                        },
                        {
                            "type": "string",
                            "$ref": "#/types/aws:apigateway/restApi:RestApi"
                        }
                    ],
                    "description": "The id of the API to connect.\n"
                },
                "stageName": {
                    "type": "string",
                    "description": "The name of a specific deployment stage to expose at the given path. If omitted, callers may select any stage by including its name as a path element after the base path.\n"
                }
            },
            "requiredInputs": [
                "restApi",
                "domainName"
            ],
            "stateInputs": {
                "description": "Input properties used for looking up and filtering BasePathMapping resources.\n",
                "properties": {
                    "basePath": {
                        "type": "string",
                        "description": "Path segment that must be prepended to the path when accessing the API via this mapping. If omitted, the API is exposed at the root of the given domain.\n"
                    },
                    "domainName": {
                        "type": "string",
                        "description": "The already-registered domain name to connect the API to.\n"
                    },
                    "restApi": {
                        "type": "string",
                        "oneOf": [
                            {
                                "type": "string"
                            },
                            {
                                "type": "string",
                                "$ref": "#/types/aws:apigateway/restApi:RestApi"
                            }
                        ],
                        "description": "The id of the API to connect.\n"
                    },
                    "stageName": {
                        "type": "string",
                        "description": "The name of a specific deployment stage to expose at the given path. If omitted, callers may select any stage by including its name as a path element after the base path.\n"
                    }
                },
                "type": "object"
            }
        },
```

pkg/codegen/schema/schema.go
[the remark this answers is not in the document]
justinvp: Yeah, a comment would be helpful. Thanks.

pgavlin:
> Can you explain why it's expected? I would expect this still to be wrapped as an Input<string> like all the other input properties. The type isn't marked as plain.

Ah, damn, I thought the change to the plain type was on the outputs. The change to an optional type is expected, but the change to a plain type is not.

pkg/codegen/internal/test/testdata/simple-enum-schema/docs/tree/v1/rubbertree.md
pgavlin: @lukehoban @cnunciato any opinions on this one? My changes currently cause enum-typed properties to be rendered using their underlying type in the Go docs. Might be tricky to fix.

pgavlin:
@justinvp addressed your feedback; PTAL

PR #7086, opened 2021-05-19
Add gzip compression in filestate backend

### LINKED ISSUE #5469: Support optional compression of remote state JSON
When using a remote state backend, some resources might cause the state file to grown significantly, causing `pulumi up` to take 10s of minutes instead of seconds.

One such example is using `yaml.ConfigFile` to deploy [cert-manager](/jetstack/cert-manager)'s single YAML deployment to kubernetes.
The YAML file defines a series of kubernetes CRDs with accompanying documentation, which means it weights a hefty 1.6Mb. Applying it leads to a 10Mb JSON state file (from ~500kb), which gets read and written for each new resource defined in the file, causing the whole operation to take a significant amount of time. 

The remote state file is currently stored as plain JSON. Compressing it with gzip would already speed things up significantly.

awoimbee:
# Description
Add gzip compression in the filestate backend.
Compression can be turned off by setting `gzip` to `false` in the `localBackend` `struct` (default is `true`).
Fixes 

Fixes #5469

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -3,3 +3,6 @@
 - [cli] Fix a regression caused by [#6893](https://github.com/pulumi/pulumi/pull/6893) that stopped stacks created
   with empty passphrases from completing successful pulumi commands when loading the passphrase secrets provider.
   [#6976](https://github.com/pulumi/pulumi/pull/6976)
+- [backend] Add gzip compression to filestate backend.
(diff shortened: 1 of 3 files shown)


awoimbee:
This PR will improve be opened from another repository in the future.
In the meantime, I'm still interested in some guidance since this will be my first PR to this repo.

leezen:
@awoimbee Tests-wise, I think we'd want to see that this successfully writes both compressed and uncompressed files correctly and with the correct file extensions.

PR #7087, opened 2021-05-19
WIP: make codegen generate `${fn}Output` helper forms lifted to Input args and Output result

### LINKED ISSUE #5758: Functions/datasources don't compose well with Pulumi model - should accept Outputs

t0yv0:
# Description

Fixes #5758 - add generation of helper `${fn}Output` forms lifted to Input args and Output result.

- [x] should we emit code for funcs with no inputs or no outputs? - decided not to
- [x] typescript
  - [ ] generate doc comment?
  - [ ] inline the type into or the fundef or continue giving it distinct FuncOutputArgs name?
  - [ ] fix breakage detected on azure-native
  - [x] generated code passes runtime tests
  - [ ] pulumi-azure-native builds with the changes   
- [x] python
  - [ ] generate doc comment?  
  - [ ] generated code passes runtime tests
  - [ ] pulumi-azure-native builds with the changes   
- [x] dotnet
  - [x] generated code passes runtime tests
  - [x] pulumi-azure-native builds with the changes   
- [x] go
  - [x] discuss `${fn}Apply` naming
  - [x] generated code passes runtime tests
  - [x] pulumi-azure-native builds with the changes   

## Checklist


t0yv0:
Curiously, getting an error trying to test plain properties, courtesy of this code:

```
func bindFunction(token string, spec FunctionSpec, types *types) (*Function, error) {
	var inputs *ObjectType
	if spec.Inputs != nil {
		ins, err := types.bindObjectType(token+"Args", *spec.Inputs, false /*supportsPlainProperties*/)
```

Continue assuming plain properties are not allowed in FunctionSpec.

lukehoban:
Curious - why did we choose `getFooApply` over the proposed `getFooOutput`. The latter explains what's different to the user - it takes and returns outputs. The former explains how it's implemented - which feels harder to explain. The user question this feature addresses is "how do I pass an output to getFoo?" - so a name that answers that clearly seems ideal.

t0yv0:
Still open to correction of naming. The biggest problem is in Go codegen that is the most name-hungry.

https://github.com/pulumi/pulumi/pull/7087/files#diff-26e6320810312bc9a6fe1a3e39634bc577531a8a3aa4a9fc0325bdcbaff9df5d

In current version these are the names used by typical Go codegen:

```
[nix-shell:~/pulumi2/pkg/codegen/internal/test/testdata/output-funcs]$ grep '^func.[^(]\|type' listStorageAccountKeys.go  | grep -v //
grep '^func.[^(]\|type' listStorageAccountKeys.go  | grep -v //
func ListStorageAccountKeys(ctx *pulumi.Context, args *ListStorageAccountKeysArgs, opts ...pulumi.InvokeOption) (*ListStorageAccountKeysResult, error) {
type ListStorageAccountKeysArgs struct {
type ListStorageAccountKeysResult struct {
func ListStorageAccountKeysApply(ctx *pulumi.Context, args ListStorageAccountKeysApplyInput, opts ...pulumi.InvokeOption) ListStorageAccountKeysResultOutput {
type ListStorageAccountKeysApplyInput interface {
type ListStorageAccountKeysApplyArgs struct {
type ListStorageAccountKeysApplyOutput struct { *pulumi.OutputState }
type ListStorageAccountKeysResultOutput struct { *pulumi.OutputState }
func init() {
```

If we name it `ListStorageAccountKeysOutput` then we get names such as `ListStorageAccountKeysApplyOutputInput` and `ListStorageAccountKeysApplyOutputOutput`.

Using Apply sidesteps this problem, and I've moved the other languages accordingly.

There are however other ways forward, including fighting go codegen. I've been reluctant to go deep with changes to go codegen names as it seems to emit names that have to magically link up; it's not very aware of the target language structure but is rather a straight string emitter. We'd need more tests to get confidence it all aligns. That being sad, it may be worth it to do just that.

Another option is using `fnOutput` everywhere but special-casing things just for Go.

pkg/codegen/internal/test/testdata/output-funcs/dotnet/funcWithAllOptionalInputs.cs
```diff
@@ -0,0 +1,74 @@
...
+{
+    public static class FuncWithAllOptionalInputs
+    {
+        /// <summary>
+        /// Check codegen of functions with all optional inputs.
+        /// </summary>
+        public static Task<FuncWithAllOptionalInputsResult> InvokeAsync(FuncWithAllOptionalInputsArgs? args = null, InvokeOptions? options = null)
+            => Pulumi.Deployment.Instance.InvokeAsync<FuncWithAllOptionalInputsResult>("madeup-package:codegentest:funcWithAllOptionalInputs", args ?? new FuncWithAllOptionalInputsArgs(), options.WithVersion());
+
+        public static Output<FuncWithAllOptionalInputsResult> Apply(FuncWithAllOptionalInputsApplyArgs? args = null, InvokeOptions? options = null)
```
mikhailshilkov, line 20: I suppose this should be called `Invoke`

lukehoban:
> If we name it ListStorageAccountKeysOutput then we get names such as ListStorageAccountKeysApplyOutputInput and ListStorageAccountKeysApplyOutputOutput.

> Using Apply sidesteps this problem, and I've moved the other languages accordingly.

Got it.  Though this feels like the tail wagging the dog a bit (the names that are causing us to change this are esoteric things users will never use - vs. the name that users will interact with all the time)  I think we should find a way to make `Output` work.  

I don't think we actually need `ListStorageAccountKeysApplyInput` and `ListStorageAccountKeysApplyOutput` at all.  I think `ListStorageAccountKeysApply` (which should be `ListStorageAccountKeysOutput`) can take a  `ListStorageAccountKeysApplyArgs` (actually `ListStorageAccountKeysOutputArgs`) as an argument - which is much more consistent with the existing signatures.

t0yv0:
You are right! I was able to inline the implementation and get rid of extra types. In the last commit we have:

```
grep '^func.[^(]\|type' listStorageAccountKeys.go  | grep -v //
func ListStorageAccountKeys(ctx *pulumi.Context, args *ListStorageAccountKeysArgs, opts ...pulumi.InvokeOption) (*ListStorageAccountKeysResult, error) {
type ListStorageAccountKeysArgs struct {
type ListStorageAccountKeysResult struct {
func ListStorageAccountKeysOutput(ctx *pulumi.Context, args ListStorageAccountKeysOutputArgs, opts ...pulumi.InvokeOption) ListStorageAccountKeysResultOutput {
type ListStorageAccountKeysOutputArgs struct {
type ListStorageAccountKeysResultOutput struct { *pulumi.OutputState }
func init() {
```

Updating other languages to match.

pkg/codegen/internal/test/testdata/output-funcs/dotnet/funcWithAllOptionalInputs.cs
t0yv0: Good idea, implemented.

t0yv0:
Closed in favor of other PRs.

PR #7095, opened 2021-05-20
[dotnet/sdk] Use source context with serilog

### LINKED ISSUE #7093: [dotnet/sdk] Serilog logging should set `SourceContext`
I'm using `config.Development.yml` to configure Serilog logging in a generic host:
```yaml
Serilog:
  MinimumLevel:
    Override:
      Microsoft: Information
      Microsoft.Hosting.Lifetime: Warning
  WriteTo:
    - Name: Seq
      Args:
        serverUrl: http://localhost:5341
```
But I can't control the dotnet automation api logging since it doesn't generate a source context:

image

This should be a matter of using `ILogger<T>` instead of `ILogger` where type `T` becomes the source context, like the following:
```c#
public sealed partial class DeployCommand : AsyncCommandBase<DeployCommand.Settings>
{
    public DeployCommand(IOptions<Config> options, ILogger<DeployCommand> logger, IServiceProvider serviceProvider)
        : base(options, logger) { }

    protected override async Task<int> OnExecuteAsync(CommandContext context, Settings settings)
    {
        Logger.LogInformation("Deploying resources...");
        using var totalTimeLogger = new ElapsedTimeLogger(Logger, "Deployed resources");
        ...
    }
}

public abstract class AsyncCommandBase<TSettings> : AsyncCommand<TSettings> where TSettings : CommandSettings
{
    protected AsyncCommandBase(IOptions<Config> options, ILogger logger)
    {
        Config = options.Value;
        Logger = logger;
    }

    protected Config Config { get; }
    protected ILogger Logger { get; }
}
```

gitfool:
# Description

Fixes #7093.

@mikhailshilkov @orionstudt PTAL

## Checklist

### DIFF
--- a/sdk/dotnet/Pulumi/Deployment/Deployment.cs
+++ b/sdk/dotnet/Pulumi/Deployment/Deployment.cs
@@ -5,7 +5,10 @@ using System.Collections.Concurrent;
 using System.Diagnostics.CodeAnalysis;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Extensions.Configuration;
 using Pulumi.Testing;
+using Serilog;
+using Serilog.Events;
 
 namespace Pulumi
 {
@@ -82,6 +85,13 @@ namespace Pulumi
         private readonly bool _isDryRun;
         private readonly ConcurrentDictionary<string, bool> _featureSupport = new ConcurrentDictionary<string, bool>();
 
+#if NET5_0_OR_GREATER
+        private Serilog.ILogger _serilogger;
+#else
+#nullable disable
+        private Serilog.ILogger _serilogger;
+#nullable restore
+#endif
         private readonly ILogger _logger;
         private readonly IRunner _runner;
 
@@ -122,13 +132,15 @@ namespace Pulumi
             _stackName = stack;
             _projectName = project;
 
-            Serilog.Log.Debug("Creating Deployment Engine.");
-            this.Engine = new GrpcEngine(engine);
-            Serilog.Log.Debug("Created Deployment Engine.");
+            InitSerilogger();
 
-            Serilog.Log.Debug("Creating Deployment Monitor.");
+            _serilogger.Debug("Creating Deployment Engine.");
+            this.Engine = new GrpcEngine(engine);
+            _serilogger.Debug("Created Deployment Engine.");
+
+            _serilogger.Debug("Creating Deployment Monitor.");
             this.Monitor = new GrpcMonitor(monitor);
-            Serilog.Log.Debug("Created Deployment Monitor.");
+            _serilogger.Debug("Created Deployment Monitor.");
 
             _runner = new Runner(this);
             _logger = new Logger(this, this.Engine);
@@ -143,6 +155,7 @@ namespace Pulumi
         /// </summary>
         internal Deployment(IEngine engine, IMonitor monitor, TestOptions? options)
         {
+            InitSerilogger();
             _isDryRun = options?.IsPreview ?? true;
             _stackName = options?.StackName ?? "stack";
             _projectName = options?.ProjectName ?? "project";
@@ -156,6 +169,7 @@ namespace Pulumi
         string IDeployment.StackName => _stackName;
         bool IDeployment.IsDryRun => _isDryRun;
 
+        Serilog.ILogger IDeploymentInternal.Serilogger => _serilogger;
         ILogger IDeploymentInternal.Logger => _logger;
         IRunner IDeploymentInternal.Runner => _runner;
 
@@ -163,6 +177,25 @@ namespace Pulumi
         {
             get => Stack;
             set => Stack = value;
+        }
+
+#if NET5_0_OR_GREATER
+        [MemberNotNull(nameof(_serilogger))]
+#endif
+        private void InitSerilogger()
+        {
+            var verboseLogging = !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("PULUMI_DOTNET_LOG_VERBOSE"));
+
+            var configRoot = new ConfigurationBuilder()
+                .AddEnvironmentVariables()
+                .Build();
+
+            _serilogger = new LoggerConfiguration()
+                .MinimumLevel.Is(verboseLogging ? LogEventLevel.Verbose : LogEventLevel.Fatal)
+                .ReadFrom.Configuration(configRoot)
+                .WriteTo.Console()
+                .CreateLogger()
+                .ForContext<Deployment>();
         }
 
         private async Task<bool> MonitorSupportsFeature(string feature)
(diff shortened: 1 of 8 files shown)


sdk/dotnet/Pulumi/Deployment/Deployment_Run.cs
```diff
@@ -184,22 +194,18 @@ private static async Task<ImmutableArray<Resource>> TestAsync(IMocks mocks, Func
         // this method *must* remain marked async
         // in order to protect the scope of the AsyncLocal Deployment.Instance we cannot elide the task (return it early)
         // if the task is returned early and not awaited, than it is possible for any code that runs before the eventual await
-        // to be executed synchronously and thus have multiple calls to one of the Run methods affecting eachothers Deployment.Instance
+        // to be executed synchronously and thus have multiple calls to one of the Run methods affecting each others Deployment.Instance
         internal static async Task<int> CreateRunnerAndRunAsync(
             Func<Deployment> deploymentFactory,
             Func<IRunner, Task<int>> runAsync)
         {
-            var enableVerboseLogging = Environment.GetEnvironmentVariable("PULUMI_DOTNET_LOG_VERBOSE");
```
orionstudt, line 192: I don't think the declaration should be pulled out of here. Having the declaration in here means that consequent runs from Automation API will get their own setting.

For instance, by doing it in the static constructor if the first run in Automation API has verbose logging enabled than all subsequent runs in the same process will have it enabled and will have no control over it since we only check the environment variable once.

Now, it looks like the way our code currently existed also had a problem - since we were just instantiating over the static serilog logger it was entirely possible that we would overwrite the logger from a currently ongoing concurrent Automation API run. So that was a problem too.

To resolve both of these issues I think we should keep the environment variable check and the instantiation in this method, and we should see if we can refactor the logger to be provided into the DeploymentInstance constructor so that it belongs only to that run.

gitfool:
@orionstudt I've force pushed a change to move all Serilog static references to a `Deployment` [instance reference](https://github.com/pulumi/pulumi/blob/6cc0a25192af3a5d5581286558ad28941d0d2aa6/sdk/dotnet/Pulumi/Deployment/Deployment.cs#L86-L92) that's initialized in [Deployment.InitSerilogger](https://github.com/pulumi/pulumi/blob/6cc0a25192af3a5d5581286558ad28941d0d2aa6/sdk/dotnet/Pulumi/Deployment/Deployment.cs#L183). That's a good starting point but I think this will not work as you expect since this is still where it references the global static logger and/or overwrites it. I have an idea though...

gitfool:
@orionstudt I now I understand everything that was happening with my application and think these changes address the issues.

Pulumi was using the current global Serilog logger, unless `PULUMI_DOTNET_LOG_VERBOSE` was specified, in which case it overwrote the global Serilog logger and configured a minimum debug log level. Before the dotnet automation api, the current global Serilog logger would default to the internal [`SilentLogger.Instance`](https://github.com/serilog/serilog/blob/2aabe64a323ff449a2a483968e03d57e4d0bb19d/src/Serilog/Log.cs#L43) and so effectively there would be no Serilog logger output unless `PULUMI_DOTNET_LOG_VERBOSE` was specified. However, using the dotnet automation api, I'm creating a global Serilog logger at application bootstrap that writes to console, which had the side effect that Pulumi would then also generate Serilog output, even when `PULUMI_DOTNET_LOG_VERBOSE` was not specified.

These changes herein switch Pulumi to always create its own Serilog logger per `Deployment` [instance](https://github.com/pulumi/pulumi/blob/eabdd3a016c404773b25c79d91cfa2f4cd7f05f2/sdk/dotnet/Pulumi/Deployment/Deployment.cs#L88-L94) that's always initialized in [Deployment.InitSerilogger](https://github.com/pulumi/pulumi/blob/eabdd3a016c404773b25c79d91cfa2f4cd7f05f2/sdk/dotnet/Pulumi/Deployment/Deployment.cs#L185-L199), which sets the `MinimumLevel` accordingly, adds more flexibility via configuration through environment variables, and of course sets the source context!

sdk/dotnet/Pulumi/Pulumi.csproj
```diff
@@ -29,6 +29,10 @@
     <ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
   </PropertyGroup>
 
+  <ItemGroup>
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
```
orionstudt, line 33: I think this is a very big reference to add. What functionality is this change giving us? I didn't think any of your Serilog changes required an AspNetCore reference. I think we should be wary of this in the core lib

orionstudt: The only thing I can think of is the `using Microsoft.Extensions.Configuration;` that was added in `Deployment.cs`? If that's it I would just take which `Microsoft.Extensions.Configuration.*` dependencies we need directly instead of adding the framework reference.

sdk/dotnet/Pulumi/Deployment/Deployment_Run.cs
```diff
@@ -184,19 +182,11 @@ private static async Task<ImmutableArray<Resource>> TestAsync(IMocks mocks, Func
...
-            if (enableVerboseLogging != null && enableVerboseLogging != "")
-            {
-                Serilog.Log.Logger = new Serilog.LoggerConfiguration().MinimumLevel.Debug().WriteTo.Console().CreateLogger();
```
t0yv0, line 195: ❤️ removing global updates

t0yv0:
Thank you! Great change, and sounds like significant investigative work went into it.

Waiting on @orionstudt question about the necessity of the new framework dependency; then happy to run acceptance tests and merge up.

sdk/dotnet/Pulumi/Pulumi.csproj
gitfool: I was trying to avoid referencing a framework dependent version of [Microsoft.Extensions.Configuration.EnvironmentVariables](Microsoft.Extensions.Configuration.EnvironmentVariables).

gitfool:
@t0yv0 yeah I had to step through Serilog initialization. Gotta love dotnet debugging. Addressed concerns; ready for final review!

sdk/dotnet/Pulumi/Deployment/Deployment.cs
```diff
@@ -82,6 +85,9 @@ internal static IDeploymentInternal InternalInstance
         private readonly bool _isDryRun;
         private readonly ConcurrentDictionary<string, bool> _featureSupport = new ConcurrentDictionary<string, bool>();
 
+#nullable disable // remove when MemberNotNull available
+        private Serilog.ILogger _serilogger;
+#nullable restore
```
vipentti, line 90: Maybe this could be simplified to, if the idea is to workaround "Non-nullable field '_serilogger' must contain a non-null value when exiting constructor"

```suggestion
        private Serilog.ILogger _serilogger = null!;
```

t0yv0 (approved):

sdk/dotnet/Pulumi/Deployment/Deployment.cs
gitfool: Good idea; that worked.

gitfool:
Unrelated failure: [pulumi/pulumi/runs/2642199786](https://github.com/pulumi/pulumi/runs/2642199786?check_suite_focus=true])
> Error: connect ECONNREFUSED 52.241.88.84:443

t0yv0:
Rerunning, it's probably fine.

PR #7097, opened 2021-05-20
Allow editable installs of dev Python packages

jen20:
# Description

This commit adds a fallback for the README definition in the generated setup.py files for Python SDKs, thus allowing editable installs of packages which not yet been built.

## Checklist

Codegen tests do not appear to test setup.py generation.

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -15,6 +15,8 @@
   [#7079](https://github.com/pulumi/pulumi/pull/7079)
   [#7080](https://github.com/pulumi/pulumi/pull/7080)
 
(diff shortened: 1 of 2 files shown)


lukehoban (approved):
The change itself LGTM.

Is this the only issue that currently prevents the source Python package from being installed?

jen20:
@lukehoban Yes - the missing README gave an ENOENT and fails when trying to install the package from source in place (useful for examples in a component monorepo etc), so adding an obviously-fake README seems like the best fallback (rather than encoding path information to read from the repository root, or copying the top-level README during generation). There's a second issue which I'll follow up with relating to the version placeholder, but that actually does not block installation.

PR #7102, opened 2021-05-21
Use input type variants when creating discriminated unions

### LINKED ISSUE #7101: [programgen/go] Discriminated union creation code doesn't compile
Since https://github.com/pulumi/pulumi/pull/6430 we support discriminated unions. However, Go programgen code instantiating discriminated types fails to compile. Specifically, the code doesn't use the `...Args` input variants for the underlying types. See https://www.pulumi.com/docs/reference/pkg/azure-native/documentdb/databaseaccount/#cosmosdbdatabaseaccountcreatemax for an eample.

## Expected behavior
programgen generates valid code

## Current behavior
```
BackupPolicy: documentdb.PeriodicModeBackupPolicy{
	PeriodicModeProperties: documentdb.PeriodicModeProperties{
		BackupIntervalInMinutes:        240,
		BackupRetentionIntervalInHours: 8,
	},
	Type: "Periodic",
},
```
when in fact it should be:
```
BackupPolicy: &documentdb.PeriodicModeBackupPolicyArgs{
	PeriodicModeProperties: &documentdb.PeriodicModePropertiesArgs{
		BackupIntervalInMinutes:        pulumi.Int(240),
		BackupRetentionIntervalInHours: pulumi.Int(8),
	},
	Type: pulumi.String("Periodic"),
},
```

viveklak:
# Description
Use input type variants when creating discriminated unions in go programgen.

Fixes #7101 

## Checklist

Tested manually since azure-native tests are currently disabled.

No changelog required.


mikhailshilkov:
A test seems to be failing? Related - should we add a test for change? (maybe the fix of (1) does (2))

squaremo:
Looks like this rebases cleanly, for the minute -- and that the problem this solves is still present. Could it be revived @viveklak? I can review it. (As a side-note: I reckon it's worth recapitulating the explanation given at #7101 in the commit message, so it's right there in the git history.)

viveklak:
> Looks like this rebases cleanly, for the minute -- and that the problem this solves is still present. Could it be revived @viveklak? I can review it. (As a side-note: I reckon it's worth recapitulating the explanation given at #7101 in the commit message, so it's right there in the git history.)

Apologies - I can't recall the details of this PR anymore. Fairly certain there were some nuances I hit while fixing the failing test so I think there is more work to be done here. I am happy to close this PR if someone closer to programgen is willing to take over (@AaronFriel/ @iwahbe / @dixler perhaps?). I don't have bandwidth for this at the moment unfortunately.

Issue #7106, opened 2021-05-21
nodejs/typescript sdk doesn't process Map type field in resource args


Dmitry-Danchenko:
nodejs/typescript sdk doesn't process Map type field in resource args.

I'm trying to create Kafka topic using typescript SDK:

<code>
import * as kafka from "@pulumi/kafka";

new kafka.Topic("test-topic", {
    name : "test-topic",
    partitions: 4,
    replicationFactor: 1,
    config: new Map<string, any>([["retention.ms", "1200000"]])
  });
 
</code>

## Expected behavior

It should create topic with config: {"retention.ms", "1200000"}

```bash
pulumi up --diff
Previewing update (dev):
  pulumi:pulumi:Stack: (same)
    [urn=urn:pulumi:dev::kafka_migrate::pulumi:pulumi:Stack::kafka_migrate-dev]
    + kafka:index/topic:Topic: (create)
        [urn=urn:pulumi:dev::kafka_migrate::kafka:index/topic:Topic::test-topic]
        [provider=urn:pulumi:dev::kafka_migrate::pulumi:providers:kafka::kafka-dmp::c8998bd0-719a-46ef-8c9b-6d983fbadfbd]
        config           : {
            retention.ms: "1200000"
        }
        name             : "test-topic"
        partitions       : 4
        replicationFactor: 1
Resources:
    + 1 to create
    2 unchanged
```

## Current behavior

It creates Topic with an empty config:

```bash
pulumi up --diff
Previewing update (dev):
  pulumi:pulumi:Stack: (same)
    [urn=urn:pulumi:dev::kafka_migrate::pulumi:pulumi:Stack::kafka_migrate-dev]
    + kafka:index/topic:Topic: (create)
        [urn=urn:pulumi:dev::kafka_migrate::kafka:index/topic:Topic::test-topic]
        [provider=urn:pulumi:dev::kafka_migrate::pulumi:providers:kafka::kafka-dmp::c8998bd0-719a-46ef-8c9b-6d983fbadfbd]
        config           : {
        }
        name             : "test-topic"
        partitions       : 4
        replicationFactor: 1
Resources:
    + 1 to create
    2 unchanged
```

## Steps to reproduce

https://github.com/Dmitry-Danchenko/pulumi-kafka-bug-example

## Context (Environment)

I can't correctly create/manage Kafka Topics.

## Affected feature

Kafka plugin isn't working properly:
https://www.pulumi.com/docs/intro/cloud-providers/kafka/

[...]

PR #7107, opened 2021-05-21
fix-serialize-property-map

Dmitry-Danchenko:
# Description

nodejs/typescript sdk doesn't process Map type field in resource args

Fixes # (issue)

[#7106] https://github.com/pulumi/pulumi/issues/7106

## Checklist

### DIFF
--- a/sdk/nodejs/runtime/rpc.ts
+++ b/sdk/nodejs/runtime/rpc.ts
@@ -419,6 +419,19 @@ export async function serializeProperty(ctx: string, prop: Input<any>, dependent
         return result;
     }
 
+    if (prop instanceof Map) {
+        const result: any = {};
+        for (const k of prop.keys()) {
+            if (excessiveDebugOutput) {
+                log.debug(`Serialize property [${ctx}]: map[${k}] element`);
+            }
+            // When serializing maps, we serialize any undefined values as `null`. This matches JSON semantics.
+            const elem = await serializeProperty(`${ctx}[${k}]`,prop.get(k), dependentResources);
+            result[k] = (elem === undefined ? null : elem);
+        }
+        return result;
+    }
+
     return await serializeAllKeys(prop, {});
 
     async function serializeAllKeys(innerProp: any, obj: any) {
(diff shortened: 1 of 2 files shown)


sdk/nodejs/runtime/rpc.ts
```diff
@@ -419,6 +419,19 @@ export async function serializeProperty(ctx: string, prop: Input<any>, dependent
         return result;
     }
 
+    if (prop instanceof Map) {
```
justinvp, line 422: Our SDK code generator for Node.js emits map types as `{[key: string]: T}` rather than `Map<string, T>`, so it's unclear to me if support for serializing `Map`s is necessary.

https://github.com/pulumi/pulumi/blob/67bcd73a15cf718e1ac2dccddb6d5d5bb3d9504d/pkg/codegen/nodejs/gen.go#L216-L217

sdk/nodejs/runtime/rpc.ts
```diff
@@ -419,6 +419,19 @@ export async function serializeProperty(ctx: string, prop: Input<any>, dependent
...
+        const result: any = {};
+        for (const k of prop.keys()) {
+            if (excessiveDebugOutput) {
+                log.debug(`Serialize property [${ctx}]: map[${k}] element`);
+            }
+            // When serializing maps, we serialize any undefined values as `null`. This matches JSON semantics.
+            const elem = await serializeProperty(`${ctx}[${k}]`,prop.get(k), dependentResources);
+            result[k] = (elem === undefined ? null : elem);
```
justinvp, line 430: This doesn't look right. I just tried it out and calling `JSON.stringify` on a `Map` with a `undefined` value, and the entire key/value pair is excluded from the resulting JSON string.

```
$ node
Welcome to Node.js v14.16.1.
Type ".help" for more information.
> const m = new Map();
undefined
> m.set("foo", undefined);
Map(1) { 'foo' => undefined }
> JSON.stringify(m);
'{}'
```

If we decide we do want to add support for serializing `Map`, I think we want to behave the same as `serializeAllKeys` below.

PR #7115, opened 2021-05-23
Send plugin install output to stderr

joeduffy:
We currently send plugin install output to stdout. This interferes
with --json (#5747), automation API scenarios, and in general is bad
CLI hygiene. This change sends plugin output to stdout instead.

Two questions:

1. Is there a reason this isn't the preferred fix? I can easily believe
  so, since we were exploring alternative ideas in the linked issue.

2. How should I test this? Where are the current plugin installation
  test cases? I clearly can't add a dependency on AWS here -- is
  there a "simpler" plugin that I can install for test purposes?

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -3,3 +3,6 @@
 
 ### Bug Fixes
 
+- [cli] - Send plugin install output to stderr, so that it doesn't
+  clutter up --json, automation API scenarios, and so on.
+  [#7115](https://github.com/pulumi/pulumi/pull/7115)
--- a/pkg/engine/plugins.go
+++ b/pkg/engine/plugins.go
@@ -16,6 +16,7 @@ package engine
 
 import (
 	"fmt"
+	"os"
 	"sort"
 
 	"github.com/blang/semver"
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,14 @@
 CHANGELOG
 =========
 
+## Unreleased
```
stack72, line 4: These need to go into CHANGELOG_PENDING rather than CHANGELOG as we automate the release notes from that file

EvanBoyle (approved):
The fix looks good to me. I can't think of any reason why we wouldn't do this. 

I'm not aware of any e2e tests we have for plugin installation. You could probably pattern match against some of the other e2e CLI tests we have for things like history and install a simple plugin like random: https://github.com/pulumi/pulumi/blob/f6fea7fafb2a9def1eb7eee3c6e7cb02007574b8/tests/history_test.go#L72

Since installing plugins modifies global state, will probably want to set PULUMI_HOME to a tmp dir.

joeduffy:
> You could probably pattern match against some of the other e2e CLI tests we have for things like history and install a simple plugin like random

I thought of that, but doesn't that conceptually add a circular dependency in our repos? That seems...yucky. I had hoped we would have a "no-op" plugin or something we could use for test purposes. I'm kind of surprised we don't have any plugin tests! Since we don't have anything, I may just go as-is, and file an issue to improve our test coverage here. Before giving up and committing, @justinvp I know you've done most work in the plugin system, am I missing an obvious way to test this out?

PR #7136, opened 2021-05-25
Fix non-deterministic ordering of generated Python code

t0yv0:
# Description

Currently the JSON literals generated for Python codegen do not have consistent ordering of list elements. This is unfortunate for our CI as repeat builds generate code diffs. The fix sorts the data before splicing it into JSON.

## Checklist

### DIFF
--- a/pkg/codegen/python/gen_resource_mappings.go
+++ b/pkg/codegen/python/gen_resource_mappings.go
@@ -18,6 +18,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"sort"
 )
 
 // Generates code to build and regsiter ResourceModule and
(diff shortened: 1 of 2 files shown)


pkg/codegen/python/gen_resource_mappings.go
```diff
@@ -88,6 +95,7 @@ func allResourceModuleInfos(root *modContext) []resourceModuleInfo {
 	for _, mctx := range root.walkSelfWithDescendants() {
 		result = append(result, collectResourceModuleInfos(mctx)...)
 	}
+	sort.Sort(resourceModuleInfoByFqn(result))
```
lukehoban, line 98: Can this use `sort.Slice` to avoid the need for the intermediate struct type?

lukehoban (approved):

lukehoban:
Is it possible to add tests?

stack72 (approved):

PR #7137, opened 2021-05-25
Support using pre-signed URLs for Azure Storage issued by the Pulumi Service

praneetloke:
# Description

As you know Pulumi Service supports [self-hosted](https://www.pulumi.com/docs/guides/self-hosted/) options. This means a customer can self-host Pulumi Service on Azure. For policy pack publishing to work with Azure Storage, the signed URL retrieved from the service must have a header called `x-ms-blob-type` that is set to `BlockBlob`. This is an exception with Azure Storage only. Since the header is a custom header only used by Azure Storage service, it is safe to include it for all of them.

Fixes `N/A`

## Checklist

~~- [ ] I have added tests that prove my fix is effective or that my feature works~~ I have not added a test to this repo but we do have a test in the service that exercises this for signed URLs.

### DIFF
--- a/pkg/backend/httpstate/client/client.go
+++ b/pkg/backend/httpstate/client/client.go
@@ -616,10 +616,8 @@ func (pc *Client) PublishPolicyPack(ctx context.Context, orgName string,
 		return "", errors.Wrapf(err, "Failed to upload compressed PolicyPack")
 	}
 
-	if resp.RequiredHeaders != nil {
-		for k := range resp.RequiredHeaders {
-			putReq.Header.Add(k, resp.RequiredHeaders[k])
-		}
+	for k := range resp.RequiredHeaders {
+		putReq.Header.Add(k, resp.RequiredHeaders[k])
 	}
 
 	_, err = http.DefaultClient.Do(putReq)
--- a/pkg/backend/httpstate/client/client.go
+++ b/pkg/backend/httpstate/client/client.go
@@ -607,16 +607,21 @@ func (pc *Client) PublishPolicyPack(ctx context.Context, orgName string,
 	}
 
 	//
-	// Step 2: Upload the compressed PolicyPack directory to the presigned S3 URL. The PolicyPack is
-	// now published.
+	// Step 2: Upload the compressed PolicyPack directory to the pre-signed object storage service URL.
+	// The PolicyPack is now published.
 	//
 
-	putS3Req, err := http.NewRequest(http.MethodPut, resp.UploadURI, dirArchive)
+	putReq, err := http.NewRequest(http.MethodPut, resp.UploadURI, dirArchive)
 	if err != nil {
 		return "", errors.Wrapf(err, "Failed to upload compressed PolicyPack")
 	}
 
-	_, err = http.DefaultClient.Do(putS3Req)
+	// Add a custom header for Azure Storage so that if the pre-signed URL is for their
+	// storage service, the request doesn't fail due to it being required.
+	// See x-ms-blob-type on https://docs.microsoft.com/en-us/rest/api/storageservices/put-blob
+	putReq.Header.Add("x-ms-blob-type", "BlockBlob")
+
+	_, err = http.DefaultClient.Do(putReq)
 	if err != nil {
 		return "", errors.Wrapf(err, "Failed to upload compressed PolicyPack")
 	}
--- a/pkg/backend/httpstate/client/client.go
+++ b/pkg/backend/httpstate/client/client.go
@@ -616,8 +616,8 @@ func (pc *Client) PublishPolicyPack(ctx context.Context, orgName string,
 		return "", errors.Wrapf(err, "Failed to upload compressed PolicyPack")
 	}
 
-	for k := range resp.RequiredHeaders {
-		putReq.Header.Add(k, resp.RequiredHeaders[k])
+	for k, v := range resp.RequiredHeaders {
+		putReq.Header.Add(k, v)
 	}
 
 	_, err = http.DefaultClient.Do(putReq)
--- a/pkg/backend/httpstate/client/client.go
+++ b/pkg/backend/httpstate/client/client.go
@@ -616,10 +616,11 @@ func (pc *Client) PublishPolicyPack(ctx context.Context, orgName string,
 		return "", errors.Wrapf(err, "Failed to upload compressed PolicyPack")
 	}
 
-	// Add a custom header for Azure Storage so that if the pre-signed URL is for their
-	// storage service, the request doesn't fail due to it being required.
-	// See x-ms-blob-type on https://docs.microsoft.com/en-us/rest/api/storageservices/put-blob
-	putReq.Header.Add("x-ms-blob-type", "BlockBlob")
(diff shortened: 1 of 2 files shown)


pkg/backend/httpstate/client/client.go
```diff
@@ -607,16 +607,21 @@ func (pc *Client) PublishPolicyPack(ctx context.Context, orgName string,
...
+	putReq, err := http.NewRequest(http.MethodPut, resp.UploadURI, dirArchive)
 	if err != nil {
 		return "", errors.Wrapf(err, "Failed to upload compressed PolicyPack")
 	}
 
-	_, err = http.DefaultClient.Do(putS3Req)
+	// Add a custom header for Azure Storage so that if the pre-signed URL is for their
+	// storage service, the request doesn't fail due to it being required.
+	// See x-ms-blob-type on https://docs.microsoft.com/en-us/rest/api/storageservices/put-blob
+	putReq.Header.Add("x-ms-blob-type", "BlockBlob")
```
praneetloke, line 622: I didn't add any conditional checks to include the header because storage accounts might have a custom domain and therefore the detection may not be 100% accurate. If you have other suggestions, please let me know.

praneetloke:
Also do you think a changelog entry is warranted for this change?

pkg/backend/httpstate/client/client.go
lukehoban: Maybes it's too much - but would it make sense for the service to return back required headers to include as part of the CreatePolicyPackResponse above?  Then this wouldn't need to be hard coded - and could be provided by the service which knows what kind of storage backend is being used?

Also - does this break S3 requests?  I thought the headers with part of the presigned signature and couldn't be modified like this?

praneetloke: That's a good idea to include required headers in the response. I am wondering if we could simply return the backend object storage scheme? That way, the only header we would add is the custom header for Azure Storage. It would be sort of explicit.

About S3 requests and pre-signed URL restrictions: I believe if we were to set one of the [supported headers](https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/s3-example-presigned-urls.html) ([gocloud ref](https://github.com/google/go-cloud/blob/master/blob/blob.go#L1059)), then, only the value that was given at the time of URL generation will be allowed. I will double-check that, though.

lukehoban: >  I am wondering if we could simply return the backend object storage scheme? That way, the only header we would add is the custom header for Azure Storage. It would be sort of explicit.

Curious what you have in mind here?  You mean the client would key off of the scheme to decide what headers to include?  Still feels a little oddly coupled, but definitely workable if that's sufficiently easier than leaving all the logic about what URLs and headers to POST to in the service.

praneetloke: > You mean the client would key off of the scheme to decide what headers to include?

Yes, this is what I am thinking.

> Still feels a little oddly coupled, but definitely workable if that's sufficiently easier than leaving all the logic about what URLs and headers to POST to in the service.

Yep agreed. Maybe this is what you were originally suggesting too, but we could return the headers with both the name **and** the value that need to be set by the client. Otherwise, the CLI will need to know what value to set anyway for a header and so we don't gain much. But it seems like a bad idea to include something in the response that the CLI implicitly trusts and round-trips to the signed URL. Rather I am fine with being explicit on the client-side about what header it will include based on the scheme. What do you think? The alternative is to leave this as-is for now, i.e. include this header for all signed URLs since it is a custom header.

praneetloke: @leezen and I talked about the options here. We are going to go the route of having the service return the headers that the CLI should add in order to use the pre-signed URL. It will be in favor of us not wanting to couple the CLI with the storage mechanism. I'll update this PR once I have the service changes ready.

sdk/go/common/apitype/policy.go
```diff
@@ -46,6 +46,9 @@ type CreatePolicyPackRequest struct {
 type CreatePolicyPackResponse struct {
 	Version   int    `json:"version"`
 	UploadURI string `json:"uploadURI"`
+	// RequiredHeaders represents headers that the CLI must set in order
+	// for the upload to succeed.
+	RequiredHeaders map[string]string `json:"headers,omitempty"`
```
lukehoban, line 51: We typical try to keep the json and go property names aligned modulo Pascal vs camel casing. Should this just be `Headers`?

pkg/backend/httpstate/client/client.go
```diff
@@ -607,16 +607,20 @@ func (pc *Client) PublishPolicyPack(ctx context.Context, orgName string,
...
 	//
 
-	putS3Req, err := http.NewRequest(http.MethodPut, resp.UploadURI, dirArchive)
+	putReq, err := http.NewRequest(http.MethodPut, resp.UploadURI, dirArchive)
 	if err != nil {
 		return "", errors.Wrapf(err, "Failed to upload compressed PolicyPack")
 	}
 
-	_, err = http.DefaultClient.Do(putS3Req)
+	for k := range resp.RequiredHeaders {
```
lukehoban, line 619: Is it true that this works for `nil` headers? (As would be returned by a service that doesn't yet support or need to populate headers?).

pkg/backend/httpstate/client/client.go
```diff
@@ -607,16 +607,20 @@ func (pc *Client) PublishPolicyPack(ctx context.Context, orgName string,
...
+		putReq.Header.Add(k, resp.RequiredHeaders[k])
```
lukehoban, line 620: Total nit - this could be `k, v := range` and then `.Add(k, v)` here.

lukehoban (approved):

sdk/go/common/apitype/policy.go
praneetloke: oh shoot. I totally overlooked naming the json property name after my IDE auto-filled the `omitempty`. But yes, I'll camel-case this.

pkg/backend/httpstate/client/client.go
praneetloke: That's right https://play.golang.org/p/2U7KzvDu0lj

sdk/go/common/apitype/policy.go
praneetloke, line 51:
```suggestion
	RequiredHeaders map[string]string `json:"requiredHeaders,omitempty"`
```

pkg/backend/httpstate/client/client.go
praneetloke, line 619:
```suggestion
	for k, v := range resp.RequiredHeaders {
```

pkg/backend/httpstate/client/client.go
praneetloke, line 620:
```suggestion
		putReq.Header.Add(k, v)
```

praneetloke:
Added the `impact-no-changelog` label since the PR doesn't change the behavior of a feature for a user.

praneetloke:
The failed macos build is unrelated to the change in this PR. It passed previously on a restart. I'll skip restarting it again.

Issue #7072, opened 2021-05-17
Support methods for multi-language components


justinvp:
Components often have methods that provide additional functionality associated with an instance of a component. We will add a new capability in the schema and Provider interface to support resource methods.

## Design

Resource methods will be implemented by adding a new field, `methods`, to `schema.ResourceSpec`. This field will map from method names to functions. Functions will be identified using a JSON path, which must refer to a function in the containing schema. If a method map refers to a function that does not exist or refers to an entity that is not a function, `schema.ImportSpec` will return an error. By convention, method names should be in `camelCase`, and functions that implement methods should use a token of the form `pkgName:path/to/module:ResourceName/methodName`. A function that implements a resource method for a resource type T must have an input field with the name `__self__` of the resource type T. The implementation of a resource method should depend only on properties of a resource type that are present in the schema, as the value of the `__self__` parameter will be a resource reference that can only be resolved to its serializable state.

### Code Generation

A resource method definition in a Pulumi schema should generate an appropriately-scoped method or function in each SDK. For all of our currently-supported languages, this should be a member function on the appropriate resource type.

### `Call` Provider gRPC

While methods could be implemented with the existing `Invoke` method with some additional arguments, instead we will add a new `Call` gRPC method specifically for methods. `Call` will take similar arguments as `Invoke` and additionally accept the monitor address and component resource options, to be able to create child resources of a component inside Call.

### SDKs

We will provide helpers in each SDK to make it easier to implement `Call` for methods, similar to what we’ve done for `Construct`, adapting the raw gRPC values into the Pulumi SDK model in each language.

## Primary Work Items for 7/8

- [x] #7141
- [x] #7363
- [x] #7379
- [x] #7364
- [x] #7377
- [x] #7378
- [x] #7437
- [x] #7438

## M59

- [x] #7555
- [x] #7488
- [ ] #7582
- [ ] #7606

lukehoban:
> Resource methods will be implemented by adding a new field, methods, to schema.ResourceSpec. This field will map from method names to functions.

We likely want to support module- level functions as well as methods. We support module-level functions as invokes today, but will need to allow them to opt-in to being `Call`s instead I assume?

t0yv0:
> additional functionality associated with an instance of a component

What is meant by an instance of a component? Does the method implementation operate on read-only fields from a serialized representation of the object implementing the component, or does it operate on an actual object that has continuous identity in the provider process address space and can therefore mutate fields?

[...]

PR #7141, opened 2021-05-26
Add schema support for methods

justinvp:
The implementation of this is based on what was outlined in the original design doc.

One question I had while implementing this: Why not add the method definitions directly in the resource's `"methods"` section instead of referencing functions defined in the top-level `"functions"` section?

Part of #7072

### DIFF
--- a/pkg/codegen/schema/schema.go
+++ b/pkg/codegen/schema/schema.go
@@ -321,6 +321,15 @@ type Resource struct {
 	Language map[string]interface{}
 	// IsComponent indicates whether the resource is a ComponentResource.
 	IsComponent bool
+	// Methods is the list of methods for the resource.
+	Methods []*Method
+}
+
+type Method struct {
+	// Name is the name of the method.
+	Name string
+	// Function is the function definition for the method.
+	Function *Function
 }
 
 // Function describes a Pulumi function.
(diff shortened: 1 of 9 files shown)


pkg/codegen/schema/schema.go
```diff
@@ -339,6 +348,8 @@ type Function struct {
 	DeprecationMessage string
 	// Language specifies additional language-specific data about the function.
 	Language map[string]interface{}
+	// IsMethod indicates whether the function is a method of a resource.
+	IsMethod bool
```
justinvp, line 352: This will be useful for the forthcoming codegen changes, to avoid emitting a top-level function if the function is being used as a method.

pgavlin:
> Why not add the method definitions directly in the resource's "methods" section instead of referencing functions defined in the top-level "functions" section?

I think I got too cute with the original design. Adding the method definitions in the resource seems fine. @lukehoban any thoughts?

pgavlin (approved):
LGTM as written, but I would also be fine with moving things around as you describe.

justinvp:
> I think I got too cute with the original design. Adding the method definitions in the resource seems fine. @lukehoban any thoughts?

Although, one reason in favor of the original design is that it does keep all of the function tokens together in one place (which effectively matches the `switch` that'd be inside the provider implementation).

justinvp:
I'm going to leave this as-is (original design). We can revisit if we find the schema authoring of these to be too unnatural, ahead of making the functionality broadly available.

PR #7142, opened 2021-05-26
Populate package.json version to match provider version

liamawhite:
As the version is hardcoded, whenever I rebuild I get:

```
Diagnostics:
  pulumi:providers:tsb (tsb):
    error: pulumi:providers:tsb resource 'tsb's property 'version' value {${VERSION}} has a problem: could not parse provider version: Invalid character(s) found in major number "${VERSION}"
```

Is there any reason we can't just populate this with the provider version?

### DIFF
--- a/pkg/codegen/nodejs/gen.go
+++ b/pkg/codegen/nodejs/gen.go
@@ -1603,7 +1603,7 @@ func genNPMPackageMetadata(pkg *schema.Package, info NodePackageInfo) string {
 	// Create info that will get serialized into an NPM package.json.
 	npminfo := npmPackage{
 		Name:        packageName,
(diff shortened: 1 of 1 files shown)


pgavlin (commented):
I would love it if we could make this change, but unfortunately, this approach does not work with the way that we determine the version of a Node package prior to publishing. We derive the version of each package from the Git tag associated with the commit being published, and we use `${VERSION}` as a reliable pattern in `package.json` that allows us to stamp in the released version.

All that said, I do think that we could probably fix things here with a little bit of work on our end. @stack72 what do you think about using `pulumictl` to deserialize `package.json` and stamp in the correct version prior to publishing?

PR #7144, opened 2021-05-26
put nodejs resource registration logs behind excessiveDebugOutput flag

EvanBoyle:
I was looking at some program traces and noticed hundreds of calls to `/pulumirpc.Engine/Log`. When I remove debug logs for duplicate resource registrations, this basic program cuts it's runtime from 10s down to 3s.

```ts
import * as kubernetes from "@pulumi/kubernetes";

console.log(aws);
console.log(kubernetes);

export const foo = {};
```

Looking at the traces, there are ~800 logging calls for duplicate registrations, each of which is an RPC request to the engine. This change puts those specific logging calls behind the excessive debug flag. Perhaps in theory this shouldn't be a problem, and we should be able to handle these logging requests but it does appear to be creating perf problems for the most basic of programs. 

Fixes # (issue)

## Checklist

No tests or changelog here. No behavior change, just hiding some logs behind the debug flag.

### DIFF
--- a/sdk/nodejs/runtime/rpc.ts
+++ b/sdk/nodejs/runtime/rpc.ts
@@ -631,7 +631,9 @@ export function register<T extends { readonly version?: string }>(source: Map<st
                 // In this case, we might legitimately get mutliple registrations of the same resource.  It should not
                 // matter which we use, so we can just skip re-registering.  De-serialized resources will always be
                 // instances of classes from the first registered package.
-                log.debug(`skip re-registering already registered ${registrationType} ${key}@${item.version}.`);
+                if (excessiveDebugOutput) {
+                    log.debug(`skip re-registering already registered ${registrationType} ${key}@${item.version}.`);
+                }
                 return false;
             }
         }
@@ -640,7 +642,9 @@ export function register<T extends { readonly version?: string }>(source: Map<st
         source.set(key, items);
     }
 
-    log.debug(`registering ${registrationType} ${key}@${item.version}`);
+    if (excessiveDebugOutput) {
+        log.debug(`skip re-registering already registered ${registrationType} ${key}@${item.version}.`);
+    }
     items.push(item);
     return true;
 }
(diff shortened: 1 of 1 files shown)


sdk/nodejs/runtime/rpc.ts
```diff
@@ -640,7 +642,9 @@ export function register<T extends { readonly version?: string }>(source: Map<st
         source.set(key, items);
     }
 
-    log.debug(`registering ${registrationType} ${key}@${item.version}`);
+    if (excessiveDebugOutput) {
+        log.debug(`skip re-registering already registered ${registrationType} ${key}@${item.version}.`);
```
lukehoban, line 646: Looks like a copy/paste error?

```suggestion
        log.debug(`registering ${registrationType} ${key}@${item.version}`);
```

lukehoban (approved):

t0yv0:
Hm, nice find, so these cost `(/   7000  800.0) = 8.75ms` per call? This is interesting, taking a note.

Is there a reason `excessiveDebugOutput` is not part of `log.debug` that is, should we silence just these calls or all calls.

t0yv0 (approved):

EvanBoyle:
> Is there a reason excessiveDebugOutput is not part of log.debug that is, should we silence just these calls or all calls.

`log.debug` is a public api so probably not prudent to make behavior changes. Probably we should create an internal logging interface that does exactly what you describe. We more or less do this for pulumi/pulumi and it doesn't seem to negatively impact supportability too much.

komalali (approved):

t0yv0:
We can also experiment with different transports for logs than gRPC if this has such overhead.

komalali:
>Hm, nice find, so these cost (/ 7000 800.0) = 8.75ms per call? This is interesting, taking a note.

@t0yv0 according to the trace itself it appears the log calls take 1ms `/pulumirpc.Engine/Log (1ms)`

pgavlin (approved):

EvanBoyle:
Flaky archive test causing test failure in the mac build. Filed https://github.com/pulumi/pulumi/issues/7147

t0yv0:
TBC in 
https://github.com/pulumi/pulumi/issues/7146

PR #7153, opened 2021-05-27
Filestate backend gzip compression

### LINKED ISSUE #5469: Support optional compression of remote state JSON

awoimbee:
# Description
Replacement for #7086.
Fixes #5469

Motivation:
Stack file size explosion with lots of k8s resources (helm charts are the worst offender): we've got stacks above 30MB ! (from 16MB in the linked issue). Pulumi becomes a real chore to use when it spends so much time transferring files.

Solution:
Enable compressed (gzip) files reading and add an environment var to write compressed files.
The default behavior is still to write plain json, so backwards compatibility is assured. The default behavior could then change in a later (breaking) update.

## Checklist

### DIFF
--- a/pkg/backend/filestate/state.go
+++ b/pkg/backend/filestate/state.go
@@ -163,7 +163,7 @@ func (b *localBackend) getCheckpoint(stackName tokens.QName) (*apitype.Checkpoin
 	if err != nil {
 		return nil, err
 	}
-	bytes, err = inflateBytes(bytes)
+	bytes, err = maybeInflateBytes(bytes)
 	if err != nil {
 		return nil, err
 	}
@@ -306,7 +306,7 @@ func (b *localBackend) backupStack(name tokens.QName) error {
 	if err != nil {
 		return err
 	}
-	byts, err = inflateBytes(byts)
+	byts, err = maybeInflateBytes(byts)
 	if err != nil {
 		return err
 	}
@@ -419,7 +419,7 @@ func (b *localBackend) getHistory(name tokens.QName, pageSize int, page int) ([]
 		if err != nil {
 			return nil, errors.Wrapf(err, "reading history file %s", filepath)
 		}
-		b, err = inflateBytes(b)
+		b, err = maybeInflateBytes(b)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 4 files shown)


pkg/backend/filestate/state.go
```diff
@@ -260,7 +283,9 @@ func (b *localBackend) removeStack(name tokens.QName) error {
 func backupTarget(bucket Bucket, file string) string {
 	contract.Require(file != "", "file")
 	bck := file + ".bak"
-	err := renameObject(bucket, file, bck)
+	err := bucket.Delete(context.TODO(), bck)
```
awoimbee, line 286: This is the part I least like about this PR, but I think it should not cause problems.
This `Delete` is here so we don't have stale `.json.bak` next to newer `.json.gz.bak` and inversely.

awoimbee:
Hi @stack72 @lukehoban 
Any news on this PR ? It's been nearly a month, I'm ready to do any modification necessary.

I've been using a test build of pulumi based on this PR for a while. I have had no issues. Compression makes the FileState backend a lot faster, even with smaller stacks.
Also, I believe that this upgrade to the FileSate backend doesn't devalue the paid (SaaS) plan.

pkg/backend/filestate/state.go
Frassle: But the renameObject below is going to write over bck anyway? This delete looks superfluous.

jsravn:
What's the status of this PR? We really need this with a self hosted backend.

awoimbee:
>What's the status of this PR?

Honestly it needs some work before getting merged. But anyways it's not getting much traction from the community nor from the maintainers.
If like me you can't live without compressed stacks, check out the builds @ https://github.com/extrality/pulumi. But beware, once enlightened you will not be able to bear the official pulumi builds.

pyoio:
Is there an ETA on this (or #9610) being reviewed -> released? We're stuck using GCS backend for the foreseeable future and local executions of pulumi are uploading gigabytes per invocation (the state JSON is about 14mb) when updating only a couple of parameters 😭

PR #7158, opened 2021-05-27
[codegen/go] Fix emitted type of resources and types

justinvp:
In Go, resource types are modeled as pointers, but there were cases where the type was not being emitted as a pointer, leading to marshaling errors in programs. Additionally, array and map values that are external references were being emitted as pointers, but only _resources_ should be pointers (not external types), regardless of whether the resource type is external or local.

This fixes the underlying Go marshaling issues with https://github.com/pulumi/pulumi-eks/issues/566 and @lblackstone's repro in https://github.com/pulumi/pulumi-eks/issues/577

Part of https://github.com/pulumi/pulumi-eks/issues/566
Part of https://github.com/pulumi/pulumi-eks/issues/577

### DIFF
--- a/pkg/codegen/go/gen.go
+++ b/pkg/codegen/go/gen.go
@@ -221,22 +221,18 @@ func (pkg *pkgContext) plainType(t schema.Type, optional bool) string {
 		return pkg.plainType(t.ElementType, optional)
 	case *schema.ArrayType:
 		typ = "[]"
-		if pkg.isExternalReference(t.ElementType) {
-			typ += "*"
-		}
 		typ += pkg.plainType(t.ElementType, false)
 		return typ
 	case *schema.MapType:
 		typ = "map[string]"
-		if pkg.isExternalReference(t.ElementType) {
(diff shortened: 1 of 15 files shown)


pkg/codegen/go/gen.go
```diff
@@ -914,8 +910,9 @@ func (pkg *pkgContext) genOutputTypes(w io.Writer, t *schema.ObjectType, details
 			outputType, applyType := pkg.outputType(p.Type, true), pkg.plainType(p.Type, true)
 			deref := ""
 			// If the property was required, but the type it needs to return is an explicit pointer type, then we need
-			// to derference it.
-			if p.IsRequired && applyType[0] == '*' {
+			// to derference it, unless it is a resource type which should remain a pointer.
```
viveklak, line 913:
```suggestion
			// to dereference it, unless it is a resource type which should remain a pointer.
```

lblackstone (approved):

viveklak (approved):

PR #7162, opened 2021-05-27
[automation/nodejs] - Only log, don't error on unparsed events

komalali:
# Description

I had thought this issue was resolved, but alas it is not.

In the absence of a repro, I've decided to remove the error on this line and change it to a warning. In most cases all we need is the summary event from the preview (when a user hasn’t provided an onEvent callback), and it seems more reasonable to only error if we are unable to get the summary rather than if we are unable to read any random event. Hopefully this warning will allow us to gather more data on the issue if it continues to pop up, but at the very least folks who are not using `onEvent` will be able to let their programs complete without error.

Related to: https://github.com/pulumi/pulumi/issues/6768

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,7 +1,5 @@
 ### Improvements
 
-
-  
 - [codegen] - Encrypt input args for secret properties.
   [#7128](https://github.com/pulumi/pulumi/pull/7128)
 
@@ -10,3 +8,6 @@
 - [cli] - Send plugin install output to stderr, so that it doesn't
   clutter up --json, automation API scenarios, and so on.
   [#7115](https://github.com/pulumi/pulumi/pull/7115)
+
+- [auto/nodejs] - Emit warning instead of breaking on parsing JSON events for automation API.
+  [#7162](https://github.com/pulumi/pulumi/pull/7162)
--- a/sdk/nodejs/automation/stack.ts
+++ b/sdk/nodejs/automation/stack.ts
@@ -354,13 +354,13 @@ Event: ${line}\n${e.toString()}`);
         }
 
         if (!summaryEvent) {
-            throw new Error("No summary of changes.");
+            log.warn("Failed to parse summary event, but preview succeeded. PreviewResult `changeSummary` will be empty.");
         }
 
         return {
             stdout: preResult.stdout,
             stderr: preResult.stderr,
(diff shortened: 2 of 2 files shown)


EvanBoyle (commented):
On the surface, the change looks good. Have we considered any of the following:

1. Is it safe to continue parsing events in this state? 
2. Is there any downstream functionality that might fail as a result of missing events (one that comes to mind is permaurl parsing, another is preview resource summary)? Does any of this need to be updated to be tolerant of continued processing in the event of failure?

komalali:
>Is it safe to continue parsing events in this state?

Yes, just because one event didn't parse doesn't mean others will fail. Plus, we *need* to continue parsing for the summaryEvent.

>Is there any downstream functionality that might fail as a result of missing events (one that comes to mind is permaurl parsing, another is preview resource summary)? Does any of this need to be updated to be tolerant of continued processing in the event of failure?

I wasn't aware that we do any permaurl parsing or that it had anything to do with engine events. Where does this happen?
We already have a check in place if the summary event doesn't exist. I think it's fair to error if that specific event fails to parse, because it is required for the `PreviewResult`. I'm still not sure why this error happens but since the summary event is so short I find it is probably more unlikely that it would get oddly truncated. I could be wrong here but I'd like to see the failures to re-evaluate.

EvanBoyle (approved):
Ok. I guess I should've asked "if we get a bad event, will the rest of the events also be bad?". It sounds like we believe the answer to that is no.

I do still think there is some mild weirdness here where we could potentially have an update that succeeds, but the user gets a failure due to the summary event being missing if the event was corrupt. We don't have any mechanism of distinguishing update failures vs automation api bugs/machinery issues. 

It sounds like you have thought through all of this though, so I will give it the 🚢

komalali:
>Ok. I guess I should've asked "if we get a bad event, will the rest of the events also be bad?". It sounds like we believe the answer to that is no.
Yeah, when I had opened up this logging for debugging purposes the user reported that they got 2 of these errors (out of some 800 event messages) so yes, not all succeeding events will fail to parse.

>I do still think there is some mild weirdness here where we could potentially have an update that succeeds, but the user gets a failure due to the summary event being missing if the event was corrupt.
Yeah, fair enough. I changed the error on missing summaryevent to log a warning instead and made the `changeSummary` property optional on `PreviewResult`.

Issue #7149, opened 2021-05-27
Loading nodejs mutli-lang provider has ~10s of overhead


EvanBoyle:
I created two multi-lang providers that create identical resources. One implemented in Go, and one implemented in Node. I added some instrumentation into the `Registry` and provider loader components of the CLI and observed about 10s of overhead when loading the node plugin as compared to the go plugin (go is < 1s). 

To measure this, I logged timestamps before and after each provider load within the registry:

https://github.com/pulumi/pulumi/blob/73a66f48ead301bacf1d7d2833c016cf803b57c1/pkg/resource/deploy/providers/registry.go#L80

(both user programs in examples below are using generated typescript sdks)

Results for Go:

```
I0526 19:09:25.566732   70321 registry.go:81] loading provider, gobench, 1622081365
I0526 19:09:25.659098   70321 registry.go:84] done loading provider, gobench, 1622081365
I0526 19:09:25.964986   70321 registry.go:81] loading provider, aws, 1622081365
I0526 19:09:33.797189   70321 registry.go:84] done loading provider, aws, 1622081373
 
Outputs:
    bucket: "page-a40b1c4"
    url   : "page-a40b1c4.s3-website-us-west-2.amazonaws.com"

Resources:
    + 5 created

Duration: 14s

```

results for typescript:

```

I0526 19:10:16.489986   70340 registry.go:81] loading provider, tsbench, 1622081416
I0526 19:10:27.226615   70340 registry.go:84] done loading provider, tsbench, 1622081427
I0526 19:10:27.674548   70340 registry.go:81] loading provider, aws, 1622081427
I0526 19:10:37.692740   70340 registry.go:84] done loading provider, aws, 1622081437
 
Diagnostics:
  pulumi:pulumi:Stack (simple-tsbench):
    user program start
    1622081415.741
    user program imports
    1622081416.25
    user program done
    1622081416.278
 
    pulumi server.ts main
    1622081426.971
    pulumi server.ts port allocated, grpc up
    1622081427.215
    pulumi server.ts getplugininfo
    1622081427.222
    pulumi server.ts checkconfig
    1622081427.228
    pulumi server.ts configure
    1622081427.299
    pulumi server.ts construct
    1622081427.431
    pulumi server.ts constructimpl
    1622081427.432
    pulumi provider construct
    1622081427.434
    provider component constructor
    1622081427.434
    provider component constructor done
    1622081427.501
 
Outputs:
    bucket: "page-b986dff"
[...]

PR #7163, opened 2021-05-28
make plugin metadata (size, install time) opt in for list operations

EvanBoyle:
# Description

This change is a simple perf optimization to speed up the process of listing plugins by excluding some metadata like size by default. 

Our strategy for finding a plugin is to first look on the path, and then to iterate through all plugins in the plugin cache (a directory). We do this for each plugin that is loaded when `NewProvider` is called. Unfortunately, the codepath that gets all plugins is shared by `pulumi plugin ls` that needs to do things like display the total size of all plugins, the size of each plugin, and when the plugin was last installed/last used.

This means that any time a plugin is loaded, we are computing the size of all plugins by recursively enumerating all folder (including all of the node_modules directories of any installed node multi-lang plugins!). For my 5 gb of node plugins this translated to 10s of overhead each time a plugin was loaded. 

This change is a very simple fix. `pulumi plugin ls` is the only code path that uses size, so we create a dedicated code path `GetPluginsWithMetadata` that populates that info, excluding from the result of `GetPlugins` by default. 

This is the least invasive option I could come up with. This struct is public API (exposed via automation api), so removing these fields from the default PluginInfo struct would not be easy. 

Fixes https://github.com/pulumi/pulumi/issues/7149
Fixes https://github.com/pulumi/astro/issues/16 (for now)

## Checklist

No changelog required
Existing plugin tests (those that directly and indirectly exercise this path) are passing, however we have no e2e tests on `plugin ls`. Not going to take on adding those as a part of this change.

### DIFF
--- a/sdk/go/common/workspace/plugins.go
+++ b/sdk/go/common/workspace/plugins.go
@@ -533,17 +533,31 @@ func GetPluginDir() (string, error) {
 	return GetPulumiPath(PluginDir)
 }
 
-// GetPlugins returns a list of installed plugins.
+// GetPlugins returns a list of installed plugins without size info and last accessed metadata.
+// Plugin size requires recursively traversing the plugin directory, which can be extremely expensive with the introduction of
+// nodejs multilang components that have deeply nested node_modules folders.
 func GetPlugins() ([]PluginInfo, error) {
 	// To get the list of plugins, simply scan the directory in the usual place.
 	dir, err := GetPluginDir()
 	if err != nil {
 		return nil, err
 	}
-	return getPlugins(dir)
+	return getPlugins(dir /* skipMetadata */, true)
 }
 
-func getPlugins(dir string) ([]PluginInfo, error) {
+// GetPluginsWithMetadata returns a list of installed plugins with metadata about size, and last access (POOR RUNTIME PERF)
+// Plugin size requires recursively traversing the plugin directory, which can be extremely expensive with the introduction of
+// nodejs multilang components that have deeply nested node_modules folders.
+func GetPluginsWithMetadata() ([]PluginInfo, error) {
+	// To get the list of plugins, simply scan the directory in the usual place.
+	dir, err := GetPluginDir()
+	if err != nil {
+		return nil, err
+	}
+	return getPlugins(dir /* skipMetadata */, false)
+}
+
+func getPlugins(dir string, skipMetadata bool) ([]PluginInfo, error) {
 	files, err := ioutil.ReadDir(dir)
 	if err != nil {
 		if os.IsNotExist(err) {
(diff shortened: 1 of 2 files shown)


sdk/go/common/workspace/plugins.go
```diff
@@ -533,17 +533,33 @@ func GetPluginDir() (string, error) {
...
+// expensive with the introduction of nodejs multilang components that have
+// deeply nested node_modules folders.
 func GetPlugins() ([]PluginInfo, error) {
 	// To get the list of plugins, simply scan the directory in the usual place.
 	dir, err := GetPluginDir()
 	if err != nil {
 		return nil, err
 	}
-	return getPlugins(dir)
+	return getPlugins(dir /* skipMetadata */, true)
```
justinvp, line 546:
```suggestion
	return getPlugins(dir, true /* skipMetadata */)
```

sdk/go/common/workspace/plugins.go
```diff
@@ -533,17 +533,33 @@ func GetPluginDir() (string, error) {
...
+// and last access (POOR RUNTIME PERF). Plugin size requires recursively traversing the
+// plugin directory, which can be extremely expensive with the introduction of
+// nodejs multilang components that have deeply nested node_modules folders.
+func GetPluginsWithMetadata() ([]PluginInfo, error) {
+	// To get the list of plugins, simply scan the directory in the usual place.
+	dir, err := GetPluginDir()
+	if err != nil {
+		return nil, err
+	}
+	return getPlugins(dir /* skipMetadata */, false)
```
justinvp, line 559:
```suggestion
	return getPlugins(dir, false /* skipMetadata */)
```

justinvp (approved):

sdk/go/common/workspace/plugins.go
```diff
@@ -533,17 +533,33 @@ func GetPluginDir() (string, error) {
 	return GetPulumiPath(PluginDir)
 }
 
-// GetPlugins returns a list of installed plugins.
+// GetPlugins returns a list of installed plugins without size info and last accessed metadata.
+// Plugin size requires recursively traversing the plugin directory, which can be extremely
+// expensive with the introduction of nodejs multilang components that have
+// deeply nested node_modules folders.
```
lukehoban, line 539: I assume there is only one caller in the world that cares about this additional metadata (plugin ls). Can't we just modify that caller to call SetFileMetadata and just always skip filling in metadata here? (Instead of forking this API).

EvanBoyle: While that is possible, what you're suggesting is to basically duplicate `workspace.getPlugins` in `cmd/plugin`. The `PluginInfo` returned does not contain the path of the plugin, which is required for `SetFileMetadata`. Doesn't really feel better than having this "fork". 

I'll leave as is for now. Like you said plugin_ls is the only consumer of this surface area, all other callers are doing the right thing, and we maintain the option to refactor later if it makes sense.

Issue #3979, opened 2020-02-26
Provider aliases aren't working


clstokes:
If I use an explicit provider (e.g. `aws.Provder()`) and then change its name it leads to all resources that use it being replaced. I tried to use `aliases` to prevent this, but aliasing the provider seems to have no effect and all resources still get replaced.

I've reproduced this across TypeScript and Go with Pulumi.

#### Version Info

```
pulumi-ts-provider-alias % pulumi version
v1.11.0
pulumi-ts-provider-alias % npm ls --depth 0
pulumi-ts-provider-alias@ /Users/clstokes/cc/scratch/pulumi-ts-provider-alias
├── @pulumi/aws@1.23.0
├── @pulumi/awsx@0.18.14
├── @pulumi/pulumi@1.11.0
└── @types/node@8.10.59
```

#### Steps to Reproduce

1. `up` the code below
1. Change the `providerName` value to `"name2"`
1. `pulumi preview --diff` and see the bucket will be replaced because the provider changes.

#### Code to Reproduce

```ts
import * as aws from "@pulumi/aws";

const providerName = "name2";
const awsProvider = new aws.Provider(providerName, {
    region: "us-west-2",
}, { aliases: [{ name: "name1" }]});
// }, { aliases: ["urn:pulumi:dev::pulumi-ts-provider-alias::pulumi:providers:aws::name1"] });

const bucket = new aws.s3.Bucket("my-bucket", {}, { provider: awsProvider });

export const bucketName = bucket.id;
```

#### `preview --diff` output after provide name change

```
pulumi-ts-provider-alias % pulumi preview --diff
Previewing update (dev):
  pulumi:pulumi:Stack: (same)
    [urn=urn:pulumi:dev::pulumi-ts-provider-alias::pulumi:pulumi:Stack::pulumi-ts-provider-alias-dev]
    +-aws:s3/bucket:Bucket: (replace)
        [id=my-bucket-a797c6e]
        [urn=urn:pulumi:dev::pulumi-ts-provider-alias::aws:s3/bucket:Bucket::my-bucket]
        [provider: urn:pulumi:dev::pulumi-ts-provider-alias::pulumi:providers:aws::name1::577bc71c-4473-4681-a2ef-6ba5f38fbed1 => urn:pulumi:dev::pulumi-ts-provider-alias::pulumi:providers:aws::name2::577bc71c-4473-4681-a2ef-6ba5f38fbed1]
      - accelerationStatus      : ""
[...]

PR #7166, opened 2021-05-28
Update Registry to respect provider aliases

EvanBoyle:
# Description

This change makes the provider registry aware of aliases. This allows provider aliases to work properly, ensuring that resources that use aliased providers don't get replaced unnecessarily.

Fixes https://github.com/pulumi/pulumi/issues/3979

## Checklist


pgavlin:
It feels like this will behave subtly differently than other aliases, and I'm not sure that's what we want.

For other resource types, an alias just means "consider this resource to have the same identity as the resource in the old state". For providers, that identity influences a couple things:

1. Which provider instance is used for a particular resource, which is managed by the provider map
2. Whether or not resources that refer to the provider will be considered as requiring replacement due to a change in their provider

I think that your changes address (1)--updating the provider map ensures that resources in the old state will treat the new provider as having the same identity as their provider if it has been aliased--but I'm not sure that it addresses (2). I think that the referenced issue requires that (2) is fixed. Can you add a test for that scenario?

EvanBoyle:
Just tested (2) out and you are correct, altering the provider results in it getting replaced but not the resources that depend on it. 

Do have any suggestions on what layer to look at for addressing (2)?

joeduffy:
Hey @EvanBoyle I was super excited to see you working on a fix for this. Any chance it will be landing soonish?

EvanBoyle:
Update, it looks like the incorrect behavior for (2) is due to https://github.com/pulumi/pulumi-terraform-bridge/issues/244

I'll write some lifecycle tests for (2) to verify.

EvanBoyle:
@pgavlin I've added a passing test that covers those two scenarios, would you mind taking another look?

pkg/resource/deploy/providers/registry.go
```diff
@@ -254,6 +254,18 @@ func (r *Registry) Check(urn resource.URN, olds, news resource.PropertyMap,
 	return inputs, nil, nil
 }
 
+// RegisterAliases takes the computed alias map, and updates the registry such that the new URNSs
+// point to the alised providers
```
pgavlin, line 258:
```suggestion
// RegisterAliases takes the computed alias map and updates the registry such that the new URNs
// point to the aliased providers
```

pkg/resource/deploy/step_generator.go
```diff
@@ -233,6 +233,9 @@ func (sg *stepGenerator) generateSteps(event RegisterResourceEvent) ([]Step, res
 		}
 	}
 
+	// register aliases with the provider registry
+	sg.deployment.providers.RegisterAliases(sg.aliased)
```
pgavlin, line 237: This is overkill: it's re-registering every alias after each resource registration. I think that we can move this into the code above that does the usual alias processing and change it s.t. it only handles a single alias at a time.

pgavlin (changes requested):

pkg/resource/deploy/providers/registry.go
```diff
@@ -254,6 +254,15 @@ func (r *Registry) Check(urn resource.URN, olds, news resource.PropertyMap,
 	return inputs, nil, nil
 }
 
+// RegisterAlias updates the registry such that the new URN points to the aliased provider
+func (r *Registry) RegisterAlias(alias, urn resource.URN) {
+	for ref, prov := range r.providers {
+		if ref.urn == alias {
+			r.setProvider(mustNewReference(urn, ref.id), prov)
+		}
+	}
```
pgavlin, line 263: In principle we shouldn't need to do this. The caller can know whether or not the aliased resource is a provider by checking the type portion of the URN. I'd prefer that we be more precise here and do those checks in the step generator s.t. the contract for this method can be "alias this provider to this URN, and a provider with the alias target must exist". I think that will make the behavior clearer in both cases.

pgavlin (approved):

PR #7167, opened 2021-05-28
[auto/python] - Add stack function overloads

komalali:
# Description

Adding function overloads for `create_stack`, `select_stack` and `create_or_select_stack` to clarify the necessary combination of args for each invocation.

## Checklist

If existing tests continue to succeed that's all the signal we need.


EvanBoyle (commented):
This is not the fix I would've expected, but I don't have a ton of context here. 

The surprising thing about the function in question is that you can pass arguments that get ignored (ie workdir for inline args). Wouldn't it be more clear for users if we instead merged all arguments (workdir and program) into opts? Then we don't have to add more surface area.

komalali:
@EvanBoyle I had modeled this after the typescript implementation but perhaps that was my mistake?

PR #7173, opened 2021-05-30
[dotnet/sdk] Add create unknown to output utilities

### LINKED ISSUE #7172: [dotnet/sdk] Add `OutputUtilities.CreateUnknown`
Following up from https://pulumi-community.slack.com/archives/CQ2QFLNFL/p1622159397052700, it would be useful to be able to create an unknown output value, which can also be returned from an awaitable value factory that is safe to use during preview; i.e. does not hang. In combination with https://github.com/pulumi/pulumi/pull/7170, the value factory can also await dependent output values as required.

gitfool:
# Description

Fixes #7172.

## Checklist


gitfool:
@mikhailshilkov PTAL

sdk/dotnet/Pulumi/Serialization/Deserializer.cs
```diff
@@ -217,14 +217,12 @@ private static Asset DeserializeAsset(Value value)
...
-            string? urn;
-            if (!TryGetStringValue(value.StructValue.Fields, Constants.ResourceUrnName, out urn))
+            if (!TryGetStringValue(value.StructValue.Fields, Constants.ResourceUrnName, out var urn))
             {
                 throw new InvalidOperationException("Value was marked as a Resource, but did not conform to required shape.");
             }
 
-            string? version;
-            if (!TryGetStringValue(value.StructValue.Fields, Constants.ResourceVersionName, out version)) {
+            if (!TryGetStringValue(value.StructValue.Fields, Constants.ResourceVersionName, out var version)) {
```
mikhailshilkov, line 225: Awesome cleanup across the files. Do you mind doing this in a separate PR? Thanks a ton!

sdk/dotnet/Pulumi/Core/Output.cs
```diff
@@ -26,6 +26,12 @@ public static Output<T> CreateSecret<T>(T value)
         public static Output<T> CreateSecret<T>(Task<T> value)
             => Output<T>.CreateSecret(value);
 
+        public static Output<T> CreateUnknown<T>(T value)
```
mikhailshilkov, line 29: Docs are missing... We should definitely explain why this is needed. I'm inclined to put it into `OutputUtilities` too, given how niche the scenario is...

sdk/dotnet/Pulumi/Serialization/Deserializer.cs
gitfool: Will do. Removed.

sdk/dotnet/Pulumi/Core/Output.cs
gitfool: Moved and documented.

gitfool:
@mikhailshilkov I've addressed your initial review.

sdk/dotnet/Pulumi/Core/OutputUtilities.cs
```diff
@@ -11,6 +12,27 @@ namespace Pulumi.Utilities
     /// </summary>
     public static class OutputUtilities
     {
+        /// <summary>
+        /// Create an unknown with the given value.
+        /// Note: generally, this should never be used since an unknown never resolves during preview.
```
t0yv0, line 17: Since we're repeating the warnings perhaps we can lift them to the class level?

sdk/dotnet/Pulumi/Core/Output.cs
```diff
@@ -327,5 +333,12 @@ void Update<X>(OutputData<X> data, ref X location)
                 (isKnown, isSecret) = OutputData.Combine(data, isKnown, isSecret);
             }
         }
+
+        internal static Output<T> Unknown() => Unknown(default!);
```
t0yv0, line 337: It's a little hard to see syntactically but it seems that the `default!` is not leaking right, the only place we use this is in `Unknown().Apply(_ => valueFactory());`? If we could inline there would be easier on the reader to see that no code ever tries to inspect the `default!`.

t0yv0 (approved):
LGTM, with the usual uneasiness about exposing more internals, but I do believe it's useful in testing..

sdk/dotnet/Pulumi/Core/Output.cs
gitfool: Done

sdk/dotnet/Pulumi/Core/OutputUtilities.cs
gitfool: Done

gitfool:
@t0yv0 I've addressed your review too.

t0yv0:
I believe the test failures are not related. Merging up.

PR #7178, opened 2021-06-01
[dotnet] Fix Resharper code issues

### LINKED ISSUE #7177: [dotnet] Fix Resharper code issues
[Resharper](https://www.jetbrains.com/resharper) can be used to highlight "code issues", a lot of which are purely cosmetic but still useful to apply for code consistency.

gitfool:
# Description

Fixes #7177.

## Checklist

### DIFF
--- a/sdk/dotnet/Pulumi.Automation/Commands/IPulumiCmd.cs
+++ b/sdk/dotnet/Pulumi.Automation/Commands/IPulumiCmd.cs
@@ -11,7 +11,7 @@ namespace Pulumi.Automation.Commands
     internal interface IPulumiCmd
     {
         Task<CommandResult> RunAsync(
-            IEnumerable<string> args,
+            IList<string> args,
             string workingDir,
             IDictionary<string, string?> additionalEnv,
             Action<string>? onStandardOutput = null,
--- a/sdk/dotnet/Pulumi.Automation/Commands/LocalPulumiCmd.cs
+++ b/sdk/dotnet/Pulumi.Automation/Commands/LocalPulumiCmd.cs
@@ -128,9 +128,9 @@ namespace Pulumi.Automation.Commands
             return alphaNumWord.IsMatch(firstArgument) ? firstArgument : "event-log";
         }
 
-        private class EventLogFile : IDisposable
+        private sealed class EventLogFile : IDisposable
         {
-            private bool _disposedValue;
+            public string FilePath { get; }
 
             public EventLogFile(string command)
             {
@@ -139,37 +139,21 @@ namespace Pulumi.Automation.Commands
                 this.FilePath = Path.Combine(logDir, "eventlog.txt");
             }
 
-            public string FilePath { get; }
-
-            protected virtual void Dispose(bool disposing)
-            {
-                if (!_disposedValue)
-                {
-                    if (disposing)
-                    {
-                        var dir = Path.GetDirectoryName(this.FilePath);
-                        try
-                        {
-                            Directory.Delete(dir, recursive: true);
-                        }
-                        catch (Exception e)
-                        {
-                            // allow graceful exit if for some reason
-                            // we're not able to delete the directory
-                            // will rely on OS to clean temp directory
-                            // in this case.
-                            Trace.TraceWarning("Ignoring exception during cleanup of {0} folder: {1}", dir, e);
-                        }
-                    }
-                    _disposedValue = true;
-                }
-            }
-
             public void Dispose()
             {
-                // Do not change this code. Put cleanup code in 'Dispose(bool disposing)' method
-                Dispose(disposing: true);
-                GC.SuppressFinalize(this);
+                var dir = Path.GetDirectoryName(this.FilePath);
+                try
+                {
+                    Directory.Delete(dir, recursive: true);
+                }
+                catch (Exception e)
+                {
+                    // allow graceful exit if for some reason
+                    // we're not able to delete the directory
+                    // will rely on OS to clean temp directory
+                    // in this case.
+                    Trace.TraceWarning("Ignoring exception during cleanup of {0} folder: {1}", dir, e);
+                }
             }
         }
     }
--- a/sdk/dotnet/Pulumi.Automation/Commands/LocalPulumiCmd.cs
+++ b/sdk/dotnet/Pulumi.Automation/Commands/LocalPulumiCmd.cs
@@ -9,7 +9,6 @@ using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
-
 using CliWrap;
 using Pulumi.Automation.Commands.Exceptions;
 using Pulumi.Automation.Events;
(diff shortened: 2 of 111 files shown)


gitfool:
@mikhailshilkov PTAL 😉

sdk/dotnet/Pulumi.Automation/WorkspaceStack.cs
```diff
@@ -616,14 +615,13 @@ await this.Workspace.RunCommandAsync(new[] { "cancel", "--stack", this.Name, "--
...
-            IEnumerable<string> args,
+            IList<string> args,
             Action<string>? onStandardOutput,
             Action<string>? onStandardError,
             Action<EngineEvent>? onEngineEvent,
             CancellationToken cancellationToken)
         {
-            var argsList = args.ToList();
-            argsList.AddRange(new List<string>() { "--stack", this.Name });
+            args = args.Concat(new[] { "--stack", this.Name }).ToList();
```
gitfool, line 624: This looked like a bug as it never passed `argsList`.

gitfool: Note: I also changed `IEnumerable<string> args` to `IList<string> args` everywhere since the enumerable was enumerated multiple times in some places, so it was being treated like a non-enumerable type anyway.

mikhailshilkov: @komalali This was added in https://github.com/pulumi/pulumi/pull/6300 - does this change look like a legit bug fix to you?

komalali: Yes, the change that actually passes in `--stack` to `RunStackCommand` looks good, thanks for catching that. 

I can't speak to the type change for `IEnumerable` to `IList` though :)

gitfool: Resharper correctly warned about the enumerable being enumerated multiple times, so I fixed that too.

mikhailshilkov:
Thank you @gitfool 

There's a scary number of changes (and types of changes). I looked through them and I _think_ they are fine. Some styles choices are debatable but I guess consistency is more important.

@t0yv0 Do you want to take a look to double-check?

lukehoban:
Is it possible to run this static analysis as part of the build?  I've not sure a one time "cleanup" of these will do much good if this isn't enforced by the build itself - as things will just naturally drift over time.

gitfool:
@lukehoban I haven't used them but JetBrains' [InspectCode](https://www.jetbrains.com/help/resharper/InspectCode.html) or better yet their [dotnet tools cli](https://www.jetbrains.com/help/resharper/ReSharper_Command_Line_Tools.html#install-and-use-resharper-command-line-tools-as-net-core-tools) could be incorporated into the ci build. An additional option could be to install the Roslyn based [StyleCopAnalyzers](https://github.com/DotNetAnalyzers/StyleCopAnalyzers) for compile-time style enforcement.

Also, note that I didn't quite "fix" everything, either with code changes or suppressing them, as I ran out of steam that day. 😩

Issues before this PR:
image
Issues after this PR:
image

gitfool:
I "fixed" the remaining issues, either with code changes or by suppressing them.

Rebased with master, resolved conflicts, and forced pushed.

sdk/dotnet/Pulumi/Serialization/Converter.cs
```diff
@@ -476,8 +476,8 @@ static bool CheckEnumType(Type targetType, Type underlyingType)
             }
         }
 
-        private static ConstructorInfo GetPropertyConstructor(Type outputTypeArg)
+        private static ConstructorInfo? GetPropertyConstructor(Type outputTypeArg)
             => outputTypeArg.GetConstructors(BindingFlags.NonPublic | BindingFlags.Public | BindingFlags.Instance).FirstOrDefault(
-                c => c.GetCustomAttributes<OutputConstructorAttribute>() != null);
+                c => c.GetCustomAttribute<OutputConstructorAttribute>() != null);
```
gitfool, line 481: This also looked like a bug as [`System.Attribute.GetCustomAttributes`](https://docs.microsoft.com/en-us/dotnet/api/system.attribute.getcustomattributes) never returns `null` so this filter is redundant and this method would always return all ctors:
> Returns ... An Attribute array that contains the custom attributes applied to element, _or an empty array if no such custom attributes exist_.

sdk/dotnet/Pulumi.Automation/Commands/LocalPulumiCmd.cs
```diff
@@ -128,9 +128,9 @@ private static string SanitizeCommandName(string? firstArgument)
             return alphaNumWord.IsMatch(firstArgument) ? firstArgument : "event-log";
         }
 
-        private class EventLogFile : IDisposable
+        private sealed class EventLogFile : IDisposable
```
gitfool, line 131: This class was over complicated, using the [dispose pattern](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose#implement-the-dispose-pattern), although without a destructor, when it didn't really need to. It's only used internally, and with a `using` statement, so I `sealed` it and stripped it back to a simple `Dispose`. I `sealed` other simple classes using `Dispose` as well.

sdk/dotnet/Pulumi.Automation/Commands/IPulumiCmd.cs
```diff
@@ -11,7 +11,7 @@ namespace Pulumi.Automation.Commands
     internal interface IPulumiCmd
     {
         Task<CommandResult> RunAsync(
-            IEnumerable<string> args,
+            IList<string> args,
```
t0yv0, line 14: Hm, this was an automatically suggested change? I wonder why it did that. It seems to me IEnumerable has fewer assumptions than IList, and if RunAsync does not need random access, the change weakens it? Just curious why it decided to change this.

sdk/dotnet/Pulumi.Automation/Commands/LocalPulumiCmd.cs
t0yv0: Ok no harm done here, simpler code is easier to read, and if it disposes twice it's idempotent so it's fine. Previous code probably was IDE-generated.

t0yv0:
Eyeballing changes, looks good to me, only one was surprising (IEnumerable -> IList).

To Luke's point, running as part of build seems possible:

https://www.jetbrains.com/help/resharper/ReSharper_Command_Line_Tools.html#run-resharper-command-line-tools

Added as https://github.com/pulumi/pulumi/issues/7229

t0yv0 (approved):

sdk/dotnet/Pulumi.Automation/Commands/IPulumiCmd.cs
gitfool: No, the issue was were multiple enumerations of an enumerable in some places. I then looked carefully at caller usage and determined that it was better to fix the warning by promoting `args` from `IEnumerable` to `IList` and doing so consistently.

t0yv0:
Awesome thank your for explaining the IEnumerable change. I'll run another round of tests just in case.

/run-acceptance-tests

t0yv0:
Python lint failures obviously unrelated. OK this is going in ! :)

Issue #7182, opened 2021-06-01
nodejs multi-language components are 200+MB


EvanBoyle:
Each nodejs component has it's own `node_modules` with some heavy dependencies such as`@pulumi/pulumi` and `grpc`. This leads to 200+MB plugins at a minimum. Install a dozen components on your system and you are at multiple GB in the plugin cache. Add to this fact that we may want to include a node runtime embedded in the plugin at some point and this starts getting out of hand.

We should investigate ways to slim this down. Some options that may be worth considering:

1. Manual audit of dependencies
2. Creating single js file distributions (no node_modules) via something like ncc https://github.com/vercel/ncc
3. provide a pre-installed pseudo-plugin "pulumi-plugin runner" that has a node runtime, common large dependencies (pulumi, grpc, provider SDKs maybe?). that could be referenced from real plugins via package.json `pulumi-plugin-runner: "../pulumi-plugin-runner`. Our plugin installation process could manage this dependency and even have multiple versions installed (one per pulumi runtime version?).

EvanBoyle:
Just tried a very basic `ncc` invocation:

```console
ncc build index.ts -o single
```

and the result is a 22MB javascript file. 90% reduction!

EvanBoyle:
@joeduffy suggested a common node modules solution with the following layout:

```
a/
    package.json
b/
    package.json
c/
    package.json
node_modules/
```

Then running `npm i a/ b/ c/` from the root.

EvanBoyle:
Got an `ncc` based plugin working on one of my projects, but it required some changes to pulumi/pulumi to exclude the typescript dependency. 

- old size: `230 MB`
- new size: `16 MB`

`ncc` does not seem to be advanced enough to exclude certain submodules from being compiled (`@pulumi/pulumi/runtime/closure' for instance), so I had to make a local build of pulumi with all of that deleted. We could probably use a fancier webpack configuration to achieve the same thing, but I think it would be nice to keep our users (and ourselves) out of the business of webpack if possible so I opened https://github.com/pulumi/pulumi/issues/7184

EvanBoyle:
https://github.com/pulumi/pulumi/pull/7188 makes them 150MB

komalali:
[...]

PR #7185, opened 2021-06-01
Add ability to skip installation (npm, venv, etc) step when installing plugins

EvanBoyle:
This change allows plugins to opt out of `npm install`, `venv`, and other installation/setup steps. This allows for instance, a single `index.js` file to be compiled and run by node as per an `ncc` build configuration: https://github.com/pulumi/pulumi/issues/7182#issuecomment-852420810

### DIFF
--- a/sdk/go/common/workspace/plugins.go
+++ b/sdk/go/common/workspace/plugins.go
@@ -374,7 +374,7 @@ func (info PluginInfo) Install(tgz io.ReadCloser) error {
 	if err != nil && !os.IsNotExist(err) {
 		return errors.Wrap(err, "loading PulumiPlugin.yaml")
 	}
-	if proj != nil {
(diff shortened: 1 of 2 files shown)


justinvp:
An alternative would be to just not include the `PulumiPlugin.yaml` file in the plugin package.

EvanBoyle:
ah, good point. there might be further design to do here especially if we go down the route of node modules sharing for common dependencies like the node runtime, pulumi, grcp, etc.

Frassle:
> ah, good point

Not really, removing the PulumiPlugin.yaml means the engine doesn't know what runtime to start the plugin as and so assumes its a native executable. A single .js file is not a native executable.

justinvp:
You could have a script (both shell script, and .bat script for Windows) that runs `node index.js`. Not saying that's ideal, but it's already possible as a workaround.

Issue #7184, opened 2021-06-01
Remove typescript dependency from core `@pulumi/pulumi` nodejs sdk


EvanBoyle:
Today our closure serialization libraries use typescript at runtime. That dependency alone is 50mb. We may want to consider isolating this dependency into a separate package outside of `@pulumi/pulumi` to make life easier for consumers (ie no crazy webpack configuration to shake this dependency out). This would involve moving the following into separate packages:

- dynamic providers
- sdk/node/runtime/closure

This would greatly contribute to improving https://github.com/pulumi/pulumi/issues/7182

EvanBoyle:
Note that this is already causing us to implement strange behavior and workarounds in other areas like the pulumi github action: 

https://github.com/pulumi/actions/blob/master/scripts/replace-ts.js

lukehoban:
Moving these would of course be a non-trivial breaking change.  Also, package like `@pulumi/aws` actually use closure serialization, so I expect they would need to import this new dependency anyway.  It might not be simple to unravel this.

An alternative would be to remove the `typescript` dependency entirely.  I am fairly certain it is only used to parse JavaScript code - and there are *much* lighter-weight options available for that.  It is used for:
* `parseFunction` - to identify lexically captured variables and whether a non-lexical this is referenced
* `rewriteSuperReferences` - to rewrite calls to `super(...)` though I don't recall exactly why this is even needed
* `isLegalFunctionName` - to determine if a string is a valid JS identifier

I believe none of these actually process TypeScript source code, and also that none of these use any type-directed logic.  So a much simpler JavaScript parser could likely be used as a replacement.

EvanBoyle:
Perhaps we should consider making this change for 4.0. 

I dug up everything we import, many types but looks like a lot of functionality as well: 

```
import {
    ClassDeclaration, isClassDeclaration, ConstructorDeclaration, isConstructorDeclaration, SyntaxKind, isExpressionStatement, isArrowFunction, createSourceFile as tsCreateSourceFile, ScriptTarget,
    ScriptKind, SourceFile, Diagnostic, forEachChild, Node, CallExpression, FunctionLikeDeclarationBase, FunctionExpression, Identifier, MethodDeclaration, ThisExpression, Block, CatchClause, PropertyAssignment,
[...]

PR #7186, opened 2021-06-01
remove all typescript dependencies

EvanBoyle:
This branch shows concretely what would need to be moved into a separate package to drop the runtime typescript dependency from `@pulumi/pulumi` a la https://github.com/pulumi/pulumi/issues/7184


lukehoban:
As noted in https://github.com/pulumi/pulumi/issues/7184, this would be a pretty significant breaking change, and I'm not even sure that things like `@pulumi/aws` can work without this (or would need to take a dependency on the new library that this is factored into).  There's a suggestion in https://github.com/pulumi/pulumi/issues/7184#issuecomment-852441101 on a non-breaking way to achieve this size reduction.

pgavlin:
Closing--I think this example has served its purpose.

PR #7188, opened 2021-06-01
skip installing dev dependencies for nodejs plugin setup

EvanBoyle:
# Description

This change skips installation of `devDependencies` by running  `yarn install --production` or `npm install --production` when installing pulumi plugins that specify the `nodejs` runtime via `PulumiPlugin.yaml`. 

In my tests, this results in a 35% reduction in size (my plugins include typescript as a dev dep) from `230MB` -> `150MB`. 

Contributes to https://github.com/pulumi/pulumi/issues/7182

## Checklist

Not a user facing features, simply an optimization.


komalali (approved):
TIL we include devDependencies in our plugins

PR #7193, opened 2021-06-02
Bump the Pulumi Service API version header to 7

praneetloke:
# Description

It ensures users are using a CLI version that supports publishing policy packs to Azure Storage.

Fixes # `N/A`

## Checklist

~~- [ ] I have added tests that prove my fix is effective or that my feature works~~ Bumping the service API version.
~~- [ ] I have updated the [CHANGELOG-PENDING](https://github.com/pulumi/pulumi/blob/master/CHANGELOG_PENDING.md) file with my change~~


chrsmith (approved):
This is a fine thing to add, I guess. But IMHO it's kind of annoying since we've bumped the version number maybe 4-5 times in the past few years, and in that same timeframe submitted maybe 1,000+ PRs.

chrsmith (commented):
Post-hoc LGTM. I didn't know we were all the way up to version 7. Fancy!

Issue #6990, opened 2021-05-07
"The Pulumi CLI encountered a fatal error. This is a bug!" while trying `pulumi import`


patrickmburrell:
OS: macOS Catalina, v10.15.7
computer: MacBook Pro (Retina, 15-inch, Late 2013)

I ran the following command:
`pulumi import aws:ec2/instance/Instance "[resource name here]" [id-here]`

I wasn't sure about the type token to use, and I couldn't find any documentation for it, so I just guessed at `aws:ec2/instance/Instance`. I got the following response:

================================================================================
The Pulumi CLI encountered a fatal error. This is a bug!
We would appreciate a report: https://github.com/pulumi/pulumi/issues/
Please provide all of the below text in your report.
================================================================================
Pulumi Version:   v3.2.1
Go Version:       go1.16.3
Go Compiler:      gc
Architecture:     amd64
Operating System: darwin
Panic:            fatal: An assertion has failed: Module member token 'aws:ec2/instance/Instance' missing module member delimiter

goroutine 1 [running]:
runtime/debug.Stack(0xc0018bf240, 0x55d05e0, 0xc0005f2bc0)
        /usr/local/Cellar/go/1.16.3/libexec/src/runtime/debug/stack.go:24 +0x9f
main.panicHandler()
        /private/tmp/pulumi-20210506-41645-1px53ac/pkg/cmd/pulumi/main.go:29 +0x76
panic(0x55d05e0, 0xc0005f2bc0)
        /usr/local/Cellar/go/1.16.3/libexec/src/runtime/panic.go:965 +0x1b9
github.com/pulumi/pulumi/sdk/v3/go/common/util/contract.failfast(...)
        /private/tmp/pulumi-20210506-41645-1px53ac/sdk/go/common/util/contract/failfast.go:23
github.com/pulumi/pulumi/sdk/v3/go/common/util/contract.Assertf(0x7ffeefbff500, 0x5993e0d, 0x38, 0xc0018bf490, 0x1, 0x1)
        /private/tmp/pulumi-20210506-41645-1px53ac/sdk/go/common/util/contract/assert.go:33 +0x1a5
github.com/pulumi/pulumi/sdk/v3/go/common/tokens.ModuleMember.Module(0x7ffeefbff5ce, 0x19, 0xc000691410, 0x0)
        /private/tmp/pulumi-20210506-41645-1px53ac/sdk/go/common/tokens/tokens.go:192 +0xd2
[...]

PR #7202, opened 2021-06-03
[cli] Protect against panic when using incorrect token type in pulumi import

stack72:
Fixes: #6990

Previously:

```
pulumi import aws:ec2/instance/Instance id-12345 test-instance
Previewing import (dev)

================================================================================
The Pulumi CLI encountered a fatal error. This is a bug!
We would appreciate a report: https://github.com/pulumi/pulumi/issues/
Please provide all of the below text in your report.
================================================================================
Pulumi Version:   3.4.0-alpha.1622658772+dbea8058.dirty
Go Version:       go1.16.4
Go Compiler:      gc
Architecture:     amd64
Operating System: darwin
Panic:            fatal: An assertion has failed: Module member token 'aws:ec2/instance/Instance' missing module member delimiter
```

After:

```
pulumi import aws:ec2/instance/Instance id-12345 test-instance
Previewing import (dev)

error: import type "aws:ec2/instance/Instance" is not a valid resource type - refer to the import section of the provider resource documentation.
```

### DIFF
--- a/pkg/engine/deployment.go
+++ b/pkg/engine/deployment.go
@@ -16,6 +16,7 @@ package engine
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/opentracing/opentracing-go"
@@ -25,6 +26,7 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/common/diag"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource/plugin"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/tokens"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/fsutil"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/result"
@@ -173,6 +175,12 @@ func newDeployment(ctx *Context, info *deploymentContext, opts deploymentOptions
 		}
 		for i := range opts.imports {
 			imp := &opts.imports[i]
+			_, err := tokens.ParseTypeToken(imp.Type.String())
+			if err != nil {
+				return nil, errors.New(fmt.Sprintf("import type %q is not a valid resource type token. "+
+					"Type tokens must be of the format <package>:<module>:<type> - "+
+					"refer to the import section of the provider resource documentation.", imp.Type.String()))
+			}
 			if imp.Provider == "" && imp.Version == nil {
 				imp.Version = defaultProviderVersions[imp.Type.Package()]
 			}
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -8,6 +8,9 @@
 - [cli] - Send plugin install output to stderr, so that it doesn't
   clutter up --json, automation API scenarios, and so on.
   [#7115](https://github.com/pulumi/pulumi/pull/7115)
+  
+- [cli] Protect against panics when using the wrong resource type with `pulumi import`
+  [#7202](https://github.com/pulumi/pulumi/pull/7202)
 
 - [auto/nodejs] - Emit warning instead of breaking on parsing JSON events for automation API.
   [#7162](https://github.com/pulumi/pulumi/pull/7162)


pkg/engine/deployment.go
```diff
@@ -173,6 +175,11 @@ func newDeployment(ctx *Context, info *deploymentContext, opts deploymentOptions
 		}
 		for i := range opts.imports {
 			imp := &opts.imports[i]
+			_, err := tokens.ParseTypeToken(imp.Type.String())
+			if err != nil {
+				return nil, errors.New(fmt.Sprintf("import type %q is not a valid resource type - "+
+					"refer to the import section of the provider resource documentation.", imp.Type.String()))
```
lukehoban, line 181: Should we note explicitly that type tokens must be of the form `<package>:<module>:<type>`?

lukehoban (approved):
Can we add a test for this?

This feels like an okay fix for now to address the common user-facing issue - but I will say I think we will need to directionally do one of two things:
1. Ensure that we validate all strings that we convert to type tokens at the point we take them from the outside world, and produce useful error messages there - and then assume that all things typed as `token.Type` are valid
2. Admit that we aren't cleanly ensuring that all things typed as `token.Type` are valid, and change the API for `token.Type` to return errors instead of asserting (which would require a *lot* of changes to consumers to handle all the new error cases.

I personally think we should be doing (1).  The changes in this PR aren't really either (1) or (2) - they are patching the problem in between these two in one place (but I'll bet there are dozens more like this as latent issues).

stack72:
> 1. Ensure that we validate all strings that we convert to type tokens at the point we take them from the outside world, and produce useful error messages there - and then assume that all things typed as `token.Type` are valid

I definitely think we need to more in advance here - we can do something at the cli layer but we'd need to make sure that work is correct. Would you be ok for me to add a follow up issue here to make sure we investigate that but proceed with this to step any panics for now?

lukehoban:
>  Would you be ok for me to add a follow up issue here to make sure we investigate that but proceed with this to step any panics for now?

Yep - happy to open a follow-up issue.  Basically any place we use `types.Token(<string>)` should actually be parsing the results and returning an error if the string is not a token.

```
$ grep -r --include "*.go" --exclude "*_test.go" "tokens.Type("  .
./sdk/go/common/resource/plugin/provider_server.go:     typ, name, parent := tokens.Type(req.GetType()), tokens.QName(req.GetName()), resource.URN(req.GetParent())
./sdk/go/common/resource/urn.go:        return tokens.Type(strings.Split(urn.URNName(), URNNameDelimiter)[2])
./sdk/go/common/resource/urn.go:        return tokens.Type(lastType)
./sdk/go/pulumi/mocks.go:       parentType := tokens.Type("")
./sdk/go/pulumi/mocks.go:       return string(resource.NewURN(tokens.QName(m.stack), tokens.PackageName(m.project), parentType, tokens.Type(typ),
./pkg/cmd/pulumi/import.go:             Type:    tokens.Type(typ),
./pkg/operations/operations_cloud_aws.go:       cloudFunctionType     = tokens.Type("cloud:function:Function")
./pkg/operations/operations_cloud_aws.go:       cloudLogCollectorType = tokens.Type("cloud:logCollector:LogCollector")
./pkg/operations/operations_cloud_aws.go:       cloudServiceType      = tokens.Type("cloud:service:Service")
./pkg/operations/operations_cloud_aws.go:       cloudTaskType         = tokens.Type("cloud:task:Task")
./pkg/operations/operations_gcp.go:     gcpFunctionType = tokens.Type("gcp:cloudfunctions/function:Function")
./pkg/operations/resources.go:                  childURN.Type() == tokens.Type(typ) &&
./pkg/operations/operations_aws.go:     awsFunctionType = tokens.Type("aws:lambda/function:Function")
./pkg/operations/operations_aws.go:     awsLogGroupType = tokens.Type("aws:cloudwatch/logGroup:LogGroup")
./pkg/resource/deploy/deployment.go:    parentType := tokens.Type("")
./pkg/resource/deploy/providers/reference.go:   return tokens.Type("pulumi:providers:" + pkg)
./pkg/resource/deploy/source_eval.go:           t = tokens.Type(req.GetType())
```

The case that causes this particular bug is in the middle of the results above.  Most of the others look like they are probably safe - but should likely still be using a parsing function and propagating an error just to be sure.

pkg/engine/deployment.go
stack72: So I thought about that but then I sort of thought how a user would think about turning:

```
aws:ec2/instance/Instance
```

into

```
aws:ec2/instance:Instance
```

Do you think `<package>:<module>:<type>` would help them see the differences?

PR #7203, opened 2021-06-03
Adding a workflow dispatch to GHA to test master builds from branches

stack72:
# Description

Fixes # (issue)

## Checklist


stack72:
Self merging as this is required to test a feature on another branch before a PR hits

PR #7208, opened 2021-06-04
[codegen/go] Don't error for 3rd party Go SDKs - assume v1 default instead

lukehoban:
Printing to stdout in `init` can interfere with the user's program behaviour, and in particular for multi-lang components leads to printing to stdout before the language host plugin can send back a port to connect on.

Before these changes, any attempt to use my provider leads to this (not also - this code previously wasn't correctly printing the error object and ended up with an errant `%v`):

```
Previewing refresh (test)

View Live: https://app.pulumi.com/lukehoban/serverless-simple/test/previews/7be2cbd8-180d-4ceb-9d5f-77d924a272f5

 
error: could not load plugin for serverless provider 'urn:pulumi:test::serverless-simple::pulumi:providers:serverless::default_0_0_1': serverless (resource) plugin [/Users/lukehoban/go/bin/pulumi-resource-serverless] wrote a non-numeric port to stdout ('failed to determine package version. defaulting to v1: %v failed to determine the package version from github.com/lukehoban/pulumi-package-aws-apigateway/sdk/go/apigateway'): strconv.Atoi: parsing "failed to determine package version. defaulting to v1: %v failed to determine the package version from github.com/lukehoban/pulumi-package-aws-apigateway/sdk/go/apigateway": invalid syntax
```

After, I get a warning but am not blocked:

```
Refreshing (test)

View Live: https://app.pulumi.com/lukehoban/serverless-simple/test/updates/10

     Type                                 Name                    Status      Info
     pulumi:pulumi:Stack                  serverless-simple-test              1 mess
     ├─ aws:sqs:Queue                     receiverQueue                       
     └─ serverless:index:Service          svc                                 
        ├─ aws:s3:Bucket                  bucket                              
 ~      ├─ aws:lambda:EventSourceMapping  receiver-sqs-0          updated     [d
 ~      ├─ aws:s3:BucketObject            payload                 updated     [d
 -      ├─ aws:lambda:Function            receiver                deleted     
 -      ├─ aws:lambda:Function            sender                  deleted     
 -      └─ aws:iam:Role                   role                    deleted     
 
Diagnostics:
  pulumi:pulumi:Stack (serverless-simple-test):
    failed to determine the package version from github.com/lukehoban/pulumi-package-aws-apigateway/sdk/go/apigateway
 
Resources:
    ~ 2 updated
    - 3 deleted
    5 changes. 4 unchanged

Duration: 4s
```

Separately of course, I don't really want this warning at all.  Avoiding the warning appears to be related to https://github.com/pulumi/pulumi/pull/6866#issuecomment-854364151.


justinvp (approved):

lukehoban:
Actually - this shouldn't be an error at all - and actually the logic here likely needs to be revisited - I've opened https://github.com/pulumi/pulumi/issues/7224 on the issue itself.  To unblock things for the common case, I'm for now going to just make this not an error/warning at all - which will allow the "normal" case of v1 3rd party packages to work - but there will be more change needed to make 3rd party packages actually work correctly - and I think for that bigger change we can likely just get rid of `PkgVersion` entirely.

lblackstone (approved):

iwahbe:
I'm going to avoid removing the returned error, since that is a breaking change. Instead, the second returned value will always be `nil`. This is document in the generated code.

justinvp:
> I think for that bigger change we can likely just get rid of PkgVersion entirely

I don't think it was intentional to export this function. We should consider making it lowercase (if we need to keep it) -- I highly doubt we'll break anyone, but it technically would be a breaking change.

iwahbe:
@justinvp We need to export it because it is called in submodules.

Issue #5236, opened 2020-08-27
    panic: fatal: A failure has occurred: Unrecognized structpb value kind in RPC[tf.Provider[aws].Invoke(aws:iam/getPolicyDocument:getPolicyDocument).args]: <nil>


maratsh:
```
const policyDoc = pulumi.output(
      aws.iam.getPolicyDocument(
        {
          statements: [
            {
              actions: [
                'sqs:ChangeMessageVisibility',
                'sqs:ChangeMessageVisibilityBatch',
                'sqs:DeleteMessage',
                'sqs:DeleteMessageBatch',
                'sqs:GetQueueAttributes',
                'sqs:ReceiveMessage',
              ],
              resources: [ args.sqsArn ],
            },
          ],
        }
      )
    );

    const policy = new aws.iam.Policy(name, {
      policy: policyDoc.json
    });
```

is throwing on pulumi up

Diagnostics:
  aws:sns:TopicSubscription (orders):
    error: transport is closing
 
  pulumi:pulumi:Stack (stack):
    panic: fatal: A failure has occurred: Unrecognized structpb value kind in RPC[tf.Provider[aws].Invoke(aws:iam/getPolicyDocument:getPolicyDocument).args]: <nil>
    goroutine 97 [running]:
    github.com/pulumi/pulumi/sdk/v2/go/common/util/contract.failfast(...)
    	/home/runner/go/pkg/mod/github.com/pulumi/pulumi/sdk/v2@v2.9.1-0.20200825190708-910aa96016cd/go/common/util/contract/failfast.go:23
    github.com/pulumi/pulumi/sdk/v2/go/common/util/contract.Failf(0x5fcabac, 0x2f, 0xc001b5cec8, 0x2, 0x2)
    	/home/runner/go/pkg/mod/github.com/pulumi/pulumi/sdk/v2@v2.9.1-0.20200825190708-910aa96016cd/go/common/util/contract/fail.go:30 +0x181
    github.com/pulumi/pulumi/sdk/v2/go/common/resource/plugin.UnmarshalPropertyValue(0xc000c4bd00, 0xc00114c050, 0x49, 0x101, 0xc000766090, 0x0, 0x0)
    	/home/runner/go/pkg/mod/github.com/pulumi/pulumi/sdk/v2@v2.9.1-0.20200825190708-910aa96016cd/go/common/resource/plugin/rpc.go:353 +0x3bd
[...]

PR #7210, opened 2021-06-04
[cli] prevent panic in pulumi python applications when apply is used incorrectly

stack72:
Fixes: #5236

This bug *only* affected Python - as other language runtimes caught the
incorrect use of apply at compile time. Given the following Pulumi program:

```
from pulumi_aws import iam, s3
my_bucket = s3.Bucket(
    "simple-bucket",
    force_destroy=True,
    acl="private",
    versioning=s3.BucketVersioningArgs(enabled=True),
)
policy_document = iam.get_policy_document(
    statements=[
        iam.GetPolicyDocumentStatementArgs(
            sid="",
            effect="Allow",
            actions=["s3:GetObject"],
            resources=[my_bucket.arn.apply(lambda arn: f"{arn}/*")],
        ),
    ]
)
```

The apply is incorrect here as the Invoke should be run as part of the apply. This
application will produce a panic message similar to the following:

```
pulumi:pulumi:Stack (my-stack-dev):
    panic: fatal: A failure has occurred: Unrecognized structpb value kind in RPC[tf.Provider[aws].Invoke(aws:iam/getPolicyDocument:getPolicyDocument).args]: <nil>
    goroutine 114 [running]:
    	/home/runner/go/pkg/mod/github.com/pulumi/pulumi/sdk/v2@v2.9.1-0.20200825190708-910aa96016cd/go/common/util/contract/failfast.go:23
    github.com/pulumi/pulumi/sdk/v2/go/common/util/contract.Failf(0x6bc984f, 0x2f, 0xc0010daec8, 0x2, 0x2)
```

In order to help the user, we are able to detect if the error is due to an Invoke and point
them in the direction of the `apply` docs:

```
    Exception: invoke of aws:iam/getPolicyDocument:getPolicyDocument failed: invocation of aws:iam/getPolicyDocument:getPolicyDocument returned an error:
    Incorrect use of `apply`. Please refer to https://www.pulumi.com/docs/intro/concepts/inputs-outputs/#apply for further details on how to use `apply`
```

# Description

Fixes # (issue)

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -13,6 +13,9 @@
   
 - [cli] Protect against panics when using the wrong resource type with `pulumi import`
   [#7202](https://github.com/pulumi/pulumi/pull/7202)
+  
+- [cli] Prevent a panic in python programs when incorrect use of `apply` is included in pulumi program
+  [#7210](https://github.com/pulumi/pulumi/pull/7210)
 
 - [auto/nodejs] - Emit warning instead of breaking on parsing JSON events for automation API.
   [#7162](https://github.com/pulumi/pulumi/pull/7162)
--- a/sdk/go/common/resource/plugin/rpc.go
+++ b/sdk/go/common/resource/plugin/rpc.go
@@ -17,6 +17,7 @@ package plugin
 import (
 	"reflect"
 	"sort"
+	"strings"
 
 	structpb "github.com/golang/protobuf/ptypes/struct"
 	"github.com/pkg/errors"
@@ -414,7 +415,8 @@ func UnmarshalPropertyValue(v *structpb.Value, opts MarshalOptions) (*resource.P
 
 			var ref resource.PropertyValue
 			if hasID {
-				ref = resource.MakeCustomResourceReference(resource.URN(urn.StringValue()), resource.ID(id), packageVersion)
+				ref = resource.MakeCustomResourceReference(resource.URN(urn.StringValue()),
+					resource.ID(id), packageVersion)
 			} else {
 				ref = resource.MakeComponentResourceReference(resource.URN(urn.StringValue()), packageVersion)
 			}
@@ -424,8 +426,20 @@ func UnmarshalPropertyValue(v *structpb.Value, opts MarshalOptions) (*resource.P
 		}
 
 	default:
-		contract.Failf("Unrecognized structpb value kind in RPC[%s]: %v", opts.Label, reflect.TypeOf(v.Kind))
-		return nil, nil
+		if opts.Label != "" && strings.Contains(opts.Label, ".Invoke") {
+			// we believe this is an issue with an apply being used incorrectly
+			// and thus should point the user at some specific documentation to help
+			// them understand the apply
+			// the check for the invoke is due to an unhandled exception in Python when the apply is incorrect
+			// in NodeJS / Go etc, this is caught at build time but in Python it is not caught until runtime
+			// the structure of the opts.Label will be
+			// tf.Provider[aws].Invoke(aws:iam/getPolicyDocument:getPolicyDocument).args (as an example)
+			return nil, errors.New("\nIncorrect use of `apply`. Please refer to " +
+				"https://www.pulumi.com/docs/intro/concepts/inputs-outputs/#apply for further details on " +
+				"how to use `apply`")
+		}
+		return nil, errors.Errorf("Unrecognized structpb value kind in RPC[%s]: %v",
+			opts.Label, reflect.TypeOf(v.Kind))
 	}
 }
 


t0yv0:
Hmm can I get some help understanding why we would disallow this invocation of get_policy_document?

```
def get_policy_document(override_json: Optional[str] = None,
                        override_policy_documents: Optional[Sequence[str]] = None,
                        policy_id: Optional[str] = None,
                        source_json: Optional[str] = None,
                        source_policy_documents: Optional[Sequence[str]] = None,
                        statements: Optional[Sequence[pulumi.InputType['GetPolicyDocumentStatementArgs']]] = None,
                        version: Optional[str] = None,
                        opts: Optional[pulumi.InvokeOptions] = None) -> AwaitableGetPolicyDocumentResult:
```

At a first glance its return type is async, something like `Awaitable[GetPolicyDocumentResult]`. So if you wanted to unpack it, you could `await` it. Why is this bad/ not allowed?

Also. I am working on a codegen PR adding the 'Output' lifted form:

```
def get_policy_document_output(override_json: Optional[Input[str]] = None,
                        override_policy_documents: Optional[Input[Sequence[str]]] = None,
                        ...
                        opts: Optional[pulumi.InvokeOptions] = None) -> Output[GetPolicyDocumentResult]:
```

Will that be allowed at top level? I feel like I need to test this out.

stack72:
Summarising what I just talked to @t0yv0 about - the reason this is not valid is that the apply returns an output and the get_policy_document only accepts plain inputs. This is not a problem in NodeJS / Dotnet / Go lanugage hosts as we catch this at build time.

In Python we don't and thus we were originally panicing - when the work to allow outputs to be passed to data sources is in place, we can remove this relaxing addition

t0yv0 (approved):
It does fix it for me but I'm digging in related PR for where in the code we could put the fix and the test for this, where it "belongs". This location here looks a little too magical.

t0yv0:
Let's merge though as this is a P1..

lukehoban:
This doesn't feel like the right fix.

The problem is not an "incorrect use of apply", it's that Outputs are not supported as arguments to invoke. 

But we also shouldn't be trying to handle this at this layer via grepping opts.Label.  We should add this check in at the `Invoke` layer, as well as in the SDK layer itself.  

If we are going to take a quick patch fix for the P1, I would just take the part of this that reports the same message we do currently but as an error instead of panic.  That seems appropriate as a change in isolation, as this is a public API that should have clean error paths for invalid inputs.

t0yv0:
@lukehoban do you have any thoughts on t0yv0/keep-unknowns-fix ?

There are some options there, including `add this check in at the Invoke layer` via RejectUnknowns: true at Invoke.

Alternatively, surprisingly, adding `KeepUnknowns: true` makes Outputs work as arguments to Invoke (in Python). Are we sure we do not support it?

t0yv0:
As we talked about let me try to repro this on master with latest fixes.

t0yv0:
I cannot reproduce the issue when using `pulumi` CLI from master. The provided program seems to work as intended (though we believe it's incorrect):

```

# Add this:
#
# pulumi.export('policy_document', policy_document)
#
# $ pulumi up --yes
#
#
# $ pulumi stack output policy_document | jq .
#
# {
#   "id": "2827767266",
#   "json": "{\n  \"Version\": \"2012-10-17\",\n  \"Statement\": [\n    {\n      \"Sid\": \"\",\n      \"Effect\": \"Allow\",\n      \"Action\": \"s3:GetObject\",\n      \"Resource\": \"arn:aws:s3:::simple-bucket-571a4c2/*\"\n    }\n  ]\n}",
#   "override_json": null,
#   "override_policy_documents": null,
#   "policy_id": null,
#   "source_json": null,
#   "source_policy_documents": null,
#   "statements": [
#     {
#       "actions": [
#         "s3:GetObject"
#       ],
#       "conditions": [],
#       "effect": "Allow",
#       "not_actions": [],
#       "not_principals": [],
#       "not_resources": [],
#       "principals": [],
#       "resources": [
#         "arn:aws:s3:::simple-bucket-571a4c2/*"
#       ],
#       "sid": ""
#     }
#   ],
#   "version": "2012-10-17"
# }
#
#
# $ pulumi stack output policy_document | jq '.json|fromjson'
#
# {
#   "Version": "2012-10-17",
#   "Statement": [
#     {
#       "Sid": "",
#       "Effect": "Allow",
#       "Action": "s3:GetObject",
#       "Resource": "arn:aws:s3:::simple-bucket-571a4c2/*"
#     }
#   ]
# }

```

t0yv0:
I think the fix https://github.com/pulumi/pulumi/pull/7214 made this work.

stack72:
Superceded via #7214

Issue #7132, opened 2021-05-25
pulumi up crashes with generic error message "transport is closing"


aureq:
When calling `pulumi up`, the command crashes and returns the following error
`Exception: invoke of aws:iam/getPolicyDocument:getPolicyDocument failed: invocation of aws:iam/getPolicyDocument:getPolicyDocument returned an error: transport is closing`

## Expected behavior

The CLI shouldn't crash and a more specific error message should be displayed.

## Current behavior

the command `pulumi up` crashes, dumps some debug information along with other error messages.

```
Previewing update (crash)

View Live: https://app.pulumi.com/aureq/aws-py-wireguard-server-transport-closing/crash/previews/a39490fe-3c1a-4315-bbad-8c8ea88baa8c

     Type                                    Name                                             Plan       Info
 +   pulumi:pulumi:Stack                     aws-py-wireguard-server-transport-closing-crash  create     3 errors; 34 messages
 +   └─ aws:secretsmanager:Secret            wireguard-vpn-secret                             create     
 +      └─ aws:secretsmanager:SecretVersion  wireguard-vpn-secret                             create     
 
Diagnostics:
  pulumi:pulumi:Stack (aws-py-wireguard-server-transport-closing-crash):
    panic: fatal: A failure has occurred: Unrecognized structpb value kind in RPC[tf.Provider[aws].Invoke(aws:iam/getPolicyDocument:getPolicyDocument).args]: <nil>
    goroutine 99 [running]:
    github.com/pulumi/pulumi/sdk/v3/go/common/util/contract.failfast(...)
        /home/runner/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.0.0/go/common/util/contract/failfast.go:23
    github.com/pulumi/pulumi/sdk/v3/go/common/util/contract.Failf(0x72dbe51, 0x2f, 0xc001c96a00, 0x2, 0x2)
        /home/runner/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.0.0/go/common/util/contract/fail.go:30 +0x18d
    github.com/pulumi/pulumi/sdk/v3/go/common/resource/plugin.UnmarshalPropertyValue(0xc002194200, 0xc001cb39f0, 0x49, 0x101, 0x0, 0xc002198170, 0x0, 0x0)
[...]

PR #7212, opened 2021-06-04
Keep unknown fix

t0yv0:
Isolating a panic from #7132.

- during discussion it came up that 7132 is not a valid program (not well typed; invoke should not accept Output that can be unknown)

- as far as I can tell from static code analysis, callers of this function (such as resmon.Invoke), marshal with KeepUnknowns = true

- if we believe that invoke should never have unknowns, should we instead do RejectUnknowns = true to fail fast?

- there may be a separate issue with list-nested unknown removal

# Description

Fixes # (issue)

## Checklist

### DIFF
--- a/sdk/go/common/resource/plugin/provider_plugin.go
+++ b/sdk/go/common/resource/plugin/provider_plugin.go
@@ -1172,6 +1172,7 @@ func (p *provider) Invoke(tok tokens.ModuleMember, args resource.PropertyMap) (r
 		Label:         fmt.Sprintf("%s.args", label),
 		KeepSecrets:   p.acceptSecrets,
 		KeepResources: p.acceptResources,
+		KeepUnknowns:  true,
 	})
 	if err != nil {
 		return nil, nil, err
--- a/sdk/go/common/resource/plugin/provider_plugin_test.go
+++ b/sdk/go/common/resource/plugin/provider_plugin_test.go
@@ -1,12 +1,18 @@
 package plugin
 
 import (
+	"context"
+	"fmt"
 	"reflect"
 	"testing"
 
+	empty "github.com/golang/protobuf/ptypes/empty"
+	structpb "github.com/golang/protobuf/ptypes/struct"
 	"github.com/stretchr/testify/assert"
+	grpc "google.golang.org/grpc"
 
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+	pulumirpc "github.com/pulumi/pulumi/sdk/v3/proto/go"
 )
 
 func TestAnnotateSecrets(t *testing.T) {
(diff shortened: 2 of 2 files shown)


t0yv0:
Folks I'll need some expert advise which way is up from here..

t0yv0:
Alt fix from Paul https://github.com/pulumi/pulumi/pull/7210

t0yv0:
If I take `RejectUnknowns: true` in resmon.Invoke the user experience is:

```
    pulumi:pulumi:Stack (repro-7132-dev):
      File "/Users/anton/.nix-profile/bin/pulumi-language-python-exec", line 92, in <module>
        loop.run_until_complete(coro)
        return future.result()
      File "/Users/anton/pulumi/sdk/python/lib/pulumi/runtime/stack.py", line 110, in run_in_stack
      File "/Users/anton/pulumi/sdk/python/lib/pulumi/runtime/stack.py", line 43, in run_pulumi_func
        func()
      File "/Users/anton/pulumi/sdk/python/lib/pulumi/runtime/stack.py", line 110, in <lambda>
      File "/Users/anton/pulumi/sdk/python/lib/pulumi/runtime/stack.py", line 133, in __init__
        func()
      File "/Users/anton/.nix-profile/bin/pulumi-language-python-exec", line 91, in <lambda>
        coro = pulumi.runtime.run_in_stack(lambda: runpy.run_path(args.PROGRAM, run_name='__main__'))
      File "/nix/store/7q3qsarzyffcm2sjb2yhjgy0d3hzdi7y-python3-3.8.8/lib/python3.8/runpy.py", line 282, in run_path
        return _run_code(code, mod_globals, init_globals,
      File "/nix/store/7q3qsarzyffcm2sjb2yhjgy0d3hzdi7y-python3-3.8.8/lib/python3.8/runpy.py", line 87, in _run_code
        exec(code, run_globals)
      File "./__main__.py", line 55, in <module>
        role_policy = iam.get_policy_document(
      File "/Users/anton/repro-7132/venv/lib/python3.8/site-packages/pulumi_aws/iam/get_policy_document.py", line 185, in get_policy_document
        __ret__ = pulumi.runtime.invoke('aws:iam/getPolicyDocument:getPolicyDocument', __args__, opts=opts, typ=GetPolicyDocumentResult).value
      File "/Users/anton/pulumi/sdk/python/lib/pulumi/runtime/invoke.py", line 163, in invoke
        raise invoke_error
    Exception: invoke of aws:iam/getPolicyDocument:getPolicyDocument failed: failed to unmarshal aws:iam/getPolicyDocument:getPolicyDocument args: unexpected unknown property value

```

pkg/resource/deploy/source_eval.go
```diff
@@ -562,7 +562,8 @@ func (rm *resmon) Invoke(ctx context.Context, req *pulumirpc.InvokeRequest) (*pu
 
 	args, err := plugin.UnmarshalProperties(
 		req.GetArgs(), plugin.MarshalOptions{
-			Label:         label,
+			Label: label,
+			// RejectUnknowns: true  //  should we just fail fast here if we see unknowns?
```
mikhailshilkov, line 566: I'd expect to reject unknowns here but I don't have a holistic picture of this option, unfotunately

t0yv0: This option simply scans the data structure and if it finds any unknowns it throws an exception. Something like this:

```
if hasUnknowns(props) {
   return fmt.Errorf("Unexpected unknown property")
} else {
   return nil
}
```

@mikhailshilkov that sounded like a decent option to me also, but now it seems that with my other fix to Marshal/Unmarshal functions landing on master, I cannot repro the bug... That is, unknowns get property filtered out and do not cause panic. Moreover, when Pulumi resolves them to known values, the example works - and passing Output to RPC function works, contrary to wha the team believes. I'm therefore hesitating to put this fix in, as it would break seemingly working examples.

sdk/go/common/resource/plugin/provider_plugin.go
```diff
@@ -1172,6 +1172,7 @@ func (p *provider) Invoke(tok tokens.ModuleMember, args resource.PropertyMap) (r
 		Label:         fmt.Sprintf("%s.args", label),
 		KeepSecrets:   p.acceptSecrets,
 		KeepResources: p.acceptResources,
+		KeepUnknowns:  true,
```
EvanBoyle, line 1175: Talking out loud here, does this mean that all provider invokes must be written in a way that understands unknowns? Is that currently the case? Should this be configurable by provider? Might be good to try this patch out with some invokes non-terraform based providers like kubernetes and azure-native.

EvanBoyle: Thinking specifically about the behavior in previews.

t0yv0: Thank you - this is what I'm trying to figure out 🙂  Should provider impl be shielded from unknowns completely, is that something we're trying to maintain? If I had confirmation here it would be easier to make decisions on the concrete code things.

In the concrete layer...  this patch is now a bit outdated as I can't trigger this path anymore. What happens is this chain:

```
pulumi-user-program --> resmon.Invoke --> provider.Invoke
```

resmon.Invoke currently does KeepUnknowns: true but provider.Invoke does KeepUnknowns: false. It's inconsistent but it only would matter if these functions actually get passed unknowns.

There was a bug in how KeepUnkonwn: false behaved that I fixed
https://github.com/pulumi/pulumi/pull/7214/files

(The function should have been better named "filterOutUnknowns").

So I think right now it works by provider.Invoke filtering out unknowns and not resmon.Invoke filtering out unknowns. If this is the right division of responsibilities we can keep it as is.

The other option mentioned in this PR was RejectUnknowns: true in resmon.Invoke. This would fail with an exception instead of filtering them out.

t0yv0:
Thanks a ton @EvanBoyle for talking it through today. Based on our conversation, closing this PR and opening two simpler / more targeted issues to track follow-up:

- https://github.com/pulumi/pulumi/issues/7272 
- https://github.com/pulumi/pulumi/issues/7273
-

PR #7228, opened 2021-06-07
bump GHA workflows to node 16.x

EvanBoyle:
Draft PR to run our test suite against node 16.x.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -28,7 +28,7 @@ jobs:
         go-version: [ 1.16.x ]
         python-version: [ 3.9.x ]
         dotnet-version: [ 3.1.x ]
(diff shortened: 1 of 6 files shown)


pgavlin:
@emiliza should we take this?

EvanBoyle:
I cannot remember why I did this, but it was a part of investigating some failure/bug report. FWIW node 16 is now active LTS as of october https://nodejs.org/en/about/releases/

emiliza:
Let's close this for now - we openly support Maintenance versions of node, so we should still test this on min-spec (though looking at that chart, perhaps we should've been pinned to 12!)

We can revisit this as part of https://github.com/pulumi/pulumi/issues/6110

PR #7058, opened 2021-05-14
Omit non-primitive provider properties from code-gen

jen20:
# Description

Following pulumi/pulumi-terraform-bridge#347, properties are generated for all provider config matching the inputs. Unfortunately this does not work for complex values generally (not only in bridged providers) since values are JSON serialized.

While a proper solution to this is designed, it's sufficient for now to stop generating non-primitive properties, which this commit does.

## Effect

Following pulumi/pulumi-terraform-bridge#347, the following schema diff is produced for `pulumi-aws`:

```diff
diff --git a/provider/cmd/pulumi-resource-aws/schema.json b/provider/cmd/pulumi-resource-aws/schema.json
index 4b9553258..72bdfca93 100644
--- a/provider/cmd/pulumi-resource-aws/schema.json
+++ b/provider/cmd/pulumi-resource-aws/schema.json
@@ -148781,6 +148781,93 @@
     },
     "provider": {
         "description": "The provider type for the aws package. By default, resources use package-wide configuration\nsettings, however an explicit `Provider` instance may be created and passed during resource\nconstruction to achieve fine-grained programmatic control over provider settings. See the\n[documentation](https://www.pulumi.com/docs/reference/programming-model/#providers) for more information.\n",
+        "properties": {
+            "accessKey": {
+                "type": "string",
+                "description": "The access key for API operations. You can retrieve this from the 'Security \u0026 Credentials' section of the AWS console.\n"
+            },
+            "allowedAccountIds": {
+                "type": "array",
+                "items": {
+                    "type": "string"
+                }
+            },
+            "assumeRole": {
+                "$ref": "#/types/aws:index/ProviderAssumeRole:ProviderAssumeRole"
+            },
+            "defaultTags": {
+                "$ref": "#/types/aws:index/ProviderDefaultTags:ProviderDefaultTags",
+                "description": "Configuration block with settings to default resource tags across all resources.\n"
+            },
+            "endpoints": {
+                "type": "array",
+                "items": {
+                    "$ref": "#/types/aws:index/ProviderEndpoint:ProviderEndpoint"
+                }
+            },
+            "forbiddenAccountIds": {
+                "type": "array",
+                "items": {
+                    "type": "string"
+                }
+            },
+            "ignoreTags": {
+                "$ref": "#/types/aws:index/ProviderIgnoreTags:ProviderIgnoreTags",
+                "description": "Configuration block with settings to ignore resource tags across all resources.\n"
+            },
+            "insecure": {
+                "type": "boolean",
[...]

PR #7230, opened 2021-06-07
Include props which have underlying strings on Go SDK providers

jen20:
## Description

This commit modifies the work in #7058 to permit properties which do not pass the test of being strings directly, but which have an underlying type of string.

When applied to `pulumi-aws`, this results in the following diff:

```diff
diff --git a/sdk/go/aws/provider.go b/sdk/go/aws/provider.go
index c32ad2367..8b4c9fd0a 100644
--- a/sdk/go/aws/provider.go
+++ b/sdk/go/aws/provider.go
@@ -21,6 +21,8 @@ type Provider struct {
 	AccessKey pulumi.StringPtrOutput `pulumi:"accessKey"`
 	// The profile for API operations. If not set, the default profile created with `aws configure` will be used.
 	Profile pulumi.StringPtrOutput `pulumi:"profile"`
+	// The region where AWS operations will take place. Examples are us-east-1, us-west-2, etc.
+	Region pulumi.StringPtrOutput `pulumi:"region"`
 	// The secret key for API operations. You can retrieve this from the 'Security & Credentials' section of the AWS console.
 	SecretKey pulumi.StringPtrOutput `pulumi:"secretKey"`
 	// The path to the shared credentials file. If not set this defaults to ~/.aws/credentials.
```

The primary purpose this is desirable is to expose Region from instances of the AWS provider.

## Checklist

Code-gen diff included in PR text.

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -3,6 +3,8 @@
 - [codegen] - Encrypt input args for secret properties.
   [#7128](https://github.com/pulumi/pulumi/pull/7128)
 
+- [codegen] - Include properties with an underlying type of string on Go provider instances.
+
 ### Bug Fixes
 
 - [cli] - Send plugin install output to stderr, so that it doesn't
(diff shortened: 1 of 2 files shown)


lukehoban (approved):
LGTM.

I'm not entirely sure why this is typed as `string` (instead of `Region`) in the Go (and .NET and Python) codegen - but it seems that it is already `string` on the input side for those SDKS - so having the output also be `string` is consistent.

PR #7239, opened 2021-06-08
[codegen/python] Update Python codegen to use a non-conflicting property name

stack72:
Fixes: https://github.com/pulumi/pulumi-aws/issues/1519

There is a realistic change that a mapped property is called `resource_name` and
this would conflict with the internal property `resource_name`.

Mypy reports an sdk error here:

```
dk/python/pulumi_aws/servicecatalog/tag_option_resource_association.py:277: error: Duplicate argument 'resource_name' in function definition
Found 1 error in 1 file (errors prevented further checking)
```

It's better to move the python internal property to something that
won't conflict, `resource_name_` in this case

# Description

Fixes # (issue)

## Checklist

### DIFF
--- a/pkg/codegen/python/gen.go
+++ b/pkg/codegen/python/gen.go
@@ -1193,7 +1193,7 @@ func (mod *modContext) genResource(res *schema.Resource) (string, error) {
 
 		if hasStateInputs {
 			for _, prop := range res.StateInputs.Properties {
-				pname := PyName(prop.Name)
+				pname := InitParamName(prop.Name)
 				ty := mod.typeString(prop.Type, true, true, true, true /*optional*/, true /*acceptMapping*/)
 				fmt.Fprintf(w, ",\n            %s: %s = None", pname, ty)
 			}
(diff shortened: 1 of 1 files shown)


justinvp (changes requested):
This is not the right fix. A blanket rename of `resource_name` to `resource_name_` is a breaking change for everyone because the parameter can be passed by name (e.g. `resource_name="foo"`).

It seems like a better fix would be adopting the same fix as in https://github.com/pulumi/pulumi/pull/5052 for `get` parameters (using `InitParamName` for the parameters to `get`).

pkg/codegen/python/gen.go
```diff
@@ -1193,7 +1193,7 @@ func (mod *modContext) genResource(res *schema.Resource) (string, error) {
 
 		if hasStateInputs {
 			for _, prop := range res.StateInputs.Properties {
-				pname := PyName(prop.Name)
+				pname := InitParamName(prop.Name)
```
justinvp, line 1196: I think we also need to use `InitParamName` when assigning the param to `__props__.__dict__` below.

pkg/codegen/python/gen.go
```diff
@@ -1215,7 +1215,7 @@ func (mod *modContext) genResource(res *schema.Resource) (string, error) {
 		if res.StateInputs != nil {
 			for _, prop := range res.StateInputs.Properties {
 				stateInputs.Add(prop.Name)
-				fmt.Fprintf(w, "        __props__.__dict__[%[1]q] = %[1]s\n", PyName(prop.Name))
+				fmt.Fprintf(w, "        __props__.__dict__[%[1]q] = %[1]s\n", InitParamName(prop.Name))
```
justinvp, line 1218:
```suggestion
				fmt.Fprintf(w, "        __props__.__dict__[%q] = %s\n", PyName(prop.Name), InitParamName(prop.Name))
```

justinvp (approved):

PR #7246, opened 2021-06-08
[sdk/dotnet] Allow automation api consumer to provide custom logger implementation for inline program

### LINKED ISSUE #7116: [dotnet/sdk] Support Serilog logger factory with inline programs
Following up from #7095, Serilog logger instances are now per `Deployment`, and the effort to enable [Serilog configuration](https://github.com/serilog/serilog/wiki/Configuration-Basics) in [Deployment.InitSerilogger](https://github.com/pulumi/pulumi/blob/cc8459b2d903468b3f283064648fb9a641ece0c9/sdk/dotnet/Pulumi/Deployment/Deployment.cs#L180-L182) works as expected; I can for example configure Serilog from my generic host in `launchSettings.json`:
```json
{
  "profiles": {
    "Development": {
      "commandName": "Project",
      "commandLineArgs": "deploy alpha aws --diff",
      "environmentVariables": {
        "DOTNET_Environment": "Development",
        "DOTNET_HostBuilder__ReloadConfigOnChange": "false",
        "PULUMI_AUTOMATION_API_SKIP_VERSION_CHECK": "true",
        "Serilog__MinimumLevel__Override__Pulumi": "Verbose",
        "Serilog__Enrich__0__Name": "With",
        "Serilog__Enrich__0__Args__enricher": "Pharos.Gemini.SourceContextUqnEnricher, gemini",
        "Serilog__WriteTo__0__Name": "Console",
        "Serilog__WriteTo__0__Args__outputTemplate": "[{Timestamp:HH:mm:ss.fff} {Level:u3}] ({SourceContextUqn}) {Message:lj}{NewLine}{Exception}",
        "Serilog__WriteTo__1__Name": "Seq",
        "Serilog__WriteTo__1__Args__serverUrl": "http://localhost:5341"
      }
    }
  }
}
```

However, there are some caveats to this approach: environment variable configuration sources are clumsy to use and effectively yield global variables in the host application and console output, although customizable, is duplicated if specified (as above).

This approach falls short of what would be _really_ useful: invert control of the Serilog logger factory so that the host of inline programs can _fully_ configure Serilog as required. This would enable reusing the same configuration sources already used by the host such as existing code or yaml configuration files. Furthermore, the output could be captured and displayed as unit / integration tests run (for inline programs only)!

orionstudt:
# Description

Fixes #7116

Allows consumer of dotnet Automation API to provide a custom deployment logger implementation so that Automation API logs can be sent to whatever logging provider the consumer uses with whatever config they desire.

## Checklist

### DIFF
--- a/sdk/dotnet/Pulumi.Automation/PreviewOptions.cs
+++ b/sdk/dotnet/Pulumi.Automation/PreviewOptions.cs
@@ -18,5 +18,7 @@ namespace Pulumi.Automation
         public bool? TargetDependents { get; set; }
 
         public PulumiFn? Program { get; set; }
+
+        public IDeploymentLogger? Logger { get; set; }
     }
 }
--- a/sdk/dotnet/Pulumi.Automation/PreviewOptions.cs
+++ b/sdk/dotnet/Pulumi.Automation/PreviewOptions.cs
@@ -19,6 +19,10 @@ namespace Pulumi.Automation
 
         public PulumiFn? Program { get; set; }
 
+        /// <summary>
+        /// A custom logger implementation that will be used for the action. Note that it will only be used
+        /// if <see cref="Program"/> is also provided.
+        /// </summary>
         public IDeploymentLogger? Logger { get; set; }
     }
 }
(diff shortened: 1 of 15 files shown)


infin8x:
Ah ok, thanks for clarifying @orionstudt!

PR #7247, opened 2021-06-08
Watch Command - Added `--path` Argument

### LINKED ISSUE #6620: Add ability to ignore files/directories within the project folder when using the `watch` command
Continuing from https://github.com/pulumi/pulumi/issues/6537, now I have a new question.

Thanks to @pierskarsenbarg, we found what was causing the initial issue.

On my system, I was using the `local` backend, and as I was deploying new stuff, state files (located in `.pulumi` folder) would change, and the watch command would pick up those changes too, hence infinite re-deploys.

The question that I now have is...

Is it possible to tell the `watch` command to ignore the `.pulumi` folder somehow? 

So, only track changes from all folders / files that are located within the Pulumi project, except the folder in which the state files are being saved.

I believe that would resolve the original issue, linked above.

Thanks in advance! 🥂

doitadrian:
# Description

Closes https://github.com/pulumi/pulumi/issues/6620

As mentioned in the linked issue, currently, the `pulumi watch` command watches the entire Pulumi project directory (recursively). With this PR, developers will be able to specify the exact paths that need to be watched.

The main motivation for this came from the fact that in our project, the application code (so, not infrastructure code) folder is located within the Pulumi project folder. Now, this wouldn't be a problem if we didn't bundle our application code with Webpack.  

This means that, while writing application code, the re-deploy process gets triggered by the `watch` command twice:

1. once for the change that the developer made in the source application code
2. once after Webpack has actually built the change, and stored the new code in a separate `code/build` folder (so, also stored within the Pulumi project folder)

For reference, I'm pasting an image I pasted in the original issue:
image

In our specific case, ideally, we'd only want Pulumi to process a re-deployment once a change has been detected in `code/build` and  `pulumi` (this is where cloud infra code is located) folders.

So, with this PR, developers will be able to run the following:

```bash
pulumi watch --path code/build --path pulumi
```

If no `--path` arguments were passed, the `watch` command acts as usual - watches the whole Pulumi project folder.

Fixes #6620

## Checklist

### DIFF
--- a/pkg/backend/backend.go
+++ b/pkg/backend/backend.go
@@ -164,7 +164,7 @@ type Backend interface {
 	// Destroy destroys all of this stack's resources.
 	Destroy(ctx context.Context, stack Stack, op UpdateOperation) (engine.ResourceChanges, result.Result)
 	// Watch watches the project's working directory for changes and automatically updates the active stack.
-	Watch(ctx context.Context, stack Stack, op UpdateOperation) result.Result
+	Watch(ctx context.Context, stack Stack, op UpdateOperation, paths []string) result.Result
 
 	// Query against the resource outputs in a stack's state checkpoint.
 	Query(ctx context.Context, op QueryOperation) result.Result
--- a/pkg/backend/backend.go
+++ b/pkg/backend/backend.go
@@ -245,6 +245,9 @@ type UpdateOptions struct {
 	AutoApprove bool
 	// SkipPreview, when true, causes the preview step to be skipped.
 	SkipPreview bool
+
+	// WatchPaths, allows specifying paths that need to be watched (instead of the whole project).
+	WatchPaths []string
 }
 
 // QueryOptions configures a query to operate against a backend and the engine.
--- a/pkg/backend/backend.go
+++ b/pkg/backend/backend.go
@@ -245,9 +245,6 @@ type UpdateOptions struct {
 	AutoApprove bool
 	// SkipPreview, when true, causes the preview step to be skipped.
 	SkipPreview bool
-
-	// WatchPaths, allows specifying paths that need to be watched (instead of the whole project).
-	WatchPaths []string
 }
 
 // QueryOptions configures a query to operate against a backend and the engine.
--- a/pkg/backend/watch.go
+++ b/pkg/backend/watch.go
@@ -69,9 +69,19 @@ func Watch(ctx context.Context, b Backend, stack Stack, op UpdateOperation, appl
 	}()
 
 	events := make(chan notify.EventInfo, 1)
-	if err := notify.Watch(path.Join(op.Root, "..."), events, notify.All); err != nil {
-		return result.FromError(err)
+
+	if len(op.Opts.WatchPaths) > 0 {
+		for _, element := range op.Opts.WatchPaths {
+			if err := notify.Watch(path.Join(op.Root, element, "..."), events, notify.All); err != nil {
+				return result.FromError(err)
+			}
+		}
+	} else {
+		if err := notify.Watch(path.Join(op.Root, "..."), events, notify.All); err != nil {
+			return result.FromError(err)
+		}
 	}
+
 	defer notify.Stop(events)
 
 	fmt.Printf(op.Opts.Display.Color.Colorize(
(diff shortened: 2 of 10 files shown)


doitadrian:
Although this works (tested manually), there are a couple of things that could probably be checked.

1. There weren't any tests for the `watch` command in the repository. I think I'm gonna need help with those, not sure how we'd want to test that. Is there any particular approach we might want to use here?

2. Note that I added an additional `WatchPaths` property to the `UpdateOperation.Opts` struct. That seemed the most appropriate place for me (out of other options that I managed to see). But I have a feeling that this might not belong here. Should we do this differently?

Sorry if my code wasn't too sharp, this is actually my first time working with Golang 😅 .

Thanks!

pkg/backend/backend.go
```diff
@@ -245,6 +245,9 @@ type UpdateOptions struct {
 	AutoApprove bool
 	// SkipPreview, when true, causes the preview step to be skipped.
 	SkipPreview bool
+
+	// WatchPaths, allows specifying paths that need to be watched (instead of the whole project).
+	WatchPaths []string
```
lukehoban, line 250: Instead of putting this on `UpdateOption`, let's put this as a separate argument to `Watch` on the backend interface.  This option isn't relevant to most updates, and is really not about the update part of the watch operation at all - so will be easier to keep it watch-specific.

pkg/backend/watch.go
```diff
@@ -67,9 +67,19 @@ func Watch(ctx context.Context, b Backend, stack Stack, op UpdateOperation, appl
...
+			if err := notify.Watch(path.Join(op.Root, element, "..."), events, notify.All); err != nil {
+				return result.FromError(err)
+			}
+		}
+	} else {
+		if err := notify.Watch(path.Join(op.Root, "..."), events, notify.All); err != nil {
+			return result.FromError(err)
+		}
 	}
+
```
lukehoban, line 82: We could probably unify these two branches, and just have the default value (either provided here, or by the CLI code) be `[]string{""}`, so that the code in the first path works in all cases.

lukehoban (commented):
Two suggested code refactorings - but the feature addition LGTM.  Thanks! :tada:

pkg/backend/watch.go
doitadrian: Sounds good!

pkg/backend/backend.go
doitadrian: Yeah, I doubted this was the right place to put this in the first place, but said whatever, we can discuss this along the way.

All in all sounds good, will take a look!

pkg/backend/watch.go
doitadrian: I added the `[]string{""}` you proposed and unified the mentioned branching.

pkg/backend/backend.go
doitadrian: I added `paths []string` as the last parameter on the `Watch` function.

Had to add it in multiple places. Not sure if I got them all right, but it seems it's good. 

Can you please double-check?

Thank you!

doitadrian:
@lukehoban I checked both, and I believe this might be it. 

I'd be glad to add further amends if needed.

pkg/cmd/pulumi/watch.go
```diff
@@ -143,6 +145,9 @@ func newWatchCmd() *cobra.Command {
 		}),
 	}
 
+	cmd.PersistentFlags().StringArrayVarP(
+		&pathArray, "path", "", []string{""},
```
lukehoban, line 149: One additional question as I looked deeper at this: What exactly are the allowed paths?  Can they be files or only folders? (it looks like the implementation only supports folders?).  Should we support files as well?  Can they be relative or absolute?  (it looks like the implementation only supports relative folders?). I'm okay supporting only a subset of these initially - but if so, we should be clear on what is supported in the help text.  Ideally though, we would support relative/absolute and files/folders.

doitadrian: Checked all of your comments - all sounds good.

Initially, relative folders/files were supported, but now we also support absolute folders/files.

I updated the command descriptions, so now it should be clear to the user.

Hope it's fine but let me know if further amends are needed.

Thanks! 🍻

doitadrian: Just wanted to let you know that I'm patiently waiting for an additional review here.

I'm sure you're all super busy, so, whenever you get a chance to check this, it's cool. 🙂 

Thank you! 🍺

lukehoban (approved):

PR #7248, opened 2021-06-08
Add trace proxying to fix sub-process trace collection into files

### LINKED ISSUE #7109: Appdash tracing not at parity with Zipkin tracing, loses sub-process traces
Currently documented approach to tracing is losing data from sub-process traces. The workaround is to use Zipkin-based tracing, that seems to be undocumented (perhaps intentionally?).

I have added a proposed PR to the docs repo that documents using Zipkin-based tracing to a local Jaeger instance. 

https://github.com/pulumi/pulumi-hugo/pull/266

## Expected behavior

Passing `--tracing=file:./up.trace` collects correlated data from CLI and plugins.

## Current behavior

Passing `--tracing=file:./up.trace` collects correlated data from CLI only.

## Steps to reproduce

Run `pulumi up --tracing=file:./up.trace` and observe missing traces from the language and resource plugins.

https://www.pulumi.com/docs/troubleshooting/#performance describes the trace viewing process.

t0yv0:
Fixes for sub-process trace collection and Go language plugin tracing.

# Description

- proxy app dash server to collect sub-process traces
- for go language plugin, fix trace command line propagation
- for go language plugin, add an example sub-trace on subprocess Exec
- verify that file-based tracing now captures that example sub-trace with a test

Fixes #7109 

Side note: I noticed that we're killing plugins with SIGKILL and the subprocess cannot capture that, therefore `main` finalizers do not really run, in particular, it's impossible to run `defer CloseTracing()` to close the global span in a plugin. That is probably something we had to live with. Finalizers can go inside gRPC endpoints or gRPC interceptors instead.

## Checklist


t0yv0:
Didn't quite understand the CI failures, let's try again after merging master.

sdk/go/pulumi-language-go/main.go
```diff
@@ -198,10 +199,21 @@ func (host *goLanguageHost) GetRequiredPlugins(ctx context.Context,
 		return nil, err
 	}
 
+	args := []string{"list", "-m", "-json", "-mod=mod", "all"}
+
+	tracingSpan, _ := opentracing.StartSpanFromContext(ctx,
+		fmt.Sprintf("%s %s", gobin, strings.Join(args, " ")),
+		opentracing.Tag{Key: "component", Value: "exec.Command"},
+		opentracing.Tag{Key: "command", Value: gobin},
+		opentracing.Tag{Key: "args", Value: args})
```
lukehoban, line 208: Possibly for later - but I wonder if we should have a generic wrapper over `exec.Command` that does this tracing - and use that everywhere we do process spawns?

tests/integration/integration_go_test.go
```diff
@@ -586,3 +589,61 @@ func TestComponentProviderSchemaGo(t *testing.T) {
...
+		Tracing:                fmt.Sprintf("file:./pulumi.trace"),
+		PreviewCompletedHook:   searchForGoListTrace,
+	}
+
+	integration.ProgramTest(t, opts)
+
+	if foundTrace == nil {
+		t.Errorf("Did not find a trace for `go list -m -json -mod=mod all` command")
+	}
+}
```
lukehoban, line 649: Awesome to have test coverage on this tracking Infrastructure.

lukehoban (commented):
🎉

sdk/go/pulumi-language-go/main.go
t0yv0: I like that idea https://github.com/pulumi/pulumi/issues/7311 put it in the backlog for the moment.

PR #7254, opened 2021-06-09
Ensure `Output.from_input({})` returns `{}` instead of `[]`

### LINKED ISSUE #7252: [sdk/python] `Output.from_input({})` regression - returns `[]` instead of `{}`
The changes in https://github.com/pulumi/pulumi/pull/7175 introduced a regression where `Output.from_input({})` was changed to return `[]` instead of `{}`.  This was not caught by any existing tests, but triggered failures in `pulumi_kuberentes` as seen in https://github.com/pulumi/pulumi-kubernetes/issues/1602.

lukehoban:
Fixes #7252.

Subtly, `Output.all(**{})` by design returns `Output([])`, because the `all` function itself cannot distinguish between this and `Output.all()`.  So we must special case empty dicts to avoid calling `all` on them.

We were missing tests for empty arrays and dicts, so these have been added.

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -5,4 +5,4 @@
 
 ### Bug Fixes
 
-
+- [sdk/python] - Fix regression in behaviour for `Output.from_input({})`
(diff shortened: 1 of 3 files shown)


justinvp (approved):

lukehoban:
The test failure appears to be unrelated - tracking that in https://github.com/pulumi/pulumi/issues/7255.

PR #7262, opened 2021-06-09
Seal internal members of Resource to prevent accidental inherit/override

t0yv0:
Seal members of Resource class to prevent accidental inherit+override bugs as in https://github.com/pulumi/examples/issues/1017

I expect this is incomplete as we'd need to modify codegen also to rename things like `Urn` property correct?

Is there an easy way to trigger tests of downstream codegen impacts, like I sometimes see automatically happen on codegen PRs..

# Description

Fixes # (issue)

## Checklist


t0yv0:
Hm, that's not it. Trying to remember how C# worked. So sounds like properties are non-overridable already. So that is not what's happening in the bug.

Issue #7112, opened 2021-05-21
Error on BranchProtection resource


kateetrant:
## Expected behavior
pulumi imports BranchProtection resource after retrieval without error

## Current behavior
Errors on import. Subsequent import fails on preview as it is already imported

## Steps to reproduce

1. run ` pulumi import github:index/branchProtection:BranchProtection repo.repo repo.repo:development`
2. agree to import

*Stack Trace and relevant info*
```
Pulumi Version:   v3.3.0
Go Version:       go1.16.3
Go Compiler:      gc
Architecture:     amd64
Operating System: darwin
Panic:            invalid Go source code:

package main

import (

"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
"github.com/pulumi/pulumi-github/sdk/v4/go/github"
)
func main() {
pulumi.Run(func(ctx *pulumi.Context) error {
_, err := github.NewBranchProtection(ctx, "repo_repo", &github.BranchProtectionArgs{
AllowsDeletions: pulumi.Bool(false),
AllowsForcePushes: pulumi.Bool(false),
EnforceAdmins: pulumi.Bool(false),
Pattern: pulumi.String("development"),
PushRestrictions: interface{}{
nil,
nil,
pulumi.String("MDQ6VGVhbTQwMjg0ODk="),
},
RepositoryId: pulumi.String("MDEwOlJlcG9zaXRvcnkyMTg4MzY5NDY="),
RequireSignedCommits: pulumi.Bool(false),
}, pulumi.Protect(true))
if err != nil {
return err
}
return nil
})
}


goroutine 1 [running]:
runtime/debug.Stack(0xc002212a88, 0x56d7840, 0xc000d0bd28)
	/usr/local/Cellar/go/1.16.3/libexec/src/runtime/debug/stack.go:24 +0x9f
main.panicHandler()
	/private/tmp/pulumi-20210520-6659-1j9ek9d/pkg/cmd/pulumi/main.go:29 +0x76
panic(0x56d7840, 0xc000d0bd28)
	/usr/local/Cellar/go/1.16.3/libexec/src/runtime/panic.go:965 +0x1b9
github.com/pulumi/pulumi/pkg/v3/codegen/go.GenerateProgram(0xc000ec1800, 0x691f620, 0xc001125ce0, 0x53a15c0, 0xc002cbada0, 0x1, 0x1)
[...]

PR #7265, opened 2021-06-10
[cli/import] - Handle panic during import codegen

komalali:
# Description

This PR is to handle a panic that is caused by a failure during go codegen when using the `pulumi import` command. Unfortunately, `import` is impossible to write tests for since it is dependent on a resource existing in a provider. The [original PR](https://github.com/pulumi/pulumi/pull/4765) to add the import command included no tests.

I have also been unable to repro this particular failure manually since I do not have the correct access to use the `PushRestrictions` feature for `BranchProtection` in Github as I do not have an organization using the team or enterprise tiers.

I tested this change by manually adding a panic to the `generateImportedDefinitions` function and observing the resulting output. 

Related to: #7112 

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -8,7 +8,7 @@
 
 - [codegen] - Include properties with an underlying type of string on Go provider instances.
 
-- [cli] - Provide a more useful error instead of panicking when codegen fails during import.
+- [cli] - Provide a more helpful error instead of panicking when codegen fails during import.
   [#7265](https://github.com/pulumi/pulumi/pull/7265)
 
 ### Bug Fixes
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -8,6 +8,9 @@
 
 - [codegen] - Include properties with an underlying type of string on Go provider instances.
 
+- [cli] - Provide a more useful error instead of panicking when codegen fails during import.
+  [#7265](https://github.com/pulumi/pulumi/pull/7265)
+
 ### Bug Fixes
 
 - [sdk/python] - Fix regression in behaviour for `Output.from_input({})`
--- a/pkg/cmd/pulumi/import.go
+++ b/pkg/cmd/pulumi/import.go
@@ -198,6 +198,20 @@ func generateImportedDefinitions(out io.Writer, stackName tokens.QName, projectN
 	snap *deploy.Snapshot, programGenerator programGeneratorFunc, names importer.NameTable,
 	imports []deploy.Import, protectResources bool) (bool, error) {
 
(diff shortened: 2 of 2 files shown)


CHANGELOG_PENDING.md
```diff
@@ -8,6 +8,9 @@
 
 - [codegen] - Include properties with an underlying type of string on Go provider instances.
 
+- [cli] - Provide a more useful error instead of panicking when codegen fails during import.
```
lukehoban, line 11:
```suggestion
- [cli] - Provide a more helpful error instead of panicking when codegen fails during import.
```

Minor: this can probably be in the "bug fixes" category?

lukehoban (approved):
> handle a panic that is caused by a failure during go codegen

Is there a reason we can't have codegen return a real error instead of panicng?  It wasn't entirely clear here what call in the code here is panicing, and why that is expected such that we need to add a panic handler at this layer?

Assuming there is a good need to still be panicing in the codegen, this change certainly looks like it helps improve the UX for `pulumi import`.

> Unfortunately, import is impossible to write tests for since it is dependent on a resource existing in a provider. The original PR to add the import command included no tests.

Technically I believe we could import a resource like `random.RandomString` which doesn't have any backing cloud provider state (or have a "dummy" resource provider that we test against).  But I agree we can tackle test coverage here independently of this PR.

CHANGELOG_PENDING.md
komalali: Heh, I put it in improvements because I felt like we aren't actually fixing the bug, but rather improving the error reported by the bug. But yeah, could go either way I suppose :)

komalali:
> It wasn't entirely clear here what call in the code here is panicing, and why that is expected such that we need to add a panic handler at this layer? 

So, this is the line that panics: https://github.com/pulumi/pulumi/blob/3e3e4f2deb7a35fbf069b82e35242e39272c6b02/pkg/codegen/go/gen_program.go#L92

You're right though, it should just return an error. The reason that I didn't fix it at that layer is because [docgen has existing code](https://github.com/pulumi/pulumi-terraform-bridge/blob/18062f55f11271a7bb298b2de1a335a7cce8fd46/pkg/tfgen/docs.go#L1084-L1163) to work around this, and I wanted to avoid working in multiple layers and potentially alter the behavior of existing tooling to get to resolution here.

The reason we're getting to this panic, in this case, is not at the codegen layer, it's something that happens higher up the chain (possibly with the readResource call and the state generated) but I have been unable to investigate that since I don't have the correct access to this resource to observe the panic myself.

In any case, I was trying to resolve the panic without diving into multiple other layers when I can't even repro the error.

EvanBoyle (approved):

PR #7279, opened 2021-06-11
Tracing enhancements for CLI perf metrics capture

t0yv0:
I'm finding I need a few more hooks to enable trace collections easily in the integration tests. Everything in this PR is opt-in. If you don't set the magic env vars nothing should be affected for our regular users.

# Description

- Allows to run ProgramTest with `Tracing: "file:./pulumi-{command}.trace` resulting in multiple files depending on which command is being traced.

- Allows to set PULUMI_TRCING_TAG_FOO env vars to add tags to the top-level trace span

- Allows to set PULUMI_TRACING_MEMSTATS_POLL_INTERVAL=100ms to poll runtime.ReadMemStats and record in the trace

- Records os, arch, num_cpu in the trace

Fixes # (issue)

## Checklist

### DIFF
--- a/pkg/testing/integration/program.go
+++ b/pkg/testing/integration/program.go
@@ -222,7 +222,7 @@ type ProgramTestOptions struct {
 	//
 	// Template `{command}` syntax will be expanded to the current
 	// command name such as `pulumi-stack-rm`. This is useful for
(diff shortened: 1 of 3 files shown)


justinvp (approved):
LGTM

PR #7288, opened 2021-06-14
[sdk/python] Ensure Output objects are not iterable

### LINKED ISSUE #5028: Output in Python allows infinite iteration
**Summary**

Iterating through acm.Certificate.domain_validation_options results in an infinite loop that creates a memory leak that eats all available memory.

**System**
- Ubuntu 19.04
- Python 3.7.4
- Terraform 0.12.25 (probably not relevant)
- Pulumi 2.6.1

**Example code**

```python

import pulumi
from pulumi_aws import ec2, acm, lb, route53

# Configuration
config = pulumi.Config()
MANAGED_DOMAINS = config.require_object('managed_domains')
EXTRA_DOMAINS = config.require_object('extra_domains')
ALL_DOMAINS = EXTRA_DOMAINS + MANAGED_DOMAINS

# Create new zones
zones = {}
for domain in MANAGED_DOMAINS:
  zone = route53.Zone(domain, name=domain)
  zones[domain] = zone

# Create ACM generated certificate
cert = acm.Certificate("cert",
    domain_name=ALL_DOMAINS[0],
    tags={
        "Environment": "test",
    },
    validation_method="DNS",
    subject_alternative_names=ALL_DOMAINS[1:]
)

# Add the validation DNS records to each managed domain
cert_records = []
for record_to_add in cert.domain_validation_options:
  if record_to_add.name not in MANAGED_DOMAINS:
    continue
  record = route53.Record(f"acm-verify-{record_to_add.name}",
      name=record_to_add.resourceRecordName,
      ttl=60,
      type=record_to_add.resourceRecordType,
      records=[record_to_add.resourceRecordValue],
      zone_id=zones[record_to_add.name])
  cert_records += [record]

# Wait for verification of ACM cert
cert_validation = acm.CertificateValidation(
  "cert-validation",
  certificate_arn=cert.arn
)

```

**Debugging**

Theres probably a better way to do this, but for me to debug the out of memory issue, I moved `venv/bin/python3` to `venv/bin/python3.orig` and put the following at `venv/bin/python3`:

```bash
#!/bin/bash

ARGS=( "$@" )

echo "Env" > args_used
env >> args_used
echo "Run" >> args_used
echo "python3.orig -m pdb ${ARGS[@]}" >> args_used
sleep 3600
```

Then:
- run `pulumi preview -d -v 9 --logtostderr`
- switch to second terminal, `cat args_used`
- copy and export `PULUMI_CONFIG` (using single quotes to keep formatting)
- copy and paste the `python3.orig -m pdb ...` command to get a pdb shell 
  - note you can also use `gdb -ex r --args python3.orig ...` or other tools at this point
- set a breakpoint for the line `for record_to_add in cert.domain_validation_options` (line 34 for me):
```
b <path/to/__main__.py:34>
```
- run using command `r` in pdb
- code will break
- interact with code using `interact`
- try the following code in the interpreter:
```python
for foo in cert.domain_validation_options:
    print("Infinite loop")
```
- watch your memory usage grow with top

**Possibly related issues**
Looks like there is a redesign of `aws_acm_certificate` occurring in terraform, one feature of which is that `domain_validation_options` will be calculated during plan.
- https://github.com/terraform-providers/terraform-provider-aws/issues/13053
- https://github.com/terraform-providers/terraform-provider-aws/pull/14199

lukehoban:
Although `Output` objects can never correctly support iteration, Python will see the implementation of `__getitem__` and try to iterate the object, leading to an infinite loop.  To prevent this, we need to explicitly implement `__iter__` and make it return a `TypeError` to prevent iteration (and offer a useful error message).

Fixes #5028.

### DIFF
--- a/sdk/python/lib/test/test_output.py
+++ b/sdk/python/lib/test/test_output.py
@@ -193,3 +193,33 @@ class OutputFromInputTests(unittest.TestCase):
         self.assertIsInstance(x_val, OutputFromInputTests.FooArgs)
         self.assertIsInstance(x_val.nested, OutputFromInputTests.NestedArgs)
         self.assertEqual(x_val.nested.hello, "world")
+
+class Obj:
+    def __init__(self, x: str):
+        self.x = x
+
+class OutputHoistingTests(unittest.TestCase):
+    @pulumi_test
+    async def test_item(self):
+        o = Output.from_input([1,2,3])
+        x = o[0]
+        x_val = await x.future()
+        self.assertEqual(x_val, 1)
+
+    @pulumi_test
+    async def test_attr(self):
+        o = Output.from_input(Obj("hello"))
+        x = o.x
+        x_val = await x.future()
+        self.assertEqual(x_val, "hello")
+
+    @pulumi_test
+    async def test_no_iter(self):
+        x = Output.from_input([1,2,3])
+        errored = False
+        try:
+            for i in x:
+                print(i)
+        except TypeError:
+            errored = True
+        self.assertTrue(errored)    
--- a/sdk/python/lib/test/test_output.py
+++ b/sdk/python/lib/test/test_output.py
@@ -216,10 +216,6 @@ class OutputHoistingTests(unittest.TestCase):
     @pulumi_test
     async def test_no_iter(self):
         x = Output.from_input([1,2,3])
-        errored = False
-        try:
+        with self.assertRaises(TypeError):
             for i in x:
                 print(i)
-        except TypeError:
-            errored = True
-        self.assertTrue(errored)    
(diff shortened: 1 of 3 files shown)


sdk/python/lib/test/test_output.py
```diff
@@ -193,3 +193,33 @@ async def test_unwrap_input_type_nested(self):
...
+    @pulumi_test
+    async def test_no_iter(self):
+        x = Output.from_input([1,2,3])
+        errored = False
+        try:
+            for i in x:
+                print(i)
+        except TypeError:
+            errored = True
+        self.assertTrue(errored)
```
justinvp, line 225: Can this be simplified?

```suggestion
        with self.assertRaises(TypeError):
            for i in x:
                print(i)
```

justinvp (approved):

PR #7295, opened 2021-06-14
[sdk/python] Reduce `log.debug` calls for improved performance

### LINKED ISSUE #7292: [sdk/python] `log.debug` calls in SDK measurably impact program performance
The following Python program profiled with [pyinstrument](https://github.com/joerick/pyinstrument):

```python
from pyinstrument import Profiler
profiler = Profiler()
profiler.start()

from pulumi_aws import s3
bucket = s3.Bucket("mybucket")
for x in range(5000):
    s3.BucketObject(f"{x}", bucket=bucket.id, content="hello")

profiler.stop()
with open("profile.html", 'w') as f:
    f.write(profiler.output_html())
```

Run with `pulumi preview` yields the following:

Screen Shot 2021-06-14 at 6 19 23 AM

We should reduce the amount of `log.debug` calls to improve program performance.

justinvp:
`log.debug` messages are part of an update and get sent to the service. Reducing the number of `log.debug` calls can significantly improve the performance of a program. This commit changes the debug logging in the Python SDK to be similar to the debug logging in the Node.js SDK.

 - An `excessive_debug_output` variable has been added (that can be set to `True` during debugging), but otherwise will avoid a large number of `log.debug` calls.
- Some unhelpful/redundant `log.debug` calls have been deleted.

Fixes #7292

---

The following program run with `pulumi preview` (plus the fix from #7293):

```python
from pulumi_aws import s3
bucket = s3.Bucket("mybucket")
for x in range(5000):
    s3.BucketObject(f"{x}", bucket=bucket.id, content="hello")
```

Duration Before: 35.8 seconds
Duration After: 6.35 seconds

29.45 second improvement!

### Before

Screen Shot 2021-06-14 at 7 31 41 AM

### After

Screen Shot 2021-06-14 at 7 31 57 AM

### DIFF
--- a/sdk/python/lib/pulumi/runtime/resource.py
+++ b/sdk/python/lib/pulumi/runtime/resource.py
@@ -81,7 +81,6 @@ async def prepare_resource(res: 'Resource',
                            opts: Optional['ResourceOptions'],
                            typ: Optional[type] = None) -> ResourceResolverOperations:
     from .. import Output  # pylint: disable=import-outside-toplevel
-    log.debug(f"resource {props} preparing to wait for dependencies")
     # Before we can proceed, all our dependencies must be finished.
     explicit_urn_dependencies = []
     if opts is not None and opts.depends_on is not None:
@@ -154,7 +153,6 @@ async def prepare_resource(res: 'Resource',
         if not alias_val in aliases:
             aliases.append(alias_val)
 
-    log.debug(f"resource {props} prepared")
     return ResourceResolverOperations(
         parent_urn,
         serialized_props,
@@ -311,7 +309,6 @@ def read_resource(res: 'CustomResource',
     # that we are populating the Resource object with properties associated with an already-live resource.
     #
     # Same as below, we initialize the URN property on the resource, which will always be resolved.
-    log.debug("preparing read resource for RPC")
     (resolve_urn, res.__dict__["urn"]) = resource_output(res)
 
     # Furthermore, since resources being Read must always be custom resources (enforced in the
@@ -328,7 +325,6 @@ def read_resource(res: 'CustomResource',
 
     async def do_read():
         try:
-            log.debug(f"preparing read: ty={ty}, name={name}, id={opts.id}")
             resolver = await prepare_resource(res, ty, True, False, props, opts, typ)
 
             # Resolve the ID that we were given. Note that we are explicitly discarding the list of
@@ -419,7 +415,6 @@ def register_resource(res: 'Resource',
     # Simply initialize the URN property and get prepared to resolve it later on.
     # Note: a resource urn will always get a value, and thus the output property
     # for it can always run .apply calls.
-    log.debug("preparing resource for RPC")
     (resolve_urn, res.__dict__["urn"]) = resource_output(res)
 
     # If a custom resource, make room for the ID property.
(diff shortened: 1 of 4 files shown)


sdk/python/lib/pulumi/runtime/resource.py
```diff
@@ -81,7 +81,6 @@ async def prepare_resource(res: 'Resource',
                            opts: Optional['ResourceOptions'],
                            typ: Optional[type] = None) -> ResourceResolverOperations:
     from .. import Output  # pylint: disable=import-outside-toplevel
-    log.debug(f"resource {props} preparing to wait for dependencies")
```
justinvp, line 84: Deleting the `log.debug` calls at the beginning/end of this function, because we don't do it in the Node.js SDK

sdk/python/lib/pulumi/runtime/resource.py
```diff
@@ -311,7 +309,6 @@ def read_resource(res: 'CustomResource',
     # that we are populating the Resource object with properties associated with an already-live resource.
     #
     # Same as below, we initialize the URN property on the resource, which will always be resolved.
-    log.debug("preparing read resource for RPC")
```
justinvp, line 314: Redundant with the earlier `log.debug` call at the beginning of the function.

sdk/python/lib/pulumi/runtime/resource.py
```diff
@@ -328,7 +325,6 @@ def read_resource(res: 'CustomResource',
 
     async def do_read():
         try:
-            log.debug(f"preparing read: ty={ty}, name={name}, id={opts.id}")
```
justinvp, line 331: Doesn't really add much new information beyond what was logged at the beginning of the function, and Node.js doesn't have this additional logging.

sdk/python/lib/pulumi/runtime/resource.py
```diff
@@ -419,7 +415,6 @@ def register_resource(res: 'Resource',
     # Simply initialize the URN property and get prepared to resolve it later on.
     # Note: a resource urn will always get a value, and thus the output property
     # for it can always run .apply calls.
-    log.debug("preparing resource for RPC")
```
justinvp, line 422: Redundant with the `debug.log` call at the beginning of the function.

sdk/python/lib/pulumi/runtime/resource.py
```diff
@@ -434,7 +429,6 @@ def register_resource(res: 'Resource',
 
     async def do_register():
         try:
-            log.debug(f"preparing resource registration: ty={ty}, name={name}")
```
justinvp, line 437: Similarly redundant with the earlier `log.debug` call -- and Node.js doesn't have this additional log call.

sdk/python/lib/pulumi/runtime/settings.py
```diff
@@ -31,6 +31,10 @@
 _GRPC_CHANNEL_OPTIONS = [('grpc.max_receive_message_length', _MAX_RPC_MESSAGE_SIZE)]
 
 
+# excessive_debug_output enables, well, pretty excessive debug output pertaining to resources and properties.
+excessive_debug_output = False
```
justinvp, line 35: This is equivalent to what we have in the Node.js SDK:

https://github.com/pulumi/pulumi/blob/7820d2c17b40ca2345f4215b3a6789744e2784e6/sdk/nodejs/runtime/settings.ts#L32-L35

EvanBoyle: an env var might be easier for users, but this is what we do in node 🤷

EvanBoyle (approved):
🐎

joeduffy:
@justinvp TIL we log synchronously. Makes sense -- don't want to lose messages -- but the same way we coordinate synchronous execution and shutdown with other RPCs, etc, we could do the same here and possibly bank even more.

PR #7297, opened 2021-06-14
[codegen/pcl] Allow missing object properties.

pgavlin:
Add an option to allow missing object properties. This will prevent us
from losing examples once resource typechecking is fixed (it is
currently unintentionally disabled because the resource inputs object
type has an unexpected shape).


pgavlin:
(related to https://github.com/pulumi/pulumi/pull/7059)

lblackstone (approved):

PR #7319, opened 2021-06-16
Add a convert-trace command.

pgavlin:
This command converts an appdash trace into a pprof file for use with
`go tool pprof`. Spans are converted into stacks by sampling each root
span at a given rate and recording the stack of subspans at each sample.

These changes also replace the conditional addition of experimental and
debug commands with conditional visibility. Experimental and debug
commands will always be available, but will be hidden unless the
appropriate environment variables are set.


t0yv0:
Pretty cool, I can use this to learn how to add a command into pulumi CLI also, useful. I'll only get to a proper review on Monday sorry due to PTO.

pgavlin:
Here are Appdash and pprof views for an example trace:

## Appdash

Screen Shot 2021-06-16 at 12 36 03

## pprof

Screen Shot 2021-06-16 at 12 34 52
Screen Shot 2021-06-16 at 12 34 45
Screen Shot 2021-06-16 at 12 34 38

pgavlin:
> I'll only get to a proper review on Monday sorry due to PTO.

No rush!

pkg/cmd/pulumi/convert-trace.go
```diff
@@ -0,0 +1,284 @@
+// Copyright 2016-2018, Pulumi Corporation.
```
lblackstone, line 1: nit: Date
```suggestion
// Copyright 2016-2021, Pulumi Corporation.
```

pkg/cmd/pulumi/convert-trace.go
```diff
@@ -0,0 +1,284 @@
...
+var urnFieldRe = regexp.MustCompile(`urn:"([^"]+)"`)
+var tokFieldRe = regexp.MustCompile(`tok:"([^"]+)"`)
+
+func extractResourceType(t *appdash.Trace) string {
+	grpcPayload := findGRPCPayload(t)
+	matches := urnFieldRe.FindStringSubmatch(grpcPayload)
+	if len(matches) == 0 {
+		return ""
+	}
+	urn := resource.URN(matches[1])
```
lblackstone, line 85: I believe `FindStringSubmatch` will always return 0 or 2 results for the current regexp, but it would probably be good to check for `len(matches) < 2` just to be safe.

pkg/cmd/pulumi/convert-trace.go
```diff
@@ -0,0 +1,284 @@
...
+	return string(urn.Type())
+}
+
+func extractFunctionToken(t *appdash.Trace) string {
+	grpcPayload := findGRPCPayload(t)
+	matches := tokFieldRe.FindStringSubmatch(grpcPayload)
+	if len(matches) == 0 {
+		return ""
+	}
+	return matches[1]
```
lblackstone, line 98: Same here

pkg/cmd/pulumi/convert-trace.go
```diff
@@ -0,0 +1,284 @@
...
+			if err := readTrace(args[0], store); err != nil {
+				return err
+			}
+
+			roots, err := traceRoots(store)
+			if err != nil {
+				return err
+			}
+
+			const quantum = 500 * time.Microsecond
```
lblackstone, line 205: Should this be configurable?

lblackstone (changes requested):
A few minor suggestions, but LGTM overall

pkg/cmd/pulumi/convert-trace.go
pgavlin: This will always return 0 or 2 results--I don't think that we need to sanity check.

pkg/cmd/pulumi/convert-trace.go
pgavlin: That's a great idea; I'll add a knob.

lblackstone (approved):

PR #7323, opened 2021-06-16
[sdk/dotnet] - Fix swallowed nested exceptions with inline program so they correctly bubble to consumer

### LINKED ISSUE #7322: [sdk/dotnet] - Inline Programs swallow exceptions in nested task without indicating error
Somewhat related to #7050

@gitfool pointed out that after the fix for #7050 brought in with #7299 he was still seeing weird behavior. After some discovery:

When an inline program throws an exception from a nested task (such as an `Output<T>`) that exception will be swallowed and the consumer will have no indication of the error. Their call to `.PreviewAsync(..)` or `.UpAsync(..)` will execute without throwing.

## Steps to reproduce

Can be easily reproduced with the following inline program:

```csharp
var outputTask = Task.Run((Func<string>)(() => throw new FileNotFoundException()));
var programWithOutput = PulumiFn.Create(() =>
{
    var output = Output.Create(outputTask);
    {
        ["output"] = output,
    };
});
```

orionstudt:
# Description

Fixes #7322

This PR:
- ensures the exit code properly bubbles up so that `CommandException` would be thrown in the scenario described in #7322
- does additional capturing so that the explicit exception will be thrown for inline consumers, rather than the generic `CommandException`. If there is a single nested exception that was swallowed, it captures that. If there was more than 1 it re-throws as an `AggregateException` containing all of the swallowed exceptions.

## Checklist


orionstudt:
Alright did a little refactor with the latest commit to pull all of the exception complexity out of the inline program `PulumiFn` implementations. All of the exception capturing now happens in a single place in `Deployment.RunInlineAsync` so that it is a bit easier to reason about.

sdk/dotnet/Pulumi.Automation/PulumiFn.cs
```diff
@@ -18,7 +17,7 @@ internal PulumiFn()
         {
         }
 
-        internal abstract Task<ExceptionDispatchInfo?> InvokeAsync(IRunner runner, CancellationToken cancellationToken);
+        internal abstract Task<int> InvokeAsync(IRunner runner, CancellationToken cancellationToken);
```
t0yv0, line 20: What does the int stand for, perhaps a quick comment here? Exit code?

orionstudt: yea it should be the exit code returned from one of the `IRunner.RunAsync(...)` overloads. Depending on the shape of the inline function (generic, service provider, or delegate) it needs to use a different overload so `PulumiFn` needs to handle the actual call to `IRunner`. I can add a quick comment

sdk/dotnet/Pulumi/Deployment/Deployment.Runner.cs
```diff
@@ -29,6 +30,9 @@ private class Runner : IRunner
             /// exiting once the set becomes empty.
             /// </summary>
             private readonly Dictionary<Task, List<string>> _inFlightTasks = new Dictionary<Task, List<string>>();
+            private readonly List<Exception> _exceptions = new List<Exception>();
```
t0yv0, line 33: Hrm. Is this useful? This made me wonder about why we have this entire class in here, and it sounds like at a very high level it's there because "new Resource()" completes in a fire-and-forget manner and the background exceptions need to go somewhere, so presumably they go here? Is lifecycle of Runner pretty contained so we won't have this list growing indefinitely in a server or smth?

sdk/dotnet/Pulumi/Deployment/Deployment_Inline.cs
```diff
@@ -41,20 +42,50 @@ private Deployment(InlineDeploymentSettings settings)
...
-                    exceptionDispatchInfo = await func(runner).ConfigureAwait(false);
-                    return 1;
+                    int? exitCode = null;
+                    try
+                    {
+                        exitCode = await runnerFunc(runner).ConfigureAwait(false);
+
+                        // if there was swallowed exceptions from the in-flight tasks we want to either capture
+                        // if it is single or re-throw as an aggregate exception if there is more than 1
+                        if (runner.SwallowedExceptions.Count == 1)
```
t0yv0, line 60: Aha here we read from this. Thinking.

sdk/dotnet/Pulumi/Deployment/Deployment.Runner.cs
orionstudt: Lifecycle of the `IRunner` is tied to the lifecycle of the `Deployment` [here](https://github.com/pulumi/pulumi/blob/98211718ad2b4466c2efc3b452a4c52b0e468872/sdk/dotnet/Pulumi/Deployment/Deployment_Run.cs#L190), so only for the duration of the linked method and then it will no longer be referenced. So the exception list should _at most_ be the exceptions only from a single deployment before the `IRunner` is de-referenced and hit by GC

t0yv0:
The changes look good to me especially simplifying exception unwrapping.

Note to self: I'd like to dig later and understand why we have this code in the first place. I'm a little uncomfortable about all these errors landing in a kitchen-sink owned by Runner object and especially if somehow Runner is used concurrently by several resources, will not there be confusion in unnecessary waits and possibly mis-attribution of exceptions? It's confusing to me why e.g.  WhileRunningAsync is called from so many methods here instead of say being called once from the CLI in the lifecycle to prevent premature exit. We could also think about grouping background work and exceptions into a property of the resource itself e.g.:

```
var r = new MyResource(); // returns immediately, but starts background work
await r.BackgroundTasks; // waits till all resource-related background work finished; rethrow exception if it failed.

// implicitly, make CLI `await r.BackgroundTasks` also to avoid premature exit
```

In the above notation --^ it's very explicit which background work was scoped to which resource constructor.

orionstudt:
@t0yv0 I 100% agree that the `Runner` implementation needs a refactor. The only reason I've been hesitant to do it is because it heavily impacts both the CLI scenarios and the Automation API scenarios. I believe the way it currently exists is a remnant of the initial CLI SDK implementation.

t0yv0:
Oh, interesting. Are there pulumi issues about it yet, I'll ask around internally. We have a similar highly interesting code across all SDKs that I'm very worried about, perhaps at some point we can spin off a coordinated refactor.

t0yv0:
You're right that we need good test cov on Automation API first and confidence in that so we can attempt these refactors safely...

t0yv0 (approved):

orionstudt:
I'm not sure if there is an existing issue for this specific piece of code - but I agree that there's gotta be a less confusing way to deal with all of those background tasks in both the CLI & Automation API scenarios. It can definitely be error-prone, though, and it's a very important part of the .NET SDK.

My hesitance to dig deeper into `Runner`'s handling of the backgrounds tasks is probably why I missed these swallowed exceptions in the initial Automation API implementation in the first place so.. whoops, lol.

orionstudt:
CLI scenario has the benefit of not needing to throw the explicit exception to the caller, so it was able to rely on just keeping track of whether errors were thrown at all and returning the appropriate error code - which is the purpose of the [this flag here](https://github.com/pulumi/pulumi/blob/98211718ad2b4466c2efc3b452a4c52b0e468872/sdk/dotnet/Pulumi/Deployment/IEngineLogger.cs#L9). So it cared only that those exceptions existed and not that they were swallowed.

orionstudt:
One thing to keep in mind with your proposed change is with a line like this:
```csharp
await r.BackgroundTasks;
```
Is that if there is an exception in those tasks, it will be thrown because we are actually explicitly `await`ing the background tasks now. This is a change of behavior in the CLI scenario, which isn't currently actually throwing those exceptions. So we'll have to keep that in mind when it comes time for a refactor

t0yv0:
Yup. The change may be crafted such as it's a non-change in behavior, but the crux of my idea was giving the user an option to concisely access resource-related background tasks - and await them or harvest their exceptions if the user wants to.

t0yv0:
Anyway, thanks for the work here, as always! I will merge once the tests pass.

PR #7327, opened 2021-06-17
Stop printing value if it looks like a secret

### LINKED ISSUE #7301: config that looks like secret should not be printed
## Steps to reproduce

Actual:

```
$ pulumi config set some_password $verysensible
error: config value 'verysecretpassword' looks like a secret; rerun with --secret to encrypt it, or --plaintext if you meant to store in plaintext
```

Expected

```
$ pulumi config set some_password $verysensible
error: config value for 'some_password' looks like a secret; rerun with --secret to encrypt it, or --plaintext if you meant to store in plaintext
```

If it looks like a secret, please don't print it. Else, that security feature quickly becomes a security vulnerability

JasonWhall:
# Description

Stops printing secret value on `pulumi config set` if it thinks it may be a secret value. Instead, print the config key name

Fixes #7301

## Checklist


justinvp (approved):
LGTM, thanks!

PR #7343, opened 2021-06-21
[sdk/go] Specify known when creating outputs for construct

### LINKED ISSUE #7318: Output as component input leads to unknown value
When providing an `output` into a multi-language component, the value is treated as unknown or undefined and leads to unexpected behavior.

Per @justinvp, these lines appear to be the culprit:

https://github.com/pulumi/pulumi/blob/258fb00bc2ecbd489af6d694a2204468cc7ca729/sdk/go/pulumi/provider.go#L238

https://github.com/pulumi/pulumi/blob/258fb00bc2ecbd489af6d694a2204468cc7ca729/sdk/go/pulumi/provider.go#L188

(Apologies for brevity of issue, lots of detail in a private slack conversation.)

justinvp:
If an input is an unknown value or contains unknowns, specify `false` for `known` when creating the output.

Added integration tests to ensure this is behaving as expected for all languages.

Fixes #7318


tests/integration/integration_test.go
```diff
@@ -695,3 +694,43 @@ func testComponentProviderSchema(t *testing.T, path string, env ...string) {
...
+		{
+			componentDir: "testcomponent-go",
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.componentDir, func(t *testing.T) {
+			pathEnv := pathEnv(t,
+				filepath.Join("..", "testprovider"),
+				filepath.Join(testDir, test.componentDir))
+			integration.ProgramTest(t, &integration.ProgramTestOptions{
```
t0yv0, line 722: Cool. Interesting so this was just bowing up before the change, no need for extra asserts?

t0yv0 (approved):
LGTM. The more I see the one-liner fix with 1,880 lines of per-language tests, the more I worry about our future supporting N languages 🙂

tests/integration/integration_test.go
justinvp: Right. Before the change, in the case of the Go-based provider, the `Apply` inside the provider's component implementation will run because it thinks the value is known, which will panic and fail the test. With the fix, the `Apply` won't run because we properly pass along unknown, so no panic, and the test succeeds. I wasn't really sure if there was a better way to test this from an integration test.

PR #7345, opened 2021-06-21
Fix watch command in macos arm64 builds

komalali:
# Description

https://github.com/pulumi/pulumi/pull/7247 added the ability to specify paths in `pulumi watch`, but this PR didn't update the alternative handling we have for `darwin/arm64` where `watch` isn't supported. This led to a [failure](https://github.com/pulumi/pulumi/runs/2878255663?check_suite_focus=true#step:8:77) in p/p master.

This PR corrects the issue by updating the darwin/arm64 function signature.

## Checklist

### DIFF
--- a/pkg/backend/watch_darwin_arm64.go
+++ b/pkg/backend/watch_darwin_arm64.go
@@ -21,6 +21,6 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/result"
 )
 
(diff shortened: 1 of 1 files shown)


lblackstone (approved):

pgavlin (approved):

komalali:
Test failure is unrelated

PR #7347, opened 2021-06-21
Read passphrase from the terminal when rotating.

pgavlin:
Rotating a passphrase requires that the old passphrase is available via
one of the `PULUMI_CONFIG_PASSPHRASE` or `PULUMI_CONFIG_PASSPHRASE_FILE`
environment variables. This confuses `readPassphrase` when reading a new
passphrase, since that function checks the aforementioned environment
variables prior to reading from the console. The overall effect is that
it is impossible to rotate the passphrase for a stack using the
passphrase provider. These changes fix this by always reading from the
console when rotating a passphrase.


lukehoban (approved):
LGTM. 

Do we have any way to test this?

pgavlin:
> Do we have any way to test this?

Working on an integration test.

Issue #7333, opened 2021-06-17
NodeJS Automation API - detect side by side @pulumi/pulumi and fail early


EvanBoyle:
Addresses the underlying issue behind one of the reports in https://github.com/pulumi/pulumi/issues/6998

When two copies of pulumi/pulumi are loaded in an automation api inline program, there can be problems with global state and the outstanding work queue. We should try to detect this scenario and fail fast.

hzshlomi:
Hi Evan, I saw the patch is storing a random value in the environment and my question is - do you intent to also prohibit same version parallel runs? asking because a previously working program just stopped running throwing the `Detected multiple versions of '@pulumi/pulumi' in use in an inline automation api program` error message.

EvanBoyle:
@hzshlomi parallel same version runs of inline programs are not supported for nodejs at the moment: https://github.com/pulumi/pulumi/issues/5449

There is some global state in the nodejs sdk, and parallel executions can cause strange and undefined behavior leading to  inconsistent and hard to diagnose errors.

hzshlomi:
Thanks for the reply and the pointer. There is yet another usage scenario that is prohibited by this implementation - the case where you want to wrap up common functions in a helper library. the pulumi sdk would be `require`d more than once and complain, is that correct?

EvanBoyle:
> the case where you want to wrap up common functions in a helper library
I believe you should still be able to use the `resolutions` field of `package.json` to force usage of a single pulumi/pulumi. Alternatively, if this is a helper library you are in control of, you can either declare pulumi as a peer dependency, or curry in the pulumi modules (pass the instance of pulumi as a parameter to your helper library/functions). 

If you have an example/minimal repro happy to help get you unblocked here.

hzshlomi:
Thanks again :) i think i will just go ahead and export pulumi/pulumi for the helper lib user to require.. i guess it will do for now

PR #7349, opened 2021-06-22
detect nodejs side by side pulumi for inline automation programs and fail fast

EvanBoyle:
# Description

This change detects when nodejs automation api inline programs load multiple instances of `@pulumi/pulumi` and triggers an early failure with a helpful error message:

```console
      pulumi:pulumi:Stack (inlineNode-dev):
          error: Unhandled exception: Error: Detected multiple versions of '@pulumi/pulumi' in use in an inline automation api program
          Use the yarn 'resolutions' field to resolve this: https://github.com/pulumi/pulumi/issues/5449.
```

Using multiple versions of pulumi/pulumi in inline programs can cause collisions in global state. Until we address https://github.com/pulumi/pulumi/issues/5449, inline node automation users need to pin to a single version of pulumi/pulumi to avoid issues like those seen in https://github.com/pulumi/pulumi/issues/6998

This change set adds a module level global (a random number) that each instance of pulumi/pulumi will write into the environment. We detect side by side scenarios by comparing the module level global to the value in the environment every time we observe settings, or do a resource operation (invoke, register resource, get, etc). 

Fixes https://github.com/pulumi/pulumi/issues/7333
## Checklist

### DIFF
--- a/sdk/nodejs/runtime/settings.ts
+++ b/sdk/nodejs/runtime/settings.ts
@@ -239,6 +239,8 @@ export function hasMonitor(): boolean {
  * getMonitor returns the current resource monitoring service client for RPC communications.
  */
 export function getMonitor(): Object | undefined {
+    // pre-emptive fail fast check for node inline programs
+    runSxSCheck();
     if (monitor === undefined) {
         const addr = options().monitorAddr;
         if (addr) {
@@ -329,6 +331,8 @@ export function serialize(): boolean {
 
  */
 function options(): Options {
+    // pre-emptive fail fast check for node inline programs
+    runSxSCheck();
     // The only option that needs parsing is the parallelism flag.  Ignore any failures.
     let parallel: number | undefined;
     const parallelOpt = process.env[nodeEnvKeys.parallel];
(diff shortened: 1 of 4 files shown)


komalali (approved):
LGTM

sdk/nodejs/runtime/settings.ts
```diff
@@ -545,3 +549,37 @@ export function monitorSupportsSecrets(): Promise<boolean> {
...
+    const envSxS = process.env[pulumiSxSEnv];
+    process.env[pulumiSxSEnv] = sxsRandomIdentifier;
+
+    if (!isInline) {
+        return;
+    }
+
+    // if we see a different identifier, another version of pulumi has been loaded and we should fail.
+    if (!!envSxS && envSxS !== sxsRandomIdentifier) {
+        throw new Error("Detected multiple versions of '@pulumi/pulumi' in use in an inline automation api program.\n" +
```
joeduffy, line 582: Would be good to think about how to possibly link to docs with an example resolution in errors like this.

Issue #7012, opened 2021-05-10
Accept `Output[Resource]` in ~all places where `Resource` is accepted in core SDK


lukehoban:
There are a variety of reasons why users may end up with an `Output[Resource]` instead of a `Resource`, and with multi-lang components, this is now also much more common.  However, in several places in the core SDKs, only a `Resource` is accepted, and thus these APIs are hard to use with multi-lang components.

We should accept `Output[Resource]` in all of the following (unless there is a very good reason we can't), in all language SDKs:
* `parent` resource option
* `provider` resource option
* `providers` resource option
* `dependsOn` resource option (already supports `Output<Resource>` in Node.js)

- Go: https://github.com/pulumi/pulumi/pull/7350
- [ ] TypeScript TODO
- [ ] Python TODO
- [ ] .NET TODO

benesch:
For posterity's sake, #5642 has been tracking the `dependsOn` case.

t0yv0:
One implementation issue here is determining child resource set. Currently resource C constructor mutates parent P resource property `__childResources` which gets read in `getAllTransitivelyReferencedCustomResourceURNs` called from `registerResource`, `readResource`. 

In the current implementation it seems possible already that P may complete `registerResource(P)` call from `P.constructor()` before `C.constructor()` registers itself as a child, so P registers with an incomplete child-set. 

If we make parent option an `Input[Resource]` and spawn a thread to add a child to it, we might defer child registration until this Input resolves, which may be even after `C.constructor()` finishes. Moreover it's not clear what should wait for this thread (C and P constructions would be considered complete), so this might introduce races on whether the C-P relation is registered or not.

t0yv0:
This is currently blocked until we have a go-ahead for breaking
changes for these reasons:

- there is no non-breaking path forward for Node JS implementing
  input-ty Parent and Providers deferred options. These are needed for
  transformations and for supporting public interface such as
  [getProvider](https://github.com/pulumi/pulumi/blob/master/sdk/nodejs/resource.ts#L189)

- we cannot ship the feature in Go and have it unimplemented in Node.
  Go [PR](https://github.com/pulumi/pulumi/pull/7411) is ready and
  could be satisfactory with minor revisions

- Fundamentally the reason changes are possible in Go is that the
  runtime provides means to spin-await a deferred value onto the
  caller's stack `Promise<T> -> T`. We have similar facilities in all
  runtimes except Node

- Partial implementations that

[...]

PR #7350, opened 2021-06-22
Go support for ResourceInput versions of ResourceOption

t0yv0:
# Description

Fixes Go subset of #7012 

## Checklist

# Design:

## Questions

- [ ] Naming of newly introduced API elements - please feel free to suggest better naming
- [ ] Should we fail on Unknown or attempt to handle and proceed?
- [ ] Is it appropriate to `await` outputs inside `FooOption` or is this too eager/preventing some parallelism?

## Decided: not embedding ResourceInput interface in Resource interface

This seems difficult. If we embed ResourceInput in Resource so that Resource values can automatically be promoted to ResourceInput, then we break code that creates structs that embed ResourceState and expect to implement Resource that way. If we implement ElementType() on ResourceState itself, it will not be aware of the context where it's embedded in, and it will return Resource type; this seems to break `awaitInputs` and/or serialization tests in a way I was not able to fully resolve.

### DIFF
--- a/sdk/go/pulumi/resource.go
+++ b/sdk/go/pulumi/resource.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"fmt"
 	"reflect"
-	"runtime/debug"
 
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
 )
@@ -406,6 +405,29 @@ func Provider(r ProviderResource) ResourceOrInvokeOption {
 	})
 }
 
+// Similar to Provider, but accepts ProviderResourceInput or ProviderResourceOutput.
+func ProviderInput(pri ProviderResourceInput) resourceOrInvokeOption {
+	return resourceOrInvokeOption(func(ctx context.Context, ro *resourceOptions, io *invokeOptions) error {
+
+		// Less obvious than ParentInput that we should
+		// force-await here, but doing so for simplicity of
+		// implementation.
+		p, deps, err := awaitProviderResourceOutput(ctx, pri.ToProviderResourceOutput())
+		if err != nil {
+			return err
+		}
+
+		switch {
+		case ro != nil:
+			ro.DependsOn = append(ro.DependsOn, deps...)
+			return Providers(p).applyResourceOption(ctx, ro)
+		case io != nil:
+			io.Provider = p
+		}
+		return nil
+	})
+}
+
 // ProviderMap is an optional map of package to provider resource for a component resource.
 func ProviderMap(o map[string]ProviderResource) ResourceOption {
 	return resourceOption(func(ctx context.Context, ro *resourceOptions) error {
@@ -516,9 +538,23 @@ func awaitResourceInputMAGIC(ctx context.Context, ri ResourceInput) (Resource, [
 	resource, isResource := result.(Resource)
 
 	if !isResource {
-
-		fmt.Printf("Stack: %s\n", string(debug.Stack()))
 		return nil, nil, fmt.Errorf("ResourceInput resolved to a value that is not a Resource but a %v",
+			reflect.TypeOf(result))
+	}
+
+	return resource, deps, err
+}
+
+func awaitProviderResourceOutput(ctx context.Context, pri ProviderResourceInput) (ProviderResource, []Resource, error) {
+	result, known, _, deps, err := pri.ToProviderResourceOutput().await(ctx)
+
+	if !known {
+		return nil, nil, fmt.Errorf("Encountered unknown ProviderResourceInput, this is currently not supported")
+	}
+
+	resource, isResource := result.(ProviderResource)
+	if !isResource {
+		return nil, nil, fmt.Errorf("ProviderResourceInput resolved to a value that is not a Resource but a %v",
 			reflect.TypeOf(result))
 	}
 
--- a/sdk/go/pulumi/resource.go
+++ b/sdk/go/pulumi/resource.go
@@ -172,6 +172,7 @@ type ComponentResource interface {
 // be used for a given resource by passing it in ResourceOpt.Provider.
 type ProviderResource interface {
 	CustomResource
+	ProviderResourceInput
 
 	getPackage() string
 }
--- a/sdk/go/pulumi/resource.go
+++ b/sdk/go/pulumi/resource.go
@@ -443,6 +443,25 @@ func ProviderMap(o map[string]ProviderResource) ResourceOption {
 	})
 }
 
+// Similar to ProviderMap, but accepts ProviderResourceInput instead of ProviderResource.
+func ProviderInputMap(inputMap map[string]ProviderResourceInput) ResourceOption {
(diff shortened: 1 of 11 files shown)


sdk/go/pulumi/resource.go
```diff
@@ -369,36 +431,95 @@ func Providers(o ...ProviderResource) ResourceOption {
...
+	}
+	return &plainResourceInput{resource}, nil
+}
+
+// TODO - helps incremental refactoring, but can we completely remove
+// this function?
+func awaitResourceInputMAGIC(ctx context.Context, ri ResourceInput) (Resource, []Resource, error) {
+	result, known, _, deps, err := ri.ToResourceOutput().await(ctx)
+
+	if !known {
```
t0yv0, line 511: Should we in fact support unknowns instead of failing, for example skipping unknown resources from lists, or making unknown parent as "no parent" (which can be tricky as preview/update will re-parent, change URN)

sdk/go/pulumi/resource.go
```diff
@@ -227,79 +229,109 @@ type invokeOptions struct {
...
 // DependsOn is an optional array of explicit dependencies on other resources.
 func DependsOn(o []Resource) ResourceOption {
-	return resourceOption(func(ro *resourceOptions) {
+	return resourceOption(func(ctx context.Context, ro *resourceOptions) error {
 		ro.DependsOn = append(ro.DependsOn, o...)
+		return nil
+	})
+}
+
+// Like DependsOn, but accepts ResourceInptu and ResourceOutput.
```
pgavlin, line 308:
```suggestion
// Like DependsOn, but accepts ResourceInput and ResourceOutput.
```

sdk/go/pulumi/resource.go
```diff
@@ -227,79 +229,109 @@ type invokeOptions struct {
...
+func DependsOnInputs(o []ResourceInput) ResourceOption {
+	return deferResourceOption(func(ctx context.Context) (ResourceOption, error) {
+		// Similarly to ParentInput, we force-await any
+		// ResourceOutputs passed in right here, instead of
+		// trying to lazily await them as needed downstream.
```
pgavlin, line 313: Can you elaborate on why you chose this approach, how it interacts with `deferResourceOption`, and why you chose this approach rather than handling this as part of [`prepareResourceInputs`](https://github.com/pulumi/pulumi/blob/398fb2f8523e9d5c4fa0a064aaeab7ecdd70ee82/sdk/go/pulumi/context.go#L937)?

t0yv0: Editing the type of resourceOptions struct to become lazy in the Parent and similar options and propagating it into prepareResourceInputs and other call sites was the initial approach in the early commits here. I think it generates a larger code diff and makes the logic of those call sites a tad harder to read, while they still essentially force-awaits these options at least the Parent options, every time they are called.

deferResourceOption is a way to postpone the effect of force-awaiting Parent until the option is used, instead of being on the callstack of where the option is defined. So you don't force-await parent simply by calling ParentInput(x) but only when ParentInput(x) is used.

t0yv0: The downside with this approach is that it's less visible lexically where exactly the force-awaiting happens. I've been wondering important that is to us. If that's more important I can factor back to that approach.

t0yv0: Another question I had, is there a solution that does not `await` these inputs at all? Could we chain them into an ApplyT somehow so that the laws of ApplyT such as dependency propagation would be satisfied by construction? I came into this problem expecting to do this, but could not work it out, and ended with await + explicit deps prop instead.

pgavlin: > The downside with this approach is that it's less visible lexically where exactly the force-awaiting happens. I've been wondering important that is to us. If that's more important I can factor back to that approach.

I do think that making it more lexically clear when things are awaited is important. As it stands, it's not clear to me whether or not these `Input`s are awaited inside or outside of e.g. [this goroutine](https://github.com/pulumi/pulumi/blob/398fb2f8523e9d5c4fa0a064aaeab7ecdd70ee82/sdk/go/pulumi/context.go#L562-L632). Awaiting these `Input`s _outside_ the goroutine is inconsistent with when we await other inputs, and will unintentionally limit concurrency.

pgavlin: > Another question I had, is there a solution that does not await these inputs at all? Could we chain them into an ApplyT somehow so that the laws of ApplyT such as dependency propagation would be satisfied by construction? I came into this problem expecting to do this, but could not work it out, and ended with await + explicit deps prop instead.

Honestly, I'm not sure offhand. Eventually this will boil down into explicit `await`s, but gathering deps via the usual combinators would be useful if it is indeed possible without too much additional complexity.

t0yv0: Hrm, what are the usual combinators? That can be helpful.

Regarding the L562 goroutine. No, the parent-like inputs are not awaited in that goroutine. There are things that depend on resolved parent before we get to this goroutine. applyTransformations, mergeProviders, makeResourceState, collapseAliases. IS reordering these and/or pushing them down into L562 goroutine a safe thing to do? Are there any side-effects to watch out for when reordering?

It was getting pretty hairy but I can try to prepare a version of this PR where we try to push this down into the L562 goroutine, and we can review that then.

pgavlin: > There are things that depend on resolved parent before we get to this goroutine. applyTransformations, mergeProviders, makeResourceState, collapseAliases

Hm, that could be problematic. It looks like we have similar dependencies in e.g. the Node SDK, which observes fields of `parent` inside the `Resource` constructor. That's going to cause some pain when we need to make similar changes in that SDK.

If we have time, it might be best to take a small step back and enumerate more completely the places where we need to accept `Input<Resource>` across the various SDKs and the implications of the higher-level change.

t0yv0: Node SDK is going to be more painful indeed, from my brief foray there. It is stricter in the types, so we cannot force-await the inputs on the same stack frame.

The enumeration I'm going with is I believe complete in #7012 - the ticket description. 

One implication I foresee for cases with Parent `Input<Resource>` is that error detection that used to be performed in-band on the strict Parent Resource option will now be deferred into Input-await layer, and any errors encountered there may shift, so a constructor that used to throw will not throw but produce a resource with, say, `Output<URN>` resolving with an error.

The PM input into this issue is that the Go part is top priority, the other languages are lesser priority.

What would you advise @pgavlin ? At my level of familiarity with the SDKs I'm unable to foresee more implications by analysis, but I can put the work into putting up updated PR for Go and adding PRs for other languages, and report on the implications I discover in the process of implementing. Should I go ahead with that? The other question, do you feel we should defer pushing Go in until other language SDKs implications are worked out (see the PM note above).

t0yv0:
Closing in favor of https://github.com/pulumi/pulumi/pull/7411

Issue #6894, opened 2021-04-27
Config secret info isn't included in Construct calls


justinvp:
This is a follow-up from #3139

When the engine calls a provider's `Construct`, it passes along the config as `map[config.Key]string` (e.g. raw string values). It does not specify which config values are secrets.

We should consider adding an additional field to the [`ConstructRequest`](https://github.com/pulumi/pulumi/blob/eb2dcf24884bab173af175cd41d8279e07797a75/sdk/proto/provider.proto#L250-L272): a list of config keys that are secrets, so that when a provider rehydrates the config in the various language SDKs, we can mark them as secrets that will emit warnings when calling the non-secret-returning methods on `Config`.

Issue #7127, opened 2021-05-24
Re-enable warning when a secret config is read as a non-secret


justinvp:
We recently added a warning when reading a secret config value as a non-secret (#6896 #7078 #7079 #7080). However, one issue we've run into is that many of our provider SDKs greedily read config values in a `config` module using the non-secret config APIs, which results in a warning if such values happen to be saved as secrets -- a warning that users can't really do anything about (see #7126). We're going to temporarily disable the new warning for now. This tracks re-enabling the warning without this issue.

farvour:
Thank you. This was more of an annoyance than anything. Pulumi already detects with a value is secret encoded via the presence of the `secret` key on a value.. when the future is materialized. I don't see the purpose in telling users to specifically require a secret anyway since the underlying program should not care if it comes from a configuration whether it was sensitive or not [unless they care to check in the code themselves].

Examples where this was a nuisance:

```    warning: Configuration 'aws-rds-py:aws_rds_instances' value is a secret; use `require_secret_object` instead of `require_object```

But we intentionally nest secrets within a value of the mapping/object:

```  aws-rds-py:aws_rds_instances:
    foobar:
      master_password:
        secure: v1:zspecialsecuretlao23thj08923fn092n10r91h20jn109cn0jt09j089ajsd90th98h23589
      parameter_group_parameters:
      - apply_method: pending-reboot
        name: max_connections
        value: 200```

justinvp:
> I don't see the purpose in telling users to specifically require a secret

If you read the object via `require_object`, the object will be deserialized and available to the program with plaintext values. `master_password` will be plaintext and passing it as an input to any resource will likely result in the value being stored in the state unencrypted. To avoid this, use `require_object_secret`, which returns the object wrapped as `Output[T]` marked as a secret, and pass individual values in the object elsewhere with `apply` (e.g. `foobar.apply(lambda fb: fb.master_password)`) to maintain the value's secretness. Alternatively, continue using `require_object`, but use `output.Secret(foobar.master_password)` before passing it as an input to a resource, if you want it to to remain a secret as it flows throughout the rest of the system.

farvour:
Hmmm good point and I like that second option there as it keeps the secret within a normal configuration flexible. Interestingly enough I have been using Output.secret to generate secrets from vault using hvac directly as futures, but never thought to use this paradigm with the base configuration system.

Thanks!

justinvp:
[...]

PR #7358, opened 2021-06-23
Include config secret info in Construct calls

EvanBoyle:
# Description

This change updates the construct proto and provider server implementations to be aware of which config values are secret. This change adds some tests that verify that, and while I was in there I also added some tests that verify that secret component outputs are indeed stored as secrets. 

Currently this metadata isn't really consumed, but it is there for when we re-enable warnings where users read secrets into plaintext values: https://github.com/pulumi/pulumi/issues/7127

Given that the target functionality is currently disabled, I'll leave this out of the changelog for now. 

Fixes https://github.com/pulumi/pulumi/issues/6894

## Checklist


tests/integration/construct_component/dotnet/Component.cs
```diff
@@ -16,6 +16,9 @@ class Component : Pulumi.ComponentResource
     [Output("childId")]
     public Output<string> ChildId { get; private set; } = null!;
 
+     [Output("secret")]
```
justinvp, line 19: Super nitpick (I'm sorry):

```suggestion
    [Output("secret")]
```

justinvp (approved):
LGTM. Thanks!

PR #7363, opened 2021-06-24
Initial support for resource methods (authoring from Node.js, calling from Python)

justinvp:
This PR adds initial support for resource methods (via a new `Call` gRPC method similar to `Invoke`), support for implementing methods in Node.js, and support for calling methods in Python.

The PR is broken up into separate commits for easier review.

I've included a basic integration test, but plan on adding more test coverage in follow-up PRs (i.e. creating resources from methods, returning errors, and passing plain values).

I'll be opening separate PRs on the codegen and docgen changes. And follow-up PRs with support for calling and authoring methods in the other languages.

Part of #7072


sdk/proto/provider.proto
```diff
@@ -40,6 +40,9 @@ service ResourceProvider {
     // of responses.
     rpc StreamInvoke(InvokeRequest) returns (stream InvokeResponse) {}
 
+    // Call dynamically executes a method in the provider associated with a component resource.
+    rpc Call(CallRequest) returns (CallResponse) {}
```
justinvp, line 44: @pgavlin, I'm curious to hear your thoughts on `Call`. The original design was for methods to be implemented with `Invoke` (and we'd augment `Invoke` with the additional information needed to be able to create resources). However, @lukehoban and I were wondering if it'd be better to have a separate RPC, since we'd be adapting the raw RPC values into the Pulumi SDK model in each language, just like `Construct`. Which then leaves `Invoke` for leaf functions that just provide raw compute / data sources.

I've added `Call` in this PR, and think it's probably best to have it, but part of me still feels a little unsure. On the one hand, I do think it simplifies the implementation in the providers: it's more clear when implementing `Call` that the context and values of the call are that of the Pulumi SDK model. On the other hand, we're now bifurcating functions listed in the schema where some are implemented with `Invoke` (leaf functions that _don't_ accept inputs and don't create resources, i.e. raw compute / data sources) and others are implemented with `Call` (methods & leaf functions that do accept inputs and create resources).

If we're going to keep `Call` one question I have is what do we do about mocking. Our mocking APIs for `Invoke` are actually named `Call` (e.g. [Node.js](https://github.com/pulumi/pulumi/blob/d525a8b3318bdc2f9292bc06c106ec8ce6ee0b92/sdk/nodejs/runtime/mocks.ts#L88), [Python](https://github.com/pulumi/pulumi/blob/d525a8b3318bdc2f9292bc06c106ec8ce6ee0b92/sdk/python/lib/pulumi/runtime/mocks.py#L100), [Go](https://github.com/pulumi/pulumi/blob/d525a8b3318bdc2f9292bc06c106ec8ce6ee0b92/sdk/go/pulumi/mocks.go#L19), [.NET](https://github.com/pulumi/pulumi/blob/d525a8b3318bdc2f9292bc06c106ec8ce6ee0b92/sdk/dotnet/Pulumi/Testing/IMocks.cs#L27)), so I don't think we'd be able to distinguish these unless we came up with a different name for `Call` (or used a different name for the mock API). I was thinking we'd just have both `Invoke` and `Call` pass through to the mock's `Call` (similar to how both `registerResource` and `readReasource` are passed through to the mock [`newResource`](https://github.com/pulumi/pulumi/blob/d525a8b3318bdc2f9292bc06c106ec8ce6ee0b92/sdk/nodejs/runtime/mocks.ts#L96)).

sdk/python/lib/pulumi/runtime/invoke.py
```diff
@@ -143,3 +145,114 @@ async def do_rpc():
...
+        try:
+            # Construct a provider reference from the given provider, if one is available on the resource.
+            provider_ref, version = None, ""
+            if res is not None:
+                if res._provider is not None:
+                    provider_urn = await res._provider.urn.future()
+                    provider_id = (await res._provider.id.future()) or rpc.UNKNOWN
+                    provider_ref = f"{provider_urn}::{provider_id}"
+                    log.debug(f"Call using provider {provider_ref}")
+                version = res._version or ""
```
justinvp, line 179: Regular invokes accept an `InvokeOptions` which allows specifying the provider and version.

Having a generated method accept `InvokeOptions` as an argument felt unnatural to me since, if you wanted to specify a custom provider, you would have done so when creating the resource itself.

So I've made it so that the provider / version is picked up from the resource, which is passed into this function as part of the generated SDK code, e.g.:

```python
    def get_kubeconfig(__self__,
                       profile_name: Optional[str] = None,
                       role_arn: Optional[str] = None) -> pulumi.Output['Cluster.GetKubeconfigResult']:
        __args__ = dict()
        __args__['__self__'] = __self__
        __args__['profileName'] = profile_name
        __args__['roleArn'] = role_arn
        return pulumi.runtime.call('eks:index:Cluster/getKubeconfig', __args__, res=__self__, typ=Cluster.GetKubeconfigResult)
```

There are two cases that I need to handle in subsequent changes:

1. If we are going to use `Call` for leave functions that accept inputs / create resources, we will want to support `InvokeOptions` for that, so I'll need to update this function to be able to pass either a resource or `InvokeOptions` to specify the custom provider.

2. If the instance of the resource was rehydrated from a resource reference, I don't think the `res._provider` or `res._version` will be available to us. For that case, we'll need to make sure we can fill in that information from the built-in `pulumi:pulumi:getResource`.

tests/integration/construct_component_methods/testcomponent/index.ts
```diff
@@ -0,0 +1,92 @@
...
+    public readonly first!: pulumi.Output<string>;
+    public readonly second!: pulumi.Output<string>;
+
+    constructor(name: string, args: ComponentArgs, opts?: pulumi.ComponentResourceOptions) {
+        if (opts?.urn) {
+            args = <any>{
+                first: undefined,
+                second: undefined,
+            };
+        }
```
justinvp, line 21: This and the `if` on line 26 are annoying. This is to be able to support recreating an instance of `Component` that comes back as `__self__` resource reference. I wish there was a cleaner way to enable this for components, so authors won't have to handle this.

tests/integration/construct_component_methods/testcomponent/index.ts
```diff
@@ -0,0 +1,92 @@
...
+            version: this.version,
+            construct: (name, type, urn) => {
+                switch (type) {
+                    case "testcomponent:index:Component":
+                        return new Component(name, <any>undefined, { urn });
+                    default:
+                        throw new Error(`unknown resource type ${type}`);
+                }
+            },
+        });
```
justinvp, line 53: Similarly, it's annoying that component authors would have to write this in order to provide support for methods (for `__self__` to come back as a resource reference.

I'm actually wondering whether we could/should implicitly register any resource the first time it is created (in the `Resource` base class), so component authors don't have to even think about or write this.

tests/integration/construct_component_methods/testcomponent/index.ts
```diff
@@ -0,0 +1,92 @@
...
+    async call(token: string, inputs: pulumi.Inputs): Promise<provider.InvokeResult> {
+        switch (token) {
+            case "testcomponent:index:Component/getMessage":
+                const self = inputs.__self__;
+                if (!pulumi.Output.isInstance(self)) {
+                    throw new Error(`__self__ is not an instance of Output for method ${token}`);
+                }
+                return {
+                    outputs: {
+                        message: self.apply(s => (<Component>s).getMessage(inputs.name)),
```
justinvp, line 78: This is annoying. `inputs.__self__` is an `Output` here (I think because the resource reference has a dependency on itself, so in our `server.ts` provider implementation we wrap it in an `Output` rather than the plain value; need to double check, though).

If my suspicion for why this is an output is correct, I'm considering special casing this so we don't wrap `__self__` as an `Output` so you don't have to use an `apply` here.

(Though, ultimately, the long term plan is that you wouldn't have to write this code yourself and we'd automatically be able to dispatch for you).

pkg/resource/deploy/source_eval.go
```diff
@@ -596,6 +596,89 @@ func (rm *resmon) Invoke(ctx context.Context, req *pulumirpc.InvokeRequest) (*pu
 	return &pulumirpc.InvokeResponse{Return: mret, Failures: chkfails}, nil
 }
 
+// Call dynamically executes a method in the provider associated with a component resource.
+func (rm *resmon) Call(ctx context.Context, req *pulumirpc.CallRequest) (*pulumirpc.CallResponse, error) {
```
pgavlin, line 600: This looks good. I'd love to see some inner-loop tests in `pkg/engine/lifecycletest`.

sdk/proto/provider.proto
pgavlin: Briefly answered earlier.

I do think that we should keep `Call`: there is additional info that we need in this method that we really don't need for PoD functions (e.g. input dependencies, monitor info), and I agree that this simplifies the implementation.

> I was thinking we may be able to get by having both Invoke and Call pass through to the mock's Call (similar to how both registerResource and readReasource are passed through to the mock newResource).

This would be my preference for now. It might be the case that in the future we need to separate remote component resource registration and `Call` out into their own mocks, as the shapes of the inputs are a bit different (i.e. the user's mocks might actually want to see `Input`s rather than a bunch of data + dependency info).

sdk/python/lib/pulumi/runtime/invoke.py
pgavlin: > If we are going to use Call for leave functions that accept inputs / create resources, we will want to support InvokeOptions for that, so I'll need to update this function to be able to pass either a resource or InvokeOptions to specify the custom provider.

I believe that we will use Call for this as well, yes. I think that we can probably get away with eliding options from methods, but providing them for module-level functions.

tests/integration/construct_component_methods/testcomponent/index.ts
pgavlin: Hm, yeah, these are bothersome. I'll give this some thought.

tests/integration/construct_component_methods/testcomponent/index.ts
pgavlin: > I'm actually wondering whether we could/should implicitly register any resource the first time it is created (in the Resource base class), so component authors don't have to even think about or write this.

Ah, and this is a problem b/c component authors need these to be registered and don't have the benefit of the schema-generated registrations. I think that in the longer term, we can probably address this with code generation as well, but what you suggest would probably work for now. Given that the registrations would be taking place in the implementing component, I don't think that it would be likely (or perhaps possible) to see a reference to a resource prior to some instance of that resource being created.

pgavlin (approved):

sdk/proto/provider.proto
```diff
@@ -112,6 +115,37 @@ message InvokeResponse {
...
+message CallRequest {
+    // ArgumentDependencies describes the resources that a particular argument depends on.
+    message ArgumentDependencies {
+        repeated string urns = 1; // A list of URNs this argument depends on.
+    }
+
+    string tok = 1;                                        // the function token to invoke.
+    google.protobuf.Struct args = 2;                       // the arguments for the function invocation.
+    map<string, ArgumentDependencies> argDependencies = 3; // a map from argument keys to the dependencies of the argument.
+    string provider = 4;                                   // an optional reference to the provider to use for this invoke.
```
EvanBoyle, line 127: Does this provider refer to the component provider? 

How does this interact with the provider map that is supplied on construct? How does a call impl retrieve that? 

https://github.com/pulumi/pulumi/blob/634e97cd55d8ad0347924b535f3f0daf6517a6c1/sdk/proto/provider.proto#L269

EvanBoyle: Along the same lines, do resources created within call auto-parent to the component? Should this be configurable? 

Do we need to provide ways to expose all resource properties to call?

EvanBoyle: Also config, and config secret keys.

tests/integration/construct_component_methods/testcomponent/index.ts
EvanBoyle: >Similarly, it's annoying that component authors would have to write this in order to provide support for methods (for __self__ to come back as a resource reference.

Another option is to just require that call authors import their own generated SDKs. I have done this in working with astro.

tests/integration/construct_component_methods/testcomponent/index.ts
EvanBoyle: Hm, is this the reason why we have changes elsewhere to wait for outstanding RPC requests? This effectively forces you to create resources inside of apply.

sdk/nodejs/provider/server.ts
```diff
@@ -350,8 +316,67 @@ class Server implements grpc.UntypedServiceImplementation {
...
+        // in its own context, possibly using Node's `createContext` API to avoid modifying global state:
+        // https://nodejs.org/api/vm.html#vm_vm_createcontext_contextobject_options
+        const res = this.constructCallQueue.then(() => this.callImpl(call, callback));
+        // tslint:disable:no-empty
+        this.constructCallQueue = res.catch(() => {});
+        return res;
+    }
+
+    async callImpl(call: any, callback: any): Promise<void> {
+        try {
```
EvanBoyle, line 343: We need to wrap this in a global uncaught handler that resolves the gRPC call as was done in https://github.com/pulumi/pulumi/pull/7365

sdk/nodejs/provider/server.ts
```diff
@@ -401,6 +426,46 @@ class Server implements grpc.UntypedServiceImplementation {
...
+                                req.getMonitorendpoint(), req.getDryrun());
+
+    const pulumiConfig: {[key: string]: string} = {};
+    const rpcConfig = req.getConfigMap();
+    if (rpcConfig) {
+        for (const [k, v] of rpcConfig.entries()) {
+            pulumiConfig[k] = v;
+        }
+    }
+    runtime.setAllConfig(pulumiConfig);
```
EvanBoyle, line 442: Will need to add configSecretKeys here https://github.com/pulumi/pulumi/pull/7358

sdk/nodejs/resource.ts
```diff
@@ -809,9 +809,9 @@ export class ComponentResource<TData = any> extends Resource {
         // for a component resource.  The component is just used for organizational purposes and does
         // not correspond to a real piece of cloud infrastructure.  As such, changes to it *itself*
         // do not have any effect on the cloud side of things at all.
-        super(type, name, /*custom:*/ false, /*props:*/ remote ? args : {}, opts, remote);
-        this.__registered = remote;
-        this.__data = remote ? Promise.resolve(<TData>{}) : this.initializeAndRegisterOutputs(args);
+        super(type, name, /*custom:*/ false, /*props:*/ remote || opts?.urn ? args : {}, opts, remote);
+        this.__registered = remote || !!opts?.urn;
+        this.__data = remote || opts?.urn ? Promise.resolve(<TData>{}) : this.initializeAndRegisterOutputs(args);
```
EvanBoyle, line 814: These changes shouldn't be necessary if the caller specifies `remote: true`. This would be taken care of automatically if provider implementers import their own SDKs.

tests/integration/construct_component_methods/testcomponent/index.ts
EvanBoyle: nvm, I see that was just a refactor to keep the monitor alive. 

This makes me nervous though. Would be great to see a non-trivial test that creates a few resources within the apply (ideally with some artificially latency introduced). I wouldn't be surprised if the current approach could occasionally result in dropped dependencies.

tests/integration/construct_component_methods/testcomponent/index.ts
EvanBoyle: Again, I think this is something that can be improved by importing the generated SDK for a component. While the component instance will be different from the provider component class, both should fulfill the same public interface (which is a requirement for methods I believe). Then we can add some framework magic to do something like `ComponentFunction.prototype.apply( this=SDKInstance, args...)`

EvanBoyle (approved):
Overall the approach appears to make sense. A couple of comments to consider, and some tests/cases that should be added. Fine by me to merge as is, but we should open issues to track follow ups first. There are two changes that should be merged in from recent node runtime work. 

Summary of main concerns:

1. Tests for methods that create resources, with artificial latency injected. (concerns about resources created within apply). 
2. Plan for supporting resource options for call (resource options inherited from parent somehow? explicitly provided?).

tests/integration/construct_component_methods/testcomponent/index.ts
justinvp: I've fixed this by ensuring the resource reference remains a plain value.

sdk/proto/provider.proto
justinvp: > Does this provider refer to the component provider?

No, it's the provider where the `Call` is implemented. This is the same as with `Invoke`. When making an `Invoke` or `Call`, you call it on the monitor, and you can include the provider/version to use:

https://github.com/pulumi/pulumi/blob/398fb2f8523e9d5c4fa0a064aaeab7ecdd70ee82/sdk/proto/resource.proto#L26

And the engine will lookup the specified provider (otherwise using the default provider) and forward the call:

https://github.com/pulumi/pulumi/blob/398fb2f8523e9d5c4fa0a064aaeab7ecdd70ee82/pkg/resource/deploy/source_eval.go#L559-L561

https://github.com/pulumi/pulumi/blob/398fb2f8523e9d5c4fa0a064aaeab7ecdd70ee82/sdk/proto/provider.proto#L37

> How does this interact with the provider map that is supplied on construct? How does a call impl retrieve that?

It's unrelated. There's not currently a way to fetch a component's map of providers outside of when the component is being registered. If there are particular providers that you want to use when creating resources from methods, you'll either have to have the component save them in the state, or pass them in as arguments.

> Along the same lines, do resources created within call auto-parent to the component? Should this be configurable?

No, there's no auto-parenting.

> Do we need to provide ways to expose all resource properties to call?

I'm not sure what you mean. For methods, there is a `__self__` argument that is a resource reference to the component, so when it is deserialized, it will have all of the component's state available for use.

> config secret keys.

Yep, I added this while rebasing.

sdk/nodejs/provider/server.ts
justinvp: I included this when I rebased against master.

sdk/nodejs/provider/server.ts
justinvp: I included this when I rebased against master.

sdk/proto/provider.proto
EvanBoyle: From a practical perspective:

```ts
const x = new Cluster("foo", { providers: { aws: explicitAwsProvider }});
const y  = x.AddSubnet();
```

I would find it very surprising as a user that resources created in `x.AddSubnet()` used the default AWS provider. The interaction of resources options provided to a component and subsequent usage of `Call` is something we should make an explicit design decision on. 

Personally, this seems like it will be common enough that telling users to transfer resource options into state in order for `Call` to have access is way too painful.

pgavlin: > Personally, this seems like it will be common enough that telling users to transfer resource options into state in order for Call to have access is way too painful.

I agree. We should match the existing behavior to the greatest extent possible, which may mean exploring additional options for persistence. This might force the issue a bit on using a resource reference -> object table to ensure that any resource that has been marshaller in or out of a process is always represented by the same object (which would of course simplify other aspects of authoring as well).

justinvp: > This might force the issue a bit on using a resource reference -> object table to ensure that any resource that has been marshaller in or out of a process is always represented by the same object

Yes, this would definitely make things easier for implementers. Let me explore this a bit in the next day or two. If we're going to make a change along these lines, it'll be in a subsequent PR(s).

justinvp:
Follow-up items are being tracked in #7072

PR #7364, opened 2021-06-24
[codegen/python] Emit resource methods

justinvp:
Adds support for emitting resource methods for Python.

Note: The downstream codegen tests are currently failing because we merged #7363 which adds the new `Call` RPC which isn't implemented in those providers yet. But I don't expect this change to result in any diffs to our providers as none of them have methods defined in their schema.

Part of #7072


pkg/codegen/internal/test/testdata/simple-methods-schema/python/pulumi_example/foo.py
```diff
@@ -0,0 +1,142 @@
...
+        :param pulumi.ResourceOptions opts: Options for the resource.
+        """
+        opts = pulumi.ResourceOptions.merge(opts, pulumi.ResourceOptions(id=id))
+
+        __props__ = FooArgs.__new__(FooArgs)
+
+        return Foo(resource_name, opts=opts, __props__=__props__)
+
+    @pulumi.output_type
+    class BarResult:
```
justinvp, line 95: There are a few different options on where to emit the result class for methods. I considered emitting them in the same module as the resource, prefixed with the resource name (e.g. `FooBarResult` for a resource `Foo`). I also considered putting them in some other module like `results` (similar to what we do for output types with `outputs`). In the end, nesting the result classes within the resource class felt best (e.g. `Foo.BarResult`), which is what I've done here. This avoids adding more noise to the module for types that you'll rarely (if ever) need to refer to directly.

pkg/codegen/python/gen.go
```diff
@@ -1275,6 +1291,152 @@ func (mod *modContext) genProperties(w io.Writer, properties []*schema.Property,
 	}
 }
 
+func (mod *modContext) genMethods(w io.Writer, res *schema.Resource) {
```
justinvp, line 1294: I was originally going to modify `genFunction` to also support emitting methods, but decided to have a separate implementation for methods to avoid unnecessary diffs with #7087. After that lands, we may be able to do some consolidation.

pkg/codegen/internal/test/testdata/simple-methods-schema/python/pulumi_example/foo.py
pgavlin: Sounds good to me. In what circumstances might you need to refer to these types directly?

pgavlin (approved):

pkg/codegen/internal/test/testdata/simple-methods-schema/python/pulumi_example/foo.py
justinvp: > In what circumstances might you need to refer to these types directly?

I can only really come up with contrived examples where maybe you declare a variable with a type hint ahead of calling the method and assigning the result to the variable at a later point.

```python
result: Optional[pulumi.Output[Foo.BarResult]] = None
# ... sometime later ...
result = foo.bar() 
```

The vast majority of the time you would be calling the method and using the result directly, with no need to refer to the type because it'd be inferred from the return type hint on the method:

```python
result = foo.bar()
```

Issue #6991, opened 2021-05-07
Nodejs multi-language components hang if error is thrown from an `.apply` within the provider impl


EvanBoyle:
I narrowed down the hangs I was experiencing to a throwing an error inside of an apply in the provider component implementation. I have a method on my component that generates an output via an apply and throwing an error within does the trick consistently:

```ts
    getUsesMap(uses: pulumi.Resource[]): pulumi.Output<{ [key: string]: string}> {
        return pulumi.output(pulumi.all(uses).apply(resources => {
            // adding this line causes the hang
            throw new Error("this is intentional");
            ....
```

I did some debugging to try and see where the hang was coming from. I discovered that in the repro, the uncaught error handler is being hit, but the assumption in this comment seems to be incorrect:

https://github.com/pulumi/pulumi/blob/b4323ea437fd06293ecf5f8b3529376759367d5d/sdk/nodejs/provider/server.ts#L445-L446

The only way I was able to fix the hang was by manually draining the work queue and exiting the process on the first sign of error as you can see in this draft PR:

https://github.com/pulumi/pulumi/pull/6989

I tried to run `disconnect()` instead which waits for outstanding work to complete, but after 15+ minutes the process was still hung, presumably with some outstanding work in the rpcKeepalive queue. 

I tried to hang a catch handler off of the constructImpl promise that would properly respond to the gRPC request with an error, but this catch block never gets hit:

https://github.com/pulumi/pulumi/blob/b4323ea437fd06293ecf5f8b3529376759367d5d/sdk/nodejs/provider/server.ts#L262

PR #7365, opened 2021-06-25
Fix a hang in nodejs remote components when an error is thrown within an apply

EvanBoyle:
# Description

The uncaught handler wasn't terminating the outstanding gRPC request causing a hang. This change adds/removes an uncaught handler around constructImpl calls that bubbles the uncaught error back via the gRPC callback. 

Fixes # https://github.com/pulumi/pulumi/issues/6991

## Checklist

### DIFF
--- a/sdk/nodejs/provider/server.ts
+++ b/sdk/nodejs/provider/server.ts
@@ -275,9 +275,6 @@ class Server implements grpc.UntypedServiceImplementation {
         const uncaughtHandler = (err: Error) => {
             if (!this.uncaughtErrors.has(err)) {
                 this.uncaughtErrors.add(err);
-                // Use `pulumi.log.error` here to tell the engine there was a fatal error, which should
-                // stop processing subsequent resource operations.
-                log.error(err.stack || err.message || ("" + err));
             }
             // bubble the uncaught error in the user code back and terminate the outstanding gRPC request.
             callback(err, undefined);
--- a/sdk/nodejs/provider/server.ts
+++ b/sdk/nodejs/provider/server.ts
@@ -37,13 +37,15 @@ const statusproto = require("../proto/status_pb.js");
 class Server implements grpc.UntypedServiceImplementation {
     readonly engineAddr: string;
     readonly provider: Provider;
+    readonly uncaughtErrors: Set<Error>;
 
     /** Queue of construct calls. */
     constructQueue = Promise.resolve();
 
-    constructor(engineAddr: string, provider: Provider) {
+    constructor(engineAddr: string, provider: Provider, uncaughtErrors: Set<Error>) {
         this.engineAddr = engineAddr;
         this.provider = provider;
+        this.uncaughtErrors = uncaughtErrors;
     }
 
     // Satisfy the grpc.UntypedServiceImplementation interface.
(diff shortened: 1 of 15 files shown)


tests/integration/integration_nodejs_test.go
```diff
@@ -1017,3 +1017,35 @@ func TestComponentProviderSchemaNode(t *testing.T) {
 	}
 	testComponentProviderSchema(t, path)
 }
+
+// Test throwing an error within an apply in a remote component written in nodejs.
+// The provider should return the error and shutdown gracefully rather than hanging.
+func TestConstructNodeErrorApply(t *testing.T) {
+	if runtime.GOOS == WindowsOS {
+		t.Skip("Temporarily skipping test on Windows")
+	}
```
justinvp, line 1026: Can we try running it on Windows?

You'll need to update `build.proj` like:

https://github.com/pulumi/pulumi/blob/634e97cd55d8ad0347924b535f3f0daf6517a6c1/build.proj#L283-L286

and

https://github.com/pulumi/pulumi/blob/634e97cd55d8ad0347924b535f3f0daf6517a6c1/build.proj#L298

sdk/nodejs/provider/server.ts
```diff
@@ -267,6 +269,23 @@ class Server implements grpc.UntypedServiceImplementation {
...
+        // to avoid a hang.
+        const uncaughtHandler = (err: Error) => {
+            if (!this.uncaughtErrors.has(err)) {
+                this.uncaughtErrors.add(err);
+                // Use `pulumi.log.error` here to tell the engine there was a fatal error, which should
+                // stop processing subsequent resource operations.
+                log.error(err.stack || err.message || ("" + err));
+            }
+            // bubble the uncaught error in the user code back and terminate the outstanding gRPC request.
+            callback(err, undefined);
```
justinvp, line 283: I'm wondering if it should be the following:

```suggestion
            if (!this.uncaughtErrors.has(err)) {
                this.uncaughtErrors.add(err);
                // bubble the uncaught error in the user code back and terminate the outstanding gRPC request.
                callback(err, undefined);
            }
```

I'd added the `log.error` elsewhere, but I think that was misguided -- despite what was originally documented, I don't think it'll stop processing subsequent resource operations. So not sure it does much good here (unless `callback(err, undefined)` doesn't end up showing the error in the `pulumi up` output).

tests/integration/construct_component_error_apply/nodejs/component.ts
```diff
@@ -0,0 +1,19 @@
+// Copyright 2016-2020, Pulumi Corporation.  All rights reserved.
```
justinvp, line 1:
```suggestion
// Copyright 2016-2021, Pulumi Corporation.  All rights reserved.
```

tests/integration/construct_component_error_apply/nodejs/index.ts
```diff
@@ -0,0 +1,6 @@
+// Copyright 2016-2020, Pulumi Corporation.  All rights reserved.
```
justinvp, line 1:
```suggestion
// Copyright 2016-2021, Pulumi Corporation.  All rights reserved.
```

tests/integration/construct_component_error_apply/testcomponent/index.ts
```diff
@@ -0,0 +1,47 @@
+import * as pulumi from "@pulumi/pulumi";
```
justinvp, line 1: The original file is missing the copyright.

```suggestion
// Copyright 2016-2021, Pulumi Corporation.  All rights reserved.

import * as pulumi from "@pulumi/pulumi";
```

justinvp (approved):
LGTM

sdk/nodejs/provider/server.ts
EvanBoyle: Fair enough, it doesn't seem to make a difference in the output. However, the callback should not be made inside the if (this would equate to a gRPC hang in some instances).

EvanBoyle:
Flakey codegen test is unrelated. Merging.

Issue #7359, opened 2021-06-24
Fix Go SDK to be compatible with changes in #7059


lukehoban:
With the changes in #7059, any attempt to use the new features causes a panic.

```
panic: fatal: An assertion has failed: map[]: cannot assign map[string]providerconstruct.TestObjInput to map[string]providerconstruct.TestObj
    goroutine 13 [running]:
    github.com/pulumi/pulumi/sdk/v3/go/common/util/contract.failfast(...)
    	/Users/Luke/Code/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.4.0/go/common/util/contract/failfast.go:23
    github.com/pulumi/pulumi/sdk/v3/go/common/util/contract.Assertf(0x18b7e00, 0x1a05ac9, 0x1a, 0xc0004a8fc0, 0x3, 0x3)
    	/Users/Luke/Code/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.4.0/go/common/util/contract/assert.go:33 +0x1a5
    github.com/pulumi/pulumi/sdk/v3/go/pulumi.marshalInputAndDetermineSecret(0x18b7e80, 0x0, 0x1b0aba0, 0x18b7e20, 0x1, 0x1, 0x0, 0x181cc53, 0xe, 0x0, ...)
    	/Users/Luke/Code/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.4.0/go/pulumi/rpc.go:288 +0x514
    github.com/pulumi/pulumi/sdk/v3/go/pulumi.marshalInput(0x18b7e80, 0x0, 0x1b0aba0, 0x18b7e20, 0x181cc01, 0xe, 0x0, 0x0, 0x1b0aba0, 0x18b7e20, ...)
    	/Users/Luke/Code/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.4.0/go/pulumi/rpc.go:170 +0x73
    github.com/pulumi/pulumi/sdk/v3/go/pulumi.marshalInputs.func1(0x181cc6b, 0xe, 0x18b7e80, 0x0, 0x1b0aba0, 0x18b7e20, 0x195, 0x0)
    	/Users/Luke/Code/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.4.0/go/pulumi/rpc.go:77 +0x9f
[...]

PR #7369, opened 2021-06-25
Preliminary fix for #7359.

pgavlin:
These changes contain a preliminary fix for #7359 in the Go SDK. The fix
handles input values that are nested one level deep within maps and
arrays, but does not handle other cases of nested input types.


pgavlin:
Working on adding some tests for this now.

justinvp (approved):
Aside from some tests, LGTM.

Issue #2814, opened 2019-06-11
Using S3 backend with multiple projects is unpleasant


eropple:
Current situation: I've got a number of...I keep wanting to call them _stacks_, but in Pulumi that word means something (though [the documentation](https://pulumi.io/reference/organizing-stacks-projects/) also confusingly overloads the term!), so let's call them _projects_...encompassing different bits of functionality. While I want to use the Pulumi SaaS, I need to prove that we can do our thing by using the S3 backend first.

Right now, the experience of doing so feels pretty bad. And while I'm gung-ho about Pulumi, I feel queasy about maybe having it too early here because while the actual resource/state management seems to be very mature I don't know if I can adequately manage this stuff _and_ provide a quality, DRY developer experience without writing a bunch of should-be-superfluous wrapper scripts around Pulumi itself.

## Current options ##
- I can tell my developers that they need to switch stacks to something like `dev.build.my-service` (or something equally long and gross), and that their configs will be stored in `Pulumi.dev.build.my-service.yaml` (barf). Then I can pass similarly grody `pulumi.StackReference` names in our code. The IAM policies necessary to effectively control access to these states are also going to be kinda gross.
- I can use separate `pulumi login`s for separate services, but `pulumi login` is stateful at a user level (this strikes me as no bueno, FWIW, but is probably a separate issue) and the S3 backend can't talk to a different `pulumi login`/"backend" (another term that seems aggressively overloaded) so `my-service` couldn't talk to `core-infra` even if it wanted to.

## Desired Outcomes ##
- We have a `name` field in `Pulumi.yaml`. I would like the contents of that field to be namespaced into the `.pulumi/stacks` directory. For example, for `my-service`, I want `dev` to be at `.pulumi/stacks/my-service/dev.json`.
- I would like this name to be part of the name used for `pulumi.StackReference`s. I should be able to go something like

```ts
// while working in project `my-service`, stack `dev` - `my-service/dev`
const env = config.get("env") || pulumi.getStack().split("/").last();
const core = new pulumi.StackReference(`core-infra/${env}`); // `core-infra/dev`
```

...and just have it work. (I recognize that this potentially collides with the organization stuff in the SaaS product, but I can't come up with a better way to do this that respects that `.pulumi` really wants to think of itself as a file system.

Thanks, folks. Appreciate your hard work.

ellismg:
[...]

PR #3120, opened 2019-08-20
Simple project support for local backends

Frassle:
This is probably a bit messy but it gets some basic project support into the local backend. It can still list and load old stacks which aren't associated with a project. Ideally the whole project idea should probably be lifted up so project handling is generalised but that's probably awkward while also keeping back compat with all the current file stacks not under project folders.

There's still some issues, rename isn't working quite right (history files don't move). But want to get some eyes on this early to check the approach is sound before spending time getting it good enough to merge.

PR #7372, opened 2021-06-28
Add project support for local backend

foriequal0:
# Description

Add project support for local backend. I'm not sure local project would need organization too.
I've put new project enabled local stacks under `.pulumi/projects/{stacks,locks,...}`.

Fixes: #2814

## Checklist

This is not a finished PR. I just wanted to get feedbacks before I work in depth.

### DIFF
--- a/pkg/backend/filestate/backend.go
+++ b/pkg/backend/filestate/backend.go
@@ -53,7 +53,6 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/common/tokens"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/cmdutil"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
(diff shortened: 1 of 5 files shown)


liubog2008:
Any progress of this PR?

sammck:
Is this project being maintained?

Frassle:
> Is this project being maintained?

I think it's clear from other activity on this repo that the project is maintained. If you mean the file backend support in particular, then yes while it has been bereft of work recently it is now one of our priorities for this year.

This is not the only PR raised to add project support to the file backend, I raised one myself a few years back (https://github.com/pulumi/pulumi/pull/3120). We know that project support is highly requested and it is being worked on.

@foriequal0 thank you for your efforts here, but I'm going to close this PR for now. Pulumi are going to make much larger changes to the backend interface which will include support for projects, but also trying to unify some other differences between the cloud and file backends.

PR #7374, opened 2021-06-28
[codegen/python] - Include python enum docstrings

komalali:
# Description

Fixes # https://github.com/pulumi/pulumi/issues/7375

## Checklist

### DIFF
--- a/pkg/codegen/internal/test/testdata/simple-enum-schema/python/pulumi_plant/tree/v1/_enums.py
+++ b/pkg/codegen/internal/test/testdata/simple-enum-schema/python/pulumi_plant/tree/v1/_enums.py
@@ -27,8 +27,11 @@ class RubberTreeVariety(str, Enum):
     types of rubber trees
     """
     BURGUNDY = "Burgundy"
+    "A burgundy rubber tree."
     RUBY = "Ruby"
+    "A ruby rubber tree."
     TINEKE = "Tineke"
+    "A tineke rubber tree."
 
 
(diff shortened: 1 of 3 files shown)


pkg/codegen/internal/test/testdata/simple-enum-schema/python/pulumi_plant/tree/v1/_enums.py
```diff
@@ -27,8 +27,11 @@ class RubberTreeVariety(str, Enum):
     types of rubber trees
     """
     BURGUNDY = "Burgundy"
+    "A burgundy rubber tree."
```
t0yv0, line 30: Here it is :)

t0yv0: I wonder if you need `"""` to indicate it's a docstring to the tooling. Otherwise LGTM!

t0yv0 (approved):

pkg/codegen/internal/test/testdata/simple-enum-schema/python/pulumi_plant/tree/v1/_enums.py
komalali: Right, this is my fix :)

komalali: ah good point!

PR #7377, opened 2021-06-28
[sdk/nodejs] Support for calling methods

justinvp:
Support for calling methods from Node.js.

Part of #7072

### DIFF
--- a/sdk/nodejs/runtime/invoke.ts
+++ b/sdk/nodejs/runtime/invoke.ts
@@ -23,7 +23,7 @@ import { InvokeOptions } from "../invoke";
 import * as log from "../log";
 import { Inputs, Output } from "../output";
 import { debuggablePromise } from "./debuggable";
-import { deserializeProperties, serializeProperties } from "./rpc";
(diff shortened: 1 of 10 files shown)


EvanBoyle (approved):
overall these changes look good. runtime/* has been a been a bug farm and can fail in exciting ways so it would be good to see some tests that specifically exercise the `call` failure path if at all possible. We have some runtime register resource tests that do this:

https://github.com/pulumi/pulumi/blob/e955a6b06a19f9222286618600b84f581bc931ec/sdk/nodejs/tests/runtime/props.spec.ts#L361

sdk/nodejs/runtime/invoke.ts
```diff
@@ -101,27 +101,27 @@ export async function streamInvoke(
         const req = createInvokeRequest(tok, serialized, provider, opts);
 
         // Call `streamInvoke`.
-        const call = monitor.streamInvoke(req, {});
```
pgavlin, line 104: Did this change b/c of a lint issue w. shadowing?

pgavlin (approved):

sdk/nodejs/runtime/invoke.ts
justinvp: Yes

PR #7378, opened 2021-06-28
[codegen/nodejs] Emit resource methods

justinvp:
Adds support for emitting resource methods for Node.js.

Part of #7072

---

This is ready for review.

Note: there is an issue in this PR that I'm still investigating. The schema defines a method taking both an inputty and plain argument for the same type, e.g.:

```typescript
    export interface BarArgs {
        baz?: pulumi.Input<inputs.nested.BazArgs>;
        bazPlain?: inputs.nested.BazArgs;
    }
```

But currently only the `inputs.nested.Baz` is being emitted:

```typescript
    export interface Baz {
        hello?: string;
        world?: string;
    }
```


pkg/codegen/nodejs/gen.go
```diff
@@ -717,6 +793,47 @@ func (mod *modContext) genResource(w io.Writer, r *schema.Resource) error {
...
+					continue
+				}
+				args = append(args, arg)
+			}
+			if len(args) > 0 {
+				comment := fun.Inputs.Comment
+				if comment == "" {
+					comment = fmt.Sprintf("The set of arguments for the %s.%s method.", name, method.Name)
+				}
+				mod.genPlainType(w, methodName+"Args", comment, args, true, false, 1)
```
t0yv0, line 814: When working on codegen, these positional booleans were making it harder for me to read and keep track of what's this doing. Don't fix in this PR but consider maybe we should name them in a helper record or something so call  sites are easier to read.

t0yv0:
The pulumi-aws diff is weird.. perhaps pulumi-aws is behind master in the codegen it's using so we have spurious diffs not coming from the changes here.

pkg/codegen/internal/test/testdata/simple-methods-schema/nodejs/foo.ts
```diff
@@ -0,0 +1,117 @@
...
+        }
+        super(Foo.__pulumiType, name, inputs, opts);
+    }
+
+    /**
+     * A description of bar.
+     */
+    bar(args: Foo.BarArgs): pulumi.Output<Foo.BarResult> {
+        return pulumi.runtime.call("example::Foo/bar", {
+            "__self__": this,
```
t0yv0, line 62: Nit: if there's never any renaming here, could we just pack this up into `withSelf(this, args)` or even `{'self': this, 'args': args}` and have `call` unpack. It's no big deal I guess since methods are relatively rare, I've just been cognizant of generated code size.

t0yv0 (approved):
LGTM. I'm least familiar with Node JS codegen so reviewing lightly here.

justinvp:
There's still an issue when the schema defines a method taking both an inputty and plain argument for a type, the plain type is emitted, but not the inputty type. I'm not sure if this is even specific to methods or a general issue. Either way, this doesn't block mainline scenarios where the args are just primitive types, so I'm going to go ahead and merge this and finish investigating/addressing this issue in a subsequent PR. I've added this as a follow-up task in #7072.

PR #7383, opened 2021-06-29
[codegen/go] - Inputty Go enums

### LINKED ISSUE #7351: Generated enums in Go are not "Inputty"
Given the following schema definition for an input property

```
                "volumeType": {
                    "$ref": "#/types/x:index:EbsVolumeType",
                    "type": "string"
                },
```

And the following (shortened) definition of the referenced type:

```
        "x:index:EbsVolumeType": {
            "type": "string",
            "enum": [
                {
                    "name": "gp2",
                    "value": "gp2"
                },
                {
                    "name": "io1",
                    "value": "io1"
                },
            ]
        },
```

The Go code generator generates the following entry in to the input-typed arguments structure for the resource with the input property:

```
	VolumeType EbsVolumeType `pulumi:"volumeType"`
```

However, `EbsVolumeType` is a type alias for `pulumi.String`, which is not an `Input`.

Consequently, `Output` types cannot be supplied here, which is unexpected.

komalali:
# Description

Fixes #7351 

## Checklist

### DIFF
--- a/pkg/codegen/go/doc.go
+++ b/pkg/codegen/go/doc.go
@@ -88,8 +88,8 @@ func (d DocLanguageHelper) GetLanguageTypeString(pkg *schema.Package, moduleName
 		glog.Errorf("cannot calculate type string for type %q. could not find a package for module %q", t.String(), moduleName)
 		os.Exit(1)
 	}
-	if _, ok := t.(*schema.EnumType); ok {
-		return modPkg.inputType(t)
+	if typ, ok := t.(*schema.EnumType); ok {
+		return modPkg.tokenToEnum(typ.Token)
 	}
 	return modPkg.typeString(t)
 }
--- a/pkg/codegen/go/doc.go
+++ b/pkg/codegen/go/doc.go
@@ -88,9 +88,6 @@ func (d DocLanguageHelper) GetLanguageTypeString(pkg *schema.Package, moduleName
 		glog.Errorf("cannot calculate type string for type %q. could not find a package for module %q", t.String(), moduleName)
 		os.Exit(1)
 	}
-	if typ, ok := t.(*schema.EnumType); ok {
-		return modPkg.tokenToEnum(typ.Token)
-	}
 	return modPkg.typeString(t)
 }
 
--- a/pkg/codegen/go/gen_test.go
+++ b/pkg/codegen/go/gen_test.go
@@ -182,4 +182,39 @@ func TestEnumUsage(t *testing.T) {
 			return nil
 		}, pulumi.WithMocks("project", "stack", mocks(1))))
 	})
+
+	t.Run("EnumOutputs", func(t *testing.T) {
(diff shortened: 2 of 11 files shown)


komalali:
I have done my best to update downstream providers, but unfortunately, there are issues with [k8s](https://github.com/pulumi/pulumi-kubernetes/issues/1645) and [azure-native](https://github.com/pulumi/pulumi-azure-native/pull/974/checks?check_run_id=3011412146) that are unlikely to be resolved within the timeline needed for this PR.

I request that you review my PR based on the downstream codegen tests that do work, and the codegen tests within the pulumi/pulumi repo which test both the code generation as well as the usability of the generated enums.

pkg/codegen/go/gen_test.go
```diff
@@ -124,21 +125,25 @@ func TestEnumUsage(t *testing.T) {
...
-				size := all[3].(*int)
-				typ := all[4].(string)
+				size := all[3].(*plant.ContainerSize)
+				typ := all[4].(tree.RubberTreeVariety)
 				assert.Equal(t, *material, "ceramic", "unexpected material on resource: %v", urn)
 				assert.Equal(t, *color, "Magenta", "unexpected color on resource: %v", urn)
-				assert.Equal(t, *size, 22, "unexpected size on resource: %v", urn)
-				assert.Equal(t, typ, "Ruby", "unexpected type on resource: %v", urn)
+				assert.Equal(t, *size, plant.ContainerSize(22), "unexpected size on resource: %v", urn)
+				assert.Equal(t, typ, tree.RubberTreeVarietyRuby, "unexpected type on resource: %v", urn)
```
komalali, line 146: Note that it is no longer possible to pass in untyped strings/integers to strict enums. This is in line with how strings must be wrapped in `pulumi.String` to function correctly.

emiliza:
@pgavlin can you take a look? 👀

pkg/codegen/go/doc.go
```diff
@@ -88,8 +88,8 @@ func (d DocLanguageHelper) GetLanguageTypeString(pkg *schema.Package, moduleName
 		glog.Errorf("cannot calculate type string for type %q. could not find a package for module %q", t.String(), moduleName)
 		os.Exit(1)
 	}
-	if _, ok := t.(*schema.EnumType); ok {
-		return modPkg.inputType(t)
+	if typ, ok := t.(*schema.EnumType); ok {
+		return modPkg.tokenToEnum(typ.Token)
```
pgavlin, line 92: What happens if we delete these lines entirely? I have a feeling that the call to `typeString` below will do the right thing.

pgavlin (approved):
Looks great! Just one question about a small simplification.

pkg/codegen/go/doc.go
komalali: Indeed!

t0yv0:
Props for tests that *run* generated code! Awesome.

PR #7388, opened 2021-06-29
Add flag for printing full stack names

svangordon-fruit:
# Description

Add a global flag, `--full-name`, that will print the fully qualified stack name, instead of possibly eliding the org or project name.

### Output before:
```
❯ pulumi stack --show-name
full-name-test
```
### Output after:
```
❯ pulumi stack --show-name --full-name
fooOrg/barProject/full-name-test
```

Fixes # (issue)

## Checklist

I wasn't able to find tests that exercise the `pulumi stack --show-name` command with a service backend, but I've included a unit test for `cloudBackendReference.String()` 

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -12,7 +12,7 @@
 - [sdk/go] - Add preliminary support for unmarshaling plain arrays and maps of output values.
   [#7369](https://github.com/pulumi/pulumi/pull/7369)
 
-- [cli] - Add global flag to print fully qualified stack names.
+- [cli] - Add global flag to print fully qualified stack names. [#7388](https://github.com/pulumi/pulumi/pull/7388)
 
 ### Bug Fixes
 
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -12,6 +12,8 @@
 - [sdk/go] - Add preliminary support for unmarshaling plain arrays and maps of output values.
   [#7369](https://github.com/pulumi/pulumi/pull/7369)
 
+- [cli] - Add global flag to print fully qualified stack names.
+
 ### Bug Fixes
 
 - [sdk/dotnet] - Fix swallowed nested exceptions with inline program, so they correctly bubble to the consumer.
--- a/pkg/cmd/pulumi/pulumi.go
+++ b/pkg/cmd/pulumi/pulumi.go
@@ -180,6 +180,8 @@ func NewPulumiCmd() *cobra.Command {
 		"Enable verbose logging (e.g., v=3); anything >3 is very verbose")
 	cmd.PersistentFlags().StringVar(
 		&color, "color", "auto", "Colorize output. Choices are: always, never, raw, auto")
+	cmd.PersistentFlags().BoolVar(&httpstate.PrintFullStackNames, "full-name", false,
+		"Print fully qualified stack names")
 
 	// Common commands:
 	//     - Getting Started Commands:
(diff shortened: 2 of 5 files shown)


CHANGELOG_PENDING.md
```diff
@@ -12,6 +12,8 @@
 - [sdk/go] - Add preliminary support for unmarshaling plain arrays and maps of output values.
   [#7369](https://github.com/pulumi/pulumi/pull/7369)
 
+- [cli] - Add global flag to print fully qualified stack names.
```
lukehoban, line 15:
```suggestion
- [cli] - Add global flag to print fully qualified stack names. [#7388](https://github.com/pulumi/pulumi/pull/7388)
```

pkg/cmd/pulumi/pulumi.go
```diff
@@ -180,6 +180,8 @@ func NewPulumiCmd() *cobra.Command {
 		"Enable verbose logging (e.g., v=3); anything >3 is very verbose")
 	cmd.PersistentFlags().StringVar(
 		&color, "color", "auto", "Colorize output. Choices are: always, never, raw, auto")
+	cmd.PersistentFlags().BoolVar(&httpstate.PrintFullStackNames, "full-name", false,
```
lukehoban, line 183: I don't think we'll want to couple this flag directly into global state in the `httpstate` module.  Some more notes in summary, but would love to find a way to scope this a bit more if possible.

lukehoban (commented):
> `pulumi stack --show-name`

Is this the only place this change is intended to change behaviour? Or does this stack display aim to impact how stacks are rendered elsewhere?

Due to the issues noted below on coupling this flag to global state in the `httpstate` module, I'm inclined to see if we can just make this new behaviour the default.  I'm unsure exactly what set of things will be impacted by that though. 

If we can't make this the default, my inclination would be to enable this behind an environment variable instead of a flag, to avoid having to thread that flag all the way through the backend interfaces.  

If the only place we *want* to change the behaviour is in `pulumi stack`, is there a way to scope the fix here to just that function - instead of having to plumb through to full backend interface (just pass this as an option to a single backend function?)

svangordon-fruit:
@lukehoban I believe that we could make this behavior the default, and it wouldn't break anything w/in the client. If anyone has written external tooling that assumes that `pulumi stack ls` is outputting names in a specific format, it would break that tooling. Also, I'm not sure if you have any thoughts about developer experience for the wider Pulumi community -- making that the default would help me, but I'm not sure if it would make things better or worse for everyone else.

pgavlin:
I think that scoping this without using some sort of package-global state is likely to require some refactoring. That refactoring might be a good thing, ultimately: it's always bothered me that we call `CurrentUser` (which can perform network I/O!) from within `String`, where we have no opportunity to fail, cancel, etc.

> If the only place we want to change the behaviour is in pulumi stack, is there a way to scope the fix here to just that function - instead of having to plumb through to full backend interface (just pass this as an option to a single backend function?)

One way we could achieve this and make some progress towards the refactoring mentioned above:

- Move the current implementation of `String` to a function like `DisplayName(ctx context.Context, abbreviate bool) (string, error)`
- Have `String` call `DisplayName(context.Background(), true)`

Callers of `DisplayName` would then be able to control the format of the returned name. Eventually, we would want to move all callers of `String` to use `DisplayName` and pass flags as appropriate.

lunaris:
Apologies; I didn't spot this when I implemented #15857 -- the `-Q` or `--fully-qualify-stack-names` flag should cover this use case :smile:

PR #2630, opened 2019-04-05
Allow querying of stack state snapshots

hausdorff:

PR #7389, opened 2021-06-29
Fix pylint issues

### LINKED ISSUE #7387: Python lint failures in CI
Screen Shot 2021-06-29 at 3 31 28 PM

justinvp:
Fixes #7387

### DIFF
--- a/sdk/python/lib/pulumi/resource.py
+++ b/sdk/python/lib/pulumi/resource.py
@@ -816,12 +816,6 @@ class CustomResource(Resource):
     dynamically loaded plugin for the defining package.
     """
 
-    __pulumi_type: str
-    """
-    Private field containing the type ID for this object. Useful for implementing `isInstance` on
-    classes that inherit from `CustomResource`.
-    """
-
     def __init__(self,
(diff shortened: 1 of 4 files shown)


sdk/python/lib/pulumi/resource.py
```diff
@@ -816,12 +816,6 @@ class CustomResource(Resource):
     dynamically loaded plugin for the defining package.
     """
 
-    __pulumi_type: str
```
justinvp, line 819: This was added in https://github.com/pulumi/pulumi/pull/2630 (specifically https://github.com/pulumi/pulumi/pull/2630/commits/44f0a4f7f78c9d3f10db3fd9102642118c19a22e) to set the stage for being able to filter resources by type for `pulumi query`.

The Node.js code generator will emit static `isInstance` methods that make use of the equivalent field in Node.js, but looks like we never got around to implementing the equivalent functionality for Python. Since this isn't being used, I'm deleting it. We can add it back if/when we actually need it.

t0yv0 (approved):
🚢

PR #7394, opened 2021-06-30
Demo an issue with naive Input/Output wrappers for interfaces

t0yv0:
# Description

Demonstrates an issue with naive Input/Output wrappers around an Example type that happens to be an interface.

```
go test -run TestInputOutputWrappersForInterfaces
panic: interface conversion: pulumi.example is not pulumi.Input: missing method ElementType

goroutine 20 [running]:
github.com/pulumi/pulumi/sdk/v3/go/pulumi.awaitInputs(0x1b8a418, 0xc0001a8000, 0x1979f20, 0xc0004385f0, 0x16, 0x196dd00, 0xc000438600, 0x194, 0x9a00000, 0x0, ...)
	/Users/anton/pulumi/sdk/go/pulumi/types.go:691 +0x280d
github.com/pulumi/pulumi/sdk/v3/go/pulumi.toOutputTWithContext.func3(0xc0001cb840, 0x1979f20, 0xc0004385f0, 0x1b8b3a0, 0xc000436460, 0x1b8a418, 0xc0001a8000, 0x196dd00, 0xc000438600, 0x194, ...)
	/Users/anton/pulumi/sdk/go/pulumi/types.go:828 +0x137
created by github.com/pulumi/pulumi/sdk/v3/go/pulumi.toOutputTWithContext
	/Users/anton/pulumi/sdk/go/pulumi/types.go:820 +0x216
exit status 2
FAIL	github.com/pulumi/pulumi/sdk/v3/go/pulumi	0.170s
```

In the real tasks I've encountered this trying to build a ResourceInput wrapper for the Resource interface type. The workaround is to not embed XInput interface in X interface, but then some explicit conversion methods are needed. Do we have a convention for these? Or should we consider this a bug we can fix in awaitInputs and make the example work as written. 

## Checklist

### DIFF
--- a/sdk/go/pulumi/types.go
+++ b/sdk/go/pulumi/types.go
@@ -687,8 +687,13 @@ func awaitInputs(ctx context.Context, v, resolved reflect.Value) (bool, bool, []
 		// but `T` does (https://golang.org/ref/spec#Method_sets).
 		// In this case, dereference the pointer to get at its actual value.
 		if v.Kind() == reflect.Ptr && valueType.Kind() != reflect.Ptr {
-			v = v.Elem()
-			contract.Assert(v.Interface().(Input).ElementType() == valueType)
+			if v.Elem().CanInterface() {
+				input, ok := v.Elem().Interface().(Input)
+				if ok && input.ElementType() == valueType {
+					v = v.Elem()
+					contract.Assert(v.Interface().(Input).ElementType() == valueType)
+				}
+			}
 		}
 
 		// If we are assigning the input value itself, update the value type.
@@ -696,7 +701,7 @@ func awaitInputs(ctx context.Context, v, resolved reflect.Value) (bool, bool, []
 			valueType = v.Type()
 		} else {
 			// Handle pointer inputs.
-			if v.Kind() == reflect.Ptr {
+			if v.Kind() == reflect.Ptr && valueType.Kind() == reflect.Ptr {
 				v, valueType = v.Elem(), valueType.Elem()
 
 				resolved.Set(reflect.New(resolved.Type().Elem()))
@@ -715,7 +720,15 @@ func awaitInputs(ctx context.Context, v, resolved reflect.Value) (bool, bool, []
 	}
 
 	known, secret, deps, err := true, false, make([]Resource, 0), error(nil)
-	switch v.Kind() {
+
+	var shape reflect.Kind
+	if v.Kind() == resolved.Kind() {
+		shape = v.Kind()
+	} else {
+		shape = 9999 // trigger the default case
+	}
+
+	switch shape {
 	case reflect.Interface:
 		if !v.IsNil() {
 			return awaitInputs(ctx, v.Elem(), resolved)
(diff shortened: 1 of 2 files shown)


sdk/go/pulumi/types.go
```diff
@@ -687,16 +687,21 @@ func awaitInputs(ctx context.Context, v, resolved reflect.Value) (bool, bool, []
 		// but `T` does (https://golang.org/ref/spec#Method_sets).
 		// In this case, dereference the pointer to get at its actual value.
 		if v.Kind() == reflect.Ptr && valueType.Kind() != reflect.Ptr {
-			v = v.Elem()
-			contract.Assert(v.Interface().(Input).ElementType() == valueType)
+			if v.Elem().CanInterface() {
```
t0yv0, line 690: This change essentially copies the Assert into the if condition. The test made assert fail here, though it is not the relevant case, not what the comment describes.

sdk/go/pulumi/types.go
```diff
@@ -687,16 +687,21 @@ func awaitInputs(ctx context.Context, v, resolved reflect.Value) (bool, bool, []
...
+			}
 		}
 
 		// If we are assigning the input value itself, update the value type.
 		if assignInput {
 			valueType = v.Type()
 		} else {
 			// Handle pointer inputs.
-			if v.Kind() == reflect.Ptr {
+			if v.Kind() == reflect.Ptr && valueType.Kind() == reflect.Ptr {
```
t0yv0, line 704: Without the extra condition `valueType.Elem()` fails on the example test.

sdk/go/pulumi/types.go
```diff
@@ -715,7 +720,15 @@ func awaitInputs(ctx context.Context, v, resolved reflect.Value) (bool, bool, []
 	}
 
 	known, secret, deps, err := true, false, make([]Resource, 0), error(nil)
-	switch v.Kind() {
+
+	var shape reflect.Kind
+	if v.Kind() == resolved.Kind() {
```
t0yv0, line 725: Code below seems to assume pretty strongly that `v.Kind() == resolved.Kind()` as it simultaneously recurs over these. I think this should be safe to not recur when this is not the case, as in our example. It's tricky though..

t0yv0:
I think I found a fairly surgical fix to awaitInputs that lets my test pass. We need to move cautiously there though.. this function is a little big. I suspect it is implementing the non-type-preserving transformation such as for example converting `StringInput` to `string` in certain positions such as `FooInput` property to `FooArgs` property. This was not obvious to me from the start, and is a source of some difficulty here. I'd love some day to rework this, specify, factor and test this out so it is more obviously correct.

For the moment, let's see if the test suites pass with these fixes, then I'd like some really solid review on this.

justinvp:
I asked it over in the other PR, but curious about this: https://github.com/pulumi/pulumi/pull/7584#discussion_r687062672

lunaris:
Can we/should we close this?

lunaris:
Closing as this seems stale. Feel free to challenge if I'm wrong here!

t0yv0:
https://github.com/pulumi/pulumi/pull/16697 yeah the issue is still around

Issue #6912, opened 2021-04-29
`pulumi up --logflow` causes node multi-lang component provider to hang


EvanBoyle:
Repro:

1. create a component from typescript boilerplate
2. `pulumi up --logflow` and observe no progress (I think this interferes with the provider printing out the port)

t0yv0:
Cannot reproduce on 

```
pulumi version
v3.5.1
```

EvanBoyle:
Just gave this a try, as I was surprised this didn't repro given nothing has changed in the boilerplate. Turns out my original repro description was wrong. Specifically it is the combination of two CLI flags:

```console
$ pulumi up --skip-preview --logflow --logtostderr
```

Reproduces for me after cloning the boilerplate repro, building the provider, and running the program with the latest pulumi nodejs sdk.

EvanBoyle:
Somehow the logflow flag is getting picked up as the engine address which causes log messages to hang. The runtime options when specifying both flags end up as: 

```
    {
      project: 'simple',
      stack: 'abc',
      dryRun: false,
      queryMode: false,
      parallel: 2147483647,
      monitorAddr: '127.0.0.1:52743',
      engineAddr: '--logtostderr',
      syncDir: undefined,
      cacheDynamicProviders: true,
      testModeEnabled: false,
      legacyApply: false
    }
```

Will prep a fix

t0yv0:
Unfortunately having to reopen due to the fix causing https://github.com/pulumi/pulumi/issues/7642 - going to revert that fix.

Issue #7402, opened 2021-06-30
Add full provider flagset to multi-language providers, automation api lang servers


EvanBoyle:
Providers currently ignore flags like `--logflow` `--logtostderr` and `-v=10`. We should support the same set of flags in our provider boilerplate. Automation API inline language servers have the same issue. Dynamic providers also do not support these flags and should be considered. 

Would be helpful to make these changes across the board to assist with observability and supportability.

PR #7404, opened 2021-06-30
specify plugin arguments before other flags

EvanBoyle:
# Description

This change alters the order in which we provide arguments to plugins. Previously we were adding flags like `--logflow` ahead of standard plugin args like the engine address. This causes hangs in nodejs multi-lang provider servers which don't have super robust argument parsing (also applies to dynamic providers) and expect the first argument to be the engine grpc address.

This fix is just a bandaid, with a test to ensure we don't regress. I've filed a follow up issue here to add the full flagset to all multi-lang provider servers: https://github.com/pulumi/pulumi/issues/7402

Fixes https://github.com/pulumi/pulumi/issues/6912

## Checklist

### DIFF
--- a/sdk/go/common/resource/plugin/plugin.go
+++ b/sdk/go/common/resource/plugin/plugin.go
@@ -280,6 +280,7 @@ func newPlugin(ctx *Context, pwd, bin, prefix string, args, env []string, option
 // execPlugin starts the plugin executable.
 func execPlugin(bin string, pluginArgs []string, pwd string, env []string) (*plugin, error) {
 	var args []string
+	args = append(args, pluginArgs...)
 	// Flow the logging information if set.
 	if logging.LogFlow {
 		if logging.LogToStderr {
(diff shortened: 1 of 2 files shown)


t0yv0 (approved):
Thanks Evan, awesome. And thanks for backlogging the proper fix.

t0yv0:
```
  pulumi:pulumi:Stack (construct_component_nodejs-p-it-fv-az201-5-nodejs-c0541807):
    usage: testcomponent.py [-h] engine
    testcomponent.py: error: unrecognized arguments: --logtostderr
 
Resources:
    + 1 created

Duration: 3s


--- FAIL: TestConstructNode (44.19s)
    --- FAIL: TestConstructNode/testcomponent-python (8.23s)
```

May be related. Let me check this out quickly since you're on vac.

t0yv0:
Closing in favor of https://github.com/pulumi/pulumi/pull/7412 where this is included.

PR #7410, opened 2021-07-01
[sdk/go] Add stack output helpers for numeric type

SDuck4:
# Description

When using the Pulumi stack, some stack output types are `int` or `float64`(Ex. Service port).
However, there is no function to get stack output into numeric types.
So, I add stack output helpers for numeric type.

## Checklist


sdk/go/pulumi/stack_reference.go
```diff
@@ -39,6 +39,24 @@ func (s *StackReference) GetIDOutput(name StringInput) IDOutput {
 	}).(IDOutput)
 }
 
+// GetFloat64Output returns a stack output keyed by the given name as an Float64Output
+func (s *StackReference) GetFloat64Output(name StringInput) Float64Output {
+	return s.GetOutput(name).ApplyT(func(out interface{}) float64 {
+		var res float64
+		if out != nil {
+			res = out.(float64)
```
EvanBoyle, line 47: This type conversion will panic if out is say, a `string` instead of `float64`. It is not uncommon for users to try and do something like this so we should catch and return an error rather than crashing the runtime. The applier func accepted by `ApplyT` has a form that can propagate errors just for this purpose:

https://github.com/pulumi/pulumi/blob/d4180d9f616100c959c74013ae5191bb2703bc0f/sdk/go/pulumi/types.go#L380

sdk/go/pulumi/stack_reference.go
```diff
@@ -39,6 +39,24 @@ func (s *StackReference) GetIDOutput(name StringInput) IDOutput {
...
+		}
+		return res
+	}).(Float64Output)
+}
+
+// GetIntOutput returns a stack output keyed by the given name as an IntOutput
+func (s *StackReference) GetIntOutput(name StringInput) IntOutput {
+	return s.GetFloat64Output(name).ApplyT(func(out float64) int {
+		return int(out)
```
EvanBoyle, line 56: Similar to the feedback above, we should catch and propagate the error here. In addition, we should just use `strconv.Atoi` directly rather than delegating to `s.GetFloat64Output`: 

https://pkg.go.dev/strconv#example-Atoi

Reason being that in the error case, the user who is trying to get an `int` will get an error message complaining that the value could not be parsed as a float64 which would be pretty confusing.

EvanBoyle (commented):
Thanks for the contribution! Two suggestions for your consideration.

sdk/go/pulumi/stack_reference.go
SDuck4: I modified the code to return the error in case of a type mismatch. Please check it.

sdk/go/pulumi/stack_reference.go
SDuck4: Now I understand the error catch and propagation.
I modified the code to process the type conversion within the function not delegating to `s.GetFloat64Output` and return error when failed.
Please check it.

EvanBoyle (approved):
Thanks for the contribution!

PR #7411, opened 2021-07-01
[Blocked] Go support for ResourceInput versions of ResourceOption - take 2

t0yv0:
Status: blocked as discussed in #7012

# Description

This is an alternative version of https://github.com/pulumi/pulumi/pull/7350 that implements the Go subset of #7012

See "real-ish" example in https://github.com/t0yv0/test-pulumi-parent-input-option/blob/main/examples/simple-go/main.go#L18

Where this is going: as we discussed with @pgavlin there was a desire to move all `awaits` that can block in `ctx.ReadResource` and `ctx.RegisterResource` on waiting on Outputs (hidden in Inputs) into separate go routines - currently these are anonymous `go func()` kickoffs - let's call them *resolvers* at the end of those funcs. Before this change, all awaits happen there. The change introduces more inputs, most importantly in the Parent option. So this PR attempts to make sure that ParentInput is awaited there also.

The challenges are refactoring code that depends on options.Parent safely and reordering the effects.

I think this work is getting there with several caveats that may or may not be show-stopper:

- [x] Promises added to compensate for races on `transformations, aliases, providers` properties

- [ ] Transformations admit ResourceOption and can be a source of more inputs. In the current code it simply fails with an exception if a user attempts to pass Inputs in transformations "not implemented yet". We could instead await those, since it would seem *applyTransformations* now runs on the *resolver* goroutine.

- [ ] Changes to the error path. Errors such as incorrect options previously detected in-band by registerResource/readResource are moved off into *resolver* routine and will be surfacing differently.

- [ ] IDEA (not done yet) - at the cost of some more defensive code, I can make sure these changes only happen if the resource options do indeed contain Inputs, and otherwise make the code behave exactly as before; this will ensure legacy programs are not affected

- [ ] TODO s3.BucketOutput and friends does not auto-cast to ResourceInput, but likely should, if we can; to simplify the user code

A way to avoid these difficulties would be awaiting option inputs in a blocking fashion inside registerResource/readResource, and not in the *resolver* goroutine. I'm told we do not want that but would benefit from fully understanding why - resolutions to known=true will need to be handled, anything else?

Fixes # (issue)

## Checklist


sdk/go/pulumi/context.go
```diff
@@ -220,7 +220,10 @@ func (ctx *Context) Invoke(tok string, args interface{}, result interface{}, opt
 	options := &invokeOptions{}
 	for _, o := range opts {
 		if o != nil {
-			o.applyInvokeOption(options)
+			err := o.applyInvokeOptionAfterAwaitingInputs(ctx.ctx, options)
```
t0yv0, line 223: Introducing a new input await here --^

sdk/go/pulumi/context.go
```diff
@@ -399,8 +382,14 @@ func (ctx *Context) ReadResource(
 			return
 		}
 
+		transformedOpts, transformedProps, err := ctx.transformOptionsAndProps(t,
```
t0yv0, line 385: Introducing another input await here --^

sdk/go/pulumi/context.go
```diff
@@ -530,58 +519,44 @@ func (ctx *Context) registerResource(
...
 		var err error
+		deps := make(map[string][]Resource)
+
 		defer func() {
 			resState.resolve(ctx, ctx.DryRun(), err, inputs, urn, resID, state, deps)
 			ctx.endRPC(err)
 		}()
 
+		transformedOpts, transformedProps, err :=
+			ctx.transformOptionsAndProps(t, name, resStateBuilder, props, opts...)
```
t0yv0, line 546: And another await here --^

sdk/go/pulumi/context.go
```diff
@@ -675,7 +695,10 @@ func applyTransformations(t, name string, props Input, resource Resource, opts [
 
 		res := transformation(args)
 		if res != nil {
-			resOptions := merge(res.Opts...)
+			resOptions, err := tryMergeWithoutAwaiting(res.Opts...)
```
t0yv0, line 698: Currently will fail if asked to await, however we can switch this to await here. This runs within contexts that are resolving resources already.

sdk/go/pulumi/resource.go
```diff
@@ -217,6 +219,39 @@ type resourceOptions struct {
 	Version string
 }
 
+func awaitResourceInput(ctx context.Context, ri ResourceInput) (Resource, []Resource, error) {
+	result, known, _, deps, err := ri.ToResourceOutput().await(ctx)
+
+	if !known {
+		return nil, nil, fmt.Errorf("Encountered unknown ResourceInput, this is currently not supported")
```
t0yv0, line 226: Not yet obvious to me of whether we should try to support it somehow already.  I need figure out an example representative of the use case.

sdk/go/pulumi/resource.go
```diff
@@ -217,6 +219,39 @@ type resourceOptions struct {
...
+	}
+
+	return resource, deps, err
+}
+
+func awaitProviderResourceInput(ctx context.Context, pri ProviderResourceInput) (ProviderResource, []Resource, error) {
+	result, known, _, deps, err := pri.ToProviderResourceOutput().await(ctx)
+
+	if !known {
+		return nil, nil, fmt.Errorf("Encountered unknown ProviderResourceInput, this is currently not supported")
```
t0yv0, line 243: Same concern.

t0yv0:
Manual example works, but tests indicate provider transformation facility breaking, invetigating.

t0yv0:
The newly introduced promises were never initializing for newDependecyResource references. The timing out examples involved a multi-lang scenario where RegisterResource was called in Go with a Parent option pointing to an URN allocated by Node, calling newDependencyResource, and stalling awaiting its transformations, aliases, providers. The above commit adds an init step for these, let's see what the test suite tells us now.

lunaris:
Closing this as it seems blocked somewhat indefinitely. That said, it seems we also worked around it satisfactorily in #13592, so hopefully we don't need to resurrect this to move forward :tada:

PR #7433, opened 2021-07-06
[codegen] Unify SDK codegen testing

pgavlin:
Rather than duplicating the list of tests and codegen driver across each
SDK, move its definition into `pkg/codegen/internal/test`. This has a
few notable benefits:

- All SDK code generators will be tested against each test. Though some
  tests may exercise a particular code generator more than others, the
  extra coverage will be generally beneficial.
- Adding a new test is simpler, as only a single file needs to be
  changed.
- All SDKs now honor the `PULUMI_ACCEPT` environment variable for
  updating baselines.
- Codegen tests now validate all generated files instead of only a
  particular subset.


pgavlin:
@komalali this is what I had in mind for a uniform driver for the SDK codegen tests. I think that we should do something similar for the program codegen tests--wdyt? We could also take this approach for a certain set of more unittestish bits, e.g. generating type names from schema types.

pgavlin:
I'm going to split this into two commits so that it's actually reviewable in the web UI 😬

pgavlin:
Okay, this is now two commits. The first commit contains the test code changes; the second updates the baselines.

komalali (approved):
This is great! Yeah, I agree we should use a similar strategy for programgen tests (and really anything that has to be replicated across languages)

PR #7437, opened 2021-07-06
[sdk/go] Support for calling methods

justinvp:
Support for calling methods from Go.

Part of #7072


sdk/go/pulumi/context.go
```diff
@@ -308,6 +308,182 @@ func (ctx *Context) Invoke(tok string, args interface{}, result interface{}, opt
 	return nil
 }
 
+// Call will invoke a provider call function, identified by its token tok.
+//
+// output is used to determine the output type to return; self is optional for methods.
+func (ctx *Context) Call(tok string, args Input, output Output, self Resource, opts ...InvokeOption) (Output, error) {
```
justinvp, line 314: I'd love feedback on the signature. In particular `output Output`. It is used to pass a "dummy" output, which is only used to be able to determine the type of the output to return. The actual output we return is a new one that we create.

For example:

```go
func (c *Component) GetMessage(ctx *pulumi.Context, args *ComponentGetMessageArgs) (ComponentGetMessageResultOutput, error) {
 	out, err := ctx.Call("testcomponent:index:Component/getMessage", args, ComponentGetMessageResultOutput{}, c)
 	if err != nil {
 		return ComponentGetMessageResultOutput{}, err
 	}
 	return out.(ComponentGetMessageResultOutput), nil
 }
```

Is this weird? Is there a "better" way?

t0yv0: `reflect.Type` and pass `reflect.TypeOf(ComponentGetMessageResultOutput{})` comes to mind as an option.

t0yv0: For `self Resource`, is only the URN/id actually used?

sdk/go/pulumi/context.go
```diff
@@ -308,6 +308,182 @@ func (ctx *Context) Invoke(tok string, args interface{}, result interface{}, opt
...
+			}
+			resolvedArgs["__self__"] = resolvedSelf
+			argDeps["__self__"] = deps
+		}
+
+		// Marshal all properties for the RPC call.
+		keepUnknowns := ctx.DryRun()
+		rpcArgs, err := plugin.MarshalProperties(
+			resolvedArgs,
+			plugin.MarshalOptions{KeepSecrets: true, KeepUnknowns: keepUnknowns, KeepResources: ctx.keepResources})
```
t0yv0, line 381: I was poking around messy bugs with MarshalOptions recently. I think the code here says this: if ctx.DryRun()=true (preview?), preserve as-is, otherwise try to filter unknowns out (which afaik can change the "shape" of the data by e.g. removing required properties). Is that the desired behavior?

There is also an option to RejectUnknowns=true which will fail with an exception if it sees any unknowns.

What I'm thinking could be an alternative here. If `ctx.DryRun() = false` (we're applying changes), be strict and RejectUnknowns since there should not be any. If `ctx.DryRun() = true` need to choose between `KeepUnknowns=true` and `KeepUnknowns=false` depending on whether the implementing code knows how to handle them. Not knowing much about this,`KeepUnknowns=true` seems like a better default to me since I don't fully trust the filtering process.

sdk/go/pulumi/context.go
```diff
@@ -308,6 +308,182 @@ func (ctx *Context) Invoke(tok string, args interface{}, result interface{}, opt
...
+				output.getState().reject(err)
+			}
+
+			// Allocate storage for the unmarshalled output.
+			var secret bool
+			dest := reflect.New(output.ElementType()).Elem()
+			known := !outprops.ContainsUnknowns()
+			secret, err = unmarshalOutput(ctx, resource.NewObjectProperty(outprops), dest)
+			if err != nil {
+				output.getState().reject(err)
```
t0yv0, line 437: reject() might be called twice here after L428 is that ok?

sdk/go/pulumi/context.go
```diff
@@ -556,8 +735,12 @@ func (ctx *Context) registerResource(
 	// Merge providers.
 	providers := mergeProviders(t, options.Parent, options.Provider, options.Providers)
 
+	// Get the provider for the resource.
+	provider := getProvider(t, options.Provider, providers)
```
t0yv0, line 739: This seems like a bugfix for provider not being pulled out of providers map when options.Provider is nil?

sdk/go/pulumi/context.go
```diff
@@ -1032,14 +1231,14 @@ func getTimeouts(custom *CustomTimeouts) *pulumirpc.RegisterResourceRequest_Cust
 
 // getOpts returns a set of resource options from an array of them. This includes the parent URN, any dependency URNs,
 // a boolean indicating whether the resource is to be protected, and the URN and ID of the resource's provider, if any.
-func (ctx *Context) getOpts(t string, providers map[string]ProviderResource, opts *resourceOptions, remote bool,
-) (URN, []URN, bool, string, map[string]string, bool, ID, []string, []string, string, error) {
+func (ctx *Context) getOpts(t string, provider ProviderResource, opts *resourceOptions, remote bool,
+) (URN, []URN, bool, string, map[string]string, bool, ID, []string, []string, error) {
```
t0yv0, line 1235: 10-tuples remind me of Haskell pain from previous job, sitting there counting out positions to the 7th argument during a 3am debugging session. Could we use a record to name the fields? I can undertake a separate PR 😅

t0yv0:
Looks good to me, neat, consistent with the other things I've seen. Comments mostly nit-ty and an opportunity to educate me. 

Now, this only modifies the Go SDK client to the Engine to issue Calls, assume the impl of Calls is in separate PR?

t0yv0 (approved):

sdk/go/pulumi/context.go
justinvp: > ctx.DryRun()=true (preview?)

Yep, `DryRun` means preview.

> What I'm thinking could be an alternative here.

I like the idea of being strict and setting `RejectUnknowns=true` when `DryRun` is false. When `DryRun` is false there should not be any unknown values, so whether we're filtering or strictly rejecting, it shouldn't matter either way, and rejecting would be a good assertion to have.

However, the code here is based on the exact same marshaling options used when registering/reading a resource:

https://github.com/pulumi/pulumi/blob/2c0d6924cbf374931857c17de8d15712d3b6d057/sdk/go/pulumi/context.go#L1155-L1162

My preference is to not introduce `RejectUnknowns` as part of this change, and instead consider switching to using it here (and elsewhere) consistently in a subsequent change.

sdk/go/pulumi/context.go
justinvp: Great catch! I'll address this.

sdk/go/pulumi/context.go
justinvp: It's so I can save the `provider` in the ResourceState, so I can use it subsequently inside `Call` when a `self` argument is passed to it, to pass along any explicit provider or version for the resource. That way, we don't have to expose `InvokeOptions` on a method to be able to specify those things, instead we can just implicitly use the provider/version of the resource.

sdk/go/pulumi/context.go
justinvp: > `reflect.Type` and pass `reflect.TypeOf(ComponentGetMessageResultOutput{})` comes to mind as an option.

Yep, certainly an option. It's a more primitive type and may reduce some confusion around "Why am I passing an `Output` here?" On the other hand, then they have to write `reflect.TypeOf(ComponentGetMessageResultOutput{})` when we could just do that for them. And someone may end up passing some other `reflect.Type` completely unrelated to an output.

I think I'll stick with what I've got as it forces passing an output. Users aren't typically going to be using `Call` directly anyway (instead they'll be using generated methods/functions in our SDKs).

> For `self Resource`, is only the URN/id actually used?

It'll be used as the `__self__` argument, which is marshaled as a resource reference (which ends up using the URN/id). It's also used for the provider/version options, if they're available on the resource.

sdk/go/pulumi/context.go
t0yv0: Trying here https://github.com/pulumi/pulumi/pull/7457

PR #7438, opened 2021-07-06
[codegen/go] Emit resource methods

justinvp:
Adds support for emitting resource methods for Go.

Note: The downstream codegen tests are currently failing because we merged #7363 which adds the new `Call` RPC which isn't implemented in those providers yet. But I don't expect this change to result in any diffs to our providers as none of them have methods defined in their schema.

Part of #7072


t0yv0 (approved):
LGTM.

PR #7441, opened 2021-07-07
Turn off fail-fast for codegen tests

komalali:
### Problem

Currently, all downstream codegen test jobs get canceled if just one fails. 

### Solution

This PR changes the behavior so that some tests may run to completion even if others fail. 

### Reasoning

While we strive to always have all of the downstream codegen jobs run to completion, there are occasions where certain providers cannot be updated due to extenuating circumstances. In those situations, it is better to see the result of **some** downstream provider codegen tests rather than none at all.


pgavlin (approved):
:100:

PR #7443, opened 2021-07-07
[sdk/python] Translate property dep keys.

pgavlin:
The map of property dependencies the Python SDK receives from the
engine uses `camelCase` property keys, but the SDK performs lookups
using `snake_case` property keys. Fix this by translating the map's keys
from `camelCase` to `snake_case` prior to performing any lookups.


pgavlin:
FYI @jen20

lukehoban (approved):
Would be great to follow up with test coverage here - and likely a reminder to add additional coverage of potential name translation issues in other places.

jen20:
This looks promising! I’ll test it locally.

Issue #5582, opened 2020-10-15
[Automation API, multi-language components] Provider config statically set when package is imported for node, python, dotnet


EvanBoyle:
TLDR; provider config for node.js, python, and c# uses global state and is incompatible with Automation API inline programs. 

Provider config is instantiated statically at package import time:

```
import * as aws from "@pulumi/aws";

// any values here are static from the time the package was imported.
aws.config; 
```

See the implementation: https://github.com/pulumi/pulumi-aws/blob/master/sdk/nodejs/config/vars.ts

For traditional CLI driven pulumi programs this is fine as the order of ops is typically:

1. `pulumi config set aws:region us-west-2` from the CLI
2. `pulumi up` use CLI to exec a typescript program, this serialized config into environment variables
3. Pulumi up starts a langhost that execs the program with the appropriate config.
4. Program runs and `import * as aws from "@pulumi/aws"` establishes static config that cannot change.
5. Calls to `aws.config.*` have up to date config values.

With inline program the flow is problematic since we run a single program to set config and run the program:

1. We hit the import section of the program. `import * as aws from "@pulumi/aws"` is the first thing that runs, solidifying all values in `aws.config.*` (ie anything in the `aws:` config namespace is undefined).
2. A workspace and stack are created.
3. Config is set on the stack: `stack.setConfig("aws:region", { value: "us-west-2" });`
4. Program is run (`stack.up()`) and that config is serialized and handed off to the program.
5. The program gets the correct config via environment variables, but any provider config like that read from the awsx package here will return `undefined` as it was statically read from the import before we did any config setting: https://github.com/pulumi/pulumi-awsx/blob/b515c09c32196286380dc08b362c53272b63d6a3/nodejs/awsx/utils.ts#L137-L140

Our go SDK does the right thing here are exposes functions. This was the config is read fresh when the value in question is referenced: https://github.com/pulumi/pulumi-aws/blob/master/sdk/go/aws/config/config.go#L13

For node it appears we can use property getters to fix this in a non-breaking way.
Python definitely has the same behavior: https://github.com/pulumi/pulumi-aws/blob/master/sdk/python/pulumi_aws/config/vars.py
I'm not familiar with the lifecycle of C# static classes, but it looks like it's probably an issue here as well: https://github.com/pulumi/pulumi-aws/blob/master/sdk/dotnet/Config/Config.cs

### Languages
- [x] Dotnet (#7549)
- [x] Node.js (#7530)
- [x] Python (#7447)

[...]

PR #7447, opened 2021-07-07
[codegen/python] - Implement dynamic config-getters

komalali:
# Description

This PR implements dynamic provider config getters for python. To avoid a breaking change, we define a class that extends the ModuleType type and implements property getters for each config key. We then overwrite the `__class__` attribute of the current module as described in the proposal for [PEP-549](https://www.python.org/dev/peps/pep-0549/).

To continue to get autocomplete for the config keys, we now emit a `__init__.pyi` stub file in the config module which declares the variables and associated types.

I have added tests for the codegen side of things, but I'm struggling to figure out how to test the functionality in isolation. I used the following Automation API script to locally validate the result, but it requires a locally built pulumi_aws based on the codegen in this PR.

```python
import pulumi
from pulumi import automation as auto
import pulumi_aws as aws

def pulumi_program():
    pulumi.export("region", aws.config.region)

project_name = "dynamic_config_test"
stack_name = "dev"
stack = auto.create_or_select_stack(stack_name=stack_name,
                                    project_name=project_name,
                                    program=pulumi_program)

stack.workspace.install_plugin("aws", "v4.11.0")

stack.set_config("aws:region", auto.ConfigValue(value="us-west-1"))
up_res = stack.up()
assert up_res.outputs["region"].value == "us-west-1"

stack.set_config("aws:region", auto.ConfigValue(value="us-east-1"))
up_res = stack.up()
assert up_res.outputs["region"].value == "us-east-1"

stack.destroy()
```

Python part of #5582 

#### Notable Changes

- [The actual codegen changes](https://github.com/pulumi/pulumi/pull/7447/files#diff-f96a068940ee11b51a8bb4e6bbd2289ff1d18dc300bb90f527868eee0a0097f1)
- [Generated `vars.py`](https://github.com/pulumi/pulumi/pull/7447/files#diff-d68481e7a1f76424dec1fed17000dd21648cf430ec75301f9d559c6c45cce70d)
- [Generated `config/__init__.py`](https://github.com/pulumi/pulumi/pull/7447/files#diff-6d2b007846270d7d7c8c011f38a3715869d2c0a63d23a35b50458cfdd9ed0e5f)
- [Generated `config/__init__.pyi`](https://github.com/pulumi/pulumi/pull/7447/files#diff-d7a17d5ae669f6507c6e2e112c0a77b86ad8b634fce20c4be9e4f0b756a53c00)

## Checklist


pkg/codegen/internal/test/testdata/provider-config-schema/python/pulumi_configstation/config/vars.py
```diff
@@ -0,0 +1,52 @@
...
+
+import types
+
+__config__ = pulumi.Config('configstation')
+
+
+class _ExportableConfig(types.ModuleType):
+    @property
+    def favorite_plants(self) -> Optional[Sequence[str]]:
+        return __config__.get('favoritePlants')
```
justinvp, line 21: I'm not sure if this is related to your changes, but this doesn't look right.

Since the type hint is `Optional[Sequence[str]]`, I would expect this to use `__config__.get_object` like the following (and possibly cast the result to make mypy happy, if needed):

```suggestion
    def favorite_plants(self) -> Optional[Sequence[str]]:
        return __config__.get_object('favoritePlants')
```

Looking at AWS, for Node.js you can see it's doing the right thing:

https://github.com/pulumi/pulumi-aws/blob/1b4fa4fecdc5b89c0141f07dcc9f65a4a1465d85/sdk/nodejs/config/vars.ts#L16

```typescript
export let allowedAccountIds: string[] | undefined = __config.getObject<string[]>("allowedAccountIds");
```

But for Python, it looks like we don't return the rich objects, we just always use `get` and return the string value (which for objects, will be a JSON string):

https://github.com/pulumi/pulumi-aws/blob/1b4fa4fecdc5b89c0141f07dcc9f65a4a1465d85/sdk/python/pulumi_aws/config/vars.py#L42

```python
allowed_account_ids = __config__.get('allowedAccountIds')
```

Unfortunately, I don't think we can change this without it being breaking for our existing providers. If we want to return actual objects, we'll probably want a Python setting in the schema so providers can opt-out (or opt-in depending on what we want the default behavior to be).

Or we just maintain the existing behavior and type all these as `Optional[str]`.

pkg/codegen/internal/test/testdata/provider-config-schema/python/pulumi_configstation/config/vars.py
```diff
@@ -0,0 +1,52 @@
...
+
+    @property
+    def favorite_sandwich(self) -> Optional[pulumi.InputType['outputs.Sandwich']]:
+        """
+        omg my favorite sandwich
+        """
+        return __config__.get('favoriteSandwich')
```
justinvp, line 28: Two issues:

1. If we want to return the object, it should be typed as `Optional['outputs.Sandwich']` (remove the `pulumi.InputType`)
2. It's going to need to use `get_object`, but that alone isn't enough. `get_object` uses `json.loads(v)` to deserialize the JSON string to an object, but this'll just return a `dict`, so we'll need to convert the dict to the output type.
    - For inputs, we have a helper ([`pulumi.runtime.to_json`](https://github.com/pulumi/pulumi/blob/a952140b34c667d005116ee53be08077b5863948/sdk/python/lib/pulumi/runtime/_json.py#L40)) that helps serialize an input type to JSON, which is used by the `Provider` resource class (e.g. [in AWS](https://github.com/pulumi/pulumi-aws/blob/1b4fa4fecdc5b89c0141f07dcc9f65a4a1465d85/sdk/python/pulumi_aws/provider.py#L456)).
    - Perhaps we need a similar `pulumi.runtime.from_json` that can be used here, if we want to return these objects. It'd have to work similar to how we unmarshal RPC values into outputs, to instantiate the right output types based on the type information.
        - We might need `pulumi.runtime.from_json` anyway for https://github.com/pulumi/pulumi/issues/6259, if we want to expose the non-string outputs on the `Provider` resource, if we're not doing the serialization/deserialization at the Go -> gRPC boundary.

justinvp (commented):
Two initial comments after taking a quick look

pkg/codegen/internal/test/testdata/provider-config-schema/python/pulumi_configstation/config/vars.py
komalali: Interesting... yeah I was suspicious of the types that it was generating, especially the `pulumi.InputType`. I wonder if this is why the [existing config variables](https://github.com/pulumi/pulumi-aws/blob/master/sdk/python/pulumi_aws/config/vars.py) don't specify types at all.

pkg/codegen/internal/test/testdata/provider-config-schema/python/pulumi_configstation/config/vars.py
komalali: Thanks for clarifying this, I think it makes sense to just type them all as Optional[str] for now, so we can maintain the current behavior consistently for automation API and multilang. 

I'll open up a follow-up issue that we can consider to return the rich objects and/or the output types.

I guess this begs the question, what's the point of the `config/outputs` at all if we're not using them anywhere? I imagine they are just being generated automatically as a result of the Input/Output typing work that you did last year but we're not actually using them... right?

pkg/codegen/internal/test/testdata/provider-config-schema/python/pulumi_configstation/config/__init__.py
```diff
@@ -0,0 +1,8 @@
+# coding=utf-8
+# *** WARNING: this file was generated by test. ***
+# *** Do not edit by hand unless you're certain you know what you are doing! ***
+
+import sys
+from .vars import _ExportableConfig
+
+sys.modules[__name__].__class__ = _ExportableConfig
```
t0yv0, line 8: 🤯

t0yv0:
I feel a little lost situating what this is for, do we have a ticket list? I'll try to read again next week with a fresh head. By the implementation of it, it sounds like schema can define config variables, types and defaults, and the PR makes these "type-safely" (I'd say statically) available to the autogenerated Python resource provider code. What's tripping me up is "dynamic", use of auto API, and what exactly are the bw-compat concerns.

komalali:
@t0yv0 this is for the python part of #5582 - sorry, I had that in the PR description but I guess it got dropped at some point. The problem is defined pretty well in the issue, let me know if you need any more information.

pkg/codegen/internal/test/testdata/provider-config-schema/python/pulumi_configstation/config/vars.py
justinvp: > I guess this begs the question, what's the point of the `config/outputs` at all if we're not using them anywhere? I imagine they are just being generated automatically as a result of the Input/Output typing work that you did last year but we're not actually using them... right?

Yeah, likely.

komalali: To close the loop here, I have opted to maintain the status quo as fixing the types would involve a breaking change. I opened an issue to track returning the correct types: https://github.com/pulumi/pulumi/issues/7493

pkg/codegen/python/gen.go
```diff
@@ -760,7 +770,34 @@ func (mod *modContext) genConfig(variables []*schema.Property) (string, error) {
 			configFetch += " or " + v
 		}
 
-		fmt.Fprintf(w, "%s = %s\n", PyName(p.Name), configFetch)
+		fmt.Fprintf(w, "%s@property\n", indent)
+		fmt.Fprintf(w, "%sdef %s(self) -> %s:\n", indent, PyName(p.Name), mod.typeString(p.Type, false, true))
```
justinvp, line 774: Shouldn't the type always be `str` (or `Optional[str]`)? Also, it might be worth adding a brief comment on why its typed as `str` (for historical reasons to maintain backcompat).

justinvp (approved):
LGTM otherwise.

It might be worth adding a comment that explains briefly why we're generating the `_ExportableConfig` and using `sys.modules[__name__].__class__ = _ExportableConfig`, and why we're generating the stub file.

pkg/codegen/python/gen.go
komalali: whoopsies

PR #7452, opened 2021-07-08
[experimental] Add support for YAML schemas.

pgavlin:
These changes provide an alternative method for authoring Pulumi
schemas. Rather than hewing closely to the design of OpenAPI, this
package chooses a YAML-based approach that allows for a much more
concise description of a Pulumi schema. Although this package is
intended to cover the majority of use cases, it is not intended to have
feature parity with the JSON-based schema language, and omits some
advanced features for the sake of simplicity.

The two most striking differences are the representations of package
members and property types.

Rather than using separate, flat namespaces for types, resources, and
functions, a YAML schema uses a single, hierarchical namespace. The type
of a member is indicated using a YAML tag; members may be object types,
enum types, resources, components, functions, or modules.

YAML schemas also make use of tags and the module tree to simplify type
references. A type reference may be a primitive type, constructed type,
or a reference to an object type, enum type, or resource. Primitive
types are booleans, strings, integers, and numbers; constructued types
are inputs, arrays, maps, and unions. At the outermost level, a
property's type may also be optional, in which case the property does
not require a value.

# Description

Fixes # (issue)

## Checklist

### DIFF
--- a/pkg/codegen/schema/schema.go
+++ b/pkg/codegen/schema/schema.go
@@ -1655,14 +1655,17 @@ func (t *types) bindEnumValues(values []*EnumValueSpec, typ Type) ([]*Enum, erro
 				return nil, errorMessage(spec.Value, typ.String())
 			}
 		case IntType:
-			v, ok := spec.Value.(float64)
-			if !ok {
+			switch v := spec.Value.(type) {
+			case float64:
+				if math.Trunc(v) != v || v < math.MinInt32 || v > math.MaxInt32 {
+					return nil, errors.Errorf("cannot assign enum value of type 'number' to enum of type 'integer'")
+				}
+				spec.Value = int32(v)
+			case int:
+				spec.Value = int32(v)
+			default:
 				return nil, errorMessage(spec.Value, typ.String())
 			}
-			if math.Trunc(v) != v || v < math.MinInt32 || v > math.MaxInt32 {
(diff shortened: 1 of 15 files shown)


mikhailshilkov:
Could you elaborate on the target usage scenarios? Is this an internal format, or will we suggest users to use it for multi-language components and native providers? If the latter, did we consider providing comparable APIs in our core languages rather than YAML?

pgavlin:
> Could you elaborate on the target usage scenarios? Is this an internal format, or will we suggest users to use it for multi-language components and native providers?

To clarify: this is an exploratory experiment, not a proposal. If it generates reasonable ideas for improvements, I would suggest that we incorporate those improvements into our existing schema support. The original goal was to make it easier for Pulumi engineers to write schemas for our testbed.

> If the latter, did we consider providing comparable APIs in our core languages rather than YAML?

Doubtful. The goal for core languages is to automatically generate schemas in the most idiomatic way possible (e.g. from a .d.ts file for Node, from tooling comments for Go, from attributes for .NET).

pgavlin:
For example, this work led to this PR, which is a simple improvement that should increase quality-of-life for manually-authored schemas: https://github.com/pulumi/pulumi/pull/7461

pgavlin:
Closing this--not moving forward with this for now.

PR #7457, opened 2021-07-08
Reject unknowns in context.go when not in preview

t0yv0:
# Description

Based on a conversation in https://github.com/pulumi/pulumi/pull/7437#discussion_r665748931 we might prefer to be strict and fail with an exception in the Go SDK when unknowns are being marshaled in a context that does not expect them. Currently the behavior instead is to filter them out of the data structure. There were issues around this such as https://github.com/pulumi/pulumi/pull/7214

There are many other places in the codebase that have the similar problem, but I only attempt to edit context.go here to keep the change small and see this passes CI checks.

## Checklist

### DIFF
--- a/sdk/go/pulumi/context.go
+++ b/sdk/go/pulumi/context.go
@@ -253,10 +253,12 @@ func (ctx *Context) Invoke(tok string, args interface{}, result interface{}, opt
 		resolvedArgsMap = resolvedArgs.ObjectValue()
 	}
 
-	keepUnknowns := ctx.DryRun()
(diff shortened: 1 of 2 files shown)


justinvp (commented):
Being more strict seems reasonable to me. There shouldn't be any unknowns during actual deployments.

@pgavlin, any concerns about switching from `KeepUnknowns = false` to `RejectUnknowns = true` during deployments in the Go SDK?

pgavlin (approved):
This seems reasonable to me. I'll be interested to see if we hit any of these in outer-loop testing.

PR #7461, opened 2021-07-08
[schema] Add `inouts` to the schema JSON.

pgavlin:
These changes add support for new fields, `inouts` and `requiredInouts`,
to support simpler definitions of resources that share property
definitions between their inputs and outputs. This is intended to make
hand-authored schemas less tedious to write.

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -7,6 +7,9 @@
 - [sdk/go] - Support for calling resource methods
   [#7437](https://github.com/pulumi/pulumi/pull/7437)
 
+- [schema] - Support for inputs in schema JSON
+  [#7461](https://github.com/pulumi/pulumi/pull/7461)
+
 ### Bug Fixes
 
 - [codegen/go] - Reimplement strict go enums to be Inputs.
--- a/pkg/codegen/internal/test/testdata/schema/bad-inouts-1.json
+++ b/pkg/codegen/internal/test/testdata/schema/bad-inouts-1.json
@@ -1,0 +1,20 @@
+{
+  "version": "0.0.1",
+  "name": "example",
+  "resources": {
+    "example::Resource": {
+      "inoutProperties": {
+        "foo": {
+          "type": "frob"
+        }
+      },
+      "type": "object"
+    }
+  },
+  "language": {
+    "csharp": {},
+    "go": {},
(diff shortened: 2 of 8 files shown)


pgavlin:
Windows build failure looks like a flake.

mikhailshilkov:
IMO, having both "properties" and "inouts" may also be more confusing for schema consumers and authors. Especially, because type definitions are different from resource definitions.

What should I do if my "inout" is optional as input but required as output?

Have we considered reifying all properties into a single collection `properties` and then having a flag like `readonly: true`, similar to Open API specs?

mikhailshilkov:
As a data point, I think we wouldn't be able to use `inouts` in Azure/Google native providers because the types of properties have to be different: we model input and output types as separate schema types (`Foo` and `FooResponse`) as responses often have more properties than requests.

pgavlin:
> What should I do if my "inout" is optional as input but required as output?

In that case you'd need to define the input and output properties separately as is done today.

> Have we considered reifying all properties into a single collection properties and then having a flag like readonly: true, similar to Open API specs?

Interesting--do you have a link for this?

> As a data point, I think we wouldn't be able to use inouts in Azure/Google native providers because the types of properties have to be different: we model input and output types as separate schema types (Foo and FooResponse) as responses often have more properties than requests.

Yeah, I suspected that this would be the case. It's also the case that this isn't often applicable for TF providers, as input and output types often differ slightly. I expect, though, that this will be enough for simple, hand-written schemas like the ones we use for tests. Though maybe that isn't enough to warrant its inclusion.

mikhailshilkov:
> Interesting--do you have a link for this?

Yes, any Azure API spec, e.g. the one below. The type is used for both requests and responses, it defines `properties`, most of them as both inputs and outputs, but `fqdn` is marked as `readOnly`, so it's an output only:

https://github.com/Azure/azure-rest-api-specs/blob/cb835f4a8f3b32919a7819d49b19a48f4aa1168b/specification/containerservice/resource-manager/Microsoft.ContainerService/stable/2021-05-01/managedClusters.json#L2262-L2307

mikhailshilkov:
This PR is very stale and it doesn't sound like we are going to move this direction any time soon. Closing it for now, it's still referenced from https://github.com/pulumi/home/issues/1649.

PR #7463, opened 2021-07-08
[codegen/test] Optionally skip testcases.

pgavlin:
Add support for skipping testcases when run against specific languages.

### DIFF
--- a/pkg/codegen/internal/test/sdk_driver.go
+++ b/pkg/codegen/internal/test/sdk_driver.go
@@ -4,12 +4,14 @@ import (
 	"path/filepath"
 	"testing"
 
+	"github.com/pulumi/pulumi/pkg/v3/codegen"
 	"github.com/stretchr/testify/assert"
 )
 
 type sdkTest struct {
 	Directory   string
 	Description string
+	Skip        codegen.StringSet
 }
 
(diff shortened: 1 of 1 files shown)


viveklak (approved):

komalali (approved):

pkg/codegen/internal/test/sdk_driver.go
```diff
@@ -53,11 +55,11 @@ var sdkTests = []sdkTest{
...
 	},
-	// TODO[pulumi/pulumi#7072]: Enable test after codegen support for all languages lands.
-	// {
-	// 	Directory:   "simple-methods-schema",
-	// 	Description: "Simple schema with methods",
-	// },
+	{
+		Directory:   "simple-methods-schema",
+		Description: "Simple schema with methods",
+		Skip:        codegen.NewStringSet("dotnet"),
```
justinvp, line 61: I think you'll need to skip `"docs"` as well, or add the test data files for them.

justinvp (approved):

pkg/codegen/internal/test/sdk_driver.go
pgavlin: ah, good catch!

PR #7470, opened 2021-07-08
[codegen] Add type name generation tests.

pgavlin:
The inputs and expected outputs for the tests are encoded using a
schema. Each property present in the schema forms a testcase; the
expected outputs for each language are stored in each property's
`Language` field with the language name "test". Expected outputs can be
regenerated using `PULUMI_ACCEPT`.

### DIFF
--- a/pkg/codegen/internal/test/testdata/types.json
+++ b/pkg/codegen/internal/test/testdata/types.json
@@ -1,0 +1,224 @@
+{
+  "name": "typetests",
+  "version": "0.0.1",
+  "meta": {
+    "moduleFormat": "(.*)"
+  },
+  "config": {},
+  "types": {
+    "typetests::primitives": {
+      "properties": {
+        "boolean": {
+          "type": "boolean",
+          "language": {
+            "test": {
+              "expected": {
+                "dotnet": {
+                  "input": "Input\u003cbool\u003e?",
+                  "plain": "bool?"
+                },
+                "go": {
+                  "input": "pulumi.BoolPtrInput",
+                  "plain": "*bool"
+                },
+                "nodejs": {
+                  "input": "pulumi.Input\u003cboolean\u003e | undefined",
+                  "plain": "boolean | undefined"
+                },
+                "python": {
+                  "input": "Optional[pulumi.Input[bool]]",
+                  "plain": "Optional[bool]"
+                }
+              }
+            }
+          }
+        },
+        "integer": {
+          "type": "integer",
+          "language": {
+            "test": {
+              "expected": {
+                "dotnet": {
+                  "input": "Input\u003cint\u003e?",
+                  "plain": "int?"
+                },
+                "go": {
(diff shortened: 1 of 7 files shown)


pgavlin:
Wanted to get the infra in place ASAP; we can expand the set of tests ad infinitum once this is merged.

pkg/codegen/internal/test/testdata/types.json
```diff
@@ -0,0 +1,224 @@
...
+            }
+          }
+        },
+        "requiredNumber": {
+          "type": "number",
+          "language": {
+            "test": {
+              "expected": {
+                "dotnet": {
+                  "input": "Input\u003cdouble\u003e",
```
t0yv0, line 142: This can be `<` and '>' I think the JSON parser will read it just fine.

pkg/codegen/internal/test/testdata/types.json
```diff
@@ -0,0 +1,224 @@
+{
+  "name": "typetests",
+  "version": "0.0.1",
+  "meta": {
+    "moduleFormat": "(.*)"
+  },
+  "config": {},
+  "types": {
+    "typetests::primitives": {
```
t0yv0, line 9: I can understand what this file asserts, so that is good. Does it cover all primitives? Where do we define all the primitive types, could we link it here?

t0yv0 (approved):
Implementation looks a bit hairy but sounds like we need these tests? Let's ship it.

pkg/codegen/internal/test/testdata/types.json
pgavlin: The escaping was performed automatically by the JSON marshaller--I'll see if I can disable it.

pkg/codegen/internal/test/testdata/types.json
pgavlin: It covers all of the simple primitives. I'll link them in the description for this case.

t0yv0: Awesome thanks.

pkg/codegen/internal/test/testdata/types.json
t0yv0: Perhaps not worth it, just aesthetic.

pgavlin: For reference, see https://golang.org/pkg/encoding/json/#HTMLEscape

I was able to disable the escaping.

pgavlin:
> Implementation looks a bit hairy but sounds like we need these tests? Let's ship it.

I may have gotten a little too cute with the implementation. My guiding principle was "make it easy to add new tests and to regenerate baselines for existing tests", which is why I chose to keep the test case + baseline definitions in JSON. IMO writing types in JSON is easier + clearer than writing them in Go, though it does mean that we can't test types that aren't expressible in a schema.

t0yv0:
I could follow the JSON code. I need to hack more on this Go code to follow it better.

PR #7471, opened 2021-07-09
Respect package import alias for top-level module

jen20:
# Description

This commit modifies the Go code generator to use configured aliases for the root package name of a Go module. This is useful when a version 2 package is present, as it prevents generation of identifiers such as "v2" which were produced by the old logic.

## Checklist

No test coverage appears to exist for this path.


justinvp (approved):
LGTM. Thanks!

PR #7472, opened 2021-07-09
Stop producing pointers for map and slice elements

jen20:
# Description

This commit modifies Go program generation to prevent producing array and slice object elements as pointers in args structures, which fails at runtime and does not make sense in any case. For example, in the case of a type defined like this in schema:

```json
"statements": {
  "type": "array",
  "items": {
    "$ref": "/aws/v4.8.0/schema.json#/types/aws:iam/getPolicyDocumentStatement:getPolicyDocumentStatement"
  }
},
```

The following (which fails at runtime) was produced before this change:

```go
Statements []*iam.GetPolicyDocumentStatement `pulumi:"statements"`
```

And the following is produced after after this change:

```go
Statements []iam.GetPolicyDocumentStatement `pulumi:"statements"`
```

Test expectations are updated accordingly.

## Checklist

### DIFF
--- a/pkg/codegen/go/gen.go
+++ b/pkg/codegen/go/gen.go
@@ -396,15 +396,9 @@ func (pkg *pkgContext) typeStringImpl(t schema.Type, argsType bool) string {
 		return pkg.tokenToEnum(t.Token)
 	case *schema.ArrayType:
 		typ := "[]"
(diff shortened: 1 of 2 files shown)


justinvp (approved):
LGTM. Thanks!

PR #7473, opened 2021-07-09
Fix Node.js intermediate module import path generation

jen20:
# Description

This commit fixes code generation for intermediate module paths to produce valid TypeScript identifiers.

Before this change, the following (non-compilable) import was produced in `./jetstack/certmanager/acme`:

```
import * as jetstack/certmanager/acme/v1alpha2 from "./jetstack/certmanager/acme/v1alpha2";
```

After this change, the following import is produced:

```
import * as v1alpha2 from "./v1alpha2";
```

This example is repeated at each level of the module tree. Test expectations are adjusted to reflect this change.

## Checklist


justinvp (approved):
LGTM. Thanks!

PR #7474, opened 2021-07-09
Update package import handling for Node.js

jen20:
# Description

This pull request includes two related changes - the second builds upon the first.

1. Allow non-pulumi imports for Node.js

  Currently the code generator is assuming that Node.js dependencies are following a naming scheme that is prefixed with `@pulumi/`. If this is not the case the generated import statement is incorrect.

  This commit adds a map `ProviderNameToModuleName` to the language definition that allows you to map the name of the extracted provider of a dependency to a module name that the generator now uses to create the import statement.

2.  Prepend "pulumi" to import names in Node.js SDK

  It is common when writing multi-language components to have a module name which conflicts with a provider name. This can produce unusable code, since you cannot simultaneously import a package as `aws` and have a namespace `aws`, for example.

  This commit makes this situation much less likely, by renaming the imported identifier for providers to `pulumiX` where it would previously have been `x`.

  This has an unfortunate side effect of making the tags in the documentation slightly uglier, since import statements for external packages are now of the form `import * as pulumiAws from "@pulumi/aws"`. I don't see a way to discern whether code generation is for SDKs vs examples however, and short of plumbing that through, I don't see a way around this, so test expectations are updated accordingly.

## Checklist

Not a user-facing change.

### DIFF
--- a/pkg/codegen/nodejs/gen.go
+++ b/pkg/codegen/nodejs/gen.go
@@ -32,6 +32,7 @@ import (
 	"unicode"
 
 	"github.com/pkg/errors"
+
 	"github.com/pulumi/pulumi/pkg/v3/codegen"
 	"github.com/pulumi/pulumi/pkg/v3/codegen/schema"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
@@ -144,6 +145,7 @@ func (mod *modContext) objectType(pkg *schema.Package, tok string, input, args, 
 
 	namingCtx, pkgName, external := mod.namingContext(pkg)
 	if external {
+		pkgName = fmt.Sprintf("pulumi%s", title(pkgName))
 		root = "types.output."
 		if input {
 			root = "types.input."
@@ -165,6 +167,10 @@ func (mod *modContext) objectType(pkg *schema.Package, tok string, input, args, 
 func (mod *modContext) resourceType(r *schema.ResourceType) string {
 	if strings.HasPrefix(r.Token, "pulumi:providers:") {
 		pkgName := strings.TrimPrefix(r.Token, "pulumi:providers:")
+		if pkgName != mod.pkg.Name {
+			pkgName = fmt.Sprintf("pulumi%s", title(pkgName))
+		}
+
 		return fmt.Sprintf("%s.Provider", pkgName)
 	}
 
@@ -172,7 +178,10 @@ func (mod *modContext) resourceType(r *schema.ResourceType) string {
 	if r.Resource != nil {
 		pkg = r.Resource.Package
 	}
-	namingCtx, pkgName, _ := mod.namingContext(pkg)
+	namingCtx, pkgName, external := mod.namingContext(pkg)
+	if external {
+		pkgName = fmt.Sprintf("pulumi%s", title(pkgName))
+	}
 
 	modName, name := namingCtx.tokenToModName(r.Token), tokenToName(r.Token)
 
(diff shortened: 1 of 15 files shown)


pkg/codegen/nodejs/gen.go
```diff
@@ -1000,7 +1009,15 @@ func (mod *modContext) getTypeImportsForResource(t schema.Type, recurse bool, ex
...
-			externalImports.Add(fmt.Sprintf("import * as %[1]s from \"@pulumi/%[1]s\";", pkg))
+			var nodePackageInfo NodePackageInfo
+			if languageInfo, hasLanguageInfo := mod.pkg.Language["nodejs"]; hasLanguageInfo {
+				nodePackageInfo = languageInfo.(NodePackageInfo)
+			}
+			if imp, ok := nodePackageInfo.ProviderNameToModuleName[pkg]; ok {
+				externalImports.Add(fmt.Sprintf("import * as %[1]s from \"%[2]s\";", fmt.Sprintf("pulumi%s", title(pkg)), imp))
+			} else {
+				externalImports.Add(fmt.Sprintf("import * as %[1]s from \"@pulumi/%[2]s\";", fmt.Sprintf("pulumi%s", title(pkg)), pkg))
+			}
```
justinvp, line 1020: Nit: The index args aren't necessary, right?

```suggestion
			if imp, ok := nodePackageInfo.ProviderNameToModuleName[pkg]; ok {
				externalImports.Add(fmt.Sprintf("import * as %s from \"%s\";", fmt.Sprintf("pulumi%s", title(pkg)), imp))
			} else {
				externalImports.Add(fmt.Sprintf("import * as %s from \"@pulumi/%s\";", fmt.Sprintf("pulumi%s", title(pkg)), pkg))
			}
```

pkg/codegen/nodejs/gen.go
```diff
@@ -1012,7 +1029,7 @@ func (mod *modContext) getTypeImportsForResource(t schema.Type, recurse bool, ex
 		// If it's from another package, add an import for the external package.
 		if t.Resource != nil && t.Resource.Package != mod.pkg {
 			pkg := t.Resource.Package.Name
-			externalImports.Add(fmt.Sprintf("import * as %[1]s from \"@pulumi/%[1]s\";", pkg))
+			externalImports.Add(fmt.Sprintf("import * as %[1]s from \"@pulumi/%[2]s\";", fmt.Sprintf("pulumi%s", title(pkg)), pkg))
```
justinvp, line 1032: Nit: Same here

```suggestion
			externalImports.Add(fmt.Sprintf("import * as %s from \"@pulumi/%s\";", fmt.Sprintf("pulumi%s", title(pkg)), pkg))
```

justinvp, line 1032: Shouldn't this use `ProviderNameToModuleName` as well?

justinvp (commented):
@jen20, some questions/comments, but LGTM otherwise.

pkg/codegen/nodejs/gen.go
jen20: They're not strictly necessary - they were carried over from somewhere else. They can be removed.

pkg/codegen/nodejs/gen.go
jen20: Good question - that hasn't proven necessary in practice though it is possible there are cases where it matters. I'll take a second pass at this and update the PR tomorrow.

jen20: Yup, you're right - addressed in 88c9fdf.

jen20: Addressed in 88c9fdf.

pkg/codegen/nodejs/gen.go
jen20: Addressed in 88c9fdf.

justinvp (approved):
LGTM. Thanks!

Issue #7454, opened 2021-07-08
Panic when generating dotnet SDK for pulumi-aws


komalali:
Missed this on my earlier review but it looks like the [fix PR](https://github.com/pulumi/pulumi/pull/7451) is causing a panic in pulumi-aws when running `make build_dotnet` - I was able to replicate this locally as well. 
https://github.com/pulumi/pulumi/runs/3013744300#step:8:4796

Full trace:
```
panic: unexpected type <nil>

goroutine 1 [running]:
github.com/pulumi/pulumi/pkg/v3/codegen/dotnet.(*modContext).typeString(0xc00a0d01c0, 0x0, 0x0, 0x8813fc0, 0x6, 0xc000000001, 0x8815db4, 0x6)
        /Users/komal/go/pkg/mod/github.com/pulumi/pulumi/pkg/v3@v3.6.2-0.20210708031642-ccae7c80f8c9/codegen/dotnet/gen.go:420 +0x1d90
github.com/pulumi/pulumi/pkg/v3/codegen/dotnet.(*modContext).unionTypeString(0xc00a0d01c0, 0xc00799fdc0, 0x8813fc0, 0x6, 0x1, 0xc00e000101, 0x8815db4)
        /Users/komal/go/pkg/mod/github.com/pulumi/pulumi/pkg/v3@v3.6.2-0.20210708031642-ccae7c80f8c9/codegen/dotnet/gen.go:266 +0x1ba
github.com/pulumi/pulumi/pkg/v3/codegen/dotnet.(*modContext).typeString(0xc00a0d01c0, 0x9461c38, 0xc00799fdc0, 0x8813fc0, 0x6, 0x1, 0x10edece, 0xc0079a980f)
        /Users/komal/go/pkg/mod/github.com/pulumi/pulumi/pkg/v3@v3.6.2-0.20210708031642-ccae7c80f8c9/codegen/dotnet/gen.go:398 +0x112d
github.com/pulumi/pulumi/pkg/v3/codegen/dotnet.(*modContext).typeString(0xc00a0d01c0, 0x9461b48, 0xc00ce8c560, 0x8813fc0, 0x6, 0x1, 0x2, 0x0)
        /Users/komal/go/pkg/mod/github.com/pulumi/pulumi/pkg/v3@v3.6.2-0.20210708031642-ccae7c80f8c9/codegen/dotnet/gen.go:319 +0x127f
[...]

PR #7475, opened 2021-07-09
[codegen] Add more type tests.

pgavlin:
- Add another regression test for #7454
- Add tests for singly-nested plain collections

### DIFF
--- a/pkg/codegen/internal/test/testdata/types.json
+++ b/pkg/codegen/internal/test/testdata/types.json
@@ -485,6 +485,40 @@
         "requiredString"
       ]
     },
+    "typetests:regressions:inputObject": {
+      "description": "Regression test for bad input types for {\"type\": \"object\"}",
+      "properties": {
+        "object": {
+          "type": "object",
+          "additionalProperties": {
+            "type": "string"
+          },
+          "language": {
+            "test": {
+              "expected": {
+                "dotnet": {
+                  "input": "InputMap<string>",
+                  "plain": "ImmutableDictionary<string, string>?"
+                },
+                "go": {
+                  "input": "pulumi.StringMapInput",
+                  "plain": "map[string]string"
+                },
+                "nodejs": {
+                  "input": "pulumi.Input<{[key: string]: pulumi.Input<string>}> | undefined",
+                  "plain": "{[key: string]: string} | undefined"
+                },
+                "python": {
+                  "input": "Optional[pulumi.Input[Mapping[str, pulumi.Input[str]]]]",
+                  "plain": "Optional[Mapping[str, str]]"
+                }
+              }
+            }
+          }
+        }
+      },
+      "type": "object"
+    },
     "typetests:regressions:pulumi/pulumi/7454": {
       "description": "Regression test for pulumi/pulumi#7454",
       "properties": {
--- a/pkg/codegen/internal/test/testdata/types.json
+++ b/pkg/codegen/internal/test/testdata/types.json
@@ -6,6 +6,69 @@
   },
   "config": {},
   "types": {
+    "typetests::plainCollections": {
+      "description": "Tests for singly-nested plain collections",
+      "properties": {
+        "array": {
+          "type": "array",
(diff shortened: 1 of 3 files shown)


pkg/codegen/schema/schema.go
```diff
@@ -743,7 +743,8 @@ func (pkg *Package) MarshalJSON() (bytes []byte, err error) {
 				continue
 			}
 
-			o, err := pkg.marshalObject(t)
+			// Use the input shape when marshaling in order to get the plain annotations right.
+			o, err := pkg.marshalObject(t.InputShape, false)
```
t0yv0, line 747: These booleany call sites will be tough reading some day. `/* plain */ false` or `marshalObjectInputty`?

pkg/codegen/internal/test/testdata/types.json
```diff
@@ -421,6 +484,43 @@
...
+            "type": "array",
+            "items": {
+              "type": "string"
+            }
+          },
+          "language": {
+            "test": {
+              "expected": {
+                "dotnet": {
+                  "input": "InputList<ImmutableArray<string>>",
```
t0yv0, line 503: I'm having to do some memorization here, is this table driven?

Assuming `InputList<X>` ~ `Input<Seq<X>>`, this is `Input<Seq<Seq<string>>>`. Why is this not `Input<Seq<Input<Seq<Input<str>>>>>` as in the other languages?

 
A(S)

pkg/codegen/internal/test/testdata/types.json
```diff
@@ -421,6 +484,43 @@
...
+                  "plain": "ImmutableArray<ImmutableArray<string>>"
+                },
+                "go": {
+                  "input": "pulumi.StringArrayArrayInput",
```
t0yv0, line 507: I might need a build a reference for these for myself :) I dread remembering which type of values this input carries.

pkg/codegen/schema/schema.go
```diff
@@ -794,8 +795,8 @@ func (pkg *Package) marshalObjectData(
 	}, nil
 }
 
-func (pkg *Package) marshalObject(t *ObjectType) (ComplexTypeSpec, error) {
-	data, err := pkg.marshalObjectData(t.Comment, t.Properties, t.Language)
+func (pkg *Package) marshalObject(t *ObjectType, plain bool) (ComplexTypeSpec, error) {
```
t0yv0, line 798: These changes are not really fixing a bug in prod code but just enabling the test right? This code does not run in prod contest? Could we move it in a file that only compiles for tests, or mark somehow? Always less spooky reviewing PRs that do not edit prod code.

t0yv0 (approved):
Looking good just had some questions.

pkg/codegen/internal/test/testdata/types.json
pgavlin: > Why is this not Input<Seq<Input<Seq<Input<str>>>>> as in python + nodejs?

That's a great question--and one I don't have an answer for. My original changes to reify input types did change the definition to what you'd expect, but that turns out to be a breaking change because (among other things) this no longer compiles:

```csharp
InputList<ImmutableArray<string>> = foo.someNestedArrayInputProperty;
```

pkg/codegen/schema/schema.go
pgavlin: This code could run in prod, but nothing currently uses it. It is generally useful code, so I'd prefer to leave it where it is and add tests for it specifically.

pkg/codegen/internal/test/testdata/types.json
pgavlin: Yup. Can't wait for the day that this is `pulumi.Input[[]pulumi.Input[[]pulumi.Input[string]]]`

t0yv0: Ok so this is consistent with node+py. The .NET is the outlier 🧠 💥 here. I will follow up on that.

PR #7476, opened 2021-07-09
Add issue management Actions

infin8x:
# Description

Adding some issue management automation, including:

- Automatically closing issues that have had the `awaiting-feedback` label for more than 14 days

### DIFF
--- a/.github/workflows/issue-management.yml
+++ b/.github/workflows/issue-management.yml
@@ -2,7 +2,7 @@ name: Issue management
 
 on:
   schedule:
-  - cron: '41 8 * * *'
+  - cron: '12 13 * * *'
 
 jobs:
   stale:
@@ -18,4 +18,5 @@ jobs:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
         only-labels: awaiting-feedback
         stale-issue-message: 'Since we haven''t heard back from you, closing this issue for now. Please feel free to comment with more information and we can get this reopened.'
+        stale-issue-label: awaiting-feedback-stale
         days-before-stale: 14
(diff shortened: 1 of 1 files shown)


.github/workflows/issue-management.yml
```diff
@@ -0,0 +1,21 @@
...
+    permissions:
+      issues: write
+      pull-requests: write
+
+    steps:
+    - uses: actions/stale@v3
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        only-labels: awaiting-feedback
+        stale-issue-message: 'Since we haven''t heard back from you, closing this issue for now. Please feel free to comment with more information and we can get this reopened.'
```
emiliza, line 20: Should we also start adopting `stale-issue-label` to better track which issues get auto-closed out?

infin8x: Ah, that's probably a good idea. Perhaps `awaiting-feedback-stale`?

.github/workflows/issue-management.yml
```diff
@@ -0,0 +1,21 @@
+name: Issue management
+
+on:
+  schedule:
+  - cron: '41 8 * * *'
```
emiliza, line 5: Curious about the thoughts about this time, I can get behind 8:41am pst but these crons observe utc, so we should adjust this if that was the intent (Unless you really want 1am pst?)

infin8x: ah, Actions started doing that in template workflows because they were DDoSing themselves by setting every workflow to run at, 9am Pacific I think 😂. Lemme set to something early morning Pacific.

emiliza (approved):

PR #7479, opened 2021-07-09
Use variable over placeholder for Python version

jen20:
This commit modifies the generation of `setup.py` to use Python variables as the source for the package version and plugin version instead of placeholder strings. This has the effect of making the packages installable via the `-e` flag directly from their source directory rather than requiring a build step, which is useful while developing a plugin and examples in tandem.

## Checklist

Not a user-facing change.


justinvp (approved):
LGTM, Thanks!

@stack72, FYI, this is the change that will require `Makefile` updates for all our providers.

PR #7480, opened 2021-07-09
Add newline to generated package.json files

jen20:
This commit adds a newline to the end of the package.json files generated by Pulumi codegen, such that they can be installed in place without modification.

## Checklist

Not a user-facing change.


justinvp (approved):
LGTM. Thanks!

Issue #3767, opened 2020-01-17
Python: follow up on type issues


EvanBoyle:
There were a few type issues discovered as a part of enabling mypy in #3710 and #3758. 

1. Follow up on the encoding of `Unknown` and `None` WRT `Output`: https://github.com/pulumi/pulumi/pull/3758#discussion_r367570819
2. Label `T` in Output as covariant: https://github.com/pulumi/pulumi/pull/3758#discussion_r367579715. If I try enabling this, mypy complains about covariant parameters in `.apply`. This needs some further investigation.

lukehoban:
> If I try enabling this, mypy complains about covariant parameters in .apply.

Curious - which lines of code triggered these errors?

EvanBoyle:
Another one to look into: https://github.com/pulumi/pulumi/pull/3758/files/9008ac25d834e04a643a1d96b45d6122183af1be#r368084010

lukehoban:
We've made many improvements to these types - and remaining improvements are not "critical" and easy to make incrementally.

PR #7483, opened 2021-07-10
Make Output[T] covariant in Python SDK

### LINKED ISSUE #6843: Consider making Output[T] covariant in python mypy
Original user issue reported as follows:

```
# mypy .
helper_k8s.py:193: error: Argument "string_data" to "Secret" has incompatible type "Output[Mapping[str, str]]"; expected "Union[Mapping[str, Union[str, Awaitable[str], Output[str]]], Awaitable[Mapping[str, Union[str, Awaitable[str], Output[str]]]], Output[Mapping[str, Union[str, Awaitable[str], Output[str]]]], None]"
Found 1 error in 1 file (checked 7 source files)                                                                                                                                                              
~4m-23.0s
```

The user expects `Output[Mapping[str, str]]` to be a subtype of `Output[S]` since `Mapping[str, str]` is a subtype of `S`, with `S = Mapping[str, Union[str, Awaitable[str], Output[str]]]`. That is to say, the user expects the `Output` to be covariant. The workaround is to apply an identity lambda, so `x.apply(lambda v: v)` instead of `x`. Playground listing below:

```python
from pulumi import Output
from typing import Mapping, Union, Awaitable, Generic, TypeVar


S = Mapping[str, Union[str, Awaitable[str], Output[str]]]


def f1(x: Mapping[str, str]) -> S:
    return x


# def f2(x: Output[Mapping[str, str]]) -> Output[S]:
#     return x
#
# error: Incompatible return value type (got "Output[Mapping[str, str]]", expected "Output[Mapping[str, Union[str, Awaitable[str], Output[str]]]]")
# note: Perhaps you need a type annotation for "x"? Suggestion: "Output[Mapping[str, Union[str, Awaitable[str], Output[str]]]]"


def f3(x: Output[Mapping[str, str]]) -> Output[S]:
    return x.apply(lambda v: v)


T1 = TypeVar('T1')
T2 = TypeVar('T2', contravariant=True)
T3 = TypeVar('T3', covariant=True)


class InvariantBox(Generic[T1]):
    pass


class ContravariantBox(Generic[T2]):
    pass


class CovariantBox(Generic[T3]):
    pass


# def f4(x: InvariantBox[Mapping[str, str]]) -> InvariantBox[S]:
#     return x
#
# error --^


# def f5(x: ContravariantBox[Mapping[str, str]]) -> ContravariantBox[S]:
#     return x
#
# error --^


def f6(x: CovariantBox[Mapping[str, str]]) -> CovariantBox[S]:
    return x
``` 

## Affected feature

Usability of k8s libs.

## Discussion

There might be far-reaching implications of marking Output as a covariant, so it might not be feasible, or else not a good idea.

## References

- https://github.com/pulumi/pulumi/issues/3767
- https://github.com/pulumi/pulumi/pull/3758#discussion_r367579715

benesch:
In short, this allows subtyping to correctly "propagate" through an Output;
if Cow is a subtype of Animal, this commit makes it so Output[Cow] is
treated as a subtype of Output[Animal].

Without this change, users of the Python SDK occasionally contend with a
confusing error message that is resolved by adding a call to
`.apply(lambda x: x)` to satisfy mypy.

Touches #3767.
Fix #6843.

cc @lukehoban @t0yv0


orionstudt:
If this is approved I'd like to do similar in the dotnet SDK :) Similarly, if this is approved, is there any reason to not do the same in `Input<T>`?

benesch:
In Python, Input[T] is already covariant as a result of its definition
as Union[T, Awaitable[T], 'Output[T]'].

benesch:
On Wed, Jul 14, 2021 at 11:00 AM Josh Studt ***@***.***>
wrote:

> If this is approved I'd like to do similar in the dotnet SDK :) Similarly,
> if this is approved, is there any reason to not do the same in Input<T>?
>
In Python, Input[T] is already covariant as a result of its definition as
Union[T, Awaitable[T], 'Output[T]'].

orionstudt:
@benesch Makes sense for python. I don't think it is already covariant in the dotnet SDK. Sorry I deleted my comment though, thought it made more sense to make it on the issue instead of your PR

pgavlin (approved):
LGTM.

Should we also add a contravariant type variable for `Inputs` in a future change? cc @lukehoban @t0yv0 @komalali @justinvp

t0yv0:
I was slightly scared to recommend going ahead with this as I don't feel I have a good handle on all the implications, but I guess we will find out now that it is in :)

`Input[T]` and `Output[T]` should both have the same variance (if any) imo, that'd be covariant, because T occurs in a positive position in both (contrast with a negative position in a function type e.g. `T -> int`).

lukehoban:
> I was slightly scared to recommend going ahead with this as I don't feel I have a good handle on all the implications, but I guess we will find out now that it is in :)

Are we running mypy as part of our build to catch any possible impacts of this?

benesch:
> `Input[T]` and `Output[T]` should both have the same variance (if any) imo, that'd be covariant, because T occurs in a positive position in both (contrast with a negative position in a function type e.g. `T -> int`).

Agree! `pulumi.Input[T]` is already working correctly (covariantly) by my read.

> Are we running mypy as part of our build to catch any possible impacts of this?

The python/sdk library definitely gets a mypy run via the `lint` target. Not sure about downstream providers though.

pgavlin:
> Agree! pulumi.Input[T] is already working correctly (covariantly) by my read.

Perfect. Covariance and contravariance is something I've always struggled to keep straight.

justinvp:
Note that we do look at the type hints at runtime and it was necessary to pass a definition for `T` to `get_type_hints` for it to work.

https://github.com/pulumi/pulumi/blob/0bcca3883e47bc99daf566ae6cd66281e389c378/sdk/python/lib/pulumi/_types.py#L725-L728

I think this was only needed for `Input[T]` and `InputType[T]` and was unrelated to `Output[T]`. The existing tests should cover us here, but I'm curious to see how running the examples with this change fairs. We may end up having to add a definition for `T_co` in the same way.

Issue #7367, opened 2021-06-25
Deteriorated IDE experience for python (autocomplete)


komalali:
We recently made a change, specifically https://github.com/pulumi/pulumi/pull/6827 that has led to a bit of a deteriorated IDE experience. We also don't have `class` docstrings, only `__init__` docstrings, which causes a bit of a weird experience in VSCode. The issues are detailed below.

#### Module Autocomplete

If we `import pulumi_aws as aws`, autocomplete appears to work fine.

<img width="600" alt="Screen Shot 2021-06-25 at 11 21 18 AM" src="https://user-images.githubusercontent.com/24326455/123455955-e9945800-d5a7-11eb-97bf-73b741e9c00a.png">
<img width="652" alt="Screen Shot 2021-06-25 at 11 21 27 AM" src="https://user-images.githubusercontent.com/24326455/123455961-eac58500-d5a7-11eb-9dfb-142f96298cb3.png">
<img width="660" alt="Screen Shot 2021-06-25 at 11 21 50 AM" src="https://user-images.githubusercontent.com/24326455/123455969-ec8f4880-d5a7-11eb-8c66-9028bbc21ab4.png">

On the other hand, if we do `from pulumi_aws import iam`, we get the initial autocomplete for `iam`, but no further autocomplete for the module itself.

<img width="725" alt="Screen Shot 2021-06-25 at 11 20 42 AM" src="https://user-images.githubusercontent.com/24326455/123456038-003aaf00-d5a8-11eb-8f58-b7edf421b2f9.png">
<img width="594" alt="Screen Shot 2021-06-25 at 11 20 59 AM" src="https://user-images.githubusercontent.com/24326455/123456055-06309000-d5a8-11eb-9ae2-1525cb3b59a4.png">

This behavior seems related to the changes in https://github.com/pulumi/pulumi/pull/6827

##### Module Autocomplete

If we `import pulumi_aws as aws`, we don't get any autocomplete for the modules, only the direct exports from the parent module. Again, I would attribute this to the changes in https://github.com/pulumi/pulumi/pull/6827

<img width="712" alt="Screen Shot 2021-06-25 at 11 28 25 AM" src="https://user-images.githubusercontent.com/24326455/123464050-eef6a000-d5b1-11eb-9fb6-6f9f66fc7c3d.png">

If we continue on, we do get autocomplete for the exports in `iam` itself.
[...]

PR #7487, opened 2021-07-12
Fix IDE completion issues for lazy imports

t0yv0:
# Description

Fixes  module completion part of #7367

The issues with sub-package discovery are fixed for VSCode and PyCharm.

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -2,3 +2,7 @@
 
 ### Bug Fixes
 
+- [codegen/python] - Recover good IDE completion experience over
+  module imports that was compromised when introducing the lazy import
+  optimization.
+  [#7487](https://github.com/pulumi/pulumi/pull/7487)
--- a/pkg/codegen/internal/test/testdata/nested-module-thirdparty/python/foo_bar/__init__.py
+++ b/pkg/codegen/internal/test/testdata/nested-module-thirdparty/python/foo_bar/__init__.py
@@ -9,7 +9,8 @@ from .provider import *
 
 # Make subpackages available:
 if typing.TYPE_CHECKING:
-    import foo_bar.deeply as deeply
+    import foo_bar.deeply as __deeply
+    deeply = __deeply
 else:
     deeply = _utilities.lazy_import('foo_bar.deeply')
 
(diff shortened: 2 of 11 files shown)


justinvp (approved):
I haven't had a chance to try this out myself, but I'll take your word for it. LGTM.

The underscore prefixed names (e.g. `__<whatever>`) don't show up in completion drop downs, right?

komalali (approved):
I would **love** to see some screenshots of before/after as part of fixes like this since it's one of the few ways to test this (aside from running it locally myself). But yes, as @justinvp said I'll take your word that it works 😄

t0yv0:
> The underscore prefixed names (e.g. __<whatever>) don't show up in completion drop downs, right?

They do in PyCharm but not in VSCode :/

I'll try to add screenshots here, and for completeness try the .pyi solution also. The currently suggested hack works, but I've no idea why it works, so it might stop working randomly as the IDEs update.

komalali:
>They do in PyCharm but not in VSCode :/

Things that might help:
* Using just one underscore `_iam` pycharm *should* deprioritize it if not remove it completely.
* Declaring `__all__` can help exclude any unwanted exports.

t0yv0:
I've packaged this up in https://github.com/t0yv0/pulumi-7367/tree/main/after-7487 so you can try it out easily.

t0yv0:
I think I'll check this in when tests pass (for lack of time pursuing the better fix), but I backlogged https://github.com/pulumi/pulumi/issues/7690

t0yv0:
Re:

```
Things that might help:

Using just one underscore _iam pycharm should deprioritize it if not remove it completely.
Declaring __all__ can help exclude any unwanted exports
```

It tried `_iam` but it seems to show up about the same as `__iam` in pycharm. Delcaring `__all__` sounds a bit complicated - just as complicated as using pyi files, we need to know exactly what we're exporting, statically. Right now the export set is somewhat dynamic (reexport this and that).

komalali:
I pulled this down locally and gave it a little run-through. The `__iam` is sufficiently deprioritized that it's not an issue (shows up last in the list). The overall experience in pycharm and vscode is much improved from where it was before this change. Thanks for fixing this @t0yv0!

t0yv0:
Thanks for verifying!

justinvp:
I cloned your test repo and tried it out with VS Code and it looked great. Thanks for fixing this up!

PR #7488, opened 2021-07-12
Add methods lifecycle tests

justinvp:
Add inner-loop tests for methods in `pkg/engine/lifecycletest`

Part of #7072

### DIFF
--- a/pkg/engine/lifeycletest/pulumi_test.go
+++ b/pkg/engine/lifeycletest/pulumi_test.go
@@ -2192,3 +2192,167 @@ func TestComponentOutputs(t *testing.T) {
 	}
 	p.Run(t, nil)
 }
+
+// Test calling a method.
+func TestSingleComponentMethodDefaultProviderLifecycle(t *testing.T) {
+	var urn resource.URN
+
+	loaders := []*deploytest.ProviderLoader{
+		deploytest.NewProviderLoader("pkgA", semver.MustParse("1.0.0"), func() (plugin.Provider, error) {
+			construct := func(monitor *deploytest.ResourceMonitor,
+				typ, name string, parent resource.URN, inputs resource.PropertyMap,
+				options plugin.ConstructOptions) (plugin.ConstructResult, error) {
+
+				var err error
+				urn, _, _, err = monitor.RegisterResource(tokens.Type(typ), name, false, deploytest.ResourceOptions{
+					Parent:  parent,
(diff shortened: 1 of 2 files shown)


pgavlin (approved):

pkg/engine/lifeycletest/pulumi_test.go
```diff
@@ -2192,3 +2192,167 @@ func TestComponentOutputs(t *testing.T) {
 	}
 	p.Run(t, nil)
 }
+
+// Test calling a method.
+func TestSingleComponentMethodDefaultProviderLifecycle(t *testing.T) {
+	var urn resource.URN
```
t0yv0, line 2198: This trips me up why this would be a shared variable, just abbreviating for convenience? 

My expectation would be that executing Call would identify the resource instance somehow by passing a unique identifier such as URN somewhere, instead we have:

```
	outs, _, _, err := monitor.Call("pkgA:m:typA/methodA", resource.PropertyMap{
			"name": resource.NewStringProperty("Alice"),
		}, "", "")
```

And the mock ProviderLoader relies on the "global" `var urn` to find the resource. That deviates from the usage in actual prod code in this manner, or am I misunderstanding something? Thanks.

pkg/engine/lifeycletest/pulumi_test.go
```diff
@@ -2192,3 +2192,167 @@ func TestComponentOutputs(t *testing.T) {
...
+				return plugin.ConstructResult{
+					URN:     urn,
+					Outputs: outs,
+				}, nil
+			}
+
+			call := func(monitor *deploytest.ResourceMonitor, tok tokens.ModuleMember, args resource.PropertyMap,
+				info plugin.CallInfo, options plugin.CallOptions) (plugin.CallResult, error) {
+
+				_, _, _, err := monitor.RegisterResource("pkgA:m:typC", "resA", true, deploytest.ResourceOptions{
```
t0yv0, line 2323: It's probably fine but I don't see asserts that resA is registered with child-parent being recognized. 

It seemed to me a bit of a pain to test so I was hacking test helpers along these lines:

https://github.com/pulumi/pulumi/pull/7411/files#diff-e703abf84e2aadde2ed61f393d3856f324b84ace6eee460f8a3d55084c9b262bR577

If this resonates, I can lift that out of that PR and try to add?

PR #7497, opened 2021-07-13
[codegen/nodejs] Fix required input annotations.

pgavlin:
Ensure that required input properties that are _not_ mentioned in the
NodeJS-specific required inputs metadata are made optional.


pgavlin:
The majority of the adds in this change are test-related. The fix itself is very small, and is confined to `pkg/codegen/nodejs/gen.go`.

lblackstone (approved):

komalali (approved):

PR #7499, opened 2021-07-13
Emit To[ElementType]PtrOutput methods for go enum output types

### LINKED ISSUE #7496: [codegen/go] - Emit To[Type]PtrOutput methods for enum output types

komalali:
# Description

Fixes #7496 

## Checklist

### DIFF
--- a/pkg/codegen/go/gen.go
+++ b/pkg/codegen/go/gen.go
@@ -813,7 +813,7 @@ func (pkg *pkgContext) genEnumType(w io.Writer, name string, enumType *schema.En
 	inputType := pkg.inputType(enumType)
 	pkg.genEnumInputFuncs(w, name, enumType, elementType, inputType)
 
-	pkg.genEnumOutputTypes(w, name)
+	pkg.genEnumOutputTypes(w, name, elementType, goElementType)
 	pkg.genEnumInputTypes(w, name, enumType, goElementType)
 
 	details := pkg.detailsForType(enumType)
@@ -864,7 +864,9 @@ func (pkg *pkgContext) genEnumType(w io.Writer, name string, enumType *schema.En
 	return nil
 }
 
-func (pkg *pkgContext) genEnumOutputTypes(w io.Writer, name string) {
+func (pkg *pkgContext) genEnumOutputTypes(w io.Writer, name, elementType, goElementType string) {
+	asFuncName := Title(strings.Replace(elementType, "pulumi.", "", -1))
+
 	fmt.Fprintf(w, "type %sOutput struct{ *pulumi.OutputState }\n\n", name)
 	genOutputMethods(w, name, name, false)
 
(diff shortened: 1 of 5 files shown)


pkg/codegen/go/gen.go
```diff
@@ -782,6 +782,7 @@ func (pkg *pkgContext) genEnumType(w io.Writer, name string, enumType *schema.En
 	case "number":
 		goElementType = "float64"
 	}
+	asFuncName := Title(strings.Replace(elementType, "pulumi.", "", -1))
```
pgavlin, line 785:
```suggestion
	asFuncName := strings.TrimPrefix(elementType, "pulumi.")
```

I don't think we need the call to `Title` here, do we? If we expect `elementType` to always be `pulumi.Foo`, then the result of trimming the prefix will always be exported. Or are there cases where we don't expect to see that prefix?

pkg/codegen/go/gen.go
```diff
@@ -893,6 +915,20 @@ func (pkg *pkgContext) genEnumOutputTypes(w io.Writer, name string) {
...
+	fmt.Fprintf(w, "return o.To%sPtrOutputWithContext(context.Background())\n", asFuncName)
+	fmt.Fprint(w, "}\n\n")
+
+	fmt.Fprintf(w, "func (o %[1]sPtrOutput) To%[2]sPtrOutputWithContext(ctx context.Context) %[3]sPtrOutput {\n", name, asFuncName, elementType)
+	fmt.Fprintf(w, "return o.ApplyTWithContext(ctx, func(_ context.Context, e *%s) *%s {\n", name, goElementType)
+	fmt.Fprintf(w, "if e == nil {\n")
+	fmt.Fprintf(w, "return nil\n")
+	fmt.Fprintf(w, "}\n")
+	fmt.Fprintf(w, "s := %s(*e)\n", goElementType)
+	fmt.Fprintf(w, "return &s\n")
```
pgavlin, line 928: tiny nit: 
```suggestion
	fmt.Fprintf(w, "v := %s(*e)\n", goElementType)
	fmt.Fprintf(w, "return &v\n")
```

pkg/codegen/go/gen.go
```diff
@@ -878,6 +879,27 @@ func (pkg *pkgContext) genEnumOutputTypes(w io.Writer, name string) {
...
+	fmt.Fprint(w, "}\n\n")
+
+	fmt.Fprintf(w, "func (o %[1]sOutput) To%[2]sPtrOutput() %[3]sPtrOutput {\n", name, asFuncName, elementType)
+	fmt.Fprintf(w, "return o.To%sPtrOutputWithContext(context.Background())\n", asFuncName)
+	fmt.Fprint(w, "}\n\n")
+
+	fmt.Fprintf(w, "func (o %[1]sOutput) To%[2]sPtrOutputWithContext(ctx context.Context) %[3]sPtrOutput {\n", name, asFuncName, elementType)
+	fmt.Fprintf(w, "return o.ApplyTWithContext(ctx, func(_ context.Context, e %s) *%s {\n", name, goElementType)
+	fmt.Fprintf(w, "s := %s(e)\n", goElementType)
+	fmt.Fprintf(w, "return &s\n")
```
pgavlin, line 899: tiny nit:
```suggestion
	fmt.Fprintf(w, "v := %s(e)\n", goElementType)
	fmt.Fprintf(w, "return &v\n")
```

pgavlin (approved):
Just a few nits; LGTM otherwise.

pkg/codegen/go/gen.go
komalali: You're probably right! This was just moving around some existing code.

PR #7500, opened 2021-07-13
Fix rendering of diffs for resource without DetailedDiffs

### LINKED ISSUE #7486: pulumi 3.6.1 doesn't show diffs
I use Pulumi dynamic provider and before 3.6.1 (I was using 3.6.0) `pulumi up` was showing diffs like the following:

```
$ pulumi up
Previewing update (dev):
     Type                               Name                                          Plan       Info
     pulumi:pulumi:Stack                my-dev
 ~   └─ pulumi-nodejs:dynamic:Resource  myorg:users:User:Admin    update     [diff: ~firstName]

Resources:
    ~ 1 to update
    1 unchanged

Do you want to perform this update? details
  pulumi:pulumi:Stack: (same)
    [urn=urn:pulumi:dev::XXX::pulumi:pulumi:Stack::my-dev]
    ~ pulumi-nodejs:dynamic:Resource: (update)
        [id=myorg:users:User:Admin]
        [urn=urn:pulumi:dev::XXX::pulumi-nodejs:dynamic:Resource::myorg:users:User:Admin]
      ~ user                   : {
          ~ firstName      : "John" => "Joe"
        }
```

However, after upgrading to 3.6.1, the detailed preview doesn't show the diff but the whole existing resource data

This issue has been also raised on Slack: https://pulumi-community.slack.com/archives/C84L4E3N1/p1625741083479000

lukehoban:
The diff rendering logic tests whether the `DetailedDiff` is `nil` to determine whether to use it for rendering or defer to older older supported approaches to computing diffs.

The new logic added in https://github.com/pulumi/pulumi/pull/7226 could lead to replacing a `nil` `DetailedDiff` with an empty `DetailedDiff`, which would make the rendering logic believe that a `DetailedDiff` was present but empty, instead of missing entirely.

This change ensures that we maintain nil-ness of the `DetailedDiff` when we transform it for handling of `replaceOnChanges`.

Also adds tests for this and other cases on `applyReplaceOnChanges`.

Fixes #7486.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@ CHANGELOG
 - [sdk/python] - Fix an issue where dependency keys were incorrectly translates to camelcase
   [#7443](https://github.com/pulumi/pulumi/pull/7443)
 
+- [cli] - Fix rendering of diffs for resource without DetailedDiffs
+  [#7500](https://github.com/pulumi/pulumi/pull/7500)
+
 ## 3.6.0 (2021-06-30)
 
 ### Improvements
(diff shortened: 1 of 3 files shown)


pgavlin (approved):
Good find. LGTM.

lblackstone (approved):

PR #7502, opened 2021-07-13
Set PULUMI_CONFIG_PASSPHRASE for Automation Tests

hossambarakat:
# Description

I have fixed the dotnet Automation.Tests that were failing because of missing PULUMI_CONFIG_PASSPHRASE environment variable.

## Checklist

The change is fixing tests and no impact on existing behaviour

There is no changes related to calling Pulumi Service API


komalali (approved):

t0yv0:
Thanks Komal sorry late for reviewing here. LGTM.

PR #7509, opened 2021-07-13
[codegen/schema] YAML {un,}marshaling support.

pgavlin:
These changes add support for unmarshaling and marshaling package
schemas using YAML instead of JSON. Language-specific data is
canonically JSON. Users of the `*Spec` types will need to update the
types of the the their `Language` values to use the new
`schema.RawMessage` type instead of `json.RawMessage`: the former has
support for YAML while the latter does not.


pgavlin:
Here's the test schema: [pkg/codegen/internal/test/testdata/simple-yaml-schema/schema.yaml](https://github.com/pulumi/pulumi/pull/7509/files#diff-30a7952ac39d0e9e935e51e700556f42713fa82a2eb3ea21ae319a641051ac85).

The rest of the changes are localized to [`schema.go`](https://github.com/pulumi/pulumi/pull/7509/files#diff-39fe226c4621511c7bf43ed09c55cb1949143530e09be475118e17c6d7d61581).

pkg/codegen/schema/schema.go
```diff
@@ -1082,151 +1135,149 @@ type TypeSpec struct {
...
+	Environment []string `json:"environment,omitempty" yaml:"environment,omitempty"`
 	// Language specifies additional language-specific data about the default value.
-	Language map[string]json.RawMessage `json:"language,omitempty"`
+	Language map[string]RawMessage `json:"language,omitempty" yaml:"language,omitempty"`
 }
 
 // PropertySpec is the serializable form of an object or resource property.
 type PropertySpec struct {
-	TypeSpec
+	TypeSpec `yaml:",inline"`
```
komalali, line 1169: What does the `yaml:",inline"` do?

pkg/codegen/internal/test/testdata/simple-yaml-schema/schema.yaml
```diff
@@ -0,0 +1,126 @@
...
+          "$ref": "#/resources/example::Resource"
+    outputs:
+      properties:
+        result:
+          "$ref": "#/resources/example::Resource"
+language:
+  csharp: {}
+  go: {
+    "importBasePath": "github.com/pulumi/pulumi/pkg/v3/codegen/internal/test/testdata/simple-resource-schema/go/example"
+  }
```
komalali, line 124: Could this be:
```suggestion
  go:
    importBasePath: "github.com/pulumi/pulumi/pkg/v3/codegen/internal/test/testdata/simple-resource-schema/go/example"
```

pkg/codegen/internal/test/testdata/simple-yaml-schema/schema.yaml
```diff
@@ -0,0 +1,126 @@
...
+  example::ObjectWithNodeOptionalInputs:
+    properties:
+      foo:
+        type: string
+      bar:
+        type: integer
+    required: [foo]
+    language:
+      nodejs:
+        "requiredInputs": []
```
komalali, line 53: Could this be unquoted? (Trying to see if there's a reason it's expressed with the quotes)
```suggestion
        requiredInputs: []
```

komalali (approved):

pkg/codegen/schema/schema.go
pgavlin: It tells the YAML marshaler that the fields from `TypeSpec` will appear inline in the serialized form of `PropertySpec`. Without this, we'd need something like

```yaml
property:
    TypeSpec:
        type: string
```

pkg/codegen/internal/test/testdata/simple-yaml-schema/schema.yaml
pgavlin: It could. I left it as-is because of this:

> Language-specific data is canonically JSON

In accordance with that statement, language-specific data is always serialized and deserialized as JSON, even if it appears in a YAML document.

pkg/codegen/internal/test/testdata/simple-yaml-schema/schema.yaml
pgavlin: Same answer as above re: canonical representation of language-specific data.

komalali: In that case, shouldn't it be:
```
nodejs: {
	"requiredInputs": []
}
```

PR #7516, opened 2021-07-13
Move docstring to class docstring for Python

t0yv0:
# Description

Improves IDE (PyCharm, VSCode) experience by splitting resource doc strings that were annotating the `__init__` constructor methods into the common part that is now a class docstring, and the remainder part that continues to annotate the `__init__` methods.

Now in PyCharm one can F1 on `aws.iam.Role` and see the docs, and in VSCode one can hover over the class name and similarly see the docs; constructor docs are less busy.

What I am curious about is the impact on doc generation with Sphinx - how can we test/verify that? The comments indicated that this was previously done this way to support Sphinx doc gen.

There remains some tension between what Sphinx wants and what IDEs accept. For example, the RST parameter doc notation is not currently recognized by either IDEs, so parameter doc lookups are failing.

Fixes the class docstring part of #7367 

## Checklist

### DIFF
--- a/pkg/codegen/python/gen.go
+++ b/pkg/codegen/python/gen.go
@@ -1006,6 +1006,9 @@ func (mod *modContext) genResource(res *schema.Resource) (string, error) {
 
 	// Produce a class definition with optional """ comment.
 	fmt.Fprintf(w, "class %s(%s):\n", name, baseType)
+
+	mod.genResourceClassDocstring(w, res)
+
 	if res.DeprecationMessage != "" && mod.compatibility != kubernetes20 {
 		escaped := strings.ReplaceAll(res.DeprecationMessage, `"`, `\"`)
 		fmt.Fprintf(w, "    warnings.warn(\"\"\"%s\"\"\", DeprecationWarning)\n\n", escaped)
@@ -1037,7 +1040,7 @@ func (mod *modContext) genResource(res *schema.Resource) (string, error) {
 	// Emit an __init__ overload that accepts the resource's inputs as function arguments.
 	fmt.Fprintf(w, "    @overload\n")
 	emitInitMethodSignature("__init__")
-	mod.genInitDocstring(w, res, resourceArgsName, false /*argsOverload*/)
+	mod.genResourceInitDocstring(w, res, resourceArgsName, false /*argsOverload*/)
 	fmt.Fprintf(w, "        ...\n")
 
 	// Emit an __init__ overload that accepts the resource's inputs from the args class.
@@ -1050,7 +1053,7 @@ func (mod *modContext) genResource(res *schema.Resource) (string, error) {
 		fmt.Fprintf(w, "                 args: %s,\n", resourceArgsName)
 	}
 	fmt.Fprintf(w, "                 opts: Optional[pulumi.ResourceOptions] = None):\n")
-	mod.genInitDocstring(w, res, resourceArgsName, true /*argsOverload*/)
+	mod.genResourceInitDocstring(w, res, resourceArgsName, true /*argsOverload*/)
 	fmt.Fprintf(w, "        ...\n")
 
 	// Emit the actual implementation of __init__, which does the appropriate thing based on which
@@ -1933,15 +1936,35 @@ func recordProperty(prop *schema.Property, snakeCaseToCamelCase, camelCaseToSnak
 	}
 }
 
+// `genResourceClassDocstring` emits the class docstring for the given
+// resource type.
+//
+// Sphinx (the documentation generator that we use to generate Python
+// docs) does not draw a distinction between documentation comments on
+// the class itself and documentation comments on the `__init__`
+// method of a class. The docs repo instructs Sphinx to concatenate
+// the two together.
+//
+// However, IDEs benefit from class docstrings. In PyCharm, emitting a
+// class docstring enables better `F1` help accessor when hovering
+// over a class name. VSCode experience is similarly improved.
+func (mod *modContext) genResourceClassDocstring(w io.Writer, res *schema.Resource) {
+	b := &bytes.Buffer{}
+
+	// If this resource has documentation, write it at the top of the docstring, otherwise use a generic comment.
+	if res.Comment != "" {
+		fmt.Fprintln(b, codegen.FilterExamples(res.Comment, "python"))
+	} else {
+		fmt.Fprintf(b, "Create a %s resource with the given unique name, props, and options.\n", tokenToName(res.Token))
+	}
+
+	// printComment handles the prefix and triple quotes.
+	printComment(w, b.String(), "        ")
+}
+
 // genInitDocstring emits the docstring for the __init__ method of the given resource type.
 //
-// Sphinx (the documentation generator that we use to generate Python docs) does not draw a
-// distinction between documentation comments on the class itself and documentation comments on the
-// __init__ method of a class. The docs repo instructs Sphinx to concatenate the two together, which
-// means that we don't need to emit docstrings on the class at all as long as the __init__ docstring
-// is good enough.
-//
-// The docstring we generate here describes both the class itself and the arguments to the class's
+// The docstring we generate here describes the arguments to the class's
 // constructor. The format of the docstring is in "Sphinx form":
 //   1. Parameters are introduced using the syntax ":param <type> <name>: <comment>". Sphinx parses this and uses it
 //      to populate the list of parameters for this function.
@@ -1951,16 +1974,9 @@ func recordProperty(prop *schema.Property, snakeCaseToCamelCase, camelCaseToSnak
 //
 // This function does the best it can to navigate these constraints and produce a docstring that
 // Sphinx can make sense of.
-func (mod *modContext) genInitDocstring(w io.Writer, res *schema.Resource, resourceArgsName string, argOverload bool) {
+func (mod *modContext) genResourceInitDocstring(w io.Writer, res *schema.Resource, resourceArgsName string, argOverload bool) {
 	// b contains the full text of the docstring, without the leading and trailing triple quotes.
 	b := &bytes.Buffer{}
-
-	// If this resource has documentation, write it at the top of the docstring, otherwise use a generic comment.
-	if res.Comment != "" {
-		fmt.Fprintln(b, codegen.FilterExamples(res.Comment, "python"))
-	} else {
-		fmt.Fprintf(b, "Create a %s resource with the given unique name, props, and options.\n", tokenToName(res.Token))
-	}
 
 	// All resources have a resource_name parameter and opts parameter.
 	fmt.Fprintln(b, ":param str resource_name: The name of the resource.")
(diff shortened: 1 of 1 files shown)


justinvp (commented):
> What I am curious about is the impact on doc generation with Sphinx - how can we test/verify that? The comments indicated that this was previously done this way to support Sphinx doc gen.

I don't think we need to hold ourselves back for Sphinx doc gen. We originally generated Python-specific docs for all our Python provider SDKs with Sphinx (with customizations) using this: https://github.com/pulumi/docs/blob/master/tools/pydocgen/pydocgen/__init__.py

However, we've now moved away from that and now [generate _resource-oriented_ docs](https://github.com/pulumi/docs/tree/123f85ae7cf03fa4e371c1d9a32b4fec510823f7/tools/resourcedocsgen) for our provider SDKs (that combine docs for all languages into single set of docs).

The only thing we're using Sphinx (our pydocgen tool) for is the core `pulumi` SDK package, `pulumi_policy` (and for some reason `pulumi_terraform`, which should prob move to just resource-oriented docs): https://github.com/pulumi/docs/blob/123f85ae7cf03fa4e371c1d9a32b4fec510823f7/scripts/generate_python_docs.sh#L9-L13

So my suggestion would be we optimize the generated code for provider SDKs to work well with popular IDEs.

In the short term, we still need to keep `pulumi` and `pulumi_policy` working with our Sphinx-based pydocgen tool, but we could consider moving to a more standard way of generating Python package docs (that doesn't necessarily have to fit in perfectly with the rest of our site), if it'd help us improve the IDE experience for these packages.

komalali (commented):
From then downstream codegen https://github.com/pulumi-bot/pulumi-random/compare/91eb5ac..d7fa2ab it looks like the indentation might be off. Also would love to see the tests here (running `PULUMI_ACCEPT=true go test ./...` in codegen/python should do the trick)

komalali:
>RST parameter doc notation is not currently recognized by either IDEs

In my experience, this is the format recognized by JetBrains products, but not by VSCode (prefers google style docstrings). There is however conflict between the two, since Jetbrains products don't parse google-style docstrings.

IntelliJ:
Screen Shot 2021-07-13 at 2 50 35 PM

VSCode:
Screen Shot 2021-07-13 at 2 53 38 PM


I'm also wondering if we should keep the description in both the constructor and the class docstring, since now we don't get the information about the class when using the constructor:

Before:
Screen Shot 2021-07-13 at 2 58 49 PM

After:
Screen Shot 2021-07-13 at 2 56 08 PM

Note: The before is actually still "incorrect". If there is an empty between the description and the parameters list, we get the "most correct" form.
Screen Shot 2021-07-13 at 3 00 51 PM

t0yv0:
Good catch on indent @komalali fixed above.

t0yv0:
I'm not sure about including both places - the Sphinx comment scared me that that'd duplicate something in the docs, and duplicating in source is also not ideal, but we can do it if it helps. I can run a few more experiments in the IDEs tomorrow.

komalali:
>I'm not sure about including both places - the Sphinx comment scared me that that'd duplicate something in the docs

As @justinvp mentioned, we don't use sphinx to generate documentation for the SDKs anymore. You can also test the docs by running `PULUMI_ACCEPT=true go test ./...` in codegen/docs to see exactly what the docs look like.

>duplicating in source is also not ideal

We must do what is required to make this the *most* seamless development scenario for the users. If that comes at the cost of repeating ourselves, I'm okay with it (especially since this is generated code and we aren't actually repeating ourselves).

t0yv0:
This makes sense, mostly, though it would be great to have generated code sizings here, I can see if I can get them. Bloating package size is not free for UX. We've been at a place where azure-native was simply unusable for some users on account of package size (doing a bit better now).

pkg/codegen/python/gen.go
```diff
@@ -1933,15 +1936,35 @@ func recordProperty(prop *schema.Property, snakeCaseToCamelCase, camelCaseToSnak
...
 
-// genInitDocstring emits the docstring for the __init__ method of the given resource type.
+// `genResourceClassDocstring` emits the class docstring for the given
+// resource type.
+//
+// Sphinx (the documentation generator that we use to generate Python
+// docs) does not draw a distinction between documentation comments on
+// the class itself and documentation comments on the `__init__`
+// method of a class. The docs repo instructs Sphinx to concatenate
+// the two together.
```
komalali, line 1946: I think any reference to how sphinx generates these things is now out of date for any of the provider SDK docs. We currently only use sphinx to generate the pulumi and pulumi-policy SDKs (which are all hand-written and unaffected by this code).

komalali:
Yeah that's fair. Would love to see what sort of effect the bloat has on size/speed 🙏🏽

t0yv0:
The one-sentence summary is not where the bloat is, basically the interesting part is including examples in docstrings. For example for iam.Role:

```
class Role(pulumi.CustomResource):
        """
        Provides an IAM role.

        > **NOTE:** If policies are attached to the role via the `iam.PolicyAttachment` resource and you are modifying the role `name` or `path`, the `force_detach_policies` argument must be set to `true` and applied before attempting the operation otherwise you will encounter a `DeleteConflict` error. The `iam.RolePolicyAttachment` resource does not have this requirement.

        > **NOTE:** If you use this resource's `managed_policy_arns` argument or `inline_policy` configuration blocks, this resource will take over exclusive management of the role's respective policy types (e.g., both policy types if both arguments are used). These arguments are incompatible with other ways of managing a role's policies, such as `iam.PolicyAttachment`, `iam.RolePolicyAttachment`, and `iam.RolePolicy`. If you attempt to manage a role's policies by multiple means, you will get resource cycling and/or errors.

        ## Example Usage
        ### Basic Example

        ```python
        import pulumi
        import json
        import pulumi_aws as aws

        test_role = aws.iam.Role("testRole",
            assume_role_policy=json.dumps({
                "Version": "2012-10-17",
                "Statement": [{
                    "Action": "sts:AssumeRole",
                    "Effect": "Allow",
                    "Sid": "",
                    "Principal": {
                        "Service": "ec2.amazonaws.com",
                    },
                }],
            }),
            tags={
                "tag-key": "tag-value",
            })
        ```
        ### Example of Using Data Source for Assume Role Policy

        ```python
        import pulumi
        import pulumi_aws as aws

        instance_assume_role_policy = aws.iam.get_policy_document(statements=[aws.iam.GetPolicyDocumentStatementArgs(
            actions=["sts:AssumeRole"],
            principals=[aws.iam.GetPolicyDocumentStatementPrincipalArgs(
                type="Service",
                identifiers=["ec2.amazonaws.com"],
            )],
        )])
        instance = aws.iam.Role("instance",
            path="/system/",
            assume_role_policy=instance_assume_role_policy.json)
        ```
        ### Example of Exclusive Inline Policies

        This example creates an IAM role with two inline IAM policies. If someone adds another inline policy out-of-band, on the next apply, the provider will remove that policy. If someone deletes these policies out-of-band, the provider will recreate them.

        ```python
        import pulumi
        import json
        import pulumi_aws as aws

        inline_policy = aws.iam.get_policy_document(statements=[aws.iam.GetPolicyDocumentStatementArgs(
            actions=["ec2:DescribeAccountAttributes"],
            resources=["*"],
        )])
        example = aws.iam.Role("example",
            assume_role_policy=data["aws_iam_policy_document"]["instance_assume_role_policy"]["json"],
            inline_policies=[
                aws.iam.RoleInlinePolicyArgs(
                    name="my_inline_policy",
                    policy=json.dumps({
                        "Version": "2012-10-17",
                        "Statement": [{
                            "Action": ["ec2:Describe*"],
                            "Effect": "Allow",
                            "Resource": "*",
                        }],
                    }),
                ),
                aws.iam.RoleInlinePolicyArgs(
                    name="policy-8675309",
                    policy=inline_policy.json,
                ),
            ])
        ```
        ### Example of Removing Inline Policies

        This example creates an IAM role with what appears to be empty IAM `inline_policy` argument instead of using `inline_policy` as a configuration block. The result is that if someone were to add an inline policy out-of-band, on the next apply, the provider will remove that policy.

        ```python
        import pulumi
        import pulumi_aws as aws

        example = aws.iam.Role("example",
            assume_role_policy=data["aws_iam_policy_document"]["instance_assume_role_policy"]["json"],
            inline_policies=[aws.iam.RoleInlinePolicyArgs()])
        ```
        ### Example of Exclusive Managed Policies

        This example creates an IAM role and attaches two managed IAM policies. If someone attaches another managed policy out-of-band, on the next apply, the provider will detach that policy. If someone detaches these policies out-of-band, the provider will attach them again.

        ```python
        import pulumi
        import json
        import pulumi_aws as aws

        policy_one = aws.iam.Policy("policyOne", policy=json.dumps({
            "Version": "2012-10-17",
            "Statement": [{
                "Action": ["ec2:Describe*"],
                "Effect": "Allow",
                "Resource": "*",
            }],
        }))
        policy_two = aws.iam.Policy("policyTwo", policy=json.dumps({
            "Version": "2012-10-17",
            "Statement": [{
                "Action": [
                    "s3:ListAllMyBuckets",
                    "s3:ListBucket",
                    "s3:HeadBucket",
                ],
                "Effect": "Allow",
                "Resource": "*",
            }],
        }))
        example = aws.iam.Role("example",
            assume_role_policy=data["aws_iam_policy_document"]["instance_assume_role_policy"]["json"],
            managed_policy_arns=[
                policy_one.arn,
                policy_two.arn,
            ])
        ```
        ### Example of Removing Managed Policies

        This example creates an IAM role with an empty `managed_policy_arns` argument. If someone attaches a policy out-of-band, on the next apply, the provider will detach that policy.

        ```python
        import pulumi
        import pulumi_aws as aws

        example = aws.iam.Role("example",
            assume_role_policy=data["aws_iam_policy_document"]["instance_assume_role_policy"]["json"],
            managed_policy_arns=[])
        ```

        ## Import

        IAM Roles can be imported using the `name`, e.g.

        ```sh
         $ pulumi import aws:iam/role:Role developer developer_name
        ```
        """
```

VSCode renders the markdown but PyCharm doesn't. In either case the viewport is not optimized for such a wall of text. Could we emit HTTP links instead to the examples?

pkg/codegen/python/gen.go
t0yv0: I'll update the PR tomorrow, thank you!

komalali:
> VSCode renders the markdown but PyCharm doesn't. In either case the viewport is not optimized for such a wall of text. Could we emit HTTP links instead to the examples?

Yeah, agreed that the examples are largely the reason for giant docstrings. And the discrepancy between the two IDEs is pretty silly. The idea to emit HTTP links is intriguing, not sure how I feel about completely removing examples from the source though. Perhaps we can compromise by only putting the full description in the class docstring and then just the generic one-liner description in the constructors?

komalali:
Actually, if the size bloat is considerable, perhaps we can be a little clever here to sustain the full description in both places without actually repeating ourselves. The docstring is just the `__doc__` attribute on the class object (or the constructor object).

Screen Shot 2021-07-13 at 3 32 45 PM

We can just emit the docstring as a variable and then assign it to both the constructor and the class:
```python
_cat_docstring = "I am a cat"

class Cat:
	def __init__(self):
		return

Cat.__doc__ = _cat_docstring
Cat.__init__.__doc__ = _cat_docstring
```

lunaris:
Given how long this has been open, I'd be minded to close since it appears (?) it's not been an issue over the last few years? Or should we rebase and get this finished?

t0yv0:
I never quite finished assessing how things work in IntelliJ (not a user), could be worth another look if it's still a problem. I recall substantial differences in what IntelliJ liked vs what VSCode liked in these annotations.

PR #7523, opened 2021-07-14
Add force flag for RemoveStack

yarinm:
# Description
I'm in the need for forcibly delete certain stacks from the state, even though they have resources in it.
This is relevant for us for kubeneretes stacks that we wish to delete because we are going to destroy the infra stack anyway.

## Checklist

### DIFF
--- a/sdk/go/auto/workspace.go
+++ b/sdk/go/auto/workspace.go
@@ -17,8 +17,10 @@ package auto
 import (
 	"context"
 
+	"github.com/pulumi/pulumi/sdk/v3/go/auto/optremove"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/apitype"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/workspace"
+
 	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
 )
 
@@ -84,7 +86,7 @@ type Workspace interface {
 	// SelectStack selects and sets an existing stack matching the stack name, failing if none exists.
 	SelectStack(context.Context, string) error
 	// RemoveStack deletes the stack and all associated configuration and history.
-	RemoveStack(context.Context, string) error
+	RemoveStack(context.Context, string, ...optremove.Option) error
 	// ListStacks returns all Stacks created under the current Project.
 	// This queries underlying backend and may return stacks not present in the Workspace.
 	ListStacks(context.Context) ([]StackSummary, error)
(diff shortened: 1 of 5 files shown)


EvanBoyle (commented):
Unfortunately this is a breaking change as implemented, which we should try to avoid. 

Adding variadic functional options is usually how we work around this for other methods.

yarinm:
@EvanBoyle changed the PR so it will not be a breaking change

sdk/go/auto/workspace.go
```diff
@@ -109,6 +109,22 @@ type Workspace interface {
 	StackOutputs(context.Context, string) (OutputMap, error)
 }
 
+type RemoveStackOption interface {
+	getRemoveStackFlag() string
+}
+
+func RemoveStackForce() RemoveStackOption {
+	return &removeStackFlagOption{flag: "--force"}
```
EvanBoyle, line 117: This implementation is too specific to the fact that the fact that the automation api uses the CLI. Rather than flag, let's create a specific force option. It might look something along the lines of:

```go
type removeStackOptions {
  Force bool
}

type RemoveStackOptions interface {
	applyRemoveStackOption(*removeStackOptions)
}

type removeStackOption func(*removeStackOptions)

func (o removeStackOption) applyRemoveStackOption(opts *removeStackOptions) {
	o(opts)
}

func Force() RemoveStackOptions {
	return removeStackOption(func(wo *removeStackOptions) {
		wo.Force = true
	})
}
```

In addition, let's extract this option code into a separate package in `sdk/go/auto/optremove` to keep the auto namespace clean (there are likely other commands that `Force` could apply to). This is what we have done for preview, update, refresh, and destroy options: https://github.com/pulumi/pulumi/blob/c3dc2d54abf85da811de0fbaa5e98f22b0f054c8/sdk/go/auto/optup/optup.go#L17

EvanBoyle (commented):
Thanks for making this into a non-breaking change! Just one request to move these options into a subpackage and restructure them slightly.

sdk/go/auto/workspace.go
yarinm: Done

EvanBoyle: Looks like we should be able to clean up the original options and remove them from `workspace.go`. Would you mind doing that?

yarinm: Yeah sorry about that, forgot to clean that up

komalali (changes requested):
Looks like there's a [linting error](https://github.com/pulumi/pulumi/pull/7523/checks?check_run_id=3179186486#step:4:10), the `optremove.go` file needs to be run through gofmt (it's using spaces instead of tabs).

komalali (approved):

EvanBoyle:
@yarinm can you merge in the latest changes from master so we can get a clean test run?

PR #7527, opened 2021-07-14
update gRPC client to grpc-dotnet

JayDouglass:
# Description

Fixes 
- Missing libgrpc dependency for Apple silicon (used by Grpc.Core)

## Checklist

There are already test cases that cover this change

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -4,6 +4,7 @@
 
 - [schema] - Support for unmarshaling/marshaling schemas from/to YAML
   [#7509](https://github.com/pulumi/pulumi/pull/7509)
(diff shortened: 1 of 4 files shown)


mikhailshilkov:
@JayDouglass Thank you for the PR. The change seems to be causing a bunch of gRPC-related test failures, see here: https://github.com/pulumi/pulumi/runs/3121291298?check_suite_focus=true

t0yv0:
Right, for example `dotnet test --filter ConcurrentUpdateExceptionIsThrown`

I poked at it a little bit but could not figure out quickly.

emiliza:
Closing out this PR due to inactivity. I've opened up https://github.com/pulumi/pulumi/issues/8476 to track this work

PR #7529, opened 2021-07-14
Fixes for C# concurrency bugs in 7492

### LINKED ISSUE #7492: [auto/dotnet] - Flakey Automation API tests for C#
Since the PR to [Fix swallowed nested exceptions in inline programs](https://github.com/pulumi/pulumi/pull/7323), we have been seeing increased [flakey tests](https://github.com/pulumi/pulumi/runs/3049702475#step:17:97) for the dotnet automation API tests. The stack trace is below:

```
  Failed Pulumi.Automation.Tests.LocalWorkspaceTests.StackLifecycleInlineProgram [5 s]
  Error Message:
   System.Collections.Generic.KeyNotFoundException : The given key 'System.Threading.Tasks.UnwrapPromise`1[System.Threading.Tasks.VoidTaskResult]' was not present in the dictionary.
  Stack Trace:
     at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
   at Pulumi.Deployment.Runner.<>c__DisplayClass13_0.<<WhileRunningAsync>g__HandleCompletion|0>d.MoveNext() in /_/sdk/dotnet/Pulumi/Deployment/Deployment.Runner.cs:line 147
--- End of stack trace from previous location where exception was thrown ---
   at Pulumi.Deployment.Runner.WhileRunningAsync() in /_/sdk/dotnet/Pulumi/Deployment/Deployment.Runner.cs:line 181
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw(Exception source)
   at Pulumi.Deployment.<>c__DisplayClass63_0.<<RunInlineAsync>b__1>d.MoveNext() in /_/sdk/dotnet/Pulumi/Deployment/Deployment_Inline.cs:line 62
--- End of stack trace from previous location where exception was thrown ---
   at Pulumi.Automation.WorkspaceStack.PreviewAsync(PreviewOptions options, CancellationToken cancellationToken) in /_/sdk/dotnet/Pulumi.Automation/WorkspaceStack.cs:line 422
   at Pulumi.Automation.WorkspaceStack.PreviewAsync(PreviewOptions options, CancellationToken cancellationToken) in /_/sdk/dotnet/Pulumi.Automation/WorkspaceStack.cs:line 445
   at Pulumi.Automation.Tests.LocalWorkspaceTests.StackLifecycleInlineProgram() in /Users/runner/work/pulumi/pulumi/sdk/dotnet/Pulumi.Automation.Tests/LocalWorkspaceTests.cs:line 482
   at Pulumi.Automation.Tests.LocalWorkspaceTests.StackLifecycleInlineProgram() in /Users/runner/work/pulumi/pulumi/sdk/dotnet/Pulumi.Automation.Tests/LocalWorkspaceTests.cs:line 499
```

t0yv0:
# Description

- [x] reproduces the problem in a failing test
- [x] tentative fix that passes the test
- [x] fix existing test failures due to wrapping of exception in AggregateException
- [x] add a test that early termination behavior is preserved on Exception
- [x] add a test that description logging behavior is preserved

Fixes #7492 and #6377 

## Checklist


t0yv0:
@orionstudt early preview, I still have some tests to write here.

sdk/dotnet/Pulumi/Deployment/Deployment.Runner.cs
```diff
@@ -29,9 +30,12 @@ private class Runner : IRunner
             /// continuously, asynchronously loop, waiting for these tasks to complete, and only
             /// exiting once the set becomes empty.
             /// </summary>
-            private readonly Dictionary<Task, List<string>> _inFlightTasks = new Dictionary<Task, List<string>>();
+            private readonly TaskMonitoringHelper _inFlightTasks = new TaskMonitoringHelper();
             private readonly List<Exception> _exceptions = new List<Exception>();
 
+            private readonly ConcurrentDictionary<Tuple<int,string>,int> _descriptions =
```
orionstudt, line 36: you can use named tuples for this so it's a bit easier to read (though I will admit not easier to write).

Those look like:
```csharp
private readonly ConcurrentDictionary<(int NameOne, string NameTwo), int> _descriptions = ...;
```

Which can be expanded when you are using them like:
```csharp
var (nameOne, nameTwo) = this._descriptions.Keys[0];
```

orionstudt: This is a thinking man's PR so i'll try to grok the rest of it over my coffee tmr lol

sdk/dotnet/Pulumi/Deployment/TaskMonitoringHelper.cs
```diff
@@ -0,0 +1,101 @@
...
+{
+    /// Monitors dynamically added tasks for completion. Enters IDLE
+    /// state when all monitored tasks finish. Allows awaiting next
+    /// IDLE state or the first exception, whichever comes first.
+    /// Thread-safe.
+    internal class TaskMonitoringHelper
+    {
+        private readonly object _lockObject = new Object();
+        private int _activeTasks = 0;
+        private TaskCompletionSource<Exception?>? _next = null;
```
justinvp, line 17: Nit: There's no need to specify the default values (smaller IL):

```suggestion
        private int _activeTasks;
        private TaskCompletionSource<Exception?>? _next;
```

mikhailshilkov:
Probably, also fixes https://github.com/pulumi/pulumi/issues/6377 ?

sdk/dotnet/Pulumi/Deployment/Deployment.Runner.cs
```diff
@@ -80,21 +84,31 @@ Task<int> IRunner.RunAsync(Func<Task<IDictionary<string, object?>>> func, StackO
...
+                        (Tuple<int,string> key, int value) => value + 1
+                    );
+                    if (timesSeen == 1)
                     {
-                        descriptions = new List<string>();
-                        _inFlightTasks.Add(task, descriptions);
+                        task.ContinueWith(task => {
+                            _deploymentLogger.LogDebug($"Completed task: {description}");
+                            int old;
+                            _descriptions.TryRemove(key, out old);
```
orionstudt, line 109: You can declare the out variable inline to save yourself having to declare it early:
```csharp
_descriptions.TryRemove(key, out var old);
```

But since here we don't need the out variable you can omit the declaration altogether and use the discard:
```suggestion
                            _descriptions.TryRemove(key, out _);
```

sdk/dotnet/Pulumi/Deployment/TaskMonitoringHelper.cs
```diff
@@ -0,0 +1,101 @@
...
+                // Check if we are moving out of IDLE state.
+                if (_activeTasks == 0)
+                {
+                    // Fresh promise is needed.
+                    _next = new TaskCompletionSource<Exception?>();
+                }
+                _activeTasks++;
+                CheckInvariants();
+            };
+            task.ContinueWith(this.OnTaskCompleted);
```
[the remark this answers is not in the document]
t0yv0: This comment makes me smile because I really am lucky to have a teammate that cares about allocations, while also being a skeptic this will ever matter (if we put a counter on Tasks allocated in pulumi programs I doubt it would ever reach 1mln?), it's just fun.

sdk/dotnet/Pulumi/Deployment/TaskMonitoringHelper.cs
```diff
@@ -0,0 +1,101 @@
...
+
+                // If entering IDLE state or observing an exception, 
+                // notify waiters and reset state with a fresh promise.
+                if (_activeTasks == 0 || task.Exception != null)
+                {
+                    if (_next != null)
+                    {
+                        _next.SetResult(task.Exception);
+                    }
+                    _next = new TaskCompletionSource<Exception?>();
```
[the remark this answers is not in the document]
t0yv0: Point taken I was being sloppy. Let me elaborate the code.

sdk/dotnet/Pulumi.Tests/Deployment/DeploymentRunnerTests.cs
```diff
@@ -0,0 +1,195 @@
...
+                // The joinTask will await register tasks. We check here that it
+                // finishes early because task1 failed, and does not wait for task2.
+                this.RunnerResult = jointTask.ContinueWith(t => {
+                    Assert.True(task1.IsFaulted, "task1 should be IsFaulted");
+                    Assert.False(task2.IsCompleted, "task2 should not be IsCompleted yet");
+                    Assert.Contains(error, runner.SwallowedExceptions);
+                    return 0;
+                });
+
+                // Need to undo the recording of the error in the engine, otherwise
```
t0yv0, line 77: Ok this is a code smell and looks like can still fail to work and fail on CI (though seems to work locally). Let me refactor this.

t0yv0:
Ready for final review folks. Thanks a lot.

sdk/dotnet/Pulumi/Deployment/Deployment.Runner.cs
```diff
@@ -29,9 +30,12 @@ private class Runner : IRunner
...
             /// exiting once the set becomes empty.
             /// </summary>
-            private readonly Dictionary<Task, List<string>> _inFlightTasks = new Dictionary<Task, List<string>>();
+            private readonly TaskMonitoringHelper _inFlightTasks = new TaskMonitoringHelper();
             private readonly List<Exception> _exceptions = new List<Exception>();
 
+            private readonly ConcurrentDictionary<(int TaskId, string Desc),int> _descriptions =
+                new ConcurrentDictionary<(int TaskId, string Desc),int>();
+
             public ImmutableList<Exception> SwallowedExceptions => this._exceptions.ToImmutableList();
```
orionstudt, line 39: Curious if this list I put here and the `_exceptions` list are still necessary or if they can be lifted into your `TaskExceptionTracker` since that also contains a list of exceptions?

I guess this immutable list might still be necessary on the `IRunner` interface but maybe the `_exceptions` one can be lifted and the `TaskMonitoringHelper` can expose some property to access them?

orionstudt:
Had one comment but otherwise this looks good and glad someone finally got to it :)

sdk/dotnet/Pulumi/Deployment/Deployment.Runner.cs
t0yv0: That's a good point, having a look. Looks like that list access is unsynchronized but coming from multiple threads so possibly another cause of woes. And.. yeah it looks like it could be the same list that I moved into the helper class. Let me unify.

t0yv0: Ah actually no, TaskExceptionTracker it's adding and clearing it's internal buffer ~ the intent there was that it tries not to lose exceptions, but it temporarily holds them in between observing them in tasks and next time someone reads them via AwaitExceptionAsync. Exposing that list would make the interface to that class a bit more confusing.

The _exception list in Runner seems to be append-only, and I think Runner should continue to own that, if it's needed, I don't want to be tackling that one in this PR :)

I can however lock access to it so we don't race on it.

orionstudt: Roger that, makes sense.

t0yv0:
Actually, I need to get back to this and find out why exceptions are being duplicated. Just got this CI failure:

```
[xUnit.net 00:03:19.29]     Pulumi.Automation.Tests.LocalWorkspaceTests.InlineProgramExceptionPropagatesToCaller [FAIL]
  X Pulumi.Automation.Tests.LocalWorkspaceTests.InlineProgramExceptionPropagatesToCaller [1s 865ms]
  Error Message:
   Assert.Throws() Failure
Expected: typeof(System.IO.FileNotFoundException)
Actual:   typeof(System.AggregateException): One or more errors occurred. (Unable to find the specified file.) (Unable to find the specified file.)
---- System.AggregateException : One or more errors occurred. (Unable to find the specified file.) (Unable to find the specified file.)
-------- System.IO.FileNotFoundException : Unable to find the specified file.
-------- System.IO.FileNotFoundException : Unable to find the specified file.
```

t0yv0:
Speculating that the same task `RegisterTask` more than once, and the change loses the protection of Dictionary tracking tasks.  Need to return to this on Monday. Also... need to deaggregate exceptions (drop AggregatedException) and decide how to handle multiples of them in HandleExceptionAsync, figure out what the exit codes mean there.

orionstudt:
I think you'll still want AggregateException for when more than 1 task throws an exception.

t0yv0:
Investigating the failing test deeper, thankfully it reproduces locally. It seems we sometimes observe dependent tasks, which cause the same exception to be caught twice in this code:

```
   -- RegisterTask for task 35 on runner #20429945 with description
      <RegisterResourceOutputs: pulumi:pulumi:Stack-exception_inline_tstack_node-hcjhlacq>

   -- RegisterTask for task 30 on runner #20429945 with description <Output<string>>
```

Then identically the same exception was reported multiple times and wrapped as an AggregateException.

This seems like a legitimate consequence of the current class design, when we're called with `RegisterTask` it's not easy to detect if we've already been called with a dependent task.

Opting for a fairly local Distinct dedup pass here, let's see if this passes tests. Could also consider a runner-scoped table to deduplicate.

t0yv0:
OK folks finally a green checkmark from CI (though you need to restart it a few times to get past Python etc flakes...). Ready for final review!

sdk/dotnet/Pulumi/Deployment/TaskMonitoringHelper.cs
```diff
@@ -0,0 +1,191 @@
...
+        public async Task<IEnumerable<Exception>> AwaitIdleOrFirstExceptionAsync()
+        {
+            var error = errTracker.AwaitExceptionAsync();
+            var idle = idleTracker.AwaitIdleAsync();
+            var first = await Task.WhenAny((Task)error, idle).ConfigureAwait(false);
+            if (first == idle)
+            {
+                return Enumerable.Empty<Exception>();
+            }
+            var err = await error;
```
mikhailshilkov, line 42: Nit. Also, I'd use `exceptionTask` instead of `error`.
```suggestion
            return await error;
```

sdk/dotnet/Pulumi/Deployment/TaskMonitoringHelper.cs
```diff
@@ -0,0 +1,191 @@
...
+using System;
+using System.Linq;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Pulumi
+{
+    /// Monitors dynamically added tasks for completion. Enters IDLE
```
mikhailshilkov, line 12: It's not public API but anyway the convention is to wrap the comments between two

```
/// <summary>
/// </summary>
```

I believe we do so fairly consistently.

sdk/dotnet/Pulumi.Tests/Deployment/DeploymentRunnerTests.cs
```diff
@@ -0,0 +1,209 @@
...
+
+            public TerminatesEarlyOnExceptionStack()
+            {
+                var runner = Pulumi.Deployment.Instance.Internal.Runner;
+                var error = new Exception("Deliberate test error");
+
+                Task task1 = Task.FromException(error);
+                Task task2 = Task.Delay(100);
+
+                runner.RegisterTask("task1", task1);
```
mikhailshilkov, line 84: Minor thing but I would swap the order of registrations to "prove" that the exception is returned even if it's registered later.

sdk/dotnet/Pulumi/Deployment/Deployment.Runner.cs
```diff
@@ -29,10 +31,23 @@ private class Runner : IRunner
...
+            private readonly object _exceptionsLock = new object();
             private readonly List<Exception> _exceptions = new List<Exception>();
 
-            public ImmutableList<Exception> SwallowedExceptions => this._exceptions.ToImmutableList();
+            private readonly ConcurrentDictionary<(int TaskId, string Desc),int> _descriptions =
+                new ConcurrentDictionary<(int TaskId, string Desc),int>();
+
+            public ImmutableList<Exception> SwallowedExceptions
+            {
+                get {
```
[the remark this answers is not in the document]
t0yv0: Ouch. :) Damn I need to install that linter/autoformatter...

sdk/dotnet/Pulumi.Tests/Deployment/DeploymentRunnerTests.cs
```diff
@@ -0,0 +1,209 @@
...
+            public StressRunnerStack()
+            {
+                var runner = Pulumi.Deployment.Instance.Internal.Runner;
+
+                for (var i = 0; i < 100; i++)
+                {
+                    runner.RegisterTask($"task{i}", Task.Delay(100 + i));
+                }
+
+                this.RunnerResult = runner.RunAsync<EmptyStack>();
```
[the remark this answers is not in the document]
t0yv0: This test failed on master. This is our first test to reproduce the original problem. It was non-deterministic but N=100 was enough to trigger it pretty consistently. Consider it a regression test?  I an add a comment about that.

The `RunAsync<EmptyStack>()` clause indirectly invokes `WhileRunningAsync()` which is where the bug was. Should I instead make that internal and invoke that directly?

sdk/dotnet/Pulumi.Tests/Deployment/DeploymentRunnerTests.cs
```diff
@@ -0,0 +1,209 @@
...
+            }
+        }
+
+        [Fact]
+        public async Task TerminatesEarlyOnException()
+        {
+            var deployResult = await Deployment.TryTestAsync<TerminatesEarlyOnExceptionStack>(new EmptyMocks());
+            Assert.NotNull(deployResult.Exception);
+            if (deployResult.Exception! is AggregateException)
```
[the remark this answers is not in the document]
t0yv0: This is a good point. This really should be gone now, I've fixed an issue with this duplication after the test was written. Let me try to remove and see that we consistently pass without this.

sdk/dotnet/Pulumi.Tests/Deployment/DeploymentRunnerTests.cs
```diff
@@ -0,0 +1,209 @@
...
+            public TerminatesEarlyOnExceptionStack()
+            {
+                var runner = Pulumi.Deployment.Instance.Internal.Runner;
+                var error = new Exception("Deliberate test error");
+
+                Task task1 = Task.FromException(error);
+                Task task2 = Task.Delay(100);
+
+                runner.RegisterTask("task1", task1);
+                runner.RegisterTask("task2", task2);
```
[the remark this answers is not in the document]
t0yv0: Hm, this is a good idea that has not occurred to me. Sounds like every Output participates in `RegisterTask`? Let me play with this.

sdk/dotnet/Pulumi.Tests/Deployment/DeploymentRunnerTests.cs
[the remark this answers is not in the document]
t0yv0: Ok this is pretty interesting I investigated a bit further here. 

The `RunAsync<EmptyStack>` is essential to the repro. In fact it seems that I can only repro the exception when more than one stack is being deployed against the same Runner object (and same Deployment object). So to reproduce the issue we need a race of two concurrent `WhileRunningAsync()` threads on the same runner. In fact even with these changes here, we will not see exceptions, but we might have race conditions in this case of mis-attributing exceptions (or spurious waits). If the single runner is servicing two stacks, it's going to mixup their exceptions.

So taking a big step back. Do we believe it's an error to be running two stacks simultaneously against the single Deployment? If it's an error, how can we best prevent that from happening?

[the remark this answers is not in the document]
orionstudt: I missed this on my review. To @mikhailshilkov 's point I didn't even know this was possible. I don't think a consumer would be able to do this because the access modifiers would prevent them from accessing the Runner.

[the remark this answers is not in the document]
orionstudt: Conceptually I think it is an error. Currently a Deployment is an instance of a pulumi update action against a single stack, and the Runner is an implementation detail of that Deployment - and it is 1:1. I don't think it would ever make sense the way this it is currently architected to have 2 update actions executing against a single Deployment.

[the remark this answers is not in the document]
orionstudt: So the issue is that we need to find a way to repro the original issue which was a flake race condition caused by access to the inner task dictionary of the Runner? It's always difficult to try to repro a flakey issue but make we could instantiate a Runner for testing purposes and just throw N tasks at it? Maybe you don't need a Deployment to repro since this issue was isolated to the Runner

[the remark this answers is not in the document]
t0yv0: Hrm. That could just be a bad test case. Maybe this situation is unrealistic but under some other situations this condition can arise as well. I'm leaning a little to accepting changes here as they make the locking a little more foolproof?

Let me have one more look at CI failures that manifested with the exception we're fixing here.

Also. We could put a little check into WhileRunningAsync to make sure it's never invoked concurrently (one instance at a time), and fail loudly if it does. If we believe it should never happen, but if we manage to do it somehow anyway, that'd catch it.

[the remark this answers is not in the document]
orionstudt: Probably wouldn't hurt anything. Yea when I made that original Automation API PR I specifically refactored the core SDK so that all pulumi execution enters [right here](https://github.com/pulumi/pulumi/blob/81f87b0c6ac5a2b52268b46599affdf79d9e6db7/sdk/dotnet/Pulumi/Deployment/Deployment_Run.cs#L186), even the Automation API enters there. And you'll see the first thing it does is instantiate a Deployment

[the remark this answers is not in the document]
t0yv0: I've put a PR here with the defensive check https://github.com/pulumi/pulumi/pull/7597 - locally it does not seem to detect it, let's see on CI... Also I locally sporadically get "key not found" in the tests, without triggering the concurrent WRA. So basically I have a poor repro test here, there must be other ways to repro the situation.

t0yv0:
Ah the duplication of exceptions is still present in CI (and can't for the life of me get it locally). OK I need to work through this.

```

   xUnit.net 00:00:02.08]     Pulumi.Tests.DeploymentRunnerTests.TerminatesEarlyOnException [FAIL]
  X Pulumi.Tests.DeploymentRunnerTests.TerminatesEarlyOnException [47ms]
  Error Message:
   The collection was expected to contain a single element, but it contained more than one element.
  Stack Trace:
     at Pulumi.Tests.DeploymentRunnerTests.TerminatesEarlyOnExceptionStack.<>c__DisplayClass4_0.<.ctor>b__0(Task`1 t) in


   xUnit.net 00:00:17.76]     Pulumi.Tests.DeploymentRunnerTests.TerminatesEarlyOnException [FAIL]
  Failed Pulumi.Tests.DeploymentRunnerTests.TerminatesEarlyOnException [90 ms]
  Error Message:
   Assert.IsType() Failure
Expected: Pulumi.RunException
Actual:   System.AggregateException
  Stack Trace:
     at Pulumi.Tests.DeploymentRunnerTests.TerminatesEarlyOnException() in /Users/runner/work/pulumi/pulumi/sdk/dotnet/Pulumi.Tests/Deployment/DeploymentRunnerTests.cs:line 47
--- End of stack trace from previous location where exception was thrown ---


```

t0yv0:
Allright folks I've played with timeouts until I could reliably reproduce exception duplication locally, and it's an artifact of badly written test. Here's what was happening exactly:

1. Stack had outputs, and was participating in `void IDeploymentInternal.RegisterResourceOutputs(Resource resource, Output<IDictionary<string, object?>> outputs)`, which is a registered task also (with the runner)

2. RegisterResourceOutputs task was failing with "Deliberate test error", so it was reported via the runner twice, once from the inner task and once from RegisterResourceOutputs 

3. Normally, error deduplication works just fine to remove the dup reporting here; however since the test was using EmptyStack, it introduced a race with two WhileRunningAsync instances active on the same runner; under certain circumstances the two instances grabbed one copy of the exception each, so deduplication did not apply

4. Based on conversation here and aux PR we believe WhileRunningAsync never runs more than one copy, so it was a borked test

5. Last version of test does not have this problem

sdk/dotnet/Pulumi.Tests/Deployment/DeploymentRunnerTests.cs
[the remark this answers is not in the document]
t0yv0: Yeah that PR basically passes! So far so good.

t0yv0:
Allright, sorry for many detours here.. So my repro was bad, but I think this change still helps. On master I get sporadic test failure. With the change the tests are deterministic for me.

@mikhailshilkov please take another look? Appreciate it.

sdk/dotnet/Pulumi.Tests/Deployment/DeploymentRunnerTests.cs
```diff
@@ -0,0 +1,174 @@
...
+                for (var i = 0; i < 100; i++)
+                {
+                    runner.RegisterTask($"task{i}", Task.Delay(100 + i));
+                }
+
+                // This introduces a race between two simultaneous
+                // `WhileRunningAsync`, the second one being implicit
+                // in runing this stack, and is necessary for the
+                // repro.
+                ((Deployment.Runner)runner).WhileRunningAsync();
```
mikhailshilkov, line 42: Is that even legal to do so? Should we prohibit this? Is there no way to repro in terms of user API?

(Sorry if I missed your earlier comment about that)

sdk/dotnet/Pulumi.Tests/Deployment/DeploymentRunnerTests.cs
```diff
@@ -0,0 +1,174 @@
...
+                TState state,
+                Exception exc,
+                Func<TState, Exception, string> formatter)
+            {
+                var msg = formatter(state, exc);
+                Write($"{level} {eventId} {msg}");
+            }
+
+            public IEnumerable<String> Messages {
+                get {
```
mikhailshilkov, line 118: I see you are persistent about this :)

mikhailshilkov (approved):
Your changes LGTM. Just one question about the test.

sdk/dotnet/Pulumi.Tests/Deployment/DeploymentRunnerTests.cs
t0yv0: Go habits die hard. I write more GO here than anything else.. Damn it.

sdk/dotnet/Pulumi.Tests/Deployment/DeploymentRunnerTests.cs
t0yv0: Yes it's illegal. This can't happen. I could not repro this using legal code unfortunately. Let me just drop this test and check the changes in with no repro of the original problem. I have some indirect indication that the changes are helping (`dotnet test` becomes stable on my machine).

PR #7530, opened 2021-07-14
[codegen/nodejs] - Dynamic provider config getters

komalali:
# Description

Nodejs part of #5582 

## Checklist

### DIFF
--- a/pkg/codegen/internal/test/testdata/provider-config-schema/nodejs/config/vars.ts
+++ b/pkg/codegen/internal/test/testdata/provider-config-schema/nodejs/config/vars.ts
@@ -5,18 +5,68 @@ import * as pulumi from "@pulumi/pulumi";
 import { input as inputs, output as outputs } from "../types";
 import * as utilities from "../utilities";
 
+declare var exports: any;
 let __config = new pulumi.Config("configstation");
 
-export let favoritePlants: string[] | undefined = __config.getObject<string[]>("favoritePlants");
+export declare const favoritePlants: string[] | undefined;
+Object.defineProperty(exports, "favoritePlants", {
+    get() {
+        return __config.getObject<string[]>("favoritePlants");
+    },
+    enumerable: true,
+});
+
 /**
  * omg my favorite sandwich
  */
-export let favoriteSandwich: outputs.config.Sandwich | undefined = __config.getObject<outputs.config.Sandwich>("favoriteSandwich");
-export let isMember: boolean | undefined = __config.getObject<boolean>("isMember") || true;
-export let kids: outputs.Child | undefined = __config.getObject<outputs.Child>("kids");
-export let name: string | undefined = __config.get("name");
-export let numberOfSheep: number | undefined = __config.getObject<number>("numberOfSheep");
+export declare const favoriteSandwich: outputs.config.Sandwich | undefined;
+Object.defineProperty(exports, "favoriteSandwich", {
+    get() {
+        return __config.getObject<outputs.config.Sandwich>("favoriteSandwich");
+    },
+    enumerable: true,
+});
+
+export declare const isMember: boolean | undefined;
+Object.defineProperty(exports, "isMember", {
+    get() {
+        return __config.getObject<boolean>("isMember") || true;
+    },
+    enumerable: true,
+});
+
+export declare const kids: outputs.Child | undefined;
+Object.defineProperty(exports, "kids", {
+    get() {
+        return __config.getObject<outputs.Child>("kids");
+    },
+    enumerable: true,
+});
+
+export declare const name: string | undefined;
+Object.defineProperty(exports, "name", {
+    get() {
+        return __config.get("name");
+    },
+    enumerable: true,
+});
+
+export declare const numberOfSheep: number | undefined;
+Object.defineProperty(exports, "numberOfSheep", {
+    get() {
+        return __config.getObject<number>("numberOfSheep");
+    },
+    enumerable: true,
+});
+
 /**
  * This is a huge secret
  */
-export let secretCode: string | undefined = __config.get("secretCode") || utilities.getEnv("SECRET_CODE", "MY_SUPER_SECRET_CODE");
+export declare const secretCode: string | undefined;
+Object.defineProperty(exports, "secretCode", {
+    get() {
+        return __config.get("secretCode") || utilities.getEnv("SECRET_CODE", "MY_SUPER_SECRET_CODE");
+    },
+    enumerable: true,
+});
+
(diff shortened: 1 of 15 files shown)


pkg/codegen/nodejs/gen.go
```diff
@@ -1158,6 +1158,8 @@ func (mod *modContext) genConfig(w io.Writer, variables []*schema.Property) erro
 
 	mod.genHeader(w, mod.sdkImports(referencesNestedTypes, true), externalImports, imports)
 
+	fmt.Fprintf(w, "declare var exports: any;\n")
+
 	// Create a config bag for the variables to pull from.
 	fmt.Fprintf(w, "let __config = new pulumi.Config(\"%v\");\n", mod.pkg.Name)
```
justinvp, line 1164: Since we're making changes here, should we go ahead and make this `const`?

```suggestion
	fmt.Fprintf(w, "const __config = new pulumi.Config(\"%v\");\n", mod.pkg.Name)
```

pkg/codegen/internal/test/testdata/provider-config-schema/nodejs/config/vars.ts
```diff
@@ -5,18 +5,68 @@ import * as pulumi from "@pulumi/pulumi";
...
-export let numberOfSheep: number | undefined = __config.getObject<number>("numberOfSheep");
+export declare const favoriteSandwich: outputs.config.Sandwich | undefined;
+Object.defineProperty(exports, "favoriteSandwich", {
+    get() {
+        return __config.getObject<outputs.config.Sandwich>("favoriteSandwich");
+    },
+    enumerable: true,
+});
+
+export declare const isMember: boolean | undefined;
```
justinvp, line 30: Unrelated to your change (and not something we have to do as part of this change), but I see that this is never going to be `undefined` because of the default value. Maybe file an issue to consider fixing the type in such cases?

pkg/codegen/nodejs/gen.go
```diff
@@ -1181,8 +1183,13 @@ func (mod *modContext) genConfig(w io.Writer, variables []*schema.Property) erro
 			configFetch += " || " + v
```
justinvp, line 1183: Unrelated to your change, but just wanted to point it out while looking over the code here.

The comment above this line says:

```typescript
// Note: this logic isn't quite correct, but already exists in all of the TF-based providers.
// Specifically, this doesn't work right if the first value is set to false but the default value
// is true.
```

We should be able to address this by using the `??` operator instead of `||`. Maybe worth considering as a separate change?

pkg/codegen/internal/test/testdata/provider-config-schema/nodejs/config/vars.ts
komalali: I'll just fix it here

pkg/codegen/nodejs/gen.go
komalali: Yeah I didn't mess with it because I figured the note was there for a reason but I can just fix it here.

pkg/codegen/internal/test/testdata/provider-config-schema/nodejs/config/vars.ts
justinvp: I wasn't referring to whether the property was required, I was referring to the default value.

`isMember`'s getter is:

```typescript
return __config.getObject<boolean>("isMember") ?? true;
```

Because of the `?? true` it will never return `undefined`.

`name`'s getter is:

```typescript
return __config.get("name");
```

It can still return `string | undefined` because `Config.get` returns `string | undefined`.

komalali: Ah right okay fair enough, missed that about the default values but that's a fair point.

But `name` is required... so it won't ever actually return `undefined` - right? In [python](https://github.com/pulumi/pulumi/blob/master/pkg/codegen/internal/test/testdata/provider-config-schema/python/pulumi_configstation/config/__init__.pyi#L24) this is typed as `str` (not `Optional[str]`) - so I thought it should be the same here...

justinvp: > But `name` is required... so it won't ever actually return `undefined` - right?

It can return `undefined` if someone doesn't have that config value set.

Ah, I missed this in the python review. Since it's just calling `get` it should be `Optional[str]` not `str`. We should fix that. (Sorry, I didn't connect the dots w.r.t. `str` vs. `Optional[str]` in python until this discussion).

> Or should it be `__config.require()` instead?

It likely should, but that would be a behavior change that we should consider separately. (Not sure if we'd ever change it, to be honest).

In AWS, for Node.js, there is an overlay that adds require.ts that can be used to check:

https://github.com/pulumi/pulumi-aws/blob/master/sdk/nodejs/config/require.ts

komalali: Got it. I thought the required checking was not an overlay and a byproduct of setting the config var as required. Okay thanks will fix up!

pkg/codegen/nodejs/gen.go
```diff
@@ -1175,14 +1177,20 @@ func (mod *modContext) genConfig(w io.Writer, variables []*schema.Property) erro
 			if err != nil {
 				return err
 			}
-			// Note: this logic isn't quite correct, but already exists in all of the TF-based providers.
-			// Specifically, this doesn't work right if the first value is set to false but the default value
-			// is true.
-			configFetch += " || " + v
+			configFetch += " ?? " + v
```
t0yv0, line 1180: That's null coalescing operator? Awesome.

pkg/codegen/nodejs/gen.go
```diff
@@ -1175,14 +1177,20 @@ func (mod *modContext) genConfig(w io.Writer, variables []*schema.Property) erro
...
+		}
+		optType := codegen.OptionalType(p)
```
t0yv0, line 1182: Put this in the else branch?

t0yv0 (approved):
Wunderbar, I am learning some TypeScript tricks here, thanks.

justinvp (approved):
LGTM

With the TypeScript upgrade, it's worth noting that it's no longer possible to set these values (given the new TypeScript => JavaScript transpiled code), so it'll no longer be possible to do things like [this](https://github.com/lukehoban/pulumi-package-aws-apigateway/blob/a6ceea9081a4161939670707be20403d1e48123f/provider/cmd/pulumi-resource-apigateway/restAPI.ts#L89-L90). But this change addresses the problem that that's working around, so there's no longer a need to do that.

t0yv0:
It's great not to be able to do things like that.

Issue #7493, opened 2021-07-12
[codegen/python] - Provider config getters should return rich objects


komalali:
Since the type hint is `Optional[Sequence[str]]`, I would expect this to use `__config__.get_object` like the following (and possibly cast the result to make mypy happy, if needed):

```suggestion
    def favorite_plants(self) -> Optional[Sequence[str]]:
        return __config__.get_object('favoritePlants')
```

Looking at AWS, for Node.js you can see it's doing the right thing:

https://github.com/pulumi/pulumi-aws/blob/1b4fa4fecdc5b89c0141f07dcc9f65a4a1465d85/sdk/nodejs/config/vars.ts#L16

```typescript
export let allowedAccountIds: string[] | undefined = __config.getObject<string[]>("allowedAccountIds");
```

But for Python, it looks like we don't return the rich objects, we just always use `get` and return the string value (which for objects, will be a JSON string):

https://github.com/pulumi/pulumi-aws/blob/1b4fa4fecdc5b89c0141f07dcc9f65a4a1465d85/sdk/python/pulumi_aws/config/vars.py#L42

```python
allowed_account_ids = __config__.get('allowedAccountIds')
```

Unfortunately, I don't think we can change this without it being breaking for our existing providers. If we want to return actual objects, we'll probably want a Python setting in the schema so providers can opt-out (or opt-in depending on what we want the default behavior to be).

Or we just maintain the existing behavior and type all these as `Optional[str]`.

_Originally posted by @justinvp in https://github.com/pulumi/pulumi/pull/7447#discussion_r667129900_

PR #7536, opened 2021-07-15
[codegen/python] - Fix required vs default values in provider config

komalali:
# Description

Fixes the type annotations for config variables with default values. In doing this, I realized that some of the existing type annotations were really broken, for instance:

```python
@property
def is_member(self) -> str:
    return __config__.get('isMember') or True
```

This will actually return a string or a bool... so the type hint is definitely wrong (and it's also just bizarre behavior). 

I decided to fix the type annotations and corresponding `config.get` functions for primitive types. This seems like a palatable breaking change, not quite going to the extents of the more complex objects which are still tracked in #7493 which would be a more obviously breaking change for users.

## Checklist

Not updating CHANGELOG-PENDING as this is a follow-on from #7447 

### DIFF
--- a/pkg/codegen/python/gen.go
+++ b/pkg/codegen/python/gen.go
@@ -774,14 +774,7 @@ func (mod *modContext) genConfig(variables []*schema.Property) (string, error) {
 			configFetch += " or " + v
 		}
 
-		// For historical reasons and to maintain backwards compatibility, the config variables for python
-		// are always typed as `Optional[str`] or `str` since the getters only use config.get().
-		// To return the rich objects would be a breaking change, tracked in https://github.com/pulumi/pulumi/issues/7493
-		typeString := "Optional[str]"
-		if p.IsRequired() {
-			typeString = "str"
-		}
-
+		typeString := genConfigVarType(p)
 		fmt.Fprintf(w, "%s@property\n", indent)
 		fmt.Fprintf(w, "%sdef %s(self) -> %s:\n", indent, PyName(p.Name), typeString)
 		dblIndent := strings.Repeat(indent, 2)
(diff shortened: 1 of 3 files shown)


t0yv0:
Before looking at the code... even without type checking, API that returns "string" or bool like this doesn't make sense so you're saying you fix the API here? That'd be wonderful.

pkg/codegen/python/gen.go
```diff
@@ -794,6 +783,49 @@ func (mod *modContext) genConfig(variables []*schema.Property) (string, error) {
...
+	// are typed as `Optional[str`] or `str` for complex objects since the getters only use config.get().
+	// To return the rich objects would be a breaking change, tracked in https://github.com/pulumi/pulumi/issues/7493
+	typeString := "str"
+	switch codegen.UnwrapType(configVar.Type) {
+	case schema.BoolType:
+		typeString = "bool"
+	case schema.IntType:
+		typeString = "int"
+	case schema.NumberType:
+		typeString = "float"
```
t0yv0, line 820: really nitty but I'd prefer `default: typeString = "str"` to make it really obvious we're handling all cases.

pkg/codegen/python/gen.go
```diff
@@ -794,6 +783,49 @@ func (mod *modContext) genConfig(variables []*schema.Property) (string, error) {
...
+		getFunc = "get_bool"
+	case schema.IntType:
+		getFunc = "get_int"
+	case schema.NumberType:
+		getFunc = "get_float"
+	}
+
+	configFetch := fmt.Sprintf("__config__.%s('%s')", getFunc, configVar.Name)
+	if configVar.DefaultValue != nil {
+		v, err := getDefaultValue(configVar.DefaultValue, unwrappedType)
```
t0yv0, line 800: So I've looked at `getDefaultValue` func. How do we know that the `DefaultValue` it receives is conformant to `schema.Type`? It does not seem that it currently checks or returns any error in that case. Is this checked somewhere else? Presumably we would want to let the schema author know they're issuing a malformed schema if they specify a default value that does not typecheck against the type they specify.

```
func getDefaultValue(dv *schema.DefaultValue, t schema.Type) (string, error) {
```

t0yv0: This is off topic for the PR but I'm still curious. If we don't know I can take a ticket to the side to verify this separately.

t0yv0 (approved):

justinvp (approved):
LGTM

I'm slightly hesitant about the breaking change, but the current behavior with non-string default values is just wrong (in some cases you'll get a string, in other cases you'll get some other type), so I'm in favor of fixing that, and it seems reasonable to just go ahead and make all primitives return the right type while making that fix.

pkg/codegen/python/gen.go
komalali: I'm not sure about this, feel free to dig in 😄

PR #7541, opened 2021-07-15
[sdk] Wait on remote component dependencies

### LINKED ISSUE #7540: [sdk/nodejs] `dependsOn` of a multi-language component resource doesn't wait on the children of the component
If two multi-lang components `First` and `Second` are defined, and `First` internally creates a `FirstNested` component which has children, a Node.js program which construct a `First` and then a `Second` that `dependsOn: [first]` will not wait to start constructing `Second` until `First` and all it's children are constructed.

```ts
import * as xyz from "@pulumi/xyz";

const first = new xyz.First("test");

const second = new xyz.Second("test", {
    sgId: "sg-02fb5e24966a9604e",
}, {
    dependsOn: [first],
})
```

Note that `Second` and it's children are being constructed before children of `First` are done being constructed.

image (1)

This is different than the Python and Go SDKs, and not expected behaviour for the Node.js SDK.

lukehoban:
When a resource `dependsOn` a remote component, we were not correctly waiting on it, because we were skipping over waiting on components, and only waiting on their custom resource children.  For remote components, we do not know the children, but waiting on the remote component will always wait on all children.

This fixes the Node.js and .NET SDKs which both contain this approach to collecting transitive dependencies.  The Go and Python SDKs use a different approach which did not have this bug. https://github.com/pulumi/pulumi/issues/7542 is tracking aligning the implementations as a follow-up.

Fixes #7540.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@ CHANGELOG
 - [codegen/go] - Emit To[ElementType]Output methods for go enum output types
   [#7499](https://github.com/pulumi/pulumi/pull/7499)
 
+- [sdk/nodejs] - Wait on remote componet dependencies
+  [#7541](https://github.com/pulumi/pulumi/pull/7541)
+
 ## 3.6.1 (2021-07-07)
 
 ### Improvements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ CHANGELOG
 - [codegen/go] - Emit To[ElementType]Output methods for go enum output types
   [#7499](https://github.com/pulumi/pulumi/pull/7499)
 
-- [sdk/nodejs] - Wait on remote componet dependencies
+- [sdk/nodejs] - Wait on remote component dependencies
   [#7541](https://github.com/pulumi/pulumi/pull/7541)
 
 ## 3.6.1 (2021-07-07)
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,9 @@ CHANGELOG
 - [codegen/go] - Emit To[ElementType]Output methods for go enum output types
   [#7499](https://github.com/pulumi/pulumi/pull/7499)
 
+- [sdk/nodejs] - Wait on remote componet dependencies
```
metral, line 22:
```suggestion
- [sdk/nodejs] - Wait on remote component dependencies
```

sdk/nodejs/runtime/resource.ts
```diff
@@ -584,30 +584,39 @@ function addAll<T>(to: Set<T>, from: Set<T>) {
...
     // To do this, first we just get the transitively reachable set of resources (not diving
     // into custom resources).  In the above picture, if we start with 'Comp1', this will be
     // [Comp1, Cust1, Comp2, Cust2, Cust3]
     const transitivelyReachableResources = await getTransitivelyReferencedChildResourcesOfComponentResources(resources);
 
-    const transitivelyReachableCustomResources = [...transitivelyReachableResources].filter(r => CustomResource.isInstance(r));
+    // Then we filter to only include Custom and Remote resources.
+    const transitivelyReachableCustomResources =
+        [...transitivelyReachableResources]
+        .filter(r => CustomResource.isInstance(r) || (r as ComponentResource).__remote.valueOf);
```
pgavlin, line 619: Can you explain the `.valueOf` idiom?

pgavlin (approved):

sdk/nodejs/runtime/resource.ts
lukehoban: Fixed.

PR #7545, opened 2021-07-16
[codegen/hcl2] Improve ConversionFrom perf.

pgavlin:
- Lazily produce conversion failure diagnostics. This lowers the
  allocation volume and cuts down on execution time by avoiding the
  conversion of source and dest types to strings.
- Add a fast path for union conversions that checks if the source type
  is identical to any of the union's element types. Type equality
  checks are generally much faster than type conversion checks.

These changes lead to a significant speedup in codegen time in
azure-native.

### DIFF
--- a/pkg/codegen/hcl2/model/type.go
+++ b/pkg/codegen/hcl2/model/type.go
@@ -45,7 +45,6 @@ type Type interface {
 
 	equals(other Type, seen map[Type]struct{}) bool
 	conversionFrom(src Type, unifying bool, seen map[Type]struct{}) (ConversionKind, lazyDiagnostics)
-	//conversionTo(dest Type, seen map[Type]struct{}) (ConversionKind, lazyDiagnostics)
 	string(seen map[Type]struct{}) string
 	unify(other Type) (Type, ConversionKind)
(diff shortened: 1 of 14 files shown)


pkg/codegen/hcl2/model/type.go
```diff
@@ -42,7 +44,8 @@ type Type interface {
 	String() string
 
 	equals(other Type, seen map[Type]struct{}) bool
-	conversionFrom(src Type, unifying bool, seen map[Type]struct{}) (ConversionKind, hcl.Diagnostics)
+	conversionFrom(src Type, unifying bool, seen map[Type]struct{}) (ConversionKind, lazyDiagnostics)
+	//conversionTo(dest Type, seen map[Type]struct{}) (ConversionKind, lazyDiagnostics)
```
mikhailshilkov, line 48: Do we need this line?

mikhailshilkov (approved):
LGTM. I'm surprised to learn this has a lot of impact, great catch!

mikhailshilkov:
There's a lint error by the way. Not sure if the test failure may be related?

PR #7546, opened 2021-07-16
[codegen/go] Improve codegen performance.

pgavlin:
- Track which languages have been imported for a package. If a language
  has already been imported, do not re-run its importers.
- Track which package contexts have been loaded in the Go code
  generator, and do not reload a context that already exists.

These changes shave a significant amount of time off of codegen in
azure-native.


mikhailshilkov (approved):
LGTM, thank you!

PR #7547, opened 2021-07-16
[codegen/python] Improve program gen perf.

pgavlin:
- Only build casing tables once per package
- Right-size buffers in name generation

These changes lead to a significant speedup in example gen for
azure-native.

### DIFF
--- a/pkg/codegen/python/python.go
+++ b/pkg/codegen/python/python.go
@@ -106,8 +106,10 @@ func pyName(name string, legacy bool) string {
 	var result strings.Builder           // The components of the name, joined together with underscores.
 	var currentComponent strings.Builder // The characters composing the current component being built
 
-	result.Grow(len(name) + len(name)/5)
-	currentComponent.Grow(len(name) + len(name)/5)
+	// Preallocate enough space for the name + 5 underscores. '5' is based on a wild guess that most names will consist
+	// of 5 or fewer words.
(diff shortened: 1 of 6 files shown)


pkg/codegen/python/python.go
```diff
@@ -105,6 +105,10 @@ func pyName(name string, legacy bool) string {
 
 	var result strings.Builder           // The components of the name, joined together with underscores.
 	var currentComponent strings.Builder // The characters composing the current component being built
+
+	result.Grow(len(name) + len(name)/5)
+	currentComponent.Grow(len(name) + len(name)/5)
```
komalali, line 110: What's the significance of `len(name) + len(name)/5`? Could we add a comment for it?

komalali (approved):

mikhailshilkov (approved):

pkg/codegen/python/python.go
pgavlin: Yep--sorry for the omission. Late-night coding :P

justinvp:
I think we can simply kill all the Python casing table -related code. It's been on my TODO list to go back and clean all this up after the 3.0 release.

pgavlin:
> I think we can simply kill all the Python casing table -related code. It's been on my TODO list to go back and clean all this up after the 3.0 release.

Sweet. Let's do that as a follow-up.

PR #7549, opened 2021-07-16
[codegen/dotnet] - Dynamic provider config getters

### LINKED ISSUE #5582: [Automation API, multi-language components] Provider config statically set when package is imported for node, python, dotnet

komalali:
# Description

Fixes #5582 

Since this is incredibly difficult to write automated tests for, I confirmed that this works as expected locally by running the following program:
```csharp
using System;
using System.Threading.Tasks;
using Pulumi.Automation;

namespace InlineProgram
{
    class Program
    {
        static async Task Main(string[] args)
        {
            // define our pulumi program "inline"
            var program = PulumiFn.Create(() =>
            {
                // export the website url
                {
                    ["currentRegion"] = Pulumi.Aws.Config.Region,
                };
            });

            var projectName = "dynamic_config_test_dotnet";
            var stackName = "dev";

            var stackArgs = new InlineProgramArgs(projectName, stackName, program);
            var stack = await LocalWorkspace.CreateOrSelectStackAsync(stackArgs);

            await stack.Workspace.InstallPluginAsync("aws", "v4.11.0");

            await stack.SetConfigAsync("aws:region", new ConfigValue("us-west-1"));
            var result = await stack.UpAsync();
            Console.WriteLine(result.Outputs["currentRegion"].Value as String == "us-west-1" ? "test passed" : "fail");

            await stack.SetConfigAsync("aws:region", new ConfigValue("us-east-1"));
            result = await stack.UpAsync();
            Console.WriteLine(result.Outputs["currentRegion"].Value as String == "us-east-1" ? "test passed" : "fail");

            await stack.DestroyAsync();
        }
    }
}
```

## Checklist

- [x] I have added (codegen) tests that prove my fix is effective or that my feature works


justinvp (approved):
LGTM

t0yv0 (approved):

Issue #7422, opened 2021-07-02
[sdk/dotnet] Possible race condition results in `NullReferenceException`


gitfool:
Recently I've been experiencing random failures with `NullReferenceException` which looks due to a race condition:
```
Diagnostics:
  pulumi:pulumi:Stack (aws-eks-alpha):
    error: Running program 'D:\Devel\Mps\devops-gemini-pulumi\Gemini\bin\Debug\gemini.dll' failed with an unhandled exception:
    System.NullReferenceException: Object reference not set to an instance of an object.
       at Task<HashSet<string>> Pulumi.Deployment.GetAllTransitivelyReferencedCustomResourceUrnsAsync(HashSet<Resource> resources)+(CustomResource r) => { } in /_/sdk/dotnet/Pulumi/Deployment/Deployment_Prepare.cs:line 152
       at bool System.Linq.Enumerable+SelectEnumerableIterator<TSource, TResult>.MoveNext()
       at Task<TResult[]> System.Threading.Tasks.Task.WhenAll<TResult>(IEnumerable<Task<TResult>> tasks)
       at async Task<HashSet<string>> Pulumi.Deployment.GetAllTransitivelyReferencedCustomResourceUrnsAsync(HashSet<Resource> resources) in /_/sdk/dotnet/Pulumi/Deployment/Deployment_Prepare.cs:line 153
       at async Task<PrepareResult> Pulumi.Deployment.PrepareResourceAsync(string label, Resource res, bool custom, bool remote, ResourceArgs args, ResourceOptions options) in /_/sdk/dotnet/Pulumi/Deployment/Deployment_Prepare.cs:line 80
       at async Task<(string urn, string id, Struct data, ImmutableDictionary<string, ImmutableHashSet<Resource>> dependencies)> Pulumi.Deployment.RegisterResourceAsync(Resource resource, bool remote, Func<string, Resource> newDependency, ResourceArgs args, ResourceOptions options) in /_/sdk/dotnet/Pulumi/Deployment/Deployment_RegisterResource.cs:line 27
       at async Task<(string urn, string id, Struct data, ImmutableDictionary<string, ImmutableHashSet<Resource>> dependencies)> Pulumi.Deployment.ReadOrRegisterResourceAsync(Resource resource, bool remote, Func<string, Resource> newDependency, ResourceArgs args, ResourceOptions options) in /_/sdk/dotnet/Pulumi/Deployment/Deployment_ReadOrRegisterResource.cs:line 73
[...]

PR #7553, opened 2021-07-16
Detect null pointer exceptions in #7422

t0yv0:
# Description

Fixes # (issue)

## Checklist

### DIFF
--- a/sdk/dotnet/Pulumi/Resources/Resource.cs
+++ b/sdk/dotnet/Pulumi/Resources/Resource.cs
@@ -20,46 +20,70 @@ namespace Pulumi
         /// The child resources of this resource.  We use these (only from a ComponentResource) to
         /// allow code to dependOn a ComponentResource and have that effectively mean that it is
         /// depending on all the CustomResource children of that component.
-        /// 
+        ///
         /// Important!  We only walk through ComponentResources.They're the only resources that
         /// serve as an aggregation of other primitive(i.e.custom) resources.While a custom resource
         /// can be a parent of other resources, we don't want to ever depend on those child
         /// resource.  If we do, it's simple to end up in a situation where we end up depending on a
         /// child resource that has a data cycle dependency due to the data passed into it. An
         /// example of how this would be bad is:
-        /// 
+        ///
         /// <c>
         ///     var c1 = new CustomResource("c1");
         ///     var c2 = new CustomResource("c2", { parentId = c1.id }, { parent = c1 });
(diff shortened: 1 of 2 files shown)


t0yv0:
CC @gitfool

t0yv0:
This runs counter to expectations.. Our own test suite fails on this with:

```
xUnit.net 00:01:06.39]     Pulumi.Automation.Tests.LocalWorkspaceTests.InlineProgramAllowsParallelExecution [FAIL]
  X Pulumi.Automation.Tests.LocalWorkspaceTests.InlineProgramAllowsParallelExecution [3s 194ms]
  Error Message:
   System.AggregateException : One or more errors occurred. (Attempting to repeatedly set `Urn` property on Pulumi.Stack resource. Please report to https://github.com/pulumi/pulumi/issues)
---- System.Exception : Attempting to repeatedly set `Urn` property on Pulumi.Stack resource. Please report to https://github.com/pulumi/pulumi/issues
  Stack Trace:
```

sdk/dotnet/Pulumi/Resources/Resource.cs
```diff
@@ -20,46 +20,70 @@ public class Resource
...
         [Output(Constants.UrnPropertyName)]
-        public Output<string> Urn { get; private protected set; } = null!;
+        public Output<string> Urn {
+            get
+            {
+                if (_urn != null)
+                {
+                    return _urn;
+                }
+                throw new Exception($"Uninitialized `Urn` property on {this} resource. " +
```
gitfool, line 72: This should be wrapped in an `else` block.

sdk/dotnet/Pulumi/Resources/Resource.cs
```diff
@@ -20,46 +20,70 @@ public class Resource
...
+                                    "Please report to https://github.com/pulumi/pulumi/issues");
+            }
+            private protected set
+            {
+                if (_urn == null)
+                {
+                    _urn = value;
+                }
+                throw new Exception($"Attempting to repeatedly set `Urn` property on {this} resource. " +
```
gitfool, line 81: This should be wrapped in an `else` block.

t0yv0: Ah certainly, thanks :/ here it's not just style is it.

gitfool: The first else on the getter wasn't needed... I'm still waking up! 😉

t0yv0: Fixed up, let's see if the CI passes possibly.

PR #7555, opened 2021-07-16
[sdk/python] Support for implementing methods in provider

justinvp:
Part of #7072


sdk/python/lib/pulumi/provider/server.py
```diff
@@ -55,10 +54,20 @@ class ProviderServicer(ResourceProviderServicer):
     engine_address: str
     provider: Provider
     args: List[str]
+    lock: asyncio.Lock
+
 
-    # NOTE: remove @_asynchronized when we can avoid modifying globals in the method body.
-    @_asynchronized
```
t0yv0, line 60: I was prematurely getting happy that we factored out the global state mod :) 

But looks like you just need to use the same lock object in both functions. In that case you could also just make it a parameter of the modifier so you say `@asynchronized(lambda self: self.lock)` and avoid the noise here. Doesn't matter too much just a tad more readable.

sdk/python/lib/pulumi/provider/server.py
```diff
@@ -92,36 +101,40 @@ async def Construct(self, request: proto.ConstructRequest, context) -> proto.Con
...
-            k: ProviderServicer._construct_output(the_input, deps=deps(k))
+            k: await ProviderServicer._create_output(the_input, deps=deps(k))
             for k, the_input in
             rpc.deserialize_properties(request.inputs, keep_unknowns=True).items()
         }
 
     @staticmethod
-    def _construct_output(the_input: Any, deps: Set[Resource]) -> Any:
+    async def _create_output(the_input: Any, deps: Set[str]) -> Any:
+        is_resource_reference = (known_types.is_resource(the_input)
```
t0yv0, line 121: This condition implies roughly this:

```
r: Resource
_create_output(r, deps=[r.urn])
```

So it depends only on its own `urn`?

It's probably fine but its getting tricky, perhaps a separate def to detect if this?

sdk/python/lib/pulumi/provider/server.py
```diff
@@ -158,6 +171,86 @@ async def _construct_response(self, result: ConstructResult) -> proto.ConstructR
...
+            return await self._call(request, context)
+        finally:
+            self.lock.release()
+
+    async def _call(self, request: proto.CallRequest, context):
+        # pylint: disable=unused-argument
+        assert isinstance(request, proto.CallRequest), \
+            f'request is not CallRequest but is {type(request)} instead'
+
+        pulumi.runtime.settings.reset_options(
```
t0yv0, line 189: As before, we don't reset them back to original values do we, when done? That is nothing like:

```
with temp_runtime_settings(...):
    ....
```

Perhaps this is fine..

Also would we be interested in removing the lock actually?

1. the best would be to get to explicit parameter passing; perhaps a lot of work/ breaking?

2. we could try to use https://www.python.org/dev/peps/pep-0555/#semantics-and-higher-level-api coroutine-scoped context vars;  the change probably would be lexically very small, however I'd want to personally test out how context vars behave to confirm the assumptions before shipping something like that, it sounds a bit risky

sdk/python/lib/pulumi/runtime/proto/provider_pb2.pyi
```diff
@@ -49,29 +49,62 @@ class ConstructResponse:
...
-        def __init__(self, urns: List[str]) -> void:
+        def __init__(self, urns: List[str]) -> None:
             pass
 
     urn: str
     state: Struct
     stateDependencies: Dict[str,PropertyDependencies]
 
 
+class CallRequest:
```
t0yv0, line 66: Sorry for this stuff here, still haven't gotten to auto-generate it, it's somewhere in the backlog.

tests/integration/construct_component_methods/testcomponent-python/testcomponent.py
```diff
@@ -0,0 +1,95 @@
...
+    @pulumi.getter
+    def first(self) -> pulumi.Output[str]:
+        return pulumi.get(self, "first")
+
+    @property
+    @pulumi.getter
+    def second(self) -> pulumi.Output[str]:
+        return pulumi.get(self, "second")
+
+    def get_message(self, name: pulumi.Input[str]) -> pulumi.Output[str]:
```
t0yv0, line 50: We don't annotate these specially like `@pulumi.gettter`? Eg. the annotation is just by the virtue of including it in Provider? This is fine, just curious.

t0yv0, line 50: I'm hoping/guessing the `self` here exercises the "do not wrap in output" Resource  reference path I had qns on earlier, if so we're good with testing it..

t0yv0 (approved):

sdk/python/lib/pulumi/provider/server.py
justinvp: I was originally planning to do that, but I'm hesitant because I'm not sure it'll use the same event loop as what will be used by the gRPC server.

t0yv0: We could first move off threaded grpc server onto evented (asyncio) one? That'd remove threads.

t0yv0: Or even in that case, there's a possibility of context switches in how we handle the request?

sdk/python/lib/pulumi/provider/server.py
justinvp: We'll want to remove the lock eventually.

tests/integration/construct_component_methods/testcomponent-python/testcomponent.py
justinvp: No annotation necessary for these.

justinvp: Right. It's really for line 87-89:

```python
        comp: Component = args["__self__"]
         outputs = {
             "message": comp.get_message(args["name"])
         }
```

So you can call `comp.get_message(...)` and don't have to do `comp.apply(lambda v: v.get_message(...))`.

PR #7558, opened 2021-07-16
[docgen/go] - Fix go docgen links for pulumi types

komalali:
# Description

The go docgen was incorrectly using the provider major version to create links to Pulumi SDK godocs for pulumi type links. Any ideas for how I could do this in a less static way? Do we have access to the pulumi SDK major version anywhere? If there are no quick answers I'd prefer we merge this to fix the broken links and make it perfect later.

Fixes https://github.com/pulumi/docs/issues/6240

## Checklist

### DIFF
--- a/pkg/codegen/go/doc.go
+++ b/pkg/codegen/go/doc.go
@@ -28,6 +28,8 @@ import (
 	"github.com/pulumi/pulumi/pkg/v3/codegen/schema"
 )
 
+const pulumiSDKVersion = "v3"
+
 // DocLanguageHelper is the Go-specific implementation of the DocLanguageHelper.
 type DocLanguageHelper struct {
 	packages map[string]*pkgContext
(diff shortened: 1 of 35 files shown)


pkg/codegen/go/doc.go
```diff
@@ -36,14 +36,9 @@ type DocLanguageHelper struct {
...
-func (d DocLanguageHelper) GetDocLinkForPulumiType(pkg *schema.Package, typeName string) string {
-	moduleVersion := ""
-	if pkg.Version != nil {
-		if pkg.Version.Major > 1 {
-			moduleVersion = fmt.Sprintf("v%d/", pkg.Version.Major)
-		}
-	}
-	return fmt.Sprintf("https://pkg.go.dev/github.com/pulumi/pulumi/sdk/%sgo/pulumi?tab=doc#%s", moduleVersion, typeName)
+func (d DocLanguageHelper) GetDocLinkForPulumiType(_ *schema.Package, typeName string) string {
+	pulumiModuleVersion := "v3"
```
justinvp, line 40: Pull this out into a `const` at the top of the file, or if not, make it a `const` here?

justinvp (approved):
LGTM. I'm not aware of a built-in way to determine this dynamically. Maybe @pgavlin knows.

pgavlin:
> Do we have access to the pulumi SDK major version anywhere? If there are no quick answers I'd prefer we merge this to fix the broken links and make it perfect later.

We do not, but I think there is a quick answer: we can add it to the [Go-specific package info](https://github.com/pulumi/pulumi/blob/master/pkg/codegen/go/importer.go#L23-L48). Maybe something like `PulumiSDKVersion`?

komalali:
@pgavlin I added it to the go-specific package info as well, but it will default to the specified version in docgen if it isn't supplied. Keeps it flexible without causing breaking changes to existing providers :)

pgavlin (approved):

PR #7562, opened 2021-07-17
[sdk/dotnet] Clean obj & bin dirs when building

justinvp:
This should avoid occasional errors like:

```
attempting to find a local Pulumi NuGet package yielded [/opt/pulumi/nuget/Pulumi.3.8.0-alpha.1626397740.nupkg /opt/pulumi/nuget/Pulumi.3.8.0-alpha.1626408887.nupkg] results
```

When there are more then one versions leftover from prior builds.

@stack72, I noticed `dotnet clean` was already in the Makefile, but commented out. Any downside you're aware of to uncommenting?

### DIFF
--- a/build.proj
+++ b/build.proj
@@ -63,6 +63,8 @@
       <VersionPart>$(Version.Split("-")[0])</VersionPart>
       <VersionPrefix>$(VersionPart.Replace("v",""))</VersionPrefix>
     </PropertyGroup>
+    <Exec Command="dotnet clean"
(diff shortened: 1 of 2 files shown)


lukehoban (approved):

mikhailshilkov (approved):
LGTM

justinvp:
Sigh. Turns out `dotnet clean` doesn't remove any `*.nupkg` files in the `/bin/Debug` directories.

PR #7572, opened 2021-07-19
Fix tests on master

t0yv0:
# Description

Codegen tests are sensitive to dependency versions it seems, we need to bump this.

## Checklist

### DIFF
--- a/pkg/codegen/internal/test/testdata/external-resource-schema/go/example/argFunction.go
+++ b/pkg/codegen/internal/test/testdata/external-resource-schema/go/example/argFunction.go
@@ -4,7 +4,7 @@
 package example
 
 import (
-	"github.com/pulumi/pulumi-random/sdk/v2/go/random"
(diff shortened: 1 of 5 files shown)


EvanBoyle (commented):
This failure looks like a legit bug that needs to be fixed. The tests in question reference external types via schema:

https://github.com/pulumi/pulumi/blob/ed9124972cabd3b4979e6997927cba55add6ff0c/pkg/codegen/internal/test/testdata/external-resource-schema/schema.json#L8

The types in the tests in question are for v2 of the random provider.

t0yv0:
Ah good catch.

t0yv0:
So this passes locally if I remove `pulumi-plugin-random` from PATH per Evan's suggestion. CI occurrence:

https://github.com/pulumi/pulumi/pull/7529/checks?check_run_id=3106762602#step:17:3236

Annoying that presence of the plugin on the path affects test results here but not as big of a deal as broken tests on master. Closing for now.

PR #7573, opened 2021-07-19
[codegen/model] Speed up union types.

pgavlin:
The goal of these changes is to reduce the allocation volume incurred by
union types.

Union types currently allocate a large volume of data when they are
created in order to deduplicate element types and sort element types
into a canonical order. The deduplication is important as it allows for
a fast path during comparison based on the number of elements, but the
sorting--also used to simplify equality checks--generally costs more
than it is worth, as it requires stringifying each element type.

These changes replace the string-based deduplication process with either
a simple O(N^2) approach for small numbers of element types or a more
complex hashtable-based approach for larger numbers of element types.
The same approach is used for equality checks.

These changes speed up the code generation process by decreasing the
amount of time spent in the allocator and garbage collector.

### DIFF
--- a/pkg/codegen/hcl2/model/type.go
+++ b/pkg/codegen/hcl2/model/type.go
@@ -76,7 +76,6 @@ const (
 	hashKindConst
 	hashKindList
 	hashKindMap
-	hashKindNamed
 	hashKindObject
 	hashKindOpaque
 	hashKindOutput
--- a/pkg/codegen/hcl2/model/type.go
+++ b/pkg/codegen/hcl2/model/type.go
@@ -15,6 +15,8 @@
 package model
 
 import (
+	_ "unsafe" // for linkname
+
 	"github.com/hashicorp/hcl/v2"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 )
@@ -43,11 +45,82 @@ type Type interface {
 	ConversionFrom(src Type) ConversionKind
 	String() string
 
+	hash(stack objTypeSet) uint32
 	equals(other Type, seen map[Type]struct{}) bool
 	conversionFrom(src Type, unifying bool, seen map[Type]struct{}) (ConversionKind, lazyDiagnostics)
 	string(seen map[Type]struct{}) string
 	unify(other Type) (Type, ConversionKind)
 	isType()
+}
+
+type objTypeSet map[*ObjectType]struct{}
+
+func (s *objTypeSet) add(t *ObjectType) bool {
+	m := *s
+	if m == nil {
+		m = map[*ObjectType]struct{}{}
+		*s = m
+	}
+
+	_, has := m[t]
+	m[t] = struct{}{}
+	return has
+}
+
+func (s objTypeSet) delete(t *ObjectType) {
+	delete(s, t)
+}
+
+const (
+	hashKindNone uint32 = iota + 1
+	hashKindConst
+	hashKindList
+	hashKindMap
+	hashKindNamed
+	hashKindObject
+	hashKindOpaque
+	hashKindOutput
+	hashKindPromise
+	hashKindSet
+	hashKindTuple
+	hashKindUnion
+)
+
+//go:linkname hashString runtime.stringHash
+func hashString(s string, seed uintptr) uintptr
+
+type fnvHash uint32
+
+func newFNV() fnvHash {
+	return fnvHash(2166136261)
+}
+
+func (f *fnvHash) sum() uint32 {
+	return uint32(*f)
+}
+
+func (f *fnvHash) addString(s string) {
+	f.addUint32(uint32(hashString(s, 0)))
+}
+
+func (f *fnvHash) addUint32(c uint32) {
+	h := uint32(*f)
+	h ^= c & 0xff
+	h *= 16777619
+	h ^= (c >> 8) & 0xff
+	h *= 16777619
+	h ^= (c >> 16) & 0xff
+	h *= 16777619
+	h ^= (c >> 24) & 0xff
+	h *= 16777619
+	*f = fnvHash(h)
+}
+
+func hashCombine(code1, code2 uint32) uint32 {
+	fnv := newFNV()
+	fnv.addUint32(code1)
+	fnv.addUint32(code2)
+	return fnv.sum()
 }
 
 var (
--- a/pkg/codegen/hcl2/model/type_object.go
+++ b/pkg/codegen/hcl2/model/type_object.go
@@ -35,7 +35,9 @@ type ObjectType struct {
 	Annotations []interface{}
 
 	propertyUnion Type
-	s             string
+
+	s string
+	h uint32
 }
 
 // NewObjectType creates a new object type with the given properties and annotations.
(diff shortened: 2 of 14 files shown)


pgavlin:
No rush on the PR here--I'd like to get it in at some point in the near future, but I'd also like to add some more tests to validate the changes.

mikhailshilkov:
Thank you for caring about codegen perf! I'm curious if you have any preliminary measurements of the speedup effect?

pkg/codegen/hcl2/model/type.go
```diff
@@ -43,13 +45,84 @@ type Type interface {
...
 	equals(other Type, seen map[Type]struct{}) bool
 	conversionFrom(src Type, unifying bool, seen map[Type]struct{}) (ConversionKind, lazyDiagnostics)
 	string(seen map[Type]struct{}) string
 	unify(other Type) (Type, ConversionKind)
 	isType()
 }
 
+type objTypeSet map[*ObjectType]struct{}
+
+func (s *objTypeSet) add(t *ObjectType) bool {
```
t0yv0, line 58: Something just slightly weird here, I'm more used to set.Add to return true if the element is added, here it's inverse, it returns false. E.g. https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.hashset-1.add?view=net-5.0

pkg/codegen/hcl2/model/type.go
```diff
@@ -43,13 +45,84 @@ type Type interface {
...
+	hashKindPromise
+	hashKindSet
+	hashKindTuple
+	hashKindUnion
+)
+
+//go:linkname hashString runtime.stringHash
+func hashString(s string, seed uintptr) uintptr
+
+type fnvHash uint32
```
t0yv0, line 92: Considered linking in fnv1a from https://github.com/segmentio/fasthash ? Don't know how conservative we are about taking on little dependencies. If you think we prefer rolling our own a quick comment on why in here would be good.

pkg/codegen/hcl2/model/type_object.go
```diff
@@ -80,6 +82,32 @@ func (t *ObjectType) Traverse(traverser hcl.Traverser) (Traversable, hcl.Diagnos
...
+		fnv := newFNV()
+		fnv.addUint32(hashKindObject)
+
+		keys := make([]string, 0, len(t.Properties))
+		for k := range t.Properties {
+			keys = append(keys, k)
+		}
+		sort.Strings(keys)
+
+		for _, k := range keys {
```
t0yv0, line 101: I don't know if we ever have a thing with super-fat objects, but you can consider giving up hashing past 8th field or such.

pkg/codegen/hcl2/model/type_table.go
```diff
@@ -0,0 +1,140 @@
+package model
+
+type typeTable struct {
```
t0yv0, line 3: Ha, so we need this because go does not let us build hashmaps with custom hash/equality? That seems to be true, I googled for libraries, but there is nothing out there, how sad.

pkg/codegen/hcl2/model/type_union.go
```diff
@@ -32,38 +33,69 @@ type UnionType struct {
...
+				}
+			}
 		}
 	}
-	elementTypes = elementTypes[:dst]
 
 	if len(elementTypes) == 1 {
+		tt := reflect.ValueOf(elementTypes[0])
+		if !tt.Elem().IsValid() {
+			panic("what?")
```
t0yv0, line 97: Just to be on the safe side let's put a bit more descriptive string into the panic :)

pkg/codegen/hcl2/model/type_object.go
```diff
@@ -89,6 +117,15 @@ func (t *ObjectType) equals(other Type, seen map[Type]struct{}) bool {
...
 	}
+
+	otherObject, ok := other.(*ObjectType)
+	if !ok {
+		return false
+	}
+	if len(t.Properties) != len(otherObject.Properties) {
+		return false
+	}
+
```
t0yv0, line 128: There is something slightly spooky semantically here (before the changes). The function trying to prove `t.equals(x)` decides that it's true if it needs to prove `t.equals(y)` goal as part of the proof...

It sounds to me it would prove `A=B` from:

```
type A struct {
    x: int
    y: A
}

type B struct {
    x: int
    y: string
}
```

Equality on these infinite coinductive structures is tricky. I wonder if I am missing something and this actually works, or we just do not hit corner cases in practice.

pkg/codegen/hcl2/model/type_union.go
```diff
@@ -142,8 +206,32 @@ func (t *UnionType) equals(other Type, seen map[Type]struct{}) bool {
...
-		if !t.equals(otherUnion.ElementTypes[i], seen) {
+	if t.hash(nil) != otherUnion.hash(nil) {
+		return false
+	}
+
+	if len(t.ElementTypes) < 16 {
+		for _, t := range t.ElementTypes {
+			found := false
+			for _, u := range otherUnion.ElementTypes {
+				if t.equals(u, seen) {
```
t0yv0, line 217: This was here before the changes also, but it scares me we propagate the same `seen` map that seems to pertain to `t`, but we have substituted `t` with shadowing in `for _, t := range t.ElementTypes`.

t0yv0:
Pat this basically LGTM but two things:

- I'm not very good at proofreading the imperative code for roll-your-own dict and such; if you need solid review there I can give it another concerted effort, but we'd benefit from more reviewers; if I was writing the code I'd be second guessing myself and writing some heavy tests for it; same goes for equals overloads

- I have concerns something funky is going on with type equality before the changes, the `seen` map

This kind of code would be easy to test out if we had solid random generators for the domain of types.

```
prop1(randomType1, randomType2 Type) bool {
    return naiveEq(randomType1, randomType2) = fastEq(randomType1, randomType2)
}
```

And same for testing hashtables against maps, etc.

pkg/codegen/hcl2/model/type_object.go
pgavlin: I do think that this is subtly broken, and has been since we started handling recursive types. What we really want to say is that the following two types are equal:

```
struct {
    x int
    y self
}

struct {
    x int
    y self
}
```

Note that I'm deliberately avoiding the use of names here, as the type system doesn't give names to object types. I considered making that change, but it's quite a bit more disruptive. It might be worth doing, however, as it may reduce the equality problem to "just check the names".

pkg/codegen/hcl2/model/type_union.go
pgavlin: oh, oops--this was debugging code that should be removed 😬

pkg/codegen/hcl2/model/type_table.go
pgavlin: That's correct. Waiting for generics on that one...

pkg/codegen/hcl2/model/type.go
pgavlin: > If you think we prefer rolling our own a quick comment on why in here would be good.

That's a good find. I rolled this one together because
a. FNV-1a is simple
b. The [stdlib implementation](https://pkg.go.dev/hash/fnv@go1.16.6#New32a) doesn't have a convenience method for adding `Uint32` values to the hash, and writing a full FNV-1a implementation is ~as much code as writing a convenience method
c. The ability to use the internal implementation of string hashing is useful for performance.

All that said, I think that we can probably use the package you've linked w/o issue. Really wish the stdlib had the convenience method I wanted...

pgavlin:
> I'm curious if you have any preliminary measurements of the speedup effect?

I don't. I intend to add some microbenchmarks as part of the tests for this to help track type system perf.

pgavlin:
> if I was writing the code I'd be second guessing myself and writing some heavy tests for it; same goes for equals overloads

Yup, hence my first comment on the PR:

> No rush on the PR here ... I'd also like to add some more tests to validate the changes.

pkg/codegen/hcl2/model/type_table.go
t0yv0: Sadness.. the universe is cold and unforgiving.

mikhailshilkov:
@pgavlin @t0yv0 Should we close this PR as stale, or do you think it's still revivable?

t0yv0:
I like this PR. It feels like quite a bit of work to come back to this. This feels to me in the category of both quite a bit involved and foundational tech many pieces depend on. Given this I'd love to elaborate with a few more tests and a bit less laconic style, as well as maybe see if we can get by without unsafe linker tricks for super fast hashing.

So the question is should we do this now? Or can this be parked until such time as we hit a wall. Do we have a motivating example, something with a union type with too many cases, that makes the current implementation impractical for users? Or the measurements asked for before?

Or perhaps there's a HCL maintenance backlog somewhere this could go on?

lunaris:
Going to close this as stale for now -- it sounds like the risks are non-zero and we can always revive it later on if this becomes an issue.

PR #7577, opened 2021-07-20
Set UseSharedCompilation to false

mikhailshilkov:
Fix https://github.com/pulumi/pulumi-azure-native/issues/985 (once it propagates to azure-native)

This is only needed in Azure Native but as we only compile a single project for each provider, there should be no downside to enabling it for all providers.


t0yv0:
DO I read this right that this means a little bit slower compile times / more resources, but fixes the mentioned overflow bug? SGTM. 🚢

t0yv0 (approved):

justinvp (approved):
LGTM

mikhailshilkov:
> a little bit slower compile times / more resources,

If I read it right, it gets slower compile times for projects that are compiled sequentially one after the other which shouldn't be super relevant.

PR #7579, opened 2021-07-20
[sdk/{go,dotnet] Unmarshal invalid assets.

pgavlin:
The two more strongly-typed Pulumi SDKs curently fail with an error
during unmarshaling when attempting to marshal a value that is not an
asset into an asset-typed location (e.g. an asset-typed resource
output property). While this behavior is reasonable on its face, it
gives rise to practical challenges when dealing with TF-provider-backed
resources that have asset-typed properties. When such a resource is
refreshed, the values of its asset-typed properties are replaced with
non-asset values, as the TF bridge can't currently create a resonable
stand-in asset value.

For example, consider an S3 bucket object:

```

const bucket = new aws.s3.Bucket("my-bucket");
new aws.s3.BucketObject("my-object", {
    source: new pulumi.FileAsset("some/file"),
});
```

Upon creation, the value of the input property `source` will be a file
asset backed by the path `some/file`. The bridge will propagate this
value to the `source` output property; this propagation is safe because
the resource was just created and so the output property must have the
value that was passed by the program.

Now, let some actor apply out-of-band changes to the contents of the
bucket object s.t. the `source` property changes when the object is
refreshed. In that case, the `source` property will be a string value
which the bridge is unable to interpret as an asset. The next time the
Pulumi program is run, the Go or .NET SDK will attempt to deserialize
the string into an asset-typed property and will fail.

With these changes, the deserialization would not fail, and would
instead create an asset or archive value that will fail to marshal if
passed to another resource. Users can avoid these errors by not passing
asset or archive outputs to other resources/stack outputs.

These changes unblock users who are hitting
https://github.com/pulumi/pulumi-aws/issues/1521.

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -4,5 +4,8 @@
 - [sdk/python] - Support for authoring resource methods in Python.
   [#7555](https://github.com/pulumi/pulumi/pull/7555)
 
+- [sdk/{go,dotnet}] - Admit non-asset/archive values when unmarshaling into assets and archives.
+  [#7579](https://github.com/pulumi/pulumi/pull/7579)
+
 ### Bug Fixes
 
--- a/sdk/dotnet/Pulumi.Automation/Pulumi.Automation.xml
+++ b/sdk/dotnet/Pulumi.Automation/Pulumi.Automation.xml
@@ -739,6 +739,12 @@
             A Pulumi program as an inline function (in process).
             </summary>
         </member>
+        <member name="M:Pulumi.Automation.PulumiFn.InvokeAsync(Pulumi.IRunner,System.Threading.CancellationToken)">
+            <summary>
+            Invoke the appropriate run function on the <see cref="T:Pulumi.IRunner"/> instance. The exit code returned
+            from the appropriate run function should be forwarded here as well.
+            </summary>
+        </member>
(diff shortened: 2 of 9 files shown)


lblackstone (approved):
Code changes LGTM, but was curious about the following:

> Users can avoid these errors by not passing
asset or archive outputs to other resources/stack outputs.

Is this a common scenario? Do we need to add any documentation about this?

pgavlin:
> Is this a common scenario? Do we need to add any documentation about this?

I think that this is a _very_ uncommon scenario, esp. since this is already broken today.

pgavlin:
> Do we need to add any documentation about this?

I realize I didn't really answer this question--we _should_ document this, yes. Not sure where to add that documentation, but we should do it.

mikhailshilkov (approved):
Makes sense to me

pgavlin:
> we should document this, yes. Not sure where to add that documentation, but we should do it.

@lukehoban any ideas on where we ought to document this?

lukehoban:
> @lukehoban any ideas on where we ought to document this?

I would expect somewhere on https://www.pulumi.com/docs/intro/concepts/assets-archives/.   I think we could just add a note there saying that asset objects should be used as inputs to Pulumi programs, but should generally not be read from the outputs of Pulumi programs, and that behaviour may be undefined when assets are read as outputs.

Issue #5474, opened 2020-09-29
pulumi stack output --show-secrets incorrectly displays special characters


ghost:
I'm using @pulumi/random to generate password for my rds instance. The value I got was something like `pass&word`.  When I use command `pulumi stack output --show-secrets` it displayed the value `pass\u0026word`

## Expected Behavior

Show correct secret value in correct encoding, i.e. `pass&word`

## Current Behavior

Shows incorrect secret value , i.e. `pass\u0026word`

## Steps to Reproduce

1.  clone https://github.com/akhanqi/secrets-pulumi-bug
2. run `pulumi up`
3. run `pulumi stack output --show-secret`
4. check that output has \u003c or other unicode characters

> OUTPUT    VALUE
    password  {"id":"none","length":30,"lower":true,"minLower":0,"minNumeric":0,"minSpecial":30,"minUpper":0,"number":true,"result":"_}}%-++{:_(-[%\u003c\u003e?-\u003c{})\u003e=+\u0026-#]$","special":true,"upper":true,"urn":"urn:pulumi:dev::pulumi-secrets-bug::random:index/randomPassword:RandomPassword::password"}

jccr:
I rely on `pulumi stack output --show-secrets --json` as part of my CI workflow, and I've recently hit this issue when wrapping a value with an `&` character with `pulumi.secret()`

This became a showstopper for me. I think I can try to workaround this in my CI script to unescape the value, but this is inconvenient.

@stack72 any word on the secret handling improvement work?

PR #7593, opened 2021-07-21
[cli] - Don't escape special characters when printing JSON

komalali:
# Description

Fixes https://github.com/pulumi/pulumi/issues/5474

## Checklist

### DIFF
--- a/pkg/cmd/pulumi/util.go
+++ b/pkg/cmd/pulumi/util.go
@@ -776,12 +776,11 @@ func (cancellationScopeSource) NewScope(events chan<- engine.Event, isPreview bo
 }
 
 func makeJSONString(v interface{}) (string, error) {
-	out := bytes.NewBuffer([]byte{})
-	encoder := json.NewEncoder(out)
+	var out bytes.Buffer
+	encoder := json.NewEncoder(&out)
 	encoder.SetEscapeHTML(false)
 	encoder.SetIndent("", "  ")
-	err := encoder.Encode(v)
-	if err != nil {
+	if err := encoder.Encode(v); err != nil {
 		return "", err
 	}
 	return out.String(), nil
--- a/pkg/cmd/pulumi/util.go
+++ b/pkg/cmd/pulumi/util.go
@@ -775,13 +775,25 @@ func (cancellationScopeSource) NewScope(events chan<- engine.Event, isPreview bo
 	return c
 }
 
+func makeJSONString(v interface{}) (string, error) {
+	out := bytes.NewBuffer([]byte{})
+	encoder := json.NewEncoder(out)
+	encoder.SetEscapeHTML(false)
+	encoder.SetIndent("", "  ")
+	err := encoder.Encode(v)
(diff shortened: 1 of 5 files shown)


EvanBoyle:
Huh, I wonder if the failures seen in https://github.com/pulumi/pulumi/issues/6768 are at all related to this category of bug occurring elsewhere in the engine event code.

I do see unicode in the error trace posted by one of the users (`\u003c`) in that thread.

komalali:
>Huh, I wonder if the failures seen in #6768 are at all related to this category of bug occurring elsewhere in the engine event code.

The escaped unicode is certainly present in those engine events and likely should be corrected. I don't believe that's the source of that error though. Every single diagnostic event contains escaped unicode that you referenced, but they rarely result in an error. And I haven't seen any evidence that the break in json resulting in the error is at the location of the escaped unicode.

pkg/cmd/pulumi/util.go
```diff
@@ -775,13 +775,25 @@ func (cancellationScopeSource) NewScope(events chan<- engine.Event, isPreview bo
...
+	encoder.SetIndent("", "  ")
+	err := encoder.Encode(v)
+	if err != nil {
+		return "", err
+	}
+	return out.String(), nil
+}
+
 // printJSON simply prints out some object, formatted as JSON, using standard indentation.
 func printJSON(v interface{}) error {
```
t0yv0, line 791: Nit: perhaps put "indent" in the name of the func to further emphasize what it does.

t0yv0 (approved):
LGTM

pkg/cmd/pulumi/util.go
```diff
@@ -775,13 +775,25 @@ func (cancellationScopeSource) NewScope(events chan<- engine.Event, isPreview bo
 	return c
 }
 
+func makeJSONString(v interface{}) (string, error) {
+	out := bytes.NewBuffer([]byte{})
+	encoder := json.NewEncoder(out)
```
pgavlin, line 780: The zero value for `bytes.Buffer` is a usable buffer:

```suggestion
	var out bytes.Buffer
	encoder := json.NewEncoder(&out)
```

pkg/cmd/pulumi/util.go
```diff
@@ -775,13 +775,25 @@ func (cancellationScopeSource) NewScope(events chan<- engine.Event, isPreview bo
...
+	encoder.SetEscapeHTML(false)
+	encoder.SetIndent("", "  ")
+	err := encoder.Encode(v)
+	if err != nil {
```
pgavlin, line 784:
```suggestion
	if err := encoder.Encode(v); err != nil {
```

pgavlin (approved):
LGTM overall--two minor suggestions.

pkg/cmd/pulumi/util.go
komalali: Since this function already existed and is used in many places in the CLI I'd rather not change them all here :)

PR #7597, opened 2021-07-21
Try testing for concurrent WRA

t0yv0:
# Description

Defensive check for concurrent WhileRunningAsync. Locally this does not seem to detect any violations. Let us see on CI.

Fixes # (issue)

## Checklist


t0yv0:
OK this behaves as expected and the defensive check is not needed.

PR #7598, opened 2021-07-21
Add ProviderMain function

jen20:
This commit adds a new counterpart to `ComponentMain` which accepts functional options for specifying callback functions. Currently it supports `construct` (for components) and `call` (for methods), but is extensible in a non-breaking fashion in future to support all other provider methods as they become desirable to implement.

The original `ComponentMain` still exists, though it may be desirable to deprecate it in future in favor of `ProviderMain`.

## Checklist

Existing test uses new function.

Not a user facing change.

N/A.

### DIFF
--- a/pkg/resource/provider/component_provider.go
+++ b/pkg/resource/provider/component_provider.go
@@ -32,6 +32,33 @@ type componentProvider struct {
 	version   string
 	schema    []byte
 	construct provider.ConstructFunc
+	call      provider.CallFunc
+}
+
+type Options struct {
+	Name      string
+	Version   string
+	Schema    []byte
+	Construct provider.ConstructFunc
+	Call      provider.CallFunc
+}
+
+// MainWithOptions is an entrypoint for a resource provider plugin that implements `Construct` and optionally also
+// `Call` for component resources.
+//
+// Using it isn't required but can cut down significantly on the amount of boilerplate necessary to fire up a new
+// resource provider for components.
+func MainWithOptions(opts Options) error {
+	return Main(opts.Name, func(host *HostClient) (pulumirpc.ResourceProviderServer, error) {
+		return &componentProvider{
+			host:      host,
+			name:      opts.Name,
+			version:   opts.Version,
+			schema:    opts.Schema,
+			construct: opts.Construct,
+			call:      opts.Call,
+		}, nil
+	})
 }
 
 // ComponentMain is an entrypoint for a resource provider plugin that implements `Construct` for component resources.
@@ -82,7 +109,10 @@ func (p *componentProvider) Configure(ctx context.Context,
 // Construct creates a new instance of the provided component resource and returns its state.
 func (p *componentProvider) Construct(ctx context.Context,
 	req *pulumirpc.ConstructRequest) (*pulumirpc.ConstructResponse, error) {
-	return provider.Construct(ctx, req, p.host.conn, p.construct)
+	if p.construct != nil {
+		return provider.Construct(ctx, req, p.host.conn, p.construct)
+	}
+	return nil, status.Error(codes.Unimplemented, "Construct is not yet implemented")
 }
 
 // CheckConfig validates the configuration for this provider.
@@ -155,6 +185,9 @@ func (p *componentProvider) Invoke(ctx context.Context,
 // Call dynamically executes a method in the provider associated with a component resource.
 func (p *componentProvider) Call(ctx context.Context,
 	req *pulumirpc.CallRequest) (*pulumirpc.CallResponse, error) {
+	if p.call != nil {
+		return provider.Call(ctx, req, p.host.conn, p.call)
+	}
 	return nil, status.Error(codes.Unimplemented, "Call is not yet implemented")
 }
 
--- a/tests/integration/component_provider_schema/testcomponent-go/main.go
+++ b/tests/integration/component_provider_schema/testcomponent-go/main.go
@@ -21,9 +21,14 @@ func main() {
 	if _, ok := os.LookupEnv("INCLUDE_SCHEMA"); ok {
 		schema = `{"hello": "world"}`
 	}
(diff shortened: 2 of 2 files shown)


pkg/resource/provider/component_provider.go
```diff
@@ -32,6 +32,46 @@ type componentProvider struct {
 	version   string
 	schema    []byte
 	construct provider.ConstructFunc
+	call      provider.CallFunc
+}
+
+type ProviderOption func(*componentProvider)
+
+func WithConstructFunc(f provider.ConstructFunc) ProviderOption {
```
justinvp, line 40: If we stick with functional options, should these be simply `WithConstruct` and `WithCall` without the `Func` suffix?

pkg/resource/provider/component_provider.go
```diff
@@ -82,7 +122,10 @@ func (p *componentProvider) Configure(ctx context.Context,
 // Construct creates a new instance of the provided component resource and returns its state.
 func (p *componentProvider) Construct(ctx context.Context,
 	req *pulumirpc.ConstructRequest) (*pulumirpc.ConstructResponse, error) {
-	return provider.Construct(ctx, req, p.host.conn, p.construct)
+	if p.construct != nil {
+		return provider.Construct(ctx, req, p.host.conn, p.construct)
+	}
+	return nil, status.Error(codes.Unimplemented, "Call is not yet implemented")
```
justinvp, line 128:
```suggestion
	return nil, status.Error(codes.Unimplemented, "Construct is not yet implemented")
```

pkg/resource/provider/component_provider.go
```diff
@@ -32,6 +32,46 @@ type componentProvider struct {
...
+
+// ProviderMain is an entrypoint for a resource provider plugin that implements `Construct` and optionally also
+// `Call` for component resources.
+//
+// Using it isn't required but can cut down significantly on the amount of boilerplate necessary to fire up a new
+// resource provider for components.
+//
+// `WithConstructFunc` and `WithCallFunc` can be used to provide implementations of the `Construct` and `Call` functions
+// respectively. This is also extensible for future operations without breaking changes.
+func ProviderMain(name, version string, schema []byte, funcs ...ProviderOption) error {
```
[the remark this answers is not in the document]
t0yv0: I don't have a strong sense of our Go style yet, I'd be ok with the struct at this point, but no preference.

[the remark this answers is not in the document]
t0yv0: A bit more thinking here..

I have a problem: I hack in Emacs and do not rely on completion much. The stylistic questions like this probably are best resolved with reference to IDE experience of a novice user. Which is more discoverable? Struct is probably pretty discoverable via completion.  Functions that produce ProviderOption are perhaps a tiny bit less discoverable?

Java patterns I have seen that go out of the way to provide "builders" like this:

```
ProviderBuilder().WithCallFunc().WithConstructFunc().WithSchema().Build().Main()
```

The idea is that objet method completion gives you only the options you could set, while being "fluid."

t0yv0:
LGTM, deferring to Justin on the style discussion.

pkg/resource/provider/component_provider.go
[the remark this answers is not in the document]
justinvp: > Which is more discoverable? Struct is probably pretty discoverable via completion.

I agree, another reason I lean towards the struct. (Thanks for mentioning this!)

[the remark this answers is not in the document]
pgavlin: Yeah, I'm still torn on these. The functional options style is perhaps slightly less annoying to write, but in large packages it is tough from a discoverability standpoint. I'm fine with a struct.

pkg/resource/provider/component_provider.go
jen20: I'd say no, since `WithConstruct` in this case would be used for an object that implemented the interface (similar to ServeFunc etc in the sodlib).

jen20:
I'm fine with the idea of a struct instead of functional options. I'd really like to see that approach extended to replace `ResourceOptions` too, since they are similarly undiscoverable.

pkg/resource/provider/component_provider.go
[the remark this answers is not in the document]
jen20: I've converted this to use a struct.

pkg/resource/provider/component_provider.go
```diff
@@ -32,6 +32,33 @@ type componentProvider struct {
...
+	Construct provider.ConstructFunc
+	Call      provider.CallFunc
+}
+
+// ProviderMain is an entrypoint for a resource provider plugin that implements `Construct` and optionally also
+// `Call` for component resources.
+//
+// Using it isn't required but can cut down significantly on the amount of boilerplate necessary to fire up a new
+// resource provider for components.
+func ProviderMain(opts Options) error {
```
justinvp, line 51: Looks like there's a lint warning:

> resource/provider/component_provider.go:51:6: func name will be used as provider.ProviderMain by other packages, and that stutters; consider calling this Main (golint)

We already have `Main`, so we could ignore the lint warning.

Another option is to rename this to something like `MainWithOptions` (is that the typical naming convention when adding such an "overload" in Go?).

justinvp (approved):
LGTM otherwise

pkg/resource/provider/component_provider.go
jen20: I don't really have any opinion on which way to go here.

joeduffy: `MainWithOptions` seems good to me!

emiliza: When you get a chance - can get get this last comment addressed @jen20? Would love to get this PR merged in, thanks!

jen20: I've adjusted the code and commit message, and force pushed a new commit.

justinvp (approved):
LGTM. Thanks!

Issue #3620, opened 2019-12-09
Provide a way to get all stacks from `stack ls`


lukehoban:
We recently made a change to improve the performance of `stack ls` - but as part of that also changed the correctness of `stack ls -a` to not report all stacks if there are more than 300. A warning is issued when there are more - but it does not indicate what users can do to actually get all stacks. 

We should:
1. Provide a way to actually get all stacks
2. Possibly make that the default again for `stack ls -a`
3. If we do continue to limit by default, include enough information in warning for users to actually get all stacks (or limit query by project If necessary). 

Cross ref [slack discussion](https://pulumi-community.slack.com/archives/C84L4E3N1/p1575851212127100?thread_ts=1575805569.114600&channel=C84L4E3N1&message_ts=1575851212.127100). 

Other notes:

> I would recommend adding the 300 Max to the documentation so users will not rely on it (as we did) with their automation tools.
> if i use a tag filter with the CLI will it still hide results? Cause i still see the warning. Where as if i use the project  filter i don't

jasminen:
@lukehoban Thanks for addressing this issue! 

I did some tests and indeed filtering by tag will hide some results even if result set size is < 300 . My guess is that all stacks are returned and then filtered by tag - which hits the Max 300 limit.

PR #7600, opened 2021-07-21
Paginate backend list operations

chrsmith:
# Description

Today, the "list user stacks" API that powers `pulumi stack ls` in the `httpstate` backend is not paginated. So the Pulumi Service that powers the command needs to do a lot of work, fetching all available stacks.

This has caused some problems for dear https://app.pulumi.com, because some of our more prolific customers have 10,000+ stacks! Fetching all of those for `plumi stack ls --all` in a single HTTP request will likely timeout. (This is why the Pulumi Service has a cap that returns a maximum of 300 results by default.)

We will shortly update the Pulumi Service API so the "list user stacks" API will be paginated. This means, that at the very least, we can address this problem by simply having `httpstate`/`httpstate/client` packages be smart enough to repeatedly call the API endpoint and gather up all stacks before returning.

However, this PR goes one step further, and updates the `backend.Backend` interface to support pagination natively. That is, updating methods like `ListStacks(...)` and `ListPolicyPacks(...)` to accept/return continuation tokens. This way, CLI commands will be able to render partial results while loading large sets of data. (e.g. printing stacks as they are returned, rather than waiting for all 10,000+ stacks to be loaded first.)

I looked into it, but actually implementing the "list stacks as they get loaded" would be a bit much to add to this PR. We'd need to get fancy with the virtual terminal and rerender lines if the column widths change.

Anyways, let me know how you would like to proceed.

- Do we like the design for how pagination will be supported? The naive approach would be something like `PageOptions{ PageNumber int, PageSize int }`. However, that actually hoists a lot of implementation requirements to backend implementation. So just having an opaque cursor `ContinuationToken` approach is arguably more flexible. (Read: I really do _not_ want to expose page numbers and offsets from Pulumi Service APIs.)

- Do we want to avoid updating the `backend.Backend` interface to support pagination? This is a breaking change to any custom Pulumi backend people may have built. But at the same time, it's super-straight forward to fix by just ignoring the input continuation token and returning `nil` for the output continuation token. (See the changes to `filestate` and `httpstate` backends in this PR.)

Fixes https://github.com/pulumi/pulumi/issues/3620.

## Checklist

- [⚠️ ] I have added tests that prove my fix is effective or that my feature works

I only added a test for `pulumi stack ls`, but if you'd like me to follow the same pattern for making `pulumi policy ls` and `pulumi policy group ls` testable, I'd be happy to add tests for those too.

I just am not sure if the churn/refactoring (see `stack_ls_test.go`) is worth it, but you tell me.

- [✅   ] I have updated the [CHANGELOG-PENDING]

- [✅ ] Yes, there are changes in this PR that warrants bumping the Pulumi Service API version

> Done. But note that the Pulumi Service-side of these changes hasn't been merged yet. (Doing more testing of this CLI with users that have access to bazillions of stacks.)


pkg/backend/httpstate/client/client.go
```diff
@@ -208,25 +208,27 @@ type ListStacksFilter struct {
...
 	}
 
 	var resp apitype.ListStacksResponse
 	if err := pc.restCall(ctx, "GET", "/api/user/stacks", queryFilter, nil, &resp); err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
-	return resp.Stacks, nil
+	return resp.Stacks, resp.ContinuationToken, nil
```
chrsmith, line 231: The `ContinuationToken` field was added earlier, in https://github.com/pulumi/pulumi/pull/7510.

pkg/backend/backend.go
```diff
@@ -109,6 +109,12 @@ type ListStacksFilter struct {
 	TagValue     *string
 }
 
+// ContinuationToken is an opaque string used for paginated backend requests. If non-nil, means
+// there are more results to be returned and the continuation token should be passed into a
+// subsequent call to the backend method. A nil continuation token means all results have been
+// returned.
+type ContinuationToken *string
```
pgavlin, line 116: does this need to be a `*string`? I'd imagine that an empty string would be good enough to say "no continuation token".

pkg/cmd/pulumi/policy_ls.go
```diff
@@ -51,31 +53,44 @@ func newPolicyLsCmd() *cobra.Command {
...
+					return err
+				}
+
+				allPolicyPacks = append(allPolicyPacks, resp.PolicyPacks...)
+
+				if outContToken == nil {
+					break
+				}
+				inContToken = outContToken
 			}
```
pgavlin, line 74: feels like it _might_ be worth writing a little helper to encapsulate the token pattern, e.g.

```go
func listWithContinuation(list func(token backend.ContinuationToken) (backend.ContinuationToken, error) (err error) {
	var token backend.ContinuationToken
	for {
		token, err = list(token)
		if err != nil {
			return err
		}
		if token == nil {
			return nil
		}
	}
}
```

Then callers could be something like 

```go
var allPolicyPacks []apitype.PolicyPackWithVersions
if err := listWithContinuation(func (token backend.ContinuationToken) (backend.ContinuationToken, error) {
	resp, outContToken, err := b.ListPolicyPacks(ctx, orgName, inContToken)
	allPolicyPacks = append(allPolicyPacks, resp.PolicyPacks...)
	return outContToken, err
}); err != nil {
	return err
}
```

Honestly, though, it's a pretty simple pattern and I don't feel strongly about this.

pgavlin (approved):
Two nits, LGTM otherwise

pkg/backend/backend.go
chrsmith: I went back and forth a lot on this. The empty string does seem reasonable enough to imply no continuation token. But I erred on keeping this a pointer because `if token == nil` seemed slightly preferable to `if token == ""`, IMHO.

But if you really don't like the design I can change it.

pkg/cmd/pulumi/policy_ls.go
chrsmith: That's a reasonable suggestion, but I'm inclined to keep this as-is. We only would use this in 3x places, and one of those will 🤞 be replaced with a much slicker "output stacks while we receive them" rather than batching up all the results synchronously before proceeding.

chrsmith:
Rebased with master and force-pushed. (Since the CLI was shipped since the PR was created, there were conflicts in `CHANGELOG_PENDING.md.)

PR #7604, opened 2021-07-21
[sdk/go] Fix methods panic when marshaling `self`

### LINKED ISSUE #7603: [sdk/go] Panic when calling a method on a resource that implements `pulumi.Input`
We're seeing a panic when calling a method on a resource that implements `pulumi.Input`, which will be the case for all code-generated resources.

```
panic: reflect: nil type passed to Type.AssignableTo
goroutine 39 [running]:
reflect.(*rtype).AssignableTo(0x18d7ca0, 0x0, 0x0, 0xc0005be310)
	/usr/local/go/src/reflect/type.go:1445 +0xd3
github.com/pulumi/pulumi/sdk/v3/go/pulumi.marshalInputAndDetermineSecret(0x18d7ca0, 0xc0005be310, 0x0, 0x0, 0x1, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/Users/justin/go/src/github.com/pulumi/pulumi/sdk/go/pulumi/rpc.go:206 +0x31a5
github.com/pulumi/pulumi/sdk/v3/go/pulumi.marshalInput(0x18d7ca0, 0xc0005be310, 0x0, 0x0, 0x1, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/Users/justin/go/src/github.com/pulumi/pulumi/sdk/go/pulumi/rpc.go:171 +0x73
github.com/pulumi/pulumi/sdk/v3/go/pulumi.(*Context).Call.func1(0x0, 0x0, 0x0)
	/Users/justin/go/src/github.com/pulumi/pulumi/sdk/go/pulumi/context.go:367 +0x236
github.com/pulumi/pulumi/sdk/v3/go/pulumi.(*Context).Call.func2(0xc0005d8000, 0x1954dda, 0x28, 0x1a36f70, 0xc0005be460, 0xc000112640)
	/Users/justin/go/src/github.com/pulumi/pulumi/sdk/go/pulumi/context.go:457 +0x122
created by github.com/pulumi/pulumi/sdk/v3/go/pulumi.(*Context).Call
	/Users/justin/go/src/github.com/pulumi/pulumi/sdk/go/pulumi/context.go:421 +0x225
```

justinvp:
Pass a valid `destType` to `marshalInput` to avoid panicing when the `self` resource implements `pulumi.Input`.

This issue would have been caught by testing the generated code, in addition to comparing it to what is expected.

Fixes #7603

### DIFF
--- a/sdk/go/pulumi/context.go
+++ b/sdk/go/pulumi/context.go
@@ -364,7 +364,7 @@ func (ctx *Context) Call(tok string, args Input, output Output, self Resource, o
 		// If we have a value for self, add it to the arguments.
 		if self != nil {
 			var deps []URN
-			resolvedSelf, selfDeps, err := marshalInput(self, anyType, true)
+			resolvedSelf, selfDeps, err := marshalInput(self, reflect.TypeOf(self), true)
 			if err != nil {
 				return nil, fmt.Errorf("marshaling __self__: %w", err)
 			}
--- a/sdk/go/pulumi/context.go
+++ b/sdk/go/pulumi/context.go
@@ -364,7 +364,7 @@ func (ctx *Context) Call(tok string, args Input, output Output, self Resource, o
 		// If we have a value for self, add it to the arguments.
 		if self != nil {
 			var deps []URN
-			resolvedSelf, selfDeps, err := marshalInput(self, nil, true)
+			resolvedSelf, selfDeps, err := marshalInput(self, anyType, true)
 			if err != nil {
 				return nil, fmt.Errorf("marshaling __self__: %w", err)
 			}
--- a/tests/integration/construct_component_methods/go/main.go
+++ b/tests/integration/construct_component_methods/go/main.go
@@ -72,6 +72,10 @@ func (o ComponentGetMessageResultOutput) Message() pulumi.StringOutput {
 	return o.ApplyT(func(v ComponentGetMessageResult) string { return v.Message }).(pulumi.StringOutput)
 }
 
+func (*Component) ElementType() reflect.Type {
+	return reflect.TypeOf((*Component)(nil))
+}
+
 func init() {
 	pulumi.RegisterOutputType(ComponentGetMessageResultOutput{})
 }
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -17,6 +17,9 @@
 
 - [sdk/go] - Fix target and replace options for the Automation API.
   [#7426](https://github.com/pulumi/pulumi/pull/7426)
-  
+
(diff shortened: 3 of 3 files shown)


tests/integration/construct_component_methods/go/main.go
```diff
@@ -72,6 +72,10 @@ func (o ComponentGetMessageResultOutput) Message() pulumi.StringOutput {
 	return o.ApplyT(func(v ComponentGetMessageResult) string { return v.Message }).(pulumi.StringOutput)
 }
 
+func (*Component) ElementType() reflect.Type {
+	return reflect.TypeOf((*Component)(nil))
+}
```
justinvp, line 77: The test fails after adding this (making `*Component` implement `pulumi.Input`), and succeeds after the fix in `context.go`.

sdk/go/pulumi/context.go
```diff
@@ -364,7 +364,7 @@ func (ctx *Context) Call(tok string, args Input, output Output, self Resource, o
 		// If we have a value for self, add it to the arguments.
 		if self != nil {
 			var deps []URN
-			resolvedSelf, selfDeps, err := marshalInput(self, nil, true)
+			resolvedSelf, selfDeps, err := marshalInput(self, anyType, true)
```
pgavlin, line 367: Feels like we ought to have a better type than `anyType` for `resolvedSelf`. Is there a reason that we can't use the type of the resource itself?

justinvp: Ah, fair point. It doesn't really make a difference here and I was just following [some](https://github.com/pulumi/pulumi/blob/ece9f2fb30598ed4796214f2c91290f31a63a742/sdk/go/pulumi/context.go#L246) [other](https://github.com/pulumi/pulumi/blob/ece9f2fb30598ed4796214f2c91290f31a63a742/sdk/go/pulumi/context.go#L1400) [cases](https://github.com/pulumi/pulumi/blob/ece9f2fb30598ed4796214f2c91290f31a63a742/sdk/go/pulumi/rpc.go#L261) where we use `anyType` with `marshalInput` when the type doesn't matter, but I don't see any reason we can't use the type of the resource itself here. Fixed.

pgavlin (approved):

tests/integration/construct_component_methods/go/main.go
t0yv0: Input is such a scary interface. I'm developing an irrational fear that I'm not sure what happens when we declare something an input. I need to work on making this specific :)

t0yv0 (approved):
LGTM with a bit of unfortunate FUD about inputs.

PR #7606, opened 2021-07-22
[codegen/dotnet] Emit resource methods

justinvp:
Part of #7072

---

Note: There's currently a bug when a method's inputs references a nested type. For example:

```csharp
    public sealed class BarArgs : Pulumi.CallArgs
    {
        [Input("baz")]
        public Input<Pulumi.Example.Nested.Inputs.BazArgs>? Baz { get; set; }

        [Input("bazPlain")]
        public Pulumi.Example.Nested.Inputs.BazArgs? BazPlain { get; set; }

        [Input("bazRequired", required: true)]
        public Input<Pulumi.Example.Nested.Inputs.BazArgs> BazRequired { get; set; } = null!;
    }
```

`BarArgs` is referencing `Pulumi.Example.Nested.Inputs.BazArgs`, but that type is not being emitted, instead we're getting this:

```csharp
namespace Pulumi.Example.Nested.Inputs
{
    public sealed class Baz : Pulumi.InvokeArgs
    {
        [Input("hello")]
        public string? Hello { get; set; }

        [Input("world")]
        public string? World { get; set; }

        public Baz()
        {
        }
    }
}
```

I haven't had a chance to look into this deeply to understand the root cause. It's something to do with how we generate nested types for function inputs and currently it always emits the plain shape rather than the input shape. Need to generate the input shape if it's used by a method.

I'd sort of like to merge the change in this PR as-is (modulo any review feedback) which represents the core method support for .NET, and file an issue about the nested type and address that subsequently. Edit: #7651


pkg/codegen/dotnet/gen.go
```diff
@@ -1081,6 +1143,70 @@ func (mod *modContext) genResource(w io.Writer, r *schema.Resource) error {
 		}
 	}
 
+	// Generate method types.
+	genMethodTypes := func(method *schema.Method) error {
```
justinvp, line 1147: I was originally planning to emit the `<Method>Args` and `<Method>Result` classes as nested classes within the resource's class, e.g.:

```csharp
public class MyResource : ComponentResource
{
    public Output<FooResult> Foo(FooArgs args)
        => ...

    public sealed class FooArgs : Pulumi.CallArgs
    {
        ...
    }

    public sealed class FooResult
    {
    }
}
```

Which is similar to what we've done for Node.js and Python, and appropriately "namespaces" the types to avoid any naming conflicts for similarly named methods on different resources within the same Pulumi module.

However, using these as nested classes from .NET is more annoying which isn't an issue in practice with TS/JS and Python. In .NET you have to do:

```
var res = new MyResource(...);
var result = res.Foo(MyResource.FooArgs(...)); // Kinda annoying
```

When it'd be much nicer to use as:

```
var res = new MyResource(...);
var result = res.Foo(FooArgs(...)); // Nicer
```

This PR currently does the latter nicer approach (non-nested).

**What do folks think?** 

There is some risk that within a Pulumi module, there could be resources that have methods with the same name which then would result in conflicts. If that becomes a problem, we'd have to come up with a way to address it, either by nesting the classes, prefixing the resource name (e.g. `<Resource><Method>Args`), or providing a way to customize the naming from the schema. However, I wonder how common such conflicts would be in practice (same method name on multiple resources in the same module).

Note that in Go, we currently emit these prefixed as `<Resource><Method>Args` and `<Resource><Method>Result`, to prevent such conflicts, but this is also annoying to use, so if we proceed with the unnested, un-resource-prefixed names in C#, I'll open a PR to remove the `<Resource>` prefix from the Go codegen.

justinvp: On the other hand, there's already precedent for prefixing the resource name for nested types, e.g. `Bucket`'s `Website` property accepts a `BucketWebsiteArgs` type. This is making me lean towards making the args/result types non-nested and prefixed with the resource name, e.g.:

```csharp
var res = new MyResource(...);
var result = res.Foo(new MyResourceFooArgs(...));
```

It isn't as nice to use, but it's more consistent with what we've done elsewhere and reduces the likelihood of conflicts.

t0yv0: Flattening them out into method params without an intermediate class is out of question? After all these are methods, so one expects to call them without much ceremony? Most of these will have 1-2 parameters?

pkg/codegen/internal/test/testdata/simple-methods-schema/dotnet/Utilities.cs
```diff
@@ -0,0 +1,87 @@
...
+                {
+                    return value;
+                }
+            }
+            return null;
+        }
+
+        static string[] trueValues = { "1", "t", "T", "true", "TRUE", "True" };
+        static string[] falseValues = { "0", "f", "F", "false", "FALSE", "False" };
+        public static bool? GetEnvBoolean(params string[] names)
```
t0yv0, line 28: Interesting what this file is doing here :) Some artifact of codegen I assume.

pkg/codegen/internal/test/testdata/simple-methods-schema/dotnet/Foo.cs
```diff
@@ -0,0 +1,150 @@
...
+            var merged = ComponentResourceOptions.Merge(defaultOptions, options);
+            // Override the ID if one was specified for consistency with other language SDKs.
+            merged.Id = id ?? merged.Id;
+            return merged;
+        }
+
+        /// <summary>
+        /// A description of bar.
+        /// </summary>
+        public Pulumi.Output<FooBarResult> Bar(FooBarArgs args)
```
t0yv0, line 42: Should this be `FooBarArgs? args` here? Since `args ?? new FooBarArgs()` is trying to deal with nullables?

pkg/codegen/internal/test/testdata/simple-methods-schema/dotnet/Foo.cs
```diff
@@ -0,0 +1,150 @@
...
+        ///
+        /// <param name="name">The unique name of the resource</param>
+        /// <param name="args">The arguments used to populate this resource's properties</param>
+        /// <param name="options">A bag of options that control this resource's behavior</param>
+        public Foo(string name, FooArgs? args = null, ComponentResourceOptions? options = null)
+            : base("example::Foo", name, args ?? new FooArgs(), MakeResourceOptions(options, ""), remote: true)
+        {
+        }
+
+        private static ComponentResourceOptions MakeResourceOptions(ComponentResourceOptions? options, Input<string>? id)
```
t0yv0, line 27: This could go in SDK instead of codegen'd code or no?

t0yv0 (approved):
Code itself LGTM - these codegen changes are always a bit spooky to review, but the test seems to cover "no arg" and "lots of arg" cases, and I tried to look through the code to notice if anything was off.

I think since this does not possibly break providers the bar is a bit lower on trying to find all bugs appriori 🚢

pkg/codegen/dotnet/gen.go
justinvp: > Flattening them out into method params without an intermediate class is out of question? After all these are methods, so one expects to call them without much ceremony? Most of these will have 1-2 parameters?

It's a great question. From a usage standpoint, I'd love for these to be params. I think the main downside is we don't provide much control over the order of these as specified in the schema, which makes adding new optional params in a compatible way problematic. I believe this is why we've always emitted args classes for functions, rather than flattened into params -- so what we're doing for methods is just consistent with the approach we've taken for functions.

Note that we do flatten these to method args in Python only, but force the args to be specified by keyword rather than positionally, so if new args are added, existing code continues to work. For .NET, there's no way to force params to be specified by name, though.

For methods, we could decide to be different from what we've done for functions, and flatten them out into method params, and just live with the limited versionablity, i.e. if you want to add a new optional arg in a compatible way, you'd have to add a new method instead of adding it to an existing method.

pkg/codegen/internal/test/testdata/simple-methods-schema/dotnet/Foo.cs
justinvp: No, the signature is right. There are required inputs for this method, so it should be `FooBarArgs`. The `args ?? new FooBarArgs()` is the same thing we do for functions:

https://github.com/pulumi/pulumi/blob/4a035c44aa80a31be0af942db438b430e3ed3a57/pkg/codegen/dotnet/gen.go#L1116

While the C# compiler will help prevent you from passing `null` here with warnings, someone could force it, or a null value could come in from another language, so this is ensures it's initialized to an instance of the appropriate `Args` class. When the args class is serialized, it will throw errors if a required input isn't set:

https://github.com/pulumi/pulumi/blob/4a035c44aa80a31be0af942db438b430e3ed3a57/sdk/dotnet/Pulumi/Resources/InputArgs.cs#L57-L60

pkg/codegen/internal/test/testdata/simple-methods-schema/dotnet/Utilities.cs
justinvp: We just always emit this file in full for SDKs, and don't filter out its members if they aren't used.

pkg/codegen/internal/test/testdata/simple-methods-schema/dotnet/Foo.cs
justinvp: To reduce the size of the generated code? I don't see why not aside from having to require a version of the `Pulumi` package where it exists. Might be easier to move it to the `Utilities` class that gets generated, so then there's only one emitted rather than one for each resource class.

t0yv0: Yes, pretty much reducing the size of generated code is what I have in mind here, but this might be negligible. Which package it belongs will have more to do with how stable the logic is and where it's best versioned (which probably is here, not SDK). "only one emitted rather than one for each resource class." that sounds good if there's a material reduction on SDK size, which again there probably isn't. Thanks.

pkg/codegen/dotnet/gen.go
t0yv0: I think consistency of methods and functions is important to keep, as well as consistency across languages. I filed the issue citing your excellent notes in https://github.com/pulumi/pulumi/issues/7664 - it probably has no legs but just in case we one day find a way forward on this, I wanted to keep a note in the tracker.

PR #7617, opened 2021-07-22
Demonstrate depends_on ignoring ComponentResource

t0yv0:
# Description

Passing ComponentResource to depends_on ignores its URN when talking to the engine. 

Is this a bug or a feature?

## Checklist


t0yv0:
Copying discussion from elsewhere. It was brought up that ComponentResources are "just containers" so maybe this is by design, and instead dependsOn(componentResource) should send dependencies on `componentResource.Children` or some such. On the other hand Justin brought up https://github.com/pulumi/pulumi/pull/7541 so maybe we want this after all?

I'm leaving it here to have another look with fresh eyes next week.

t0yv0:
Looks like it's the behavior we want.

PR #7624, opened 2021-07-23
Rehydrate provider resources in `Construct`

### LINKED ISSUE #7231: [go] Provider resources passed to multi-lang components cannot be recovered as (e.g.) `*aws.Provider`
The following code inside a multi-language component results in the error below.

```go
	providerInstance := component.GetProvider("aws::")
	provider := providerInstance.(*aws.Provider)
	component.AWSProfile = provider.Profile
```

```
panic: interface conversion: pulumi.ProviderResource is *pulumi.ProviderResourceState, not *aws.Provider
```

This prevents being able to access the provider configuration off of the provider instance passed in by the user.

justinvp:
Previously, any provider resource passed to multi-lang components would be instantiated as a `DependencyProviderResource` inside `Construct`, which prevents the component from being able to easily access the provider's state as an instance of of the provider (e.g. `*aws.Provider`).

This change attempts to rehydrate the provider resource in the same way that resource references are rehydrated, if it's been registered, s.t. the specific provider resource type is instantiated with its state. Otherwise falling back to returning `DependencyProviderResource`.

Broken up into separate commits to make it a little easier to review:

1. [Go](https://github.com/pulumi/pulumi/commit/cc5fe0a650b32c56a6c3dd4876713c326dee9af1)
2. [Node.js](https://github.com/pulumi/pulumi/commit/db4899055ac606e4f45d5b31d28c48e1e36305d3)
3. [Python](https://github.com/pulumi/pulumi/commit/895bcb0501db5239062aff32385f12c0f53f4b81)

Fixes #7231


sdk/nodejs/resource.ts
```diff
@@ -1037,15 +1037,22 @@ export class DependencyProviderResource extends ProviderResource {
...
         (<any>this).urn = new Output(<any>this, Promise.resolve(urn), Promise.resolve(true), Promise.resolve(false), Promise.resolve([]));
         (<any>this).id = new Output(<any>this, Promise.resolve(id), Promise.resolve(true), Promise.resolve(false), Promise.resolve([]));
     }
 }
+
+/**
+ * parseResourceReference parses the URN and ID out of the provider reference.
+ * @internal
+ */
+export function parseResourceReference(ref: string): [string, string] {
```
t0yv0, line 1050: 👍 separate func makes it readable at the callsite

tests/integration/construct_component_provider/testcomponent-go/main.go
```diff
@@ -0,0 +1,88 @@
...
+func NewComponent(ctx *pulumi.Context, name string, opts ...pulumi.ResourceOption) (*Component, error) {
+	component := &Component{}
+	err := ctx.RegisterComponentResource("testcomponent:index:Component", name, component, opts...)
+	if err != nil {
+		return nil, err
+	}
+
+	// Test that we're indeed getting back an instance of `Provider` with its state.
+	provider := component.GetProvider("testcomponent::").(*Provider)
+	component.Message = provider.Message
```
t0yv0, line 37: This sort of assignment is implied in base constructor call in other langs?

t0yv0:
I think I can follow a little - the tests assert that Message is populated but it only happens through the provider, not directly. What surprised me is that there seems to be no state handling in the diff, beyond (resName, resType, URN). I think the changes make sure that the ConstructProvider call happens:

```
			return resourcePackage.ConstructProvider(ctx, resName, string(resType), string(ref.URN))
```

That seems to build an instance with the right functionality via the provider map - so far so good; and it calls ctx.registerResource, which calls ctx.getResource which gets state back from the engine? This is the mechanism that gets the state out?

LGTM

t0yv0 (approved):

tests/integration/construct_component_provider/testcomponent-go/main.go
justinvp: The other langs do it explicitly in their constructors:

Python:

https://github.com/pulumi/pulumi/pull/7624/files#diff-357f441447e8aab4d038abbb5a01d00141a8e9d20d5651e8fc9dae378a075f7dR46

```python
         provider = self.get_provider("testcomponent::")
         if not isinstance(provider, Provider):
             raise Exception(f"provider is not an instance of Provider: {provider}")
         self.message = provider.message
         self.register_outputs({
             "message": self.message,
         })
```

Node.js:

https://github.com/pulumi/pulumi/pull/7624/files#diff-88176009303a404a3c6fa167d388546cd26c977e414ff1b342afe5ad0b8cf4d1R28

```typescript
         const provider = this.getProvider("testcomponent::");
         if (!(provider instanceof Provider)) {
             throw new Error("provider is not an instance of Provider");
         }
         this.message = provider.message;
         this.registerOutputs({
             message: provider.message,
         });
```

Node.js differs slightly with it happening in `initialize` rather than the constructor itself, following the pattern from #3676.

justinvp:
> This is the mechanism that gets the state out?

Yep!

PR #7628, opened 2021-07-23
Fanout build experiment

t0yv0:
An experiment in spreading test execution across GHA workers to speed up CI time.

- [ ] Windows build is not yet split


t0yv0:
Tagging interested parties. This brute-force method achieves total CI time speedup (let's wait out to measure again though). 

The one number I've seen was 36 min (vs 1h 30 min baseline).

The downsides are pretty significant waste in repeated build+test on the parallel workers. If GHA action time translates into cost for us, there is additional cost here. 

I'm curious if we want to push forward with this, I could work on propagating the same `go test` wrapper into the Windows build.

The idea would be that this is a temporary solution that alleviates a real pain point until the proper solution of the factored build arrives.

t0yv0:
53 min on macos-lastest with testset = integration is the critical path.

This does not get us to 30 unfortunately.

Next step I think would be to double run the native tests, move them off the crit path.

t0yv0:
Errgh, breaking out native tests into their own runner, "integration" still critical path with 49m 19s on mac, 45m on Linux. Need to try to break them up (by language?).

t0yv0:
45 min with a bit more careful bin-packing ..

EvanBoyle (approved):
🎉  let's give it a try!

PR #7630, opened 2021-07-24
Set the package on DependencyProviderResource

### LINKED ISSUE #6895: `DependencyProviderResource` should specify a valid package string instead of blank value
When initializing a `DependencyProviderResource`, blank values are passed to the base class's constructor:

https://github.com/pulumi/pulumi/blob/eb2dcf24884bab173af175cd41d8279e07797a75/sdk/nodejs/resource.ts#L1007-L1009

`ProviderResource`'s constructor:

https://github.com/pulumi/pulumi/blob/eb2dcf24884bab173af175cd41d8279e07797a75/sdk/nodejs/resource.ts#L750-L753

So an empty string is passed for `pkg` which will be returned from the `ProviderResource.getPackage()` method.

This may cause issues if the `DependencyProviderResource` is added to `ComponentResourceOptions` passed to other components in a `Construct` call, as the `pkg` is used as part of normalizing the providers:

https://github.com/pulumi/pulumi/blob/eb2dcf24884bab173af175cd41d8279e07797a75/sdk/nodejs/resource.ts#L358

https://github.com/pulumi/pulumi/blob/eb2dcf24884bab173af175cd41d8279e07797a75/sdk/nodejs/resource.ts#L952

We should be able to parse the package from the passed-in URN for `DependencyProviderResource`.

We should also consider specifying the type for `DependencyResource` from the URN, instead of passing a blank value to the base class's constructor.

This applies to all language SDKs.

justinvp:
When initializing `DependencyProviderResource` (Node.js, Python, and .NET), pass the package to the base constructor instead of an empty string s.t. the provider is usable when its package is read.

Note that the Go SDK was already doing this (line 133):

https://github.com/pulumi/pulumi/blob/eaf78edfef11c7ab97dc90b2bafd3c9fcc7eedf7/sdk/go/pulumi/resource.go#L127-L135

... so this is making the other language SDKs consistent.

Also worth noting that the likelihood of this being a problem is much less now after #7624, but it's still worth fixing because it's still possible to fallback to `DependencyProviderResource` when we can't rehydrate a provider.

Fixes #6895

### DIFF
--- a/sdk/dotnet/Pulumi/Resources/DependencyProviderResource.cs
+++ b/sdk/dotnet/Pulumi/Resources/DependencyProviderResource.cs
@@ -14,15 +14,9 @@ namespace Pulumi
     internal sealed class DependencyProviderResource : ProviderResource
     {
         public DependencyProviderResource(string reference)
-            : base(package: "", name: "", args: ResourceArgs.Empty, dependency: true)
+            : base(package: GetPackage(reference), name: "", args: ResourceArgs.Empty, dependency: true)
         {
-            var lastSep = reference.LastIndexOf("::", StringComparison.Ordinal);
-            if (lastSep == -1)
-            {
-                throw new ArgumentException($"Expected \"::\" in provider reference ${reference}.");
-            }
-            var urn = reference.Substring(0, lastSep);
-            var id = reference.Substring(lastSep + 2);
+            var (urn, id) = ParseReference(reference);
 
             var resources = ImmutableHashSet.Create<Resource>(this);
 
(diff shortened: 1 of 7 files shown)


sdk/dotnet/Pulumi/Resources/DependencyProviderResource.cs
```diff
@@ -32,5 +26,29 @@ public DependencyProviderResource(string reference)
             var idData = OutputData.Create(resources, id, isKnown: true, isSecret: false);
             this.Id = new Output<string>(Task.FromResult(idData));
         }
+
+        private static string GetPackage(string reference)
+        {
+            var (urn, _) = ParseReference(reference);
+            var urnParts = urn.Split("::");
```
justinvp, line 33: We really should consolidate the URN parsing in Node.js, Python, and .NET, similar to [what we have in Go](https://github.com/pulumi/pulumi/blob/master/sdk/go/common/resource/urn.go). I considered doing that as part of this, but it was looking like a larger change than I wanted to bite off.

emiliza:
@pgavlin eyes please!

pgavlin (approved):

t0yv0:
Sorry late review here, LGTM.

PR #7631, opened 2021-07-24
Allow setting plugin name and version for Node.js

jen20:
# Description

This commit adds two new fields to the Node package info struct to permit setting the plugin name if it differs from the package name, and the version if it differs from the package version. This was already supported by the loader.

## Checklist

No test coverage of this area.

Not a user-facing change.

N/A.

### DIFF
--- a/pkg/codegen/nodejs/importer.go
+++ b/pkg/codegen/nodejs/importer.go
@@ -52,8 +52,12 @@ type NodePackageInfo struct {
 	DisableUnionOutputTypes bool `json:"disableUnionOutputTypes,omitempty"`
 	// An indicator for whether the package contains enums.
 	ContainsEnums bool `json:"containsEnums,omitempty"`
-	// A map allowing you to map the name of a provider to the name of the module encapsulating the provider
+	// A map allowing you to map the name of a provider to the name of the module encapsulating the provider.
 	ProviderNameToModuleName map[string]string `json:"providerNameToModuleName,omitempty"`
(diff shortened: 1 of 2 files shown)


pkg/codegen/nodejs/importer.go
```diff
@@ -52,8 +52,12 @@ type NodePackageInfo struct {
 	DisableUnionOutputTypes bool `json:"disableUnionOutputTypes,omitempty"`
 	// An indicator for whether the package contains enums.
 	ContainsEnums bool `json:"containsEnums,omitempty"`
-	// A map allowing you to map the name of a provider to the name of the module encapsulating the provider
+	// A map allowing you to map the name of a provider to the name of the module encapsulating the provider.
 	ProviderNameToModuleName map[string]string `json:"providerNameToModuleName,omitempty"`
+	// The name of the plugin, which might be different from the package name.
+	PluginName string `json:"pluginName,omitempty"`
+	// The version of the plugin, which might be different from the version of the package..
+	PluginVersion string `json:"pluginVersion,omitempty"`
```
justinvp, line 60: Do you have an example of a situation where this would be used?

jen20: Largely for if you publish a package from changed language stubs that does not require a changed provider binary (and thus users don't need to download a potentially large binary again). All this does is expose existing functionality for use, not adding the actual functionality.

justinvp (approved):

PR #7632, opened 2021-07-24
Do not build multiple times in CI

t0yv0:
Let's try to make more sense out of our makefiles and CI interaction.

# The Problem

Before the changes there's evidence we're building twice when CI runs `make test_all`:

```
[nix-shell:~/pulumi]$ make --debug=b test_all -n | grep 'bdist_wheel'
make --debug=b test_all -n | grep 'bdist_wheel'
cd ./env/src && pipenv run python setup.py build bdist_wheel --universal
cd ./env/src && pipenv run python setup.py build bdist_wheel --universal

[nix-shell:~/pulumi]$ make --debug=b test_all -n | grep 'dotnet build'
make --debug=b test_all -n | grep 'dotnet build'
dotnet build dotnet.sln /p:Version=3.9.0-alpha.1627080558+fbf44bfa
dotnet build dotnet.sln /p:Version=3.9.0-alpha.1627080558+fbf44bfa
```

This happens because top-level Make recurs twice, two separate sub-processes of Make, and they cannot deduplicate the 'build' target between 'install' and 'test_all' targets.

The reason for this is interplay of rules from `common.mk` and extra specifiers on the Makefile.

# The Fix

I am attempting to move the rule that top-level "test_all" needs "sdk/*/install" from Makefile to common.mk, introducing a helper target "install_test_all", and effectively deduplicating the double recursive Make call.

I also attempted to remove the separate "install" step from CI since we're going to run it anyway, but that failed. Looks like test_all is missing a dependency on some installation steps.


t0yv0:
Closing in favor of https://github.com/pulumi/pulumi/pull/7638

Issue #3468, opened 2019-11-07
Unable to assign unique name to Dynamic providers in Pulumi Python


Ralnoc:
While building dynamic providers, it would be extremely helpful to be able to identify which provider is actually implementing an action being taken. Unfortunately, at least in the Pulumi Python, all Dynamic Providers are typed as `pulumi-python:dynamic:Resource`. 

First, I wanted to check to see if I missed something that would allow us to change that identifier. However if it is currently not possible, it would be extremely helpful to be able to do so.

pgavlin:
It is not currently possible to change that identifier, no.

However, it would be a relatively easy change to allow you to specify the last piece of the type (i.e. `Random`). Would that be sufficient for your use case?

pgavlin:
(we can't change the first or second pieces of the type token because they're used by the plugin system to locate which plugin to use for the resource)

Ralnoc:
Absolutely! Allow us to assign a unique name to the final name for the type would work great.

schollii:
@pgavlin I'm not familiar with the pulumi code base but I've programed in python and many other languages for many years so I might be able to help implement this. If you can put a link to a file where this label is currently implemented, I can have a look.

Asamsig:
This was feature implemented [here](https://github.com/pulumi/pulumi/pull/7633).
See this [comment](https://github.com/pulumi/pulumi/pull/7633#issuecomment-920472853) for further confirmation.

So this issue can be closed.

iwahbe:
Fixed by https://github.com/pulumi/pulumi/pull/7633

PR #7633, opened 2021-07-24
Added support for custom naming of dynamic provider resource

jancespivo:
# Description

Now there is not possible to change a name of dynamic provider resource without copying a code of the `pulumi.sdk.python.lib.pulumi.dynamic.dynamic.Resource` and changing the hard-coded name `"pulumi-python:dynamic:Resource"`. I successfully uses this proposal to make it possible.

Usage:
```python
class CustomResource(
    Resource, name="my-custom-provider:CustomResource"
):
   ...
```

## Checklist

### DIFF
--- a/sdk/python/lib/pulumi/dynamic/dynamic.py
+++ b/sdk/python/lib/pulumi/dynamic/dynamic.py
@@ -242,6 +242,9 @@ class Resource(CustomResource):
     Resource represents a Pulumi Resource that incorporates an inline implementation of the Resource's CRUD operations.
     """
 
+    def __init_subclass__(cls, name: str ='dynamic:Resource'):
+        cls._resource_type_name = name
+
     def __init__(self,
                  provider: ResourceProvider,
                  name: str,
@@ -261,4 +264,4 @@ class Resource(CustomResource):
         props = cast(dict, props)
         props[PROVIDER_KEY] = serialize_provider(provider)
 
-        super().__init__("pulumi-python:dynamic:Resource", name, props, opts)
+        super().__init__(f"pulumi-python:{self._resource_type_name}", name, props, opts)
--- a/sdk/python/lib/pulumi/dynamic/dynamic.py
+++ b/sdk/python/lib/pulumi/dynamic/dynamic.py
@@ -242,8 +242,10 @@ class Resource(CustomResource):
     Resource represents a Pulumi Resource that incorporates an inline implementation of the Resource's CRUD operations.
     """
 
-    def __init_subclass__(cls, name: str ='dynamic:Resource'):
-        cls._resource_type_name = name
+    def __init_subclass__(cls, module: str = '', name: str ='Resource'):
+        if module != '':
+            module = '/' + module
+        cls._resource_type_name = f"dynamic{module}:{name}"
 
     def __init__(self,
                  provider: ResourceProvider,
--- a/sdk/python/lib/pulumi/dynamic/dynamic.py
+++ b/sdk/python/lib/pulumi/dynamic/dynamic.py
@@ -15,7 +15,7 @@
 import asyncio
 import base64
 import pickle
-from typing import Any, Optional, List, TYPE_CHECKING, no_type_check, cast
+from typing import Any, ClassVar, Optional, List, TYPE_CHECKING, no_type_check, cast
 
 import dill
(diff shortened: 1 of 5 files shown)


lukehoban:
Thanks for the PR!  I believe this would address https://github.com/pulumi/pulumi/issues/3468.  We likely want to add the same support for Node.js as well as part of any update here.

sdk/python/lib/pulumi/dynamic/dynamic.py
```diff
@@ -234,6 +234,9 @@ class Resource(CustomResource):
     Resource represents a Pulumi Resource that incorporates an inline implementation of the Resource's CRUD operations.
     """
 
+    def __init_subclass__(cls, name: str ='dynamic:Resource'):
+        cls._resource_type_name = name
+
```
pgavlin, line 239: I'd prefer that we keep the `dynamic` portion of the type token. Removing that portion of the token makes it dangerous for us to add any additional builtin resources to the `pulumi-python` provider. If you want to change the module, we could instead do something like this:

```suggestion
    def __init_subclass__(cls, module: str = '', name: str ='Resource'):
        if module != '':
            module = '/' + module
        cls._resource_type_name = f"dynamic{module}:{name}"

```

This would let you form type tokens like `pulumi-python:dynamic/my-module:Resource`. The `pulumi-python` provider would then be able to safely add other resources outside of the `dynamic` namespace in the future.

pgavlin (changes requested):
Can we add some tests for this as well? An integration test like [`tests/integration/dynamic/python`](https://github.com/pulumi/pulumi/tree/master/tests/integration/dynamic/python) should work. To add a test, you can just copy that folder to a new folder (e.g. `tests/integration/dynamic/python-type-name`) and add a new test method to `tests/integration/integration_python_test.go`, e.g. by copying and modifying the method for [`tests/integration/dynamic/python`](https://github.com/pulumi/pulumi/blob/master/tests/integration/integration_python_test.go#L389-L408).

Please let me know if you need any more info on how to add a test!

@lukehoban I think that we can file a follow-up issue for Node.

sdk/python/lib/pulumi/dynamic/dynamic.py
```diff
@@ -234,6 +234,11 @@ class Resource(CustomResource):
     Resource represents a Pulumi Resource that incorporates an inline implementation of the Resource's CRUD operations.
     """
 
+    def __init_subclass__(cls, module: str = '', name: str ='Resource'):
```
pgavlin, line 237:
```suggestion
    def __init_subclass__(cls, module: str = '', name: str = 'Resource'):
```

pgavlin (approved):
One nit, but LGTM otherwise!

jancespivo:
@pgavlin I've added a test. It is my first contribution in go and I don't know how to run tests. I've tried to run `go test` in the integration directory, but it print `PASS` regardless the new test should pass or not. Please help :) Thx

tests/integration/integration_python_test.go
```diff
@@ -407,6 +407,25 @@ func TestDynamicPython(t *testing.T) {
...
+	var randomVal string
+	integration.ProgramTest(t, &integration.ProgramTestOptions{
+		Dir: filepath.Join("dynamic", "python-resource-type-name"),
+		Dependencies: []string{
+			filepath.Join("..", "..", "sdk", "python", "env", "src"),
+		},
+		ExtraRuntimeValidation: func(t *testing.T, stack integration.RuntimeValidationStackInfo) {
+			urn_s = stack.Outputs["urn"].(string)
+			urn := resource.URN(urn_s)
+	        typ := urn.Type()
```
pgavlin, line 422:
```suggestion
			urnOut = stack.Outputs["urn"].(string)
			urn := resource.URN(urnOut)
			typ := urn.Type()
```

pgavlin (approved):
One small set of suggestions for the test changes, but this looks great!

You can ensure that your test runs by running `go test -tags=all -run TestCustomResourceTypeNameDynamicPython ` from within the `tests/integration` directory.

infin8x:
@pgavlin @lukehoban it looks like this did indeed fix https://github.com/pulumi/pulumi/issues/3468 - shall I resolve it or is there more work tracked there?

PR #7638, opened 2021-07-26
Avoid triple-building projects in the solution

t0yv0:
# Description

Another experiment in speeding up builds. The idea is to add a switch that makes `test_all` target avoid installing, since CI runs `install_all` before `test_all`, to avoid triple-building-and-installing.

Before:

```
$ make --debug=b TEST_ALL_DEPS= install_all -n | grep 'bdist_wheel'
cd ./env/src && pipenv run python setup.py build bdist_wheel --universal

$ make --debug=b test_all -n | grep 'bdist_wheel'
cd ./env/src && pipenv run python setup.py build bdist_wheel --universal
cd ./env/src && pipenv run python setup.py build bdist_wheel --universal

# CI would rebuild Python, .NET etc 3 times.
```

After:
```
make --debug=b TEST_ALL_DEPS= test_all -n | grep 'bdist_wheel'
```

With the extra magic flag test_all does not depend on build/install, which happened in the previous install_all step.

This shaves off a few minutes from the build time. Will not notice on 1:30 builds but once we get builds to 30 min, 3 min will be 10% which is noticeable.

## Checklist


pgavlin:
> since CI runs install_all before test_all, to avoid triple-building-and-installing.

Why not just have CI run `test_all`?

t0yv0:
test_all still builds everything twice on master.

komalali (approved):
Clever. I'm down to give it a try 👍🏽 🤞🏽

t0yv0:
Credit to https://stackoverflow.com/questions/50633083/makefile-ignore-dependency-when-running-target

Thanks. I'm going to check this in as I'm pretty confident it shouldn't break anything existing. We can always back out.

PR #7640, opened 2021-07-26
Revert flag parse ordering change

t0yv0:
Fixes https://github.com/pulumi/pulumi/issues/7642

# Description

In a recent PR we changed the order in which `pulumi` CLI passes CLI arguments to plugins.

```
$ pulumi preview --logflow --verbose=1 --debug --logtostderr

# Before the change:
pulumi-language-go --logtostderr -v=1 -root=... 127.0.0.1:64391
pulumi-resource-azure-native --logtostderr -v=1 127.0.0.1:64391

# After the change:
pulumi-language-go -root=... 127.0.0.1:64471 --logtostderr -v=1
pulumi-resource-azure-native 127.0.0.1:64471 --logtostderr -v=1
```

Unfortunately as reported by @viveklak this broke `--logflow` which became a no-op for Go-implemented plugins such as `azure-native`. I have traced it through `pkg/resource/provider/main.go` users `flag`. We also use `glog` which relies on `flag` internally (`init()`). At the moment, `flag` considers `pulumi-resource-azure-native 127.0.0.1:63937 -v=9` to have "no flags" and have two arguments, "127.0.0.1:63937" and "-v=9".

History:

- Original issue as https://github.com/pulumi/pulumi/pull/7404
- Merged as https://github.com/pulumi/pulumi/pull/7412/files

See also:

- #6912 
- #7402

Fixes # (issue)

## Checklist


viveklak (approved):
Thanks @t0yv0. I haven't spent much mental bandwidth on making/suggesting the right fix here. At the moment we rely heavily on the logflow args for native providers to dump the raw http payloads to track down issues reported by users so reverting seems valuable.

t0yv0:
I'll check with Evan - I can probably follow up and re-fix the `pulumi up --logflow causes node multi-lang component provider to hang #6912` some other way for node. @EvanBoyle

EvanBoyle:
Looks like we need tests in this area in the future. Feel free to revert, but also reopen the original issue and drop a note in if you don't mind.

Issue #7595, opened 2021-07-21
Azure-Native Go SDK is broken after upgrading to 3.7.0


mikhailshilkov:
Unfortunately, we don't check this during PR build, so it's already merged (but not released yet). The nightly Go SDK Build job is failing with

```
# github.com/pulumi/pulumi-azure-native/sdk/go/azure/datafactory
go/azure/datafactory/pulumiTypes.go:149793:58: undefined: SSISExecutionParameterResponseArgsMapInput
# github.com/pulumi/pulumi-azure-native/sdk/go/azure/datamigration
go/azure/datamigration/pulumiTypes.go:26278:51: undefined: DatabaseTableResponseArgsArrayInput
# github.com/pulumi/pulumi-azure-native/sdk/go/azure/documentdb
go/azure/documentdb/pulumiTypes.go:17348:44: undefined: CompositePathResponseArgsArrayInput
# github.com/pulumi/pulumi-azure-native/sdk/go/azure/insights
go/azure/insights/pulumiTypes.go:22991:62: undefined: WorkbookTemplateLocalizedGalleryArgsArrayInput
# github.com/pulumi/pulumi-azure-native/sdk/go/azure/webpubsub
go/azure/webpubsub/pulumiTypes.go:4653:58: undefined: EventHandlerTemplateResponseArgsArrayInput
```

See https://github.com/pulumi/pulumi-azure-native/runs/3116673664?check_suite_focus=true#step:16:27

PR #7643, opened 2021-07-26
[codegen/go] Unify input method generation.

pgavlin:
The code for input method generation was duplicated. These changes
remove the duplicated code.

This is prep for fixing #7595.

### DIFF
--- a/pkg/codegen/go/gen.go
+++ b/pkg/codegen/go/gen.go
@@ -706,27 +706,6 @@ func (pkg *pkgContext) getInputUsage(name string) string {
 	}, "\n")
 }
 
-// genResourceContainerInput handles generating container (slice/map) wrappers around
-// resources to facilitate external references.
-func genResourceContainerInput(w io.Writer, name, receiverType, elementType string) {
-	fmt.Fprintf(w, "func (%s) ElementType() reflect.Type {\n", receiverType)
-	fmt.Fprintf(w, "\treturn reflect.TypeOf((%s)(nil))\n", elementType)
-	fmt.Fprintf(w, "}\n\n")
-
-	fmt.Fprintf(w, "func (i %s) To%sOutput() %sOutput {\n", receiverType, Title(name), name)
-	fmt.Fprintf(w, "\treturn i.To%sOutputWithContext(context.Background())\n", Title(name))
-	fmt.Fprintf(w, "}\n\n")
-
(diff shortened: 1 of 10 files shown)


lblackstone (approved):

EvanBoyle:
Sorry, not super familiar with this area. Do you have pointers to the test cases that were originally written to cover this?

pgavlin:
> Sorry, not super familiar with this area. Do you have pointers to the test cases that were originally written to cover this?

They're the ones that are failing :)

I haven't yet updated the baselines for the diffs. There is an expected diff in the body of the `ElementType` method, which changes like so:

```diff
--- Expected
+++ Actual
@@ -131,3 +131,3 @@
func (PetArray) ElementType() reflect.Type {
-	return reflect.TypeOf(([]*Pet)(nil))
+	return reflect.TypeOf((*[]*Pet)(nil)).Elem()
}
```

Both implementations produce the same result, but the latter is a pattern that is usable with all types (incl. interface types), which is why the code generator uses it.

EvanBoyle:
> Both implementations produce the same result, but the latter is a pattern that is usable with all types (incl. interface types), which is why the code generator uses it.

Do we have tests that actually exercise any of this newly generated code at runtime? (just to assert correctness of the implementation of `ElementType`). Bringing this up as I was just looking at a bug last week which falls in this category of defect/coverage gap.

pgavlin:
> Do we have tests that actually exercise any of this newly generated code at runtime? (just to assert correctness of the implementation of ElementType). Bringing this up as I was just looking at a bug last week which falls in this category of defect/coverage gap.

No. We don't really have a framework for doing that at the moment.

EvanBoyle:
> No. We don't really have a framework for doing that at the moment.

Not sure we need a framework per se, however we could just import the generated code and run some of it maybe just hand authored files in `pkg/codegen/internal/test/runtime`. Then we can just import the example packages.

pgavlin:
> Not sure we need a framework per se, however we could just import the generated code and run some of it maybe just hand authored files in pkg/codegen/internal/test/runtime. Then we can just import the example packages.

Hm, that might work in the immediate term. I do think that we need some sort of a framework here similar to the integration testing framework (indeed, probably building upon it) to exercise the code in real Pulumi programs.

EvanBoyle (approved):

PR #7647, opened 2021-07-26
[codegen/python] Delete dead code around casing tables

justinvp:
Coincident with the release of Pulumi 3.0, we updated the provider SDK codegen for Python to no longer use casing tables for translating Python snake_case names to Pulumi camelCase names (and vice versa). Instead, the mapping is encoded in decorators applied on class properties.

Some of the code that was used to generate and use the casing tables has persisted. This commits removes this code, as it's no longer necessary, and will improve the quality of our generated examples.

Follow-up from https://github.com/pulumi/pulumi/pull/7547#issuecomment-881659505

### DIFF
--- a/pkg/codegen/docs/gen.go
+++ b/pkg/codegen/docs/gen.go
@@ -48,13 +48,6 @@ var (
 	packagedTemplates  map[string][]byte
 	docHelpers         map[string]codegen.DocLanguageHelper
 
-	// The following property case maps are for rendering property
-	// names of nested properties in Python language with the correct
-	// casing.
-	snakeCaseToCamelCase map[string]string
-	camelCaseToSnakeCase map[string]string
-	seenCasingTypes      codegen.Set
-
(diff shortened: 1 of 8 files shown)


pgavlin:
> Coincident with the release of Pulumi 3.0, we updated the provider SDK codegen for Python to no longer use casing tables for translating Python snake_case names to Pulumi camelCase names (and vice versa). Instead, the mapping is encoded in decorators applied on class properties.
> 
> Some of the code that was used to generate and use the casing tables has persisted. This commits removes this code, as it's no longer necessary, and will improve the quality of our generated examples.

So is it correct that we no longer use these casing tables at all, even in maps?

justinvp:
> So is it correct that we no longer use these casing tables at all, even in dictionaries?

Correct. If you pass a dict, it'll do the translation using the type/name information of the associated `@input_type` class for the property. And if the property is intended to be a dictionary, we do no translation.

pgavlin (approved):

Issue #7407, opened 2021-07-01
Uniform authoring for program codegen tests


emiliza:
Currently: 
We hand author PCL and compare outputs (diff tests)

Things to consider:
- Look at how we’re doing comparison testing - decide on just one
- Scenarios to try: Re-baseline, failures, specific test
- Create set of utility functions and docs
- Start using existing utility functions
- Run all tests against all test schemas

PR #7657, opened 2021-07-27
[codegen] Unify program codegen testing

komalali:
# Description

Inspired by https://github.com/pulumi/pulumi/pull/7433

Fixes: #7407 

## Checklist


pkg/codegen/hcl2/binder_test.go
```diff
@@ -1,51 +0,0 @@
-package hcl2
```
komalali, line 1: This file is deleted because the functionality tested here is already tested via the programgen tests and it was causing an import cycle error.

pkg/codegen/hcl2/binder_schema_test.go
```diff
@@ -1,18 +1,22 @@
-package hcl2
+package hcl2_test
```
komalali, line 1: Some rejigging required by `import cycle not allowed in test` linting error

pkg/codegen/hcl2/binder_schema.go
```diff
@@ -28,7 +28,7 @@ import (
 	"github.com/zclconf/go-cty/cty"
 )
 
-type packageSchema struct {
+type PackageSchema struct {
```
t0yv0, line 31: Now that this is exported, do we need to add `// internal` or some such markers so users don't depend on it with the expectation of full support (no breaking changes, etc)? Or this is implied already?

t0yv0 (approved):
An awesome and much needed refactor.

pkg/codegen/hcl2/binder_schema_test.go
pgavlin: Ah, did this cause a cycle between `codegen/hcl2` and `codegen/internal/test`?

pkg/codegen/hcl2/binder_schema.go
pgavlin: > Now that this is exported, do we need to add // internal or some such markers so users don't depend on it with the expectation of full support (no breaking changes, etc)? Or this is implied already?

The canonical way to make this internal would be to factor it out into a new package, `hcl2/internal/whatever`, and export it from that package. I have a _slight_ preference for that approach here, but we kind of already treat these packages as internal interfaces that we're free to break at will, so I don't feel strongly.

pgavlin: (and when I say canonical, I mean that the `go` tool would refuse to allow packages outside of `hcl2` to link to `hcl2/internal/whatever`)

pkg/codegen/hcl2/binder_schema_test.go
komalali: indeed

pgavlin:
Note: this will fix https://github.com/pulumi/pulumi/issues/7407

pkg/codegen/hcl2/binder_schema.go
komalali: yikes, yeah I dunno if I want to go down that refactor rabbit hole right now. The issue is really that the program driver needs `BindProgram`, so it imports hcl2. https://github.com/pulumi/pulumi/pull/7657/files#diff-723d6e036de4c1d99d5d402b6dc8988d390ba50413f3d465a7b97056378e42bfR164

And hcl2 needs test.NewHost, so it imports test.

I added a commit that refactors `NewHost` into a new `utils` package to avoid the cycle. Is that preferable to what I had?

pkg/codegen/hcl2/binder_test.go
pgavlin: Can we restore these tests now that the cycle has been broken?

pgavlin (approved):
One question re: restoring a deleted file, but LGTM otherwise!

justinvp (approved):

PR #7659, opened 2021-07-27
Continue build fanout: split integration tests by tag 

t0yv0:
# Description

Since integration tests were on the critical path, this further splits them by language tag to fan them out across workers. Some care is needed to utilize the workers well and not request too many. For starters - one worker per language.

Fixes # (issue)

## Checklist


t0yv0:
OK results here are really bothersome.

Ubuntu runner: fantastic, critical path is 30 minutes 🎉 

However, Mac runners are taking forever since we seem not to have enough throughput. The mac jobs are waiting on available runners.

t0yv0:
Closing, this is turning problematic and not worth additional complexity.

PR #7661, opened 2021-07-27
Fixing WaitGroup related panics in the Go SDK

### LINKED ISSUE #7275: Issue with the work for waiting async tasks in Go SDK
Experienced by @stack72 - seeing this on the pulumi-docker repo after upgrading to Pulumi v3.4.0 when using a pulumi-docker go program
```
Diagnostics:
  pulumi:pulumi:Stack (dockerfile-go-p-it-fv-az290-4-dockerfile-d29d3b03):
    panic: sync: WaitGroup is reused before previous Wait has returned
    goroutine 1 [running]:
    sync.(*WaitGroup).Wait(0xc00039c9d8)
    	/opt/hostedtoolcache/go/1.16.4/x64/src/sync/waitgroup.go:132 +0xae
    github.com/pulumi/pulumi/sdk/v3/go/pulumi.(*Context).wait(0xc00039c8c0, 0x0, 0x0)
    	/home/runner/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.4.0/go/pulumi/context.go:156 +0x50
    github.com/pulumi/pulumi/sdk/v3/go/pulumi.RunWithContext(0xc00039c8c0, 0xe32308, 0xc00003824f, 0xd)
    	/home/runner/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.4.0/go/pulumi/run.go:112 +0x2a6
    github.com/pulumi/pulumi/sdk/v3/go/pulumi.RunErr(0xe32308, 0x0, 0x0, 0x0, 0x0, 0x0)
    	/home/runner/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.4.0/go/pulumi/run.go:84 +0x23f
    github.com/pulumi/pulumi/sdk/v3/go/pulumi.Run(0xe32308, 0x0, 0x0, 0x0)
    	/home/runner/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.4.0/go/pulumi/run.go:41 +0x59
    main.main()
    	/home/runner/go/src/stackName-1623065997245879475/main.go:9 +0x3e
    exit status 2
```

Possibly related to:  https://github.com/pulumi/pulumi/pull/6983

t0yv0:
# Description

Fixes #7275 

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -4,3 +4,7 @@
 
 - [cli] - Respect provider aliases
   [#7166](https://github.com/pulumi/pulumi/pull/7166)
+
+- [sdk/go] - Fix panics caused by logging from `ApplyT`, affecting
+  `pulumi-docker` and potentially other providers
+  [#7661](https://github.com/pulumi/pulumi/pull/7661)
--- a/sdk/go/pulumi/context.go
+++ b/sdk/go/pulumi/context.go
@@ -58,7 +58,7 @@ type Context struct {
 	rpcsLock sync.Mutex // a lock protecting the RPC count and event.
 	rpcError error      // the first error (if any) encountered during an RPC.
 
-	join sync.WaitGroup // the waitgroup for non-RPC async work associated with this context
+	join workGroup // the waitgroup for non-RPC async work associated with this context
 
 	Log Log // the logging interface for the Pulumi log stream.
 }
--- a/sdk/go/pulumi/context_test.go
+++ b/sdk/go/pulumi/context_test.go
@@ -1,0 +1,106 @@
+// Copyright 2016-2021, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package pulumi
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// The test is extracted from a panic using pulumi-docker and minified
+// while still reproducing the panic. The issue is that the resource
+// constructor `NewImage` processes `StringInput` and logs into a
+// captured `*pulumi.Context` from `ApplyT`. The user program passes a
+// vanilla `String`. The `ApplyT` is not tracked against the context
+// join group, but the logging is, which causes the logging statement
+// to appear as "dynamic" work that appeared unexpectedly after "all
+// work was done", and race with the program completion `Wait`.
+func TestLoggingFromApplyCausesNoPanics(t *testing.T) {
+	// Usually panics on iteration 100-200
+	for i := 0; i < 1000; i++ {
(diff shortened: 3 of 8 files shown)


t0yv0:
Discussion on possible approaches to fix this:

- The fact that `String(x).Apply(foo)` is unaware of pulumi.Context and does not participate in the join group is a very big footgun but do we need breaking changes to fix it? Say allow only `ctx.String()` with ctx being pulumi.Context to construct these? Or somehow infer "current active pulumi.Context" automagically? https://github.com/jtolio/gls very controversial tool comes to mind again.

- For the particular case of `Docker` resource, we had a chance in `RegisterOutputs` to make triple-sure the work of computing the resource outputs is registered with the context join group. Perhaps we could admit a fix along those lines, and that would make sure Wait() is waiting on ApplyT that calculates resource outputs in this case.

- We could add some locking to the WaitGroup and try to handle the situation of `Wait()` and `Add(1)` racing in WaitGroup state 0 better  handled specifically for the case of logging. That is, we'd accept that there is a race, by design, and it's indeterminate if the `ctx.Log` work should be in or out of the set of "all work" that has too be awaited, but we could resolve the situation gracefully by for example ignoring the race, and ignoring whether logs propagated or not.

t0yv0:
Copying from the ticket the minimal Go program that helped me understand when/why Go runtime throws a panic:

```
package main

import (
	"fmt"
	"sync"
	"time"
)

// Note that calls with a positive delta that occur when the counter is zero must happen before a Wait.

func main() {
	wg := &sync.WaitGroup{}
	wg.Add(1)

	go func() {
		fmt.Printf("Adding\n")
		wg.Add(-1)
		wg.Add(1)
		fmt.Printf("Added\n")
	}()

	wg.Wait()
	time.Sleep(10)
}
```

Note that unfortunately the panic is attributed to `Wait()` call not `Add` call.

pgavlin:
Oof. This is pretty bad. Aside from the panic, this provides an easy route for asynchronous work to go untracked.

One thing we could consider to ameliorate the problem of untracked asynchronous work is the addition of a fast path in `ApplyT` for the case that the receiver is already resolved. This fast path would promptly run the callback and create and return an already-resolved output if the result of the callback does not contain any `Output`s.

> We could add some locking to the WaitGroup and try to handle the situation of Wait() and Add(1) racing in WaitGroup state 0 better handled specifically for the case of logging. That is, we'd accept that there is a race, by design, and it's indeterminate if the ctx.Log work should be in or out of the set of "all work" that has too be awaited, but we could resolve the situation gracefully by for example ignoring the race, and ignoring whether logs propagated or not.

This seems like the best way to resolve the panic. I think that if we do not want to refactor, we will need to atomically swap in a new `WaitGroup` that is never `Wait`ed immediately prior to calling `Wait` on the original `WaitGroup`.

t0yv0:
Ok let me try something along these lines.

t0yv0:
@pgavlin PTAL.. 

Catching/recovering from panics on `sync.WaitGroup` proved to be not workable. It throws panics in Add in addition to Wait. I was also not clear if it leaves the WaitGroup in a good state after a panic. My attempts to work that out ended in live-lock or deadlock on the tests.

Because of this, I rolled our own version of `sync.WaitGroup` that does not have the race detection enabled. In the situation where previously we panicked, the thread doing Add will wait on the lock held by the thread doing Wait. Then the thread doing Wait will complete successfully. Basically Pulumi program may finish with the extra racing work being ignored.

Unfortunately this does not discriminate between the use for `ctx.Log` and use for `Output` - it seems tricky to do so, perhaps involving rebuilding the race detection in our clone of `sync.WaitGroup`.

I'd tackle fast-path for ApplyT in a separate PR.

pgavlin (approved):
Okay, this seems reasonable to me. Thanks for the tests here!

PR #7662, opened 2021-07-27
[codegen/go] Unify some output type generation

pgavlin:
These changes take a step towards simplifying and unifying the
generation of output types in the Go SDKs, especially for pointer,
array, and map outputs. This code was previously duplicated amongst the
various specialized output type generators, which led to inconsistencies
between the various implementaitons.

This is prep work for fixing #7595.

### DIFF
--- a/pkg/codegen/go/gen.go
+++ b/pkg/codegen/go/gen.go
@@ -738,6 +738,72 @@ func genInputImplementation(w io.Writer, name, receiverType, elementType string,
 	}
 }
 
+func genOutputType(w io.Writer, baseName, elementType string, ptrMethods, resourceType bool) {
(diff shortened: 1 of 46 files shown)


pgavlin:
I'm going to work on adding some behavioral tests here as well, as there are some changes to the implementations of the various `To*Output` methods.

t0yv0:
I don't feel I understand the inputs/outputs in Go SDK well enough yet to review this, still need to work through what fail in https://github.com/pulumi/pulumi/pull/7394 - I'm still quite confused there.

viveklak (approved):
To me this looks like a pretty good net improvement. I inspected the test outputs and I didn't see anything concerning (other than fixing some glaring issues with the elementtype for nested container types).

komalali (approved):

Issue #7663, opened 2021-07-27
Eliminate threading from the Python SDK


t0yv0:
Please consider eliminating threading use from the Python SDK in favor of heavily leaning on asynchronous IO.

Fix outline:

1. switch to AIO-based gRPC clients for the engine and monitor, making those async - see https://grpclib.readthedocs.io/en/latest/

```
python -m grpc_tools.protoc -I./ --python_out="$TEMP_DIR" --grpclib_python_out="$TEMP_DIR" *.proto
```

2. switch mocks to async IO

3. search for all instances of code similar to the following `await asyncio.get_event_loop().run_in_executor(None, do_invoke)` and replace with forking AIO coroutines; currently `run_in_executor` uses default ThreadPoolExecutor, we essentially jump threads on every Invoke, RegisterResource, and so on, to make sure gRPC calls are non-blocking

4. Search for ThreadPoolExecutor and replace the callsites, which appear to be gRPC *server* instances, with AIO-based servers

We could break this up as appropriate.

If this becomes a breaking change, we could build out a parallel set of classes such as AIOMonitorClient and gradually transition code, with env vars to back out.

## Affected feature

As an immediate benefit this will remove issues related to several AIO loops on different threads https://github.com/pulumi/pulumi/issues/6561

Indirectly, this will also contribute to scalability and stability of the SDK. Python AIO loops should be capable of cheaper context switches than threading. Note that Python programs using threading still enter the Global Interpreter Lock when entering Python code and thus only speed things up for the situation of waiting on IO in native code.

t0yv0:
CC @emiliza logging this for later as discussed.

t0yv0:
AS @komalali rightly points out we also need to tackle e #6762  before or concurrently with this, since putting more things onto the shared AIO loop on the main thread will introduce accidental waits unless #6762 can distinguish pulumi tasks from non-pulumi tasks (such as internal tasks of the gRPC servers and clients).

ahmed-wajid:
This keeps raising sometimes with flaky Pytests. But meanwhile, it raises pretty often when we run Automation API, with celery for async tasks. 

```
Resources:
    + 6 created
    17 unchanged

Duration: 1m57s
[...]

Issue #6762, opened 2021-04-13
[sdk/python] - Improve await of outstanding tasks


komalali:
Currently, we [collect all remaining async tasks](https://github.com/pulumi/pulumi/blob/0c4e1a33e058a5b129ba009c5a58793200c09fff/sdk/python/lib/pulumi/runtime/stack.py#L65-L87) in the event loop and await them prior to exit. We introduced this via https://github.com/pulumi/pulumi/pull/6606. This is not great, because there may be async tasks from the user that have nothing to do with Pulumi and we'll end up waiting maybe forever.

A better solution would collect the outstanding tasks that we would like to await (mostly `Output`/`apply`-related tasks) and *only* await those. 

## Expected behavior

The expected behavior is for `pulumi preview` to succeed quickly in presence of unrelated long-running tasks.

## Current behavior

Current behavior is for `pulumi preview`, and indeed driver program execution, hangs until unrelated long-running tasks complete.

## Steps to reproduce

Simple program to repro:

```
"""An AWS Python Pulumi program"""

import asyncio
import pulumi
from pulumi_aws import s3

async def t():
    print('start')
    await asyncio.sleep(10000)
    print('end')

loop = asyncio.get_running_loop()
loop.create_task(t())

# Create an AWS resource (S3 Bucket)
bucket = s3.Bucket('my-bucket')

# Export the name of the bucket
pulumi.export('bucket_name', bucket.id)
```

The more interesting reproduction case involves Automation API. Starting from inline_program example:

https://github.com/pulumi/automation-api-examples/tree/main/python/inline_program

If we inject the `create_task` into main.py, `def pulumi_program`, then we can reproduce the issue. Interestingly if we inject create_task at top-level, the issue does not happen. This seems to be affecting only the scope of inline program, which may have its own event loop (and/or process or thread).

## Context (Environment)

This has affected @t0yv0  when working on multi-lang features, in the context of a provider running gRPC AIO server. The Python primitives we use to await RPC calls inadvertently grab tasks pertaining to the gRPC server loop, which never complete by design. Specifically this primitive is responsible:

```
# pulumi/runtime/stack.py
async def run_pulumi_func(func: Callable):
  ...
```

This can be traced down to using ```_all_tasks = asyncio.Task.all_tasks``` and awaiting them; that is, the code references the thread-level event loop and inspects all running tasks.

It seems to be invoked in at least LanguageServer.

For mutli-lang component purposes we might have a workaround.

## Affected feature

[...]

PR #7666, opened 2021-07-28
Remove threading from mock tests to fix AIO waits on diff event loop

### LINKED ISSUE #6561: Flakey python test errors with "Task attached to a different loop"

t0yv0:
# Description

Fixes #6561

I could not reproduce 6561 locally except by replacing `sync_await._ensure_event_loop()` with `get_event_loop`. These refactored tests pass under plain `get_event_loop`. This is because all action now happens on the main thread, unlike on ThreadPoolExecutor as in the product. 

So it is unfortunate that the refactored tests are working in a context that's different from product context.

On the upside, I believe they will stop flaking up the PR verification. This is a big relief.

TO reduce blast radius, I refactored the test suite to use Pytest fixtures so that the modifications to the default executor are "local" to the test suite and do not bleed into other tests.

## Checklist

### DIFF
--- a/sdk/python/lib/test_with_mocks/resources.py
+++ b/sdk/python/lib/test_with_mocks/resources.py
@@ -1,6 +1,21 @@
+# Copyright 2016-2021, Pulumi Corporation.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 from typing import Optional
 
 import pulumi
+
 
 class MyComponent(pulumi.ComponentResource):
     outprop: pulumi.Output[str]
@@ -36,9 +51,6 @@ class Module(pulumi.runtime.ResourceModule):
             raise Exception(f"unknown resource type {typ}")
 
 
-pulumi.runtime.register_resource_module("aws", "ec2/instance", Module())
-
-
 class MyCustom(pulumi.CustomResource):
     instance: pulumi.Output
     def __init__(self, resource_name, props: Optional[dict] = None, opts = None):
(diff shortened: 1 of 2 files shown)


t0yv0:
Ironic, but it fails on *other* flakes.

komalali:
I think the windows failures are not flakes but some sort of weird windows fail that happens sometimes. Maybe an OOM? It's impossible to see the output of those runs. The [raw logs](https://pipelines.actions.githubusercontent.com/I4JhI7AD9DIK0DK9Dg4gHSXXNoWQh7TNpUrWvWXAHzW5wLBXtg/_apis/pipelines/1/runs/17731/signedlogcontent/20?urlExpires=2021-07-28T20%3A57%3A36.2870347Z&urlSigningMethod=HMACV1&urlSignature=oeLXqWhFkmESbMPLXkHkRcrby7AeSAad%2BSK8KZswLY0%3D) don't show anything

komalali:
I have to admit it makes me *really* nervous to fix this at the test level. I would probably prefer to skip all these tests to prevent PR flakey-ness rather than do this and consider #6561 "fixed". Curious to hear what @justinvp thinks.

t0yv0:
OOM quite possibly could kill logs...

Note that the test still has coverage for intended Pulumi semantics, just under one thread vs multiple thread... Seems like it has some coverage value.. I acknowledge the concern though.

The real concern is - do we have consensus on removing threading (https://github.com/pulumi/pulumi/issues/7663) and how fast can we get that in, if we do want it.

I think if we could commit to removing threading from product soon-ish, we could check this in, consider "flaky/test" problem closed, but reopen the bug for #6561 user portions, or just not close it, and move forward here.

komalali:
I'm definitely on board to remove threading. Or more specifically, I'm on board to move to the asyncio grpc interface. I foresee some complications, specifically we will need to rework the "await all remaining async work" since the server itself will be a neverending async process which has led to hangs before. 

But regardless, I think this is work we should do because of the exact error this PR is attempting to remedy.

t0yv0:
Oh that's an excellent comment we have the https://github.com/pulumi/pulumi/issues/6762 problem. I'll link that into the ticket as well.. This needs to be taken care of also!

justinvp (approved):
Like @komalali, I'm a bit nervous about the test changes because the main point of these tests is to have coverage of how [users would write unit tests using mocks](https://www.pulumi.com/docs/guides/testing/unit/) ([example](https://github.com/pulumi/examples/tree/master/testing-unit-py)). With these changes, the tests are no longer representative of what customers would write when writing their own tests using mocks. (Or do you think we need to rewrite our docs and example to use the `ImmediateExecutor`?)

However, CI flakiness is not good, so I'm OK with taking this change temporarily until we can address the problem for real. Part of that would be changing this back to the way it was. Which issue is tracking this? #7663?

The alternative is to disable these tests and have an issue open that tracks re-enabling the tests when the underlying issue is addressed, but then we'd have no coverage at all of this area in the meantime.

t0yv0:
I'll just keep #6561 open.

PR #7671, opened 2021-07-29
Experiment with reducing test parallelism to free up runners

t0yv0:
# Description

Gets us to 2 test subsets and parallel runners instead of 4. Hypothesis here is that because of repeat work in the parallel runners, this will help the total CI throughput and reduce "waiting for Mac OS runner" delays.

Fixes # (issue)

## Checklist


t0yv0:
Looks like we're doing kind of okay in current configuration, closing.

PR #7677, opened 2021-07-29
[sdk/python] Handle unknown results from methods

### LINKED ISSUE #7670: [sdk/python] Methods are not correctly bubbling up unknown values
The Python SDK is not correctly handling unknown values returned from methods. This results in errors being raised in the generated `Result` class's constructor when it checks the type of an individual value.

For example:

```python
@pulumi.output_type
class GetMessageResult:
    def __init__(__self__, message=None):
        if message and not isinstance(message, str):
            raise TypeError("Expected argument 'message' to be a str") # <-- this will be raised when message is unknown
        pulumi.set(__self__, "message", message)

    @property
    @pulumi.getter
    def message(self) -> str:
        return pulumi.get(self, "message")
```

To address this, the SDK needs to check for any unknown values after deserializing the RPC values, mark the returned output as unknown, and not attempt to instantiate the `Result` class.

justinvp:
This fixes handling of unknown results from methods in the Python SDK. Node.js and Go were already handling this appropriately.

This adds tests in each of the languages. The test program passes an unknown value to the provider, and asserts that an `apply` is not run for it, and then returns the value, and the program asserts that an `apply` does not run on the result. The Python test program fails before the fix and succeeds after.

Fixes #7670


sdk/python/lib/pulumi/runtime/invoke.py
```diff
@@ -237,11 +239,12 @@ def do_rpc_call():
...
 
-                # If typ is not None, call translate_output_properties to instantiate any output types.
-                value = rpc.translate_output_properties(deserialized, lambda prop: prop, typ) if typ else deserialized
+                if is_known:
+                    # If typ is not None, call translate_output_properties to instantiate any output types.
+                    value = rpc.translate_output_properties(deserialized, lambda p: p, typ) if typ else deserialized
 
             resolve_value.set_result(value)
-            resolve_is_known.set_result(True)
+            resolve_is_known.set_result(is_known)
```
t0yv0, line 247: I feel we're coding against a very low-level interface here so it sounds like it was an easy mistake to make.

tests/integration/construct_component_methods_unknown/testcomponent-python/testcomponent.py
```diff
@@ -0,0 +1,80 @@
...
+class Component(pulumi.ComponentResource):
+    def __init__(self,
+                 resource_name: str,
+                 opts: Optional[pulumi.ResourceOptions] = None) -> None:
+        super().__init__("testcomponent:index:Component", resource_name, {}, opts)
+
+    def get_message(self, echo: pulumi.Input[str]) -> pulumi.Output[str]:
+        message = pulumi.Output.from_input(echo)
+        message.apply(lambda v: panic("should not run (echo)"))
+        return message
```
t0yv0, line 36: Could make it a bit stronger by `return  message.apply(lambda v: panic("should not run (echo)"))`? So we don't rely on "we wait for all outstanding applies". I should perhaps not be skeptical that it works, but I've just been chasing bugs in that in Go SDK.

t0yv0:
Curious, the integration tests are fine here but how do you feel about using Mock-based tests instead?  I think this could run against a MockMonitor and exercise the changes in rpc.py but maybe I'm mistaken. They're a bit more tedious but run faster. Or would that lose essential coverage?

t0yv0 (approved):

pgavlin (approved):

tests/integration/construct_component_methods_unknown/testcomponent-python/testcomponent.py
justinvp: Done

justinvp:
> how do you feel about using Mock-based tests

Good point. We could probably have Mock-based tests for this, but I think I'd still additionally want integration tests to test E2E. Also, at some point soon I hope to be able to add SDK generation for many of these integration tests (generate SDKs for the provider resources based on a schema, to fully exercise the codegen + runtime), so in some ways this is setting the stage for that.

PR #7681, opened 2021-07-29
Fix reflecting list type annotations

### LINKED ISSUE #7313: Incorrect type annotation in Python dynamic provider leads to hang
Deploying https://github.com/pulumi/examples/blob/master/aws-py-ec2-provisioners/ leads to a hang.

Verbose logging shows that the following errors are reported before the hang:

```
  File "/Users/lukehoban/go/src/github.com/pulumi/examples/aws-py-ec2-provisioners/venv/lib/python3.9/site-packages/pulumi/runtime/rpc_manager.py", line 67, in rpc_wrapper
    result = await rpc
  File "/Users/lukehoban/go/src/github.com/pulumi/examples/aws-py-ec2-provisioners/venv/lib/python3.9/site-packages/pulumi/runtime/resource.py", line 542, in do_register
    rpc.resolve_outputs(res, resolver.serialized_props, resp.object, deps, resolvers, transform_using_type_metadata)
  File "/Users/lukehoban/go/src/github.com/pulumi/examples/aws-py-ec2-provisioners/venv/lib/python3.9/site-packages/pulumi/runtime/rpc.py", line 697, in resolve_outputs
    translated_value = translate_output_properties(value, translate_to_pass, types.get(key),
  File "/Users/lukehoban/go/src/github.com/pulumi/examples/aws-py-ec2-provisioners/venv/lib/python3.9/site-packages/pulumi/runtime/rpc.py", line 643, in translate_output_properties
    element_type = _get_list_element_type(typ)
  File "/Users/lukehoban/go/src/github.com/pulumi/examples/aws-py-ec2-provisioners/venv/lib/python3.9/site-packages/pulumi/runtime/rpc.py", line 79, in _get_list_element_type
    raise AssertionError(f"Unexpected type. Expected 'list' got '{typ}'")
AssertionError: Unexpected type. Expected 'list' got '<class 'list'>'
```

These errors are triggered by the following line - in particular the fact that it is `pulumi.Output[list]` instead of `pulumi.Output[List[Dict[str, str]]]`.  Changing to the latter fixes this, but that should not be necessary - "incorrect" type annotations should not lead to a failure and certainly not to a hang.

https://github.com/pulumi/examples/blob/master/aws-py-ec2-provisioners/provisioners.py#L160:L160

This also appears to be a regression, as this same program worked when originally added.

Finally - this program does not appear to be being run in the `examples` test suite.  The morally equivalent TypeScript example is being run, so we should have all the required machinery there to put this under test.

t0yv0:
# Description

This makes the example from https://github.com/pulumi/pulumi/issues/7313 pass (examples/aws-py-ec2-provisioners).

Fixes #7313 

It did not hang this time but it did fail with a relatively obscure stack trace. It works with the fix.

## Checklist

### DIFF
--- a/sdk/python/lib/pulumi/runtime/rpc.py
+++ b/sdk/python/lib/pulumi/runtime/rpc.py
@@ -68,6 +68,11 @@ def _get_list_element_type(typ: Optional[type]) -> Optional[type]:
     if typ is None:
         return None
 
+    # Annotations not specifying the element type are assumed by mypy
+    # to signify Any element type. Follow suite here.
+    if typ in {list, List, Sequence, abc.Sequence}:
+        return cast(Any, type)
+
     # If typ is a list, get the type for its values, to pass
     # along for each item.
     origin = _types.get_origin(typ)
--- a/sdk/python/lib/pulumi/runtime/rpc.py
+++ b/sdk/python/lib/pulumi/runtime/rpc.py
@@ -69,7 +69,7 @@ def _get_list_element_type(typ: Optional[type]) -> Optional[type]:
         return None
 
     # Annotations not specifying the element type are assumed by mypy
-    # to signify Any element type. Follow suite here.
+    # to signify Any element type. Follow suit here.
     if typ in {list, List, Sequence, abc.Sequence}:
         return cast(type, Any)
 
--- a/sdk/python/lib/pulumi/runtime/rpc.py
+++ b/sdk/python/lib/pulumi/runtime/rpc.py
@@ -71,7 +71,7 @@ def _get_list_element_type(typ: Optional[type]) -> Optional[type]:
     # Annotations not specifying the element type are assumed by mypy
     # to signify Any element type. Follow suite here.
     if typ in {list, List, Sequence, abc.Sequence}:
-        return cast(Any, type)
+        return cast(type, Any)
 
     # If typ is a list, get the type for its values, to pass
     # along for each item.
--- a/sdk/python/lib/test/runtime/test_rpc.py
+++ b/sdk/python/lib/test/runtime/test_rpc.py
@@ -1,0 +1,36 @@
+# Copyright 2016-2021, Pulumi Corporation.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from collections import abc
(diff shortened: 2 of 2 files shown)


sdk/python/lib/test/runtime/test_rpc.py
```diff
@@ -0,0 +1,36 @@
...
+    assert rpc._get_list_element_type(typing.Sequence) == Any
+
+    assert rpc._get_list_element_type(typing.List[Any]) == Any
+    assert rpc._get_list_element_type(typing.Sequence[Any]) == Any
+
+    assert rpc._get_list_element_type(typing.List[int]) == int
+    assert rpc._get_list_element_type(typing.Sequence[int]) == int
+
+    assert rpc._get_list_element_type(typing.List[typing.List[int]]) == typing.List[int]
+    assert rpc._get_list_element_type(typing.Sequence[typing.Sequence[int]]) == typing.Sequence[int]
```
komalali, line 36: What happens if it's `typing.List[typing.List]]`?

komalali (approved):

sdk/python/lib/test/runtime/test_rpc.py
t0yv0: Good question. I'll add. It does no recursive descend so I think it just returns X from `List[X]` without unpacking. The outer function does descend, do next time we'll be called with X=List and infer Any.

sdk/python/lib/pulumi/runtime/rpc.py
```diff
@@ -68,6 +68,11 @@ def _get_list_element_type(typ: Optional[type]) -> Optional[type]:
     if typ is None:
         return None
 
+    # Annotations not specifying the element type are assumed by mypy
+    # to signify Any element type. Follow suit here.
+    if typ in {list, List, Sequence, abc.Sequence}:
+        return cast(type, Any)
```
justinvp, line 74: Based on the usage, and how we do similar checks for dicts, we could return `None` here. But returning `Any` also works (since all the places that use this will convert `Any` to `None` and aligning with mypy makes sense to me.

justinvp, line 74: I'm good with the way you have it so feel free to ignore...

Just wondering if we could put this after the origin check so we don't have to have the `typ in {...}` twice? Granted I have not tried it out.

```python
def _get_list_element_type(typ: Optional[type]) -> Optional[type]:
    if typ is None:
        return None

    # If typ is a list, get the type for its values, to pass
    # along for each item.
    origin = _types.get_origin(typ)
    if typ is list or origin in {list, List, Sequence, abc.Sequence}:
        args = _types.get_args(typ)
        # Return the element type, if available, otherwise Any to be consistent with
        # how mypy treats annotations that do not specify an element type.
        return args[0] if args else cast(type, Any)

    raise AssertionError(f"Unexpected type. Expected 'list' got '{typ}'")
```

justinvp (approved):

sdk/python/lib/pulumi/runtime/rpc.py
t0yv0: This doesn't quite work since `_types.get_origin(Sequence) is None`

PR #7684, opened 2021-07-29
Use `provider.MainWithOptions` to reduce boilerplate in integration tests

justinvp:
We can use the new `provider.MainWithOptions` from #7598 to reduce boilerplate in some of our testcomponent providers.

Also, while cleaning up here, I took this as an opportunity to replace use of `github.com/pkg/errors` in the `tests` dir to use the built-in functionality of the Go standard library.

### DIFF
--- a/tests/examples/examples_test.go
+++ b/tests/examples/examples_test.go
@@ -4,6 +4,7 @@ package examples
 
 import (
 	"bytes"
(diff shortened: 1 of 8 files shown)


t0yv0 (approved):
Neat! 🚢

PR #7692, opened 2021-07-31
Fix Go generation for top-level module resources in a provider with a dash in name

mikhailshilkov:
This will unblock https://github.com/pulumi/pulumi-aws-native/issues/6

Unlike `azure-native` and `google-native`, `aws-native` has types/resources in the root (index) module. Our go module doesn't handle this correctly and generates code like

```go
type Analyzer struct {
	pulumi.CustomResourceState

    // ...
	Tags aws-native.TagArrayOutput `pulumi:"tags"`
    // ...
}
```

This fix will replace `aws-native` to `aws` in the snippet above. Code generation succeeds and the code compiles.


justinvp (approved):
LGTM

pgavlin (approved):

pgavlin:
Surprised that there isn't a test failure in here--I thought that we had tests that hit this case. Would you mind adding one?

mikhailshilkov:
@pgavlin I added tests, PTAL

justinvp (approved):

PR #7695, opened 2021-08-02
Go install command is given proper DOTNET_VERSION

jetvova:
# Description

Fixed small typo in build script. Previously, the -ldflags would be given an incorrect empty version.


stack72 (approved):
Nice spot here @jetvova!

PR #7698, opened 2021-08-02
Update of CONTRIBUTING.md to reflect 3.0 changelog

T-Vova:
CONTRIBUTING.md updated to require NodeJS v11.x rather than v10.x, as Pulumi 3.0 no longer supports the latter.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,7 +9,7 @@ For larger features, we'd appreciate it if you open a [new issue](https://github
 To hack on Pulumi, you'll need to get a development environment set up. You'll want to install the following on your machine:
 
 - Go 1.16
-- NodeJS 11.X.X or later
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -9,7 +9,7 @@ For larger features, we'd appreciate it if you open a [new issue](https://github
 To hack on Pulumi, you'll need to get a development environment set up. You'll want to install the following on your machine:
 
 - Go 1.16
-- NodeJS 10.X.X or later
+- NodeJS 11.X.X or later
```
justinvp, line 12: It's really 12 or later. 12 is the current Maintenance LTS version. https://nodejs.org/en/about/releases/

justinvp (approved):
LGTM otherwise

PR #7699, opened 2021-08-02
Fixed failed test: test_runtime_errors()

T-Vova:
# Description

Added an additional regex check to test_runtime_errors() in test_errors.py to support dotnet's unique unhandled error message.


t0yv0:
Hm, no change in tests? This passed before and after? The interesting combination is not being exercised?

PR #7700, opened 2021-08-03
python: Use Sequence rather than List in Resource fields

### LINKED ISSUE #7693: Function parameters in the Python SDK should use `Iterable` or `Sequence` whenever possible
Many of the function parameters in the Python API specify parameters of type `list[T]`, e.g.:

  * https://github.com/pulumi/pulumi/blob/29fa23d6d9e78009cc63b64387303f32be472dee/sdk/python/lib/pulumi/resource.py#L393
  * https://github.com/pulumi/pulumi/blob/29fa23d6d9e78009cc63b64387303f32be472dee/sdk/python/lib/pulumi/resource.py#L396
  * https://github.com/pulumi/pulumi/blob/29fa23d6d9e78009cc63b64387303f32be472dee/sdk/python/lib/pulumi/resource.py#L407
  * https://github.com/pulumi/pulumi/blob/29fa23d6d9e78009cc63b64387303f32be472dee/sdk/python/lib/pulumi/resource.py#L400

I think in all of the cases above, all that's really required is `Iterable[T]`. Using `list[T]` can be painful, because `list[T]` is invariant, and so a parameter that accepts `list[Resource]` will reject e.g. an argument of type `list[CustomResource]`.

This is the same issue as https://github.com/pulumi/pulumi/issues/5278, but for the code in the core SDK, rather than the codegen'd providers.

benesch:
# Description

Similar to #5282, but for core SDK types. The tl;dr is that because
Sequence[T] is covariant, constructing resources becomes much more
ergonomic.

Fix #7693.

## Checklist

### DIFF
--- a/sdk/python/lib/pulumi/runtime/resource.py
+++ b/sdk/python/lib/pulumi/runtime/resource.py
@@ -15,7 +15,7 @@ import asyncio
 import os
 import traceback
 
-from typing import Optional, Any, Callable, List, NamedTuple, Dict, Set, Tuple, Union, TYPE_CHECKING, cast, Mapping
+from typing import Optional, Any, Callable, List, NamedTuple, Dict, Set, Tuple, Union, TYPE_CHECKING, cast, Mapping, Sequence
 from google.protobuf import struct_pb2
 import grpc
 
@@ -644,14 +644,14 @@ class RegisterResponse:
 def convert_providers(
         provider: Optional['ProviderResource'],
         providers: Optional[Union[Mapping[str, 'ProviderResource'],
-                                  List['ProviderResource']]]) -> Mapping[str, 'ProviderResource']:
+                                  Sequence['ProviderResource']]]) -> Mapping[str, 'ProviderResource']:
     if provider is not None:
         return convert_providers(None, [provider])
 
     if providers is None:
         return {}
 
-    if not isinstance(providers, list):
+    if isinstance(providers, Mapping):
         return providers
 
     result = {}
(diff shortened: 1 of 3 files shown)


sdk/python/lib/pulumi/runtime/resource.py
```diff
@@ -644,14 +644,14 @@ def __init__(self,
 def convert_providers(
         provider: Optional['ProviderResource'],
         providers: Optional[Union[Mapping[str, 'ProviderResource'],
-                                  List['ProviderResource']]]) -> Mapping[str, 'ProviderResource']:
+                                  Iterable['ProviderResource']]]) -> Mapping[str, 'ProviderResource']:
```
komalali, line 647: Is there a reason you used `Iterable` here instead of `Sequence`? Isn't `Mapping` itself a subclass of `Iterable`? Does that mean this is just `Optional[Iterable['ProviderResource']]`?

benesch: Oops, no, just an oversight. I can change this to `Sequence` if you'd prefer, but all `Sequence[T]`s implement `Iterable[T]` so it's not too much different.

`Mapping` *is* a subclass of `Iterable`, but `Mapping[str, ProviderResource]` impls `Iterable[[str, ProviderResource]]`. The idea here is that this function takes a list-like *or* a dict-like.

komalali: Right right, makes sense.

Yeah, I think let's just change this to `Sequence` to keep things consistent. Otherwise, the change looks good to me.

benesch: Sure thing, done!

benesch:
Thanks very much!

PR #7704, opened 2021-08-04
[sdk/nodejs] Account for outstanding async work done in `prepareResource`

EvanBoyle:
Fixes https://github.com/pulumi/pulumi/issues/6998

Turns out we weren't adding `prepareResource` work the the keep-alive queue. We were waiting until `runAsyncResourceOp` which happens in a `.then` chained off of `prepareResource`. The automation api shutdown routine, `runtime.disconnect`, was unaware of this outstanding work as a result.

### DIFF
--- a/sdk/nodejs/runtime/resource.ts
+++ b/sdk/nodejs/runtime/resource.ts
@@ -420,10 +420,10 @@ export function registerResource(res: Resource, t: string, name: string, custom:
 async function prepareResource(label: string, res: Resource, custom: boolean, remote: boolean,
                                props: Inputs, opts: ResourceOptions): Promise<ResourceResolverOperation> {
 
-    // add an entry to the rpc queue while we prepare the request. 
+    // add an entry to the rpc queue while we prepare the request.
     // automation api inline programs that don't have stack exports can exit quickly. If we don't do this,
     // sometimes they will exit right after `prepareResource` is called as a part of register resource, but before the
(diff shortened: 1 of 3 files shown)


komalali (approved):

sdk/nodejs/runtime/resource.ts
```diff
@@ -563,6 +569,9 @@ async function prepareResource(label: string, res: Resource, custom: boolean, re
         }
     }
 
+    // free the RPC queue
+    done();
```
lukehoban, line 573: Don't we need to put this in a `finally` to ensure it runs even if there are errors?  It looks like all other places we use `rpcKeepAlive` do that?

Also - this pattern seems fundamentally somewhat flimsy.  This may be a necessary patch to take - but I expect we'll need to revisit whether this approach overall is the right foundation for ensuring we track all async work correctly.

EvanBoyle: Good catch. Follow up here: https://github.com/pulumi/pulumi/pull/7707

> this pattern seems fundamentally somewhat flimsy
Agree, I called this out in slack. Unfortunately our current architecture forces us to account for async work as we allow reusing the process across multiple automation api updates. Open to ideas or brainstorming here. Perhaps we could move some of this accounting into the debuggablePromise implementation, but that just sort of shifts the problem around.

PR #7707, opened 2021-08-04
free prepResource in finally

EvanBoyle:
follow up from https://github.com/pulumi/pulumi/pull/7704

RPC queue should be freed in a finally block in `prepareResource`


lukehoban (approved):

EvanBoyle:
Unrelated failure, got passing runs for win + mac + ubuntu between two different runs.

Issue #7000, opened 2021-05-08
Allow -s commands without a Pulumi.yaml file


joeduffy:
If I try to run

```
$ pulumi cancel -s ...some stack...
```

The CLI tries to look for a `Pulumi.yaml` file and fails:

```
error: no Pulumi project found in the current working directory
```

This command should work fine without needing to be within a valid Pulumi project.

I vaguely remember issues like this in the past, but can't find an issue (open or closed) tracking it.

stack72:
The issue before was for Pulumi stack rm without the need for the Pulumi.yaml file

EvanBoyle:
@joeduffy how would you expect this to work? I guess this would only work if the specified stack was "fully qualified" ie `pulumi cancel -s org/proj/stack`.

joeduffy:
> I guess this would only work if the specified stack was "fully qualified"

Agree, since otherwise we don't know the project name. I thought we do this for other stack commands already, though, like `pulumi stack rm -s org/proj/stack`?

joeduffy:
(That's why I labeled it a bug, by the way; cancel shouldn't need the project file, just as removing a stack doesn't either.)

EvanBoyle:
Thanks for the clarification, added the bug label back!

EvanBoyle:
@joeduffy went to try and fix this and discovered it actually works as expected for all the commands that I tested (rm, cancel, history). Going to close this out. I will note that you get the original error described in the issue if you don't use a fully qualified stack name: 

```
$ pulumi cancel -s dev
error: no Pulumi project found in the current working directory

$ pulumi cancel -s org/proj/dev
This will irreversibly cancel the currently running update for 'dev'!
Please confirm that this is what you'd like to do by typing ("dev"): dev
error: [409] Conflict: The Update has already completed

```

joeduffy:
> I will note that you get the original error described in the issue if you don't use a fully qualified stack name:

Should we improve the error message? That bit wasn't clear to me and, although it's obvious in hindsight you would need to supply a fully qualified name, apparently it wasn't obvious to me when I tried to do this and failed!

EvanBoyle:
👍  https://github.com/pulumi/pulumi/pull/7708

PR #7708, opened 2021-08-04
improve error message when flag passed to -s is not fully qualified

EvanBoyle:
follow up from https://github.com/pulumi/pulumi/issues/7000


komalali (approved):

EvanBoyle:
unrelated test failure

Issue #6968, opened 2021-05-05
[programgen/go] Go example for azure-native is invalid


mikhailshilkov:
I had to exclude the `azure-native` example of programgen for Go [here](https://github.com/pulumi/pulumi/blob/20bf4c8b677cab117348450e3162636b680ecdde/pkg/codegen/go/gen_program_test.go#L31) in https://github.com/pulumi/pulumi/pull/6430 because it generates invalid Go code (the problem was already there before the PR, the PR introduced a test that surfaces it). Go program gen seems to fail to generate an array of object types correctly.

This affects quite a few examples in the Azure Native schema.

PR #7711, opened 2021-08-05
fix multiple edge cases in go program gen

EvanBoyle:
Fixes multiple issues in go program gen that regressed in https://github.com/pulumi/pulumi/pull/7059

Fixes https://github.com/pulumi/pulumi/issues/6968 

The azure native example now generates correctly with the exception of some plain types not getting wrapped in pulumi types which was also a regression from 7059. 

Fixes are:

1. Use appropriate array types for input args
2. Fix some situations where Args suffixes were not being added appropriately
3. Fix some some typing issues within generated apply functions
4. add modules to root level types (ie `aws.GetAmiArgs`)

@pgavlin is opening follow up issues for the remaining regressions from 7059

update: currently blocked by https://github.com/pulumi/pulumi/issues/7713

### DIFF
--- a/pkg/codegen/go/gen.go
+++ b/pkg/codegen/go/gen.go
@@ -31,9 +31,11 @@ import (
 	"strings"
 	"unicode"
 
+	"github.com/hashicorp/hcl/v2"
 	"github.com/pkg/errors"
 
 	"github.com/pulumi/pulumi/pkg/v3/codegen"
+	"github.com/pulumi/pulumi/pkg/v3/codegen/hcl2"
 	"github.com/pulumi/pulumi/pkg/v3/codegen/schema"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 )
@@ -155,11 +157,11 @@ func (pkg *pkgContext) tokenToType(tok string) string {
 		}
 	}
 
-	if mod == "" {
-		mod = components[0]
-	}
 	if mod == pkg.mod {
 		return name
+	}
+	if mod == "" {
+		mod = components[0]
 	}
 
 	mod = strings.Replace(mod, "/", "", -1) + "." + name
@@ -405,7 +407,15 @@ func (pkg *pkgContext) typeStringImpl(t schema.Type, argsType bool) string {
 		typ := "map[string]"
 		return typ + pkg.typeStringImpl(t.ElementType, argsType)
 	case *schema.ObjectType:
-		return pkg.resolveObjectType(t)
+		typ := pkg.resolveObjectType(t)
+		// handle top-level/namespaceless args
+		if argsType && !strings.Contains(typ, ".") {
+			prefix := pkg.getImportPrefix(t.Token, t)
+			if prefix != "" {
+				return prefix + "." + typ
+			}
+		}
+		return typ
 	case *schema.ResourceType:
 		return "*" + pkg.resolveResourceType(t)
 	case *schema.TokenType:
@@ -446,6 +456,30 @@ func (pkg *pkgContext) typeStringImpl(t schema.Type, argsType bool) string {
 	}
 
 	panic(fmt.Errorf("unexpected type %T", t))
+}
+
+func (pkg *pkgContext) getImportPrefix(tok string, typ schema.Type) string {
+	var tokenRange hcl.Range
+	pkgName, module, _, _ := hcl2.DecomposeToken(tok, tokenRange)
+	// namespaceless invokes
+	if module == "" || strings.HasPrefix(module, "/") || strings.HasPrefix(module, "index/") {
+		module = pkgName
+	}
+	info, ok := pkg.pkg.Language["go"].(GoPackageInfo)
+	if !ok {
+		return module
+	}
+
+	if m, ok := info.ModuleToPackage[module]; ok {
+		module = m
+	}
+
+	imp := fmt.Sprintf("%s/%s", info.ImportBasePath, module)
+	if alias, ok := info.PackageImportAliases[imp]; ok {
+		imp = alias
+	}
+
+	return strings.Split(imp, "/")[0]
 }
 
 func (pkg *pkgContext) typeString(t schema.Type) string {
--- a/pkg/codegen/go/gen.go
+++ b/pkg/codegen/go/gen.go
@@ -155,12 +155,13 @@ func (pkg *pkgContext) tokenToType(tok string) string {
 		}
 	}
 
-	if mod == pkg.mod {
-		return name
-	}
 	if mod == "" {
 		mod = components[0]
 	}
+	if mod == pkg.mod {
+		return name
+	}
+
 	mod = strings.Replace(mod, "/", "", -1) + "." + name
 	return strings.Replace(mod, "-provider", "", -1)
 }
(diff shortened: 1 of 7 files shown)


pkg/codegen/internal/test/testdata/aws-webserver.pp.go
```diff
@@ -27,8 +27,8 @@ func main() {
 		}
 		opt0 := true
 		ami, err := aws.GetAmi(ctx, &GetAmiArgs{
-			Filters: []GetAmiFilter{
-				GetAmiFilter{
+			Filters: []GetAmiFilterArgs{
```
EvanBoyle, line 30: This is a regression I am introducing... Patch likely needs a little more work.

EvanBoyle: Fixed this, as well as the missing module prefix.

EvanBoyle:
Seems to be a case of fix one thing break another. The code that dealt with namespaceless types was removed and unified with some of the code that does SDK generation... I can understand the desire to reuse code but this makes things pretty brittle and difficult to change. 

Have some ideas on how to fix the immediate problem and will give that another try tomorrow.

EvanBoyle:
Looks like everything is now passing, except for the azure native downstream tests

https://github.com/pulumi/pulumi/pull/7711/checks?check_run_id=3258759554#step:8:273

However, I think this is related to https://github.com/pulumi/pulumi/issues/7595

cc @mikhailshilkov to confirm.

mikhailshilkov:
> However, I think this is related to #7595

Indeed, azure-native downstream failure is unrelated to your changes

mikhailshilkov (approved):
This looks great to me, thank you!

pgavlin (commented):
The string/suffix matching worries me. It's not a reliable test for the cases we're looking for--it may catch types that happen to end in one of the suffixes being checked w/o there being any relationship between the suffix and the type (e.g. imagine an object type with the token `pkg:module:fooArgs`). I'd prefer that we move away from string matching/suffix matching and towards using the structure of the types themselves (i.e. the `hcl2.Type` values and their annotations)` to make these checks.

pkg/codegen/go/gen.go
```diff
@@ -401,7 +407,15 @@ func (pkg *pkgContext) typeStringImpl(t schema.Type, argsType bool) string {
...
 	case *schema.ObjectType:
-		return pkg.resolveObjectType(t)
+		typ := pkg.resolveObjectType(t)
+		// handle top-level/namespaceless args
+		if argsType && !strings.Contains(typ, ".") {
+			prefix := pkg.getImportPrefix(t.Token, t)
+			if prefix != "" {
+				return prefix + "." + typ
+			}
+		}
```
pgavlin, line 417: Feels like we could be more precise here: we'll need the prefix if the referring module does not match the referent module, so it seems like that should be the condition (rather than `argsType && !strings.Contains(top, ".")`).

pkg/codegen/go/gen_program_expressions.go
```diff
@@ -588,7 +593,22 @@ func (g *generator) argumentTypeName(expr model.Expression, destType model.Type,
...
+		suffixes := []string{
+			"Args", "Array", "Map", "Ptr", "Bool", "Int", "String", "Any",
+		}
+		hasSuffix := false
+		isInvokeArg := strings.Contains(typ, ".Lookup") || strings.Contains(typ, ".Get")
+		for _, s := range suffixes {
+			if strings.HasSuffix(typ, s) {
+				hasSuffix = true
+			}
+		}
```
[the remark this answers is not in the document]
EvanBoyle: This addresses the azure native example generating incorrect types (leaving off the `Args` suffix for some of the below such as `networking.ForwardingConfiguration`. Without this patch, the generated code is for the fully prompt version of the struct but the actual signature is the args struct with input fields. 

I am just now noticing, that all of these properties seem to use discriminated unions in the schema (and all of these properties loose inputtyness of the underlying string values at this layer as well). I bet that is a hint to the "proper" fix, but it is unclear to me what layer to make that in.

```diff
@@ -13,5 +13,5 @@
     &network.RoutingRuleArgs{
-					RouteConfiguration: &network.ForwardingConfigurationArgs{
+					RouteConfiguration: network.ForwardingConfiguration{
       OdataType: "#Microsoft.Azure.FrontDoor.Models.FrontdoorForwardingConfiguration",
-						BackendPool: &network.SubResourceArgs{
+						BackendPool: network.SubResource{
        Id: "/subscriptions/subid/resourceGroups/rg1/providers/Microsoft.Network/frontDoors/frontDoor1/backendPools/backendPool1",
@@ -30,5 +30,5 @@
       Actions: pulumi.Array{
-							&cdn.DeliveryRuleCacheExpirationActionArgs{
+							cdn.DeliveryRuleCacheExpirationAction{
         Name: "CacheExpiration",
-								Parameters: &cdn.CacheExpirationActionParametersArgs{
+								Parameters: cdn.CacheExpirationActionParameters{
          CacheBehavior: "Override",
@@ -39,5 +39,5 @@
        },
-							&cdn.DeliveryRuleResponseHeaderActionArgs{
+							cdn.DeliveryRuleResponseHeaderAction{
         Name: "ModifyResponseHeader",
-								Parameters: &cdn.HeaderActionParametersArgs{
+								Parameters: cdn.HeaderActionParameters{
          HeaderAction: "Overwrite",
@@ -48,5 +48,5 @@
        },
-							&cdn.DeliveryRuleRequestHeaderActionArgs{
+							cdn.DeliveryRuleRequestHeaderAction{
         Name: "ModifyRequestHeader",
-								Parameters: &cdn.HeaderActionParametersArgs{
+								Parameters: cdn.HeaderActionParameters{
          HeaderAction: "Overwrite",
@@ -59,5 +59,5 @@
       Conditions: pulumi.Array{
-							&cdn.DeliveryRuleRemoteAddressConditionArgs{
+							cdn.DeliveryRuleRemoteAddressCondition{
         Name: "RemoteAddress",
-								Parameters: &cdn.RemoteAddressMatchConditionParametersArgs{
+								Parameters: cdn.RemoteAddressMatchConditionParameters{
          MatchValues: []string{
```

pkg/codegen/go/gen.go
EvanBoyle: So, I agree with you here but my perception of the problem is that we have intertwined the SDK generation and program gen codepaths in a very complicated and brittle way. 

I attempted to make the fix at this layer originally by reversing these two if statements:

https://github.com/pulumi/pulumi/blob/97183b83a38385f06d3bbd4ff6999eeec2e5624e/pkg/codegen/go/gen.go#L1835-L1840

However this codepath is shared with the SDK generator and breaks all SDK gen for `pulumiTypes`. The declared types become `type aws.Foo struct`...

At this layer there is also no context from the calling package (`main`).

pkg/codegen/go/gen.go
```diff
@@ -444,6 +458,30 @@ func (pkg *pkgContext) typeStringImpl(t schema.Type, argsType bool) string {
...
+		module = m
+	}
+
+	imp := fmt.Sprintf("%s/%s", info.ImportBasePath, module)
+	if alias, ok := info.PackageImportAliases[imp]; ok {
+		imp = alias
+	}
+
+	return strings.Split(imp, "/")[0]
+}
```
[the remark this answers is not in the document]
EvanBoyle: This was previously a part of a method of the `generator` before it was deleted in #7059. It now lies in the intertwined codepath that services both retrieving types for program gen and generating sdk types and input types. 

The closest thing seems to be `tokenToPackage` but this produces incorrect results. ie instead of `aws` it returns `/getamifilter`.

lunaris:
Going to close this for now as it looks quite risky in its current form/like we'd need to think about a larger refactoring; we can revisit/reopen as and when necessary.

Issue #7717, opened 2021-08-06
Example generation panics when trying to build pulumi-aws SDKs


komalali:
Error:

```
panic: interface conversion: schema.Type is *schema.MapType, not *schema.OptionalType

goroutine 1 [running]:
github.com/pulumi/pulumi/pkg/v3/codegen/schema.(*types).bindProperties(0xc00e727c70, 0xc00564f290, 0xc005480200, 0xc, 0x10, 0xc0068ed500, 0x89460cd, 0x6, 0x0, 0x0, ...)
	/Users/stack72/Code/go/pkg/mod/github.com/pulumi/pulumi/pkg/v3@v3.7.1-0.20210714212650-083fc64ff547/codegen/schema/schema.go:1982 +0xdd5
github.com/pulumi/pulumi/pkg/v3/codegen/schema.bindResource(0xc0017b0ea0, 0x19, 0xc0047acd80, 0xcda, 0xc00564f290, 0x0, 0x0, 0xc005480200, 0xc, 0x10, ...)
	/Users/stack72/Code/go/pkg/mod/github.com/pulumi/pulumi/pkg/v3@v3.7.1-0.20210714212650-083fc64ff547/codegen/schema/schema.go:2229 +0x132
github.com/pulumi/pulumi/pkg/v3/codegen/schema.bindResources(0xc003fe3890, 0xc00e727c70, 0xc0102261b0, 0x17c, 0xc0069fc5d0, 0x0, 0x0, 0x0, 0x0)
	/Users/stack72/Code/go/pkg/mod/github.com/pulumi/pulumi/pkg/v3@v3.7.1-0.20210714212650-083fc64ff547/codegen/schema/schema.go:2326 +0x1b8
github.com/pulumi/pulumi/pkg/v3/codegen/schema.importSpec(0x893f860, 0x3, 0x9218a20, 0xc, 0x8acc1e0, 0x55, 0xc001810020, 0x2, 0x2, 0x8982ca2, ...)
	/Users/stack72/Code/go/pkg/mod/github.com/pulumi/pulumi/pkg/v3@v3.7.1-0.20210714212650-083fc64ff547/codegen/schema/schema.go:1395 +0x548
github.com/pulumi/pulumi/pkg/v3/codegen/schema.ImportSpec(...)
[...]

PR #7718, opened 2021-08-06
[programgen] - Fix a panic in property binding

komalali:
# Description

Fixes: https://github.com/pulumi/pulumi/issues/7717

## Checklist


stack72 (approved):

pgavlin:
Curious: do we know what triggered this issue? Given the code that does the property binding--which wraps all property types in an optional--I would have thought that this situation was not possible.

Issue #4698, opened 2020-05-26
Start linting all TypeScript code with ESLint


mikhailshilkov:
[TSLint](https://github.com/palantir/tslint) that we currently use in several places is deprecated in favor of ESLint.

I suggest using ESLint consistently:
- [x] Replace TSLint with ESLint in pulumi/pulumi (https://github.com/pulumi/pulumi/pull/7719)
- [ ] Replace TSLint with ESLint in pulumi/examples
- [ ] Introduce ESLint in all provider repositories starting with providers with any hand-written mixins

stack72:
Kinda tracked here - https://github.com/pulumi/examples/issues/520

horacehylee:
Hi, may I volunteer to help with this issue?

mikhailshilkov:
@horacehylee Definitely! Thank you!

horacehylee:
@mikhailshilkov I have raised a PR for migrating to ESlint, could you please take a look?

PR #7719, opened 2021-08-07
Use ESlint instead of TSlint

horacehylee:
# Description

Migrated TSlint configs to ESlint ones using [tslint-to-eslint-config](https://github.com/typescript-eslint/tslint-to-eslint-config) tool, and refined the configs to better match the current coding style.

Changes:
- [member-delimiter-style](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/member-delimiter-style.md#options) that as suggested default for type definition to be  with `semicolon`
- Indentation fixes that is enforced by [eslint-indent](https://eslint.org/docs/rules/indent#options)
- Added dependencies for ESlint with Typescript
- Removed TSlint

Fixes [#4698](https://github.com/pulumi/pulumi/issues/4698) (Start linting all TypeScript code with ESLint)

## Checklist


sdk/nodejs/runtime/settings.ts
```diff
@@ -32,7 +32,7 @@ const grpcChannelOptions = { "grpc.max_receive_message_length": maxRPCMessageSiz
 /**
  * excessiveDebugOutput enables, well, pretty excessive debug output pertaining to resources and properties.
  */
-export let excessiveDebugOutput: boolean = false;
+export const excessiveDebugOutput: boolean = false;
```
t0yv0, line 35: This might be substantive? 

```
`const` is a signal that the identifier won't be reassigned. `let` is a signal that the variable may be reassigned,
```

Will this prevent a user from reassigning this value?

t0yv0:
This is looking very helpful 👍 

Almost all changes appear to be strictly lint-level (not substantive) except one change from `let` to `const`, is that correct? This is the only issue I'd like resolved before approving.

Also I've not used these tools in a long time so can't comment on the tslint config, but I'd vote to check this in and continue iterating as needed.

horacehylee:
The `prefer-const` config is enabled for both ESlint and TSlint config, which should prefer to use `const` than `let`

sdk/nodejs/runtime/settings.ts
justinvp: Good catch.

I think the original intention was to allow it to be assigned. Though, I suspect nobody is setting it these days (besides maybe a Pulumi engineer on a rare occasion).

With ES6 modules, you aren't allowed to reassign an exported value from outside the module (imports are a read-only live view of the value), so it might be worth considering changing it to `const` to discourage anyone setting it from outside the module going forward.

Changing it to `const` does mean it isn't useful unless you're modifying the source directly (either by modifying a copy of `@pulumi/pulumi` inside `node_modules` or using `yarn link` locally). If being able to set it is missed, we could consider adding a `setExcessiveDebugOutput` function, or perhaps better, switch to using an environment variable that could work across all SDKs.

justinvp (approved):
Some minor comments. LGTM otherwise. Thanks, @horacehylee!

sdk/nodejs/runtime/settings.ts
t0yv0: In that case let's accept this change for now.

sdk/nodejs/.eslintrc.js
```diff
@@ -0,0 +1,193 @@
+module.exports = {
+    "env": {
+        "browser": true,
```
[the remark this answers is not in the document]
horacehylee: Good catch, removed

CHANGELOG_PENDING.md
```diff
@@ -18,3 +18,6 @@
 
 - [auto/nodejs] - Fix a case where inline programs could exit with outstanding async work.
   [#7704](https://github.com/pulumi/pulumi/pull/7704)
+
+- [auto/nodejs] - Use ESlint instead of TSlint
```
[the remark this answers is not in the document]
horacehylee: Thanks updated.

sdk/nodejs/runtime/settings.ts
horacehylee: > If being able to set it is missed, we could consider adding a setExcessiveDebugOutput function, or perhaps better, switch to using an environment variable that could work across all SDKs.

Could we open an issue for it? I could also work on this issue as well as a follow up.

PR #7731, opened 2021-08-10
[codegen/docs] Emit docs for methods

### LINKED ISSUE #7648: [codegen/docs] Docgen for methods
Update the doc generator to emit methods for resources. The output should be similar to functions, but on the resource page itself. For each resource method, emit the signature (in all languages), and a table of the inputs/outputs.

justinvp:
Screenshot of contrived example adding some methods to the `RandomId` resource:

Screen Shot 2021-08-10 at 2 14 59 PM

Fixes #7648


pkg/codegen/docs/gen_method.go
```diff
@@ -0,0 +1,329 @@
...
+
+		n := len(params)
+		if n == 0 {
+			functionParams[lang] = ""
+			continue
+		}
+
+		for i, p := range params {
+			if err := templates.ExecuteTemplate(b, paramTemplate, p); err != nil {
+				panic(err)
```
t0yv0, line 299: Nit: could use normal err handling here?

t0yv0 (approved):
🚢

pkg/codegen/docs/gen_method.go
justinvp: I agree with the nitpick, and normally I would prefer the standard err handling, but this is patterned off of what is [done](https://github.com/pulumi/pulumi/blob/85c15b05e4b3b8f97375211209d3b1507a31d9fe/pkg/codegen/docs/gen.go#L1032-L1041) [elsewhere](https://github.com/pulumi/pulumi/blob/85c15b05e4b3b8f97375211209d3b1507a31d9fe/pkg/codegen/docs/gen.go#L1260-L1269) [in](https://github.com/pulumi/pulumi/blob/85c15b05e4b3b8f97375211209d3b1507a31d9fe/pkg/codegen/docs/gen_function.go#L269-L278) this package, and replumbing it to do the standard thing would be a larger change than I'd like to bite off right now. So I'm going to keep it as-is.

t0yv0: Right, there's probably a reason for it (like the errors being almost impossible).

Issue #7542, opened 2021-07-15
Consistently implement (and document) depends on a component across languages


lukehoban:
Today. using `dependsOn: [component]` behaves differently depending on whether `component` is a remote or non-remote component, and on what language the `dependsOn` is used in:
1. Depending on a remote component will always wait on all of its children, regardless of the language of the host or guest resource (one exception to this is being fixed in https://github.com/pulumi/pulumi/pull/7541).
2. Depending on a non-remote component in Node.js will:
a. In Node.js and .NET: wait on all of its transitive custom and remote resource children.
b. In Go and Python: not wait on any of its children.

We should ensure this is implemented consistently across language runtimes.

emiliza:
From planning: Initial task- sketch a design, don't just copy pasta node work
cc: @t0yv0 @pgavlin

clstokes:
Per the end user, a fix for Go is more important than Python here, but fixes for both are needed.

pgavlin:
I think that porting the existing logic is fine, but the implementations will definitely be a bit unwieldy.

As far as a longer-term solution for the resource model itself, how about stating that depending on a component depends on all of the children that were known at the time the component's `RegisterResourceOutputs` call was made? This is a behavior that can be implemented in the engine. I believe that it is flexible enough to accommodate even those components that create resources in applies (e.g. Helm charts), provided that those resources are designed s.t. they do not call `RegisterResourceOutputs` until any resource-creating applies have run (e.g. for a Node Helm chart, `RRO` would depend on the `ready` property).

t0yv0:
1. Notes on non-determinism 

I've found this in the docs: "Also inside of its constructor, allocate any child resources, passing the parent option as appropriate to ensure component resource children are parented correctly." Do we actually prohibit or discourage creating children outside the constructor? This seems possible and a source of non-determinism.

It feels a little unfortunate to be stating that our desired semantic is non-deterministic ("all of the children that were known at the time of RegisterResourceOutputs call"), since valid programs can declare children before or after the call. Perhaps this could be slightly amended to making sure that programs that accept children after such time fail with an exception, to direct users to fix it? Still even then, the failure would be non-deterministic.

This may be wishful thinking (out of reach), but is it quite impossible to make sure language SDKs lexically mark off a block where children may be allocated, so we know precisely (as an event) when all children are finished being added and disallow any new children past that point. This can be constructor, or "RegisterResourceOutputs", or something similar.

[...]

PR #7732, opened 2021-08-10
[sdk/python] Transitive component dependencies.

pgavlin:
Implement Node/.NET-style dependency semantics for component resources.
Depending on a component implicitly depends on all of the component's
children. The exact set of children depends on exactly when the
component resource is observed.

Part of #7542.

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -4,5 +4,9 @@
   child resources.
   [#7732](https://github.com/pulumi/pulumi/pull/7732)
 
+- [sdk/python] - Depending on a component now depends on the transitive closure of its
+  child resources.
+  [#7732](https://github.com/pulumi/pulumi/pull/7732)
+
 ### Bug Fixes
 
--- a/sdk/python/lib/pulumi/runtime/resource.py
+++ b/sdk/python/lib/pulumi/runtime/resource.py
@@ -15,7 +15,7 @@ import asyncio
 import os
 import traceback
 
-from typing import Optional, Any, Callable, List, NamedTuple, Dict, Set, Tuple, Union, TYPE_CHECKING, cast, Mapping, Sequence
+from typing import Optional, Any, Callable, List, NamedTuple, Dict, Set, Tuple, Union, TYPE_CHECKING, cast, Mapping, Sequence, Iterable
 from google.protobuf import struct_pb2
 import grpc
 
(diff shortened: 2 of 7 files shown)


sdk/python/lib/pulumi/runtime/resource.py
```diff
@@ -661,15 +657,37 @@ def convert_providers(
     return result
 
 
-async def _resolve_depends_on_urns(options: 'ResourceOptions') -> List[str]:
+async def _add_dependency(deps: Set[str], res: 'Resource'):
+    from .. import ComponentResource # pylint: disable=import-outside-toplevel
+
+    if isinstance(res, ComponentResource):
```
justinvp, line 663: Should we include a comment along the lines of the following somewhere in the Python SDK?

https://github.com/pulumi/pulumi/blob/e567b4762fe0dbb88ec35db2f3677a95cfdd715b/sdk/nodejs/runtime/resource.ts#L600-L627

justinvp (approved):
LGTM

sdk/python/lib/pulumi/runtime/resource.py
```diff
@@ -661,15 +657,37 @@ def convert_providers(
...
+        for child in res._childResources:
+            await _add_dependency(deps, child)
+        if not res._remote:
+            return
+
+    urn = await res.urn.future()
+    if urn:
+        print(f"adding dep {urn}")
```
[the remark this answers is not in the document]
pgavlin: Ah, nope

sdk/python/lib/pulumi/runtime/resource.py
```diff
@@ -661,15 +657,37 @@ def convert_providers(
...
 
 
-async def _resolve_depends_on_urns(options: 'ResourceOptions') -> List[str]:
+async def _add_dependency(deps: Set[str], res: 'Resource'):
+    from .. import ComponentResource # pylint: disable=import-outside-toplevel
+
+    if isinstance(res, ComponentResource):
+        for child in res._childResources:
+            await _add_dependency(deps, child)
+        if not res._remote:
```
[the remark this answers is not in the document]
pgavlin: That's correct. We do not add in-process components to the list of dependencies. I'm not sure that there would be any harm in doing so, but this is consistent with Node.

sdk/python/lib/pulumi/runtime/resource.py
pgavlin: Yes, that's a good idea.

CHANGELOG_PENDING.md
```diff
@@ -1,49 +1,5 @@
 ### Improvements
 
-- [cli] - Fix the preview experience for unconfigured providers. Rather than returning the
```
pgavlin, line 3: Not sure what's up with the CL. I'll get this fixed before merging.

PR #7736, opened 2021-08-10
[sdk/python] Fix pulumi.property's default value handling

### LINKED ISSUE #7385: Pulumi input_type decorator default property value
## problem
Using pulumi decorator for input_type and setting property default value doesn't seem like its getting respected.

## Steps to reproduce minimal
```
import pulumi

@pulumi.input_type
class Test:
    a: pulumi.Input[str] = pulumi.property("a", default="foo")
    b: pulumi.Input[str] = pulumi.property("b", default="bar")

ham = Test()

print(ham.a)
# actual: "a"
# expected "foo"

print(ham.b)
# actual: "b"
# expected "bar"

```

Expected: default value of "foo", and "bar" for a, and b respectively
Actual: property name is set as default instead of default value

Maybe I am not understanding the use of pulumi.input_type correctly?  Thank you for your help and time.

justinvp:
We were incorrectly setting the name of the attribute rather than the intended default value.

Fixes #7385


t0yv0 (approved):
🚢

PR #7737, opened 2021-08-10
[sdk/go] Transitive component dependencies.

pgavlin:
Implement Node/.NET-style dependency semantics for component resources.
Depending on a component implicitly depends on all of the component's
children. The exact set of children depends on exactly when the
component resource is observed.

Part of #7542.

### DIFF
--- a/sdk/go/pulumi/resource.go
+++ b/sdk/go/pulumi/resource.go
@@ -16,6 +16,7 @@ package pulumi
 
 import (
 	"reflect"
+	"sync"
 
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
 )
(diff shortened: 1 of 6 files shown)


sdk/go/pulumi/rpc.go
```diff
@@ -63,14 +63,68 @@ func mapStructTypes(from, to reflect.Type) func(reflect.Value, int) (reflect.Str
...
+
+func (s urnSet) values() []URN {
+	values := make([]URN, 0, len(s))
+	for v := range s {
+		values = append(values, v)
+	}
+	return values
+}
+
+func addDependency(ctx context.Context, deps urnSet, res Resource) error {
```
t0yv0, line 91: I guess the Context here in the signature is a good hint that this func will be awaiting things.

sdk/go/pulumi/rpc.go
```diff
@@ -63,14 +63,68 @@ func mapStructTypes(from, to reflect.Type) func(reflect.Value, int) (reflect.Str
...
+	if _, custom := res.(CustomResource); !custom {
+		for _, child := range res.getChildren() {
+			if err := addDependency(ctx, deps, child); err != nil {
+				return err
+			}
+		}
+		if !res.isRemoteComponent() {
```
t0yv0, line 98: I'm a bit confused here. For non-isRemoteComponent, we skip adding `res.URN()` to the `deps`? Why?

sdk/go/pulumi/resource.go
```diff
@@ -162,6 +204,12 @@ type Resource interface {
 
 	// addTransformation adds a single transformation to the resource.
 	addTransformation(t ResourceTransformation)
+
+	// markRemoteComponent marks this resource as a remote component resource.
+	markRemoteComponent()
+
+	// isRemoteComponent returns true if this is a local (i.e. in-process) component resource.
```
t0yv0, line 211: This reads counter-intuitive to me. `isRemoteComponent=true` indicates this is a local component resource. Would it be more intuitive to have that isRemoteComponent=false`? Perhaps better naming alltogether?

sdk/go/pulumi/rpc.go
```diff
@@ -63,14 +63,68 @@ func mapStructTypes(from, to reflect.Type) func(reflect.Value, int) (reflect.Str
...
+	return ok
+}
+
+func (s urnSet) union(other urnSet) {
+	for v := range other {
+		s.add(v)
+	}
+}
+
+func (s urnSet) values() []URN {
```
t0yv0, line 83: Might sort it to make tests deterministic/stable.

sdk/go/pulumi/rpc_test.go
```diff
@@ -829,3 +830,56 @@ func TestInvalidArchive(t *testing.T) {
 	_, _, err = marshalInput(d, archiveType, true)
 	assert.Error(t, err)
 }
+
+func TestDependsOnComponent(t *testing.T) {
```
t0yv0, line 834: Maybe this is our chance to be just a little more verbose and record the reference knowledge here, for example I'm still a little confused about the different treatment for e.g. CustomResource, we could add a test to make it obvious it's intentional, and add a few comments here. Alternatively if we have a good doc reference for intended behavior, we could link it here.

sdk/go/pulumi/resource.go
pgavlin: Ah, the comment is outdated. Sorry about the confusion here--the method name is accurate.

sdk/go/pulumi/rpc.go
```diff
@@ -63,14 +63,68 @@ func mapStructTypes(from, to reflect.Type) func(reflect.Value, int) (reflect.Str
...
+	values := make([]URN, 0, len(s))
+	for v := range s {
+		values = append(values, v)
+	}
+	return values
+}
+
+func addDependency(ctx context.Context, deps urnSet, res Resource) error {
+	if _, custom := res.(CustomResource); !custom {
```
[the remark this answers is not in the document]
pgavlin: Correct.

sdk/go/pulumi/rpc.go
pgavlin: b/c a local component isn't really part of the dependency graph--instead, it's a stand-in for its descendants. I don't think that there would be any harm in adding local components to the deps, but we've already set a precedent of not doing so in the Node SDK.

sdk/go/pulumi/rpc.go
```diff
@@ -63,14 +63,68 @@ func mapStructTypes(from, to reflect.Type) func(reflect.Value, int) (reflect.Str
...
+			if err := addDependency(ctx, deps, child); err != nil {
+				return err
+			}
+		}
+		if !res.isRemoteComponent() {
+			return nil
+		}
+	}
+
+	urn, _, _, err := res.URN().awaitURN(ctx)
```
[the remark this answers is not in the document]
pgavlin: This is kind of a lame answer, but the URN will be the empty string if `isKnown` is `false` and we've already used this pattern elsewhere, and I'd rather be consistent with the existing code.

sdk/go/pulumi/rpc.go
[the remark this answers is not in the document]
pgavlin: I'll add a similar comment here.

sdk/go/pulumi/rpc.go
pgavlin: I went back and forth on that. All consumers of this will sort the list, so I didn't sort it here, but there's not really any harm in doing so. WDYT?

PR #7739, opened 2021-08-11
[WIP] Fix corruption of workspace files under heavy contention

### LINKED ISSUE #7738: Corrupt workspace settings files under contention
When many separate Pulumi CLI processes heavily contend for the same folder workspace, some of them are prone to fail when reading corrupt JSON settings files. Currently these files are used to associate a stack name with a given folder. 

The core issue here is that `ioutil.WriteFile` is not atomic. It's unclear if the scenario is realistic enough to justify bringing additional dependencies that aim to provide an atomic version of this function:

- https://github.com/google/renameio but no Windows support?
- https://github.com/natefinch/atomic

## Steps to reproduce

A test will be added in the accompanying PR.

t0yv0:
# Description

I would like to make sure the test passes on Windows here. We should also weigh in the benefit of fixing this versus bringing in the additional dependency. The scenario is rather unlikely in practice.

Fixes #7738

## Checklist


EvanBoyle (approved):
So, generally this dependency seems fairly isolated and I am fine with bringing it in. However I will note two things:

1. sum of dependent code is <100LoC
2. the dependency has no no tests and no CI
3. it is permissibly licensed

Given the above, it may be worth considering either forking the dep or rewriting it within the pulumi/pulumi codebase.

lunaris:
Thoughts on closing this, or potentially accepting the decent (but not perfect) improvement of reimplementing it using the features added in the linked #8393 (`atomicWriteFile`)?

lunaris:
Closing since this doesn't appear to be a priority; we can revisit later if it is, potentially using the work in #8393

t0yv0:
https://github.com/pulumi/pulumi/pull/8393 is not really atomic in general as in non atomic on Windows but since it's not a priority closing is the right call.

PR #7746, opened 2021-08-11
Re-introduce the fix for 7734 and mitigate CI hangs

t0yv0:
# Description

Re-add commits fixing 7734 and fix the test contamination issue that caused master builds to hang. Basically RPC_MANAGER state got polluted between the tests.

Fixes # 7734

## Checklist


t0yv0:
When investigating hangs on master, I found that there is a test bleed e.g.:

```
pytest -v lib/test/test_monitor_termination.py # ok
pytest -v lib/test/test_next_serialize.py # ok
pytest -v lib/test/test_monitor_termination.py lib/test/test_next_serialize.py # fail
```

Debugging revealed that the bleed amounts to the new exception gets stuck in RPC_MANAGER, and re-thrown when other tests do `wait_for_rpcs`.

The suggested workaround is to clear RPC_MANAGER in the Pulumi test decorator as part of cleanup after a test.

With this working, the next issue was that `pytest` invocations print a lot of errors about background Tasks failing with the new exception and not being observed/handled. Added the logging suppression borrowed from `sdk/python/cmd/pulumi-language-python-exec`.

sdk/python/lib/pulumi/runtime/mocks.py
```diff
@@ -21,22 +21,30 @@
...
+    """
 
     @functools.wraps(fn)
     def wrapper(*args, **kwargs):
         from .. import Output  # pylint: disable=import-outside-toplevel
-        _sync_await(run_pulumi_func(lambda: _sync_await(Output.from_input(fn(*args, **kwargs)).future())))
+
+        try:
+            _sync_await(run_pulumi_func(lambda: _sync_await(Output.from_input(fn(*args, **kwargs)).future())))
+        finally:
```
t0yv0, line 46: If this is too much blast radius, I can make it local to the test that need it. It does however seem reasonable to increase test isolation.

justinvp (approved):
LGTM

PR #7755, opened 2021-08-12
Allow Python dynamic provider resources to be constructed outside of `__main__`

### LINKED ISSUE #7453: Dynamic resources in Python don't work outside of __main__
Instantiating a dynamic resource in Python outside of the `__main__` module fails. This was filed as #4100 but that issue was incorrectly closed, in my opinion! The repro there is totally legit and I am copy/pasting it below.

## Steps to reproduce

Run `pulumi up` on this program:

```py
# __main__.py
import custom_resource

# custom_resource.py
import pulumi
import pulumi.dynamic


class CustomDynamicResourceProvider(
    pulumi.dynamic.ResourceProvider,
):
    pass


pulumi.dynamic.Resource(
    provider=CustomDynamicResourceProvider(),
    name='my-resource',
    props={},
)
```

It will incorrectly fail:

```
Previewing update (dev)

View Live: https://app.pulumi.com/benesch/pulumi-4100/dev/previews/b9c38f39-280d-40d2-9bed-e6c5ccd7636c

     Type                               Name             Plan       Info
 +   pulumi:pulumi:Stack                pulumi-4100-dev  create     
     └─ pulumi-python:dynamic:Resource  my-resource                 1 error
 
Diagnostics:
  pulumi-python:dynamic:Resource (my-resource):
    error: Exception calling application: Program run without the Pulumi engine available; re-run using the `pulumi` CLI
```

I've put this together in a Git repository, if that's easier: https://github.com/benesch/pulumi-4100

As far as I can tell, all the dynamic resource examples in the pulumi/examples repository do so from `__main__` so this bug is never observed.

cc @leezen @ilaif

lukehoban:
The underlying library `dill` that we use for serializing dynamic providers into Pulumi state for Python dynamic providers serializes classes differently depending on whether they are in `__main__` or in another module.  We need the by-value serialization to be applied in all cases.

https://github.com/uqfoundation/dill/issues/424 is tracking adding the ability into `dill` to specify this by-value serialization explicitly, but until then, we will temporarily re-write the `__module__` of a provider class prior to serialization, so that `dill` behaves as we need for the dynamic provider use case.

Fixes #7453.

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -3,3 +3,5 @@
 
 ### Bug Fixes
 
+- [sdk/python] Allow Python dynamic provider resources to be constructed outside of `__main__`. 
+  [#7755](https://github.com/pulumi/pulumi/pull/7755)
--- a/sdk/python/lib/pulumi/dynamic/dynamic.py
+++ b/sdk/python/lib/pulumi/dynamic/dynamic.py
@@ -225,7 +225,7 @@ def serialize_provider(provider: ResourceProvider) -> str:
     pty = type(provider)
     oldptymod = pty.__module__
     try:
-        # TODO[uqfoundation/dill#424]: Dill currently only serializes classes by value when they are in `__main__`. 
+        # TODO[uqfoundation/dill#424]: Dill currently only serializes classes by value when they are in `__main__`.
         # Since we need to ensure classes are serialized by value for dynamic provider serialization to work, we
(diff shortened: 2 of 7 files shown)


t0yv0 (approved):
LGTM

benesch:
This is wonderful! Thanks very much, @lukehoban!

justinvp (approved):

PR #7761, opened 2021-08-13
[backend/http] Better lease renewal diagnosability

### LINKED ISSUE #3055: Log when the renewal of the lease token fails
An other issue has caused the CLI to stall and therefore not renew the lease token in time. We should log when the renewal of the lease token fails for better visibility in the CLI.

pgavlin:
- Add context to the lease renewal failure error
- Spew to the -v=3 logs when lease renewal fails

Fixes #3055.


chrsmith (approved):
LGTM

t0yv0 (approved):
🚀

PR #7762, opened 2021-08-13
Fix null exceptions when reading unknown outputs

t0yv0:
# Description

Based on investigating a user's issue, there are situations when unknown Output<T> are awaited into `default!` values causing confusing NPEs. I am making an edit here that forces call sites of the GetValueAsync to make a decision, and adjusting one callsite that triggered the NPE I've been observing. 

Open to feedback here.

My FP background means I'm biased to making sure undefined values are tracked in the type system somehow and callsites are forced to make a decision, but if this is not how we do things here, I'm open to a different approach.

Relates to: https://github.com/pulumi/pulumi/issues/7422

## Checklist


sdk/dotnet/Pulumi/Core/InputList.cs
```diff
@@ -147,7 +147,8 @@ IEnumerator IEnumerable.GetEnumerator()
 
         public async IAsyncEnumerator<Input<T>> GetAsyncEnumerator(CancellationToken cancellationToken)
         {
-            var data = await _outputValue.GetValueAsync().ConfigureAwait(false);
```
t0yv0, line 150: This changes behavior, but is reasonable? If someone uses an unknown list as an GetAsyncEnumerator, this will treat it as empty. Before, it would have thrown an exception. We have options here: we can restore previous behavior with `whenUnknown: default!` or we can cook up a more descriptive exception.

sdk/dotnet/Pulumi/Core/InputMap.cs
```diff
@@ -113,7 +114,8 @@ IEnumerator IEnumerable.GetEnumerator()
 
         public async IAsyncEnumerator<Input<KeyValuePair<string, V>>> GetAsyncEnumerator(CancellationToken cancellationToken)
         {
-            var data = await _outputValue.GetValueAsync().ConfigureAwait(false);
+            var data = await _outputValue.GetValueAsync(whenUnknown: ImmutableDictionary<string, V>.Empty)
```
t0yv0, line 117: Similar comment applies as in InputList.

sdk/dotnet/Pulumi/Core/OutputUtilities.cs
```diff
@@ -61,7 +61,7 @@ public static async Task<bool> GetIsKnownAsync<T>(Output<T> output)
         /// </summary>
         /// <param name="output">The <see cref="Output{T}"/> to evaluate.</param>
         public static Task<T> GetValueAsync<T>(Output<T> output)
```
t0yv0, line 63: Not changing the public exported method (yet). Though an argument can be made for removing this from the public API as it is dangerous.

sdk/dotnet/Pulumi/Deployment/Deployment.EnginerLogger.cs
```diff
@@ -141,7 +141,7 @@ private static async Task<string> TryGetResourceUrnAsync(Resource? resource)
                 {
                     try
                     {
-                        return await resource.Urn.GetValueAsync().ConfigureAwait(false);
+                        return await resource.Urn.GetValueAsync(whenUnknown: "").ConfigureAwait(false);
```
t0yv0, line 144: No change in behavior it seems because of the try/catch and return "".

sdk/dotnet/Pulumi/Deployment/Deployment_Prepare.cs
```diff
@@ -40,8 +40,12 @@ public partial class Deployment
             // If no parent was provided, parent to the root resource.
             LogExcessive($"Getting parent urn: t={type}, name={name}, custom={custom}, remote={remote}");
             var parentUrn = options.Parent != null
-                ? await options.Parent.Urn.GetValueAsync().ConfigureAwait(false)
-                : await GetRootResourceAsync(type).ConfigureAwait(false);
+                ? await options.Parent.Urn.GetValueAsync(whenUnknown: "").ConfigureAwait(false)
```
t0yv0, line 43: This seemed like a reasonable thing to default to branch 2 when options.Parent is not null but unknown; however I'm not sure. Perhaps better to revert to original behavior simply with `default!`.

sdk/dotnet/Pulumi/Deployment/Deployment_Prepare.cs
```diff
@@ -97,8 +101,8 @@ public partial class Deployment
             var uniqueAliases = new HashSet<string>();
             foreach (var alias in res._aliases)
             {
-                var aliasVal = await alias.ToOutput().GetValueAsync().ConfigureAwait(false);
-                if (uniqueAliases.Add(aliasVal))
+                var aliasVal = await alias.ToOutput().GetValueAsync(whenUnknown: "").ConfigureAwait(false);
```
t0yv0, line 104: Seems reasonable to skip unknown aliases.

sdk/dotnet/Pulumi/Deployment/Deployment_Prepare.cs
```diff
@@ -121,7 +125,7 @@ void LogExcessive(string message)
         }
 
         private static Task<ImmutableArray<Resource>> GatherExplicitDependenciesAsync(InputList<Resource> resources)
-            => resources.ToOutput().GetValueAsync();
+            => resources.ToOutput().GetValueAsync(whenUnknown: ImmutableArray<Resource>.Empty);
```
t0yv0, line 128: This change in behavior is intentional.

I have been getting an issue on a very large stack privately shared by @gitfool that is difficult to minimize, but essentially it was hitting this line in preview and failing with:

```
System.InvalidOperationException: This operation cannot be performed on a default instance of ImmutableArray<T>.  Consider
initializing the array, or checking the ImmutableArray<T>.IsDefault property.
  at System.Collections.Immutable.ImmutableArray`1.ThrowInvalidOperationIfNotInitialized()
  at System.Collections.Immutable.ImmutableArray`1.System.Collections.Generic.ICollection<T>.get_Count()
  at System.Collections.Generic.HashSet`1..ctor(IEnumerable`1 collection, IEqualityComparer`1 comparer)
  at Pulumi.Deployment.PrepareResourceAsync(String label, Resource res, Boolean custom, Boolean remote, ResourceArgs args,
     ResourceOptions options) in Deployment_Prepare.cs:24
  at Pulumi.Deployment.RegisterResourceAsync(Resource resource, Boolean remote, Func`2 newDependency, ResourceArgs args,
     ResourceOptions options) in Deployment_RegisterResource.cs:27
  at Pulumi.Deployment.ReadOrRegisterResourceAsync(Resource resource, Boolean remote, Func`2 newDependency, ResourceArgs args
     , ResourceOptions options) in Deployment_ReadOrRegisterResource.cs:73
  at Pulumi.Deployment.CompleteResourceAsync(Resource resource, Boolean remote, Func`2 newDependency, ResourceArgs args,
     ResourceOptions options, ImmutableDictionary`2 completionSources) in Deployment_ReadOrRegisterResource.cs:89
  at Pulumi.Output`1.GetValueAsync() in Output.cs:153
  at Pulumi.Deployment.EngineLogger.TryGetResourceUrnAsync(Resource resource) in Deployment.EnginerLogger.cs:144
  at Pulumi.Deployment.<>c__DisplayClass63_0.<<RunInlineAsync>b__1>d.MoveNext() in Deployment_Inline.cs:62

```

sdk/dotnet/Pulumi/Deployment/Deployment_Prepare.cs
```diff
@@ -159,9 +163,9 @@ private static Task<ImmutableArray<Resource>> GatherExplicitDependenciesAsync(In
                     default: return false; // Unreachable
                 }
             });
-            var tasks = transitivelyReachableCustomResources.Select(r => r.Urn.GetValueAsync());
+            var tasks = transitivelyReachableCustomResources.Select(r => r.Urn.GetValueAsync(whenUnknown: ""));
             var urns = await Task.WhenAll(tasks).ConfigureAwait(false);
-            return new HashSet<string>(urns);
+            return new HashSet<string>(urns.Where(urn => !string.IsNullOrEmpty(urn)));
```
t0yv0, line 168: Seems reasonable to skip unknowns. But I'm open to reverting to `default!`.

sdk/dotnet/Pulumi/Deployment/Deployment_ReadOrRegisterResource.cs
```diff
@@ -46,16 +46,16 @@ public partial class Deployment
...
                 var id = result.Fields["id"].StringValue;
                 var state = result.Fields["state"].StructValue;
                 return (urn, id, state, ImmutableDictionary<string, ImmutableHashSet<Resource>>.Empty);
             }
-            
+
             if (options.Id != null)
             {
-                var id = await options.Id.ToOutput().GetValueAsync().ConfigureAwait(false);
+                var id = await options.Id.ToOutput().GetValueAsync(whenUnknown: "").ConfigureAwait(false);
```
t0yv0, line 58: No change in behavior, similar to `default!` because of `string.IsNullOrEmpty` below.

sdk/dotnet/Pulumi/Deployment/Deployment_RegisterResourceOutputs.cs
```diff
@@ -25,8 +25,8 @@ void IDeploymentInternal.RegisterResourceOutputs(Resource resource, Output<IDict
 
             // The registration could very well still be taking place, so we will need to wait for its URN.
             // Additionally, the output properties might have come from other resources, so we must await those too.
-            var urn = await resource.Urn.GetValueAsync().ConfigureAwait(false);
-            var props = await outputs.GetValueAsync().ConfigureAwait(false);
+            var urn = await resource.Urn.GetValueAsync(whenUnknown: default!).ConfigureAwait(false);
```
t0yv0, line 28: Leaving behavior as-before here, because I'm unsure. Should this not be running at all in preview? In that case we can opt for throwing an informative exception instead, saying this method is accidentally running in preview when it should not.

sdk/dotnet/Pulumi/Deployment/Deployment_RootResource.cs
```diff
@@ -38,7 +38,7 @@ public partial class Deployment
 
         private async Task<string> SetRootResourceWorkerAsync(Stack stack)
         {
-            var resUrn = await stack.Urn.GetValueAsync().ConfigureAwait(false);
+            var resUrn = await stack.Urn.GetValueAsync(whenUnknown: default!).ConfigureAwait(false);
```
t0yv0, line 41: Also, here, leaving previous behavior as I'm unsure.

sdk/dotnet/Pulumi/Resources/ProviderResource.cs
```diff
@@ -53,8 +53,8 @@ public ProviderResource(string package, string name, ResourceArgs args, CustomRe
 
             if (provider._registrationId == null)
             {
-                var providerUrn = await provider.Urn.GetValueAsync().ConfigureAwait(false);
-                var providerId = await provider.Id.GetValueAsync().ConfigureAwait(false);
+                var providerUrn = await provider.Urn.GetValueAsync(whenUnknown: default!).ConfigureAwait(false);
```
t0yv0, line 56: Leaving previous behavior here.

t0yv0:
@gitfool CC

sdk/dotnet/Pulumi/Core/OutputUtilities.cs
mikhailshilkov: It was added not that long ago - mainly for unit tests I believe. It's in our "semi-public" OutputUtilities and has a comment "please don't use unless you know what you are doing".

mikhailshilkov (approved):
This looks reasonable to me. I'm a bit nervous about every behavior change as I'm not sure I can think of all the possible way they could affect other things. But I haven't seen anything obviously wrong.

PR #7764, opened 2021-08-15
Support native ES module for node sdk

lewisl9029:
# Description

Node has had [native ES module support](https://nodejs.org/api/packages.html#packages_package_json_and_file_extensions) since v14. Users can specify `"type": "module"` in their package.json to have all .js files treated as ES modules instead of as CommonJS modules. 

When I tried to add Pulumi to a project with `"type": "module"` configured however, I get the following error on `pulumi up`:

```
error: Running program '<redacted>\app\infrastructure\index.js' failed with an unhandled exception:
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: <redacted>\app\infrastructure\index.js
require() of ES modules is not supported.
require() of <redacted>\app\infrastructure\index.js from <redacted>\app\node_modules\@pulumi\pulumi\cmd\run\run.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename index.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from <redacted>\app\package.json.
```

This is obviously not ideal given that Node is steadily moving towards making ES modules a first-class citizen, and likely won't be stopping anytime soon.

This PR is an attempt to make the minimum set of changes possible to support native ES modules in the Pulumi Node SDK. 

As can be seen the comments in [this commit](https://github.com/pulumi/pulumi/commit/77e051b199a5e88d8489f8cdd3934a3ebcfa40ad), I ended up having to jump through a few hoops to work around TypeScript transpiling away dynamic `import` calls into `require`s. The solution is definitely a hack, but was the best I could come up with. Happy to take suggestions on alternatives.

I've also included [a new test](https://github.com/pulumi/pulumi/pull/7764/commits/fe9a75b6269dbdad18aca2e357c1d9d37254633c) to verify that the functionality works. All existing tests seem to be passing as well, outside of a small subset of tests that fail with this error:

```
error: PULUMI_ACCESS_TOKEN must be set for login during non-interactive CLI sessions
```

But I'm assuming these failures don't have anything to do with my changes since they occur on a clean master branch as well.

## Checklist


lukehoban (approved):
Thanks @lewisl9029, and apologies for the *very* long delay getting back to you.

I spent some time with this today, and the changes here look great.  There's a couple of additional things I'd like to do as follow-up, but those can build on top of this work.  I'll work on getting this merged and then following up with those additional pieces.

PR #7766, opened 2021-08-16
Fixed nil pointer error with BindExpressionText and typeExpr

T-Vova:
# Description

`BindExpressionText()` has the potential to return a nil `typeExpr`, which leads to the subsequent `typeExpr.Type()` call panicking. A guard condition is added to return with an error if `typeExpr` is found to be nil after calling `BindExpressionText()`.

Before, the error was identified by the coverage tracker in this way:
```
{
      "Reason": "runtime error: invalid memory address or nil pointer dereference",
      "Count": 1
},
```
Now, it is more specific:
```
{
      "Reason": "cannot bind expression: \u003canonymous\u003e:0,61-62: Missing attribute separator; Expected a newline or comma to mark the beginning of the next attribute.",
      "Count": 1
},
```

### DIFF
--- a/pkg/codegen/hcl2/binder.go
+++ b/pkg/codegen/hcl2/binder.go
@@ -15,6 +15,7 @@
 package hcl2
 
 import (
+	"fmt"
 	"os"
 	"sort"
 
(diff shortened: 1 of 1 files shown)


stack72 (approved):

stack72:
Thanks @T-Vova - the downstream Azure Native build is not related to this so I am going to merge this in

PR #7770, opened 2021-08-16
[sdk/go] retract v3.10.1

pgavlin:
This version introduced a critical bug in resource argument marshaling
that caused empty properties to be marshaled as null values for any
resource that depends on other resources.

### DIFF
--- a/sdk/go.mod
+++ b/sdk/go.mod
@@ -2,6 +2,8 @@ module github.com/pulumi/pulumi/sdk/v3
 
 go 1.16
 
+retract v3.10.1 // This version introduced a critical bug in property marshaling.
+
 require (
 	github.com/Microsoft/go-winio v0.4.14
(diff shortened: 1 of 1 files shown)


pgavlin:
For reference: https://golang.org/ref/mod#go-mod-file-retract

justinvp (approved):
I'm in favor of retracting, and if we decide to do so, this change LGTM.

justinvp:
```
go: errors parsing go.mod:
Error: /home/runner/work/pulumi/pulumi/sdk/go.mod:5: unknown directive: retract
```

🤔

pgavlin:
> The retract directive was added in Go 1.16. Go 1.15 and lower will report an error if a retract directive is written in the main module's go.mod file and will ignore retract directives in go.mod files of dependencies.

...well, this might sink the idea of retracting :(

Issue #7622, opened 2021-07-23
Recursive types not supported in Go codegen


mikhailshilkov:
Unfortunately, we haven't been building the Go SDK of Google Native yet, so I don't know when this got broken (potentially, long time ago). Anyway, filing an issue to investigate. All other SDKs are fine, so I assume there's an issue with Go code generation specifically.

```
Error: 	/home/runner/work/pulumi-google-native/pulumi-google-native/sdk/go/google/bigquery/v2/pulumiTypes.go:16327:6: QueryParameterTypeResponse refers to
Error: 	/home/runner/work/pulumi-google-native/pulumi-google-native/sdk/go/google/bigquery/v2/pulumiTypes.go:16327:6: QueryParameterTypeResponse
go/google/bigquery/v2/pulumiTypes.go:19181:6: invalid recursive type StandardSqlDataTypeResponse
Error: 	/home/runner/work/pulumi-google-native/pulumi-google-native/sdk/go/google/bigquery/v2/pulumiTypes.go:19181:6: StandardSqlDataTypeResponse refers to
Error: 	/home/runner/work/pulumi-google-native/pulumi-google-native/sdk/go/google/bigquery/v2/pulumiTypes.go:19181:6: StandardSqlDataTypeResponse
# github.com/pulumi/pulumi-google-native/sdk/go/google/dialogflow/v3
go/google/dialogflow/v3/pulumiTypes.go:6603:6: GoogleCloudDialogflowCxV3IntentInput redeclared in this block
	previous declaration at go/google/dialogflow/v3/pulumiTypes.go:6355:6
# github.com/pulumi/pulumi-google-native/sdk/go/google/dialogflow/v3beta1
go/google/dialogflow/v3beta1/pulumiTypes.go:6627:6: GoogleCloudDialogflowCxV3beta1IntentInput redeclared in this block
	previous declaration at go/google/dialogflow/v3beta1/pulumiTypes.go:6379:6
# github.com/pulumi/pulumi-google-native/sdk/go/google/transcoder/v1beta1
go/google/transcoder/v1beta1/pulumiTypes.go:1784:69: cannot use v.Inputs (type []AudioChannelInput) as type []AudioChannelInput in return argument
go/google/transcoder/v1beta1/pulumiTypes.go:1808:6: AudioChannelInput redeclared in this block
	previous declaration at go/google/transcoder/v1beta1/pulumiTypes.go:1717:6
```

[...]

PR #7779, opened 2021-08-18
[codegen/go] Fix generation of nested collection types and remove invalid args types

EvanBoyle:
# Description

https://github.com/pulumi/pulumi/pull/7059/files#diff-434bc124a144c56427b080a04745148c2798b7026c4c6416f68795ddd069a997R329-R418 introduced changes that generate what appear to be invalid collection types (`ArgsArray`, `ArgsArrayMap`, etc). 

The fix here is twofold:

1. Trim `Args` suffixes from the type names in nested collection type generation
2. Break nested collection type generation into two steps: collect and dedupe types, and generate types. 

Diffs for kubernetes and terraform based providers look clean. The failing downstream is azure native which is expected as there were some breaking changes to schema that will be corrected in https://github.com/pulumi/pulumi-azure-native/pull/1077

Fixes https://github.com/pulumi/pulumi/issues/7595

## Checklist


mikhailshilkov:
This makes sense to me but I don't have a ton of context on collection types in our Go SDK, so I'd love others to review.

Does this also fix compilation errors in aws-native and google-native? (see https://github.com/pulumi/pulumi/issues/7622)

t0yv0:
Looks reasonable at a glance.

pgavlin (approved):

PR #7787, opened 2021-08-18
stabilize go codegen for nested collection types

EvanBoyle:
Fixes non-determinism introduced in https://github.com/pulumi/pulumi/pull/7779 by sorting. 

Tested this locally against azure native and it results in no diff (other than sorting a few declarations). 

Iterating over maps when generating code is a bag sad. 

## Checklist


justinvp (approved):

EvanBoyle:
Go codegen tests passing. Going to merge.

Issue #7780, opened 2021-08-18
Deletion order does not take into account (transitive) children of multi-language components


lukehoban:
The dependencies recorded in Pulumi state for resources that depend on multi-language components include the component itself, but not all of its transitive children.  This is different from in-language component dependencies, which will include all transitive children in the direct dependencies list.

Because of this, deletion order does not ensure that children of a dependent component are deleted prior to the resource that depends on that component.  

For example - given the following resource graph - where `First` and `Second` are multi-language components, constructed from an outer Pulumi program:

```
 +   pulumi:pulumi:Stack                repro-preview-repro  created     
 +   ├─ xyz:index:First                 test                 created     
 +   │  └─ xyz:index:FirstNested        test-nested          created     
 +   │     ├─ aws:ec2:NetworkInterface  test-nested-eni2     created     
 +   │     ├─ aws:ec2:NetworkInterface  test-nested-eni1     created     
 +   │     └─ aws:ec2:SecurityGroup     test-nested-sg       created     
 +   └─ xyz:index:Second                test                 created     
 +      ├─ aws:ec2:SecurityGroupRule    test-rule-7          created     
 +      ├─ aws:ec2:SecurityGroupRule    test-rule-4          created     
 +      ├─ aws:ec2:NetworkInterface     test-eni             created     
 +      ├─ aws:ec2:SecurityGroupRule    test-rule-1          created     
 +      ├─ aws:ec2:SecurityGroupRule    test-rule-0          created     
 +      ├─ aws:ec2:SecurityGroupRule    test-rule-5          created     
 +      ├─ aws:ec2:SecurityGroupRule    test-rule-9          created     
 +      ├─ aws:ec2:SecurityGroupRule    test-rule-6          created     
 +      ├─ aws:ec2:SecurityGroupRule    test-rule-3          created     
 +      ├─ aws:ec2:SecurityGroupRule    test-rule-8          created     
 +      └─ aws:ec2:SecurityGroupRule    test-rule-2          created    
```

When this is deleted, the `SecurityGroupRule::test-rule-0` resource has a dependency on `First::test`, but this does not ensure that deletion of `SecurityGroupRule::test-rule-0` happens *before* deletion of `SecurityGroup::test-nested-sg`, which is necessary for correctness.  If these had been all in-language components, `SecurityGroupRule::test-rule-0` would have had a listed dependency on `SecurityGroup::test-nested-sg` explicitly (along with all other transitive children of it's dependencies).

[...]

PR #7788, opened 2021-08-18
Include transitive children in dependency list for deletes

lukehoban:
Fixes deletion order issues outlined in #7780 related to dependencies on multi language components not ensuring that all (transitive) children of the dependency are treated as dependencies. 

Since the state file is not guaranteed to include all transitive dependencies explicitly specified (because dependencies on multi-langauge components will only include the dependency on the component itself, not all of it's transitive children), this PR enlightens the computation of dependencies for deletion ordering to expand the transitive dependency to include transitive children of direct dependencies.

Planned follow-ups:
- Adjust step generator to ensure that the state file *does* always capture the complete transitive list of dependencies
- Add test coverage for deletion order at the lifecycle test and integration test layers as well


CHANGELOG_PENDING.md
```diff
@@ -5,9 +5,15 @@
 
 ### Bug Fixes
 
+- [sdk/python] Allow Python dynamic provider resources to be constructed outside of `__main__`. 
+  [#7755](https://github.com/pulumi/pulumi/pull/7755)
+
```
justinvp, line 10: Bad merge?

pkg/resource/graph/dependency_graph.go
```diff
@@ -92,7 +93,23 @@ func (dg *DependencyGraph) DependenciesOf(res *resource.State) ResourceSet {
...
+		// Include all resources that are dependencies of the resource
+		if dependentUrns[candidate.URN] {
+			set[candidate] = true
+			// All transitive children of the dependency that are before this resource in the topological sort are
+			// also implicitly dependencies. This is necessary because for remote components, the dependencies will
+			// not include the transitive set of children directly, but will include the parent component. We must
+			// walk that component's children here to ensure they are treated as dependencies. Transitive children
+			// of the dependency that are after the resource in the topological sort are not included as this could
+			// lead to cycles in the dependency order.
+			for _, transitiveCandidateIndex := range dg.childrenOf[candidate.URN] {
```
pgavlin, line 105: Looks like we're including children of custom resources here. I think that we should only include children of component resources.

justinvp (approved):
LGTM

pgavlin (approved):

Issue #7742, opened 2021-08-11
add `PluginServerUrl` option to Automation API plugin acquisition


jaxxstorm:
If a user wants to install a custom plugin or multi language component binary, they need to install the resource plugin.

Generally, this is handled by the `--server` flag, however this is not plumbed through the automation API in all languages:

Example, in Go:

https://github.com/pulumi/pulumi/blob/013036064f8a49fca6842179a40e005da06a806a/sdk/go/auto/local_workspace.go#L375

## Affected feature
Automation API

orionstudt:
Hi @jaxxstorm I added & referenced a PR to add this functionality to the .NET Automation API as we need it internally.

PR #7796, opened 2021-08-19
[auto/dotnet] - plugin installation options: exact version, server

orionstudt:
# Description

Add functionality to .NET Automation API to allow for the `--exact` and `--server` arguments when doing `pulumi plugin install`. Related to #7742.

## Checklist


sdk/dotnet/Pulumi.Automation/Workspace.cs
```diff
@@ -259,8 +259,9 @@ public Task SelectStackAsync(string stackName)
         /// <param name="name">The name of the plugin.</param>
         /// <param name="version">The version of the plugin e.g. "v1.0.0".</param>
         /// <param name="kind">The kind of plugin e.g. "resource".</param>
+        /// <param name="options">Any additional plugin installation options.</param>
         /// <param name="cancellationToken">A cancellation token.</param>
-        public abstract Task InstallPluginAsync(string name, string version, PluginKind kind = PluginKind.Resource, CancellationToken cancellationToken = default);
+        public abstract Task InstallPluginAsync(string name, string version, PluginKind kind = PluginKind.Resource, PluginInstallOptions? options = null, CancellationToken cancellationToken = default);
```
justinvp, line 264: Technically, this is a binary breaking change. Though, I don't know if it'd affect anyone in practice since folks generally compile the source program each time, and it's unlikely anyone is passing a `cancellationToken` or overriding `InstallPluginAsync`.

If we did want to make it a binary compatible change, we'd need to add another overload. I can imagine two different approaches.

## Approach 1

Keep the existing method as-is, but change `abstract` to `virtual` and add a new `virtual` overload with the new parameter that throws:

```c#
public abstract class Workspace : IDisposable
{
    // Existing method changed from `abstract` to `virtual` and calls the new overload.
    public virtual Task InstallPluginAsync(string name, string version, PluginKind kind = PluginKind.Resource, CancellationToken cancellationToken = default)
        => InstallPluginAsync(name, version, kind, null, cancellationToken);

    // New virtual overload, that just throws at runtime.
    public virtual Task InstallPluginAsync(string name, string version, PluginKind kind = PluginKind.Resource, PluginInstallOptions? options = null, CancellationToken cancellationToken = default)
        => throw new NotSupportedException($"{nameof(InstallPluginAsync)} must be overridden");
}

public sealed class LocalWorkspace : Workspace
{
    public override Task InstallPluginAsync(string name, string version, PluginKind kind = PluginKind.Resource, PluginInstallOptions? options = null, CancellationToken cancellationToken = default)
    {
        // ... current implementation in PR ...
    }
}
```

The upside of this approach is is only requires subclasses to override a single virtual method. The downside is that these are no longer abstract, so there's no way to force a subclass to implement it, aside from the runtime exception. Most of this is a moot point because in practice only our own `LocalWorkspace` will be subclassing it.

## Approach 2

Keep the existing methods as-is (remains `abstract`) and add a new virtual overload with the new option that calls the existing abstract overload, ignoring the passed-in `PluginInstallOptions`. It'd be up to a subclass to override either the existing abstract overload alone to support the current functionality, or both to be able to support the new `PluginInstallOptions`.

```c#
public abstract class Workspace : IDisposable
{
    // Existing abstract method with no changes.
    public abstract Task InstallPluginAsync(string name, string version, PluginKind kind = PluginKind.Resource, CancellationToken cancellationToken = default);

    // New virtual method that can be overridden to support the new options.
    public virtual Task InstallPluginAsync(string name, string version, PluginKind kind = PluginKind.Resource, PluginInstallOptions? options = null, CancellationToken cancellationToken = default)
    {
        // By default, call the existing abstract `InstallPluginAsync` overload, which ignores `PluginInstallOptions`.
        // A subclass can override both to add support for `PluginInstallOptions`.
        return InstallPluginAsync(name, version, kind, cancellationToken);
    }
}

public sealed class LocalWorkspace : Workspace
{
    public override Task InstallPluginAsync(string name, string version, PluginKind kind = PluginKind.Resource, CancellationToken cancellationToken = default)
        => InstallPluginAsync(name, version, kind, null, cancellationToken);

    public override Task InstallPluginAsync(string name, string version, PluginKind kind = PluginKind.Resource, PluginInstallOptions? options = null, CancellationToken cancellationToken = default)
    {
        // ... current implementation in PR ...
    }
}
```

If we do want to make a binary compatible change, is there a preference on which approach we'd want to take?

t0yv0: ```Though, I don't know if it'd affect anyone in practice since folks generally compile the source program each time, and it's unlikely anyone is passing a cancellationToken or overriding InstallPluginAsync.```

That's what I thought too, so it though we don't care? @mikhailshilkov for another opinion. Would help to know our stance on this so we can review PRs accordingly.

mikhailshilkov: `LocalWorkspace` has the same breaking change, right? Why do you think it's unlikely that somebody uses it in an incompatible way?

sdk/dotnet/Pulumi.Automation/PluginInstallOptions.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace Pulumi.Automation
+{
+    public class PluginInstallOptions
+    {
+        /// <summary>
+        /// If <c>true</c>, force installation of an exact version match (usually >= is accepted).
+        /// <para/>
+        /// Defaults to <c>false</c>.
+        /// </summary>
+        public bool ExactVersion { get; set; } = false;
```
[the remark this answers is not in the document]
orionstudt: Probably not :)

sdk/dotnet/Pulumi.Automation/Workspace.cs
orionstudt: Yea I was thinking about this and kind of regretting that I went with the optional `= default` for our `CancellationToken`s instead of doing separate overloads to begin with.

I'm up for either suggested approach, whichever you guys decide is best.

sdk/dotnet/Pulumi.Automation/PublicAPI.Unshipped.txt
```diff
@@ -0,0 +1,8 @@
+Pulumi.Automation.PluginInstallOptions
```
orionstudt, line 1: Separate question, should API changes go straight into the `Shipped` file for a PR or is it correct to have them here in `Unshipped`?

sdk/dotnet/Pulumi.Automation/Workspace.cs
orionstudt: Could we solve it by keeping the change as I made it, and then adding an extension to support the old overload?

sdk/dotnet/Pulumi.Automation/PublicAPI.Unshipped.txt
t0yv0: Sorry for the delay here, I'll check on that and get back to you. 

We have an internal release process doc https://github.com/pulumi/home/wiki/Producing-the-Pulumi-CLI---SDKs and it does not have any step for moving Unshipped->Shipped (an oversight perhaps). So I think it should go straight into the Shipped file for now.

CC @emiliza --^ alternative here is that we fixup our release process to do this moving of Unshipped to Shipped. Context: these text files are integrated with CI-based API surface checker in .NET so any changes to public API surface need to be marked in these files to be considered intentional.

sdk/dotnet/Pulumi.Automation/Workspace.cs
t0yv0: @orionstudt your suggestion LGTM, no binary compat I think but the old source code using the old form of the method will compile without changes. @justinvp @mikhailshilkov pinging is that allright? You both have written a lot more C# code than I have, I'm mostly just interested in moving the fix forward.

mikhailshilkov: Yeah, it's binary-breaking but I'd be fine to ship this with an explicit shout-out in the changelog.

justinvp: > but the old source code using the old form of the method will compile without changes

That's true for callers, but technically anyone can subclass `Workspace` and an override of `InstallPluginAsync` in a subclass won't be source compatible with the new signature. That said, I highly doubt anyone outside this assembly is subclassing `Workspace`. So I'm good with the binary break and breaking other subclasses of `Workspace`, as long as we maintain source compat for callers.

In terms of maintaining source compat, @orionstudt, is there any benefit to the extension method over keeping the old overload on the base class (no longer abstract with implementation that calls the new overload)?

```c#
public abstract class Workspace
{
    // Existing method, no longer abstract, calls new abstract overload.
    public Task InstallPluginAsync(string name, string version, PluginKind kind = PluginKind.Resource, CancellationToken cancellationToken = default)
        => InstallPluginAsync(name, version, kind, null, cancellationToken);

    // New abstract overload.
    public abstract Task InstallPluginAsync(string name, string version, PluginKind kind = PluginKind.Resource, PluginInstallOptions? options = null, CancellationToken cancellationToken = default);
}
```

orionstudt: @justinvp Probably not, I actually would prefer the approach you just outlined because an extension would likely be in a separate file. Better that things remain close together. Assuming we are OK with potentially breaking the (highly unlikely) subclasses of `Workspace`?

justinvp: > Edit: I am getting this alert when trying to go with that approach.

Does the analyzer complain still when the defaults are removed from the old overload?

```c#
public abstract class Workspace
{
    public Task InstallPluginAsync(string name, string version, PluginKind kind, CancellationToken cancellationToken)
        => InstallPluginAsync(name, version, kind, null, cancellationToken);

    public abstract Task InstallPluginAsync(string name, string version, PluginKind kind = PluginKind.Resource, PluginInstallOptions? options = null, CancellationToken cancellationToken = default);
}
```

orionstudt: Ok should be good now I just made the params on the back-compat overload non-optional.

sdk/dotnet/Pulumi.Automation/PublicAPI.Unshipped.txt
orionstudt: Sounds good the changes have been moved to `Shipped`

CHANGELOG_PENDING.md
```diff
@@ -1,5 +1,7 @@
 ### Improvements
 
+- [auto/dotnet] - Add support for `--exact` and `--server` with `pulumi plugin install` via Automation API.
```
justinvp, line 3: Mind adding a note about the signature change for `InstallPluginAsync` (binary breaking, but largely source compatible)?

justinvp (approved):
LGTM

CHANGELOG_PENDING.md
orionstudt: Done!

t0yv0:
Thanks so much @orionstudt

PR #7781, opened 2021-08-18
Update x/sys for sdk to support go 1.17

roopakv:
# Description

This updates x/sys for the sdk and therefor adds support for go 1.17. More info here https://github.com/Homebrew/homebrew-core/pull/83413

This also ran go mod tidy to clean up the go.sum file :)

## Checklist

- [x] I have added tests that prove my fix is effective or that my feature works

PR #7799, opened 2021-08-19
Fix order of go mod tidy

t0yv0:
# Description

Fixes issues similar to the one surfacing in:
https://github.com/pulumi/pulumi/pull/7781

Basically `go mod tidy` fetches transitive dependencies, if it's placed after edits that point to the local SDK, then any new dependencies from the local SDK propagate automatically into integration tests without having to explicitly declare them. 

Questions:

- do we want that? or we actually want to explicitly declare those in tests projects

- is it safe to remove `go mod download` - I think `go mod tidy` includes that work

## Checklist

### DIFF
--- a/pkg/testing/integration/program.go
+++ b/pkg/testing/integration/program.go
@@ -1933,12 +1933,6 @@ func (pt *ProgramTester) prepareGoProject(projinfo *engine.Projinfo) error {
 		}
 	}
 
-	// initial tidy to resolve dependencies
-	err = pt.runCommand("go-mod-tidy", []string{goBin, "mod", "tidy"}, cwd)
-	if err != nil {
-		return err
-	}
-
 	// link local dependencies
 	for _, pkg := range pt.opts.Dependencies {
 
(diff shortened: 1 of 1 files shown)


lblackstone (approved):
LGTM

> do we want that? or we actually want to explicitly declare those in tests projects

I'd keep the current behavior for now, and we can follow up in another PR if desired. @justinvp might have more thoughts on that.

> is it safe to remove go mod download - I think go mod tidy includes that work

Yes, that's right

t0yv0:
Fascinating, I'm still confused by the Go ecosystem a bit.

If test or library T declares dependency on Pulumi SDK P, but P depends on Q for its internal workings, it feels weird to be asking T declare dependency on Q? If T code does not reference anything from Q? Should not transitive dependency resolution kick in in this case?

But I really am not sure and I don't have a strong opinion here. @justinvp LMK if you just want me to back out this change and I will.

justinvp:
> @justinvp LMK if you just want me to back out this change and I will.

No, this change LGTM! Maybe I misunderstood the comments above? I'm fine with this change, I was just thinking we'd continue specifying things like the following in tests for now:

https://github.com/pulumi/pulumi/blob/da2bd2ab043f2588b3a136846da1b97505326f25/tests/integration/integration_go_test.go#L25-L27

t0yv0:
Oh yes.. I think without the change, golang.org/x/sys upgrade of Pulumi SDK failed and forced one to edit or add golang.org/x/sys to go.sum of integration tests, which feels ODD. SO let's keep it.

PR #7814, opened 2021-08-20
Implement 5758 {fn}_output codgen for Python

t0yv0:
# Description

Implements #5758 for Python.

Temporarily based on t0yv0/5758-go-3 until that gets merged into master, then will rebase. The base contains shared schema tests from before.

## Checklist


t0yv0:
Expecting runtime checks fail on Windows due to paths, if so will fix then... Don't have a Windows VM handy.

PR #7817, opened 2021-08-20
Iwahbe/2715/add pulumi about command

### LINKED ISSUE #2715: Add command to output information for support
When troubleshooting issues we often need to ask users for additional information. This information typically comes from many different sources - the stack, dependencies, etc.

We should add a command or `pulumi help --support-info` (or similar) argument that gathers all of this information automatically for users to submit when they need help or support. 

Items that should be gathered:
- version info - of cli and plugins used
- OS version
- app dependency info - `package.json`, `requirements.txt`, etc
- stack info - name, resource list, any `pending_operations`, etc.
- log location information (probably not the log data themselves but provide the location in case the user wants to provide them too)
- the backend in use - `app.pulumi.com`, local, s3, etc.

This information should be output to stdout so that users can review and "scrub" information they want to omit before sharing.

iwahbe:
# Description

Adds a new command: `pulumi about` which should print out the information necessary to debug a failure in the pulumi CLI
Fixes #2715 

## Checklist


pkg/cmd/pulumi/about.go
```diff
@@ -0,0 +1,304 @@
...
+			),
+		}
+
+	return cmd
+}
+
+func formatPluginAbout() error {
+	var plugins []workspace.PluginInfo
+	var err error
+	plugins, err = workspace.GetPluginsWithMetadata()
```
emiliza, line 124: Would be great to display just the plugins and versions being used for the project, not all plugins installed

pkg/cmd/pulumi/about.go
```diff
@@ -0,0 +1,304 @@
...
+				}
+				fmt.Print("\n")
+
+				var proj *workspace.Project
+				var pwd string
+				if proj, pwd, err = readProject(); err != nil {
+					err = errors.Wrap(err, "Failed to read project for diagnosis")
+					warn(err, &opts)
+				} else {
+					fmt.Printf("This is a %s project.\n\n", proj.Runtime.Name())
```
emiliza, line 85: Also would be nice to get runtime version

pkg/cmd/pulumi/about.go
```diff
@@ -0,0 +1,304 @@
...
+		})
+	}
+	cmdutil.PrintTable(cmdutil.Table{
+		Headers: []string{"NAME", "VERSION"},
+		Rows:    rows,
+	})
+	return nil
+}
+
+func formatHostAbout() error {
```
emiliza, line 154: We don't have lots of test coverage for the CLI - let's take this opportunity to get some tests added :D

pkg/cmd/pulumi/about.go
iwahbe: The runtime version of what? Can you please clarify?

emiliza: Version of node/python/etc. Figure let's collect all the helpful telemetry while we're here

pkg/cmd/pulumi/about.go
```diff
@@ -0,0 +1,407 @@
...
+	"github.com/pulumi/pulumi/sdk/v3/go/common/diag/colors"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/util/cmdutil"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/util/executable"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/workspace"
+	"github.com/pulumi/pulumi/sdk/v3/python"
+)
+
+func newAboutCmd() *cobra.Command {
+	short := "Print information about the pulumi enviroment."
```
joeduffy, line 44:
```suggestion
	short := "Print information about the Pulumi enviroment."
```

pkg/cmd/pulumi/about.go
```diff
@@ -0,0 +1,407 @@
...
+		}),
+	})
+
+}
+
+func formatLogAbout() {
+	logDir := flag.Lookup("log_dir")
+	if logDir != nil && logDir.Value.String() != "" {
+		fmt.Printf("Pulumi locates it's logs in %s\n", logDir)
+	} else if runtime.GOOS != "windowns" {
```
joeduffy, line 319:
```suggestion
	} else if runtime.GOOS != "windows" {
```

pkg/cmd/pulumi/about.go
```diff
@@ -0,0 +1,407 @@
...
+			[]string{"Go Compiler", runtime.Compiler},
+		}),
+	})
+
+}
+
+func formatLogAbout() {
+	logDir := flag.Lookup("log_dir")
+	if logDir != nil && logDir.Value.String() != "" {
+		fmt.Printf("Pulumi locates it's logs in %s\n", logDir)
```
joeduffy, line 318:
```suggestion
		fmt.Printf("Pulumi locates its logs in %s\n", logDir)
```

pkg/cmd/pulumi/about.go
```diff
@@ -0,0 +1,407 @@
...
+	}
+
+	version = strings.TrimSpace(version)
+
+	fmt.Printf("This project is a %s project (%s %s)\n", proj.Runtime.Name(), ex, version)
+	return nil
+}
+
+func warn(err error, opts *display.Options) {
+	msg := fmt.Sprintf("%swarning:%s %s\n",
```
joeduffy, line 390: In theory this could be something like

```go
cmdutil.Diag.Warningf(&diag.Diag{Message: err.Error()})
```

pkg/cmd/pulumi/about.go
```diff
@@ -0,0 +1,546 @@
...
+	// We want Version to conform to the semvar format: v0.0.0
+	Version string `json:"version"`
+}
+
+func getProjectRuntimeAbout(proj *workspace.Project) (projectRuntimeAbout, error) {
+	var ex, version string
+	var err error
+	var out []byte
+	language := proj.Runtime.Name()
+	switch language {
```
justinvp, line 474: This is fine for now, but just pointing out that this should probably live in the language host, i.e. the CLI could ask the language host for runtime information (as well as dependency information).

We have a similar TODO for `pulumi new` where we're basically using a switch on the lang, but the functionality should really live in the language host: https://github.com/pulumi/pulumi/issues/1334

https://github.com/pulumi/pulumi/blob/784c661c0cfb96f8fa609a45fc66735678dfa613/pkg/cmd/pulumi/new.go#L577

Maybe we should add a similar TODO here (and in `formatProgramDependenciesAbout` above)?

justinvp (commented):
Some comments/questions, but otherwise LGTM!

For future PRs, it'd be nice to show some example output of running the command in the PR description.

It might be fun to add some color to the output when in interactive mode, but we can always do that later.

I ran `pulumi about -j` in an empty dir and got this:

```json
$ pulumi about -j
{
  "plugins": null,
  "host": {
    "os": "darwin",
    "version": "11.4",
    "arch": "x86_64"
  },
  "backend": {
    "name": "pulumi.com",
    "url": "https://app.pulumi.com/justinvp",
    "user": "justinvp"
  },
  "currentStack": null,
  "cliAbout": {
    "version": "3.11.0-alpha.1629745492+784c661c",
    "goVersion": "go1.17",
    "goCompiler": "gc"
  },
  "runtime": null,
  "errors": [
    {},
    {},
    {}
  ]
}
```

Two questions:

1. Should we omit empty/nil values from the JSON output (e.g. "plugins", "currentStack", "runtime") via `omitempty` in the `json` struct tags?
2. What's up with the errors array showing three empty objects? Should these just be strings?

The non-JSON output is:

```
$ pulumi about
CLI          
Version      3.11.0-alpha.1629745492+784c661c
Go Version   go1.17
Go Compiler  gc

Host     
OS       darwin
Version  11.4
Arch     x86_64

Backend  
Name     pulumi.com
URL      https://app.pulumi.com/justinvp
User     justinvp

Pulumi locates its logs in /var/folders/4q/8cmjgthn18x9gwgq4fvf4q7w0000gn/T/ by default
warning: Failed to get information about the plugin: no Pulumi.yaml project file found (searching upwards from /Users/justin/Desktop/2021/08/23/empty). If you have not created a project yet, use `pulumi new` to do so
warning: Failed to read project: no Pulumi.yaml project file found (searching upwards from /Users/justin/Desktop/2021/08/23/empty). If you have not created a project yet, use `pulumi new` to do so
warning: Failed to get information about the current stack: no Pulumi.yaml project file found (searching upwards from /Users/justin/Desktop/2021/08/23/empty). If you have not created a project yet, use `pulumi new` to do so
```

pkg/cmd/pulumi/about.go
iwahbe: That makes sense. I'll do that.

iwahbe:
I assumed that `error` would serialize to string. That was my bad. I included the empty lists and `null` values because I find it simpler to read.  If I see `currentStack: null` then I know that this is the correct printout for the current stack, I just didn't find such a stack this run. Likewise `errors: []` tells me that there are no errors, which is different from not noticing any errors.

pkg/cmd/pulumi/about.go
```diff
@@ -0,0 +1,546 @@
...
+
+// This does not have an associated struct. It also does not make sense to
+// serialize.
+func formatProgramDependenciesAbout(language, root string) (string, error) {
+	var depInfo string
+	switch language {
+	case langNodejs:
+		depInfo = "package.json"
+	case langPython:
+		depInfo = "requirements.txt"
```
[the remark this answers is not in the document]
iwahbe: I can't think of any reason we don't run the command directly. It's a good idea and much simpler for the user. I'll implement it.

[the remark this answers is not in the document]
iwahbe: For python, we can directly access the interpreter buried in the virtual environment. Something like `./venv/bin/python3 -m list`.

pkg/cmd/pulumi/about_test.go
```diff
@@ -0,0 +1,66 @@
+// Copyright 2016-2021, Pulumi Corporation.
+//
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
```
justinvp, line 4: Nit: There's still an extra line in here:

```suggestion
// Copyright 2016-2021, Pulumi Corporation.
//
// Licensed under the Apache License, Version 2.0 (the "License");
```

tests/about_test.go
```diff
@@ -0,0 +1,62 @@
+// Copyright 2016-2021, Pulumi Corporation.
+//
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
```
justinvp, line 4: Nit: Delete extra line

```suggestion
// Copyright 2016-2021, Pulumi Corporation.
//
// Licensed under the Apache License, Version 2.0 (the "License");
```

pkg/cmd/pulumi/about.go
```diff
@@ -0,0 +1,923 @@
+// Copyright 2016-2021, Pulumi Corporation.
+//
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
```
justinvp, line 4: Nit: Delete extra line

```suggestion
// Copyright 2016-2021, Pulumi Corporation.
//
// Licensed under the Apache License, Version 2.0 (the "License");
```

pkg/cmd/pulumi/about.go
```diff
@@ -0,0 +1,923 @@
...
+
+	var result []byte
+	if result, err = cmd.Output(); err != nil {
+		return "", err
+	}
+	return string(result), nil
+}
+
+func getPythonProgramDependencies(proj *workspace.Project, rootDir string,
+	transative bool) ([]programDependencieAbout, error) {
```
justinvp, line 517: Nit: transative -> transitive (applies throughout)

pkg/cmd/pulumi/about.go
```diff
@@ -0,0 +1,923 @@
...
+			})
+		}
+	}
+	return packages, nil
+}
+
+func getNodeProgramDependencies(rootDir string, transitive bool) ([]programDependencieAbout, error) {
+	// Neither "yarn list" or "npm ls" can describe what packages are required
+	// (direct dependencies). Only what packages they have installed (transitive
+	// dependencies). This means that to accuratly report only direct
```
justinvp, line 596: Nit:

```suggestion
	// dependencies). This means that to accurately report only direct
```

justinvp (approved):
LGTM

PR #7821, opened 2021-08-23
Satisfy Python lint: factor out URN parsing in Python and other fixes

t0yv0:
# Description

Python LINT started flagging duplicate code in our codebase, including on master, so I'm starting to clean up. This PR just lifts URN parsing into a shared function.

Fixes # (issue)

## Checklist


justinvp (approved):
LGTM, thanks!

t0yv0:
I'll need to work through test failures and maybe add a few tests, but thanks. Not trying to change any behavior here.

PR #7825, opened 2021-08-23
Implement 5758 {fn}_output codgen for Python

### LINKED ISSUE #5758: Functions/datasources don't compose well with Pulumi model - should accept Outputs

t0yv0:
# Description

Fixes #5758  for Python

## Checklist


pkg/codegen/internal/test/testdata/output-funcs/py_tests/funcWithAllOptionalInputs.py
```diff
@@ -0,0 +1,63 @@
...
+    if opts.version is None:
+        opts.version = _utilities.get_version()
+    __ret__ = pulumi.runtime.invoke('madeup-package:codegentest:funcWithAllOptionalInputs', __args__, opts=opts, typ=FuncWithAllOptionalInputsResult).value
+
+    return AwaitableFuncWithAllOptionalInputsResult(
+        r=__ret__.r)
+
+
+@_utilities.lift_output_func(func_with_all_optional_inputs)
+def func_with_all_optional_inputs_output(a: Optional[pulumi.Input[Optional[str]]] = None,
```
justinvp, line 60: Should we include the docstring of the function?

pkg/codegen/internal/test/testdata/output-funcs/py_tests/getIntegrationRuntimeObjectMetadatum.py
```diff
@@ -0,0 +1,96 @@
...
+        opts.version = _utilities.get_version()
+    __ret__ = pulumi.runtime.invoke('azure-native:codegentest:getIntegrationRuntimeObjectMetadatum', __args__, opts=opts, typ=GetIntegrationRuntimeObjectMetadatumResult).value
+
+    return AwaitableGetIntegrationRuntimeObjectMetadatumResult(
+        next_link=__ret__.next_link,
+        value=__ret__.value)
+
+
+@_utilities.lift_output_func(get_integration_runtime_object_metadatum)
+def get_integration_runtime_object_metadatum_output(factory_name: Optional[pulumi.Input[str]] = None,
```
justinvp, line 91: Is the plan to have the args of the `_output` variant follow the exact same pattern as the source function (where all args are optional with `= None`, w/ pending runtime check to ensure required args are set) for consistency, or do we want to use the "improved" signature uses by methods and args classse, where we force args be specified by keyword (via a `*` arg, and sort required args first and annotate as non-optional, followed by optional args set to `= None`)? The former (what you have currently in this PR) sounds fine to me, but I was just wondering.

t0yv0: I see it's the same comment you had in https://github.com/pulumi/pulumi/pull/7825

I think we should opt for consistency here and consider switching to the improved signature in both versions of the function later, which would be a breaking change for consumers calling arguments positionally, right? 

I started trying to jot down an issue but I think it's best if you do it, I have some doubt if it is unconditionally a good idea. The '*' trick to require call sites to use keywords is a good idea in my book, in fact I'd require all methods to use keywords if it was up to me (like in Smalltalk) but what I have doubts about is sorting the required arguments. 

Sorting.. seems weird to me. Either the schema has order, in which case the schema authors can arrange the most aesthetically pleasing order of these positional arguments, or it does not have order. In case it does not have order, sorting makes arbitrary decisions that can feel ugly. If all args required keywords the caller could make the decision? 

Checking up how it works, perhaps this:

```
def foo_bar(*, a, b=None, c=None):
    pass

# foo_bar()
# TypeError: foo_bar() missing 1 required keyword-only argument: 'a

foo_bar(a=1)
```

Also thinking out loud, we might need the order in the schema at least for deciding in which order we document parameters. We could have data on which parameters are most frequently set to make them float up in the docs.

t0yv0: Also let me actually test if positionally passing arguments even works right now through the `lift_output_func` combinator.. I don't think I was careful enough to test this...

t0yv0: That didn't work in fact.. added some tests now and fixed the decorator to use `inspect` to bind correctly.

t0yv0:
Docstring type signatures are approximate (str) but they were also approximate before the change, I'd prefer to tackle that later when optimizing docstrings for IDEs.

t0yv0:
@justinvp PTAL

PR #7829, opened 2021-08-24
Auto dotnet plugin install server

t0yv0:
# Description

Same as https://github.com/pulumi/pulumi/pull/7796 just with master merged, trying to get the CI checks to work.

## Checklist


t0yv0:
```
  === FAIL: integration TestLargeResourceDotNet (unknown)

```
Believe this to be flaky.

PR #7834, opened 2021-08-25
Implemented `sha1()` support for Code Generator

T-Vova:
# Description
Implemented Terraform's `sha1()` function for all 4 languages, solving a total of 20 code generation errors in the AWS provider

### DIFF
--- a/pkg/codegen/dotnet/gen_program_expressions.go
+++ b/pkg/codegen/dotnet/gen_program_expressions.go
@@ -325,7 +325,7 @@ func (g *generator) GenFunctionCallExpression(w io.Writer, expr *model.FunctionC
 		g.Fgen(w, ")")
 	case "sha1":
 		// Assuming the existence of the following helper method located earlier in the preamble
-		g.Fgenf(w, "sha1(%v)", expr.Args[0])
+		g.Fgenf(w, "ComputeSHA1(%v)", expr.Args[0])
 	default:
 		g.genNYI(w, "call %v", expr.Name)
 	}
--- a/pkg/codegen/dotnet/gen_program_expressions.go
+++ b/pkg/codegen/dotnet/gen_program_expressions.go
@@ -242,6 +242,7 @@ var functionNamespaces = map[string][]string{
 	"readFile": {"System.IO"},
 	"toJSON":   {"System.Text.Json", "System.Collections.Generic"},
 	"toBase64": {"System"},
+	"sha1":     {"System.Security.Cryptography", "System.Text"},
 }
 
 func (g *generator) genFunctionUsings(x *model.FunctionCallExpression) []string {
@@ -322,6 +323,9 @@ func (g *generator) GenFunctionCallExpression(w io.Writer, expr *model.FunctionC
 		g.Fgen(w, "JsonSerializer.Serialize(")
 		g.genDictionary(w, expr.Args[0])
 		g.Fgen(w, ")")
+	case "sha1":
+		// Assuming the existence of the following helper method located earlier in the preamble
+		g.Fgenf(w, "sha1(%v)", expr.Args[0])
 	default:
 		g.genNYI(w, "call %v", expr.Name)
 	}
--- a/pkg/codegen/dotnet/utilities.go
+++ b/pkg/codegen/dotnet/utilities.go
@@ -129,10 +129,10 @@ func getHelperMethodIfNeeded(functionName string) (string, bool) {
 	switch functionName {
 	case "sha1":
 		methodBody =
-			`private static string sha1(string input) {
+			`private static string ComputeSHA1(string input) {
 		return BitConverter.ToString(
-			new SHA1Managed().ComputeHash(Encoding.UTF8.GetBytes(%v))
-		).Replace("-","").ToLower());
+			SHA1.Create().ComputeHash(Encoding.UTF8.GetBytes(%v))
+		).Replace("-","").ToLowerInvariant());
 	}`
 	default:
 		methodBody = ""
--- a/pkg/codegen/dotnet/utilities.go
+++ b/pkg/codegen/dotnet/utilities.go
@@ -131,7 +131,7 @@ func getHelperMethodIfNeeded(functionName string) (string, bool) {
 		methodBody =
 			`private static string ComputeSHA1(string input) {
 		return BitConverter.ToString(
-			SHA1.Create().ComputeHash(Encoding.UTF8.GetBytes(%v))
+			SHA1.Create().ComputeHash(Encoding.UTF8.GetBytes(input))
 		).Replace("-","").ToLowerInvariant());
 	}`
 	default:
--- a/pkg/codegen/dotnet/utilities.go
+++ b/pkg/codegen/dotnet/utilities.go
@@ -15,10 +15,11 @@
 package dotnet
 
 import (
-	"github.com/pulumi/pulumi/pkg/v3/codegen"
 	"regexp"
 	"strings"
 	"unicode"
+
+	"github.com/pulumi/pulumi/pkg/v3/codegen"
 
 	"github.com/pkg/errors"
 )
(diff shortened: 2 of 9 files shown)


pkg/codegen/dotnet/utilities.go
```diff
@@ -118,3 +119,23 @@ func makeSafeEnumName(name, typeName string) (string, error) {
...
+// encapsulate them as their own function in the preamble.
+func getHelperMethodIfNeeded(functionName string) (string, bool) {
+	var methodBody string
+
+	switch functionName {
+	case "sha1":
+		methodBody =
+			`private static string sha1(string input) {
+		return BitConverter.ToString(
+			new SHA1Managed().ComputeHash(Encoding.UTF8.GetBytes(%v))
```
justinvp, line 134: The latest .NET version marks `SHA1Managed` obsolete and says to just use `SHA1.Create`:

https://github.com/dotnet/runtime/blob/3633a7d0930abaca701385c1059b80ca157e98c6/src/libraries/System.Security.Cryptography.Algorithms/src/System/Security/Cryptography/SHA1Managed.cs#L9

https://github.com/dotnet/runtime/blob/35cf750ea50f13f215e9eb6d3ecb6dea48b081d1/src/libraries/Common/src/System/Obsoletions.cs#L75

Let's just go ahead and do that here.

```suggestion
			SHA1.Create().ComputeHash(Encoding.UTF8.GetBytes(%v))
```

pkg/codegen/dotnet/utilities.go
```diff
@@ -118,3 +119,23 @@ func makeSafeEnumName(name, typeName string) (string, error) {
...
+// Provides code for a method which will be placed in the program preamble if deemed
+// necessary. Because many Terraform functions are complex, it is much prettier to
+// encapsulate them as their own function in the preamble.
+func getHelperMethodIfNeeded(functionName string) (string, bool) {
+	var methodBody string
+
+	switch functionName {
+	case "sha1":
+		methodBody =
+			`private static string sha1(string input) {
```
justinvp, line 132: The [convention](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions) in .NET is to use PascalCasing for methods. Having it all lowercase feels out of place. We can't use `SHA1` as that conflicts with the type in the `System.Security.Cryptography` namespace. How about naming it `ComputeSHA1`?

```suggestion
			`private static string ComputeSHA1(string input) {
```

pkg/codegen/dotnet/utilities.go
```diff
@@ -118,3 +119,23 @@ func makeSafeEnumName(name, typeName string) (string, error) {
...
+		return BitConverter.ToString(
+			new SHA1Managed().ComputeHash(Encoding.UTF8.GetBytes(%v))
+		).Replace("-","").ToLower());
```
justinvp, line 135: `ToLower()` uses the casing rules for the current culture. We can avoid that by using `ToLowerInvariant()`.

```suggestion
		).Replace("-","").ToLowerInvariant());
```

justinvp (commented):
Sorry for the "drive by" review, but noticed the PR and took a quick look and had some comments on the .NET implementation.

T-Vova:
> Sorry for the "drive by" review, but noticed the PR and took a quick look and had some comments on the .NET implementation.
No problem. Thanks for the review!

pkg/codegen/dotnet/gen_program_expressions.go
```diff
@@ -322,6 +323,9 @@ func (g *generator) GenFunctionCallExpression(w io.Writer, expr *model.FunctionC
 		g.Fgen(w, "JsonSerializer.Serialize(")
 		g.genDictionary(w, expr.Args[0])
 		g.Fgen(w, ")")
+	case "sha1":
+		// Assuming the existence of the following helper method located earlier in the preamble
+		g.Fgenf(w, "sha1(%v)", expr.Args[0])
```
justinvp, line 328: This will need to be changed as well:

```suggestion
		g.Fgenf(w, "ComputeSHA1(%v)", expr.Args[0])
```

pkg/codegen/dotnet/utilities.go
```diff
@@ -118,3 +119,23 @@ func makeSafeEnumName(name, typeName string) (string, error) {
...
+// encapsulate them as their own function in the preamble.
+func getHelperMethodIfNeeded(functionName string) (string, bool) {
+	var methodBody string
+
+	switch functionName {
+	case "sha1":
+		methodBody =
+			`private static string ComputeSHA1(string input) {
+		return BitConverter.ToString(
+			SHA1.Create().ComputeHash(Encoding.UTF8.GetBytes(%v))
```
lukehoban, line 134: Is the `%v` intentional here?  It looks like this should be referencing `input`?

pkg/codegen/go/gen_program_utils.go
```diff
@@ -54,3 +54,32 @@ func (p *promptToInputArrayHelper) getPromptItemType() string {
...
+				} else {
+					panic(err.Error())
+				}
+			}`
+	case "sha1":
+		methodBody =
+			`func sha1(input string) pulumi.StringPtrInput {
+				inputBytes := []byte(input)
+				return pulumi.String(fmt.Sprintf("%x", sha1.Sum(inputBytes)))
+			}`
```
lukehoban, line 80: Interestingly, this one requires no error handling, so could have just been inlined. But most likely nicer to read the way you have structured this.

pkg/codegen/nodejs/gen_program.go
```diff
@@ -179,7 +179,7 @@ func (g *generator) genPreamble(w io.Writer, program *hcl2.Program) {
 			continue
 		}
 		as := makeValidIdentifier(path.Base(pkg))
-		if as != pkg {
+		if as != pkg || pkg == "crypto" {
```
lukehoban, line 182: Why is this needed?

lukehoban (commented):
Will be great to rebase this after merging the other PRs it build on top of, but other than the note about the .NET helper, these changes look great.

pkg/codegen/nodejs/gen_program.go
T-Vova: `import * from "crypto"` doesn't compile properly when the `ComputeSHA1` method is added, and so I force it to use `import * as crypto from "crypto"` instead.

pkg/codegen/dotnet/utilities.go
T-Vova: Done.

pkg/codegen/go/gen_program_utils.go
T-Vova: I thought the same. If the inlined method ends up being too long, moving it into a helper is preferable for the users of the example.

pkg/codegen/go/gen_program_utils.go
```diff
@@ -54,3 +54,32 @@ func (p *promptToInputArrayHelper) getPromptItemType() string {
...
+					return pulumi.String(string(fileData[:]))
+				} else {
+					panic(err.Error())
+				}
+			}`
+	case "sha1":
+		methodBody =
+			`func sha1(input string) pulumi.StringPtrInput {
+				inputBytes := []byte(input)
+				return pulumi.String(fmt.Sprintf("%x", sha1.Sum(inputBytes)))
```
pgavlin, line 79: I don't think that this does what you want it to do:
- This won't compile, since `sha1` in `sha1.Sum` binds to `func sha1`. See https://play.golang.org/p/2ZCM_uilVdp for an example of the error here.
- `fmt.Sprintf("%x", someBytes)` won't format the bytes as a hex string--you need to use [`encoding/hex.EncodeToString](https://pkg.go.dev/encoding/hex#EncodeToString) to do that.

The `sha1Hash` function in this code should do what you want: https://play.golang.org/p/GF-55iRvw00

pgavlin (changes requested):

pgavlin (approved):
LGTM, though we should land https://github.com/pulumi/pulumi/pull/7806 first and then rebase this on top.

stack72:
Merging this as this is a flaky test

PR #7838, opened 2021-08-25
iwahbe/1515/remove opt pulumi from the build system

### LINKED ISSUE #1515: Remove the need for `/opt/pulumi` when hacking on Pulumi
When we added `/opt/pulumi` we had good reasons.  The CLI had a tight dependency on the layout of an install (`bin` and `node_modules` needed to be next to one another). Since then, we've removed these dependencies, but the wart of having a different folder we install to has lingered.

Our split of `build` and `install` targets makes things even worse because you can end up with versions of `pulumi` as well as the resource providers and language providers in both `GOBIN` and `/opt/pulumi` which an lead to some gnarly issues depending on how you've set up your `$PATH`.

We should be able to completely remove `/opt/pulumi` in favor of just installing everything into `GOBIN`.

At a high level we should do the following:

1. Remove the need for `/opt/pulumi/bin` by collapsing the build and install targets together, and just always install into `GOBIN` via (`go install`).  We'll need to place a few small additional support scripts (e.g. the `-exec` scripts some language hosts use) into `GOBIN` as well, but that doesn't seem like the end of the world.

2. Remove the need for `/opt/pulumi/node_modules`. As part of the above, when we collapse build and install into the same target, we should stop installing NPM packages into `/opt/pulumi/node_modules` in favor of just `yarn linking` the version in the bin folder. This is much more in line with how other projects handle the issue.

iwahbe:
# Description

Fixes #1515

Move:
- [X] `/opt/pulumi/bin`
- [X] `/opt/pulumi/node_modules`
- [x] `/opt/pulumi/nuget`

## Checklist

  + If this passes the tests that already exist, then it works. 


iwahbe:
/run-acceptence/-tests

justinvp (approved):
LGTM

@stack72, off the top of your head, is there anything you can think of that might potentially break that we'll need to fix-up as part of this change? Right now PR jobs are passing.

.github/workflows/master.yml
```diff
@@ -261,23 +261,17 @@ jobs:
         run: |
           make dist
         env:
-          PULUMI_NODE_MODULES: ${{ runner.temp }}/opt/pulumi/node_modules
```
stack72, line 264: Is there any reason why we need to remove this and not update it?

.github/workflows/master.yml
```diff
@@ -261,23 +261,17 @@ jobs:
...
           PULUMI_LOCAL_NUGET: ${{ runner.temp }}/opt/pulumi/nuget
-          PULUMI_ROOT: ${{ runner.temp }}/opt/pulumi
```
stack72, line 266: Same question - is this not needed now?

.gitpod/Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM gitpod/workspace-full
 USER gitpod
 
-ENV PATH="/opt/pulumi:/opt/pulumi/bin:$PATH"
+ENV PATH="/opt/pulumi:$PATH"
```
stack72, line 4: this one still seems to include /opt/pulumi

CHANGELOG_PENDING.md
```diff
@@ -16,6 +16,9 @@
 - [codegen/schema] Add a `pulumi schema check` command to validate package schemas.
   [#7865](https://github.com/pulumi/pulumi/pull/7865)
 
+- [build] - Remove the need for `/opt/pulumi` when building.
```
stack72, line 19: I don't think this is changelog worthy

build/common.mk
```diff
@@ -168,9 +167,8 @@ test_fast::
 
 install::
 	$(call STEP_MESSAGE)
-	@mkdir -p $(PULUMI_BIN)
-	@mkdir -p $(PULUMI_NODE_MODULES)
-	@mkdir -p $(PULUMI_NUGET)
+	@mkdir -p $(GOBIN)
```
stack72, line 170: I think we should check that this doesn't exist - we don't want to kill anyones GOBIN

build/common.mk
```diff
@@ -183,12 +181,7 @@ test_all::
...
 ifneq ($(NODE_MODULE_NAME),)
 install::
-	[ ! -e "$(PULUMI_NODE_MODULES)/$(NODE_MODULE_NAME)" ] || rm -rf "$(PULUMI_NODE_MODULES)/$(NODE_MODULE_NAME)"
-	mkdir -p "$(PULUMI_NODE_MODULES)/$(NODE_MODULE_NAME)"
-	cp -r bin/. "$(PULUMI_NODE_MODULES)/$(NODE_MODULE_NAME)"
-	cp yarn.lock "$(PULUMI_NODE_MODULES)/$(NODE_MODULE_NAME)"
-	rm -rf "$(PULUMI_NODE_MODULES)/$(NODE_MODULE_NAME)/node_modules"
-	cd "$(PULUMI_NODE_MODULES)/$(NODE_MODULE_NAME)" && \
+	cd bin && \
 	yarn install --prefer-offline --production && \
```
stack72, line 185: so where will this install to now?

build/tmp_dir.go
```diff
@@ -0,0 +1,10 @@
+package main
+
+import (
```
stack72, line 3: what is this for?

stack72 (commented):
I've left a few questions inline and @justinvp also had a question around makefile targets

I am presuming you were able to run this end to end?

build/common.mk
iwahbe: I'm pretty sure that when `mkdir` is called on a folder that already exists, nothing happens. That is certainly true on macOS.

stack72: 🆗

stack72:
@iwahbe there are still a few questions above - can we see if theres a resolution so we can shore this work up? :)

build/common.mk
iwahbe: The install is in place.

.github/workflows/master.yml
iwahbe: It shouldn't be needed.

.github/workflows/master.yml
iwahbe: Because we install our node_modules where we build them and then link. We don't put them in an external directory. `PULUMI_NODE_MODULES` is no longer used.

build/tmp_dir.go
iwahbe: This behavior has changed after clarification from @justinvp, this file has been deleted and we just put our nuget files in  `sdk/dotnet/.nuget`.

iwahbe:
@stack72 I think I have answered your questions. Let me know if that is not the case.

iwahbe:
Closing in favor of  https://github.com/pulumi/pulumi/pull/8437.

PR #7844, opened 2021-08-25
"make lint" returns an accurate status code

iwahbe:
# Description

This is mainly to allow `make lint` to be usable in git hooks. 

## Checklist


emiliza (approved):
Thanks for fixing this! Works as expected.

PR #7852, opened 2021-08-26
[engine] Include acyclic children in delete deps.

### LINKED ISSUE #7833: Handle component child / component dependent registration races when expanding deletion deps
See https://github.com/pulumi/pulumi/pull/7791#issuecomment-902915872 for details.

pgavlin:
Consider the following Pulumi program:

```typescript
class MyResource extends pulumi.CustomResource {
    // ...
}

class MyComponent extends pulumi.ComponentResource {
    constructor(name: string) {
        super("myComponent", name);

        this.child = new MyResource("child", {}, {parent: this});
    }
}

// Create a component resource. The constructor returns promptly, but
// the registration of the component itself happens asynchronously.
// Similarly, its child resource's constructor returns promptly, but
// its registration happens asynchronously. The SDK guarantees that the
// latter happens after the former by waiting on the parent's URN prior
// to registration (note that this is necessary in order for the runtime
// to fill out the `parent` field of the registration request).
const component = new MyComponent("component");

// Create a resource that depends on the component resource. The
// registration of this resource happens asynchronously, and races with
// the asynchronous registration of the component's child.
const res = new MyResource("resource", {}, {dependsOn: component});
```

Due to the races described in the comments, even when all of a
component's children are constructed prior to exiting the component's
constructor (e.g. not in applies, thens, etc.) we cannot guarantee that
the children have been registered prior to any resources that depend on
the component. Because of this, the approach that #7788 takes for
expanding dependencies on components to dependencies on their children
is not necessarily reliable, as it only adds children that appear prior
to dependents in the snapshot.

These changes adjust the dependency graph to add all children of a
component that would not create cycles in the dependency graph.

Fixes #7833.

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -3,3 +3,6 @@
 
 ### Bug Fixes
 
+- [engine] Include acyclic children in dependency list for deletes.
+  [#7852](https://github.com/pulumi/pulumi/pull/7852)
+
--- a/pkg/engine/lifeycletest/pulumi_test.go
+++ b/pkg/engine/lifeycletest/pulumi_test.go
@@ -35,6 +35,7 @@ import (
 	. "github.com/pulumi/pulumi/pkg/v3/engine"
 	"github.com/pulumi/pulumi/pkg/v3/resource/deploy"
 	"github.com/pulumi/pulumi/pkg/v3/resource/deploy/deploytest"
+	"github.com/pulumi/pulumi/pkg/v3/resource/deploy/providers"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/diag"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/diag/colors"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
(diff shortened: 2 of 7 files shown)


t0yv0:
Making progress on verifying this. I've verified `DependingOn` and that `DependingOn` is the inverse transitive closure of `DependsOn`.. However it appears that the order of nodes returned by `DependingOn` is not observing the order of dependencies in the list it returns, counterexample:


```
[nix-shell:~/pulumi/pkg/resource/graph]$ go test -run  TestRapidDependingOnOrdered -rapid.checks=100
go test -run  TestRapidDependingOnOrdered -rapid.checks=100
--- FAIL: TestRapidDependingOnOrdered (0.18s)
    dependency_graph_rapid_test.go:372: [rapid] failed after 4 tests: (*T).FailNow() called
        To reproduce, specify -run="TestRapidDependingOnOrdered" -rapid.failfile="TestRapidDependingOnOrdered-20210830143011-33973.fail" (or -rapid.seed=9040361521867128843)
        Failed test output:
    dependency_graph_rapid_test.go:373: [rapid] draw universe: []*resource.State{(*resource.State)(0xc000cc6000), (*resource.State)(0xc000cc6120), (*resource.State)(0xc000cc6240)}
    dependency_graph_rapid_test.go:374: Checking universe: [
        
        urn:pulumi:a::b::c:d:e::a1
        
        
        urn:pulumi:a::b::c:d:e::a2 parent=urn:pulumi:a::b::c:d:e::a1 deps=[urn:pulumi:a::b::c:d:e::a1, ]
        
        
        urn:pulumi:a::b::c:d:e::a3 parent=urn:pulumi:a::b::c:d:e::a1
        
        
        ]
    dependency_graph_rapid_test.go:189: transitiveDependenciesOf(urn:pulumi:a::b::c:d:e::a2, urn:pulumi:a::b::c:d:e::a1)
    dependency_graph_rapid_test.go:189: transitiveDependenciesOf(urn:pulumi:a::b::c:d:e::a2, urn:pulumi:a::b::c:d:e::a3)
    dependency_graph_rapid_test.go:189: transitiveDependenciesOf(urn:pulumi:a::b::c:d:e::a3, urn:pulumi:a::b::c:d:e::a1)
    dependency_graph_rapid_test.go:196: Inspecting DependingOn(urn:pulumi:a::b::c:d:e::a1) = [
        
        urn:pulumi:a::b::c:d:e::a2 parent=urn:pulumi:a::b::c:d:e::a1 deps=[urn:pulumi:a::b::c:d:e::a1, ]
        
        
        urn:pulumi:a::b::c:d:e::a3 parent=urn:pulumi:a::b::c:d:e::a1
        
        
        ]
    dependency_graph_rapid_test.go:200: 
        	Error Trace:	dependency_graph_rapid_test.go:200
        	            				dependency_graph_rapid_test.go:375
        	            				engine.go:276
        	            				engine.go:158
        	            				engine.go:94
        	            				dependency_graph_rapid_test.go:372
        	            				dependency_graph_rapid_test.go:179
        	Error:      	Should be true
        	Test:       	TestRapidDependingOnOrdered
        	Messages:   	urn:pulumi:a::b::c:d:e::a3 should appear before urn:pulumi:a::b::c:d:e::a2
FAIL
exit status 1
FAIL	github.com/pulumi/pulumi/pkg/v3/resource/graph	0.418s
```

lunaris:
After discussing with @pgavlin, I think we can close this for now in light of https://github.com/pulumi/pulumi/issues/7833#issuecomment-917076458

PR #7855, opened 2021-08-26
5758 fix for .NET

### LINKED ISSUE #5758: Functions/datasources don't compose well with Pulumi model - should accept Outputs

t0yv0:
# Description

Fixes #5758  for C#/.NET

- [ ] make sure the go test runs dotnet tests over the generated code
- [ ] make sure the generated code links against in-source Pulumi .NET SDK
- [ ] note the difficulty mocking functions, create an issue somewhere

## Checklist


t0yv0:
Close in favor of https://github.com/pulumi/pulumi/pull/7899

Issue #7623, opened 2021-07-23
Invalid Go code is generated in pulumi-aws


stack72:
This is not a new issue but we have not been building the generated go sdk for the pulumi-aws provider. After removing all of the go sdk and letting it be regenerated, we are actually generating invalid code:

```
# github.com/pulumi/pulumi-aws/sdk/v4/go/aws/kinesisanalyticsv2
go/aws/kinesisanalyticsv2/pulumiTypes.go:2388:43: impossible type assertion:
	ApplicationApplicationConfigurationSqlApplicationConfigurationOutput does not implement pulumi.Output (missing ApplyT method)
go/aws/kinesisanalyticsv2/pulumiTypes.go:2396:43: impossible type assertion:
	ApplicationApplicationConfigurationSqlApplicationConfigurationOutput does not implement pulumi.Output (missing ApplyT method)
go/aws/kinesisanalyticsv2/pulumiTypes.go:2459:11: cannot use v.Input (type *ApplicationApplicationConfigurationSqlApplicationConfigurationInput) as type *ApplicationApplicationConfigurationSqlApplicationConfigurationInput in return argument:
	*ApplicationApplicationConfigurationSqlApplicationConfigurationInput is pointer to interface, not interface
go/aws/kinesisanalyticsv2/pulumiTypes.go:2466:11: cannot use v.Outputs (type []ApplicationApplicationConfigurationSqlApplicationConfigurationOutput) as type []ApplicationApplicationConfigurationSqlApplicationConfigurationOutput in return argument
go/aws/kinesisanalyticsv2/pulumiTypes.go:2503:11: cannot use v.Input (type *ApplicationApplicationConfigurationSqlApplicationConfigurationInput) as type *ApplicationApplicationConfigurationSqlApplicationConfigurationInput in return argument:
	*ApplicationApplicationConfigurationSqlApplicationConfigurationInput is pointer to interface, not interface
go/aws/kinesisanalyticsv2/pulumiTypes.go:2513:11: cannot use v.Outputs (type []ApplicationApplicationConfigurationSqlApplicationConfigurationOutput) as type []ApplicationApplicationConfigurationSqlApplicationConfigurationOutput in return argument
go/aws/kinesisanalyticsv2/pulumiTypes.go:2527:6: ApplicationApplicationConfigurationSqlApplicationConfigurationInput redeclared in this block
	previous declaration at go/aws/kinesisanalyticsv2/pulumiTypes.go:2363:6
[...]

PR #7857, opened 2021-08-27
[codegen/go] detect and avoid collisions in resource and type code generation

EvanBoyle:
# Description
This change detects and avoids colliding type declarations as each type and resource actually generates multiple declarations in our Go type system (Input, Output, Args, etc.). 

Two types or resources named `Type` and `TypeInput` generate overlapping types in our go SDK. In this example, we'd get `type TypeInput interface` generated from the first resource, and `type TypeInput struct`. These overlapping types cause a compilation error due to duplicate identifiers. 

This change computes and stores the set of names for resources and types ahead of generation. If it detects a conflict, it attemps to resolve it via suffixing. If unable to resolve it, it panics as we're unable to generate correct code. 

This is just a first pass at this to unblock AWS and GCP. 

Fixes https://github.com/pulumi/pulumi/issues/7623
Fixes https://github.com/pulumi/pulumi/issues/7622 (half of it at)

The key question I have is around the appending of `Type` when we have duplicate type tokens. Duplicate type tokens are not supported under this change. In practice, I don't see this causing diffs in the codegen below. @pgavlin @justinvp thoughts on this? 

## Checklist


t0yv0:
Following the discussion from last Friday and https://github.com/pulumi/pulumi/pull/7865 I am now wondering whether 7865 includes a schema check that would report these collisions - and I think no, it does not. So one possible follow up item is to add those checks into "schema check" tool. However this is still a meta-problem, that won't fix the actual providers. LMK if I can personally help with any of the work here.

EvanBoyle:
Ok, @pgavlin and I talked and he showed me some concrete places (in kube where the downstream codegen tests were broken last week) where we depend on duplicate types. I put together  https://github.com/pulumi/pulumi/commit/4311a024cd5d120623936cb8ad8987149a2a5d22 to add back support for duplicate tokens across types and resources. Kubernetes diffs are now clean. 

I still don't have ideas or answers for Luke's questions on preventing accidental changes when types are added or removed.

t0yv0:
@EvanBoyle for my 2c, I'd vote to take this change if it fixes the immediate pressing issue. Re Luke's comment on stability over time, it feels important but we do not have evidence yet of it biting in practice so perhaps something more strategic that we'd better look when someone has runway to take a long holistic look at codegen? I'd personally not block on it.

Now I don't quite follow the discussion on duplicate type/resource tokens, this is not making sense to me at the moment, which might be okay for now...

EvanBoyle:
@t0yv0 I'm inclined to do do the same. 

To elaborate on the duplicate token issue. Right now it is possible to have two identical tokens in the same package. One for a resource and one for a type:

https://github.com/pulumi/pulumi/blob/master/pkg/codegen/internal/test/testdata/resource-args-python/schema.json#L5-L15

For some languages this is fine because the types get generated into separate namespaces, but for go we have some kinda hacky logic that detects this and adds a `Type` suffix to the type. In practice we use this for things like `Pod` and `MutatingWebHook` in our kubernetes provider (I do not have background on why).

The collision detection logic works by computing the expected permutations based on resource and type tokens in the schema and looking for collisions. Duplicate type tokens kinda messes up this logic, so I had to add special case handling that specifically looks for duplicate type tokens, adds them to a map, and then uses the original renaming logic for those tokens.

pkg/codegen/go/gen.go
```diff
@@ -2513,6 +2563,73 @@ func generatePackageContextMap(tool string, pkg *schema.Package, goInfo GoPackag
...
+				if suffix != "" {
+					pkg.renamed[originalNames[i]] = names[i]
+				}
+			}
+		default:
+			return
+		}
+	}
+
+	for _, t := range pkg.Types {
```
t0yv0, line 2629: Is this deterministic or we need a sort here --^

pkg/codegen/go/gen.go
```diff
@@ -1295,6 +1293,9 @@ func (pkg *pkgContext) getDefaultValue(dv *schema.DefaultValue, t schema.Type) (
 
 func (pkg *pkgContext) genResource(w io.Writer, r *schema.Resource, generateResourceContainerTypes bool) error {
 	name := resourceName(r)
```
t0yv0, line 1295: This reads weird and will confuse me when working with this code later. Does the place that calls `resourceName` func in `genResourceModule` also needs the renamed treatment? The renamed treatment could also be packaged into some func accepting pkg and `*schema.Resource`.

I'm noticing `pkg.renamed` as present before, but not used here.. Rename resourceName to `resourceNameBeforeNameDisambiguation`?

pkg/codegen/go/gen.go
```diff
@@ -2513,6 +2563,73 @@ func generatePackageContextMap(tool string, pkg *schema.Package, goInfo GoPackag
 		scanResource(r)
 	}
 
+	// compute set of names generated by a type
+	// handling any potential collisions via remapping along the way
+	scanType := func(t schema.Type) {
+		switch t := t.(type) {
+		case *schema.ObjectType:
```
t0yv0, line 2570: No other types allocate names except ObjectType? Enums don't?

t0yv0 (approved):
Evan you've requested a review so it basically LGTM I'm happy we're addressing a pressing issue. I've left a few comments around lines I'm unsure about. In very unsure about codegen validity in general and the kind of 'by-exemplar' testing included likely does not uncover corner cases; it's likely we will have to revisit codegen heavily, including these changes introduced here. Since the large uncertainty I'd welcome a second reviewer here.

pkg/codegen/go/gen.go
EvanBoyle: Unfortunately, like duplicate type names we also have some bespoke logic (and tests) in place for enums:

https://github.com/pulumi/pulumi/blob/4311a024cd5d120623936cb8ad8987149a2a5d22/pkg/codegen/go/gen.go#L1951

This change doesn't touch that, and is a targeted fix for the issues we're seeing in aws/gcp. Agree that we will likely need to cover more cases in the future, but will leave that as a future exercise as bug reports come in.

pkg/codegen/go/gen.go
EvanBoyle: The schema binder handles sorting resource and type arrays before we get to this point: https://github.com/pulumi/pulumi/blob/76ee1b8ccfee815eb315d9e0e0ddfaaf505c472b/pkg/codegen/schema/schema.go#L1506

pkg/codegen/go/gen.go
EvanBoyle: Good catch, refactored and updated the tests.

EvanBoyle:
unrelated dotnet test failure, going to merge.

Issue #7434, opened 2021-07-06
Deeper fix of Go SDK to be compatible with changes in #7059


emiliza:
Based on @pgavlin's investigation from https://github.com/pulumi/pulumi/issues/7359 (with relation to the changes in #7059)

### M61/62
- [x] https://github.com/pulumi/pulumi/issues/7846 (Merged)
PR: https://github.com/pulumi/pulumi/pull/7861
- [x] https://github.com/pulumi/pulumi/issues/7847 (Merged)
PR: https://github.com/pulumi/pulumi/pull/7861
- [x] https://github.com/pulumi/pulumi/issues/7845 (Merged)
PR: https://github.com/pulumi/pulumi/pull/7870
- [x] https://github.com/pulumi/pulumi/issues/7848 (Merged) 
PR: https://github.com/pulumi/pulumi/pull/7928
PR: https://github.com/pulumi/pulumi/pull/7975
- [x] https://github.com/pulumi/pulumi/issues/7849 (Merged)
PR: https://github.com/pulumi/pulumi/pull/7958
- [x] Aiming for dev build to validate above ~9/15 (pushed from 9/1)
- [x] https://github.com/pulumi/pulumi/issues/7850 (Merged)
PR: https://github.com/pulumi/pulumi/pull/7925
- [x] https://github.com/pulumi/pulumi/issues/7851 (Merged)
PR: https://github.com/pulumi/pulumi/pull/7926
- [x] https://github.com/pulumi/pulumi/issues/8007 (Merged)

### M63
- [x] https://github.com/pulumi/pulumi/pull/7959
- [x] https://github.com/pulumi/pulumi/issues/8165
- [x] https://github.com/pulumi/pulumi/issues/8166
- [x] https://github.com/pulumi/pulumi/issues/8189

### M64
- [x] https://github.com/pulumi/pulumi/issues/8006
- [x] https://github.com/pulumi/pulumi/issues/8155
- [x] https://github.com/pulumi/pulumi/issues/8156

### Defer for TBD
-  .NET SDK unmarshaling (used in MLC provider)

t0yv0:
What does it mean for an output value to be marshaled? Does marshaling always await the output value until it resolves to a concrete value and set of dependencies and then passes that over the boundary? E.g. in Haskell notation this is what we're actually passing?

```
data ResolvedOutput a = Unknown | Output IsSecret a [Dependency]
```

[...]

PR #7861, opened 2021-08-28
Initial support for (un)marshaling output values

### LINKED ISSUE #7846: Expand the definition of `resource.Output` in the Go SDK
Pre-requisite for: #7434

Expand the definition of `resource.Output` in the Go SDK to additional information about the output, i.e. dependencies and secretness

### LINKED ISSUE #7847: Add support in the core Go RPC code for marshaling and unmarshaling output values
Part of https://github.com/pulumi/pulumi/issues/7434

 Add support in the core Go RPC code for marshaling and unmarshaling output values
- Output values should be marshaled as special objects ala archives, assets, and resource refs
- Output values should be unmarshaled as `resource.Output` values

justinvp:
This change expands the definition of `resource.Output` in the Go SDK with additional information about the output, i.e. dependencies and secretness, and adds support in the core Go RPC code for (un)marshaling output values.

Output values are marshaled as special objects ala archives, assets, and resource refs and are unmarshaled as `resource.Output` values.

Note: The current plan is to scope this to only be used when marshaling inputs to a multi-lang component and for method calls using `Call`.

Subsequent PRs will add:
 - A monitor feature for output values, which will initially be disabled by default but available to turn on via an envvar
 - A way for component providers to indicate support for accepting output values (for `Construct` and `Call`)
 - Support for (un)marshaling output values in each language SDKs
 - E2E tests
 - Turn the monitor feature on by default (w/ env var to disable)

Fixes #7846
Fixes #7847

Part of #7434

### DIFF
--- a/sdk/go/common/resource/plugin/rpc.go
+++ b/sdk/go/common/resource/plugin/rpc.go
@@ -72,7 +72,7 @@ func MarshalProperties(props resource.PropertyMap, opts MarshalOptions) (*struct
 	for _, key := range props.StableKeys() {
 		v := props[key]
 		logging.V(9).Infof("Marshaling property for RPC[%s]: %s=%v", opts.Label, key, v)
-		if v.IsOutput() && !opts.KeepOutputValues {
+		if v.IsOutput() && !v.OutputValue().Known && !opts.KeepOutputValues {
 			logging.V(9).Infof("Skipping output property for RPC[%s]: %v", opts.Label, key)
 		} else if opts.SkipNulls && v.IsNull() {
 			logging.V(9).Infof("Skipping null property for RPC[%s]: %s (as requested)", opts.Label, key)
--- a/sdk/go/common/resource/plugin/rpc.go
+++ b/sdk/go/common/resource/plugin/rpc.go
@@ -152,15 +152,15 @@ func MarshalPropertyValue(key resource.PropertyKey, v resource.PropertyValue,
 		return nil, nil // return nil and the caller will ignore it.
 	} else if v.IsOutput() {
 		if !opts.KeepOutputValues {
+			result := v.OutputValue().Element
 			if !v.OutputValue().Known {
 				// Unknown outputs are marshaled the same as Computed.
-				computed := resource.MakeComputed(v.OutputValue().Element)
-				return MarshalPropertyValue(computed, opts)
-			} else if v.OutputValue().Secret {
-				secret := resource.MakeSecret(v.OutputValue().Element)
-				return MarshalPropertyValue(secret, opts)
+				result = resource.MakeComputed(result)
 			}
-			return MarshalPropertyValue(v.OutputValue().Element, opts)
+			if v.OutputValue().Secret {
+				result = resource.MakeSecret(result)
+			}
+			return MarshalPropertyValue(result, opts)
 		}
 		obj := resource.PropertyMap{
 			resource.SigKey: resource.NewStringProperty(resource.OutputValueSig),
@@ -456,15 +456,19 @@ func UnmarshalPropertyValue(key resource.PropertyKey, v *structpb.Value,
 			}
 
 			if !opts.KeepOutputValues {
+				result := &value
 				if !known {
-					v := structpb.Value{
+					result, err = UnmarshalPropertyValue(&structpb.Value{
 						Kind: &structpb.Value_StringValue{StringValue: UnknownStringValue},
+					}, opts)
+					if err != nil {
+						return nil, err
 					}
-					return UnmarshalPropertyValue(&v, opts)
-				} else if secret {
-					return unmarshalSecretPropertyValue(value, opts), nil
 				}
-				return &value, nil
+				if secret && result != nil {
+					result = unmarshalSecretPropertyValue(*result, opts)
+				}
+				return result, nil
 			}
 
 			var dependencies []resource.URN
--- a/sdk/go/common/resource/plugin/rpc.go
+++ b/sdk/go/common/resource/plugin/rpc.go
@@ -1,4 +1,4 @@
-// Copyright 2016-2018, Pulumi Corporation.
+// Copyright 2016-2021, Pulumi Corporation.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -39,6 +39,7 @@ type MarshalOptions struct {
 	RejectAssets       bool   // true if we should return errors on Asset and Archive values.
 	KeepResources      bool   // true if we are keeping resoures (otherwise we return raw urn).
 	SkipInternalKeys   bool   // true to skip internal property keys (keys that start with "__") in the resulting map.
+	KeepOutputValues   bool   // true if we are keeping output values.
 }
 
 const (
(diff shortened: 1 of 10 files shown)


sdk/go/common/resource/plugin/rpc.go
```diff
@@ -148,12 +149,35 @@ func MarshalPropertyValue(v resource.PropertyValue, opts MarshalOptions) (*struc
...
+		if !opts.KeepOutputValues {
+			if !v.OutputValue().Known {
+				// Unknown outputs are marshaled the same as Computed.
+				computed := resource.MakeComputed(v.OutputValue().Element)
+				return MarshalPropertyValue(computed, opts)
+			} else if v.OutputValue().Secret {
+				secret := resource.MakeSecret(v.OutputValue().Element)
+				return MarshalPropertyValue(secret, opts)
+			}
+			return MarshalPropertyValue(v.OutputValue().Element, opts)
```
pgavlin, line 161: How do we marshal secret unknowns today? Do those even come up?

sdk/go/common/resource/properties.go
```diff
@@ -84,7 +84,10 @@ type Computed struct {
 // encountered, it means the resource has not yet been created, and so the output value is unavailable.  Note that an
 // output property is a special case of computed, but carries additional semantic meaning.
 type Output struct {
-	Element PropertyValue // the eventual value (type) of the output property.
+	Element      PropertyValue // the value of this output if it is resolved.
+	Known        bool          `json:"-"` // true if this output's value is known.
+	Secret       bool          `json:"-"` // true if this output's value is secret.
+	Dependencies []URN         `json:"-"` // the dependencies associated with this output.
```
pgavlin, line 90: The `json` tags are interesting--what's the idea here?

pgavlin (commented):
A+++ test coverage. A few questions.

pgavlin:
Including reified `Output` values--metadata and all--in resource inputs could subsume the information we serialize in several different places (both at the RPC layer and at the statefile layer). I'd love to see a design doc that covers the various pieces of this work and considers whether or not we could/should simplify things by replacing that info with reified outputs. Should we work together on this @justinvp ?

> Note: The current plan is to scope this to only be used when marshaling inputs to a multi-lang component and for method calls using Call.
> ...
> A way for component providers to indicate support for accepting output values (for Construct and Call)

I'd suggest that we expand the scope of this from "component providers" to "providers". In either case, we'll need to lower reified outputs to simple property values for providers that don't support reified outputs, and I think we'll end up treating them uniformly in the provider SDK anyways (and indeed, it looks like that's already the case in this PR).

sdk/go/common/resource/properties.go
justinvp: I didn't originally have the `json` tags, but found that we have this test:

https://github.com/pulumi/pulumi/blob/7ebac460a3c9ca1c2d6625f31e17297985d47ab9/pkg/resource/stack/deployment_test.go#L333

Which fails otherwise because it isn't expecting the new fields.

justinvp:
> Including reified `Output` values--metadata and all--in resource inputs could subsume the information we serialize in several different places (both at the RPC layer and at the statefile layer). I'd love to see a design doc that covers the various pieces of this work and considers whether or not we could/should simplify things by replacing that info with reified outputs. Should we work together on this @justinvp ?

Yes, definitely interested in this. Note that in the short term, we've scoped this to just what is needed to address the multi-lang issue noted in #7434, both to deliver a solution in a short timeframe and to reduce risk associated with a larger change. The aim is that we can initially roll this out to address the multi-lang problem in a way that doesn't preclude us from doing the larger cleanup/consolidation later. Let's sync up so we're on the same page w.r.t. the short-term and long-term plans, and how that might affect the design.

> I'd suggest that we expand the scope of this from "component providers" to "providers". In either case, we'll need to lower reified outputs to simple property values for providers that don't support reified outputs, and I think we'll end up treating them uniformly in the provider SDK anyways (and indeed, it looks like that's already the case in this PR).

I was thinking to initially scope this (both to reduce timeline and blast radius), the language SDKs would only keep output values when marshaling inputs for `RegisterResource` calls when `remote` is `true` and for calls to `Call`, if the monitor says it supports them. And then having the engine only pass along the output values to a provider's `Construct` and `Call`, if the provider says it can accept them, and not any of the other provider RPCs. And in the future, we could consider passing along output values from the language SDKs in more places, and similarly have a way for the provider to say it can accept them in other RPCs. My main concern is someone having a provider that has both custom resources and components, and the regular CRUD operations of the provider not being able to handle the output values yet.

t0yv0:
Reading the changes they look good, but I have this https://github.com/pulumi/pulumi/pull/7871 failure:

```
[nix-shell:~/pulumi/sdk/go/common/resource/plugin]$ go test . -run TestOutputValueTurnaround
go test . -run TestOutputValueTurnaround
--- FAIL: TestOutputValueTurnaround (0.00s)
    rpc_rapid_test.go:55: [rapid] failed after 0 tests: 
        	Error Trace:	rpc_rapid_test.go:71
        	            				engine.go:276
        	            				engine.go:186
        	            				engine.go:123
        	            				engine.go:94
        	            				rpc_rapid_test.go:55
        	Error:      	Not equal: 
        	            	expected: resource.PropertyValue{V:resource.Output{Element:resource.PropertyValue{V:interface {}(nil)}, Known:false, Secret:false, Dependencies:[]resource.URN{}}}
        	            	actual  : resource.PropertyValue{V:resource.Output{Element:resource.PropertyValue{V:interface {}(nil)}, Known:false, Secret:false, Dependencies:[]resource.URN(nil)}}
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -7,4 +7,3 @@
        	            	   Secret: (bool) false,
        	            	-  Dependencies: ([]resource.URN) {
        	            	-  }
        	            	+  Dependencies: ([]resource.URN) <nil>
        	            	  }
        	Test:       	TestOutputValueTurnaround
```

Am I doing something wrong?

t0yv0:
For the long term... I was also trying to read up more on what's the grammar for PropertyValues and where do we use them.  So they're definitely used in communicating RegisterResource and similar calls. Are they used in resource diff? Are they used in state snapshots - seeing first mention here that seems interesting.

I'll do a bit more reading but anything you can drop into design docs or developer doc helps.

t0yv0 (approved):

sdk/go/common/resource/plugin/rpc.go
justinvp: I didn't think they came up in practice -- the [Node.js](https://github.com/pulumi/pulumi/blob/e696fb6c5017b0ebffcb56069438b01dfa5ab089/sdk/nodejs/runtime/rpc.ts#L345-L354), [Python](https://github.com/pulumi/pulumi/blob/e696fb6c5017b0ebffcb56069438b01dfa5ab089/sdk/python/lib/pulumi/runtime/rpc.py#L264-L272), and [.NET](https://github.com/pulumi/pulumi/blob/e696fb6c5017b0ebffcb56069438b01dfa5ab089/sdk/dotnet/Pulumi/Serialization/Serializer.cs#L147-L157) SDKs will marshal an unknown secret as just an unknown value without the secret information. However, the [Go](https://github.com/pulumi/pulumi/blob/e696fb6c5017b0ebffcb56069438b01dfa5ab089/sdk/go/pulumi/rpc.go#L223-L225) SDK will keep it as an unknown value wrapped as a secret, so it is possible. I updated the code here to do the same.

justinvp:
> Are they used in resource diff? Are they used in state snapshots

Yes, they're used for diffs and snapshots. For now, I'm scoping when output values are kept s.t. they do not show up in diffs or snapshots.

PR #7865, opened 2021-08-28
[codegen/schema] Add a schema checker

pgavlin:
- Change the schema package to report semantic errors as diagnostics
  rather than Go errors
- Add a `pulumi schema check` command to the CLI for static checking of
  package schemas

The semantic checker can be extended in the future to add support for
target-specific checks.

### DIFF
--- a/pkg/cmd/pulumi/schema.go
+++ b/pkg/cmd/pulumi/schema.go
@@ -1,0 +1,35 @@
+// Copyright 2016-2021, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


pkg/cmd/pulumi/schema.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/spf13/cobra"
+)
+
+func newSchemaCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "schema",
+		Short: "Analyze package schemas",
+		Long: `Analyze package schemas
+
+Subcommands of this command can be used to analyze Pulumi package schemas. This can be useful to check hand-authored
```
t0yv0, line 28: If this is the correct doc page for what this tool is about, adding a link can be very handy: https://www.pulumi.com/docs/guides/pulumi-packages/schema/

t0yv0 (approved):
Read through, LGTM, send those diags out.. 🚢

justinvp (approved):
LGTM

EvanBoyle (commented):
Looks great! Would be nice to have a small harness that evaluates a test schema, and checks against an expected set of diagnostic errors. Seems necessary if we are going to expect the providers team (and eventually customers) to depend on the output of this tool.

PR #7868, opened 2021-08-30
[sdk/python] Stack preview: enable json output

sticha:
# Description

The pulumi preview CLI command has an option that allows the preview diffs, operations, and overall output to be returned as JSON.
This change allows setting this option also when using the Python Automation API.

Fixes # (issue)

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -13,4 +13,7 @@
 - [codegen/schema] Add a `pulumi schema check` command to validate package schemas.
   [#7865](https://github.com/pulumi/pulumi/pull/7865)
 
+- [sdk/python] - Automation API: Enable JSON output for stack preview.
+  [#7868](https://github.com/pulumi/pulumi/pull/7868)
+
 ### Bug Fixes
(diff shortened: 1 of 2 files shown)


EvanBoyle:
@sticha Have you tried using the structured engine events API? You can pass an `on_event` handler into your preview that will stream engine events containing detailed information: 

https://github.com/pulumi/pulumi/blob/1b67c9ee520ceef61cc00749f7ec2508d4bd52fd/sdk/python/lib/pulumi/automation/_stack.py#L275

sticha:
@EvanBoyle good point, i have also considered it but i would prefer getting the complete json at the end of a preview execution (as this is already a feature of the preview command of the pulumi cli).
However, should there be a reason against including my change, it would probably be possible to build the contents of the json by listening for ResOutputsEvents.

> @sticha Have you tried using the structured engine events API? You can pass an `on_event` handler into your preview that will stream engine events containing detailed information:
> 
> https://github.com/pulumi/pulumi/blob/1b67c9ee520ceef61cc00749f7ec2508d4bd52fd/sdk/python/lib/pulumi/automation/_stack.py#L275

komalali:
@sticha thanks for the PR! With the addition of structured logging, we actually [removed the json flag](https://github.com/pulumi/pulumi/pull/6436) from our Go automation API and to stay consistent we will not accept this PR. For some context, our current implementation for `--json` in preview is not ideal for long-running processes (the big json blob at the end gives no insight to incremental progress), and we intend to reexamine that implementation to bring it more in line with incremental structured logging as we look into implementing `--json` for other commands like `up` and `destroy`.

PR #7870, opened 2021-08-30
Add monitor feature for output values

### LINKED ISSUE #7845: Add a monitor feature for output values
Pre-requisite for: https://github.com/pulumi/pulumi/issues/7434

This feature will be disabled by default, and should be toggleable via an envvar for testing purposes.

justinvp:
This change does two things:

1. Adds a monitor feature for output values, which is disabled by default, but can be enabled via `PULUMI_ENABLE_OUTPUT_VALUES` envvar. When this is enabled, it indicates that the monitor supports unmarshaling output values.

2. Adds a new `acceptOutputs` field to the provider's `ConfigureResponse` that a provider can set to true to indicate that the engine can pass output values to the provider.

To initially scope the use of the feature:

- In the resource monitor, we only keep output values when unmarshaling properties in `RegisterResource` when `remote` is `true` (for multi-lang components) and `Call`.

- In the code that calls a provider, we only keep output values when marshaling properties for `Construct` and `Call` when the provider has indicated it accepts output values.

Fixes #7845

Part of #7434


t0yv0 (approved):
LGTM, proto changes should be safe right. One thing to verify is Go defaulting behavior over newly added bool fields. DisableOutputValues defaults to false, KeepOutputValues defaults to false, so the defaults don't agree, which might not be relevant if we cover all constructor sites and are not affected by Go defaulting.

iwahbe (approved):
This seems right to me.

pgavlin (approved):
LGTM! Took me a minute to grok the polarity of the internal option vs. the envvar :)

PR #7872, opened 2021-08-30
Fix rmdir DeprecationWarning

garrett-hopper:
```
(node:346333) [DEP0147] DeprecationWarning: In future versions of Node.js, fs.rm
sive: true }) will be removed. Use fs.rm(path, { recursive: true }) instead
(Use `node --trace-deprecation ...` to show where the warning was created)
```

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -11,3 +11,5 @@
   resources. [#7784](https://github.com/pulumi/pulumi/pull/7784)
 
 ### Bug Fixes
+
+- Fix Node `fs.rmdir` DeprecationWarning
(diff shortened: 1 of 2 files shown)


emiliza:
Hi!  Thank you for the contribution - unfortunately we cannot merge this as-is.  We're currently building with node 14.x (see broken test [here](https://github.com/pulumi/pulumi/runs/3471416171?check_suite_focus=true)) and `fs.rm` isn't available until [15.x](https://nodejs.org/docs/latest-v15.x/api/fs.html)

garrett-hopper:
I don't know that there's a way to prevent the deprecation notice while still supporting 14.x, so this will just have to wait until 14.x is no longer being used. (Feel free to close if that's not expected to happen anytime soon.)

PR #7873, opened 2021-08-30
Add workflow commands test on windows, unix, lint

iwahbe:
# Description

Allows Pulumi members to run `/run-acceptance-tests windows-only`, `/run-acceptance-tests unix-only`, `/run-acceptance-tests lint-only` as a comment and run the expected workflow. This is to make it easier to develop when tests are failing on a specific platform. 

## Checklist


iwahbe:
/run-unix-tests

This is a test of the system.

PR #7885, opened 2021-09-01
Run workflow from tags

iwahbe:
# Description

Attempt to run test workflows by adding tags. 
This includes 3 tags: `ci/lint`, `ci/build-windows`, and `ci/build-unix`. When these tags are added to a PR, they cause the CI to activate the relevant test. Other tests are skipped. This has no effect on test triggers already in place. 

## Checklist

       This needs to be tested manually. 

### DIFF
--- a/.github/workflows/run-build-and-acceptance-tests.yml
+++ b/.github/workflows/run-build-and-acceptance-tests.yml
@@ -70,7 +70,6 @@ jobs:
       - name: Lint ${{ matrix.directory }}
         run: |
           cd ${{ matrix.directory }} && golangci-lint run -c ../.golangci.yml
-      - name: Remove Label ci/lint
   sdk-lint:
     if: |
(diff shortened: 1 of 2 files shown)


.github/workflows/run-build-and-acceptance-tests.yml
```diff
@@ -32,6 +34,8 @@ jobs:
 
             [1]: ${{ steps.vars.outputs.run-url }}
   go-lint:
+    if: |
+      github.event.label.name == 'ci/lint' || github.event.action != 'labeled'
```
emiliza, line 38: I'm not terribly keen on this being a label (which we're using primarily for metadata), having 2 different ways to execute tests could be confusing and would prefer for these to be slash commands

iwahbe:
Decided not to use this strategy.

PR #7887, opened 2021-09-01
Fix 7862

### LINKED ISSUE #7862: Pulumi v3.10.1+ hangs forever when previewing
`pulumi preview` in v3.10.1+ hangs forever if you use a `pulumi_eks.Cluster` as the parent of another resource. Here's the smallest demonstration I could come up with:

```py
cluster = eks.Cluster(
    "example",
    vpc_id="doesntmatter",
    skip_default_node_group=True,
)

aws.eks.NodeGroup(
    "example",
    cluster_name="example",
    instance_types=["t3.micro"],
    node_role_arn=cluster.instance_roles[0].arn,
    subnet_ids=["doesntmatter"],
    scaling_config=aws.eks.NodeGroupScalingConfigArgs(
        desired_size=1,
        max_size=1,
        min_size=1,
    ),
    opts=pulumi.ResourceOptions(
        # Comment out the following line to make preview work.
        parent=cluster,
    ),
)
```

I've put together a Git repository that packages this reproducer up at https://github.com/benesch/pulumi-bug-eks. Instructions are in the README there.

I'm almost positive that https://github.com/pulumi/pulumi/commit/ecb98b66fd45c1cd5958bfd912eee8b8ce51a09d is the culprit, based on the fact that it is the _only_ commit in v3.10.1 that touches the Python SDK. From a quick skim, I wouldn't be surprised if there's additional trickiness that needs to happen when a remote resource is used as the parent of a local resource.

Not setting the cluster as the parent of the node group is a perfectly fine workaround for us for now. Just lost a fair amount of hair tracking this one down. :D 

cc @pgavlin

t0yv0:
# Description

Fixes #7862 

- [x] Still need to add a test, but with this change the user's problem does not repro.

## Checklist


sdk/python/lib/pulumi/runtime/resource_cycle_breaker.py
```diff
@@ -0,0 +1,69 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import TYPE_CHECKING, Set, Iterable
+
+if TYPE_CHECKING:
+    from .. import Resource
+
+
+_DEPENDENCIES_PROPERTY = '_direct_computed_dependencies'
```
t0yv0, line 21: So this algorithm needs to store the state somewhere. Options to put the state in:

- some property of Resource objects (currently implemented)
- global table (I'm ever-so-slightly worried about leaks and re-entrancy in some auto API cases)
- table scoped to the Pulumi program run that is executing - this would perhaps be ideal, but where do we put such state, which object/class?

EvanBoyle: @pgavlin has talked about adding a kv store as in the engine a first class feature in the pulumi RPC interface to service scenarios just like this. But I don't personally have anything against the approach you've taken to solve the customer issue right now.

t0yv0: Yeah! Hm. This would indeed come handy one day. However as things stand with Py SDK I'd be a bit worried about relying on async KV store for core functionality.. For example I'd love to first get rid of threading and have a really foolproof async situation. in current design core functionality of Py SDK is to order the RegisterResource calls "right" before they hit the engine.

So far I can think of no downside piggybacking on the Resource prop - we're not ever pickling these objects right.. But if needed/wanted I can also explicitly declare this prop on the Resource class so it's more official this is being used.

EvanBoyle (approved):

PR #7891, opened 2021-09-02
Turn off grpc info log in Provider

### LINKED ISSUE #7156: Provider gRPC client logging causes failures.
I'm trying to write my own provider using the boilerplate template but I think there is a breaking bug. From what I can tell the provider is expected to output to stdout the port it is listening on, but what appears to happen is that grpc logging beats it to it.

When I run Pulumi up I get:

```
pulumi:providers:tsb (tsb):
    error: tsb (resource) plugin [/Users/liamwhite/go/bin/pulumi-resource-tsb] wrote a non-numeric port to stdout ('2021-05-27T13:44:04.073942Z	info	[core]parsed scheme: ""'): strconv.Atoi: parsing "2021-05-27T13:44:04.073942Z\tinfo\t[core]parsed scheme: \"\"": invalid syntax
```

I can fix this by manually configuring the grpc logger to either info log to stderr (or discard).

```
// NewHostClient dials the target address, connects over gRPC, and returns a client interface.
func NewHostClient(addr string) (*HostClient, error) {
	grpclog.SetLoggerV2(grpclog.NewLoggerV2(os.Stderr, os.Stderr, os.Stderr)) // <-----------
	conn, err := grpc.Dial(
		addr,
		grpc.WithInsecure(),
		grpc.WithUnaryInterceptor(rpcutil.OpenTracingClientInterceptor()),
		rpcutil.GrpcChannelOptions(),
	)
...
```

The only problem with sending to stderr is that it shows up in diagnostics so it may be preferable to `io.Discard` instead.

liamawhite:
Signed-off-by: Liam White <liam@tetrate.io>

# Description

As described in #7156. When I run Pulumi up I get:

```
pulumi:providers:tsb (tsb):
    error: tsb (resource) plugin [/Users/liamwhite/go/bin/pulumi-resource-tsb] wrote a non-numeric port to stdout ('2021-05-27T13:44:04.073942Z	info	[core]parsed scheme: ""'): strconv.Atoi: parsing "2021-05-27T13:44:04.073942Z\tinfo\t[core]parsed scheme: \"\"": invalid syntax
```

This is due to grpc logging to stdout before the provider writes the port to stdout. This PR explicitly sets the GRPC defaults to prevent users from overriding them and breaking their providers.

Fixes #7156

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works -> There are no tests in this section of the code, but I have verified manually that it solves my problem.

- [ ] I have updated the [CHANGELOG-PENDING](https://github.com/pulumi/pulumi/blob/master/CHANGELOG_PENDING.md) file with my change. -> unsure if this requires a changelog update

### DIFF
--- a/pkg/resource/provider/host.go
+++ b/pkg/resource/provider/host.go
@@ -15,6 +15,8 @@
 package provider
 
 import (
+	"io/ioutil"
+	"os"
 	"strings"
 
 	"github.com/pulumi/pulumi/sdk/v3/go/common/diag"
@@ -24,6 +26,7 @@ import (
 	lumirpc "github.com/pulumi/pulumi/sdk/v3/proto/go"
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/grpclog"
 )
 
 // HostClient is a client interface into the host's engine RPC interface.
@@ -34,6 +37,9 @@ type HostClient struct {
 
 // NewHostClient dials the target address, connects over gRPC, and returns a client interface.
 func NewHostClient(addr string) (*HostClient, error) {
+	// GRPC info logging to stdout introduces a race condition with the printing of the port, so drop them.
+	// See https://github.com/pulumi/pulumi/issues/7156
+	grpclog.SetLoggerV2(grpclog.NewLoggerV2(ioutil.Discard, os.Stderr, os.Stderr))
 	conn, err := grpc.Dial(
 		addr,
 		grpc.WithInsecure(),
--- a/pkg/resource/provider/host.go
+++ b/pkg/resource/provider/host.go
@@ -37,9 +37,9 @@ type HostClient struct {
 
 // NewHostClient dials the target address, connects over gRPC, and returns a client interface.
 func NewHostClient(addr string) (*HostClient, error) {
-	// GRPC info logging to stdout introduces a race condition with the printing of the port, so drop them.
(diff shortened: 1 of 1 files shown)


pkg/resource/provider/host.go
```diff
@@ -34,6 +37,9 @@ type HostClient struct {
 
 // NewHostClient dials the target address, connects over gRPC, and returns a client interface.
 func NewHostClient(addr string) (*HostClient, error) {
+	// GRPC info logging to stdout introduces a race condition with the printing of the port, so drop them.
+	// See https://github.com/pulumi/pulumi/issues/7156
+	grpclog.SetLoggerV2(grpclog.NewLoggerV2(ioutil.Discard, os.Stderr, os.Stderr))
```
lukehoban, line 42: It should not be the case that grpclog logs to stdout by default.  Per https://github.com/grpc/grpc-go/blob/v1.40.0/grpclog/loggerv2.go#L123-L146, logs are only to stderr by default, and only on errors by default.

If you are seeing something different than that, I suspect it must be due to using some different version of that module of some environmental configuration?

That said, since we are particularly incompatible with sending info logs to stdout, it does potentially make sense to try to override this - but perhaps we should set it back to the same thing grpclog does by default (`Discard`, `Discard`, `Stderr`)?

liamawhite: updated

PR #7896, opened 2021-09-02
Validate Name, Version and Enviroment

### LINKED ISSUE #7893: pulumi schema check does not fail on empty name or empty version
## Steps to reproduce

1. echo "{}" > schema.json
2. pulumi schema check schema.json

Expected: 
error: empty name
Actual: 
Nothing, check was clean

iwahbe:
# Description

Adds validators for `Package.Name`, corrects the validator for `Package.Version`, and adds a validator for `Default.Environment`

Fixes #7893

## Checklist


iwahbe:
The `output-funcs` test relies on the `name` of each package being `py_tests`. That was the cause of the confusion.

pkg/codegen/schema/schema.go
```diff
@@ -1417,6 +1422,8 @@ func bindSpec(spec PackageSpec, languages map[string]Language, loader Loader) (*
 		} else {
 			version = &v
 		}
+	} else {
+		diags = diags.Append(errorf("#/version", "No version provided"))
```
pgavlin, line 1426: This will not work without downstream code changes. We allow schemas to elide the version argument to handle situations where the version is not known at the time the schema is generated, which is the case in ~all of our providers. All of our providers pull their version from a git tag, which is not available at the time the schema is generated; it's complicated.

pkg/codegen/schema/schema.go
```diff
@@ -2023,6 +2033,9 @@ func bindDefaultValue(path string, value interface{}, spec *DefaultSpec, typ Typ
 		for name, raw := range spec.Language {
 			language[name] = json.RawMessage(raw)
 		}
+		if spec.Environment == nil {
+			diags = diags.Append(errorf(path, "Default must specify an environment"))
+		}
```
pgavlin, line 2038: This is not correct: it is valid for defaults to provide only static values and no env vars. It seems reasonable to require one or the other, though.

pgavlin (changes requested):

pkg/codegen/schema/schema.go
iwahbe: I may have read this wrong. [The schema docs](https://www.pulumi.com/docs/guides/pulumi-packages/schema/) is a little confusing.

pkg/codegen/schema/schema.go
iwahbe: This flatly contradicts what the [docs](https://www.pulumi.com/docs/guides/pulumi-packages/schema/) say about the `package.version` property. Is this a distinction we make for our own packages, or a general rule (in which case we should update the docs)?

iwahbe: Issue in Hugo here: https://github.com/pulumi/pulumi-hugo/issues/570

pgavlin: > Is this a distinction we make for our own packages

This, basically. I think it's okay to leave the documentation as-is.

pkg/codegen/schema/schema.go
pgavlin: Ah, no, this is my mistake. We spread default information out into a couple of places, and if a `DefaultSpec` is provided, then the `Environment` field should be non-empty. With that in mind, I think the rule should be 

```suggestion
		if len(spec.Environment) == 0 {
			diags = diags.Append(errorf(path, "default must specify an environment"))
		}
```

pkg/codegen/schema/schema.go
```diff
@@ -1878,6 +1885,9 @@ func (t *types) bindType(path string, spec TypeSpec, inputShape bool) (result Ty
 		var discriminator string
 		var mapping map[string]string
 		if spec.Discriminator != nil {
+			if spec.Discriminator.PropertyName == "" {
+				diags = diags.Append(errorf(path, "Discriminator must provide a property name"))
```
pgavlin, line 1889:
```suggestion
				diags = diags.Append(errorf(path, "discriminator must provide a property name"))
```

pkg/codegen/schema/schema.go
```diff
@@ -1408,6 +1408,11 @@ func errorf(path, message string, args ...interface{}) *hcl.Diagnostic {
 func bindSpec(spec PackageSpec, languages map[string]Language, loader Loader) (*Package, hcl.Diagnostics, error) {
 	var diags hcl.Diagnostics
 
+	// Validate that there is a name
+	if spec.Name == "" {
+		diags = diags.Append(errorf("#/name", "No name provided"))
```
pgavlin, line 1413:
```suggestion
		diags = diags.Append(errorf("#/name", "no name provided"))
```

pkg/cmd/pulumi/schema_check.go
```diff
@@ -71,6 +72,9 @@ func newSchemaCheckCommand() *cobra.Command {
 			diagWriter := hcl.NewDiagnosticTextWriter(os.Stderr, nil, 0, true)
 			wrErr := diagWriter.WriteDiagnostics(diags)
 			contract.IgnoreError(wrErr)
+			if err == nil && len(diags) != 0 {
+				return errors.New("schema validation failed")
```
pgavlin, line 76:
```suggestion
			if err == nil && drags.HasErrors() {
				return errors.New("schema validation failed")
```

pkg/codegen/internal/test/testdata/output-funcs/getIntegrationRuntimeObjectMetadatum.json
```diff
@@ -134,9 +134,7 @@
         }
       },
       "type": "object",
-      "required": [
-        "type"
-      ]
+      "required": ["type"]
```
pgavlin, line 137: Is the reformatting in these files a function of your editor?

t0yv0:
Since Pat is already here taking a raincheck and I'm a bit underwater with reviews, will review tomorrow..

pkg/codegen/internal/test/testdata/output-funcs/getIntegrationRuntimeObjectMetadatum.json
iwahbe: Sorry, yes. It won't happen again. I can undo them if that is helpful.

pgavlin: > Sorry, yes. It won't happen again. It will revert in the next commit.

Oh, no worries--I was just wondering if it was intentional. I don't really care about the formatting here as long as we're consistent, and I'm happy to pick whatever approach works most naturally for us when we need to edit there files :)

pkg/cmd/pulumi/schema_check.go
pgavlin: Looks like this change didn't get applied?

t0yv0:
LGTM modulo Pat's comment, should the clause short-circuit on len(diags) = 0 or just errors? assuming there are warnings and errors. We should not stop if just warnings?

Also. Related thought. 

This uses JSON schema meta-schema:

https://github.com/pulumi/pulumi/pull/7952

Potentially the same could be accomplished by adding validation patterns http://json-schema.org/understanding-json-schema/reference/regular_expressions.html#example  and riding on generic JSON schema validation checks.

Don't wait though till that lands, check this in.

t0yv0 (approved):

pgavlin (approved):
LGTM!

PR #7899, opened 2021-09-02
5758 for C#/.NET

### LINKED ISSUE #5758: Functions/datasources don't compose well with Pulumi model - should accept Outputs

t0yv0:
# Description

Fixes #5758  for .NET/C#

## Checklist


pkg/codegen/internal/test/testdata/external-resource-schema/dotnet/ArgFunction.cs
```diff
@@ -13,6 +13,18 @@ public static class ArgFunction
...
             => Pulumi.Deployment.Instance.InvokeAsync<ArgFunctionResult>("example::argFunction", args ?? new ArgFunctionArgs(), options.WithVersion());
+
+        public static Output<ArgFunctionResult> Invoke(ArgFunctionOutputArgs? args = null, InvokeOptions? options = null)
+        {
+            args = args ?? new ArgFunctionOutputArgs();
+            return Pulumi.Output.All(
+                args.Name.Box()
+            ).Apply(a => {
+                    var args = new ArgFunctionArgs();
+                    a[0].Set(args, nameof(args.Name));
```
t0yv0, line 24: The Box business is somewhat messy but that's what I ended up with to workaround the problem that (1) tuple helpers `Output[A] * Output[B] -> Output[A*B]` only work up to 8, and mixing them up gets ugly really fast, while the `All` combinator used here requires all the arguments to have the same type. In addition, the `a[0].Set()` pattern uses reflection as I didn't find a way that C# type system understands to map this back to the properties without double-stating the type of the property  in the code (ouch). I've been really really wishing this was in F# when writing this..

Now I was somewhat hesitant but if this is a problem for reviewers I can rework this, now that the Go codegen test actually *runs unit tests* against the generated code, I've got a higher confidence in iterating on this.

pkg/codegen/internal/test/testdata/external-resource-schema/dotnet/ArgFunction.cs
```diff
@@ -13,6 +13,18 @@ public static class ArgFunction
     {
         public static Task<ArgFunctionResult> InvokeAsync(ArgFunctionArgs? args = null, InvokeOptions? options = null)
             => Pulumi.Deployment.Instance.InvokeAsync<ArgFunctionResult>("example::argFunction", args ?? new ArgFunctionArgs(), options.WithVersion());
+
+        public static Output<ArgFunctionResult> Invoke(ArgFunctionOutputArgs? args = null, InvokeOptions? options = null)
```
t0yv0, line 17: It's called .Invoke but .InvokeOutput is also a possibility to match what we have in other languages. Votes?

pkg/codegen/internal/test/testdata/external-resource-schema/dotnet/ArgFunction.cs
```diff
@@ -26,6 +38,16 @@ public ArgFunctionArgs()
         }
     }
 
+    public sealed class ArgFunctionOutputArgs
```
t0yv0, line 41: *OutputArgs naming convention is something we chose here, can chose differently.

pkg/codegen/internal/test/testdata/external-resource-schema/dotnet/Utilities.cs
```diff
@@ -60,6 +62,57 @@ public static InvokeOptions WithVersion(this InvokeOptions? options)
             };
         }
 
+        public static Input<Dictionary<string,T>> ToDict<T>(this InputMap<T> inputList)
+        {
+            return inputList.Apply(v => new Dictionary<string,T>(v));
+        }
+
+        public static List<T> ToList<T>(this IEnumerable<T> elements)
```
t0yv0, line 70: Special casing for these in codegen is definitely a code smell; I wonder if there is any better way.

pkg/codegen/internal/test/testdata/output-funcs/dotnet/funcWithDefaultValue.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        public FuncWithDefaultValueArgs()
+        {
+            B = "b-default";
+        }
+    }
+
+    public sealed class FuncWithDefaultValueOutputArgs
+    {
+        [Input("a", required: true)]
```
justinvp, line 51: These attributes aren't used on the `XxxOutputArgs` variants, right? Should we not emit them?

pkg/codegen/internal/test/testdata/external-resource-schema/dotnet/ArgFunction.cs
justinvp: Seems reasonable to me.

justinvp (approved):
Some comments, but LGTM otherwise.

pkg/codegen/internal/test/testdata/output-funcs/dotnet/funcWithConstInput.cs
```diff
@@ -0,0 +1,31 @@
...
+namespace Pulumi.MadeupPackage.Codegentest
+{
+    public static class FuncWithConstInput
+    {
+        /// <summary>
+        /// Codegen demo with const inputs
+        /// </summary>
+        public static Task InvokeAsync(FuncWithConstInputArgs? args = null, InvokeOptions? options = null)
+            => Pulumi.Deployment.Instance.InvokeAsync("madeup-package:codegentest:funcWithConstInput", args ?? new FuncWithConstInputArgs(), options.WithVersion());
+    }
```
mikhailshilkov, line 19: Why no Invoke variant?

pkg/codegen/internal/test/testdata/output-funcs/dotnet/funcWithDictParam.cs
```diff
@@ -0,0 +1,84 @@
...
+        }
+    }
+
+    public sealed class FuncWithDictParamOutputArgs
+    {
+        [Input("a")]
+        private InputMap<string>? _a;
+        public InputMap<string> A
+        {
+            get => _a ?? (_a = new InputMap<string>());
```
mikhailshilkov, line 60: Could we do
```suggestion
            get => _a ??= new InputMap<string>();
```

pkg/codegen/internal/test/testdata/output-funcs/dotnet/listStorageAccountKeys.cs
```diff
@@ -0,0 +1,103 @@
...
+    public static class ListStorageAccountKeys
+    {
+        /// <summary>
+        /// The response from the ListKeys operation.
+        /// API Version: 2021-02-01.
+        /// </summary>
+        public static Task<ListStorageAccountKeysResult> InvokeAsync(ListStorageAccountKeysArgs args, InvokeOptions? options = null)
+            => Pulumi.Deployment.Instance.InvokeAsync<ListStorageAccountKeysResult>("azure-native:codegentest:listStorageAccountKeys", args ?? new ListStorageAccountKeysArgs(), options.WithVersion());
+
+        public static Output<ListStorageAccountKeysResult> Invoke(ListStorageAccountKeysOutputArgs args, InvokeOptions? options = null)
```
mikhailshilkov, line 21: Are we missing a doc comment here?

mikhailshilkov (commented):
This looks exciting!

pkg/codegen/internal/test/testdata/external-resource-schema/dotnet/ArgFunction.cs
t0yv0: I'm leaning InvokeOutput too since in other langs we have _output etc, which Luke also preferred.

pkg/codegen/internal/test/testdata/dash-named-schema/dotnet/Utilities.cs
```diff
@@ -60,6 +62,57 @@ public static InvokeOptions WithVersion(this InvokeOptions? options)
             };
         }
 
+        public static Input<Dictionary<string,T>> ToDict<T>(this InputMap<T> inputList)
```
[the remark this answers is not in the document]
t0yv0: I'd like them in SDK too but I was worried about the rollout order, that is, what if users get upgraded Azure-Native Package before they get an upgraded SDK package? This is not a worry, e.g. we have dep constraints to solve this? I can move to SDK right away then.

pkg/codegen/internal/test/testdata/external-resource-schema/dotnet/ArgFunction.cs
mikhailshilkov: Other languages have a very different invoke syntax so I'm not sure that should be a large factor. I expect we'd advise this new form by default - unless folks need the result for control flow, so having a simple unambiguous name would be my preference.

Resource constructors return outputs and they don't have `Output` in their name, so I don't see much difference here.

justinvp: FWIW, I don't feel strongly about this -- I could go either way. `Invoke` is the nicer name, but I do worry slightly that `Invoke` will be seen as the blocking variant of `InvokeAsync`, which isn't the case.

pkg/codegen/internal/test/testdata/dash-named-schema/dotnet/Utilities.cs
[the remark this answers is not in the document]
mikhailshilkov: They'll have to upgrade the core SDK if they want the new invoke syntax. If they don't use it - I think it will just work? I feel like we did these things in the past but I may be wrong.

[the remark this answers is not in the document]
t0yv0: Let's try it I guess.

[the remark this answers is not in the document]
justinvp: If we're adding it to the SDK, can we name it `ToDictionary` rather than `ToDict` to align with the naming in the framework, e.g. https://source.dot.net/#q=todict?

pkg/codegen/internal/test/testdata/output-funcs/dotnet/funcWithDictParam.cs
t0yv0: That's something I'm going to leave for now; it came from `genInputType` independently from my changes, and doesn't look trivial to fix up.

pkg/codegen/internal/test/testdata/output-funcs/dotnet/Assert.cs
```diff
@@ -0,0 +1,34 @@
+// Copyright 2016-2021, Pulumi Corporation
+
+using System;
+using System.Threading.Tasks;
+
+using FluentAssertions;
+
+namespace Pulumi.MadeupPackage.Codegentest
```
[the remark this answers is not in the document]
t0yv0: Wont' fix this time, this runs against some assumptions in the JSON schema that's reused across langs, and I didn't find quick-and-dirty way to make codegen rename this.

pkg/codegen/internal/test/testdata/output-funcs/dotnet/funcWithDefaultValue.cs
t0yv0: My initial instinct (now added to the PR) is to drop them as they're not enforced, but alternatively we can use this to try to enforce required values via reflection. I think that's what inheriting from InvokeArgs used to do, but I got rid of it because as written it was not letting inputs pass through. Let me see..

pkg/codegen/internal/test/testdata/output-funcs/dotnet/funcWithConstInput.cs
t0yv0: So. This function does not have a return type, it's returning void. When I was developing this for the other languages, I was slightly intimidated by covering these cases correctly, so I did a search on AWS and Azure, and having found *no functions* that return void, I decided to just not handle it.

Now that I have automated testing on the gen'd code, I'm less intimidated about it, and I can bring this feature in as a subsequent set of PRs across languages. However d'you know actual use case that we really need it? Did I miss some Azure or AWS functions with void returns, or have some been added lately?

mikhailshilkov: No, I don't really know why such a function would be needed, as functions are supposed to be side-effect-free. I'm fine with ignoring this use case.

t0yv0: I missed that - we need to upgrade our specs or else I need to RTFM. In that case this is just a bad test case. I can later on update the test case to just test the paths with output values we care about.

pkg/codegen/internal/test/testdata/output-funcs/dotnet/CodegenTests.cs
```diff
@@ -0,0 +1,192 @@
...
+        {
+            Func<string,Func<FuncWithDefaultValueOutputArgs>,Task> check = (
+                (expected, args) => Assert
+                .Output(() => FuncWithDefaultValue.Invoke(args()).Apply(x => x.R))
+                .ResolvesTo(expected)
+            );
+
+            // Since A is required, not passing it is an exception.
+            // Perhaps this should be rejected by the typechecker
+            // instead? Why is A optional statically?
```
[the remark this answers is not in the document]
t0yv0: Yea.. probably not. Thinking about this with fresh covid brains, the only way in C# would be positional arguments in the constructor, but we ran into problems on how to order those, similar tow what Justin has been pointing out in Python etc. I think we should leave as is. I will remove the questions from the comment.

pkg/codegen/internal/test/testdata/output-funcs/dotnet/funcWithDefaultValue.cs
t0yv0: Never mind. The required arguments are enforced by dynamic checks already on the underlying function. So I think skipping the Input is ok.

pkg/codegen/internal/test/testdata/output-funcs/dotnet/listStorageAccountKeys.cs
t0yv0: True, added.

t0yv0:
Going back and forth on naming, sorry.. It's really easy to change before we ship but becomes more of an issue with backwards compatibility later: 

In the last version I considered @mikhailshilkov preferences that is we have:

```
F.Invoke(new FInputArgs() {...})
```

The argument that you'd want to recommend this .Invoke form as primary is a good argument to keep it shorter. 

Now I'm opting for the Args class naming that is hopefully entirely unsurprising:

```
F.Invoke(new FInvokeArgs() {...})
```

`FArgs` would have been better but it's taken for InvokeAsync and we cannot change (BW compat).


This is a bit out of tune with Python and Go (and upcoming node), but that may be ok:

```
func FuncWithDefaultValueOutput(ctx *pulumi.Context, args FuncWithDefaultValueOutputArgs, opts ...pulumi.InvokeOption) FuncWithDefaultValueResultOutput {
```

t0yv0:
@justinvp @mikhailshilkov last chance to vote on naming then I'll merge. Thanks a lot!

mikhailshilkov:
I'm happy with the naming from your last comment 👍  Maybe ask in Slack so that everyone had a chance to chime in?

Is it already too late to change Python and Go? (has it shipped in some providers?)

t0yv0:
Discussed naming on Slack, keeping naming as-is.

Now. Downstream codegen tests fail. I *think* this is because since moving helper types like Box to the SDK, the downstream codegen does not check against latest version of the SDK. This sounds safe -ish but I'm hesitant to merge as then all PRs will invalidate the downstream codegen checks?

```
Error: /home/runner/work/pulumi/pulumi-aws/sdk/dotnet/ServiceCatalog/GetConstraint.cs(79,25): error CS1061: 'Input<string>' does not contain a definition for 'Box' and no accessible extension method 'Box' accepting a first argument of type 'Input<string>' could be found (are you missing a using directive or an assembly reference?) [/home/runner/work/pulumi/pulumi-aws/sdk/dotnet/Pulumi.Aws.csproj]
```

Thoughts on the best way to proceed here?

mikhailshilkov:
Can we ship the helpers first and release the core SDK? I'd expect this would fix downstream tests.

t0yv0:
That sounds good. Let's do this. Start with https://github.com/pulumi/pulumi/pull/7934

t0yv0:
@mikhailshilkov @justinvp apologies for force-push here, there are no changes to the logic I have just brought the PR up-to-date with the latest SDK codegen test framework. Also of note (called out in CHANGELOG_PENDING.md) - the generated code relies on helper functions that are only available in Pulumi .NET SDK 3.13 or higher. If this is allright I would like to merge.

pkg/codegen/dotnet/gen.go
```diff
@@ -1230,6 +1249,34 @@ func (mod *modContext) genResource(w io.Writer, r *schema.Resource) error {
...
+}
+
+func allOptionalInputs(fun *schema.Function) bool {
+	allOptional := true
+	if fun.Inputs != nil {
+		for _, prop := range fun.Inputs.Properties {
+			allOptional = allOptional && !prop.IsRequired()
+		}
+	}
+	return allOptional
```
pgavlin, line 1277: Nit:

```suggestion
	if fun.Inputs != nil {
		for _, prop := range fun.Inputs.Properties {
			if prop.IsRequired() {
				return false
			}
		}
	}
	return true
```

pgavlin (approved):
LGTM. This looks like pretty much exactly what I'd expect 👍

mikhailshilkov (approved):
🎉

pkg/codegen/dotnet/gen.go
t0yv0: Coming shortly in follow-up PR.

t0yv0:
Windows had a failure:

```
   https://github.com/pulumi/pulumi/pull/7899/checks?check_run_id=3702752010

   --- FAIL: TestEngineEventPerf (53.94s) (?)

```

Running again to see if flaky :/ if not flaky will merge up.

t0yv0:
These downstream checks are wonderful! Caught regression on pulumi-aws, investigating:

```
Error: /home/runner/work/pulumi/pulumi-aws/sdk/dotnet/AutoScaling/Inputs/GetAmiIdsFilter.cs(13,25): error CS0263: Partial declarations of 'GetAmiIdsFilterArgs' must not specify different base classes [/home/runner/work/pulumi/pulumi-aws/sdk/dotnet/Pulumi.Aws.csproj]
```

t0yv0:
I've isolated the issue to a naming collision of helper types under `"compatibility": "tfbridge20"` which is still in use under pulumi-aws. Doing a bit of research on this flag. Noting that we have rather limited coverage of codegen under this flag,  only under pkg/codegen/internal/test/testdata/aws.json.

t0yv0:
Couple of notes here, and apologies to reviewers - I can open a fresh PR if this is helpful since I'm having to change the approach from the ground up. Two issues came up when inspecting downstream failure:

1. `aws:getAmiIds` was not really passing unit tests due to Apply in .NET *not* converting `getAmiIdsFilter` from input-enabled .NET type to plain .NET type. I am very glad this got caught and I reworked the solution to not rely on Apply for the implementation. Instead, it relies on existing serde capability which should, in principle, serialize input and plan version of the same type to identical value (modulo dependencies). There remain two todos here I am working on:

- [ ] enhance unit test to exercise getAmiIdsFilter

- [ ] fix the implementation of the new Invoke primitive to propagate the meta-data such as resource deps into Output result, to make it work "as if" it was an Apply, and cover this with a test

2. There was a new name conflict only showing up under  `"compatibility": "tfbridge20"` flag as present in AWS. This involves naming the `getAmiIdsFilter` type in .NET. Under this flag, the plain version got named `GetAmiIdsFilterArgs` and the input-enabled version got named `GetAmiIdsFilterArgs` also in the Inputs namespace. The current workaround renames the new version to `GetAmiIdsFilterInputArgs`. Naively doing so resulted in breaking changes in pulumi-aws provider which was not acceptable, so I started tracking additional information in type details to detect when we would be dealing with types that would only be emitted to support the new `fn.Invoke` feature. The resulting workaround is not fully general but appears to do a good job of pulumi-aws. Note that both the problem and workaround *only* affect `tfbridge20` style providers. TODOs:

- [ ] check other downstream checks for absence of breaking changes

- [ ] get team feedback on the workaround and potentially come up with a better solution

- [ ] get team feedback on *FilterArgs suffix, perhaps use better naming here

pgavlin:
>  reworked the solution to not rely on Apply for the implementation. Instead, it relies on existing serde capability which should, in principle, serialize input and plan version of the same type to identical value

Does this mean that users cannot use `Apply` with the existing (i.e. non-`Output`) signatures for these functions?

t0yv0:
> Does this mean that users cannot use Apply with the existing (i.e. non-Output) signatures for these functions?

With currently published software, the users can call `getAmiIds` right now via apply and set existing GetAmiIdsFilterArgs that is plain-typed. If the users have some Outputs that they want to transform into GetAmiIdsFilterArgs, they need to do so explicitly via Apply/All.

The difficulty is strictly with the new convenience glue types we are generating here, GetAmiIdsFilterInputArgs (inputty version of GetAmiIdsFilterArgs) didn't work as advertised until the latest changes.

sdk/dotnet/Pulumi/Deployment/Deployment_Invoke.cs
```diff
@@ -16,21 +18,70 @@ Task IDeployment.InvokeAsync(string token, InvokeArgs args, InvokeOptions? optio
...
+            //
+            // - the provider is responsible for deciding whether the
+            //   `Output<T>` is secret and known, and may add
+            //   additional dependencies
+            //
+            // This means that presence of secrets or unknowns in the
+            // `args` does not guarantee the result is secret or
+            // unknown, which differs from Pulumi SDKs that choose to
+            // implement these invokes via `Apply` (currently Go and
+            // Python).
```
justinvp, line 48: I thought Node.js will use `.apply`? Or is this comment just because the Node.js functionality hasn't landed yet?

This difference across language runtimes is unfortunate. Remind me again why we can't implement this via `Apply` in .NET?

sdk/dotnet/Pulumi/Resources/InvokeArgs.cs
```diff
@@ -13,10 +13,6 @@ public abstract class InvokeArgs : InputArgs
 
         private protected override void ValidateMember(Type memberType, string fullName)
```
justinvp, line 14: I'd have to double check, but after removing the implementation below, I suspect no subclass actually has an implementation and `ValidateMember` could be nuked entirely. But that can be cleaned up separately.

t0yv0: I'll try that.

pkg/codegen/internal/test/testdata/output-funcs-tfbridge20/dotnet/GetAmiIds.cs
```diff
@@ -0,0 +1,194 @@
...
+
+        [Input("filters")]
+        private InputList<Inputs.GetAmiIdsFilterInputArgs>? _filters;
+
+        /// <summary>
+        /// One or more name/value pairs to filter off of. There
+        /// are several valid keys, for a full reference, check out
+        /// [describe-images in the AWS CLI reference][1].
+        /// </summary>
+        public InputList<Inputs.GetAmiIdsFilterInputArgs> Filters
```
t0yv0, line 115: @justinvp this is the pointer to the counter-example where Apply approach broke down.

Using terminology from the doc by @pgavlin : https://github.com/pulumi/pulumi/blob/master/developer-docs/architecture/type-system.md#inputshapet

Suppose we have ideal type X, and we have 

- `dotnetType(X) = GetAmiIdsArgs`
-  then `dotnetType(inputShape(X)) = GetAmiIdsInvokeArgs`
- But as part of covering the fields of this record, we discover `dotnetType(Y) = Inputs.GetAmiIdsFilterArgs` 
- and `dotnetType(inputShape(Y)) = Inputs.GetAmiIdsFilterInputArgs`
- `X.Filters` is of type type `List<Y>`

Now with the Apply approach, we need to build a .NET function `Inputs.GetAmiIdsInvokeArgs -> Output<GetAmiIdsArgs>`. The previous implementation was not cutting it. It stumbled on auto-converting `Inputs.GetAmiIdsFilterInputArgs` to `Inputs.GetAmiIdsFilterArgs`.

t0yv0: I'd continue arguing that codegen or else SDK needs to reliably be able to do conversions of the form `inputShape(X) -> Input<X>`  in the runtime of the target language, of which the above is a special case :) However poking at the existing SDK it did not sound trivial. I'd have to be reflection, and I don't think it's built yet. Then I noticed that serde has this capability (sort of) and changed the approach.

sdk/dotnet/Pulumi/Deployment/Deployment_Invoke.cs
t0yv0: Answered in another comment below. LMK if you think you see a way forward with Apply. If we could deeply cross-convert plain and inputty types we still use Apply.  It's looking like quite bit of work though. I considered doing it by serializing and deserializing but at first glance even that is non-trivial with current API.

Issue #7878, opened 2021-08-31
Consider upgrading NodeJS to 14.x.x or 16.x.x in `pulumi/pulumi:latest`


aureq:
Currently, the container image `pulumi/pulumi:latest` has NodeJS 12.x.x installed ([Ref](https://github.com/pulumi/pulumi/blob/master/docker/pulumi/Dockerfile#L31)). This version of NodeJS is currently in maintenance mode until April 30th 2022 ([Ref](https://nodejs.org/en/about/releases/)).

For this container image, it would be nice to see NodeJS 14.x.x so Pulumi users would be able to leverage `es2020` features ([Ref](https://node.green/#ES2020)).

Notes:
* NodeJS 14.x.x will begin its maintenance phase on October 19th 2021, so considering NodeJS 16.x.x may be better.
* `pulumi/pulumi-nodejs:latest` already uses NodeJS 14.17.5

stack72:
We can't use Node 16.x yet as we don't use that in our main build - I will upgrade to 14.x now

PR #7904, opened 2021-09-03
Upgrade pulumi docker image to use NodeJS 14.x

stack72:
Fixes: #7878

also, we stopp building the pulumi actions docker container as we are
now EOL as per 31st August


emiliza (approved):

stack72:
Merging as macOS test is unrelated

PR #7912, opened 2021-09-07
Fixed "unknown property '0'" error

T-Vova:
# Description
No longer throws an error when encountering the [legacy HCL indexation operator](https://github.com/hashicorp/hcl/blob/main/hclsyntax/spec.md) ".0"

The following HCL example now successfully converts.

Original:
```hcl
...
    access_policy {
        tenant_id = azurerm_log_analytics_cluster.example.identity.0.tenant_id
        object_id = azurerm_log_analytics_cluster.example.identity.0.principal_id

        key_permissions = [
            "get",
            "unwrapkey",
            "wrapkey",
        ]
    }
}
...
```
Result:
```ts
...
        {
            tenantId: exampleCluster.identity.apply(identity => identity[0].tenantId),
            objectId: exampleCluster.identity.apply(identity => identity[0].principalId),
            keyPermissions: [
                "get",
                "unwrapkey",
                "wrapkey",
            ],
        },
    ],
...
```

### DIFF
--- a/pkg/codegen/hcl2/model/type_object.go
+++ b/pkg/codegen/hcl2/model/type_object.go
@@ -70,9 +70,10 @@ func (t *ObjectType) Traverse(traverser hcl.Traverser) (Traversable, hcl.Diagnos
 	}
 
 	if keyType == NumberType {
-
 		// The key is of NumberType, which means that it refers to the legacy HCL indexation syntax (".0", ".1", etc)
 		return t, nil
+	}
+
 	// The key is of PropertyType, which means that it is most likely a property (".skuName", "tenantId", etc)
 	keyString, err := convert.Convert(key, cty.String)
 	contract.Assert(err == nil)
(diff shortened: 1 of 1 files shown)


pkg/codegen/hcl2/model/type_object.go
```diff
@@ -69,15 +69,25 @@ func (t *ObjectType) Traverse(traverser hcl.Traverser) (Traversable, hcl.Diagnos
...
+		contract.Assert(err == nil)
+
+		propertyName := keyString.AsString()
+		propertyType, hasProperty := t.Properties[propertyName]
+		if !hasProperty {
+
+			// The specific key can't be found in our property list.
+			return DynamicType, hcl.Diagnostics{unknownObjectProperty(propertyName, traverser.SourceRange())}
+		}
+		return propertyType, nil
```
pgavlin, line 89: Nit: it's idiomatic in Go to omit the `else` statement if the `else` follows an `if` that always returns.

```suggestion
	// The key is of PropertyType, which means that it is most likely a property (".skuName", "tenantId", etc)
	keyString, err := convert.Convert(key, cty.String)
	contract.Assert(err == nil)

	propertyName := keyString.AsString()
	propertyType, hasProperty := t.Properties[propertyName]
	if !hasProperty {
		// The specific key can't be found in our property list.
		return DynamicType, hcl.Diagnostics{unknownObjectProperty(propertyName, traverser.SourceRange())}
	}
	return propertyType, nil
```

pgavlin (approved):
One nit; LGTM otherwise!

stack72:
Merging as unrelated flaky test failures but approved

PR #7914, opened 2021-09-07
Fix 7913 Python codegen issue with double-quote inside docstrings

### LINKED ISSUE #7913: Python codegen does not escape docstrings correctly in all cases
Python codegen does not escape docstrings correctly in all cases.

For example this code is wrong:

```
"""Corresponds to "cloudaudit.googleapis.com/activity""""
DATA_ACCESS = "DATA_ACCESS"
```

Context:

https://github.com/pulumi/pulumi-google-native/issues/195

## Steps to reproduce

1.
2.
3.
4.

Expected: 
Actual:

t0yv0:
# Description

Fixes #7913 

## Checklist


komalali:
@t0yv0 what do you think about just adding double quotes into one of the existing test schemas? I feel like this particular error doesn't need an entirely new set of schema.

komalali (approved):
LGTM except maybe just adding a relevant description to the existing enum schema rather than adding a new schema/test set.

mikhailshilkov:
Agree with @komalali here - I think the tests have maintenance cost so we should try to find the balance in quantity vs quality

t0yv0:
Yeah no problem, I'll rework, this is a bit noisy for sure.

PR #7915, opened 2021-09-07
Fix lint

t0yv0:
# Description

Fixed build failures around Go lint. Apparently there is a change between 1.16 and 1.17 on the conditional compilation syntax.

```
cd pulumi/sdk
docker run --rm -v $(pwd)/..:/pulumi -w /pulumi/sdk golangci/golangci-lint:latest golangci-lint run -c ../.golangci.yml
```

Fixes derived by running `gofmt -w -s` on offending files in the same container.

## Checklist

### DIFF
--- a/pkg/backend/watch.go
+++ b/pkg/backend/watch.go
@@ -1,3 +1,4 @@
+//go:build !darwin || !arm64
 // +build !darwin !arm64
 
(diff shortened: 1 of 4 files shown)


justinvp (approved):
LGTM

PR #7916, opened 2021-09-07
Retry http downloads that intermittently fail builds

### LINKED ISSUE #7892: Flaky  codegen/dotnet TestGeneratePackage/External_resource_schema
```
=== FAIL: codegen/dotnet TestGeneratePackage/External_resource_schema (337.06s)
    sdk_driver.go:124: 
        	Error Trace:	sdk_driver.go:124
        	Error:      	Received unexpected error:
        	            	error binding resource example::Component: failed to bind properties for example::Component: error binding type for property "storageClasses": resolving package /kubernetes/v2.7.3/schema.json#/resources/kubernetes:storage.k8s.io%2Fv1:StorageClass: failed to install plugin kubernetes-2.7.3: untarring file /Users/runner/.pulumi/plugins/resource-kubernetes-v2.7.3/pulumi-resource-kubernetes: stream error: stream ID 3; PROTOCOL_ERROR
        	Test:       	TestGeneratePackage/External_resource_schema
    --- FAIL: TestGeneratePackage/External_resource_schema (337.06s)

```

## Steps to reproduce

https://github.com/pulumi/pulumi/runs/3489651662

t0yv0:
We repeatedly hit issues with flaky tests around codegen due to the automatic plugin installation pass.  

This change adds  byte-count check and naive retry with 100ms fixed backoff.  Investigation of the underlying code revealed that `.Download()` does retry already on status codes but it returns lazily consumed HTTP body stream. Therefore we hit intermittent errors when during the untar operation on this network stream there is a network protocol error and some data fails to load.

The fix eagerly pulls the data into a temp file and retries if the file has a different size from the one promised by the HTTP server.

This was tough to test outside our CI, trying a loop of N=1000 does not detect the problem, manually simulating network issues does not either.

What we could do instead:

- if codegen code didn't do automatic plugin installation at all, perhaps this would be simpler/better? but definitely breaking

- bake this behavior into `pkgPlugin.Download()` rather than the call site
 
# Description

Fixes #7892

## Checklist


pkg/codegen/schema/loader.go
```diff
@@ -38,7 +43,7 @@ func (l *pluginLoader) getPackage(key string) (*Package, bool) {
 }
 
 // ensurePlugin downloads and installs the specified plugin if it does not already exist.
-func (l *pluginLoader) ensurePlugin(pkg string, version *semver.Version) error {
+func ensurePlugin(pkg string, version *semver.Version) error {
```
pgavlin, line 46: Did you move this off of `pluginLoader` b/c it doesn't actually need any of the loader's fields?

pkg/codegen/schema/loader.go
```diff
@@ -51,12 +56,66 @@ func (l *pluginLoader) ensurePlugin(pkg string, version *semver.Version) error {
...
+		}
+		defer tarball.Close()
+		copiedByteCount, err := io.Copy(dst, tarball)
+		if err != nil {
+			return err
+		}
+		if copiedByteCount != expectedByteCount {
+			return fmt.Errorf("Expected %d bytes but copied %d when downloading plugin %s",
+				expectedByteCount, copiedByteCount, pkgPlugin)
+		}
```
pgavlin, line 74: Curious: have we hit this in practice?

pkg/codegen/schema/loader.go
```diff
@@ -51,12 +56,66 @@ func (l *pluginLoader) ensurePlugin(pkg string, version *semver.Version) error {
...
+		for attempt := 0; ; attempt++ {
+			tempFile, err := tryDownloadToFile()
+			if err == nil {
+				return tempFile, nil
+			}
+
+			if err != nil && attempt >= 5 {
+				return tempFile, err
+			}
+			time.Sleep(100 * time.Microsecond)
```
pgavlin, line 104: Should we use exponential rather than linear backoff here?

pkg/codegen/schema/loader.go
```diff
@@ -51,12 +56,66 @@ func (l *pluginLoader) ensurePlugin(pkg string, version *semver.Version) error {
...
 			return errors.Wrapf(err, "failed to download plugin: %s", pkgPlugin)
 		}
-		if err := pkgPlugin.Install(tarball); err != nil {
+		defer os.Remove(tarball)
+		reader, err := os.Open(tarball)
+		if err != nil {
+			return errors.Wrapf(err, "failed to open downloaded plugin: %s", pkgPlugin)
+		}
+		if err := pkgPlugin.Install(reader); err != nil {
 			return errors.Wrapf(err, "failed to install plugin %s", pkgPlugin)
```
pgavlin, line 119: Just for clarification, are we downloading to a file s.t. we can retry just the download rather than the whole call to `Install`?

pgavlin (approved):

pkg/codegen/schema/loader.go
t0yv0: Yeah, not sure. Well the CI is hitting something like "connection error" when it's still reading tar contents. Checking byte count sounds like a reasonable thing to introduce, but was not able to repro, so I'm not sure if we ran this, would we hit this check or would we fail before it with "connection error" while copying the bytes, probably the later.

pkg/codegen/schema/loader.go
t0yv0: Yeah, to make it easier to test in my hacky test which I'm not checking in. I can revert.

pkg/codegen/schema/loader.go
t0yv0: Hm I wasn't sure we could retry Install but we could do it that way too guess, it may be idempotent. Conceptually it feels slightly cleaner to attribute the error to Download since it's a network error. But there's a cost to the temp file.  Let's try this but if there's reliability or disk space issues around temp files we can switch to retrying install.

pkg/codegen/schema/loader.go
t0yv0: Ah, sure, might as well.

pkg/codegen/schema/loader.go
pgavlin: SGTM

PR #7917, opened 2021-09-07
Don't fail if one subset fails

komalali:
# Description

Github actions defaults to `fail-fast: true` for build matrices. However, we'd like to see the results of all the runs even if one fails. This PR fixes the issue. 

Similar to https://github.com/pulumi/pulumi/pull/7441

## Checklist


iwahbe (approved):

t0yv0 (approved):
Ship it. It will have a bit adverse impact on throughput (mac workers) but I'm fine with trying it.

PR #7918, opened 2021-09-07
Bold in-progress diffs diffrently

### LINKED ISSUE #7337: Distinguish complete (`created`) from in-progress (`creating) tasks
When looking at pulumi console output it's often difficult to see what's in progress (green `creating`) vs what's finished (green `created`), particularly when there's a lot resources been created:

image

I'd make `creating...` in blue. This would make it really obvious to see  what's outstanding.

iwahbe:
# Description

This feature makes it easier to see which diffs are in progress vs completed. It does this by **bold**ing the colors of steps that are in progress.  

Fixes #7337

## Checklist


susanev:
@iwahbe thanks for chatting with me today.

here's my recommendation: lets leave all colors alone except for the +, -, and status.  and because of the accessibility issues here lets use the colors you picked for created and deleted (and corresponding +,-), and have creating and deleting be the same white as the other text in the table. folks will be able to more easily scan the green/white and red/white vs green/light-green and red/light-red.

what do you think?

iwahbe:
I think that folks will be more able to scan the green/neutral and red/neutral then the color mix. The problem I have is that except at stack creation and stack destruction, most commands will invoke a combination of create, delete, update and replace CRUD operations, each represented with their own color. My example didn't have that because I'm not playing with a complicated stack. If we implement your suggestion, we will get create visibility into what is in progress vs done, but will loose visibility into what type of operations are in progress. 

Right now I have it set up to split the difference. The user can choose if they want to see full color (difference between progress and pending within a color), or partial color (only completed operations are colored). 

@susanev How does that look from your end?

susanev:
lets def decide on one change together. i dont htink this situation calls for allowing folks to pick two different solutions. imma take a closer look at all the colors we are using so we can figure out a way to go with your solution but also not introduce contrast issues. ill get back to ya as soon as i can.

susanev:
im not gonna approve cause i havent reviewed the code, but based on the video you sent im good with this change. thx for doing this work!!

CHANGELOG_PENDING.md
```diff
@@ -1,5 +1,8 @@
 ### Improvements
 
+- [cli/diffs] - Add color to indicate if a action is in progress.
```
komalali, line 3:
```suggestion
- [cli] - Differentiate in-progress actions by bolding output. 
```

pkg/backend/display/diff.go
```diff
@@ -196,7 +196,7 @@ func renderSummaryEvent(action apitype.UpdateKind, event engine.SummaryEventPayl
 
 				// Print a summary of the changes of this kind
 				fprintIgnoreError(out, opts.Color.Colorize(
-					fmt.Sprintf("    %s%d %s%s%s\n", op.Prefix(), c, planTo, opDescription, colors.Reset)))
+					fmt.Sprintf("    %s%d %s%s%s\n", op.Prefix(true /* done color */), c, planTo, opDescription, colors.Reset)))
```
komalali, line 199: Nit: [our pattern](https://github.com/pulumi/pulumi/blob/master/pkg/codegen/python/gen.go#L571) is to use the function parameter name with no spaces.
```suggestion
					fmt.Sprintf("    %s%d %s%s%s\n", op.Prefix(true /*done*/), c, planTo, opDescription, colors.Reset)))
```

pkg/backend/display/progress.go
```diff
@@ -1390,8 +1390,8 @@ func (display *ProgressDisplay) getStepOp(step engine.StepEventMetadata) deploy.
 	return op
 }
 
-func (display *ProgressDisplay) getStepOpLabel(step engine.StepEventMetadata) string {
-	return display.getStepOp(step).Prefix() + colors.Reset
+func (display *ProgressDisplay) getStepOpLabel(step engine.StepEventMetadata, done bool) string {
+	return display.getStepOp(step).Prefix(done /* color done */) + colors.Reset
```
komalali, line 1394: Nit: No need when you're not passing through a true/false
```suggestion
	return display.getStepOp(step).Prefix(done) + colors.Reset
```

pkg/backend/display/rows.go
```diff
@@ -478,7 +480,7 @@ func getDiffInfo(step engine.StepEventMetadata, action apitype.UpdateKind) strin
 
 func writePropertyKeys(b io.StringWriter, keys []string, op deploy.StepOp) {
 	if len(keys) > 0 {
-		writeString(b, strings.Trim(op.Prefix(), " "))
+		writeString(b, strings.Trim(op.Prefix(true /* color done */), " "))
```
komalali, line 483: Same nit as above
```suggestion
		writeString(b, strings.Trim(op.Prefix(true /*done*/), " "))
```

pkg/cmd/pulumi/pulumi.go
```diff
@@ -106,6 +106,7 @@ func NewPulumiCmd() *cobra.Command {
 					return err
 				}
 			}
+			cmdFlag = cmd.Flag("color-level")
```
komalali, line 109: Is this a leftover?
```suggestion
```

pkg/engine/diff.go
```diff
@@ -124,7 +124,7 @@ func GetResourcePropertiesSummary(step StepEventMetadata, indent int) string {
 	writeString(&b, getIndentationString(indent, op, false))
 
 	// First, print out the operation's prefix.
-	writeString(&b, op.Prefix())
+	writeString(&b, op.Prefix(true /* done color */))
```
komalali, line 127:
```suggestion
	writeString(&b, op.Prefix(true /*done*/))
```

sdk/go/common/diag/colors/colors.go
```diff
@@ -196,7 +200,7 @@ var (
 	Underline = command("underline")
 )
 
-// Basic colors.
+// Basic
```
komalali, line 203: Curious why you removed the ` colors.` from the comment here?

komalali (commented):
This looks great @iwahbe ! Just a couple of minor clean-up questions and then should be good to go.

iwahbe:
@komalali Thanks for pointing out the convention. 
> Is this a leftover?
This was a good catch!
> Curious why you removed the colors. from the comment here?
Because it made less sense in an earlier commit. I should have restored it. (and now I did)

justinvp:
I haven't had a chance to try this out. Mind sharing the video, or a screenshot, or https://asciinema.org?

iwahbe:
> I haven't had a chance to try this out. Mind sharing the video, or a screenshot, or https://asciinema.org?
I'd be happy to. 
https://user-images.githubusercontent.com/22222529/135171528-b71759bf-cfc5-4a6e-84fc-b00425aad17a.mov

komalali (approved):
LGTM!

Issue #7890, opened 2021-09-01
Go program hangs when previewing


t0yv0:
## Steps to reproduce

```go
package main

import (
	aws_eks "github.com/pulumi/pulumi-aws/sdk/v4/go/aws/eks"
	"github.com/pulumi/pulumi-aws/sdk/v4/go/aws/iam"
	"github.com/pulumi/pulumi-eks/sdk/go/eks"
	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
)

func main() {
	pulumi.Run(func(ctx *pulumi.Context) error {
		args := &eks.ClusterArgs{}

		args.SkipDefaultNodeGroup = pulumi.Bool(true)
		args.VpcId = pulumi.String("doesntmatter")

		cluster, err := eks.NewCluster(ctx, "example", args)
		if err != nil {
			return err
		}

		_, err = aws_eks.NewNodeGroup(ctx,
			"example-ng",
			&aws_eks.NodeGroupArgs{
				ClusterName:   pulumi.String("example"),
				InstanceTypes: pulumi.ToStringArray([]string{"t3.micro"}),
				SubnetIds:     pulumi.ToStringArray([]string{"doesntmatter"}),
				ScalingConfig: &aws_eks.NodeGroupScalingConfigArgs{
					DesiredSize: pulumi.Int(1),
					MaxSize:     pulumi.Int(1),
					MinSize:     pulumi.Int(1),
				},
				NodeRoleArn: cluster.InstanceRoles.
					ApplyT(func(roles *[]iam.Role) pulumi.StringOutput {
						return (*roles)[0].Arn
					}).
					ApplyT(func(x interface{}) string {
						return x.(string)
					}).(pulumi.StringOutput),
			},
			pulumi.Parent(cluster))

		if err != nil {
			return err
		}

		return nil
	})
}

```

Expected: preview completes
Actual: preview hangs

Frassle:
Updating the code for the current sdk:
```
package main

import (
	aws_eks "github.com/pulumi/pulumi-aws/sdk/v5/go/aws/eks"
	"github.com/pulumi/pulumi-aws/sdk/v5/go/aws/iam"
	"github.com/pulumi/pulumi-eks/sdk/go/eks"
	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
)

func main() {
	pulumi.Run(func(ctx *pulumi.Context) error {
		args := &eks.ClusterArgs{}

[...]

PR #7925, opened 2021-09-08
[sdk/nodejs] Marshal output values

### LINKED ISSUE #7850: Node SDK marshalling (used in SDK)
Part of #7434

          consumed by the engine
          a Pulumi Output

justinvp:
This change adds support for marshaling outputs as output values in the Node.js SDK.

Fixes #7850

Part of #7434

### DIFF
--- a/sdk/nodejs/runtime/resource.ts
+++ b/sdk/nodejs/runtime/resource.ts
@@ -1,4 +1,4 @@
-// Copyright 2016-2018, Pulumi Corporation.
+// Copyright 2016-2021, Pulumi Corporation.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -205,7 +205,7 @@ export function readResource(res: Resource, t: string, name: string, props: Inpu
 
     const preallocError = new Error();
     debuggablePromise(resopAsync.then(async (resop) => {
-        const resolvedID = await serializeProperty(label, id, new Set());
+        const resolvedID = await serializeProperty(label, id, new Set(), { keepOutputValues: false });
         log.debug(`ReadResource RPC prepared: id=${resolvedID}, t=${t}, name=${name}` +
             (excessiveDebugOutput ? `, obj=${JSON.stringify(resop.serializedProps)}` : ``));
 
@@ -510,7 +510,11 @@ async function prepareResource(label: string, res: Resource, custom: boolean, re
 
         // Serialize out all our props to their final values.  In doing so, we'll also collect all
         // the Resources pointed to by any Dependency objects we encounter, adding them to 'propertyDependencies'.
-        const [serializedProps, propertyToDirectDependencies] = await serializeResourceProperties(label, props);
+        const [serializedProps, propertyToDirectDependencies] = await serializeResourceProperties(label, props, {
+            // To initially scope the use of this new feature, we only keep output values when
+            // remote is true (for multi-lang components).
+            keepOutputValues: remote,
+        });
 
         // Wait for the parent to complete.
         // If no parent was provided, parent to the root resource.
@@ -720,7 +724,7 @@ async function resolveOutputs(res: Resource, t: string, name: string,
                 // input prop the engine didn't give us a final value for.  Just use the value passed into the resource
                 // after round-tripping it through serialization. We do the round-tripping primarily s.t. we ensure that
                 // Output values are handled properly w.r.t. unknowns.
-                const inputProp = await serializeProperty(label, props[key], new Set());
+                const inputProp = await serializeProperty(label, props[key], new Set(), { keepOutputValues: false });
                 if (inputProp === undefined) {
                     continue;
                 }
--- a/sdk/nodejs/runtime/resource.ts
+++ b/sdk/nodejs/runtime/resource.ts
@@ -600,7 +600,8 @@ function addAll<T>(to: Set<T>, from: Set<T>) {
     }
 }
 
-async function getAllTransitivelyReferencedResourceURNs(resources: Set<Resource>): Promise<Set<string>> {
+/** @internal */
+export async function getAllTransitivelyReferencedResourceURNs(resources: Set<Resource>): Promise<Set<string>> {
     // Go through 'resources', but transitively walk through **Component** resources, collecting any
     // of their child resources.  This way, a Component acts as an aggregation really of all the
     // reachable resources it parents.  This walking will stop when it hits custom resources.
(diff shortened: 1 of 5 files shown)


sdk/nodejs/runtime/resource.ts
```diff
@@ -510,7 +510,11 @@ async function prepareResource(label: string, res: Resource, custom: boolean, re
 
         // Serialize out all our props to their final values.  In doing so, we'll also collect all
         // the Resources pointed to by any Dependency objects we encounter, adding them to 'propertyDependencies'.
-        const [serializedProps, propertyToDirectDependencies] = await serializeResourceProperties(label, props);
+        const [serializedProps, propertyToDirectDependencies] = await serializeResourceProperties(label, props, {
+            // To initially scope the use of this new feature, we only keep output values when
+            // remote is true (for multi-lang components).
+            keepOutputValues: remote,
```
justinvp, line 516: I've gone back-and-forth on whether the SDK should always keep output values (if the monitor supports it), leaving it to the engine to handle whether output values should be kept when unmarshaling.

However, to really scope and de-risk the change, for now, in the SDKs, if the monitor supports it, we'll only keep output values when marshaling values for `RegisterResource` when `remote` is `true` and for `Call`. (Additionally, initially, the engine will only keep output values during unmarshaling for `RegisterResource` when `remote` is true, and for `Call`). In the future, if/when we adopt output values more broadly we can update the SDKs to keep output values in more cases, and similarly have the engine do the same.

sdk/nodejs/runtime/rpc.ts
```diff
@@ -340,7 +361,34 @@ export async function serializeProperty(ctx: string, prop: Input<any>, dependent
...
+        });
+
+        if (opts?.keepOutputValues && await monitorSupportsOutputValues()) {
+            const dependencies = new Set<string>();
+            for (const resource of propResources) {
+                const urn = await serializeProperty(`${ctx} dependency`, resource.urn, new Set<Resource>(), {
+                    keepOutputValues: false,
+                });
+                dependencies.add(urn);
+            }
```
justinvp, line 375: I don't think this is right, but just want to confirm.

@pgavlin, we need to do the same dependency expansion here, as we do here:

https://github.com/pulumi/pulumi/blob/c247f6e283fb94cbce5a3e73ebe226e34ed99d31/sdk/nodejs/runtime/resource.ts#L552-L558

Correct?

Otherwise, the property dependencies for a property may contain more dependencies (expanded) than the union of all nested output value dependencies for the property, when those sets should be equal.

sdk/nodejs/runtime/resource.ts
t0yv0: The concern is roll-out order? So basically we need all our clients to update to new monitor that understands the extended protocol before SDKs can assume it and use the new protocol?

sdk/nodejs/tests/runtime/props.spec.ts
```diff
@@ -131,6 +132,55 @@ describe("runtime", () => {
     });
 
     describe("transferProperties", () => {
+        describe("output values", () => {
```
t0yv0, line 135: Nice!

sdk/nodejs/tests/runtime/props.spec.ts
```diff
@@ -131,6 +132,55 @@ describe("runtime", () => {
...
+                            {
+                                const resources = deps.map(dep => new DependencyResource(dep));
+                                yield {
+                                    name: `Output(${JSON.stringify(deps)}, ${JSON.stringify(tv.value)}, ` +
+                                        `isKnown=${isKnown}, isSecret=${isSecret})`,
+                                    input: new Output(resources, Promise.resolve(tv.value), Promise.resolve(isKnown),
+                                        Promise.resolve(isSecret), Promise.resolve([])),
+                                    expected: {
+                                        [runtime.specialSigKey]: runtime.specialOutputValueSig,
+                                        ...isKnown && { value: tv.expected },
```
t0yv0, line 160: Haven't seen this ... TS feature, nice!

t0yv0 (approved):

sdk/nodejs/runtime/resource.ts
justinvp: > The concern is roll-out order?

No. The feature support checking will take care of this. SDKs will only send output values if the monitor says it supports them, and then engine will only pass output values to a provider if it says it accepts them. And even when accepting output values, if the code that's accepting the output values doesn't have `KeepOutputValues` specified when unmarshaling, it will get back the same values as previously.

I'm mainly concerned that output values will somehow show up in places that can't fully accommodate them (yet), both (1) within an SDK and (2) within the engine. Perhaps I'm being overly paranoid, but given the short timeframe for rolling this out, and being involved with a similar feature (resource references) that was enabled more broadly and had a bit of a bug tail, I'm inclined to really scope down where the feature is enabled as much as possible initially.

I'm less concerned about (1), because I've done a scrub of the places where we marshal and keep output values in the SDKs. For example, there are places in the SDK where we roundtrip inputs (i.e. marshal and then unmarshal), e.g. for mocks-based testing. In that case, since the mock monitor doesn't support the feature, it should all just work as expected.

Slightly more concerned about (2). When unmarshaling the RPC values, if `KeepOutputValues` isn't true, we should get the exact same behavior as before (any output values will be unmarshaled in the old way) so we should be covered. However, if somehow output values _did_ show up, such as enabling it for _all_ `RegisterResource` calls, I'm concerned the output values might somehow mess up areas that haven't been looked at closely, such as diffs, where previously `resource.Output` was treated as unknown.

Again, these concerns are theoretical because I don't think we actually have any problems here, I'm just aiming to reduce risk.

sdk/nodejs/runtime/rpc.ts
justinvp: I'm going to go ahead and add a commit that does the expansion.

t0yv0:
Approval stands but exercise caution around dependency information. I see no obvious issue with the changes, but it's a high impact area where I'm awaiting finalized spec plus have two outstanding P1-ish fixes around looping in SDK dependency expansion for Go and C# (Python P1 got fixed), here's the Go one:

https://github.com/pulumi/pulumi/issues/7890 

Node was exempt from the P1 because it implements the resource differently (in-process), but might still be liable.

sdk/nodejs/runtime/rpc.ts
```diff
@@ -340,7 +362,44 @@ export async function serializeProperty(ctx: string, prop: Input<any>, dependent
...
+                await serializeProperty(`${ctx} dependency`, resource.urn, urnDeps, {
+                    keepOutputValues: false,
+                });
+            }
+            for (const resource of urnDeps) {
+                propResources.add(resource);
+                dependentResources.add(resource);
+            }
+
+            const dependencies = await getAllTransitivelyReferencedResourceURNs(propResources);
```
pgavlin, line 386: Excellent catch here.

pgavlin (approved):

PR #7926, opened 2021-09-08
[sdk/python] Marshal output values

### LINKED ISSUE #7851: Python SDK marshalling (used in SDK)
Part of #7434

          consumed by the engine
          a Pulumi Output

justinvp:
This change adds support for marshaling outputs as output values in the Python SDK.

Fixes #7851

Related to #7925
Part of #7434

### DIFF
--- a/sdk/python/lib/pulumi/runtime/rpc.py
+++ b/sdk/python/lib/pulumi/runtime/rpc.py
@@ -40,19 +40,40 @@ UNKNOWN = "04da6b54-80e4-46f7-96ec-b56ff0331ba9"
 """If a value is None, we serialize as UNKNOWN, which tells the engine that it may be computed later."""
 
 _special_sig_key = "4dabf18193072939515e22adb298388d"
-"""_special_sig_key is sometimes used to encode type identity inside of a map. See pkg/resource/properties.go."""
+"""
+_special_sig_key is sometimes used to encode type identity inside of a map.
+See sdk/go/common/resource/properties.go.
+"""
 
 _special_asset_sig = "c44067f5952c0a294b673a41bacd8c17"
-"""special_asset_sig is a randomly assigned hash used to identify assets in maps. See pkg/resource/asset.go."""
+"""
+special_asset_sig is a randomly assigned hash used to identify assets in maps.
+See sdk/go/common/resource/asset.go.
+"""
 
 _special_archive_sig = "0def7320c3a5731c473e5ecbe6d01bc7"
-"""special_archive_sig is a randomly assigned hash used to identify assets in maps. See pkg/resource/asset.go."""
+"""
+special_archive_sig is a randomly assigned hash used to identify assets in maps.
+See sdk/go/common/resource/asset.go.
+"""
 
 _special_secret_sig = "1b47061264138c4ac30d75fd1eb44270"
-"""special_secret_sig is a randomly assigned hash used to identify secrets in maps. See pkg/resource/properties.go"""
+"""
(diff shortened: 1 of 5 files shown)


sdk/python/lib/pulumi/runtime/rpc.py
```diff
@@ -260,7 +287,28 @@ async def serialize_property(value: 'Input[Any]',
...
         is_secret = await output._is_secret
-        value = await serialize_property(output.future(), deps, input_transformer, typ)
+        value = await serialize_property(output.future(), deps, input_transformer, typ, keep_output_values=False)
+
+        if keep_output_values and await settings.monitor_supports_output_values():
+            # TODO expand the dependencies
+            dependencies: Set[str] = set()
+            for resource in value_resources:
+                urn = await serialize_property(resource.urn, deps, input_transformer, keep_output_values=False)
+                dependencies.add(cast(str, urn))
```
justinvp, line 297: Similar question here as https://github.com/pulumi/pulumi/pull/7925/files#discussion_r704783076

sdk/python/lib/test/test_next_serialize.py
```diff
@@ -1434,3 +1435,42 @@ def transformer(key: str) -> str:
...
+                    obj["secret"] = is_secret
+                if deps:
+                    obj["dependencies"] = deps
+
+                inputs = {"value": Output(resources, future(value), future(is_known), future(is_secret))}
+                expected = struct_pb2.Struct()
+                expected["value"] = obj
+
+                actual = await rpc.serialize_properties(inputs, {}, keep_output_values=True)
+                self.assertDictEqual(json_format.MessageToDict(expected), json_format.MessageToDict(actual))
```
t0yv0, line 1471: Didn't use assertDictEqual much before. This is a deep comparison right?

sdk/python/lib/pulumi/runtime/rpc.py
```diff
@@ -226,13 +252,14 @@ async def serialize_property(value: 'Input[Any]',
...
+            obj["assets"] = await serialize_property(asset_archive.assets, deps, input_transformer,
+                keep_output_values=False)
         elif hasattr(value, "path"):
             file_archive = cast('FileArchive', value)
-            obj["path"] = await serialize_property(file_archive.path, deps, input_transformer)
+            obj["path"] = await serialize_property(file_archive.path, deps, input_transformer, keep_output_values=False)
         elif hasattr(value, "uri"):
             remote_archive = cast('RemoteArchive', value)
-            obj["uri"] = await serialize_property(remote_archive.uri, deps, input_transformer)
+            obj["uri"] = await serialize_property(remote_archive.uri, deps, input_transformer, keep_output_values=False)
```
t0yv0, line 262: False intentional everywhere? Would it matter if it was `keep_output_values` from the param for symmetry?

t0yv0 (approved):
Props for table-driven tests. Exercise caution around dependency expansion.

pgavlin (commented):
LGTM overall. Not much to add beyond what Anton mentioned and your note re: expanding dependencies.

sdk/python/lib/pulumi/runtime/rpc.py
justinvp: Actually, I'm going to do this as a quick follow-up. The resource cycle breaker added in #7887 makes adopting this inside the RPC code a little more difficult because it needs a `from_resource: 'Resource'`. I opened https://github.com/pulumi/pulumi/issues/7977 and updated the TODO to link to the issue.

PR #7933, opened 2021-09-09
Added a buildkite detector for detecting the correct env vars in CI

arwilczek90:
# Description

This adds a custom CI detector for the Buildkite CI/CD system.

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,4 +1,5 @@
 ### Improvements
+* Adds CI detector for Buildkite
 
 ### Bug Fixes
 
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,5 +1,5 @@
 ### Improvements
-* Adds CI detector for Buildkite
+* Adds CI detector for Buildkite [#7933](https://github.com/pulumi/pulumi/pull/7933)
 
 ### Bug Fixes
 
--- a/sdk/go/common/util/ciutil/buildkite.go
+++ b/sdk/go/common/util/ciutil/buildkite.go
@@ -1,4 +1,4 @@
-// Copyright 2016-2019, Pulumi Corporation.
+// Copyright 2016-2021, Pulumi Corporation.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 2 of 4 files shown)


sdk/go/common/util/ciutil/buildkite.go
```diff
@@ -0,0 +1,55 @@
...
+	v.BuildID = os.Getenv("BUILDKITE_BUILD_ID")
+	// https://buildkite.com/docs/pipelines/environment-variables#bk-env-vars-buildkite-build-number
+	v.BuildNumber = os.Getenv("BUILDKITE_BUILD_NUMBER")
+	// https://buildkite.com/docs/pipelines/environment-variables#bk-env-vars-buildkite-build-url
+	v.BuildURL = os.Getenv("BUILDKITE_BUILD_URL")
+	// https://buildkite.com/docs/pipelines/environment-variables#bk-env-vars-buildkite-message
+	// This is usually the commit message but can be other messages.
+	v.CommitMessage = os.Getenv("BUILDKITE_MESSAGE")
+	// https://buildkite.com/docs/pipelines/environment-variables#bk-env-vars-buildkite-pull-request
+	v.PRNumber = os.Getenv("BUILDKITE_PULL_REQUEST")
```
praneetloke, line 41: The Buildkite docs state `BUILDKITE_PULL_REQUEST` will be `"false"` for non-PR builds. Can you please add a check here to skip setting the `PRNumber` property if the value is the string `"false"`?

praneetloke: Adding this check will also make sure that `PRNumber` is indeed an empty string for non-PR builds which you are checking for at https://github.com/pulumi/pulumi/pull/7933/files#diff-a12ba82451d68131c5f6bb2e2546494b3cbf954c5aac38f7ee575b982134d228R45.

sdk/go/common/util/ciutil/buildkite.go
```diff
@@ -0,0 +1,55 @@
...
+	// https://buildkite.com/docs/pipelines/environment-variables#bk-env-vars-buildkite-commit
+	v.SHA = os.Getenv("BUILDKITE_COMMIT")
+
+	// If PRNumber is set to anything but empty string its a pull request, else use the value supplied by buildkite
+	// This is manually set because to buildkite a PR is a webhook which is the same as a manual push.
```
praneetloke, line 46:
```suggestion
	// This is manually set because to Buildkite a PR is a webhook which is the same as a manual push.
```

praneetloke (changes requested):

CHANGELOG_PENDING.md
```diff
@@ -1,4 +1,5 @@
 ### Improvements
+* Adds CI detector for Buildkite
```
emiliza, line 2: Please also update this to include a link to the pull request (look at CHANGELOG.md for reference)

sdk/go/common/util/ciutil/buildkite.go
arwilczek90: I've fixed this and also combined the if statements to make the logic easier.

sdk/go/common/util/ciutil/buildkite.go
arwilczek90: fixed, thanks for spotting this.

CHANGELOG_PENDING.md
arwilczek90: Done, thank you.

praneetloke (approved):

t0yv0 (approved):

Issue #7435, opened 2021-07-06
Single return values for methods and functions


justinvp:
**Update:** #8111 introduced a language flag (`liftSingleValueMethodReturns`) that can be set to `true` in the schema for each language to opt-in to "lifting" single-valued method returns to their return type. While this doesn't implement the design outlined in this issue, the opt-in flag is a practical change to opt-in to this behavior in the meantime, which doesn't prevent us from doing the full fix discussed in this issue, which we still would like to do. The downside is that it applies to _all_ methods in the schema.

---

Currently, methods and functions that return single values must return the value as part of a `Result` type.

For example, an `eks.Cluster` component that has a `getKubeconfig` method that returns a string value might be defined in the schema like:

```json
{
    "resources": {
        "eks:index:Cluster": {
            "isComponent": true,
            "methods": {
                "getKubeconfig": "eks:index:Cluster/getKubeconfig"
            }
        }
    },
    "functions": {
        "eks:index:Cluster/getKubeconfig": {
            "inputs": {
                "properties": {
                    "__self__": {
                        "$ref": "#/resources/eks:index:Cluster"
                    },
                    "profileName": {
                        "type": "string"
                    },
                    "roleArn": {
                        "type": "string"
                    }
                },
                "required": ["__self__"]
            },
            "outputs": {
                "properties": {
                    "value": {
                        "type": "string"
                    }
                },
                "required": ["value"]
            }
        }
    }
}
```

With resulting Python codegen that looks something like:

```python
class Cluster(pulumi.ComponentResource):
    @pulumi.output_type
    class GetKubeconfigResult:
        @property
        @pulumi.getter
        def value(self) -> str:
            return pulumi.get(self, "value")

    def get_kubeconfig(self,
                       profile_name: Optional[pulumi.Input[str]] = None,
                       role_arn: Optional[pulumi.Input[str]] = None) -> pulumi.Output['Cluster.GetKubeconfigResult']:
        ...
```

This isn't an ideal object model. Ideally, there'd be a way to specify that `getKubeconfig` has a single return value, so that there'd be no need for the `GetKubeconfigResult` class and the method could return `pulumi.Output[str]` directly rather than `pulumi.Output['Cluster.GetKubeconfigResult']`:

```python
class Cluster(pulumi.ComponentResource):
    def get_kubeconfig(self,
                       profile_name: Optional[pulumi.Input[str]] = None,
[...]

PR #7937, opened 2021-09-09
 Support "lifting" single-valued method returns to their return type

jen20:
# Description

This commit allows lifting single valued method returns rather than using an output structure in each of the three languages we are currently targeting.

## Checklist

No current test coverage - I think this can be added to method tests however.

Not a user-facing change

N/A

### DIFF
--- a/pkg/codegen/go/gen.go
+++ b/pkg/codegen/go/gen.go
@@ -108,6 +108,9 @@ type pkgContext struct {
 	// Name overrides set in GoPackageInfo
 	modToPkg         map[string]string // Module name -> package name
 	pkgImportAliases map[string]string // Package name -> import alias
+
+	// Determines whether to make single-return-value methods return an output struct or the value
+	liftSingleValueMethodReturns bool
 }
 
 func (pkg *pkgContext) detailsForType(t schema.Type) *typeDetails {
@@ -1529,6 +1532,8 @@ func (pkg *pkgContext) genResource(w io.Writer, r *schema.Resource, generateReso
 		methodName := Title(method.Name)
 		f := method.Function
 
+		shouldLiftReturn := pkg.liftSingleValueMethodReturns && f.Outputs != nil && len(f.Outputs.Properties) == 1
+
 		var args []*schema.Property
 		if f.Inputs != nil {
 			for _, arg := range f.Inputs.InputShape.Properties {
@@ -1547,6 +1552,8 @@ func (pkg *pkgContext) genResource(w io.Writer, r *schema.Resource, generateReso
 		var retty string
 		if f.Outputs == nil {
 			retty = "error"
+		} else if shouldLiftReturn {
+			retty = fmt.Sprintf("(%s, error)", pkg.outputType(f.Outputs.Properties[0].Type))
 		} else {
 			retty = fmt.Sprintf("(%s%sResultOutput, error)", name, methodName)
 		}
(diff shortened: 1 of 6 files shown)


pkg/codegen/python/gen.go
```diff
@@ -1528,14 +1539,24 @@ func (mod *modContext) genMethods(w io.Writer, res *schema.Resource) {
...
-			ret = "return "
 		}
-		fmt.Fprintf(w, "        %spulumi.runtime.call('%s', __args__, res=__self__%s)\n", ret, fun.Token, typ)
+
+		if method.Function.Outputs == nil {
+			fmt.Fprintf(w, "        pulumi.runtime.call('%s', __args__, res=__self__%s)\n", fun.Token, typ)
+		} else if shouldLiftReturn {
+			// Store the return in a variable and return the property output
+			fmt.Fprintf(w, "        res = pulumi.runtime.call('%s', __args__, res=__self__%s)\n", fun.Token, typ)
+			fmt.Fprintf(w, "        return res.%s\n", PyName(fun.Outputs.Properties[0].Name))
```
justinvp, line 1554: Nit: It's possible (however unlikely) that someone could have `res` as a parameter name. Using an even more unlikely name like `__result__` would help mitigate.

No need to change it in your PR. I'm going to follow-up soon with a change that adds some tests and adds .NET, and while doing that, I'll consider tweaking this.

pkg/codegen/python/gen.go
```diff
@@ -1528,14 +1539,24 @@ func (mod *modContext) genMethods(w io.Writer, res *schema.Resource) {
 		}
 
 		// Now simply call the function with the arguments.
-		var typ, ret string
+		var typ string
 		if retTypeNameQualified != "" {
```
justinvp, line 1543: No need for a change to this PR, just point out: When `shouldLiftReturn` is true, we could consider prefixing the nested result class with an underscore to make it more of an implementation detail, or even consider emitting it inside the method itself to further hide it. But it doesn't matter much in the grand scheme of things, so I'm inclined to keep it as-is for now.

justinvp (approved):
LGTM, thanks! I tried these changes out on our EKS library, and they look good to me.

I'll follow-up with a change that adds a similar change for .NET (which we'll need), along with some tests. I'll also look into adjusting the generated docs to show the appropriate return type.

Note that while this doesn't implement the design as discussed in #7435 (changing `FunctionSpec.Outputs` to accept `*TypeSpec`, and all the fall-out from that), doing this via language flag is a good practical change to opt-in to this behavior in the meantime, and doesn't prevent us from doing the full fix for #7435 in the future.

pkg/codegen/python/gen.go
jen20: Yes, that's a good point and not one I'd considered. A prefix is definitely a better idea here.

pkg/codegen/go/gen.go
```diff
@@ -1682,7 +1708,11 @@ func (pkg *pkgContext) genResource(w io.Writer, r *schema.Resource, generateReso
 	fmt.Fprintf(w, "\tpulumi.RegisterOutputType(%sOutput{})\n", name)
 	for _, method := range r.Methods {
 		if method.Function.Outputs != nil {
-			fmt.Fprintf(w, "\tpulumi.RegisterOutputType(%s%sResultOutput{})\n", name, Title(method.Name))
+			if len(method.Function.Outputs.Properties) == 1 {
```
justinvp, line 1711: One of the existing tests failed because it's missing checking `pkg.liftSingleValueMethodReturns` here.

```
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -178,4 +178,4 @@
        	            	 	pulumi.RegisterOutputType(FooOutput{})
        	            	-	pulumi.RegisterOutputType(FooBarResultOutput{})
        	            	-	pulumi.RegisterOutputType(FooGenerateKubeconfigResultOutput{})
        	            	+	pulumi.RegisterOutputType(fooBarResultOutput{})
        	            	+	pulumi.RegisterOutputType(fooGenerateKubeconfigResultOutput{})
        	            	 }
        	Test:       	TestGeneratePackage/Simple_schema_with_methods
        	Messages:   	example/foo.go
Error:     helpers.go:121: example/foo.go did not agree
    --- FAIL: TestGeneratePackage/Simple_schema_with_methods (0.05s)
```

justinvp:
Replaced by #8111

PR #7938, opened 2021-09-09
Emit schema.Package.Version when possible

iwahbe:
# Description

We currently offload adding the `Package.Version` information to makefiles in our SDKs. This means that package information which is specified in a `schema.json` is ignored in favor of another process. This PR has version information emitted if and only if `Package.Version` is supplied. This means
- `version.txt` for `dotnet`
- `setup.py` accurately contains a version for `python`
- `package.json` accurately reports it's version for `nodejs`.

If `Package.Version` is not supplied, no behavior is changed. 

This has 2 major motivations:
1. Using the schema version field should populate to the project as expected. Having the version set to `"0.0.0"` is not expected behavior. 
2. We should attempt to emit valid SDKs when possible. Running `dotnet build` requires `version.txt` to exist. This prevents special cases from developing SDKs and testing our SDK codegen. 

## Pending Questions

- [ ] How to handle the distinction python makes between `VERSION` and `PLUGIN_VERSION`

## Checklist

- [ ] Decide if we should use `npm` or `yarn` for testing (discussed for #7989).


pkg/codegen/dotnet/gen.go
```diff
@@ -1979,6 +1979,13 @@ func genPackageMetadata(pkg *schema.Package, assemblyName string, packageReferen
 
 	files.add(assemblyName+".csproj", projectFile)
 	files.add("logo.png", logo)
+
+	// Generate version.txt
+	versionFile := "version.txt"
+	if _, exists := files[versionFile]; !exists && pkg.Version != nil {
+		files[versionFile] = []byte(pkg.Version.String())
```
pgavlin, line 1986: Is this to handle the case that the caller has already added a `version.txt`?

pkg/codegen/python/gen.go
```diff
@@ -1896,9 +1896,18 @@ func genPackageMetadata(
...
+	// If `pkg` supplies a version, use that. Otherwise, create a constant for
+	// the version number to replace during build.
+	pypiVersion := "0.0.0"
+	pkgVersion := "0.0.0"
+	if pkg.Version != nil {
+		pkgVersion = pkg.Version.String()
+	}
+	// TODO: puipyVersion is the pythonic version of pkgVersion. The code to
+	// transform it is in pulumi/pulumictl. Should that be extracted to
+	// somewhere like pulumi/pulumi/sdk/python.go?
```
pgavlin, line 1908: Yes, it does seem like we should extract that code and put it somewhere common. Not sure if `pulumi/pulumi/sdk` is the right place for that, but I also don't have a better suggestion offhand :) @komalali  @t0yv0 ?

pkg/codegen/dotnet/gen.go
iwahbe: That was my intention, specifically if someone passes a `version.txt` to `GeneratePackage` through `extraFiles`. 

BTW: this is unique for `.NET` because every other generator already has the associated file, and we are just slotting in the version.

pkg/codegen/python/gen.go
t0yv0: No strong opinions.. looks like pulumictl depends on pkg https://github.com/pulumi/pulumictl/blob/master/go.mod#L11 so if the least common ancestor is pkg we can put it there; alternatively factor into its own little dedicated GH repo and package. The barriers to that are pleasantly low. If it's really small then "a little dulication is better than a little dependency" applies here too so we can inline a copy.

t0yv0:
SO I dug around how these versions are used and filed https://github.com/pulumi/pulumi/issues/7955 with questions. The fact Python (and Node?) auto-install Go plugins binaries in their setup.py strikes me as a bit of an unfortunate choice.

Presumably this is the distinction:

```How to handle the distinction python makes between VERSION and PLUGIN_VERSION``` 

The `PLUGIN_VERSION` determines which Go plugin to auto-install.

**Comment by t0yv0** [comment]
I like where this is going since Makefile-based manipulations of VERSION were giving me some headache. Before we ship this it would be good for @mikhailshilkov to make sure this causes no extra problems for provider builds.

**Comment by mikhailshilkov** [comment]
The reason we put this version placeholder initially was to avoid version-related diffs on every commit. Has this been mitigated somehow? Adding @stack72 to chime in too.

**Comment by iwahbe** [comment]
> The reason we put this version placeholder initially was to avoid version-related diffs on every commit. Has this been mitigated somehow? Adding @stack72 to chime in too.

@mikhailshilkov  The mitigation is that if you don't specify version in `schema.json`, we will still generate a placeholder `"version": "${VERSION"` in `package.json`. This change has no effect unless a version is supplied at generation time.

**Comment by mikhailshilkov** [comment]
@iwahbe Thank you for the clarification! Do I understand correctly that this is mostly to clean up the tests but likely won't affect the providers?

**Comment by iwahbe** [comment]
> @iwahbe Thank you for the clarification! Do I understand correctly that this is mostly to clean up the tests but likely won't affect the providers?

It is primarily about cleaning up tests and internal consistency. It should also help decrease the amount of makefile necessary for multi-lang.

**Comment by infin8x** [comment]
This looks great! We'd talked about doing something similar in https://github.com/pulumi/pulumi/issues/7589 - am I understanding correctly that this PR achieves the "set the version in a single location" part of that issue?

**Comment by iwahbe** [comment]
> This looks great! We'd talked about doing something similar in #7589 - am I understanding correctly that this PR achieves the "set the version in a single location" part of that issue?

@infin8x Yes. This has you set the version in the schema, and it propagates it everywhere.

**Comment by iwahbe** [comment]
> The reason we put this version placeholder initially was to avoid version-related diffs on every commit. Has this been mitigated somehow? Adding @stack72 to chime in too.

Also responding to @t0yv0, it seems like the correct behavior is to set `PLUGIN_VERSION` to a python specific version of `version.Version`. This doesn't work well for testing, since `version.Version` changes with each commit. I can think of several solutions, none of which I love.
1. We could set it to the previous version release (which would work but break all tests each release).
2. We could detect the test environment and spit out `"0.0.0"` only in test, otherwise giving the desired behavior.
3. We could punt, leaving this `"0.0.0"`

Personally I lean towards 2, but I'm unsure. Does anyone see a better solution?

**Comment by pgavlin** [comment]
> This doesn't work well for testing, since version.Version changes with each commit. I can think of several solutions, none of which I love.

Note that this is only the case for workflows that do this explicitly by using `ldflags` to set `version.Version`. These workflows are likely to be similar to our own gitops-style workflow where the schema version will be unset. It feels like if the schema version is set, then it should align with `version.Version` and we should be able to emit a real value for `PLUGIN_VERSION`.

**Comment by t0yv0** [comment]
@iwahbe I'm not really super sure, how does our system do it in practice?

I peeked at https://github.com/pulumi/pulumi-aws/blob/master/sdk/python/setup.py#L11 and I see 0.0.0. That seems to work?

Also in my investigation https://github.com/pulumi/pulumi/issues/7955 it turns out that Node JS packages are the *other* kind of packages that auto-install the plugin, yet they have no PULUMI_VERSION distinction? How do they do it? I can't seem to find the mechanism of how we hook into npm install.

**Comment by iwahbe** [comment]
> I peeked at https://github.com/pulumi/pulumi-aws/blob/master/sdk/python/setup.py#L11 and I see 0.0.0. That seems to work?

Taking a look around [pulumi-aws](https://github.com/pulumi/pulumi-aws) I saw what you did, but when ran `make build` in the aws repo, a `setup.py` file was generated with `VERSION = "4.20.0"` and `PLUGIN_VERSION = "4.20.0"`. That gets setup [here](https://github.com/pulumi/pulumi-aws/blob/c3fc1c5696d95dfcabb59a822c2460680a2c0f29/Makefile#L51).

**Comment by iwahbe** [comment]
At this point I'm fairly convinced that the correct behavior (when not testing) is to create a pythonic version of `version.Version` and stick it in `PLUGIN_VERSION`. This accommodates our current setup, where we mirror the package version with the pypi release version, only changing it to accommodate the idiosyncrasies of pypi. This is what the current commit does. 

When testing, we emit `"0.0.0"`, because we don't want to see diffs on our commit hashes in `codegen/internal/tests`. Also, `version.Version == ""` under test.

File: pkg/codegen/python/gen.go (hunk 2 above)
Comment (iwahbe) [reply pkg/codegen/python/gen.go]: I just duplicated the code. 
Taking a closer look, most of the code dealt with the current git tree being dirty, and generating a semver version. Since we are provided a semver version already, I just help with the pythonic parts. It's not long. I located it in `pulumi/pkg/codegen/python/utilities.go`.

File: pkg/codegen/python/gen.go [hunk 3]
```diff
@@ -1896,9 +1897,20 @@ func genPackageMetadata(
...
+	// the version number to replace during build.
+	packageVersion := "0.0.0"
+	if pkg.Version != nil {
+		packageVersion = pythonVersion(pkg.Version)
+	}
+	pluginVersion := "0.0.0"
+	if version.Version != "" {
+		// This happens in test builds
+		pluginVersion = version.Version
+	}
```
pgavlin, line 1910: I don't think that we want to use `version.Version` here: `"github.com/pulumi/pulumi/pkg/v3/version"` is the version of the Pulumi SDK itself, and is unrelated to the schema version. Should this be

```suggestion
	pluginVersion := "0.0.0"
	if pkg.Version != nil {
		pluginVersion = pkg.Version.String()
	}
```

iwahbe: Your right. I got the version of the `pulumi-resource-*` confused with the version of `pulumi` itself.

t0yv0 (approved):

iwahbe:
@t0yv0 I want to make sure the tests pass, then I'll merge.

PR #7943, opened 2021-09-10
Do not generate Go inputty types for output-only schema types

mikhailshilkov:
These changes prevent the Go SDK generator from generating input
types that are never referenced by rest of the generated SDK. For
example, if an object type never appears in a resource input property,
these changes will prevent an input type from being generated for
that object type. In further pursuit of this goal, these changes prevent
the generation of input/output types for function inputs and arguments
if function output versions are disabled or if the function does not need
an output form.

In order to preserve backwards compatibility, the extra input types can
be retained by setting the `generateExtraInputTypes` property in the
Go language options for a Pulumi package.

The goal of these changes is to reduce the size of Go SDK, particularly
for Azure-Native.

Related to https://github.com/pulumi/pulumi/issues/8194


pgavlin:
A note on the status of this draft:

I could not find a reasonable path forward without doing some pretty gnarly refactoring in the Go SDK generator. The goal of the refactoring is to accurately track the set of input and output types that are used by the schema so that callers can opt out of generating additional types. This accounting is relatively simple: most of the implementation is in the (new) file gen_types.go, with some additional code in generatePackageContextMap.

As part of the refactoring, I've also attempted to unify the generation of input and output types, as this unification is a natural consequence of the new way of tracking input and output types. This is where things get hairy:

1. Enum input/output type generation is almost entirely distinct from other types
2. `Input<Optional<Resource<token>>` is strange, to say the least, in that it generates a `ResourcePtrInput` type despite the fact that resources are _already_ represented using pointer types (and can therefore already represent the absence of a value)

I think I've addressed most of the inconsistencies caused by (1), but the manner in which they're currently addressed probably causes breaking changes. Those breaks aren't required, and can be resolved with some further (if unfortunate) special-casing for enums.

(2) remains a problem. The types we generate in the SDK and the type names we generate from `Package.inputType` are not the same, and this is... not great. Resolving that inconsistency may be the right path forward here.

As these changes are intended to produce no breaking changes (only _textual_ diffs due to reordering of types in the output), both (1) and (2) need to be fully addressed in that mindset, which may require reintroducing some additional special-cased logic.

pgavlin:
(oh, and tests are failing b/c I deliberately haven't updated SDK baselines)

pgavlin:
Note for reviewers: these changes are broken up into a single commit per refactoring. Only the last commit changes the expected test output--every other commit causes no diffs in the SDK codegen (whether because the change itself produces no diffs or because the test schemas have been updated to set the `generateExtraInputTypes` property).

mikhailshilkov:
@pgavlin Is this ready for review? Or do you want to take care of downstream failures first?

pgavlin:
> Is this ready for review? Or do you want to take care of downstream failures first?

This is ready for review. I'll take care of the downstream failures concurrently.

pkg/codegen/go/gen.go
```diff
@@ -39,9 +39,35 @@ import (
...
+
+	output      bool
+	ptrOutput   bool
+	arrayOutput bool
+	mapOutput   bool
+}
+
+func (d *typeDetails) mark(input, output bool) {
+	d.input = d.input || input
+	d.output = d.output || input || output
```
mikhailshilkov, line 55: Why do you set `output` to true if `input` is true? Are you planning to remove unreferenced output types later on?

mikhailshilkov (commented):
The PR makes sense directionally. Thank you for splitting the commits - it's very useful. My main worry is that this produces invalid code in azure-native (I checked aws-native manually and it has the same problem).

pkg/codegen/go/gen.go
pgavlin: Generating the Go form of an input type requires that we also generate the Go form of the corresponding output type. This is due to the way we define input type interfaces, which reference output types:

```
type StringInput interface {
 	pulumi.Input

 	ToStringOutput() StringOutput
 	ToStringOutputWithContext(context.Context) StringOutput
 }
```

pgavlin:
Okay, I think that this PR is ready to go.

@pulumi/platform-integrations tfgen will need to be updated to set the new `generateExtraInputTypes` property to `true` in the Go language info for the package in generated schemas in order to avoid changes in generated SDKs.

@pulumi/native-providers any packages that you'd like to exclude from SDK changes will also need to set that property to `true`.

mikhailshilkov:
I measured the impact on azure-native and this PR cuts about 38% in size which is awesome! I'm eager to get this merged!

pkg/codegen/go/gen.go
```diff
@@ -3086,13 +3177,17 @@ func generatePackageContextMap(tool string, pkg *schema.Package, goInfo GoPackag
...
 	// input or output property type metadata, in case they have
 	// types used in array or pointer element positions.
-	for _, f := range pkg.Functions {
-		parentOptional := false
-		if f.Inputs != nil {
-			populateDetailsForPropertyTypes(seenMap, f.Inputs.Properties, parentOptional)
-		}
-		if f.Outputs != nil {
-			populateDetailsForPropertyTypes(seenMap, f.Outputs.Properties, parentOptional)
+	if !goInfo.DisableFunctionOutputVersions || goInfo.GenerateExtraInputTypes {
```
iwahbe, line 3180: Do we have a test for how these interact?

pkg/codegen/go/gen.go
```diff
@@ -3293,6 +3388,7 @@ func GeneratePackage(tool string, pkg *schema.Package) (map[string][]byte, error
 
 			importsAndAliases := map[string]string{}
 			pkg.getImports(r, importsAndAliases)
+			importsAndAliases["github.com/pulumi/pulumi/sdk/v3/go/pulumi"] = ""
```
iwahbe, line 3391: Can you explain why this has moved? I would expect _functions_ to use it at least. (It seems like I'm wrong though.)

iwahbe (commented):
Overall, this PR looks great. I have a couple of questions though (mostly for my own understanding).

pkg/codegen/go/gen.go
```diff
@@ -39,9 +39,39 @@ import (
 )
 
 type typeDetails struct {
-	ptrElement   bool
-	arrayElement bool
-	mapElement   bool
+	input      bool
```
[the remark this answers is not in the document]
pgavlin: Not all combinations are valid--I'll add a comment describing those that are.

pkg/codegen/go/gen.go
pgavlin: Yes: enums do not require the `pulumi` SDK if they do not have input or output versions.

t0yv0 (approved):
Not finding any obvious issues, this LGTM on the strength of downstream checks.

It's mind-bending trying to prove that populateDetailsForTypes does the exact same inclusion logic as actual code generator. But if it didn't in practice we'd see a downstream failing check.

If we ever attempt to refactor to a two-pass (or N-pass) design that builds up a Go model in-memory before emitting this feature would be so much easier; for aux types we'd just generate them lazily on demand the first time code generation references them. That'd be precise and foolproof.

pgavlin:
> It's mind-bending trying to prove that populateDetailsForTypes does the exact same inclusion logic as actual code generator. But if it didn't in practice we'd see a downstream failing check.
> 
> If we ever attempt to refactor to a two-pass (or N-pass) design that builds up a Go model in-memory before emitting this feature would be so much easier; for aux types we'd just generate them lazily on demand the first time code generation references them. That'd be precise and foolproof.

I agree with all of this. And indeed, that sort of a refactor is akin to the second road I went down with this PR: https://github.com/pulumi/pulumi/compare/52d01bb915276b016886eef38c898216c981e10f...pgavlin/go-type-instantiations

Unfortunately, it was difficult to validate those changes due to the textual differences they produced. I'm somewhat motivated to adjust the existing codegen output to match the output those changes would have produced, then try to land that branch. Not anytime particularly soon, though.

t0yv0:
Yeah we need this change asap, I'm in favor. LMK if need another review or dig in on some specifics.

iwahbe (approved):
Except my previous items, I think this looks good. Like @t0yv0, I'm relying heavily on downstream tests to supplement correctness checks. I think that the change would result in compile time errors more than runtime errors, so downstream is a powerful check here.

PR #7842, opened 2021-08-25
Disable default completion command

EthanOrlander:
# Description

Fixes  [7841](https://github.com/pulumi/pulumi/issues/7841)

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works

- [x] I have updated the [CHANGELOG-PENDING](https://github.com/pulumi/pulumi/blob/master/CHANGELOG_PENDING.md) file with my change

- [ ] Yes, there are changes in this PR that warrants bumping the Pulumi Service API version

PR #7950, opened 2021-09-10
chore: upgrade cobra to v1.2.1

EthanOrlander:
# Description

Upgrades cobra to v1.2.1 to give access to disabling default completion command (as Pulumi overrides it). Required for https://github.com/pulumi/pulumi/pull/7842

## Checklist


pgavlin:
@EthanOrlander are there significant changes from cobra v1.0.0? Anything behavioral differences need to worry about?

AaronFriel:
We're now on v1.5.0 if I read our dependencies correctly.

Issue #7361, opened 2021-06-24
Flaky TestBrokenDynamicProvider


t0yv0:
https://github.com/pulumi/pulumi/runs/2898508112

panic: TestBrokenDynamicProvider: test timed out after 600 seconds, suspect pulumi hanging

Note: failure only on Mac OS runner, perhaps something network-related blowing the timeout?

If this happens again please +1 the issue.

emiliza:
+1 just happened again 😞 

https://github.com/pulumi/pulumi/runs/3549833224?check_suite_focus=true

PR #7951, opened 2021-09-10
Lower the BrokenDynamicProvider regression test from integration to mock

t0yv0:
# Description

Fixes a flaky problem. I have verified that breaking the product makes the test catch the hang as before, so we still have the coverage we want. The solution is to avoid integration test overhead and flakiness and just test this in pytest as as a max-1s mock test.

Fixes https://github.com/pulumi/pulumi/issues/7361

## Checklist


komalali (approved):
:shipit:

emiliza (approved):

PR #7952, opened 2021-09-10
[schema] Add the Pulumi Package metaschema.

pgavlin:
The Pulumi Package metaschema is a JSON schema definition that describes
the format of a Pulumi Package schema. The metaschema can be used to
validate certain basic properties of a Pulumi Package schema, including
(but not limited to):

- data types (e.g. is this property a string?)
- data formats (e.g. is this string property a valid regex?)
- object shapes (e.g. is this object missing required properties?)

The schema binder has been updated to use the metaschema as its first
validation pass.

In addition to its use in the binder, the metaschema has its own page in
the developer documentation. This page is generated using a small tool,
jsonschema2md.go.

### DIFF
--- a/developer-docs/providers/metaschema.md
+++ b/developer-docs/providers/metaschema.md
@@ -1,0 +1,801 @@
+# Pulumi Package Metaschema
+
+A description of the schema for a Pulumi Package
+
+`object`
+
+## Properties
+
+---
+
+### `attribution`
+
+Freeform text attribution of derived work, if required.
+
+`string`
+
+---
+
+### `config`
+
+The package's configuration variables.
+
+`object`
+
+#### Properties
+
+---
+
+##### `required`
+
+A list of the names of the package's required configuration variables.
+
+`array`
+
+Items: `string`
+
+---
+
+##### `variables`
+
+A map from variable name to propertySpec that describes a package's configuration variables.
+
+`object`
+
+Additional properties: [Property Definition](#property-definition)
+
+---
+
+---
+
+### `description`
+
+The description of the package. Descriptions are interpreted as Markdown.
+
+`string`
+
+---
+
+### `functions`
+
+A map from token to functionSpec that describes the set of functions defined by this package.
+
+`object`
+
+Property names: [Token](#token)
+
+Additional properties: [Function Definition](#function-definition)
+
+---
+
+### `homepage`
+
+The package's homepage.
+
+`string`
+
+---
+
+### `keywords`
+
+The list of keywords that are associated with the package, if any.
+
+`array`
+
+Items: `string`
+
+---
+
+### `language`
+
+Additional language-specific data about the package.
+
+`object`
+
+---
+
+### `license`
+
+The name of the license used for the package's contents.
+
+`string`
+
+---
+
+### `logoUrl`
+
+The URL of the package's logo, if any.
+
+`string`
+
+---
+
+### `meta`
+
+Format metadata about this package.
+
+`object`
+
+#### Properties
+
+---
+
+##### `moduleFormat` (_required_)
+
+A regex that is used by the importer to extract a module name from the module portion of a type token. Packages that use the module format "namespace1/namespace2/.../namespaceN" do not need to specify a format. The regex must define one capturing group that contains the module name, which must be formatted as "namespace1/namespace2/...namespaceN".
+
+`string`
+
+Format: `regex`
+
+---
+
+---
+
+### `name` (_required_)
+
+The unqualified name of the package (e.g. "aws", "azure", "gcp", "kubernetes", "random")
+
+`string`
+
+Pattern: `^[^0-9][-a-zA-Z0-9]*$`
+
+---
+
+### `pluginDownloadUrl`
+
+The URL to use when downloading the provider plugin binary.
+
+`string`
+
+---
+
+### `provider`
+
+The provider type for this package.
+
+[Resource Definition](#resource-definition)
+
+---
+
+### `repository`
+
+The URL at which the package's sources can be found.
+
+`string`
+
+---
+
+### `resources`
+
+A map from type token to resourceSpec that describes the set of resources and components defined by this package.
+
+`object`
+
+Property names: [Token](#token)
+
+Additional properties: [Resource Definition](#resource-definition)
+
+---
+
+### `types`
+
+A map from type token to complexTypeSpec that describes the set of complex types (i.e. object, enum) defined by this package.
+
+`object`
+
+Property names: [Token](#token)
+
+Additional properties: [Type Definition](#type-definition)
+
+---
+
+### `version`
+
+The version of the package. The version must be valid semver.
+
+`string`
+
+Pattern: `^v?(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$`
+
+---
+
+## Alias Definition
+
+`object`
+
+### Properties
+
+---
+
+#### `name`
+
+The name portion of the alias, if any
+
+`string`
+
+---
+
+#### `project`
+
+The project portion of the alias, if any
+
+`string`
+
+---
+
+#### `type`
+
+The type portion of the alias, if any
+
+`string`
+
+---
+
+## Array Type
+
+A reference to an array type. The "type" property must be set to "array" and the "items" property must be present. No other properties may be present.
+
+`object`
+
+### Properties
+
+---
+
+#### `items` (_required_)
+
+The element type of the array
+
+[Type Reference](#type-reference)
+
+---
+
+#### `type` (_required_)
+
+Constant: `"array"`
+
+---
+
+## Enum Type Definition
+
+Describes an enum type
+
+`object`
+
+### Properties
+
+---
+
+#### `enum` (_required_)
+
+The list of possible values for the enum
+
+`array`
+
+Items: [Enum Value Definition](#enum-value-definition)
+
+---
+
+#### `type` (_required_)
+
+The underlying primitive type of the enum
+
+`string`
+
+Enum: `"boolean"` | `"integer"` | `"number"` | `"string"`
+
+---
+
+## Enum Value Definition
+
+`object`
+
+### Properties
+
+---
+
+#### `deprecationMessage`
+
+Indicates whether or not the value is deprecated.
+
+`string`
+
+---
+
+#### `description`
+
+The description of the enum value, if any. Interpreted as Markdown.
+
+`string`
+
+---
+
+#### `name`
+
+If present, overrides the name of the enum value that would usually be derived from the value.
+
+`string`
+
+---
+
+#### `value` (_required_)
+
+The enum value itself
+
+`boolean` | `integer` | `number` | `string`
+
+---
+
+## Function Definition
+
+Describes a function.
+
+`object`
+
+### Properties
+
+---
+
+#### `deprecationMessage`
+
+Indicates whether or not the function is deprecated
+
+`string`
+
+---
+
+#### `description`
+
+The description of the function, if any. Interpreted as Markdown.
+
+`string`
+
+---
+
+#### `inputs`
+
+The bag of input values for the function, if any.
+
+[Object Type Details](#object-type-details)
+
+---
+
+#### `language`
+
+Additional language-specific data about the function.
+
+`object`
+
+---
+
+#### `outputs`
+
+The bag of output values for the function, if any.
+
+[Object Type Details](#object-type-details)
+
+---
+
+## Map Type
+
+A reference to a map type. The "type" property must be set to "object" and the "additionalProperties" property may be present. No other properties may be present.
+
+`object`
+
+### Properties
+
+---
+
+#### `additionalProperties`
+
+The element type of the map. Defaults to "string" when omitted.
+
+[Type Reference](#type-reference)
+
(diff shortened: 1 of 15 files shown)


pkg/codegen/internal/test/testdata/simple-enum-schema/schema.json
```diff
@@ -197,13 +197,6 @@
         }
       ]
     },
-    "plant:tree/v1:VarietiesSupported": {
-      "type": "array",
-      "items": {
-        "$ref": "#/types/plant:tree/v1:RubberTreeVariety"
-      },
-      "description": "The rubber tree varieties supported"
-    },
```
komalali, line 206: Why did this get removed 🤔?

pgavlin: Array type definitions aren't used by anything: only object and enum type definitions are processed by the schema binder. As such, the metaschema only allows object and enum type definitions in the `types` map. This schema was failing validation because of this array type definition, so I removed it.

developer-docs/providers/metaschema.md
```diff
@@ -0,0 +1,789 @@
+# Pulumi Package Metaschema
```
t0yv0, line 1: Is this auto-generated or hand-authored?

pkg/codegen/schema/pulumi.json
```diff
@@ -0,0 +1,504 @@
...
+                                "type": "string"
+                            }
+                        },
+                        "required": ["value"]
+                    }
+                }
+            },
+            "required": ["type", "enum"]
+        },
+        "resourceSpec": {
```
t0yv0, line 411: I'm guessing this is the hand-authored part. I can read it easier! Thanks.

t0yv0:
OK I've read this again. I'm absolutely loving the meta schema. I can consume it - the JSON notation is not the best (give me OCaml type defs) but I can read it fine. We're getting somewhere!

pkg/codegen/internal/test/testdata/simple-enum-schema/schema.json
mikhailshilkov: I was wondering about this the other day. CloudFormation schema does have explicit type definitions for array types. I modeled them as inline types for now. Alternatively, we could extend Pulumi schema to support those.

pkg/codegen/schema/pulumi.json
```diff
@@ -0,0 +1,504 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "https://github.com/pulumi/pulumi/blob/master/pkg/codegen/schema.json",
+    "title": "Pulumi Package Metaschema",
+    "description": "A description of the schema for a Pulumi Package",
+    "type": "object",
+    "properties": {
+        "name": {
+            "description": "The unqualified name of the package (e.g. \"aws\", \"azure\", \"gcp\", \"kubernetes\", \"random\")",
```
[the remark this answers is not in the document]
pgavlin: Good idea. We should definitely figure out what an appropriate pattern is for `name`. Probably reasonable to start with what we have in the package portion of the `token` regex.

pkg/codegen/schema/pulumi.json
```diff
@@ -0,0 +1,504 @@
...
+            "type": "string"
+        },
+        "version": {
+            "description": "The version of the package. The version must be valid semver.",
+            "type": "string",
+            "pattern": "^v?(?P<major>0|[1-9]\\d*)\\.(?P<minor>0|[1-9]\\d*)\\.(?P<patch>0|[1-9]\\d*)(?:-(?P<prerelease>(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\\.(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?$"
+        },
+        "description": {
+            "description": "The description of the package.",
```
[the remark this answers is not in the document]
pgavlin: Excellent question. It does.

pkg/codegen/schema/pulumi.json
```diff
@@ -0,0 +1,504 @@
...
+        "pluginDownloadUrl": {
+            "description": "The URL to use when downloading the provider plugin binary.",
+            "type": "string"
+        },
+        "meta": {
+            "description": "Format metadata about this package.",
+            "type": "object",
+            "properties": {
+                "moduleFormat": {
+                    "description": "A regex that is used by the importer to extract a module name from the module portion of a type token. Packages that use the module format \"namespace1/namespace2/.../namespaceN\" do not need to specify a format. The regex must define one capturing group that contains the module name, which must be formatted as \"namespace1/namespace2/...namespaceN\".",
```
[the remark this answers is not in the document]
pgavlin: In principle any regex is supported here... are you looking for examples of expressions that are in active use?

pkg/codegen/schema/pulumi.json
pgavlin: That's right.

developer-docs/providers/metaschema.md
pgavlin: This is generated by the `jsonschema2md` tool that is part of this PR: https://github.com/pulumi/pulumi/pull/7952/files#diff-98fe9f26fbfe318ee4b2589b4c08d4d693cea4f721b10cc3be93b659e73b47dc

pkg/codegen/internal/test/testdata/simple-enum-schema/schema.json
pgavlin: Yeah, I've wondered about whether or not we need this. How is it used in CloudFormation?

@komalali your question reminded me that technically this is a breaking change for schema authors, as their previously compliant schemas will now be rejected by the binder. Do you think that we should allow these types, and maybe warn on them instead of just ignoring them?

mikhailshilkov: Example:

```json
    "CertificatePolicyList" : {
      "type" : "array",
      "items" : {
        "$ref" : "#/definitions/PolicyInformation"
      }
    },
    "PolicyInformation" : {
      "description" : "Structure that contains X.509 Policy information.",
      "type" : "object",
...
```

pkg/codegen/schema/pulumi.json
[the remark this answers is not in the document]
mikhailshilkov: Consider adding this to the description.

pkg/codegen/schema/pulumi.json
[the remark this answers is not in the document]
mikhailshilkov: Maybe? I feel this may be confusing for schema users.

pkg/codegen/schema/pulumi.json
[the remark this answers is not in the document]
pgavlin: Yup

pgavlin:
The codegen failure is unrelated.

pkg/codegen/schema/pulumi.json
```diff
@@ -0,0 +1,509 @@
...
+                    "title": "Primitive Type",
+                    "description": "A reference to a primitive type.",
+                    "type": "object",
+                    "properties": {
+                        "type": {
+                            "description": "The primitive type, if any",
+                            "type": "string",
+                            "enum": ["boolean", "integer", "number", "string"]
+                        },
+                        "additionalProperties": false,
```
t0yv0, line 156: I had trouble parsing this. What is this for? Note it's in the "properties" context. Should it be moved higher up to the object x where x.description == "A reference to a primitive type"?

pkg/codegen/schema/pulumi.json
```diff
@@ -0,0 +1,509 @@
...
+                    "type": "object",
+                    "properties": {
+                        "type": {
+                            "const": "array"
+                        },
+                        "items": {
+                            "description": "The element type of an array",
+                            "$ref": "#/$defs/typeSpec"
+                        },
+                        "additionalProperties": false,
```
t0yv0, line 175: One more.

pkg/codegen/schema/pulumi.json
pgavlin: Ah, it should be up one level, yes. Good catch!

t0yv0: If it's actually "additionalProperties" in the object language, what does false mean in the meta-language?

pgavlin: Oh, wait, I take that back

pkg/codegen/schema/pulumi.json
```diff
@@ -0,0 +1,509 @@
...
+                },
+                {
+                    "title": "Map Type",
+                    "description": "A reference to a map type.",
+                    "type": "object",
+                    "properties": {
+                        "type": {
+                            "const": "object"
+                        },
+                        "additionalProperties": {
```
t0yv0, line 189: Similarly, is this nested correctly?

pkg/codegen/schema/pulumi.json
```diff
@@ -0,0 +1,509 @@
...
+                            "type": "string"
+                        },
+                        "$ref": {
+                            "description": "A reference to a type in this or another document. For example, the built-in Archive, Asset, and Any\ntypes are referenced as \"pulumi.json#/Archive\", \"pulumi.json#/Asset\", and \"pulumi.json#/Any\", respectively.\nA type from this document is referenced as \"#/types/pulumi:type:token\".\nA type from another document is referenced as \"path#/types/pulumi:type:token\", where path is of the form:\n  \"/provider/vX.Y.Z/schema.json\" or \"pulumi.json\" or \"http[s]://example.com/provider/vX.Y.Z/schema.json\"\nA resource from this document is referenced as \"#/resources/pulumi:type:token\".\nA resource from another document is referenced as \"path#/resources/pulumi:type:token\", where path is of the form:\n  \"/provider/vX.Y.Z/schema.json\" or \"pulumi.json\" or \"http[s]://example.com/provider/vX.Y.Z/schema.json\"",
+                            "type": "string",
+                            "format": "uri-reference"
+                        },
+                        "additionalProperties": false,
```
t0yv0, line 212: One more.

pkg/codegen/schema/pulumi.json
pgavlin: This line prevents the schema from matching if the object has a property named `additionalProperties`

pgavlin: This gets a bit hard to read because the names used by the metaschema to describe the package schema and the names used by the package schema to describe a package overlap.

t0yv0: Okay. So 'false' in the JSON schema meta-language means "never match". This makes some sense. I'll need to adjust the coverage script to take care of this.

t0yv0 (approved):
:shipit:

pkg/codegen/schema/pulumi.json
```diff
@@ -0,0 +1,509 @@
...
+    "additionalProperties": false,
+    "required": [
+        "name"
+    ],
+    "$defs": {
+        "token": {
+            "title": "Token",
+            "type": "string",
+            "$comment": "In the regex below, the 'module' portion of the token is optional. However, a missing module component creates a '::', which breaks URNs ('::' is the URN delimiter). We have many test schemas that use an empty module component successfully, as they never create URNs; while these are _probably_ the only places that need updating, it might be possible that there are module-less type tokens in the wild elsewhere and we may need to remain compatible with those tokens.",
+            "pattern": "^[^0-9][-a-zA-Z0-9]*:([^0-9][a-zA-Z0-9._/]*)?:[^0-9][a-zA-Z0-9._/]*$"
```
iwahbe, line 133: This lets patterns in that we don't want. For example: `-:-:-`, `:::::`, `::::`, `.:.:.`.

I don't know if we want a pattern like `foo:ba.:thing.` (note the `.` at the end), but that matches as well.

pkg/codegen/schema/pulumi.json
pgavlin: Yep. I've updated the description of the type reference schema to explain how types are discriminated.

iwahbe (changes requested):
This looks very good. My main concern is the pattern for `name` and `token`. I think it lets things through that we don't want.

pkg/codegen/schema/pulumi.json
```diff
@@ -0,0 +1,509 @@
...
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "https://github.com/pulumi/pulumi/blob/master/pkg/codegen/schema.json",
+    "title": "Pulumi Package Metaschema",
+    "description": "A description of the schema for a Pulumi Package",
+    "type": "object",
+    "properties": {
+        "name": {
+            "description": "The unqualified name of the package (e.g. \"aws\", \"azure\", \"gcp\", \"kubernetes\", \"random\")",
+            "type": "string",
+            "pattern": "^[^0-9][-a-zA-Z0-9]*$"
```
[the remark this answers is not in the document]
pgavlin: This is taken from the pattern for the package portion of tokens. I think that your suggestion is fine; I'll update both.

pkg/codegen/schema/pulumi.json
pgavlin: > This lets patterns in that we don't want. For example: -:-:-, :::::, ::::, .:.:..

Yes, this is intentional. The pattern here is not as strict as we necessarily want it to be, but I'm very confident that it will not fail any existing schemas. I intend to tighten this pattern with future changes, but I want to be more certain that we won't accidentally start rejecting schemas that are accepted today.

I think we can improve our confidence here by looking at the set of names the code generators can currently handle properly and then tightening this based on those observations, but I am not going to do so with this PR.

pgavlin: (I should note that it feels very very likely that there aren't any packages out here that we'd accidentally reject by tightening some of the aspects you mention, but in general we should be pretty conservative by default)

pgavlin: *are, not aren't

iwahbe (approved):

PR #7958, opened 2021-09-13
[sdk/go] Marshal output values

### LINKED ISSUE #7849: Go SDK marshalling (used in SDK)
Part of #7434

          consumed by the engine
          a Pulumi Output

justinvp:
This change adds support for marshaling outputs as output values in the Go SDK.

Fixes #7849

Part of #7434

### DIFF
--- a/sdk/go/pulumi/provider_test.go
+++ b/sdk/go/pulumi/provider_test.go
@@ -1277,8 +1277,14 @@ func TestConstructResult(t *testing.T) {
 	resolvedProps, _, _, err := marshalInputs(state)
 	assert.NoError(t, err)
 
-	assert.Equal(t, resource.NewPropertyMapFromMap(map[string]interface{}{
-		"foo":       "hi",
-		"someValue": "something",
-	}), resolvedProps)
+	assert.Equal(t, resource.PropertyMap{
+		"foo": resource.NewOutputProperty(resource.Output{
+			Element: resource.NewStringProperty("hi"),
+			Known:   true,
+		}),
+		"someValue": resource.NewOutputProperty(resource.Output{
+			Element: resource.NewStringProperty("something"),
+			Known:   true,
+		}),
+	}, resolvedProps)
 }
--- a/sdk/go/pulumi/rpc_test.go
+++ b/sdk/go/pulumi/rpc_test.go
@@ -1289,3 +1289,251 @@ func TestOutputValueMarshallingNested(t *testing.T) {
 		})
 	}
 }
+
+type rubberTreeArgs struct {
+	Size *TreeSize `pulumi:"size"`
+}
+type RubberTreeArgs struct {
+	Size TreeSizePtrInput
+}
+
+func (RubberTreeArgs) ElementType() reflect.Type {
+	return reflect.TypeOf((*rubberTreeArgs)(nil)).Elem()
+}
+
+type TreeSize string
+
+const (
+	TreeSizeSmall  = TreeSize("small")
+	TreeSizeMedium = TreeSize("medium")
+	TreeSizeLarge  = TreeSize("large")
+)
+
+func (TreeSize) ElementType() reflect.Type {
+	return reflect.TypeOf((*TreeSize)(nil)).Elem()
+}
+
+func (e TreeSize) ToTreeSizeOutput() TreeSizeOutput {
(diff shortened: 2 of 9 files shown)


sdk/go/pulumi/provider_test.go
```diff
@@ -490,8 +490,14 @@ func TestConstructResult(t *testing.T) {
 	resolvedProps, _, _, err := marshalInputs(state)
 	assert.NoError(t, err)
 
-	assert.Equal(t, resource.NewPropertyMapFromMap(map[string]interface{}{
-		"foo":       "hi",
-		"someValue": "something",
-	}), resolvedProps)
+	assert.Equal(t, resource.PropertyMap{
+		"foo": resource.NewOutputProperty(resource.Output{
```
justinvp, line 494: This change requires some explanation...

For those who aren't already familiar, the Go SDK marshaling goes through two layers:

1. First the Go SDK types are marshaled into `resource.PropertyValue` values. That's what this PR is changing.
2. Then when marshaling for gRPC calls, the `resource.PropertyValue` values are marshaled into the gRPC structures. Support for marshaling output values at this layer was already implemented in #7861.

During (2), when the option to `KeepOutputValues` isn't set, the same values as before these changes are returned.

For (1), when marshaling `Output`, the behavior of this PR is currently to always return an output value. (It's up to the `KeepOutputValues` option in (2) to determine whether the output values are kept or not when going across the gRPC boundary.)

This does lead to this unusual case for this test where we have some input values that are `StringOutput` that are being passed to `marshalInputs`, and since we're now returning `Output` as an output value, we need to change the expected values for this test.

I can imagine an alternative behavior for `marshalInputs`, where if the `Output`'s value is known, isn't a secret, and doesn't have any dependencies, we could just return the same value as before, so this test wouldn't have to change. We could similarly do that when the output is a secret and doesn't have any dependencies, just return the same value that would be returned previously. And we could even do it for unknowns when there aren't any dependencies. I don't feel strongly about this either way, as it doesn't really affect anything in practice (aside from this test), but curious what other folks think.

sdk/go/pulumi/rpc_test.go
```diff
@@ -880,3 +885,56 @@ func TestDependsOnComponent(t *testing.T) {
...
+					}
+					if known {
+						expected.Element = value.expected
+					}
+
+					name := fmt.Sprintf("value=%v, known=%v, secret=%v, deps=%v", value, known, secret, deps)
+					t.Run(name, func(t *testing.T) {
+						inputs := Map{"value": out}
+						expected := resource.PropertyMap{"value": resource.NewOutputProperty(expected)}
+						actual, _, _, err := marshalInputs(inputs)
```
t0yv0, line 932: There's no checks on the `_` and `_` - secret and deps, but perhaps this is okay/trivial.

t0yv0 (approved):
LGTM based on table-driven checks.

justinvp:
Test failure looks legitimate. I'll take a look.

t0yv0:
> Avoid infinite loop when an output's value is an Input

This sounded interesting, does it have a test case or implicitly covered by existing tests? I very briefly glanced over the tests and didn't see it.

justinvp:
> > Avoid infinite loop when an output's value is an Input
> 
> This sounded interesting, does it have a test case or implicitly covered by existing tests? I very briefly glanced over the tests and didn't see it.

There was an existing codegen test (`TestEnumUsage/Success`) that was failing with an infinite loop, which was happening when an output value itself is an `Input` (which was happening with one of the enums). I've extracted this out into it's own unit test. The "Avoid unnecessary ToOutput conversions" commit avoids the infinite loop.

justinvp:
@t0yv0, @pgavlin, please take a look

t0yv0 (approved):
Thanks Justin the test case will serve as a great reminder if we accidentally make changes on this again.

Reading into it a bit it seems that:

`type TreeSize string` is used in `TreeSizePtrInput` position.

And `TreeSize` implements both `TreeSizePtrInput` and `TreeSizeInput`, but "prefers" the second? It needs to implement `ElementType()` of `Input` interface and it does implement it pointing to `TreeSize` type not `*TreeSize` type. 

It was unexpected to me that this is even possible/used (two separate Input implementations), with one being potentially unsound, taking notes.

PR #7959, opened 2021-09-13
[codegen/go] Emit input type registrations

justinvp:
This change depends on #7928 and adds `RegisterInputType` calls for generated input types.

TODO
- [ ] Need to add registration calls for resource-related input types

Part of #7848
Part of #7434


pgavlin (approved):

pgavlin:
Looks like `azure-native` is still on an older version of `github.com/pulumi/pulumi/sdk/v3`, which is causing build failures in the generated code.

PR #7967, opened 2021-09-15
Iwahbe/7881/thread replace on chages through sdks

### LINKED ISSUE #7881: ReplaceOnChanges isn't respected in Go SDK
I'm trying to cause pulumi recreate my AKS cluster if the agentPoolProfile vm size has changed.

I tried using `pulumi.ReplaceOnChanges([]string{"agentPoolProfiles[0].vmSize"})` but it seems that it isn't respected at all. This is the error from pulumi:
```
azure-native:containerservice/v20210501:ManagedCluster (aks):
    error: Code="PropertyChangeNotAllowed" Message="Changing property 'agentPoolProfile.vmSize' is not allowed." Target="agentPoolProfile.vmSize"
```

When I try and use `pulumi.IgnoreChanges([]string{"agentPoolProfiles[0].vmSize"})` that seems to work so there must be some issue with ReplaceOnChanges

My pulumi CLI and SDK is v3.11.0 and using Go
azure native sdk is v1.23.1

### LINKED ISSUE #7922: ReplaceOnChanges isn't respected in .NET SDK
Similar to https://github.com/pulumi/pulumi/issues/7881, `ReplaceOnChanges` doesn't seem to work in .NET either. Reported in https://github.com/pulumi/pulumi-azure-native/issues/1117#issuecomment-913243480 and I can reproduce this myself.

I'd expect the option to be set somewhere in [this file](https://github.com/pulumi/pulumi/blob/master/sdk/dotnet/Pulumi/Deployment/Deployment_RegisterResource.cs) but it's not there.

iwahbe:
# Description

We finish threading `replaceOnChanges` through the Go and .NET SDKs. 

Fixes #7922
Fixes #7881

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -11,5 +11,5 @@
 - [cli] Use json.Unmarshal instead of custom parser
   [#7954](https://github.com/pulumi/pulumi/pull/7954)
 
-- [sdk] - Thread replaceOnChanges through Go and .NET
+- [sdk/{go,dotnet}] - Thread replaceOnChanges through Go and .NET
   [#7967](https://github.com/pulumi/pulumi/pull/7967)
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -10,3 +10,6 @@
 
 - [cli] Use json.Unmarshal instead of custom parser
   [#7954](https://github.com/pulumi/pulumi/pull/7954)
+
+- [sdk] - Thread replaceOnChanges through Go and .NET
+  [#7967](https://github.com/pulumi/pulumi/pull/7967)
--- a/sdk/dotnet/Pulumi/Deployment/Deployment_RegisterResource.cs
+++ b/sdk/dotnet/Pulumi/Deployment/Deployment_RegisterResource.cs
@@ -92,10 +92,12 @@ namespace Pulumi
             };
 
             if (customOpts != null)
+            {
                 request.AdditionalSecretOutputs.AddRange(customOpts.AdditionalSecretOutputs);
+                request.ReplaceOnChanges.AddRange(customOpts.ReplaceOnChanges);
+            }
 
             request.IgnoreChanges.AddRange(options.IgnoreChanges);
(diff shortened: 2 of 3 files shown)


sdk/dotnet/Pulumi/Deployment/Deployment_RegisterResource.cs
```diff
@@ -95,6 +95,7 @@ private static void PopulateRequest(RegisterResourceRequest request, PrepareResu
                 request.AdditionalSecretOutputs.AddRange(customOpts.AdditionalSecretOutputs);
 
             request.IgnoreChanges.AddRange(options.IgnoreChanges);
+            request.ReplaceOnChanges.AddRange(options.ReplaceOnChanges);
```
iwahbe, line 98: I would really like conformation here. I have been unable to test through a gRPC version change and don't speak C#.

iwahbe: Thanks to @t0yv0 and @mikhailshilkov helping me, this has been tested and seems to work!

CHANGELOG_PENDING.md
```diff
@@ -10,3 +10,6 @@
 
 - [cli] Use json.Unmarshal instead of custom parser
   [#7954](https://github.com/pulumi/pulumi/pull/7954)
+
+- [sdk] - Thread replaceOnChanges through Go and .NET
```
justinvp, line 14: Super nitpick, but this could be:

```suggestion
- [sdk/{go,dotnet}] - Thread replaceOnChanges through Go and .NET
```

justinvp (approved):
LGTM

Any idea why the tests added in #7226 didn't catch this? Are we missing test coverage?

iwahbe:
> Any idea why the tests added in #7226 didn't catch this? Are we missing test coverage?
Because this was a bug in the gRPC connection, I think it would need to be an integration test. I don't see any for `replaceOnChanges`. We are defiantly missing coverage. 

I'm going to merge this now, because we have a user blocked on it. I'll follow up with an issue for code coverage.

PR #7931, opened 2021-09-09
Run type checker against all languages

iwahbe:
# Description

Fixes #7801

I add tests for each generated SDK. 
- [X] `nodejs`
  This is implemented by invoking the `ts-node` cli over `npm`. I think this is the right approach, as we only have one program looking over our `.ts` code.
- [X] `python`
  We don't type check, only parse. This is done with the builtin module `py_compile`.
- [X] `dotnet`
  Currently, we invoke `dotnet build`. If you has a better (faster) solution, please let me know. 
- [X] `go`
  We also build go. This bother's me less, as its fast. 
- [ ] docs
I'm not sure what tests to run. If you can think of any, let me know. 

## Checklist

- [X] I have added tests that prove my fix is effective or that my feature works
  That's more or less the feature. 

- [ ] I have updated the [CHANGELOG-PENDING](https://github.com/pulumi/pulumi/blob/master/CHANGELOG_PENDING.md) file with my change

- [ ] Yes, there are changes in this PR that warrants bumping the Pulumi Service API version

PR #7971, opened 2021-09-15
5758 for node 

### LINKED ISSUE #5758: Functions/datasources don't compose well with Pulumi model - should accept Outputs

t0yv0:
Builds #5758/node on top of https://github.com/pulumi/pulumi/pull/7989 

This is the codegen change for Node that provides fnOutput helpers.

- [x] make sure unit tests are complete and run against the generated code
- [x] Refactoring `output-funcs` sub-schemas into a "standard" codegen example 
- [x] Extending TestSDKCodegen to be able to typecheck generated code (for Node) - overlaps with @iwahbe work in https://github.com/pulumi/pulumi/pull/7931
- [x] Extending TestSDKCodegen further to run hand-written unit tests against the generated code

# Description

Fixes #5758

## Checklist


pkg/codegen/internal/test/helpers.go
```diff
@@ -102,12 +111,119 @@ func loadDirectory(fs map[string][]byte, root, path string) error {
...
+			if err = os.Remove(path); err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
+// Reads `.sdkcodegenignore` file if present to use as loadDirectory ignore func.
```
t0yv0, line 141: Perhaps trying to be too cute here, something like .gitignore with patterns that exclude files from codegen content-checking and auto-deletion machinery, and especially the PULUMI_ACCEPT=1 machinery.

Files that need this:

- node_modules
- any hand-written test files that are put in alongside the generated files

pkg/codegen/internal/test/sdk_driver.go
```diff
@@ -4,23 +4,73 @@ import (
...
-	Then        map[string]ThenFunc
+
+	// Decides which programming languages to test the code
+	// generation for; defaults to all supported languages.
+	IncludeLanguage func(lang string) bool
+
+	// Hooks that run additional validations after the code has
+	// been generated. Hooks are usually language-specific and are
+	// included in this map by language.
+	HooksByLanguage map[string][]SdkTestHook
```
t0yv0, line 28: This is perhaps too verbose but I was going for marking typecheck and test phases separately so they show up as separate steps in `go test -test.v`.

pkg/codegen/internal/test/sdk_driver.go
```diff
@@ -4,23 +4,73 @@ import (
...
+type SdkTestEnv struct {
+	T *testing.T
+
+	// The directory with the generated code, typically `$TestDir/$Language`.
+	CodeDir string
+}
+
+func (env *SdkTestEnv) NewCommandInDir(dir string, name string, arg ...string) *exec.Cmd {
+	d := filepath.Join(env.CodeDir, dir)
+	env.T.Logf("cd %s && %s %s", d, name, strings.Join(arg, " "))
```
t0yv0, line 48: Helpers for invoking sub-commands. I really like having the command printed out so I can  just copy and try it in my shell if I want to retry that particular step.

pkg/codegen/internal/test/sdk_driver.go
```diff
@@ -112,37 +161,86 @@ var sdkTests = []sdkTest{
...
-	testDir := filepath.Join("..", "internal", "test", "testdata")
+	TestSDKCodegenWithOptions(t, &TestSDKCodegenOptions{
+		Language:   language,
+		GenPackage: genPackage,
+	})
+}
 
-	for _, tt := range sdkTests {
+// A variant of TestSDKCodegen with more options.
+func TestSDKCodegenWithOptions(t *testing.T, rawOptions *TestSDKCodegenOptions) {
```
t0yv0, line 171: Perhaps a bit overkill to parameterize by SdkTest[] set, I need to look if I can shoehorn output-funcs test into a normal SdkTest in the default list.

pkg/codegen/internal/test/testdata/output-funcs/nodejs/codegenTests.ts
```diff
@@ -0,0 +1,121 @@
+/*
+
+import 'mocha';
```
t0yv0, line 3: TBD make this work.

pkg/codegen/internal/test/testdata/output-funcs/nodejs/package.json
```diff
@@ -0,0 +1,19 @@
+{
+    "name": "@pulumi/mypkg",
+    "version": "${VERSION}",
```
t0yv0, line 3: @iwahbe this is here --^ ugly but hasn't caused compilation problems for tsc yet.

pkg/codegen/internal/test/testdata/output-funcs/nodejs/tsconfig.json
```diff
@@ -0,0 +1,32 @@
...
+        "declaration": true,
+        "sourceMap": true,
+        "stripInternal": true,
+        "experimentalDecorators": true,
+        "noFallthroughCasesInSwitch": true,
+        "forceConsistentCasingInFileNames": true,
+        "strict": true
+    },
+    "files": [
+        "codegenTests.ts",
```
t0yv0, line 16: I had to jump through hoops to get codegenTests.ts listed here. Could we do `**/*.ts` pattern instead? Or this is quite intentional for codegen to list files by name.

pkg/codegen/nodejs/gen.go
```diff
@@ -121,7 +121,7 @@ func (mod *modContext) tokenToModName(tok string) string {
 
 func (mod *modContext) namingContext(pkg *schema.Package) (namingCtx *modContext, pkgName string, external bool) {
 	namingCtx = mod
-	if pkg != nil && pkg != mod.pkg {
+	if pkg != nil && pkg.Name != mod.pkg.Name {
```
t0yv0, line 124: 💥  Really need to understand this one. It might not be correct. But codegen breaks badly without the change.

pkg/codegen/nodejs/gen.go
```diff
@@ -1026,7 +1075,7 @@ func (mod *modContext) getTypeImportsForResource(t schema.Type, recurse bool, ex
 		return true
 	case *schema.ObjectType:
 		// If it's from another package, add an import for the external package.
-		if t.Package != nil && t.Package != mod.pkg {
+		if t.Package != nil && t.Package.Name != mod.pkg.Name {
```
t0yv0, line 1078: 💥 another problematic spot

pkg/codegen/nodejs/gen.go
```diff
@@ -1042,7 +1091,7 @@ func (mod *modContext) getTypeImportsForResource(t schema.Type, recurse bool, ex
 		return true
 	case *schema.ResourceType:
 		// If it's from another package, add an import for the external package.
-		if t.Resource != nil && t.Resource.Package != mod.pkg {
+		if t.Resource != nil && t.Resource.Package.Name != mod.pkg.Name {
```
t0yv0, line 1094: 💥  And another.

pkg/codegen/nodejs/gen_test.go
```diff
@@ -28,3 +37,71 @@ func TestGenerateTypeNames(t *testing.T) {
...
 }
+
+// Resolves modules via Yarn and compiles the code with TypeScript
+// linking against in-repo Pulumi Node SDK.
+func compileHook() test.SdkTestHook {
+	return test.SdkTestHook{
+		Name: "compile",
+		RunHook: func(env *test.SdkTestEnv) {
+			env.Command("yarn", "install")
+			env.Command("yarn", "link", "@pulumi/pulumi")
```
t0yv0, line 48: This ensures the build and test runs against in-repo Node SDK. CC @iwahbe

pkg/codegen/nodejs/gen_test.go
```diff
@@ -28,3 +37,71 @@ func TestGenerateTypeNames(t *testing.T) {
...
+
+func generatePackage(tool string, pkg *schema.Package, extraFiles map[string][]byte) (map[string][]byte, error) {
+	p := *pkg
+	if len(p.Language) > 0 {
+		panic(fmt.Sprintf("%v", p.Language))
+	}
+	if extraFiles == nil {
+		extraFiles = make(map[string][]byte)
+	}
+	nodePkgInfo := NodePackageInfo{
```
t0yv0, line 64: This might need to go in schema.json if it can.

pkg/codegen/nodejs/gen_test.go
```diff
@@ -28,3 +37,71 @@ func TestGenerateTypeNames(t *testing.T) {
...
+			"mocha":        "^3.5.0",
+		},
+	}
+	p.Language["nodejs"] = nodePkgInfo
+	return GeneratePackageWithOptions(&GeneratePackageOptions{
+		Tool:       tool,
+		Pkg:        &p,
+		ExtraFiles: extraFiles,
+		ExtraFilesInPackageMetadata: []string{
+			"codegenTests.ts",
```
t0yv0, line 78: The ugly hook to get codegenTests.ts included in the TS compilation.

pkg/codegen/nodejs/gen_test.go
```diff
@@ -28,3 +37,71 @@ func TestGenerateTypeNames(t *testing.T) {
...
+
+	test.TestSDKCodegenWithOptions(t, &test.TestSDKCodegenOptions{
+		Language:    "nodejs",
+		GenPackage:  generatePackage,
+		TestRootDir: testRootDir,
+		SDKTests: []test.SdkTest{
+			{
+				Directory:   "output-funcs",
+				Description: "output-funcs",
+				IncludeLanguage: func(lang string) bool {
```
t0yv0, line 94: This I think can be removed with a small tweak.

pkg/codegen/internal/test/testdata/output-funcs/nodejs/package.json
iwahbe: It has no problems for `tsc`, but it will cause `npm install` to do nothing.

pkg/codegen/nodejs/gen_test.go
iwahbe: Check out our [schema docs](https://www.pulumi.com/docs/guides/pulumi-packages/schema/#packagelanguage), which are mostly up to date. It includes `packageName` and `devDependencies` for `nodejs` projects.

pkg/codegen/internal/test/helpers.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2016-2021, Pulumi Corporation.
+// Copyright 2016-2020, Pulumi Corporation.
```
iwahbe, line 1: Why are we reverting the year?

iwahbe (approved):
I'm ok with the work made here. I'm not sure if I like it going into t0yv0/codegen-testing-upgrades though.

PR #7975, opened 2021-09-16
[sdk/go] Unmarshal output values in component provider

### LINKED ISSUE #7848: Go SDK unmarshalling (used in MLC provider)
Part of #7434

          consumed by the engine
          a Pulumi Output

justinvp:
This adds support for unmarshaling output values in the `CopyTo` helper used when copying inputs to an args struct. More details here: https://github.com/pulumi/pulumi/pull/7928#issuecomment-916227805

Note: I'd like to add more test cases and clean up the implementation a bit, but I'll do that in a subsequent change.

Fixes #7848
Part of #7434

### DIFF
--- a/sdk/go/pulumi/provider.go
+++ b/sdk/go/pulumi/provider.go
@@ -16,13 +16,15 @@ package pulumi
 
 import (
 	"context"
+	"errors"
+	"fmt"
 	"reflect"
 	"sort"
 	"strings"
 
-	"github.com/pkg/errors"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource/plugin"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 
 	pulumirpc "github.com/pulumi/pulumi/sdk/v3/proto/go"
 	"google.golang.org/grpc"
@@ -48,26 +50,31 @@ func construct(ctx context.Context, req *pulumirpc.ConstructRequest, engineConn 
 	}
 	pulumiCtx, err := NewContext(ctx, runInfo)
 	if err != nil {
-		return nil, errors.Wrap(err, "constructing run context")
+		return nil, fmt.Errorf("constructing run context: %w", err)
 	}
 
 	// Deserialize the inputs and apply appropriate dependencies.
 	inputDependencies := req.GetInputDependencies()
 	deserializedInputs, err := plugin.UnmarshalProperties(
 		req.GetInputs(),
-		plugin.MarshalOptions{KeepSecrets: true, KeepResources: true, KeepUnknowns: req.GetDryRun()},
+		plugin.MarshalOptions{
+			KeepSecrets:      true,
+			KeepResources:    true,
+			KeepUnknowns:     req.GetDryRun(),
+			KeepOutputValues: true,
+		},
 	)
 	if err != nil {
-		return nil, errors.Wrap(err, "unmarshaling inputs")
+		return nil, fmt.Errorf("unmarshaling inputs: %w", err)
 	}
 	inputs := make(map[string]interface{}, len(deserializedInputs))
 	for key, value := range deserializedInputs {
 		k := string(key)
-		var deps []Resource
+		var deps urnSet
 		if inputDeps, ok := inputDependencies[k]; ok {
-			deps = make([]Resource, len(inputDeps.GetUrns()))
-			for i, depURN := range inputDeps.GetUrns() {
-				deps[i] = pulumiCtx.newDependencyResource(URN(depURN))
+			deps = urnSet{}
+			for _, depURN := range inputDeps.GetUrns() {
+				deps.add(URN(depURN))
 			}
 		}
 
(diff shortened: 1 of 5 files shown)


sdk/go/pulumi/provider.go
```diff
@@ -190,7 +197,22 @@ func createProviderResource(ctx *Context, ref string) (ProviderResource, error)
 
 type constructInput struct {
 	value resource.PropertyValue
-	deps  []Resource
+	deps  urnSet
```
t0yv0, line 200: 👍

sdk/go/pulumi/provider.go
```diff
@@ -202,21 +224,316 @@ func constructInputsMap(ctx *Context, inputs map[string]interface{}) (Map, error
...
 
-		output := ctx.newOutput(resultType, ci.deps...)
+		output := ctx.newOutput(resultType, ci.Dependencies(ctx)...)
 		output.getState().resolve(value, known, secret, nil)
 		result[k] = output
 	}
 	return result, nil
 }
 
+func gatherDeps(v resource.PropertyValue, deps urnSet) {
```
t0yv0, line 242: The switch is complete? This is more my Go competency question I guess. I find this:

```
// PropertyValue is the value of a property, limited to a select few types (see below).
type PropertyValue struct {
	V interface{}
}
```

I'm trying to "see below" but being left a bit fuzzy with the exact set of types we allow here. In e.g. F# the type def itself would be a tight closed grammar (I think TypeScript can do it to).

For example I don't see Asset/Archive here but that's probably ok if they don't carry deps.

justinvp: I think so. I'm basing this off of two things:

1. [`PropertyValue.ContainsUnknowns`](https://github.com/pulumi/pulumi/blob/4f3f3666952b016c017e546ba4f40270246fa029/sdk/go/common/resource/properties.go#L377) and [`PropertyValue.ContainsSecrets`](https://github.com/pulumi/pulumi/blob/4f3f3666952b016c017e546ba4f40270246fa029/sdk/go/common/resource/properties.go#L395) for drilling into `PropertyValue`s:

2. How we gather dependencies in SDKs during marshaling. For example, [in Go](https://github.com/pulumi/pulumi/blob/4f3f3666952b016c017e546ba4f40270246fa029/sdk/go/pulumi/rpc.go#L231) and the other SDKs, we only add to the dependencies when marshaling `Output`s and resource references (adding the resource's URN as a dependency).

sdk/go/pulumi/provider.go
```diff
@@ -202,21 +224,316 @@ func constructInputsMap(ctx *Context, inputs map[string]interface{}) (Map, error
...
+			gatherDeps(e, deps)
+		}
+	case v.IsObject():
+		for _, e := range v.ObjectValue() {
+			gatherDeps(e, deps)
+		}
+	}
+}
+
+func copyInputTo(ctx *Context, v resource.PropertyValue, dest reflect.Value) error {
```
t0yv0, line 266: Please help me understand why this is fundamentally different from this functionality in the RPC: 

```
func unmarshalPropertyValue(ctx *Context, v resource.PropertyValue) (interface{}, bool, error) {
```

t0yv0:
I'll try to do another review pass through the large code addition to see if I find any issues.. 

I understand the urgency and appreciate the "clean up the implementation a bit" later comment, just hoping we can make it "clean up the implementation a lot" if possible, one day?

Sheer volume of newly added code means our serde framework is not properly factored.. this code double-states structural traversals.  Naively we're doing something like this over some grammar of types:

```Protobuf <-> PropertyValue <-> interface{} (Go value) ```

And ideally we'd have the serde framework give us `reflect.Type -> Serializer`, and when we need to add support for a new type we'd just RegisterSerializer(..) or something, and add a random test case generator into the type soup to make sure it still turns around.

**Comment by t0yv0** [comment]
Starting to understand a bit better from a more thorough review, give me more time here disregard prev comment..

File: sdk/go/pulumi/provider.go [hunk 4]
```diff
@@ -202,21 +224,316 @@ func constructInputsMap(ctx *Context, inputs map[string]interface{}) (Map, error
...
+			return err
+		}
+		dest.Set(result)
+		return nil
+	}
+
+	if dest.Type().Implements(inputType) {
+		// Try to determine the input type from the interface.
+		if it, ok := inputInterfaceTypeToConcreteType.Load(dest.Type()); ok {
+			inputType := it.(reflect.Type)
```
t0yv0, line 326: Rambling here, this is getting interesting right here:

Yay so we're deserializing PropertyValue into dest placeholder of  type `dest.Type() = Input[T]`, which is hairy because `Input` is an  interface.

No matter, homegrown typeclasses to the rescue (`inputInterfaceTypeToConcreteType`)!

Now we found the default type `inputType = D` that implements `D <: Input[T]`. We know that a value of type `D` can be put into a placeholder of type `Input[T]`.

So now we will just deserialize `v` of type `resource.PropertyValue` into a placeholder of type `D`.

QQ: do we need brand new code to do this last bit? I thought the codebase already knows how to do this? Or is it non-trivial to "view" a placeholder of type `Input[T]` as a placeholder of type `D`?

t0yv0 (approved):
OK yeah LGTM basically, tried to grind through didn't find anything jumping out as problematic.. To answer my own questions:

- I think this change teaches Go to deserialize (from PropertyValue) Input[T] values via default type maps, and also auto-coerces Output values, which is net new functionality
 
- Because the overarching work-stream is *VERY* conservative, this new functionality is scoped into providers rather than general RPC framework

- The hope is that given time once we develop airtight testing capability around the RPC framework and ability to detect regressions, we'd be able to unify/simplify this code 
-

sdk/go/pulumi/provider.go
justinvp: > do we need brand new code to do this last bit?

I suspect we can indeed unify/simplify the code here, and I'll be taking a look at that subsequently.

justinvp:
I added a commit that adds support for plain structs. I am planning to merge this today and will be looking into simplifying/cleaning up the implementation subsequently.

PR #7985, opened 2021-09-16
[codegen/go] - Resolve type name collisions

### LINKED ISSUE #7966: [codegen/go] - Regression in enum name generation
There was a regression to enum name generation in the go SDK caused by https://github.com/pulumi/pulumi/pull/7857 - this is currently affecting azure-native and preventing an update.

For posterity, a description of the issue:
Consider a resource `azure-native:network:Endpoint` an object type `azure-native:network:Endpoint` and an enum `azure-native:network:EndpointType`.

In our current code generation, since there is a collision between the resource name and the object name, the object type is renamed `EndpointType`, which causes a collision with the enum `EndpointType`. As a result, the enum should be renamed `EndpointTypeEnum`. This was working as expected until the linked PR was checked in, but now the enum is not being correctly renamed, resulting in a collision between the two types.

komalali:
# Description

Extends https://github.com/pulumi/pulumi/pull/7857 to include enums and fixes a regression in enum name collision detection

Fixes #7966 

## Checklist


pkg/codegen/go/gen.go
```diff
@@ -165,6 +165,44 @@ func (pkg *pkgContext) tokenToType(tok string) string {
...
+	mod, name := pkg.tokenToPackage(tok), components[2]
+
+	modPkg, ok := pkg.packages[mod]
+	name = Title(name)
+
+	if ok {
+		newName, renamed := modPkg.renamed[name]
+		if renamed {
+			name = newName
+		} else if modPkg.duplicateTokens[tok] {
```
t0yv0, line 190: Looks like this is the changed line.

pkg/codegen/go/gen.go
```diff
@@ -165,6 +165,44 @@ func (pkg *pkgContext) tokenToType(tok string) string {
...
+			// If the package containing the enum's token already has a resource or type with the
+			// same name, add an `Enum` suffix.
+			name += "Enum"
```
t0yv0, line 193: `modPkg.renamed[name]` edit not needed here? I have not fully grasped what that does.

pkg/codegen/go/gen.go
```diff
@@ -2594,35 +2591,80 @@ func generatePackageContextMap(tool string, pkg *schema.Package, goInfo GoPackag
...
+		}
+
 		switch t := t.(type) {
 		case *schema.ObjectType:
 			pkg := getPkgFromToken(t.Token)
-			// maintain support for duplicate tokens for types and resources in Kubernetes
+			// maintain support for duplicate tokens for types and resources
 			if resSeen[t.Token] {
 				pkg.duplicateTokens[t.Token] = true
-				return
```
t0yv0, line 2603: Intentional right?

t0yv0 (approved):
LGTM ~ trusting the added test without the codegen change result in broken tests so we guard against regressions here. TYVM.

pkg/codegen/go/gen.go
komalali: yes! without this the type names are never processed and the enum never sees the collision

t0yv0:
I'm noticing force-push wipes GH history of what I reviewed and what I didn't. I've used force-push quite a bit myself before to clean commit history but it sounds like there's reasons to take it easy once we have reviewers.

komalali:
uh sorry i was in the middle of rebasing to fix the conflicts when you reviewed

t0yv0:
NP

pkg/codegen/go/gen.go
```diff
@@ -2594,35 +2591,80 @@ func generatePackageContextMap(tool string, pkg *schema.Package, goInfo GoPackag
 	// compute set of names generated by a type
 	// handling any potential collisions via remapping along the way
 	scanType := func(t schema.Type) {
+		getNames := func(name, suffix string) []string {
+			names := []string{}
+			names = append(names, name+suffix)
+			names = append(names, name+suffix+"Input")
+			names = append(names, name+suffix+"Output")
+			return names
```
pgavlin, line 2599: Not super important--especially since this is just code getting moved around--but returning a slice literal is a bit shorter and can avoid unnecessary allocations from `append`:

```suggestion
			return []string{name+suffix, name+suffix+"Input", name+suffix+"Output"}
```

pgavlin (approved):
I think the fix looks good. Just a couple stylistic nits.

pkg/codegen/go/gen.go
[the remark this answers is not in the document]
komalali: Yeah TBH this code gives me a headache. I'll try taking another pass at it.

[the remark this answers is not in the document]
komalali: Okay I tried, but due to the order of operations it is not feasible to remove `duplicateTokens` without reworking a bunch of this code. I'm going to merge this as-is and we can take another pass at this separately at another time.

komalali:
omfg the collisions 🤯

PR #7986, opened 2021-09-16
[automation-api] Exclude tests from test_fast.

pgavlin:
The Automation API tests take long enough that thay don't really fit
into `test_fast`. These changes move those tests to `test_all`.

### DIFF
--- a/scripts/test_subsets.py
+++ b/scripts/test_subsets.py
@@ -21,6 +21,9 @@ TEST_SUBSETS = {
         'github.com/pulumi/pulumi/tests/integration'
     ],
     'auto-and-lifecycletest': [
+        'auto-dotnet',
+        'auto-nodejs',
+        'auto-python',
         'github.com/pulumi/pulumi/sdk/v3/go/auto',
         'github.com/pulumi/pulumi/pkg/v3/engine/lifeycletest'
     ],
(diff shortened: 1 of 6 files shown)


pgavlin:
@komalali this is something I've been thinking about doing for a while now. The automation API tests feel more like the integration tests to me on a few axes:

- they need to run actual Pulumi programs
- they take longer than a unit test
- they depend on the Pulumi CLI

I think that given all of these attributes, it's more appropriate to move them from `test_fast` to `test_all`. What do you think?

komalali:
LGTM but looks like there's maybe been an [update to pylint](https://pypi.org/project/pylint/#history) which is throwing new errors. We might want to pin pylint (or fix the code)

pgavlin:
> We might want to pin pylint (or fix the code)

Where should we pin it?

komalali:
>Where should we pin it?

I think 2.10.2 since 2.11.0 was just released a few hours ago and lines up with these new errors showing up.

komalali:
Huh. Weird that didn't work. It was definitely 2.11 that added it. https://pylint.pycqa.org/en/latest/whatsnew/2.11.html?highlight=consider-using-f-string#:~:text=Added%20consider-using-f-string%3A%20Emitted%20when%20.format()%20or%20%27%25%27%20is%20being%20used%20to%20format%20a%20string.

pgavlin:
> Huh. Weird that didn't work. It was definitely 2.11 that added it. 

Trying a different syntax now...

komalali (approved):

justinvp (approved):

PR #7989, opened 2021-09-16
Codegen testing upgrades

t0yv0:
# Description

When working on #5758 I've tried introducing unit testing over generated code, that now conflicts with type-checking generated code. This PR generalizes TestSDKCodegen into multiple passes and starts adding test passes. It also fixes a few issues I found.. I prefer compiling files in-place rather then making a type-check/* copy - this is for when you want to open an editor on the compiled project and work on it directly.

- [x] go code not compiling files deeply 
- [x] go not compiling against in-repo pulumi go sdk
- [x] node not compiling against in-repo pulumi node sdk
- [ ] dotnet not compiling against in-repo pulumi dotnet sdk - I'll take this as follow-up
- [ ] python not compiling against in-repo pulumi python sdk - irrelevant for now until we have real tests

After this I'm hoping to follow up to make "output-funcs" a "normal" schema just like all the rest.

Fixes # (issue)

## Checklist


pkg/codegen/dotnet/gen_test.go
```diff
@@ -15,21 +13,27 @@ import (
...
+		Language:   "dotnet",
+		GenPackage: generatePackageWithVersion,
+		Checks: map[string]test.CodegenCheck{
+			"dotnet/compile": typeCheckGeneratedPackage,
+		},
+	})
+}
+
+// TODO replace this with GeneratePackage when https://github.com/pulumi/pulumi/pull/7938 lands.
+func generatePackageWithVersion(tool string, pkg *schema.Package, extraFiles map[string][]byte) (map[string][]byte, error) {
```
t0yv0, line 26: It looks like either this version or the original code works fine, not sure which to prefer.

pkg/codegen/internal/test/sdk_driver.go
```diff
@@ -59,17 +71,6 @@ var sdkTests = []sdkTest{
 	{
 		Directory:   "resource-args-python",
 		Description: "Resource args with same named resource and type",
-		Then: map[string]ThenFunc{
```
t0yv0, line 62: This is now done by just a generic "go/test" check.

pkg/codegen/internal/test/sdk_driver.go
```diff
@@ -145,31 +157,72 @@ func TestSDKCodegen(t *testing.T, language string, genPackage GenPkgSignature, c
 				schemaPath = filepath.Join(dirPath, "schema.yaml")
 			}
 
-			files, err := GeneratePackageFilesFromSchema(schemaPath, genPackage)
+			files, err := GeneratePackageFilesFromSchema(schemaPath, opts.GenPackage)
 			require.NoError(t, err)
 
-			// Check output is valid code (will type-check). If code will not
```
t0yv0, line 151: I'm actually actively preferring it the other way. PULUMI_ACCEPT=1 should let the system materialize the codegen outputs so we can see them in git diff. If subsequent checks fail that's good too.

pkg/codegen/internal/test/testdata/.sdkcodegenignore
```diff
@@ -0,0 +1,10 @@
+__pycache__
+bin
+node_modules
```
t0yv0, line 3: This is kinda like poor-man's gitignore that excludes paths from the codegen diff-checking algo.

pkg/codegen/internal/test/testdata/dash-named-schema/go/foo/submodule1/init.go
```diff
@@ -6,8 +6,8 @@ package submodule1
 import (
 	"fmt"
 
+	"dash-named-schema/foo"
 	"github.com/blang/semver"
-	"github.com/pulumi/pulumi/pkg/v3/codegen/internal/test/testdata/dash-named-schema/go/foo"
```
t0yv0, line 10: This now is guaranteed not to grab things off github.

pkg/codegen/internal/test/testdata/nested-module-thirdparty/go/foo/deeply/nested/module/init.go
```diff
@@ -32,7 +32,7 @@ func (m *module) Construct(ctx *pulumi.Context, name, typ, urn string) (r pulumi
 }
 
 func init() {
-	version, err := foo - bar.PkgVersion()
```
t0yv0, line 35: I do no think this previous code compiled.. "foo - bar" is output for identifier "foo-bar". Ouch. 

The reason it passed compile check is that compile check was not deeply compiling ( -v all).

pkg/testing/integration/command.go
```diff
@@ -83,9 +83,15 @@ func RunCommand(t *testing.T, name string, args []string, wd string, opts *Progr
 		t.Logf("Invoke '%v' failed: %s\n", command, cmdutil.DetailedError(runerr))
 
 		if !opts.Verbose {
+			stderr := opts.Stderr
```
t0yv0, line 86: Avoid a panic here under some circumstances.

pkg/codegen/go/gen_test.go
```diff
@@ -145,166 +175,166 @@ func (mocks) Call(args pulumi.MockCallArgs) (resource.PropertyMap, error) {
...
-				assert.Equal(t, typ, tree.RubberTreeVarietyBurgundy, "unexpected type on resource: %v", urn)
-				wg.Done()
-				return nil
-			})
-			wg.Wait()
-			return nil
-		}, pulumi.WithMocks("project", "stack", mocks(1))))
-	})
-}
+// func TestEnumUsage(t *testing.T) {
```
t0yv0, line 178: TODO I need to move these into in-place Go tests so that normal "test" pass can pick them up.

pgavlin:
> I prefer compiling files in-place rather then making a type-check/* copy - this is for when you want to open an editor on the compiled project and work on it directly.

I'm a fan of fewer copies here, too--I find it makes it easier to understand what exactly is under test--but does this imply that we'll need appropriate .gitignore files to handle the build detritus?

t0yv0:
```
I'm a fan of fewer copies here, too--I find it makes it easier to understand what exactly is under test--but does this imply that we'll need appropriate .gitignore files to handle the build detritus?
```

Yes precisely. There's a `.gitignore` file and `.sdkcodegenignore` file added in this PR at what I hope is the appropriate level of hierarchy.

pgavlin:
> There's a .gitignore file and .sdkcodegenignore file added in this PR at what I hope is the appropriate level of hierarchy.

Okay, perfect. I saw the latter but must have skipped over the former.

t0yv0:
This should be ready 🤞 let's see how tests do.

pkg/codegen/internal/test/testdata/nested-module-thirdparty/go/foo/deeply/nested/module/init.go
iwahbe: Thanks for pointing this out!!!

pkg/codegen/nodejs/gen_test.go
```diff
@@ -17,34 +15,57 @@ import (
...
 	cmdOptions := integration.ProgramTestOptions{
 		Verbose: true,
 		Stderr:  &stderr,
 		Stdout:  &stdout,
 	}
 
-	// TODO remove when https://github.com/pulumi/pulumi/pull/7938 lands
-	file := filepath.Join(pwd, "package.json")
-	oldFile, err := ioutil.ReadFile(file)
+	// TODO: previous attempt used npm. It may be more popular and
```
iwahbe, line 36: Do we want to support compiling our codegen output with `npm`? 
I'm thinking about the experience for authoring multi-lang components. If we say that `yarn` is required, then this decision makes sense. If we say you can also use `npm`, then I think we should test the more brittle system. I genuinely don't know our policy. Thoughts?

pkg/codegen/internal/test/helpers.go
```diff
@@ -142,28 +263,20 @@ func RewriteFilesWhenPulumiAccept(t *testing.T, dir, lang string, actual map[str
...
+	_, err := os.ReadDir(baseline)
 	switch {
 	case err == nil:
-		for _, e := range entries {
-			err = os.RemoveAll(filepath.Join(baseline, e.Name()))
-			require.NoError(t, err)
-		}
+		ignore, err := loadIgnoreMap(baseline)
+		require.NoError(t, err)
+		err = removeFilesFromDirUnlessIgnored(baseline, baseline, ignore)
```
iwahbe, line 271: I don't understand why we would want to leave some files behind. This makes the test result dependent on local files, perhaps generated from a previous test. Shouldn't these tests be entirely driven from schema.

t0yv0: Thank you for this comment it's worth thinking through and considering alternatives.

1. I find it awfully convenient to be able to add hand-authored files into the code-generated test tree, specifically test files defining unit (mock) tests over the generated code. Being in the same tree makes it easy to open up that tree in an editor and for it to feel like a normal project when editing those tests. So I would like to keep this capability.

2. The point you bring up is very important also - the rewrite procedures should completely remove files generated from a previous test, otherwise we cannot trust that we are actually checking the code-gen. Before the change, this was done by removing the folder whole-sale, which was trivially correct but conflicting with *1*. After the change, this is relying on `.sdkcodegenignore` patterns to teach the system which sub-trees to leave, assuming (but not knowing for sure) that no code-generated files would match the `.sdkcodegenignore` patterns. This also has the side-effect of leaving build files like node_modules alone when doing codegen rewrites, which is hopefully benign (can be removed by `git clean -fxd`).

How else could we accomplish 1+2?

a. Identify code-generated files by headers "// this file was generated", though this is problematic with json files

b. Make the code-generator emit a list of files "generated.txt" when generating, and clean only those

c. Slight variation on (1) where we copy hand-authored files from a "mixin" folder, such as:

```
rm -rf $codedir
mkdir $codedir
generate-code --to $codedir
cp -r $codedir-mixin/ $codedir/ # copy test files into the tree
cd $codedir && go test # etc
``` 

I personally think we might be ok as-is for now, but option (b) is attractive as it is just as ergonomic and has files in one place while being fully precise. Thoughts?

pkg/codegen/nodejs/gen_test.go
t0yv0: Oh you've mentioned that, it's a compelling argument I'm just waiting on #7938 to work and then also we need to figure out how to make NPM link against the in-source SDK (similar to yarn link), in the meanwhile the yarn change gets this rolling. I can create an issue to remember to switch back?

pkg/codegen/internal/test/helpers.go
pgavlin: > This also has the side-effect of leaving build files like node_modules alone when doing codegen rewrites, which is hopefully benign (can be removed by git clean -fxd).

I find this pretty worrisome from the point of view of type-checking the generated code. It feels like it's difficult to guarantee that leaving these files behind won't affect type checking. For example, if we ignore `go.mod` and it doesn't get overwritten before the next round of type checking, the results of type checking might not really be representative.

I should note that this shouldn't bite us in terms of overall validation--CI will always start from a clean state--but I do find it incredibly irritating when tests pass locally and fail in CI due to the sorts of environmental issues that are possible when we leave these things behind.

iwahbe: > I should note that this shouldn't bite us in terms of overall validation--CI will always start from a clean state--but I do find it incredibly irritating when tests pass locally and fail in CI due to the sorts of environmental issues that are possible when we leave these things behind.

Yes

I'm not sure _a_ and _b_ are viable. They works for files we generate directly, but fail for files generated by other processes (like `yarn.lock`). Solving this problem was why I had the `$TEST/typecheck` folder. I would advocate for _c_. It's less ergonomic, but easier to implement and obviously correct.

t0yv0: So about that.. 

`go clean -fxd` removes everything that's not checked into git, whether it's .gitignore'd or not. And even with current changes, we assume that checked in changes are intentional (and will match CI). That command is muscle memory for me and I use it a lot to make sure I'm 'testing from clean'. I'd not want to make it part of the encoded test flow because sometimes you don't want that.. for example you have dirty edits to a test file in progress or you just don't want to blow the build cache on iterative changes.

Would it help if I checked in a developer-oriented makefile in this folder with some of these commands, like `clean::` could do `git clean -fxd`; and we could have a helper command for the PULUMI_ACCEPT=1 run to make it more discoverable?

> For example, if we ignore go.mod and it doesn't get overwritten before the next round of type checking, the results of type checking might not really be representative.

`go.mod` is a special snowflake. In current version it is always overwritten by the `typeCheckGeneratedPackage` pass. Because I did not find a way to make codegen itself generate it as one of the official outputs (search for go.mod yields nothing around gen.go). So this particular example is not possible.

pkg/codegen/nodejs/gen_test.go
iwahbe: I'll make a note of it in #7938, I don't think it needs a separate issue. 

> then also we need to figure out how to make NPM link against the in-source SDK (similar to yarn link)

`npm` also has a [link feature](https://docs.npmjs.com/cli/v7/commands/npm-link). I assume we would just use that. 

I'm happy to merge this part as is, and fix it later.

pkg/codegen/internal/test/helpers.go
t0yv0: Let me come back to this on Monday and make a call. If I cannot convince you, the right thing is to cede ergonomics in favor of a simpler and more robust foolproof solution.

t0yv0:
@iwahbe and @pgavlin PTAL. The blanket fool-proof deletion on PULUMI_ACCEPT=1 is back. Unit test files are now copied in from `go-extras` and similar. Otherwise we keep the in-tree approach and .sdkcodegenignore cuteness. Minor inconvenience: when editing unit test files in-tree, one has to remember to copy the edits back into go-extras and check them in.

This looks like an ok compromise?

I can also bring back Ian's original approach of running extra checks on a completely separate copy of the workspace (separate from the one checked into Git for understanding codegen).

pgavlin:
I think that this has already been suggested in this PR, but looking at the code--`loadDirectory` and its `ignore` parameter in particular--I'm wondering if we should have the test framework emit a "manifest.json" file or something that lists all of the files that were emitted by the generator when run under `PULUMI_ACCEPT`. We could then drive the loading etc. based on that, though we'd still want to obliterate everything prior to emitting new baselines.

t0yv0:
Yes, it was suggested in the context of deletion lists but we can adopt it just for loadDirectory, and simultaneously get rid of .sdkcodegenignore rules. That'd be more foolproof. I will add.

t0yv0:
Last (necessarily large) commit adds manifests and removes `.sdkcodegenignore`. @pgavlin

pkg/codegen/dotnet/gen_test.go
```diff
@@ -15,21 +13,30 @@ import (
...
+	test.TestSDKCodegen(t, &test.SDKCodegenOptions{
+		Language:   "dotnet",
+		GenPackage: generatePackageWithVersion,
+		Checks: map[string]test.CodegenCheck{
+			"dotnet/compile": typeCheckGeneratedPackage,
+		},
+	})
+}
+
+// TODO replace this with GeneratePackage when https://github.com/pulumi/pulumi/pull/7938 lands.
```
iwahbe, line 25: https://github.com/pulumi/pulumi/pull/7938 has landed. It would be great to eliminate this, but we could also do it later.

pkg/codegen/go/gen_test.go
```diff
@@ -66,38 +62,76 @@ func TestGoPackageName(t *testing.T) {
...
+	alreadyHaveGoMod, err := test.PathExists(goMod)
+
+	require.NoError(t, err)
+	if alreadyHaveGoMod {
+		err := os.Remove(goMod)
+		require.NoError(t, err)
 	}
-	t.Logf("*** Testing go in dir: %q ***", root)
 
+	runCommand(t, "go_mod_init", codeDir, goExe, "mod", "init", inferModuleName(codeDir))
```
iwahbe, line 110: Is the correct response to delete and then renew `go.mod` or is it to create only if it doesn't exist.

pkg/codegen/internal/test/sdk_driver.go
```diff
@@ -137,7 +149,7 @@ type checkPackageSignature = func(t *testing.T, pwd string)
 //
 // The schema is the only piece that must be manually authored. Once the schema has been written, the expected outputs
 // can be generated by running `PULUMI_ACCEPT=true go test ./..." from the `pkg/codegen` directory.
-func TestSDKCodegen(t *testing.T, language string, genPackage GenPkgSignature, checkPackage checkPackageSignature) {
+func TestSDKCodegen(t *testing.T, opts *SDKCodegenOptions) {
```
iwahbe, line 152: We should probably add an explanation on the test manifest and how to add mix-in files.

iwahbe (dismissed):
I want to see the tests pass, but I'm happy with the strategy used.

pkg/codegen/dotnet/gen_test.go
t0yv0: Done.

pkg/codegen/internal/test/sdk_driver.go
t0yv0: Done, PTAL.

pkg/codegen/go/gen_test.go
t0yv0: This is a tricky one. It's not coming from the codegen, and in the latest deletion scheme, it'll never be coming from old runs of codegen (which is why I was worried and making sure to delete it). It may however come from go-extras mixins. If it does the user probably intends that one to be used. I can switch to creating it only if it does not exist.

t0yv0: Added the change PTAL.

t0yv0:
@iwahbe could we merge now? PTAL. The tests pass but I cancelled downstream checks to free up CI workers to do other verification, as we know these changes are test-only and should not affect downstream codegen.

pkg/codegen/internal/test/helpers.go
```diff
@@ -76,55 +78,85 @@ func LoadFiles(dir, lang string, files []string) (map[string][]byte, error) {
...
+	buf := &bytes.Buffer{}
+	enc := json.NewEncoder(buf)
+	enc.SetIndent("", "  ")
+	err := enc.Encode(cm)
+	if err != nil {
+		return err
 	}
-	return fs, nil
+	data := buf.Bytes()
+	return ioutil.WriteFile(filepath.Join(dir, "codegen-manifest.json"), data, 0600)
```
pgavlin, line 139: Nit: rather than emitting to an in-memory buffer, we could emit directly to the file:

```go
	sort.Strings(cm.EmittedFiles)

	f, err := os.Create(filepath.Join(dir, "codegen-manifest.json"), 0600)
	if err != nil {
		return err
	}
	defer f.Close()

 	enc := json.NewEncoder(f)
 	enc.SetIndent("", "  ")
	return enc.Encode(cm)
```

pkg/codegen/internal/test/helpers.go
```diff
@@ -76,55 +78,85 @@ func LoadFiles(dir, lang string, files []string) (map[string][]byte, error) {
...
-		if !assert.Contains(t, actual, name) || !assert.Equal(t, string(file), string(actual[name]), name) {
-			t.Logf("%s did not agree", name)
+		_, inActual := actual[name]
+		if inActual {
+			if !assert.Equal(t, string(file), string(actual[name]), name) {
+				t.Logf("%s did not agree", name)
+				ok = false
+			}
+		} else {
+			t.Logf("File %s was expected but is missing from the actual fileset", name)
```
pgavlin, line 153:
```suggestion
		if contents, inActual := actual[name]; inActual {
			if !assert.Equalf(t, string(file), string(contents), "%s did not agree", name) {
				ok = false
			}
		} else {
			t.Logf("File %s was expected but is missing from the actual fileset", name)
```

pgavlin (approved):
Two nits; LGTM otherwise.

t0yv0:
@iwahbe I can't merge - this is really annoying. 

```
Merging is blocked
Merging can be performed automatically once the requested changes are addressed.
```

@pgavlin I badly need this merged so I'll take the nit fix into the follow up PR that depends on this.

t0yv0:
Never mind. Got it merged. I figured out how to dismiss the blocking review from GH POV. Thanks. If there's a remaining problem LMK and I will add a new PR.

PR #7993, opened 2021-09-17
iwahbe/7858/fix nodejs hypen imports

### LINKED ISSUE #7858: Nodejs codegen produces incorrect import syntax for azure-native
When implementing the SDK generator for custom component providers, the nodejs codegen produces the following syntax for the azure native package:

```
import * as pulumiAzure-native from "@pulumi/azure-native";
```

As you can see the `pulumiAzure-native` contains a hyphen which is syntactically incorrect.

## Steps to reproduce

1. Create a custom component resource that uses the `@pulumi/azure-native` pkg
2. Run the `make generate` which ultimately executes `nodejsgen` against the `schema.json` to generate the SDKs
3. Check the output SDK for nodejs with incorrect import sysntax

Expected: possible correct syntax would be `import * as pulumiAzureNative from "@pulumi/azure-native";`
Actual: `import * as pulumiAzure-native from "@pulumi/azure-native";`

dotnet, go and python SDK generators seem to output in the correct format.

iwahbe:
# Description

Encapsulate the creation of external package names (`pulumi%s`) and convert them to pascal case (correctly handling `-`). 

Fixes #7858

## Checklist


pkg/codegen/nodejs/gen.go
```diff
@@ -45,6 +45,13 @@ type typeDetails struct {
 	inputType  bool
 }
 
+// title Capitalizes the first rune in s.
+//
+// Note: this is expected to work on strings which are not valid identifiers.
+// Example:
+// 	"hello"   => "Hello"
+// 	"hiAlice" => "HiAlice"
+// 	"hi.Bob"  => "Hi.Bob"
```
komalali, line 54: ❤️

pkg/codegen/nodejs/gen.go
```diff
@@ -69,6 +76,30 @@ func camel(s string) string {
...
+		} else {
+			split[len(split)-1] = append(split[len(split)-1], r)
+		}
+	}
+	words := make([]string, len(split))
+	for i, v := range split {
+		words[i] = title(string(v))
+	}
+	return strings.Join(words, "")
+}
```
komalali, line 95: Since there's already a `camel` function couldn't `pascal` just be `title(camel(s))`?

iwahbe: No. I had that thought, but it doesn't work out. `camel` does conversions like `JSONObject` to `jsonObject`. `title(camel("JSONObject"))` would be `JsonObject` and we want it to remain unchanged. `camel` also has no impact on `-`, which we want `pascal` to do.

komalali (approved):

pkg/codegen/nodejs/gen.go
komalali: Expanding on the comment for `pascal` to include what you just said would be awesome!

PR #7997, opened 2021-09-18
Implemented filebase64sha256 & fileArchive support for codegen

T-Vova:
Implemented filebase64sha256 support for code generator, and fileArchive support for Go.
Pairs with: https://github.com/pulumi/pulumi-terraform-bridge/pull/391

### DIFF
--- a/pkg/codegen/dotnet/utilities.go
+++ b/pkg/codegen/dotnet/utilities.go
@@ -128,6 +128,12 @@ func getHelperMethodIfNeeded(functionName string) (string, bool) {
 		return `private static string ReadFileBase64(string path) {
 		return Convert.ToBase64String(System.Text.UTF8.GetBytes(File.ReadAllText(path)))
 	}`, true
+	case "filebase64sha256":
+		return `private static string ComputeFileBase64Sha256(string path) {
+		var fileData = System.Text.UTF8.GetBytes(File.ReadAllText(path))
+		var hashData = SHA256.Create().ComputeHash(fileData)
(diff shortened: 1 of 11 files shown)


pkg/codegen/dotnet/utilities.go
```diff
@@ -129,6 +129,12 @@ func getHelperMethodIfNeeded(functionName string) (string, bool) {
 		return `private static string ReadFileBase64(string path) {
 		return Convert.ToBase64String(System.Text.UTF8.GetBytes(File.ReadAllText(path)))
 	}`, true
+	case "filebase64sha256":
+		return `private static string ComputeFileBase64Sha256(string path) {
+		var fileData = System.Text.UTF8.GetBytes(File.ReadAllText(path))
+		var hashData = SHA256.Create().ComputeHash(fileData)
+		return Convert.ToBase64String(BitConverter.ToString(hashData).Replace("-","").ToLowerInvariant())
```
pgavlin, line 136: This is different from `sha1` below in that it converts the result of `BitConverter.ToString` to base64. Is that intentional? It would be helpful to include a description of the behavior of `filebase64sha256` in the PR comment (and perhaps a link to the source docs).

T-Vova: You were correct, there was a mistake. `filebase64sha256` is supposed to take in a file, convert it into a hash, and then convert those hash bytes into a Base64 string.

stack72 (approved):

PR #8002, opened 2021-09-20
[apitype] Add a JSON schema for deployments.

pgavlin:
This schema can be used to validate the contents of a Pulumi deployment.
If a deployment validates against this schema, it should be considered
syntactically valid, though it may contain certain classes of semantic
errors (e.g. references to unknown resources in dependency lists,
dependency cycles, etc.).

This schema is not yet used for validation in practice and may not be
entirely accurate.

These changes also add this schema (and the schemas on which it depends)
to the developer documentation. jsonschema2md.go has been updated to
support multi-file schemas.

### DIFF
--- a/sdk/go/common/apitype/property-values.json
+++ b/sdk/go/common/apitype/property-values.json
@@ -1,0 +1,231 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "https://github.com/pulumi/pulumi/blob/master/sdk/go/common/apitype/property-values.json",
+    "$anchor": "property_value",
+    "title": "Pulumi Property Value",
+    "description": "A schema for Pulumi Property values.",
+    "oneOf": [
+        {
+            "title": "Primitive property values",
+            "type": ["null", "boolean", "number", "string"],
+            "not": { "$ref": "#unknown_value" }
+        },
+        {
+            "$anchor": "unknown_value",
+            "title": "Unknown property values",
+            "const": "04da6b54-80e4-46f7-96ec-b56ff0331ba9"
+        },
+        {
+            "title": "Array property values",
+            "type": "array",
+            "items": {
+                "$ref": "#property_value"
+            }
+        },
+        {
+            "title": "Object property values",
+            "type": "object",
+            "additionalProperties": {
+                "$ref": "#property_value"
+            },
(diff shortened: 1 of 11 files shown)


pgavlin:
NOTE: the schema -> Markdown work in this PR is ongoing. Looks like there are some problems handling `oneOf`...

pgavlin:
> Looks like there are some problems handling oneOf...

Nah, I think I was wrong here. The Markdown renderer just doesn't make the nesting super obvious 😬

sdk/go/common/apitype/property-values.json
```diff
@@ -0,0 +1,232 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "https://github.com/pulumi/pulumi/blob/master/sdk/go/common/apitype/property-values.json",
+    "$anchor": "property_value",
+    "title": "Pulumi Property Value",
+    "description": "A schema for Pulumi Property values.",
```
t0yv0, line 6: OK this schema sounds extremely important in here, that is the piece of knowledge I really wanted to have formalized, props..

t0yv0:
Where do these "deployment" JSON objects fit in the architecture? e.g.

https://pulumi-developer-docs.readthedocs.io/en/latest/architecture/overview.html?highlight=deployment#the-deployment-engine

Are these state snapshots? Are these used on the wire between Deployment Engine and State Storage Backend? Are these wire-only or they are persisted forms.

t0yv0 (approved):
I'm fine merging in WIP almost-correct non-binding documentation provided we mark it as such in the documents itself that it is WIP.

pgavlin:
> Are these state snapshots? Are these used on the wire between Deployment Engine and State Storage Backend? Are these wire-only or they are persisted forms.

They are state snapshots, yes.

pgavlin:
Just rebased this on top of https://github.com/pulumi/pulumi/pull/8009, which helpfully found a pretty obvious error in the JSON schema for property values. Working through another failure now.

komalali (approved):
👏🏽 This is awesome! And useful in [areas of automation API](https://github.com/pulumi/pulumi/blob/master/sdk/nodejs/automation/stack.ts#L644) where we haven't fully defined this structure.

PR #8009, opened 2021-09-21
[testing] Add rapid generators for PropertyValues.

pgavlin:
And use those generators to test property value serialization and
deserialization paths.

### DIFF
--- a/pkg/resource/stack/deployment_test.go
+++ b/pkg/resource/stack/deployment_test.go
@@ -450,7 +450,7 @@ func TestSerializePropertyValue(t *testing.T) {
 
 func TestDeserializePropertyValue(t *testing.T) {
 	rapid.Check(t, func(t *rapid.T) {
-		v := ValueGenerator(6).Draw(t, "property value")
+		v := ObjectValueGenerator(6).Draw(t, "property value")
 		_, err := DeserializePropertyValue(v, config.NopDecrypter, config.NopEncrypter)
 		assert.NoError(t, err)
 	})
@@ -498,36 +498,36 @@ func TestRoundTripPropertyValue(t *testing.T) {
 	})
 }
 
-// UnknownGenerator generates the unknown resource.PropertyValue.
-func UnknownGenerator() *rapid.Generator {
+// UnknownObjectGenerator generates the unknown object value.
+func UnknownObjectGenerator() *rapid.Generator {
 	return rapid.Custom(func(t *rapid.T) interface{} {
 		return rapid.Just(computedValuePlaceholder).Draw(t, "unknowns")
 	})
 }
 
-// BoolGenerator generates boolean resource.PropertyValues.
-func BoolGenerator() *rapid.Generator {
+// BoolObjectGenerator generates boolean object values.
+func BoolObjectGenerator() *rapid.Generator {
 	return rapid.Custom(func(t *rapid.T) interface{} {
 		return rapid.Bool().Draw(t, "booleans")
 	})
 }
 
-// NumberGenerator generates numeric resource.PropertyValues.
-func NumberGenerator() *rapid.Generator {
+// NumberObjectGenerator generates numeric object values.
+func NumberObjectGenerator() *rapid.Generator {
 	return rapid.Custom(func(t *rapid.T) interface{} {
 		return rapid.Float64().Draw(t, "numbers")
 	})
 }
 
-// StringGenerator generates string resource.PropertyValues.
-func StringGenerator() *rapid.Generator {
+// StringObjectGenerator generates string object values.
+func StringObjectGenerator() *rapid.Generator {
 	return rapid.Custom(func(t *rapid.T) interface{} {
 		return rapid.String().Draw(t, "strings")
 	})
 }
 
-// TextAssetGenerator generates textual *resource.Asset values.
-func TextAssetGenerator() *rapid.Generator {
+// TextAssetObjectGenerator generates textual asset object values.
+func TextAssetObjectGenerator() *rapid.Generator {
 	return rapid.Custom(func(t *rapid.T) interface{} {
 		return map[string]interface{}{
 			resource.SigKey:            resource.AssetSig,
@@ -536,17 +536,17 @@ func TextAssetGenerator() *rapid.Generator {
 	})
 }
 
-// AssetGenerator generates *resource.Asset values.
-func AssetGenerator() *rapid.Generator {
-	return TextAssetGenerator()
+// AssetObjectGenerator generates asset object values.
+func AssetObjectGenerator() *rapid.Generator {
+	return TextAssetObjectGenerator()
 }
 
-// LiteralArchiveGenerator generates *resource.Archive values with literal archive contents.
-func LiteralArchiveGenerator(maxDepth int) *rapid.Generator {
+// LiteralArchiveObjectGenerator generates archive object values with literal archive contents.
+func LiteralArchiveObjectGenerator(maxDepth int) *rapid.Generator {
 	return rapid.Custom(func(t *rapid.T) map[string]interface{} {
 		var contentsGenerator *rapid.Generator
 		if maxDepth > 0 {
-			contentsGenerator = rapid.MapOfN(rapid.StringMatching(`^(/[^[:cntrl:]/]+)*/?[^[:cntrl:]/]+$`), rapid.OneOf(AssetGenerator(), ArchiveGenerator(maxDepth-1)), 0, 16)
+			contentsGenerator = rapid.MapOfN(rapid.StringMatching(`^(/[^[:cntrl:]/]+)*/?[^[:cntrl:]/]+$`), rapid.OneOf(AssetObjectGenerator(), ArchiveObjectGenerator(maxDepth-1)), 0, 16)
 		} else {
 			contentsGenerator = rapid.Just(map[string]interface{}{})
 		}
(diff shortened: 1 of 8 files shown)


pgavlin:
Note that the roundtrip test is likely to fail. The serializer currently cannot roundtrip output values, as it serializes them as simple unknowns. cc @justinvp

sdk/go/common/resource/testing/rapid.go
```diff
@@ -0,0 +1,226 @@
...
+// IDGenerator generates legal resource.ID values.
+func IDGenerator() *rapid.Generator {
+	return rapid.Custom(func(t *rapid.T) resource.ID {
+		return resource.ID(rapid.String().Draw(t, "ids").(string))
+	})
+}
+
+// SemverStringGenerator generates legal semver strings.
+func SemverStringGenerator() *rapid.Generator {
+	return rapid.StringMatching(`^v?(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$`)
```
t0yv0, line 40: L40 is good for T-shirt design

sdk/go/common/resource/testing/rapid.go
```diff
@@ -0,0 +1,226 @@
...
+}
+
+// UnknownPropertyGenerator generates the unknown resource.PropertyValue.
+func UnknownPropertyGenerator() *rapid.Generator {
+	return rapid.Custom(func(t *rapid.T) resource.PropertyValue {
```
t0yv0, line 45: Could use `rapid.Just(resource.MakeComputed(resource.NewStringProperty("")))`? The outer Custom wrapper it seems to me makes it go over fresh new values instead of reusing a single value. The difference wouldn't be observable if consumers never used pointer equality on the value, but maybe they do.

sdk/go/common/resource/testing/rapid.go
```diff
@@ -0,0 +1,226 @@
...
+		return resource.NewArchiveProperty(ArchiveGenerator(maxDepth).Draw(t, "archives").(*resource.Archive))
+	})
+}
+
+// ResourceReferenceGenerator generates resource.ResourceReference values.
+func ResourceReferenceGenerator() *rapid.Generator {
+	return rapid.Custom(func(t *rapid.T) resource.ResourceReference {
+		return resource.ResourceReference{
+			URN:            URNGenerator().Draw(t, "referenced URN").(resource.URN),
```
t0yv0, line 130: I'd worry about implicit empty values like "" if they're allowed/can occur in any of these fields, we'd need to prod rapid to make them appear more often. Random strings shrink to "" but it might miss them easily, and regex-driven strings don't cover "" if it does not match.

pkg/resource/stack/deployment_test.go
```diff
@@ -435,3 +438,196 @@ func TestDeserializeInvalidResourceErrors(t *testing.T) {
...
+		require.NoError(t, err)
+
+		var wireObject interface{}
+		err = json.Unmarshal(wire, &wireObject)
+		require.NoError(t, err)
+
+		deserialized, err := DeserializePropertyValue(wireObject, config.NopDecrypter, config.NopEncrypter)
+		require.NoError(t, err)
+
+		resource_testing.AssertEqualPropertyValues(t, replaceOutputsWithComputed(original), deserialized)
```
t0yv0, line 496: 👍

sdk/go/common/resource/plugin/rpc_rapid_test.go
```diff
@@ -17,70 +17,298 @@ package plugin
...
+		}
+	})
+}
+
+// AssetGenerator generates *resource.Asset values.
+func AssetGenerator() *rapid.Generator {
+	return TextAssetGenerator()
+}
+
+// LiteralArchiveGenerator generates *resource.Archive values with literal archive contents.
```
t0yv0, line 141: These aren't *resource.Archive values but *structpb.Value values right that represent the gRPC wire form of the *resource.Archive?

t0yv0 (approved):
I like where this is going.

Let me reconstruct the narrative here.

So Pulumi has an important type PropertyValue that:

- is the basis of $lang - engine RegisterResource protocol

- is the basis of the diff algo on resources

- figures in state files

The grammar of these is defined by resource.PropertyValue constructors and soon-to-be-added matching JSON schema.

These representations figure in the PR:

- native strongly-typed PropertyValue representation
- JSON-like representation (parsed as `map[string]interface{}` in Go but equivalently serializable to JSON)
- protobuf pb.Struct representation

The PR tests the turnaround properties on these representations.

The representations not covered pertain to no-Go language SDKs; essentially every supported  $lang talks the pb.Struct representation.

I was a bit confused why we need to define rapid generators over JSON and protobuf repr's if we believe these values only ever originate from the strongly-typed representation but it makes sense if we need to double-check the wire repr or we plant to evolve the wire repr somewhat separately and not just as a rigid function fo the strongly-typed repr.

sdk/go/common/resource/testing/rapid.go
pgavlin: I've used a `Custom` wrapper here to ensure that all of the `PropertyGenerator` functions return a `Generator` that returns values of type `resource.PropertyValue`. This would be clearer if we had generics (indeed, I sort of want to fork `rapid` and port it to use Go generics just to see what the usability is like).

sdk/go/common/resource/testing/rapid.go
```diff
@@ -0,0 +1,226 @@
...
+		return resource.NewAssetProperty(AssetGenerator().Draw(t, "assets").(*resource.Asset))
+	})
+}
+
+// LiteralArchiveGenerator generates *resource.Archive values with literal archive contents.
+func LiteralArchiveGenerator(maxDepth int) *rapid.Generator {
+	return rapid.Custom(func(t *rapid.T) *resource.Archive {
+		var contentsGenerator *rapid.Generator
+		if maxDepth > 0 {
+			contentsGenerator = rapid.MapOfN(rapid.StringMatching(`^(/[^[:cntrl:]/]+)*/?[^[:cntrl:]/]+$`), rapid.OneOf(AssetGenerator(), ArchiveGenerator(maxDepth-1)), 0, 16)
```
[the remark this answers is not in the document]
pgavlin: It might make a difference in terms of allocation volume--it'd be worth experimenting.

sdk/go/common/resource/testing/rapid.go
pgavlin: > I'd worry about implicit empty values like "" if they're allowed/can occur in any of these fields, we'd need to prod rapid to make them appear more often. Random strings shrink to "" but it might miss them easily, and regex-driven strings don't cover "" if it does not match.

Absolutely. I think that for the most part we're in the clear with these generators. The one exception that comes to mind is the `parentType` portion of a URN, which is allowed to be empty but will currently always be a legal type token. I'll adjust that.

sdk/go/common/resource/testing/rapid.go
```diff
@@ -0,0 +1,226 @@
...
+	return rapid.Custom(func(t *rapid.T) resource.PropertyValue {
+		return resource.NewSecretProperty(&resource.Secret{
+			Element: PropertyValueGenerator(maxDepth-1).Draw(t, "secret element").(resource.PropertyValue),
+		})
+	})
+}
+
+// PropertyValueGenerator generates arbitrary resource.PropertyValues. The maxDepth parameter controls the maximum
+// number of times the generator may recur.
+func PropertyValueGenerator(maxDepth int) *rapid.Generator {
```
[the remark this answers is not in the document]
pgavlin: Once this PR lands I intend to use the grammar here to validate the JSON schema :D

pkg/resource/stack/deployment_test.go
```diff
@@ -435,3 +438,196 @@ func TestDeserializeInvalidResourceErrors(t *testing.T) {
...
+func TestDeserializePropertyValue(t *testing.T) {
+	rapid.Check(t, func(t *rapid.T) {
+		v := ValueGenerator(6).Draw(t, "property value")
+		_, err := DeserializePropertyValue(v, config.NopDecrypter, config.NopEncrypter)
+		assert.NoError(t, err)
+	})
+
+}
+
+func replaceOutputsWithComputed(v resource.PropertyValue) resource.PropertyValue {
```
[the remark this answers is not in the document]
pgavlin: Yeah, I agree. I didn't want to add that with this PR, though :)

sdk/go/common/resource/properties_diff.go
```diff
@@ -342,6 +342,33 @@ func (v PropertyValue) DeepEquals(other PropertyValue) bool {
...
+		}
+		if vo.Secret != oo.Secret {
+			return false
+		}
+
+		if len(vo.Dependencies) != len(oo.Dependencies) {
+			return false
+		}
+		for i, dep := range vo.Dependencies {
+			if dep != oo.Dependencies[i] {
```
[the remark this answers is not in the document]
pgavlin: There are assumed to be sorted, yes--I'll add a comment. We should enforce that property somewhere (cc @justinvp).

pgavlin:
> I was a bit confused why we need to define rapid generators over JSON and protobuf repr's if we believe these values only ever originate from the strongly-typed representation but it makes sense if we need to double-check the wire repr or we plant to evolve the wire repr somewhat separately and not just as a rigid function fo the strongly-typed repr.

I added those parallel grammars because I have a hard time trusting that roundtrip/turnaround testing is sufficiently foolproof. I think that sort of testing leaves the door open for bugs in the encoder that match bugs in the decoder s.t. they cause errors that the test cannot detect--and indeed, I think we've seen that before with resource references. Using a parallel grammar in principle guards against this scenario.

sdk/go/common/resource/plugin/rpc_rapid_test.go
pgavlin: Ah, oops--I meant to fix up these comments.

sdk/go/common/resource/testing/rapid.go
```diff
@@ -0,0 +1,226 @@
...
+)
+
+// TypeGenerator generates legal tokens.Type values.
+func TypeGenerator() *rapid.Generator {
+	return rapid.Custom(func(t *rapid.T) tokens.Type {
+		return tokens.Type(rapid.StringMatching(`^[a-zA-Z][-a-zA-Z0-9_]*:([^0-9][a-zA-Z0-9._/]*)?:[^0-9][a-zA-Z0-9._/]*$`).Draw(t, "type token").(string))
+	})
+}
+
+// URNGenerator generates legal resource.URN values.
```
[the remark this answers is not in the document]
t0yv0: Nice!

t0yv0:
Awesome in that case a bit of redundancy is intentional and will help us make sure we only make changes intentionally here.

PR #8010, opened 2021-09-21
Switch from golint to revive

iwahbe:
# Description

Update the go linter from `golint` to `revive`. 

## Checklist


justinvp (approved):
Thanks!

PR #8014, opened 2021-09-21
Enable output values by default

### LINKED ISSUE #8007: Output values: Enable feature by default
Enable the feature by default and change the polarity of the envvar (from `PULUMI_ENABLE_OUTPUT_VALUES` to `PULUMI_DISABLE_OUTPUT_VALUES`).

Part of #7434

justinvp:
Enable output values by default in the resource monitor and change the polarity of the envvar from `PULUMI_ENABLE_OUTPUT_VALUES` to `PULUMI_DISABLE_OUTPUT_VALUES`.

We've had a cron job running in the `pulumi/examples` repro with the feature both enabled and disabled, and everything is looking good on those runs (as expected), so we're comfortable enabling it by default.

Fixes #8007

### DIFF
--- a/sdk/nodejs/tests/runtime/props.spec.ts
+++ b/sdk/nodejs/tests/runtime/props.spec.ts
@@ -161,12 +161,32 @@ describe("runtime", () => {
                                         ...isSecret && { secret: isSecret },
                                         ...(deps.length > 0) && { dependencies: deps },
                                     },
+                                    expectedRoundTrip: new Output(resources,
+                                        Promise.resolve(isKnown ? tv.expected : undefined), Promise.resolve(isKnown),
+                                        Promise.resolve(isSecret), Promise.resolve([])),
                                 };
                             }
                         }
                     }
                 }
             }
+
+            async function assertOutputsEqual<T>(a: Output<T>, e: Output<T>) {
+                async function urns(res: Set<Resource>): Promise<Set<string>> {
+                    const result = new Set<string>();
+                    for (const r of res) {
+                        result.add(await r.urn.promise());
+                    }
+                    return result;
+                }
+
+                assert.deepStrictEqual(await urns(a.resources()), await urns(e.resources()));
+                assert.deepStrictEqual(await a.isKnown, await e.isKnown);
+                assert.deepStrictEqual(await a.promise(), await e.promise());
+                assert.deepStrictEqual(await a.isSecret, await e.isSecret);
+                assert.deepStrictEqual(await urns(await a.allResources!()), await urns(await e.allResources!()));
+            }
+
             for (const test of generateTests()) {
                 it(`marshals ${test.name} correctly`, asyncTest(async () => {
                     runtime._setTestModeEnabled(true);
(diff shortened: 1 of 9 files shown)


t0yv0 (approved):
📿

pgavlin (approved):

justinvp:
@pgavlin, @t0yv0, please take another look. Some of the existing MLC tests failed when output values were enabled by default and the subsequent commits here address the problems.

1. When marshaling unknown output values and the `KeepOutputValues` option isn't set, we'd fall back to marshaling the value as if it was `resource.MakeComputed(v.OutputValue().Element)`. However, this is not the behavior we want because that ends up being marshaled as a null value when `v.OutputValue().Element` is `nil`:

     https://github.com/pulumi/pulumi/blob/0c0684af5cdca67617343f5e975d861f4d4e743b/sdk/go/common/resource/plugin/rpc.go#L247-L250

    To address this, I've changed it to always fall back to marshaling `resource.MakeComputed(resource.NewStringProperty(""))` so that the marshaled value is the unknown string sentinel value (consistent with what the SDKs do).

2. `MarshalProperties` was skipping `v.IsOutput()` property values when marshaling properties to providers that do not accept outputs. More details are in the commit message, but to address this, I've added a new `MarshalOptions.DontSkipOutputs` option, which is now specified when marshaling properties in the engine when calling a provider's `Construct` and `Call`.

3. There are cases inside the Node.js and Python SDKs where the serialized inputs are deserialized and returned as outputs. When that happens, we were erroring when there were output values because we hadn't implemented the output value deserialization yet in these SDKs. I've added commits that do that.

sdk/nodejs/tests/runtime/props.spec.ts
```diff
@@ -161,12 +161,32 @@ describe("runtime", () => {
                                         ...isSecret && { secret: isSecret },
                                         ...(deps.length > 0) && { dependencies: deps },
                                     },
+                                    expectedRoundTrip: new Output(resources,
+                                        Promise.resolve(isKnown ? tv.expected : undefined), Promise.resolve(isKnown),
+                                        Promise.resolve(isSecret), Promise.resolve([])),
```
t0yv0, line 166: The last element is weird, I am guessing this corresponds to
```
    /** @internal */
    public constructor(
        resources: Set<Resource> | Resource[] | Resource,
        promise: Promise<T>,
        isKnown: Promise<boolean>,
        isSecret: Promise<boolean>,
        allResources: Promise<Set<Resource> | Resource[] | Resource> | undefined) {
```

Is it possible to skip it (default undefined)? 

It's weird that the constructor has two parameters where resources can be passed in but there are probably reasons.

justinvp: > It's weird that the constructor has two parameters where resources can be passed in but there are probably reasons.

Yeah, it's a little strange. It was introduced in #3680. I'm not entirely familiar with the backstory, but looks like it was done this way to address some SxS problems.

> Is it possible to skip it (default undefined)?

We might be able to make it optional with a default value of undefined.

PR #8017, opened 2021-09-22
Use case insensitive checks when detecting duplicate type tokens

EvanBoyle:
# Description
https://github.com/pulumi/pulumi/pull/7857 introduced some logic to detect duplicate type tokens. However, the check was not case insensitive and apparently we have duplicate type tokens in different casings out in the wild such as the AWS provider:

- `"aws:ec2/NetworkInterfaceAttachment:NetworkInterfaceAttachment"`
- `"aws:ec2/networkInterfaceAttachment:NetworkInterfaceAttachment"`

This breaks our duplicate detection logic. 

This is a tactical fix with passing tests that unblocks the AWS issue. My guess is that we should probably convert all tokens to lowercase representation at schema binding time, but I will leave that to the platform team as a design exercise after we fix the user-facing P1. 

Fixes https://github.com/pulumi/pulumi-aws/issues/1632

## Checklist


iwahbe (approved):
I'm not sure what the `Then` clause is needed for. The output will already be compiled as a test of validity. What additional tests are we running?

Otherwise I approve this hot fix.

EvanBoyle:
> I'm not sure what the Then clause is needed for. 

I am not sure either. This test case is a copy of an existing test case verbatim, with a case insensitive schema change. It is possible that the `Then` clause should've been cleaned up when the compilation changes were made but was missed.

komalali (approved):

EvanBoyle:
azure native failures due to known issue that will be fixed in https://github.com/pulumi/pulumi/pull/7985

komalali:
Yes but the failures in https://github.com/pulumi/pulumi/runs/3670061955 look related to these changes

EvanBoyle:
Indeed, was just a test bug. Forgot to capitalize a second reference to the type: fb05dbb

PR #8025, opened 2021-09-22
Fix master build

t0yv0:
# Description

Follow up to undo the effects of a recent bad merge.

## Checklist

### DIFF
--- a/pkg/codegen/internal/test/sdk_driver.go
+++ b/pkg/codegen/internal/test/sdk_driver.go
@@ -71,17 +71,6 @@ var sdkTests = []sdkTest{
 	{
 		Directory:   "resource-args-python-case-insensitive",
 		Description: "Resource args with same named resource and type case insensitive",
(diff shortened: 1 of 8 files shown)


pkg/codegen/internal/test/sdk_driver.go
```diff
@@ -71,17 +71,6 @@ var sdkTests = []sdkTest{
 	{
 		Directory:   "resource-args-python-case-insensitive",
 		Description: "Resource args with same named resource and type case insensitive",
-		Then: map[string]ThenFunc{
-			"go": func(t *testing.T, testDir string) {
-				cmd := exec.Command("go", "test", "./...")
-				cmd.Dir = filepath.Join(testDir, "go-program")
```
t0yv0, line 77: Let me check if these go-program things are migrated into unit tests.

komalali (approved):

PR #8027, opened 2021-09-22
Update SDK version in pulumi/pkg

komalali:
# Description

Fixes # (issue)

## Checklist


stack72 (approved):

pkg/go.mod
```diff
@@ -34,7 +34,7 @@ require (
 	github.com/opentracing/opentracing-go v1.1.0
 	github.com/pgavlin/goldmark v1.1.33-0.20200616210433-b5eb04559386
 	github.com/pkg/errors v0.9.1
-	github.com/pulumi/pulumi/sdk/v3 v3.3.1
+	github.com/pulumi/pulumi/sdk/v3 v3.13.0
```
t0yv0, line 37: Curious, does this one matter if it gets replaced by "replace" directive?

komalali: replace only works locally when pointing to another dir

pgavlin: I think Anoton's question was along the lines of "how does the version in this stanza interact with a `replace` directive that mentions `github.com/pulumi/pulumi/sdk/v3`". @t0yv0 is that right?

t0yv0: Right, I perhaps naively assume that a replace directive of the sort makes it irrelevant what version we put in here so we might as well put v0.0.0 but perhaps there's a situation where this version is in fact relevant.

pgavlin: > Right, I perhaps naively assume that a replace directive of the sort makes it irrelevant what version we put in here so we might as well put v0.0.0 but perhaps there's a situation where this version is in fact relevant.

`replace` directives are only respected within the context of the repository that contains the `go.mod` file. They do not affect consumers of the go.mod file. From https://golang.org/ref/mod#go-mod-file-replace:

> replace directives only apply in the main module’s go.mod file and are ignored in other modules. See Minimal version selection for details.

See https://github.com/pulumi/pulumi/issues/6383 for some details on how this affects our setup.

pgavlin: Or to answer more directly: the version we have in here is relevant to users that reference `github.com/pulumi/pulumi/pkg/v3` without also specifying a `replace` for `github.com/pulumi/pulumi/sdk/v3` (i.e. the mainline scenario).

PR #8032, opened 2021-09-22
[auto/python] - Fix a bug in printing stack.

komalali:
# Description

This repro'd with `print(stack)` in a local program so I added that to the test.

## Checklist

### DIFF
--- a/sdk/python/lib/test/automation/test_local_workspace.py
+++ b/sdk/python/lib/test/automation/test_local_workspace.py
@@ -327,7 +327,7 @@ class TestLocalWorkspace(unittest.TestCase):
         stack_name = stack_namer(project_name)
         work_dir = test_path("data", project_name)
         stack = create_stack(stack_name, work_dir=work_dir)
-        print(stack)
+        self.assertIsNone(print(stack))
 
         config: ConfigMap = {
             "bar": ConfigValue(value="abc"),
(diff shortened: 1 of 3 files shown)


justinvp (approved):

pgavlin:
> This repro'd with print(stack) in a local program so I added that to the test.

What exactly was the issue? Did `print(stack)` throw?

komalali:
@pgavlin Yes, `print(stack)` threw for local programs because the `__repr__` function was looking for the program's name without checking if there was a program at all: https://github.com/pulumi/pulumi/pull/8032/files#diff-333d7645bb6845c14765cb4e3f3fac2b0fbe65587203a09cf07d5170e17ac2dfL104

sdk/python/lib/test/automation/test_local_workspace.py
```diff
@@ -327,6 +327,7 @@ def test_stack_lifecycle_local_program(self):
         stack_name = stack_namer(project_name)
         work_dir = test_path("data", project_name)
         stack = create_stack(stack_name, work_dir=work_dir)
+        print(stack)
```
pgavlin, line 330: Maybe add a note here to the effect of "`print(stack)` should not throw"

pgavlin (approved):
LGTM

sdk/python/lib/test/automation/test_local_workspace.py
komalali: Added an `assert` to clarify it's a part of the test.

Issue #7804, opened 2021-08-20
[Epic] Code generation quality


EvanBoyle:
Code generation quality spans (currently) four languages, SDK generation, and examples generation. We have substantial gaps in coverage that have led to some painful regressions in recent history. We have opportunities to make targeted testing infrastructure improvements to make it safer to make changes in this area, and to discover bugs sooner.

This becomes more important as we think about community and multi-language authors consuming this surface area, not just core pulumi team members.

While there are some advanced techniques we could use such as generative property testing, we have some low hanging fruit we should tackle first:

- [x] https://github.com/pulumi/pulumi/issues/7801
- [x] https://github.com/pulumi/pulumi/issues/7802

Adding an output compilation step in our tests will bring more confidence when making changes, and prevent future regressions. A campaign to bring more examples under the suite will simultaneously expand coverage and improve the quality of our documentation. 

If we make it this far, we may want to consider closing the loop by running a preview on some of the generated programs. However, there is more complexity there (acquiring a provider binary, cloud credentials, etc) which is not well suited to the current pulumi/pulumi test infrastructure.

iwahbe:
I'm keeping a running list of issues found from #7802:
- [ ] https://github.com/pulumi/pulumi/issues/8062

PR #8036, opened 2021-09-23
iwahbe/7802/compile program generator test output

### LINKED ISSUE #7802: Compile output from program generator tests
`TestGenerateProgram` and the common machinery in [`TestProgramCodegen`](https://github.com/pulumi/pulumi/blob/1409d884383c9e2f8bee4e2a20bac31fd36104f9/pkg/codegen/internal/test/program_driver.go#L104) only does string matching as verification. It does not exercise the generated program or even make sure that it compiles.

Unfortuantely, it seems like the default way that we write these tests makes to overlook errors as we generate test output:

1. Author test schema
2. Write test harness
3. Run test suite with PULUMI_ACCEPT=true go test ./... to generate files to disk.

No where in this flow is process to make sure that the generated code works or compiles.

We should add an additional step at the end of these tests to use language specific tooling, to compile, typecheck, or build whatever artifacts are appropriate for each supported language.

Follow up to errors introduced by  https://github.com/pulumi/pulumi/pull/7059 related to https://github.com/pulumi/pulumi/issues/7735

iwahbe:
# Description

We run a compiler over our generated code. 

Fixes #7802

## Checklist


iwahbe:
Azure native downstream failure is unrelated.

iwahbe:
`azure-native`  downstream failures are unrelated.

pkg/codegen/dotnet/gen_program_test.go
```diff
@@ -1,11 +1,97 @@
...
+
+	// Add dependencies (based on directory name)
+	if pkg, pkgVersion := packagesFromTestName(filepath.Base(dir)); pkg != "" {
+		err = integration.RunCommand(t, "create dotnet project",
+			[]string{ex, "add", "package", pkg, "--version", pkgVersion},
+			dir, &integration.ProgramTestOptions{})
+		assert.NoError(t, err, "Failed to add dependency %q %q", pkg, pkgVersion)
+	} else {
+		err = integration.RunCommand(t, "add sdk ref",
+			[]string{ex, "add", "reference", "../../../../../../sdk/dotnet/Pulumi"},
```
t0yv0, line 57: This is nice! I'd want that in SDK codegen tests also, to link against in-repo Pulumi SDK.

pkg/codegen/dotnet/gen_program_test.go
```diff
@@ -1,11 +1,97 @@
...
+	var err error
+	dir := filepath.Dir(path)
+
+	ex, err := executable.FindExecutable("dotnet")
+	assert.NoError(t, err, "Failed to find dotnet executable")
+
+	projectFile := filepath.Join(dir, filepath.Base(dir)+".csproj")
+	if _, err := ioutil.ReadFile(projectFile); os.IsNotExist(err) {
+		defer func() {
+			err = os.Remove(projectFile)
```
t0yv0, line 39: Philosophically I have a slight preference for keeping this files in rather than always automatically deleting them, in case something is failing I would like to have the ability to interact with the generated program as a normal program, edit it and experiment with it.

pkg/codegen/dotnet/gen_program_test.go
```diff
@@ -1,11 +1,97 @@
...
+// Example:
+// 	"aws-eks-pp" => ("Pulumi.Aws", 4.21.1)
+// 	"azure-sa-pp" => ("Pulumi.Azure", 4.21.1)
+// 	"resource-options-pp" => ("","")
+//
+// Note: While we could instead do this by using the generateMetaData function
+// for each language, we are trying not to expand the functionality under test.
+func packagesFromTestName(name string) (string, string) {
+	if strings.Contains(name, "aws") {
+		return "Pulumi.Aws", "4.21.1"
```
t0yv0, line 86: Hmm, this is interesting.. Do these tests truly depend on these providers.

pkg/codegen/go/gen_program_test.go
```diff
@@ -31,56 +41,72 @@ func TestCollectImports(t *testing.T) {
...
+	_, err = ioutil.ReadFile("go.mod")
+	if os.IsNotExist(err) {
+		defer func() {
+			// If we created the module, we also remove the module
+			err = os.Remove(filepath.Join(dir, "go.mod"))
+			assert.NoError(t, err)
+			err = os.Remove(filepath.Join(dir, "go.sum"))
+			assert.NoError(t, err)
+		}()
+		err = integration.RunCommand(t, "generate go.mod",
```
t0yv0, line 96: Similarly to the compile check for SDK codegen, this needs `go mod edit` pointing to in-repo Go SDK with a replace directive.

pkg/codegen/internal/test/Makefile
```diff
@@ -0,0 +1,33 @@
+.PHONY = update_random update_aws update_azure update_azure_native update_kubernetes
+
+RANDOM_SCHEMA       := "https://raw.githubusercontent.com/pulumi/pulumi-random/master/provider/cmd/pulumi-resource-random/schema.json"
```
t0yv0, line 3: Is the choice for a floating reference (master) as opposed to a fixed reference (v1.2.3.) intentional here?

Does it conflict with choosing to pin versions of packages in a Go file above?

pkg/codegen/internal/test/Makefile
```diff
@@ -0,0 +1,33 @@
+.PHONY = update_random update_aws update_azure update_azure_native update_kubernetes
```
t0yv0, line 1: I'd consider rewriting this makefile in Go can be very much worth it, it is a little extra effort but is much more easily cross-platform.

pkg/codegen/internal/test/Makefile
```diff
@@ -0,0 +1,33 @@
...
+update_azure:
+	cd testdata && curl ${AZURE_SCHEMA} > azure.json
+	$(call add_version,azure,4.18.0)
+
+update_azure_native:
+	cd testdata && curl ${AZURE_NATIVE_SCHEMA} > azure-native.json
+	$(call add_version,azure-native,1.29.0)
+
+update_kubernetes:
+	cd testdata && curl ${KUBERNETES_SCHEMA} > kubernetes.json
```
t0yv0, line 32: I'm still wondering if we depend on all the entire schemata or on some subset. I had a script lying around for extracting a subset of a schema by closing over references found in a type or two. It may be a legitimate strategy to depend on the entire schema but I'm thinking ahead to dependency loops between this package and the providers, there may be some slight downsides to that, some interesting maintenance situations where we can't build+ test one without a past version of the other.

pkg/codegen/internal/test/program_driver.go
```diff
@@ -19,78 +19,147 @@ import (
...
-		ProgramFile: "secret",
+		Name:        "aws-secret",
 		Description: "Secret",
 	},
 	{
-		ProgramFile: "functions",
+		Name:        "functions",
 		Description: "Functions",
+		SkipCompile: codegen.NewStringSet("go", "dotnet"),
+		// Blocked on go:
```
t0yv0, line 131: This looks very unfortunate, worth a new issue?

pkg/codegen/nodejs/gen_program_test.go
```diff
@@ -1,11 +1,71 @@
...
+		err = os.RemoveAll(nodeModules)
+		assert.NoError(t, err, "Failed to delete %s", nodeModules)
+		packageJSON := filepath.Join(dir, "package.json")
+		err = os.Remove(packageJSON)
+		assert.NoError(t, err, "Failed to delete %s", packageJSON)
+		yarnLock := filepath.Join(dir, "yarn.lock")
+		err = os.Remove(yarnLock)
+		assert.NoError(t, err, "Failed to delete %s", yarnLock)
+	}()
+	err = integration.RunCommand(t, "yarn add and install",
```
t0yv0, line 48: Not seeing `yarn link @pulumi/pulumi` somewhere here, which would make it reference in-repo SDK.

pkg/codegen/nodejs/gen_program_test.go
```diff
@@ -1,11 +1,71 @@
...
+		[]string{ex, "add", pkg}, dir, &integration.ProgramTestOptions{})
+	assert.NoError(t, err, "Could not install package: %q", pkg)
+	err = integration.RunCommand(t, "tsc check",
+		[]string{ex, "run", "tsc", "--noEmit", filepath.Base(path)}, dir, &integration.ProgramTestOptions{})
+	assert.NoError(t, err, "Failed to build %q", path)
+}
+
+// packagesFromTestName attempts to figure out what package should be imported
+// from the name of the test.
+func packagesFromTestName(name string) (string, string) {
```
t0yv0, line 58: Noting more pins here.

pkg/codegen/go/gen_program_test.go
```diff
@@ -31,56 +41,72 @@ func TestCollectImports(t *testing.T) {
...
+		err = integration.RunCommand(t, "go tidy",
+			[]string{ex, "mod", "tidy"},
+			dir, &integration.ProgramTestOptions{})
+		assert.NoError(t, err)
+	} else {
+		assert.NoError(t, err)
 	}
-	t.Fatalf("test file not found")
-	return nil
+	err = integration.RunCommand(t, "test build", []string{ex, "build", "-v", "all"},
```
t0yv0, line 107: Thanks for -v all, sounds pretty foolproof for compiling the recursively nested files if any.

t0yv0:
This looks very helpful! Thank you for taking this on. This puts you ahead of me in the area of programgen familiarity as I had not had a chance to use it and work with it. The changes look pretty good to me, I would love us to start having the compile-check coverage. 

My comments are more of conversation starters; if something resonates as a good idea but takes more than a few minutes to address, let's backlog and cross-ref in this style (as Justin showed me), when next working on features in the area of programgen we can pick these up naturally: 

```
	// TODO[pulumi/pulumi#7127]: Re-enabled the warning.
```

pkg/codegen/dotnet/gen_program_test.go
iwahbe: They depend on them for type information, so they really do depend on them. It's unfortunate but I can't think of a way around it right now without rewriting the tests.

pgavlin: I'd really prefer that we not do this via string matching on filenames:
- it imposes implicit structure
- it's fragile with respect to the contents of the tests

Instead, I'd rather we drive this from the program code generators themselves. However, I realize that's a bigger change, so I think that this is okay provided that we file an issue to come back and clean this up.

pkg/codegen/internal/test/program_driver.go
iwahbe: My plan is to do a pass after this merges and file an issue for each case that doesn't compile.

pkg/codegen/dotnet/gen_program_test.go
iwahbe: We can do that, but it's slightly more complicated. In the SDK codegen comp tests we rely on the SDK generator to build the package (including it's metadata). My desire was to avoid modifying the generated package metadata. There it's a trade off, changing what part of the system is put under test. Here we need to manually build the metadata (so we may as well do it right).

pkg/codegen/internal/test/Makefile
iwahbe: I think the solution to this is to depend only on pinned versions of the schema (which we need to do anyway). Upgrading the schema these tests depend on will need to be a manual process. As long as our providers exist and git keeps their release history, we should be fine. If we ever reach a point were we can't use the old schema, all these tests will need to be rewritten anyway, so I don't think we need to worry about it.

pkg/codegen/internal/test/Makefile
iwahbe: No. This has changed to a fixed version.

iwahbe:
@t0yv0 I rewrote the make file in `go`. This gives us a single source of truth for schema version, and will ensure tests only run with the correct schema.

pkg/codegen/internal/test/Makefile
komalali: >If we ever reach a point were we can't use the old schema, all these tests will need to be rewritten anyway

FWIW, this isn't true. We've updated the schema when we have added new features, and we certainly don't have to rewrite all the tests to update the schema version.

iwahbe: Sorry @komalali, what I wrote was unclear. I meant if we ever made a breaking change to schemas, such that the current version of codegen is not compatible with the schemas currently used in the tests, we would need to rewrite the tests anyway.

pkg/codegen/internal/test/Makefile
pgavlin: Is this actually used cross-platform? I'd prefer not to change the way we build things for one specific case.

pgavlin: (even if it is used cross-platform, I don't think that this is the right time or place to add more ways that SW gets built in this repo--I'd much rather stick with the status quo for now)

pkg/codegen/internal/test/program_driver.go
```diff
@@ -104,62 +173,41 @@ type langConfig struct {
...
-					outputFile: "main.go",
-				}
-			case "dotnet":
-				cfg = langConfig{
-					extension:  "cs",
-					outputFile: "MyStack.cs",
-				}
-			default:
-				t.Fatalf("language %s not recognized", language)
+			testDir := filepath.Join(testdataPath, tt.Name+"-pp")
```
pgavlin, line 192: Thank you for restructuring the test cases into directories 👍

pgavlin: Might be worth going even further and putting each language's baseline in its own directory ala SDK tests

pkg/codegen/dotnet/gen_program_test.go
pgavlin: +1

iwahbe: The system now creates metadata files before each test. This way after a test you can `cd` into the test area and manually build and modify the test. 

Because these files are recreated after each test, metadata files are still ephemeral.

pkg/codegen/internal/test/program_driver.go
iwahbe: New issues are here: https://github.com/pulumi/pulumi/issues/7804#issuecomment-928135309

.gitignore
```diff
@@ -24,6 +24,15 @@ coverage.cov
...
 pkg/codegen/internal/test/testdata/*/typecheck/**
 
+# We don't need to keep testing python __pycache__ either either
+pkg/codegen/internal/**/__pycache__
+pkg/codegen/internal/test/testdata/*-pp/*/*.csproj
+pkg/codegen/internal/test/testdata/*-pp/*/go.mod
+pkg/codegen/internal/test/testdata/*-pp/*/go.sum
+pkg/codegen/internal/test/testdata/*-pp/*/Program.cs
+pkg/codegen/internal/test/testdata/*-pp/*/package.json
+pkg/codegen/internal/test/testdata/*-pp/*/go.sum
```
pgavlin, line 34: I thought that the `__pycache__` line at the bottom would catch `__pycache__` at any level of nesting. Does it not?

pkg/codegen/dotnet/gen_program_test.go
```diff
@@ -1,11 +1,96 @@
...
+	dir := filepath.Dir(path)
+
+	ex, err := executable.FindExecutable("dotnet")
+	assert.NoError(t, err, "Failed to find dotnet executable")
+
+	// We create a new cs-project each time the test is run.
+	projectFile := filepath.Join(dir, filepath.Base(dir)+".csproj")
+	programFile := filepath.Join(dir, "Program.cs")
+	if err = os.Remove(projectFile); !os.IsNotExist(err) {
+		assert.NoError(t, err)
```
pgavlin, line 39: Nit: `require.NoError` will terminate the test; `assert.NoError` will continue.

```suggestion
		require.NoError(t, err)
```

pkg/codegen/dotnet/gen_program_test.go
```diff
@@ -1,11 +1,96 @@
...
+	}
+	if err = os.Remove(programFile); !os.IsNotExist(err) {
+		assert.NoError(t, err)
+	}
+	err = integration.RunCommand(t, "create dotnet project",
```
pgavlin, line 44: Nice to see the `integration` framework coming in handy here :)

pkg/codegen/dotnet/gen_program_test.go
pgavlin: Same feedback throughout wherever errors really ought to terminate the test. TBH, it might be best to have the `Check` functions return an `error` and let the caller `assert` on it.

pkg/codegen/dotnet/gen_program_test.go
```diff
@@ -1,11 +1,96 @@
...
+// packagesFromName attempts to figure out what package should be imported from
+// the name of a test.
+//
+// Example:
+// 	"aws-eks-pp" => ("Pulumi.Aws", 4.21.1)
+// 	"azure-sa-pp" => ("Pulumi.Azure", 4.21.1)
+// 	"resource-options-pp" => ("","")
+//
+// Note: While we could instead do this by using the generateMetaData function
+// for each language, we are trying not to expand the functionality under test.
```
pgavlin, line 82: Can we link to the cleanup issue here?

pkg/codegen/importer/testdata/cases.json
```diff
@@ -588,16 +588,20 @@
...
+                            {
+                                "key": "foobar",
+                                "value": "foobar"
+                            },
+                            {
+                                "key": "foobar",
+                                "value": "foobar"
+                            }           
+                        ]
+                    }
```
pgavlin, line 604: Is the indentation off here? The diff looks odd but I can't quite tell if this is intentional.

pgavlin (approved):
A few nits, but LGTM overall!

.gitignore
iwahbe: I think it does.

pkg/codegen/importer/testdata/cases.json
iwahbe: It's not intentional. `cases.json` does indentation vie tabs, and I put spaces. It was right for the size of tabs as displayed on my editor, but wrong for GitHub. The next push will be all tabs.

pkg/codegen/dotnet/gen_program_test.go
iwahbe: I'll fix the `require/assert`. 

I find the `require`'s clearer then returning errors. Since the compile check is the last check performed, its semantically indistinct. Being able to `assert` is helpful, for example in `defer`ed cleanups.

PR #8039, opened 2021-09-23
Simplify output-funcs codegen test

t0yv0:
# Description

Refactors tests for #5758 to take advantage of the new test framework that consolidates compile and test-checking generated code. Most of the changes are mechanical via PULUMI_ACCEPT=1. The interesting changes are consolidating the JSON schemas used for testing output-funcs into a single `schema.json` and the other interesting bits are highlighted with comments in the PR.

## Checklist


pkg/codegen/internal/test/sdk_driver.go
```diff
@@ -196,7 +220,9 @@ func TestSDKCodegen(t *testing.T, opts *SDKCodegenOptions) { // revive:disable-l
 	testDir := filepath.Join("..", "internal", "test", "testdata")
 
 	for _, tt := range sdkTests {
-		t.Run(tt.Description, func(t *testing.T) {
+		t.Run(tt.Directory, func(t *testing.T) {
```
t0yv0, line 223: Note: I'm finding it easier to type `go test -run TestGeneratePackage/output-funcs` by folder name, not description.

pkg/codegen/internal/test/testdata/provider-config-schema/python/pulumi_configstation/__init__.py
```diff
@@ -17,7 +17,7 @@
 
 _utilities.register(
     resource_modules="""
-null
+[]
```
t0yv0, line 20: Python code in utilities.py was actually treating null/None and [] differently. This is correct.

pkg/codegen/python/gen_resource_mappings.go
```diff
@@ -89,7 +89,7 @@ func makeResourceModuleInfo(pkg, mod, fqn string) resourceModuleInfo {
 }
 
 func allResourceModuleInfos(root *modContext) []resourceModuleInfo {
-	var result []resourceModuleInfo
+	result := []resourceModuleInfo{}
```
t0yv0, line 92: This is the change that avoids sending nil to Python code.

pkg/codegen/python/gen_test.go
```diff
@@ -75,6 +71,11 @@ func pyCompileCheck(t *testing.T, codeDir string) {
 	cmdOptions := integration.ProgramTestOptions{}
 	err = filepath.Walk(codeDir, func(path string, info filesystem.FileInfo, err error) error {
 		require.NoError(t, err) // an error in the walk
+
+		if info.Mode().IsDir() && info.Name() == "venv" {
```
t0yv0, line 75: I noticed that it tries to recur into `venv` and spends a lot of time in there, short-circuit it.

pkg/codegen/internal/test/sdk_driver.go
```diff
@@ -44,85 +44,109 @@ var sdkTests = []sdkTest{
...
 	},
 	{
 		Directory:   "hyphen-url",
 		Description: "A resource url with a hyphen in it's path",
+		Skip:        codegen.NewStringSet("python/test"),
+	},
+	{
+		Directory:        "output-funcs",
+		Description:      "Tests targeting the $fn_output helper code generation feature",
+		SkipCompileCheck: codegen.NewStringSet(dotnet),
```
t0yv0, line 149: I intend to fix .NET compilation rebasing https://github.com/pulumi/pulumi/pull/7899 onto this work shortly after it's merged.

pkg/codegen/internal/test/testdata/.gitignore
```diff
@@ -2,3 +2,7 @@ package-lock.json
 go.sum
 go.mod
 */go/tests
+*/python/tests
```
t0yv0, line 5: These kinds of files are copied from python-extras; ignore them in `python/` folders that do get checked in.

pkg/codegen/internal/test/sdk_driver.go
```diff
@@ -207,8 +231,21 @@ type SDKCodegenOptions struct {
 func TestSDKCodegen(t *testing.T, opts *SDKCodegenOptions) { // revive:disable-line
 	testDir := filepath.Join("..", "internal", "test", "testdata")
 
-	for _, tt := range sdkTests {
-		t.Run(tt.Description, func(t *testing.T) {
+	// Motivation for flagging parallelism: not all tests are
+	// parallel-safe yet (for example, codegen/docs tests fail),
+	// and there are concerns about memory utilizaion in CI. It
+	// can be a nice feature for developing though.
+	parallel := os.Getenv("PULUMI_PARALLEL_SDK_CODEGEN_TESTS") != ""
```
t0yv0, line 238: YMMV but it helps a lot on my laptop to use that plus `-parallel 4`.

pkg/codegen/python/gen_test.go
```diff
@@ -59,164 +56,160 @@ func TestRelPathToRelImport(t *testing.T) {
 }
 
 func TestGeneratePackage(t *testing.T) {
+	// To speed up these tests, we will generate one common
```
t0yv0, line 59: I found it to help a little bit, though it reduces isolation, I think it's worth it. The test suite rebuilds it on every run of the suite, but that's better than one per test.

pkg/codegen/python/gen_test.go
```diff
@@ -59,164 +56,160 @@ func TestRelPathToRelImport(t *testing.T) {
...
+	if err != nil {
+		return "", err
+	}
+	return filepath.Join(hereDir, "venv"), nil
+}
 
-		info, _ := pkg.Language["python"].(PackageInfo)
+// To serialize shared `venv` operations; without the lock running
+// tests with `-parallel` causes sproadic failure.
+var venvMutex *sync.Mutex = &sync.Mutex{}
```
t0yv0, line 96: Can't do `pip install -e .` in parallel on a shared venv, withohut a mutex. Unfortunate.

pkg/codegen/python/gen_test.go
```diff
@@ -59,164 +56,160 @@ func TestRelPathToRelImport(t *testing.T) {
...
 		}
+		return nil
+	})
+	require.NoError(t, err)
 
-		_, err = writer.Write([]byte(funcCode))
-		return err
-	}
+	args := append([]string{"-m", "py_compile"}, pythonFiles...)
+	test.RunCommand(t, "python syntax check", codeDir, "python", args...)
```
t0yv0, line 163: Batching this up is a real perf win.

t0yv0:
@iwahbe PTAL and let me know what the plan is with the VERISON change that got reverted from master. I understand we are disabling .NET compilation checks because of this revert. Is there a way forward to get that back in soon or should we instead work around and enable .NET compilation checks without it?

pkg/codegen/internal/test/sdk_driver.go
iwahbe: I agree that parallelism should be flagged because of memory utilization. If parallelism is known to break certain tests, we should disable it for those tests. That way we can run `PULUMI_PARALLEL_SDK_CODEGEN_TESTS=true go test ./...`.

pkg/codegen/internal/test/testdata/output-funcs/go/mypkg/funcWithListParam.go
```diff
@@ -42,14 +40,14 @@ func FuncWithListParamOutput(ctx *pulumi.Context, args FuncWithListParamOutputAr
 
 type FuncWithListParamOutputArgs struct {
 	A pulumi.StringArrayInput `pulumi:"a"`
-	B pulumi.StringPtrInput `pulumi:"b"`
+	B pulumi.StringPtrInput   `pulumi:"b"`
```
iwahbe, line 43: Do we know why the whitespace changed?

iwahbe (approved):
I'm happy with this.

 WRT the `dotnet` tests. I would like to get the version generating code merged back in, but I'm not sure exactly when that will happen. I'm going to push a workaround to reenable that soon.

pkg/codegen/internal/test/testdata/output-funcs/go/mypkg/funcWithListParam.go
t0yv0: I think something is applying gofmt now?

A large benefit of this change is that these tests now use official codegen API. Before that, the tests hooked into undocumented internals to build out half-baked support for testing function codegen. Perhaps official API does a gofmt pass.

pkg/codegen/internal/test/sdk_driver.go
t0yv0: I will take that as follow up.

PR #8049, opened 2021-09-23
[codegen/go] Rewrite cyclic types.

### LINKED ISSUE #7622: Recursive types not supported in Go codegen

pgavlin:
When computing the type name for a field of an object type, we must
ensure that we do not generate invalid recursive struct types. A struct
type T contains invalid recursion if the closure of its fields and its
struct-typed fields' fields includes a field of type T. A few examples:

Directly invalid:

	type T struct { Invalid T }

Indirectly invalid:

	type T struct { Invalid S }

	type S struct { Invalid T }

In order to avoid generating invalid struct types, we replace all
references to types involved in a cyclical definition with *T. The
examples above therefore become:

1.
```
       type T struct { Valid *T }
```

2.
```
       type T struct { Valid *S }

       type S struct { Valid *T }
```

We do this using a rewriter that turns all fields involved in reference
cycles into optional fields.

These changes also include two enhancements to the SDK codegen test
driver in the interest of making debugging more convenient.

1. If the -sdk.run flag is passed, the driver will only run the test
   with the given directory name
2. If the -sdk.no-checks flag is passed, the driver will not run
   post-generation checks

This allows the use of e.g. `dlv` to debug a particular case without
editing the driver itself, for example:

	$ go test -c -gcflags="all=-N -l" .
	$ dlv exec ./go.test -- -test.run TestGeneratePackage -sdk.run=cyclic-types -sdk.no-checks

Fixes #7622.


pkg/codegen/internal/test/testdata/cyclic-types/go/example/pulumiTypes.go
```diff
@@ -0,0 +1,581 @@
...
+	return o.ApplyT(func(v *IndirectCycleS) *IndirectCycleT {
+		if v == nil {
+			return nil
+		}
+		return v.Foo2
+	}).(IndirectCycleTPtrOutput)
+}
+
+type IndirectCycleT struct {
+	Foo3 *IndirectCycleS `pulumi:"foo3"`
```
t0yv0, line 439: both IndirectCycleS and IndirectCycleT properties get '*' treatment, so this is symmetric/deterministic, good!

t0yv0:
This looks well motivated. Coupe of notes.

1. I'm struggling to find time before Monday to give this the review it deserves, would it be ok to take a rain check on the 24h review SLA or is there urgency to get it out now.

2. Being conservative about rollout: with the recent track record of codegen changes causing some problems for providers, should we do some manual rebuilds of azure-native and google-native that report having this issue, and perhaps one other provider that *does not* have this issue to check for any breaking changes? I could help with that. Or do we trust the logic + "downstream codegen checks" here to be sufficient.

3. Is this fully non-breaking (beside the strictly necessary addition of * that the user notices in the API)? Are all derived forms working? 

4. Thinking out loud: substituting T for *T in a struct field, during recursive serialization *T will serialize to the same form as T so we're likely okay there. I'd be curious to check the generated Input helper forms, as well as cases where *T, as well as interactions of this pass with situations where T field is marked optional or pointer in the schema (so the default projection is already *T), whether the treatment is symmetric/deterministic/stable wrt changes of schema. Preliminary looks good - I'm glancing through tests and finding answers to some of these, more on  Monday.

t0yv0:
>  really solving the problem at the wrong layer

This certainly resonates and this is why I feel we need to think about unexpected potential interactions with other features a little harder, to be on the safe side here.

pgavlin:
> I'm struggling to find time before Monday to give this the review it deserves, would it be ok to take a rain check on the 24h review SLA or is there urgency to get it out now.

No worries--it's more important that this lands reliably and with robust review than that it lands now.

> Being conservative about rollout: with the recent track record of codegen changes causing some problems for providers, should we do some manual rebuilds of azure-native and google-native that report having this issue, and perhaps one other provider that does not have this issue to check for any breaking changes? I could help with that. Or do we trust the logic + "downstream codegen checks" here to be sufficient.

I think that looking at the changed files and trying to build those packages that saw changes should be sufficient.

> Is this fully non-breaking (beside the strictly necessary addition of * that the user notices in the API)? Are all derived forms working?

In some aspects this is non-breaking:
- property access expressions should continue to work, since Go uses the same syntax for accesses that automatically dereference pointers and those that do not
- existing inputs will be okay, since a `TInput` is assignable to a `TPtrInput`

Importantly, though, existing _outputs_ may break if they are passed to `ApplyT`, since the element type of the output is now `*T` instead of `T`. I don't think that there's any way to avoid this beyond changing the dynamic type checker for `ApplyT`, which is pretty unpalatable.

pkg/codegen/go/gen.go
```diff
@@ -1878,6 +1880,111 @@ func init() {
...
+
+	// If the current path does not involve a cycle, recur into the current object type.
+	if len(cycle) == 0 {
+		rewriteCyclicFields(rewritten, path, objectType)
+		return
+	}
+
+	// If we've found a cycle, mark each property involved in the cycle as optional.
+	//
+	// NOTE: this overestimates the set of properties that must be marked as optional. For example, in case (2) above,
```
t0yv0, line 1960: 👍

pkg/codegen/go/gen.go
```diff
@@ -1878,6 +1880,111 @@ func init() {
...
+		return
+	}
+
+	path = append(path, op)
+
+	// Check the current path for cycles by crawling backwards until reaching the start of the path
+	// or finding a property that is a member of the current object type.
+	var cycle []objectProperty
+	for i := len(path) - 1; i > 0; i-- {
+		if path[i].object == objectType {
```
t0yv0, line 1946: Correct to use pointer equality here? one physical object per logical type in the process?

pkg/codegen/go/gen.go
```diff
@@ -1878,6 +1880,111 @@ func init() {
...
+		return
+	}
+
+	// If we've found a cycle, mark each property involved in the cycle as optional.
+	//
+	// NOTE: this overestimates the set of properties that must be marked as optional. For example, in case (2) above,
+	// only one of T.Invalid or S.Invalid needs to be marked as optional in order to break the cycle. However, choosing
+	// a minimal set of properties that is also deterministic and resilient to changes in visit order is difficult and
+	// seems to add little value.
+	for _, p := range cycle {
```
t0yv0, line 1964: Would it make sense to log any time we do this cycle rewrite? E.g. which type and field? I'm not sure how this surfaces to codegen users but that sounds valuable to see what's happening, as a user of codegen.

t0yv0 (approved):
Code LGTM, I read through the recursion.

Test failure unrelated (merge master should fix).

Reassuring zero diff on pulumi-aws.

Expect to have very low blast radius since few cycles are identified in existing schmas. Logging rewrites might help "see" the blast radius.

I may try this out on pulumi-gcp later on today just to get in the habit of checking codegen PRs against affected providers.

pkg/codegen/go/gen.go
pgavlin: For object types, yes. The schema binder ensures that reference equality indicates type identity for these types.

pkg/codegen/go/gen.go
pgavlin: Logging might be useful, yes, but we don't really have a logging framework for the code generators, and I'd prefer not to just log to stdout/stderr. File this one under "enhancements" :)

t0yv0:
For `dlv exec ./go.test -- -test.run TestGeneratePackage -sdk.run=cyclic-types -sdk.no-checks` - not sure if this landed yet or is sitting in my straggler PRs, but I renamed the tests so that one can say `-run TestGeneratePackage/cyclic-types` e.g. test DIR ID is used in the test name, not description. The sdk.no-checks is useful though.

pgavlin:
>  I renamed the tests so that one can say -run TestGeneratePackage/cyclic-types e.g. test DIR ID is used in the test name, not description. The sdk.no-checks is useful though.

TBH I was not certain that `-run` worked with subtests--I've had mixed results with that in the past. Must've been user error, though :)

pgavlin:
(in any case, I've removed the new flag)

komalali (approved):
LGTM but you'll have to regen the codegen tests (if you've already pulled in changes in master)

PR #8053, opened 2021-09-24
[sdk/{nodejs,python}] Fix errors when testing remote components with mocks

### LINKED ISSUE #8050: [sdk/{nodejs|python}] Mock monitor indicates it supports output values leading to errors in unit tests
v3.13.0 of the Python and Node.js SDKs have support for serializing output values as special objects when the monitor supports the feature (it's disabled by default in v13.3 of the CLI unless the `PULUMI_ENABLE_OUTPUT_VALUES` envvar is set to a truthy value) when serializing inputs for remote components and method calls.

When mocks are used to unit test remote components, any outputs in the remote component's inputs will be serialized as output values because the mock monitor in both of these SDKs currently indicates it supports all features:

https://github.com/pulumi/pulumi/blob/0bcf475573f96f09d4fa8375fef1d935f439f82f/sdk/python/lib/pulumi/runtime/mocks.py#L228-L230

https://github.com/pulumi/pulumi/blob/0bcf475573f96f09d4fa8375fef1d935f439f82f/sdk/nodejs/runtime/mocks.ts#L208-L213

This can lead to errors like the following in Python when running the tests:

```
future: <Task finished name='Task-41' coro=<Output.__init__.<locals>.is_value_known() done, defined at /workspace/target/venv/lib/python3.9/site-packages/pulumi/output.py:93> exception=AssertionError('Unrecognized signature when unmarshalling resource property')>
```

What's happening is that an output in one of the of the inputs is being serialized as an output value, and then when the inputs are being deserialized inside the mock monitor implementation, the deserialization fails because support for deserializing output values has not been implemented yet in these SDKs.

We already need to add support for deserializing output values in these SDKs (for other reasons) and it's already implemented and will be landing as part of #8014. This will avoid the "unrecognized signature" error during deserialization, but this isn't the best fix for this particular issue.

A better fix is to change the mock monitor to indicate it does not support output values. That way, `Output` instances won't start showing up in the mock's `newResource`/`new_resource` inputs where they weren't before.

justinvp:
v3.13.0 introduces support for serializing outputs in inputs as special output value objects in the Node.js and Python SDKs when serializing inputs for remote components and method calls. This functionality is currently disabled by default in the engine (setting the `PULUMI_ENABLE_OUTPUT_VALUES` envvar to a truthy value enables it).

However, unit testing remote components with mocks results in errors being raised in v3.13.0, related to the new output value support. This is due to the mock monitor implementation saying it supports all features (which now includes output values), so the SDK serializers are serializing outputs as output values, which the mock monitor can't handle correctly.

This change addresses the issue by updating the mock monitor implementation in the Node.js and Python SDKs to indicate the specific features that are supported, excluding support for output values. New tests with mocks fail before the change and pass after.

Fixes #8050

### DIFF
--- a/sdk/nodejs/runtime/mocks.ts
+++ b/sdk/nodejs/runtime/mocks.ts
@@ -206,8 +206,14 @@ export class MockMonitor {
     }
 
     public supportsFeature(req: any, callback: (err: any, innerResponse: any) => void) {
+        const id = req.getId();
+
+        // Support for "outputValues" is deliberately disabled for the mock monitor so
+        // instances of `Output` don't show up in `MockResourceArgs` inputs.
+        const hasSupport = id === "secrets" || id === "resourceReferences";
+
         callback(null, {
-            getHassupport: () => true,
+            getHassupport: () => hasSupport,
         });
     }
 }
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -13,3 +13,6 @@
 
 - [codegen/go] - Fix resolution of enum naming collisions
   [#7985](https://github.com/pulumi/pulumi/pull/7985)
+
+- [sdk/{nodejs,python}] - Fix errors when testing remote components with mocks.
+  [#8053](https://github.com/pulumi/pulumi/pull/8053)
--- a/sdk/nodejs/tests_with_mocks/mocks.spec.ts
+++ b/sdk/nodejs/tests_with_mocks/mocks.spec.ts
@@ -42,6 +42,14 @@ pulumi.runtime.setMocks({
         case "pkg:index:MyCustom":
             assert.strictEqual(args.custom, true);
             return { id: args.name + "_id", state: args.inputs };
+        case "pkg:index:MyRemoteComponent":
+            return {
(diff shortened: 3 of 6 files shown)


sdk/nodejs/runtime/mocks.ts
```diff
@@ -206,8 +206,14 @@ export class MockMonitor {
     }
 
     public supportsFeature(req: any, callback: (err: any, innerResponse: any) => void) {
+        const id = req.getId();
+
+        // Support for "outputValues" is deliberately disabled for the mock monitor so
+        // instances of `Output` don't show up in `MockResourceArgs` inputs.
+        const hasSupport = id === "secrets" || id === "resourceReferences";
```
justinvp, line 213: Instead of blanket support for all features, I've changed it to be explicit about what the mock monitor supports. These are based off what the actual engine supports:

https://github.com/pulumi/pulumi/blob/b8b6f3dd70b2e35e15a4a5e69a9dadebf8be5932/pkg/resource/deploy/source_eval.go#L539-L546

t0yv0 (approved):
LGTM

PR #8055, opened 2021-09-24
Do report stdout and stderr in case of failure

t0yv0:
# Description

This change makes sure that we do get to see stdout/stderr of a failing command. I think the original code is a typo.

## Checklist

### DIFF
--- a/pkg/codegen/internal/test/helpers.go
+++ b/pkg/codegen/internal/test/helpers.go
@@ -335,7 +335,6 @@ func RunCommand(t *testing.T, name string, cwd string, exec string, args ...stri
 		append([]string{exec}, args...),
 		wd,
 		&cmdOptions)
-	require.NoError(t, err)
 	if err != nil {
 		stdout := stdout.String()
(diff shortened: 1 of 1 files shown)


pkg/codegen/internal/test/helpers.go
```diff
@@ -345,6 +344,7 @@ func RunCommand(t *testing.T, name string, cwd string, exec string, args ...stri
 		if len(stderr) > 0 {
 			t.Logf("stderr: %s", stderr)
 		}
+		require.NoError(t, err)
```
pgavlin, line 347: We can replace the `require.NoError(t, err)` and the existing `if` with a call to `assert.NoError` and a check on its return value:

```suggestion
		if !assert.NoError(t, err) {
```

`assert` does not fail a test promptly; `require` does.

pgavlin (changes requested):

t0yv0:
Applied, PTAL.

iwahbe (approved):

Issue #8058, opened 2021-09-24
Project configuration (`Pulumi.yaml`) option to refresh before every `preview`/`update`/`destroy`


infin8x:
Following up on a suggestion in https://github.com/pulumi/pulumi/issues/2247#issuecomment-925366911, let's introduce a new config option that will cause the engine to run a refresh before every `preview`/`update`/`destroy` operation. This should be logically equivalent to the user remembering to always add `--refresh` to those operations.

Suggested option name/type: `alwaysRefresh` (boolean)

stack72:
Hi @infin8x 

So I need a little clarification on the work here - should the `Pulumi.yaml` option be used over the cli args? I.e. the user has specifically set this value for the entire project

Paul

jre21:
Generally in this situation, command-line flags should take precedence over the config file.  That way a user can set a default behavior but still override it for a specific execution.  The most common approach I've seen in other projects is:

- `pulumi update --refresh`: perform a refresh before updating
- `pulumi update --no-refresh`: update without performing a refresh
- `pulumi update`: use the behavior defined in the configuration file (or the global default if the config doesn't specify a behavior)

Another common pattern to keep in mind for if you do decide to change the default behavior in 4.0: many projects will issue a depreciation notice for a few releases prior to changing a default.  This notice would typically be printed during program execution, state what behavior is changing, and direct users who wish to retain the current behavior to set it explicitly set it in their config file (usually with a note that doing so will silence the deprecation warning).  The notice should _not_ be displayed to users who already have explicit configuration.  Git took this approach, for example, prior to changing the default setting for `push.default` in their 2.0 release.

infin8x:
@stack72 good question. I agree with @jre21's suggestion - CLI flags take precedence over the config file.

Thanks @jre21 for the broader suggestion on deprecation workflow.

PR #8071, opened 2021-09-27
[cli] Add the ability to control auto-refresh of stacks by Pulumi.yaml

stack72:
Fixes: #8058

A user can now specify the following in their Pulumi.yaml

```
options:
  refresh: always
```

This will cause the derivative stacks to refresh by default when running an update, preview
or destroy command. We can override this with `--refresh=false`

the specific CLI commands still override this argument. Therefore:

* If a user sets refresh: always, they can override this behaviour with `--refresh==false`
* if a user sets `--refresh` or `--refresh=true` then it will use that by default
* if no cli args are passed but `refresh: always` then we will perform a refresh
* the default behaviour still exists of no refresh

https://asciinema.org/a/2BQoAe9gx9E07gBiM9h9A1cra

## Checklist

### DIFF
--- a/pkg/cmd/pulumi/destroy.go
+++ b/pkg/cmd/pulumi/destroy.go
@@ -209,7 +209,7 @@ func newDestroyCmd() *cobra.Command {
 		&parallel, "parallel", "p", defaultParallel,
 		"Allow P resource operations to run in parallel at once (1 for no parallelism). Defaults to unbounded.")
 	cmd.PersistentFlags().StringVarP(
-		&refresh, "refresh", "r", "no",
+		&refresh, "refresh", "r", "",
 		"Refresh the state of the stack's resources before this update")
 	cmd.PersistentFlags().Lookup("refresh").NoOptDefVal = "true"
 	cmd.PersistentFlags().BoolVar(
--- a/pkg/cmd/pulumi/destroy.go
+++ b/pkg/cmd/pulumi/destroy.go
@@ -41,7 +41,7 @@ func newDestroyCmd() *cobra.Command {
 	var diffDisplay bool
 	var eventLogPath string
 	var parallel int
-	var refresh bool
+	var refresh string
 	var showConfig bool
 	var showReplacementSteps bool
 	var showSames bool
@@ -143,10 +143,14 @@ func newDestroyCmd() *cobra.Command {
 				targetUrns = append(targetUrns, resource.URN(t))
 			}
 
+			refreshOption, err := getRefreshOption(proj, refresh)
+			if err != nil {
+				return result.FromError(err)
+			}
 			opts.Engine = engine.UpdateOptions{
 				Parallel:                  parallel,
 				Debug:                     debug,
-				Refresh:                   refresh,
+				Refresh:                   refreshOption,
 				DestroyTargets:            targetUrns,
 				TargetDependents:          targetDependents,
 				UseLegacyDiff:             useLegacyDiff(),
@@ -204,9 +208,10 @@ func newDestroyCmd() *cobra.Command {
 	cmd.PersistentFlags().IntVarP(
 		&parallel, "parallel", "p", defaultParallel,
 		"Allow P resource operations to run in parallel at once (1 for no parallelism). Defaults to unbounded.")
-	cmd.PersistentFlags().BoolVarP(
-		&refresh, "refresh", "r", false,
+	cmd.PersistentFlags().StringVarP(
+		&refresh, "refresh", "r", "no",
 		"Refresh the state of the stack's resources before this update")
+	cmd.PersistentFlags().Lookup("refresh").NoOptDefVal = "true"
 	cmd.PersistentFlags().BoolVar(
 		&showConfig, "show-config", false,
 		"Show configuration keys and variables")
--- a/pkg/cmd/pulumi/util.go
+++ b/pkg/cmd/pulumi/util.go
@@ -839,3 +839,26 @@ func checkDeploymentVersionError(err error, stackName string) error {
 	}
 	return errors.Wrap(err, "could not deserialize deployment")
 }
+
+func getRefreshOption(proj *workspace.Project, refresh string) (bool, error) {
+	// we want to check for an explicit --refresh or a --refresh=true or --refresh=false
+	// no is the default we assign to the refresh default to be able to understand
+	// the difference between when the user actually interacted with the cli argument
+	// and the default functionality today
+	if refresh != "" && refresh != "no" {
+		refreshDetails, boolErr := strconv.ParseBool(refresh)
+		if boolErr != nil {
+			// the user has passed a --refresh but with a random value that we don't support
+			return false, errors.New("unable to determine value for --refresh")
+		}
+		return refreshDetails, nil
+	}
+
+	// the user has not specifically passed an argument on the cli to refresh but has set a Project option to refresh
+	if proj.Options != nil && proj.Options.Refresh == "always" {
+		return true, nil
+	}
+
+	// the default functionality right now is to always skip a refresh
+	return false, nil
+}
--- a/pkg/cmd/pulumi/util.go
+++ b/pkg/cmd/pulumi/util.go
@@ -842,10 +842,10 @@ func checkDeploymentVersionError(err error, stackName string) error {
 
 func getRefreshOption(proj *workspace.Project, refresh string) (bool, error) {
 	// we want to check for an explicit --refresh or a --refresh=true or --refresh=false
-	// no is the default we assign to the refresh default to be able to understand
-	// the difference between when the user actually interacted with the cli argument
+	// refresh is assigned the empty string by default to distinguish the difference between
+	// when the user actually interacted with the cli argument (`NoOptDefVal`)
 	// and the default functionality today
-	if refresh != "" && refresh != "no" {
+	if refresh != "" {
 		refreshDetails, boolErr := strconv.ParseBool(refresh)
 		if boolErr != nil {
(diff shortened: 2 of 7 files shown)


pkg/cmd/pulumi/destroy.go
```diff
@@ -207,6 +208,10 @@ func newDestroyCmd() *cobra.Command {
 	cmd.PersistentFlags().BoolVarP(
 		&refresh, "refresh", "r", false,
 		"Refresh the state of the stack's resources before this update")
+	cmd.PersistentFlags().BoolVar(
+		&skipRefresh, "skip-refresh", false,
+		"Skip the refreshing of the state of the stack's resources before this update. Typically used "+
+			"when overriding a Pulumi project setting")
```
pgavlin, line 214: I wonder if we could piggyback off of the existing `refresh` flag here. Is there any way to differentiate between a flag that is unset and therefore has its default value and a flag that is set but has the same value as its default? That would allow us to use `--refresh=false` rather than `--skip-refresh`.

pkg/cmd/pulumi/util.go
```diff
@@ -839,3 +839,23 @@ func checkDeploymentVersionError(err error, stackName string) error {
...
+	if refresh {
+		return true
+	}
+
+	// the user has specified `--skip-refresh`
+	if skipRefresh {
+		return false
+	}
+
+	// the user has not specifically asked for a refresh or to skip-refresh
```
pgavlin, line 854:
```suggestion
	// the user has not specifically asked for a refresh or to skip-refresh, but has set the `autoRefresh` option in their Pulumi.yaml
```

lukehoban:
> A user can now set autoRefresh: true in Pulumi.yaml

Two thoughts on this:
1. Should we move these sorts of options under some container, instead of at top level?  It feels like over time we're going to accrue a lot of these - and that it would be better not to pollute the top level of the file. 
2. Is `autoRefresh` the right name for this?  It's not what we call the flag - we just use `refresh`.  Should this be `refresh` or `refreshOnUpdate`?

komalali:
@lukehoban 

> Is autoRefresh the right name for this? It's not what we call the flag - we just use refresh. Should this be refresh or refreshOnUpdate?

Is the refresh only on `Update` or is also on `Destroy`? TBH I think `autoRefresh` is clearer than either of the two options you've suggested.

pkg/cmd/pulumi/destroy.go
stack72: I am happy to change this of course but `--refresh=false` doesn't feel as nice to me as `--skip-refresh` :) But 🤷

komalali: Agree that `skip-refresh` is "clearer" but since we already have a bool flag for `refresh` and passing `refresh=false` is already a valid use of the flag let's stick to that. Otherwise we'll kind of have 2 ways of specifying "don't refresh"

pkg/cmd/pulumi/util.go
```diff
@@ -839,3 +839,26 @@ func checkDeploymentVersionError(err error, stackName string) error {
...
+func getRefreshOption(proj *workspace.Project, refresh string) bool {
+	// we want to check for an explicit --refresh or a --refresh=true or --refresh=false
+	// no is the default we assign to the refresh default to be able to understand
+	// the difference between when the user actually interacted with the cli argument
+	// and the default functionality today
+	if refresh != "" && refresh != "no" {
+		refreshDetails, boolErr := strconv.ParseBool(refresh)
+		if boolErr != nil {
+			// the user has passed a --refresh but with a random value that we don't support
+			return false
```
komalali, line 852: Maybe surface an error here since the user might be confused by this behavior

pkg/cmd/pulumi/util.go
```diff
@@ -839,3 +839,26 @@ func checkDeploymentVersionError(err error, stackName string) error {
 	}
 	return errors.Wrap(err, "could not deserialize deployment")
 }
+
+func getRefreshOption(proj *workspace.Project, refresh string) bool {
+	// we want to check for an explicit --refresh or a --refresh=true or --refresh=false
+	// no is the default we assign to the refresh default to be able to understand
+	// the difference between when the user actually interacted with the cli argument
+	// and the default functionality today
+	if refresh != "" && refresh != "no" {
```
pgavlin, line 848: I'm concerned that we're not surfacing an error when faced with unexpected input, and that we're using a sort of random value (`"no"`) rather than a standard bool as the default here.

How about defining the `refresh` flag using [`VarP`](https://pkg.go.dev/github.com/spf13/pflag#FlagSet.Var), and defining an implementation of [`Value`](https://pkg.go.dev/github.com/spf13/pflag#Value) with "optional boolean" semantics? e.g.

```go
type optionalBoolean struct {
	value *bool
}

var _ = pflag.Value((*optionalBoolean)(nil))

func (b *optionalBoolean) String() string {
	if b.value == nil {
		return "<unset>"
	}
	return strconv.FormatBool(*b.value)
}

func (b *optionalBoolean) Set(s string) error {
	v, err := strconv.ParseBool(s)
	if err != nil {
		return err
	}
	b.value = &v
}

func (b *optionalBoolean) Type() string {
	return "bool"
}

...

func newFooCmd() {
	var refresh optionalBoolean
	var cmd = &cobra.Command{
		...
	}
	cmd.PersistentFlags().VarP(
 		&refresh, "refresh", "r",
 		"Refresh the state of the stack's resources before this update")
}

File: sdk/go/common/workspace/project.go [hunk 5]
```diff
@@ -58,6 +58,11 @@ type ProjectBackend struct {
 	URL string `json:"url,omitempty" yaml:"url,omitempty"`
 }
 
+type ProjectOptions struct {
+	// Refresh is the ability to always run a refresh as part of a pulumi update / preview / destroy
+	Refresh *string `json:"refresh,omitempty" yaml:"refresh,omitempty"`
```
komalali, line 63:
```suggestion
	Refresh string `json:"refresh,omitempty" yaml:"refresh,omitempty"`
```

pkg/cmd/pulumi/util.go
```diff
@@ -839,3 +839,26 @@ func checkDeploymentVersionError(err error, stackName string) error {
...
+		refreshDetails, boolErr := strconv.ParseBool(refresh)
+		if boolErr != nil {
+			// the user has passed a --refresh but with a random value that we don't support
+			return false
+		}
+		return refreshDetails
+	}
+
+	// the user has not specifically passed an argument on the cli to refresh but has set a Project option to refresh
+	if proj.Options != nil && proj.Options.Refresh != nil && *proj.Options.Refresh == "always" {
```
komalali, line 858: If we make `Refresh` a `string` instead of `*string` we can simplify this:

```suggestion
	if proj.Options != nil && proj.Options.Refresh == "always" {
```

komalali (commented):
Mostly LGTM but let's just keep `Refresh` as a string instead of a pointer to avoid oddities since we don't *need* `nil` here.
Also returning an error when not being able to parse a bool would be more obvious to the user.

pkg/cmd/pulumi/util.go
komalali: @stack72 and I worked together to come up with a solution that essentially equates to the same thing. If you'd rather we can use the value `"<unset>"` instead of `no`. And yeah I agree it should return an error if we can't parse the value. But otherwise, these are just different ways of doing the same thing.

komalali: And I kinda think the way it currently is is easier to read :)

pgavlin: > And I kinda think the way it currently is is easier to read :)

I'll admit that I don't love the way the `flag`/`pflag` packages deal with user-defined parsing--I'd rather be able to just supply a function. The downside I see to the current approach is that it puts the error reporting and handling in an awkward place--at the point the value is calculated, rather than at the point the flag is parsed.

> If you'd rather we can use the value "<unset>" instead of "no"

If we stick with a plain string, is there any reason not to use `""` for the default value? I don't see a semantic difference between `""` and `"no"`, and `"no"` has the downside of being something of a one-off. FWIW, we could use "" in the `Value` approach as well.

sdk/go/common/workspace/project.go
stack72: 👍

pkg/cmd/pulumi/util.go
stack72: 👍

pkg/cmd/pulumi/util.go
komalali: I didn't say `""` I said `"<unset>"` as in your example code

pkg/cmd/pulumi/util.go
stack72: done!

pkg/cmd/pulumi/util.go
komalali: I hear you and I respectfully think this is fine 🙏🏽

pgavlin: > I didn't say "" I said "<unset>" as in your example code

Yeah, I know--I was making another suggestion--sorry for the lack of clarity!

komalali (approved):

pkg/cmd/pulumi/util.go
komalali: Yeah I think we arrived at `"no"` from somewhat of a meandering path and looking over it again it's true that `""` works just as well and is clearer and avoids the double `if refresh != "" && refresh != "no"`. 

Pretty sure we started with the empty string and I somehow convinced @stack72 that it wouldn't work. But it does!

PR #8085, opened 2021-09-28
Update tests/go.mod

justinvp:


komalali:
I think this might also need a `go mod download` in `tests`? Or a `make ensure` at root for good measure?

justinvp:
> I think this might also need a `go mod download` in `tests` (`go.sum` is not updated)? Or a `make ensure` at root for good measure?

This is the change after running `make ensure` in root.

Issue #8092, opened 2021-09-28
Fail release action early if the git branch is dirty


komalali:
Currently, if a dirty git branch is tagged and released it will proceed halfway through release and fail on releasing the python SDK (but CLI + nodejs SDK will be released). We should add a step to the process that fails immediately if the git branch is dirty.

Issue #8061, opened 2021-09-27
LocalWorkspace supports stack outputs flaky


t0yv0:
```
     Error: Timeout of 120000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
      at listOnTimeout (internal/timers.js:557:17)
      at processTimers (internal/timers.js:500:7)
```

https://github.com/pulumi/pulumi/pull/8051

## Steps to reproduce

1.
2.
3.
4.

Expected: 
Actual:

komalali:
Not the same exact failure, but similar in that it's automation api nodejs test failing with a timeout: https://github.com/pulumi/pulumi/runs/3738395273

This is super weird! Has there been a regression in this area? These tests have worked reliably with no flakes for the past year-ish.

PR #8093, opened 2021-09-28
Check worktree is clean before releasing

komalali:
# Description

* Adds `lint` step as required before release.
* Adds a new job `check-clean-worktree` to ensure the git work tree is clean before releasing.

I elected to add a separate job for this step to avoid doing it 8 times as part of build-and-test.

I unfortunately don't know how to test this without a release. However, this is already a part of the `master` workflow: https://github.com/pulumi/pulumi-aws/blob/master/.github/workflows/master.yml#L68-L69

Fixes https://github.com/pulumi/pulumi/issues/8092

## Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -186,7 +186,7 @@ jobs:
   publish-binaries:
     name: Publish Binaries
     runs-on: macos-latest
-    needs: [check-clean-worktree, lint, build-and-test, windows-build]
+    needs: [lint, build-and-test, windows-build]
     strategy:
       matrix:
         go-version: [ 1.16.x ]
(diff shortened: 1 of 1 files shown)


pgavlin (changes requested):
> I elected to add a separate job for this step to avoid doing it 8 times as part of build-and-test.

I don't think that this is going to give the results that we want: I think that this is only going to check that the worktree is clean immediately after cloning the repository, which will always be true. I think that we're going to need to do this as part of each leg after building, as that is where we can end up leaving the worktree dirty.

pgavlin:
(I think that checking for a clean worktree is pretty quick, so I don't think we'll notice the overhead of rerunning it)

komalali:
>I think that we're going to need to do this as part of each leg after building, as that is where we can end up leaving the worktree dirty.

Thanks you for clarifying! That makes way more sense 😅

pgavlin (approved):

komalali:
Unrelated test failure captured in https://github.com/pulumi/pulumi/issues/8061#issuecomment-929700982

Issue #7976, opened 2021-09-16
Flaky TestArchiveTarFiles test


justinvp:
```
=== FAIL: go/common/resource TestArchiveTarFiles (13.62s)
    asset_test.go:375: 
        	Error Trace:	asset_test.go:375
        	Error:      	Expected nil, but got: incorrect blob size for test-results/39d66487-faf8-49f0-a9b7-a6c651c0edff.json: expected 1208454, got 1208454: archive/tar: write too long
        	Test:       	TestArchiveTarFiles
```

praneetloke:
I've disabled this test with #8102.

PR #8102, opened 2021-09-29
[codegen/docs] Generate a package tree that can be serialized as JSON

praneetloke:
# Description

This PR adds a new exported function in the docs generator that can be used by the `resourcedocsgen` tool to retrieve a package tree that can be emitted somewhere for our docs site. The JSON file is used to render some UI components.

Along with this, I have also updated the docs template bundler to be invoked programmatically. It's useful for the SDK generator tests we run from `docs/gen_test.go` so that it doesn't end up using a stale `packaged.go` which is git ignored.

Fixes https://github.com/pulumi/pulumi-service/issues/7346

## Checklist

  - **Note**: I didn't do this because this has no end user impact. Let me know if you still would like me to include this in the pending list


pkg/codegen/docs/bundler/bundler.go
```diff
@@ -0,0 +1,113 @@
+package bundler
```
praneetloke, line 1: All of this was moved from `docs/bundler.go`.

pkg/codegen/docs/package_tree.go
```diff
@@ -0,0 +1,75 @@
...
+// navigable tree format starting from the top-level/index/root
+// of a package.
+type PackageTreeItem struct {
+	Name     string            `json:"name"`
+	Type     entryType         `json:"type"`
+	Link     string            `json:"link"`
+	Children []PackageTreeItem `json:"children,omitempty"`
+}
+
+func generatePackageTree(rootMod modContext) ([]PackageTreeItem, error) {
```
praneetloke, line 25: I could also make this a receiver function of the `modContext` type if you think that reads better. I started going down this route and didn't think it was bad enough to have to change to that.

pkg/codegen/docs/gen.go
```diff
@@ -1831,9 +1827,7 @@ func generateModulesFromSchemaPackage(tool string, pkg *schema.Package) map[stri
 	return modules
 }
 
-// GeneratePackage generates the docs package with docs for each resource given the Pulumi
-// schema.
-func GeneratePackage(tool string, pkg *schema.Package) (map[string][]byte, error) {
+func Initialize(tool string, pkg *schema.Package) {
```
praneetloke, line 1830: Adding this initialize allows to use the generated modules to do other things with it as separate operations rather a single function. It seemed like a good idea to decompose them into separate things.

justinvp (approved):

praneetloke:
The downstream resource docs workflow failure is expected because of some changes. The `resourcedocsgen` tool in `docs` repo uses a local path override for `pulumi/pulumi`. So there will be a dependency problem until the associated PR is https://github.com/pulumi/docs/pull/6530 merged, which is also the PR that actually uses these changes. So yeah these are inter-dependent right now in their PR workflows.

As for https://github.com/pulumi/pulumi/issues/7976, it seems to be a flaky test. See https://github.com/pulumi/pulumi/issues/7976.

praneetloke:
Alright the only failing PR workflow is the resource docs test. As mentioned before, that's ok to ignore for this PR. Note that all SDK gen unit tests that include docs generation passed. I'll merge this.

PR #8103, opened 2021-09-29
[codegen] Rename the PCL package.

pgavlin:
It's just confusing that PCL lives in a package named `hcl2`.


pgavlin:
Note that this rename is technically a breaking change, but it's a breaking change to an internal package. We'll need to update `pulumi-terraform-bridge` once this is in, as it consume the PCL package as part of `tf2pulumi`.

pgavlin:
The codegen tests are failing because I `gofmt`'d the baselines 🤦

komalali (approved):
The codegen tests are failing now because of the break in the pulumi-terraform-bridge. We should make that change a fast-follow.

pgavlin:
https://github.com/pulumi/pulumi-terraform-bridge/pull/395

PR #8108, opened 2021-09-30
[testing] Stack context for value generation

pgavlin:
These changes add a context type to `resource/testing` that can be used to
generate property values that are valid with respect to certain rules. The
context represents a Pulumi stack, and contains a project name, stack name,
and a list of resources.

- URNs generated using a context will always use the context's project
  and stack name
- Resource references generated using a context will always refer to
  resources in the context's resource list
- Output values generated using a context will always pull dependencies
  from the context's resource list

### DIFF
--- a/sdk/go/common/resource/testing/rapid.go
+++ b/sdk/go/common/resource/testing/rapid.go
@@ -263,6 +263,7 @@ func resourceReferenceGenerator(ctx *StackContext) *rapid.Generator {
 	return rapid.Custom(func(t *rapid.T) resource.ResourceReference {
 		r := resourceGenerator.Draw(t, "referenced resource").(*resource.State)
 
+		// Only pull the resource's ID if it is a custom resource. Component resources do not have IDs.
 		var id resource.PropertyValue
 		if r.Custom {
 			id = rapid.OneOf(UnknownPropertyGenerator(), rapid.Just(resource.NewStringProperty(string(r.ID)))).Draw(t, "referenced ID").(resource.PropertyValue)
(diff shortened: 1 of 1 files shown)


sdk/go/common/resource/testing/rapid.go
```diff
@@ -9,6 +9,105 @@ import (
...
+	return ctx.stackName
+}
+
+// Resources returns the context's resources.
+func (ctx *StackContext) Resources() []*resource.State {
+	return ctx.resources
+}
+
+// Append creates a new context that contains the current context's resources and the given list of resources.
+func (ctx *StackContext) Append(r ...*resource.State) *StackContext {
```
t0yv0, line 41: No mutation, thanks!

pgavlin: Figured you'd like that one 😝

sdk/go/common/resource/testing/rapid.go
```diff
@@ -125,32 +237,65 @@ func ArchivePropertyGenerator(maxDepth int) *rapid.Generator {
...
+	}
+
 	return rapid.Custom(func(t *rapid.T) resource.ResourceReference {
-		id := rapid.OneOf(UnknownPropertyGenerator(), StringPropertyGenerator()).Draw(t, "referenced ID").(resource.PropertyValue)
-		if id.IsString() && id.StringValue() == "" {
-			id = resource.PropertyValue{}
+		r := resourceGenerator.Draw(t, "referenced resource").(*resource.State)
+
+		var id resource.PropertyValue
+		if r.Custom {
```
t0yv0, line 267: If r.Custom == false, we get id set to default empty value? Why is that? Feels a bit odd, perhaps worth a quick comment here.

t0yv0 (approved):
Nice.

sdk/go/common/resource/testing/rapid.go
pgavlin: This is because component resources do not have IDs, and we use the absence of a value in the `ID` field to indicate that.

PR #8109, opened 2021-10-01
Fix Go automation API --target / --replace args

jdartigalongue:
# Description

This MR fixes an issue in the Pulumi Go automation API where --target / --replace args are not correctly formated, causing the CLI to not recognize the flag :

Error: unknown flag: --target urn:pulumi:stack::project::pulumi:providers:kubernetes::k8s

Fixes # (issue)

## Checklist


komalali (approved):
Thanks for the fix @jdartigalongue !

PR #8111, opened 2021-10-01
Support "lifting" single-valued method returns to their return type

justinvp:
Replaces #7937. I've already reviewed James' commits. This PR does the following:

- Addresses the merge conflict and bring this up-to-date with master
- Fixes an issue preventing an existing test from passing (40b33ed34c3e198c286302f5c738632ea503504d)
- Addresses some feedback (c8d6bf1d8ae2355717fb2f27b631db132c1d5aad)
- Adds .NET codegen (85c9afa77827cd8f961751b63a9487bb0f98b116)
- Adds docgen (96397ed2524b833e11c8e377da8b66987b4f2bf7)
- Adds codegen tests (9ed99c15c6ca96016ad12d4d404cf540167de978)

As I mentioned in #7937, while this doesn't implement the design as discussed in #7435 (changing `FunctionSpec.Outputs` to accept `*TypeSpec`, and all the fall-out from that), doing this via language flag is a good practical change to opt-in to this behavior in the meantime, and doesn't prevent us from doing the full fix for #7435 in the future.


pgavlin (approved):
> doing this via language flag is a good practical change to opt-in to this behavior in the meantime, and doesn't prevent us from doing the full fix for #7435 in the future.

It feels a little weird to duplicate this flag across all the languages, but I suppose that we already do this for e.g. the `Compatibility` flag.

pkg/codegen/internal/test/testdata/simple-methods-schema-single-value-returns/go/example/foo.go
```diff
@@ -0,0 +1,116 @@
...
+func (FooArgs) ElementType() reflect.Type {
+	return reflect.TypeOf((*fooArgs)(nil)).Elem()
+}
+
+func (r *Foo) GetKubeconfig(ctx *pulumi.Context, args *FooGetKubeconfigArgs) (pulumi.StringOutput, error) {
+	out, err := ctx.Call("example::Foo/getKubeconfig", args, fooGetKubeconfigResultOutput{}, r)
+	if err != nil {
+		return pulumi.StringOutput{}, err
+	}
+	return out.(fooGetKubeconfigResultOutput).Kubeconfig(), nil
```
t0yv0, line 48: Would have been possible to omit fooGetKubeconfigResultOutput type entirely? Save space..

t0yv0:
I'd be conservative merging these PRs since for unrelated reasons the downstream codegen checks are not informative right now, but I understand there is some urgency here.

pkg/codegen/internal/test/testdata/simple-methods-schema-single-value-returns/go/example/foo.go
justinvp: Not based on the current implementation of the SDK's `Context.Call`, which needs the output:

https://github.com/pulumi/pulumi/blob/230f4ee006995b47751c9ee4f44643ea3427fb82/sdk/go/pulumi/context.go#L331

The SDK itself doesn't know how to lift the single value, so it's handled here in the codegen.

I don't think we have to be too concerned about the size just yet as there are very few uses of methods out there, and even with broader adoption, I doubt it'll be used so much as to cause size issues (famous last words).

t0yv0: Ah that's right, these are methods, not functions (of which there's a few).

PR #8116, opened 2021-10-01
Fix 8110

### LINKED ISSUE #8110: Node codegen emits incorrectly nested enum-array type literals
## Hello!

## Issue details

In certain circumstances of nested arrays/unions nodejs codegen emits the wrong type. 

### Steps to reproduce

Consider:

```
    "deploymentTypes": {
      "type": "array",
      "items": {
       "oneOf": [
        {
         "type": "string"
        },
        {
         "$ref": "#/types/azure-native:network:ConfigurationType"
        }
       ]
     },
```

The intent is `Array<Union<string,ConfigurationType>>`.

But the generated code is:

https://github.com/pulumi/pulumi-azure-native/blob/master/sdk/nodejs/network/listNetworkManagerDeploymentStatus.ts#L33

```typescript
string | enums.network.v20210201preview.ConfigurationType[]
```

Which gets interpreted as `Union<string,Array<ConfigurationType>>`.

Expected: `(string | enums.network.v20210201preview.ConfigurationType)[]`
Actual: `string | enums.network.v20210201preview.ConfigurationType[]`

t0yv0:
# Description

Fix Node codegen to not create incorrectly parsed type references for unions and optionals.

Fixes #8110

Based on discussion, some effort was put into minimally inserting parens, rather than inserting them everywhere. 

## Checklist


pgavlin (commented):
This feels like an annoying question, but how hard would it be to only generate the parens when they're necessary? It'd minimize the diffs and would keep the types a little less noisy.

t0yv0:
Not annoying at all I was asking the same thing.

It would also make diff checking easier.

It's some work mostly to find TypeScript grammar precedence rules. If someone remembers how those work that could be a great help.

t0yv0:
Not seeing BNF in https://www.typescriptlang.org/docs/handbook/2/everyday-types.html

But at first glance, `[]` has higher precedence than `|` and we don't have other "operators" at play so I can just try to handle those.

pkg/codegen/internal/tstypes/tstypes_test.go
```diff
@@ -0,0 +1,181 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/davecgh/go-spew/spew"
+
+	"github.com/stretchr/testify/assert"
+	"pgregory.net/rapid"
+)
+
+func TestParenInsert(t *testing.T) {
```
t0yv0, line 27: Admittedly verbose but foolproof. Works on a simplified grammar but can be easily extended if we run into more grammar peculiarities. This checks that unparsing and auto-parenthesizing the AST into tokens gets parsed back into the same ast.

pkg/codegen/internal/tstypes/tstypes_test.go
```diff
@@ -0,0 +1,181 @@
...
+		if err != nil {
+			t.Error(err)
+			return
+		}
+		t.Logf("parsed: %s", spew.Sdump(parsed))
+		assert.Equal(t, example, parsed)
+	})
+}
+
+func astGenerator() *rapid.Generator {
```
t0yv0, line 48: Lib remark: shrinking trees seems not entirely satisfactory, I'd like to compare to what Haskell has and if necessary send a patch upstream. Once a bad tree T has been found, I'd like it to be minimized heavily by pruning branches that do not matter for the test before presenting to the user.

pkg/codegen/internal/test/testdata/regress-node-8110/go/my8110/pulumiTypes.go
```diff
@@ -0,0 +1,64 @@
+// *** WARNING: this file was generated by test. ***
+// *** Do not edit by hand unless you're certain you know what you are doing! ***
```
t0yv0, line 2: Non-node code noisy, relatively low value, we should probably support skipping codegen by lang.

pkg/codegen/internal/tstypes/tstypes.go
```diff
@@ -0,0 +1,222 @@
...
+)
+
+type typeToken struct {
+	kind  typeTokenKind
+	value string
+}
+
+type typeScriptTypeUnparser struct{}
+
+func (u *typeScriptTypeUnparser) unparse(ast TypeAst) []typeToken {
```
pgavlin, line 173: are these methods just for scoping? if so, these can take a value receiver:

```suggestion
func (u typeScriptTypeUnparser) unparse(ast TypeAst) []typeToken {
```

pgavlin (approved):

PR #8117, opened 2021-10-01
[testing] Support *resource.State generation

pgavlin:
These changes add support for generating *resource.State values to the
rapid testing framework. If states are generated within a *StackContext,
they are guaranteed to only refer to other resources that exist within
that context.

### DIFF
--- a/pkg/resource/stack/deployment_test.go
+++ b/pkg/resource/stack/deployment_test.go
@@ -25,6 +25,8 @@ import (
 	"github.com/stretchr/testify/require"
 	"pgregory.net/rapid"
 
+	"github.com/pulumi/pulumi/pkg/v3/resource/deploy"
+	"github.com/pulumi/pulumi/pkg/v3/secrets/b64"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/apitype"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource/config"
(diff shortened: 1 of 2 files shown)


pgavlin:
I think that I've still got one or two of the generators involved in producing valid URNs subtly wrong. No rush on this PR.

pkg/resource/stack/deployment_test.go
```diff
@@ -456,6 +458,38 @@ func TestDeserializePropertyValue(t *testing.T) {
...
+			isProvider := rapid.Bool().Draw(t, "is provider").(bool)
+			context = context.Append(context.ResourceStateGenerator(isProvider, 6).Draw(t, fmt.Sprintf("resource %v", i)).(*resource.State))
+		}
+
+		// finally, serialize the resources as a deployment
+		snap := deploy.Snapshot{
+			SecretsManager: b64.NewBase64SecretsManager(),
+			Resources:      context.Resources(),
+		}
+		_, err := SerializeDeployment(&snap, nil, false)
```
t0yv0, line 488: 👍 

There's presumably DeserializeDeployment somewhere also?

sdk/go/common/resource/testing/rapid.go
```diff
@@ -412,3 +460,235 @@ func propertyValueGenerator(ctx *StackContext, maxDepth int) *rapid.Generator {
...
+			ID:                      id,
+			Inputs:                  inputs,
+			Outputs:                 outputs,
+			Parent:                  parent,
+			Protect:                 protect,
+			External:                external,
+			Dependencies:            sortedDeps(dependencySet),
+			InitErrors:              initErrors,
+			Provider:                provider,
+			PropertyDependencies:    propertyDependencies,
```
t0yv0, line 687: Curious the code above makes it look like PropertyDependencies is a computed field fully determined by Inputs. We chose to include it in the model for redundancy?

t0yv0 (approved):

sdk/go/common/resource/testing/rapid.go
pgavlin: > Curious the code above makes it look like PropertyDependencies is a computed field fully determined by Inputs. We chose to include it in the model for redundancy?

Until @justinvp's recent work, we did not have a first-class representation of `Output` values in the `resource.PropertyValue` system, so all dependency information was lifted to the top level during marshaling. We can probably have the SDKs stop producing this field if the resource monitor they're connected to supports `Output` values. We could probably also move the responsibility for consuming these values into the resource monitor from the engine (e.g. the resource monitor could wrap any top-level property in `Inputs` that has a non-empty set of dependencies in `PropertyDependencies` with an `Output` value that has the same set of dependencies).

pkg/resource/stack/deployment_test.go
pgavlin: Yep.

sdk/go/common/resource/testing/rapid.go
t0yv0: Ah that totally makes sense. I wouldn't change this just yet until the new changes "settle in" and become the new normal so to speak.

lunaris:
Closing this since we appear to have done OK in its absence -- that said we can always revisit if we want to adopt more property-style testing in core/for these use cases.

PR #8016, opened 2021-09-21
Cancel checks on outdated PR commits

t0yv0:
# Description

The motivation is included in the comment.

Fixes # (issue)

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works

- [ ] I have updated the [CHANGELOG-PENDING](https://github.com/pulumi/pulumi/blob/master/CHANGELOG_PENDING.md) file with my change

- [ ] Yes, there are changes in this PR that warrants bumping the Pulumi Service API version

PR #8118, opened 2021-10-01
Revert "Cancel checks on outdated PR commits (#8016)"

emiliza:
This reverts commit a9c9853e85fb628979569701e8714f1694a6a719.

# Description
Per [docs](https://docs.github.com/en/actions/learn-github-actions/contexts) `github.head_ref` is only available via `pull_request` or `pull_request_target`.   

This breaks as `/run-acceptance-tests` is triggered via `repository_dispatch`, eg:
https://github.com/pulumi/pulumi/actions/runs/1295914761/workflow

To get unblocked, let's revert and figure out an alternative way to do concurrency later.

cc: @t0yv0


stack72 (approved):
...

Issue #6954, opened 2021-05-02
Provide naming guidance/recommendations for SDK packages


infin8x:
Since Pulumi Packages will sometimes be used directly from language package registries (npmjs, NuGet Gallery, PyPI), we should help package authors land in the pit of success when naming their Pulumi Package (and the resultant SDK packages created by the `pulumi package publish` command (#6928)).

## Goals of package naming

- The Pulumi namespace (`@pulumi` on npm) or pulumi-bot account (NuGet, PyPI) should own “common” names where possible, so that it’s easy for users to fall in the pit of success if they install packages directly from the CLI without searching the registry. e.g. `@pulumi/eks`, Pulumi.Eks, or pulumi-eks should all point to the official 1st-party Pulumi Package for EKS, if possible. 
- We should try to differentiate Pulumi-owned packages from community-contributed packages, so it's easier for users of both types of packages to know where to go for support, where to look for docs, etc.

## Proposal

- All 1st-party Pulumi Packages are owned by a Pulumi-owned account (we already have this today - `pulumi-bot`)
- All 1st-party Pulumi Packages start with `@pulumi/` (on npm), `Pulumi.` (on NuGet), and `pulumi-` (on PyPI)
  - For npm, we guarantee this because we own the `@pulumi/` namespace
  - For NuGet.org, we can guarantee this using [namespace reservation](https://docs.microsoft.com/en-us/nuget/nuget-org/id-prefix-reservation), if we don't already
  - PyPI doesn't seem to have a solution, per [this thread](https://discuss.python.org/t/namespace-support-in-pypi/1609/42)
  - For Go, this is solved automatically as long as all official Pulumi packages come from the `pulumi` GitHub organization
- For 3rd-party Pulumi Packages, we should recommend (and the tooling should generate) names of this format:
  - npm: `@your-namespace/pulumi-[package-name]`
  - NuGet: `[your-namespace].Pulumi.[package-name]`
  - PyPI: `[your-namespace]-pulumi-[package-name]`
  - Go modules: n/a, this is determined by the URL from which the module is served; we recommend repo names of `[username/organization]/pulumi-[package-name]`
- If a package name has multiple words, separate those words with hyphens e.g. "pulumi-aws-quickstart-redshift"

## Other notes:
- Reference internal slack thread: https://pulumi.slack.com/archives/C01AP60KH5F/p1633149434021800
- Proposal for plugin naming

infin8x:
### Notes from the review 5/13

[...]

PR #8127, opened 2021-10-04
Add support for pulumiplugin.json in .NET and Go

### LINKED ISSUE #7244: .NET: Support custom server URLs for plugins
.NET doesn't currently have a way to indicate that a package has a plugin from a custom server URL.

- [langhost] Allow non `Pulumi.` prefixed packages when searching for packages that need plugins in `GetRequiredPlugins`.

### LINKED ISSUE #7245: Go: Support custom server URLs for plugins
Go doesn't currently have a way to indicate that a module has a plugin from a custom server URL.

- [langhost] Allow non pulumi prefixed modules when searching for plugins in `GetRequiredPlugins`.

iwahbe:
I add support for generating pulumiplugin.json in both .NET and Go.
Python already has this, and nodejs has it embedded in package.json.

I also standardized our `fs` wrapper on `map[string][]byte`. Otherwise
there would be 5 identical copies in our codebase.

# Description

Fixes #7245
Fixes #7244

## Checklist

- [X] Update codegen to generate `pulumipackage.json`.
- [ ] Update langhost to use new information
  - [ ] Go: Allow non pulumi prefixed modules when searching for plugins in `GetRequiredPlugins`.
  - [ ] .NET: Allow non Pulumi. prefixed packages when searching for packages that need plugins in `GetRequiredPlugins`.

### DIFF
--- a/pkg/codegen/docs/gen.go
+++ b/pkg/codegen/docs/gen.go
@@ -38,7 +38,6 @@ import (
 	"github.com/pulumi/pulumi/pkg/v3/codegen/nodejs"
 	"github.com/pulumi/pulumi/pkg/v3/codegen/python"
 	"github.com/pulumi/pulumi/pkg/v3/codegen/schema"
-	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 )
 
 var (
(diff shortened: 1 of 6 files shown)


iwahbe:
https://github.com/pulumi/pulumi/issues/6954 specifies what package names we allow.

PR #8130, opened 2021-10-04
[sdk/python] - Python 3.10 compatibility

### LINKED ISSUE #8129: Running `pulumi up` in new project with python 3.10 prints deprecation warning
## Hello!

### Steps to reproduce

1. Install [python 3.10](https://www.python.org/downloads/)
2. Run `pulumi new aws-python`
3. Run `pulumi up`
4. Observe output

Expected: No deprecation warning
Actual:

```
Updating (dev)

View Live: https://app.pulumi.com/komalali/blah/dev/updates/1

     Type                 Name       Status      Info
 +   pulumi:pulumi:Stack  blah-dev   created     2 messages
 +   └─ aws:s3:Bucket     my-bucket  created

Diagnostics:
  pulumi:pulumi:Stack (blah-dev):
    /opt/pulumi/bin/pulumi-language-python-exec:71: DeprecationWarning: There is no current event loop
      loop = asyncio.get_event_loop()

Outputs:
    bucket_name: "my-bucket-bc117c7"

Resources:
    + 2 created

Duration: 9s
```

This is due to the fact that [asyncio.get_event_loop()](https://github.com/pulumi/pulumi/blob/480173a57f7360c4f261c1b28efbb02cd4381bb5/sdk/python/cmd/pulumi-language-python-exec#L71) now prints a deprecation warning when there is no running loop. See [the docs](https://docs.python.org/3.10/library/asyncio-eventloop.html#:~:text=Deprecated%20since%20version%203.10%3A%20Deprecation%20warning%20is%20emitted%20if%20there%20is%20no%20running%20event%20loop.%20In%20future%20Python%20releases%2C%20this%20function%20will%20be%20an%20alias%20of%20get_running_loop().) for more info.

komalali:
# Description

Fixes #8129 

## Checklist

### DIFF
--- a/sdk/python/cmd/pulumi-language-python-exec
+++ b/sdk/python/cmd/pulumi-language-python-exec
@@ -69,17 +69,8 @@ if __name__ == "__main__":
 
     successful = False
 
-    # asyncio.get_running_loop was only added in python 3.7 but we still support python 3.6
-    # In python 3.10, asyncio.get_event_loop prints a deprecation warning if no loop is present
-    # This code will be cleaned up as part of https://github.com/pulumi/pulumi/issues/8131
-    if sys.version_info[0] == 3 and sys.version_info[1] < 7:
-        loop = asyncio.get_event_loop()
-    else:
-        try:
-            loop = asyncio.get_running_loop()
-        except RuntimeError:
-            loop = asyncio.new_event_loop()
-            asyncio.set_event_loop(loop)
+    loop = asyncio.new_event_loop()
+    asyncio.set_event_loop(loop)
 
     # We are (unfortunately) suppressing the log output of asyncio to avoid showing to users some of the bad things we
     # do in our programming model.
--- a/sdk/python/cmd/pulumi-language-python-exec
+++ b/sdk/python/cmd/pulumi-language-python-exec
@@ -76,6 +76,10 @@ if __name__ == "__main__":
         loop = asyncio.get_event_loop()
     else:
         try:
+            # The docs for get_running_loop are somewhat misleading because they state:
+            # This function can only be called from a coroutine or a callback. However, if the function is
+            # called from outside a coroutine or callback (the standard case when running `pulumi up`), the function
+            # raises a RuntimeError as expected and falls through to the exception clause below.
             loop = asyncio.get_running_loop()
         except RuntimeError:
             loop = asyncio.new_event_loop()
--- a/sdk/python/cmd/pulumi-language-python-exec
+++ b/sdk/python/cmd/pulumi-language-python-exec
@@ -69,8 +69,17 @@ if __name__ == "__main__":
 
     successful = False
 
-    loop = asyncio.new_event_loop()
-    asyncio.set_event_loop(loop)
+    # asyncio.get_running_loop was only added in python 3.7 but we still support python 3.6
+    # In python 3.10, asyncio.get_event_loop prints a deprecation warning if no loop is present
+    # This code will be cleaned up as part of https://github.com/pulumi/pulumi/issues/8131
+    if sys.version_info[0] == 3 and sys.version_info[1] < 7:
+        loop = asyncio.get_event_loop()
+    else:
+        try:
+            loop = asyncio.get_running_loop()
+        except RuntimeError:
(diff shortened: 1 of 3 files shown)


komalali:
Actually, I need to rethink this slightly. `asyncio.get_running_loop()` was only added in python 3.7 so this will break 3.6 compatibility which we are not *immediately* ready for.

komalali:
Okay, fixed for now and opened https://github.com/pulumi/pulumi/issues/8131 and added a comment to clean up once we get around to it.

t0yv0 (approved):
Looks pretty good! 

Would you like to try running a test matrix with 3.10 (and/or 3.6)? 

```
  build_and_test:
    name: Build & Test
    strategy:
      matrix:
        platform: [ ubuntu-latest, macos-latest ]
        go-version: [ 1.16.x ]
        python-version: [ 3.9.x ]
        dotnet-version: [ 3.1.x ]
        node-version: [ 14.x ]

```

Can try it as a one-off at least.

komalali:
I'll try it. Historically we've had issues with the actual testing harness when trying to run on python 3.6 (rather than the tests themselves). But it's useful signal nonetheless.

sdk/python/cmd/pulumi-language-python-exec
```diff
@@ -68,7 +68,18 @@ if __name__ == "__main__":
...
-    loop = asyncio.get_event_loop()
+
+    # asyncio.get_running_loop was only added in python 3.7 but we still support python 3.6
+    # In python 3.10, asyncio.get_event_loop prints a deprecation warning if no loop is present
+    # This code will be cleaned up as part of https://github.com/pulumi/pulumi/issues/8131
+    if sys.version_info[0] == 3 and sys.version_info[1] < 7:
+        loop = asyncio.get_event_loop()
+    else:
+        try:
+            loop = asyncio.get_running_loop()
```
pgavlin, line 79: The [docs for `get_running_loop`](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_running_loop) state that this function can only be called from a coroutine or a callback. Are we in one of those contexts here?

komalali: I don't think we ever would be but I wasn't trying to make any assumptions and merely mirror the functionality of `asyncio.get_event_loop()` in a manner that cooperates with python 3.10. I would think just these two lines would be sufficient:

komalali:
Hah github runners don't have python 3.10 yet.

sdk/python/cmd/pulumi-language-python-exec
komalali: I'm gonna go ahead and just do what I said in the comment above.

sdk/python/cmd/pulumi-language-python-exec
```diff
@@ -68,7 +68,9 @@ if __name__ == "__main__":
         os.chdir(args.pwd)
 
     successful = False
-    loop = asyncio.get_event_loop()
+
+    loop = asyncio.new_event_loop()
+    asyncio.set_event_loop(loop)
```
pgavlin, line 73: This isn't quite what the original code was doing either--the call to `get_event_loop` would either get the event loop if one existed or create a new event loop if there wasn't one. I'm not sure that we _need_ that functionality--it may be that we're never on an event loop when we start and what you have here is sufficient--but I think that we should make sure we understand and document the circumstances under which we're calling these APIs.

pgavlin: (I think that we used `get_event_loop` b/c we weren't totally confident of the conditions under which this code may be called)

komalali: I understand what get_event_loop does. Thanks.

komalali: >If there is no current event loop set in the current OS thread, the OS thread is main, and set_event_loop() has not yet been called, asyncio will create a new event loop and set it as the current one.

https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop

pgavlin (approved):
LGTM--thanks for expanding that comment.

PR #8136, opened 2021-10-05
Do not share maps so tests can run in parallel

t0yv0:
Running under `PULUMI_PARALLEL_SDK_CODEGEN_TESTS=1` is nice in development but it used to crash docs/ tests. The crash was resulting from concurrent access to module-static state. This is now a struct private to the test.

# Description

Fixes # (issue)

## Checklist

### DIFF
--- a/pkg/codegen/docs/gen.go
+++ b/pkg/codegen/docs/gen.go
@@ -41,37 +41,23 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 )
 
-var (
-	// modules is a map of a module name and information
-	// about it. This is crux of all API docs generation
-	// as the modContext carries information about the resources,
-	// functions, as well other modules within each module.
-	modules map[string]*modContext
+// Populated in auto-generated `packaged.go`
+var packagedTemplates map[string][]byte
 
-	supportedLanguages = []string{"csharp", "go", "nodejs", "python"}
-	snippetLanguages   = []string{"csharp", "go", "python", "typescript"}
-	templates          *template.Template
-	packagedTemplates  map[string][]byte
-	docHelpers         map[string]codegen.DocLanguageHelper
+func init() {
+	packagedTemplates = map[string][]byte{}
+}
 
-	// The language-specific info objects for a certain package (provider).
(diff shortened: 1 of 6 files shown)


pkg/codegen/docs/gen.go
```diff
@@ -363,8 +383,8 @@ func resourceName(r *schema.Resource) string {
 	return strings.Title(tokenToName(r.Token))
 }
 
-func getLanguageDocHelper(lang string) codegen.DocLanguageHelper {
-	if h, ok := docHelpers[lang]; ok {
+func getLanguageDocHelper(dctx *docGenContext, lang string) codegen.DocLanguageHelper {
```
pgavlin, line 386: Nit: I have a _slight_ preference for making this (and other functions that take a `*docGenContext` as their first param) methods on `*docGenContext`

pgavlin (approved):

pkg/codegen/internal/test/sdk_driver.go
```diff
@@ -246,10 +246,8 @@ type SDKCodegenOptions struct {
 func TestSDKCodegen(t *testing.T, opts *SDKCodegenOptions) { // revive:disable-line
 	testDir := filepath.Join("..", "internal", "test", "testdata")
 
-	// Motivation for flagging parallelism: not all tests are
-	// parallel-safe yet (for example, codegen/docs tests fail),
-	// and there are concerns about memory utilizaion in CI. It
-	// can be a nice feature for developing though.
+	// Motivation for flagging: concerns about memory utilizaion
+	// in CI. It can be a nice feature for developing though.
 	parallel := os.Getenv("PULUMI_PARALLEL_SDK_CODEGEN_TESTS") != ""
```
iwahbe, line 251:
```suggestion
	parallel := cmdutil.IsTruthy(os.Getenv("PULUMI_PARALLEL_SDK_CODEGEN_TESTS"))
```
I don't love that setting `PULUMI_PARALLEL_SDK_CODEGEN_TESTS` to `"false"` makes it true.

iwahbe (approved):
LGTM

One thought. Is it work defaulting this to on and then setting `PULUMI_PARALLEL_SDK_CODEGEN_TESTS=false` in the CI environment?

PR #8171, opened 2021-10-07
Iwahbe/8000/uprgade go versions

### LINKED ISSUE #8000: Upgrade to Go 1.17
https://golang.org/doc/go1.17

iwahbe:
# Description

Fixes #8000

## Checklist


pgavlin:
> The rest don't really matter I'd argue as they are test cases and we support go 1.14+

I might even go the other way and say that we _want_ to leave the test-cases as-is for exactly this reason.

PR #8187, opened 2021-10-11
Fix default package name

### LINKED ISSUE #8181: Go codegen doesn't handle hyphenated package names
When I use this in `schema.json` for a package named `aws-apigateway`:

```yaml
  go:
    importBasePath: github.com/lukehoban/pulumi-package-aws-apigateway/sdk/go/apigateway
```

I end up with `go/apigateway/restAPI.go` with package `aws`.

If I use this:

```yaml
  go:
    importBasePath: github.com/lukehoban/pulumi-package-aws-apigateway/sdk/go/apigateway
    rootPackageName: apigateway
```

I end up with `go/restAPI.go` with package `apigateway`.

But neither of these seems to be "right" - I want (and kind of assume I should get by default) `go/apigateway/restAPI.go` with package `apigateway`.

It feels like we should make one or both of these changes:
1. The default package name should not be `strings.Split(name, "-")[0]` but perhaps `strings.ReplaceAll(name, "-", "")`
2. If `importBasePath` is specified, the last part of the path (`apigateway` in the examples above) should be used as the package name, instead of "guessing" it from the `strings.Split(name, "-")[0]` logic.  Or at least, that logic should apply to the final part of the user-provided import base path.

iwahbe:
Resulting behavior: 

| name          | importBasePath                                        | rootPackageName | resulting root | resulting name                    |
|:-------------:|:-----------------------------------------------------:|:---------------:|----------------|:---------------------------------:|
| `hello-world` | ` `                                                   | ` `             | `hello-world`  | `helloworld`                      |
| `hello-world` | `github.com/org/my-pulumi-package-aws-thing/my-thing` | ` `             | `mything`      | `mything`                         |
| `hello-world` | `github.com/org/my-pulumi-package-aws-thing/thing`    | `bar`           | ` `            | `bar`   |
| ` `           | `github.com/org/my-pulumi-package-aws-thing/thing`    | `bar`           | ` `            | `bar`  |
| `hello-world` | ` `                                                   | `bar`           | ` `            | `bar`   |
| ` `           | ` `                                                   | `bar`           | ` `            | `bar`    |

# Description

Corrects the default behavior for deriving package names from the `name`, `importBasePath` and `rootPackageName`.

Fixes #8181

## Checklist


iwahbe:
We assume that the last segment of `importBasePath` is a valid go identifier. This results in invalid code generated from paths like `"simple-enum-schema/plant-provider"`.

Issue #8177, opened 2021-10-08
Make ts-node a peer dependency


dmattia:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

the nodejs sdk specifies ts-node and typescript versions in its dependencies: https://github.com/pulumi/pulumi/blob/a92a005d6897506d3ea40307dfd37788c51a4dcf/sdk/nodejs/package.json#L24-L25

I've been hitting issues using `"ts-node": "=8.10.2"` in my own project where I frequently see errors when I run `pulumi up` like:

<img width="1273" alt="Screen Shot 2021-10-08 at 11 12 31 AM" src="https://user-images.githubusercontent.com/8922077/136596102-14e457b7-14d5-4b74-864f-975cd46133c9.png">

I can work around this by using yarn resolutions, like:

```json
"resolutions": {
    "ts-node": "=8.10.2"
  },
```

but this affects ts-node in my entire yarn workspace, which is not what I want.

By making ts-node a peer dependency (still with a minimum version specified), it would be easier to bring my own version I want to use.

### Affected area/feature

Pulumi nodejs sdk

komalali:
Hi! This is an interesting idea. FWIW, there is context on why we have avoided updating ts-node thus far and are pinned to this version: https://github.com/pulumi/pulumi/issues/4876

dmattia:
A message from the Slack has another workaround idea (https://pulumi-community.slack.com/archives/C84L4E3N1/p1633725538463200): 

```text
we have similar issues very often - compiling typescript works, ts-node throws weird errors. We tweaked our tsconfig to match ts-node requirements closer, but we still get them.
We are now investigating disabling ts-node entirely and running @pulumi/pulumi on compiled code. Oddly, a service fails without an error, though it deploys with TS enable and with the CLI …
Btw, we’re using ts-node 10.* - had fewer issues than with 8.*
```

I can try running on already built typescript code, but that is still unideal.

dixler:
@dmattia thanks again for the PR. We're trying to think about ways to introduce the changes described in #10122 without breaking people's projects if we want to do this before 4.0.

cc @RobbieMcKinstry  @t0yv0

julienp:
Done in https://github.com/pulumi/pulumi/pull/15622

Issue #4876, opened 2020-06-23
Revisit ts-node version


leezen:
Currently, [we're using an older version of ts-node](https://github.com/pulumi/pulumi/pull/4007) because it provides caching. Recent updates suggest that performance has improved and we should consider upgrading.

stack72:
@justinvp / @lukehoban / @leezen what are your thoughts on the version of ts-node we should try? We can try and send a PR to ensure tests all pass as expected

justinvp:
Just a note: If more recent versions of ts-node don't pan out, we should look into babel-node.

justinvp:
> what are your thoughts on the version of ts-node we should try?

We should try the latest version, e.g. `^10.2.0`, and see if CI passes and also check to see if there is a perf difference before/after the change. And assuming all is good, merge and then also check a run of the examples repo to further verify nothing is broken by the upgrade.

iwahbe:
It turns out that there is still a major (2.5x) performance regression with upgrading to `^10.2.0`.

delitescere:
@iwahbe Thanks for checking this out! I'm curious, what does that performance regression translate to, into real terms?

I am assuming the limiting factor in a `pulumi up` is generally going to be the speed of provisioning cloud resources, so the speed of ts-node isn't going to be a constant factor.

Is the performance regression you observed based on:
1. a microbenchmark
1. "a `pulumi preview` that took 10 seconds now takes 25 seconds"
1. "Pulumi Node.js SDK tests that took 10 minutes now take 25 minutes"
1. "a `pulumi up` that took 5 minutes now takes 12 minutes"

The first two are (imo) non-issues. The third is undesirable and the fourth is unacceptable. However, the "inability to do things that the old ts-node doesn't support" is also a significant impact.

lukehoban:
There's more details on the performance issues related to this in https://github.com/pulumi/pulumi/issues/3671#issuecomment-592810579. 

[...]

PR #8199, opened 2021-10-12
Make ts-node and typescript peer dependencies

dmattia:
# Description

Fixes https://github.com/pulumi/pulumi/issues/8177
Related to https://github.com/pulumi/pulumi/issues/4876

Newer ts-node and typescript versions can make configurations not work with Pulumi at all right now. In my case, I am developing in a monorepo that uses newer versions of ts-node and typescript, and these older versions cause my code to not be able to run. See the more detailed descriptions I left in those two Issues linked above for what exactly is breaking

## Checklist

This is not an application code change

I'm not sure where to update this version, but this deserves probably a minor bump in the node sdk

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 CHANGELOG
 =========
 
+# 3.35.5 (2022-07-01)
+
+### Misc
+
+- Testing fork
+
 # 3.35.3 (2022-07-01)
 
 ### Improvements


lukehoban:
It does seem like a good idea to allow these dependencies to be controlled by consumers of this library.  That said, I suspect this change as-is will break all existing Pulumi+TypeScript users, as nothing will provide the peer dependencies.  For new users, we would need to include versions of these dependencies in our templates.  But for existing users, this would require every one of them to make edits to their own package.json in order to adopt this.  I suspect we could not take a change like that until 4.0, unless we could figure out a way to avoid that breaking change.

Unfortunately, neither npm or yarn supports the ability to provide a default value, but allow consumers to override this, which is what we really want here to ensure this can be made more flexibile without a breaking change.  https://github.com/yarnpkg/berry/issues/1001 is tracking what appears to be the needed feature in yarn, but does not look to be imminent (and we likely couldn't rely on any new release of these tools as they wouldn't be widely adopted yet).  

@pgavlin @justinvp @dmattia Any thoughts on how we might be able to achieve this without a breaking change?  Or if not, is this something we should consider for 4.0?

dmattia:
If anyone else using yarn v2 wants to use this fork, you can by setting a resolution to it in your top level `package.json` like:

```json
"resolutions": {
    "@pulumi/pulumi": "exec:./use-pulumi-fork.js"
  },
```

And then create a file `use-pulumi-fork.js` with the contents:

```javascript
/**
 * Script to force yarn v2 to use the dmattia/pulumi fork of the Pulumi NodeJS SDK.
 *
 * Because pulumi/pulumi does not have a package.json file in its root, yarn (and npm)
 * do not know how to import it as a dependency from a normal git reference.
 *
 * This script just points to the dmattia/pulumi Github repository and builds the node
 * archive from the `/sdk/nodejs` folder inside the fork.
 *
 * Pull Request to track progress of the fork: https://github.com/pulumi/pulumi/pull/8199
 */

const pathToRepo = path.join(execEnv.tempDir, 'repo');
const pathToSdkDir = path.join(pathToRepo, 'sdk/nodejs');
const pathToArchive = path.join(execEnv.tempDir, 'archive.tgz');

// Clone the repository
child_process.execFileSync(`git`, [
  `clone`,
  `git@github.com:dmattia/pulumi`,
  pathToRepo,
]);

// Install the dependencies
child_process.execFileSync(`yarn`, [`install`], {
  cwd: pathToSdkDir,
});

// Pack a specific workspace
child_process.execFileSync(`yarn`, [`pack`, `--out`, pathToArchive], {
  cwd: pathToSdkDir,
});

// Send the package content into the build directory
child_process.execFileSync(`tar`, [
  `-x`,
  `-z`,
  `--strip-components=1`,
  `-f`,
  pathToArchive,
  `-C`,
  execEnv.buildDir,
]);
```

PupoSDC:
Hello!

I'm just giving this a nudge to try to see if thsi does not get forgotten ;) 

TS 3.7 is getting pretty outdated, and we are struggling quite a bit migrating some code to use pulumi as it uses features that 3.7 didn't have resulting in some cryptic errors that only show up once you do "pulumi up".

it would be great for DX if this could be merged at some point!

cheers!

EvanBoyle:
related to https://github.com/pulumi/pulumi/issues/7184

dmattia:
@EvanBoyle @lukehoban Could I get a clarification on why this PR was closed? This PR is certainly related to issue https://github.com/pulumi/pulumi/issues/7184, as I commented myself in https://github.com/pulumi/pulumi/issues/7184#issuecomment-983831283, as it is an attempted solution.

I've been using a fork of Pulumi with this change for a few months now (https://www.npmjs.com/package/@transcend-io/pulumi/v/1.0.5) and it gets around the problems #7184 mentions and is the only solution I have right now to make typechecking work with newer typescript/ts-node versions.

dmattia:
Re-opened fixed up version in https://github.com/pulumi/pulumi/pull/10122

PR #8202, opened 2021-10-12
[auto/dotnet] - pulumi state delete & unprotect functionality

orionstudt:
# Description

Adds `pulumi state delete` and `pulumi state unprotect` functionality to the .NET Automation API.

## Checklist


orionstudt:
Any thoughts on this API design? @mikhailshilkov @t0yv0 @justinvp 

I thought trying to name the methods something like `StateDelete` or `StateUnprotect` would be awkward so instead I separated them into a module on `WorkspaceStack` so they are invoked like `stack.State.Delete(....)`

Also should you be able to take this action on the root `Workspace` and provide a stack name as well?

pgavlin:
> Any thoughts on this API design? @mikhailshilkov @t0yv0 @justinvp

+ @komalali 

> I thought trying to name the methods something like StateDelete or StateUnprotect would be awkward so instead I separated them into a module on WorkspaceStack so they are invoked like stack.State.Delete(....)

This seems pretty reasonable to me. I think this is new ground for the Automation API in general, but I think that the general shape should translate well to all of our current platforms.

PR #8214, opened 2021-10-13
Pin grpc version

iwahbe:
# Description

As @komalali recommended, I have pinned the grpc version. 

Fixes # (issue)

## Checklist


lionello:
Should this be pinned to v1.3.x indefinitely? I'm having build issues because of this. https://github.com/grpc/grpc-node/issues/2002

iwahbe:
It should not be pinned indefinitely. I opened https://github.com/pulumi/pulumi/issues/12312 to track this.

PR #8216, opened 2021-10-13
Update workflows to pin goreleaser to v0.181.1

emiliza:
~Updating from `macos-latest` which currently is picking up macos-11 workers.  This is a temporary workaround due to Goreleaser not being able to be installed on Mac Os 11 see: https://github.com/goreleaser/goreleaser/issues/2575~

edit: ...because there's a better way to solve it, pin the dependency - v0.181.1 has the right assets we need

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -112,7 +112,7 @@ jobs:
       - name: Run GoReleaser
         uses: goreleaser/goreleaser-action@v2
         with:
-          version: latest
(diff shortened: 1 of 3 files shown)


emiliza:
From discord sounds like there's a fix in flight. I'll move this PR to standby in case we need to cut a release before goreleaser's updated 🍀

t0yv0:
Thanks for pushing this in to unblock an out-of-schedule release today.

PR #8219, opened 2021-10-14
Simplify the .NET implementation of secret inputs

t0yv0:
# Description

Following discussion on https://github.com/pulumi/pulumi/pull/7128

It seems that `.Apply(Ouput.CreateSecret)` at least compiles. It should do the same thing in theory. Can we think of failure modes? I could add unit tests at runtime to explore further.

Fixes # (issue)

## Checklist


Frassle:
dotnet code has moved repo so closing this off, reopen in pulumi-dotnet if still relevant

PR #8234, opened 2021-10-15
Programgen updates to utilize fn.Output forms in examples and API docs (7949)

### LINKED ISSUE #7949: Update program gen to employ new $fn_output forms

t0yv0:
# Description

The goal is to teach programgen to emit fn.Output forms of invokes to simplify example generation.

Fixes #7949

## Checklist


t0yv0:
With the latest change the example generates valid code. But I think there might still be problems with examples that nest. I think that `argumentTypeName` is inaccurate in the general case, since it needs to make assumptions that are matching the assumptions of SDK codegen in `func (mod *modContext) typeName(t *schema.ObjectType, state, input, args bool)`

For example:

https://github.com/pulumi/pulumi/pull/7899/files#diff-c44352515fd31f42be144a763ab5771fe41195a15c2c1213d41a8cbc94dd69d4R220

Those assumptions are getting some complexity due to flags and bw-compatib constraints.

Thinking out loud here; the program code generator found and is reading the schema, presumably from the generated pulumi-aws package? Should we also embed mappings from `model.Type` to the target language nominal type coordinates (FQN) into the generated package? Right now SDK codegen and program gen assume that these are fixed rules (functions) from the schema into these, but ultimately it's a property of the generated package it seems, that can vary version by version.


```csharp

        using Pulumi;
        using Aws = Pulumi.Aws;
        
        class MyStack : Stack
        {
            public MyStack()
            {
                var privateS3VpcEndpoint = new Aws.Ec2.VpcEndpoint("privateS3VpcEndpoint", new Aws.Ec2.VpcEndpointArgs
                {
                    VpcId = aws_vpc.Foo.Id,
                    ServiceName = "com.amazonaws.us-west-2.s3",
                });
                var privateS3PrefixList = Aws.Ec2.GetPrefixList.Invoke(new Aws.Ec2.GetPrefixListInvokeArgs
                {
                    PrefixListId = privateS3VpcEndpoint.PrefixListId,
                });
                var bar = new Aws.Ec2.NetworkAcl("bar", new Aws.Ec2.NetworkAclArgs
                {
                    VpcId = aws_vpc.Foo.Id,
                });
                var privateS3NetworkAclRule = new Aws.Ec2.NetworkAclRule("privateS3NetworkAclRule", new Aws.Ec2.NetworkAclRuleArgs
                {
                    NetworkAclId = bar.Id,
                    RuleNumber = 200,
                    Egress = false,
                    Protocol = "tcp",
                    RuleAction = "allow",
                    CidrBlock = privateS3PrefixList.Apply(privateS3PrefixList => privateS3PrefixList.CidrBlocks?[0]),
                    FromPort = 443,
                    ToPort = 443,
                });
            }
        
        }
```

pkg/codegen/go/gen_program_expressions.go
```diff
@@ -814,9 +858,14 @@ func (g *generator) genApply(w io.Writer, expr *model.FunctionCallExpression) {
 	isInput := false
 	retType := g.argumentTypeName(nil, then.Signature.ReturnType, isInput)
 	// TODO account for outputs in other namespaces like aws
-	typeAssertion := fmt.Sprintf(".(%sOutput)", retType)
-	if !strings.HasPrefix(retType, "pulumi.") {
-		typeAssertion = fmt.Sprintf(".(pulumi.%sOutput)", Title(retType))
+	var typeAssertion string
+	if retType == "[]string" {
```
t0yv0, line 862: Calling this out, something is really patchy/incomplete here even after the fix in the PR. Does it ring any bells?

pgavlin:
> this is ready for review, PTAL. 

Thanks for the heads up!

> I can break out per lang or rejigger commits if this is too much. This passes basic tests for me and I think does the trick.

If it's not too much trouble, would you mind factoring the changes into a commit per-language?

t0yv0:
Yes, totally, even separate PRs perhaps. I'll factor and rebase tomorrow.

t0yv0:
- [x] Go https://github.com/pulumi/pulumi/pull/8431
- [ ] Python
- [ ] Node
- [ ] .NET

t0yv0:
CLose in favor of rebased per-language PRs.

PR #8243, opened 2021-10-18
Some improvements to the vscode devcontainer

Frassle:
* Installed the go, dotnet, and python extensions by default.
* Set the gopls settings to ignore copies of modules in build directories.
* Set XDG_CACHE_DIR/CONFIG_DIR so that vscode can install go tools.
* Use "containerEnv" instead of "runArgs" to set the container environment.
* Turn on gopls "experimentalWorkspaceModule" so we can open at the root pulumi directory.
* Use "postCreateCommand" to restore all dotnet projects so ominisharp doesn't complain about missing types on startup.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -32,8 +32,12 @@ RUN mkdir -p /go/bin \
 
 USER $USER_NAME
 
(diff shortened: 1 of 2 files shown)


Frassle:
If you open the pulumi folder in vscode it _should_ pop up a message saying it's found a devcontainer and ask if you want to reopen in that container. If you don't get that you can do it via the command bar (Ctrl+Shift+P) and search for "Open workspace in container..."

PR #8260, opened 2021-10-19
[cli, testing, github] Gather code coverage data in CI.

pgavlin:
These changes add support for gathering code coverage data during tests.

For tests that do not involve the Pulumi CLI, this is straightforward: all of the ecosystems we target already support gathering coverage data, and we follow the rules accordingly. Support for each language is broken out into its own commit.

For tests that do involve the Pulumi CLI, the picture is a bit more complicated. Go does not make it trivial to perform a coverage-instrumented build (`go build` does not have a `-cover` flag, for example). In lieu of official support, we abuse `go test -c` and `TestMain` to produce a build of the CLI that supports collecting and reporting coverage data.


stack72:
My only comment here is that I really don't think this should be part of the Pulumi/Pulumi pkg mod - this feels like it should be created elsewhere. This is an internal utility that is now going to be in our dependency chain

pgavlin:
> My only comment here is that I really don't think this should be part of the Pulumi/Pulumi pkg mod - this feels like it should be created elsewhere. This is an internal utility that is now going to be in our dependency chain

I assume you're referring to the coverage merger? If so, I'm happy to move that out into a different repo (maybe pulumictl, since we already install that everywhere)

pgavlin:
This is ready for review. I'd like to get this landed early next week (ideally Monday). The latest changes depend on https://github.com/pulumi/pulumictl/pull/42.

pkg/cmd/pulumi/main_test.go
```diff
@@ -0,0 +1,81 @@
+package main
```
justinvp, line 1: Nit: copyright

justinvp (approved):
LGTM

PR #8262, opened 2021-10-19
[codegen/docs] Sort the package tree items

praneetloke:
# Description

Sort the modules, resources and functions independently as individual groups to maintain the order of listing modules first followed by resources and then the functions.

Fixes https://github.com/pulumi/registry/issues/211

## Checklist

  - Not a CLI or SDK change
  - None

### DIFF
--- a/pkg/codegen/docs/gen_test.go
+++ b/pkg/codegen/docs/gen_test.go
@@ -19,6 +19,7 @@
 package docs
 
 import (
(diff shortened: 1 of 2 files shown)


cnunciato:
@praneetloke am I reading it right that this would sort just the top-level items, but not nested ones? Or would all children be sorted as well?

Issue #7813, opened 2021-08-20
[codegen/schema] Add package metadata to the Pulumi schema spec


praneetloke:
## Affected feature

Today when we generate resource docs, the docs generator uses a [lookup map](https://github.com/pulumi/pulumi/blob/master/pkg/codegen/docs/gen.go#L62) for mapping a package's name to its "display friendly" name. For example, `aws` -> `Amazon Web Services` etc. It would be good to not rely on the lookup table and instead let that information flow from the schema spec itself so that it becomes the source of truth for that information. I am proposing that we add a `title` property at the root of the spec.

If we could also add `category`, `publisher`, whether or not a package uses TF bridge and the package's availability status (preview vs. ga). `category` would simply be a single string value of a package's category. It's how we categorize packages today on the [API Reference](https://www.pulumi.com/docs/reference/pkg/) and the [Cloud Providers](https://www.pulumi.com/docs/intro/cloud-providers/) pages. As for whether or not a package uses TF bridge we have a few native providers and it would be helpful to discern them from the TF bridge-based providers. The package's availability status is pretty self-explanatory. The `description` property already mentions the preview status of a provider so maybe it's ok to not have a separate field?

Overall, these are the fields I talked about: `title`, `category`, `publisher`, `usesTfBridge`, `availabilityStatus` (the property names are purely suggestions only)

infin8x:
I'd like to propose a slightly different set of schema additions:

| Property            | Type                                       | Required | Description                                                                                                                       |
|---------------------|--------------------------------------------|----------|-----------------------------------------------------------------------------------------------------------------------------------|
| `publisher`         | `string`                                   | Yes      | Publisher is the name of the person or organization that authored and published the package.                                      |
| `displayName`       | `string`                                   | No       | DisplayName is the human-friendly name of the package, used when the package is displayed on Pulumi.com                           |

Then, to @justinvp's suggestion [in Slack](https://pulumi.slack.com/archives/CBUP505U7/p1629513669109600?thread_ts=1629422079.092700&cid=CBUP505U7), I wonder if we can introduce some structure into the `keywords`? For example, could we have keywords like:

- `category:cloud`
- `kind:[terraform-provider|native-provider|component]`

And finally, I wonder if we can drive the availability status (e.g. "public preview" vs. "generally available") off of SemVer, instead of as a distinct metadata item.

@praneetloke @justinvp would love your thoughts?

praneetloke:
[...]

PR #8276, opened 2021-10-20
Add DisplayName and Publisher to the schema

praneetloke:
# Description

Just as the title states. Note that I didn't add anything for identifying if a package is a component. I see that we have an `IsComponent` on a `ResourceSpec` but not on the `PackageSpec` itself. However, just as with the suggestions for `category` being in the `keywords` with a GH-label like format I think we could do `kind/component` in the `keywords` and therefore, not have to add anything explicitly. Let me know if you think otherwise.

These additions will help ensuring that we generate package documentation from the schema and don't have to rely on users to specify them manually while running the various docs generator tools.

Related to #7813

## Checklist

  - I didn't add tests. Let me know if you think these changes warrant a test.
- [ ] ~Yes, there are changes in this PR that warrants bumping the Pulumi Service API version~
  - No changes related to the service.

### DIFF
--- a/pkg/codegen/schema/pulumi.json
+++ b/pkg/codegen/schema/pulumi.json
@@ -10,6 +10,10 @@
             "type": "string",
             "pattern": "^[a-zA-Z][-a-zA-Z0-9_]*$"
         },
+        "displayName": {
+            "description": "The human-friendly name of the package.",
+            "type": "string"
+        },
         "version": {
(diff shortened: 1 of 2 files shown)


praneetloke:
I wasn't sure if updates to the schema get advertised in the changelog. So please let me know if I should add an entry for this.

PR #8293, opened 2021-10-26
7947: docgen for $fnOutput forms

### LINKED ISSUE #7947: Update docs to use new fn_output forms for funcalls so that this shows up in API reference documentation
Following up on #5758 we need to make a change to pkg/codegen/docs so that it uses the new fn_output forms as appropriate.

t0yv0:
# Prerequisites

Prerequisites before this can be merged:

- [x] pulumi-azure-native updated to support output-versioned invokes for C#. How to check: observe GetGallery.Invoke being emitted in addition to GetGallery.InvokeAsync in https://github.com/pulumi/pulumi-azure-native/blob/master/sdk/dotnet/Compute/GetGallery.cs#L18 for example

- [x] pulumi-aws updated to support output-versioned invokes for C#. How to check: observe GetAmi.Invoke being emitted in addition to GetAmi.InvokeAsync in https://github.com/pulumi/pulumi-aws/blob/master/sdk/dotnet/Ec2/GetAmi.cs#L77 for example

# Description

Fixes #7947 

The change is intended to affect function documentation pages such as the following one, to guide the user on the availability of fn.Output helper forms.

**Before**: https://www.pulumi.com/registry/packages/aws/api-docs/getamiids/#getamiidsfilter 

**After**: http://pulumi-docs-origin-pr-6704-d6d3b74c.s3-website.us-west-2.amazonaws.com/registry/packages/aws/api-docs/ec2/getamiids/

It currently modifies the signature emission under "Using getAmiIds" to showcase the new form, and adds a paragraph explaining that there are two forms to choose from. The change intentionally does not duplicate the arguments and result tables for the output form (this would mostly duplicate information).

It is a bit unclear how to best support the user in documenting the helper types. To take the above example the page is listing `GetAmiIdsFilter` as helper type. The naming of the helper types gets a bit confusing and non-intuitive in the various language targets:

```
       Python     : Input[Optional[Sequence[InputType[GetAmiIdsFilterArgs]]]] (InputType = Union[T, Mapping[str, Any]])
       Go         : GetAmiIdsFilterArrayInput ~ GetAmiIdsFilterArray = []GetAmiIdsFilterInput
       TypeScript : Input<Array<Input<GetAmiIdsFilterArgs>>>
       C#         : InputList<GetAmiIdsFilterInputArgs>
```

In languages like TypeScript and Python, the type names do not matter much, but they do in Go and C#. There is some hope that the related changes to programgen https://github.com/pulumi/pulumi/pull/8234 will simplify the examples. I am leaning to backlogging further work on documenting these helper types, but open to suggestions here.

- [x] TODO NeedsOutputVersion now has a Go-specific flag, make use of that to not generate misleading docs for providers such as azure-native that opt out 

## Checklist


t0yv0:
Documented unrelated CI failure    https://github.com/pulumi/pulumi/issues/8382 - merging

Issue #7900, opened 2021-09-02
Pulumi C#/.NET hangs when previewing a program due to dependency cycles


t0yv0:
This issue stems from cross-checking the https://github.com/pulumi/pulumi/issues/7862 problem in C#/.NET. The problem does exist.

## Steps to reproduce

```csharp
using Pulumi;
using Pulumi.Aws.S3;
using Pulumi.Eks;

class MyStack : Stack
{
    public MyStack()
    {
        var cluster = new Cluster("example",
                                  new ClusterArgs()
                                  {
                                      VpcId = "doesntmatter",
                                      SkipDefaultNodeGroup = true,
                                  });

        new Pulumi.Aws.Eks.NodeGroup(
            "example-ng",
            new Pulumi.Aws.Eks.NodeGroupArgs()
            {
                ClusterName = "example",
                InstanceTypes = "t3.micro",
                NodeRoleArn = cluster.InstanceRoles.Apply(xs => xs[0].Arn),
                SubnetIds = "doesntmatter",
                ScalingConfig = new Pulumi.Aws.Eks.Inputs.NodeGroupScalingConfigArgs()
                {
                    DesiredSize = 1,
                    MaxSize = 1,
                    MinSize = 1,
                },
            },
            new Pulumi.CustomResourceOptions()
            {
                Parent = cluster,
            }
        );
    }
}
```

Expected: pulumi preview works
Actual: pulumi preview hangs

PR #8316, opened 2021-10-30
Enable output marshaling in .NET

### LINKED ISSUE #8157: Output Values: .NET SDK marshaling (used in SDK)
Part of: #7434

This issue is the .NET equivalent for:
- https://github.com/pulumi/pulumi/pull/7925
- https://github.com/pulumi/pulumi/pull/8014/commits/0c7d0343c498ecc213070a63792868e66cd48175
- https://github.com/pulumi/pulumi/pull/8014/commits/fa645feb9dd479d4094d6395ea045cc8c48df97e

We can model the returned special value off of Secretes and Resource References: https://github.com/pulumi/pulumi/blob/d39a14432fb51f8af3adb9d4c4e8fcb77c3710e3/sdk/dotnet/Pulumi/Serialization/Serializer.cs#L151-L157

iwahbe:
# Description

Fixes #8157 

## Checklist


sdk/dotnet/Pulumi/Serialization/Serializer.cs
```diff
@@ -137,17 +140,49 @@ public Serializer(bool excessiveDebugOutput)
...
+
+                    var depencdencies = await Deployment.GetAllTransitivelyReferencedResourceUrnsAsync(propResources);
+                    var builder = ImmutableDictionary.CreateBuilder<string, object?>();
+                    builder.Add(Constants.SpecialSigKey, Constants.SpecialOutputValueSig);
+                    if (isKnown)
+                        builder.Add("value", value);
+                    if (isSecret)
+                        builder.Add("secret", isSecret);
+                    if (depencdencies.Count > 0)
+                        builder.Add("dependencies", depencdencies.ToArray());
```
t0yv0, line 179: typo `depencdencies`?

sdk/dotnet/Pulumi/Serialization/Serializer.cs
```diff
@@ -137,17 +140,49 @@ public Serializer(bool excessiveDebugOutput)
...
+                    foreach (var resource in propResources)
+                    {
+                        var childURNResolver = new Serializer(_excessiveDebugOutput, keepOutputValues: false);
+                        await childURNResolver.SerializeAsync($"{ctx} dependency", resource.Urn, keepResources);
+                        urnDeps.UnionWith(childURNResolver.DependentResources);
+                    }
+                    propResources.UnionWith(urnDeps);
+                    this.DependentResources.UnionWith(urnDeps);
+
+                    var depencdencies = await Deployment.GetAllTransitivelyReferencedResourceUrnsAsync(propResources);
```
t0yv0, line 171: Do we need the transitive closure here?  Is that how the other SDKs do it?

I also wonder if this leads to hangs with cycles, exacerbating https://github.com/pulumi/pulumi/issues/7900

sdk/dotnet/Pulumi/Deployment/Deployment_Call.cs
```diff
@@ -53,8 +53,10 @@ private Output<T> Call<T>(string token, CallArgs args, Resource? self, CallOptio
             }
 
             var (serialized, argDependencies) = await SerializeFilteredPropertiesAsync(
-    				$"call:{token}",
-    				argsDict, _ => true, keepResources: true).ConfigureAwait(false);
+                    $"call:{token}",
+                    argsDict, _ => true,
+                    keepResources: true,
+                    keepOutputValues: true).ConfigureAwait(false);
```
t0yv0, line 59: So Call assumes the monitor supports output values? If memory serves me correctly this might be reasonable but good to double-check.

t0yv0 (approved):
LGTM with some questions.

sdk/dotnet/Pulumi/Serialization/Serializer.cs
iwahbe: This is how the other SDKs do it: https://github.com/pulumi/pulumi/blob/e629bc32d5afe9615869060235b907fe6e92f0b2/sdk/nodejs/runtime/rpc.ts#L386. `GetAllTransitivelyReferencedResourceUrnsAsync` should handle cycles: https://github.com/pulumi/pulumi/blob/e629bc32d5afe9615869060235b907fe6e92f0b2/sdk/dotnet/Pulumi/Deployment/Deployment_Prepare.cs#L183

sdk/dotnet/Pulumi/Deployment/Deployment_Call.cs
iwahbe: That is my understanding. It is how it works in the other SDKs.

sdk/dotnet/Pulumi/Serialization/Serializer.cs
```diff
@@ -134,25 +134,56 @@ public Serializer(bool excessiveDebugOutput)
...
+
+                    var dependencies = await Deployment.GetAllTransitivelyReferencedResourceUrnsAsync(propResources);
+                    var builder = ImmutableDictionary.CreateBuilder<string, object?>();
+                    builder.Add(Constants.SpecialSigKey, Constants.SpecialOutputValueSig);
+                    if (isKnown)
+                        builder.Add(Constants.ValueName, value);
+                    if (isSecret)
+                        builder.Add(Constants.SecretName, isSecret);
+                    if (dependencies.Count > 0)
+                        builder.Add(Constants.DependenciesName, dependencies.ToImmutableArray<object>());
```
justinvp, line 175:
```suggestion
                        builder.Add(Constants.DependenciesName, dependencies.ToImmutableArray());
```

sdk/dotnet/Pulumi/Serialization/Serializer.cs
```diff
@@ -134,25 +134,56 @@ public Serializer(bool excessiveDebugOutput)
                 {
                     Log.Debug($"Serialize property[{ctx}]: Recursing into Output");
                 }
-
                 var data = await output.GetDataAsync().ConfigureAwait(false);
                 this.DependentResources.AddRange(data.Resources);
+                var propResources = new HashSet<Resource>();
+                propResources.UnionWith(data.Resources);
```
justinvp, line 140: Also, wondering if it'd be more clear named: `propResources` => `outputResources`.

justinvp (commented):
I have some questions about the serialization/deserialization that I'll spend a little time looking into in more detail and report back.

Otherwise, looking good, aside from a bunch of nitpicky suggestions.

sdk/dotnet/Pulumi/Serialization/Serializer.cs
iwahbe: It's `propResources` because it makes it easier to look between the C# and TypeScript implementations. That's the variable name that `rpc.ts` uses.

sdk/dotnet/Pulumi/Serialization/Serializer.cs
```diff
@@ -134,25 +134,56 @@ public Serializer(bool excessiveDebugOutput)
                 {
```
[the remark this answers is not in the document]
t0yv0: I was wondering what is the concern here. Would anything notice the difference if the serialized form has a `v` or an empty output? Is this just to optimize the payload size a bit?

If we could formalize equivalence "up to autoconvert" for `Output(v, known=True, deps=[], secret=False) ~ v` we could be tolerant when deserializing, and not care in which shape the things are serialized, would that be simpler? Or I am missing something that truly cares which form it is?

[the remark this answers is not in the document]
justinvp: > Is this just to optimize the payload size a bit?

Right. It's not really a problem, it's just a little unfortunate that every single value ends up as an output value. In any case, I went ahead and added a commit that avoids emitting an output value when it's not needed.

More problematic: I've found that in the .NET SDK `Input<T>` (and `InputList<T>`, `InputMap<V>`, and `InputUnion<T0, T1>`) eagerly stores the value of `T` as an `Output<T>`:

https://github.com/pulumi/pulumi/blob/0a38bc295cfc5ae3c325dcece169da5a5e80fffc/sdk/dotnet/Pulumi/Core/Input.cs#L22-L31

So a test like this fails:

```csharp
[Fact]
public static Task TestSecretInInputList()
    => RunInNormal(async () =>
    {
        var input = new InputList<string>{ Output.CreateSecret("hello") };
        var expected = ImmutableArray<object>.Empty.Add(
            ImmutableDictionary<string, object>.Empty
                .Add(Constants.SpecialSigKey, Constants.SpecialOutputValueSig)
                .Add("value", "hello")
                .Add("secret", true));

        var s = new Serializer(excessiveDebugOutput: false);
        var actual = await s.SerializeAsync("", input, keepResources: true, keepOutputValues: true);

        Assert.Equal(expected, actual!);
    });
```

I'm expecting the serialized value to look like (the value inside the list is an output value):

```json
[
    {
        "4dabf18193072939515e22adb298388d": "d0e6a833031e9bbcd3f4e8bde6ca49a4",
        "value": "hello",
        "secret": true
    }
]
```

But actually getting (the top-level value is an output value):

```json
{
    "4dabf18193072939515e22adb298388d": "d0e6a833031e9bbcd3f4e8bde6ca49a4",
    "value": ["hello"],
    "secret": true
}
```

Which basically defeats the purpose of why we added output values to begin with.

It turns out that regular .NET programs (that aren't even using MLCs) end up having undesirable previews because of this behavior with `Input<T>`. For example:

```csharp
using Pulumi;
using Pulumi.Random;
using Pulumi.Aws.S3;

class MyStack : Stack
{
    public MyStack()
    {
        var randString = new RandomString("str", new RandomStringArgs { Length = 10 });
        var bucket = new Bucket("bucket", new BucketArgs {
            Tags = {
                { "hello", "world" },
                { "hi", "there" },
                { "foo", randString.Id },
            }
        });
    }
}
```

Has the following preview diff:

```
        tags        : output<string>
```

When we want it to be like this:

```
        tags        : {
            hello     : "world"
            hi        : "there"
            foo       : output<string>
        }

```

My suggestion is we merge this PR as-is (modulo addressing any review feedback) and open an issue to track this issue with `Input<T>` eagerly converting values to `Output<T>`. And fix that separately. I'm thinking the fix will involve just avoiding/deferring the eager `Output<T>` conversion as much as possible, and only do it when needed.

justinvp:
Ian looked over the commits I added going to merge this. I'll open an issue that tracks addressing the issue with `Input<T>` et. al.

PR #8339, opened 2021-11-02
Fix 8322

### LINKED ISSUE #8322: Output-versioned function invoke fails to preview in C#/.NET
## Hello!

## Issue details

It seems that output-versioned invokes get called in preview with unknown arguments, while they should not. Seems to be a similar issue as https://github.com/pulumi/pulumi/issues/8172 for Python

### Steps to reproduce

```
var resourceGroup = new ResourceGroup("rg");

var workspace = new Workspace("loganalytics", new WorkspaceArgs
{
    ResourceGroupName = resourceGroup.Name,
    Sku = new WorkspaceSkuArgs { Name = "PerGB2018" },
    RetentionInDays = 30,
});

var workspaceSharedKeys = GetSharedKeys.Invoke(new GetSharedKeysInvokeArgs
{
    ResourceGroupName = resourceGroup.Name,
    WorkspaceName = workspace.Name,
});
```

Expected: preview suceeds
Actual: preview fails

t0yv0:
# Description

Fixes #8322 

In other languages like TypeScript, we implement the Output-versioned invokes using Apply. This means in preview, if any arguments are unknown, monitor is not called at all (short-circuit). This was missing in C#, causing breaking previews.

I believe this works but this continues to be a bit awkward at implementation level. Relative to TypeScript and Python, .NET SDK feels a bit lacking in power:

- cannot implement this feature using Apply as it's unclear that the SDK has power to transform inputShape(Args) to Input(Args)

- therefore emulating Apply features indirectly in the serialization related machinery

- decided against detecting unknowns on the native representation (input to the serializer), as doing so involves peeking under lazy Inputs and potentially triggering side-effects, and also it seems a bit tricky to recognize all values correctly

- therefore opted to detecting unknowns AFTER serialization in the serialized form, before it gets projected into protobuf form; this *seems* to be well defined, but needs eyes as the not supported by type system (`object?` trees)

## Checklist


sdk/dotnet/Pulumi/Deployment/Deployment_Invoke.cs
```diff
@@ -81,27 +96,8 @@ private async Task<OutputData<T>> RawInvoke<T>(string token, InvokeArgs args, In
...
-            Log.Debug(label);
-
-            // Be resilient to misbehaving callers.
-            // ReSharper disable once ConstantNullCoalescingCondition
-            args ??= InvokeArgs.Empty;
-
-            // Wait for all values to be available, and then perform the RPC.
-            var argsDict = await args.ToDictionaryAsync().ConfigureAwait(false);
-
-            var keepResources = await this.MonitorSupportsResourceReferences().ConfigureAwait(false);
```
t0yv0, line 96: Note: slight reordering here, in that MonitorSupportsResourceReferences call will now go before args.ToDictionaryAsync() await, and before   Log.Debug(label); hopefully this is benign.

sdk/dotnet/Pulumi/Deployment/Deployment_Serialization.cs
```diff
@@ -85,5 +96,24 @@ public partial class Deployment
                 propertyToDependentResources = PropertyToDependentResources;
             }
         }
+
+        private readonly struct RawSerializationResult
```
t0yv0, line 100: Helper.. Alternatively can go for SerializationResult<T>, or tuples.

pgavlin:
> cannot implement this feature using Apply as it's unclear that the SDK has power to transform inputShape(Args) to Input(Args)

This has come up before, but I want to double-check something for my own understanding: does this mean that without output-shaped functions, .NET users cannot use `Apply` to achieve the same result?

t0yv0:
> This has come up before, but I want to double-check something for my own understanding: does this mean that without output-shaped functions, .NET users cannot use Apply to achieve the same result b/c the C# type system can't express what we want?

The users can always use Apply to emulate fn.Output form by coding the transformation from inputs to Args. That is always possible. However they need to code the transformation in each specific case separately.

What is limited is our generic programming model, not the specific programming model. We're having difficulty here writing code that will work correctly *for all* possible concrete Args types. 

In TypeScript, the advanced type system takes care of it by being able to type-check the generic programs.

However this is not the only way. In languages with simpler (or no) type systems, reflective programming achieves the same effect.

Basically if we had something like this:

```csharp
PlainShape<T1,T2>(T1 term) Input<T2> 
PlainShapeReflective(Type t1, Type t2, object term) object // or this
```

Then we could implement like this:

```csharp
public Task<GetAmiIdsResult> InvokeAsync(GetAmiIdsArgs args, InvokeOptions? options = null)

public Output<GetAmiIdsResult> Invoke(GetAmiIdsInvokeArgs args, InvokeOptions? options = null)
{
   PlainShape<GetAmiIdsInvokeArgs, GetAmiIdsArgs>(args).Apply(options => InvokeAsync(args, options))
}
```

Which I think would be cleaner, definitely avoid bugs like the one fixed here. However I think .NET SDK does not yet know how to do something like `PlainShape`, and it's a bit of work to teach it.

pgavlin:
> The users can always use Apply to emulate fn.Output form by coding the transformation from inputs to Args. That is always possible. However they need to code the transformation in each specific case separately, akin to this:

Okay, this is very helpful. Thanks for clarifying!

sdk/dotnet/Pulumi/Deployment/Deployment_Invoke.cs
```diff
@@ -55,7 +55,22 @@ private async Task<OutputData<T>> RawInvoke<T>(string token, InvokeArgs args, In
...
             // tracking, which is a good future direction also for
             // `Invoke`.
-            var result = await InvokeRawAsync(token, args, options).ConfigureAwait(false);
+
+            var keepResources = await this.MonitorSupportsResourceReferences().ConfigureAwait(false);
+            var serializedArgs = await SerializeInvokeArgs(token, args, keepResources);
+
+            // Short-circuit actually invoking if `Unknowns` are
+            // present in `args`, otherwise preview can break.
+            if (serializedArgs.PropertyValues.Values.Any(Serializer.ContainsUnknowns))
```
justinvp, line 64: Nit: Wondering if it could be:

```suggestion
            if (Serializer.ContainsUnknowns(serializedArgs.PropertyValues))
```

Shorter and avoids the enumerator allocation for `Values` and the delegate allocation for the method group passed to `Any`.

sdk/dotnet/Pulumi/Deployment/Deployment_Invoke.cs
```diff
@@ -55,7 +55,22 @@ private async Task<OutputData<T>> RawInvoke<T>(string token, InvokeArgs args, In
...
+            {
+                return new OutputData<T>(resources: ImmutableHashSet.Create<Resource>(),
```
justinvp, line 66:
```suggestion
                return new OutputData<T>(resources: ImmutableHashSet<Resource>.Empty,
```

justinvp (approved):
I'm still a bit hesitant about how we handle this in .NET (relying on serialization layer rather than `Apply` like the other languages), but I understand the reasons.

Some comments/suggestions, but otherwise LGTM.

sdk/dotnet/Pulumi/Serialization/Serializer.cs
```diff
@@ -385,6 +385,29 @@ internal static Value CreateValue(object? value)
...
+                int _ => false,
+                double d => false,
+                bool b => false,
+                string s => s == Constants.UnknownValue,
+                ImmutableArray<object> list => list.Any(ContainsUnknowns),
+                ImmutableDictionary<string, object> dict =>
+                    dict.Values.Any(ContainsUnknowns) ||
+                    dict.Keys.Contains(Constants.UnknownValue),
+                _ => throw new InvalidOperationException("Unsupported value when converting to protobuf: " + value.GetType().FullName),
+            };
```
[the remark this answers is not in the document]
t0yv0: You've heard Alan Perlis quote that "functional programmers know the Value of everything but the Cost of nothing"? This is very much the situation here :) I'll humor you and incorporate the changes but as you point it's a wash, say Pulumi program has 10,000 outputs that go through this,  10 * 1e-9 * 10000 = 1E-5 seconds savings, this is going to be noise for a Pulumi program like that. I'm very curious if someone runs programs that either have lots of outputs or very deep trees being passed around; understanding those scenarios would actually help construct some benchmarks to be mindful of this.

sdk/dotnet/Pulumi/Deployment/Deployment_Invoke.cs
t0yv0: IT can since PropertyValues happens to be ImmutableDictionary<string,object> which is allowed by the grammar understood by ContainsUnknowns. I was typechecking it using a different type in my head though, have not thought of it.

sdk/dotnet/Pulumi/Serialization/Serializer.cs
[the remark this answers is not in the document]
t0yv0: I'd like to leave 'unsupportedValue' check in place though to remind us when the assumptions of the grammar break. If we introduce new kidns of values there that contain unknowns this will help find/fix the code.

sdk/dotnet/Pulumi/Serialization/Serializer.cs
```diff
@@ -385,6 +385,29 @@ internal static Value CreateValue(object? value)
...
+            {
+                null => false,
+                int _ => false,
+                double d => false,
+                bool b => false,
+                string s => s == Constants.UnknownValue,
+                ImmutableArray<object> list => list.Any(ContainsUnknowns),
+                ImmutableDictionary<string, object> dict =>
+                    dict.Values.Any(ContainsUnknowns) ||
+                    dict.Keys.Contains(Constants.UnknownValue),
```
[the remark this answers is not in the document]
t0yv0: Thanks, noted!

PR #8341, opened 2021-11-02
[codegen/go] Slightly simplify enum type gen.

pgavlin:
- Remove enumElementType, whose functionality duplicates work done in
  argsTypeImpl
- Clarify the args- vs. Go-ness of the two element types

This is prep work for #7943. These changes should not produce codegen diffs.


justinvp (approved):
LGTM

PR #8344, opened 2021-11-03
Fix cmdutil.PrintTable to handle ansi escapes and non-byte glyphs

Frassle:
Fixes two bugs in how padding was calculated in PrintTable.

Firstly we remove all ANSI escape codes from the string before measuring
how wide it is. Secondly we measure glyph count (using rivo/uniseg) not
byte or rune count of the string.

Together these fix the padding/alignment issues I saw when using
PrintTable with plan output. They also slightly change the layout of
"pulumi stack", for example the below is printed with current master and
has 6 characters of space for padding between SecurityGroup and
web-secgrp:

```
Current stack resources (4):
    TYPE                                        NAME
    pulumi:pulumi:Stack                         aws-cs-webserver-test
    ├─ aws:ec2/securityGroup:SecurityGroup      web-secgrp
    ├─ aws:ec2/instance:Instance                web-server-www
    └─ pulumi:providers:aws                     default_4_25_0
```

While printed with this commit you only get 2 characters of space for
padding (which is correct, the column gap is set to "  "):
```
Current stack resources (4):
    TYPE                                    NAME
    pulumi:pulumi:Stack                     aws-cs-webserver-test
    ├─ aws:ec2/securityGroup:SecurityGroup  web-secgrp
    ├─ aws:ec2/instance:Instance            web-server-www
    └─ pulumi:providers:aws                 default_4_25_0
```

### DIFF
--- a/sdk/go/common/util/cmdutil/console.go
+++ b/sdk/go/common/util/cmdutil/console.go
@@ -140,6 +140,9 @@ func (table Table) String() string {
 	return table.ToStringWithGap("  ")
 }
 
+// 7-bit C1 ANSI sequences
+var ansiEscape *regexp.Regexp = regexp.MustCompile(`\x1B(?:[@-Z\\-_]|\[[0-?]*[ -/]*[@-~])`)
+
 func (table *Table) ToStringWithGap(columnGap string) string {
 	columnCount := len(table.Headers)
 
@@ -152,9 +155,6 @@ func (table *Table) ToStringWithGap(columnGap string) string {
 	}}
 
 	allRows = append(allRows, table.Rows...)
-
-	// 7-bit C1 ANSI sequences
-	ansiEscape := regexp.MustCompile(`\x1B(?:[@-Z\\-_]|\[[0-?]*[ -/]*[@-~])`)
 
 	for rowIndex, row := range allRows {
 		columns := row.Columns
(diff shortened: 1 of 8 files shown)


sdk/go/common/util/cmdutil/console.go
```diff
@@ -152,6 +153,9 @@ func (table *Table) ToStringWithGap(columnGap string) string {
 
 	allRows = append(allRows, table.Rows...)
 
+	// 7-bit C1 ANSI sequences
+	ansiEscape := regexp.MustCompile(`\x1B(?:[@-Z\\-_]|\[[0-?]*[ -/]*[@-~])`)
```
iwahbe, line 157: I would consider making `ansiEscape` a global constant. That way it only needs to be compiled once, and not on each call.

iwahbe (approved):
Besides one nit, this looks good (when the tests pass)

pgavlin (commented):
Could we factor the string measurement code out into its own function for easier unit testing? Especially given the regex, which I'm moderately leery of :)

Frassle:
That would be something like:
```
func MeasureText(text string) int { 
  clean := ansiEscape.ReplaceAllString(text, "")
  return uniseg.GraphemeClusterCount(clean)
}
```
?

pgavlin:
Yup, that's right. Nothing fancy, just something a little bit more amenable to unit testing

t0yv0:
Diving deep here, thank you!

PR #8347, opened 2021-11-03
Delete redundant `docker` step from release workflow

justinvp:


.github/workflows/release.yml
```diff
@@ -78,22 +78,6 @@ jobs:
           pulumictl create choco-deploy $(pulumictl get version --language generic -o)
         env:
           GITHUB_TOKEN: ${{ secrets.PULUMI_BOT_TOKEN}}
-  docker:
```
stack72, line 81: we need this block - we just need to remove the `verify-containers`

stack72 (approved):

Issue #3989, opened 2020-02-27
[CLI] Configure to use organization by default


SpainTrain:
## Problem

Currently, running a command like `pulumi new` creates the project and stack in one's personal Pulumi account.  All of my projects and stacks are in our Pulumi organization, so this creates some overhead moving stacks into the organization manually.

**EDIT**

```
pulumi new -s <org>/<stackname>
```

is possible, and solves the problem.  However, this does still require remembering to add the flag every time, even if every project/stack is always made in the same org (and my memory isn't amazing :grin:).

## Feature Suggestion

~~Provide a CLI command and/or flag and/or environment variable which sets the organization/account to be used by commands.~~

**EDIT**

Provide an environment variable and/or wizard option to set default org.

## Related Issues

- https://github.com/pulumi/pulumi/issues/3539 (Requests a superset of what this issue is requesting)
- https://github.com/pulumi/pulumi/issues/1421 (Provides guidance on how to get around part of this limitation)

## Other Background

Based on Slack thread: https://pulumi-community.slack.com/archives/C84L4E3N1/p1582825218176200?thread_ts=1582824975.176100&cid=C84L4E3N1

justinvp:
It is possible to do this today with:

```bash
pulumi new -s <org>/<stackname>
```

kevink:
Would it make any sense to put the target org and stackname in the pulumi.yml like the way we can config the default aws region?

ringods:
I raised #3747 a while back regarding working against multiple backends. Maybe add this ticket to the `Related Issues` in the description of this issue.

clstokes:
Related to https://github.com/pulumi/pulumi/issues/4826

PR #8352, opened 2021-11-03
[CLI] Adding the ability to create a default org for backends that support orgs

stack72:
Fixes: #3989

For backends that support orgs (currently, Pulumi SaaS and SaaS self hosted)
we can now attribute a default org to use with that backend.

We have a new command:

```
pulumi org set-default [org name]
```

That will allow any stacks to be created by the user within that org by default

```
$ pulumi org get-default

demo-org
```

```
$ pulumi org

Current Backend: https://api.pulumi.com
Default Org: demo-org
```

We can also specify an org name as part of a login

```
$ pulumi login --org-name demo-org
```

```
pulumi login --org-name demo-org
Manage your Pulumi stacks by logging in.
Run `pulumi login --help` for alternative login options.
Enter your access token from https://app.pulumi.com/account/tokens
    or hit <ENTER> to log in using your browser                   :
We've launched your web browser to complete the login process.

Waiting for login to complete...

Logged in to pulumi.com as stack72 (https://app.pulumi.com/stack72)

~/code/test-pulumi-org-set
▶ cat ~/.pulumi/config.json
{
    "backends": {
        "https://api.pulumi.com": {
            "defaultOrg": "demo-org"
        }
    }
}
```

[asciicast](https://asciinema.org/a/9TQjcB0QGFGzj6MMzdfYUkwDc)


pkg/cmd/pulumi/login.go
```diff
@@ -145,6 +159,8 @@ func newLoginCmd() *cobra.Command {
 	}
 
 	cmd.PersistentFlags().StringVarP(&cloudURL, "cloud-url", "c", "", "A cloud URL to log in to")
+	cmd.PersistentFlags().StringVar(&defaultOrg, "org-name", "", "A default org to associate with the login. "+
+		"Please note, this is only available for Pulumi SaaS and Pulumi Self Hosted backends")
```
lukehoban, line 163: We should probably use the same terminology as our documentation (https://www.pulumi.com/docs/intro/concepts/state/#backends):

```suggestion
		"Please note, this is only available in backends which support organizations, including the hosted and self-service Pulumi Service backends")
```

pkg/cmd/pulumi/new.go
```diff
@@ -170,9 +170,12 @@ func runNew(args newArgs) error {
 	// otherwise `getStack` will fail to detect the project folder and fail.
 	// The main purpose of this lookup is getting a proper start with a project
 	// created via the web app.
+	// TODO - WE NEED TO FIGURE OUT THE ORDER OF OPERATIONS HERE WHEN SETTING DEFAULT ORG!!
+	// HERE!!!!
```
lukehoban, line 174: Is this TODO still needed?

pkg/cmd/pulumi/org.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/pulumi/pulumi/pkg/v3/backend/httpstate"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/util/cmdutil"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/workspace"
+)
+
+func newOrgCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "org",
+		Short: "Manage Organisation configuration",
+		Long: "Manage Organisation configuration.\n" +
```
lukehoban, line 33: I believe we use "Organization" everywhere in our product, and it looks like the paragraph here uses both spellings :-).

pkg/cmd/pulumi/org.go
```diff
@@ -0,0 +1,150 @@
...
+			"\n" +
+			"Use this command to manage any specific organization configuration, " +
```
lukehoban, line 35:
```suggestion
			"Use this command to manage organization configuration, " +
```

pkg/cmd/pulumi/org.go
```diff
@@ -0,0 +1,150 @@
...
+			"e.g. setting the default organization for a backend",
+		Args: cmdutil.NoArgs,
+	}
+
+	cmd.AddCommand(newOrgSetCmd())
+	cmd.AddCommand(newOrgGetCmd())
+
+	return cmd
```
lukehoban, line 43: Currently I believe all our commands actually do something.  Should `pulumi org` do something?  Should it be the same as `pulumi org get`?

Are there other organization-related things we expect to add in the future here?  I can imagine things for configuring teams might end up being subcommands.  If so:
1. Should `pulumi org` produce an overall summary of organization-related settings
2. Should `get`/`set` be `set-default`/`get-default` or similar?

I'm not certain either way on this - but curious how we think we'll evolve this top level command?

pkg/cmd/pulumi/org.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+func newOrgSetCmd() *cobra.Command {
+	var orgName string
+
+	var cmd = &cobra.Command{
+		Use:   "set [NAME]",
+		Args:  cmdutil.MaximumNArgs(1),
+		Short: "Set the default org for the current backend",
```
lukehoban, line 52: We mix "org" and "organization" - in plain english, should we consistently use "organization"?

pkg/cmd/pulumi/org.go
```diff
@@ -0,0 +1,150 @@
...
+		Long: "Set the default org for the current backend.\n" +
+			"\n" +
+			"This command is used to set the default organization in which to create \n" +
+			"projects and stacks for the current backend.\n" +
+			"\n" +
+			"Currently, only the Pulumi SaaS and Pulumi Self Hosted solutions, support setting \n" +
+			"organizations. If you try and set a default organization for a backend that does not \n" +
```
lukehoban, line 59: Similar to notes above - suggest that we (a) frame this in terms of backends supporting organizations, not about supporting this feature (b) note that the organizations that currently support organizations are the managed and self-hosted Pulumi Service backends (using the same terminology as https://www.pulumi.com/docs/intro/concepts/state/#backends).

pkg/cmd/pulumi/org.go
```diff
@@ -0,0 +1,150 @@
...
+
+func newOrgGetCmd() *cobra.Command {
+	var cmd = &cobra.Command{
+		Use:   "get [NAME]",
+		Args:  cmdutil.MaximumNArgs(1),
+		Short: "Get the default org for the current backend",
+		Long: "Get the default org for the current backend.\n" +
+			"\n" +
+			"This command is used to get the default organization for which and stacks are created in " +
+			"the current backend.\n" +
```
lukehoban, line 110: Currently this command returns both the default org and the name of the backend.  Should this return just the default org as a plain string (for scripting scenarios), and allow `pulumi org` to return the richer information this command currently returns?

sdk/go/common/workspace/creds.go
```diff
@@ -251,3 +251,123 @@ func StoreCredentials(creds Credentials) error {
 
 	return nil
 }
+
```
lukehoban, line 254: Can we add test coverage for the functionality here?

pkg/cmd/pulumi/org.go
```diff
@@ -0,0 +1,150 @@
...
+					return err
+				}
+			}
+
+			return nil
+		}),
+	}
+
+	cmd.PersistentFlags().StringVar(&orgName,
+		"org-name", "", "The name of the default organization")
```
lukehoban, line 97: Do we use this pattern of allowing positional or flag-based args anywhere else?  Given that the positional arg takes away the ability to do anything else here anyway, it feels like this isn't really useful/necessary?

pkg/backend/httpstate/backend.go
```diff
@@ -229,6 +229,10 @@ func loginWithBrowser(ctx context.Context, d diag.Sink, cloudURL string, opts di
 	return New(d, cloudURL)
 }
 
+func SetDefaultOrg(url string, orgName string) error {
+	return workspace.SetBackendConfigDefaultOrg(url, orgName)
```
lukehoban, line 233: Should this verify that the org is a real org the user has access to?  (Or at least warn if it is not?)

I'm not certain either way.  It would be useful feedback in case of typos, but may preclude using this command before creating an org?

pkg/cmd/pulumi/org.go
```diff
@@ -0,0 +1,150 @@
+// Copyright 2016-2018, Pulumi Corporation.
```
iwahbe, line 1:
```suggestion
// Copyright 2016-2021, Pulumi Corporation.
```

pkg/cmd/pulumi/org.go
stack72: I like this idea to be able to get a good set of data and then get just return the plain string. I also like the get-default and set-default - it makes it more specific what they do!

pkg/cmd/pulumi/login.go
```diff
@@ -145,6 +159,8 @@ func newLoginCmd() *cobra.Command {
 	}
 
 	cmd.PersistentFlags().StringVarP(&cloudURL, "cloud-url", "c", "", "A cloud URL to log in to")
+	cmd.PersistentFlags().StringVar(&defaultOrg, "org-name", "", "A default org to associate with the login. "+
```
justinvp, line 162: Apologies if this has already been discussed, but I am wondering if this should be named `default-org`?

```
pulumi login --default-org demo-org
```

Or maybe even just `org`?

```
pulumi login --org demo-org
```

sdk/go/common/workspace/creds.go
stack72: I will open an issue to add tests here - we may need to think about how to do that better as this writes a file to disk

stack72:
@lukehoban thank you for the detailed review here - I liked all of the suggestions. I am going to have to open a follow up issue to add the tests here as we need to think about the fact it writes to a file on disk

pkg/cmd/pulumi/login.go
```diff
@@ -145,6 +159,9 @@ func newLoginCmd() *cobra.Command {
...
+		"Please note, this is only available in backends which support organizations, including the hosted and "+
```
justinvp, line 163: Should we error if this option is specified with a backend that doesn't support it?

pkg/go.mod
```diff
@@ -130,8 +132,10 @@ require (
 	github.com/pierrec/lz4 v2.6.0+incompatible // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
 	github.com/rivo/uniseg v0.2.0 // indirect
+	github.com/rogpeppe/go-internal v1.6.1 // indirect
 	github.com/russross/blackfriday/v2 v2.0.1 // indirect
 	github.com/ryanuber/go-glob v1.0.0 // indirect
+	github.com/ryboe/q v1.0.15 // indirect
```
justinvp, line 138: What is causing these changes to the `go.mod`?

pkg/cmd/pulumi/util.go
```diff
@@ -862,3 +862,20 @@ func getRefreshOption(proj *workspace.Project, refresh string) (bool, error) {
...
 }
+
+func buildStackName(stackName string) string {
+	if strings.Count(stackName, "/") == 2 {
+		return stackName
+	}
+
+	defaultOrg, err := workspace.GetBackendConfigDefaultOrg()
+	if err != nil {
+		return ""
```
justinvp, line 873: Should `buildStackName` return the error?

pkg/cmd/pulumi/org.go
```diff
@@ -0,0 +1,155 @@
...
+			}
+
+			orgName = args[0]
+
+			currentBe, err := currentBackend(displayOpts)
+			if err != nil {
+				return err
+			}
+			if !currentBe.SupportsOrganizations() {
+				return fmt.Errorf("Unable to create a default organization for backend type: %s",
```
justinvp, line 92:
```suggestion
				return fmt.Errorf("Unable to set a default organization for backend type: %s",
```

?

pkg/cmd/pulumi/org.go
```diff
@@ -0,0 +1,155 @@
...
+			return nil
+		}),
+	}
+
+	return cmd
+}
+
+func newOrgGetDefaultCmd() *cobra.Command {
+	var cmd = &cobra.Command{
+		Use:   "get-default [NAME]",
```
justinvp, line 115: Does `get-default` accept an argument?

justinvp (approved):
LGTM otherwise.

pkg/cmd/pulumi/login.go
stack72: I like the rename to `default-org` as it's clearer

pkg/go.mod
stack72: Fixed!

pkg/cmd/pulumi/org.go
stack72: It does not -good call!

pkg/cmd/pulumi/org.go
stack72: 🆗

pkg/cmd/pulumi/login.go
stack72: good idea!

stack72:
There is a further issue on way to change the other areas where stacks are used

PR #8359, opened 2021-11-05
Implement the --exclude-protected feature

### LINKED ISSUE #6539: Add `--exclude-protected` flag to `pulumi destroy`
It's generally easy to tear-down and rebuild resources if none of them have to be protected, however with certain resource types, they might be more cumbersome and you want to ensure you don't delete them as part of the process (such as Azure KeyVault).

Working with larger amounts of resources makes it hard to tear-down and rebuild resources when one is protected, since you have to manually paste a huge list of target URNs to the `pulumi destroy` command. It would be better if I could simply say `pulumi destroy --exclude-protected` so that all unprotected resources get destroyed.

It could be implemented in such a way that it either 1) detects and prevents deletion of resources the protected resource is dependent on or 2) it would require you to manually protect those resources as well in order to successfully perform the deletion.  In my case, out of 27 resources in a given stack, I only protect the KeyVault (and the ResourceGroup, since the KeyVault slightly depends on it). 

## Affected feature

Pulumi CLI - destroy command.

iwahbe:
This piggybacks on the same machinery used by the --target flag. By
examining the stack, we find a list of all resources managed by
Pulumi (in that stack). We then form them into a DAG, and mark all
resources as either protected or unprotected.

A resource is protected it has the `Protect` flag set, is has a child
that is protected or has a dependency that is protected. It is unprotected otherwise.

We then pass the urns of unprotected resources to the update options
passed to the destroy operation in the same way that `--target` does.

# Description

Fixes #6539

## Checklist


pgavlin:
> A resource is protected it has the Protect flag set or is has a child
with the protect flag set. It is unprotected otherwise.

Do we also treat resources that are dependencies of protected resources as protected? It feels like we'll need to, as we can't destroy those resources w/o first destroying the protected resources.

iwahbe:
> Do we also treat resources that are dependencies of protected resources as protected? It feels like we'll need to, as we can't destroy those resources w/o first destroying the protected resources.

We currently allow destroying dependencies of protected resources (we don't mark them as protected). My understanding was that dependencies encode only the data necessary to create a resource. I investigated after your comment, and found that this created some errors with invalid snapshots. 
```
error: post-step event returned an error: failed to verify snapshot: resource urn:pulumi:dev::test-protect::aws:s3/bucket:Bucket::my-2bucket dependency urn:pulumi:dev::test-protect::aws:s3/bucket:Bucket::my-bucket refers to missing resource
```
I will rectify this to protect dependencies of protected products.

pkg/cmd/pulumi/destroy.go
```diff
@@ -149,6 +150,34 @@ func newDestroyCmd() *cobra.Command {
...
+					// destroy a protected resource.
+					return nil
+				}
+				for _, r := range unprotected {
+					targetUrns = append(targetUrns, r.URN)
+				}
+			}
+
+			if targets != nil && len(*targets) > 0 && excludeProtected {
+				return result.FromError(errors.New("You cannot specify --target and --exclude-protected"))
```
t0yv0, line 179: Cannot specify it "yet"? Is there a fundamental problem here?

pkg/cmd/pulumi/destroy.go
```diff
@@ -149,6 +150,34 @@ func newDestroyCmd() *cobra.Command {
 			if err != nil {
 				return result.FromError(err)
 			}
+
+			var protectedCount int
+			if excludeProtected {
```
t0yv0, line 155: Separate function here? I get a bit hard to follow. Stylistic preference but I'm a fan of smaller named func.

pkg/cmd/pulumi/destroy.go
iwahbe: `--exclude-protected` means destroy everything that is not protected. `--target` means destroy this specific thing. Any flag specified by `--target` would either be invalid (because it is protected) or would already be targeted by the destroy. 

Because these flag behaviors don't make sense together, I forbid specifying both of them.

t0yv0:
This is awesome! This is an awesome feature. I've just  recently been proofreading some PRs around dependency handling and bugs in dependency handling  and being quite confused. I'd like to work together to make this as explicit as possible.

One PR I'd like to get merged before is:
https://github.com/pulumi/pulumi/pull/8360

Then just like in 8360, I think we can possibly exploit DependencyGraph to encapsulate ordering assumptions as well as topsort and most importantly the definition of what do we mean by dependencies in there. 

DependencyGraph currently is build on this base relation:

C child depends on P=C.Parent unless nil
R resource depends on R.Provider unless nil
R resource depends on explicit dependencies R.Dependencies

The transitive closure of the relation gives the "indirect" dependency relation. "dg.DependenciesOf" and "dg.DependsOn" query these relations.

Presumably here, we could:

- load DependencyGraph on all resources
- find the set of protected resources
- use DependencyGraph to find `protected_and_deps = union([dg.DependingOn(x) for x in protected])`
- compute set difference `to_delete = all_resources - protected_and_deps`
- use DependencyGraph to order to_delete set for correct deletion ordering
- delete it

t0yv0:
```--exclude-protected means destroy everything that is not protected. --target means destroy this specific thing. Any flag specified by --target would either be invalid (because it is protected) or would already be targeted by the destroy.

Because these flag behaviors don't make sense together, I forbid specifying both of them.```

Except --target has  --target-dependents  which presumably uses graph computation that could use restricting - that's what I was thinking, but when you put it this way it's probably not worth it, reasonable to prohibit.

**Comment by iwahbe** [comment]
@t0yv0 I would be happy to work with you to ensure everything is clear. 
> DependencyGraph currently is build on this base relation:
>
> C child depends on P=C.Parent unless nil
> R resource depends on R.Provider unless nil
> R resource depends on explicit dependencies R.Dependencies
> 
> The transitive closure of the relation gives the "indirect" dependency relation. "dg.DependenciesOf" and "dg.DependsOn" > query these relations.

I would like this to be true, but I don't think it is. `dg.DependenciesOf` doesn't handle parents at all. `dg.DependingOn` gives only direct parents (grandparent does not show as depended on). 

> Presumably here, we could:
> 
> load DependencyGraph on all resources
> find the set of protected resources
> use DependencyGraph to find protected_and_deps = union([dg.DependingOn(x) for x in protected])
> compute set difference to_delete = all_resources - protected_and_deps

I think we could do this once #8360 lands. 

> use DependencyGraph to order to_delete set for correct deletion ordering
> delete it

We shouldn't need to do this. The engine should be able to handle this.

File: pkg/cmd/pulumi/destroy.go [hunk 3]
```diff
@@ -257,3 +286,52 @@ func newDestroyCmd() *cobra.Command {
...
+			rProtected[r] = true
+			transitiveProtected.UnionWith(rProtected)
+		}
+	}
+	allResources := graph.NewResourceSetFromArray(resources)
+	return allResources.SetMinus(transitiveProtected).ToArray(), transitiveProtected.ToArray()
+}
+
+// Returns the number of protected resources that remain. Appends all unprotected resources to `targetUrns`.
+func handleExcludeProtected(s backend.Stack, targetUrns *[]resource.URN) (int, error) {
```
t0yv0, line 323: Nit (this is mostly fine, perhaps I have bad FP habits here): This one signature is a bit funky (C-style?) it needs targetUrns to be empty when started, and it populates it when done, so why not  move `*[]resource.URN` to the result of the function rather than a parameter?

pkg/resource/graph/dependency_graph.go
```diff
@@ -123,6 +123,67 @@ func (dg *DependencyGraph) DependenciesOf(res *resource.State) ResourceSet {
 	return set
 }
 
+// `TransitiveDependenciesOf` calculates the set of resources that `r` depends
```
t0yv0, line 126: It's awesome that this model passes, consider checking it into the "rapid" file: 

```
func TestRapidTransitiveDependenciesOf(t *testing.T) {
	graphCheck(t, func(t *rapid.T, universe []*resource.State) {
		expectedInTDepsOf := transitively(universe)(expectedDependenciesOf)
		dg := NewDependencyGraph(universe)
		for _, a := range universe {
			tda := dg.TransitiveDependenciesOf(a)
			for _, b := range universe {
				assert.Equalf(t,
					expectedInTDepsOf(a, b),
					tda[b],
					"Mismatch on a=%v, b=%b",
					a.URN,
					b.URN)
			}
		}
	})
}
```

The rapid file does not yet test providers so if there's any subtleties with providers won't get caught yet but it will afterward when we extend the random generator.

pkg/cmd/pulumi/destroy.go
```diff
@@ -257,3 +286,52 @@ func newDestroyCmd() *cobra.Command {
...
+// We rely on the fact that `resources` is topologically sorted with respect to
+// its dependencies. This function understands that providers live outside this
+// topological sort.
+func seperateProtected(resources []*resource.State) (
+	/*unprotected*/ []*resource.State /*protected*/, []*resource.State) {
+	dg := graph.NewDependencyGraph(resources)
+	transitiveProtected := graph.ResourceSet{}
+	for _, r := range resources {
+		if r.Protect {
+			rProtected := dg.TransitiveDependenciesOf(r)
```
t0yv0, line 313: So thinking out loud here. Are you sure we're specifying what we want here. The code is traversing TransitiveDependenciesOf and considers them protected, which is the inverse dependency order from DependingOn. So in practice given your own example:

// A
// B: Parent = A
// C: Parent = A, Protect = True
//
// -->
//
// Unprotected: B
// Protected: A, C

We're extending protection to parent of C, so A and C stick around, good.

However we're not extending protection to the children, like this:

// A
// B: Parent = A
// C: Parent = A, Protect = True
// D: Parent = C
//
// -->
//
// Unprotected: B, D
// Protected: A, C

So presumably we'd nuke C but leave D hanging with a deleted parent? is this right?

Actually yeah it seems right. Awesome.

t0yv0 (approved):
LGTM!

pkg/cmd/pulumi/destroy.go
```diff
@@ -257,3 +286,52 @@ func newDestroyCmd() *cobra.Command {
...
+//
+// -->
+//
+// Unprotected: B
+// Protected: A, C
+//
+// We rely on the fact that `resources` is topologically sorted with respect to
+// its dependencies. This function understands that providers live outside this
+// topological sort.
+func seperateProtected(resources []*resource.State) (
```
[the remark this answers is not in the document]
iwahbe: I love the idea here. I think it sufficiently local that I will keep it as is. I will definitely use this for other PRs though.

Issue #6283, opened 2021-02-08
pulumi destroy with --target errors and leaves stack in unworkeable state


thrixton:
Targeting a resource (with dependent resources) for destroy errors and leaves the stack in an unworkeable state
## Example
Given the resource tree below
Deployment ==> Service
Service ==> Ingress

If I do a `pulumi destroy -t (urn of Deployment)`, we get the error below
```
error: post-step event returned an error: failed to verify snapshot: resource urn:pulumi:<snip>-ingress dependency urn:pulumi:<snip>worker refers to missing resource
```
When I do a `pulumi stack export --file stack.json` and look at the resulting file, I can see that there is a resource (my ingress) which still refers to my destroyed resource.
E.g. From the example above, Ingress still exists in the stack and refers to Deployment in the `dependencies` and `propertyDependencies.spec` elements.

I then remove the lines from my `stack.json` and do a `pulumi stack import --file stack.json` and it then works as per expected.

## Expected Behavior
Either dependent resources are destroyed (Service and Ingress)  
or
Dependent resources are left in place in a pending state and references are cleaned up in the stack

## Context (Environment)
"@pulumi/pulumi": "^2.20.0"
"@pulumi/kubernetes": "^2.8.0"

> NOTE: This may be an issue with the kubernetes provider rather than pulumi as it's not strictly necessary to remove dependent resources as their requirements may be satisfied at a future point.

alex-hunt-materialize:
This also happens when passing `--target-dependents` during the destroy.

The expected behavior should probably be:

1. For `pulumi destroy --target $SOME_URN`, it should abort immediately if any dependents exist.
2. For `pulumi destroy --target $SOME_URN --target-dependents`, it should destroy the dependents first, followed by the target resource, and it should never corrupt the stack.

alex-hunt-materialize:
Additionally, it seems to remove items from the state which still exist in reality. It crashed with:
```
error: post-step event returned an error: failed to verify snapshot: resource urn:pulumi:alexhunt::mzcloud::kubernetes:batch/v1beta1:CronJob::frontegg-sync-job dependency urn:pulumi:alexhunt::mzcloud::kubernetes:core/v1:Secret::mzcloud-alexhunt refers to missing resource
```
and the RDS instance which both of those depend on is gone from the state. The RDS instance itself still exists in AWS.

alex-hunt-materialize:
This may also be related to https://github.com/pulumi/pulumi/issues/3878

[...]

PR #8360, opened 2021-11-05
Fix issue with --target deletion dependent calculation

joeduffy:
The code that computed `--target` deletion dependents was not correct. It only used parent/child component relationships, but did not respect actual DAG dependencies (which are more common). As a result, it could erroneously leave hanging references to resources that no longer exist after performing a `pulumi destroy --target X` operation. This manifested in bugs like https://github.com/pulumi/pulumi/issues/6283, which is fixed by this change. The solution is to compute the (transitive!) dependency graph correctly, factoring in both parent/child, as well as explicit and implicit, dependencies. The existing logic does the correct thing once we do this. I've also added tests for this area, including regression tests that cover transitive dependency relationships, as well as ones that would cause an infinite loop given a naive implementation.

This came up during a customer call today and I was genuinely curious, and after seeing it in action, the fix seemed fairly self-evident to me, so I rolled with it. Happy to leave it to someone else to fix if my approach is not desirable.


t0yv0:
This is awesome, thank you, will work through this shortly.

CC @emiliza 

Tangentially related - we have a stalled thread in https://github.com/pulumi/pulumi/pull/7852 that was working through the implications of dependency edges via Component resources. I think this is not in play here, but there's the `--target` parameter in play instead that I was not aware of. I'd like to model-check this functionality also (in addition to unit tests) just as https://github.com/pulumi/pulumi/pull/7859 was fruitful in finding bugs in 7852. 

I'd push to invest a bit of time here to get the dependency story rock-solid.

pgavlin:
>  It only used parent/child component relationships, but did not respect actual DAG dependencies

😬 

> The solution is to compute the (transitive!) dependency graph correctly, factoring in both parent/child, as well as explicit and implicit, dependencies

Would we be able to use the preexisting [`DependencyGraph`](https://pkg.go.dev/github.com/pulumi/pulumi/pkg/v3@v3.17.0/resource/graph#DependencyGraph) code to compute this?

joeduffy:
TIL `DependencyGraph`! @pgavlin I tried going that route, thinking I could use [`DependingOn`](https://github.com/pulumi/pulumi/blob/pkg/v3.17.0/pkg/resource/graph/dependency_graph.go#L23) -- since we are looking for dependents, not dependencies -- but that doesn't seem to include parent/child logic the same way that [`DependenciesOf` does](https://github.com/pulumi/pulumi/blob/7e2c74d58c989db04c8b877e9f4a2f7361c3b461/pkg/resource/graph/dependency_graph.go#L113-L116). I may very well be out of my league on this one; do you know if this is intentional?

t0yv0:
Consider https://github.com/pulumi/pulumi/pull/8399

pkg/resource/graph/dependency_graph.go
```diff
@@ -20,7 +20,8 @@ type DependencyGraph struct {
 // order with respect to the snapshot dependency graph.
 //
 // The time complexity of DependingOn is linear with respect to the number of resources.
-func (dg *DependencyGraph) DependingOn(res *resource.State, ignore map[resource.URN]bool) []*resource.State {
+func (dg *DependencyGraph) DependingOn(res *resource.State,
+	ignore map[resource.URN]bool, includeChildren bool) []*resource.State {
```
t0yv0, line 24: I'm wondering whether this should be `includeChildren=true` always.  It's great to be conservative with the change here so we minimize blast radius, but we should follow up to figure this out what the implications here are..

joeduffy: That was my initial instinct too, but was trying to reduce risk with the change. For what it’s worth, I tried running the unit tests with this always set to true and, somewhat terrifyingly, they passed. @pgavlin @justinvp What is the intended behavior here?

iwahbe:
>  but that doesn't seem to include parent/child logic the same way that [`DependenciesOf` does](https://github.com/pulumi/pulumi/blob/7e2c74d58c989db04c8b877e9f4a2f7361c3b461/pkg/resource/graph/dependency_graph.go#L113-L116). I may very well be out of my league on this one; do you know if this is intentional?

[`DependenciesOf`](https://github.com/pulumi/pulumi/blob/7e2c74d58c989db04c8b877e9f4a2f7361c3b461/pkg/resource/graph/dependency_graph.go#L113-L116) only handles the immediate parent, even though the rest of the function is for transitive dependencies. I would be very curious why this does not handle transitive parents.

t0yv0:
I'd love to merge it with https://github.com/pulumi/pulumi/pull/8399 addition which is relevant to the bug here. Transitivity seems fairly relevant to the bug here, if parent resource is included in --target then why not grandparent? But does the addendum sound too controversial?

Let me quickly check again about merging as-is.

@iwahbe thanks for you review on https://github.com/pulumi/pulumi/pull/8404 - let's finish this out and when Pat is back cross-check it so we can resolve the questions around DependenciesOf intuition.

t0yv0:
Ok I've model checked the code in this PR against this:

```
// Model `DependingOn` as the inverse transitive closure of `dependenciesOf`.
func expectedDependingOn(universe []*resource.State, includeChildren bool) R {
	if includeChildren {

		dependingOn := inverse(transitively(universe)(
			union(hasProvider, hasDependency)))

		return func(a, b *resource.State) bool {

			if dependingOn(a, b) {
				return true
			}

			if isParent(b, a) {
				return true
			}

			for _, x := range universe {
				if dependingOn(x, b) && isParent(x, a) {
					return true
				}
			}

			return false
		}

	} else {
		return inverse(transitively(universe)(union(hasProvider, hasDependency)))
	}
}

```

In prolog notation easiest to read:

```

  dependingOnIncludingChildren(a, b) -: dependingOn(a, b).
  dependingOnIncludingChildren(a, b) -: isParent(b, a).  
  dependingOnIncludingChildren(a, b) -: dependingOn(x, b), isParent(x, a).  

```

It doesn't go into grand parents but it makes some sense, no harm in this. Approving.

t0yv0 (approved):

joeduffy:
Thanks @t0yv0, fully agree with the potential broader point you’re making. The code has to compute the transitive set because the DependencyGraph wasn’t doing it (I found out because the original fix did this and when I switched to DG, it did not pass the new tests). Thanks for merging as is, and definitely a fan of doing the deeper fixing/refactoring you suggest.

t0yv0:
Would you like a point release for this @joeduffy or should this wait for the next cycle? e.g. how urgent is the bug?

Pat should be back on Monday.

t0yv0:
I think I see what you're saying, I missed that, since the code does a transitive set over DependingOn, the user will not observe this difference! My tests are at the wrong level so to say. Thanks for that point.

joeduffy:
It’s ok to wait for the next release. Thank you for asking!

PR #8383, opened 2021-11-09
Move gopls settings to .vscode from .devcontainer

Frassle:
These gopls settings should apply to any vscode instance, not just ones in devcontainers.


danielrbradley (approved):
This makes sense. Assume the JSON comments are still fine in this context?

Frassle:
Yeh vscode doesn't have any issue with comments in setting files.

https://code.visualstudio.com/docs/languages/json#_json-with-comments

PR #8384, opened 2021-11-09
Improve vscode dotnet support

Frassle:
# Description

Improve support for dotnet (mostly) in vscode and the devcontainer.

This installs a test adapter for dotnet in the devcontainer so that our C# tests show in the vscode test explorer.
This also sets some default settings for onmisharp and the test-adapter to focus on the main sdk .sln.
Finally this changes the devcontainer startup to call "make ensure" rather than "dotnet restore" on every csproj. The main dotnet projects are now restored as part of "make ensure". The sum result of this is that once the devcontainer is started up there are working environments for all the languages (dotnet has projects restored, go has deps downloaded, python has a venv setup etc).

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -18,8 +18,7 @@
 
     "extensions": ["golang.go", "ms-dotnettools.csharp", "ms-python.python"],
 
-    // We want to dotnet restore all projects on startup so that omnisharp doesn't complain about lots of missing types on startup.
-    "postCreateCommand": "find -name \"*.??proj\" | xargs -L1 dotnet restore",
+    "postCreateCommand": "make ensure",
 
     "settings": {
         "extensions.ignoreRecommendations": true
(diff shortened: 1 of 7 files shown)


Frassle:
Hmm dotnet restore seems sad at /opt/pulumi/nuget not existing on the first restore. Couple of choices
A) Fix up the workflow files to ensure that's created before restore (I've done a few of these, but there are some that don't use the env var that I've missed in https://github.com/pulumi/pulumi/pull/8384/commits/93337c78a74d4f2929497dc8d534951919932ef9)
B) Can we just get rid of PULUMI_LOCAL_NUGET, it doesn't seem to be needed when running on workstations not clear why it would be needed in CI.

t0yv0 (approved):
Looks pretty harmless unfortunately I'm not a VS code user yet so not sure how to verify easily. I say let's check this in and if there's any problems we can follow up or back out.

Apologies for the very late review here!

PR #8389, opened 2021-11-09
[codegen/docs] Use go:embed for templates and use DisplayName from schema for titles

praneetloke:
# Description

* Now that the schema supports a `DisplayName` property in the package spec, the docs generator can use that when building docs if it is not an empty string. Still falling back to the lookup map.
* Add examples of tags that can be used in `Keywords`. Related to https://github.com/pulumi/pulumi/issues/7813#issuecomment-964505013
* I have also removed the need to generate the git-ignored templates bundle file (It was my doing over a year ago :)). Basically, doing so will fix the issue where sometimes people update a template and forget to run `make generate`. It also allows us to drop the local override in `docs` and pin the `pulumi/pulumi` version and therefore not need to clone this repo in order just so that one could API docs.

## Checklist

  - The change to not generate the templates bundle file is already covered by several docs generator tests. I do see those passing on my local machine (despite removing the templates bundle file, of course!)
  - Code changes are limited to the docs generator


.github/workflows/run-docs-generation.yml
```diff
@@ -93,6 +93,12 @@ jobs:
           # If generating docs for more providers here, be sure to update
           # the description of the draft PR that is opened in the next step.
           pushd docs
+
+          pushd tools/resourcedocsgen
+          go mod edit -replace github.com/pulumi/pulumi/pkg/v3=../../../pulumi/pkg
```
praneetloke, line 98: I did this because I plan to pin the versions used by the `resourcedocsgen` tool in `docs`. For the sake of this workflow, it does need the local override so that the templates used by it are what's in the PR current branch always.

praneetloke:
Hmm now I remember why I bundled the templates in the first place. Let me think about it a bit more before I mark this as ready for review.

praneetloke:
Alright this is ready for review. tl;dr; used `go:embed`.

When `bundler.go` was first authored over a year ago we needed a way to embed the packaged templates such that alternate entrypoints into the docs generator can still work. There were a few ways to embed static files and most of them involved embedding a byte-array serialized version of the templates. We could have done that using one of the many third-party libraries but we also didn't need most of the functionalities those libraries offered. Most of them were geared towards running a static file server. We just needed a file with the byte-arrays so a rudimentary bundler was implemented which worked ok. But this is preventing us from offering a nicer dev/contribution experience in our docs ecosystem. `go:embed` plays nicely with `text/template` in that it works with a virtual FS out-of-box.

praneetloke:
The failed PR jobs seem unrelated to the changes in this PR. The one that looks suspicious to me are all the `codegen/schema` failures in this job https://github.com/pulumi/pulumi/runs/4160301528?check_suite_focus=true#step:20:28425. Not sure what's going on there. Should I just re-run all failed jobs to clear that up?

justinvp (approved):
LGTM. Thanks for updating this to use `go:embed`!

praneetloke:
@justinvp thank you for approving! Should I rerun all failed jobs in this PR?

justinvp:
I'm re-running the jobs. If we get the same failures we'll have to look more closely.

praneetloke:
All jobs passed! I'll merge this.

Issue #8186, opened 2021-10-11
[auto/*] - Avoid mutating global state when creating or selecting stacks


komalali:
## Hello!

- Vote on this issue by adding a 👍 reaction
- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details

Currently, automation API's `createStack`, `selectStack` and variants of the functionality also mutates global workspace state by setting the created or selected stack as "current". We did [work](https://github.com/pulumi/pulumi/issues/5257) for the 3.0 launch to start using the `--stack` flag in most of the commands implemented via automation API, with the exception of `create` and `select`.

In practice, this means that the functions `createStack`, `selectStack` and `createOrSelectStack` cannot be run in parallel, as each invocation will overwrite the `workspace.json` file to set the stack as current and result in a [corrupted json file](https://github.com/pulumi/pulumi/issues/6287).

## Solution

One possible solution is prototyped in https://github.com/pulumi/pulumi/compare/komalali/auto-stack-select?expand=1

The solution involves adding a `setCurrent` flag to `pulumi stack init` which defaults to True. This flag is then piped through to automation api. If `setCurrent` is False, there is no mutation of the workspace json file and the `createStack`, `selectStack` and `createOrSelectStack` functions can be run in parallel with no issues.

seunggs:
Hi is there any traction on this? I'm building an app that manages infrastructure for clients and without parallel stack select/create, my app will error out if two clients sign up around the same time. This is particularly troublesome since provisioning an eks cluster, for example, the process can be running for more than half an hour.

seunggs:
Any update on this?

mussa572:
Any update on this , I am facing the similar issue

Moshemo:
Hello,
Any updates on this issue ?

zbuchheit:
I was running into an issue with this. My solution was to make a directory per stack, which in turn created a workspaces file per stack fixing the global state issue for me.

PR #8393, opened 2021-11-09
Improve corrupt workspace settings experience

### LINKED ISSUE #6287: Confusing error if `~/.pulumi/workspaces` file is corrupted
Pulumi uses `~/.pulumi/workspaces` to keep track of what stack is currently selected when running `up`, `pre`, etc. If the current project's workspace gets corrupted somehow, then the user will be presented with a confusing error message and no clear path to fix their situation

## Expected Behavior

If a workspace file is corrupt, then `pulumi` should provide a better error message to the user about which file is corrupt and how it's used. 

Alternatively, if the workspace file is _only_ used for determining the currently selected file, then the user should be presented with the same option as `pulumi stack select` which fixes the corrupt file.

## Current Behavior

If a project's workspace file is corrupt, the user is presented with:

```
aws-ts-webserver % pulumi up
error: invalid character 'a' looking for beginning of value
```

## Steps to Reproduce

1. use an existing Pulumi project or create a new one
1. `pulumi stack init my-corrupt-stack-file`
2. `ls -ltr ~/.pulumi/workspaces`
    - copy the most recently updated file
3. `echo 'blah' >> filename-copied-from-previous-step.json`
4. `pulumi up`
    - observe: error shown above

## Context (Environment)

```
% pulumi version
v2.20.0
```

iwahbe:
This improvement comes in two parts.

1. The error message for a corrupt workspace settings file now clearly
indicates both the file and the parsing error.

2. Writing the workspace settings is now atomic. This prevents
corruption from multiple concurrent calls.

# Description

Fixes #6287

This change also lessons the impact of #8186, though it is still something we need to address. 

## Checklist

### DIFF
--- a/sdk/go/common/workspace/workspace.go
+++ b/sdk/go/common/workspace/workspace.go
@@ -19,6 +19,7 @@ import (
 	"crypto/sha1"
 	"encoding/hex"
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
@@ -105,7 +106,7 @@ func NewFrom(dir string) (W, error) {
 
 	err = w.readSettings()
 	if err != nil {
-		return nil, errors.Wrap(err, "unable to read workspace settings")
+		return nil, fmt.Errorf("unable to read workspace settings: %w", err)
 	}
 
 	upsertIntoCache(dir, w)
--- a/sdk/go/common/workspace/workspace.go
+++ b/sdk/go/common/workspace/workspace.go
@@ -25,7 +25,6 @@ import (
 	"strings"
 	"sync"
 
-	"github.com/docker/docker/pkg/ioutils"
 	"github.com/pkg/errors"
 
 	"github.com/pulumi/pulumi/sdk/v3/go/common/tokens"
@@ -140,7 +139,30 @@ func (pw *projectWorkspace) Save() error {
 	if err != nil {
 		return err
 	}
-	return ioutils.AtomicWriteFile(settingsFile, b, 0600)
+	return atomicWriteFile(settingsFile, b)
+}
+
+// atomicWriteFile provides a rename based atomic write through a temporary file.
+func atomicWriteFile(path string, b []byte) error {
+	tmp, err := ioutil.TempFile("", filepath.Base(path))
+	if err != nil {
+		return errors.Wrapf(err, "failed to create temporary file %s", path)
+	}
+	defer func() { contract.Ignore(os.Remove(tmp.Name())) }()
+
+	if err = tmp.Chmod(0600); err != nil {
+		return errors.Wrap(err, "failed to set temporary file permission")
+	}
+	if _, err = tmp.Write(b); err != nil {
+		return errors.Wrap(err, "failed to write to temporary file")
+	}
+	if err = tmp.Sync(); err != nil {
+		return err
+	}
+	if err = tmp.Close(); err != nil {
+		return err
+	}
+	return os.Rename(tmp.Name(), path)
 }
 
 func (pw *projectWorkspace) readSettings() error {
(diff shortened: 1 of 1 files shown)


sdk/go/common/workspace/workspace.go
```diff
@@ -139,8 +139,30 @@ func (pw *projectWorkspace) Save() error {
...
+	if _, err = tmp.Write(b); err != nil {
+		return errors.Wrap(err, "failed to write to temporary file")
+	}
+	if err = tmp.Sync(); err != nil {
+		return err
+	}
+	if err = tmp.Close(); err != nil {
+		return err
+	}
+	return os.Rename(tmp.Name(), path)
```
mjudeikis, line 165: `os.Rename` works only if you use the same filesystem. in the OS'es where tmpfs is used and you trying to move somewhere else - this should fail.  
https://stackoverflow.com/questions/42392600/oserror-errno-18-invalid-cross-device-link
https://github.com/jupyterlab/jupyterlab/issues/5781

In the example for fedora:
```
 df -T /tmp/go-build* | awk '{print $2}' | tail -n1
tmpfs
df -T /usr/local/bin/stuff | awk '{print $2}' | tail -n1                                                                                                                                                                                                                                         
btrfs
```
In this scenario, this should fail so you might need to do one more sync from one file to other or check underlays fs types (atomic file writes are not so easy :/ ) 

p.s sorry if not the place and not the case, just came across this random PR on the internet :)

t0yv0:
@iwahbe so this looks pretty cool but have you tested this on Windows?

This resonates with a very similar PR I tried to undertake. Note that it has a test that we might want to run here as well. 

https://github.com/pulumi/pulumi/pull/7739

The problem with that PR is that I dug up a discussion from people who spent a good bit of time on this problem, concluding that atomic file writes are not possible on windows:

https://github.com/google/renameio/pull/20

That being said we might merge this if it works "most of the time", I mean it can still be an improvement without being fully atomic on Windows.

Discussion at https://github.com/golang/go/issues/8914 is also gold...

I wonder what's the right solution on Windows?

t0yv0:
We also seem to be using this file locking lib in Pulumi already: https://github.com/gofrs/flock/blob/master/flock_winapi.go

t0yv0:
Copying from team discussion: believe os.Rename to be atomic in the same volume on Windows, hence `ioutil.TempFile(filepath.Dir(path), filepath.Base(path))` is trying to use the dir of the destination to stay on the same volume.

t0yv0 (approved):

sdk/go/common/workspace/workspace.go
```diff
@@ -105,7 +105,7 @@ func NewFrom(dir string) (W, error) {
 
 	err = w.readSettings()
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, "unable to read workspace settings")
```
t0yv0, line 108: `fmt.Fprintf("unable to read workspace settings: %w", err)` I thought maybe is the more modern form (nit)

t0yv0:
Not to discount the statements on Windows support from google/renameio project, but reading the fine print it seems that the WIndows support does not offer the atomic guarantees under high load. The scenarios in Pulumi application will not involve high contention. At any rate the PR improves the status QUO.

sdk/go/common/workspace/workspace.go
justinvp: Aside: I realize it's a low-value return, but at some point I'd love to remove our dependency on `github.com/pkg/errors` and move to the more modern built-in equivalents throughout our codebase. Maybe when we have a little downtime near the holidays we can look into it. Seems like there are tools that can help with this: https://xdg.me/rewriting-go-with-ast-transformation/

PR #8395, opened 2021-11-10
Update command respects `--target-dependents`

### LINKED ISSUE #5869: pulumi up --target doesn't support create dependent targets
## Expected Behavior

When run `pulumi up -t urn --target-dependents`, it should create all dependent targets.

## Current Behavior

```
Previewing update (prod)

View Live: https://app.pulumi.com/xdays/foobar/prod/previews/11111111-7ea8-4b7a-affc-c6f55c4a33fc

     Type                              Name             Plan       Info
     pulumi:pulumi:Stack               foobar-prod                 1 error
     └─ foobar:aws:VPC                 vpc-prod
        └─ aws:ec2:Vpc                 vpc-prod-vpc
 +         └─ aws:ec2:InternetGateway  vpc-prod-vpc-ig  create     1 error

Diagnostics:
  aws:ec2:InternetGateway (vpc-prod-vpc-ig):
    error: Resource 'urn:pulumi:prod::foobar::foobar:aws:VPC$aws:ec2/vpc:Vpc$aws:ec2/internetGateway:InternetGateway::vpc-prod-vpc-ig' depends on 'urn:pulumi:prod::foobar::foobar:aws:VPC$aws:ec2/vpc:Vpc::vpc-prod-vpc' which was was not specified in --target list.

  pulumi:pulumi:Stack (foobar-prod):
    error: preview failed

```

### LINKED ISSUE #8209: Target dependencies doesn't update dependents
## Hello!

## Issue details

When running `pulumi up --target *urn* --target-dependents`, the dependents are not being updated.

### Steps to reproduce

1. Clone this repo: https://github.com/pierskarsenbarg/target-test
2. Follow instructions in [README.md](https://github.com/pierskarsenbarg/target-test/blob/main/README.md)
3. Uncomment tag lines in both `Bucket` and `BucketObject` ([these lines](https://github.com/pierskarsenbarg/target-test/blob/main/index.ts#L7-L11) and [these lines](https://github.com/pierskarsenbarg/target-test/blob/main/index.ts#L17-L19))
4. Run `pulumi up --target urn:pulumi:dev::target-test::aws:s3/bucket:Bucket::bucket --target-dependents` and see that the preview only mentions the bucket resource. 
5. Select no (i.e. don't deploy changes)
6. Re-comment [these lines](https://github.com/pierskarsenbarg/target-test/blob/main/index.ts#L7-L11) and [these lines](https://github.com/pierskarsenbarg/target-test/blob/main/index.ts#L17-L19) and uncomment [this line](https://github.com/pierskarsenbarg/target-test/blob/main/index.ts#L21) (so mark the bucket as a parent of the bucket object)
7. Run `pulumi up`
8. Uncomment the tag lines again ([these lines](https://github.com/pierskarsenbarg/target-test/blob/main/index.ts#L7-L11) and [these lines](https://github.com/pierskarsenbarg/target-test/blob/main/index.ts#L17-L19))
9. Run `pulumi up --target urn:pulumi:dev::target-test::aws:s3/bucket:Bucket::bucket --target-dependents` and see in the preview that only the bucket is marked for update

Expected: 

I would expect that the preview would show both the bucket and the bucketobject resources are marked for update

Actual: 
On the bucket resource is marked for update

iwahbe:
This change enables `pulumi up -t ${URN} --target-dependents`. Previously `--target-dependents` was ignored for non destroy operations. 

# Description

Fixes #8209
Fixes #5869

## Checklist

   Tests are located in `pkg/engine/lifeycletest/target_test.go` as part of `TestUpdateTarget`. 


pkg/resource/deploy/step_generator.go
```diff
@@ -72,10 +72,34 @@ func (sg *stepGenerator) isTargetedUpdate() bool {
...
 }
 
+// isTargetedForUpdateOrAddIfDep checks if a target should be updated, updating
+// `sg.isTargetedForUpdateOrAddIfDep` as necessary for `--target-dependents` to be
+// accurate.
+func (sg *stepGenerator) isTargetedForUpdateOrAddIfDep(urn resource.URN, dependencies []resource.URN) bool {
+	if sg.isTargetedForUpdate(urn) {
+		return true
+	}
+	// TODO; does sg.opts.TrustDependencies play a role here?
```
iwahbe, line 90: I think `TrustDependencies` is about parallelization, but I'm not sure. I don't think we need to check `TrustDependencies`.

pkg/resource/deploy/step_generator.go
```diff
@@ -440,7 +464,7 @@ func (sg *stepGenerator) generateSteps(event RegisterResourceEvent) ([]Step, res
 
 		// If the user requested only specific resources to update, and this resource was not in
 		// that set, then do nothin but create a SameStep for it.
-		if !sg.isTargetedForUpdate(urn) {
+		if !sg.isTargetedForUpdateOrAddIfDep(urn, goal.Dependencies) {
```
t0yv0, line 467: Wondering what's goal.Dependencies here in the context, immediate or transitive dependencies of resource identified by urn, or something else? How does it relate to ResourceGraph?

Also I have a premonition that the engine on the update path observes the dependency graph incrementally; does it observe enough of it to do a good job here? Probably yes but good to think through.

t0yv0:
High-level tests for the feature please, could also explore if this obeys transitive deps in those tests...

I do realize you may be seeking some preliminary feedback on the approach from Pat before doing the tests, that's great, can wait. Just given my level of familiarity tests do help reviewing a lot. Thanks.

pkg/resource/deploy/step_generator.go
iwahbe: I'm pretty sure that the dependency graph is observed incrementally, but in topologically sorted order. At least, that is the assumption I am working under. It's fine if `goal.Dependencies` is local (not transitive). Transitive parents will already have been registered in previous calls.

pgavlin: > I'm pretty sure that the dependency graph is observed incrementally, but in topologically sorted order

This is correct.

> It's fine if goal.Dependencies is local (not transitive). Transitive parents will already have been registered in previous calls.

`goal.Dependencies` is indeed non-transitive. Anything that a resource R depends on--whether the edge is a dependency edge, a parent edge, or a provider edge--must be registered prior to R itself.

pkg/resource/deploy/step_generator.go
pgavlin: `TrustDependents` is about ordering--basically, we used to have Python SDKs that didn't report correct dependency info, so we used that flag to work around that. You shouldn't need to check it here, since this is more of a structural thing.

pkg/resource/deploy/step_generator.go
```diff
@@ -72,10 +72,34 @@ func (sg *stepGenerator) isTargetedUpdate() bool {
...
+	// TODO; does sg.opts.TrustDependencies play a role here?
+	if !sg.opts.TargetDependents || dependencies == nil {
+		return false
+	}
+	for _, dep := range dependencies {
+		if sg.updateTargetsOpt[dep] {
+			sg.updateTargetsOpt[urn] = true
+			return true
+		}
+	}
```
pgavlin, line 99: You should check the dependency graph implementation used by `--target-dependents` in the delete case and see if it considers a resource's children and a provider's associated resources to be dependents (I _think_ it does, but I'm uncertain). If it does, then this code should also check the target set for the resource's parent and provider.

iwahbe: Good catch. The destroy version doesn't work that way. It instead runs the normal target destroy step generator. If it catches a resource to destroy not targeted and `--target-dependents` is not set, it errors. 

I need to check for parents and providers for the behavior to work the same way.

pkg/resource/deploy/step_generator.go
```diff
@@ -72,10 +72,34 @@ func (sg *stepGenerator) isTargetedUpdate() bool {
...
+// Note: to accommodate the `--target-dependents` flag,
+// `isTargetedForUpdateOrAddIfDep` should probably be called instead.
 func (sg *stepGenerator) isTargetedForUpdate(urn resource.URN) bool {
 	return sg.updateTargetsOpt == nil || sg.updateTargetsOpt[urn]
 }
 
+// isTargetedForUpdateOrAddIfDep checks if a target should be updated, updating
+// `sg.isTargetedForUpdateOrAddIfDep` as necessary for `--target-dependents` to be
+// accurate.
+func (sg *stepGenerator) isTargetedForUpdateOrAddIfDep(urn resource.URN, dependencies []resource.URN) bool {
```
t0yv0, line 86: Nit: stylistically I would perhaps break it into two functions if it does have to mutate `sg.updateTargetsOpt`. It "looks" like a predicate because it returns bool, but it does mutate. I read the comment, then read the implementation, then had to read all the sites that touch `updateTargetsOpt` to figure out what it does. It seems like it does a sensible thing after all. 

Perhaps something like this?

```go
if sg.isAnyTargetedForUpdate(dependencies) {
   sg.targetForUpdate(urn) // more obviously mutating
}
sg.isTargetedForUpdate(urn)
```

iwahbe: I spent way too long trying to come up with a name that made sense for this function. I was hoping it would read `(isTargetedForUpdate)Or(AddIfDep)`. The problem is that I have two separate ideas (`isTargetedForUpdate` and `addIfDep`) that I want to always be called together. I forced this by putting them in one function, but then there are naming problems. I can break these out, I'm just worried that someone (or me later) will forget to call one. Thanks for the feedback.

t0yv0: Eh, don't sweat it it's a really small thing. Code will get rewritten... ultimately our defense is code reviews and thoughtful tests that remind when something is off.

pkg/resource/deploy/step_generator.go
```diff
@@ -72,8 +72,51 @@ func (sg *stepGenerator) isTargetedUpdate() bool {
...
+// returns true.
+func (sg *stepGenerator) isTargetedForUpdate(
+	urn, parent resource.URN, provider string, dependencies []resource.URN) bool {
+	if sg.updateTargetsOpt == nil || sg.updateTargetsOpt[urn] {
+		return true
+	}
+	if provider != "" {
+		res, err := providers.ParseReference(provider)
+		contract.AssertNoError(err)
+		dependencies = append(dependencies, res.URN())
```
pgavlin, line 86: I would avoid modifying the input slice like this. It shouldn't be visible to callers--and won't be if the slice gets copied during the `append`--but instead, I'd rather we manually check `sg.updateTargetsOpt` for the URN.

pkg/resource/deploy/step_generator.go
```diff
@@ -72,8 +72,51 @@ func (sg *stepGenerator) isTargetedUpdate() bool {
 	return sg.updateTargetsOpt != nil || sg.replaceTargetsOpt != nil
 }
 
-func (sg *stepGenerator) isTargetedForUpdate(urn resource.URN) bool {
-	return sg.updateTargetsOpt == nil || sg.updateTargetsOpt[urn]
+// isTargetedForUpdate returns if urn is target for update. The function accommodates
+// `--target-dependents`. `targetDependentsForUpdate` should probably be called if this function
+// returns true.
+func (sg *stepGenerator) isTargetedForUpdate(
```
pgavlin, line 78: Offhand it looks like these changes always use the `--target-dependents` behavior--I would have expected us to terminate early if `sg.opts.TargetDependents` is false.

pkg/resource/deploy/step_generator.go
```diff
@@ -72,8 +72,51 @@ func (sg *stepGenerator) isTargetedUpdate() bool {
...
+		if dep != "" && sg.updateTargetsOpt[dep] {
+			return true
+		}
+	}
+	return false
+}
+
+// targetDependentsForUpdate sets the dependents of urn to be updated, if performing a targeted
+// update.
+func (sg *stepGenerator) targetDependentsForUpdate(
```
pgavlin, line 101: I don't think we need this function. This is adding all of a resource's dependencies to the target set; I think that we just need to add the resource itself.

pkg/resource/deploy/step_generator.go
```diff
@@ -490,9 +534,11 @@ func (sg *stepGenerator) generateSteps(event RegisterResourceEvent) ([]Step, res
 	// We will also not record this non-created resource into the checkpoint as it doesn't actually
 	// exist.
 
-	if !sg.isTargetedForUpdate(urn) &&
-		!providers.IsProviderType(goal.Type) {
-
+	isTargeted := sg.isTargetedForUpdate(urn, goal.Parent, goal.Provider, goal.Dependencies)
+	if isTargeted {
+		sg.targetDependentsForUpdate(urn, goal.Parent, goal.Provider, goal.Dependencies)
+	}
```
pgavlin, line 540: Let's pull this earlier (above the `hasOld` case) so we can handle the check and the bookkeeping once.

pkg/resource/deploy/step_generator.go
```diff
@@ -72,8 +72,34 @@ func (sg *stepGenerator) isTargetedUpdate() bool {
 	return sg.updateTargetsOpt != nil || sg.replaceTargetsOpt != nil
 }
 
-func (sg *stepGenerator) isTargetedForUpdate(urn resource.URN) bool {
-	return sg.updateTargetsOpt == nil || sg.updateTargetsOpt[urn]
+// isTargetedForUpdate returns if urn is target for update. The function accommodates
+// `--target-dependents`. `targetDependentsForUpdate` should probably be called if this function
+// returns true.
+func (sg *stepGenerator) isTargetedForUpdate(
+	urn, parent resource.URN, provider string, dependencies []resource.URN) bool {
```
pgavlin, line 79: Nit: maybe use a single `*resource.State` parameter and pull properties off of that:

```suggestion
func (sg *stepGenerator) isTargetedForUpdate(r *resource.State) bool {
```

pgavlin (approved):
One nit, LGTM otherwise!

PR #8404, opened 2021-11-11
Dependency graph formalization and PBT tests

t0yv0:
# Description

Beginnings of formalizing and PBT-testing the dependency graph. This has been a starting point of many experiments for me in verifying PRs touching dependencies, and has been very useful to me, I think we can grow formalizations from here. I suggest checking it in as it will continue catching interesting changes even in the current form.

Fixes # (issue)

## Checklist

### DIFF
--- a/pkg/resource/graph/dependency_graph_rapid_test.go
+++ b/pkg/resource/graph/dependency_graph_rapid_test.go
@@ -1,0 +1,364 @@
+// Copyright 2016-2021, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Model-checks dependency_graph functionality against simple models
+// using property-based testing.
+//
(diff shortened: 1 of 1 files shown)


pkg/resource/graph/dependency_graph_rapid_test.go
```diff
@@ -0,0 +1,329 @@
...
+				if hasDependency(a, b) {
+					assert.Truef(t, aD[b],
+						"DependenciesOf(%v) is missing a dependecy %v",
+						a.URN, b.URN)
+				}
+				if aD[b] {
+					assert.True(t, expectedDependenciesOf(a, b),
+						"DependenciesOf(%v) includes an unexpected %v",
+						a.URN, b.URN)
+				}
```
iwahbe, line 94: I might be confused, but this doesn't seem to provide assertions for transitive dependencies of `a`. It feels like this is asserting `DependenciesOf(x) = list(x.parent, x.Dependencies.flatten(), x.provider).filter(x != nil)`.

t0yv0: That's exactly true, if we do not consider component resources ( just custom resources Custom: true). If it provided assertions for transitive dependencies they would fail, because that's not what DependenciesOf computes at the moment.

t0yv0: Looking at the DependenciesOf code though, the for loop is awfully confusing, it looks like this was intended to maybe be transitive but isn't (even in the case of ignoring the Component Resource subtleties)....

t0yv0:
Model extended to cover `includingChildren` that was just recently introduced.. FWIW..

iwahbe (approved):
LGTM

t0yv0:
Thanks Ian. I'm writing a quick doc for us to follow up on the underlying issues later this week, but will check this in for now, a few more tests don't hurt to fix the current behavior in place at least.

PR #8406, opened 2021-11-12
Update error handling

iwahbe:
This is the result of a change applied via `go-rewrap-errors`.

The one place where the new interface was not complient was the
`errors.Cause` method. This was copied directly from the go/pkg/errors
implementation.

# Description

Fixes #Modernize error handling in `pkg`. 

## Checklist

### DIFF
--- a/pkg/backend/filestate/backend.go
+++ b/pkg/backend/filestate/backend.go
@@ -322,7 +322,7 @@ func (b *localBackend) GetStack(ctx context.Context, stackRef backend.StackRefer
 	stackName := stackRef.Name()
 	snapshot, path, err := b.getStack(stackName)
 	switch {
-	case gcerrors.Code(legacyErrorCause(err)) == gcerrors.NotFound:
+	case gcerrors.Code(errors.Unwrap(err)) == gcerrors.NotFound:
 		return nil, nil
 	case err != nil:
 		return nil, err
--- a/pkg/backend/filestate/backend.go
+++ b/pkg/backend/filestate/backend.go
@@ -17,6 +17,7 @@ package filestate
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/url"
 	"os"
@@ -28,7 +29,7 @@ import (
 	"time"
 
 	"github.com/gofrs/uuid"
-	"github.com/pkg/errors"
+
 	user "github.com/tweekmonster/luser"
 	"gocloud.dev/blob"
 	_ "gocloud.dev/blob/azureblob" // driver for azblob://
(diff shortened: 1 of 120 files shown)


pkg/backend/filestate/state.go
```diff
@@ -441,3 +441,34 @@ func (b *localBackend) addToHistory(name tokens.QName, update backend.UpdateInfo
 	checkpointFile := fmt.Sprintf("%s.checkpoint.json", pathPrefix)
 	return b.bucket.Copy(context.TODO(), checkpointFile, b.stackPath(name), nil)
 }
+
+// This was copied from
+// https://github.com/pkg/errors/blob/5dd12d0cfe7f152f80558d591504ce685299311e/errors.go#L264-L288.
+func legacyErrorCause(err error) error {
```
justinvp, line 447: Is this actually necessary?

I haven't actually tried it out, so it would be good to confirm, but it looks like this is only used for errors returned from `listBucket`, and we're wrapping the error in on place in `listBucket`:

https://github.com/pulumi/pulumi/blob/070757c7a28dff940697ffc0f76f7d01361cd42e/pkg/backend/filestate/bucket.go#L80

I don't believe this wrapped error is going to have a `Cause` method since it's now using `fmt.Errorf` to wrap the original error rather than `errors.Wrap`.

Instead of `legacyErrorCause` shouldn't we be using the built-in [`errors.Unwrap`](https://pkg.go.dev/errors#Unwrap) to get the underlying error (taking care to handle the case where it isn't wrapped)?

justinvp:
Can the `go.mod` be updated to remove `github.com/pkg/errors`?

pkg/backend/filestate/state.go
iwahbe: It doesn't seem to be necessary. I went as far down he call stack as I could (until interfaces). I think it's fine if tests pass.

pkg/backend/filestate/backend.go
```diff
@@ -321,7 +322,7 @@ func (b *localBackend) GetStack(ctx context.Context, stackRef backend.StackRefer
 	stackName := stackRef.Name()
 	snapshot, path, err := b.getStack(stackName)
 	switch {
-	case gcerrors.Code(errors.Cause(err)) == gcerrors.NotFound:
+	case gcerrors.Code(errors.Unwrap(err)) == gcerrors.NotFound:
```
justinvp, line 325: `errors.Unwrap` returns `nil` if the error isn't wrapped (i.e. doesn't have an `Unwrap` method):

https://cs.opensource.google/go/go/+/refs/tags/go1.17.3:src/errors/wrap.go;l=14-22

Given the way `getStack` is implemented, I think we need to loop through calling `Unwrap` until it returns nil, and then return the previous err. Similar to what `errors.Is` and `errors.As` does:

https://cs.opensource.google/go/go/+/refs/tags/go1.17.3:src/errors/wrap.go;l=14-22

Right?

iwahbe: It looked to me like we wrapped any returned error, but I may have missed something. I can add that functionality.

iwahbe:
> Mind creating a shared function and using it everywhere in this file where we're calling `gcerrors.Code`? I just want to proactively guard against any potential future changes that would result in multiple wrapped errors.

Done

justinvp (approved):

PR #8407, opened 2021-11-12
Remove use of /opt/pulumi for nuget packages

Frassle:
Stop using /opt/pulumi/nuget, this makes some progress towards fixing

There are three main changes here:
1) We no longer pack on build, instead requiring an explicit pack command
(which runs in relase configuration). This should be a small speed improvement
for standard builds.
2) The integration tests no longer rely on a nuget package, instead we just
reference the project directly. This removes the need for a local nuget feed
that had to be setup and installed into.
3) As we've got rid of the local nuget feed `make install` simply runs "dotnet
pack" and `make publish` picks up the .nupkg from the sdk bin folder.

The one workflow this changes is if people have pulumi programs outside
pulumi/pulumi that they we're using `make install` to get a build of the
dotnet sdk into. Anyone using that workflow instead needs to either:
1) Manually copy the .nupkg from the build folder to wherever they have
setup their local nuget feed.
2) Change that project to refer to Pulumi via a project reference.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -160,8 +160,6 @@ jobs:
         uses: actions/setup-dotnet@v1
         with:
           dotnet-version: ${{ matrix.dotnet-version }}
-      - run: mkdir -p ${{ runner.temp }}/opt/pulumi/nuget
-      - run: dotnet nuget add source ${{ runner.temp }}/opt/pulumi/nuget
       - name: Set up Node ${{ matrix.node-version }}
         uses: actions/setup-node@v1
         with:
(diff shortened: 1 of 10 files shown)


Frassle:
> System.IO.FileLoadException: Could not load file or assembly 'Pulumi, Version=2.15.6.0, Culture=neutral, PublicKeyToken=null'. The located assembly's manifest definition does not match the assembly reference. (0x80131040)

Might have to think this some more. Because we've got some tests that use nuget packages that then refer to pulumi I think project references are tripping it up. I'll close this off to not clutter things and try a different approach next week.

t0yv0:
When working with ProjectReference locally I typically have to edit Pulumi.Core  `<AssemblyVersion>` attribute to get past as similar warning, perhaps this can help here.

Issue #8156, opened 2021-10-06
Output Values: Python SDK unmarshaling (used in MLC provider)


emiliza:
Part of: #7434

Issue #8155, opened 2021-10-06
Output Values: Node SDK unmarshaling (used in MLC provider)


emiliza:
Part of: #7434

PR #8410, opened 2021-11-13
Deprecate dependency maps when using output values

justinvp:
This PR is part of addressing @pgavlin's [feedback](https://github.com/pulumi/pulumi/pull/8205#discussion_r728459058).

When sending output values to `RegisterResource` (and then `Construct`) for remote components and `Call` for methods, there's no need to also send a dependencies map, because the output values themselves contain the dependencies. This PR deprecates the use of the dependencies map when output values are used.

## RegisterResource

Unfortunately, we cannot simply elide the dependencies map when calling `RegisterResource` for remote components because the monitor will [fill-in empty property dependencies with explicitly set dependencies](https://github.com/pulumi/pulumi/blob/164a2ec818994db6ac6491fa315a5ed44071a5e8/pkg/resource/deploy/source_eval.go#L926-L931). Since we want to continue working well with older CLIs, we need the SDKs to continue to send the property dependencies to `RegisterResource`. But we'd also like the monitor to not pass the property dependencies to `Construct` if there's no need to. To do that, the SDKs set a new `hasOutputs` field on the `RegisterResourceRequest` message when the properties are serialized keeping output values. This tells the monitor it can exclude passing the property dependencies to the provider's `Construct`.

When the monitor calls the provider's `Construct`, if the provider does not support output values and the property dependencies are empty, we "polyfill" the map based on the dependencies gathered in the properties (from output values and resource refs).

## Call

For `Call`, we don't have to worry about older CLIs filling in empty dependencies maps, so we can simply omit sending the map from the SDKs.

---

This PR is broken up into several commits to make it easier to review.

Part of #8155 and #8156


pgavlin:
> Since we want to continue working well with older CLIs

Okay, so we did decide not to take a breaking change here?

pkg/resource/deploy/source_eval.go
```diff
@@ -993,11 +997,14 @@ func (rm *resmon) RegisterResource(ctx context.Context,
...
-			PropertyDependencies: propertyDependencies,
-			Providers:            providerRefs,
+			Aliases:      aliases,
+			Dependencies: dependencies,
+			Protect:      protect,
+			Providers:    providerRefs,
+		}
+		if !hasOutputs {
+			// Only include property dependencies when we don't have output values in the properties.
+			options.PropertyDependencies = propertyDependencies
```
pgavlin, line 1007: Could we instead iterate the property dependencies and wrap each corresponding value with an `Output` that carries its deps? Then we could leave the `PropertyDependencies` out altogether from the call to `Construct`, I think.

PR #8419, opened 2021-11-15
[sdk/dotnet] Avoid eager conversion to Output<T> in Input<T>

### LINKED ISSUE #8418: [sdk/dotnet] `Input<T>` eagerly converts values to `Output<T>` leading to undesirable preview diffs
In the .NET SDK, `Input<T>` (and `InputList<T>`, `InputMap<V>`, `InputUnion<T0, T1>`, and `InputJson`) eagerly stores the value of `T` as an `Output<T>`, which leads to undesirable preview diffs.

https://github.com/pulumi/pulumi/blob/0a38bc295cfc5ae3c325dcece169da5a5e80fffc/sdk/dotnet/Pulumi/Core/Input.cs#L22-L31

For example, consider this program:

```csharp
using Pulumi;
using Pulumi.Random;
using Pulumi.Aws.S3;

class MyStack : Stack
{
    public MyStack()
    {
        var randString = new RandomString("str", new RandomStringArgs { Length = 10 });
        var bucket = new Bucket("bucket", new BucketArgs {
            Tags = {
                { "hello", "world" },
                { "hi", "there" },
                { "foo", randString.Id },
            }
        });
    }
}
```

Expected preview diff:

```
        tags        : {
            hello     : "world"
            hi        : "there"
            foo       : output<string>
        }
```

Actual:

```
        tags        : output<string>
```

This also means that we have the same issue with programs that create multi-lang components, even with output values added in #8316.

To address this, we need to fix `Input<T>` et. al. to not eagerly convert values to `Output<T>`. Instead, internally, it can keep a discriminated union of the plain `T` and `Output<T>` and only convert to `Output<T>` when needed.

justinvp:
This changes `Input<T>`, `InputList<T>`, `InputMap<V>`, and `InputUnion<T0, T1>` to no longer eagerly convert plain values to `Output<T>`.

Fixes pulumi/pulumi-dotnet#22

### DIFF
--- a/sdk/dotnet/Pulumi/Core/Input.cs
+++ b/sdk/dotnet/Pulumi/Core/Input.cs
@@ -1,7 +1,8 @@
-// Copyright 2016-2019, Pulumi Corporation
+// Copyright 2016-2021, Pulumi Corporation
 
 using System;
 using System.Threading.Tasks;
+using OneOf;
 
 namespace Pulumi
 {
@@ -13,6 +14,7 @@ namespace Pulumi
     internal interface IInput
     {
         IOutput ToOutput();
+        object? Value { get; }
     }
 
     /// <summary>
@@ -21,29 +23,34 @@ namespace Pulumi
     /// </summary>
     public class Input<T> : IInput
     {
-        /// <summary>
-        /// Technically, in .net we can represent Inputs entirely using the Output type (since
-        /// Outputs can wrap values and promises).  However, it would look very weird to state that
-        /// the inputs to a resource *had* to be Outputs. So we basically just come up with this
-        /// wrapper type so things look sensible, even though under the covers we implement things
-        /// using the exact same type
-        /// </summary>
-        private protected Output<T> _outputValue;
+        private readonly OneOf<T, Output<T>> _value;
+
+        private protected Input(T value)
+            => _value = OneOf<T, Output<T>>.FromT0(value);
 
         private protected Input(Output<T> outputValue)
-            => _outputValue = outputValue ?? throw new ArgumentNullException(nameof(outputValue));
+            => _value = OneOf<T, Output<T>>.FromT1(outputValue ?? throw new ArgumentNullException(nameof(outputValue)));
+
+        private protected virtual Output<T> ToOutput()
+            => _value.Match(v => Output.Create(v), v => v);
+
+        private protected virtual object? Value
+            => _value.Value;
 
         public static implicit operator Input<T>(T value)
-            => Output.Create(value);
+            => new Input<T>(value);
 
         public static implicit operator Input<T>(Output<T> value)
             => new Input<T>(value);
 
         public static implicit operator Output<T>(Input<T> input)
-            => input._outputValue;
+            => input.ToOutput();
 
         IOutput IInput.ToOutput()
             => this.ToOutput();
+
+        object? IInput.Value
+            => this.Value;
     }
 
     public static class InputExtensions
--- a/sdk/dotnet/Pulumi.Tests/Serialization/MarshalOutputTests.cs
+++ b/sdk/dotnet/Pulumi.Tests/Serialization/MarshalOutputTests.cs
@@ -91,6 +91,21 @@ namespace Pulumi.Tests.Serialization
             },
             new object[]
             {
+                new InputList<string> { "hello" },
+                ImmutableArray<object>.Empty.Add("hello")
+            },
+            new object[]
+            {
+                new InputList<string> { Output.Create("hello") },
+                ImmutableArray<object>.Empty.Add("hello")
+            },
+            new object[]
+            {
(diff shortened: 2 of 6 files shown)


sdk/dotnet/Pulumi/Core/Input.cs
```diff
@@ -21,29 +23,34 @@ internal interface IInput
...
     {
-        /// <summary>
-        /// Technically, in .net we can represent Inputs entirely using the Output type (since
-        /// Outputs can wrap values and promises).  However, it would look very weird to state that
-        /// the inputs to a resource *had* to be Outputs. So we basically just come up with this
-        /// wrapper type so things look sensible, even though under the covers we implement things
-        /// using the exact same type
-        /// </summary>
-        private protected Output<T> _outputValue;
+        private readonly OneOf<T, Output<T>> _value;
```
t0yv0, line 26: Ok so this is readonly means that InputList completely ignores this slot right, it just has its own slot called _value and overrides the accessors.. Sorry I have not worked much with inheritance and struggling a bit here.

sdk/dotnet/Pulumi/Core/Input.cs
```diff
@@ -21,29 +23,34 @@ internal interface IInput
...
 
         private protected Input(Output<T> outputValue)
-            => _outputValue = outputValue ?? throw new ArgumentNullException(nameof(outputValue));
+            => _value = OneOf<T, Output<T>>.FromT1(outputValue ?? throw new ArgumentNullException(nameof(outputValue)));
+
+        private protected virtual Output<T> ToOutput()
+            => _value.Match(v => Output.Create(v), v => v);
+
+        private protected virtual object? Value
+            => _value.Value;
```
t0yv0, line 38: Because virtual, it's no longer true the type of this is OneOf<T, Output<T>>, it can be something else, which is taken advantage of in InputList..

t0yv0: I think it might be worth annotating this prop with a note that this object is how serializer sees it, and that it can be different from  `OneOf<T, Output<T>>` introducing transforms. I also wonder what about deserializer. The hook can break symmetry e.g. can InputMap deserialize from what it serializes to? Or are we riding on the understanding that InputMaps never deserialize? Where you might be doing this is for example in writing typed provider mocks (not a feature yet), but when you're mocking a function F you might want to write your mock in terms of typed F.args, but mock framework would need to reconstruct F.args from the serial form.

t0yv0:
Thanks for the detailed explanation the ticket, I worked through this a bit and it was helpful!

sdk/dotnet/Pulumi/Core/Input.cs
justinvp: Yes, that's right. It's unfortunate that this slot (field) can't be used by `InputList<T>` (or the map), but I have to do something like this since `InputList<T>` is `Input<ImmutableArray<T>>` rather than `Input<ImmutableArray<Input<T>>>` as you pointed out. Since the subclasses aren't using the field anymore, they don't need access to it, so I've tightened up the visibility and made it readonly.

sdk/dotnet/Pulumi/Core/Input.cs
justinvp: > Because virtual, it's no longer true the type of this is OneOf<T, Output>, it can be something else, which is taken advantage of in InputList

Right.

> I think it might be worth annotating this prop with a note

Yes, sounds good!

> I also wonder what about deserializer. The hook can break symmetry e.g. can InputMap deserialize from what it serializes to? Or are we riding on the understanding that InputMaps never deserialize?

This is a good question. I don't think the InputMap is going to show up during deserialization. The deserializer deserializes protobuf structs as `ImmutableDictionary<string, object?>`:

https://github.com/pulumi/pulumi/blob/f98d39b84bc0e3a78a13dc09f80730e39a297b10/sdk/dotnet/Pulumi/Serialization/Deserializer.cs#L75-L100

Typically, when you have an input typed as `InputMap<string>` the output will be `Output<ImmutableDictionary<string, string>>`.

> Where you might be doing this is for example in writing typed provider mocks (not a feature yet)

Hmmm. This is a good question. This shouldn't be a problem with our current `IMocks` abstraction, but typed provider mocks is interesting to think about. This will also likely come up when we get around to implementing nice provider APIs for .NET (for multi-lang components) where we'll want to take serialized inputs and deserialize them into the inputs that a component expects, which could be `InputList<T>` or `InputMap<V>`. Let me think about this some more. I suspect we'll be able to make this work without too much trouble.

t0yv0: Thanks. My general intuition (from before Pulumi) is that serializers always end up needing to be symmetrical after more and more features pile up. Curious where we land on this.

Frassle:
dotnet code has moved repo so closing this off, reopen in pulumi-dotnet if still relevant

PR #8424, opened 2021-11-16
Add tests that return failures from Call

### LINKED ISSUE #7649: Methods: Additional test coverage
- [x] #7701
- [ ] #8424

justinvp:
This PR adds some integration tests that return failures from `Call`, addressing a couple issues found in the process:

- [sdk/nodejs] - Allow returning failures from Call in the provider without setting result outputs.
- [sdk/go] - Allow specifying Call failures from the provider.

Fixes #7649


sdk/nodejs/provider/server.ts
```diff
@@ -391,14 +391,17 @@ class Server implements grpc.UntypedServiceImplementation {
 
             const resp = new provproto.CallResponse();
 
-            const [ret, retDependencies] = await runtime.serializeResourceProperties(`call(${req.getTok()})`, result.outputs);
-            const returnDependenciesMap = resp.getReturndependenciesMap();
-            for (const [key, resources] of retDependencies) {
-                const deps = new provproto.CallResponse.ReturnDependencies();
-                deps.setUrnsList(await Promise.all(Array.from(resources).map(r => r.urn.promise())));
-                returnDependenciesMap.set(key, deps);
+            if (result.outputs) {
```
t0yv0, line 394: In case result.outputs is falsey, we don't need to resp.setReturn()? Use default value?

CHANGELOG_PENDING.md
```diff
@@ -20,6 +20,12 @@
 - [sdk/nodejs] - Unmarshal output values in component provider.
   [#8205](https://github.com/pulumi/pulumi/pull/8205)
 
+- [sdk/nodejs] - Allow returning failures from Call in the provider without setting result outputs.
```
t0yv0, line 23: OK this here explains it.

t0yv0 (approved):

PR #8430, opened 2021-11-16
[FIX] unprotect cmd example

raphaelauv:
when I run `pulumi up` with a protected resource pulumi give an example , but that example is not working

```
    error: Preview failed: unable to delete resource "urn:pulumi:XXXXXX::XXXX::gcp:component:all_resource$gcp:storage/bucket:Bucket::bucket_XXXXX"
    as it is currently marked for protection. To unprotect the resource, either remove the `protect` flag from the resource in your Pulumiprogram and run `pulumi up` or use the command:
    `pulumi state unprotect urn:pulumi:XXXX::XXXX::gcp:component:all_resource$gcp:storage/bucket:Bucket::bucket_XXXXXX`
```

If I run the example cmd , it throw : 

```log
error: No such resource "urn:pulumi:XXXX::XXX::gcp:component:all_resource:storage/bucket:Bucket::bucket_XXXXX" exists in the current state
```

If i put around the urn some **`**  it work  : 

```
pulumi state unprotect `urn:pulumi:XXXX::XXXX::gcp:component:all_resource$gcp:storage/bucket:Bucket::bucket_XXXXXX`
```

# Description

Fixes # (issue)

## Checklist

### DIFF
--- a/pkg/resource/deploy/step.go
+++ b/pkg/resource/deploy/step.go
@@ -341,7 +341,7 @@ func (s *DeleteStep) Apply(preview bool) (resource.Status, StepCompleteFunc, err
 			"as it is currently marked for protection. To unprotect the resource, "+
 			"either remove the `protect` flag from the resource in your Pulumi"+
 			"program and run `pulumi up` or use the command:\n"+
-			"`pulumi state unprotect `%s``", s.old.URN, s.old.URN)
+			"`pulumi state unprotect '%s'`", s.old.URN, s.old.URN)
 	}
 
 	// Deleting an External resource is a no-op, since Pulumi does not own the lifecycle.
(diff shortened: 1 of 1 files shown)


svangordon-fruit:
If I had to guess, this is related to escaping of special characters in the URN (especially `$`). IIRC, in the docs it's recommended to surround URNs in single quotes (`'`) to avoid variable interpolation. Single quotes might be more portable across shells than backticks.

pkg/resource/deploy/step.go
```diff
@@ -341,7 +341,7 @@ func (s *DeleteStep) Apply(preview bool) (resource.Status, StepCompleteFunc, err
 			"as it is currently marked for protection. To unprotect the resource, "+
 			"either remove the `protect` flag from the resource in your Pulumi"+
 			"program and run `pulumi up` or use the command:\n"+
-			"`pulumi state unprotect %s`", s.old.URN, s.old.URN)
+			"`pulumi state unprotect `%s``", s.old.URN, s.old.URN)
```
iwahbe, line 344:
```suggestion
			"`pulumi state unprotect '%s'`", s.old.URN, s.old.URN)
```
I agree with @svangordon-fruit. We should use `'`. Backtick doesn't work on bash or zsh or fish, off the top of my head.

iwahbe (changes requested):

iwahbe (approved):
LGTM

PR #8431, opened 2021-11-16
Go implementation of fn.Output program gen (#7949)

### LINKED ISSUE #7949: Update program gen to employ new $fn_output forms

t0yv0:
# Description

Fixes #7949 for Go

## Checklist

### DIFF
--- a/pkg/codegen/go/gen_program_expressions.go
+++ b/pkg/codegen/go/gen_program_expressions.go
@@ -195,7 +195,18 @@ func (g *generator) GenFunctionCallExpression(w io.Writer, expr *model.FunctionC
 		if module == "" {
 			module = pkg
 		}
-		name := fmt.Sprintf("%s.%s", module, fn)
+		isOut, outArgs, outArgsType := pcl.RecognizeOutputVersionedInvoke(expr)
+		if isOut {
+			outTypeName, err := outputVersionFunctionArgTypeName(outArgsType)
+			if err != nil {
+				panic(fmt.Errorf("Error when generating an output-versioned Invoke: %w", err))
+			}
+			g.Fgenf(w, "%s.%sOutput(ctx, ", module, fn)
+			g.genObjectConsExpressionWithTypeName(w, outArgs, outArgsType, outTypeName)
+		} else {
+			g.Fgenf(w, "%s.%s(ctx, ", module, fn)
+			g.Fgenf(w, "%.v", expr.Args[1])
+		}
 
 		optionsBag := ""
 		var buf bytes.Buffer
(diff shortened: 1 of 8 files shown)


pkg/codegen/go/gen_program_expressions.go
```diff
@@ -814,9 +858,14 @@ func (g *generator) genApply(w io.Writer, expr *model.FunctionCallExpression) {
 	isInput := false
 	retType := g.argumentTypeName(nil, then.Signature.ReturnType, isInput)
 	// TODO account for outputs in other namespaces like aws
-	typeAssertion := fmt.Sprintf(".(%sOutput)", retType)
-	if !strings.HasPrefix(retType, "pulumi.") {
-		typeAssertion = fmt.Sprintf(".(pulumi.%sOutput)", Title(retType))
+	var typeAssertion string
+	if retType == "[]string" {
+		typeAssertion = ".(pulumi.StringArrayOutput)"
```
t0yv0, line 863: This is far from complete here, I'm adding just enough to pass the tests. Any thoughts on how to factor this out? We need to backlog this.

pkg/codegen/internal/test/program_driver.go
```diff
@@ -49,8 +49,11 @@ var programTests = []programTest{
 	{
 		Name:        "aws-fargate",
 		Description: "AWS Fargate",
+
+		// TODO[pulumi/pulumi#8440]
```
t0yv0, line 53: So this is back to where master was, but now with a note linking that we have an issue. Go code does not compile here on master.

pkg/codegen/go/gen_program_expressions.go
```diff
@@ -195,7 +195,18 @@ func (g *generator) GenFunctionCallExpression(w io.Writer, expr *model.FunctionC
...
 		}
-		name := fmt.Sprintf("%s.%s", module, fn)
+		isOut, outArgs, outArgsType := pcl.RecognizeOutputVersionedInvoke(expr)
+		if isOut {
+			outTypeName, err := outputVersionFunctionArgTypeName(outArgsType)
+			if err != nil {
+				panic(fmt.Errorf("Error when generating an output-versioned Invoke: %w", err))
+			}
+			g.Fgenf(w, "%s.%sOutput(ctx, ", module, fn)
+			g.genObjectConsExpressionWithTypeName(w, outArgs, outArgsType, outTypeName)
```
t0yv0, line 205: Special casing here to force a custom type name computation. This is tying back to the unfortunate business of type names being custom for inputty- and normal version of function arg types.

t0yv0:
@pgavlin rebased on master(+pcl change). This is slightly complicated because of (custom) type names for inputty args. The PR essentially addresses the problems head on, it is not the most nice and clean solution.

pkg/codegen/go/gen_program_expressions.go
pgavlin: Not offhand. Definitely need to backlog this. In general we need to go take another hard look at the Go program generator and its approach to naming.

pgavlin (approved):

PR #8433, opened 2021-11-16
Programgen support for fn_output forms in Python (#7949)

### LINKED ISSUE #7949: Update program gen to employ new $fn_output forms

t0yv0:
# Description

Fixes #7949 for Python

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,5 +1,9 @@
 ### Improvements
 
+- [codegen/python] - Program generator now uses `fn_output` forms where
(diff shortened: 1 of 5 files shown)


t0yv0:
@pgavlin this one up next. It's really simple compared to Go and .NET. The ProgramGen framework does not yet support running unit tests on top of the generated code, but I tried a program that hits the example code against mocks and it seems reasonable. The Python structures automagically pass through inputs and it works out, there is no new name for inputty arg structures in Python.

pgavlin (approved):

PR #8434, opened 2021-11-16
Programgen support for fnOutput forms in node (#7949)

### LINKED ISSUE #7949: Update program gen to employ new $fn_output forms

t0yv0:
# Description

Fixes #7949 for Node

## Checklist

### DIFF
--- a/pkg/codegen/hcl2/model/expression.go
+++ b/pkg/codegen/hcl2/model/expression.go
@@ -1034,7 +1034,13 @@ func (x *FunctionCallExpression) Typecheck(typecheckOperands bool) hcl.Diagnosti
 	typecheckDiags := typecheckArgs(rng, x.Signature, x.Args...)
 	diagnostics = append(diagnostics, typecheckDiags...)
 
-	x.Signature.ReturnType = liftOperationType(x.Signature.ReturnType, x.Args...)
+	// Unless the function is already automatically using an
+	// Output-returning version, modify the signature to account
+	// for automatic lifting to Promise or Output.
+	_, isOutput := x.Signature.ReturnType.(*OutputType)
+	if !isOutput {
+		x.Signature.ReturnType = liftOperationType(x.Signature.ReturnType, x.Args...)
+	}
 	return diagnostics
 }
 
--- a/pkg/codegen/internal/test/testdata/aws-fargate-pp/nodejs/aws-fargate.ts
+++ b/pkg/codegen/internal/test/testdata/aws-fargate-pp/nodejs/aws-fargate.ts
@@ -4,9 +4,9 @@ import * as aws from "@pulumi/aws";
 const vpc = aws.ec2.getVpc({
     "default": true,
 });
-const subnets = vpc.then(vpc => aws.ec2.getSubnetIds({
-    vpcId: vpc.id,
-}));
+const subnets = aws.ec2.getSubnetIdsOutput({
+    vpcId: vpc.then(vpc => vpc.id),
+});
 // Create a security group that permits HTTP ingress and unrestricted egress.
 const webSecurityGroup = new aws.ec2.SecurityGroup("webSecurityGroup", {
     vpcId: vpc.then(vpc => vpc.id),
@@ -43,7 +43,7 @@ const taskExecRolePolicyAttachment = new aws.iam.RolePolicyAttachment("taskExecR
 });
 // Create a load balancer to listen for HTTP traffic on port 80.
 const webLoadBalancer = new aws.elasticloadbalancingv2.LoadBalancer("webLoadBalancer", {
-    subnets: subnets.then(subnets => subnets.ids),
+    subnets: subnets.ids,
     securityGroups: [webSecurityGroup.id],
 });
 const webTargetGroup = new aws.elasticloadbalancingv2.TargetGroup("webTargetGroup", {
@@ -85,7 +85,7 @@ const appService = new aws.ecs.Service("appService", {
     taskDefinition: appTask.arn,
     networkConfiguration: {
         assignPublicIp: true,
-        subnets: subnets.then(subnets => subnets.ids),
+        subnets: subnets.ids,
         securityGroups: [webSecurityGroup.id],
     },
     loadBalancers: [{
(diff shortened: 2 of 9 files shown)


pkg/codegen/hcl2/model/expression.go
```diff
@@ -1034,7 +1034,13 @@ func (x *FunctionCallExpression) Typecheck(typecheckOperands bool) hcl.Diagnosti
...
 	diagnostics = append(diagnostics, typecheckDiags...)
 
-	x.Signature.ReturnType = liftOperationType(x.Signature.ReturnType, x.Args...)
+	// Unless the function is already automatically using an
+	// Output-returning version, modify the signature to account
+	// for automatic lifting to Promise or Output.
+	_, isOutput := x.Signature.ReturnType.(*OutputType)
+	if !isOutput {
+		x.Signature.ReturnType = liftOperationType(x.Signature.ReturnType, x.Args...)
+	}
```
pgavlin, line 1043: This is interesting. I think that the lifting behavior is wrong in general here: essentially, it's lifting based on the assumption that all parameter types are prompt, so any eventual inputs must transform eventual outputs. Instead, the lifting should be conditional on the argument type not being assignable to the parameter type. In other words, a call that passes an `Output<T>` arg to an `Output<T>` parameter does not require lifting, but a call that passes an `Output<T>` arg to a `T` parameter does.

pkg/codegen/pcl/invoke.go
```diff
@@ -92,22 +77,140 @@ func (b *binder) bindInvokeSignature(args []model.Expression) (model.StaticFunct
...
+	return b.makeSignature(argsType, model.NewOutputType(returnType)), nil
+}
+
+// Detects invoke calls that use an output version of a function.
+func IsOutputVersionInvokeCall(call *model.FunctionCallExpression) bool {
+	if call.Name == Invoke {
+		// Currently binder.bindInvokeSignature will assign
+		// either DynamicType, a Promise<T>, or an Output<T>
+		// for the return type of an invoke. Output<T> implies
+		// that an output version has been picked.
```
pgavlin, line 168: Do you think it would it be worth using a different name for the output-form invoke intrinsic, or does that seem overcomplicated?

pkg/codegen/internal/test/testdata/aws-fargate-pp/nodejs/aws-fargate.ts
```diff
@@ -4,9 +4,9 @@ import * as aws from "@pulumi/aws";
 const vpc = aws.ec2.getVpc({
     "default": true,
 });
-const subnets = vpc.then(vpc => aws.ec2.getSubnetIds({
-    vpcId: vpc.id,
-}));
+const subnets = aws.ec2.getSubnetIdsOutput({
+    vpcId: vpc.then(vpc => vpc.id),
+});
```
pgavlin, line 9: This is interesting--so it looks like we're choosing the output form if there are any eventuals in the args, not just `Output`s. IMO we should adjust that so that we only choose the output form if there are `Output`s in the args (consistent with my general opinion that we should only use `Output<>` in our examples where we must).

pkg/codegen/pcl/invoke.go
t0yv0: I don't have the compiler pipeline in my head so difficult to judge, I could try to explore this? Where would such a change go?

The current change fits rather well in `bindInvokeSignature`, so I looked up call sites of that, but what I find is:

```	// Define the invoke function.
	b.root.DefineFunction(Invoke, model.NewFunction(model.GenericFunctionSignature(b.bindInvokeSignature)))
```

It is defining an (overloaded) function Invoke. That's not the right place to define it as `InvokeOutput` or some such it seems, since "overloading" happens elsewhere.

Do you believe things might fit better if we had a new `InvokeOutput` intrinsic?

pkg/codegen/internal/test/testdata/aws-fargate-pp/nodejs/aws-fargate.ts
t0yv0: Hm okay, I think I see. So it should continue lifting over promises but not over outputs. I can work this out tomorrow morning.

pkg/codegen/pcl/invoke.go
pgavlin: > Do you believe things might fit better if we had a new InvokeOutput intrinsic?

Nope, I think what you've laid out in your comment has convinced me that it wouldn't. Thanks!

pkg/codegen/internal/test/testdata/aws-fargate-pp/nodejs/aws-fargate.ts
t0yv0: This one is done.

pkg/codegen/hcl2/model/expression.go
pgavlin: Filed https://github.com/pulumi/pulumi/issues/8439 to track.

pgavlin (approved):
LGTM. We can fix https://github.com/pulumi/pulumi/issues/8439 later on if you'd like.

pkg/codegen/hcl2/model/expression.go
t0yv0: I'm looking at it now briefly now.

Tried this but it fails:

```go

	diags := typecheckArgs(rng, x.Signature, x.Args...)
	lifted := liftOperationType(x.Signature.ReturnType, x.Args...)

	if !lifted.equals(x.Signature.ReturnType, nil) {
		fmt.Printf("LIFTED BUT DIAGS: %v\n", diags)
	}

	if diags.HasErrors() {
		x.Signature.ReturnType = lifted
	}

```

```
=== RUN   TestGenerateProgram/AWS_EKS
LIFTED BUT DIAGS: no diagnostics
```

t0yv0: It's super tricky, `x.Parameters[0].Type == Dynamic` in the failing tests. So we don't know the real parameter type at this point. I'll need help on  #8439 later on, I'm also confused why this code is modifying the signature. It's kind of surprising, we'll need to put some comments and point to the overview of the Typecheck pass and how it interplays with other passes I think, to make this easy to understand.

PR #8437, opened 2021-11-17
Move `/opt/pulumi` to `$HOME/.pulumi`

iwahbe:
# Description

`/opt/pulumi` is legacy of our build system we are trying to fix. We still need a place to put executables and artifacts. Luckily, `~/.pulumi` already exists. This moves our build artifacts from `/opt/pulumi.{bin,nuget,node_modules}` to `~/.pulumi/{bin,nuget,node_modules}`. This fixes the insane permissions problems of `/opt/pulumi`, as `~/.pulumi` is normal user accessible space. 

Fixes # (issue)

## Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -41,7 +41,7 @@ We use `make` as our build system, so you'll want to install that as well, if yo
 We build Pulumi in `$PULUMI_ROOT`, which defaults to `$HOME/.pulumi`. If you would like to build Pulumi in another location, you do so by setting `$PULUMI_ROOT`. 
 
 ```bash
-export PATH=$HOME/.pulumi:$HOME/.pulumi/bin:$PATH
+export PATH=$HOME/.pulumi/bin:$PATH
 ```
 
 You'll also need to make sure your maximum open file descriptor limit is set to 5000 at a minimum.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -38,12 +38,10 @@ If you have a web browser, you can get a fully pre-configured Pulumi development
 
 We use `make` as our build system, so you'll want to install that as well, if you don't have it already. We have extremely limited support for doing development on Windows (the bare minimum for us to get Windows validation of `pulumi`) so if you're on windows, we recommend that you use the [Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/install-win10). We'd like to [make this better](https://github.com/pulumi/pulumi/issues/208) so feel free to pitch in if you can.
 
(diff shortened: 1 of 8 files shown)


CONTRIBUTING.md
```diff
@@ -38,12 +38,10 @@ If you have a web browser, you can get a fully pre-configured Pulumi development
...
 
-For historical reasons (which we'd [like to address](https://github.com/pulumi/pulumi/issues/1515)) our build system requires that the folder `/opt/pulumi` exists and is writable by the current user. If you'd like, you can override this location by setting `PULUMI_ROOT` in your environment. The build is known to fail if this doesn't exist, so you'll need to create it first.
+We build Pulumi in `$PULUMI_ROOT`, which defaults to `$HOME/.pulumi`. If you would like to build Pulumi in another location, you do so by setting `$PULUMI_ROOT`. 
 
 ```bash
-mkdir /opt/pulumi
-sudo chown <your_user_name>: /opt/pulumi
-export PATH=/opt/pulumi:/opt/pulumi/bin:$PATH
+export PATH=$HOME/.pulumi:$HOME/.pulumi/bin:$PATH
```
Frassle, line 44: I don't think we currently put anything at "pulumi" directly, just inside "pulumi/bin" so can remove one of these path entries.

Frassle (approved):
I think this is a reasonable step to make the build system better. It still has some downsides that /opt/pulumi had in that we need to fix up path and install the sdks in a global location for tests and have appropriate global settings for it (e.g. you'll have to run `dotnet nuget add source $HOME/.pulumi/nuget` before running the tests) but it's an improvement.

Frassle:
Note that CI is still configured to use /opt/pulumi after this change.

iwahbe:
> Note that CI is still configured to use /opt/pulumi after this change.

Yup. We are pretty careful to set `PULUMI_ROOT` in CI, so it should be fine.

justinvp (approved):
Looks reasonable to me.

One thing worth noting is that our install script also installs pulumi to `$HOME/.pulumi/bin`:

https://github.com/pulumi/get.pulumi.com/blob/c7e7379ef8053f0dab4fc6d49d9fcfe58f5282e1/dist/install.sh#L133

So now when you build locally, the locally built binaries will be copied over anything that's been installed via the install script. Which is probably fine. In the past, I've kept around an installed version in `$HOME/.pulumi/bin` and have my `PATH` setup so that I can easily delete the locally built `pulumi` binary in `/opt/pulumi/bin` to be able to use the released version in `$HOME/.pulumi/bin`. Now I'll either have to set `PULUMI_ROOT` when building locally (to keep around my local install), or just delete `$HOME/.pulumi/bin` and reinstall a released version in order to use it over a local build.

iwahbe:
I actually like that this will install to the real default install location. It makes sense that without special work, a `make install` command will install as close as possible to the real install. It also makes it harder to accidentally shadow the version of Pulumi you are working on with a production copy.

PR #8448, opened 2021-11-18
Preview of update plans

### LINKED ISSUE #2318: Enable constraining an update or refresh to a preview result

Frassle:
# Description

This PR adds the concept of "update plans" to pulumi. Plans can be generated via a preview command with the extra argument "--save-plan=<file_path>.json". This will run the preview and save the set of expected operations to the given JSON file. The user can then run update with "--plan=<file_path>.json" to run a constrained update. This will only allow operations that are either recorded in the plan file or are deemed a safe subset of those operations.

Some examples of safe subsets:
* If a plan file says nothing about resource R but the update doesn't want to change anything about resource R this is not a constraint error.
* If a plan says resource R should be updated to value X, but at update time it's already at value X this is not a constraint error.

Using this feature requires the environment variable PULUMI_EXPERIMENTAL to be true.

Fixes #2318

## Checklist

## Todo

- [X] Auto refresh in update if --constraint-file is given
- [x] Work out what to do about auto preview in update if --constraint-file is given


pkg/cmd/pulumi/preview.go
```diff
@@ -223,6 +235,13 @@ func newPreviewCmd() *cobra.Command {
 	cmd.PersistentFlags().BoolVar(
 		&configPath, "config-path", false,
 		"Config keys contain a path to a property in a map or list to set")
+	if hasExperimentalCommands() {
+		cmd.PersistentFlags().StringVar(
+			&planFilePath, "save-constraints", "",
+			"Save the operations proposed by the preview to a constraints file at the given path")
```
infin8x, line 241:
```suggestion
			&planFilePath, "save-plan", "",
			"Save the operations proposed by the preview to a plan file at the given path")
```

pkg/cmd/pulumi/up.go
```diff
@@ -505,6 +522,14 @@ func newUpCmd() *cobra.Command {
 		&yes, "yes", "y", false,
 		"Automatically approve and perform the update after previewing it")
 
+	if hasExperimentalCommands() {
+		cmd.PersistentFlags().StringVar(
+			&planFilePath, "constraints-file", "",
+			"Path to a constraints file to use for the update. The update will not "+
+				"perform operations that exceed its constraints (e.g. replacements instead of updates, or updates instead"+
+				"of sames).")
```
infin8x, line 530:
```suggestion
			&planFilePath, "plan-file", "",
			"Path to a plan file to use for the update. The update will not "+
				"perform operations that exceed its plan (e.g. replacements instead of updates, or updates instead"+
				"of sames).")
```

Also, what does "sames" mean in this context?

infin8x (commented):
After some discussions with @gthuang and `@joeduffy`, I think we want to call these "Update Plans" after all. Tried to propose changes here.

pkg/cmd/pulumi/up.go
Frassle: "sames" is that we don't expect any change

pkg/backend/apply.go
```diff
@@ -196,10 +197,13 @@ func PreviewThenPromptThenExecute(ctx context.Context, kind apitype.UpdateKind,
 	// Preview the operation to the user and ask them if they want to proceed.
 
 	if !op.Opts.SkipPreview {
-		changes, res := PreviewThenPrompt(ctx, kind, stack, op, apply)
+		plan, changes, res := PreviewThenPrompt(ctx, kind, stack, op, apply)
 		if res != nil || kind == apitype.PreviewUpdate {
 			return changes, res
 		}
+
+		// TODO(pdg-plan): should this check for an existing plan?
```
Frassle, line 205: We need more tests for how plans work when doing the auto preview as part of up. I'm not sure we should be replacing the plan here, but also because of mutation I don't think we can just run with the original plan object (*shakes fist at mutability by default*)

pkg/cmd/pulumi/util.go
```diff
@@ -862,10 +865,43 @@ func getRefreshOption(proj *workspace.Project, refresh string) (bool, error) {
...
+	if err != nil {
+		return err
+	}
+	defer contract.IgnoreClose(f)
+
+	deploymentPlan, err := stack.SerializePlan(plan, enc, showSecrets)
+	if err != nil {
+		return err
+	}
+	return json.NewEncoder(f).Encode(deploymentPlan)
```
lukehoban, line 888: This is not pretty-printing - so we get one long line.  We should likely do similar to https://github.com/pulumi/pulumi/blob/fraser/planPrototype/sdk/go/common/workspace/creds.go#L227:L227 to ensure we get pretty-printed JSON output here for readability.

sdk/go/common/apitype/plan.go
```diff
@@ -0,0 +1,78 @@
...
+	Version int             `json:"version"`
+	Plan    json.RawMessage `json:"plan"`
+}
+
+// DeploymentPlanV1 is the serializable version of a deployment plan.
+type DeploymentPlanV1 struct {
+	// TODO(pdg-plan): should there be a message here?
+
+	// Manifest contains metadata about this plan.
+	Manifest ManifestV1 `json:"manifest" yaml:"manifest"`
```
lukehoban, line 70: When I'm using the feature, I see this manifest having empty value.  I assume we should be populating this?

```json
    "manifest": {
        "time": "0001-01-01T00:00:00Z",
        "magic": "",
        "version": ""
    },
```

sdk/go/common/apitype/plan.go
```diff
@@ -0,0 +1,78 @@
...
+	// Any environment variables that were set when the plan was created. Values are encrypted.
+	EnvironmentVariables map[string][]byte `json:"environmentVariables,omitempty"`
+	// The configuration in use during the plan.
+	Config map[string]ConfigValue `json:"config"`
```
lukehoban, line 74: I'm currently seeing this show as `null`, even though there is config in my environment.

```json
    "config": null,
```

```
$ pulumi config
KEY         VALUE
aws:region  us-west-2
```

lukehoban: Related - `null` should not be a valid value here.  I think if we are using this, it should be `omitempty` so that we don't include it when not needed, or else the map should always be initialized to `map[string]ConfigValue{}` so that this is `"config": {},`

pkg/cmd/pulumi/util.go
```diff
@@ -862,10 +865,43 @@ func getRefreshOption(proj *workspace.Project, refresh string) (bool, error) {
 		return true, nil
 	}
 
-	// the default functionality right now is to always skip a refresh
+	// if the user has passed a plan file then the default behaviour is to refresh
+	if planFilePath != "" {
+		return true, nil
+	}
```
lukehoban, line 871: I think it is quite surprising to couple `--plan` to `--refresh`.  These are orthogonal options, and users can certainly choose to use both together, but to have the former imply the latter doesn't feel right.

I'm fairly sure we will not want to do this - but I'm interested to understand the motivating scenario for this better.

lukehoban (commented):
Trying this out locally - ran into a few blocking issues.

First - an issue with stack exports.  If I deploy this:

```ts
const bucket = new aws.s3.Bucket("my-bucket", {});
export const bucketName = bucket.id;
```

Then run `pulumi preview --save-plan plan.json` then immediately `pulumi up --plan plan.json`, I get:

```
➜  simple PULUMI_EXPERIMENTAL=true /opt/pulumi/bin/pulumi preview --save-plan plan.json
Previewing update (dev)

View Live: https://app.pulumi.com/lukehoban/simple-foo/dev/previews/66981c6d-e229-45ae-a086-a010d3e0c0b5

     Type                 Name            Plan
     pulumi:pulumi:Stack  simple-foo-dev
     └─ aws:s3:Bucket     my-bucket

Resources:
    2 unchanged

➜  simple PULUMI_EXPERIMENTAL=true /opt/pulumi/bin/pulumi up --plan plan.json
Previewing update (dev)

View Live: https://app.pulumi.com/lukehoban/simple-foo/dev/previews/f6898a61-2352-40cc-a9f9-2104b6ce8291

     Type                 Name            Plan        Info
 ~   pulumi:pulumi:Stack  simple-foo-dev  refresh     1 error
 ~   └─ aws:s3:Bucket     my-bucket       refresh

Diagnostics:
  pulumi:pulumi:Stack (simple-foo-dev):
    error: resource violates plan: properties changed: -bucketName


error: Expected resource operations for urn:pulumi:dev::simple-foo::pulumi:pulumi:Stack::simple-foo-dev but none were seen.
```

Second - an issue that I suspect is due to the refresh piece (I have notes in the review feedback that I think we need to decouple this - but if we are considering keeping this, we'll need to ensure it doesn't trigger the issue here):

If I deploy this:

```ts
const bucket = new aws.s3.Bucket("my-bucket", {});
```

And then change it to:

```ts
```ts
    website: {
        indexDocument: "index.html",
    },
});
```

Then run `pulumi preview --save-plan plan.json` then immediately `pulumi up --plan plan.json`, I get:

```
➜  simple PULUMI_EXPERIMENTAL=true /opt/pulumi/bin/pulumi up --plan plan.json
Previewing update (dev)

View Live: https://app.pulumi.com/lukehoban/simple-foo/dev/previews/cf933799-a8ec-4d61-9fe6-a73232576220

     Type                 Name            Plan        Info
     pulumi:pulumi:Stack  simple-foo-dev
 ~   └─ aws:s3:Bucket     my-bucket       refresh     1 error

Diagnostics:
  aws:s3:Bucket (my-bucket):
    error: resource violates plan: properties changed: -acl, -websiteEndpoint, -bucketDomainName, -requestPayer, -bucketRegionalDomainName, -corsRules, -tags, -region, -lifecycleRules, -versioning, -tagsAll, -bucket, -loggings, -websiteDomain, -id, -forceDestroy, -accelerationStatus, -grants, -hostedZoneId, -arn, +website


error: Expected resource operations for urn:pulumi:dev::simple-foo::aws:s3/bucket:Bucket::my-bucket but none were seen.
```

All of these changed properties are results of the refresh modifying the state, but don't seem to be captured into the plan, so that the plan doesn't actually capture what the combined operation is going to do, so the corresponding update can't work.

Frassle:
> First - an issue with stack exports. If I deploy this:

Yup there's nothing handling RegisterResourceOutputs calls right now. I hadn't clicked that these we're such a separate code path, adding some tests for this now.

> Second - an issue that I suspect is due to the refresh piece 

This might be due to refresh, I'll need to do some more working out exactly how all these bits behave. I would of thought a refresh after up would always leave the stack in the same state, but this is suggesting that's not true?

lukehoban:
> I would of thought a refresh after up would always leave the stack in the same state, but this is suggesting that's not true?

Yes, unfortunately that is not true today.  Providers are free to take actions in a refresh that they didn't take during create, and quite a lot of common Terraform providers in particular do differ between these.

sdk/go/common/apitype/plan.go
Frassle: Yeh this needs filling in, it's a stub put in by @pgavlin. I'm not sure about "magic" but I guess it makes sense to set time to current time and and version to 1.

sdk/go/common/apitype/plan.go
Frassle: I'll fix up filling this in and add `omitempty`

sdk/go/common/apitype/plan.go
Frassle: Ah I see this is the same structure as for stack files so I'll go see how those get filled in

pkg/resource/deploy/deployment_executor.go
```diff
@@ -236,7 +236,15 @@ func (ex *deploymentExecutor) Execute(callerCtx context.Context, opts Options, p
 				}
 
 				if event.Event == nil {
-					return false, ex.performDeletes(ctx, updateTargetsOpt, destroyTargetsOpt)
+					res := ex.performDeletes(ctx, updateTargetsOpt, destroyTargetsOpt)
```
Frassle, line 239: I think I added this while debugging some unplanned delete bugs. I think it's safe to stay here, means you get a nicer error if deletes do fail and it just translated into a Bail later on anyway.

pkg/resource/deploy/deployment_executor.go
```diff
@@ -477,7 +518,6 @@ func (ex *deploymentExecutor) refresh(callerCtx context.Context, opts Options, p
 	}
 
 	// Make sure if there were any targets specified, that they all refer to existing resources.
-	targetMapOpt := createTargetMap(opts.RefreshTargets)
```
Frassle, line 480: I could revert this, just moved it down 50 lines to where it was actually used while I was looking at this code a lot.

pkg/resource/deploy/step_generator.go
```diff
@@ -335,8 +378,16 @@ func (sg *stepGenerator) generateSteps(event RegisterResourceEvent) ([]Step, res
 		// invalid (they got deleted) so don't consider them. Similarly, if the old resource was External,
 		// don't consider those inputs since Pulumi does not own them. Finally, if the resource has been
 		// targeted for replacement, ignore its old state.
-		if recreating || wasExternal || sg.isTargetedReplace(urn) {
-			inputs, failures, err = prov.Check(urn, nil, goal.Properties, allowUnknowns)
+		if recreating || wasExternal || sg.isTargetedReplace(urn) || !hasOld {
```
Frassle, line 381: The `!hasOld` should be safe here, in the case where hasOld == false, oldChecked will be nil and goal.properties == inputs.

pkg/cmd/pulumi/up.go
```diff
@@ -299,10 +317,11 @@ func newUpCmd() *cobra.Command {
...
 		opts.Engine = engine.UpdateOptions{
-			LocalPolicyPacks: engine.MakeLocalPolicyPacks(policyPackPaths, policyPackConfigPaths),
-			Parallel:         parallel,
-			Debug:            debug,
-			Refresh:          refreshOption,
+			LocalPolicyPacks:  engine.MakeLocalPolicyPacks(policyPackPaths, policyPackConfigPaths),
+			Parallel:          parallel,
+			Debug:             debug,
+			Refresh:           refreshOption,
+			ExperimentalPlans: hasExperimentalCommands(),
```
t0yv0, line 324: Should this check for `planFilePath != ""`?

pkg/backend/backend.go
```diff
@@ -161,7 +161,7 @@ type Backend interface {
 	RenameStack(ctx context.Context, stack Stack, newName tokens.QName) (StackReference, error)
 
 	// Preview shows what would be updated given the current workspace's contents.
-	Preview(ctx context.Context, stack Stack, op UpdateOperation) (engine.ResourceChanges, result.Result)
+	Preview(ctx context.Context, stack Stack, op UpdateOperation) (*deploy.Plan, engine.ResourceChanges, result.Result)
```
t0yv0, line 164: This code structure is a tad confusing to me. Backend says it's an interface for `pluggable backend implementations of the Pulumi Cloud Service.` but it's not just storage functionality that's in the interface it's the engine functionality? And an implementation of this interface wraps the engine? Interesting.

pkg/resource/deploy/deployment_executor.go
```diff
@@ -267,8 +275,38 @@ func (ex *deploymentExecutor) Execute(callerCtx context.Context, opts Options, p
 		res = ex.checkTargets(opts.UpdateTargets, OpUpdate)
 	}
 
+	// Check that we did operations for everything expected in the plan. We mutate ResourcePlan.Ops as we run
```
t0yv0, line 278: Spend a bunch of time reading this.. I think it makes sense given the explanation of how it mutates. Find ResourcePlans with pending ops (shouldn't be any), fail if any of them is not an unneeded delete. Why `res == nil` check?

pkg/resource/deploy/step_generator.go
```diff
@@ -179,6 +179,42 @@ func (sg *stepGenerator) GenerateSteps(event RegisterResourceEvent) ([]Step, res
 		contract.Assert(len(steps) == 0)
 		return nil, res
 	}
+
+	// Check each proposed step against the relevant resource plan, if any
+	for _, s := range steps {
```
t0yv0, line 184: This all looks reasonable - just stylistically I'm wondering if some of it could be pulled out into pure functions for clarity? I'm trying to calibrate my biases vs what's the preferred style here (or idiomatic Go..).. 

What seems to be going on here, there's a few side effects.

If `sg.deployment.plan != nil` then []Step is checked against it, with failures surfacing as result.Errorf. Could this be a pure func? Or is `resourcePlan.Ops = resourcePlan.Ops[1:]` also meaningfully mutates/ pops off Op? The order of the Ops and the order of the Steps must match then?

Then the other part is in ExperimentalPlans we  push the op `resourcePlan.Ops = append(resourcePlan.Ops, s.Op())`.

Somehow this is in the same for loop. I'm now wondering if that's really essential for this to be in the same for loop like the op queue appended to by L214 needs sometimes to be popped off in L191, like the side-effect interaction is essential?

t0yv0: Also it seems that GenerateSteps has only *one* call site in the codebase. Does this logic better fit inside GenerateSteps because sequencing is really important, or could it be factored into "CheckPlans" or such and integrated where GenerateSteps call site is?

sdk/go/common/apitype/plan.go
```diff
@@ -0,0 +1,89 @@
...
+	// the type of resource.
+	Type tokens.Type `json:"type"`
+	// the name for the resource's URN.
+	Name tokens.QName `json:"name"`
+	// true if this resource is custom, managed by a plugin.
+	Custom bool `json:"custom"`
+	// the resource properties that will be changed.
+	CheckedInputs map[string]interface{} `json:"checkedInputs,omitempty"`
+	// the resource properties that will be changed.
+	InputDiff PlanDiffV1 `json:"inputDiff,omitempty"`
```
t0yv0, line 32: Perhaps different doc comment between CheckedInputs vs InputDiff? A bit of elaboration here on how these are different could be really helpful to the reader. Presumably InputDiff=d(old, new), but are CheckedInputs total inputs in old or new version? Or they're partial?

sdk/go/common/apitype/plan.go
```diff
@@ -0,0 +1,89 @@
...
+	// the resource outputs that will be changed.
+	OutputDiff *PlanDiffV1 `json:"outputDiff,omitempty"`
```
t0yv0, line 34: Is the *PlanDiffV1 intentional vs PlanDiffV1 in InputDiff? Extra meaning for nil?

sdk/go/common/apitype/plan.go
```diff
@@ -0,0 +1,89 @@
...
+// DeploymentPlanV1 is the serializable version of a deployment plan.
+type DeploymentPlanV1 struct {
+	// TODO(pdg-plan): should there be a message here?
+
+	// Manifest contains metadata about this plan.
+	Manifest ManifestV1 `json:"manifest" yaml:"manifest"`
+	// Any environment variables that were set when the plan was created. Values are encrypted.
+	EnvironmentVariables map[string]string `json:"environmentVariables,omitempty"`
+	// The configuration in use during the plan.
+	Config config.Map `json:"config,omitempty"`
```
t0yv0, line 85: Similarly encrypted? This and capturing EnvironmentVariables makes me a little worried we might be grabbing information not quite needed for operation of the feature, and accidentally grabbing sensitive information and sending it places.

sdk/go/common/resource/properties_diff.go
```diff
@@ -392,3 +392,271 @@ func (v PropertyValue) DeepEquals(other PropertyValue) bool {
...
+		}
+		vr := v.ResourceReferenceValue()
+		or := other.ResourceReferenceValue()
+
+		if vr.URN != or.URN {
+			return false
+		}
+
+		vid, oid := vr.ID, or.ID
+		if vid.IsComputed() && oid.IsComputed() {
```
t0yv0, line 626: && or || here?

pkg/resource/deploy/plan.go
```diff
@@ -0,0 +1,612 @@
...
+		Version: version.Version,
+		// Plugins: sm.plugins, - Explicitly dropped, since we don't use the plugin list in the manifest anymore.
+	}
+	manifest.Magic = manifest.NewMagic()
+
+	environmentVariables := make(map[string]string)
+	for _, e := range os.Environ() {
+		// golang has so many questionable design decisions, lets just hope no one puts '=' in there env var names.
+		pair := strings.SplitN(e, "=", 2)
+		environmentVariables[pair[0]] = pair[1]
```
t0yv0, line 47: There was a comment elsewhere these were encrypted (at the struct property), but there's no encryption I can see. So passwords from env vars would go into EnvironmentVariables prop. Should we change the comments here?

pkg/resource/deploy/plan.go
```diff
@@ -0,0 +1,612 @@
...
+// Goal is a desired state for a resource object.  Normally it represents a subset of the resource's state expressed by
+// a program, however if Output is true, it represents a more complete, post-deployment view of the state.
+type GoalPlan struct {
+	// the type of resource.
+	Type tokens.Type
+	// the name for the resource's URN.
+	Name tokens.QName
+	// true if this resource is custom, managed by a plugin.
+	Custom bool
+	// the resource's checked input properties that we saw during preview.
```
t0yv0, line 124: Ah looks like here the doc comments are different between CheckedInputs and InputDiff.

pkg/resource/deploy/plan.go
```diff
@@ -0,0 +1,612 @@
...
+	CheckedInputs resource.PropertyMap
+	// the resource's checked input properties we expect to change.
+	InputDiff PlanDiff
+	// the resource's output properties we expect to change (only set for RegisterResourceOutputs)
+	OutputDiff *PlanDiff
```
t0yv0, line 129: *PlanDiff intentional vs PlanDiff? So for non RegisterResourceOutputs it's nil?

pkg/resource/deploy/plan.go
```diff
@@ -0,0 +1,612 @@
...
+
+// A ResourcePlan represents the planned goal state and resource operations for a single resource. The operations are
+// ordered.
+type ResourcePlan struct {
+	Goal    *GoalPlan
+	Ops     []StepOp
+	Outputs resource.PropertyMap
+}
+
+func (rp *ResourcePlan) diffURNs(a, b []resource.URN) (message string, changed bool) {
```
t0yv0, line 215: Are these sorted/deduped, that is are these `[]` slices actually representing sets? Should we be defensive and sort/dedup here?

pkg/resource/deploy/plan.go
```diff
@@ -0,0 +1,612 @@
...
+		stringsA[i] = string(key)
+	}
+	stringsB := make([]string, len(a))
+	for i, key := range b {
+		stringsB[i] = string(key)
+	}
+	return rp.diffStrings(stringsA, stringsB)
+}
+
+func (rp *ResourcePlan) diffStrings(a, b []string) (message string, changed bool) {
```
t0yv0, line 239: Ah ok set logic here. "diffStringSets"? perhaps a more pointed name.

pkg/resource/deploy/plan.go
```diff
@@ -0,0 +1,612 @@
...
+		outputDiff := PlanDiff{}
+		if err := checkDiff(oldOutputs, newOutputs, outputDiff); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (rp *ResourcePlan) checkGoal(
```
t0yv0, line 491: This code is written with early termination semantic (return first error), would it be sometimes helpful to collect and present all errors or it does not matter?

t0yv0 (approved):
Finished reading through - LGTM, added a few comments/ questions. My familiarity with some of the code is a bit light, esp around stateful struct updates + flow through the system. I believe @pgavlin was intending to do a review pass here also.

pkg/backend/backend.go
Frassle: Urgh don't even get me started on this! The whole thing is a mad blob of everything and I need to gut a load of it to get project support in local backends sorted.

pkg/cmd/pulumi/up.go
Frassle: Nope, even if we don't have a plan this turns on the generation of plans so that the inline preview for up generates a plan and passes it on to the actual update operation. That is we actually *always* use plans even if no plan file is given.

pkg/resource/deploy/deployment_executor.go
Frassle: Not really any point checking for plan failures if we're going to fail the deployment anyway. Chances are if the deployment failed a load of expected ops didn't happen so this would just be noise. That should probably be a comment here...

pkg/resource/deploy/step_generator.go
Frassle: > Then the other part is in ExperimentalPlans we push the op resourcePlan.Ops = append(resourcePlan.Ops, s.Op()).

Note that is the generation of *newPlans* vs the modification of the *currentPlan*. So yes all the mutation here is kinda needed.

> or could it be factored into "CheckPlans" or such and integrated where GenerateSteps call site is?

It would be really invasive to push it down into generateSteps where the step list is made.

sdk/go/common/apitype/plan.go
Frassle: Ah yes, I should update CheckedInputs to point out its temporary as well until we get fixed autonames.

sdk/go/common/apitype/plan.go
Frassle: Huh yeh I think at one point nil meant something special. I think its just the same as empty now I'll try and clean up.

sdk/go/common/apitype/plan.go
Frassle: Hmm no it probably should be!

sdk/go/common/resource/properties_diff.go
Frassle: I think `||` actually. There's some redundancy above with IsComputed checks at the top of this function doing `&&` and `||`

sdk/go/common/apitype/plan.go
Frassle: Ah wait Config already tracks what is secure or not so the secret config values should already be encrypted.

pkg/resource/deploy/plan.go
Frassle: config values are encrypted/decrypted as used. env vars are encrypted when we write them out. We _could_ change this so we pass a crypter in to NewPlan and just encrypt the envvars straight away.

pkg/resource/deploy/plan.go
Frassle: Yes they are sets that's why we make the two string maps and then diff that.

pkg/resource/deploy/plan.go
Frassle: Done

pkg/resource/deploy/plan.go
Frassle: Probably, need to think about how best to display large error sets to the user. I think this will be ok for preview release.

pkg/cmd/pulumi/preview.go
```diff
@@ -200,6 +207,26 @@ func newPreviewCmd() *cobra.Command {
...
+					if err != nil {
+						return result.FromError(err)
+					}
+					if err = writePlan(planFilePath, plan, encrypter, showSecrets); err != nil {
+						return result.FromError(err)
+					}
+
+					// Write out message on how to use the plan
+					var buf bytes.Buffer
+					writer := bufio.NewWriter(&buf)
```
pgavlin, line 221: Nit: no need to wrap this in a `bufio.Writer`--you can just use `&buf` directly.

pkg/cmd/pulumi/preview.go
```diff
@@ -223,6 +250,13 @@ func newPreviewCmd() *cobra.Command {
 	cmd.PersistentFlags().BoolVar(
 		&configPath, "config-path", false,
 		"Config keys contain a path to a property in a map or list to set")
+	if hasExperimentalCommands() {
+		cmd.PersistentFlags().StringVar(
+			&planFilePath, "save-plan", "",
+			"Save the operations proposed by the preview to a plan file at the given path")
+	}
```
pgavlin, line 257: Nit: rather than conditionally defining this command based on `hasExperimentalCommands`, I have a slight preference for hiding it based on `hasExperimentalCommands`: that approach doesn't require setting the envvar to use the command, only to discover it.

pkg/cmd/pulumi/util.go
```diff
@@ -866,6 +869,36 @@ func getRefreshOption(proj *workspace.Project, refresh string) (bool, error) {
...
+	defer contract.IgnoreClose(f)
+
+	deploymentPlan, err := stack.SerializePlan(plan, enc, showSecrets)
+	if err != nil {
+		return err
+	}
+	encoder := json.NewEncoder(f)
+	encoder.SetIndent("", "  ")
+	return encoder.Encode(deploymentPlan)
```
pgavlin, line 885: We might want to standardize on indentation here. `stack export` uses four spaces, for example.

pkg/resource/deploy/deployment_executor.go
```diff
@@ -267,8 +275,41 @@ func (ex *deploymentExecutor) Execute(callerCtx context.Context, opts Options, p
...
+	// so by the time we get here everything in the map should have an empty ops list (except for unneeded
+	// deletes). We skip this check if we already have an error, chances are if the deployment failed lots of
+	// operations wouldn't have got a chance to run so we'll spam errors about all of those failed operations
+	// making it less clear to the user what the root cause error was.
+	if res == nil && ex.deployment.plan != nil {
+		for urn, resourcePlan := range ex.deployment.plan.ResourcePlans {
+			if len(resourcePlan.Ops) != 0 {
+				if len(resourcePlan.Ops) == 1 && resourcePlan.Ops[0] == OpDelete {
+					// We haven't done a delete for this resource check if it was in the snapshot,
+					// if it's already gone this wasn't done because it wasn't needed
```
pgavlin, line 288: How does this scenario arise?

pkg/resource/deploy/plan.go
pgavlin: Just curious whether or not we need the envvars? More information is probably good, but I might leave them out of the preview if we're not yet sure how we'll use them.

pkg/resource/deploy/plan.go
```diff
@@ -0,0 +1,605 @@
...
+func (plan *Plan) Clone() *Plan {
+	return copystructure.Must(copystructure.Copy(plan)).(*Plan)
+}
+
+// PlanDiff holds the results of diffing two object property maps.
+type PlanDiff struct {
+	Adds    resource.PropertyMap   // the resource's properties we expect to add.
+	Deletes []resource.PropertyKey // the resource's properties we expect to delete.
+	Updates resource.PropertyMap   // the resource's properties we expect to update.
+}
```
pgavlin, line 68: I'm a little surprised by the way we're storing diffs.

I think that using the inputs like this imposes a very strong requirement that all changes are visible via the inputs--i.e. there are no input vs. state differences. This is different from the model used by providers, where diffs can be observed either from input <-> input or from input <-> state. An alternative approach would be to store the detailed diff as returned by the provider (polyfilling a detailed diff where necessary), then checking that the detailed diff from the update is constrained to the detailed diff from the preview, with old values sourced from the old inputs or old state as directed by the detailed diff.

I think that what is implemented might be okay assuming that in the vast majority of cases all input <-> state differences are essentially input <-> input differences b/c the differing property has the same value in the state and the inputs (though that assumption requires that the provider's `Read` method can accurately reflect values from the state back to the inputs).

I was also a little surprised that we're storing a flat list of top-level diffs rather than the tree diff (ala `resource.ValueDiff`). Is there a benefit to the flat list? Storing the tree diff might make error presentation a little more precise.

pkg/resource/deploy/deployment_executor.go
Frassle: If you planned to delete a resource, then someone else deletes it, then you run your plan. End result is the same, the resource is deleted so it didn't feel like it should be an error case.

pkg/resource/deploy/plan.go
Frassle: They were in the prototype branch and I just kept them, I'm not sure if we would ever do anything with them. There's so many other ways users could cause their programs to be non-deterministic.

pkg/resource/deploy/plan.go
Frassle: > I think that using the inputs like this imposes a very strong requirement that all changes are visible via the inputs

What would an input state difference be? I'm not sure I follow how we could have a state difference and not also have an input difference?

> Is there a benefit to the flat list?

Depends on what semantics we're going for. Like if you have a dict property and you run a plan to go from`{a: 1, b: 2}` to `{a: 1, c: 3}` is that recording "update to `{a: 1, c: 3}`" or "delete key b, add key c with 3"? Because the former only allows a plan to execute if it ends up with the value `{a: 1, c: 3}` but the later would allow someone to come and change the dict to something like `{b: 2, d: 4}` and the plan would allow the value `{c: 3, d: 4}` but not allow the value `{a: 1, c: 3}`.

Might be that second semantic is useful but it seemed more complex to code and explain and I don't have a concrete scenario that would use it so I err'ed on simplicity.

pgavlin: > What would an input state difference be? I'm not sure I follow how we could have a state difference and not also have an input difference?

Consider a resource `R` with `(inputs, state) => ({foo: bar}, {foo: bar})`. If we refresh `R` and the provider returns `({foo: bar}, {foo: baz})`, the provider is still allowed to indicate that `R` needs to be updated on a subsequent run even if the program presents the inputs `{foo: bar}` b/c `foo` has different properties in the inputs and the state. This is arguably a bug, and I'm not sure it happens in practice, but it is a "feature" of the model that we might need to consider.

More common is the case where an input has no corresponding entry in the state, so diffs are only detectable from input to input (e.g. k8s secrets).

> the later would allow someone to come and change the dict to something like {b: 2, d: 4} and the plan would allow the value {c: 3, d: 4} but not allow the value {a: 1, c: 3}.

I think the latter would still allow `{a: 1, c: 3}` b/c the `same` on `c` is constrained to the `add` in the diff. I think the argument here is the same as for the analogous case at top level in that we want to allow the base state to differ as long as the diffs are compatible.

mikhailshilkov: This is good feedback and it sounds like we need to give it a thought. Would it be okay to go ahead with this PR to kick off examples testing and do this work as a follow-up?

pgavlin: Yep, definitely. We can probably even ship the preview as-is and revisit this while we gather more data.

Issue #7897, opened 2021-09-02
Panic when using `iam.RoleArrayOutput.Index` in Go


t0yv0:
Hitting an issue with cluster.InstanceRoles.Index in Go against pulumi-eks and pulumi-aws. I suspect this might be a codegen or SDK related issue so filling here:

## Steps to reproduce

```go

package main

import (
	"github.com/pulumi/pulumi-eks/sdk/go/eks"
	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
)

func main() {
	pulumi.Run(func(ctx *pulumi.Context) error {
		cluster, err := eks.NewCluster(ctx, "example", &eks.ClusterArgs{})
		if err != nil {
			return err
		}
		cluster.InstanceRoles.Index(pulumi.Int(0))
		return nil
	})
}
```

```
    panic: interface conversion: pulumi.Output is pulumi.AnyOutput, not iam.RoleOutput
    goroutine 1 [running]:
    github.com/pulumi/pulumi-aws/sdk/v4/go/aws/iam.RoleArrayOutput.Index(0xc0003d44d0, 0x20c4080, 0x2077960, 0xc0005ea300)
    	/Users/anton/go/pkg/mod/github.com/pulumi/pulumi-aws/sdk/v4@v4.18.0/go/aws/iam/role.go:665 +0x1a6
    main.main.func1(0xc0004aa140, 0x1f6d936, 0x13)
    	/Users/anton/tmp/repro-7862/go1/main.go:21 +0xc5
    github.com/pulumi/pulumi/sdk/v3/go/pulumi.RunWithContext(0xc0004aa140, 0x1fa7b60, 0xc000042087, 0x2)
    	/Users/anton/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.11.0/go/pulumi/run.go:103 +0x206
    github.com/pulumi/pulumi/sdk/v3/go/pulumi.RunErr(0x1fa7b60, 0x0, 0x0, 0x0, 0x0, 0x0)
    	/Users/anton/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.11.0/go/pulumi/run.go:84 +0x23f
    github.com/pulumi/pulumi/sdk/v3/go/pulumi.Run(0x1fa7b60, 0x0, 0x0, 0x0)
[...]

PR #8449, opened 2021-11-18
[codegen/go] Remove ResourcePtr input/output types

pgavlin:
These changes remove the `Ptr` variants of input/ouptut types for
resources. A `TPtr` input or output is normally generated for `T` if `T`
is present in an `optional(input(T))` or `optional(output(T))` and if
the Go representation for `T` is not nilable. The generation of `Ptr`
variants for resource types breaks the latter rule: the canonical
representation of a resource type named `Foo` is a pointer to a struct
type named `Foo` (i.e. `*Foo`). `Foo` itself is not a resource, as it
does not implement the Go `Resource` interface. Because this
representation already accommodates `nil` to indicate the lack of a
value, we need not generate `FooPtr{Input,Output}` types.

Besides being unnecessary, the implementation of `Ptr` types for
resources was incorrect. Rather than using `**Foo` as their element
type, these types use `*Foo`--identical to the element type used for
the normal input/output types. Furthermore, the generated code for
at least `FooOutput.ToFooPtrOutputWithContext` and `FooPtrOutput.Elem`
was incorrect, making these types virtually unusable in practice.

Finally, these `Ptr` types should never appear on input/output
properties in practice, as the logic we use to generate input and output
type references never generates them for `optional({input,output}(T)).
Instead, it generates references to the standard input/output types.

Though this is _technically_ a breaking change--it changes the set of
exported types for any package that defines resources--I believe that in
practice it will be invisible to users for the reasons stated above.
These types are not usable, and were never referenced.

This is preparatory work for #7943.

### DIFF
--- a/pkg/codegen/go/gen.go
+++ b/pkg/codegen/go/gen.go
@@ -784,16 +784,14 @@ type genInputImplementationArgs struct {
 	elementType     string
 	ptrMethods      bool
 	toOutputMethods bool
-	resourceType    bool
 }
 
-func genInputImplementation(w io.Writer, name, receiverType, elementType string, ptrMethods, resourceType bool) {
+func genInputImplementation(w io.Writer, name, receiverType, elementType string, ptrMethods bool) {
 	genInputImplementationWithArgs(w, genInputImplementationArgs{
 		name:            name,
 		receiverType:    receiverType,
 		elementType:     elementType,
 		ptrMethods:      ptrMethods,
-		resourceType:    resourceType,
 		toOutputMethods: true,
 	})
 }
@@ -802,14 +800,9 @@ func genInputImplementationWithArgs(w io.Writer, genArgs genInputImplementationA
 	name := genArgs.name
 	receiverType := genArgs.receiverType
 	elementType := genArgs.elementType
-	resourceType := genArgs.resourceType
 
 	fmt.Fprintf(w, "func (%s) ElementType() reflect.Type {\n", receiverType)
-	if resourceType {
-		fmt.Fprintf(w, "\treturn reflect.TypeOf((*%s)(nil))\n", elementType)
-	} else {
-		fmt.Fprintf(w, "\treturn reflect.TypeOf((*%s)(nil)).Elem()\n", elementType)
-	}
+	fmt.Fprintf(w, "\treturn reflect.TypeOf((*%s)(nil)).Elem()\n", elementType)
 	fmt.Fprintf(w, "}\n\n")
 
 	if genArgs.toOutputMethods {
@@ -837,15 +830,11 @@ func genInputImplementationWithArgs(w io.Writer, genArgs genInputImplementationA
 	}
 }
 
-func genOutputType(w io.Writer, baseName, elementType string, ptrMethods, resourceType bool) {
+func genOutputType(w io.Writer, baseName, elementType string, ptrMethods bool) {
 	fmt.Fprintf(w, "type %sOutput struct { *pulumi.OutputState }\n\n", baseName)
 
 	fmt.Fprintf(w, "func (%sOutput) ElementType() reflect.Type {\n", baseName)
-	if resourceType {
-		fmt.Fprintf(w, "\treturn reflect.TypeOf((*%s)(nil))\n", elementType)
-	} else {
-		fmt.Fprintf(w, "\treturn reflect.TypeOf((*%s)(nil)).Elem()\n", elementType)
-	}
+	fmt.Fprintf(w, "\treturn reflect.TypeOf((*%s)(nil)).Elem()\n", elementType)
 	fmt.Fprintf(w, "}\n\n")
 
 	fmt.Fprintf(w, "func (o %[1]sOutput) To%[2]sOutput() %[1]sOutput {\n", baseName, Title(baseName))
@@ -869,8 +858,8 @@ func genOutputType(w io.Writer, baseName, elementType string, ptrMethods, resour
 	}
 }
 
-func genArrayOutput(w io.Writer, baseName, elementType string, resourceType bool) {
-	genOutputType(w, baseName+"Array", "[]"+elementType, false, resourceType)
+func genArrayOutput(w io.Writer, baseName, elementType string) {
+	genOutputType(w, baseName+"Array", "[]"+elementType, false)
 
 	fmt.Fprintf(w, "func (o %[1]sArrayOutput) Index(i pulumi.IntInput) %[1]sOutput {\n", baseName)
 	fmt.Fprintf(w, "\treturn pulumi.All(o, i).ApplyT(func (vs []interface{}) %s {\n", elementType)
@@ -879,8 +868,8 @@ func genArrayOutput(w io.Writer, baseName, elementType string, resourceType bool
 	fmt.Fprintf(w, "}\n\n")
 }
 
-func genMapOutput(w io.Writer, baseName, elementType string, resourceType bool) {
-	genOutputType(w, baseName+"Map", "map[string]"+elementType, false, resourceType)
+func genMapOutput(w io.Writer, baseName, elementType string) {
+	genOutputType(w, baseName+"Map", "map[string]"+elementType, false)
 
 	fmt.Fprintf(w, "func (o %[1]sMapOutput) MapIndex(k pulumi.StringInput) %[1]sOutput {\n", baseName)
 	fmt.Fprintf(w, "\treturn pulumi.All(o, k).ApplyT(func (vs []interface{}) %s{\n", elementType)
@@ -889,8 +878,8 @@ func genMapOutput(w io.Writer, baseName, elementType string, resourceType bool) 
 	fmt.Fprintf(w, "}\n\n")
 }
 
-func genPtrOutput(w io.Writer, baseName, elementType string, resourceType bool) {
-	genOutputType(w, baseName+"Ptr", "*"+elementType, false, resourceType)
+func genPtrOutput(w io.Writer, baseName, elementType string) {
+	genOutputType(w, baseName+"Ptr", "*"+elementType, false)
 
 	fmt.Fprintf(w, "func (o %[1]sPtrOutput) Elem() %[1]sOutput {\n", baseName)
 	fmt.Fprintf(w, "\treturn o.ApplyT(func(v *%[1]s) %[1]s {\n", baseName)
@@ -956,7 +945,7 @@ func (pkg *pkgContext) genEnum(w io.Writer, enumType *schema.EnumType) error {
 
 		fmt.Fprintf(w, "type %[1]sArray []%[1]s\n\n", name)
 
-		genInputImplementation(w, name+"Array", name+"Array", "[]"+name, false, false)
+		genInputImplementation(w, name+"Array", name+"Array", "[]"+name, false)
 	}
 
 	// Generate the map input.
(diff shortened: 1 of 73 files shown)


pgavlin:
These changes also fix the element types for `FooArray` and `FooMap` where `Foo` is a resource type. Prior to these changes, the element types were `[]Foo` and `map[string]Foo`; they are now `[]*Foo` and `map[string]*Foo`. The prior element type prevented these input/output/args types from being usable in practice.

mikhailshilkov:
@pgavlin I don't see much changes in https://github.com/pulumi-bot/pulumi-azure-native/compare/c62ea8d3cd6..828691f7e76
Am I jumping the gun or looking at a wrong thing?

pgavlin:
> @pgavlin I don't see much changes in https://github.com/pulumi-bot/pulumi-azure-native/compare/c62ea8d3cd6..828691f7e76
> Am I jumping the gun or looking at a wrong thing?

Ah, I think there's not much of a change there b/c azure-native already disables these types by leaving the `GenerateResourceContainerTypes` option on the Go language config unset.

pgavlin:
...which makes me even more confident that we should take this change, as it further limits the potential set of breaks.

iwahbe:
@pgavlin, Will this change leave `Ptr` types for non-resource types? It looks like it does. I'm wondering if there is any `"Ptr"` handling code we can remove with this change.

pgavlin:
> @pgavlin, Will this change leave Ptr types for non-resource types? It looks like it does. I'm wondering if there is any "Ptr" handling code we can remove with this change.

It will, yes.

pgavlin:
(we continue to need `Ptr` types to represent optional object- and enum-typed inputs and outputs)

pkg/codegen/go/gen.go
```diff
@@ -1704,42 +1692,26 @@ func (pkg *pkgContext) genResource(w io.Writer, r *schema.Resource, generateReso
...
-		// Emit the resource pointer input type.
-		fmt.Fprintf(w, "type %sPtrInput interface {\n", name)
-		fmt.Fprintf(w, "\tpulumi.Input\n\n")
-		fmt.Fprintf(w, "\tTo%[1]sPtrOutput() %[1]sPtrOutput\n", name)
-		fmt.Fprintf(w, "\tTo%[1]sPtrOutputWithContext(ctx context.Context) %[1]sPtrOutput\n", name)
-		fmt.Fprintf(w, "}\n\n")
-		ptrTypeName := camel(name) + "PtrType"
-		fmt.Fprintf(w, "type %s %sArgs\n\n", ptrTypeName, name)
-		genInputImplementation(w, name+"Ptr", "*"+ptrTypeName, "*"+name, false, true)
+	genInputImplementation(w, name, "*"+name, "*"+name, false)
```
t0yv0, line 1695: Is this intentional that the last flag becomes `false`?

```
func genInputImplementation(w io.Writer, name, receiverType, elementType string, ptrMethods, resourceType bool) {
```

We're removing `ptrMethods` so

```
func genInputImplementation(w io.Writer, name, receiverType, elementType string, resourceType bool) {
```

So the flag `false` refers to `resourceType` param. It used to be `true`.

t0yv0: Ah never mind.

```
- func genInputImplementation(w io.Writer, name, receiverType, elementType string, ptrMethods, resourceType bool) {
+ func genInputImplementation(w io.Writer, name, receiverType, elementType string, ptrMethods bool) {
```

So false is ptrMethods. That sounds right.

pkg/codegen/internal/test/testdata/cyclic-types/go/example/provider.go
```diff
@@ -48,7 +48,7 @@ type ProviderInput interface {
 }
 
 func (*Provider) ElementType() reflect.Type {
-	return reflect.TypeOf((*Provider)(nil))
+	return reflect.TypeOf((**Provider)(nil)).Elem()
```
t0yv0, line 51: This is a no-op change right? It seems to me before and after is the same. Would it be possible to special-case the generator a bit to avoid the change. It's a minor thing here but it's really annoying in provider upgrades to be seeing  more diffs than strictly needed for the change.

pkg/codegen/internal/test/testdata/external-resource-schema/go/example/cat.go
```diff
@@ -88,35 +88,6 @@ func (i *Cat) ToCatOutputWithContext(ctx context.Context) CatOutput {
...
 
-func (i *Cat) ToCatPtrOutput() CatPtrOutput {
-	return i.ToCatPtrOutputWithContext(context.Background())
-}
-
-func (i *Cat) ToCatPtrOutputWithContext(ctx context.Context) CatPtrOutput {
-	return pulumi.ToOutputWithContext(ctx, i).(CatPtrOutput)
-}
-
-type CatPtrInput interface {
```
t0yv0, line 99: OK awesome good riddance.

pkg/codegen/internal/test/testdata/external-resource-schema/go/example/cat.go
```diff
@@ -230,15 +167,15 @@ func (o CatArrayOutput) ToCatArrayOutputWithContext(ctx context.Context) CatArra
 }
 
 func (o CatArrayOutput) Index(i pulumi.IntInput) CatOutput {
-	return pulumi.All(o, i).ApplyT(func(vs []interface{}) Cat {
-		return vs[0].([]Cat)[vs[1].(int)]
+	return pulumi.All(o, i).ApplyT(func(vs []interface{}) *Cat {
```
t0yv0, line 170: I'll be curious if this here closes https://github.com/pulumi/pulumi/issues/7897

pkg/codegen/internal/test/testdata/external-resource-schema/go/example/cat.go
```diff
@@ -250,18 +187,16 @@ func (o CatMapOutput) ToCatMapOutputWithContext(ctx context.Context) CatMapOutpu
 }
 
 func (o CatMapOutput) MapIndex(k pulumi.StringInput) CatOutput {
-	return pulumi.All(o, k).ApplyT(func(vs []interface{}) Cat {
-		return vs[0].(map[string]Cat)[vs[1].(string)]
+	return pulumi.All(o, k).ApplyT(func(vs []interface{}) *Cat {
```
t0yv0, line 190: Would be curious to check this too.

pkg/codegen/internal/test/testdata/resource-property-overlap/dotnet/Rec.cs
```diff
@@ -13,7 +13,7 @@ namespace Pulumi.Example
     public partial class Rec : Pulumi.CustomResource
     {
         [Output("rec")]
-        public Output<Pulumi.Example.Outputs.Rec?> Rec { get; private set; } = null!;
+        public Output<Pulumi.Example.Rec?> Rec { get; private set; } = null!;
```
t0yv0, line 16: This is curious. Self-reference?

pkg/codegen/internal/test/testdata/resource-property-overlap/go/example/rec.go
```diff
@@ -13,7 +13,7 @@ import (
 type Rec struct {
 	pulumi.CustomResourceState
 
-	Rec RecPtrOutput `pulumi:"rec"`
+	Rec RecOutput `pulumi:"rec"`
```
t0yv0, line 16: INteresting.

pkg/codegen/internal/test/testdata/resource-property-overlap/schema.json
```diff
@@ -5,7 +5,7 @@
     "example::Rec": {
       "properties": {
         "rec": {
-          "$ref": "#/types/example::Rec"
+          "$ref": "#/resources/example::Rec"
```
t0yv0, line 8: 🤔 it sounds like this used to be an invalid schema?

pkg/codegen/internal/test/testdata/external-resource-schema/go/example/cat.go
t0yv0: I tried unit testing this but alas Go compilation is disabled because it generates reference against non-existing type:

```
        "requiredMetadataArray": {
          "type": "array",
          "items": {
            "$ref": "/kubernetes/v3.7.0/schema.json#/types/kubernetes:meta%2Fv1:ObjectMeta"
          }
        },
```

Becomes `RequiredMetadataArray metav1.ObjectMetaArrayInput` which is a dangling ref.

t0yv0:
I've attached a unit test in https://github.com/pulumi/pulumi/pull/8462/files and I'd like to understand why it's timing out.

pkg/codegen/internal/test/testdata/cyclic-types/go/example/provider.go
pgavlin: This is a no-op change, yes. 

> Would it be possible to special-case the generator a bit to avoid the change.

It would, though that sort of special-casing is what I was hoping to avoid with these changes (I removed the special-case code that produced the original output).

pkg/codegen/internal/test/testdata/resource-property-overlap/dotnet/Rec.cs
pgavlin: Yep.

pkg/codegen/internal/test/testdata/resource-property-overlap/schema.json
pgavlin: I _think_ that this was originally an unexpected schema. We've always allowed for references to undefined types in order to support overlays. With @lblackstone 's recent changes to allow overlays to be represented in a first class fashion in the schema, we might be able to change that.

pgavlin:
> I've attached a unit test in https://github.com/pulumi/pulumi/pull/8462/files and I'd like to understand why it's timing out.

I don't want to block on that, frankly. This work is setup for https://github.com/pulumi/pulumi/pull/7943, and I'd rather get it merged and investigate the timeout after that's done.

t0yv0:
Okay.. well do we have any evidence (test) that the generated Index and MapIndex forms are working at all, before or after the change? I'm looking around for examples and I can't any, and anything I try fails.

t0yv0:
Need another half-our to cross check, sounds like I've found enough evidence to file some bugs against mainline and proceed here, one sec.

t0yv0:
Cross-checked this a bit. We can ship based on this:

Before the change:
- BucketArray.ToBucketArrayOutput panics, workaround ApplyT targeting `*[]s3.BucketArray`
- BucketArrayOutput.Index panics (unusable)

After the change:
- BucketArray.ToBucketArrayOutput generates possibly corrupt values that may lead to hangs (investigating out of band)
- BucketArrayOutput.Index works correctly but can propagate the error from above

t0yv0 (approved):

pgavlin:
@emiliza just wanted to make sure that we're okay with the breaking change on this one

pgavlin:
cc @stack72 as well

t0yv0:
I've filed follow up as https://github.com/pulumi/pulumi/issues/8477 

New information: the scope of the remaining problem is thankfully a lot smaller than I thought. It affects URN property only, and is not related to Map/Array helpers. The problem is present on master and remains unchanged with this PR.

PR #8455, opened 2021-11-19
[backend/filestate] Don't unwrap go-cloud errors

### LINKED ISSUE #8461: --create flag does not work with pulumi stack select with a filestate backend in v3.18.0
## Hello!

## Issue details

Pulumi CLI fails with the following error when running pulumi stack select dev2 -c to select a stack but first creating it, if it does not exist already.

```
pulumi stack select dev2 -c --cwd C:\GitHub\temp-pulumi-stacks\pr-test\
error: failed to load checkpoint: blob (key ".pulumi/stacks/dev2.json") (code=NotFound): CreateFile C:\temp\.pulumi\stacks\dev2.json: The system cannot find the file specified.
```

### Steps to reproduce

1. Create a new project with a filestate backend
2. Run `pulumi stack select <a non-existent stack> --create` from the project folder

Expected: Stack is created, if it does not exist and then selected as the current stack
Actual: An error is encountered (see above)

praneetloke:
# Description

Pulumi CLI fails with the following error when running `pulumi stack select dev2 -c` to select a stack but first creating it, if it does not exist already.

```
pulumi stack select dev2 -c --cwd C:\GitHub\temp-pulumi-stacks\pr-test\
error: failed to load checkpoint: blob (key ".pulumi/stacks/dev2.json") (code=NotFound): CreateFile C:\temp\.pulumi\stacks\dev2.json: The system cannot find the file specified.
```

Fixes #8461 (Related to #8406)

## Checklist

### DIFF
--- a/pkg/backend/filestate/backend.go
+++ b/pkg/backend/filestate/backend.go
@@ -882,14 +882,3 @@ func (b *localBackend) UpdateStackTags(ctx context.Context,
 	// The local backend does not currently persist tags.
 	return errors.New("stack tags not supported in --local mode")
 }
-
-// Returns the original error in the chain. If `err` is nil, nil is returned.
-func drillError(err error) error {
-	e := err
-	for errors.Unwrap(e) != nil {
-		if u := errors.Unwrap(e); u != nil {
-			e = u
-		}
-	}
-	return e
-}
--- a/pkg/backend/filestate/backend.go
+++ b/pkg/backend/filestate/backend.go
@@ -323,7 +323,7 @@ func (b *localBackend) GetStack(ctx context.Context, stackRef backend.StackRefer
 	snapshot, path, err := b.getStack(stackName)
 
 	switch {
-	case gcerrors.Code(drillError(err)) == gcerrors.NotFound:
+	case gcerrors.Code(err) == gcerrors.NotFound:
 		return nil, nil
 	case err != nil:
 		return nil, err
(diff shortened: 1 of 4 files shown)


pkg/backend/filestate/backend.go
```diff
@@ -323,7 +323,7 @@ func (b *localBackend) GetStack(ctx context.Context, stackRef backend.StackRefer
 	snapshot, path, err := b.getStack(stackName)
 
 	switch {
-	case gcerrors.Code(drillError(err)) == gcerrors.NotFound:
+	case gcerrors.Code(err) == gcerrors.NotFound:
```
praneetloke, line 326: The problem seems to be that we are passing the unwrapped error to go-cloud's `Code` func which tries to find a matching error that it can assign as `gcerr.Error`. So when it is passed an already-unwrapped error, it fails return the error code that we expect to compare against.

praneetloke:
I didn't update the changelog, but I think this PR does warrant an entry.

pkg/backend/filestate/backend.go
justinvp: Ah, `gcerrors` itself is wrapping errors, so when we unwrap the chain, it prevents `gcerrors.Code` from finding its wrapped error. This is my bad, I had suggested unwrapping the full chain to @iwahbe since we are also wrapping, not realizing that `gcerrors` itself is wrapping and relying on it. Sorry about that @iwahbe.

iwahbe (approved):
Test without fix correctly show failure and this fixes the problem on my machine. I'm happy here.

justinvp (approved):

PR #8462, opened 2021-11-19
Add a unit test to resourcePtr branch

t0yv0:
# Description

Fixes # (issue)

## Checklist

### DIFF
--- a/pkg/codegen/internal/test/testdata/resource-property-overlap/go-extras/tests/go_test.go
+++ b/pkg/codegen/internal/test/testdata/resource-property-overlap/go-extras/tests/go_test.go
@@ -39,7 +39,7 @@ func TestArrayOutputIndex(t *testing.T) {
 
 		pulumi.All(u1, u2).ApplyT(func(all []interface{}) int {
 			urn1 := all[0].(pulumi.URN)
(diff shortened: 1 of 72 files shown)


pkg/codegen/internal/test/testdata/resource-property-overlap/go-extras/tests/go_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
+
+	"resource-property-overlap/example"
+)
+
+func TestArrayOutputIndex(t *testing.T) {
```
t0yv0, line 17: This actually times out. I can't figure it out why. @pgavlin 

Before your PR it was failing with a different error.

From what I can tell `example.RecArray{rec}.ToRecArrayOutput().Index(pulumi.Int(0)).` may be coming back with a *Rec that is not initialized as a resource, so it's URN is never populated.

Accidental struct copy?

t0yv0:
I figured it out I think.

PR #8467, opened 2021-11-19
Tidy up generator trees

### LINKED ISSUE #8464: FAIL: resource/stack TestSerializePropertyValue
## Hello!

## Issue details

Something to investigate (also failing on master): either a problem with the test, or we have regressed on serialization ability.

https://github.com/pulumi/pulumi/runs/4257430169?check_suite_focus=true. 

```
=== FAIL: resource/stack TestSerializePropertyValue (190.46s)
Error:     deployment_test.go:444: [rapid] failed after 88 tests: (*T).FailNow() called
        To reproduce, specify -run="TestSerializePropertyValue" -rapid.failfile="TestSerializePropertyValue-20211118222509-76075.fail" (or -rapid.seed=1314149388578197325)
        Failed test output:
```

### Steps to reproduce

1.
2.
3.
4.

Expected: 
Actual:

t0yv0:
# Description

Fixes #8464 

I played with #8464 repro which reproduces on master but I could not get the Go system to print the generated value or the error.

I was also noticing how slow it was generating the first example.

I also noticed that it doesn't reproduce if depth is lowered to 5.

The repro is basically this:

```
git co f98d39b84bc0e3a78a13dc09f80730e39a297b10 -b repro
cd pkg/resource/stack
go test -test.v -run="TestSerializePropertyValue" -rapid.seed=1314149388578197325

=== RUN   TestSerializePropertyValue
    deployment_test.go:444: [rapid] failed after 0 tests: (*T).FailNow() called
        To reproduce, specify -run="TestSerializePropertyValue" -rapid.failfile="TestSerializePropertyValue-20211119164007-93661.fail" (or -rapid.seed=1314149388578197325)
        Failed test output:
--- FAIL: TestSerializePropertyValue (105.05s)
FAIL

```

It is frying CPU for 105s.

It is quite concerning that I cannot print the value out. I am still playing with that a bit.

On a slight tangent, I found a way to generate recursive trees in Rapid that shrinks to smaller trees correctly and does not waste as many generator objects. Introducing this mechanically makes the test complete in <2s and I can no longer reproduce the problem. It might be possible that the reason for this is also accidentally lowering maxDepth by 1, but I think we want to take the change anyway.

## Checklist

### DIFF
--- a/sdk/go/common/resource/testing/rapid.go
+++ b/sdk/go/common/resource/testing/rapid.go
@@ -2,7 +2,7 @@
 package testing
 
 import (
-	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/assert"
 	"pgregory.net/rapid"
 
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
@@ -197,7 +197,7 @@ func StringPropertyGenerator() *rapid.Generator {
 func TextAssetGenerator() *rapid.Generator {
 	return rapid.Custom(func(t *rapid.T) *resource.Asset {
 		asset, err := resource.NewTextAsset(rapid.String().Draw(t, "text asset contents").(string))
-		require.NoError(t, err)
+		assert.NoError(t, err)
 		return asset
 	})
 }
@@ -223,7 +223,7 @@ func LiteralArchiveGenerator(maxDepth int) *rapid.Generator {
 		gen0: rapid.Custom(func(t *rapid.T) *resource.Archive {
 			contents := emptyContentsGen.Draw(t, "literal archive contents").(map[string]interface{})
 			archive, err := resource.NewAssetArchive(contents)
-			require.NoError(t, err)
+			assert.NoError(t, err)
 			return archive
 		}),
 		gen1: func(self *rapid.Generator) *rapid.Generator {
@@ -233,7 +233,7 @@ func LiteralArchiveGenerator(maxDepth int) *rapid.Generator {
 			return rapid.Custom(func(t *rapid.T) *resource.Archive {
 				content := contentsGen.Draw(t, "literal archive contents").(map[string]interface{})
 				archive, err := resource.NewAssetArchive(content)
-				require.NoError(t, err)
+				assert.NoError(t, err)
 				return archive
 			})
 		},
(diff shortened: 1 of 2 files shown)


t0yv0:
This modification still fails for me:

```
func TestSerializePropertyValue(t *testing.T) {
	gen := resource_testing.PropertyValueGenerator(6)
	rapid.Check(t, func(t *rapid.T) {
		gen.Draw(t, "property value")
	})
}
```

Basically it seems that `gen.Draw` is able to "quietly kill the test" and call `t.FailNow()` for some reason. This is not great, it is some issue in `rapid`. But at least it is reassuring that we do not have a failure of our property hiding in here.

pgavlin:
> Basically it seems that gen.Draw is able to "quietly kill the test" and call t.FailNow() for some reason. This is not great, it is some issue in rapid. But at least it is reassuring that we do not have a failure of our property hiding in here.

Have you tried running this under a debugger and setting a breakpoint on `runtime.gopanic`?

t0yv0:
Ha, well learning to use a debugger is still in my backlog..

pgavlin:
I generally use [`dlv`](https://github.com/go-delve/delve) directly from a terminal after building the test binary in debug mode:

```bash
> go test -c -gcflags="all=-N -l" .
> dlv exec ./foo.test -- -test.run TheTestToRun
(dlv) b runtime.gopanic
(dlv) c
```

t0yv0:
Ok yeah that worked. It points to LiteralArchiveGenerator using `require.NoError` where it probably shouldn't. Tinkering.

t0yv0:
OK @pgavlin this is better: https://github.com/pulumi/pulumi/pull/8469

sdk/go/common/resource/testing/rapid.go
```diff
@@ -351,64 +375,87 @@ func outputPropertyGenerator(ctx *StackContext, maxDepth int) *rapid.Generator {
...
+				arrayPropertyGenerator(self),
+				objectPropertyGenerator(self),
+				outputPropertyGenerator(ctx, self),
+				secretPropertyGenerator(self),
+			)
+		},
+	})
+}
 
-	if ctx == nil || len(ctx.Resources()) > 0 {
```
t0yv0, line 401: This should have stayed.

t0yv0:
Played with it a bit more, it's curious .. actually no improvement on shrinking behavior - the old code is also left-biased via OneOf so that works out well; but the other surprising part was that this version is slower. Go might like the short-lived or stack-allocated generators better. I's interesting. Anyway, closing, not an improvement.

t0yv0:
For future ref

```
go test -test.v -run="TestSerializePropertyValue" -rapid.seed=1314149388578197325
```

```
dlv test -- -test.v -run="TestSerializePropertyValue" -rapid.seed=1314149388578197325
```

Issue #8477, opened 2021-11-22
Converting a resource value X with .ToXOutput() loses URN information


t0yv0:
## Hello!

- Vote on this issue by adding a 👍 reaction
- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details

If a resource value gets passed through an `Output` helper type, specifically if it goes through a `pulumi.ToOutputWithContext`, it loses information about the resource URN.  The issue is specific to URN property. Other "normal" output properties seem to be propagated as one would expect. 

The issue is expected to affect every resource. 

The conditions where this was discovered are slightly synthetic, it is not yet clear which user scenarios this would affect.

### Steps to reproduce

```
package main

import (
	"github.com/pulumi/pulumi-aws/sdk/v4/go/aws/s3"
	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
)

func main() {
	pulumi.Run(func(ctx *pulumi.Context) error {
		bucket, err := s3.NewBucket(ctx, "my-bucket", nil)
		if err != nil {
			return err
		}

		ctx.Export("bucketUrn1", bucket.URN())
		urn2 := bucket.ToBucketOutput().
			ApplyT(func(x *s3.Bucket) pulumi.URNOutput {
				return x.URN()
			}).
			ApplyT(func(x interface{}) pulumi.URN {
				return x.(pulumi.URN)
			}).(pulumi.URNOutput)

		ctx.Export("bucketUrn2", urn2)
		return nil
	})
}
```

Then `pulumi up`.

Expected: exported the same URN in both bucket1Urn and bucket2Urn slots
Actual: `creating error: program failed: waiting for RPCs: unexpected unknown property value for "bucketUrn2"`

justinvp:
This no longer repros with:

```
	github.com/pulumi/pulumi-aws/sdk/v6 v6.2.1
	github.com/pulumi/pulumi/sdk/v3 v3.86.0
```

```sh
$ pulumi up
Previewing update (dev)

View in Browser (Ctrl+O): url

     Type                 Name              Plan
 +   pulumi:pulumi:Stack  proj-dev  create
 +   └─ aws:s3:Bucket     my-bucket         create


Outputs:
    bucketUrn1: "urn:pulumi:dev::proj::aws:s3/bucket:Bucket::my-bucket"
    bucketUrn2: "urn:pulumi:dev::proj::aws:s3/bucket:Bucket::my-bucket"

Resources:
[...]

PR #8479, opened 2021-11-22
Add a unit test to check Array/Index

t0yv0:
# Description

Adds a unit test  to check that `XArray{x}.ToXArrayOutput().Index(pulumi.Int(0)) == x` over the generated code.

Fixes # (issue)

## Checklist


t0yv0:
@pgavlin in this form it's actually passing. Consider adding, if not I'll follow up to add as a separate PR once yours lands on master. My first attempt failed because it ran into https://github.com/pulumi/pulumi/issues/8477 which is unrelated as far as I can see.

PR #8488, opened 2021-11-24
Specify fields on engine diff

### LINKED ISSUE #8487: ReplaceOnChanges option isn't applied if provider's diff is unknown
The AWS Native provider is somewhat simplistic in diff calculation: it returns `DiffResponse_DIFF_UNKNOWN` and relies on the engine to calculate the actual detailed diff (which is a supported scenario, if I read the docs correctly):

https://github.com/pulumi/pulumi-aws-native/blob/dc7de1c4676854d8cbaaf7a475112f125e52fe9b/provider/pkg/provider/provider.go#L582-L585

It works for the UI experience, but apparently it doesn’t work for `replaceOnChanges` option that users set in their programs as reported in https://github.com/pulumi/pulumi-aws-native/issues/240 

I reproduced it with basic TS and Python programs, e.g.

```ts
import * as aws from "@pulumi/aws-native";

const logGroup = new aws.logs.LogGroup("log-test", {
    retentionInDays: 30,
}, { replaceOnChanges: ["retentionInDays"]});
```

Deploy, change the value of `retentionInDays`, run preview.

I'm advocating for the possibility of having a diff by default, so that new providers don’t have to dive into the details of diff calculation unless they have to. I understand that the engine doesn’t have all the information that the provider could use - but I also think the engine can be a good-enough baseline.

iwahbe:
# Description

Computes the specific keys for an engine diff. 

Fixes #8487 

## Checklist


t0yv0:
Shall we add some tests?

CHANGELOG_PENDING.md
```diff
@@ -8,5 +8,9 @@
 
 ### Bug Fixes
 
+- [cli/engine] - Accuretly computes the fields changed when diffing with unhelpful providers. This
```
Frassle, line 11:
```suggestion
- [cli/engine] - Accurately computes the fields changed when diffing with unhelpful providers. This
```

Frassle (commented):
A test to check that this does fix #8487 would be nice.

sdk/go/common/resource/plugin/provider.go
```diff
@@ -217,6 +217,34 @@ type DiffResult struct {
...
+	result := DiffResult{}
+	if !obj.AnyChanges() {
+		result.Changes = DiffNone
+		return result
+	}
+
+	ignoreSet := make(map[string]bool, len(ignoreChanged))
+	for _, v := range ignoreChanged {
+		ignoreSet[v] = true
+	}
```
pgavlin, line 234: I don't think that this is going to do quite what you want. The comment on `Goal.IgnoreChanges` is unfortunately misleading (PR to update it is [here](https://github.com/pulumi/pulumi/pull/8491)): its contents are a list of property paths, not just a list of top-level property keys. [`processIgnoreChanges`](https://github.com/pulumi/pulumi/blob/master/pkg/resource/deploy/step_generator.go#L1164-L1201) has an example of how to work with these paths. I think that the most straightforward way to get this right is to call `processIgnoreChanges` _again_ on `oldInputs` and `newInputs` prior to computing the diff (`processIgnoreChanges` has already been run once using the original inputs from the program; rerunning is necessary b/c `newInputs` is the result of `Check` and may differ from the original inputs due to e.g. default application by the provider). This will also ensure that we don't return `DiffSome` if all changes are ignored.

pgavlin (changes requested):

sdk/go/common/resource/plugin/provider.go
```diff
@@ -217,6 +218,53 @@ type DiffResult struct {
 	DeleteBeforeReplace bool                    // if true, this resource must be deleted before recreating it.
 }
 
+// Computes the detailed diff of Updated keys.
+func NewDetailedDiffFromObjectDiff(diff *resource.ObjectDiff) map[string]PropertyDiff {
+	return objectDiffToDetailsDiff("", diff)
+}
+
+func objectDiffToDetailsDiff(prefix string, diff *resource.ObjectDiff) map[string]PropertyDiff {
```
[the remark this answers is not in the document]
pgavlin: Also, I think that if we're going to use this function in product code, we need to flesh it out to incorporate adds and deletes as well. It was limited to test code before b/c it is not complete.

pgavlin (commented):
I like the direction, but I think we need to flesh out `objectDiffToDetailedDiff`

sdk/go/common/resource/plugin/provider.go
```diff
@@ -217,6 +218,85 @@ type DiffResult struct {
...
 
+// Computes the detailed diff of Updated, Added and Deleted keys.
+func NewDetailedDiffFromObjectDiff(diff *resource.ObjectDiff) map[string]PropertyDiff {
+	if diff == nil {
+		return map[string]PropertyDiff{}
+	}
+	return objectDiffToDetailedDiff("", diff)
+}
+
+func objectDiffToDetailedDiff(prefix string, diff *resource.ObjectDiff) map[string]PropertyDiff {
```
pgavlin, line 229: Nit: I think passing in an accumulator for the result in these functions will simplify the code by removing the need for the various copy loops.

```suggestion
func objectDiffToDetailedDiff(result map[string]PropertyDiff, prefix string, diff *resource.ObjectDiff) {
```

pkg/resource/deploy/step_generator.go
```diff
@@ -1132,10 +1132,17 @@ func diffResource(urn resource.URN, id resource.ID, oldInputs, oldOutputs,
 		return diff, err
 	}
 	if diff.Changes == plugin.DiffUnknown {
-		if oldInputs.DeepEquals(newInputs) {
```
pgavlin, line 1135: I'd consider adding some tests to `sdk/go/common/resource` to ensure that `PropertyValue.Diff` always returns a nil diff if `PropertyValue.DeepEquals` returns `true` and always returns a non-nil diff if `PropertyValue.Equals` returns `false`.

pgavlin (approved):
LGTM overall. One suggestion for a simplification + one suggestion for some extra testing.

PR #8492, opened 2021-11-24
Fixes 8403 name conflicts in Go codegen

### LINKED ISSUE #8403: Name collision on Go codegen code prevents pulumi-mongodbatlas upgrade
## Hello!

## Issue details

https://github.com/pulumi/pulumi-mongodbatlas/runs/4182589280?check_suite_focus=true

Auto-naming gives a function result type the XResult name, but it also gives a `results:` property type an `XResult` name; in TypeScript we get two similarly named types in distinct modules, but in Go they conflict over a slot in the shared namespace and do not compile:

```

export function getCloudProviderSnapshotRestoreJobs(..): Promise<GetCloudProviderSnapshotRestoreJobsResult> {

export interface GetCloudProviderSnapshotRestoreJobsResult {
    ...
    readonly results: outputs.GetCloudProviderSnapshotRestoreJobsResult[];
}


outputs.ts:

export interface GetCloudProviderSnapshotRestoreJobsResult {
   ...
}
```

This is a problem with codegen but apparently has a low impact as 29/30 provider upgrades did not hit it.

### Steps to reproduce

1.
2.
3.
4.

Expected: 
Actual:

t0yv0:
# Description

Fixes #8403

```go

func GetCustomDbRoles(ctx *pulumi.Context, args *GetCustomDbRolesArgs, opts ...pulumi.InvokeOption) (*GetCustomDbRolesResult, error)

// -->

func LookupCustomDbRoles(ctx *pulumi.Context, args *LookupCustomDbRolesArgs, opts ...pulumi.InvokeOption) (*LookupCustomDbRolesResult, error)

// A collection of values returned by getCustomDbRoles.
type LookupCustomDbRolesResult struct {
	// The provider-assigned unique ID for this managed resource.
	Id        string `pulumi:"id"`
	ProjectId string `pulumi:"projectId"`
	// A list where each represents a custom db roles.
	Results []GetCustomDbRolesResult `pulumi:"results"` // Now GetCustomDbRolesResult != LookupCustomDbRolesResult
}

```
## Checklist


pkg/codegen/go/gen.go
```diff
@@ -3124,15 +3126,9 @@ func generatePackageContextMap(tool string, pkg *schema.Package, goInfo GoPackag
 
 		if f.Inputs != nil {
 			pkg.names.Add(name + "Args")
-			if originalName != name {
```
t0yv0, line 3127: I believe this code was not doing what it appears to have been doing. Functions generate helper FArgs and FResult types but those types currently do not consult pkg.renamed tables, they continue generating as FArgs/FResult, but F the function name itself does consult those tables. So what this actually was doing, it would rename an unrelated type to collide with FArgs/FResult if there was one. Let's see what the downstream checks say, I expect there shouldn't be diffs based on this removal

pkg/codegen/go/gen.go
```diff
@@ -3110,8 +3110,10 @@ func generatePackageContextMap(tool string, pkg *schema.Package, goInfo GoPackag
 		pkg.functions = append(pkg.functions, f)
 
 		name := tokenToName(f.Token)
-		originalName := name
-		if pkg.names.Has(name) {
+
+		if pkg.names.Has(name) ||
```
t0yv0, line 3114: Renaming to kick in not only based on function name conflict, but also on FResult name conflict (relevant for the bug) and FArgs thrown in.

pkg/codegen/go/gen.go
```diff
@@ -3110,8 +3110,10 @@ func generatePackageContextMap(tool string, pkg *schema.Package, goInfo GoPackag
...
+			pkg.names.Has(name+"Args") ||
+			pkg.names.Has(name+"Result") {
 			switch {
```
iwahbe, line 3117: I think it would make sense to add a default case to the switch:
```go
default:
    return fmt.Errorf("failed to acquire unique name for %s", name)
```
Right now it looks like this can fail silently.

iwahbe (approved):
Besides one comment, LGTM.

pkg/codegen/go/gen.go
t0yv0: Afraid that might break working programs if conflict detection is inaccurate (i.e. the conflicting type didn't really get generated). Perhaps superstitious.. A warning would be best. I'll look back at this after reInvent.

PR #8494, opened 2021-11-24
Run validate function even if dryRun=true

Frassle:
Hit this while trying to add some validation checks to runs of Preview while using plans. Seems one test actually was assuming this was the case already and just hasn't been running it's expected validate function.

### DIFF
--- a/pkg/engine/lifeycletest/test_plan.go
+++ b/pkg/engine/lifeycletest/test_plan.go
@@ -96,11 +96,11 @@ func (op TestOp) RunWithContext(
 	_, res := op(info, ctx, opts, dryRun)
 	contract.IgnoreClose(journal)
 
-	if dryRun {
-		return nil, res
-	}
 	if validate != nil {
 		res = validate(project, target, journal.Entries(), firedEvents, res)
+	}
+	if dryRun {
+		return nil, res
 	}
 
 	snap := journal.Snap(target.Snapshot)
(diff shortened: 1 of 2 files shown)


pkg/engine/lifeycletest/test_plan.go
```diff
@@ -96,12 +96,12 @@ func (op TestOp) RunWithContext(
 	_, res := op(info, ctx, opts, dryRun)
```
iwahbe, line 96: Do we need to check `res`? We didn't before, but it seems like we should.

Frassle: No we feed it directly into validate which should check it (or validate is nil in which case we return it)

pkg/engine/lifeycletest/test_plan.go
```diff
@@ -210,7 +210,8 @@ func (p *TestPlan) Run(t *testing.T, snapshot *deploy.Snapshot) *deploy.Snapshot
 		if !step.SkipPreview {
 			previewSnap := CloneSnapshot(t, snap)
 			previewTarget := p.GetTarget(previewSnap)
-			_, res := step.Op.Run(project, previewTarget, p.Options, true, p.BackendClient, step.Validate)
+			// Don't run validate on the preview step
+			_, res := step.Op.Run(project, previewTarget, p.Options, true, p.BackendClient, nil)
```
pgavlin, line 214: Wondering if we should add a `dryRun`/`preview` flag to the validator and then allow it to hit the preview here?

pgavlin (approved):
LGTM; one question re: running the validator during a step's preview.

pkg/engine/lifeycletest/test_plan.go
Frassle: Could do, but nothing would use it right now. Would just result in a lot of `if dryRun { return }` lines added to the current validator set.

PR #8498, opened 2021-11-26
[sdk/dotnet] Improve collection initializers

### LINKED ISSUE #8497: [sdk/dotnet] Improve collection initializers
## Hello!

## Issue details

The current [`InputList`](https://github.com/pulumi/pulumi/blob/4d4ff9f1d6885e374b0113e98b7d015a52753ac8/sdk/dotnet/Pulumi/Core/InputList.cs#L47) and [`InputMap`](https://github.com/pulumi/pulumi/blob/4d4ff9f1d6885e374b0113e98b7d015a52753ac8/sdk/dotnet/Pulumi/Core/InputMap.cs#L43) collection initializers are clunky to work with when combining different sources of values.

With a few tweaks we can exploit the existing conversion operators to get much more flexibility and convenient usability.

gitfool:
Fixes #8497. See the [InputMapCollectionInitializers](https://github.com/pulumi/pulumi/blob/af99d3cbe087d38f40a64e7d50701cb197c3e8c2/sdk/dotnet/Pulumi.Tests/Core/InputTests.cs#L49) and [InputListCollectionInitializers](https://github.com/pulumi/pulumi/blob/af99d3cbe087d38f40a64e7d50701cb197c3e8c2/sdk/dotnet/Pulumi.Tests/Core/InputTests.cs#L83) tests for examples of flexibility and usability.

### DIFF
--- a/sdk/dotnet/Pulumi.Automation/Pulumi.Automation.xml
+++ b/sdk/dotnet/Pulumi.Automation/Pulumi.Automation.xml
@@ -1370,34 +1370,12 @@
             <param name="force">A boolean indicating whether the deletion should be forced.</param>
             <param name="cancellationToken">A cancellation token.</param>
         </member>
-        <member name="M:Pulumi.Automation.WorkspaceStackState.UnprotectAsync(System.String)">
-            <summary>
-            Unprotect a resource in a stack's state.
-            This command clears the ‘protect’ bit on the provided resource <paramref name="urn"/>, allowing the resource to be deleted.
-            </summary>
-            <param name="urn">The Pulumi URN to be unprotected.</param>
-        </member>
         <member name="M:Pulumi.Automation.WorkspaceStackState.UnprotectAsync(System.String,System.Threading.CancellationToken)">
             <summary>
             Unprotect a resource in a stack's state.
             This command clears the ‘protect’ bit on the provided resource <paramref name="urn"/>, allowing the resource to be deleted.
             </summary>
             <param name="urn">The Pulumi URN to be unprotected.</param>
-            <param name="cancellationToken">A cancellation token.</param>
-        </member>
-        <member name="M:Pulumi.Automation.WorkspaceStackState.UnprotectAsync(System.Collections.Generic.IEnumerable{System.String})">
-            <summary>
-            Unprotect resources in a stack's state.
-            This command clears the ‘protect’ bit on the provided resource <paramref name="urns"/>, allowing those resources to be deleted.
-            </summary>
-            <param name="urns">The Pulumi URNs to be unprotected.</param>
-        </member>
-        <member name="M:Pulumi.Automation.WorkspaceStackState.UnprotectAsync(System.Collections.Generic.IEnumerable{System.String},System.Threading.CancellationToken)">
-            <summary>
-            Unprotect resources in a stack's state.
-            This command clears the ‘protect’ bit on the provided resource <paramref name="urns"/>, allowing those resources to be deleted.
-            </summary>
-            <param name="urns">The Pulumi URNs to be unprotected.</param>
             <param name="cancellationToken">A cancellation token.</param>
         </member>
         <member name="M:Pulumi.Automation.WorkspaceStackState.UnprotectAllAsync(System.Threading.CancellationToken)">
--- a/sdk/dotnet/Pulumi.Tests/Core/InputTests.cs
+++ b/sdk/dotnet/Pulumi.Tests/Core/InputTests.cs
@@ -1,5 +1,6 @@
 ﻿// Copyright 2016-2019, Pulumi Corporation
 
+using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Threading.Tasks;
 using Xunit;
@@ -45,6 +46,22 @@ namespace Pulumi.Tests.Core
             });
 
         [Fact]
+        public Task InputMapCollectionInitializers()
+            => RunInPreview(async () =>
+            {
+                var map = new InputMap<string>
+                {
+                    { "K1", "V1" },
+                    { "K2", Output.Create("V2") },
+                    new Dictionary<string, string> { { "K3", "V3" }, { "K4", "V4"} },
+                    Output.Create(new Dictionary<string, string> { ["K5"] = "V5", ["K6"] = "V6" }.ToImmutableDictionary())
+                };
+                var data = await map.ToOutput().DataTask.ConfigureAwait(false);
+                Assert.Equal(6, data.Value.Count);
+                Assert.Equal(new Dictionary<string, string> { ["K1"] = "V1", ["K2"] = "V2", ["K3"] = "V3", ["K4"] = "V4", ["K5"] = "V5", ["K6"] = "V6" }, data.Value);
+            });
+
+        [Fact]
         public Task InputMapUnionInitializer()
             => RunInPreview(async () =>
             {
@@ -61,7 +78,24 @@ namespace Pulumi.Tests.Core
                 Assert.True(data.Value.ContainsValue("testValue"));
                 Assert.True(data.Value.ContainsValue(123));
             });
-        
+
+        [Fact]
+        public Task InputListCollectionInitializers()
+            => RunInPreview(async () =>
+            {
+                var list = new InputList<string>
+                {
+                    "V1",
+                    Output.Create("V2"),
+                    new[] { "V3", "V4" },
+                    new List<string> { "V5", "V6" },
+                    Output.Create(ImmutableArray.Create("V7", "V8"))
+                };
+                var data = await list.ToOutput().DataTask.ConfigureAwait(false);
+                Assert.Equal(8, data.Value.Length);
+                Assert.Equal(new[] { "V1", "V2", "V3", "V4", "V5", "V6", "V7", "V8" }, data.Value);
+            });
+
         [Fact]
         public Task InputListUnionInitializer()
             => RunInPreview(async () =>
--- a/sdk/dotnet/Pulumi/Core/InputList.cs
+++ b/sdk/dotnet/Pulumi/Core/InputList.cs
@@ -66,13 +66,6 @@ namespace Pulumi
         }
 
         /// <summary>
-        /// Concatenates the values in this list with the value in <paramref name="other"/>,
-        /// returning the concatenated sequence in a new <see cref="InputList{T}"/>.
-        /// </summary>
-        public InputList<T> Concat(Input<T> other)
-            => Output.Concat(_outputValue, other);
-
-        /// <summary>
         /// Concatenates the values in this list with the values in <paramref name="other"/>,
         /// returning the concatenated sequence in a new <see cref="InputList{T}"/>.
         /// </summary>
--- a/sdk/dotnet/Pulumi/Core/InputList.cs
+++ b/sdk/dotnet/Pulumi/Core/InputList.cs
@@ -55,29 +55,29 @@ namespace Pulumi
         {
         }
 
-        public void Add(Input<T> value)
+        public void Add(params Input<T>[] inputs)
         {
-            _outputValue = Concat(value);
+            _outputValue = Concat(inputs);
         }
 
-        public void Add(InputList<T> values)
+        public void Add(InputList<T> inputs)
         {
-            _outputValue = Concat(values);
+            _outputValue = Concat(inputs);
         }
 
         /// <summary>
-        /// Concatenates the values in this list with the given <paramref name="value"/>,
+        /// Concatenates the values in this list with the value in <paramref name="other"/>,
         /// returning the concatenated sequence in a new <see cref="InputList{T}"/>.
         /// </summary>
-        public InputList<T> Concat(Input<T> value)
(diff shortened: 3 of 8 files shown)


sdk/dotnet/Pulumi/Core/InputList.cs
```diff
@@ -55,19 +55,29 @@ private InputList(Output<ImmutableArray<T>> values)
         {
         }
 
-        public void Add(params Input<T>[] inputs)
```
gitfool, line 58: This never worked as intended; `params` does not avoid multiple invocations for each elements, so I removed it.

sdk/dotnet/Pulumi.Automation/Pulumi.Automation.xml
```diff
@@ -1370,13 +1370,6 @@
             <param name="force">A boolean indicating whether the deletion should be forced.</param>
             <param name="cancellationToken">A cancellation token.</param>
         </member>
-        <member name="M:Pulumi.Automation.WorkspaceStackState.UnprotectAsync(System.String)">
-            <summary>
-            Unprotect a resource in a stack's state.
-            This command clears the ‘protect’ bit on the provided resource <paramref name="urn"/>, allowing the resource to be deleted.
-            </summary>
-            <param name="urn">The Pulumi URN to be unprotected.</param>
-        </member>
```
gitfool, line 1379: Unrelated changes that are made automatically and should have been previously committed.

sdk/dotnet/Pulumi.Automation/Pulumi.Automation.xml
```diff
@@ -1385,21 +1378,6 @@
...
-            <param name="urns">The Pulumi URNs to be unprotected.</param>
-        </member>
-        <member name="M:Pulumi.Automation.WorkspaceStackState.UnprotectAsync(System.Collections.Generic.IEnumerable{System.String},System.Threading.CancellationToken)">
-            <summary>
-            Unprotect resources in a stack's state.
-            This command clears the ‘protect’ bit on the provided resource <paramref name="urns"/>, allowing those resources to be deleted.
-            </summary>
-            <param name="urns">The Pulumi URNs to be unprotected.</param>
-            <param name="cancellationToken">A cancellation token.</param>
-        </member>
```
gitfool, line 1402: Unrelated changes that are made automatically and should have been previously committed.

sdk/dotnet/Pulumi/Core/InputList.cs
```diff
@@ -55,19 +55,29 @@ private InputList(Output<ImmutableArray<T>> values)
...
+        public void Add(Input<T> value)
         {
-            // Make an Output from the values passed in, mix in with our own Output, and combine
-            // both to produce the final array that we will now point at.
-            _outputValue = Output.Concat(_outputValue, Output.All(inputs));
+            _outputValue = Concat(value);
         }
 
+        public void Add(InputList<T> values)
```
gitfool, line 63: Exploits existing conversion operators for collection initializers.

sdk/dotnet/Pulumi/Core/InputMap.cs
```diff
@@ -55,7 +55,14 @@ public void Add(string key, Input<V> value)
         {
             var inputDictionary = (Input<ImmutableDictionary<string, V>>)_outputValue;
             _outputValue = Output.Tuple(inputDictionary, value)
-                                 .Apply(x => x.Item1.Add(key, x.Item2));
+                .Apply(x => x.Item1.Add(key, x.Item2));
+        }
+
+        public void Add(InputMap<V> values)
```
gitfool, line 61: Exploits existing conversion operators for collection initializers.

sdk/dotnet/Pulumi/PublicAPI.Unshipped.txt
```diff
@@ -1,2 +0,0 @@
-Pulumi.ResourceOptions.ReplaceOnChanges.get -> System.Collections.Generic.List<string>
-Pulumi.ResourceOptions.ReplaceOnChanges.set -> void
```
gitfool, line 2: Unrelated changes that should have been previously moved from unshipped to shipped.

gitfool: I added my public api changes directly to shipped to avoid such artifacts.

sdk/dotnet/Pulumi/Core/InputList.cs
justinvp, line 58: Changing this is a binary breaking change. Let's keep the existing params overload.

sdk/dotnet/Pulumi/Core/InputList.cs
```diff
@@ -55,19 +55,29 @@ private InputList(Output<ImmutableArray<T>> values)
...
+
         /// <summary>
-        /// Concatenates the values in this list with the values in <paramref name="other"/>,
+        /// Concatenates the values in this list with the given <paramref name="values"/>,
         /// returning the concatenated sequence in a new <see cref="InputList{T}"/>.
         /// </summary>
-        public InputList<T> Concat(InputList<T> other)
-            => Output.Concat(_outputValue, other._outputValue);
+        public InputList<T> Concat(InputList<T> values)
+            => Output.Concat(_outputValue, values._outputValue);
```
justinvp, line 80: Please revert the parameter name change. While it's unlikely to impact anyone, it is technically a breaking change.

gitfool:
@justinvp I've applied your suggested changes.

justinvp:
Thanks, @gitfool!

I don't know what the original intention was behind the use of `params` in `InputList<T>.Add`. Adding a collection to an existing collection via `Add` (and therefore also collection initializers) is a little unusual (I'd expect to do that through an `AddRange` method). That said, I do see how being able to add various types of collections as part of initializing a collection is useful for Pulumi in particular (thanks for the examples in the tests that demonstrate this), and we already allow doing this through the `Add(params)`, so I'm on board with these changes. cc: @Frassle and @mikhailshilkov in case they have any other thoughts.

gitfool:
Note: I do have a practical case in mind that lead me to submitting this - succinctly combining dependencies in their various forms.

justinvp (approved):

sdk/dotnet/Pulumi/Core/InputList.cs
```diff
@@ -57,11 +57,21 @@ private InputList(Output<ImmutableArray<T>> values)
...
+        public void Add(InputList<T> inputs)
+        {
+            _outputValue = Concat(inputs);
+        }
+
+        /// <summary>
+        /// Concatenates the values in this list with the value in <paramref name="other"/>,
+        /// returning the concatenated sequence in a new <see cref="InputList{T}"/>.
+        /// </summary>
+        public InputList<T> Concat(Input<T> other)
```
Frassle, line 72: I'd normally (and at least in dotnet F# it seems this way) use Concat for two sequence parameters, and Add/Append for a sequence and a single value.

sdk/dotnet/Pulumi/Core/Output.cs
```diff
@@ -101,8 +101,11 @@ public static Output<string> Format(FormattableString formattableString)
                 string.Format(formattableString.Format, objs.ToArray()));
         }
 
+        internal static Output<ImmutableArray<T>> Concat<T>(Output<ImmutableArray<T>> values, Output<T> value)
```
Frassle, line 104: Again I would think this would be better named Append

sdk/dotnet/Pulumi.Tests/Core/InputTests.cs
```diff
@@ -44,6 +45,22 @@ public Task MergeInputMaps()
...
 
+        [Fact]
+        public Task InputMapCollectionInitializers()
+            => RunInPreview(async () =>
+            {
+                var map = new InputMap<string>
+                {
+                    { "K1", "V1" },
+                    { "K2", Output.Create("V2") },
+                    new Dictionary<string, string> { { "K3", "V3" }, { "K4", "V4"} },
```
Frassle, line 56: I find it slightly strange that we're allowing the adding of a collection here? None of the standard dotnet collections support this use of collection initializers afaik.

sdk/dotnet/Pulumi/Core/InputList.cs
```diff
@@ -57,11 +57,21 @@ private InputList(Output<ImmutableArray<T>> values)
 
         public void Add(params Input<T>[] inputs)
         {
-            // Make an Output from the values passed in, mix in with our own Output, and combine
-            // both to produce the final array that we will now point at.
-            _outputValue = Output.Concat(_outputValue, Output.All(inputs));
+            _outputValue = Concat(inputs);
         }
 
+        public void Add(InputList<T> inputs)
```
Frassle, line 63: I would expect this to be called AddRange or Concat but I'm guessing you've named it Add explicitly so you can use it in initializer statements. As I said above I think that's a little unusual compared to the normal dotnet collections.

sdk/dotnet/Pulumi/Core/InputMap.cs
Frassle: Again I'd expect this to be called AddRange.

t0yv0:
@justinvp checking in on this, can we merge this as-is or should take some changes first? I'm on board with the changes for the functionality, I don't index on the C# being idiomatic as highly, so long as this is discoverable.

sdk/dotnet/Pulumi/Core/InputList.cs
gitfool: Happy to rename things if that helps.

sdk/dotnet/Pulumi/Core/Output.cs
gitfool: Happy to rename things if that helps.

sdk/dotnet/Pulumi.Tests/Core/InputTests.cs
gitfool: Yes, this is more about making the collection initializers more user friendly than about following other dotnet conventions. These collections are used extensively by resources so it helps a lot with inline resource instantiation and therefore simplifies dotnet code using pulumi libraries.

sdk/dotnet/Pulumi/Core/InputList.cs
gitfool: Yes, this is more about making the collection initializers more user friendly than about following other dotnet conventions. These collections are used extensively by resources so it helps a lot with inline resource instantiation and therefore simplifies dotnet code using pulumi libraries.

sdk/dotnet/Pulumi/Core/InputList.cs
gitfool: This can actually be removed with the `params` signature restored since it uses the other overload.

sdk/dotnet/Pulumi/Core/Output.cs
gitfool: This can actually be removed with the params signature restored since it uses the other overload.

sdk/dotnet/Pulumi/Core/InputList.cs
gitfool: (Which highlights the mismatch from using `params` since it's called multiple times by the collection initializer, converted to an `InputList` from an array of _one_ item, and then passed to `Concat`.)

sdk/dotnet/Pulumi/Core/InputList.cs
gitfool: Changed to delegate to `AddRange` so intention is clearer.

gitfool:
I made some follow-up changes to hopefully address concerns but still augment the collection initializers, albeit in a non-standard way, bearing in mind the extensive use of these collections.

gitfool:
@t0yv0 the many acceptance test failures look unrelated to this change.

sdk/dotnet/Pulumi.Tests/Core/InputTests.cs
Frassle: Do you have any examples of real world uses cases using this. We're not against adding a non-standard Add method but we'd like to see that it is genuinely useful to have. If there's anything in pulumi/examples, or a new example we could add there to show off the wins of having these methods that would be really good.

gitfool: Yes, mainly with kubernetes resources with multiple dependencies, e.g.:
```c#
var certManagerRelease = new Release("cert-manager",
    new ReleaseArgs
    {
        Namespace = "cert-manager",
        Name = "cert-manager",
        RepositoryOpts = new RepositoryOptsArgs { Repo = "https://charts.jetstack.io" },
        Chart = "cert-manager",
        Version = K8sConfig.CertManagerChartVersion,
        Values = certManagerValues,
        Atomic = true,
        SkipCrds = true
    },
    new CustomResourceOptions { DependsOn = { certManagerCrds.Ready(), certManagerNs, kubePrometheusStackCrds.Ready() }, Provider = k8sProvider });
```
This does not compile without my proposed changes.

t0yv0:
Thank you for this contribution 🎉 

It got accepted via https://github.com/pulumi/pulumi/pull/8582

PR #8507, opened 2021-11-29
Allow interpolating plugin custom server URLs

### LINKED ISSUE #8263: Allow interpolating `--server`/`pluginDownloadURL` to support GitHub releases and other plugin hosts
Pulumi supports a `--server` flag to `pulumi plugin install` and a `pluginDownloadUrl` in a pakcages schema.json that implicitly sets the `--server` flag for automatic plugin acquisition associated with that packge.

Today, this server URL is composed with the target plugin name in a fixed way.  For example - to use https://github.com/sacloud/pulumi-sakuracloud, the `--server` flag must be passed `--server https://github.com/sacloud/pulumi-sakuracloud/releases/download/0.6.0`, and then binaries with the right names must appear there (https://github.com/sacloud/pulumi-sakuracloud/releases/tag/0.6.0) like `pulumi-resource-sakuracloud-v0.6.0-darwin-amd64.tar.gz`.

The problem is that the download URL must be a fixed path, at which all of these binaries reside.  For GitHub releases, as seen in the example above, these binaries live in a different location depending on the release version.  This means that the schema.json must be updated every time a new version is published to point at a new location for the plugin (and there is a chicken and egg here, that this must be updated before the binary is built and published there).

It would be better if the variables used to build the binary name could also be interpolated into the server URL.  For example, if the above could be specified as:

```
--server https://github.com/sacloud/pulumi-sakuracloud/releases/download/${VERSION}
# and
"pluginDownloadURL": "https://github.com/sacloud/pulumi-sakuracloud/releases/download/${VERSION}"
```

VERSION is the most important, but ARCH and OS might also be important in case binaries are generated into different locations based on these (though that should be rare?).  PACKAGE name in principle could also be provided, though should be fixed for any use of this, so not clear it would ever be necessary to interpolate this.

justinvp:
A custom server URL can contain `${VERSION}`, `${OS}`, or `${ARCH}` (case sensitive) in the URL and these values will be interpolated when downloading the plugin.

Fixes #8263

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -6,6 +6,8 @@
   these types virtually unusable in practice.
   [#8449](https://github.com/pulumi/pulumi/pull/8449)
 
+- Allow interpolating plugin custom server URLs.
+  [#8507](https://github.com/pulumi/pulumi/pull/8507)
 ### Bug Fixes
 
 - [codegen/go] - Respect default values in Pulumi object types.
--- a/sdk/go/common/workspace/plugins.go
+++ b/sdk/go/common/workspace/plugins.go
@@ -207,6 +207,14 @@ func (info *PluginInfo) SetFileMetadata(path string) error {
 	return nil
 }
 
+func interpolateURL(serverURL string, version semver.Version, os, arch string) string {
+	replacer := strings.NewReplacer(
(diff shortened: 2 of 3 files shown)


iwahbe (approved):
This looks clean. Assuming tests pass, LGTM

PR #8512, opened 2021-11-30
Move `PULUMI_ROOT` to `$HOME/.pulumi-dev`

### LINKED ISSUE #8509: Move `PULUMI_ROOT` to `pulumi/pulumi/bin`
## Hello!

## Issue details

Build outputs into ./bin/ in the local repo (and .gitignore it).

This will require changing the default Pulumi root and the contributing.md file. 

### Affected area/feature

build system.

iwahbe:
# Description

Fixes #8509

## Checklist

### DIFF
--- a/pkg/testing/integration/program.go
+++ b/pkg/testing/integration/program.go
@@ -2008,16 +2008,8 @@ func (pt *ProgramTester) prepareDotNetProject(projinfo *engine.Projinfo) error {
 
 	localNuget := os.Getenv("PULUMI_LOCAL_NUGET")
 	if localNuget == "" {
-		// We assume that the test is run from a subdirectory of Pulumi. We then search backwards
-		// until we are in a directory called Pulumi.
-		dir, err := os.Getwd()
-		if err != nil {
-			return err
-		}
-		for dir != "" && filepath.Base(dir) != "pulumi" {
-			dir = filepath.Dir(dir)
-		}
-		localNuget = filepath.Join(dir, ".bin", "nuget")
+		home := os.Getenv("HOME")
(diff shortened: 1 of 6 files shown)


pkg/testing/integration/program.go
```diff
@@ -2008,8 +2008,16 @@ func (pt *ProgramTester) prepareDotNetProject(projinfo *engine.Projinfo) error {
 
 	localNuget := os.Getenv("PULUMI_LOCAL_NUGET")
 	if localNuget == "" {
-		home := os.Getenv("HOME")
-		localNuget = filepath.Join(home, ".pulumi", "nuget")
+		// We assume that the test is run from a subdirectory of Pulumi. We then search backwards
+		// until we are in a directory called Pulumi.
```
justinvp, line 2012: I'm not sure we can make this assumption. The `integration` package is used by other repos, and potentially repos not in the github.com/pulumi org. I guess in those cases `PULUMI_LOCAL_NUGET` would have to be set?

iwahbe: I agree that the assumption is unsafe. We also have the scenario where someone hacks on Pulumi by putting it into a `pulumi-dev` (or `Pulumi`) folder. The tests will fail there as well. The problem is that we want to put `PULUMI_ROOT` (and thus `PULUMI_LOCAL_NUGET`) in a relative directory. 

Off of the top of my head, I can't think of a way to avoid something like this and still use a relative path. As an alternative, we could use `$HOME/.pulumi/.dev` as `PULUMI_ROOT`. This would avoid this questionable lookup and also allow maintaining both a development and released instance of pulumi.

justinvp (approved):
LGTM

pgavlin (approved):

PR #8515, opened 2021-11-30
[sdk/python] Check for `pulumiplugin.json` on package lookup

### LINKED ISSUE #7243: Python: Look for non `pulumi-` prefixed packages in GetRequiredPlugins
Python currently only looks at packages prefixed with `pulumi-` when considering required plugins. We need to modify this to look at all packages that have a `pulumiplugin.json` with fallback to `pulumi-` for packages that don't yet have a `pulumiplugin.json`.

iwahbe:
# Description

Check for `pulumiplugin.json` as a way to find if a package is a pulumi package. 

Fixes #7243

## Checklist


justinvp (approved):
LGTM

PR #8516, opened 2021-12-01
[sdk/go] Check for pulumiplugin.json in `GetRequiredPlugins`

### LINKED ISSUE #7245: Go: Support custom server URLs for plugins

iwahbe:
# Description

Checks for `pulumiplugin.json` in `Dir` (from `go list -m -json -mod=mod all`). 

Fixes #7245 

## Checklist


pgavlin:
Have we validated that the module sources always include all of the files present in the module's repo? I'm wondering if we need to update our Go codegen to ensure that the `pulumiplugin.json` file is referenced from the code via a `//go:embed` directive or similar.

iwahbe:
> Have we validated that the module sources always include all of the files present in the module's repo? I'm wondering if we need to update our Go codegen to ensure that the `pulumiplugin.json` file is referenced from the code via a `//go:embed` directive or similar.

Our codegen does not currently include `pulumiplugin.json` at all for go. That will be added in another PR. Performing a quick test, it seems like files are just included. It doesn't seem like we need to include a `//go:embed` directive.

justinvp (approved):
LGTM. I've also verified that `pulumiplugin.json` file is included in the module.

PR #8517, opened 2021-12-01
[sdk/dotnet] Check for `pulumiplugin.json` in package lookup

### LINKED ISSUE #7244: .NET: Support custom server URLs for plugins

iwahbe:
# Description

Fixes #7244

## Checklist


sdk/dotnet/cmd/pulumi-language-dotnet/main.go
```diff
@@ -298,55 +299,104 @@ func (host *dotnetLanguageHost) DetermineDotnetPackageDirectory(
...
+		version: version,
+	}
+}
+
+func DeterminePluginDependency(packageDir, packageName, packageVersion string) (*pulumirpc.PluginDependency, error) {
 
 	logging.V(5).Infof("GetRequiredPlugins: Determining plugin dependency: %v, %v, %v",
 		packageDir, packageName, packageVersion)
 
 	// Check for a `~/.nuget/packages/package_name/package_version/content/version.txt` file.
```
justinvp, line 333: Nit: Maybe update this comment to mention `pulumiplugin.json` as well?

sdk/dotnet/cmd/pulumi-language-dotnet/main.go
```diff
@@ -298,55 +299,104 @@ func (host *dotnetLanguageHost) DetermineDotnetPackageDirectory(
...
-			// just ignore it.
-			logging.V(5).Infof("GetRequiredPlugins: No such file")
-			return nil, nil
-		}
-
-		// some other error.  report it as it means we can't read this important file.
-		logging.V(5).Infof("GetRequiredPlugins: err: %v", err)
+	pulumiPlugin, err := plugin.LoadPulumiPluginJSON(pulumiPluginFilePath)
+	if os.IsNotExist(err) {
+		pulumiPlugin = nil
```
justinvp, line 343: Is this necessary? Won't `pulumiPlugin` already be `nil` when `err != nil` and therefore `os.IsNotExists(err)` is `true`?

justinvp (approved):
LGTM

sdk/dotnet/cmd/pulumi-language-dotnet/main.go
iwahbe: We need the check, but not the assignment. I can fold it down to this
```go
if err != nil && !os.IsNotExist(err) {
		return nil, err
	}
```

PR #8527, opened 2021-12-02
[codegen/python] emit `pulumiplugin.json` by default

### LINKED ISSUE #8526: [codegen/python] Emit `pulumiplugin.json` by default.
## Hello!

## Issue details
Emit `pulumiplugin.json` by default.

iwahbe:
# Description

Remove the flag that makes `pulumiplugin.json` optional. We will now emit unconditionally. 

Fixes #8526 

## Checklist


pkg/codegen/internal/test/testdata/cyclic-types/python/pulumi_example/pulumiplugin.json
```diff
@@ -0,0 +1,4 @@
+{
+  "resource": true,
+  "name": "example"
+}
```
justinvp, line 4: Nit: One minor thing that's bugged me about this emitted file is that we don't include a trailing newline. As part of emitting this by default now, would you mind fixing it to include the trailing newline? Similar to #7480.

Perhaps include the newline in the result returned from the `JSON()` method?

pkg/codegen/python/gen.go
```diff
@@ -1941,15 +1941,14 @@ func genPulumiPluginFile(pkg *schema.Package) ([]byte, error) {
 	plugin := &plugin.PulumiPluginJSON{
 		Resource: true,
 		Name:     pkg.Name,
-		Version:  "${PLUGIN_VERSION}",
```
justinvp, line 1944: Nice. I like this. Now, when this codegen is adopted by the providers we aren't forced to make a change to the `Makefile` to search/replace this value. We can do it at our leisure. And we can use `jq` to add the `version` property.

I do think we'll end up wanting to update our provider `Makefiles` to do this, because it'll fix being able to automatically download plugins for dev builds of the package, but that's not a huge priority.

justinvp (approved):
LGTM

PR #8535, opened 2021-12-02
[codegen/go] Fix ElementType for nested collection types.

pgavlin:
The element types for nested collection types are currently incorrect.

For arrays, we generate an element type of `[]FooArgs`, where
`FooArgs` is an args type for `FooInput`. Instead, the element type
should be `[]Foo`, where `Foo` is the element type of `FooInput`.

For maps, we generate an element type of `FooArgs`. Instead, the element
type should be `map[string]Foo`, where `Foo` is the element type of
`FooInput`.

These changes correct the element types. This is prep for #7943.

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -21,3 +21,6 @@
 
 ### Bug Fixes
 
+- [codegen/go] - Fix `ElementType` for nested collection input and output types.
+  [#8535](https://github.com/pulumi/pulumi/pull/8535)
+
--- a/pkg/codegen/go/gen.go
+++ b/pkg/codegen/go/gen.go
@@ -2150,10 +2150,15 @@ func (pkg *pkgContext) addSuffixesToName(typ schema.Type, name string) []string 
 	return names
 }
 
+type nestedTypeInfo struct {
+	resolvedElementType string
+	names               map[string]bool
+}
+
 // collectNestedCollectionTypes builds a deduped mapping of element types -> associated collection types.
 // different shapes of known types can resolve to the same element type. by collecting types in one step and emitting types
 // in a second step, we avoid collision and redeclaration.
-func (pkg *pkgContext) collectNestedCollectionTypes(types map[string]map[string]bool, typ schema.Type) {
+func (pkg *pkgContext) collectNestedCollectionTypes(types map[string]*nestedTypeInfo, typ schema.Type) {
 	var elementTypeName string
 	var names []string
 	switch t := typ.(type) {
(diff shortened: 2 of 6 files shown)


CHANGELOG_PENDING.md
```diff
@@ -1,5 +1,8 @@
 ### Improvements
 
+- [codegen/go] - Fix `ElementType` for nested collection input and output types.
```
iwahbe, line 3: This seems like a bug to me. Can you explain why this is in "Improvements".

iwahbe (commented):
Can you explain to my why this isn't a breaking change? Otherwise this looks good.

CHANGELOG_PENDING.md
pgavlin: > This seems like a bug to me. Can you explain why this is in "Improvements".

Because I put it in the wrong place :P

pgavlin:
> Can you explain to my why this isn't a breaking change? Otherwise this looks good.

Strictly speaking, this is a breaking change. In particular, it will break code that does an `ApplyT` with a callback that returns a `Foo{Array,Map}` and type asserts to `Foo{Array,Map}MapOutput` or `Foo{Array,Map}ArrayOutput`. I _strongly_ doubt that such code exists in the wild, however: attempting to do anything with the result is unlikely to work, as the type of the resolved value of the output will not match the shape expected by its consumers if it is passed to another `Apply` or to a resource as an input. The way in which this is a breaking change is that it makes these output/input types usable by correcting the element type s.t. the resolved value of an affected output matches the shape expected by its consumers.

pgavlin:
(and because this is a break that is making previously unusable types usable, I think we should take it)

iwahbe (approved):

stack72 (approved):

lukehoban:
Is there an issue tracking the bug this fixes?  I'm curious to see concretely the code that doesn't work

pgavlin:
> Is there an issue tracking the bug this fixes? I'm curious to see concretely the code that doesn't work

There is not. This is something I came across while working on #7943.

pgavlin:
@lukehoban this is an example of some code that does not work: https://gist.github.com/pgavlin/bfcc0ba03e8a2b2e7dece4cdbb9c1b1c

The `ToSomeOtherObjectArrayArrayOutput` panics because the element type of `SomeOtherObjectArrayArrayOutput` does not match the resolved type of `SomeOtherObjectArrayArray`:

```
gh8535 ❯ ./gh8535 
panic: interface conversion: pulumi.Output is pulumi.AnyOutput, not main.SomeOtherObjectArrayArrayOutput

goroutine 1 [running]:
main.SomeOtherObjectArrayArray.ToSomeOtherObjectArrayArrayOutputWithContext(...)
	/Users/pgavlin/dev/pulumi/repros/gh8535/main.go:199
main.SomeOtherObjectArrayArray.ToSomeOtherObjectArrayArrayOutput(...)
	/Users/pgavlin/dev/pulumi/repros/gh8535/main.go:195
main.main()
	/Users/pgavlin/dev/pulumi/repros/gh8535/main.go:299 +0x145
```

pgavlin:
In contrast, the version with correct element types does not panic: https://gist.github.com/pgavlin/2511d0e4cadf1232cfd44682bd5589ae

PR #8539, opened 2021-12-03
[codegen/go] Simplify object default applies.

pgavlin:
The overall goal of this change is to avoid unnecessary conversions
between pointer and non-pointer output types. The use of these
conversions requires that we always generate the pointer variant of an
object type, which impedes #7943's ability to omit these types. Instead,
we can let the return type of the callback match the argument type and
return the dereferenced result of `foo.Defaults()` if necessary.

These changes also simplify the generated code somewhat in order to
avoid introducing additional variables.

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,5 +1,8 @@
 ### Improvements
 
+- [codegen/go] - Simplify the application of object defaults in generated SDKs.
+  [#8539](https://github.com/pulumi/pulumi/pull/8539)
+
 - [codegen/{python,dotnet}] - Emit `pulumiplugin.json` unconditionally.
   [#8527](https://github.com/pulumi/pulumi/pull/8527)
   [#8532](https://github.com/pulumi/pulumi/pull/8532)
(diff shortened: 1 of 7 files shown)


iwahbe (approved):
2 nits, Otherwise LGTM. 
1. I would love to see some comments describing `outputType` and `outputTypeImpl` (and why the both exist). 
2. My understanding is that this cleans up only our internal implementation of `Defaults`. Do we need a changelog entry for this? (I'm genuinely asking. I've heard conflicting definitions of "changelog worthy)

pgavlin:
> My understanding is that this cleans up only our internal implementation of Defaults. Do we need a changelog entry for this? (I'm genuinely asking. I've heard conflicting definitions of "changelog worthy)

These days I think that we should be adding CL entries for all codegen changes, especially to help @pulumi/platform-integrations and @pulumi/native-providers understand that sort of changes to expect in the generated SDKs.

pgavlin:
> I would love to see some comments describing outputType and outputTypeImpl (and why the both exist).

Done.

justinvp (approved):

PR #8543, opened 2021-12-03
[codegen/schema] Do not validate in ImportSchema.

### LINKED ISSUE #8541: Metaschema validation adds significant latency to import and programgen tools
## Hello!

## Issue details
With https://github.com/pulumi/pulumi/pull/7952/ we added metaschema validation as part of the codepath related to binding to the spec. With large schemas (azure-native), the schema validation can take upwards of 20 seconds. While this amortizes well for example generation and other batch operations, it imposes a large penalty for important workloads like one-off resource imports and conversion tools (arm2pulumi).

Screen Shot 2021-11-30 at 6 40 39 PM

One option is to allow disabling schema validation in some cases. There is a draft PR for that [here](https://github.com/pulumi/pulumi/pull/8542). This feels clunky and I am not sure when we would actually like to enforce metaschema validation. Should this be part of schema generation instead so we can trust that the schema has been validated against the metaschema (or at least a certain version etc.) and allow that to be trusted?

### Steps to reproduce

1. `pulumi import azure-native:storage/v20150501preview:StorageAccount <name> /subscriptions/.../resourceGroups/.../providers/Microsoft.Storage/storageAccounts/<storage-account>`
2. Wait 30 seconds for import to succeed

Expected: 
Import should be near instantaneous

Actual: 
~30+ seconds wait for the codegen for the import.

pgavlin:
Allow consumers to load package specs without validating them against
the Pulumi package metaschema. This dramatically reduces the package
load time for scenarios in which packages can be assumed to be
well-formed (e.g. program codegen + packages that are referenced by a
root schema).

Fixes #8541.

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -40,6 +40,10 @@
 - [cli] Log secret decryption events when a project uses the Pulumi Service and a 3rd party secrets provider
   [#8563](https://github.com/pulumi/pulumi/pull/8563)
 
+- [schema] Do not validate against the metaschema in ImportSpec. Clients that need to
+  validate input schemas should use the BindSpec API instead.
+  [#8543](https://github.com/pulumi/pulumi/pull/8543)
+
 ### Bug Fixes
 
 - [codegen/schema] - Error on type token names that are not allowed (schema.Name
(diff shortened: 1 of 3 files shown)


viveklak (approved):

iwahbe (approved):
LGTM

pgavlin:
Test failure is unrelated.

PR #8544, opened 2021-12-03
Pipe serverURL through register resource

### LINKED ISSUE #7242: Include custom server URLs for plugins in stack state
When determining required plugins for an existing stack, we look at providers in the stack's state.

https://github.com/pulumi/pulumi/blob/f0bfc94ef35b2d1b961e65bbbba21495a7be6619/pkg/engine/plugins.go#L97-L127

However, we currently have no way to tell whether such providers are from a plugin from a custom server URL, so we attempt to download all these plugins from the Pulumi URL.

We'll need to include the custom URL in the state to address this. One way of doing that is making it an optional input and a (required?) output, then plumb it through the plugin loader. So if a server is specified in the inputs, we would pass that to the loader. And the loader would always return the server it used to download the plugin. With this approach, we'd have to stamp the URL in the generated SDKs.

One challenge: If it's stamped as an input to a provider, this wouldn't handle the common case of default providers, since they aren't instantiated by the SDK. Which means, we may need to embed the custom server URL in every resource request like we do for the version.

iwahbe:
# Description

Fixes #7242

## Checklist


pkg/engine/deployment.go
```diff
@@ -182,8 +182,13 @@ func newDeployment(ctx *Context, info *deploymentContext, opts deploymentOptions
 					"Type tokens must be of the format <package>:<module>:<type> - "+
 					"refer to the import section of the provider resource documentation.", imp.Type.String())
 			}
-			if imp.Provider == "" && imp.Version == nil {
-				imp.Version = defaultProviderVersions[imp.Type.Package()]
+			if imp.Provider == "" {
+				if imp.Version == nil {
+					imp.Version = defaultProviderVersions[imp.Type.Package()].Version
+				}
+				if imp.ServerURL == "" {
```
justinvp, line 189: We'd discussed offline whether the name of the input property to the provider should be `serverURL` or `pluginDownloadURL` and decided to go with the latter. The latter is more verbose, but it aligns with the [name of the property in the schema](https://github.com/pulumi/pulumi/blob/c95e49e5abc9ea845f08475a3a44bcf9ea593942/pkg/codegen/schema/schema.go#L1520) and is less likely to conflict with a provider's own input properties (which very well could have a `serverURL` for its own purposes). This PR does use `pluginDownloadURL` for the input property itself, but everywhere else in the implementation is using `ServerURL`.

Two questions:

1. Should we use `PluginDownloadURL` throughout instead of `ServerURL` to be consistent and keep the concept count lower for easier readability/understandability of the code?
   - I realize we do use `ServerURL` in the [`workspace.PluginInfo`](https://github.com/pulumi/pulumi/blob/c95e49e5abc9ea845f08475a3a44bcf9ea593942/sdk/go/common/workspace/plugins.go#L93) struct and `Server` for the [`plugin.PulumiPluginJSON`](https://github.com/pulumi/pulumi/blob/eb735c7106fcbb48eb3e0edaab3d0baea47d2f0e/sdk/go/common/resource/plugin/plugin.go#L58) struct, so there is a case to be made for `ServerURL`.
1. When we expose this in the SDKs on `ResourceOptions`, do we want to expose it as `ResourceOptions.PluginDownloadURL` or `ResourceOptions.ServerURL`?

Curious if anyone else has a preference.

pkg/engine/plugins.go
```diff
@@ -283,9 +286,9 @@ func computeDefaultProviderPlugins(languagePlugins, allPlugins pluginSet) map[to
 		}
 	}
 
-	defaultProviderVersions := make(map[tokens.Package]*semver.Version)
+	defaultProviderVersions := make(map[tokens.Package]workspace.PluginInfo)
```
justinvp, line 289: Nit: ?

```suggestion
	defaultProviderInfo := make(map[tokens.Package]workspace.PluginInfo)
```

pkg/engine/deployment.go
```diff
@@ -182,8 +182,13 @@ func newDeployment(ctx *Context, info *deploymentContext, opts deploymentOptions
 					"Type tokens must be of the format <package>:<module>:<type> - "+
 					"refer to the import section of the provider resource documentation.", imp.Type.String())
 			}
-			if imp.Provider == "" && imp.Version == nil {
-				imp.Version = defaultProviderVersions[imp.Type.Package()]
+			if imp.Provider == "" {
+				if imp.Version == nil {
+					imp.Version = defaultProviderVersions[imp.Type.Package()].Version
```
justinvp, line 187: Nit: Rename `defaultProviderVersions` => `defaultProviderInfo`?

pkg/resource/deploy/source_eval.go
```diff
@@ -55,7 +55,7 @@ type EvalRunInfo struct {
 // a confgiuration map.  This evaluation is performed using the given plugin context and may optionally use the
 // given plugin host (or the default, if this is nil).  Note that closing the eval source also closes the host.
 func NewEvalSource(plugctx *plugin.Context, runinfo *EvalRunInfo,
-	defaultProviderVersions map[tokens.Package]*semver.Version, dryRun bool) Source {
+	defaultProviderVersions map[tokens.Package]workspace.PluginInfo, dryRun bool) Source {
```
justinvp, line 58: Nit: ?

```suggestion
	defaultProviderInfo map[tokens.Package]workspace.PluginInfo, dryRun bool) Source {
```

pkg/resource/deploy/source_eval.go
```diff
@@ -66,10 +66,10 @@ func NewEvalSource(plugctx *plugin.Context, runinfo *EvalRunInfo,
 }
 
 type evalSource struct {
-	plugctx                 *plugin.Context                    // the plugin context.
-	runinfo                 *EvalRunInfo                       // the directives to use when running the program.
-	defaultProviderVersions map[tokens.Package]*semver.Version // the default provider versions for this source.
-	dryRun                  bool                               // true if this is a dry-run operation only.
+	plugctx                 *plugin.Context                         // the plugin context.
+	runinfo                 *EvalRunInfo                            // the directives to use when running the program.
+	defaultProviderVersions map[tokens.Package]workspace.PluginInfo // the default provider versions for this source.
```
justinvp, line 71: Nit: ?

```suggestion
	defaultProviderInfo map[tokens.Package]workspace.PluginInfo // the default provider info for this source.
```

justinvp (commented):
Looks pretty good. Some questions and nits.

I'm less familiar with the provider registry, and haven't actually pulled down these changes to try it out. It may be worth @pgavlin taking a quick look over the changes as he's more familiar with the provider registry and default providers and may have additional feedback.

pkg/engine/lifeycletest/provider_test.go
```diff
@@ -1118,3 +1118,54 @@ func TestProviderVersionInputAndOption(t *testing.T) {
...
+		assert.NoError(t, err)
+
+		if provID == "" {
+			provID = providers.UnknownID
+		}
+
+		provRef, err := providers.NewReference(provURN, provID)
+		assert.NoError(t, err)
+
+		_, _, _, err = monitor.RegisterResource("pkgA:m:typA", "resA", true, deploytest.ResourceOptions{
```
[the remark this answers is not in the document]
iwahbe: We can and should have that test: https://github.com/pulumi/pulumi/pull/8544/commits/e525d901522c6882c5d8179a38d02dee30ff5d64

sdk/go/common/resource/properties.go
```diff
@@ -646,3 +646,47 @@ const OutputValueSig = "d0e6a833031e9bbcd3f4e8bde6ca49a4"
...
+		r:    sp.m,
+	}
+}
+
+func (sp SpecialProperties) ServerURL() SpecialProperty {
+	return SpecialProperty{
+		name: "pluginDownloadURL",
+		r:    sp.m,
+	}
+}
```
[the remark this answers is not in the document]
iwahbe: I added these helpers because I wanted to distinguish between inputs that Pulumi (the engine) cares about and general inputs (that are user defined). It also provides a single source of truth for the correct key, which prevents mistypes. We can just look up the value if you feel that would be a better option.

pkg/engine/deployment.go
iwahbe: 1. @justinvp You convinced me that `"serverURL"` had a higher chance of conflicting with user inputs/outputs. It should be impossible to conflict with the `ServerURL` name in go, unless I'm confused. Given that we have (and are locked into through serialization) two different names for the same concept, I figured that I should choose a single name (for in the code base) and stick with it. I chose `serverURL` because 2/3 of our endpoints have that name. I would be happy to change it to `PulumiPluginJSON` if you would prefer. 

2. I think `ResourceOptions.PluginDownloadURL` is probably clearer there.

I would also be curious what others think.

sdk/go/common/resource/properties.go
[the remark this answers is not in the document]
justinvp: My preference is not expose the helpers here. `sdk/go/common/resource/properties` is used across the Pulumi ecosystem to represent properties for any resource, not just providers. Adding new public APIs here with special accessors for two provider inputs doesn't feel quite right to me.

I'd be fine with having non-public helpers that can be used within the implementation. Perhaps in pkg/internal? Or no helpers at all and just access the map directly (possibly using string constants if we're concerned about mistypes).

If you are going to keep some helpers like this, one nit (related to my other comment): The accessor looking up `"pluginDownloadURL"` should probably be named `PluginDownloadURL()`. I think it's better for the key and name of the accessor to match.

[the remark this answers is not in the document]
iwahbe: That makes perfect sense. I'm going to move the helpers to `pkg/resource/deploy/providers`. There is already a helper  (`GetProviderVersion`), so the location should make sense. I'll make the name change here, and just make the general change from `ServerURL` to `PluginDownloadURL`.

justinvp (approved):
LGTM

pkg/resource/deploy/providers/registry.go
```diff
@@ -30,19 +30,45 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/common/workspace"
 )
 
-// GetProviderVersion fetches and parses a provider version from the given property map. If the version property is not
-// present, this function returns nil.
+var versionKey resource.PropertyKey = "version"
+var pluginDownloadKey resource.PropertyKey = "pluginDownloadURL"
```
pgavlin, line 34: These can be constants:

```suggestion
const versionKey resource.PropertyKey = "version"
const pluginDownloadKey resource.PropertyKey = "pluginDownloadURL"
```

pgavlin (commented):
It looks like we're considering the download URL to be part of the provider's identity for the purposes of default providers. Is that intentional?

iwahbe:
> It looks like we're considering the download URL to be part of the provider's identity for the purposes of default providers. Is that intentional?

Yes. We want `{ "foo", "1.2.3", "your-url" }` to be a different provider from `{ "foo", "1.2.3", "my-url" }`. I'm not 100% sure that I'm understanding your question though.

pgavlin:
> Yes. We want { "foo", "1.2.3", "your-url" } to be a different provider from { "foo", "1.2.3", "my-url" }.

This is what I was looking for, thanks. In that case, should we apply some canonicalization to the download URL? It feels a little awkward that e.g. `my-url/provider` and `my-url/provider/`would be considered different.

pgavlin:
We'll probably also need to incorporate the download URL (if any) into [`ProviderRequest.Name`](https://github.com/pulumi/pulumi/blob/master/pkg/resource/deploy/providers/provider.go#L61-L86), which is used to generate the name portion of the provider resource. Otherwise, I think that e.g. `{ "foo", "1.2.3", "your-url" }` and `{ "foo", "1.2.3", "my-url" }` will both attempt to register a provider named `foo_1_2_3`, which will fail due to the duplicate name.

iwahbe:
@pgavlin That makes sense. I think (https://github.com/pulumi/pulumi/pull/8544/commits/a97152ee0c3f8052137833750344a91c1fabdecb + https://github.com/pulumi/pulumi/pull/8544/commits/67538b10557d2444b17fb0db90e0b2bb0db56348) should cover both cases.

Issue #5455, opened 2020-09-25
Dynamic providers fail to load when used from multi-lang components


justinvp:
For example, consider a component written in TypeScript that uses a dynamic provider, and a Python program that uses the component. The Python program will fail during `pulumi up` with:

```
    internal/modules/cjs/loader.js:638
        throw err;
        ^
    Error: Cannot find module '@pulumi/pulumi/cmd/dynamic-provider'
        at Function.Module._resolveFilename (internal/modules/cjs/loader.js:636:15)
        at Function.resolve (internal/modules/cjs/helpers.js:33:19)
        at [eval]:1:21
        at Script.runInThisContext (vm.js:122:20)
        at Object.runInThisContext (vm.js:329:38)
        at Object.<anonymous> ([eval]-wrapper:6:22)
        at Module._compile (internal/modules/cjs/loader.js:778:30)
        at evalScript (internal/bootstrap/node.js:590:27)
        at startup (internal/bootstrap/node.js:265:9)
        at bootstrapNodeJSCore (internal/bootstrap/node.js:623:3)
```

The problem is due to the `pwd` used when launching the `pulumi-resource-pulumi-nodejs` plugin: it is currently being set to the dir of the Pulumi program.

In the case of our example, the Pulumi program is written in Python and doesn’t have a `node_modules,` so when `pulumi-resource-pulumi-nodejs` is run (which is just a bash script), it can’t find the `@pulumi/pulumi/cmd/dynamic-provider` script its looking for:

https://github.com/pulumi/pulumi/blob/6e0163b248d0fb3d6db61de6c963fb89625e3c8f/sdk/nodejs/dist/pulumi-resource-pulumi-nodejs#L1-L5

If the `pwd` is hardcoded to be the dir of the component (which has a `node_modules`), everything works as expected.

We’ll need to flow the `pwd` through somehow such that if `pulumi-resource-pulumi-nodejs` is being launched due to use from a component, it uses the `pwd` of the component, rather than the `pwd` of the program.

justinvp:
While we definitely want to make sure dynamic providers work with multi-language components, we could consider working around this in the short-term for EKS by changing the dynamic providers in EKS to be custom resources implemented via the new provider plugin (would need to ensure the eks resource plugin is installed as part of depending on `@pulumi/eks`).

delitescere:
[...]

PR #8548, opened 2021-12-06
Replace the use of pipenv with Python's venv

### LINKED ISSUE #8379: Follow up to avoid the need to globally-install Python SDK
## Hello!

## Issue details

The quick fix in https://github.com/pulumi/pulumi/pull/8361 was unfortunate. It requires Python SDK to be installed globally so that the pulumi-langage-host run from a sub-process under pytest finds it. It is possibly related to how `pipenv` works on the latest version. Alternative solutions:

1. edit the environment with `PATH` being extended with `$(pipenv venv)/bin` so that `#/usr/bin/env python` resolves to the python within the VENV managed by pipenv, that has a reference to Pulumi Python SDK already.

2. consider removing pipenv entirely

### Steps to reproduce

1.
2.
3.
4.

Expected: 
Actual:

justinvp:
We've had a few issues in the recent past related to pipenv oddities in CI which lead us to temporarily globally install the Python SDK in CI.

This change removes the use of pipenv in favor of Python's built-in venv and avoids globally installing the Python SDK.

Fixes #8379


tests/testprovider/main.go
```diff
@@ -17,25 +17,42 @@ package main
 
 import (
 	"context"
-	"crypto/rand"
```
t0yv0, line 20: Curious, are these changes related to / necessitated by pipenv removal?

justinvp: Dynamic providers typically don't work correctly when used from multi-language components (#5455). Somehow the way we were previously setting up the tests to use the shared pipenv virtualenv allowed the Python MLC tests that used dynamic providers to pass. But when I changed the tests to use their own virtual env from (python -m venv venv) I was running into errors along the lines of #5455.

Instead of trying to rework how we setup the tests to get them to pass, I changed the Python MLCs to use a resource from a testprovider rather than a resource from a dynamic provider inside the MLC. The testprovider previously only exposed a `Random` resource and as part of this change I added a new `Echo` resource. These changes were just some refactoring to more cleanly expose both `Random` and `Echo` from the testprovider.

With these changes, we could potentially do some further subsequent cleanup of the MLC tests:

- There are some tests with a MLC implemented in TypeScript that use a dynamic provider, and we have some hacks in the integration test package to make this work. But it is ugly and not representative of real-world usage (you have to have a `node_modules` dir in your program directory with the dependencies needed by the MLC, even if your program  is written in another language). We could remove those hacks and replace the use of the dynamic provider with the new `Echo` resource in the testprovider.

- Go doesn't have dynamic providers and in order to implement the equivalent MLC in Go we had to expose an equivalent `Echo` resource from the MLC provider itself (this was before I had added the testprovider). We could instead clean these Go test MLCs up to not do that, which would significantly reduce the amount of code in the MLC, and instead use the new `Echo` resource in the testprovider.

- When we get around to implementing support for authoring MLC in .NET, we'll be able to use the new `Echo` resource in the testprovider.

PR #8549, opened 2021-12-07
Set less agressive timeouts for HTTP retries

chrsmith:
The Pulumi CLI's `httpstate` backend retries HTTP requests if it sees a 5xx error. This presumably makes stack updates a little more reliable in the face of transient network errors and things.

However, in practice the default values for the delay/retry logic are much too aggressive to actually be useful.
https://github.com/pulumi/pulumi/blob/a952140b34c667d005116ee53be08077b5863948/sdk/go/common/util/retry/until.go#L34-L38

Here are some timestamps I'm seeing for repeated HTTP requests that arise from failures. The first value is the initial 5xx, and the next four are the additional attempts to the same URL. Notice how all 5 requests are attempted in less than 2 seconds. (In-line with the default retry values... though not accounting for the time it takes to actually send the request, and for the server to send the response, etc.)

```
2021-12-06T23:08:53.497787Z
2021-12-06T23:08:53.678658Z
2021-12-06T23:08:53.941169Z
2021-12-06T23:08:54.328391Z
2021-12-06T23:08:54.866719Z
```

```
2021-12-06T23:08:51.009979Z
2021-12-06T23:08:51.249725Z
2021-12-06T23:08:51.567933Z
2021-12-06T23:08:51.995871Z
2021-12-06T23:08:52.592026Z
```

This PR makes the timeouts much more generous, so that hopefully if there is some meaningful error on the server-side it can be corrected before the last retry is sent. (Delaying 1s, 2s, 4s, and finally 8s between subsequent attempts.)

Waiting a total of 15s before failing an HTTP request is a long time... if the end result is that a stack update fails and/or potentially leaves the checkpoint file in a bad state, I'd think we should err on the side of letting the CLI wait and retry. 

Though, perhaps that's the wrong way to think about it. What do you think?

### DIFF
--- a/pkg/backend/httpstate/client/api.go
+++ b/pkg/backend/httpstate/client/api.go
@@ -26,6 +26,7 @@ import (
 	"reflect"
 	"runtime"
 	"strings"
+	"time"
 
 	"github.com/pulumi/pulumi/sdk/v2/go/common/diag"
 
@@ -112,6 +113,18 @@ func (updateAccessToken) Kind() accessTokenKind {
 
 func (t updateAccessToken) String() string {
 	return string(t)
+}
+
+func float64Ptr(f float64) *float64 {
+	return &f
+}
+
+func durationPtr(d time.Duration) *time.Duration {
+	return &d
+}
+
+func intPtr(i int) *int {
+	return &i
 }
 
 // pulumiAPICall makes an HTTP request to the Pulumi API.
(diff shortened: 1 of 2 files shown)


t0yv0:
This function affects the following call sites in this repo:

1. https://formulae.brew.sh/api/formula/pulumi.jso
2. https://get.pulumi.com/releases/plugins
3. func pulumiAPICall (httpstate.go)
4. func (a *Asset) readURI() (*Blob, error) {
5. func (a *Archive) openURLStream

If the motivation is to change only (3) could me make it so only that scenario is affected? It really is reasonable to have longer retries if we know the performance profile of the Pulumi service, but it's a bit less reasonable to also backoff as aggressively on things like CDN lookups as this just makes diagnosing networking failures slower for the users.

chrsmith:
Thanks @t0yv0 , I should have looked at all the callers of `httputil.DoWithRetry`. I reworked the code to keep the behavior as-is for existing callers, but allow for the `httpstate` module to specify custom values for the delay in-between requests.

PTAL

leezen (approved):

t0yv0 (approved):
Awesome, thank you, LGTM.

PR #8551, opened 2021-12-07
Clone snapshot before testing with it

Frassle:
Got to love mutability by default. Just found a whole load of tests that are checking that resource snapshots haven't changed by checking they're equal to the snapshot before, but the snapshot was mutated by the update so it's trivially equal because its the same object. Changing it so we clone the snapshot before passing it in triggers a load of failures because default provider urns have been filled in by update and the test didn't expect that. This PR adds the code to clone before testing and fixes up the tests to pass again.

### DIFF
--- a/pkg/engine/lifeycletest/refresh_test.go
+++ b/pkg/engine/lifeycletest/refresh_test.go
@@ -382,6 +382,19 @@ func validateRefreshDeleteCombination(t *testing.T, names []string, targets []st
 
 	provURN := p.NewProviderURN("pkgA", "default", "")
 
+	// The new resources will of had their default provider urn filled in. We fill this in on
+	// the old resources here as well so that the equal checks below pass
+	for _, r := range snap.Resources {
+		if r.URN == provURN {
+			provRef, err := providers.NewReference(r.URN, r.ID)
+			assert.NoError(t, err)
+			for i := range oldResources {
+				oldResources[i].Provider = provRef.String()
+			}
+			break
+		}
+	}
+
 	for _, r := range snap.Resources {
 		switch urn := r.URN; urn {
 		case provURN:
@@ -576,6 +589,19 @@ func validateRefreshBasicsCombination(t *testing.T, names []string, targets []st
 
 	provURN := p.NewProviderURN("pkgA", "default", "")
 
+	// The new resources will of had their default provider urn filled in. We fill this in on
+	// the old resources here as well so that the equal checks below pass
+	for _, r := range snap.Resources {
+		if r.URN == provURN {
+			provRef, err := providers.NewReference(r.URN, r.ID)
+			assert.NoError(t, err)
+			for i := range oldResources {
+				oldResources[i].Provider = provRef.String()
+			}
+			break
+		}
+	}
+
 	for _, r := range snap.Resources {
 		switch urn := r.URN; urn {
 		case provURN:
@@ -678,7 +704,7 @@ func TestCanceledRefresh(t *testing.T) {
 		Parallel: 1,
 		Host:     deploytest.NewPluginHost(nil, nil, nil, loaders...),
 	}
-	project, target := p.GetProject(), p.GetTarget(old)
+	project, target := p.GetProject(), p.GetTarget(t, old)
 	validate := func(project workspace.Project, target deploy.Target, entries JournalEntries,
 		_ []Event, res result.Result) result.Result {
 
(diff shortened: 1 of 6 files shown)


t0yv0:
Something is not quite right with the tests yet, ping me again once the PR works out passing tests. 

```
$ git rev-parse HEAD
dded327de47cfd920686d0428c65cca701b8df43

[nix-shell:~/pulumi/pkg/engine/lifeycletest]$ go test -test.v -run TestSingleResourceDefaultProviderUpgrade
go test -test.v -run TestSingleResourceDefaultProviderUpgrade
=== RUN   TestSingleResourceDefaultProviderUpgrade
    provider_test.go:159: unexpected resource urn:pulumi:test::::pulumi:providers:pkgA::default
--- FAIL: TestSingleResourceDefaultProviderUpgrade (0.01s)
FAIL
exit status 1
FAIL	github.com/pulumi/pulumi/pkg/v3/engine/lifeycletest	0.214s


$ git checkout master
$ git rev-parse HEAD
c02bff9acf1e879e71651f4b06a5af84b3a3aac6

[nix-shell:~/pulumi/pkg/engine/lifeycletest]$ go test -test.v -run TestSingleResourceDefaultProviderUpgrade
go test -test.v -run TestSingleResourceDefaultProviderUpgrade
=== RUN   TestSingleResourceDefaultProviderUpgrade
--- PASS: TestSingleResourceDefaultProviderUpgrade (0.02s)
PASS
ok  	github.com/pulumi/pulumi/pkg/v3/engine/lifeycletest	0.226s

```

I'm quite curious about which part of a structure called Snapshot is being mutated as you point out.

```
   type Snapshot struct {
	Manifest          Manifest             // a deployment manifest of versions, checksums, and so on.
	SecretsManager    secrets.Manager      // the manager to use use when seralizing this snapshot.
	Resources         []*resource.State    // fetches all resources and their associated states.
	PendingOperations []resource.Operation // all currently pending resource operations.
   }
```

It sounds like something is shared and being completed in Resources? 

I'll run a few more experiments once the tests pass.

Frassle:
So TestSingleResourceDefaultProviderUpgrade was because the tests now don't mutate the snapshot so it was doing a destroy against a snapshot with no provider in it, and when it went to inject the default provider the NullSource it used didn't have a project name so the urn ended up as `urn:pulumi:test::::pulumi:providers:pkgA::default` instead of `urn:pulumi:test::test::pulumi:providers:pkgA::default`. I've made NullSource have a project name now, so that test should pass now.

Frassle:
> I'm quite curious about which part of a structure called Snapshot is being mutated as you point out.

Resources is what gets mutated. See [rebuildBaseState](https://github.com/pulumi/pulumi/blob/096b667ee161d9061e141a7ea3d6e863115bff9d/pkg/resource/deploy/deployment_executor.go#L521) where we re-assign prev.Resources.

pkg/resource/deploy/source_null.go
```diff
@@ -22,16 +22,19 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/result"
 )
 
-// NullSource is a singleton source that never returns any resources.  This may be used in scenarios where the "new"
+// NullSource is a source that never returns any resources.  This may be used in scenarios where the "new"
 // version of the world is meant to be empty, either for testing purposes, or removal of an existing stack.
-var NullSource Source = &nullSource{}
```
t0yv0, line 27: A little worried here since "technically" this is a breaking change if any code happens to depend on `var NullSource`. Perhaps consider leaving that var around? This package is not under `internal` and the identifier is Capitalized so it's exported.

pkg/engine/lifeycletest/refresh_test.go
```diff
@@ -382,6 +382,19 @@ func validateRefreshDeleteCombination(t *testing.T, names []string, targets []st
 
 	provURN := p.NewProviderURN("pkgA", "default", "")
 
+	// The new resources will of had their default provider urn filled in. We fill this in on
```
t0yv0, line 385: "resources will of had" <- typo

pkg/engine/lifeycletest/refresh_test.go
```diff
@@ -576,6 +589,19 @@ func validateRefreshBasicsCombination(t *testing.T, names []string, targets []st
...
```
t0yv0, line 592: If this is a copy of the above block perhaps a helper func would deduplicate.

pkg/resource/deploy/source_null.go
Frassle: I don't think anything uses this, the only reference I can find to NullSource across github.com/pulumi is this use for destroy. Actually even across all github.com search for deploy.NullSource only returns this. We could leave this in as `&nullSource{project: ""}` if we're that worried.

pkg/engine/lifeycletest/refresh_test.go
Frassle: Sorry, I don't see what the typo is. "will have had"? "would of had"?

pkg/resource/deploy/source_null.go
t0yv0: I mean something in user code that links against pulumi/pulumi. I'm definitely worried since we have significant codebases linking against us.

pkg/engine/lifeycletest/refresh_test.go
t0yv0: Possibly.. I'm not sure what you're trying to say. It seems to me the `of` word is misplaced. But I'm not a native English speaker so it's mostly up to you, just calling attention to a possible typo.

t0yv0 (approved):
LGTM with minor nits.

pkg/engine/lifeycletest/refresh_test.go
Frassle: Yes I think this is the standard native english speaker thinks/hears "will've" which sounds like "will of" but actually its "will have".

PR #8554, opened 2021-12-07
Adding a Log3rdPartyDecryptionEventType for logging 3rd party secrets provider decryption

stack72:
# Description

Fixes # (issue)

## Checklist

### DIFF
--- a/sdk/go/common/apitype/stacks.go
+++ b/sdk/go/common/apitype/stacks.go
@@ -77,6 +77,12 @@ type DecryptValueResponse struct {
 	Plaintext []byte `json:"plaintext"`
 }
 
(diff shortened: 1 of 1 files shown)


t0yv0 (approved):
The changes are trivially safe (additive) but I'm wondering if there's a link to the context for this work that'd be awesome. It looks like something is going to be sending JSON messages to log decryption events for auditing?

stack72:
> The changes are trivially safe (additive) but I'm wondering if there's a link to the context for this work that'd be awesome. It looks like something is going to be sending JSON messages to log decryption events for auditing?

So what we are doing here is exposing an endpoint in the service for anyone using the service to be able to audit when a decryption event happens for a 3rd party secret provider

t0yv0:
Perfect. 

So the struct could have been called Request but Log is fine too as it's probably no-response-API right.

PR #8555, opened 2021-12-07
[sdk/go] Add PluginDownloadURL resource option

### LINKED ISSUE #8553: [sdk/go] Support for custom server urls
## Hello!

## Issue details

Provide the go SDK hookup for https://github.com/pulumi/pulumi/pull/8544. 
This has two components: 
- Update the SDK itself to accommodate passing server urls.
- Update codegen to emit server urls when `pluginDownloadUrl` is set in the schema. 
### Affected area/feature

- go sdk
- codegen

iwahbe:
# Description

Fixes #8553

This branch has https://github.com/pulumi/pulumi/pull/8544 merged in, so we should wait until it merges before merging here. 

## Checklist


tests/integration/integration_test.go
```diff
@@ -1004,6 +1005,50 @@ func testConstructOutputValues(t *testing.T, lang string, dependencies ...string
...
+		data := &apitype.DeploymentV3{}
+		err = json.Unmarshal(deployment.Deployment, data)
+		assert.NoError(t, err)
+		urlKey := "pluginDownloadURL"
+		for _, resource := range data.Resources {
+			switch {
+			case providers.IsDefaultProvider(resource.URN):
+				assert.Equal(t, "get.com", resource.Outputs[urlKey])
+			case providers.IsProviderType(resource.Type):
+				assert.Equal(t, "get.pulumi/test/providers", resource.Outputs[urlKey])
```
justinvp, line 1022: Should we be checking for it in `resource.Inputs` here and above rather than `resource.Outputs`?

justinvp (approved):
LGTM

tests/integration/integration_test.go
iwahbe: I think it's probably a good idea to check both.

PR #8558, opened 2021-12-08
[codegen/schema] Add package name allowlist

### LINKED ISSUE #8556: Add allowlist of non-package names in Pulumi schema
## Hello!

## Issue details

Currently, we enforce that all type tokens defined in the schema are prefixed under the package name. This prevents certain use cases (CRD wrappers). We should add a field to the schema that lists additional valid prefixs. We should then only error on unexpected prefixs. 

- We need to be careful to ensure test coverage for tokens with other names. 

### Affected area/feature

schema

iwahbe:
# Description

Fixes #8556

## Checklist


pkg/codegen/schema/schema.go
```diff
@@ -1831,12 +1834,15 @@ const (
 )
 
 // Validate an individual name token.
-func (spec *PackageSpec) validateTypeToken(section, token string) hcl.Diagnostics {
+func (spec *PackageSpec) validateTypeToken(validPrefixes map[string]bool, section, token string) hcl.Diagnostics {
```
justinvp, line 1837: Mind using the same naming throughout for improved readability/understandability?

```suggestion
func (spec *PackageSpec) validateTypeToken(allowedPackageNames map[string]bool, section, token string) hcl.Diagnostics {
```

pkg/codegen/schema/schema.go
```diff
@@ -1831,12 +1834,15 @@ const (
...
 	diags := hcl.Diagnostics{}
 
-	requiredPrefix := spec.Name + ":"
 	path := memberPath(section, token)
-	if !strings.HasPrefix(token, requiredPrefix) {
+	var prefix string
+	if i := strings.Index(token, ":"); i != -1 {
+		prefix = token[:i]
+	}
+	if !validPrefixes[prefix] {
```
justinvp, line 1845:
```suggestion
	var packageName string
	if i := strings.Index(token, ":"); i != -1 {
		packageName = token[:i]
	}
	if !allowedPackageNames[packageName] {
```

pkg/codegen/schema/schema.go
```diff
@@ -1847,14 +1853,18 @@ func (spec *PackageSpec) validateTypeToken(section, token string) hcl.Diagnostic
 // This is for validating non-reference type tokens.
 func (spec *PackageSpec) validateTypeTokens() hcl.Diagnostics {
 	diags := hcl.Diagnostics{}
+	validPrefixes := map[string]bool{spec.Name: true}
+	for _, prefix := range spec.AllowedPackageNames {
+		validPrefixes[prefix] = true
+	}
```
justinvp, line 1859:
```suggestion
	allowedPackageNames := map[string]bool{spec.Name: true}
	for _, packageName := range spec.AllowedPackageNames {
		allowedPackageNames[packageName] = true
	}
```

justinvp (approved):
LGTM

pkg/codegen/schema/schema.go
iwahbe: Good point

PR #8563, opened 2021-12-08
[cli] Log Secret Decryption events when using 3rd party secrets + service

stack72:
For users who use the Pulumi Service for their state BUT a 3rd party
secrets provider, we now log when that secret decryption has happened.

We have 2 forms of logs:

* single secret when a user uses `pulumi config get x --show-secrets`
* batch decryption for when a user runs a stack export or config list

The API logs to request are fire and forget. The API response is a 204
so we don't need to do anything with the response.

# Description

Fixes # (issue)

## Checklist

### DIFF
--- a/pkg/cmd/pulumi/util.go
+++ b/pkg/cmd/pulumi/util.go
@@ -882,3 +882,34 @@ func buildStackName(stackName string) (string, error) {
 
 	return stackName, nil
 }
+
+// we only want to log a secrets decryption for a service backend project
+// we will allow any secrets provider to be used (service or self managed)
+// we will log the message and not worry about the response. The types
+// of messages we will log here will range from single secret decryption events
+// to requesting a list of secrets in an individual event e.g. stack export
+// the logging event will only happen during the `--show-secrets` path within the cli
+func log3rdPartySecretsProviderDecryptionEvent(ctx context.Context, backend backend.Stack,
+	secretName, commandName string) {
+	if stack, ok := backend.(httpstate.Stack); ok {
+		// we only want to do something if this is a service backend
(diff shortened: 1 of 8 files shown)


pkg/cmd/pulumi/util.go
```diff
@@ -882,3 +882,30 @@ func buildStackName(stackName string) (string, error) {
...
 }
+
+// we only want to log a secrets decryption for a service backend project
+// we will allow any secrets provider to be used (service or self managed)
+// we will log the message and not worry about the response. The types
+// of messages we will log here will range from single secret decryption events
+// to requesting a list of secrets in an individual event e.g. stack export
+// the logging event will only happen during the `--show-secrets` path within the cli
+func log3rdPartySecretsProviderDecryptionEvent(ctx context.Context, backend backend.Stack,
+	secretName, commandName string) {
```
justinvp, line 893: Nit: Does it ever make sense to have both `secretName` and `commandName` set, or are these mutually exclusive?

It seems like it'd be somewhat easy mix-up and set the wrong parameter. It might be worth having separate functions for the callers to call. Or at a minimum, perhaps using some asserts, e.g.:

```go
if secretName != "" {
	contract.Assert(commandName == "");
	_ = client.Log3rdPartySecretsProviderDecryptionEvent(ctx, id, secretName)
}

if commandName != "" {
	contract.Assert(secretName == "");
	_ = client.LogBulk3rdPartySecretsProviderDecryptionEvent(ctx, id, commandName)
}
```

justinvp (approved):
LGTM

pkg/cmd/pulumi/util.go
stack72: I really like this! I will get tis updated now

PR #8564, opened 2021-12-08
Use $(PYTHON) for makefile

iwahbe:
# Description

Python kept failing to build. It was because we were invoking `python` instead of `$(PYTHON)`. This fixes the problem, along with a few other small nits. 

## Checklist


t0yv0 (approved):
This all looks good.

justinvp (approved):
Thanks

PR #8565, opened 2021-12-08
Always lock in filestate backends, previously feature-flagged

### LINKED ISSUE #6536: Turn on `PULUMI_SELF_MANAGED_STATE_LOCKING=1` by default
In https://github.com/pulumi/pulumi/pull/2697 we added support for self-managed backend state locking behind the  `PULUMI_SELF_MANAGED_STATE_LOCKING=1` flag to ensure that we can get some real-world validation of this before making it the default, given it's potential impact on performance and correctness when using a local or cloud backend.  

Once we have sufficient real-world experience, we should flip this to be on by default.

t0yv0:
# Description

Fixes #6536

## Checklist

### DIFF
--- a/pkg/backend/filestate/backend.go
+++ b/pkg/backend/filestate/backend.go
@@ -280,13 +280,11 @@ func (b *localBackend) DoesProjectExist(ctx context.Context, projectName string)
 func (b *localBackend) CreateStack(ctx context.Context, stackRef backend.StackReference,
 	opts interface{}) (backend.Stack, error) {
 
-	if cmdutil.IsTruthy(os.Getenv(PulumiFilestateLockingEnvVar)) {
-		err := b.Lock(ctx, stackRef)
-		if err != nil {
-			return nil, err
-		}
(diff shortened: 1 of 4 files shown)


pkg/backend/filestate/backend.go
```diff
@@ -280,13 +280,11 @@ func (b *localBackend) DoesProjectExist(ctx context.Context, projectName string)
...
 
-	if cmdutil.IsTruthy(os.Getenv(PulumiFilestateLockingEnvVar)) {
-		err := b.Lock(ctx, stackRef)
-		if err != nil {
-			return nil, err
-		}
-		defer b.Unlock(ctx, stackRef)
+	err := b.Lock(ctx, stackRef)
+	if err != nil {
+		return nil, err
```
t0yv0, line 285: This is actually covered by `TestListStacksWithMultiplePassphrases`

justinvp (approved):
LGTM. Not sure what's going on with the coverage report.

PR #8576, opened 2021-12-09
Opt language hosts out of `PULUMI_IGNORE_AMBIENT_PLUGINS`

iwahbe:
# Description

Opt language hosts out of `PULUMI_IGNORE_AMBIENT_PLUGINS`. Currently, setting `PULUMI_IGNORE_AMBIENT_PLUGINS` breaks the Automation API because the Automation API does not have a language specific package manager to fall back to. 

## Checklist


t0yv0 (approved):
Does what it says, LGTM.

PR #8578, opened 2021-12-09
fixed obvious reference to stack name

### LINKED ISSUE #8292: Improve docs and help messages for fully-qualified stack name
## Hello!

## Issue details

As customers are using more and more the Pulumi service, and as they're migrating onto our platform, it's not always clear that a fully-qualified stack name can comprise of the project name.
This is particularly important as they often need to specify the project name rather than let our CLI magically (and sometimes incorrectly) guess it.

The most documented form of a stack name is `[<org-name>/]<stack-name>` and `project-name` is almost always ommitted.

A more helpful format could be `[<org-name>/[<project-name>/]]<stack-name>` or simply described in 2 parts as `[<org-name>/]<stack-name>` and `[<org-name>/<project-name>]<stack-name>`

For reference:

* https://github.com/pulumi/pulumi/blob/master/pkg/cmd/pulumi/stack_init.go#L38
* https://github.com/pulumi/pulumi/blob/master/pkg/cmd/pulumi/util.go#L389
* https://github.com/pulumi/pulumi/blob/master/pkg/cmd/pulumi/new.go#L531

It would be great to also update this documentation: https://github.com/pulumi/pulumi-hugo/blob/master/themes/default/content/docs/intro/concepts/stack.md?plain=1#L33

MadsRC:
# Description

Fixes #8292

Simple fix for the inconsistent documentation. Decided to go with the format of `[<org-name>/<project-name>/]<stack-name>` where the optional project-name is followed by a forward slash. In my head, it makes sense to separate project-name from stack name, and the forward slash is already used as a separator.

I'm looking at fixing the pulumi-hugo repo as well, but wanted to wait for someone to comment/approve on the format before beginning that one.

## Checklist

### DIFF
--- a/pkg/cmd/pulumi/new.go
+++ b/pkg/cmd/pulumi/new.go
@@ -536,7 +536,7 @@ func promptAndCreateStack(prompt promptForValueFunc,
 	if b.SupportsOrganizations() {
 		fmt.Print("Please enter your desired stack name.\n" +
 			"To create a stack in an organization, " +
-			"use the format <org-name>/<stack-name> (e.g. `acmecorp/dev`).\n")
+			"use the format [<org-name>/<project-name>/]<stack-name> (e.g. `acmecorp/shipping/dev`).\n")
 	}
 
 	for {
--- a/pkg/cmd/pulumi/stack_init.go
+++ b/pkg/cmd/pulumi/stack_init.go
@@ -35,7 +35,7 @@ func newStackInitCmd() *cobra.Command {
 	var stackToCopy string
 
 	cmd := &cobra.Command{
-		Use:   "init [<org-name>/]<stack-name>",
+		Use:   "init [<org-name>/<project-name>/]<stack-name>",
 		Args:  cmdutil.MaximumNArgs(1),
 		Short: "Create an empty stack with the given name, ready for updates",
 		Long: "Create an empty stack with the given name, ready for updates\n" +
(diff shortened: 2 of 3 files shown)


iwahbe:
I read `[<org-name>/<project-name>/]<stack-name>` as either provide a fully qualified name (org + project + stack) or provide **only** a stack name. 

We are following the unix convention of `[]` marking something as optional. I prefer `[<org-name>/[<project-name>/]]<stack-name>`. It allows for the case `<org-name>/<project-name>`.

svangordon-fruit:
I think that we should leave this as it is. If `to create a stack in an organization` is understood to mean `to create a stack in an organization other than the default one`, then the org-name is indeed mandatory (because if omitted, the stack will be created in the default org).

Additionally, I would discourage asking the user for a project name at this prompt. `promptAndCreateStack` is called from two places, `pulumi up` and `pulumi new`. In both places, there is already a Pulumi.yaml file with a project name in the CWD, so there is only one project name that would be valid. If the user provides any project name other than the one in the Pulumi.yaml file it will result in an error.

The user is _allowed_ to provide a project name as part of a fully qualified stack name, but in both places where this prompt is used there is no benefit to a user doing so.

iwahbe:
@svangordon-fruit `pulumi up` requires a `Pulumi.yaml` file in the CWD. `pulumi new` actually forbids it (without `--force`).

svangordon-fruit:
@iwahbe At the point where [`promptAndCreateStack` is called in `new.go`](https://github.com/pulumi/pulumi/blob/272c4643b20d40113689475fd2f9fe3134e6d431/pkg/cmd/pulumi/new.go#L257), `workspace.SaveProject` [has already been called](https://github.com/pulumi/pulumi/blob/968f6596d9cfa8f6125358f2ed62c6e93986b1ed/pkg/cmd/pulumi/up.go#L258), writing the `Pulumi.yaml` file to disk.

iwahbe:
@svangordon-fruit Thanks for pointing this out!

Frassle:
I'm going to agree with @svangordon-fruit's comments about this. I don't think it's any clearer trying to suggest users add project names in these places. I'll close this off, but thank you for raising it.

PR #8580, opened 2021-12-10
Remove .gitignore for file we no longer generate

Frassle:
Praneet removed the generation of the packaged.go file in https://github.com/pulumi/pulumi/pull/8389 but I still had it sitting around in my local repo because it was still .gitignored. Was very confusing why "make lint" kept throwing up lint error even on a fresh master checkout.
This removes the .gitignore in that folder and also an out of date comment for one test that used that old file.

### DIFF
--- a/pkg/codegen/docs/.gitignore
+++ b/pkg/codegen/docs/.gitignore
@@ -1,1 +1,0 @@
-packaged.go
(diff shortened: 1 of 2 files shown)


t0yv0 (approved):
Reasonable.

PR #8583, opened 2021-12-10
Always RegisterOutputs even if empty

### LINKED ISSUE #8273: Outputs remain in stack after corresponding resource has been deleted

## Issue details

I have written a function that creates a resource and exports a value from that resource. I call this function several times in the course of a Pulumi program, creating the desired resources and generating the desired outputs. After some refactoring, however, I determined that these resources weren't needed in the current project, so I removed the function and all the call sites.

When I ran `pulumi up`, it deleted all the resources, as expected, and the diff output also indicated that all the corresponding stack outputs were also deleted. However, running `pulumi stack` showed one of the outputs remained, even though its corresponding resource was deleted. Repeated calls to `pulumi up` showed a diff output suggesting that the remaining output was deleted, but `pulumi stack` showed that it still remained.

A bit of experimentation with adding these resources back and deleting them again suggests that it wasn't any particular resource that was triggering it, but I always had one output left behind. Since my stack had no other outputs aside from these I'm describing, there may be some kind of edge condition I was triggering at the transition from 1 to 0 outputs, but that's pure speculation on my part.

I ultimately ended up fixing the problem by running `pulumi stack export`, editing the resulting JSON, and then running `pulumi stack import`.

I unfortunately cannot provide a reproduction case, but I was using the following software:
- Pulumi CLI 3.12.0
- Pulumi Python SDK 3.13.2
- Pulumi Hashicorp Vault Provider 4.5.0; the resources in question were instances of [pulumi_vault.identity.Entity](https://www.pulumi.com/registry/packages/vault/api-docs/identity/entity/)

t0yv0:
# Description

Fixes #8273 

The behavior is modified not just for Stack but for any ComponentResource. I cannot find a reason for the original `if` statement but it does seem reasonable to expect RegisterResourceOutputs to always be called. Let us see if the test suite passes with these changes.

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -2,3 +2,6 @@
 
 ### Bug Fixes
 
+- [sdk/python] - Fixes an issue with stack outputs persisting after
+  they are removed from the Pulumi program
+  [#8583](https://github.com/pulumi/pulumi/pull/8583)
--- a/sdk/python/lib/pulumi/resource.py
+++ b/sdk/python/lib/pulumi/resource.py
@@ -921,8 +921,7 @@ class ComponentResource(Resource):
 
         :param dict output: A dictionary of outputs to associate with this resource.
         """
-        if outputs:
-            register_resource_outputs(self, outputs)
+        register_resource_outputs(self, outputs)
 
 
 class ProviderResource(CustomResource):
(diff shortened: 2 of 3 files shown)


iwahbe (approved):
LGTM. 

---

I feel especially safe doing this because it's what we do for other sdks:

https://github.com/pulumi/pulumi/blob/8a73104cd690d21860e3afc0e7a33067e8311415/sdk/go/pulumi/context.go#L803

https://github.com/pulumi/pulumi/blob/8a73104cd690d21860e3afc0e7a33067e8311415/sdk/nodejs/resource.ts#L360

If you can find an appropriate place for it (I think we have a guide for implementing language sdks), it would be great to document that the "register resource (outputs)" is a non-optional call.

christophermaier:
Thanks very much @t0yv0 🎉

PR #8593, opened 2021-12-15
Use `pulumi-plugin.json` rather than `pulumiplugin.json`

### LINKED ISSUE #8592: Python azure-native program fails when built with the latest codegen
I'm updating azure-native's codegen from 3.19.0 to what's in master (commit 278055c101b808b46bf49e24b25c86017437ebba) in [this PR](https://github.com/pulumi/pulumi-azure-native/pull/1366). After the update, the Python program stops working with this panic:

> failed to discover plugin requirements: json: Unmarshal(nil *plugin.PulumiPluginJSON)

see https://github.com/pulumi/pulumi-azure-native/runs/4532362541?check_suite_focus=true#step:21:125

Do I need to change anything in the provider code?

justinvp:
Current/older versions of the Python language host cannot read `pulumiplugin.json` without returning an error. This hasn't been a problem because none of our packages have included the file. However, we're going to start including this file by default in packages. To avoid this error on older CLIs, we'll use a different file name. And since we're introducing the use of this file in the .NET and Go language hosts, we'll use the new name for these languages as well for consistency.

Fixes #8592


justinvp:
Definitely open to suggestions on the new filename. I went with `pulumi-plugin.json` but open to better naming ideas.

iwahbe (approved):
This solution looks good to me. LGTM

PR #8594, opened 2021-12-15
Add internal links

iwahbe:
# Description

This converts non-primitive types into links for the pagination to consume. 

## Checklist

### DIFF
--- a/pkg/cmd/pulumi/doc.go
+++ b/pkg/cmd/pulumi/doc.go
@@ -413,7 +413,10 @@ func renderDocstring(w io.Writer, docstring string) error {
 	document := parser.Parse(text.NewReader(source))
 
 	r := renderer.New(
-		renderer.WithSoftBreak(false))
-	renderer := goldmark_renderer.NewRenderer(goldmark_renderer.WithNodeRenderers(util.Prioritized(r, 100)))
+		renderer.WithSoftBreak(false),
+		renderer.WithHyperlinks(true),
+	)
+	renderer := goldmark_renderer.NewRenderer(
+		goldmark_renderer.WithNodeRenderers(util.Prioritized(r, 100)))
 	return renderer.Render(w, source, document)
 }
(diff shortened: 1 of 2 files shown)


pkg/cmd/pulumi/doc.go
```diff
@@ -322,12 +382,22 @@ func genFunctionDocstring(function *schema.Function, lang string, helper codegen
...
 	}
 	app := tview.NewApplication()
 	reader := newMarkdownReader(title, docstring, styles.Pulumi, app)
+	reader.externalLinkResolver = func(link string, reader *markdownReader) (bool, error) {
+		docstring, err := r.findDocstring(link)
+		if err != nil {
+			return true, err
+		}
+		currentPage, _ := reader.rootPages.GetFrontPage()
+		reader.backstack = append(reader.backstack, location{page: currentPage})
```
pgavlin, line 397: Might be nice to fetch the span of the currently selected node (if any) and  stick that in the back stack as well

pkg/cmd/pulumi/doc_reader.go
```diff
@@ -229,25 +269,38 @@ func (r *markdownReader) InputHandler() func(event *tcell.EventKey, setFocus fun
...
 						r.backstack = r.backstack[:len(r.backstack)-1]
-						r.view.SelectSpan(last, true)
+						if last.page != "" {
+							r.rootPages.SwitchToPage(last.page)
+						}
+						if last.span != nil {
+							r.view.SelectSpan(last.span, true)
+						}
+					}
+				case 'b':
```
pgavlin, line 286: So this brings up a dialog with the back stack in it? Nice.

pgavlin: With a little work to support widgets in dialog boxes, it would be possible to make this a menu...

pgavlin (approved):
One nit re: storing spans with pages in the backstack; LGTM otherwise!

PR #8610, opened 2021-12-18
Give user defined variables precedence in GetCommandResults

### LINKED ISSUE #8356: Re-order env vars in testing
## Hello!

## Issue details
In `testing` it would be convenient if [GetCommandResults](https://github.com/pulumi/pulumi/blob/e0a231013fb1f34a7e156a44a1b91c7b05802305/sdk/go/common/testing/environment.go#L181) would add [these env vars](https://github.com/pulumi/pulumi/blob/e0a231013fb1f34a7e156a44a1b91c7b05802305/sdk/go/common/testing/environment.go#L203-L209) in the reverse order

That would enable the caller to override the env vars that the  struct adds. That would make it easier to use the test harness to run things like behavior when there’s no valid file or when the stack passphrase isn’t set.

iwahbe:
# Description

Fixes #8356 

## Checklist


t0yv0 (approved):
This is neat and tidy, but as always, would be better with a tiny unit test, just to get in the habit of recording what we want the software to do.

t0yv0 (approved):
Thanks! LGTM.

PR #8617, opened 2021-12-20
Toward replacing MSBuild with make+bash on Windows

t0yv0:
# Description

The main goal of this PR is to have CI run Windows verifications using the same set of Bash+Makefiles that we use on Mac and Linux workers, instead of using a separately coded MSBuild logic as previously. 

It turns out that a lot of tests were never run on Windows (intentionally or not) and therefore require explicit skips or fixes for non-cross-platform builds.

As a positive side-effect of this refactor, test subset splitting to speed up CI runs now works for Windows in the same way as Linux and Mac.

Another change I included here that was making it easier to debug issues is making GHA aware of individual test suite steps instead of running one large `make test_all` target. At the cost of introducing some duplication in the GHA YAML vs Makefile itself, test failures are now isolated to an explicit step that shows which test suite they belong to, which to me makes it a lot faster to read and make sense of failing test output. It also is easy to see how much time each test suite takes.

The PR does not actually remove the MSBuild project yet as the change only touches PR verification workflow for starters, and will need to be promoted to the master+nightlies+release workflows to fully deprecate MSBuild.

- [x] Needs https://github.com/pulumi/pulumi/pull/8634 to fix some failures
- [ ] Fix to use binary installer for gotestsum (source installer is slow)
   - this can be deferred for later, we're just paying extra 30sec on builds, an annoyance 
- [x] Need to check if any of newly introduced windows skips can be relaxed to actually run the tests
   - turned out a bit non-trivial, logged follow up #8646 #8647 #8648 #8649  
- [ ] Verify if we want to keep sxs tests skip
   - it appears that the test is about TypeScript compatibility between HEAD and last-shipped Pulumi; it incurs unusually high cost as it is compiling a node dependency with c/cpp node_gyp; moreover it fails on Windows. I'd recommend keeping it skipped on Windows and even perhaps moving it out of PR verification flow into nightlies/master only, unless we can figure out caching that speeds it up a lot
- [ ] Verify if OK to skip code coverage on Windows (the trick with coverage-enabled Pulumi build is not working)
  - at this point it seems diminishing returns to push for coverage collection on Windows

Fixes # (issue)

## Checklist


build/common.mk
```diff
@@ -100,18 +100,17 @@ ifeq ($(PULUMI_ROOT),)
 	PULUMI_ROOT:=$(shell realpath "$$HOME/.pulumi-dev")
 endif
 
-# Use Python 3 explicitly vs expecting that `python` will resolve to a python 3
-# runtime.
-PYTHON ?= python3
-PIP ?= pip3
+# Expecting that `python`, `pip` will resolve to a python 3 runtime.
+PYTHON ?= python
```
t0yv0, line 104: Recent generations of Ubuntu ship python3 by default in `python` position. Windows workers do not have `python3`. Seems like a reasonable new default? But if we are not sure I can replace this with checking and auto-detection.

sdk/nodejs/Makefile
```diff
@@ -18,10 +18,10 @@ include ../../build/common.mk
 export PATH:=$(shell yarn bin 2>/dev/null):$(PATH)
 
 lint::
-	./node_modules/.bin/eslint -c .eslintrc.js --ext .ts .
+	eslint -c .eslintrc.js --ext .ts .
```
justinvp, line 21: Shouldn't this be `yarn run eslint`?

sdk/nodejs/Makefile
```diff
@@ -38,37 +38,46 @@ build_plugin::
...
 	GOBIN=$(PULUMI_BIN) go install -ldflags "-X github.com/pulumi/pulumi/sdk/v3/go/common/version.Version=${VERSION}" ${LANGUAGE_HOST}
 
 install:: install_package install_plugin
 
 unit_tests:: $(TEST_ALL_DEPS)
-	$(RUN_TESTSUITE) unit-nodejs ./node_modules/.bin/nyc -x 'bin/tests/runtime/*ClosureCases*.js' -s mocha --timeout 120000 --exclude 'bin/tests/automation/**/*.spec.js' --exclude 'bin/tests/runtime/closureLoader.spec.js' 'bin/tests/**/*.spec.js'
-	$(RUN_TESTSUITE) unit-nodejs mocha --timeout 120000 'bin/tests/runtime/closureLoader.spec.js'
-	$(RUN_TESTSUITE) unit-with-mocks-nodejs ./node_modules/.bin/nyc -s mocha 'bin/tests_with_mocks/**/*.spec.js'
+	$(TESTSUITE_SKIPPED) --skip-windows unit-nodejs || \
+		nyc -x 'bin/tests/runtime/*ClosureCases*.js' -s mocha --timeout 120000 \
```
justinvp, line 51: Shouldn't this be `yarn run nyc`?

sdk/nodejs/Makefile
```diff
@@ -38,37 +38,46 @@ build_plugin::
...
+			--exclude 'bin/tests/automation/**/*.spec.js' \
+			--exclude 'bin/tests/runtime/closureLoader.spec.js' \
+			'bin/tests/**/*.spec.js'
+	$(TESTSUITE_SKIPPED) --skip-windows unit-nodejs || \
+		mocha --timeout 120000 'bin/tests/runtime/closureLoader.spec.js'
```
justinvp, line 56: `yarn run mocha`?

Makefile
```diff
@@ -81,42 +81,26 @@ test_fast:: build
...
 test_build:: $(TEST_ALL_DEPS)
-	cd tests/testprovider && go build -o pulumi-resource-testprovider
-	cd tests/integration/construct_component/testcomponent && yarn install && yarn link @pulumi/pulumi && yarn run tsc
-	cd tests/integration/construct_component/testcomponent-go && go build -o pulumi-resource-testcomponent
-	cd tests/integration/construct_component/testcomponent-python && $(PYTHON) -m venv venv && venv/bin/python -m pip install -e ../../../../sdk/python/env/src
-	cd tests/integration/construct_component_output_values/testcomponent && yarn install && yarn link @pulumi/pulumi && yarn run tsc
-	cd tests/integration/construct_component_output_values/testcomponent-go && go build -o pulumi-resource-testcomponent
-	cd tests/integration/construct_component_output_values/testcomponent-python && $(PYTHON) -m venv venv && venv/bin/python -m pip install -e ../../../../sdk/python/env/src
+	cd tests/testprovider && go build -o pulumi-resource-testprovider$(shell go env GOEXE)
+	PYTHON=$(PYTHON) ./scripts/prepare-test.sh construct_component
```
justinvp, line 85: Thanks for cleaning this up!

justinvp (approved):
Some minor nits, but otherwise LGTM

build/common.mk
pgavlin: AFAIK OS X is still shipping Python 2.7 as `python`, so I think we need to handle that case.

Makefile
t0yv0: Happy to, I was forced to because cross-platform scripting requires sourcing `activate` and doing it in Make that runs one process per line was getting very, very unreadable.

sdk/nodejs/Makefile
t0yv0: Yeah that sounds safer. The reason it works now is `export PATH:=$(shell yarn bin 2>/dev/null):$(PATH)` but might as well do `yarn run`.

build/common.mk
t0yv0: Ah, true, interesting that the tests worked on macos-latest worker, perhaps that is because we reinstall Python on the worker. As for my local tests, I seem to be isolated from this as I install Python with Nix. I can handle this case.

t0yv0:
A bit of delay here due to very unstable/flaky test results, investigating.. Possible that TestGoAliases was failing on Windows due to "temp" folder "random" name collision across several tests.

PR #8618, opened 2021-12-20
Adding API Types for Bulk Decryption of Secrets

### LINKED ISSUE #6445: `pulumi stack` gets significantly slower with more number of secrets
Executing `pulumi stack` gets significantly slower as the number of secrets increase. 

The example in **Steps to reproduce** exhibits this, but this problem gets much worse when secrets are re-used across multiple resources and end up in the stack multiple times - e.g. with `@pulumi/kubernetes.`

A trace shows a call to `api/decryptValue` at least once for each secret in the state file.

## Expected behavior

I would expect some overall decrease in performance as the number of resources in a stack increase, but the slowdown is significant when there are a large number of secrets.

## Current behavior

Results of testing the performance of `pulumi stack` with increasing number of secrets:
```
   1 secrets -  0 seconds
  10 secrets -  0 seconds
 100 secrets -  2 seconds
 500 secrets - 14 seconds
1000 secrets - 31 seconds
```

## Steps to reproduce

1. `git clone https://gist.github.com/c162a4596061b8d92f9298bb70c9bbb1 p-secret-performance`
2. `cd p-secret-performance`
2. `npm i`
3. `npm run repro`

### Output

```
% npm run repro

> repro
> ./node_modules/ts-node/dist/bin.js repro.ts

Calling [pulumi up] with [1] secrets... Done.
Calling [pulumi stack] with [1] secrets took [0] seconds.

Calling [pulumi up] with [10] secrets... Done.
Calling [pulumi stack] with [10] secrets took [0] seconds.

Calling [pulumi up] with [100] secrets... Done.
Calling [pulumi stack] with [100] secrets took [2] seconds.

Calling [pulumi up] with [500] secrets... Done.
Calling [pulumi stack] with [500] secrets took [14] seconds.

Calling [pulumi up] with [1000] secrets... Done.
Calling [pulumi stack] with [1000] secrets took [31] seconds.
``` 

## Context (Environment)

Running on macOS, but I don't expect that it matters.

stack72:
This is a set of API types that I can use to fix #6445

### DIFF
--- a/sdk/go/common/apitype/stacks.go
+++ b/sdk/go/common/apitype/stacks.go
@@ -83,6 +83,17 @@ type Log3rdPartyDecryptionEvent struct {
 	CommandName string `json:"commandName,omitempty"`
 }
 
+// BulkDecryptValueRequest defines the request body for bulk decrypting secret values.
+type BulkDecryptValueRequest struct {
+	Ciphertexts [][]byte `json:"ciphertexts"`
+}
+
+// BulkDecryptValueResponse defines the response body for bulk decrypted secret values. The key in
+// the map is the base64 encoding of the ciphertext.
+type BulkDecryptValueResponse struct {
+	Plaintexts map[string][]byte `json:"plaintexts"`
+}
+
 // ExportStackResponse defines the response body for exporting a Stack.
 type ExportStackResponse UntypedDeployment
 


sdk/go/common/apitype/stacks.go
```diff
@@ -83,6 +83,16 @@ type Log3rdPartyDecryptionEvent struct {
...
 }
 
+// BulkDecryptValueRequest defines the request body for bulk decrypting secret values.
+type BulkDecryptValueRequest struct {
+	Secrets [][]byte `json:"secrets"`
+}
+
+// BulkDecryptValueResponse defines the response body for bulk decrypted secret values.
+type BulkDecryptValueResponse struct {
+	DecryptedSecrets map[string][]byte `json:"decryptedSecrets"`
```
Frassle, line 93: Not sure I know all this well enough to +1 but seems odd to me that we pass an array in and get a map out? Also everywhere else in thus file refers to Ciphertext and Plaintext, maybe this should match?

stack72: so the input is a list of Ciphertext in []byte and the response is a map of the ciphertext and decrypted secret

mikhailshilkov: Should it be `Ciphertexts` and `Plaintexts` then?

What is the `string` value in the map if all others are `[]byte`? Is there some implied conversion?

stack72: Folks, we are getting blocked on a naming here....

there is a conversion inside the service that sends the representation back.

mikhailshilkov: @Frassle Were you thinking something like this instead?

```
type BulkDecryptValueRequest struct {
	Ciphertexts [][]byte `json:"ciphertexts"`
}
type BulkDecryptValueResponse struct {
	Plaintexts [][]byte `json:"plaintexts"
}
```

Frassle: That's what I would expect given the rest of this file. Same names as elsewhere, no questions of how byte[] is turned into string.

Frassle (approved):

PR #8627, opened 2021-12-22
[sdk] Compute full set of aliases when both parent and child are aliased

### LINKED ISSUE #6440: Resource type alias doesn't work for one aliased resource parented to another aliased resource
Here is a repro program (with `azure-nextgen 0.6.0`, but I believe this is not provider-specific):

```ts
import * as azure from "@pulumi/azure-nextgen";

const resourceGroup = new azure.resources.v20200801.ResourceGroup("rg", {
    location: "westeurope",
    resourceGroupName: "dfsfsd",
})

const vnet = new azure.network.v20200801.VirtualNetwork("vnet", {
    resourceGroupName: resourceGroup.name,
    addressSpace: {
        addressPrefixes: ["10.2.0.0/16"],
    },
    location: "westeurope",
    virtualNetworkName: "vnet14342"
}, {parent: resourceGroup});
```

Provision the resources with `pulumi up`. Now, change the code to use the `latest` instead of `v20200801` for both resources.

Both resources have aliases between those versions defined. The alias is successfully applied to `ResourceGroup` and it's not replaced. However, the alias is not applied to the `VirtualNetwork` resource. The preview shows the vnet-to-be-deleted at the top-level:

```
     pulumi:pulumi:Stack                                k8s-azure-01-infrastructure-dev             
     ├─ azure-nextgen:resources/latest:ResourceGroup    rg                                          
 +   │  └─ azure-nextgen:network/latest:VirtualNetwork  vnet                             create     
 -   └─ azure-nextgen:network/v20200701:VirtualNetwork  vnet                             delete
 ```

Workaround:

1. Change `ResourceGroup` but not `VirtualNetwork`.
2. `pulumi up` - yes to no changes.
3. Change `VirtualNetwork`.
4. `pulumi up` - yes to no changes.

### PR DESCRIPTION
**Pull request by lukehoban**
When both the parent and child were aliased, with M and N aliases respectively, we previously computed M+N aliases.  However, all pairs of parent and child aliases must also be considered, as those all represent possible names the resource might have had in past state files.  So we must compute the (M+1)*(N+1)-1 aliases for each child and parent alias (plus not aliased).

For example, for M=N=2, this change adds the 4 entries marked as "FIXED" below.

|      |child|childAlias1|childAlias2|
|------------|-----|-----------|-----------|
|parent      |NO ALIAS   |OK         |OK         |
|parentAlias1|OK   |FIXED     |FIXED     |
|parentAlias2|OK   |FIXED     |FIXED     |

Fixes included:
- [x] Node.js
- [x] Go
- [x] Python
- [x] .NET

Fixes #6440.

### CODE REVIEW

File: sdk/go/pulumi/context.go [hunk 1]
```diff
@@ -966,6 +962,30 @@ func (ctx *Context) collapseAliases(aliases []Alias, t, name string, parent Reso
 		aliasURNs = append(aliasURNs, urn)
 	}
 
+	if parent != nil {
+		parentAliases := parent.getAliases()
+		for i := range parentAliases {
```
t0yv0, line 967: Nit:

```go
    for i, parentAlias := range parent.getAliases() {
    }
```

t0yv0 (approved):
Implementation looks fine provided the root-cause analysis is correct. I am unfortunately noting that the aliases feature appears not to be documented. Places checked:

1. google `pulumi aliases`
2. pulumi.com, docs, search: "aliases"
3. https://pulumi-developer-docs.readthedocs.io/en/latest/index.html developer docs, search: aliases
4. Go SDK code, find `aliases` on ResourceOptions, no comments

I did find a blog post via 2 which was my best reference to try to understand the intent behind the feature:

https://www.pulumi.com/blog/cumundi-guest-post/

t0yv0:
Would you like us to pick up getting the rest of languages across the finish line?

lukehoban:
> the aliases feature appears not to be documented

Aliases are documented along with the rest of our supported resource options at https://www.pulumi.com/docs/intro/concepts/resources/#aliases. This section unfortunately does not explicitly call out the interaction of parent aliases on children - we should update to include that (probably after this PR lands so that we can describe the correct behaviour). 

For me - that page is the second organic Google link and the first docs site search link for the terms you linked - though that page is much too big and these resource options sections are H3s on the page which don’t lead to great search summaries. We should break that page up into many separate pages to help with search and discoverability.

t0yv0:
Ah, very good, I will go read the page carefully. Good to have it.

sdk/go/pulumi/context.go
lukehoban: Unfortunately - this is actually required to be this way because `parentAlias` is captured in a closure, and If written the way you suggest, Go overwrites the captured loop variable (so that all applys see the final element). Several other ways we could write this correctly, but this felt the simplest.

t0yv0: Ah yes, by-reference loop variable closure capture, the new 1bln mistake.. Thanks!

lukehoban:
> codecov/patch — 43.10% of diff hit (target 58.80%)

This doesn't seem correct, there are tests that cover all of the new code here.  The reports uploaded to codecov seem to be missing all of the test coverage in the TypeScript and Python tests added with this PR.  I'll assume there is an unrelated problem in the codecov analysis that we can follow up on separately.

lukehoban:
Documentation updates to include details of how parent aliases are inherited by children at https://github.com/pulumi/pulumi-hugo/pull/939.

t0yv0 (approved):

mikhailshilkov:
We are getting some extreme cases of impact of this change in the wild: https://github.com/pulumi/pulumi/issues/9089

pgavlin:
> We are getting some extreme cases of impact of this change in the wild: https://github.com/pulumi/pulumi/issues/9089

Two thoughts off the top of my head:
1. I'm unsure why we store aliases in the state file at all. We should revisit that and see whether or not it's necessary. Offhand it seems like we only need aliases in the `RegisterResourceRequest` as they're used to lookup a resource's old state.
2. We should revisit how we represent aliases on the wire (and in state files if we do in fact need them there). I suspect that we might be able to represent them in the terser format that we use in the SDKs rather than as a list of URNs, then allow consumers to expand to a list of URNs on demand.

PR #8628, opened 2021-12-22
Don't crash on null inputs for value types in IOutputCompletionSource.SetValue(OutputData<object?> data)

### LINKED ISSUE #8625: [sdk/dotnet] NullReferenceException on output type mismatch
Raised in https://github.com/pulumi/pulumi-azure-native/issues/1381 and I can reproduce it locally.

```
System.NullReferenceException: Object reference not set to an instance of an object.
       at void Pulumi.Serialization.OutputCompletionSource<T>.SetValue(OutputData<object> data)
       at async Task Pulumi.Deployment.CompleteResourceAsync(Resource resource, bool remote, Func<string, Resource> newDependency, ResourceArgs args, ResourceOptions options, ImmutableDictionary<string, IOutputCompletionSource> completionSources)
 
  azure-native:cdn:Route (FrontDoor-Route-AdministratorPrimary):
    warning: Expected System.Collections.Immutable.ImmutableArray`1[[System.Object, System.Private.CoreLib, Version=5.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]] but got System.Collections.Immutable.ImmutableDictionary`2[[System.String, System.Private.CoreLib, Version=5.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[System.Object, System.Private.CoreLib, Version=5.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]] deserializing Pulumi.AzureNative.Cdn.Route.compressionSettings
```

The context is that an output property of a resource is expected to be of type `ImmutableArray<T>` but it's actually `T` (due to a mistake in Open API).

#8286 solved the conversion exception (it's now a warning) but it looks like we fail down the line.

Frassle:
SetValue correctly annotates that the object held in data might be null.
But then ignores that annotation and tries to cast to `T`, if the object
is null this is fine for reference types which just stay null, but for
value types this throws a NullReferenceException. Simple fix to just use
`default(T)` for when `data.Value` is null, and only cast when non-null.

This has the same behaviour as before for reference types (where
`default(T)` is null) but means that value types take their default
value rather than throwing.

Fixes #8625

### DIFF
--- a/sdk/dotnet/Pulumi/Serialization/OutputCompletionSource.cs
+++ b/sdk/dotnet/Pulumi/Serialization/OutputCompletionSource.cs
@@ -45,7 +45,7 @@ namespace Pulumi.Serialization
 
         public void SetValue(OutputData<object?> data)
             => _taskCompletionSource.SetResult(new OutputData<T>(
-                _resources.Union(data.Resources), (T)data.Value!, data.IsKnown, data.IsSecret));
+                _resources.Union(data.Resources), (data.Value == null ? default(T) : (T)data.Value)! , data.IsKnown, data.IsSecret));
 
         public void TrySetDefaultResult(bool isKnown)
             => _taskCompletionSource.TrySetResult(new OutputData<T>(
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -27,3 +27,6 @@
 - [sdk/python] - Fixes an issue with stack outputs persisting after
   they are removed from the Pulumi program
   [#8583](https://github.com/pulumi/pulumi/pull/8583)
+
+- [sdk/dotnet] - Don't throw converting value types that don't match schema
+  [#8628](https://github.com/pulumi/pulumi/pull/8628)


sdk/dotnet/Pulumi/Serialization/OutputCompletionSource.cs
```diff
@@ -45,7 +45,7 @@ public void SetStringValue(string value, bool isKnown)
 
         public void SetValue(OutputData<object?> data)
             => _taskCompletionSource.SetResult(new OutputData<T>(
-                _resources.Union(data.Resources), (T)data.Value!, data.IsKnown, data.IsSecret));
+                _resources.Union(data.Resources), (data.Value == null ? default(T) : (T)data.Value)! , data.IsKnown, data.IsSecret));
```
mikhailshilkov, line 48: What exactly caused NRE in the old code? I'd guess it's `data` that would have to be `null` for NRE, which means the fix won't help?

Frassle: Nope, data.Value was null and T was a value type. `(int)null` will throw an NRE rather than returning `default(int)`.

mikhailshilkov: Oh, interesting. Thank you for the explanation!

mikhailshilkov (approved):

PR #8634, opened 2021-12-23
Fix creds file locking on Windows

t0yv0:
# Description

Investigating sporadic test failures on Windows has led me to this discovery. TestConcurrentCredentialsWrites had issues as it depended on the state of credentials file in the environment. Specifically in the lucky case the environment had no credentials therefore it was testing this no-op branch that does not demonstrate contention:

```
	if len(creds.AccessTokens) == 0 {
		err = os.Remove(credsFile)
		if err != nil && !os.IsNotExist(err) {
			return err
		}
		return nil
	}
```

Fixing the test to demonstrate contention makes it fail on Windows with "access denied" when it tries to do os.Rename.

I shopped around again for a package to do reliable atomic file writes across platforms. I have tried using this package that we rely on elsewhere in the Pulumi codebase:

```
	"github.com/gofrs/flock"
```

But it did not work for me, I would still have Windows failures.

It seems our best bet may be using file locking based on the algorithm used by Go toolchain internally. It's an `internal` package therefore cannot be reused but there is this 	"github.com/rogpeppe/go-internal/lockedfile" package that re-exports Go internals.

With this change, the test seems to work allright across platforms.

Fixes # (issue)

## Checklist

### DIFF
--- a/pkg/go.mod
+++ b/pkg/go.mod
@@ -162,3 +162,5 @@ require (
 	gopkg.in/warnings.v0 v0.1.2 // indirect
 	gopkg.in/yaml.v2 v2.2.8 // indirect
 )
+
+require github.com/rogpeppe/go-internal v1.8.1 // indirect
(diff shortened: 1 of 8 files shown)


pkg/go.mod
```diff
@@ -162,3 +162,5 @@ require (
 	gopkg.in/warnings.v0 v0.1.2 // indirect
 	gopkg.in/yaml.v2 v2.2.8 // indirect
 )
+
+require github.com/rogpeppe/go-internal v1.8.1 // indirect
```
lukehoban, line 166: Should this be moved into the main require block?

lukehoban (approved):
LGTM

pkg/go.mod
t0yv0: This change is made by `go mod tidy` command so apparently this is how it likes it, I suggest we keep it to reduce diffs from when repeatedly applying `go mod tidy`.

Issue #7269, opened 2021-06-10
Flaky TestLocalStateLocking test


t0yv0:
Flaky test: func TestLocalStateLocking(t *testing.T) {

## Steps to reproduce

https://github.com/pulumi/pulumi/runs/2784480323?check_suite_focus=true#step:17:2116

t0yv0:
https://github.com/pulumi/pulumi/runs/3195813812?check_suite_focus=true

t0yv0:
That's right. I've worked on an "atomic file write" fix in https://github.com/pulumi/pulumi/pull/7739 but it seems that Go library authors believe the atomic file write is not possible on Windows in principe. We should also consider file locks.

PR #8642, opened 2021-12-28
[backend/filestate] Allow preview on locked stack

### LINKED ISSUE #8609: preview fails if state is locked in non-service backend
## Hello!

## Issue details

When using a file backend (e.g. file, s3, etc), a `preview` fails with an error if the state is locked. This is different behavior from using the (default) Pulumi Service backend which allows the preview to run as expected.

### Steps to reproduce

1. `pulumi login --local`
2. `pulumi new aws-typescript --stack p-lock-test`
3. `echo '{"pid":1,"username":"a","hostname":"b","timestamp":"2021-12-17T13:17:34.891106-08:00"}' > ~/.pulumi/locks/p-lock-test/test.json`
2. `pulumi pre`
    - ⚠️ observe error 
3. `rm ~/.pulumi/locks/p-lock-test/test.json`

Expected: `pre` should execute 
Actual: `pre` errors with a similar error below

#### Full output

```
k8s-ts-cluster % pulumi login --local
Logged in to Camerons-MacBook-Pro.local as clstokes (file://~)

k8s-ts-cluster % echo '{"pid":1,"username":"a","hostname":"b","timestamp":"2021-12-17T13:17:34.891106-08:00"}' > ~/.pulumi/locks/p-lock-test/test.json

k8s-ts-cluster % PULUMI_SELF_MANAGED_STATE_LOCKING=1 pulumi pre
error: the stack is currently locked by 1 lock(s). Either wait for the other process(es) to end or manually delete the lock file(s).
  file:///Users/clstokes/.pulumi/locks/p-lock-test/test.json: created by a@b (pid 1) at 2021-12-17T13:17:34-08:00
```

#### Code hints

These seem to be the code blocks that differ in behavior:

- Service backend code - https://github.com/pulumi/pulumi/blob/8a73104cd690d21860e3afc0e7a33067e8311415/pkg/backend/httpstate/backend.go#L830-L839
- S3/file backend code - https://github.com/pulumi/pulumi/blob/8a73104cd690d21860e3afc0e7a33067e8311415/pkg/backend/filestate/backend.go#L457-L472

lukehoban:
The `httpstate` backend allows previews to proceed even while an update is in progress.  This is potentially problematic as the preview may be relative to a partial state of the stack, but ensures that previews will not be blocked on stacks that have long running updates (for example, to allow for concurrent PR jobs to preview changes).  This behaviour has been consistent ~forever for the `httpstate` backend.

In the `filestate `backend, we recently introduced locking, using a quite different (more coarse-grained) approach. As part of this implementation, preview was added to the list of operations that require an exclusive lock on the stack.

For consistency, we should loosen this so that preview behaves the same relative to state locking in the `filestate` and `httpstate` backends.

In the future, we may well want to tighten this up for both backends, with some additional user controls.  Also, when the update plans feature lands shortly, that will provide some additional helpful guarantees that a previous preview was not accidentally relative to a partial state.

Fixes #8609.

### DIFF
--- a/tests/stack_test.go
+++ b/tests/stack_test.go
@@ -423,9 +423,10 @@ func TestLocalStateLocking(t *testing.T) {
 	e.RunCommand("yarn", "install")
 	e.RunCommand("yarn", "link", "@pulumi/pulumi")
 
-	// Run 10 concurrent updates
 	count := 10
 	stderrs := make(chan string, count)
+
+	// Run 10 concurrent updates
 	for i := 0; i < count; i++ {
 		go func() {
 			_, stderr, err := e.GetCommandResults("pulumi", "up", "--non-interactive", "--skip-preview", "--yes")
@@ -456,6 +457,25 @@ func TestLocalStateLocking(t *testing.T) {
 			}
 		}
 	}
+
+	// Run 10 concurrent previews
+	for i := 0; i < count; i++ {
+		go func() {
+			_, stderr, err := e.GetCommandResults("pulumi", "preview", "--non-interactive")
+			if err == nil {
+				stderrs <- "" // success marker
+			} else {
+				stderrs <- stderr
+			}
+		}()
+	}
+
+	// Ensure that all of the concurrent previews succeed.
+	for i := 0; i < count; i++ {
+		stderr := <-stderrs
+		assert.Equal(t, "", stderr)
+	}
+
 }
 
 func getFileNames(infos []os.FileInfo) []string {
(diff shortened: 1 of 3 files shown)


tests/stack_test.go
```diff
@@ -456,6 +457,25 @@ func TestLocalStateLocking(t *testing.T) {
 			}
 		}
 	}
+
+	// Run 10 concurrent previews
```
lukehoban, line 461: Added test coverage here - but this test is skipped currently due to #7269.  I verified locally that this test correctly fails before the change and succeeds after the change when un-skipped.

t0yv0 (approved):
🚢

PR #8643, opened 2021-12-28
Python Auto API fix to clear RPC_MANAGER

### LINKED ISSUE #8633: Automation API: Python in-line throws `Resource Monitor has terminated, shutting down` error for all stacks after a stack error.
## Hello!

## Issue details

An automation api program with inline programs throws `Resource Monitor has terminated, shutting down` error for stacks after a stack failure and marks the stacks as failed even though the subsequent stacks successfully deploy all the resources.

### Steps to reproduce

1. Use attached program.
2. Set the azure credentials (lines 7 - 8)
3. python auto-api-program.py
4. You will see two runs of a randomly named stack.
5. The first stack will create 5 of 8 resources and fail - on purpose to cause the condition 
6. The second or more runs will create all 8 resources and indicate failure even though it's fine.
7. The stack is in a failed state.
8. To clean up, you can delete the resource group in Azure and then force delete the stack.

Expected: 
The subsequent stack.ups should not fail.
Actual: 
They fail.

[auto_api_program.py.txt](https://github.com/pulumi/pulumi/files/7770971/auto_api_program.py.txt)

t0yv0:
# Description

Clearing RPC_MANAGER prevents state contamination between several Automation API operations that intend to be distinct, such as exception poisoning from failed up 1 to otherwise successful up 2.

Fixes #8633 for Python

- [x] add test

Separate PRs needed for:

- [ ] check node
- [ ] check dotnet
- [ ] check go

## Checklist

### DIFF
--- a/sdk/python/lib/pulumi/automation/_server.py
+++ b/sdk/python/lib/pulumi/automation/_server.py
@@ -22,6 +22,7 @@ from ._workspace import PulumiFn
 from .. import log
 from ..runtime.proto import language_pb2, plugin_pb2, LanguageRuntimeServicer
 from ..runtime import run_in_stack, reset_options, set_all_config
+from ..runtime.rpc_manager import RPC_MANAGER
 from ..errors import RunError
 
 _py_version_less_than_3_7 = sys.version_info[0] == 3 and sys.version_info[1] < 7
@@ -98,6 +99,7 @@ class LanguageServer(LanguageRuntimeServicer):
             loop.close()
             sys.stdout.flush()
             sys.stderr.flush()
+            RPC_MANAGER.clear()
 
         return result
 
--- a/sdk/python/lib/pulumi/automation/_server.py
+++ b/sdk/python/lib/pulumi/automation/_server.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import asyncio
+import logging
 import sys
 import traceback
 from contextlib import suppress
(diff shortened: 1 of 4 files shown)


sdk/python/lib/pulumi/automation/_server.py
```diff
@@ -98,6 +99,7 @@ def Run(self, request, context):
             loop.close()
             sys.stdout.flush()
             sys.stderr.flush()
+            RPC_MANAGER.clear()
```
lukehoban, line 102: Should the doc comment on `clear` be updated now that this is not just used for tests?

lukehoban (approved):
Can we add test coverage for this?

joeduffy:
Note that I believe this is a problem in (at least) our NodeJS implementation too (for a different reason). Can we add tests for and fix the other runtimes too?

t0yv0:
Thanks! Adding tests now.  I will check Node JS and Go and Python as well, putting up PRs for those if needed - sounds like definitely needed for Node.

t0yv0:
Notes on suppressing unobserved task logging.. Not particularly happy about this but it seems a pragmatic way forward. 

Currently Pulumi SDK (and Automation API) use multiple event loops. Some of these are generated by threading (one loop per thread in multi-threaded servers), some are generated by our own code create_event_loop(). If any tasks on any of those loops are abandoned (not awaited or observed) and terminate with exceptions, `asyncio` generates error output at program termination warning about these tasks. This is exactly what happens in the scenario covered by the test (among other scenarios).

The workaround is to suppress logging. We do the same in pulumi-language-python.

The proper fix would be to be hygienic about handling exceptions on the loop. However that currently makes me hesitate since it appears some of the code makes incorrect assumptions about finding the right loop on the "thread-default" slot and making changes to the loop structure is prone to exacerbate race conditions or hangs. In the long term I'd like to make careful progress on that front perhaps ending in a world of 1 thread, 1 loop, and no deadlocks except when explicitly introduced by user programs:

https://github.com/pulumi/pulumi/issues/7663
https://github.com/pulumi/pulumi/issues/8650
https://github.com/pulumi/pulumi/issues/6762

Issue #8637, opened 2021-12-27
[sdk/dotnet] Implicit `pulumi:pulumi:Stack` parent is included explicitly in resource URNs in mock tests


lukehoban:
The following test:

```cs
    class AliasesStack : Stack
    {
        public AliasesStack()
        {
            var parent1 = new Pulumi.CustomResource("test:resource:type", "myres1", null, new CustomResourceOptions { });
         }
    }

    class AliasesMocks : IMocks
    {
        public Task<object> CallAsync(MockCallArgs args)
        {
            return Task.FromResult<object>(args);
        }

        public async Task<(string? id, object state)> NewResourceAsync(MockResourceArgs args)
        {
            await Task.Delay(0);
            return ("myID", new Dictionary<string, object>());
        }
    }
```

```cs
        [Fact]
        public async Task TestAliases()
        {
            var resources = await Deployment.TestAsync<Aliases.AliasesStack>(new Aliases.AliasesMocks());
            var parent1Urn = await resources[1].Urn.GetValueAsync("");
            Assert.Equal("urn:pulumi:stack::project::test:resource:type::myres1", parent1Urn);
        }
```

Fails because the parent URN is actually `"urn:pulumi:stack::project::pulumi:pulumi:Stack$test:resource:type::myres1"`.  This is not expected, as `pulumi:pulumi:Stack` is the implicit parent, and should not be included explicitly in resource URNs.

PR #8651, opened 2021-12-29
Fix master build

t0yv0:
# Description

The assert fails for reasons outlined in #8637. The issue is test-only as in a non-mock context the URNs are allocated as expected. Commenting out the assert so that master builds stop failing.

Fixes # (issue)

## Checklist


lukehoban (approved):
Thanks 🙏

Issue #5161, opened 2020-08-11
Support top-level `await` in Node.js


lukehoban:
Support for top-level await was just made available by default in Node `v14.8.0` (https://github.com/nodejs/node/pull/34558).  We should enable this to be used in Pulumi programs, and look into what we need to do in terms of Node versions, flags for older Node versions, TypeScript versions, TypeScript compilation targets, and module system configuration to be able to (a) allow users to use top-level await (b) make this the default mode for new projects and update existing examples to take advantage of this where possible.

keithharvey:
Any progress on this? Starting to port my company onto Pulumi and the combination of 2.0 forcing Promise<Type> returns and the crosswalk files being out of date creates friction.

suonto:
Has there been any progress in this?

thanhtoan1196:
any updates?

lukehoban:
This is now supported for JavaScript and pre-compiled TypeScript programs via support for ESM added in https://github.com/pulumi/pulumi/pull/7764.  Some details on how to use this in https://github.com/pulumi/pulumi/issues/4130#issuecomment-1002868483.

Additional support for features that enable this to be used via on-the-fly-compiled TypeScript programs (the most common use case in Pulumi) is coming in https://github.com/pulumi/pulumi/pull/8655.

awoimbee:
This is now supported for on the fly compiled typescript.

How to:
- update all your imports (https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c#file-esm-package-md)
- Add to `package.json`:
  ```json
  {
      # ...
      "type": "module",
      "dependencies": {
          "ts-node": "^10.4.0",
          # ...
      }
  }
  ```
- Add to `tsconfig.json`:
  ```json
      "target": "ESNext",
      "module": "ESNext",
  ```
- Add to `pulumi.yaml`:
  ```yaml
  runtime:
    name: nodejs
    options:
      # See https://github.com/TypeStrong/ts-node/issues/1007
      nodeargs: "--loader ts-node/esm --no-warnings"
  ```

lukehoban:
[...]

PR #8655, opened 2021-12-29
[langhost/node] More ESM support: `nodeargs`, main entrypoints, and tests

lukehoban:
This follows up on the changes in https://github.com/pulumi/pulumi/pull/7764 with three additions:

* Support for passing args directly to `node` so that ESM loaders and other node process level configuration can be accessed via Pulumi
* Support the same default behaviour for loading a ESM module from Pulumi as for `node <program>`, including support for resolving main entrypoint from package folder.
* Add test cases for `.js`, pre-complied `.ts` and `ts-node`-based `.ts.

This includes test cases that show how to use top-level await in Node.js (which is only possible inside ESM), addressing https://github.com/pulumi/pulumi/issues/5161.  

Using ESM via the default `ts-node`-based TypeScript support is a little tricky, as it is dependent on experimental loader hook support in Node, upon which partially in-progress ts-node support has been added.  We cannot make these the default experience yet, but the examples here show how users can configure things themselves to access these features.  Once this support solidifies and we can rely on it in all supported Node and TypeScript versions, we may be able to update templates to support more of this by default.

### DIFF
--- a/sdk/nodejs/cmd/pulumi-language-nodejs/main.go
+++ b/sdk/nodejs/cmd/pulumi-language-nodejs/main.go
@@ -83,6 +83,7 @@ func main() {
 	var typescript bool
 	var root string
 	var tsconfigpath string
+	var nodeargs string
 	flag.StringVar(&tracing, "tracing", "",
 		"Emit tracing to a Zipkin-compatible tracing endpoint")
 	flag.BoolVar(&typescript, "typescript", true,
@@ -90,6 +91,7 @@ func main() {
 	flag.StringVar(&root, "root", "", "Project root path to use")
 	flag.StringVar(&tsconfigpath, "tsconfig", "",
 		"Path to tsconfig.json to use")
+	flag.StringVar(&nodeargs, "nodeargs", "", "Arguments for the Node process")
 	flag.Parse()
 
 	args := flag.Args()
@@ -121,7 +123,7 @@ func main() {
 	// Fire up a gRPC server, letting the kernel choose a free port.
 	port, done, err := rpcutil.Serve(0, nil, []func(*grpc.Server) error{
 		func(srv *grpc.Server) error {
-			host := newLanguageHost(nodePath, runPath, engineAddress, tracing, typescript, tsconfigpath)
+			host := newLanguageHost(nodePath, runPath, engineAddress, tracing, typescript, tsconfigpath, nodeargs)
 			pulumirpc.RegisterLanguageRuntimeServer(srv, host)
 			return nil
 		},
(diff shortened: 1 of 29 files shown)


sdk/nodejs/cmd/run/run.ts
```diff
@@ -251,6 +256,28 @@ ${defaultMessage}`);
...
+                const mainPath: string = require("module").Module._findPath(path.resolve(program), null, true) || program;
+                const main = path.isAbsolute(mainPath) ? url.pathToFileURL(mainPath).href : mainPath;
+                // Workaround for typescript transpiling dynamic import into `Promise.resolve().then(() => require`
+                // Follow this issue for progress on when we can remove this:
+                // https://github.com/microsoft/TypeScript/issues/43329
+                //
+                // Workaround inspired by es-module-shims:
+                // https://github.com/guybedford/es-module-shims/blob/main/src/common.js#L21
+                // eslint-disable-next-line no-eval
+                const dynamicImport = (0, eval)("u=>import(u)");
```
lewisl9029, line 274: Some new developments since I opened the original PR: 

According to https://github.com/microsoft/TypeScript/issues/43329#issuecomment-926931937, we should be able to rid of this workaround and use `import(main)` directly if we upgraded to TypeScript 4.5 and used `module: node12`.

lukehoban: > if we upgraded to TypeScript 4.5 and used module: node12, which preserves dynamic imports instead of transpiling them into requires.

I believe `module: node12` was removed from TypeScript 4.5, and is only available in nightlies: https://github.com/microsoft/TypeScript/issues/46454.  I'm also not too worried about the workaround you used here, and shifting to a newer TypeScript version and especially newer approach to compilation of modules for this library is likely to introduce it's own interesting side effects.  I'm happy to wait on that to handle separately, most likely after a TypeScript release with this new `module` flag is released.

t0yv0:
Unfortunately running out of time to properly review this till Monday.

sdk/nodejs/cmd/pulumi-language-nodejs/main.go
```diff
@@ -513,15 +517,21 @@ func (host *nodeLanguageHost) execNodejs(
 			env = append(env, "PULUMI_NODEJS_TSCONFIG_PATH="+host.tsconfigpath)
 		}
 
+		var nodeargs []string
+		if host.nodeargs != "" {
+			nodeargs = strings.Split(host.nodeargs, " ")
```
t0yv0, line 522: Could be an issue if any args use space inside quotes or such, but presumably that'd rarely be the case.

t0yv0:
It's looking good but I'm having trouble running tests locally and verifying a few things due to https://github.com/pulumi/pulumi/issues/8678 - looking into that first.

t0yv0 (approved):
Excellent, working for me now. Thank you! LGTM.

sdk/nodejs/cmd/pulumi-language-nodejs/main.go
lukehoban: Good point.  Probably unlikely - but I've pulled in https://pkg.go.dev/github.com/google/shlex to ensure this works correctly.

tenwit:
@lukehoban Does this change (specifically the nodeargs part, allowing the passing of various --experimental-* flags to node) enable full support for tsconfig's baseUrl/paths? It seems like I'm getting closer to getting this working, but I'm not all the way there.
If it does enable better support for shiny new TS features, would it be appropriate to update the docs at https://www.pulumi.com/docs/intro/languages/javascript/#typescript to explain how to do it? I'm stuck... :(

t0yv0:
Would it be possible to file this as an issue? That'd be very helpful for us.

PR #8660, opened 2021-12-30
Deterministic builds for goreleaser

t0yv0:
# Description

Unfortunately file mtimes affect archive checksums. Before this change, running `goreleaser` twice on the same machine and same file-set would produce a new checksum for the generated archives every single run, that is, checksum was not reproducible from the file-set. The changes fix that.

It is my hope that with the changes we can continue using goreleaser for post-release utility steps rather than recoding them, by running it twice but checking checksums from the second run against the ones in the first run.

Fixes # (issue)

## Checklist


t0yv0:
Allright, looks like touch -d is not portable, have to script that differently.

Frassle (approved):
Looks sane to me

PR #8662, opened 2021-12-30
Do not polute worktree in build.yml; this fails strict goreleaser checks

t0yv0:
# Description

Tiny improvement to avoid failing `build.yml` under strict goreleaser checks that considers the newly created `goreleaser.current.yml` to be "dirty Git worktree".

Fixes # (issue)

## Checklist


t0yv0:
Close in favor of https://github.com/pulumi/pulumi/pull/8665

Issue #8629, opened 2021-12-22
Ensure code coverage results are deterministic


lukehoban:
Currently, code coverage frequently changes in files not touched by a PR.  That leads to misleading reports of coverage loss or gain, and also to an overall level of noise that's larger than the real change in most PRs.  We should audit tests to ensure we reduce coverage noise to a bare minimum, to ensure we can rely on the data we see from code coverage analysis.

lukehoban:
Also - many code coverage results just seem to be "wrong" in meaningful ways.  Almost all PRs show as reducing coverage even though they don't, miss out on coverage introduced in the tests, include loss of coverage for files unmodified in the PR, etc.  

We should do a deep review of the correctness of the data collection, and ensure it is meaningful.

lukehoban:
Marking as P1, as this leads to "build failures" on PRs, and has marked every PR as failing for the last several weeks.

leezen:
This is only leading us to fail the coverage check, which isn't currently gating merge, correct? These failures aren't being classified as build failures nor blocking the build unless I'm missing something? Agree we want to make sure we tune this such that we can rely on the check itself, but wanted to make sure I'm not missing anything.

lukehoban:
Yes - it doesn't block, but does mark every PR as failed.  And adds a lot of incorrect "noise" on every PR.  

![Uploading Screen Shot 2022-01-04 at 4.24.20 PM.png…]()

joeduffy:
It also makes the repo just red overall. For instance, Anton tried to make things green, only to be foiled by the non-deterministic and incorrect code coverage report:

![image](https://user-images.githubusercontent.com/3953235/148151082-145ea544-0a90-490c-81ce-34958f0bfd99.png)

That's as much of a bummer as the PRs looking red too.

leezen:
Removing `p1` as we've made these informational for now.

EvanBoyle:
Go 1.20 now supports coverage profiles for integration tests that span multiple binary invocations: https://go.dev/testing/coverage/

EvanBoyle:
An interesting idea if we wanted to start generating code coverage from an e2e suite such as examples we could:

- as a part of all publishing workflows, build and publish a binary that is built with coverage info (for language plugins, for all providers) and publish these to S3 under a consistent naming convention
[...]

PR #8671, opened 2021-12-31
Fix #3982, missing output which lacks newlines

### LINKED ISSUE #3982: Last Go Printf output missing if no newline
#### Problem

It seems that if the last `fmt.Printf("")` in my Go pulumi application does not end with a newline, it's output never gets printed/flushed to the user's terminal.

#### Suggestion

Print all console output.

#### Steps to reproduce

1. `preview` the code below

#### Code to reproduce

```go
package main

import (
	"fmt"

	"github.com/pulumi/pulumi-aws/sdk/go/aws/s3"
)

func main() {
		bucket, err := s3.NewBucket(ctx, "my-bucket", nil)
		if err != nil {
			return err
		}

		fmt.Printf("1 - with newline\n")
		fmt.Printf("2 - no newline")
		fmt.Printf("3 - with newline\n")
		fmt.Printf("4 - no newline")
		// fmt.Printf("\n") // <-- uncomment this to see the "4 - no newline" get printed

		return nil
	})
}
```

#### Output from the code above

```
aws-go-printf % pulumi preview
Previewing update (dev):
     Type                 Name               Plan       Info
 +   pulumi:pulumi:Stack  aws-go-printf-dev  create     2 messages
 +   └─ aws:s3:Bucket     my-bucket          create     
 
Diagnostics:
  pulumi:pulumi:Stack (aws-go-printf-dev):
    1 - with newline
    2 - no newline3 - with newline
 

Permalink: https://app.pulumi.com/clstokes/aws-go-printf/dev/previews/84e69cb9-ebe4-4ec0-becf-0ac1b91004f3
```

joeduffy:
If the last line printed to stdout or stderr was missing a
terminating newline, it would go entirely missing (in all languages).
The reason for this is a bug in the engine's handling of plugin
outputs: Go's Reader.ReadString('\n') returns a string containing what
was read and/or an error; if the string terminated in a '\n', the
error is nil, and the entire line is returned; if the stream ends,
however, a non-nil error is returned *and* what was read is returned,
even though it wasn't terminated in a newline. The fix is simple:
instead of ignoring that text, we use it, and *then* exit the read-loop.

Also added some test cases since this is subtle and easy to regress.


lukehoban (approved):
LGTM. 

Note that this fixes #3982 (not #3892).

joeduffy:
Hmm, @lukehoban, the codecov report diffs don't make sense to me. Do they to you? The lines it says I am removing seem entirely unrelated to these changes 🤔  https://app.codecov.io/gh/pulumi/pulumi/compare/8671/changes

lukehoban:
> Hmm, @lukehoban, the codecov report diffs don't make sense to me. Do they to you? 

No - they don't make sense to me either - none of the results from PRs the last month have been "correct" - we need to take deeper look at this.  I think this is in part due to the issue I opened in https://github.com/pulumi/pulumi/issues/8629, but it also seems that there are just deeper holes in the coverage reporting, which I've amended onto that issue.  For now, unfortunately, we largely have to just ignore this reporting.

PR #8672, opened 2021-12-31
Initialize stack in mocks mode

joeduffy:
This change initializes a root stack resource when run in
mocks mode, fixing #4669. This lets you use, for instance,
stack transformations during unit testing. This is effectively the
equivalent to what we did for Python in
https://github.com/pulumi/pulumi/pull/4670/files, just for Node.js.

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,3 +1,6 @@
 ### Improvements
 
 ### Bug Fixes
+
+- [sdk/nodejs] - Fix [#4669](https://github.com/pulumi/pulumi/issues/4669), enabling
+  stack transformations in unit test mocking mode [#8672](https://github.com/pulumi/pulumi/pull/8672).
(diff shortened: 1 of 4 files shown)


lukehoban (approved):
This makes sense for consistency with Python.  

One question/concern though - does this break any code which was previously using `runInPulumiStack` from within mock tests (in part to work around this)?  Or is the statefullness there skipped since there is no engine tracking the active root stack?

joeduffy:
> Does this break any code which was previously using runInPulumiStack from within mock tests (in part to work around this)? Or is the statefullness there skipped since there is no engine tracking the active root stack?

I just tried it out, and it still works. Not sure exactly how runInPulumiStack works, but it does appear you're right that it doesn't set the global state. I also couldn't find any tests for it, so not sure what the intended semantics are here, given that the callback within runInPulumiStack presumably won't have the ambient stack available to it while it runs?

gideon-maina:
@joeduffy Hello, was there a fix for this in another merged change or could you share any info on the linked issue https://github.com/pulumi/pulumi/issues/4669, as I got this issue today? Thanks.

PR #8676, opened 2022-01-04
Using a decryptAll functionality when deserializing a deployment

stack72:
Fixes: #6445

Previously, when a deserialisation operation, e.g, `pulumi stack` command
was carried out, each service based secret in the stack was decrypted one at a time
causing n times the number of API requests to the Pulumi service

This meant the following to the user:

```
Calling [pulumi stack] with [1000] secrets took [210.04] seconds.
```

With this new code, we now scan all the inputs and outputs for secrets,
and make a bulk secret decryption dequest to the service and then use
the secrets as a lookup for the deserialization operation. This means
we have a single API request to the service and now means the following
to the user:

```
Calling [pulumi stack] with [1000] secrets took [16.30] seconds.
```

# Description

Fixes # (issue)

## Checklist

### DIFF
--- a/pkg/resource/stack/deployment.go
+++ b/pkg/resource/stack/deployment.go
@@ -232,7 +232,23 @@ func DeserializeDeploymentV3(deployment apitype.DeploymentV3, secretsProv Secret
 		if err != nil {
 			return nil, err
 		}
-		dec = d
+		dec = config.NewCachedDecrypter(d)
+
+		// Do a first pass through state and collect all of the secrets that need decrypting.
+		// We will collect all secrets and decrypt them all at once, rather than just-in-time.
+		// We do this to avoid serial calls to the decryption endpoint which can result in long
+		// wait times in stacks with a large number of secrets.
+		var ciphertexts []string
+		for _, res := range deployment.Resources {
+			ciphertexts = append(ciphertexts, scanResource(res)...)
+		}
+
+		// we call BulkDecrypt to Prime the cache with the decrypted cipertexts
+		// we don't actually use the response here at this point
+		_, err = dec.BulkDecrypt(ciphertexts)
+		if err != nil {
+			return nil, err
+		}
 
 		e, err := secretsManager.Encrypter()
 		if err != nil {
@@ -397,7 +413,7 @@ func SerializePropertyValue(prop resource.PropertyValue, enc config.Encrypter,
 
 	if prop.IsSecret() {
 		// Since we are going to encrypt property value, we can elide encrypting sub-elements. We'll mark them as
-		// "secret" so we retain that information when deserializaing the overall structure, but there is no
+		// "secret" so we retain that information when deserializing the overall structure, but there is no
 		// need to double encrypt everything.
 		value, err := SerializePropertyValue(prop.SecretValue().Element, config.NopEncrypter, showSecrets)
 		if err != nil {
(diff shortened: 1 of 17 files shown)


pkg/resource/stack/deployment.go
```diff
@@ -276,10 +276,24 @@ func DeserializeDeploymentV3(deployment apitype.DeploymentV3, secretsProv Secret
 		enc = e
 	}
 
+	sc := SecretsCache{}
```
pgavlin, line 279: Do we need a new type here? Feels like we could do this without such a large refactoring by moving the caching into the Decrypter. We do something similar with the [`cachingCrypter`](https://github.com/pulumi/pulumi/blob/master/pkg/resource/stack/secrets.go#L68-L152) to avoid re-encrypting secrets during serialization.

pkg/resource/stack/deployment_test.go
```diff
@@ -404,6 +405,24 @@ func TestCustomSerialization(t *testing.T) {
...
+		Resources: []apitype.ResourceV3{
+			{
+				URN:    resource.URN(urn),
+				Type:   "aws:ebs/volume:Volume",
+				Custom: true,
+				ID:     "vol-044ba5ad2bd959bc1",
+			},
+		},
+	}, DefaultSecretsProvider)
+	//assert.Nil(t, deployment)
```
Frassle, line 421: dead comments?

sdk/go/common/resource/config/crypt.go
```diff
@@ -100,26 +125,100 @@ func NewBlindingDecrypter() Decrypter {
...
 	return "[secret]", nil
 }
 
+type CachedCrypter interface {
+	Decrypter
+}
+
+// cachedCrypter is a Decrypter that keeps track if decrypted values, which
+// can be retrieved via SecureValues().
+type cachedCrypter struct {
```
Frassle, line 153: So there's another type called `cachingCrypter` in pkg/resource/stack/secrets.go so I wonder if this should be called `cachedDecrypter` to make it clear its just one side of the crypt interface.

Frassle (commented):
Why do we have so many crypters 😭

Frassle (approved):

sdk/go/common/resource/config/crypt.go
```diff
@@ -36,6 +36,7 @@ type Encrypter interface {
 // Decrypter decrypts encrypted ciphertext to its plaintext representation.
 type Decrypter interface {
 	DecryptValue(ciphertext string) (string, error)
+	BulkDecrypt(ciphertexts []string) (map[string]string, error)
```
pgavlin, line 39: Nit: I don't think that we need this new method on the `Decrypter` interface. Instead, we could let this live on a separate interface and polyfill it for `Decrypter`s that don't implement it. Something like:

```go
type BulkDecrypter interface {
	BulkDecrypt(ciphertexts []string) (map[string]string, error)
}

func BulkDecrypt(dec Decrypter, ciphertexts []string) (map[string]string, error) {
	if bulk, ok := dec.(BulkDecrypter); ok {
		return bulk.BulkDecrypt(ciphertexts)
	}

	result := make(map[string]string, len(ciphertexts))
 	for _, c := range ciphertexts {
		plaintext, err := dec.DecryptValue(c)
		if err != nil {
			return nil, err
		}
		result[c] = plaintext
 	}
 	return result, nil
}
```

That way, not every `Decrypter` has to implement `BulkDecrypt`--only those for which it provides meaningful additional value.

pgavlin (approved):
LGTM--one nit about moving `BulkDecrypt` to its own interface.

sdk/go/common/resource/config/crypt.go
stack72: @pgavlin I know that @Frassle is going to be in this part of the code soon - he alluded to the fact about so many decrypters :) So I'd love to be able to unlock this value now and defer this to him when he's there

Frassle: I'm fine with it as is. Pauls already done the work of writing BulkDecrypt for everything and as he alluded to I want to do a cleanup pass over the whole lot anyway so if it makes sense I'll do this change then.

pgavlin: Sounds good.

PR #8682, opened 2022-01-04
Unify Map and Value number exporting

### LINKED ISSUE #8575: Inconsistent type handling in stack config
## Hello!

## Issue details

Numeric values are stored differently in the stack config file depending on whether they are defined for the top-level or nested key.

### Steps to reproduce

1. Create a new `test` project.
2. Set config values:

```
pulumi config set --path 'foo["bar"]' 100
pulumi config set 'baz' 100
```

Expected: Both values have the same type, either numeric or string in the stack config.

Actual: The top-level key has a string value while the nested key has a numeric value:

```
test:baz: "100"
test:foo:
  bar: 100
```

iwahbe:
# Description

Fixes #8575

## Checklist

### DIFF
--- a/sdk/go/common/resource/config/map.go
+++ b/sdk/go/common/resource/config/map.go
@@ -126,8 +126,6 @@ func (m Map) Get(k Key, path bool) (Value, bool, error) {
 		return NewValue(t), true, nil
 	case bool, int, uint, int32, uint32, int64, uint64, float32, float64:
 		return NewValue(fmt.Sprintf("%v", v)), true, nil
-	default:
-		fmt.Printf("Found other value %#v of type %T\n", v, v)
 	}
 
 	// Otherwise, return it as an object value.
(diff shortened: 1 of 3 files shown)


justinvp:
I'd like to take these changes for a spin, but I won't be able to get to it until tomorrow. Does this change the behavior when reading config from a Pulumi program (i.e. via `Config.get/getNumber/getBoolean/etc.`)? The language SDKs read the config values from the `PULUMI_CONFIG` envvar which contains a JSON serialized object of the config values, and expects the values to be strings. Will that still be the case after these changes?

iwahbe:
This should not change behavior. If we serialize and then deserialize the whole structure then this strategy should work. I'm not changing the behavior of nested values at all, so we must be able to deserialize and safely serialize plain numbers. I don't think anything should change here.

justinvp (approved):
LGTM

PR #8685, opened 2022-01-05
Split coverage results into projects

pgavlin:
Take advantage of [Codecov's support for subprojects](https://github.com/pulumi/pulumi/pull/8685) to split the results for this repo into results for the CLI/core packages and each SDK.

If this works, I'm hoping that it will help us better track down sources of non-determinism and confusing coverage results.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,8 +1,28 @@
 coverage:
   status:
     project:
-      default:
+      default: false
+      core:
         informational: true
+        paths:
+          - "pkg/*"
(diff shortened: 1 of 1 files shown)


pgavlin:
I'm not sure why the split isn't taking within the context of this PR, but I'd still like to merge it. Worst case scenario we break coverage reporting temporarily and back these changes out.

t0yv0 (approved):
Let's try it.

Issue #8521, opened 2021-12-01
[Codegen] generated packages do not respect the specified version


iwahbe:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

When we run through codegen, we do not generate `version.txt` for .NET. We generate `${VERSION}` for python (setup.py) and nodejs (package.json). We should stamp in the correct version at generate time.

PR #8690, opened 2022-01-05
[sdk/go] Embed PluginDownloadURL during codegen

iwahbe:
# Description

Fixes # (issue)

Embed `PluginDownloadURL`s in resource creation requests from codegen.

## Checklist


pkg/codegen/go/gen.go
```diff
@@ -1638,6 +1638,13 @@ func (pkg *pkgContext) genResource(w io.Writer, r *schema.Resource, generateReso
 		fmt.Fprint(w, "\topts = append(opts, replaceOnChanges)\n")
 	}
 
+	// Setup PluginDownloadURL. We pre-append our call to PluginDownloadURL to
+	// that if a user adds their own separate PluginDownloadURL, we won't
+	// override it.
+	if url := pkg.pkg.PluginDownloadURL; url != "" {
+		fmt.Fprintf(w, "\topts = append([]pulumi.ResourceOption{pulumi.PluginDownloadURL(%q)},  opts...)\n", url)
```
justinvp, line 1645: I wonder if we should emit a utility function in the generated `pulumiUtilities.go` for this (only if `PluginDownloadURL` is set), to reduce the amount of duplicated code for each resource.

justinvp, line 1645: @pgavlin, quick question that's orthogonal to this PR, but kind of related:

Shouldn't we be setting the default version in here like we do in the other languages, or am I missing something?

https://github.com/pulumi/pulumi/blob/684d7aad3920042ce1982b5fedddea62a6e5bb61/pkg/codegen/dotnet/gen.go#L1037

https://github.com/pulumi/pulumi/blob/684d7aad3920042ce1982b5fedddea62a6e5bb61/pkg/codegen/nodejs/gen.go#L868-L870

https://github.com/pulumi/pulumi/blob/684d7aad3920042ce1982b5fedddea62a6e5bb61/pkg/codegen/python/gen.go#L1219-L1220

Maybe we didn't do this for Go because the version info for Go isn't readily available? (Although, we did later add [`PkgVersion`](https://github.com/pulumi/pulumi/blob/684d7aad3920042ce1982b5fedddea62a6e5bb61/pkg/codegen/go/gen.go#L3610-L3623) used in resource module registrations; so perhaps we could consider using that, if it was desirable).

justinvp: We'll also need to do something similar for invokes.

iwahbe: I'll add the function to `pulumiUtilities`. It will make the code marginally smaller. It will also provide a convent place to put the default version if we want that.

pgavlin: Honestly, I'm not sure why we didn't do this in Go. It does seem like we should. I'm also not sure how we arrived at `PkgVersion`--any idea if there's a reason we didn't just embed the version statically? The reflection-based code used by `PkgVersion` is essentially static, I think.

iwahbe: We can't embed the version statically because we can't trust that the version supplied in the schema is correct. This is the same reason why https://github.com/pulumi/pulumi/issues/8521 exists, and is blocked. Injecting the version at codegen time would require changing how our provider makefile work.

pkg/codegen/go/gen.go
```diff
@@ -3630,3 +3636,40 @@ func isZero(v interface{}) bool {
 	return reflect.ValueOf(v).IsZero()
 }
 `
+	fmt.Printf("Body:\n%s\n", pkg.GenPkgDefaultOpts())
```
justinvp, line 3639: Was this line just for debugging while working on the implementation?

pkg/codegen/go/gen.go
```diff
@@ -3560,7 +3565,8 @@ func goPackage(name string) string {
 	return strings.ReplaceAll(name, "-", "")
 }
 
-const utilitiesFile = `
+func UtilitiesFile(pkg *pkgContext) string {
```
justinvp, line 3568: Nit: Any reason not to make this a method on pkgContext?

```suggestion
func (pkg *pkgContext) GenUtilitiesFile() string {
```

pkg/codegen/internal/test/testdata/azure-native-nested-types/go/azure/documentdb/pulumiUtilities.go
```diff
@@ -0,0 +1,107 @@
...
+// isZero is a null safe check for if a value is it's types zero value.
+func isZero(v interface{}) bool {
+	if v == nil {
+		return true
+	}
+	return reflect.ValueOf(v).IsZero()
+}
+
+// PkgResourceDefaultOpts provides package level defaults to pulumi.OptionResource.
+func PkgResourceDefaultOpts(opts []pulumi.ResourceOption) []pulumi.ResourceOption {
```
justinvp, line 88: I realize the existing `PkgVersion` is exported (capitalized), but I doubt that was intentional. I am wondering if these new functions should be unexported (lowercase)... I don't think anyone will need to be calling these as part of an SDK's public API directly.

(Separately, it might be worth revisiting `PkgVersion`, making it lowercase).

pkg/codegen/internal/test/testdata/azure-native-nested-types/go/azure/documentdb/pulumiUtilities.go
```diff
@@ -0,0 +1,107 @@
...
+	defaults := []pulumi.ResourceOption{}
+	// We don't add the version if we can't discover it. We don't error here
+	// because we want to allow non-traditional go project configurations.
+	if v, err := PkgVersion(); err == nil {
```
justinvp, line 92: It looks like `PkgVersion` is only returning the major version, instead of the full version. Probably best to hold off on setting the version resource option for now until we've addressed that, doing it as a separate change.

pkg/codegen/go/gen.go
```diff
@@ -3526,7 +3531,7 @@ func GeneratePackage(tool string, pkg *schema.Package) (map[string][]byte, error
 				packageRegex = fmt.Sprintf("^%s(/v\\d+)?", pkg.importBasePath)
 			}
 
-			_, err := fmt.Fprintf(buffer, utilitiesFile, packageRegex)
+			_, err := fmt.Fprintf(buffer, UtilitiesFile(pkg), packageRegex)
```
justinvp, line 3534: Do we need to make sure `pkg.needsUtils` is set to true if there are any resources or invokes? Otherwise, is it possible the utilities file won't be emitted?

iwahbe: `insertPkgDefaultsOptsCall` sets `pkg.needsUtils = true`.

pkg/codegen/internal/test/testdata/azure-native-nested-types/go/azure/documentdb/pulumiUtilities.go
iwahbe: I capitalized `Pkg%sDefaultOpts` because I assumed there was a good reason why `PkgVersion` was capitalized. I can change that.

pkg/codegen/go/gen.go
pgavlin: Ah yep, thank you for the reminder. I wonder if we should switch this to using a `//go:embed` and a parse in `init` so that we can get a real version. This would be akin to the solutions we use for the other language. More concretely, we could e.g. have the Go code generator emit a dummy `version.txt` file that is embedded into the package, then have the Makefile update the contents of that file with the actual server. The contents of the file would be parsed into a real version during the package's `init` function.

iwahbe: If we are willing to touch every Makefile to actualize that change, I think we should instead attempt to find a safe way to embed version into the binary (if at all possible).

pgavlin: > I think we should instead attempt to find a safe way to embed version into the binary (if at all possible).

I would love it if we could find a way to do this, but I'll admit that I'm pessimistic about making that happen any time soon. I think that taking a tactical fix--even though it requires updating Makefiles--is a chance for us to not let perfect be the enemy of better.

justinvp: > we could e.g. have the Go code generator emit a dummy `version.txt` file that is embedded into the package, then have the Makefile update the contents of that file with the actual server

I still don't quite follow how this would work for go modules. Say we generate an SDK for a `foo` provider, emitted at `github.com/pulumi/pulumi-foo/sdk/go/foo`. Versions are tagged with git tags (e.g. `v1.2.3` and `sdk/v1.2.3`). Wouldn't we have to commit the `version.txt` file before tagging it? I don't see how the `Makefile` helps. When I use the `foo` module from my program, our `Makefile` is not involved.

pgavlin: That's a good point. For Go, it's the push to the git remote that's the publishing step, so even if we took my suggestion we'd still be emitting the wrong version.

PR #8693, opened 2022-01-06
Redo 8633

### LINKED ISSUE #8633: Automation API: Python in-line throws `Resource Monitor has terminated, shutting down` error for all stacks after a stack error.

t0yv0:
# Description

Fixes exception isolation of Automation API inline programs. Before the fix an exception in an operation would poison subsequent operations even if caught by the user.

Unfortunately the included tests demonstrates that the issue is not yet fixed when running on Windows, where the test hangs. Filed https://github.com/pulumi/pulumi/issues/8716 to follow up, there does not appear to be an easy fix. To clarify, the Windows hang exists before and after this fix, but is likely related.

Fixes #8633 

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,3 +1,8 @@
 ### Improvements
 
 ### Bug Fixes
+
+- [auto/python] - Fixes an issue with exception isolation in a
+  sequence of inline programs that caused all inline programs to fail
+  after the first one failed
+  [#8633](https://github.com/pulumi/pulumi/pull/8633)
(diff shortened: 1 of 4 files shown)


iwahbe (approved):
LGTM.

It looks like the test failure is unrelated.

PR #8695, opened 2022-01-06
[sdk/common] allow names with hyphens

svangordon-fruit:
# Description

Allow names with hyphens.

Fixes # (issue)

## Checklist

### DIFF
--- a/sdk/go/common/tokens/names.go
+++ b/sdk/go/common/tokens/names.go
@@ -36,7 +36,7 @@ var nameRestCharRegexp = regexp.MustCompile("^" + nameRestCharRegexpPattern + "$
 var NameRegexpPattern = nameFirstCharRegexpPattern + nameRestCharRegexpPattern
 
 const nameFirstCharRegexpPattern = "[A-Za-z_.]"
-const nameRestCharRegexpPattern = `[A-Za-z0-9_.]*`
+const nameRestCharRegexpPattern = `[A-Za-z0-9_.-]*`
 
 // IsName checks whether a string is a legal Name.
 func IsName(s string) bool {
--- a/sdk/go/common/tokens/names.go
+++ b/sdk/go/common/tokens/names.go
@@ -21,7 +21,7 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 )
 
-// Name is an identifier.  It conforms to the regex [A-Za-z_.][A-Za-z0-9_]*.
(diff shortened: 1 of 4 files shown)


sdk/go/common/tokens/names.go
```diff
@@ -36,7 +36,7 @@ var nameRestCharRegexp = regexp.MustCompile("^" + nameRestCharRegexpPattern + "$
 var NameRegexpPattern = nameFirstCharRegexpPattern + nameRestCharRegexpPattern
 
 const nameFirstCharRegexpPattern = "[A-Za-z_.]"
-const nameRestCharRegexpPattern = `[A-Za-z0-9_.]*`
+const nameRestCharRegexpPattern = `[A-Za-z0-9_.-]*`
```
Frassle, line 39: Comment on line 24 and 53 needs updating to match this

svangordon-fruit: Comments updated to reference `NameRegexpPattern`

svangordon-fruit:
I notice that in `ProviderRequest.Name`, there's some string manipulation to get around `QName`s not accepting hyphens. Now that they do, that could be simplified, though that's a bit beyond the scope of this PR. (cc @iwahbe, who will have more context that I do)
https://github.com/pulumi/pulumi/blob/8e5e4caf6e405a380a64911f8a7da4ff1de80c24/pkg/resource/deploy/providers/provider.go#L79-L81

iwahbe:
I am hesitant to change how the `Name` function works. My understanding is that they are baked into state files as part of providers among other things. If we change how they are constructed I'm concerned we won't be able to update resources whose providers have changed names. 

This is from the state of one of my stacks:
```
...
"provider": "urn:pulumi:dev::website-and-lambda::pulumi:providers:aws::default_4_17_0::f4e94675-9f92-4f63-be9e-f83df11acf3c",
...
```
I'm pretty sure that `default_4_17_0` is from the `Name` function @svangordon-fruit mentioned. 

CC @pgavlin

svangordon-fruit:
Oh my bad, looking at the git blame I missed that https://github.com/pulumi/pulumi/pull/8544 was just a formatting change to those lines, and not any kind of new logic.

I agree: if it's three years old and works fine, I think it makes sense to leave it as it is.

mikhailshilkov:
@Frassle @iwahbe Could you please take another look?

Frassle (approved):
LGTM

PR #8698, opened 2022-01-06
[sdk/nodejs] Add pluginServerURL

### LINKED ISSUE #8687: [sdk/nodejs] Support for custom server urls
## Hello!

## Issue details
Provide the nodejs SDK hookup for #8544.

iwahbe:
Tests that this work are included in the PR.

# Description

Fixes #8687 

## Checklist


sdk/nodejs/runtime/resource.ts
```diff
@@ -317,6 +318,7 @@ export function registerResource(res: Resource, t: string, name: string, custom:
         req.setSupportspartialvalues(true);
         req.setRemote(remote);
         req.setReplaceonchangesList(opts.replaceOnChanges || []);
+        req.setPlugindownloadurl(opts.pluginDownloadURL || "");
```
justinvp, line 321: Another thing that needs to be done is to save the URL on the resource instance to support `call`.

Here's where the `version` is saved:

https://github.com/pulumi/pulumi/blob/6a0cbe2c0da947150b4ff195d71d4deee16f0947/sdk/nodejs/resource.ts#L361

We'll need to do the same with the URL.

And then we'll need to pass the URL in the `CallRequest`:

https://github.com/pulumi/pulumi/blob/6a0cbe2c0da947150b4ff195d71d4deee16f0947/sdk/nodejs/runtime/invoke.ts#L248-L264

justinvp (approved):
LGTM

PR #8699, opened 2022-01-06
[nodejs/sdk] GetRequiredPlugins: Return plugins even when there're errors

### LINKED ISSUE #8697: [sdk/nodejs] GetRequiredPlugins returns an empty set of plugins when there is an error
This came up while investigating https://github.com/pulumi/examples/issues/1146

`@pulumi/pulumi` and `@pulumi/aws` have a dependency on the npm `resolve` package. The most recent release of `resolve` now includes a malformed `package.json` inside a `tests` subdirectory within the package. The malformed file is causing our implementation of `GetRequiredPlugins` in the Node.js language host to return an empty set of plugins. The intended behavior is to return any plugins it was able to find and log any errors encountered along the way:

https://github.com/pulumi/pulumi/blob/684d7aad3920042ce1982b5fedddea62a6e5bb61/sdk/nodejs/cmd/pulumi-language-nodejs/main.go#L244-L249

But that's not what's happening in this case because when an error is returned from the recursive call for the `node_modules` directory, it skips appending the resulting plugins that were found:

https://github.com/pulumi/pulumi/blob/684d7aad3920042ce1982b5fedddea62a6e5bb61/sdk/nodejs/cmd/pulumi-language-nodejs/main.go#L274-L282

The fix is to remove the `continue` on line 280 inside the `if err != nil {` check.

justinvp:
The implementation of `GetRequiredPlugins` in the Node.js language host will return an empty set of plugins if it encounters errors reading `package.json` inside a subdirectory. The [intended behavior](https://github.com/pulumi/pulumi/blob/684d7aad3920042ce1982b5fedddea62a6e5bb61/sdk/nodejs/cmd/pulumi-language-nodejs/main.go#L244-L249) is to return any plugins that were found (even if there were some errors in the process) and to log any errors that occurred along the way. This change fixes the implementation to behave as intended.

Fixes #8697

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -18,3 +18,6 @@
   where the engine ignored the final line of stdout/stderr if it didn't terminate
   with a newline. 
   [#8671](https://github.com/pulumi/pulumi/pull/8671)
+
+- [nodejs/sdk] - GetRequiredPlugins: Return plugins even when there're errors.
+  [#8699](https://github.com/pulumi/pulumi/pull/8699)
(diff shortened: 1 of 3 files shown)


iwahbe (approved):
LGTM

PR #8704, opened 2022-01-07
Clean up orphan processes in integration tests

t0yv0:
# Description

Due to #8696 we may be leaking processes that never terminate and take up memory. While I do not have hard evidence, this may have something to do with OOM issues in CI especially on Windows. While fully fixing #8969 in product requires some more careful experimentation, this little patch at least makes the integration tests clean up after themselves with minimal risk. I would like to try it to see if our stability improves any.

Fixes # (issue)

## Checklist


pkg/testing/integration/command.go
```diff
@@ -56,10 +58,13 @@ func RunCommand(t *testing.T, name string, args []string, wd string, opts *Progr
 		cmd.Stdout = opts.Stdout
 		cmd.Stderr = opts.Stderr
 		runerr = cmd.Run()
+
 	} else {
 		runout, runerr = cmd.CombinedOutput()
 	}
 
+	defer orphanCleaner.killOrphanProcesses(t)
```
pgavlin, line 66: We might already have code that does what you need: https://github.com/pulumi/pulumi/blob/master/sdk/go/common/util/cmdutil/child_windows.go#L26-L73

t0yv0: Nice! Yes this is pretty much the same idea modulo logging, let me reuse.

t0yv0:
More evidence on leaked procs. GHA detects those at the end of the build. Here is a "normal" run. We have a lot of these leaked procs.

```
  https://github.com/pulumi/pulumi/runs/4767897250?check_suite_focus=true

  Cleaning up orphan processes
  Terminate orphan process: pid (5848) (pulumi-language-dotnet)
  Terminate orphan process: pid (5780) (pulumi-language-go)
  Terminate orphan process: pid (5356) (node)
  Terminate orphan process: pid (6504) (python)
  Terminate orphan process: pid (7140) (python)
  Terminate orphan process: pid (5872) (python)
  Terminate orphan process: pid (3212) (python)
  Terminate orphan process: pid (960) (python)
  Terminate orphan process: pid (6884) (python)
  Terminate orphan process: pid (1484) (pulumi-language-python)
  Terminate orphan process: pid (7128) (dotnet)
```

Really interested in trying the KillChildren workaround.

t0yv0:
It's fascinating that it didn't work, especially on Linux. Need to dig into this.

```

*** ubuntu-latest integreation
Terminate orphan process: pid (18046) (pulumi-language-dotnet)
Terminate orphan process: pid (22971) (pulumi-language-go)
Terminate orphan process: pid (27322) (node)
Terminate orphan process: pid (27332) (node)
Terminate orphan process: pid (27342) (node)
Terminate orphan process: pid (27536) (pulumi-language-nodejs)
Terminate orphan process: pid (32082) (pulumi-language-python)

*** ubuntu-latest auto
leaning up orphan processes
Terminate orphan process: pid (8044) (dotnet)

*** windows-latest integration
Terminate orphan process: pid (3756) (pulumi-language-dotnet)
Terminate orphan process: pid (1268) (pulumi-language-go)
Terminate orphan process: pid (4768) (pulumi-language-python)
Terminate orphan process: pid (6292) (dotnet)

*** windows-latest integration-and-codegen
Terminate orphan process: pid (4272) (dotnet)

*** windows-latest auto
Terminate orphan process: pid (1140) (dotnet)
Terminate orphan process: pid (2660) (conhost)
```

t0yv0:
Sounds like this is not going to work. I've done some more experimentation. On Mac and Linux the setpgid trick does not work because in the ProgramTest -> pulumi -> plugin tree,  ProgramTest does setpgid, but so does pulumi, so plugin tree ends up having a different PGID than pulumi and killing it does not work.

I'm less sure why it does not work on Windows.

I'm leaning to finding the root cause of process leaks in the product and fixing it. It's not trivial but there's some clues like TODOs around using graceful termination instead of sending SIGKILL to the plugin process.

PR #8709, opened 2022-01-08
Fixes to KillChildren function

t0yv0:
# Description

The Windows platform-specific implementation of KillChildren as well as the comment on KillChildren were a bit misleading. What happens on Mac and Linux is that typically KillChildren(pid) would kill pid as well as any direct or indirect child. This happens because PGID gets assigned to pid as well as to recursive descendants of pid (it inherits) by default, unless those sub-processes explicitly change PGID. I've added a test to demonstrate this.

With the change - to the Windows implementation - Windows version will try to kill self and descendants now also. 

Fixes # (issue)

## Checklist

### DIFF
--- a/sdk/go/common/util/cmdutil/child_windows.go
+++ b/sdk/go/common/util/cmdutil/child_windows.go
@@ -37,7 +37,7 @@ import (
 // This is the Windows-specific implementation that sends scans all
 // system processes using native syscalls (via go-ps) and attempts to
 // Kill them, aggregating any errors it encounters.
-func KillChildrenW(pid int) error {
+func KillChildren(pid int) error {
 	procs, err := ps.Processes()
 	if err != nil {
 		return err
@@ -126,6 +126,6 @@ func isDescendant(p int, parents map[int]int) bool {
 }
 
 // RegisterProcessGroup does nothing on Windows.
-func RegisterProcessGroupW(cmd *exec.Cmd) {
+func RegisterProcessGroup(cmd *exec.Cmd) {
 	// nothing to do on Windows.
 }
--- a/sdk/go/common/util/cmdutil/child_windows.go
+++ b/sdk/go/common/util/cmdutil/child_windows.go
@@ -1,4 +1,4 @@
-// Copyright 2016-2018, Pulumi Corporation.
+// Copyright 2016-2022, Pulumi Corporation.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -11,11 +11,11 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-// +build windows
 
 package cmdutil
 
 import (
+	"fmt"
 	"os"
 	"os/exec"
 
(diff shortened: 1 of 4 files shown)


sdk/go/common/util/cmdutil/testdata/processtree.go
```diff
@@ -0,0 +1,45 @@
+// Copyright 2016-2022, Pulumi Corporation.
```
t0yv0, line 1: I might need to rename it something other than .go so it does not compile with the main go tree.

sdk/go/common/util/cmdutil/child_windows.go
```diff
@@ -11,65 +11,122 @@
...
-func processExistsWithParent(pid int, ppid int) (bool, error) {
-	procs, err := ps.Processes()
-	if err != nil {
-		return false, err
+func filterDescendants(rootPid int, processes []ps.Process) []ps.Process {
+	parents := map[int]int{}
+	for _, p := range processes {
+		pid := p.Pid()
+		ppid := p.PPid()
+		// Can have PID=0 with PPID=0, ignore it.
```
Frassle, line 104: Is 0 special here or do you just mean that when PID and PPID are equal to ignore them?

sdk/go/common/util/cmdutil/testdata/processtree.go
Frassle: Do we not have other .go files under testdata folders?

sdk/go/common/util/cmdutil/child_windows.go
t0yv0: Experimentally it seems that the 0 is the special PID but I'm not sure. What I care about is there's no cycles in this graph. x.PPID=x.PID creates a cycle that can make the program below loop.

sdk/go/common/util/cmdutil/testdata/processtree.go
pgavlin: `testdata` should be ignored by the `go` tool, so this shouldn't build by default. Are you running into issues here?

sdk/go/common/util/cmdutil/child_windows.go
```diff
@@ -11,65 +11,122 @@
...
+	// observing the process in the `survivingPids`. This method
+	// would rather return a `nil` error when unsure than error
+	// out incorrectly.
+	var result error
+
+	for _, proc := range descendants {
+		_, surviving := survivingPids[proc.Pid()]
+		if surviving {
+			err, gotErr := errors[proc.Pid()]
+			if gotErr {
```
pgavlin, line 91: Tiny nit: declare + test is a bit more idiomatic than separate statements if the lifetime of the declaration is scoped to the introduced block.

```suggestion
		if _, surviving := survivingPids[proc.Pid()]; surviving {
			if err, gotErr := errors[proc.Pid()]; gotErr {
```

sdk/go/common/util/cmdutil/child_windows.go
```diff
@@ -11,65 +11,122 @@
...
+		procPid := proc.Pid()
+		toKill, err := os.FindProcess(procPid)
+		if err != nil {
+			errors[procPid] = fmt.Errorf("FindProcess(%d) failed: %w", procPid, err)
+		}
+
+		err = toKill.Kill()
+		if err != nil {
+			errors[procPid] = fmt.Errorf("proc.Kill() failed for pid=%d: %w", procPid, err)
+		}
```
pgavlin, line 65:
```suggestion
		if err = toKill.Kill(); err != nil {
			errors[procPid] = fmt.Errorf("proc.Kill() failed for pid=%d: %w", procPid, err)
		}
```

sdk/go/common/util/cmdutil/testdata/processtree.go
t0yv0: Ah nice. Confirmed, it's good.

t0yv0:
It's been interesting that the test fails on Ubuntu. Rerunning a few more times, will confirm. Would be interesting if there's differences how Mac and Ubuntu handle setpgid; but more likely it's about a poorly setup test/poll loop.

t0yv0:
This is obsolete as we now fix the underlying problem by heartbeat checks in plugins.

PR #8723, opened 2022-01-11
Remove MSBuild file no longer in use

t0yv0:
# Description

We can wait until we go through another release before merging (though that should be fine).

Fixes # (issue)

## Checklist


t0yv0:
We released without this 👍  Rebasing, running tests.

Frassle (approved):
Getting to the point this can be killed is great work @t0yv0!

justinvp (approved):
🎉

PR #8726, opened 2022-01-11
[ci] Restore code coverage data for Node/.NET

pgavlin:
PR #8617 inadvertently broke coverage collection and reporting for
NodeJS and .NET.

### DIFF
--- a/.github/workflows/test-fast.yml
+++ b/.github/workflows/test-fast.yml
@@ -335,6 +335,11 @@ jobs:
           PULUMI_NODE_MODULES: ${{ runner.temp }}/opt/pulumi/node_modules
           PULUMI_LOCAL_NUGET: ${{ github.workspace }}/nuget
           PULUMI_ROOT: ${{ runner.temp }}/opt/pulumi
+      - name: Convert Node coverage data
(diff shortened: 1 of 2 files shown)


t0yv0:
Looks safe enough. 

I'm trying to see if `$(FOO)` and `${FOO}` are any different in Make.

t0yv0 (approved):

t0yv0:
> To substitute a variable's value, write a dollar sign followed by the name of the variable in parentheses or braces: either `$(foo)' or `${foo}' is a valid reference to the variable foo. 

No difference. Good.

t0yv0:
```
python D:/a/pulumi/pulumi/scripts/skipped.py auto-dotnet || (cd Pulumi.Automation.Tests && dotnet test /p:CollectCoverage=false /p:CoverletOutput=/dotnet-auto.xml)
MSBUILD : error MSB1008: Only one project can be specified.
Switch: p:CoverletOutput=/dotnet-auto.xml
```

Something is funky on Windows. I have Windows test infra coming up, I can take this over to tinker with it and restore a working version.

PR #8728, opened 2022-01-11
Restore coverage collection for dotnet

t0yv0:
# Description

@pgavlin I dug into this, looks like the issue was that the `/p:` syntax was not parsing on Windows under Bash. The `-p:` syntax is working better for me, so I can restore the original forms. Also restoring the cov parameters that I lost accidentally in the refactor. PTAL, also curious how the CI goes.

Fixes # (issue)

## Checklist

### DIFF
--- a/sdk/dotnet/Makefile
+++ b/sdk/dotnet/Makefile
@@ -51,11 +51,11 @@ install:: build install_plugin
 
 dotnet_test:: $(TEST_ALL_DEPS)
 	$(TESTSUITE_SKIPPED) dotnet-test || \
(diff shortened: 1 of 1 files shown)


pgavlin (approved):

pgavlin:
Oddly enough, it looks like the .NET tests aren't running. Prior to these changes, I see output from `dotnet test`; with these changes I do not:

Before: https://github.com/pulumi/pulumi/runs/4745485466?check_suite_focus=true#step:36:20
After: https://github.com/pulumi/pulumi/runs/4782217598?check_suite_focus=true#step:36:19

t0yv0:
Fixed that, it was the --no-build option.

PR #8730, opened 2022-01-12
Generate scripts/install-pulumi-plugin.json

### LINKED ISSUE #8701: [codegen/nodejs] Generate `install-pulumi-plugin.js` install script
## Hello!

## Issue details

Our provider packages include a script that runs `pulumi plugin install` for the associated plugin when the package is installed. The install script isn't currently generated as part of the Node.js SDK codegen. Instead, as part of publishing the package to NPM, we have a [script](https://github.com/pulumi/scripts/blob/master/ci/add-plugin-installer-script.js) that runs which modifies the `package.json` for the package and includes the [install script](https://github.com/pulumi/scripts/blob/master/ci/install-pulumi-plugin.js).

Rather than mangling the `package.json` and injecting the install script as part of the publish process, we should consider changing this to happen during SDK codegen. This aligns with [what we do for Python](https://github.com/pulumi/pulumi/blob/684d7aad3920042ce1982b5fedddea62a6e5bb61/pkg/codegen/python/gen.go#L1971-L1976), which emits a `setup.py` which also shells out to the `pulumi plugin install` command.

The code generator should emit the install script and appropriate entry in `package.json` to have the script run during installation of the package. The existing install script does not currently support specifying the `--server` flag to `pulumi plugin install`, so as part of this, the emitted install script should include the `--server` flag if the schema has `pluginDownloadURL` set.

iwahbe:
# Description

Fixes #8701 

## Checklist


pkg/codegen/nodejs/gen.go
```diff
@@ -2127,12 +2130,9 @@ func genNPMPackageMetadata(pkg *schema.Package, info NodePackageInfo) string {
 		Homepage:    pkg.Homepage,
 		Repository:  pkg.Repository,
 		License:     pkg.License,
-		// Ideally, this `scripts` section would include an install script that installs the provider, however, doing
-		// so causes problems when we try to restore package dependencies, since we must do an install for that. So
-		// we have another process that adds the install script when generating the package.json that we actually
-		// publish.
```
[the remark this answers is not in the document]
iwahbe: I have built a project linked against a code generated module (using `yarn link`). It works without running the `install` script. `npm link` does run the `install` script. We can mitigate this issue by having the install script exit cleanly if it is passed a version of `${VERSION}`, since we don't update that until the version is known (and the package is published).

pkg/codegen/nodejs/gen.go
```diff
@@ -2525,3 +2527,39 @@ export function resourceOptsDefaults(): any {
...
 }
+
+func genInstallScript(pluginDownloadURL string) string {
+	const installScript = `"use strict";
+var childProcess = require("child_process");
+
+var args = process.argv.slice(2);
+
+// NOTE: version split is intended to prevent sed style replacement
+if (args.indexOf("${VERS" + "ION}") != -1) {
```
justinvp, line 2538: Clever. Can we get away without it, though, and just look for the `"${VERSION}"` string as-is? I don't think we're doing sed style replacement in this file anywhere. And if there is a `pluginDownloadURL` set, the URL could contain `${VERSION}` with #8507, so it's possible this file may end up having `${VERSION}` in it anyway.

justinvp, line 2538: Nit:

```suggestion
// NOTE: version split is intended to prevent sed style replacement
if (args.indexOf("${VERS" + "ION}") !== -1) {
```

iwahbe: I think that https://github.com/pulumi/pulumi/pull/8507 ensures that you are right here. Good catch.

justinvp (approved):
LGTM

@stack72, I am pretty sure that there's no need to immediately update the `publish-tfgen-package` script as this codegen change is adopted in our provider repos.

From what I can tell and from trying out the script locally, the existing script will continue to run and overwrite what's been generated from this change (which is fine). https://github.com/pulumi/scripts/blob/b6ee1692954d2b39e7a22b0db80b8cc4b735c101/ci/publish-tfgen-package#L33-L42

Eventually, when all the provider repos have been updated, we can remove those lines from `publish-tfgen-package` so that it'll just use what is generated.

t0yv0 (approved):

PR #8735, opened 2022-01-12
[codegen/dotnet] Allow generated SDKs to have a global namespace other then `Pulumi`

### LINKED ISSUE #8573: Allow NuGet packages without `Pulumi.` prefix
## Hello!

## Issue details

In support of adding 3rd-party Pulumi Packages to Pulumi Registry, we want to ensure that folks can name their NuGet packages according to NuGet best practices, where the first prefix of the package name (e.g. `Pulumi.` in the `Pulumi.AwsNative` package name) represents the owner of the package. Currently, we assume that the package prefix will always be `Pulumi.`, which is incompatible with this best practice. We should make the schema and package generation flexible enough to accept other first prefixes (e.g. `Infin8x.` in  `Infin8x.Pulumi.CoolestPackageEver`).

iwahbe:
# Description

Fixes #8573 

If a schema author wants to replace the `Pulumi` global namespace, they can add `"Pulumi": "NewNamespace"` to the `.language.csharp.namespaces` field. 

## Checklist


Frassle (commented):
I think using the name "Pulumi" is a bit confusing now, given that it doesn't really refer to the Pulumi package but instead to "whatever is the root namespace of this current package". Would it make sense to instead use something more specific to indicate that.

Given the required nature of this property (if the user doesn't give it we're going to default to "Pulumi") maybe it makes sense to just have a new field in the .json file for it "RootNamespace" would match the nomenclature used by msbuild.

mikhailshilkov:
Coming from providers where there could be 100.000+ generated lines in a PR, we had to take a practice where we would create separate commits for hand-authored vs. generated code. Would a similar practice make sense in the context of platform codegen and test cases? I _think_ it would simplify reviewing a 7k+ PR, and should be fairly scriptable for the committer.

pkg/codegen/dotnet/gen.go
```diff
@@ -1299,9 +1308,6 @@ func (mod *modContext) genFunctionFileCode(f *schema.Function) (string, error) {
 	for _, i := range imports {
 		importStrings = append(importStrings, i.SortedValues()...)
 	}
-	if f.NeedsOutputVersion() {
-		importStrings = append(importStrings, "Pulumi.Utilities")
```
mikhailshilkov, line 1303: Is this not needed anymore? Is that related to this change?

pkg/codegen/internal/test/testdata/other-owned/dotnet/Other.Example.csproj
```diff
@@ -0,0 +1,62 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
+    <Authors>Pulumi Corp.</Authors>
```
mikhailshilkov, line 5: Maybe unrelated - but should this be configurable too?

pkg/codegen/internal/test/testdata/other-owned/schema.json
```diff
@@ -0,0 +1,275 @@
...
+      },
+      "outputs": {
+        "properties": {
+          "result": {
+            "$ref": "#/resources/example::Resource"
+          }
+        }
+      }
+    },
+    "example::overlayFunction": {
```
mikhailshilkov, line 231: How do we decide what to include into this test case's schema beyond a basic case? I guess you copied it from some other test case. How did you pick one? Does every object in this schema test something meaningful?

pkg/codegen/internal/test/testdata/other-owned/schema.json
```diff
@@ -0,0 +1,275 @@
...
+      },
+      "namespaces": {
+        "Pulumi": "Other"
+      }
+    },
+    "go": {
+      "importBasePath": "github.com/pulumi/pulumi/pkg/v3/codegen/internal/test/testdata/other-owned/go/example",
+      "generateExtraInputTypes": true
+    },
+    "nodejs": {
```
mikhailshilkov, line 265: Would it be hard to skip irrelevant languages for test cases that add no value for a given test case (all languages except C# in this case)?

mikhailshilkov (commented):
LGTM overall. Agree with Fraser's comment above and asked a few noob questions.

pkg/codegen/dotnet/gen.go
iwahbe: It became unnecessary, I'm not sure when. It's not related to the change.

pkg/codegen/internal/test/testdata/other-owned/dotnet/Other.Example.csproj
iwahbe: It should be configurable. It's not strictly related, but I can add that to this PR.

iwahbe: I think this should be driven by the `publisher` field: https://www.pulumi.com/docs/guides/pulumi-packages/schema/#pulumi-package-schema

pkg/codegen/internal/test/testdata/other-owned/schema.json
iwahbe: Not really. I will add a skipping feature.

pkg/codegen/internal/test/sdk_driver.go
```diff
@@ -29,14 +29,68 @@ type sdkTest struct {
 	// are of the form "$language/$check" such as "go/compile".
 	Checks map[string]CodegenCheck
 
-	// Skip checks, identified by "$language/$check".
-	// "$language/any" is special, skipping generating the
+	// KnownBroken indicates tests that are known to be failing, identified by
+	// "$language/$check". "$language/any" is special, skipping generating the
 	// code as well as all tests.
-	Skip codegen.StringSet
+	KnownBroken codegen.StringSet
```
Frassle, line 35: I think Skip was maybe a better term than KnownBroken. We're skipping the tests for the other languages in this PR just because they're not useful rather than because there's anything broken with them.

pkg/codegen/internal/test/testdata/other-owned/go/codegen-manifest.json
```diff
@@ -0,0 +1,16 @@
+{
```
Frassle, line 1: Can't this file and all the other go + node + python files be deleted now?

pkg/codegen/internal/test/sdk_driver.go
```diff
@@ -29,14 +29,68 @@ type sdkTest struct {
...
-	// This is a helper form of `Skip`.
-	SkipCompileCheck codegen.StringSet
+	// This is a helper form of KnownBroken.
+	KnownBrokenCompile codegen.StringSet
+
+	// IntendedFor specifies languages[/tests] that this test is intended to run
+	// for. Unlike KnownBroken, IntendedFor indicates that the test should only
+	// run for the specified input. It is not used to indicate that other tests
+	// are broken.
+	IntendedFor codegen.StringSet
```
Frassle, line 45: Bit sad having two properties that both result in skipping and nothing to force them to be in sync. Could we not have just set Skip with something like "allLanguages.Subtract("dotnet")"?

pkg/codegen/dotnet/gen.go
```diff
@@ -398,7 +411,7 @@ func (mod *modContext) typeString(t schema.Type, qualifier string, input, state,
 	case *schema.ResourceType:
 		if strings.HasPrefix(t.Token, "pulumi:providers:") {
 			pkgName := strings.TrimPrefix(t.Token, "pulumi:providers:")
-			return fmt.Sprintf("Pulumi.%s.Provider", namespaceName(mod.namespaces, pkgName))
+			return mod.rootNamespace + "." + namespaceName(mod.namespaces, pkgName) + ".Provider"
```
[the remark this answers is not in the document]
iwahbe: I'll switch it back to`Sprintf`. I'm happy to use either. I generally prefer `+` when most of the string components are behind variables (instead of string literals).

pkg/codegen/internal/test/sdk_driver.go
iwahbe: I have reverted back to `Skip` and added the `allLanguages` variable (good idea!). My intent wasn't to provide an inverse to `Skip`, it was to communicate that a test had an intended audience, distinct from just being known to fail.

pkg/codegen/internal/test/testdata/simple-yaml-schema/docs/typeuses/_index.md
```diff
@@ -164,31 +164,31 @@ The TypeUses resource accepts the following [input]({{< relref "/docs/intro/conc
...
+        <span class="property-type"><a href="#object">.<wbr>Example.<wbr>Inputs.<wbr>Object<wbr>Args</a></span>
     </dt>
     <dd>{{% md %}}{{% /md %}}</dd><dt class="property-optional"
             title="Optional">
         <span id="qux_csharp">
 <a href="#qux_csharp" style="color: inherit; text-decoration: inherit;">Qux</a>
 </span>
         <span class="property-indicator"></span>
-        <span class="property-type"><a href="#rubbertreevariety">Pulumi.<wbr>Example.<wbr>Rubber<wbr>Tree<wbr>Variety</a></span>
+        <span class="property-type"><a href="#rubbertreevariety">.<wbr>Example.<wbr>Rubber<wbr>Tree<wbr>Variety</a></span>
```
t0yv0, line 191: This doesn't look good. `.Example`? Starting with a dot?

pkg/codegen/dotnet/gen.go
```diff
@@ -202,7 +205,7 @@ func (mod *modContext) tokenToNamespace(tok string, qualifier string) string {
 	components := strings.Split(tok, ":")
 	contract.Assertf(len(components) == 3, "malformed token %v", tok)
 
-	pkg, nsName := "Pulumi."+namespaceName(mod.namespaces, components[0]), mod.pkg.TokenToModule(tok)
+	pkg, nsName := mod.rootNamespace+"."+namespaceName(mod.namespaces, components[0]), mod.pkg.TokenToModule(tok)
```
t0yv0, line 208: Perhaps handle the "" case?

t0yv0 (approved):
LGTM except something funky around empty namespaces and `.Foo` refs generated starting with a dot.

pkg/codegen/dotnet/gen.go
iwahbe: Forgot to pipe `mod.rootNamespace` into the docs generation.

PR #8737, opened 2022-01-12
Try a locked go toolchain

t0yv0:
# Description

The hypothesis being tested here is that we have OOMs in CI due to more than one instance of the `go` toolchain running at the same time, and the linker being particularly hungry (~300MB RAM) on Windows. The workaround locks the `go` toolchain at OS level so only one instance of `go build` or `go run` can run at a time. 

I've tested it a bit and it seems to work as designed. It slows down Windows verification a bit, the integration tests go into 1h7min territory, but they seem to complete so far without OOM. That'd still be a win in my book if this truly works. I think this is still conceptually better than tuning down test parallelism param (currently -p 10), since tests that do not use Go actually continue to run in parallel.

What would be best is eventually modifying the tests to set runtime.options.binary in Pulumi.yaml and also set `pt.opts.RunBuild: true` in programtest. What this does is it pre-builds the user program with `go build` (now under lock), followed by actual testing of the compiled program (not under lock). This way slow cloud-wait operations can still go in parallel.

The real test is to run the CI a few times and see if the incidence of OOMs is reduced. My sample here is too small.

I'd love to merge this in and run an experiment for a few weeks to see if the stability stats look any better.

Alternative would be to open 100 branches but I do not want to clog the CI workers so real PRs wait on this.

We can always back this out easily.

Fixes # (issue)

## Checklist

### DIFF
--- a/.github/workflows/test-fast.yml
+++ b/.github/workflows/test-fast.yml
@@ -147,6 +147,16 @@ jobs:
         with:
           repo: t0yv0/goteststats
           tag: v0.0.7
+      - name: Lock go toolchain to prevent concurrent OOM on go linker
+        if: ${{ matrix.platform == 'windows-latest' }}
+        run: |
+          LOCKED_GO=$(./scripts/locked-go-toolchain.sh)
+          echo "Adding LOCKED_GO=$LOCKED_GO to PATH"
+          echo $LOCKED_GO >> $GITHUB_PATH
+      - name: Verify go toolchain modification
+        run: |
+          command -v go
+          go version
       - name: Download Pulumi Go Binaries (linux-x64)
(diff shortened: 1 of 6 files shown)


t0yv0:
Windows builds succeeded, but need a larger sample to make inference about OOM freq impact.

Ubuntu tests subject to "Test Timed Out" - probably because the Go cannot acquire a lock. Need to do "maxwait=N-min" in the tool perhaps? Or check that the tool does release the lock correctly? If this issue never hits windows tests, can also make the modification target Windows only.

t0yv0:
Fixed an issue with unlocking correctly in the upstream, trying again.

iwahbe (approved):
Worth a try!

PR #8702, opened 2022-01-07
[sdk/dotnet] Add PluginDownloadURL

iwahbe:
# Description

Fixes #8689 

## Checklist

- [x] I have added tests that prove my fix is effective or that my feature works

- [x] I have updated the [CHANGELOG-PENDING](https://github.com/pulumi/pulumi/blob/master/CHANGELOG_PENDING.md) file with my change

- [ ] Yes, there are changes in this PR that warrants bumping the Pulumi Service API version

PR #8738, opened 2022-01-12
Revert "[sdk/dotnet] Add PluginDownloadURL"

t0yv0:
Reverts pulumi/pulumi#8702 temporarily as we sort out why it broke master builds on Mac OS that were excluded from PR verification to economize on Mac workers unfortunately.


t0yv0:
I'm suspecting something is funny with `+			[]string{dotNetBin, "add", "package", dep, "-v", version, "-s", localNuget}, cwd)` but no proof yet.. The tests seem to work fine locally for me on a Mac, it's confusing.

justinvp (approved):
🤞

pkg/testing/integration/program.go
```diff
@@ -2075,7 +2075,7 @@ func (pt *ProgramTester) prepareDotNetProject(projinfo *engine.Projinfo) error {
 		version := r.Replace(file)
 
 		err = pt.runCommand("dotnet-add-package",
-			[]string{dotNetBin, "add", "package", dep, "-v", version, "-s", localNuget}, cwd)
```
iwahbe, line 2078: I agree that this is probably the problem. It works locally on my Mac, so my initial suspicion is that `localNuget` points to the wrong place on the Mac test suite. But we use `localNuget` to get the version, so it must be correct. 

---

Either way, we can revert to unblock the release and then fix the test.

iwahbe (approved):

pkg/testing/integration/program.go
t0yv0: I'm sorry about this - there's a drift between local dev env and CI env and my recent work makes it a bit worse even to hit some other objectives. Thank you for your patience.

PR #8743, opened 2022-01-13
Accept `*` as an index in property paths

### LINKED ISSUE #8197: ReplaceOnChanges - strconv.ParseInt
I'm trying to use ReplaceOnChanges for an AKS cluster to replace when the availabilityZones change. I'm doing this right now:
```
                ReplaceOnChanges =
                {
                    "agentPoolProfiles[*].availabilityZones"
                }
```

I get the error:
error: invalid array index: strconv.ParseInt: parsing "*": invalid syntax

Shouldn't I be able to use the * syntax in an array? I'm having to do this as a workaround currently:

```
  ReplaceOnChanges =
                {
                    "agentPoolProfiles[0].availabilityZones",
                    "agentPoolProfiles[1].availabilityZones"
                }
```

iwahbe:
# Description
`PropertyPaths` are used as path descriptors with wildcards (`*`). Users
know they can use `root["*"]` to wildcard on any entry in a map. Because
strings and numbers can both be used to index (map vs array), our users
expect to be able to use `root[*]` as a wildcard for an array. We should
parse `[*]` so wildcards work as expected. Internally, this will be
syntactic sugar for `["*"]`, but with much less user confusion.

Fixes #8197
See also: https://github.com/pulumi/pulumi-hugo/issues/908
## Checklist

### DIFF
--- a/sdk/go/common/resource/properties_path_test.go
+++ b/sdk/go/common/resource/properties_path_test.go
@@ -144,6 +144,30 @@ func TestPropertyPath(t *testing.T) {
 		})
 	}
 
+	simpleCases := []struct {
+		path   string
+		parsed PropertyPath
+	}{
+		{
+			`root["*"].field[1]`,
+			PropertyPath{"root", "*", "field", 1},
+		},
+		{
+			`*.bar`,
+			PropertyPath{"*", "bar"},
+		},
+	}
+
+	t.Run("Simple", func(t *testing.T) {
+		for _, c := range simpleCases {
+			t.Run(c.path, func(t *testing.T) {
+				parsed, err := ParsePropertyPath(c.path)
+				assert.NoError(t, err)
+				assert.Equal(t, c.parsed, parsed)
+			})
+		}
+	})
+
 	negativeCases := []string{
 		// Syntax errors
 		"root[",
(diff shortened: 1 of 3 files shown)


sdk/go/common/resource/properties_path_test.go
```diff
@@ -144,6 +144,30 @@ func TestPropertyPath(t *testing.T) {
 		})
 	}
 
+	simpleCases := []struct {
+		path   string
+		parsed PropertyPath
+	}{
+		{
+			`root["*"].field[1]`,
```
Frassle, line 152: Feel there should be a `root[*].field[1]` test as well given that's what this is adding support for

iwahbe: @pgavlin Which tests would you like to see added to `TestPropertyPathContains`? `TestPropertyPathContains` takes as input a `PropertyPath`. Since `foo[*].bar` and `foo["*"].bar` both parse into the same property path `PropertyPath{"foo", "*", "bar"}`, I don't think that my change introduces any new surface area for the `Contains` method.

pgavlin: > `TestPropertyPathContains` takes as input a `PropertyPath`

Ah, so it does--I did not read the setup for that test closely enough. Nevermind, then!

pgavlin (approved):

Issue #8450, opened 2021-11-18
Allow for resources to be made "immutable"


atrauzzi:
- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

When replacing managed SSL certificate(s) for a Google Cloud load balancer, the new certificate creation returns prior to the certificate being viable.  This is totally normal and outside of Pulumis control as it's a slightly delayed process as the domain(s) go through a validation process.

Unfortunately, this also means that when Pulumi creates the new certificate(s), it's deleting the existing certificate which technically should be left in place while the new one provisions.

```csharp
            var names = string.Join("", myProject.InferredRecordSets
                .Where((recordSet) => recordSet.Domain == domain)
                .Select((recordSet) => recordSet.Name)
                .OrderBy((name) => name));

            var hashedNames = BitConverter
                .ToString(MD5.Create().ComputeHash(Encoding.UTF8.GetBytes(names)))
                .Replace("-", "")
                .Substring(0, 8)
                .ToLower();

            var name = $"{domain.Replace(".", "-")}-{hashedNames}";

            return new ManagedSslCertificate(
                name,
                options: new()
                {
                    Parent = managedZone,
                },
                args: new()
                {
                    Name = name,
                    Managed = new ManagedSslCertificateManagedArgs
                    {
                        Domains = myProject.InferredRecordSets
                            .Where((recordSet) => recordSet.Domain.Equals(domain))
                            .Select((recordSet) => recordSet.Name)
                            .ToList(),
                    },
                }
            );
```

I would like to requiest that Pulumi add the ability to have a resource "orphaned" from state when a change is detected.

This would have the effect of making the resource immutable.  Building on the example above, it'd be as simple as:

```csharp
                // ...
                options: new()
                {
                    Parent = managedZone,
                    Immutable = true,
                },
                // ...
```

With this flag set to `true`, the corresponding state would be deleted and the resource would effectively be created anew.

---

@jaxxstorm [suggested on slack](https://pulumi-community.slack.com/archives/C84L4E3N1/p1637264661329800) I create this issue here. :slightly_smiling_face:

fitzoh:
+1 for this being a useful feature.
I'd suggest something other than `immutable` though, I would assume a resource being immutable means that no changes are allowed (like `ignoreChanges` for all fields, or perhaps an error if any fields are changed).

[...]

PR #8746, opened 2022-01-13
Add RetainOnDelete resource option

### LINKED ISSUE #7747: Add `RetainOnDelete` resource option
There are many cases where it is useful or necessary to allow a resource managed by Pulumi to be kept around when the Pulumi resource managing it is deleted (including when the stack it is part of is destroyed).  For example:
1. A database or persistent storage that is created by Pulumi, but which should be intentionally retained when the Pulumi stack that created it is destroyed, allowing for manual processes to deprovision, backup and/or migrate the backing data
2. Resources imported into a Pulumi program but still "owned" by some external system, which should not be destroyed as part of the Pulumi stack lifecycle
3. Resources that have been removed from a backing cloud provider, but where the resource provider does not cleanly treat this as a deletion, where the user wants to indicate they are okay with removing the resource without requiring that the resource provider's destroy operation is run

One way to support this would be a `retainOnDelete: true` resource option, which when set to true causes the resource provider `Destroy` operation to not be executed, instead just implicitly succeeding whenever a destroy on the resource is needed (including during a replace).

Frassle:
# Description

Adds a new resource option "RetainOnDelete" which if true skips calling into the provider to actually delete the resource. Instead the engine simply drops it from our state file (or replaces it in the state file in the case of replaces). Users must be aware that this "leaks" cloud resources as Pulumi is no longer tracking them, but that is the whole point of this option.

Fixes #7747 & #8450

## Checklist


CHANGELOG_PENDING.md
```diff
@@ -1,5 +1,7 @@
 ### Improvements
 
+- [sdk] - Add `RetainOnDelete` as a resource option.
```
Frassle, line 3: Open to name suggestions if anyone thinks RetainOnDelete isn't clear enough. Another idea was "OrphanOnDelete".

pkg/resource/deploy/import.go
```diff
@@ -166,7 +166,7 @@ func (i *importer) getOrCreateStackResource(ctx context.Context) (resource.URN,
 	typ, name := resource.RootStackType, fmt.Sprintf("%s-%s", projectName, stackName)
 	urn := resource.NewURN(stackName, projectName, "", typ, tokens.QName(name))
 	state := resource.NewState(typ, urn, false, false, "", resource.PropertyMap{}, nil, "", false, false, nil, nil, "",
-		nil, false, nil, nil, nil, "")
+		nil, false, nil, nil, nil, "", false)
```
Frassle, line 169: Should probably add a way to set retainOnDelete for imports, does that make sense?

Frassle:
Further thinking about this I'm wondering if this should be more closely related to our current External concept? Maybe that should be a tristate of "managed", "external", "managed but used elsewhere so don't delete"?

t0yv0:
Will be reviewing this tomorrow. There's a bit of conflict that crept in also FYI.

Frassle:
> Will be reviewing this tomorrow. There's a bit of conflict that crept in also FYI.

I'm going to see about working in @AaronFriel idea of different delete behaviors so probably makes sense to hold off for now. I'll give you a nudge when I think it's ready again.

pkg/resource/deploy/step.go
```diff
@@ -337,26 +337,34 @@ func (s *DeleteStep) Logical() bool           { return !s.replacing }
...
-				return resource.StatusOK, nil, err
-			}
+	if preview {
+		// Do nothing in preview
+	} else if s.old.External {
+		// Deleting an External resource is a no-op, since Pulumi does not own the lifecycle.
+	} else if s.old.DeleteBehaviour == resource.DeleteBehaviourDrop {
+		// Deleting a "drop on delete" is a no-op as the user has explicitly asked us to not delete the resource.
+		// But we want to diag warn that Pulumi has stopped tracking this resource
+		msg := fmt.Sprintf("Reference to %s dropped, Pulumi has not deleted this resource", s.old.ID)
```
Frassle, line 355: Open to better wording on this message. Should it be even more clear about the cost implications?

sdk/go/common/resource/delete_behaviour.go
```diff
@@ -0,0 +1,23 @@
+package resource
+
+import pulumirpc "github.com/pulumi/pulumi/sdk/v3/proto/go"
+
+type DeleteBehaviour int32
+
+const (
+	DeleteBehaviourDelete  DeleteBehaviour = DeleteBehaviour(pulumirpc.DeleteBehaviour_DELETE)
+	DeleteBehaviourDrop                    = DeleteBehaviour(pulumirpc.DeleteBehaviour_DROP)
+	DeleteBehaviourProtect                 = DeleteBehaviour(pulumirpc.DeleteBehaviour_PROTECT)
```
Frassle, line 10: Protect here is the same as the protect flag. We could with 4.0 remove the protect bool flag and just use this. Also scope here to add a new Protect mode that doesn't even allow replacement (the current protect will allow resources to be replaced, just not fully deleted). Then again maybe we should split this behavior flag into DeleteBehaviour (i.e. what to do for a Delete) and ReplaceBehaviour (what to do for a Replace)? Like would it be valid to say delete replacements but drop the final delete? Protect against replacements but allow the final delete?

mikhailshilkov:
Let's not merge this until we finalize the discussion in https://github.com/pulumi/pulumi/issues/7747. Let me know if I need to help drive to conclusion.

pkg/resource/deploy/step.go
```diff
@@ -337,26 +337,34 @@ func (s *DeleteStep) Logical() bool           { return !s.replacing }
 func (s *DeleteStep) Apply(preview bool) (resource.Status, StepCompleteFunc, error) {
 	// Refuse to delete protected resources (unless we're replacing them in
 	// which case we will of checked protect elsewhere)
-	if !s.replacing && s.old.Protect {
+	if !s.replacing && (s.old.Protect || s.old.DeleteBehaviour == resource.DeleteBehaviourProtect) {
```
t0yv0, line 340: Repeating a qn here - Protect and DeleteBehaviorProtect seem awfully similar, are they different somehow that we want both?

pkg/resource/deploy/step.go
```diff
@@ -337,26 +337,34 @@ func (s *DeleteStep) Logical() bool           { return !s.replacing }
...
 		return resource.StatusOK, nil, fmt.Errorf("unable to delete resource %q\n"+
 			"as it is currently marked for protection. To unprotect the resource, "+
```
t0yv0, line 342: Does the message to "remove `protect` flag" still apply if the cause is DeleteBehaviorProtect?

pkg/resource/deploy/step.go
t0yv0: Yup this may be a bit tricky. "Pulumi dropped the reference to %s is no longer tracking it, but it remains provisioned in the cloud"? Except it's not necessarily cloud. "remains provisioned?"

sdk/go/common/resource/delete_behaviour.go
t0yv0: Ah so it's exact same. Er, should we just use protect flag? I'm a bit too conservative perhaps - use what's out there already without changing.

t0yv0:
> our current External concept? Maybe that should be a tristate of "managed", "external", "managed but used elsewhere so don't delete"?

Any docs refs on this concept?

t0yv0:
Looked through the code, implementation is fine, just wondering about the UX on these option design, left some questions.

sdk/python/lib/pulumi/resource.py
```diff
@@ -482,6 +482,11 @@ class ResourceOptions:
     require replacement instead of update only if `"*"` is passed.
     """
 
+    retain_on_delete: Optional[bool]
+    """
+    Option to not actually delete resources from providers when the engine calls Delete.
```
t0yv0, line 487: Nit: try to explaining this in terms of what the user gets not how it's implemented. Taking a shot at it:

"If set to True, deletions retain the resource it the cloud instead of deleting it. Pulumi simply stops tracking the deleted resource."

pkg/resource/stack/deployment.go
```diff
@@ -322,6 +322,7 @@ func SerializeResource(res *resource.State, enc config.Encrypter, showSecrets bo
 		Aliases:                 res.Aliases,
 		ImportID:                res.ImportID,
 		SequenceNumber:          res.SequenceNumber,
+		RetainOnDelete:          res.RetainOnDelete,
```
t0yv0, line 325: So.. this means the flag is to be stored in the state file correct?

t0yv0 (approved):
LGTM with nits and final question on whether statefile persist of the flag should be avoided or not (see the ticket).

pkg/resource/stack/deployment.go
Frassle: Has to be. Does mean if you deploy with an old CLI it will reset back to false, but we can't get this information from anywhere else for Deletes.

t0yv0: Hmm okay I'd think for downgrading CLI we could potentially even accept "undefined behavior" here, but your'e saying even then we still need it stored here.

Also curious for my understanding - when we extend ResourceV3, the state backends like "service" just store the extra data, the schema is fully controlled by the code here? We don't need more changes?

CHANGELOG_PENDING.md
lukehoban: I prefer RetainOnDelete over OrphanOnDelete. 

Note this also aligns with CloudFormation’s similar feature: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-deletionpolicy.html. ARM and Terraform don’t have this feature. 

Were there other candidate options we should consider?

PR #8749, opened 2022-01-14
[sdk/python] Fix makefile target ordering, +x .py

AaronFriel:
* `pip` & the `wheel` package are dependencies before lint/build.
* Fix a .py file committed without the execute bit

Used these commands with `watchexec` to make a variety of changes and then backed them out to find the minimum viable patch.

`watchexec "git clean -xfd; make lint"`

`watchexec "git clean -xfd; make build_package"`

And finally:

`watchexec "git clean -xfd; make test_all"`

Initially I thought the issue was that my distro only has `/usr/bin/python3` as I was getting errors on running `python`, but it looks like the issue was primarily activating the environment and making sure dependencies were installed correctly. The primary issue was that which was not idempotent - a second `make` would succeed, but on second the `bdist_wheel` package would be set up and they would succeed.


sdk/python/Makefile
```diff
@@ -18,23 +18,24 @@ TEST_ALL_DEPS = build
 
 ensure::
 	$(PYTHON) -m venv venv
-	. venv/*/activate && python -m pip install -r requirements.txt
+	. venv/*/activate && pip install wheel
```
t0yv0, line 21: The change from `python -m pip` to `pip` breaks my environment because it picks up the global pip instead of the venv pip. D'you need the change or it was done for aesthetics?

```
[nix-shell:~/pulumi/sdk/python]$ make ensure
make ensure
ENSURE:
python3 -m venv venv
. venv/*/activate && pip install wheel
Collecting wheel
  Using cached wheel-0.37.1-py2.py3-none-any.whl (35 kB)
Installing collected packages: wheel
ERROR: Could not install packages due to an OSError: [Errno 13] Permission denied: '/nix/store/w8lqcsrwlwrvmchj46ph9880w9h4m50a-python3-3.9.6/lib/python3.9/site-packages/wheel'
Consider using the `--user` option or check the permissions.

WARNING: You are using pip version 21.1.3; however, version 21.3.1 is available.
You should consider upgrading via the '/nix/store/w8lqcsrwlwrvmchj46ph9880w9h4m50a-python3-3.9.6/bin/python3.9 -m pip install --upgrade pip' command.
make: *** [Makefile:21: ensure] Error 1

[nix-shell:~/pulumi/sdk/python]$ (. venv/*/activate && which pip)
(. venv/*/activate && which pip)
/nix/store/h5j6wymds31fgj0pdhfrdh27qhlw6d1v-python3.9-pip-21.1.3/bin/pip

```

t0yv0:
Tests don't run automatically when you use a forked repo it seems. Trying to trigger the tests.

sdk/python/Makefile
```diff
@@ -18,23 +18,24 @@ TEST_ALL_DEPS = build
...
+	. venv/*/activate && pip install -r requirements.txt
 	mkdir -p $(PYENVSRC)
 
-build_package::
+build_package:: ensure
```
t0yv0, line 25: That one is interesting - I'm wondering why our prior Makefile art does not specify dependencies at all. Given that `ensure` does not cache, this makes it run over and over. But package downloads cache. So it comes to about 3s on my machine. So I think this is a good change to take if it makes dev workflow less confusing to start with, but I wonder what others think.

sdk/python/Makefile
AaronFriel: Interesting - how is it picking up your pip over the just-sourced venv? Is that a nix issue perhaps, something in your nix env modifying the path order and taking precedent over the venv?

AaronFriel:
Replaced by #8774 so that tests can run.

PR #8750, opened 2022-01-14
[sdk/nodejs] Ensure deps installed before lint, tsc

AaronFriel:
# Description

On first run of the makefile here, if a user doesn't have `eslint` or `typescript` globally installed, `yarn run` won't succeed. If they do have those dependencies globally, it'll use their globally installed version, which is not the desired behavior.

This ensures `yarn` is run once before `lint` and `build_package` targets, and since `build_package` is a dependency for all future targets that use `yarn run` it covers the remainder of the cases.

### DIFF
--- a/sdk/nodejs/Makefile
+++ b/sdk/nodejs/Makefile
@@ -17,10 +17,13 @@ include ../../build/common.mk
 
 export PATH:=$(shell yarn bin 2>/dev/null):$(PATH)
 
-lint::
+ensure::
+	yarn
+
+lint:: ensure
 	yarn run eslint -c .eslintrc.js --ext .ts .
 
-build_package::
+build_package:: ensure
 	yarn run tsc
 	cp tests/runtime/jsClosureCases_8.js bin/tests/runtime
 	cp tests/runtime/jsClosureCases_10_4.js bin/tests/runtime


t0yv0:
Again, tests did not run, sadly.

sdk/nodejs/Makefile
```diff
@@ -17,10 +17,13 @@ include ../../build/common.mk
 
 export PATH:=$(shell yarn bin 2>/dev/null):$(PATH)
 
-lint::
+ensure::
+	yarn
```
t0yv0, line 21: Is this a shortcut for yarn install?

t0yv0: So actually this now makes yarn install run twice `make ensure`:

```
[nix-shell:~/pulumi/sdk/nodejs]$ make ensure
make ensure
ENSURE:
yarn install
yarn install v1.22.17
[1/5] 🔍  Validating package.json...
[2/5] 🔍  Resolving packages...
success Already up-to-date.
✨  Done in 0.21s.
yarn
yarn install v1.22.17
[1/5] 🔍  Validating package.json...
[2/5] 🔍  Resolving packages...
success Already up-to-date.
✨  Done in 0.21s.
```

sdk/nodejs/Makefile
```diff
@@ -17,10 +17,13 @@ include ../../build/common.mk
...
+
+lint:: ensure
```
t0yv0, line 23: Same as with Python PR, I wonder why we did not link targets before, I wonder if people have opinions on this. Personally sounds like a change that's worth taking.

sdk/nodejs/Makefile
t0yv0: Perhaps remove the relevant line from common.mk?

```
ensure::
	$(call STEP_MESSAGE)
	@if [ -e 'package.json' ]; then echo "yarn install"; yarn install; fi
``

AaronFriel:
Recreated PR as #8773.

sdk/nodejs/Makefile
AaronFriel: Will check if this line in common.mk can be removed or if we have other places it's causing a `yarn` to be run.

PR #8751, opened 2022-01-14
[sdk/dotnet] Fix failing test

AaronFriel:
# Description

This test failed and, to be honest, I struggled to find good xunit documentation.

It appears there were two issues in this test:

1. The Assert.Matches was failing on the string termination `$`.
2. The automation API output was trimmed in the test on line 42, but the accumulated lines were not. I didn't see an obvious commit including the word trim, but I'm guessing some behavior changed with automation output?

The line 42 test failed on matching the line beginning:

```
   $ curl -sSL https://get.pulumi.com | sh
```

Which when trimmed would not equal itself.

### DIFF
--- a/sdk/dotnet/Pulumi.Automation.Tests/LocalPulumiCmdTests.cs
+++ b/sdk/dotnet/Pulumi.Automation.Tests/LocalPulumiCmdTests.cs
@@ -35,7 +35,7 @@ namespace Pulumi.Automation.Tests
             Assert.Equal(0, result.Code);
 
             Assert.Matches(@"^v?\d+\.\d+\.\d+", result.StandardOutput);
-            Assert.Matches(@"^(warning: A new version of Pulumi[^\n]+\n)?$",
+            Assert.Matches(@"^(warning: A new version of Pulumi[^\n]+\n)?",
                            result.StandardError);
 
             Assert.Equal(Lines(result.StandardOutput), stdoutLines);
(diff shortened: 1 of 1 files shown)


t0yv0:
This LGTM. Sadly the tests didn't run the CI is considering you an external contributor. Maybe it's because the PR is coming form a forked repo rather than a branch.

t0yv0 (approved):
LGTM but need to pass tests on CI before merging.

mikhailshilkov:
Hmm, the master build failure looks related to this change: https://github.com/pulumi/pulumi/runs/4843576289?check_suite_focus=true

Frassle:
Yeh the type changed with the removal of Select from Lines (IEnumerable<string> -> string[]). It was probably hitting an overload explicitly for IEnumerable before.

AaronFriel:
@Frassle it also looks like the test failed because a trailing newline made it in:

```
Expected: String[] ["v3.23.0-alpha.1642436772\n"]
Actual:   List<String> ["v3.23.0-alpha.1642436772"]
```

CRLF / LF based trimming issue for Windows vs other platforms? I see the other auto tests succeeded.

@mikhailshilkov What do you think the right approach is to fix the failing Windows test, some platform specific code or make the test trim non-whitespace characters for all lines on both the expected and actual values?

mikhailshilkov:
> make the test trim non-whitespace characters for all lines on both the expected and actual values

This sounds reasonable to me

Issue #8752, opened 2022-01-14
pulumi config set is changing the entire Pulumi yaml file, not only the specific key


felipegirotti:
## Hello!

- Vote on this issue by adding a 👍 reaction
- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details

After the update to version `3.22.0` we are facing errors on our Pulumi yaml files, we perform the command `config set --path app.version {commit-hash}`. 
And now the pulumi command are changing other values we have as string to number.
We reverted to run on our pipelines the old version `3.21.0` that working fine.

### Steps to reproduce

1. Create a stack and add values number as string, e.g: `MY_CONFIG: "2020_1"`
2. Run `pulumi -s {stack} config set --path app.version latest`
3. Your file Pulumi.{stack}.yaml is changed the value `MY_CONFIG: 20201`

Expected: We expected that only change the value I set in the command
Actual: Pulumi command is changing other values

Frassle:
I think this is due to https://github.com/pulumi/pulumi/pull/8682

Issue #8745, opened 2022-01-13
config get is returning a different result to what's displayed in the portal


wsalahuddin:
We are setting a config value of the build number (default format) from azure pipeline e.g.
20220113.2
but on retrieving this config, the value is being read as 
2.02201132e+07

We only noticed this issue today, i.e. pulumi version 3.22.0 
Not sure if this is this version related or the value of the number.

Note: Pulumi portal displays the correct value, but when you copy the value using the copy icon next to the value it shows the incorrect value.

krisdehaudt:
+1, has the same issue

mikhailshilkov:
Probably related to #8752 and #8682? @iwahbe

PR #8754, opened 2022-01-14
Revert "Unify Map and Value number exporting"

iwahbe:
Reverts pulumi/pulumi#8682

Fixes https://github.com/pulumi/pulumi/issues/8745
Fixes https://github.com/pulumi/pulumi/issues/8752


mikhailshilkov (approved):
Thank you for jumping on this! Do you mind adding a changelog entry?

PR #8755, opened 2022-01-14
Lower test parallelism for Windows testing

t0yv0:
# Description

Reduce probability of Windows test worker VM OOM issue by reducing test parallelism 10 to 1 on Windows. The impact is 60min --> 77min testing duration. I have not yet seen the CI OOM with these changes, but definitely need a larger sample to see if this works or not. I still would like to merge this and continue watching the builds.

Fixes # (issue)

## Checklist

### DIFF
--- a/.github/workflows/test-fast.yml
+++ b/.github/workflows/test-fast.yml
@@ -73,6 +73,10 @@ jobs:
     runs-on: ${{ matrix.platform }}
 
     steps:
+      - name: Reduce TESTPARALLELISM on Windows to work around OOM
+        if: ${{ matrix.platform == 'windows-latest' }}
+        run: |
+          echo "TESTPARALLELISM=1" >> $GITHUB_ENV
       - name: Set PULUMI_TEST_SUBSET env var
         run: |
           echo "PULUMI_TEST_SUBSET=${{ matrix.test-subset }}" >> $GITHUB_ENV
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -62,6 +62,10 @@ jobs:
     runs-on: ${{ matrix.platform }}
 
     steps:
+      - name: Reduce TESTPARALLELISM on Windows to work around OOM
+        if: ${{ matrix.platform == 'windows-latest' }}
+        run: |
+          echo "TESTPARALLELISM=1" >> $GITHUB_ENV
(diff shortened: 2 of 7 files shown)


t0yv0:
Surprisingly TESTPARALLELISM := 5 did not significantly slow things down but we still got a hanging OOM-ish windows-latest/etc build. Fascinating.

Trying lower.

t0yv0:
This is fascinating

```
1h 16m 49s
```

Not nearly as drastic a degradation as I'd expected.

mikhailshilkov:
> The tsc command on line 71 will always use a globally installed TypeScript, regardless of the one installed by yarn on the previous line.

Is something preventing tests from running in a more parallel fashion in the current setup? 🤔

t0yv0:
> Is something preventing tests from running in a more parallel fashion in the current setup? 🤔

Yes. 

The worker VM runs out of memory and dies with non-specific hang (no indication which test). At P=10 it specifically affects Windows VMs.  

Our other axis is more VMs. We can run more VMs on each PR and fan out tests that way. The limitation is that we are at capacity on Mac VMs and that often leads to throughput logjam where all PRs are stuck waiting on free Mac VMs.

iwahbe:
Is there a reason to set `TESTPARALLELISM?=10` in every make file instead of just in `build/common.mk`

.github/workflows/test-fast.yml
```diff
@@ -73,6 +73,10 @@ jobs:
     runs-on: ${{ matrix.platform }}
 
     steps:
+      - name: Reduce TESTPARALLELISM on Windows to work around OOM
+        if: ${{ matrix.platform == 'windows-latest' }}
+        run: |
+          echo "TESTPARALLELISM=1" >> $GITHUB_ENV
```
AaronFriel, line 79: Note for self on this GitHub-ism: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable

AaronFriel (approved):
lgtm, learning more about GitHub Actions with every PR.

t0yv0:
No particular reason on build/common.mk just using what's there already.. leaving for later. Note on last build failure: it did not actually test the change (test-fast.yml@master). And it OOM'd

PR #8756, opened 2022-01-14
Mitigate OOM by killing orphan processes

t0yv0:
# Description

This is not pretty but might help. Still adversely affected by Windows OOMs issues. This mitigation tries to cleanup orphan processes that clog memory between tests. 

Fixes # (issue)

## Checklist


t0yv0:
This is doing some cleanup but still had an OOM incident.

PR #8759, opened 2022-01-14
[python/sdk] Disallow invalid provider option

### LINKED ISSUE #5639: Bug in merge method for `ResourceOptions`
Given a `ComponentResource` that creates an `AWS` resource and a `Kubernetes` resource.

Given also that the `ComponentResource` is passed a `ResourceOptions` with a keyed `providers` property set for e.g. just `Kubernetes`.

If that `ResourceOptions` object is merged with another `ResourceOptions` object using the `merge` function the `keyed` providers property is collapsed into an unkeyed `provider` (singular) property.

Because this new `provider` property is no longer keyed to just `Kubernetes` the provider would also be used for the AWS resource.

This will result in a fatal error, since the AWS resource does not understand a `Kubernetes` provider.

This directly seems to be done on purpose as can be seen here:

https://github.com/pulumi/pulumi/blob/3d2e31289aba2065853d06024a3da9adfc013202/sdk/python/lib/pulumi/resource.py#L522

iwahbe:
When a resource R is created with `ResourceOptinos(provider=P)`, and when P.pkg disagrees with the package providing R, do not create R using P. Instead create R with the provider it finds in `providers` (or default) and pass `P` as an overriding member of `self._providers` to any child resources.

This makes `ResourceOptions(provider=P)` the same as `ResourceOptions(providers=[P])`. This is desirable since `ResourceOptinos(providers=[P]).merge(None) == ResourceOptions(provider=P)` already.

This is progress towards the reasonable expectation that `ResourceOptions(**kargs).merge(None) == ResourceOptions(**kargs)`.

I want to be super careful that this changes existing behavior only in harmless ways. I don't think it is possible to avoid a (technically) breaking change without providing an explicit opt in to correct behavior (`ResourceOptionsV2`) which I do not want. 

# Description

Fixes #5639

## Checklist

### DIFF
--- a/sdk/python/lib/pulumi/resource.py
+++ b/sdk/python/lib/pulumi/resource.py
@@ -982,9 +982,12 @@ class ProviderResource(CustomResource):
         :param bool dependency: True if this is a synthetic resource used internally for dependency tracking.
         """
 
-        if opts is not None and opts.provider is not None:
-            raise TypeError(
-                "Explicit providers may not be used with provider resources")
+        if opts is not None:
+            if not isinstance(opts, ResourceOptions):
+                raise TypeError("opts must be of type ResourceOptions")
+            if opts.provider is not None:
+                raise TypeError(
+                    "Explicit providers may not be used with provider resources")
         # Provider resources are given a well-known type, prefixed with "pulumi:providers".
         CustomResource.__init__(
             self, f"pulumi:providers:{pkg}", name, props, opts, dependency)
--- a/sdk/python/lib/pulumi/resource.py
+++ b/sdk/python/lib/pulumi/resource.py
@@ -806,10 +806,7 @@ class Resource:
             else:
                 # If a provider was specified, add it to the providers map under this type's package
                 # so that any children of this resource inherit its provider.
-                type_components = t.split(":")
-                if len(type_components) == 3:
-                    [pkg, _, _] = type_components
-                    self._providers = {**self._providers, pkg: provider}
+                self._providers = {**self._providers, provider.package: provider}
         else:
             providers = convert_providers(opts.provider, opts.providers)
             self._providers = {**self._providers, **providers}
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -2,3 +2,5 @@
 
 ### Bug Fixes
 
+- [python/sdk] - Disallow passing an invalid Provider `P` with
+  `ResourceOptions(provider=P)` to a custom resource.
--- a/sdk/python/lib/pulumi/runtime/resource.py
+++ b/sdk/python/lib/pulumi/runtime/resource.py
@@ -118,7 +118,16 @@ async def prepare_resource(res: 'Resource',
 
     # Construct the provider reference, if we were given a provider to use.
     provider_ref = None
-    if custom and opts is not None and opts.provider is not None:
+
+    pkg = None
+    try:
+        i = ty.index(":")
+        pkg = ty[:i]
+    except ValueError:
+        # If we don't find the index, we silently move on.
+        pass
+
+    if custom and opts is not None and opts.provider is not None and (opts.provider.package == pkg or pkg is None):
         provider = opts.provider
 
         # If we were given a provider, wait for it to resolve and construct a provider reference from it.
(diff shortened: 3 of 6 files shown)


sdk/python/lib/pulumi/resource.py
```diff
@@ -806,10 +806,7 @@ def __init__(self,
             else:
                 # If a provider was specified, add it to the providers map under this type's package
                 # so that any children of this resource inherit its provider.
-                type_components = t.split(":")
-                if len(type_components) == 3:
-                    [pkg, _, _] = type_components
-                    self._providers = {**self._providers, pkg: provider}
+                self._providers = {**self._providers, provider.package: provider}
```
t0yv0, line 809: Aha so provider gets indexed by provider.package (the one it's for), not the <type> package.

t0yv0:
Pondering if this is a good idea. 

If users do this unintentionally by mistake, it will look like setting provider= has no effect. Can we warn? Should we throw?

Will users ever do this intentionally and we are breaking those use cases? Like can types from multiple packages be assigned into the same provider, for the provider to support multiple packages, or is this a nonsense possibility?

t0yv0:
If we do go forward we do need tests, CHANGELOG entry, and we need to figure out how this affects other language SDKs (perhaps you've done this already but I'm being lazy and not seeing it x-linked).

iwahbe:
@t0yv0 I don't like this solution, but I don't have a better solution. The problem arose from a case like this.
```python
opts = ResourceOptions(providers={"kubernetes": KubeProvider})

# ... Somewhere else

s3.Bucket("aws-bucket", opts=ResourceOptions(protected=true).merge(opts))
```
`KubeProvider` is handed to the aws-provider which results in crashing the aws provider. The root source of the problem is 

```python
opts = ResourceOptions(providers={"kubernetes": KubeProvider})
print(opts.providers)
# {"kubernetes": <object at {KubeProvider location}>}
print(opts.provider)
# None
opts = opts.merge(ResourceOptions())
print(opts.providers)
# None
print(opts.provider)
# <object at {KubeProvider location}>
```

I can think of a case where programs check `opts.providers` and `opts.provider`. It's harder for me to think of a case when a program intentionally passes a provider from the wrong package to a resource. We can't warn, because the behavior comes up from a correct program. The warning would not be actionable. 

I'm open to alternate ideas to address the problem.

t0yv0:
In your example `opts = opts.merge(ResourceOptions())` is the most surprising to me, why did it promote stuff from `providers` to `provider` losing the `"kubernetes"` label? This seems so confusing.

t0yv0:
So I was thinking maybe there is a problem with the merge method? But if your solution is well thought ought it sounds like. I'm at EOD unable to reason properly will not be able to look at it again till Wed.

iwahbe:
> In your example `opts = opts.merge(ResourceOptions())` is the most surprising to me, why did it promote stuff from `providers` to `provider` losing the `"kubernetes"` label? This seems so confusing.

I'm not sure why the promotion happens. We don't really loose the Kubernetes label (we assume in other parts of merge that we can reconstruct the map from `{provider.package: provider}`. I agree that it is super confusing. What we actually loose is that (at the moment) `providers` is a list of optional providers to use if they apply, and `provider` is a optional provider that if supplied we must use, and if we have a single member of `providers` we promote it to `provider`. 

It looks like we don't do this in Go, but we do do it in .NET.

Frassle:
> if we have a single member of providers we promote it to provider.

Agreeing with Anton, this promotion sounds awfully suspect.

t0yv0:
Promoting new reference from the Slack discussion in case this is relevant. When SDKs diverge in behavior Node SDK is more authoritative:

```
/**
 * ComponentResourceOptions is a bag of optional settings that control a component resource's behavior.
 */
export interface ComponentResourceOptions extends ResourceOptions {
    // !!! IMPORTANT !!! If you add a new field to this type, make sure to add test that verifies
    // that mergeOptions works properly for it.

    /**
     * An optional set of providers to use for child resources. Either keyed by package name (e.g.
     * "aws"), or just provided as an array.  In the latter case, the package name will be retrieved
     * from the provider itself.
     *
     * In the case of a single provider, the options can be simplified to just pass along `provider: theProvider`
     *
     * Note: do not provide both [provider] and [providers];
     */
    providers?: Record<string, ProviderResource> | ProviderResource[];

    // !!! IMPORTANT !!! If you add a new field to this type, make sure to add test that verifies
    // that mergeOptions works properly for it.
}
```

iwahbe:
Closed in favor of https://github.com/pulumi/pulumi/pull/8770

PR #8766, opened 2022-01-17
[sdk/nodejs] Support top-level default exports in ESM

### LINKED ISSUE #8725: NodeJS + ESM: no way to set an object as stack output
## Hello!

## Issue details
Directly relates to ESM: #5161, https://github.com/pulumi/pulumi/issues/4017#issuecomment-601853157

### Steps to reproduce

1. Create a typescript project (defaults to using CommonJS modules)
2. put inside `index.ts`:
    ```typescript
    async function main() { return {helloWorld: 123} };
    export default main();
    ```
3. Run `pulumi up`; stack output is `{default: {"helloWorld": 123}}`, oops, not what we want
4. Update `index.ts`:
    ```typescript
    async function main() { return {helloWorld: 123} };
    module.exports = main();
    ```
5. Run `pulumi up`; stack output is now: `{"helloWorld": 123}`; perfect !
6. Follow these steps to use ESM https://github.com/pulumi/pulumi/issues/5161#issuecomment-1010018506
7. `module.export` isn't available with ESM, only `export` works. :(

Expected: I have a way of settings an object as stack output just like `module.exports = whatever` results in outputs `{ whatever }` 
Actual: `export default whatever` results in outputs `{ default: whatever }`, I have no equivalent to  `module.exports = whatever`.

lukehoban:
In #7764 and #8655 we added support for ESM entrypoints.  However, ESM "default exports" were handled just as "normal" in Node.js dynamic import of ESM - as a `default` proeprty in the export object.

This is not a particularly useful behaviour for Pulumi program entry points, and doesn't quite match some of the special logic we apply to non-object exports in CommonJS modules (invoking exported functions, and then awaiting exports promises).

Instead, this change adds support for default exports, treating the default export (if present) as the full returned export value.

It is for now an error to have both a default export and named exports, since it is unclear what this should mean.  In the future, we could potentially relax this and define how these two sets of exports are merged.

This is technically a breaking change from the support added in the recent releases, but only in a narrow case, and in that case the Pulumi stack exports were almost certainly not what the user wanted.

Fixes #8725, which includes a motivating example where this is ~necessary.

### DIFF
--- a/tests/integration/nodejs/esm-ts-default-export/index.ts
+++ b/tests/integration/nodejs/esm-ts-default-export/index.ts
@@ -1,0 +1,3 @@
+// Copyright 2016-2021, Pulumi Corporation.  All rights reserved.
+
+export default async function() { return {helloWorld: 123} }
--- a/tests/integration/nodejs/esm-ts-default-export/index.ts
+++ b/tests/integration/nodejs/esm-ts-default-export/index.ts
@@ -1,3 +1,3 @@
-// Copyright 2016-2021, Pulumi Corporation.  All rights reserved.
+// Copyright 2016-2022, Pulumi Corporation.  All rights reserved.
 
 export default async function() { return {helloWorld: 123} }
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -20,6 +20,9 @@
 - [cli] - Add support for overriding plugin download URLs.
   [#8798](https://github.com/pulumi/pulumi/pull/8798)
 
+- [sdk/nodejs] - Support top-level default exports in ESM.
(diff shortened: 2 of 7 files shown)


tests/integration/nodejs/esm-ts-default-export/index.ts
```diff
@@ -0,0 +1,3 @@
+// Copyright 2016-2021, Pulumi Corporation.  All rights reserved.
+
+export default async function() { return {helloWorld: 123} }
```
iwahbe, line 3: Nit: newline, correct year for copyright

iwahbe (approved):
LGTM

t0yv0 (approved):

PR #8767, opened 2022-01-18
Change Output<T>.ToString() to return a message

### LINKED ISSUE #4824: Print a helpful error message in Output.ToString
Users are very commonly confused by calling `ToString()` method on outputs ([example](https://github.com/pulumi/pulumi-azure/issues/603)). They get some default string like `Pulumi.Output``1[System.String]` which is never what they expect.

Let's consider overriding `.ToString()` to print a helpful error message pointing to the docs.

Frassle:
# Description

Fixes #4824

Changes `Output<T>.ToString()` to return an informative message similar to what we return for `toString` in the node sdk.

## Checklist


mikhailshilkov (approved):
Thank you, that looks great to me

PR #8770, opened 2022-01-18
[sdk/python] Remove merge options promotion

### LINKED ISSUE #5639: Bug in merge method for `ResourceOptions`

iwahbe:
# Description

This is the simple alternative to https://github.com/pulumi/pulumi/pull/8759. I think it is more likely to break existing code, but results in ideal merge behavior. 
The semantics derive from https://github.com/pulumi/pulumi/issues/8796

Fixes #5639

## Checklist


sdk/python/lib/test/langhost/component_provider_resolution/__main__.py
```diff
@@ -0,0 +1,55 @@
...
+        }, opts=opts)
+
+class OtherProvider(ProviderResource):
+    def __init__(self, name):
+        super().__init__("other", name)
+
+class CombinedComponent(ComponentResource):
+    def __init__(self, name, opts=None):
+        super().__init__("combined:index:CombinedResource", name, opts=opts)
+        parent = ResourceOptions(parent=self).merge(opts)
```
t0yv0, line 38: Is merge typically called by end-users or we call it in the SDK? checking for refs.

t0yv0:
I found this also:

```
def convert_providers(
        provider: Optional['ProviderResource'],
        providers: Optional[Union[Mapping[str, 'ProviderResource'],
                                  Sequence['ProviderResource']]]) -> Mapping[str, 'ProviderResource']:
    """
    Merge all providers opts (opts.provider and both list and dict forms of opts.providers) into a single dict.
    """
    if provider is not None:
        return convert_providers(None, [provider])

    if providers is None:
        return {}

    if isinstance(providers, Mapping):
        return providers

    result = {}
    for p in providers:
        result[p.package] = p

    return result
```

Does that still look right? AFAIK this only affects remote=True (Multi-Language-Component resources) during processing of register_resource.

sdk/python/lib/test/langhost/component_provider_resolution/__main__.py
iwahbe: We use it in generated SDKs, but I expect end users to call it in the wild. Writing component resources which behave as expected generally involves calling `merge` manually.

t0yv0:
The changes look sane to me given our discussion but I'm a bit worried about cross-language consistency and how this compares to `export function mergeOptions` in Node for example, where `normalizeProviders` promotes 1 providers to provider.

t0yv0:
@iwahbe just to clarify this is ready for final review or you'll be making changes post the design doc discussion? Thanks a lot for taking care of this.

iwahbe:
@t0yv0 I have implemented the changes discussed. This is ready for final review.

sdk/python/lib/pulumi/resource.py
```diff
@@ -796,26 +795,23 @@ def __init__(self,
...
+        if len(type_components) == 3:
+            [pkg, _, _] = type_components
+
+        opts.provider, providers = self._get_providers(t, pkg, opts)
 
         self._protect = bool(opts.protect)
         self._provider = opts.provider if custom else None
+        if self._provider and self._provider.package != pkg:
+            action = (
+                (opts.urn is not None and "get") or
```
t0yv0, line 809: Oof.

`"get" if opts_run is None else "read" if opts_id is None else "register"`?

sdk/python/lib/pulumi/resource.py
```diff
@@ -832,6 +828,57 @@ def __init__(self,
...
+
+        # The provider supplied by the parent
+
+        # Cast is safe because the `and` will resolve to only None or the result
+        # of get_provider (which is Optional[ProviderResource]).
+        # See https://github.com/python/mypy/issues/12030
+        parent_provider = cast(Optional[ProviderResource], opts.parent and opts.parent.get_provider(t))
+        provider = opts.provider or ambient_provider or parent_provider
+
+        if pkg and opts.provider:
```
t0yv0, line 867: Condition looks correct.

CHANGELOG_PENDING.md
```diff
@@ -11,6 +11,11 @@
 
 ### Bug Fixes
 
-- [codegen/nodejs] - Generate an install script that runs `pulumi plugin install` with 
+- [sdk/python] - Prevent `ResourceOptions.merge` from promoting between the
+  `.provider` and `.providers` fields. This is a subtle breaking change. This changes
```
t0yv0, line 15: These are user facing docs. I'm wondering if we can't describe coherently a scenario that's breaking and what the user should do should we just omit the words "subtle breaking change"? Can continue presenting it as a bugfix for the interaction of the two fields (that is treat current behavior prior to fix as an error).

t0yv0 (approved):
LGTM with minor nits.

PR #8771, opened 2022-01-18
[sdk/dotnet] Fix test by ensuring both sides of test strings trimmed

AaronFriel:
Regarding:
- #8751

This code appears to hit an issue on Windows where the expected & actual values in the assertion are trimmed differently. That PR fixed the issue by not trimming the left hand side on Linux, but on Windows, a trailing `\n` is found in the expected value:

```
Expected: String[] ["v3.23.0-alpha.1642436772\n"]
Actual:   List<String> ["v3.23.0-alpha.1642436772"]
```

The types are, I believe, not relevant for the assertion though the `\n` at the end is.

This test will ensure that both sides are an IEnumerable<String> and are trimmed before asserting on. There is some underlying behavior here, likely related to Windows using CRLF line termination, which isn't relevant to the test at hand.

### DIFF
--- a/sdk/dotnet/Pulumi.Automation.Tests/LocalPulumiCmdTests.cs
+++ b/sdk/dotnet/Pulumi.Automation.Tests/LocalPulumiCmdTests.cs
@@ -38,13 +38,14 @@ namespace Pulumi.Automation.Tests
             Assert.Matches(@"^(warning: A new version of Pulumi[^\n]+\n)?",
                            result.StandardError);
 
-            Assert.Equal(Lines(result.StandardOutput), stdoutLines);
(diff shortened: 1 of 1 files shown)


AaronFriel:
@mikhailshilkov related to this, today I'm going to set up my dev env on Windows so I can reproduce bugs like this.

mikhailshilkov (approved):
LGTM

t0yv0:
Can we back this out please as it wasn't tested by CI (need to run `/run-acceptance-tests`)? and looks like there is a failure on master. Once it passes the CI tests we can merge it back. Thanks a lot! @AaronFriel

PR #8773, opened 2022-01-19
[sdk/nodejs] Ensure deps installed before lint, tsc

AaronFriel:
On first run of the makefile here, if a user doesn't have `eslint` or
`typescript` globally installed, `yarn run` won't succeed. If they do
have those dependencies globally, it'll use their locally installed
version, which is not the desired behavior.

This ensures `yarn` is run once before `lint` and `build_package`
targets, and since `build_package` is a dependency for all future
targets that use `yarn run` it covers the remainder of the cases.

Recreating #8750 with a branch on upstream vs fork to run tests.


iwahbe (approved):
LGTM

PR #8778, opened 2022-01-19
Revert "[sdk/dotnet] Fix test by ensuring both sides of test strings trimmed"

AaronFriel:
Reverts pulumi/pulumi#8771

Per @t0yv0 who observed we're still getting inconsistency in test results & an acceptance test failure on Windows even with this change. Will debug from a Windows dotnet build soon.


t0yv0 (approved):
Thanks!

PR #8783, opened 2022-01-19
Reduce windows PR testing

t0yv0:
# Description

Just as we do already for the Mac OS testing, this PR reduces the scope of PR testing to exclude most integration tests on Windows, leaving only the fall-through `etc` test subset. The reason to do this is that we are losing a lot of builds to Windows OOM issues that are making builds flaky and are not related to regressions. If there are Windows specific regressions introduced on PRs that are caught by these tests, they will be caught once master.yml workflow runs and breaks, and we will have to back out the PR. This is inconvenient but optimizes for the common case - PRs that do not break on windows will go in faster.

Fixes # (issue)

## Checklist

### DIFF
--- a/.github/workflows/test-fast.yml
+++ b/.github/workflows/test-fast.yml
@@ -62,11 +62,14 @@ jobs:
           - etc
         exclude:
           - platform: macos-latest
-        # Only run catch-all `etc` test-subset on Mac for PR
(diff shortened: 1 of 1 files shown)


justinvp (approved):

AaronFriel (approved):
lgtm,

PR #8788, opened 2022-01-19
[sdk/dotnet] Fix inconsistent automation test

AaronFriel:
This test was flakey for me when building on Windows & *nix using different versions of the dotnet framework or depending on whether I had an up to date Pulumi version installed. The regex match is to blame.

On old versions during CI when we do build a version locally, we expect a message on stdout advising that a new version is available. If the user hasn't built the language provider locally or has the current version built, we expect an empty string.

Rather than use regex, which had inconsistent matching behavior and as usual, caused two problems, we just test if the string is nonempty first.

Once and for all, this resolves this test passing the matrix consisting of:

 Pulumi version   |    OS   | .NET SDK
 ---------------- | ------- | --------
 3.21.1 (old)     | Windows | 3.1
 3.21.1 (old)     | Windows | 6.0
 3.21.1 (old)     | Linux   | 3.1
 3.21.1 (old)     | Linux   | 6.0
 3.22.2 (current) | Windows | 3.1
 3.22.2 (current) | Windows | 6.0
 3.22.2 (current) | Linux   | 3.1
 3.22.2 (current) | Linux   | 6.0

### DIFF
--- a/sdk/dotnet/Pulumi.Automation.Tests/LocalPulumiCmdTests.cs
+++ b/sdk/dotnet/Pulumi.Automation.Tests/LocalPulumiCmdTests.cs
@@ -3,6 +3,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Pulumi.Automation.Commands;
 using Xunit;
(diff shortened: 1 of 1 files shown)


AaronFriel:
@t0yv0 our test harnesses on CI do not validate the matrix I sent you, but I can assure you I personally ran each test on a Windows & Ubuntu system.

I briefly added a test that was something like `static void ValidateRegexBehavior` but decided that the assertion was much easier to reason about without the regex.

t0yv0 (approved):

t0yv0:
It's on my TODO list to start adding validators for different versions of .NET (not on PRs probably since we're out of capacity but maybe in master.yml), can look at adding 6.

PR #8791, opened 2022-01-20
Tighter test timeout

t0yv0:
# Description

From our test stats longest tests take about 4m. This PR tightens test timeout so Go test would panic and abort when a test gets stuck earlier. The hope is that this would trade nonspecific OOM black-box failures for failures in specific tests which we can then skip or debug individually.

Fixes # (issue)

## Checklist


t0yv0:
OK not practical, this ends up killing TestConstructMethodsPython after 15m in although it self-reports taking 4 min.

PR #8792, opened 2022-01-20
Introduce timeout to commands run during testing

t0yv0:
# Description

The changes will kill the sub-process and fail with timeout after (currently hardcoded) 5 min.

Driven by https://github.com/pulumi/pulumi/runs/4881353869?check_suite_focus=true timing out after 1h on `pulumi login`.

Timeout value might need tuning.

But there is hope this can increase resiliency of the test runs.

Unfortunately killing the sub-process may still leave sub-sub-processes running as orphans and consuming memory.

Fixes # (issue)

## Checklist

### DIFF
--- a/.github/workflows/run-build-and-acceptance-tests.yml
+++ b/.github/workflows/run-build-and-acceptance-tests.yml
@@ -142,7 +142,7 @@ jobs:
   test:
     name: Test
     needs: build
-    uses: pulumi/pulumi/.github/workflows/test-fast.yml@master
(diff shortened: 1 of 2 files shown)


AaronFriel:
@t0yv0 how concerned would you be about leaving orphaned resources behind if we see a spike in tests being canceled at the cutoff?

AaronFriel (approved):

t0yv0:
We're already leaving orphan processes behind, but yes, long term it's a problem. I'm thinking maybe declining this change since it's not fixing the lost builds.

PR #8798, opened 2022-01-21
Add support for overriding plugin download URLs

justinvp:
# Description

This commit adds a variable that can be set with a linker flag to override the plugin download URL for matching plugin names.

For example, compiling the Pulumi CLI with the following linker flag:

- `-X [module path]/workspace.pluginDownloadURLOverrides=^foo.*=https://foo,^bar.*=https://bar`

Would make it so that plugin names that start with "foo" are downloaded from https://foo and names that start with "bar" are downloaded from https://bar.

## Checklist


Frassle (approved):
> For example, compiling the Pulumi CLI with the following linker flag:

I was thinking why would you want to do this with a linker flag but then we don't really have a good story for global config so I guess this is as sane an option as anything else for now.

t0yv0 (approved):

justinvp:
> I was thinking why would you want to do this with a linker flag but then we don't really have a good story for global config so I guess this is as sane an option as anything else for now.

The exact same thought process went through my head.

PR #8800, opened 2022-01-21
Use Go build cache in tests

t0yv0:
# Description

Copying Go cache setup from build.yml to test.yml, since test.yml rebuilds the tests it seems like it will benefit.

Minor issue: matrix.platform != matrix.os so it won't be the same exact cache key as in build.yml.

Fixes # (issue)

## Checklist

### DIFF
--- a/.github/workflows/test-fast.yml
+++ b/.github/workflows/test-fast.yml
@@ -106,7 +106,7 @@ jobs:
           path: |
               ${{ steps.go-cache-paths.outputs.go-build }}
               ${{ steps.go-cache-paths.outputs.go-mod }}
-          key: ${{ matrix.platform }}-go-cache-${{ hashFiles('**/go.sum') }}
+          key: ${{ matrix.platform }}-${{ matrix.test-subset }}-go-test-cache-${{ hashFiles('**/go.sum') }}
       - name: Set up Python ${{ inputs.python-version }}
         uses: actions/setup-python@v1
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/test-fast.yml
```diff
@@ -95,6 +95,18 @@ jobs:
...
+          echo "::set-output name=go-build::$(go env GOCACHE)"
+          echo "::set-output name=go-mod::$(go env GOMODCACHE)"
+      - name: Go Cache
+        uses: actions/cache@v2
+        id: go-cache
+        with:
+          path: |
+              ${{ steps.go-cache-paths.outputs.go-build }}
+              ${{ steps.go-cache-paths.outputs.go-mod }}
+          key: ${{ matrix.platform }}-go-cache-${{ hashFiles('**/go.sum') }}
```
t0yv0, line 109: Possibly may need `test-subset` in the cache key.

AaronFriel: Do you think we should just refactor matrix.platform to matrix.os here and add -latest everywhere it's used so we can ensure cache reuse?

AaronFriel (approved):
LGTM, non-blocking question for re-using the cache across multiple workflows.

.github/workflows/test-fast.yml
t0yv0: It's possible.. But I'm also wondering if distinct keys are actually a good thing, since workloads exercise different subset of the builds, the contents of the cache are slightly different, and if these different workloads compete for what's written into the shared cache key maybe it's better to keep them separate. At the cost of having more data in the cache. I do not know what the limits/costs are of having more data in the cache yet, but perhaps that is not yet a factor at all?

AaronFriel: I don't yet know enough about GitHub's cache to weigh in either, good things for us to consider going forward.

PR #8804, opened 2022-01-21
automation (python): Write canonical yamls to force quotes for strings

### LINKED ISSUE #8776: aws:allowedAccountIds with leading 0 does not work with automation in python
## Hello!

## Issue details

If we try to set `aws:allowedAccountIds` in pulumi automation with python, and the account id starts with a 0, that leading 0 is dropped by pulumi cli.

### Steps to reproduce
```
    workspace = LocalWorkspace(
        env_vars={
            "PULUMI_SELF_MANAGED_STATE_LOCKING": "1"
        },
        work_dir=tmpdir,
        program=function,
        project_settings=ProjectSettings(
            name=project_name,
            runtime=ProjectRuntimeInfo(name="python"),
            backend=ProjectBackend(
                url=f"s3://{pulumi_state_bucket}"),
        ),
        stack_settings={stack_name: StackSettings(config={
            "aws:region": "eu-west-1",
            "aws:allowedAccountIds": [account_id]
        })}
    )
```
If the account_id starts with a 0, it is dropped, and you get an error on execution because the allowedAccountId does not match the current account id.
The account_id type is str in python, but written by pulumi_automation without quotes. Because automation writes the .yaml files we can't manually add quotes around the value. Automation reads and writes the value without quotes but still preserving the leading zero, but pulumi cli reads and rewrites the value without the leading 0

Related issue: #8401 
This seems to be caused by pulumi cli eagerly parsing config values as int, instead of the defined datatype. AllowedAccountIds in [pulumi-resource-aws.json](https://raw.githubusercontent.com/pulumi/pulumi-aws/master/provider/cmd/pulumi-resource-aws/schema.json)
```
            "allowedAccountIds": {
                "type": "array",
                "items": {
                    "type": "string"
                }
            },
```

As a quick-fix, adding ```canonical=True``` to ```yaml.dump()``` in [LocalWorkspace.save_stack_settings()](https://github.com/pulumi/pulumi/blob/master/sdk/python/lib/pulumi/automation/_local_workspace.py#L162) causes the values to be written in Pulumi.stack.yaml with quotes.

wannessels:
https://github.com/pulumi/pulumi/issues/8776

# Description

automation (python): write canonical yamls to force quotes for strings
Fixes #8776 

## Checklist

If you can suggest a way to write a meaningful test combining automation and cli, I will try to add it.

### DIFF
--- a/sdk/python/lib/pulumi/automation/_local_workspace.py
+++ b/sdk/python/lib/pulumi/automation/_local_workspace.py
@@ -133,7 +133,7 @@ class LocalWorkspace(Workspace):
             if found_ext == ".json":
                 json.dump(writable_settings, file, indent=4)
             else:
-                yaml.dump(writable_settings, stream=file)
+                yaml.dump(writable_settings, stream=file, canonical=True)
 
     def stack_settings(self, stack_name: str) -> StackSettings:
         stack_settings_name = get_stack_settings_name(stack_name)
@@ -159,7 +159,7 @@ class LocalWorkspace(Workspace):
             if found_ext == ".json":
                 json.dump(settings._serialize(), file, indent=4)
             else:
-                yaml.dump(settings._serialize(), stream=file)
+                yaml.dump(settings._serialize(), stream=file, canonical=True)
 
     def serialize_args_for_op(self, stack_name: str) -> List[str]:
         # Not used by LocalWorkspace


t0yv0 (approved):
This LGTM, I'm not a YAML expert but it seems it should be safe. It would be great to add a little test case before merging so we would not regress on this. @iwahbe would you be up for that or need me to? Thank you.

iwahbe:
@t0yv0 I've been unable to write a test that fails without the change but passes with it. If you have time to do so that would be great. Alternatively, if you suggest a test case I would be happy to implement it.

t0yv0:
Well I've taken a look at this library. `canonical=True` makes readability take a great hit. It's not a free change. Granted, this is automation API so it's somewhat out of the way of users interacting with it, but if someone was to debug it, this is unfortunate. Perhaps most worrisome is the problem that the report is about account_id being read incorrectly by the Go-based CLI, perhaps the fix should go there? As you see below, Python is quoting zero-prefixed strings.

```
>>> [print(yaml.dump({"foo": "bar", "account_id": "01234"}, canonical=c)) for c in [True, False]]
[print(yaml.dump({"foo": "bar", "account_id": "01234"}, canonical=c)) for c in [True, False]]
---
!!map {
  ? !!str "account_id"
  : !!str "01234",
  ? !!str "foo"
  : !!str "bar",
}

account_id: '01234'
foo: bar

```

mikhailshilkov:
Oh, the canonical output is very different from what I would expect! I guess I don't know what canonical means - is there a doc for this handy?

As it looks above, this PR sounds like a no-go to me...

PR #8805, opened 2022-01-21
Fix invoke functions without args

danielrbradley:
# Description

Example: `invoke("aws-native::getPartition")` has no additional args object passed therefore the args length is 1.

Fallback to disable rendering of the output form for this case.

Fixes #8805 

## Checklist


pkg/codegen/internal/test/testdata/aws-native-pseudo-params-pp/aws-native-pseudo-params.pp
```diff
@@ -0,0 +1,57 @@
...
+	enableLogFileValidation = true
+	includeGlobalServiceEvents = true
+	isMultiRegionTrail = true
+	cloudWatchLogsLogGroupArn = invoke("aws-native::importValue", {
+		name = "TrailLogGroupTestArn"
+	}).value
+	cloudWatchLogsRoleArn = invoke("aws-native::importValue", {
+		name = "TrailLogGroupRoleTestArn"
+	}).value
+	kmsKeyId = invoke("aws-native::importValue", {
```
t0yv0, line 25: Program generation breaks for me unless I specify "kMSKeyId" here...but only node code generation emits a legible error message. interesting..

pkg/codegen/pcl/binder_test.go
```diff
@@ -13,37 +13,48 @@ import (
 )
 
 func TestBindProgram(t *testing.T) {
-	files, err := ioutil.ReadDir(testdataPath)
+	testdata, err := ioutil.ReadDir(testdataPath)
```
t0yv0, line 16: Thanks so much for this! This test was a no-op before your fix. Quite an omission.

pkg/codegen/pcl/invoke.go
```diff
@@ -114,7 +118,7 @@ func (b *binder) zeroSignature() model.StaticFunctionSignature {
 }
 
 func (b *binder) signatureForArgs(fn *schema.Function, args model.Expression) (model.StaticFunctionSignature, error) {
-	if b.useOutputVersion(fn, args) {
+	if args != nil && b.useOutputVersion(fn, args) {
```
t0yv0, line 121: The spirit of this is reasonable.

t0yv0:
@danielrbradley do you have an issue describing the end-user problem with this? 

I am a little worried because I've added your example to programgen tests and they are all broken (or almost). Go, TS, Dotnet are broken. Python generates code but based on how others are broken it might have runtime errors I need to check. checking docs generation now.

I will be happy to follow up with PRs that demonstrate these breakages and file issues against them I'm just wondering if this fix of yours is sufficient to unblock the use case so we can backlog fixing those things, or you need program generation to work, in which case you are still blocked?

danielrbradley:
@t0yv0 the current issue is that the latest pulumi/pulumi version in aws-native causes a panic during example generation (due to accessing `args[1]`), so the impact is not being able to upgrade. 

This is my first time looking at example generation, so may well be some additional issues with cf2pulumi that's causing the invalid pcl which I don't yet understand.

I've changed direction a bit and am just emitting diagnostics if `len(args) == 1` to avoid the panic and then should be able to investigate more on the aws-native side what's going on. Have left the aws-native test setup code in there for now though.

t0yv0 (approved):
Sorry to see your example being removed? I thought that was very useful. Sorry if I was not being clear here:

- your changes should ship I think, if `invoke("foo")` without args is intended to be supported (I think it is), good changes

- the example was uncovering further bugs in specific language programgen backends. PCL was fine but the generated code would not compile. I'd like to add that to our backlog so we can fix, I was just wondering about your use case so I don't miss the actual problem you're having we should prioritize fixing that.


TYVM.

danielrbradley:
Ah ok, shall we get this merged with the basic diagnostics so we can proceed with updating aws-native? Then I'll put my example code back onto a new branch so we can take some more time to figure out what's going on there.

t0yv0:
Yes no problem merging this at all.

t0yv0:
You have access right? Sorry if you need me to click "Squash and merge" I can do that just lmk.

PR #8807, opened 2022-01-21
Make PackageName just a QName

Frassle:
Since https://github.com/pulumi/pulumi/pull/8695 Name and thus QName has
allowed hyphens, making PackageName just the same as a QName now.

### DIFF
--- a/sdk/go/common/tokens/names.go
+++ b/sdk/go/common/tokens/names.go
@@ -124,20 +124,10 @@ func (nm QName) Namespace() QName {
 	return QName(qn)
 }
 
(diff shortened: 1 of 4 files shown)


iwahbe (approved):
I'm not sure what part of `sdk/go/common`is considered part of the public API. I would be aware that this is changing the `tokens` module API. 
Otherwise, LGTM

Frassle:
> I'm not sure what part of sdk/go/common is considered part of the public API. I would be aware that this is changing the tokens module API.

Yeh I'm not sure how much to worry about that, github search didn't show any use of these in the pulumi organisation at least.

PR #8808, opened 2022-01-21
Make StackName a tokens.Name not QName

Frassle:
Many references to stack names in the codebase used `tokens.QName` as the type, but a stack name is just a `tokens.Name` it doesn't have any delimiter parts. A full stack reference like "org/project/stack" would be a `QName`. This PR simply goes through the codebase fixing up stack names to just use `Name` and following compiler errors to fix up parameter and member lists affected by that change. It's also a bit stricter about using `tokens.AsName` rather than just `tokens.Name` to cast to `Name` to ensure assert checks are hit.


pkg/backend/filestate/backend.go
```diff
@@ -254,7 +254,7 @@ func (b *localBackend) SupportsOrganizations() bool {
 }
 
 func (b *localBackend) ParseStackReference(stackRefName string) (backend.StackReference, error) {
-	return localBackendReference{name: tokens.QName(stackRefName)}, nil
+	return localBackendReference{name: tokens.AsName(stackRefName)}, nil
```
Frassle, line 257: Little concerning that we've always just cast the string to QName here. While client code hopefully called ValidateStackName before calling this nothing required that. Thinking it wouldn't hurt to add here and would save an assert panic if there are any mistakes.

Issue #5749, opened 2020-11-12
Diff not generated for protected resource which needs to be replaced


nesl247:
When a protected resource needs to be replaced, there is no diff available to see why it needs to be replaced. It only generates an error stating it cannot replace a protected resource.

## Expected Behavior
The diff should still be generated.

## Current Behavior
```
error: Preview failed: refusing to delete protected resource 'urn:pulumi:staging::shop-front::gcp:redis/instance:Instance::shop-front-ha'
```

## Steps to Reproduce
1. Create any resource as protected
2. Make a change to the resource that would require it to be replaced
3. Try to run `pulumi up`
4. Error

## Context (Environment)
In this case we are trying to deploy a change to k8s which has a redis instance in GCP in this project. I don't know what the cause for the replacement is because of this issue.

wilsoncg:
+1
I am experiencing the same bug, very confusing. Preview (& up) fails, reports that old state & new state are the same, yet wishes to perform a replacement on a protected resource & fails with no diff generated.

This is a fresh import into a stack, with a newly minted corresponding C# project. I would expect pulumi to report there is nothing to do for a 'same' operation. Why is it failing?

How should I proceed?

```
PS> pulumi preview --json
```
```javascript
{
    "config": {
        "azure:location": "WestEurope"
    },
    "steps": [
        {
            "op": "same",
            "urn": "urn:pulumi:bot::bot::pulumi:pulumi:Stack::bot-bot",
            "oldState": {
                "urn": "urn:pulumi:bot::bot::pulumi:pulumi:Stack::bot-bot",
                "custom": false,
                "type": "pulumi:pulumi:Stack"
            },
            "newState": {
                "urn": "urn:pulumi:bot::bot::pulumi:pulumi:Stack::bot-bot",
                "custom": false,
                "type": "pulumi:pulumi:Stack"
            },
            "detailedDiff": null
        }
    ],
    "diagnostics": [
        {
            "urn": "urn:pulumi:bot::bot::azure-nextgen:alertsmanagement/latest:SmartDetectorAlertRule::Failure Anomalies - RachaelBot4wwyzn",
            "message": "error: Preview failed: refusing to delete protected resource 'urn:pulumi:bot::bot::azure-nextgen:alertsmanagement/latest:SmartDetectorAlertRule::Failure Anomalies - RachaelBot4wwyzn'\n",
            "severity": "error"
        },
        {
            "message": "error: preview failed\n",
            "severity": "error"
        }
    ],
[...]

PR #8810, opened 2022-01-21
Don't allow protected resources to be replaced

Frassle:
# Description

If a resource is protected we will no longer try to create a replacement for it. Previously the engine would of tried to create the replacement then issued a delete for the protected resource and then error out, that to-be-deleted-but-protected resource would then be left in the stack state and cause every `up` afterwards to also error.

Related to https://github.com/pulumi/pulumi/issues/5749

## Checklist

### DIFF
--- a/pkg/resource/deploy/step_generator.go
+++ b/pkg/resource/deploy/step_generator.go
@@ -605,10 +605,13 @@ func (sg *stepGenerator) generateStepsFromDiff(
 			// Note that we do allow unprotecting and replacing to happen in a single update
 			// cycle, we don't look at old.Protect here.
 			if goal.Protect {
-				return nil, result.Errorf("unable to replace resource %q\n"+
+				message := fmt.Sprintf("unable to replace resource %q\n"+
 					"as it is currently marked for protection. To unprotect the resource, "+
 					"remove the `protect` flag from the resource in your Pulumi "+
 					"program and run `pulumi up`", urn)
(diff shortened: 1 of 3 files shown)


pkg/resource/deploy/step_generator.go
```diff
@@ -601,6 +601,16 @@ func (sg *stepGenerator) generateStepsFromDiff(
...
 	if diff.Changes == plugin.DiffSome {
 		if diff.Replace() {
+			// If this resource is protected we can't replace it because that entails a delete
+			// Note that we do allow unprotecting and replacing to happen in a single update
+			// cycle, we don't look at old.Protect here.
+			if goal.Protect {
+				return nil, result.Errorf("unable to replace resource %q\n"+
+					"as it is currently marked for protection. To unprotect the resource, "+
+					"either remove the `protect` flag from the resource in your Pulumi"+
+					"program and run `pulumi up`", urn)
```
iwahbe, line 611: If we have `either`, we need to provide >1 options. Is this what you meant?
```suggestion
					"program and run `pulumi up` or edit the sate and remove the protect flag manually", urn)
```

iwahbe (approved):
1 nit, otherwise LGTM. (pending tests)

pkg/resource/deploy/step_generator.go
Frassle: Oh bad copy paste from the other protect error message. The only thing that makes sense here is changing the program (because we're look at Goal.Protect)

PR #8814, opened 2022-01-21
Checkout repo before hashing go.sum for cache keys

t0yv0:
# Description

Small fix to the Go Cache stanzas in test.yml/test-fast.yml - we need code checked out before hashing go.sum takes effect.

Fixes # (issue)

## Checklist


mikhailshilkov (approved):
LGTM

Issue #7963, opened 2021-09-14
gracefully handle 429 from pulumi service backend


EvanBoyle:
The pulumi service applied generous rate limits on various endpoints. Before we can turn those on we need to handle them gracefully the CLI

emiliza:
@pgavlin can you write up some of your concerns here please?

t0yv0:
@dixler

PR #8826, opened 2022-01-24
added rate-limit response handling

dixler:
# Description
Added handler for 429 rate limit response from pulumi service.

Fixes # (issue)
https://github.com/pulumi/pulumi/issues/7963
## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -23,6 +23,9 @@
 - [sdk/nodejs] - Support top-level default exports in ESM.
   [#8766](https://github.com/pulumi/pulumi/pull/8766)
 
+- [cli] Add better error message for pulumi service rate limit responses
+  [#7963](https://github.com/pulumi/pulumi/issues/7963)
+
 ### Bug Fixes
 
 - [sdk/python] - Prevent `ResourceOptions.merge` from promoting between the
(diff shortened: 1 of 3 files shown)


iwahbe:
Is the goal to just error out, or implement exponential backoff?
CC @EvanBoyle

Frassle:
I would of thought we'd do automatic backoff in response to this status code. Otherwise we're going to start hard failing a load of deployments that previously would of been fine.

mikhailshilkov:
@EvanBoyle Is there some kind of spec of the service behavior and expectations? Ideally, the service would return `retry-after` that we would respect here? I'd say we need to update #7963 with more details before we can ship this change.

EvanBoyle:
cc @jkisk who has been driving the service work to provide more detail via https://github.com/pulumi/pulumi-service/issues/6933

My understanding of WAF and our rate limiting is that thresholds are measured over five minute periods, meaning that exponential backoff isn't feasible. I'd expect this to be "just handle the error gracefully and return a clear message to the user about what has happened" but will let Jamie provide more detail here.

dixler:
replaced handling in title with error message for more clarity.

jkisk:
> I would of thought we'd do automatic backoff in response to this status code. Otherwise we're going to start hard failing a load of deployments that previously would of been fine.

Ok, I've chatted with Evan and Fraser and given that we are planning to initially set this limit at 5x or so our current highest usage, we don't expect this to be hit.  There is a cloudwatch alarm attached so the service on call will be aware if it is triggered, and this error message will give useful feedback to the user in that case.  There is more to do here when we start passing along 429s from third parties like Github, but at that point we can pass along a custom header to differentiate those errors.

Frassle (approved):

PR #8829, opened 2022-01-25
Allow disabling default providers

### LINKED ISSUE #3383: Provide opt-in setting to disable default providers
Default providers (configured using pulumi config, and used whenever an explicit provider is not passed either directly or indirectly) are very useful for getting started, but for robust deployments, it is often desirable to ensure that providers must be passed and configured explicitly.

While it is possible to do so today, it is still very easy to accidentally allow a resource to pick up the ambient/default provider.  

We have taken to encouraging configuring the default provider with invalid configration as a workaround to ensure the default provider cannot accidentally be used - but this only works if you remember to do it for every affected provider (and is an awkward workaround).

We should instead offer an opt-in setting (either an env var or a `pulumi:something` configuration variable) that turns off default providers, making it an error to attempt to register a resource using one.

This is related to a handful of other issues around default providers - where we may want to go even further - but this specific option is a good "lock things down" opt-in we can easily offer today in advance of larger change.

Related: 
* https://github.com/pulumi/pulumi/issues/2059 (allow setting the default provider programmatically)
* https://github.com/pulumi/pulumi/issues/2246 (expose the default provider instance to programs)
* https://github.com/pulumi/pulumi/issues/2917 (adapt API shape to more clearly encourage parenting

iwahbe:
This is done with an opt-in setting in `pulumi config`. For example, to
disable default providers for `aws`, use:

```sh
pulumi config set --path pulumi:disable-default-providers[0] aws
```

To add `kubernetes` to the disabled list, use
```sh
pulumi config set --path pulumi:disable-default-providers[1] kubernetes
```

To disable all default providers, `*` can be used.

---

Under the hood, whenever we handle a default provider request (with `defaultProviders.handleRequest`), we make sure it isn't on the deny list. If it is, we replace the requested reference with a special `DenyDefaultProvider` reference. We check for this reference whenever we are about to get a provider to do actual work. By intercepting denied providers when references are created, we ensure that we never use a denied provider by accident.

# Description

Fixes #3383

## Checklist

### DIFF
--- a/pkg/resource/deploy/providers/reference.go
+++ b/pkg/resource/deploy/providers/reference.go
@@ -96,6 +96,27 @@ func (r Reference) String() string {
 	return string(r.urn) + resource.URNNameDelimiter + string(r.id)
 }
 
+const denyDefaultProviderID resource.ID = "denydefaultprovider"
+
+// DenyDefaultProvider represent a default provider that cannot be created.
+func NewDenyDefaultProvider(name tokens.QName) Reference {
+	return mustNewReference(
+		resource.NewURN("denied", "denied", "denied", "pulumi:providers:denied", name),
+		denyDefaultProviderID)
+}
+
+// Retrieves the name of the denied provider.
+//
+// Panics if called on a provider that is not a DenyDefaultProvider.
+func DeniedDefaultProvider(ref Reference) string {
+	contract.Assert(IsDefaultProvider(ref))
+	return ref.URN().Name().String()
+}
+
+func IsDenyDefaultsProvider(ref Reference) bool {
+	return ref.id == denyDefaultProviderID
+}
+
 // NewReference creates a new reference for the given URN and ID.
 func NewReference(urn resource.URN, id resource.ID) (Reference, error) {
 	if err := validateURN(urn); err != nil {
--- a/pkg/resource/deploy/providers/reference.go
+++ b/pkg/resource/deploy/providers/reference.go
@@ -105,10 +105,14 @@ func NewDenyDefaultProvider(name tokens.QName) Reference {
 		denyDefaultProviderID)
 }
 
-// Retrieves the name of the denied provider.
+// Retrieves the package of the denied provider.
+//
+// For example, if a reference to:
+// "urn:pulumi:stack::project::pulumi:providers:aws::default_4_35_0"
+// was denied, then GetDeniedDefaultProviderPkg would return "aws".
 //
 // Panics if called on a provider that is not a DenyDefaultProvider.
-func GetDeniedDefaultProviderName(ref Reference) string {
+func GetDeniedDefaultProviderPkg(ref Reference) string {
 	contract.Assert(IsDenyDefaultsProvider(ref))
 	return ref.URN().Name().String()
 }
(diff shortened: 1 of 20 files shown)


pkg/resource/deploy/source_eval.go
```diff
@@ -386,6 +396,42 @@ func (d *defaultProviders) handleRequest(req providers.ProviderRequest) (provide
...
+	pConfig, err := d.config.GetPackageConfig("pulumi")
+	if err != nil {
+		return true, err
+	}
+
+	denyCreation := false
+	if value, ok := pConfig["disable-default-providers"]; ok {
+		array := []interface{}{}
+		if !value.IsString() {
+			return true, fmt.Errorf("Unexpected endecoding of pulumi:disable-default-providers")
```
Frassle, line 411:
```suggestion
			return true, fmt.Errorf("Unexpected encoding of pulumi:disable-default-providers")
```

pkg/resource/deploy/step.go
```diff
@@ -1175,6 +1175,10 @@ func getProvider(s Step) (plugin.Provider, error) {
 	if err != nil {
 		return nil, fmt.Errorf("bad provider reference '%v' for resource %v: %v", s.Provider(), s.URN(), err)
 	}
+	if providers.IsDenyDefaultsProvider(ref) {
+		name := providers.DeniedDefaultProvider(ref)
+		return nil, fmt.Errorf("Default provider '%v' not allowed for resource '%s'", name, s.URN())
```
Frassle, line 1180:
```suggestion
		return nil, fmt.Errorf(diag.GetDefaultProviderDenied(s.URN()), name, s.URN())
```

pkg/resource/deploy/providers/reference.go
```diff
@@ -96,6 +96,27 @@ func (r Reference) String() string {
...
+func NewDenyDefaultProvider(name tokens.QName) Reference {
+	return mustNewReference(
+		resource.NewURN("denied", "denied", "denied", "pulumi:providers:denied", name),
+		denyDefaultProviderID)
+}
+
+// Retrieves the name of the denied provider.
+//
+// Panics if called on a provider that is not a DenyDefaultProvider.
+func DeniedDefaultProvider(ref Reference) string {
```
Frassle, line 111:
```suggestion
func GetDeniedDefaultProviderName(ref Reference) string {
```
I wonder if this should just be `GetProviderName` and allow any provider ref? Or maybe just use ref.URN().Name() inline?

pkg/resource/deploy/source_eval.go
```diff
@@ -386,6 +396,42 @@ func (d *defaultProviders) handleRequest(req providers.ProviderRequest) (provide
...
+		}
+		if err := json.Unmarshal([]byte(value.StringValue()), &array); err != nil {
```
Frassle, line 413: Our config is YAML why don't we allow proper array options? String will be fine for now but it would be a nice improvement to config.

Frassle (approved):

pkg/resource/deploy/source_eval.go
iwahbe: The config (`Pulumi.*.yaml`) actually has a proper array. We serialize it to json as some earlier step. I'm not sure why it isn't kept an array.

pkg/resource/deploy/providers/reference.go
iwahbe: It really should be (and now is) called `GetDeniedDefaultProviderPkg`. I'm just storing the desired property in the `ref.URN().Name()` field. I clarified the intended behavior in comments.

PR #8825, opened 2022-01-24
ci: Run windows platform tests if and only if workflow run with windows-ci: true

AaronFriel:
The `if` condition added in `test` and `test-fast`, along with the additional input, should enable Mac & Linux tests to proceed and Windows tests will be filtered out.

A scheduled job is added which runs _just_ the Windows tests, using the full test suite.

Related:

- #8820

Closes: #8824

PR #8830, opened 2022-01-25
Revert "ci: Run windows platform tests if and only if workflow run with windows-ci: true"

AaronFriel:
Reverts pulumi/pulumi#8825


dixler (approved):
> the best kind of diff is a red one

LGTM

PR #8833, opened 2022-01-26
[codegen/go] Honor import aliases for external types/resources

justinvp:
# Problem Summary

Referencing external types/resources from Google Native doesn't work correctly in Go. For example, referencing the `AuditConfig` type leads to:

image (3)

Note the import at the top is `iam` but the usage is incorrectly `iamv1`. Attempting to workaround with a local import alias doesn't help.

This is happening because Google Native defines its own package import aliases, but they aren't taken into account fully when emitting references to the external types/resources. And any locally defined import aliases are also not taken into account.

# Fix (this change)

This change fixes the Go SDK codegen to honor package import aliases when referencing external types/resources. Local aliases are preferred, otherwise, if the external package has an import alias, it is used.


justinvp:
Oops, I forgot to add the new generated test file for the other languages. Will add in the morning.

pkg/codegen/go/gen.go
```diff
@@ -2493,6 +2508,21 @@ func (pkg *pkgContext) getTypeImports(t schema.Type, recurse bool, importsAndAli
...
 				goInfo.ImportBasePath = extractImportBasePath(extPkg)
 			}
+
+			// Ensure that any package import aliases we have specified locally take precedence over those
+			// specified in the remote package.
+			if ourPkgGoInfoI, has := pkg.pkg.Language["go"]; has {
+				ourPkgGoInfo := ourPkgGoInfoI.(GoPackageInfo)
+
+				for k, v := range ourPkgGoInfo.PackageImportAliases {
+					if goInfo.PackageImportAliases == nil {
```
Frassle, line 2518: Does it matter if this is set to non-nil map if `ourPkgGoInfo.PackageImportAliases` is empty? It's odd its in the loop rather than just before and I'm not sure if that's because its important to be nil or not?

justinvp: It's not import to be nil or not -- it's just to avoid unnecessarily creating the map. I've moved this outside the loop.

iwahbe (approved):
Besides one question, LGTM

pkg/codegen/internal/test/testdata/external-resource-schema/schema.json
```diff
@@ -148,18 +182,23 @@
...
         "Pulumi.Kubernetes": "3.7.0",
         "Pulumi.Random": "4.2.0"
       }
     },
     "go": {
       "generateResourceContainerTypes": true,
-      "generateExtraInputTypes": true
+      "generateExtraInputTypes": true,
+      "packageImportAliases": {
+        "github.com/pulumi/pulumi-google-native/sdk/go/google/iam/v1": "gcpiamv1"
```
mikhailshilkov, line 194: I'm still learning our approach to codegen tests... Would it make sense to adjust one of the existing resources to test this alias? Or do intend to have a separate test case for any feature?

pkg/codegen/internal/test/testdata/external-resource-schema/python/pulumi_example/foo.py
```diff
@@ -0,0 +1,169 @@
+# coding=utf-8
```
mikhailshilkov, line 1: As far as I understand, we add `Foo` to test Go aliases. Can we avoid generating it for all other languages, as those seem to be simply noise?

pkg/codegen/go/gen.go
```diff
@@ -206,9 +206,16 @@ func (pkg *pkgContext) tokenToType(tok string) string {
...
+
+	var importPath string
+	if alias, hasAlias := pkg.pkgImportAliases[path.Join(pkg.importBasePath, mod)]; hasAlias {
+		importPath = alias
+	} else {
+		importPath = strings.ReplaceAll(mod, "/", "")
+		importPath = strings.ReplaceAll(importPath, "-", "")
+	}
+
+	return strings.ReplaceAll(importPath+"."+name, "-provider", "")
```
mikhailshilkov, line 218: It would be great to add some unit tests for these two functions while you are on it

pkg/codegen/go/gen.go
```diff
@@ -275,7 +282,15 @@ func (pkg *pkgContext) tokenToResource(tok string) string {
 	if mod == "" {
 		mod = components[0]
 	}
-	return strings.Replace(mod, "/", "", -1) + "." + name
+
+	var importPath string
+	if alias, hasAlias := pkg.pkgImportAliases[path.Join(pkg.importBasePath, mod)]; hasAlias {
+		importPath = alias
+	} else {
+		importPath = strings.ReplaceAll(mod, "/", "")
```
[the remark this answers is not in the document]
justinvp: Looking back through the change history, it looks like you added the code that replaces `"-"` in `tokenToType` (i.e. lines 214-215) in #8190. Not sure why it wasn't added here in `tokenToResource` (or `tokenToEnum` for that matter). I'll see if it's needed in those cases.

pkg/codegen/go/gen.go
justinvp: Done

pkg/codegen/go/gen.go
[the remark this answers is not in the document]
justinvp: I started working on a fix for this, but ran into .NET test failures when updating the `dashed-import-schema` test to reference resources and enums. Unfortunately, I'll have to table this for now as out of scope. I've opened https://github.com/pulumi/pulumi/issues/8905 to track.

pkg/codegen/internal/test/testdata/external-resource-schema/python/pulumi_example/foo.py
justinvp: We can't avoid it for this test as it generates SDKs for all the languages. Instead of modifying this existing test, I reverted this test change, and added a new Go-only `external-go-import-aliases` test.

Issue #8796, opened 2022-01-20
Tracking Issue: Standardize `provider` and `providers` behaviour


iwahbe:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details
After an a design discussion ([internal link](https://docs.google.com/document/d/1Z0P-XU5AW-dhwKQCWHEiau_7kIAsF9j4CUFaPlMlGj4/edit#)), we have decided to implement the following:
1. `provider` indicates that a provider must be used. This takes priority over the ambient parent map or the providers map. `provider` is added to `providers` to be discoverable by children, but does not override an already existing provider for the same package. 
    a.  Do merge with warning if a conflict exists (between provider and providers). 
    b.  Warning should indicate that this merge will be depreciated in 6 months.
2.  `providers` indicates a set of providers that may be used, by both the resource and its children. We never promote a provider from `providers` to `provider`. 
3. Both `provider` and `providers` can be given. They operate independently as described above.
4. SDKs error if `provider` does not agree with the type of the resource being created. 
a. If python is passed a map of providers, warn if a key does not match its expected value (Provider.package). We warn instead of erroring to prevent breaking programs that have unused (and incorrect) provider maps. Warning should be converted to an error for the second release of July 2022. 
b. Depreciate accepting a Mapping for the python/go SDK. 

## TODO 
- [ ] Update each SDK
a. Update semantics in each SDK
b.  Update commends in docstrings across language SDKs to speak truth-y things
  - https://github.com/pulumi/pulumi/pull/8770
  - https://github.com/pulumi/pulumi/pull/8819
  - https://github.com/pulumi/pulumi/pull/8838
  - https://github.com/pulumi/pulumi/pull/8882
- [x] https://github.com/pulumi/pulumi-hugo/pull/997
- [ ] Consider updating developer docs with gory detail of how it plays out case by case
- [ ] If we’re inspired we can stand up a proposal for cleaning it up in a breaking way for 4.0 and add it to 4.0 wait list
a. Dropping all the maps and just having lists for “providers” is an obvious cleanup even if we don’t do anything else.

PR #8838, opened 2022-01-26
[sdk/dotnet] Normalize provider providers merge

iwahbe:
# Description

Part of https://github.com/pulumi/pulumi/issues/8796.

## Checklist


sdk/dotnet/Pulumi/Resources/ComponentResourceOptions.cs
```diff
@@ -55,13 +56,10 @@ public static ComponentResourceOptions Merge(ComponentResourceOptions? options1,
...
-            ExpandProviders(options1);
-            ExpandProviders(options2);
-
             // first, merge all the normal option values over
             MergeNormalOptions(options1, options2);
 
-            options1.Providers.AddRange(options2.Providers);
+            options1.Providers = MergeProviders(options1.Providers, options2.Providers);
 
             if (options1.Providers.Count == 1)
```
Frassle, line 64: Shouldn't this go as well? I thought we weren't promoting up anymore.

iwahbe: My bad requesting a review. I'm not done with this PR yet.

Frassle: I mean lines 64-68 in the if statement. It's setting `Provider` if `Providers.Count == 1`

Frassle: Also no worries, quick early reviews don't hurt :)

sdk/dotnet/Pulumi/Resources/ComponentResourceOptions.cs
```diff
@@ -48,37 +50,48 @@ internal override ResourceOptions Clone()
...
-            {
-                if (options.Provider != null)
+        }
+
+        private static List<ProviderResource> MergeProviders(List<ProviderResource> prov1, List<ProviderResource> prov2 )
+        {
+            var l = new List<ProviderResource>();
+            var taken = new HashSet<string>();
+            Action<List<ProviderResource>> add = (list) => {
+                for(var i = list.Count-1; i >= 0; i--)
```
dixler, line 78: Just a style question since I don't typically write dotnet, but why are you going through it in reverse and not using a foreach loop?

iwahbe: I'm not using a `foreach` loop because I want to iterate through the list in reverse. Every method (as in OOP) I found for backwards iteration involved creating a stack. I saw no reason to pay for that. 

If you're asking why I'm iterating in reverse, it's because I want to give priority to later elements in the input lists. This makes it simpler with the `taken` filter.

dixler (approved):
LGTM

Issue #8821, opened 2021-12-03
iam.GetPolicyDocument example is wrong - also how to create string pointer?


ghostsquad:
## Hello!

- Vote on this issue by adding a 👍 reaction
- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details

Example shows using a go string for the SID, however, the API only accepts a `*string`.

### Steps to reproduce

Copy/paste example into a main.go, go will complain.

Expected: 

The example at least compiles.

Actual: 

![image](https://user-images.githubusercontent.com/55453295/144683474-9c521875-a9b8-42d8-b827-ef2157d875e8.png)

First problem is that pointers to strings, ints, etc aren't a normal thing to do in go, and doing in a normal go-way results in harder to maintain code with extra variables hanging around that shouldn't really be needed.

If you are familiar with the AWS Go SDK (not the pulumi provider), it provides some utility functions to make this easier.

https://docs.aws.amazon.com/sdk-for-go/api/aws/

I didn't see any way to do this using the `pulumi` built-ins either. Even `pulumi.StringPtr("1")` returns a `StringPtrInput` type.

My work around is to import the AWS SDK or a package like: https://github.com/xorcare/pointer

t0yv0:
@AaronFriel

AaronFriel:
Hey @ghostsquad, as @leezen suggested, it is indeed frustrating to deal with pointers in Go. Since you referenced the AWS SDK, I wonder if we read the same StackOverflow post, I think I came across exactly that recommendation yesterday. 

I had to write these little helpers just the other day, and hopefully shows the pattern that's helpful:

```go
func StringRef(s string) *string { return &s }

func ProtocolRef(s v1.Protocol) *v1.Protocol { return &s }

func Int32Ref(s int32) *int32 { return &s }
```

@t0yv0 I wonder if Pulumi's Go SDK should also include conversion utilities like [AWS does](https://docs.aws.amazon.com/sdk-for-go/api/aws/)?

ghostsquad:
Ya, I mean it's easy enough to write my own small functions, or (heavy weight) import the AWS SDK.

At minimum, I wanted to bring to light that it was surprising that the docs were wrong, and although I know the solution, I wanted to help anyone else who ran into this type of problem, and was confused and/or wondering if there was a canonical/idiomatic way of solving this.

I am also requesting that the documents be updated to reflect the canonical/idiomatic way to write the code.

Thank you for your time!

[...]

PR #8839, opened 2022-01-26
[codegen/go] Improve optional params in invoke

AaronFriel:
As described in #8821, docs generated for helper functions can be incorrect because optional arguments to parameter objects are not correctly handled.

Previously, code would be generated like so:

```go
    policyDocument, err := iam.GetPolicyDocument(ctx, &iam.GetPolicyDocumentArgs{
      Statements: []iam.GetPolicyDocumentStatement{
        iam.GetPolicyDocumentStatement{
          Sid: "1",
          //...
```

However "Sid" is of type `*string`.

This four helper conversion functions, to handle the primitive types we lower from, and modifies codegen to recursively apply these functions inside of an invoke call.

In the new code generation, the above is instead rendered as:

```go
    policyDocument, err := iam.GetPolicyDocument(ctx, &iam.GetPolicyDocumentArgs{
      Statements: []iam.GetPolicyDocumentStatement{
        iam.GetPolicyDocumentStatement{
          Sid: pulumi.StringRef("1"),
          //...
```

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,5 +1,17 @@
 ### Improvements
 
+- [codegen/go] - Implement go type conversions for optional string, boolean, int, and float32
+  arguments, and changes our behavior for optional spilling from variable declaration hoisting to
+  instead rewrite as calls to these functions. Fixes #8821.
+  [#8839](https://github.com/pulumi/pulumi/pull/8839)
+
+- [sdk/go] Added new conversion functions for Read methods on resources exported at the top level
+  of the Pulumi sdk. They are `StringRef`, `BoolRef`, `IntRef`, and `Float64Ref`. They are used for
+  creating a pointer to the type they name, e.g.: StringRef takes `string` and returns `*string`.
+  Data source methods which take optional strings, bools, ints, and float64 values can be set to
+  the return value of these functions. These functions will appear in generated programs as well as
+  future docs updates.
+
 - [cli] Experimental support for update plans. Only enabled when PULUMI_EXPERIMENTAL is
   set. This enables preview to save a plan of what the engine expects to happen in a file
   with --save-plan. That plan can then be read in by up with --plan and is used to ensure
--- a/pkg/codegen/go/gen_program_optionals.go
+++ b/pkg/codegen/go/gen_program_optionals.go
@@ -1,8 +1,6 @@
 package gen
 
 import (
-	"fmt"
-
 	"github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/hclsyntax"
 	"github.com/pulumi/pulumi/pkg/v3/codegen"
@@ -30,48 +28,48 @@ func (ot *optionalTemp) SyntaxNode() hclsyntax.Node {
 }
 
 type optionalSpiller struct {
-	temps []*optionalTemp
-	count int
+	invocation         *model.FunctionCallExpression
+	intrinsicConvertTo *model.Type
 }
 
-func (os *optionalSpiller) spillExpressionHelper(
-	x model.Expression,
-	destType model.Type,
-	isInvoke bool,
-) (model.Expression, hcl.Diagnostics) {
-	var temp *optionalTemp
+func (os *optionalSpiller) preVisitor(x model.Expression) (model.Expression, hcl.Diagnostics) {
 	switch x := x.(type) {
 	case *model.FunctionCallExpression:
 		if x.Name == "invoke" {
 			// recurse into invoke args
 			isOutputInvoke, _, _ := pcl.RecognizeOutputVersionedInvoke(x)
 			// ignore output-versioned invokes as they do not need converting
-			isInvoke = !isOutputInvoke
-			_, diags := os.spillExpressionHelper(x.Args[1], x.Args[1].Type(), isInvoke)
-			return x, diags
+			if !isOutputInvoke {
+				os.invocation = x
+			}
+			os.intrinsicConvertTo = nil
+			return x, nil
 		}
 		if x.Name == pcl.IntrinsicConvert {
-			// propagate convert type
-			_, diags := os.spillExpressionHelper(x.Args[0], x.Signature.ReturnType, isInvoke)
-			return x, diags
+			if os.invocation != nil {
+				os.intrinsicConvertTo = &x.Signature.ReturnType
+			}
+			return x, nil
 		}
 	case *model.ObjectConsExpression:
-		// only rewrite invoke args (required to be prompt values in Go)
-		// pulumi.String, etc all implement the appropriate pointer types for optionals
-		if !isInvoke {
+		if os.invocation == nil {
 			return x, nil
+		}
+		destType := x.Type()
+		if os.intrinsicConvertTo != nil {
+			destType = *os.intrinsicConvertTo
 		}
 		if schemaType, ok := pcl.GetSchemaForType(destType); ok {
 			if schemaType, ok := schemaType.(*schema.ObjectType); ok {
-				var optionalPrimitives []string
+				// map of item name to optional type wrapper fn
+				optionalPrimitives := make(map[string]schema.Type)
 				for _, v := range schemaType.Properties {
-					isPrimitive := false
-					switch codegen.UnwrapType(v.Type) {
-					case schema.NumberType, schema.BoolType, schema.IntType, schema.StringType:
-						isPrimitive = true
-					}
-					if isPrimitive && !v.IsRequired() {
-						optionalPrimitives = append(optionalPrimitives, v.Name)
+					if !v.IsRequired() {
+						ty := codegen.UnwrapType(v.Type)
+						switch ty {
+						case schema.NumberType, schema.BoolType, schema.IntType, schema.StringType:
+							optionalPrimitives[v.Name] = ty
+						}
 					}
 				}
 				for i, item := range x.Items {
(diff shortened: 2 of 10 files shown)


AaronFriel:
@t0yv0 or @pgavlin would appreciate your insight on why this panics

t0yv0:
Let me return to this tomorrow

pkg/codegen/internal/test/testdata/aws-get-policy-document-pp/aws-get-policy-document.pp
```diff
@@ -0,0 +1,53 @@
...
+
+    resources = [
+      "arn:aws:s3:::${var.s3_bucket_name}/home/&{aws:username}",
+      "arn:aws:s3:::${var.s3_bucket_name}/home/&{aws:username}/*",
+    ]
+  }
+}
+
+
+resource "aws_iam_policy" "example" {
```
mikhailshilkov, line 49: I think it panics because "example" is not a valid resource type.
```suggestion
resource "aws_iam_policy" "aws:iam:Policy" {
```

pkg/codegen/internal/test/testdata/aws-get-policy-document-pp/aws-get-policy-document.pp
```diff
@@ -0,0 +1,53 @@
+data "aws_iam_policy_document" "example" {
```
mikhailshilkov, line 1: This likely needs to be an `invoke`, see https://github.com/pulumi/pulumi/blob/ae247f8f2e5cb627e1ba92e2e603daf166ef37bf/pkg/codegen/internal/test/testdata/aws-fargate-pp/aws-fargate.pp#L5 for example

t0yv0:
Apologies, slipping again, will look on Monday

mikhailshilkov:
@t0yv0 I think I responded to the initial question, I'm happy to help @AaronFriel here

pkg/codegen/internal/test/testdata/aws-get-policy-document-pp/aws-get-policy-document.pp
AaronFriel: Ah, for my understanding this is because the `.pp` files are in PCL?

mikhailshilkov: Yes, I think pp = pulumi program.

AaronFriel:
@t0yv0 I've rewritten the optional spiller to track state as it recurses into an invoke, then an intrinsic `__convert`, then an object constructor. I attempted to use `replace` clauses in `go.mod` to regenerate the AWS SDK to validate, and it appears that (unrelated to my change) I can't seem to build exactly what's in `pulumi-aws/sdk/go`. Even with a clean version of pulumi SDK, I get a lot of diffs.

That said, I did run it and verified by hand that positive changes were visible in doc comments to remove hoisted variables:

```go
    _, err := aws.GetRegions(ctx, &GetRegionsArgs{
      AllRegions: pulumi.BoolRef(true),
    }, nil)
```

```go
     _, err = lb.LookupListener(ctx, &lb.LookupListenerArgs{
       LoadBalancerArn: pulumi.StringRef(selected.Arn),
       Port:            pulumi.IntRef(443),
     }, nil)
```

This looks to be a positive change everywhere I can see, but again, having difficulty even locally rebuilding the AWS provider SDK reproducibly, so I'd be hesitant to merge this without being able to do that.

CHANGELOG_PENDING.md
```diff
@@ -1,5 +1,9 @@
 ### Improvements
 
+- [codegen/go] - Implement go type conversions for optional string, boolean, int, and float32
+  arguments, which should replace hoisted variable declarations in documentation. Fixes #8821.
+  [#8839](https://github.com/pulumi/pulumi/pull/8839)
```
mikhailshilkov, line 5: Nit: I would split this into two items: one - listing all new public functions (`Float64Ref` etc) and explaining how/when to use them; two - the program gen change (which is also user-facing because it affects imports).

pkg/codegen/go/gen_program_test.go
```diff
@@ -104,8 +104,8 @@ func goCheck(t *testing.T, path string, _ codegen.StringSet) {
 	err = integration.RunCommand(t, "point towards local Go SDK",
 		[]string{ex, "mod", "edit",
 			fmt.Sprintf("--replace=%s=%s",
-				"github.com/pulumi/pulumi/sdk/v3/go/pulumi",
-				"../../../../../../sdk")},
+				"github.com/pulumi/pulumi/sdk/v3",
+				"../../../../../../../sdk")},
```
mikhailshilkov, line 108: Why has this changed?

AaronFriel: I think the Go test were running via fluke before, resolving a different "github.com/pulumi/pulumi/sdk/v3" than the local copy. Until I changed this, tests were running using the published version of our SDK, I believe? I don't know Go module resolution well enough to say, but I do know that until I changed this, tests weren't running using my local changes.

Digging into the tests, they temporarily create a `go.mod` in a subdirectory like below, then run `go mod tidy`, then add a replacement for the Pulumi SDK. But I think that last step was a no-op.

```
testdata
├── aws-optionals-pp
│   ├── go
│   │   ├── aws-optionals.go 
│   │   └── go.mod # generated at test runtime
```

From the directory containing go.mod up to `sdk` is 7 levels:

```
testdata/aws-optionals-pp/go on  AaronFriel/issue8821 [$?⇕] via  v1.17.6 
❯ ls ../../../../../../../sdk
README.md  dotnet  go  go.mod  go.sum  nodejs  proto  python
```

CHANGELOG_PENDING.md
AaronFriel: Will add in commit tomorrow morning

AaronFriel:
Todo:
- [ ] Split changelog codegen & new Go SDK APIs into two parts.

t0yv0:
Is the change ready to review? I tried reviewing lightly - it's a tough one. I had to edit this pass once before but had to move with extreme care, I don't think this is easy to reason about. I'll try again though when this is ready.

AaronFriel:
@t0yv0 yes, this should be ready to review. The optional spiller is now a very small state machine. Until it enters into an appropriate invoke, it stays in the initial state. Once it enters an invoke, upon finding a __convert it will store the return signature for exactly 1 additional node (the argument to __convert), otherwise it will reset.

pkg/codegen/go/gen_program_optionals.go
```diff
@@ -30,91 +28,137 @@ func (ot *optionalTemp) SyntaxNode() hclsyntax.Node {
...
 	x model.Expression,
 	spiller *optionalSpiller,
 ) (model.Expression, []*optionalTemp, hcl.Diagnostics) {
-	spiller.temps = nil
-	x, diags := model.VisitExpression(x, spiller.spillExpression, nil)
-
-	return x, spiller.temps, diags
+	// We want to recurse but we only want to use the previsitor, if post visitor is nil we don't
+	// recurse.
+	x, diags := model.VisitExpression(x, spiller.preVisitor, spiller.postVisitor)
```
t0yv0, line 161: There is something really confusing here that changing postVisitor from nil changes the traversal semantics of VisitExpression.

What we had before:

```
  x, diags := model.VisitExpression(x, spiller.spillExpression, nil)
```

Was exactly equivalent to:

```
x, diags := spiller.spillExpression(x)
```

So there was no recursion whatsoever.

What we have now with post != nil, is that `model.VisitExpression` will visit every sub-expression.

Is this intentional?

pkg/codegen/internal/test/testdata/aws-fargate-pp/go/aws-fargate.go
```diff
@@ -12,9 +12,8 @@ import (
 
 func main() {
 	pulumi.Run(func(ctx *pulumi.Context) error {
-		opt0 := true
```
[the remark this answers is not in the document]
AaronFriel: Yeah, this will update existing codegen. Combining the two behaviors seemed unnecessarily complicated.

t0yv0 (approved):
Tentatively approving.. the expr to generate the new funcall heuristic looks good. Just the state machine formulation breaks my brain and I'm not sure it hits exactly the intended locations for the rewrite, which is also complicated by the fact that I'm not sure which intended locations we're trying to hit.

It seems that:

- we want to limit rewrite to expressions that are directly or indirectly contribute to constructing args for an Invoke (that is not an Output-form Invoke). The *indirectly* part is net new  in the PR, hence new recursion with postvisitor != nil that was not there before

- we want to find __convert(ObjectConsExpression(..), destTy) and rewrite them.

This seems to do that. The new example code illustrates the "indirectly" part - the `sid = "1"` is part of an array expr that contributes to statements property, not a direct property of the invoke args.

```
policyDocument = invoke("aws:iam:getPolicyDocument", {
  statements = [{
    sid = "1"

    actions = [
      "s3:ListAllMyBuckets",
      "s3:GetBucketLocation",
    ]

    resources = [
      "arn:aws:s3:::*",
    ]
  }]
})

resource example "aws:iam:Policy" {
  name   = "example_policy"
  path   = "/"
  policy = policyDocument.json
}
```

So while the previous code that used recursion was easier for me to read I bet it was not reaching deep enough into the tree to fix that expr up.

I tried rewriting in `t0yv0/8821` using F#-inspired pattern recognizers but that didn't pass this test case as the intention here is to rewrite deeply it seems; the state machine fits the problem better in the sense of what the VisitExpression interface coerces.

Now... What I'm left wondering is why the optional fixup is limited to invoke sub-expression only (that distinction seems a bit fragile). I was wondering the same before this PR and generally as I recall we can't trust the expression trees to be well-typed between passes, and the current pass sequence / obligations are not yet? documented.. So that's something we need to work on .. But approving the PR.

AaronFriel:
@t0yv0 removing your draft and I believe with this nit commit, all tests should be green again.

pkg/codegen/go/gen_program_optionals.go
AaronFriel: That's correct, without the recursion we don't get into the case in the test, we need to go a few levels deeper to get into the nested __convert

Issue #7160, opened 2021-05-27
Import leads to fields that cannot be set or are required but missing


lukehoban:
We have seen several cases where `pulumi import` leads to either `this field cannot be set` or `Missing required property` errors - for example:
- https://github.com/pulumi/pulumi/issues/6432
- https://github.com/pulumi/pulumi-pagerduty/issues/52
- https://github.com/pulumi/pulumi-azuredevops/issues/56
- https://github.com/pulumi/pulumi-gcp/issues/591
- https://github.com/pulumi/pulumi/issues/7884
- https://github.com/pulumi/pulumi/issues/7978
- https://github.com/pulumi/pulumi/issues/8326
- https://github.com/pulumi/pulumi/issues/8335

This is most likely a deeper issue with `pulumi import` that we will need to address.

cperreault:
Regarding #6432 (I am the author of it under a different professional account), I finally found out that it happened when the environment variable GOOGLE_CREDENTIALS is set. I was containing the private key of a GCP service account. When it is not set, I could do `pulumi refresh` without the App Engine Admin API error!

joeduffy:
@lukehoban @pgavlin Is https://github.com/pulumi/pulumi-aws/issues/1605 and instance of this too?

lukehoban:
> Is pulumi/pulumi-aws#1605 and instance of this too?

Yes - most likely.

0x1991babe:
One naive fix here allowed me to successfully import my resource: https://github.com/pulumi/pulumi-aws/issues/1605#issuecomment-910833221. Not sure if this is a generalizable / viable approach for other resources, but appears to be sufficient for my use case.

0x1991babe:
In case anyone needs a temp workaround, here's my hacky patch. No idea if this will work for other resources, but allowed me to successfully import the two I needed:

````patch
diff --git a/pkg/resource/deploy/step.go b/pkg/resource/deploy/step.go
index 927ee3d53..9d9d8eedf 100644
--- a/pkg/resource/deploy/step.go
+++ b/pkg/resource/deploy/step.go
@@ -931,11 +931,11 @@ func (s *ImportStep) Apply(preview bool) (resource.Status, StepCompleteFunc, err
                if !ok {
                        return resource.StatusOK, nil, errors.Errorf("unknown resource type '%v'", s.new.Type)
                }
+
+               props := cleanReadInputs(s.old.Inputs)
                for _, p := range r.InputProperties {
[...]

PR #8846, opened 2022-01-27
Add option to set property names to import with

Frassle:
# Description

Adds an option to the `import` command to allow the user to specify which properties to use the import with. By default the engine uses just required fields, but this often leads to import errors. 
For example trying to import and aws ec2 instance gives the following errors from Check:
```
error: aws:ec2/instance:Instance resource 'test' has a problem: Missing required argument: "instance_type": one of `instance_type,launch_template` must be specified. Examine values at 'Instance.InstanceType'.
error: aws:ec2/instance:Instance resource 'test' has a problem: Missing required argument: "launch_template": one of `ami,instance_type,launch_template` must be specified. Examine values at 'Instance.LaunchTemplate'.
error: aws:ec2/instance:Instance resource 'test' has a problem: Missing required argument: "ami": one of `ami,launch_template` must be specified. Examine values at 'Instance.Ami'.
```

With this new feature we can ask to import the ec2 instance using the 'ami' and 'instanceType' fields and it successfully imports.

Workaround for https://github.com/pulumi/pulumi/issues/7160

## Checklist


pkg/cmd/pulumi/import.go
```diff
@@ -291,8 +299,13 @@ func newImportCmd() *cobra.Command {
 	var protectResources bool
 
 	cmd := &cobra.Command{
-		Use:   "import [type] [name] [id]",
-		Args:  cmdutil.MaximumNArgs(3),
+		Use: "import [type] [name] [id] [properties...]",
```
iwahbe, line 302: I'm surprised that this is an argument, and not specified with a flag. 
1. Given the other flags, it makes more sense in my head to have it as a flag.
2. It locks in the `pulumi import` basic (flag-less) api until we break backwards compatibility. 
3. It renders a little funny in the help menu:
```
Usage:
  pulumi import [type] [name] [id] [properties...] [flags]
```

iwahbe (approved):
Besides one comment, LGTM

pkg/cmd/pulumi/import.go
Frassle: Yeh that's a fair comment. Looking at what the args library supports it looks like we could have:
`--properties propa,propb,propc` / `-p propa,propb -p propc`
or
`--property propa --property propb --property propc` / `-p propa -p propb -p propc`

I think `--properties a,b,c` is probably the nicer choice?

pkg/cmd/pulumi/import.go
```diff
@@ -505,30 +522,29 @@ func newImportCmd() *cobra.Command {
...
+					fmt.Printf(outputResult.String())
 				}
 			}
 
-			fmt.Printf(outputResult.String())
-
 			if res != nil {
+				fmt.Print("Import failed, try specifying the set of properties to import with.\n")
+				if importFilePath == "" {
+					fmt.Print("This can be done by passing the property names as extra argumeents " +
```
mikhailshilkov, line 541:
```suggestion
					fmt.Print("This can be done by passing the property names as extra arguments " +
```

pkg/cmd/pulumi/import.go
iwahbe: > I think --properties a,b,c is probably the nicer choice?

I would agree

pkg/cmd/pulumi/import.go
```diff
@@ -531,6 +531,9 @@ func newImportCmd() *cobra.Command {
 			}
 
 			if res != nil {
+				// TODO: This helpful message prints even in the case of the user doing a successful preview and then
+				// choosing "no" to not actually do the import. We need a way to distinguish Import _failing_ vs
+				// being canceled by user request.
 				fmt.Print("Import failed, try specifying the set of properties to import with.\n")
```
iwahbe, line 537: I might be missing this check, but are we confirming that the set of properties is not provided before we display this message.

mikhailshilkov: Good point. What happens if a specify a "wrong" set of properties?

Frassle: Oh good point! Maybe something as simple as:
`fmt.Print("Import failed, try specifying a different set of properties to import with.\n") `

mikhailshilkov: I guess it could be useful to differentiate our hint in case of no user-provided properties vs. with a wrong set of properties?

Issue #8848, opened 2022-01-27
Breaking change to `check` in `provider` type


leezen:
As a result of the change to `check` in https://github.com/pulumi/pulumi/pull/8631 it now expects to accept an additional parameter. We should make this parameter optional so as not to break existing code.

_Originally posted by @leezen in https://github.com/pulumi/pulumi-eks/issues/653#issuecomment-1023519350_

PR #8849, opened 2022-01-27
Remove sequence numbers from dynamic providers

Frassle:
# Description

Remove sequence numbers from the dynamic provider interfaces. 
It was a source breaking change adding these to the interface declarations. We don't need sequence numbers yet so this can go in as a 4.0 change.

Fixes https://github.com/pulumi/pulumi/issues/8848

## Checklist

### DIFF
--- a/sdk/nodejs/provider/provider.ts
+++ b/sdk/nodejs/provider/provider.ts
@@ -158,7 +158,7 @@ export interface Provider {
      * @param olds The old input properties to use for validation.
      * @param news The new input properties to use for validation.
      */
-    check?: (urn: resource.URN, olds: any, news: any, sequenceNumber: number) => Promise<CheckResult>;
+    check?: (urn: resource.URN, olds: any, news: any) => Promise<CheckResult>;
 
     /**
      * Diff checks what impacts a hypothetical update will have on the resource's properties.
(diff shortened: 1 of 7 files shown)


sdk/nodejs/provider/provider.ts
```diff
@@ -158,7 +158,7 @@ export interface Provider {
      * @param olds The old input properties to use for validation.
      * @param news The new input properties to use for validation.
      */
-    check?: (urn: resource.URN, olds: any, news: any, sequenceNumber: number) => Promise<CheckResult>;
+    check?: (urn: resource.URN, olds: any, news: any) => Promise<CheckResult>;
```
justinvp, line 161: This file is for full fledged providers, but what about the dynamic provider?

https://github.com/pulumi/pulumi/blob/43587cee92b821335a7b5fb996c12e708b477660/sdk/nodejs/dynamic/index.ts#L121

mikhailshilkov, line 161: Are `sequenceNumbers` strictly not needed here? Or should we keep the parameter but make it optional? (which also may avoid the breaking change vs. current master)

t0yv0 (approved):

PR #8861, opened 2022-01-27
Program codegen correctly handles foreign imports

### LINKED ISSUE #8822: [codegen/nodejs] Generate correct import paths for custom packages
The import path for a package named `frobs` is currently unconditionally generated as `"@pulumi/frobs"`, even if that is not the package's import path:

```typescript
import * as frobs from "@pulumi/frobs";
```

Some relevant code [here](https://github.com/pulumi/pulumi/blob/67ac4837ec85b441b0454e73924faa19368d3d72/pkg/codegen/nodejs/gen_program.go#L162-L167).

This should be updated to use the actual import path for the package.

iwahbe:
# Description

Fixes #8822 

## Checklist


pkg/codegen/internal/test/testdata/other.json
```diff
@@ -0,0 +1,35 @@
...
+      "packageName": "@third-party/other"
+    },
+    "csharp": {
+      "rootNamespace": "ThirdParty"
+    },
+    "python": {
+      "packageName": "third_party"
+    },
+    "go": {
+      "importBasePath": "gitplace.org/thirdparty"
```
AaronFriel, line 32: nit: I'm a fan of always making sure that urls in test data either resolve to something we own _or_ are guaranteed not to resolve.

Can we change this to `git.example.org`? It's one of the IANA "special use domain names" and is guaranteed not to have meaningful content. Alternatively, anything `.test` or `.invalid`.

AaronFriel (commented):
Left one nit, otherwise looks good.

AaronFriel (approved):
LGTM

PR #8863, opened 2022-01-28
Add Go cache setup to release.yml to match build.yml

t0yv0:
# Description

Theoretically this may be responsible for checksum mismatch we have seen on master.yml and release.yml - the checksums of tested binaries for darwin don't match the checksum of released binaries.

Fixes # (issue)

## Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -181,6 +181,20 @@ jobs:
           go-version: ${{ matrix.go-version }}
       - name: Checkout Repo
         uses: actions/checkout@v2
+      - id: go-cache-paths
+        run: |
+          echo "::set-output name=go-build::$(go env GOCACHE)"
+          echo "::set-output name=go-mod::$(go env GOMODCACHE)"
+      # NOTE: Go cache key setup should match build.yml for os: darwin
+      # to make sure binary checksums match (build=release).
+      - name: Go Cache
+        uses: actions/cache@v2
+        id: go-cache
+        with:
+          path: |
+              ${{ steps.go-cache-paths.outputs.go-build }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -181,6 +181,20 @@ jobs:
...
+      # NOTE: Go cache key setup should match build.yml for os: darwin
+      # to make sure binary checksums match (build=release).
+      - name: Go Cache
+        uses: actions/cache@v2
+        id: go-cache
+        with:
+          path: |
+              ${{ steps.go-cache-paths.outputs.go-build }}
+              ${{ steps.go-cache-paths.outputs.go-mod }}
+          key: darwin-go-cache-${{ hashFiles('**/go.sum') }}
```
Frassle, line 197: Why not just match the same as build.yml and have `${{ matrix.os }}-go-cache-${{ hashFiles('**/go.sum') }}` ?

t0yv0: matrix.os is not going to be defined in release.yml

t0yv0: Actually looking at this in the morning, this is all fishy... In build.yml we have three workers all running on darwin but two of them x-compiling to windows and linux - all with cache; in release.yml we have one worker doing all the x-compiling and darwin work in one shot w/o cache. Since it's 3 vs 1 worker we can't quite match  up the cache keys. The GHA interface is limited.

Maybe I can try this:

```
key: {random} # guarantee cache miss
restore-keys:
  - dawrin-*
  - linux-*
  - windows-*
```

Or I can disable caching in build.yml for release workflows.

In either case, I want to table this so it's not on our critical path; I want to run some experiments on the side to figure this out, and then stand up a PR again.

Issue #8871, opened 2022-01-28
Unable to unprotect and Replace resource without modifying the program to have protect as false


mitchellmaler:
## Issue details
Using the automation api we would unprotect resources (similar to running the unprotect command) and then run replace on resources. With the latest pulumi version now it says you must set the protect flag to false on the resource rather than just running the unprotect command (modify state of resource). This completely breaks the workflow of leaving the resource code alone now that we would have to mark each as false. Is there a way to force the replacement of a resource that has been unprotected but the program resource still have protect as true?

### Steps to reproduce

1. Deploy resource with protect flag set to true
2. Run unprotect command on resource URN
3. Run up with target replace on unprotected resource
4. Throws error saying we must remove protect flag.

Expected: Unprotected (in state) Resource is replaced and protected
Actual: Unprotected (in state) Resource is not allowed to be replaced and throws error

This broke our workflow so now will have to just remove the flag on all our resources and leave protection off.

mitchellmaler:
The error we are seeing now is "To unprotect the resource, remove the `protect` flag from the resource in your Pulumi program and run `pulumi up`" which didn't happen before we upgraded which seems like a breaking change of some sort.

iwahbe:
It looks like https://github.com/pulumi/pulumi/commit/67ac4837ec85b441b0454e73924faa19368d3d72 introduced the change.

mitchellmaler:
Does it make sense to roll this back back and/or make this change an opt-in/feature flag that way it doesn’t cause breaking changes to existing workflows like it has for us?

Frassle:
So replace has to look at either old.Protect or goal.Protect. I thought it made more sense to look at goal.Protect because that made it possible to remove the protect flag and change something else in the same update, but yes it does break this flow. 
If we look at just old.Protect then to do a replace you'd have to remove the flag in one up (or use the command as your doing), then make the change to your program as a separate up.

Wondering if maybe the best option is look at both:
| old.Protect | goal.Protect | Replace allowed
| ------------- | ------------- | ------------- |
| False | False  | Yes |
| False | True | Yes |
| True | False | Yes |
| True | True | No |

mitchellmaler:
That makes a lot of sense. That way it would allow a replace if the current/old resource is set to false but the new one to be created will be set to true.

Frassle:
PR raised to match the behavior in the table above

mikhailshilkov:
> Wondering if maybe the best option is look at both:

[...]

PR #8873, opened 2022-01-28
Allow replace if either old or new protect is false

Frassle:
# Description

Allow a resource to be replaced if either it's old or new state has the protect flag disabled. Importantly this still doesn't allow a resource to be just deleted in the same update operation that disables it's protect flag.

Fixes https://github.com/pulumi/pulumi/issues/8871

## Checklist


pkg/resource/deploy/step_generator.go
```diff
@@ -604,7 +604,7 @@ func (sg *stepGenerator) generateStepsFromDiff(
 			// If this resource is protected we can't replace it because that entails a delete
 			// Note that we do allow unprotecting and replacing to happen in a single update
 			// cycle, we don't look at old.Protect here.
-			if goal.Protect {
```
iwahbe, line 607: Why change between `goal.Protect` and `new.Protect`? Will they always agree?

Frassle: Yeh we set new.Protect from goal.Protect higher up in the step generator.

iwahbe (approved):
LGTM

t0yv0 (approved):

PR #8881, opened 2022-01-31
Respect schema versions (toggled via flag)

iwahbe:
# Description

Fixes https://github.com/pulumi/pulumi/issues/8521

## Checklist


pkg/codegen/dotnet/gen.go
```diff
@@ -2158,11 +2158,19 @@ func genPackageMetadata(pkg *schema.Package,
...
+
+	pulumiPlugin := &plugin.PulumiPluginJSON{
 		Resource: true,
 		Name:     pkg.Name,
 		Server:   pkg.PluginDownloadURL,
-	}).JSON()
+	}
+
+	if pkg.Version != nil && codegen.RespectVersion() {
+		files.add("version.txt", []byte(pkg.Version.String()))
```
Frassle, line 2169: What is version.txt supposed to be? If we're stamping a version to the codegen'd project should it be stamped into the project.csproj as <AssemblyVersion>?

pkg/codegen/utilities.go
```diff
@@ -162,3 +163,8 @@ func ExpandShortEnumName(name string) string {
 	}
 	return name
 }
+
+func RespectVersion() bool {
```
Frassle, line 167: 1. We don't normally do env looks ups from within the engine instead doing the lookup at the outer layer and passing the result in as an option. Do you think we should do the same for codegen?
2. Why does this need a flag anyway? It looks like we're going from the nonsense of always just setting 0.0.0 to setting a proper version? (N.B. I don't know the surrounding uses cases of this code so just let me know if I'm totally missing something here)

pkg/codegen/dotnet/gen.go
iwahbe: `version.txt` is the file we use to check the version of the package (at runtime). It makes some amount of sense to stamp the version into `*.csproj`.

pkg/codegen/utilities.go
iwahbe: 1. I'm not sure where to put the options. I think providers directly import `GeneratePackage`. Adding an additional option would be a breaking change (for both internal and external users of codegen). 
2. We need the flag because when we build our own providers, we don't know the version they are being built for. We include the git commit and the timestamp as part of the version. Ideally, we could set the default to true for the introduced flag.

pkg/codegen/internal/test/testdata/simple-resource-schema/nodejs/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@pulumi/example",
-    "version": "${VERSION}",
+    "version": "1.2.3",
```
mikhailshilkov, line 3: There were a number of attempts in the past to set this during codegen but every time we concluded we needed to do that in CI to avoid spurious diffs (e.g. when each PR bumps to a git-hash-based version, new every time). Has this changed?

iwahbe: This has not changed. That is why the feature is off by default. The feature is targeted at external (non-pulumi) customers of codegen.

mikhailshilkov: Thank you, that makes sense.

pkg/codegen/utilities.go
t0yv0: Codegen has typed option bags for example: 

```
// PackageInfo tracks Python-specific information associated with a package.
type PackageInfo struct {
	// PackageName is an override for the name of the generated python package.
	PackageName string `json:"packageName,omitempty"`
	// PythonRequires determines the Python versions that the generated provider supports
	PythonRequires string            `json:"pythonRequires,omitempty"`
	Requires       map[string]string `json:"requires,omitempty"`
	// Readme contains the text for the package's README.md files.
	Readme string `json:"readme,omitempty"`
	// Optional overrides for Pulumi module names
	//
	//    { "flowcontrol.apiserver.k8s.io/v1alpha1": "flowcontrol/v1alpha1" }
	//
	ModuleNameOverrides map[string]string `json:"moduleNameOverrides,omitempty"`
	// Toggle compatibility mode for a specified target.
	// Deprecated: This bool is no longer needed since all providers now use input/output classes.
	UsesIOClasses bool `json:"usesIOClasses,omitempty"`
	// Determines whether to make single-return-value methods return an output object or the single value.
	LiftSingleValueMethodReturns bool `json:"liftSingleValueMethodReturns,omitempty"`
}

```

These options are typed per language so that's a bit annoying but it is does provide an avenue for bw-compat addition of options affecting codegen.

t0yv0:
Something going on with tests:

```
=== Failed
=== FAIL: codegen/dotnet TestGeneratePackage/simple-resource-schema (0.00s)
    sdk_driver.go:367: Simple schema with local resource properties
    sdk_driver.go:386: 
        	Error Trace:	sdk_driver.go:386
        	Error:      	Received unexpected error:
        	            	Failed to load file version.txt referenced in codegen-manifest.json: open ../internal/test/testdata/simple-resource-schema/dotnet/version.txt: no such file or directory
        	Test:       	TestGeneratePackage/simple-resource-schema
    --- FAIL: TestGeneratePackage/simple-resource-schema (0.00s)

=== FAIL: codegen/dotnet TestGeneratePackage (181.51s)
```

Perhaps the version.txt was not checked in?

t0yv0 (approved):
Left a few comments but generally this looks quite useful - LGTM!

pkg/codegen/utilities.go
iwahbe: But these bags are passed as part of the `Package` struct, and derived from the schema. I'm trying to avoid putting a `PULUMI_RESPECT_SCHEMA_VERSION` flag in the schema itself. I could create a non-serializing flag in the `Package` struct, but it seems non-optimal.

pkg/codegen/nodejs/gen.go
```diff
@@ -2143,6 +2143,10 @@ func genNPMPackageMetadata(pkg *schema.Package, info NodePackageInfo) string {
 		},
 	}
 
+	if pkg.Version != nil && info.RespectSchemaVersion {
+		npminfo.Version = pkg.Version.String()
```
justinvp, line 2147: Shouldn't we also specify the version passed to the install script on line 2135 rather than `${VERSION}`?

Also, possibly as a separate change, I'm wondering if we should update the install script to exit early if the version argument has either `${VERSION}` _or_ `0.0.0`. The first customer to use `respectSchemaVersion` will be setting it to `0.0.0` so that the generated SDK will have a valid version of 0.0.0 in `package.json` so that local npm install/link can be done using the directory containing the generated SDK (rather than having to do an additional build step and then do the local npm install/link from a separate bin directory).

pkg/codegen/dotnet/gen.go
```diff
@@ -2158,11 +2158,20 @@ func genPackageMetadata(pkg *schema.Package,
...
 		Resource: true,
 		Name:     pkg.Name,
 		Server:   pkg.PluginDownloadURL,
-	}).JSON()
+	}
+
+	lang, ok := pkg.Language["csharp"].(CSharpPackageInfo)
+	if pkg.Version != nil && ok && lang.RespectSchemaVersion {
+		files.add("version.txt", []byte(pkg.Version.String()))
+		pulumiPlugin.Version = pkg.Version.String()
```
justinvp, line 2171: Should we do the same for Go and Python's `pulumi-plugin.json` file if we're doing it for .NET?

justinvp (approved):
LGTM otherwise

stack72 (approved):
This is a great way to roll out new functionality - opt-in to start with. As this is off by default, I can't see any issues here!

PR #8882, opened 2022-01-31
Normalize providers merge

### LINKED ISSUE #8796: Tracking Issue: Standardize `provider` and `providers` behaviour

iwahbe:
# Description

This is the final SDK change.
Fixes #8796

## Checklist


sdk/go/pulumi/context.go
```diff
@@ -911,8 +925,8 @@ func applyTransformations(t, name string, props Input, resource Resource, opts [
 }
 
 // checks all possible sources of providers and merges them with preference given to the most specific
-func mergeProviders(t string, parent Resource, provider ProviderResource,
-	providers map[string]ProviderResource) map[string]ProviderResource {
+func (ctx *Context) mergeProviders(t string, parent Resource, provider ProviderResource,
```
t0yv0, line 928: Could a unit test hit up this change here or it's tedious?

t0yv0 (approved):
Change looking good, wondering how we could test it.

sdk/go/pulumi/context.go
iwahbe: It's a little tedious, but should be done: https://github.com/pulumi/pulumi/pull/8882/commits/16fcbdce4d920b86f4935c77649c9c2f3dcf8ee1.

PR #8883, opened 2022-01-31
Add wildcard for --target and family

### LINKED ISSUE #5870: pulumi up --target wildcard urn support
## Expected Behavior

Can pulumi support `pulumi up -t urn:pulumi:prod::beluga::beluga:aws:VPC::*`?

iwahbe:
# Description

Fixes #5870

## Checklist

### DIFF
--- a/pkg/cmd/pulumi/destroy.go
+++ b/pkg/cmd/pulumi/destroy.go
@@ -143,9 +143,13 @@ func newDestroyCmd() *cobra.Command {
 				return result.FromError(fmt.Errorf("getting stack configuration: %w", err))
 			}
 
+			snap, err := s.Snapshot(commandContext())
+			if err != nil {
+				return result.FromError(err)
+			}
 			targetUrns := []resource.URN{}
 			for _, t := range *targets {
-				targetUrns = append(targetUrns, resource.URN(t))
+				targetUrns = append(targetUrns, snap.GlobUrn(resource.URN(t))...)
 			}
 
 			refreshOption, err := getRefreshOption(proj, refresh)
--- a/pkg/cmd/pulumi/destroy.go
+++ b/pkg/cmd/pulumi/destroy.go
@@ -151,6 +151,12 @@ func newDestroyCmd() *cobra.Command {
 			for _, t := range *targets {
 				targetUrns = append(targetUrns, snap.GlobUrn(resource.URN(t))...)
 			}
+			if len(targetUrns) == 0 && len(*targets) > 0 {
+				if !jsonDisplay {
+					fmt.Printf("There were no resources matching the wildcards provided.\n")
+				}
+				return nil
+			}
 
 			refreshOption, err := getRefreshOption(proj, refresh)
 			if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/resource/deploy/snapshot.go
```diff
@@ -175,3 +178,45 @@ func (snap *Snapshot) VerifyIntegrity() error {
 
 	return nil
 }
+
+// Performs glob style expansion on urns that contain '*'. Each urn can be
+// expanded into 1-n actual urns, depending on what underlying resources exist
```
Frassle, line 183:
```suggestion
// expanded into 0-n actual urns, depending on what underlying resources exist
```
There might be no matches

pkg/cmd/pulumi/destroy.go
```diff
@@ -143,9 +143,13 @@ func newDestroyCmd() *cobra.Command {
 				return result.FromError(fmt.Errorf("getting stack configuration: %w", err))
 			}
 
+			snap, err := s.Snapshot(commandContext())
```
Frassle, line 146: Is there any performance concern here? Is calling Snapshot always fast? Does this cause us to fully read here and again when the engine starts? Does that matter?

iwahbe: I don't know if calling snapshot is fast, but it is cached. This change shouldn't incur a general performance cost. 
https://github.com/pulumi/pulumi/blob/eb975813f44c42cb6ea0eb02af33c339e181a95e/pkg/backend/httpstate/stack.go#L121-L133
https://github.com/pulumi/pulumi/blob/eb975813f44c42cb6ea0eb02af33c339e181a95e/pkg/backend/filestate/stack.go#L55

Frassle (approved):

pkg/resource/deploy/snapshot.go
```diff
@@ -175,3 +178,45 @@ func (snap *Snapshot) VerifyIntegrity() error {
 
 	return nil
 }
+
+// Performs glob style expansion on urns that contain '*'. Each urn can be
+// expanded into 0-n actual urns, depending on what underlying resources exist
+// in the snapshot. URNs are returned in sorted order. All returned urns are unique.
+func (snap *Snapshot) GlobUrn(urn resource.URN) []resource.URN {
+	if !strings.Contains(string(urn), "*") {
+		return []resource.URN{urn}
```
t0yv0, line 187: This case is bit weird. I think the weirdness comes from GlobUrn combining two distinct pieces of functionality - filtering `snap.Resources` and matching against a pattern. This clause returns `urn` even if it's not mentioned in `snap.Resources`. You would think that the return values `[]resource.URN` are always a subset so this may be a little surprising. I'm guessing this is exactly what you want at the call sites but it's not obvious in the func itself.

t0yv0 (approved):

pkg/resource/deploy/snapshot.go
iwahbe: It's also how a normal glob function works (for example, in sh). Expanding an argument with no wild cards returns that argument, even if the argument is not a valid path. 
```
$ echo *.txt
real_file.txt
$ echo fake.txt
fake.txt
```

PR #8884, opened 2022-01-31
Do not rebuild Go binaries in release.yml, use testsed builds

### LINKED ISSUE #8877: Checksum check failling in CI
## Hello!

## Issue details

Investigate and fix the issue with checksum check failing to validate that we're releasing byte-for-byte identical binaries in the release.yml step - identical to the ones tested in test.yml step. This check has been failing and now is made advisory. The current theory is that the differences in Go Cache setup cause the discrepancy. Suggested fix: https://github.com/pulumi/pulumi/pull/8884

See also: https://github.com/pulumi/pulumi/issues/8824 

### Steps to reproduce

1.
2.
3.
4.

Expected: 
Actual:

t0yv0:
# Description

With these changes:

- release.yml no longer rebuilds Go binaries, but releases pre-built tested binaries; it only rebuilds the tar/zip archives

- checksum mismatch on the archives now blocks any side-effecting release actions, making sure we cannot enter "partially released" state again

- checksum mismatch is a hard failure again (used to be a warning) 

- the reproduced checksum mismatch seems to be fixed

Motivation: several master.yml builds and release.yml builds during last patch releases were lost to checksum mismatch, after which we made checkum mismatch check advisory.

A bit of background on what the check is about:

- originally we built Go binaries and put them in archives in build.yml and test them with test.yml

- due to peculiarities of Goreleaser, we rebuild the archives again in release.yml and push them out 

- the checksum check looks for these archives to be byte-for-byte identical

Implementation notes

This may be expressed more cleanly with Goreleaser pro feature "prebuilt", for now it wraps the Go toolchain to pretend to build binaries by locating the prebuilt ones.

Root cause

Still do not have full certainty as to the root cause of the mismatch, but it seems that build.yml uses Go cache and release.yml does not. It seems that go.sum go mirrored to a different set of actual binaries in the cache so the Go builds give different results depending on whether the cache is used or not. 

Testing

It is currently tricky to test changes to release.yml directly; this has been tested on a modified workflow in https://github.com/pulumi/pulumi/pull/8869

After we merge, `master.yml` builds will give additional feedback before the next release.

Fixes #8877

## Checklist

### DIFF
--- a/scripts/go-wrapper.sh
+++ b/scripts/go-wrapper.sh
@@ -1,8 +1,19 @@
 #!/usr/bin/env bash
 #
-# To be used as a replacement for the `go` toolchain. The wrapper
-# unifies building binaries normally and building them with coverage
-# support using `go test -c`.
+# To be used with Goreleaser as `gobinary` implementation as a
+# replacement for the `go` toolchain.
+#
+# First function: prebuilt binaries
+#
+# The wrapper detects and returns prebuilt binaries to skip actual go
+# builds. If a binary `-o goreleaser/../some-binary` is requested but
+# `goreleaer-prebuilt/../some-binary` already exists, the prebuilt
+# binary is copied instead of building.
+#
+# Second function: coverage-enabled builds for Pulumi CLI
+#
+# This builds binaries via `go test -c` workaround. Disabled for
+# Windows builds. Only enabled on the Pulumi CLI binaries.
 
 PULUMI_TEST_COVERAGE_PATH=$PULUMI_TEST_COVERAGE_PATH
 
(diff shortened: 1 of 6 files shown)


scripts/go-wrapper.sh
```diff
@@ -1,31 +1,77 @@
...
-            ;;
-        *)
-            go "$@"
-            ;;
-    esac
-fi
+COVERPKG="$PKG,$SDK"
+
+case "$1" in
+    build)
```
AaronFriel, line 27: yuck, this wrapping business. motivation makes sense but I hope you don't mind if I still go a little "eugh" at having to do this.

AaronFriel (commented):
@t0yv0 the motivating document checks out, and not much to review except that out matrix of [linux, mac, windows] and [ARM, x86-64] lines up, I'm unfamiliar with this go wrapping business but it makes sense to me as well.

Only nit would be that if there's an issue to review or clean up the CI process, we add this business to it and if not we create one. 

From experience, CI cleanup is always a $(currentQuarter+1) task, but I think @Frassle & you are on the right track thinking about overhauling it and this just adds to the evidence pile for making time for it.

AaronFriel (approved):

t0yv0:
The issue is https://github.com/pulumi/pulumi/issues/8877 but I should perhaps spilt it into two - this only addresses the second part. 

Point taken on `go` wrapping business - this is very unfortunate, but that's the only option I found on free tier of Goreleaser to make it reuse prebuilt binaries. I can file an issue to follow up with looking at the Goreleaser PRO option that may remove the need to wrap.

You are right that we could fix this later given that the workaround of 'ignoring checksums' currently seems to be okay. I just had this partially worked out on Friday when wrestling with the release and wanted to finish it off while it's fresh.

Issue #8887, opened 2022-02-01
Changes to surveycore seem to have broken stack selection for WSL


Frassle:
## Hello!

- Vote on this issue by adding a 👍 reaction
- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details

Building from master with https://github.com/pulumi/pulumi/pull/8858 included causes stack selection to error out when a stack is selected:
```
$ pulumi up
Please choose a stack, or create a new one:  [Use arrows to move, type to filter]
> dev
  <create a new stack>
```
Hit enter
```
$ pulumi up
Please choose a stack, or create a new one:
error: getting selected stack: [405]
```

### Steps to reproduce

Looks like this only happens on WSL. My terminal in vscode docker still works fine.

PR #8888, opened 2022-02-01
Workaround for surveycore bug with empty defaults

Frassle:
Workaround for https://github.com/AlecAivazis/survey/issues/342

Fixes https://github.com/pulumi/pulumi/issues/8887

### DIFF
--- a/pkg/cmd/pulumi/util.go
+++ b/pkg/cmd/pulumi/util.go
@@ -372,10 +372,15 @@ func chooseStack(
 
 	cmdutil.EndKeypadTransmitMode()
 
+	var defaultOption interface{}
+	if current != "" {
+		defaultOption = current
+	}
+
 	option, err := display.AskSelect(survey.Select{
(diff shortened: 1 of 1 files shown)


pkg/cmd/pulumi/util.go
```diff
@@ -372,10 +372,15 @@ func chooseStack(
...
+	var defaultOption interface{}
+	if current != "" {
+		defaultOption = current
+	}
+
 	option, err := display.AskSelect(survey.Select{
 		Message: message,
 		Options: options,
-		Default: current,
+		Default: defaultOption,
```
t0yv0, line 383: So the change substitutes nil for ""?

t0yv0 (approved):
I guess this one is hard to test - so LGTM - just read through the ticket. NOTE: we could also maybe fix it upstream in the survey lib itself.

dixler (approved):
Fixes #8887

PR #8891, opened 2022-02-01
Aaron friel/issue8890

AaronFriel:
# Description

Provider plugins that publish to a third party destination, and then used via NodeJS sdk don't respect the `pluginDownloadURL` destination for finding the package.

A bit of sleuthing and I found that the variable here was set & not used after assignment.

### DIFF
--- a/sdk/nodejs/runtime/invoke.ts
+++ b/sdk/nodejs/runtime/invoke.ts
@@ -248,13 +248,13 @@ export function call<T>(tok: string, props: Inputs, res?: Resource): Output<T> {
             // Construct a provider reference from the given provider, if one is available on the resource.
             let provider: string | undefined = undefined;
             let version: string | undefined = undefined;
-            let pluginDownloadUrl: string | undefined = undefined;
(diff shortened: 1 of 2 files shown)


sdk/nodejs/runtime/invoke.ts
```diff
@@ -248,13 +248,13 @@ export function call<T>(tok: string, props: Inputs, res?: Resource): Output<T> {
             // Construct a provider reference from the given provider, if one is available on the resource.
             let provider: string | undefined = undefined;
             let version: string | undefined = undefined;
-            let pluginDownloadUrl: string | undefined = undefined;
+            let pluginDownloadURL: string | undefined = undefined;
```
AaronFriel, line 251: I renamed this just to match casing. I use the VS Code extension to show inlay hints in function calls and if the argument passed doesn't contain the name of the argument, I see a little hint that something is off:

image

iwahbe (approved):
Good catch. LGTM

PR #8892, opened 2022-02-01
[plugins] Fix plugin reinstallation.

pgavlin:
Plugin installation has an early return path in the case that the plugin
was already installed. For normal plugin installation, this is okay and
saves some work. For reinstallation, however, this defeats the purpose
by causing installation to return before doing any actual work.

With these changes, we can better accommodate developer scenarios
without requiring that plugins are available on the PATH. For example,
the following script can be used to install a plugin under development:

    tar -c [plugin files] | gzip > plugin.tgz
    pulumi plugin install [kind] [name] [version] -f ./plugin.tgz --reinstall

### DIFF
--- a/sdk/go/common/workspace/plugins.go
+++ b/sdk/go/common/workspace/plugins.go
@@ -423,7 +423,7 @@ func (info PluginInfo) installLock() (unlock func(), err error) {
 // If a failure occurs during installation, the `.partial` file will remain, indicating the plugin wasn't fully
 // installed. The next time the plugin is installed, the old installation directory will be removed and replaced with
 // a fresh install.
-func (info PluginInfo) Install(tgz io.ReadCloser) error {
+func (info PluginInfo) Install(tgz io.ReadCloser, reinstall bool) error {
 	defer contract.IgnoreClose(tgz)
 
 	// Fetch the directory into which we will expand this tarball.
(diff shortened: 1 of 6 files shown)


pgavlin:
Working on a test case for this.

sdk/go/common/workspace/plugins.go
```diff
@@ -457,11 +457,14 @@ func (info PluginInfo) Install(tgz io.ReadCloser) error {
...
+			}
+			if !reinstall {
+				// finalDir exists, there's no partial file, and we're not reinstalling, so the plugin is already
+				// installed.
 				return nil
 			}
-			return partialFileStatErr
 		}
 
 		// The partial file exists, meaning a previous attempt at installing the plugin failed.
```
iwahbe, line 470: This is no longer accurate. It could also be a reinstall scenario.

iwahbe (approved):
Except one nit, LGTM

sdk/go/common/workspace/plugins.go
```diff
@@ -423,7 +423,7 @@ func (info PluginInfo) installLock() (unlock func(), err error) {
 // If a failure occurs during installation, the `.partial` file will remain, indicating the plugin wasn't fully
 // installed. The next time the plugin is installed, the old installation directory will be removed and replaced with
 // a fresh install.
-func (info PluginInfo) Install(tgz io.ReadCloser) error {
+func (info PluginInfo) Install(tgz io.ReadCloser, reinstall bool) error {
```
t0yv0, line 426: Just curious - this is not a change we'd consider breaking? Noting it's in sdk/go and it's a public method that changes signature, technically. Thinking it through though I can't imagine there's user code that calls it.

t0yv0 (approved):
It's tedious but we could consider testing it with "pulumi plugin install random --reinstall" and looking a timestamps of the plugin files.

sdk/go/common/workspace/plugins.go
Frassle: Last I asked @mikhailshilkov about this the comment was we should probably be keeping sdk/go/common compatible.

pgavlin: I can't imagine that there's user code that uses this either. Though this is technically breaking, I personally am not concerned about it. If we want to change it, I'd propose two exported methods: `Install` with its current signature and `Reinstall` (also with the same signature).

PR #8893, opened 2022-02-01
JavaScript fnSerialization match package.json exports

### LINKED ISSUE #8710: Pulumi generated lambda fails if dependency repository package.json defines "exports"
## Hello!

## Issue details

I'm trying to use [zod](https://github.com/colinhacks/zod) for lambda input validation. This packages defines [exports](https://github.com/colinhacks/zod/blob/master/package.json#L8-L13) in it's `package.json`.  This results in the generated lambda failing with this error:

```
{
  "errorType": "Error",
  "errorMessage": "Package subpath './lib/index.js' is not defined by \"exports\" in /var/task/node_modules/zod/package.json",
  "trace": [
    "Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './lib/index.js' is not defined by \"exports\" in /var/task/node_modules/zod/package.json",
    "    at new NodeError (internal/errors.js:322:7)",
    "    at throwExportsNotFound (internal/modules/esm/resolve.js:322:9)",
    "    at packageExportsResolve (internal/modules/esm/resolve.js:545:3)",
    "    at resolveExports (internal/modules/cjs/loader.js:450:36)",
    "    at Function.Module._findPath (internal/modules/cjs/loader.js:490:31)",
    "    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:888:27)",
    "    at Function.Module._load (internal/modules/cjs/loader.js:746:27)",
    "    at Module.require (internal/modules/cjs/loader.js:974:19)",
    "    at require (internal/modules/cjs/helpers.js:93:18)",
    "    at /var/task/__index.js:23:19"
  ]
}
```

### Steps to reproduce

here is a [repro-repo](https://github.com/PupoSDC/issue-pulumi-lambdas-fail-with-zod/blob/master/index.ts).

when i execute the lambda pulumi creates I get the error mentioned above. This error can be fixed by changing the resulting code.

Code produced by Pulumi:

```js
function __f2() {
  return (function() {
    with({ zod_1: require("zod/lib/index.js") }) {
````

Code that works:

```js
function __f2() {
  return (function() {
    with({ zod_1: require("zod") }) {
````

### Other info

Because of this [issue](https://github.com/pulumi/pulumi/issues/4130) I went ahead and tried this with the alpha version, but I get the sam result for both. 

```
pulumi about
CLI          
Version      3.21.1
Go Version   go1.17.5
Go Compiler  gc

Plugins
NAME    VERSION
aws     4.33.1
docker  3.1.0
nodejs  unknown

Host     
OS       darwin
Version  12.0.1
Arch     arm64

This project is written in nodejs (/Users/pedro/.nvm/versions/node/v16.10.0/bin/node v16.10.0)

NAME            VERSION
@pulumi/aws     4.33.1
@pulumi/awsx    0.30.0
@pulumi/pulumi  3.22.0-alpha.1641528849
zod             3.11.6
@types/node     14.18.5
```

### PR DESCRIPTION
**Pull request by dixler**
# Description
if a package.json is not found for a module path, return the original path.

if the package.json has no exports field, return the original path

if there's an exported module path tied to that module path, return that exported path or throw an exception if it isn't exported.

if exports is not an object or a string, return the original path. 

Fixes #8710

## Checklist

### CODE REVIEW

File: sdk/nodejs/tests/runtime/package.spec.ts [hunk 1]
```diff
@@ -0,0 +1,138 @@
+// Copyright 2016-2018, Pulumi Corporation.
```
ericrudder, line 1: (c) of 2022 throughout?

dixler:
TODO
- [ ] clean up TODO comments, reference names, add docstrings/better comments
- [ ] add more tests
  - [ ] null export tests
  - [ ] file extension exports
  - [ ] re-read [documentation](https://nodejs.org/api/packages.html#package-entry-points) to find extra corner cases
- [ ] update LICENSE years

CHANGELOG_PENDING.md
```diff
@@ -2,3 +2,5 @@
 
 ### Bug Fixes
 
+- [sdk/nodejs] - Fix nodejs function serialization module path complies with package.json exports if exports is specified.
```
iwahbe, line 5:
```suggestion
- [sdk/nodejs] - Fix nodejs function serialization module path to comply with package.json exports if exports is specified.
```

sdk/nodejs/runtime/closure/package.ts
```diff
@@ -0,0 +1,267 @@
...
+    srcSuffix: string;
+}];
+
+function makeRule(srcPattern: string, modPattern: string): Rule {
+    const srcSplit = srcPattern.split("*"); // NodeJS doesn't error out when provided multiple '*'.
+    const modSplit = modPattern.split("*");
+    if (srcSplit.length > 2 || modSplit.length > 2) {
+        // there is undefined behavior on more than 1 "*"
+        // see https://github.com/nodejs/node/blob/b191e66ddf/lib/internal/modules/esm/resolve.js#L664
+        throw new Error("multiple wildcards in single export target specification");
```
AaronFriel, line 125: Do these diagnostic messages go to the end user somehow, if so is there diagnostics about the particular import - I'm guessing that's either srcPattern or modPattern - so the user knows which module misbehaved?

AaronFriel (approved):
Approved with a comment/question, as I'd like to see how the error messages bubble up to users.

sdk/nodejs/runtime/closure/package.ts
dixler: I'm not sure if something catches these errors. I could extract the module name by splitting the as an argument to throw a better exception internally, but I've hit my time allowance.

the exports name will refuse to wildcard match if the original string has more than 1 asterisk.
it's a string replacement in the algorithm so if the target has multiple asterisks, each asterisk would be replaced with the match.

I think it's generally a never condition given how impractical structuring your files would be in a repository like this.

PR #8899, opened 2022-02-02
Fix nodejs type mismatch

iwahbe:
There was a mismatch between `@types/node` and `typescript` on their
default versions. This was fixed by listing both of them explicitly as a
`devDependency`.

# Description

Fixes # (issue)

## Checklist

### DIFF
--- a/pkg/codegen/nodejs/gen_program_test.go
+++ b/pkg/codegen/nodejs/gen_program_test.go
@@ -1,6 +1,7 @@
 package nodejs
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


pkg/codegen/nodejs/gen_program_test.go
```diff
@@ -29,41 +30,51 @@ func nodejsCheck(t *testing.T, path string, dependencies codegen.StringSet) {
...
+	packageJSONPath := filepath.Join(dir, "package.json")
+	if err := os.Remove(packageJSONPath); !os.IsNotExist(err) {
+		require.NoError(t, err, "Failed to delete %s", packageJSONPath)
 	}
 	yarnLock := filepath.Join(dir, "yarn.lock")
 	if err := os.Remove(yarnLock); !os.IsNotExist(err) {
 		require.NoError(t, err, "Failed to delete %s", yarnLock)
 	}
 
-	err = integration.RunCommand(t, "link @pulumi/pulumi",
```
Frassle, line 43: Not linking to the local pulumi anymore?

pkg/codegen/nodejs/gen_program_test.go
```diff
@@ -29,41 +30,51 @@ func nodejsCheck(t *testing.T, path string, dependencies codegen.StringSet) {
...
-	}
-	result := make([]string, len(deps))
-	for i, d := range deps.SortedValues() {
+func nodejsPackages(t *testing.T, deps codegen.StringSet) map[string]string {
+	result := make(map[string]string, len(deps))
+	for _, d := range deps.SortedValues() {
 		r := fmt.Sprintf("@pulumi/%s", d)
 		v := func(s string) {
-			r = fmt.Sprintf("%s@%s", r, s)
+			result[r] = "^" + s
```
mikhailshilkov, line 77: Could you add variable names or comments to explain what's going on here? `v` and `s` aren't that helpful to me.

pkg/codegen/nodejs/gen_program_test.go
```diff
@@ -29,41 +30,51 @@ func nodejsCheck(t *testing.T, path string, dependencies codegen.StringSet) {
...
-		err = integration.RunCommand(t, "yarn add and install",
-			[]string{ex, "add", pkg}, dir, &integration.ProgramTestOptions{})
-		require.NoError(t, err, "Could not install package: %q", pkg)
+	pkgInfo := npmPackage{
+		Dependencies: map[string]string{
+			"@pulumi/pulumi": "latest",
+		},
+		DevDependencies: map[string]string{
+			"@types/node": "latest",
+			"typescript":  "latest",
```
mikhailshilkov, line 50: Is it possible that those `latest` versions may produce flaky results over time?

pkg/codegen/nodejs/gen_program_test.go
```diff
@@ -29,41 +30,51 @@ func nodejsCheck(t *testing.T, path string, dependencies codegen.StringSet) {
...
 
-	err = integration.RunCommand(t, "link @pulumi/pulumi",
-		[]string{ex, "link", "@pulumi/pulumi"},
-		dir, &integration.ProgramTestOptions{})
-	require.NoError(t, err, "Failed to link @pulumi/pulumi")
-	for _, pkg := range pkgs {
-		err = integration.RunCommand(t, "yarn add and install",
-			[]string{ex, "add", pkg}, dir, &integration.ProgramTestOptions{})
-		require.NoError(t, err, "Could not install package: %q", pkg)
+	pkgInfo := npmPackage{
```
mikhailshilkov, line 44: I'm curious - why have you decided to change from yarn add commands to writing the package file explicitly?

iwahbe: I think that the new form is more readable. We deal mostly with go types, so the code is cleaner to me. When I wrote the previous version, I didn't know that `npmPackage` existed.

pkg/codegen/nodejs/gen_program_test.go
iwahbe: I can pin to the current version (latest at this moment). I don't think there will be meaningful flake from TS and node, but pinning will make sure.

mikhailshilkov (approved):
LGTM

PR #8902, opened 2022-02-02
Internal to public changelog

sefi-infralight:
# Description

Fixes # (issue)

## Checklist


Frassle:
I'm assuming these were raised by mistake and were only intended for your fork of pulumi.
If this was intended for here let us know as it's very unlikely we would accept this change as is but happy to talk through what you’re trying to accomplish.

PR #8909, opened 2022-02-02
Add a `schema extract` command

t0yv0:
# Description

Adding a command to extract schema from plugins. All the functionality is already in the codebase.. This seems useful in general, though my use-case is rather specific in trying to cross-build a provider extracting a schema from a Go resource provider and building a provider SDK in another language.

Fixes # (issue)

## Checklist

### DIFF
--- a/pkg/cmd/pulumi/schema.go
+++ b/pkg/cmd/pulumi/schema.go
@@ -31,5 +31,6 @@ package schemas for errors.`,
 	}
 
 	cmd.AddCommand(newSchemaCheckCommand())
+	cmd.AddCommand(newSchemaExtractCommand())
 	return cmd
 }
--- a/pkg/cmd/pulumi/schema_extract.go
+++ b/pkg/cmd/pulumi/schema_extract.go
@@ -16,6 +16,7 @@ package main
 
 import (
 	"encoding/json"
+	"io"
 	"os"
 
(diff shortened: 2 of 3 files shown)


Frassle (approved):
I think this is fine. Just to re-echo what I said on Slack.

I think it could be helpful to have a CLI binding to the resource provider plugin model to be able to test things like Read without worrying about the rest of the system. Given schema export is actually a provider command it would make sense for this to be in that set of commands. Giving something like:
```
$ pulumi provider aws 1.0 read <json input file> 
<json encoding of read result from ResourceProvider.Read>

$ pulumi provider aws 1.0 schema
<json encoding of schema returned from ResourceProvider.GetSchema>
```

t0yv0:
Okay let's log those as new issues that's a good idea..

Copying some more feedback I got on this PR:

1. Consider `pulumi plugin schema $pluginNameOrPath` vs `pulumi schema extract ..` - some pro/con on which "tree" the command belongs to

2. Either way, test that this can work with "component packages" as defined in https://www.pulumi.com/docs/guides/pulumi-packages/ - that is not published versioned providers but local providers under development, to debug them.

Frassle:
> This looks like it is obviated by pulumi package get-schema, is that right?

Yup

PR #8910, opened 2022-02-02
[codegen/python] Correctly ref external types/resources w/ same mod name

### LINKED ISSUE #8908: [codegen/python] Externally referenced types/resources aren't emitted correctly if they have the same module name
The following schema defines a component `example:cloudtrail:Trail` that references the `TrailAdvancedEventSelector` type and `Trail` resource from `aws.cloudtrail`:

```json
{
    "name": "example",
    "resources": {
        "example:cloudtrail:Trail": {
            "inputProperties": {
                "advancedEventSelectors": {
                    "type": "array",
                    "items": {
                        "$ref": "/aws/v4.37.1/schema.json#/types/aws:cloudtrail%2FTrailAdvancedEventSelector:TrailAdvancedEventSelector"
                    }
                },
                "trail": {
                    "$ref": "/aws/v4.37.1/schema.json#/resources/aws:cloudtrail%2Ftrail:Trail"
                }
            },
            "isComponent": true
        }
    },
    "language": {
        "python": {
            "requires": {
                "pulumi": "\u003e=3.0.0,\u003c4.0.0",
                "pulumi-aws": "\u003e=4.37.1,\u003c5.0.0"
            }
        }
    }
}
```

The emitted Python code looks like this:

```python
class TrailArgs:
    def __init__(__self__, *,
                 advanced_event_selectors: Optional[pulumi.Input[Sequence[pulumi.Input['pulumi_aws.TrailAdvancedEventSelectorArgs']]]] = None,
                 trail: Optional[pulumi.Input['pulumi_aws.Trail']] = None):
```

Note the missing `cloudtrail` in between `aws.` and the `.<Type|Resource>`:

- `pulumi_aws.TrailAdvancedEventSelectorArgs` should be `pulumi_aws.cloudtrail.TrailAdvancedEventSelectorArgs`
- `pulumi_aws.Trail` should be `pulumi_aws.cloudtrail.Trail`

justinvp:
This change fixes the Python codegen to correctly emit references to external types/resources that have the same module name as the referencing resource.

Fixes #8908


pkg/codegen/python/gen.go
```diff
@@ -165,9 +165,6 @@ func (mod *modContext) modNameAndName(pkg *schema.Package, t schema.Type, input
 	}
 
 	modName = tokenToModule(token, pkg, info.ModuleNameOverrides)
-	if modName == mod.mod {
-		modName = ""
-	}
```
justinvp, line 170: This looks like it was mistakenly copied from the non-external cases:

`objectType`:

https://github.com/pulumi/pulumi/blob/7772c3e846d028266d4cc056205d64dcca9fbd9d/pkg/codegen/python/gen.go#L217-L219

`tokenToResource`:

https://github.com/pulumi/pulumi/blob/7772c3e846d028266d4cc056205d64dcca9fbd9d/pkg/codegen/python/gen.go#L274-L276

pkg/codegen/internal/test/testdata/external-python-same-module-name/schema.json
```diff
@@ -0,0 +1,27 @@
...
+                }
+            },
+            "isComponent": true
+        }
+    },
+    "language": {
+        "python": {
+            "requires": {
+                "pulumi": "\u003e=3.0.0,\u003c4.0.0",
+                "pulumi-aws": "\u003e=4.37.1,\u003c5.0.0"
```
iwahbe, line 23: Can we type `<` instead of `\u003`?

iwahbe (approved):
LGTM

pkg/codegen/internal/test/testdata/external-python-same-module-name/schema.json
justinvp: Yep. This was based on the Go-generated awsx schema, which escapes characters for HTML by default. Fixed.

PR #8914, opened 2022-02-03
Enable delaying programtest cleanup to test Cleanup

AaronFriel:
# Description

Motivating example is something like (pseudocode):

```go
programTest(t, firstStackOptions)

programTest(t, secondStackOptions)
```

Where "secondStackOptions" will refer to a very basic stack that uses a stack reference to the first. This is a much cleaner example to put at the top level of a repo or in an "examples" folder:

```
repo/
└── examples/
    ├── stackreference-consumer/
    │   ├── Pulumi.yaml
    │   └── ...
    └── stackreference-producer/
        ├── Pulumi.yaml
        └── ...
```

The goal being to ensure all stacks in the examples dir are concise representative of features users would use in projects, without clutter or subdirectories that only the test harness cares about, while still being fully tested.

### DIFF
--- a/pkg/testing/integration/program.go
+++ b/pkg/testing/integration/program.go
@@ -1055,7 +1055,11 @@ func (pt *ProgramTester) TestLifeCycleInitAndDestroy() error {
 	}
 
 	pt.TestFinished = false
-	defer pt.TestCleanUp()
+	if pt.opts.DestroyOnCleanup {
+		pt.t.Cleanup(pt.TestCleanUp)
+	} else {
+		defer pt.TestCleanUp()
+	}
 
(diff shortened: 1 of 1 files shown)


iwahbe (approved):
I was worried about introducing implicit dependencies between tests, but `T.Cleanup` will run when a subtest ends, so the scoping is fine. Given that 
```go
for _, tt := range tests {
    t.Run(tt.Name, func(t *testing.T) {
        integration.RunTest()
    }
}
```
will observe the same behavior as before, is there a significant downside to making `DestroyOnCleanup` the default behavior? This will simplify the mental model for cleanup. You would need to specify `NoParralel` manually though (on multi-stack tests).

Either way, LGTM

PR #8920, opened 2022-02-03
[sdk/go] Fixed GoLang version parsing

### LINKED ISSUE #8907: Pulumi errors when attempting to use Golang 1.18beta2
## Hello!

## Issue details

Currently Pulumi is broken when attempting to use Golang 1.18beta2.

It throws the error `error: parsing go version: Invalid character(s) found in minor number "18beta2"` when you attempt to use Pulumi in any capacity.

```
❯ pulumi new --force
Please choose a template: aws-go
A minimal AWS Go Pulumi program
This command will walk you through creating a new Pulumi project.

Enter a value or leave blank to accept the (default), and press <ENTER>.
Press ^C at any time to quit.

project name: (pulumi-go-18-issue)
project description: (A minimal AWS Go Pulumi program)
Created project 'pulumi-go-18-issue'

Please enter your desired stack name.
To create a stack in an organization, use the format <org-name>/<stack-name> (e.g. `acmecorp/dev`).
stack name: (dev)
Created stack '[redacted]/dev'

aws:region: The AWS region to deploy into: (us-east-1) us-west-2
Saved config

Installing dependencies...

error: parsing go version: Invalid character(s) found in minor number "18beta2"
```

### Steps to reproduce

1. Install golang 1.18beta2
2. Install pulumi
3. Run `pulumi new`

Expected: It provisions a pulumi project
Actual: It fails to parse the version and errors.

cstclair-tunein:
# Description

Fixes #8907 

Currently Pulumi is parsing the GoLang version number is in semver format. However: the GoLang language version does not use Semver. For example, the golang version `1.18beta2` is a valid GoLang version identifier, but is not valid semver.

I have replaced the version parsing library with a more actively maintained one: [hashicorp/go-version](https://github.com/hashicorp/go-version). This library has the benefit of properly parsing the GoLang version discussed in issue #8907.

## Checklist


iwahbe:
I'm generally fine with this. @Frassle Do you think it is work ripping out the 664 times we call into `blang/semver` that are left? It seems kind of weird to maintain 2 different semver libraries.

mikhailshilkov:
> ripping out the 664 times we call into blang/semver

Are those semver or go-version? (apparently, these are not the same thing?)

Frassle:
go-version does say it's doing semver
> Versions used with go-version must follow [SemVer](http://semver.org/).

So if nothing breaks replacing all 664 instances then I don't see why not. One up-to-date library is better than two one of which hasn't been touched for nearly two years (which would be fine if it worked but it seemingly doesn't)

mikhailshilkov:
> go-version does say it's doing semver

How do I reconcile this with the statement below?

> the golang version 1.18beta2 is a valid GoLang version identifier, but is not valid semver.

iwahbe:
@mikhailshilkov `go-version` defaults to a relaxed parse that is more relaxed then `blang/semver`'s relaxed parse. It's still a semver library.

iwahbe:
@Frassle It looks like we expose `semver.Version` all over `sdk/go/common`. I think this is a 4.0 change. We can merge the localized change as is.

iwahbe (approved):

Frassle:
> @Frassle It looks like we expose semver.Version all over sdk/go/common. I think this is a 4.0 change. We can merge the localized change as is.

Can you add that to the 4.0 tracking issue.

cstclair-tunein:
> I'm generally fine with this. @Frassle Do you think it is work ripping out the 664 times we call into blang/semver that are left? It seems kind of weird to maintain 2 different semver libraries.

It may be easier to write a wrapper util that emulates the blang/semver api and replace the module imports globally.

PR #8922, opened 2022-02-03
Import uses all input properties and warns on check failure

Frassle:
# Description

This changes the `pulumi import` command to now import using _all_ the input fields returned from `provider.Read` instead of just the ones flagged as required. This should result in less Check failures as fields in such constraints such as "one of" that aren't represented by our schema get pulled in.

Further if there are Check failures they are now treated as a warning not an error. The engine will continue to import the resource with inputs given, update the stack state, and generate the code to match. We'll print an error to the user about the issue with instructions that they will need to edit the code before it can be successfully used.

Fixes https://github.com/pulumi/pulumi/issues/7160

## Checklist


pkg/resource/deploy/step.go
```diff
@@ -942,32 +932,48 @@ func (s *ImportStep) Apply(preview bool) (resource.Status, StepCompleteFunc, err
...
+
+		if issueCheckFailures(s.deployment.Diag().Warningf, s.new, s.new.URN, failures) {
+			ref, err := providers.ParseReference(s.Provider())
+			contract.Assert(err == nil)
+
+			pkgName := ref.URN().Type().Name()
+			s.deployment.Diag().Warningf(diag.Message(s.new.URN,
+				"One or more imported inputs failed to validate. "+
+					"This is almost certainly a bug in the `%s` provider. "+
+					"The import will still proceed, but you will need to edit the generated code after copying it into your program"),
```
pgavlin, line 963: cc @lukehoban @t0yv0 @mikhailshilkov for the wording here. Though validation failures are pretty much provider bugs by definition, there may be situations in which the provider cannot do any better, and filing an issue may be a waste of time. The rest of the message looks great to me.

mikhailshilkov:
> this changes the pulumi import command to now import using all the input fields returned from provider.Read instead of just the ones flagged as required

Are there any cases when something that used to work will fail now? Something like populating mutually-exclusive values even though they are all returned from Read?

pkg/resource/deploy/step.go
mikhailshilkov: Yeah I don't feel confident that the statement "This is almost certainly a bug in the `%s` provider" is true. Were all previous import errors bugs in providers? Or do we think this PR would fix them all?

pkg/resource/deploy/step.go
```diff
@@ -942,32 +932,48 @@ func (s *ImportStep) Apply(preview bool) (resource.Status, StepCompleteFunc, err
...
+			return rst, nil, err
+		}
+
+		if issueCheckFailures(s.deployment.Diag().Warningf, s.new, s.new.URN, failures) {
+			ref, err := providers.ParseReference(s.Provider())
+			contract.Assert(err == nil)
+
+			pkgName := ref.URN().Type().Name()
+			s.deployment.Diag().Warningf(diag.Message(s.new.URN,
+				"One or more imported inputs failed to validate. "+
```
mikhailshilkov, line 961: Do you have a full example of what a user sees in this case? Are they able to identify which "one or more inputs" are problematic?

pkg/cmd/pulumi/import.go
```diff
@@ -542,14 +542,6 @@ func newImportCmd() *cobra.Command {
 				// being canceled by user request.
 				if usedProperties {
 					fmt.Print("Import failed, try specifying a different set of properties to import with.\n")
-				} else {
-					fmt.Print("Import failed, try specifying the set of properties to import with.\n")
-					if importFilePath == "" {
-						fmt.Print("This can be done by passing the property names with the --properties flag.\n")
```
mikhailshilkov, line 548: Do we think the `properties` flag is still useful then?

mikhailshilkov (commented):
This feels like a fairly significant change. What is our plan to test it?

Frassle:
> Are there any cases when something that used to work will fail now? Something like populating mutually-exclusive values even though they are all returned from Read?

Everything that used to import to the state file will still import to the state file because pretty much anything should now bar actual errors from Read or Check (Check failures are separate from actual errors).

Codegen is worse for some things 😞 

AWS bucket with master:
```
const bucket_test = new aws.s3.Bucket("bucket-test", {
    acl: "private",
    bucket: "my-bucket-631a77b",
    forceDestroy: false,
}, {
    protect: true,
});
```
with this change:
```
const bucket_test = new aws.s3.Bucket("bucket-test", {
    accelerationStatus: "",
    arn: "arn:aws:s3:::my-bucket-631a77b",
    bucket: "my-bucket-631a77b",
    corsRules: [],
    grants: [],
    hostedZoneId: "Z2F56UZL2M1ACD",
    lifecycleRules: [],
    loggings: [],
    requestPayer: "BucketOwner",
    tags: {},
    versioning: {
        enabled: false,
        mfaDelete: false,
    },
}, {
    protect: true,
});
```
which warns:
`warning: aws:s3/bucket:Bucket resource 'bucket-test' has a problem: expected acceleration_status to be one of [Enabled Suspended], got . Examine values at 'Bucket.AccelerationStatus'.`

This is just terraform weirdness that it returns missing values as zero values from Read, which we'd like to fix. There's a question of if we want to make this code more complicated to try and work around that provider bug, or if we just want to go fix that provider bug.

pkg/resource/deploy/step.go
Frassle:
```
Diagnostics:
  aws:s3:Bucket (bucket-test):
    warning: aws:s3/bucket:Bucket resource 'bucket-test' has a problem: expected acceleration_status to be one of [Enabled Suspended], got . Examine values at 'Bucket.AccelerationStatus'.
    warning: One or more imported inputs failed to validate. This is almost certainly a bug in the `aws` provider. The import will still proceed, but you will need to edit the generated code after copying it into your program
```

pkg/cmd/pulumi/import.go
Frassle: Maybe if your importing lots of the same type of resource. If there are provider bugs with some fields you can work out what field set you need by importing the first resource and then tell all the rest to just use that field set.

It does occur to me we still need the diff set logic to support that correctly though, so if we want to keep --properties I'll need to add that back in.

mikhailshilkov:
> AWS bucket with master:
> ...
> with this change:
> ...

This looks bad? Would it be an option to start with required properties only, then fallback to this new option of all properties if the former failed? Or does it make everyone's life too complicated?

Frassle:
> This looks bad? Would it be an option to start with required properties only, then fallback to this new option of all properties if the former failed? Or does it make everyone's life too complicated?

We can try that. I'm playing around with it at the moment with aws buckets and the interaction or Read, Check, and Diff is _awkward_. The property "bucket" isn't required, but Check will fill it in based on old inputs. But if you don't use that Check result and pass an empty property array as inputs to Diff it won't report any diffs! Even though oldOutputs.bucket is set and newInputs.bucket isn't. But if we _do_ take the result from check then we end up setting all the default arguments like "acl" and "forceDestroy" that don't actually need to be in the user program because they'd get filled in by a normal update anyway.

pkg/resource/deploy/step.go
pgavlin: Not exactly. After these changes, though, it is definitively the responsibility of the provider to provide the most accurate set of inputs it can from `Read`, so any difference that e.g. causes a validation failure would be a problem in the provider.

Frassle:
Just testing import with the aws-native provider for buckets and it's read is a lot more sensible:
```
$ pulumi resource read aws-native 0.11.0 aws-native:s3:Bucket my-bucket-bb06540 -c region=us-west-1 -c version=4.36.0
{
    "id": "my-bucket-bb06540",
    "inputs": {
        "bucketName": "my-bucket-bb06540"
    },
    "outputs": {
        "__inputs": {
            "plaintext": "{\"bucketName\":\"my-bucket-bb06540\"}"
        },
        "arn": "arn:aws:s3:::my-bucket-bb06540",
        "bucketName": "my-bucket-bb06540",
        "domainName": "my-bucket-bb06540.s3.amazonaws.com",
        "dualStackDomainName": "my-bucket-bb06540.s3.dualstack.us-west-1.amazonaws.com",
        "regionalDomainName": "my-bucket-bb06540.s3.us-west-1.amazonaws.com",
        "websiteURL": "http://my-bucket-bb06540.s3-website-us-west-1.amazonaws.com"
    }
}
```
And so generate code looks better:
```
const test_bucket = new aws_native.s3.Bucket("test-bucket", {bucketName: "my-bucket-bb06540"}, {
    protect: true,
});
```

I'll test aws legacy with pats tfbridges changes now.

pgavlin:
Thanks for checking those out. Nice to see that AWS Native produces good results.

I'm surprised by the `null`s in the AWS Classic bucket inputs--I agree that we can probably clean those up with some tweaks.

pkg/resource/deploy/step.go
```diff
@@ -948,32 +938,51 @@ func (s *ImportStep) Apply(preview bool) (resource.Status, StepCompleteFunc, err
...
 		} else {
 			logging.V(9).Infof("Importing %v with supplied properties: %v", s.URN(), inputProperties)
-
 			for _, p := range inputProperties {
 				k := resource.PropertyKey(p)
 				if value, has := s.old.Inputs[k]; has {
 					s.new.Inputs[k] = value
 				}
 			}
 		}
```
pgavlin, line 951: Should we change the message below if these inputs fail `Check`? These are user-provided, so it seems that a validation failure is probably not a provider bug.

pgavlin (approved):
One question about validation failure messages with user-provided inputs, LGTM otherwise.

We should follow-up with some discussion on whether or not we want to continue to allow users to provide input property values.

pkg/resource/deploy/step.go
Frassle: Yes, I think that makes sense.

Frassle:
> We should follow-up with some discussion on whether or not we want to continue to allow users to provide input property values.

It's certainly not as critical as it was. But I think it might still have some use for when you know exactly the input properties you care about.

PR #8923, opened 2022-02-03
[sdk/nodejs] Pickup provider as part of providers

### LINKED ISSUE #8921: [sdk/nodejs] Provider specified in `provider` resource option not passed to engine for remote components
## Hello!

## Issue details

When creating a remote component (i.e. MLC) from TypeScript, and passing a provider via the singular `provider` resource option like this:

```typescript
const awsProvider = aws.Provider("myprovider");
const mycomponent = new RemoteComponent("mycomponent", {}, { provider: awsProvider });
```

The provider doesn't get passed along in the [`providers` map of the `RegisterResourceRequest`](https://github.com/pulumi/pulumi/blob/8e5e4caf6e405a380a64911f8a7da4ff1de80c24/sdk/proto/resource.proto#L107).

However, if specifying the provider like the following does work:

```typescript
const awsProvider = aws.Provider("myprovider");
const mycomponent = new RemoteComponent("mycomponent", {}, { providers: { aws: awsProvider } });
```

This is a regression introduced in #8819.

The problem is related to the following code for remote components:

https://github.com/pulumi/pulumi/blob/58cd76f8cc8452ce4df6b41f399d1072420f83ee/sdk/nodejs/runtime/resource.ts#L535-L547

Previously, `expandProviders` moved `provider` to `providers`:

https://github.com/pulumi/pulumi/blob/f99affcce41eaa68a6bf6dad83c0596b408b7c1d/sdk/nodejs/resource.ts#L995-L999

But #8819 deleted that: https://github.com/pulumi/pulumi/pull/8819/files#diff-2324dd338afb732d58bf71af19ec789fb841a80e03d9d659bd05cb411e7f30d3L996-L1000

So now `providers` is empty when only the `provider` option was specified, so it does not get passed along in the `providers` map of the `RegisterResourceRequest` here:

https://github.com/pulumi/pulumi/blob/58cd76f8cc8452ce4df6b41f399d1072420f83ee/sdk/nodejs/runtime/resource.ts#L338-L341

I haven't checked yet to see whether any other language is impacted.

iwahbe:
# Description

Fixes #8921

## Checklist

### DIFF
--- a/sdk/nodejs/runtime/resource.ts
+++ b/sdk/nodejs/runtime/resource.ts
@@ -536,6 +536,8 @@ async function prepareResource(label: string, res: Resource, custom: boolean, re
         if (remote) {
             const componentOpts = <ComponentResourceOptions>opts;
             expandProviders(componentOpts);
+            // the <ProviderResource[]> casts are safe because expandProviders
+            // /always/ leaves providers as an array.
             if (componentOpts.provider !== undefined) {
                 if (componentOpts.providers === undefined) {
                     // We still want to do the promotion, so we define providers
--- a/sdk/nodejs/runtime/resource.ts
+++ b/sdk/nodejs/runtime/resource.ts
@@ -537,11 +537,14 @@ async function prepareResource(label: string, res: Resource, custom: boolean, re
             const componentOpts = <ComponentResourceOptions>opts;
             expandProviders(componentOpts);
             if (componentOpts.provider !== undefined) {
-                if ((<ProviderResource[]> componentOpts.providers).indexOf(componentOpts.provider) !== -1) {
+                if (componentOpts.providers === undefined) {
+                    // We still want to do the promotion, so we define providers
+                    componentOpts.providers = [componentOpts.provider];
+                } else if ((<ProviderResource[]> componentOpts.providers)?.indexOf(componentOpts.provider) !== -1) {
                     const pkg = componentOpts.provider.getPackage();
                     const message = `There is a conflit between the 'provider' field (${pkg}) and a member of the 'providers' map'. `;
-                    const depreciation =  "This will become an error by the end of July 2022. See https://github.com/pulumi/pulumi/issues/8799 for more details";
-                    log.warn(message+depreciation);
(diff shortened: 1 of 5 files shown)


sdk/nodejs/runtime/resource.ts
```diff
@@ -536,6 +536,16 @@ async function prepareResource(label: string, res: Resource, custom: boolean, re
         if (remote) {
             const componentOpts = <ComponentResourceOptions>opts;
             expandProviders(componentOpts);
+            if (componentOpts.provider !== undefined) {
+                if ((<ProviderResource[]> componentOpts.providers).indexOf(componentOpts.provider) !== -1) {
+                    const pkg = componentOpts.provider.getPackage();
+                    const message = `There is a conflit between the 'provider' field (${pkg}) and a member of the 'providers' map'. `;
+                    const depreciation =  "This will become an error by the end of July 2022. See https://github.com/pulumi/pulumi/issues/8799 for more details";
+                    log.warn(message+depreciation);
```
justinvp, line 544: Nit:

```suggestion
                    const deprecation =  "This will become an error by the end of July 2022. See https://github.com/pulumi/pulumi/issues/8799 for more details";
                    log.warn(message+deprecation);
```

sdk/nodejs/runtime/resource.ts
```diff
@@ -536,6 +536,16 @@ async function prepareResource(label: string, res: Resource, custom: boolean, re
...
+                } else {
+                    (<ProviderResource[]> componentOpts.providers).push(componentOpts.provider);
```
justinvp, line 546: Is `componentOpts.providers` guaranteed to be truthy here? I think we need to guard against that.

```suggestion
                    (<ProviderResource[]> componentOpts.providers)?.push(componentOpts.provider);
```

justinvp (approved):
LGTM otherwise

sdk/nodejs/runtime/resource.ts
iwahbe: We do need to guard against it, but not by bailing out. In case of a `providers === undefined`, we should set `providers = [provider]`. https://github.com/pulumi/pulumi/pull/8923/commits/a27708fc208b12a4f3861813d011465cfd3c1cb1 makes that change.

PR #8927, opened 2022-02-04
Simpler x-plat testing in release.yml; lighten etc subset

t0yv0:
# Description

This compromise is aimed at making release flow more stable, currently affected by Windows OOM and requires lots of restarts. First we move some integration work out of `etc` subset - I suspect this should mitigate the Windows OOM of the `etc` worker. Then we use test-fast.yml in releases (assuming master.yml did the full test.yml already by that time). This means that only the `etc` worker will run on Mac + Windows platforms on releases (similar to what's happening on PRs). 

Fixes # (issue)

## Checklist

### DIFF
--- a/.github/workflows/test-fast.yml
+++ b/.github/workflows/test-fast.yml
@@ -75,10 +75,6 @@ jobs:
 
     runs-on: ${{ matrix.platform }}
 
-    # [WINDOWS-CI] - Find other instances of this string to find related snippets when this issue is
-    # resolved: https://github.com/pulumi/pulumi/issues/8820
-    continue-on-error: ${{ matrix.platform == 'windows-latest' }}
-
     steps:
       - name: Checkout Repo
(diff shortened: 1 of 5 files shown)


.github/workflows/test-fast.yml
```diff
@@ -75,10 +75,6 @@ jobs:
 
     runs-on: ${{ matrix.platform }}
 
-    # [WINDOWS-CI] - Find other instances of this string to find related snippets when this issue is
-    # resolved: https://github.com/pulumi/pulumi/issues/8820
-    continue-on-error: ${{ matrix.platform == 'windows-latest' }}
```
t0yv0, line 80: This is removed since it didn't work - somehow the "lost worker" condition is not the same as an error, and this clause has no effect on what it's trying to address.

iwahbe (approved):

PR #8928, opened 2022-02-04
export codegen tests for internal use

iwahbe:
# Description

Fixes # (issue)

## Checklist


mikhailshilkov (approved):

pkg/README.md
```diff
@@ -0,0 +1,4 @@
+# pulumi/pkg
+
+The files in `pulumi/pkg` make up the internals of the Pulumi CLI. Unlike `pulumi/sdk`, all files are considered "internal", and
+we maintain no backwards compatibility guaranty on this module.
```
t0yv0, line 4: We could frame this in a positive way.

> while pulumi/sdk maintains strict backwards compatibility guarantees, code under pkg/ is handled more informally, while we avoid making breaking changes unnecessarily they might happen when they make sense.

t0yv0 (approved):

t0yv0:
Before merging give me an hour or so to try it out in my repo, I will have feedback or confirmation that it works.

t0yv0:
So the first problem is that although funcs and structs are Public, they are in an `internal` package which Go toolchain does not let us cross-reference from other packages, as in:

```
	gen_test.go:10:2: use of internal package github.com/pulumi/pulumi/pkg/codegen/internal/test not allowed
FAIL	github.com/pulumi/pulumi-java/pkg/codegen/jvm [setup failed]
```

iwahbe:
> So the first problem is that although funcs and structs are Public, they are in an `internal` package which Go toolchain does not let us cross-reference from other packages, as in:
> 
> ```
> 	gen_test.go:10:2: use of internal package github.com/pulumi/pulumi/pkg/codegen/internal/test not allowed
> FAIL	github.com/pulumi/pulumi-java/pkg/codegen/jvm [setup failed]
> ```

I have renamed `codegen/internal` to `codegen/testing`.

PR #8944, opened 2022-02-08
Add support for GitHub private releases as plugin source

### LINKED ISSUE #8943: CLI: support for private GitHub releases in `--server`/`pluginDownloadURL`
## Hello!

## Issue details

Pulumi doesn't support authentication for downloading independently hosted plugins from private GitHub releases.

### Affected area/feature

Pulumi supports three ways of downloading plugins:
- `--server` flag to `pulumi plugin install` and a `pluginDownloadUrl` in a pakcage's schema.json that implicitly sets the `--server` flag for automatic plugin acquisition associated with that package from a specific URL, for independently hosted plugins.
- downloading from https://get.pulumi.com/releases/plugins for Pulumi hosted third party plugins.
- downloading from https://github.com/pulumi/pulumi/pulumi-*/releases, for first-party plugins.

### Proposal

Add support to pass GitHub info and credentials to download plugins from private GitHub releases.

patriziobruno:
# Description

This change adds support for downloading plugins from private GitHub releases. I added the change under the feature flag `PULUMI_EXPERIMENTAL`. It works as follows:
If the plugin can't be downloaded from Pulumi's public GitHub releases, an attempt is made to download from private GitHub releases using GitHub info and credentials passed as environment variables. To be compatible with GitHub Actions, env variable names come from [official documentation](https://docs.github.com/en/actions/learn-github-actions/contexts).
- GITHUB_REPO_OWNER
- GITHUB_TOKEN 
- GITHUB_ACTOR
- GITHUB_PERSONAL_ACCESS_TOKEN (this is not an official pipeline context var)
The implementation uses either `GITHUB_TOKEN` for service2service authentication or `GITHUB_ACTOR` + `GITHUB_PERSONAL_ACCESS_TOKEN` for HTTP basic authentication. It gets the requested `releaseId` from [GitHub's Releases API](https://docs.github.com/en/rest/reference/releases#get-a-release-by-tag-name) and, if successful, downloads the release from the URL of the asset matching the expected file-name format `pulumi-<type>-<name>-v<version>-<operatingSystem>-<arch>.tar.gz`.

If this method fails, `pulumi plugin download` falls back on get.pulumi.com.

Before adding testing, I wanted to gather feedback from the maintainers on the implementation.

Fixes #8943

## Checklist


sdk/go/common/workspace/plugins.go
```diff
@@ -317,14 +318,148 @@ func (info PluginInfo) Download() (io.ReadCloser, int64, error) {
...
-		logging.V(1).Infof("cannot find plugin on github.com/pulumi/pulumi-%s/releases", info.Name)
 	}
 
 	return getPluginResponse(buildPulumiHostedPluginURL(info.Kind, info.Name, info.Version, opSy, arch))
 }
 
+func getPluginFromPrivateGitHubResponse(info PluginInfo, opSy, arch string) (io.ReadCloser, int64, error) {
+	assetName := fmt.Sprintf("pulumi-%s-%s-v%s-%s-%s.tar.gz",
+		info.Kind, info.Name, info.Version.String(), opSy, arch)
+	resp, length, err := getPluginPrivateGitHubReleaseAPIResponse(info)
```
Frassle, line 338: Why do you need to hit the releases endpoint? For pulumi plugins we expect them at "https://github.com/pulumi/pulumi-%s/releases/download/v%s/%s" and I'd assume private plugins could be the same but under a different org so "https://github.com/%s/pulumi-%s/releases/download/v%s/%s",

sdk/go/common/workspace/plugins.go
```diff
@@ -317,14 +318,148 @@ func (info PluginInfo) Download() (io.ReadCloser, int64, error) {
...
+	if resp.StatusCode < 200 || resp.StatusCode > 299 {
+		logging.V(9).Infof("%d HTTP error fetching plugin from %s", resp.StatusCode, releaseURL)
+		return nil, -1, errors.Errorf("%d HTTP error fetching plugin GitHub releases from %s", resp.StatusCode, releaseURL)
+	}
+
+	return resp.Body, resp.ContentLength, nil
+}
+
+func authenticatedGetPluginResponse(assetURL, tokenType, token, secret, username string) (io.ReadCloser, int64, error) {
+	logging.V(9).Infof("full plugin download url: %s", assetURL)
```
Frassle, line 431: The amount of general code duplication in this file is a bit sad but maybe that's something for us to cleanup after.

sdk/go/common/workspace/plugins.go
patriziobruno: That is the browser link and requires the client to be authenticated as a browser, which is not documented. The documented way of downloading from private releases is to use GitHub's releases `API` and using Basic or OAuth authentication.

patriziobruno:
@Frassle I was wondering if it made sense to put this code under a different feature flag from `PULUMI_EXPERIMENTAL`. Maybe, `PULUMI_EXPERIMENTAL_GH_PRIVATE_REL`? To let users decide which experiments they want to use and which not?

sdk/go/common/workspace/plugins.go
patriziobruno: Following up on this conversation, I did some refactoring to reduce duplication.

iwahbe (approved):
This looks good to me. LGTM

patriziobruno:
Are there any additional steps before merging this?

Frassle:
Spoken to @stack72 about this. We do want to make sure we can support your workflow @patriziobruno, but further thinking about this we're not sure the proposed download flow is one we should support. 

Would it be acceptable to use the pluginDownloadURL to point to your private github release instead of the download flow trying to work that out from env[GITHUB_REPOSITORY_OWNER]. We'd then change the pluginDownloadURL using code to add github auth tokens (if and only if the url was for something at github.com) so it could auth to that custom url.

I'm happy to do the required code changes here if that would still work for your usecase. If that _wouldn't_ work do share why, we'll think some more about what we can do.

patriziobruno:
@Frassle @stack72 the problem about the `pluginDownloadURL` approach is that endpoints pointed by a `browser_download_url` don't accept GitHub Actions tokens, nor Basic authentication. Clients need to authenticate using a browser and the cookies that GitHub.com sends, which is not doable in a pipeline. Furthermore, once the user has authenticated, how would `pulumi` CLI access the authentication cookies? Unless GitHub.com adds OAuth tokens as a means to authenticate access to their `browser_download_url`, there's no way to download a private release without using the release API.

Frassle:
Ah right I've followed through the REST calls this makes and I see what you mean there. pluginDownloadURL makes assumptions about the format of the url, github releases don't match that format even if you pre-looked up the api style url to use.

@stack72 I can't see a simple way to fit this into our pluginDownloadURL flow given the above. I think we need to give that feature a _major_ rethink and for now accept the PR as is, with the expectation that we'll have to break all custom plugin flows at some point to tidy this up (4.0 or even 5.0 change).

mikhailshilkov:
> I think we need to give that feature a major rethink and for now accept the PR as is, with the expectation that we'll have to break all custom plugin flows at some point to tidy this up (4.0 or even 5.0 change).

This sounds somewhat scary to me... Will you leave this PR's flow behind the flag until 4.0/5.0? Otherwise that breaking change may be quite hard to pull in. It would be nice to understand what that future tidy-up may look like.

Frassle:
I'm going to merge this in with the understanding that it will stay under EXPERIMENTAL until we can spend the engineering time to rework custom plugin downloads to support this properly. Some time after adding proper support for this via the custom plugin work we will remove this code flow, and your workflow will be broken if you haven't updated to use the supported custom plugin settings.

I have raised https://github.com/pulumi/pulumi/issues/9007 to track the re-design work.

Issue #8853, opened 2022-01-27
Can't disable default Kubernetes provider in 3.23.0 for yaml.ConfigFile


itay:
## Hello!

- Vote on this issue by adding a 👍 reaction
- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details

I am trying out the new Pulumi 3.23.0 capability of disabling the default providers, and have this in my stack:

```
  pulumi:disable-default-providers:
  - aws
  - kubernetes
```

This seems to work well for AWS, but I am having an odd issue with the Kubernetes one. Specifically, I have the following resource:

```
    _, err = yaml.NewConfigFile(ctx, "certmanager-deploy-file", &yaml.ConfigFileArgs{
        File: "./cert-manager.yaml",
        Transformations: []yaml.Transformation{
            // We need to make two modifications:
            // 1. Add the role ARN for IRSA
            // 2. Set the fsGroup for IRSA token mapping
            // Docs here: https://cert-manager.io/docs/configuration/acme/dns01/route53/#eks-iam-role-for-service-accounts-irsa
            func(state map[string]interface{}, opts ...pulumi.ResourceOption) {
                metadata := state["metadata"].(map[string]interface{})
                name := metadata["name"]
                if state["kind"] == "ServiceAccount" && name == "cert-manager" {
                    var annotations map[string]interface{}
                    if v, ok := metadata["annotations"]; !ok {
                        annotations = make(map[string]interface{})
                        metadata["annotations"] = annotations
                    } else {
                        annotations = v.(map[string]interface{})
                    }
                    annotations["eks.amazonaws.com/role-arn"] = irsaRole.Arn
                }
                if state["kind"] == "Deployment" && name == "cert-manager" {
                    deploymentSpec := state["spec"].(map[string]interface{})
                    template := deploymentSpec["template"].(map[string]interface{})
                    podSpec := template["spec"].(map[string]interface{})
                    podSpec["securityContext"] = map[string]interface{}{
                        "fsGroup": 1001,
                    }
                }

            },
        },
    }, pulumi.DependsOn([]pulumi.Resource{irsaRole}), pulumi.Provider(eksConfig.Provider))
    if err != nil {
        return nil, err
    }
```
    
Where eksConfig.Provider is constructed as the result of an eks.Cluster creation:

```
    k8sProvider, err := providers.NewProvider(ctx, "k8s-ssa-provider", &providers.ProviderArgs{
[...]

PR #8978, opened 2022-02-11
Improve error handling for denied Invokes

iwahbe:
Explicitly catch invokes with denied default providers and give a
helpful error message.

# Description

Improves https://github.com/pulumi/pulumi/issues/8853

## Checklist


AaronFriel (approved):

pgavlin:
Do we have/can we add tests for this?

iwahbe:
@pgavlin I can't think of a way to test this that isn't a bit contrived.

PR #8979, opened 2022-02-11
Allow attaching to running providers

Frassle:
# Description

This adds an envvar check to attach to a currently running provider rather than spawning a new provider binary. This lets you start for example the aws provider in a debugger let it start listening in on a port and then running pulumi up with an env var set to use that running provider for aws `PULUMI_PROVIDER_AWS=<port>`

Engine side part of a fix for https://github.com/pulumi/pulumi/issues/2396

## Checklist


sdk/go/common/resource/plugin/provider.go
```diff
@@ -101,6 +101,10 @@ type Provider interface {
 	// non-blocking; it is up to the host to decide how long to wait after SignalCancellation is
 	// called before (e.g.) hard-closing any gRPC connection.
 	SignalCancellation() error
+
+	// Attach triggers an attach for a currently running provider to the engine
+	// TODO It would be nice if this was a HostClient rather than the string address but due to dependency ordering we don't have access to declare that here.
+	Attach(address string) error
```
pgavlin, line 107: Can this be an implementation detail of the gRPC implementation of plugins rather than part of the `Plugin` type?

Frassle: Maybe? I thought about maybe having a GrpcProvider that was just Provider + Attach? Is there a more "go" way to do it?

Frassle:
> I also wonder if the envvar might be more ergonomic if it we a list of (name, version?, port) tuples. Something like

Yeh happy to change to that style. I'm not sure versions matter here, I see it similar to how if you have like "pulumi-resource-aws" on path we just ignore the version and use the one you've got on path. Which I guess also answers your other question of what happens if you have multiple instances, well you don't. Same as if it was on path.

sdk/go/common/resource/plugin/provider.go
pgavlin: Yeah, I'm wondering whether or not we need to expose `Attach` at all. I think we could just invoke the gRPC method directly here: https://github.com/pulumi/pulumi/pull/8979/files#diff-4372cd77587542c3e09139fb236d0bce7d0ff31c3c81864950974b1e33e5104eR146

sdk/proto/go/provider.pb.go
```diff
@@ -2171,6 +2172,8 @@ type ResourceProviderClient interface {
 	Cancel(ctx context.Context, in *empty.Empty, opts ...grpc.CallOption) (*empty.Empty, error)
 	// GetPluginInfo returns generic information about this plugin, like its version.
 	GetPluginInfo(ctx context.Context, in *empty.Empty, opts ...grpc.CallOption) (*PluginInfo, error)
+	// Attach sends the engine address to an already running plugin.
+	Attach(ctx context.Context, in *PluginAttach, opts ...grpc.CallOption) (*empty.Empty, error)
```
Frassle, line 2176: This is going to break implementers of `ResourceProviderClient`. Which if we're going to have to push out a break change we may as well change NewProviderServer to take a GrpcProvider as well.

danielrbradley (approved):

pgavlin:
> One other thing: have you tried this when running an update where multiple instances of a particular provider are needed? I'm not sure how this fits into that scenario, as today each instance of a provider implies a new process.

Still curious about this.

Frassle:
> Still curious about this.

I'll have a look. I thought we cached processes per provider name, but I'll dig into this more.

Frassle:
So with attach it will redial the grpc connection each time and call Attach (with the same engine address). So it would also call Configure multiple times.
Not sure that's an issue, but also not sure what else we could do in this case?

PR #8981, opened 2022-02-11
Allow turning ts typechecking off

dmattia:
# Description

Adds an env var `PULUMI_NODEJS_TRANSPILE_ONLY`, defaulting to `false`, that controls if Pulumi will typecheck should perform any typechecking on its own. As this default is what the only allowed value was before this change, this should not break any existing workflows.

In our case, we prefer to run Pulumi separately from `tsc` to compile our code.

## Checklist

### DIFF
--- a/sdk/nodejs/cmd/run/run.ts
+++ b/sdk/nodejs/cmd/run/run.ts
@@ -161,7 +161,7 @@ export function run(
     const tsConfigPath: string = process.env["PULUMI_NODEJS_TSCONFIG_PATH"] ?? defaultTsConfigPath;
     const skipProject = !fs.existsSync(tsConfigPath);
     
-    const typeCheck = (process.env["PULUMI_NODEJS_TYPECHECK"] ?? "true") === "true";
+    const transpileOnly = (process.env["PULUMI_NODEJS_TRANSPILE_ONLY"] ?? "false") === "true";
 
     let compilerOptions: object;
     try {
--- a/sdk/nodejs/cmd/run/run.ts
+++ b/sdk/nodejs/cmd/run/run.ts
@@ -160,6 +160,8 @@ export function run(
     const defaultTsConfigPath = "tsconfig.json";
     const tsConfigPath: string = process.env["PULUMI_NODEJS_TSCONFIG_PATH"] ?? defaultTsConfigPath;
     const skipProject = !fs.existsSync(tsConfigPath);
+    
+    const typeCheck = (process.env["PULUMI_NODEJS_TYPECHECK"] ?? "true") === "true";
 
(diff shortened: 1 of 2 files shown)


sdk/nodejs/cmd/run/run.ts
```diff
@@ -172,7 +174,7 @@ export function run(
 
     if (typeScript) {
         tsnode.register({
-            typeCheck: true,
```
Frassle, line 175: Rather than another envvar couldn't this be read from tsConfig above? Currently we pull "compilerOptions" out from that but stands to reason we could pull "typeCheck" from there as well.

dmattia: I don't believe `tsConfig` files have a `typeCheck` option, as they are defining how typechecking should be performed. In our case, we definitely want our `tsconfig.json` file to declare how to typecheck, as we still typecheck our monorepo via `tsc`, we just don't want Pulumi to also typecheck.

This is similar to how `ts-node` is often run in production settings with `--transpile-only` enabled

Frassle: Good prompt. I went and double checked what options are valid here and in tsconfig, and your right that typeCheck isn't an option in tsconfig. It also looks like the typeCheck option here is also an old deprecated option and should be replaced with transpileOnly.

sdk/nodejs/cmd/run/run.ts
```diff
@@ -172,7 +174,7 @@ export function run(
...
+            typeCheck,
```
Frassle, line 177:
```suggestion
            transpileOnly,
```

sdk/nodejs/cmd/run/run.ts
```diff
@@ -160,6 +160,8 @@ export function run(
     const defaultTsConfigPath = "tsconfig.json";
     const tsConfigPath: string = process.env["PULUMI_NODEJS_TSCONFIG_PATH"] ?? defaultTsConfigPath;
     const skipProject = !fs.existsSync(tsConfigPath);
+    
+    const typeCheck = (process.env["PULUMI_NODEJS_TYPECHECK"] ?? "true") === "true";
```
Frassle, line 164:
```suggestion
    const transpileOnly = (process.env["PULUMI_NODEJS_TRANSPILE_ONLY"] ?? "false") === "true";
```

sdk/nodejs/cmd/run/run.ts
dmattia: Oh cool, good find. I accepted the suggested changes

Frassle (approved):

PR #2736, opened 2019-05-15
Do not prompt for passphrase multiple times

ellismg:
The change does two things:

- Reorders some calls in the CLI to prevent trying to create a secrets
  manager twice (which would end up prompting twice).

- Adds a cache inside the passphrase secrets manager such that when
  decrypting a deployment, we can re-use the one created earlier in
  the update. This is sort of a hack, but is needed because otherwise
  we would fail to decrypt the deployment, meaning that if you had a
  secret value in your deployment *and* you were using local
  passphrase encryption *and* you had not set PULUMI_CONFIG_PASSPHRASE
  you would get an error asking you to do so.

Fixes #2729

PR #8986, opened 2022-02-14
Fix passphrase secrets provider prompting

### LINKED ISSUE #8173: When using `passphrase` secrets provider with `httpstate` backend, CLI does not properly prompt for passphrase
## Hello!

## Issue details

When using the `passphrase` secrets provider with the `httpstate` (enterprise) backend and no `PULUMI_CONFIG_PASSPHRASE`/`PULUMI_CONFIG_PASSPHRASE_FILE` env var, prompts for passphrase do not work as intended. Sometimes operations fail without prompting for passphrase, and sometimes operations will prompt for a passphrase and then fail as if none had been provided.

### Steps to reproduce

```shell
# No passphrase env vars set
unset PULUMI_CONFIG_PASSPHRASE
unset PULUMI_CONFIG_PASSPHRASE_FILE

# Create and enter a tempdir for the repro
rm -rf "$TMPDIR/pphrase-repro"
mkdir "$TMPDIR/pphrase-repro" && cd "$TMPDIR/pphrase-repro"

# Make sure we're on the enterprise backend (as opposed to say, local)
pulumi logout && pulumi login

pulumi version
# v3.13.2

# Create a new stack.
# You will be prompted thrice for your passphrase: twice to set it up initially, and once to write the initial config
pulumi new typescript \
  --name "pphrase-repro" \
  --stack test-1 \
  --description "A project for repro'ing passphrase issues" \
  --secrets-provider passphrase

# Run an update, skipping preview and answering yes to the prompt.
# You will be prompted once for your passphrase
pulumi up -y -f

# Write the stack state to file.
# You will not be prompted for your passphrase.
pulumi stack export --file stack.json

# Attempt to import the file.
# This will fail for lack of a passphrase, without first prompting for a passphrase.
pulumi stack import --file stack.json
# ❯ pulumi stack import --file stack.json
# error: could not deserialize deployment: constructing secrets manager of type "passphrase": unable to find either `PULUMI_CONFIG_PASSPHRASE` or `PULUMI_CONFIG_PASSPHRASE_FILE` when trying to access the Passphrase Secrets Provider; please ensure one of these environment variables is set to allow the operation to continue

# The destroy command prompts for passphrase, but then fails for lack of passphrase
pulumi destroy
# ❯ pulumi destroy
# Enter your passphrase to unlock config/secrets
#     (set PULUMI_CONFIG_PASSPHRASE or PULUMI_CONFIG_PASSPHRASE_FILE to remember):
# Previewing destroy (test-1)
#
# View Live: https://app.pulumi.com/...
#
# error: could not deserialize deployment: constructing secrets manager of type "passphrase": unable to find either `PULUMI_CONFIG_PASSPHRASE` or `PULUMI_CONFIG_PASSPHRASE_FILE` when trying to access the Passphrase Secrets Provider; please ensure one of these environment variables is set to allow the operation to continue

```

Expected: CLI should prompt for passphrase if one is required but not provided; and, after prompting for passphrase, CLI should use that passphrase for secrets
Actual: In some circumstances, CLI fails without prompting for passphrase; in other circumstances, CLI prompts for passphrase and then fails as if none had been provided.

justinvp:
In #2736, caching was introduced inside the passphrase secrets provider to avoid prompting the user for a password multiple times in the same `pulumi` command. However, the cache is not used in all cases, which leads to double prompting and/or erroring if the passphrase is not set in the `PULUMI_CONFIG_PASSPHRASE` envvar. In some cases, the user is prompted for their passphrase and then the command fails because the passphrase wasn't set via the envvar.

This change addresses these issues. The net effect is that when using the passphrase secrets provider, you'll be prompted for your passphrase up-front once (if interactive and the envvar isn't set), and subsequent uses will use the cached value and won't prompt or error if the envvar isn't set.

Implementation note: This moves the prompting code inside the `passphrase` package. I was originally trying to avoid this, to avoid the package depending on `cmdutil.ReadConsoleNoEcho` and `cmdutil.Interactive`, but this lead to unnatural contortions elsewhere. It ended up being much cleaner to have the prompting happen inside the passphrase implementation. Functions that may prompt (when interactive and the envvar isn't set) include "Prompt" in their name.

Fixes #8173

---

Examples:

```sh
# No passphrase env vars set
unset PULUMI_CONFIG_PASSPHRASE
unset PULUMI_CONFIG_PASSPHRASE_FILE

# Before: this would prompt for a password twice to set the password, and then prompt again when saving the stack
# After: prompts for password twice to setup the password, and doesn't prompt again
pulumi new go --secrets-provider passphrase

# Before: This will fail for lack of a passphrase without first prompting for a passphrase
# After: Prompts for a passphrase once, command succeeds
pulumi stack import --file stack.json

# Before: Prompts for a passphrase, but then fails for lack of passphrase
# After: Prompts for a passphrase once, command succeeds
pulumi destroy
```

### DIFF
--- a/pkg/secrets/passphrase/manager.go
+++ b/pkg/secrets/passphrase/manager.go
@@ -1,4 +1,4 @@
-// Copyright 2016-2019, Pulumi Corporation.
+// Copyright 2016-2022, Pulumi Corporation.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -16,6 +16,7 @@
 package passphrase
 
 import (
+	cryptorand "crypto/rand"
 	"encoding/base64"
 	"encoding/json"
 	"errors"
@@ -27,7 +28,9 @@ import (
 	"sync"
 
 	"github.com/pulumi/pulumi/pkg/v3/secrets"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/diag"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource/config"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/util/cmdutil"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 )
 
@@ -114,12 +117,8 @@ var cache map[string]secrets.Manager
 func NewPassphaseSecretsManager(phrase string, state string) (secrets.Manager, error) {
 	// check the cache first, if we have already seen this state before, return a cached value.
 	lock.Lock()
-	if cache == nil {
-		cache = make(map[string]secrets.Manager)
-	}
 	cachedValue := cache[state]
 	lock.Unlock()
-
 	if cachedValue != nil {
 		return cachedValue, nil
 	}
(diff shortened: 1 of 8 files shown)


pkg/secrets/passphrase/manager.go
```diff
@@ -27,7 +28,9 @@ import (
 	"sync"
 
 	"github.com/pulumi/pulumi/pkg/v3/secrets"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/diag"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource/config"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/util/cmdutil"
```
pgavlin, line 33: If we want to avoid this dependency, we could define an interface like

```go
interface Prompter {
    ReadPassword(prompt string) (string, error)
}
```

justinvp: Yeah, that's basically the approach I was going down initially (although, typed as a func rather than an interface), but then you have to pass it through multiple layers. In the end, I actually don't think it's so bad having `passphrase` use `cmdutil` as it keeps all the passphrase-related code together, and doesn't require injecting the dependency and plumbing it through.

pkg/secrets/passphrase/manager.go
```diff
@@ -111,15 +114,18 @@ func (sm *localSecretsManager) Encrypter() (config.Encrypter, error) {
...
+// clearCache is used to clear the cache, for tests.
+func clearCache() {
+	lock.Lock()
+	cache = nil
+	lock.Unlock()
+}
+
 func NewPassphaseSecretsManager(phrase string, state string) (secrets.Manager, error) {
 	// check the cache first, if we have already seen this state before, return a cached value.
 	lock.Lock()
```
pgavlin, line 126: Nit: it would be nice to remove the open-coded locking by adding some accessor methods (e.g. `getSecretsManager(state string) (secrets.Manager, bool)` and `setSecretsManager(state string) (secrets.Manager, bool)`). I think there's also some TOCTOU bugs in here...

pgavlin (approved):

PR #8995, opened 2022-02-15
Revert "Lock inside filestate.apply instead of Update/Delete/Import (#8806)"

Frassle:
This reverts commit 8b4da4eba73cb98fe4c06b44179183562f707247 (reverts #8806).


mikhailshilkov:
Given the likely user impact, I think we should changelog this

PR #9013, opened 2022-02-18
Allow trivial stack with no engine events

AaronFriel:
While testing another feature, I needed to create a stack to exercise _only_ outputs. That stack emitted no engine events and therefore failed in programtest.


iwahbe (approved):
Looks good. LGTM

Though process:
If you are relying on the program test to ensure events are created, you should use other hooks to ensure that not only are events created, but that the right events are created.

PR #9026, opened 2022-02-22
[sdk/nodejs] Fix make install

### LINKED ISSUE #9019: `make install` no longer working in `sdk/nodejs`
I'm now seeing this:

```
➜  nodejs git:(master) ✗ make install
INSTALL:
[ ! -e "/Users/lukehoban/.pulumi-dev/node_modules/@pulumi/pulumi" ] || rm -rf "/Users/lukehoban/.pulumi-dev/node_modules/@pulumi/pulumi"
mkdir -p "/Users/lukehoban/.pulumi-dev/node_modules/@pulumi/pulumi"
cp -r bin/. "/Users/lukehoban/.pulumi-dev/node_modules/@pulumi/pulumi"
cp yarn.lock "/Users/lukehoban/.pulumi-dev/node_modules/@pulumi/pulumi"
rm -rf "/Users/lukehoban/.pulumi-dev/node_modules/@pulumi/pulumi/node_modules"
cd "/Users/lukehoban/.pulumi-dev/node_modules/@pulumi/pulumi" && \
        yarn install --prefer-offline --production && \
        (yarn unlink > /dev/null 2>&1 || true) && \
        yarn link
yarn install v1.22.10
[1/5] 🔍  Validating package.json...
[2/5] 🔍  Resolving packages...
error Package "mockpackage" refers to a non-existing file '"/Users/lukehoban/.pulumi-dev/node_modules/@pulumi/pulumi/tests/mockpackage"'.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
make: *** [install] Error 1
```

This appears to have been introduced with https://github.com/pulumi/pulumi/pull/8893.  It does not affect `make publish`, and therefore does not break the workflow used in CI, but does impact the normal local development flow.

dixler:
# Description
Fixed nodejs make install script by dropping `devDependencies` in `.pulumi-dev/node_modules` as it does not use them. Yarn attempts to resolve devDependencies on prod builds. Linked a discussion in #9019. Opening an issue #9027 to add CI tests for dev scripts.

Fixes #9019

Filing #9027 #9042

## Checklist

### DIFF
--- a/sdk/nodejs/Makefile
+++ b/sdk/nodejs/Makefile
@@ -29,6 +29,7 @@ build_package:: ensure
 	cp tests/runtime/jsClosureCases_10_4.js bin/tests/runtime
 	cp -R tests/automation/data/. bin/tests/automation/data/
 	cp README.md ../../LICENSE ./dist/* bin/
+# safely remove mockpackage dependency see [pulumi/pulumi#9026]
 	cp package.json yarn.lock bin/ && cd bin && yarn remove mockpackage && rm yarn.lock
 	node ../../scripts/reversion.js bin/package.json ${VERSION}
 	node ../../scripts/reversion.js bin/version.js ${VERSION}
--- a/sdk/nodejs/Makefile
+++ b/sdk/nodejs/Makefile
@@ -28,11 +28,11 @@ build_package:: ensure
 	cp tests/runtime/jsClosureCases_8.js bin/tests/runtime
 	cp tests/runtime/jsClosureCases_10_4.js bin/tests/runtime
 	cp -R tests/automation/data/. bin/tests/automation/data/
-	cp README.md ../../LICENSE package.json ./dist/* bin/
+	cp README.md ../../LICENSE ./dist/* bin/
+	sed 's/"devDependencies"/"_devDependencies"/g' package.json > bin/package.json
 	node ../../scripts/reversion.js bin/package.json ${VERSION}
 	node ../../scripts/reversion.js bin/version.js ${VERSION}
 	cp -R proto/. bin/proto/
-	cp -R tests/mockpackage bin/tests/
 	mkdir -p bin/tests/runtime/langhost/cases/
 	find tests/runtime/langhost/cases/* -type d -exec cp -R {} bin/tests/runtime/langhost/cases/ \;
 
--- a/sdk/nodejs/Makefile
+++ b/sdk/nodejs/Makefile
@@ -29,7 +29,7 @@ build_package:: ensure
 	cp tests/runtime/jsClosureCases_10_4.js bin/tests/runtime
 	cp -R tests/automation/data/. bin/tests/automation/data/
 	cp README.md ../../LICENSE ./dist/* bin/
-	cat package.json | grep -v '"mockpackage"'> bin/package.json
+	cp package.json yarn.lock bin/ && cd bin && yarn remove mockpackage && rm yarn.lock
 	node ../../scripts/reversion.js bin/package.json ${VERSION}
 	node ../../scripts/reversion.js bin/version.js ${VERSION}
 	cp -R proto/. bin/proto/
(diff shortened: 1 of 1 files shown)


sdk/nodejs/Makefile
```diff
@@ -32,6 +32,7 @@ build_package:: ensure
 	node ../../scripts/reversion.js bin/package.json ${VERSION}
 	node ../../scripts/reversion.js bin/version.js ${VERSION}
 	cp -R proto/. bin/proto/
+	cp -R tests/mockpackage bin/tests/
```
lukehoban, line 35: This means we end up distributing mockpackage as part of our distribution. I wasn’t clear on exactly why mockpackage needed to be included as a dependency originally, but it seems somewhat unfortunate to have to actually ship it. Is there a way we can avoid the need for the mockpackage dependency in the first place?

dixler: I added mockpackage to write a test to check that function serialization properly works with a package that specifies exports. I could remove the test and the package.

I could also add additional scripting to the makefile to only add and remove the mockpackage dependency during testing.

There was a solution in https://github.com/yarnpkg/yarn/issues/3630#issuecomment-641959126 and it involves a one-liner that removes devDependencies out of our package.json so it doesn't have to be copied over. I was initially using it, but figured it might be less favorable to have long awk one-liners editing files.

dixler: I have this implementation that strips the devDependencies by renaming the field.

https://github.com/pulumi/pulumi/blob/dixler/9019/fix-make-install-strip/sdk/nodejs/Makefile#L31-L32

dixler: Using other strategy to discard devDependencies from `package.json`.

sdk/nodejs/Makefile
```diff
@@ -28,7 +28,8 @@ build_package:: ensure
 	cp tests/runtime/jsClosureCases_8.js bin/tests/runtime
 	cp tests/runtime/jsClosureCases_10_4.js bin/tests/runtime
 	cp -R tests/automation/data/. bin/tests/automation/data/
-	cp README.md ../../LICENSE package.json ./dist/* bin/
+	cp README.md ../../LICENSE ./dist/* bin/
+	sed 's/"devDependencies"/"_devDependencies"/g' package.json > bin/package.json
```
mikhailshilkov, line 32: This feels a bit awkward - both removing all `devDependencies` and leaving `_devDependencies` in the file. Would there be a way to remove just the `mockPackage` line?

dixler: ok. just pushed another solution to strip lines containing "mockpackage".

dixler:
Filed #9042 to clean up mockpackage.

@lukehoban is the approach of `yarn remove`-ing mockpackage in the build_package make target sufficient to resolve #9019?

sdk/nodejs/Makefile
```diff
@@ -28,7 +28,8 @@ build_package:: ensure
 	cp tests/runtime/jsClosureCases_8.js bin/tests/runtime
 	cp tests/runtime/jsClosureCases_10_4.js bin/tests/runtime
 	cp -R tests/automation/data/. bin/tests/automation/data/
-	cp README.md ../../LICENSE package.json ./dist/* bin/
+	cp README.md ../../LICENSE ./dist/* bin/
+	cp package.json yarn.lock bin/ && cd bin && yarn remove mockpackage && rm yarn.lock
```
iwahbe, line 32: Why do we copy `yarn.lock` and then remove it?

dixler: In order to run `yarn remove mockpackage` to remove mockpackage from `package.json` for the bin/, I put a `yarn.lock` file in the same directory. The alternative is that I find an obscure flag in yarn that ignores a yarn.lock. The reason it's being removed is because it wasn't there in the first place and I wanted to avoid adding anything that wasn't in there in the first place.

I preferred using `yarn` to remove it because it seemed safer than sed/awk/grep/etc and trying to match text.

iwahbe (approved):
I would appreciate a comment in the makefile that explains what you said in https://github.com/pulumi/pulumi/pull/9026/files#r817230665
Otherwise LGTM

Issue #9014, opened 2022-02-18
`master` build failing on Windows


lukehoban:
The last few `master` builds have failed on Windows with this:

```
        	Error:      	"{\n  \"plugins\": null,\n  \"host\": {\n    \"os\": \"Microsoft Windows Server 2022 Datacenter\",\n    \"version\": \"10.0.20348 Build 20348\",\n    \"arch\": \"x86_64\"\n  },\n  \"backend\": {\n    \"name\": \"fv-az448-481\",\n    \"url\": \"file://C:/Users/RUNNER~1/AppData/Local/Temp/test-env3127767626\",\n    \"user\": \"fv-az448-481\\\\runneradmin\"\n  },\n  \"currentStack\": null,\n  \"cliAbout\": {\n    \"version\": \"3.25.0-alpha.1645132829\",\n    \"goVersion\": \"go1.17.7\",\n    \"goCompiler\": \"gc\"\n  },\n  \"runtime\": {\n    \"language\": \"nodejs\",\n    \"executable\": \"C:\\\\hostedtoolcache\\\\windows\\\\node\\\\14.19.0\\\\x64\\\\node.exe\",\n    \"version\": \"v14.19.0\"\n  },\n  \"dependencies\": null,\n  \"errors\": [\n    \"Failed to get information about the plugin: failed to load language plugin nodejs: could not read plugin [D:\\\\a\\\\pulumi\\\\pulumi\\\\bin\\\\pulumi-language-nodejs.exe] stdout: EOF\",\n    \"Failed to get information about the Puluimi program's plugins: Could not find either C:\\\\Users\\\\RUNNER~1\\\\AppData\\\\Local\\\\Temp\\\\test-env3127767626\\\\yarn.lock or C:\\\\Users\\\\RUNNER~1\\\\AppData\\\\Local\\\\Temp\\\\test-env3127767626\\\\package-lock.json\",\n    \"Failed to get information about the current stack: No current stack\"\n  ]\n}\n" does not contain "go1.17.6"
```

E.g. https://github.com/pulumi/pulumi/actions/runs/1861124616.

AaronFriel:
Error reports that the host OS is "Windows Server 2022 Datacenter", and that OS became the default recently. Our last passing test ran on Windows Server 2019:

https://github.com/pulumi/pulumi/runs/5224033778?check_suite_focus=true#step:1:3

mikhailshilkov:
The latest run had a single error on Windows:

[...]

PR #9028, opened 2022-02-22
CI: resolve build/test failures on Windows

AaronFriel:
See: #9014 

* Ensures our builders & runners check that the version of go in the runner cache is the latest that matches the version requested (currently 1.17.x).

* Makes the failing test more resilient and more likely to pass on developer machines by making the test only check the major and minor version, which matches our CI system which expects go 1.17.x

Detailed build logs:

Builder (macos or ubuntu) builds Pulumi with go 1.17.7: https://github.com/pulumi/pulumi/runs/5271142183?check_suite_focus=true#step:7:14

Test harness (Windows) installs go 1.17.6 from cache: https://github.com/pulumi/pulumi/runs/5271213861?check_suite_focus=true#step:7:17

Test harness then fails on checking if the built binary's "pulumi about" result contains "go1.17.6", which it does not because it was built with 1.17.7: https://github.com/pulumi/pulumi/runs/5271213861?check_suite_focus=true#step:49:37

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -88,6 +88,7 @@ jobs:
         uses: actions/setup-go@v2
         with:
           go-version: ${{ inputs.go-version }}
+          check-latest: true
       - id: go-cache-paths
         run: |
           echo "::set-output name=go-build::$(go env GOCACHE)"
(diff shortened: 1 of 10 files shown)


AaronFriel:
Cleaned up github bot comments made prior to commit rewrite and rebase/force push.

iwahbe (approved):
This looks good to me. I would wait until tests are ✅ before merging though.

PR #9029, opened 2022-02-22
ci: default builder to linux, not macos

AaronFriel:
Per comment in CI workflow, ubuntu-latest is the quicker build agent. Aligns the checksums by also publishing via ubuntu-latest.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -53,7 +53,7 @@ jobs:
     strategy:
       matrix:
         include:
-          - platform: macos-latest
+          - platform: ${{ inputs.default-build-platform }}
             os: darwin
           - platform: ${{ inputs.default-build-platform }}
             os: linux
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -78,7 +78,7 @@ jobs:
           make -C sdk/${{ matrix.language}} publish
   publish-binaries:
     name: Publish Binaries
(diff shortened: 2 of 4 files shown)


jaxxstorm:
Are we sure we want to do this? We still use cgo in some dependencies, so it'll mess with the default DNS resolver for the binary on OS X

AaronFriel:
We may want to block this until we've ensured this repo & others are building on the next point releases of Go, or Go 1.18.

I believe the issues below track the root cause and solution to DNS issues under split VPNs.

See:
- Compilation of Go DNS related issues, root caused by me on 2/9: https://github.com/golang/go/issues/51127
- Go CL merged prior to 1.18 release and backported to next releases of 1.16 and 1.17 https://github.com/golang/go/issues/51153

t0yv0:
This should not be possible because of our use of CGO. It's just not going to compile a macos binary on Linux.

t0yv0:
The PR verification is not testing the changes, you need to modify the import of the build.yml to go from @this-branch instead of @master. @AaronFriel

AaronFriel:
@t0yv0 @jaxxstorm I'm closing this as I got some additional detail from a customer and a helpful Hacker News comment.

This issue in Golang causes split DNS (work VPNs, etc.) to not work correctly under some configurations on Mac if we cross compile *to* Mac. https://github.com/golang/go/issues/12524

PR #9039, opened 2022-02-23
[secrets] Fix + refactor bulk decryption.

pgavlin:
The bulk decryption logic in deployment deserialization did not recur
into slices or maps. This prevented the bulk decryption of nested secret
values. These changes fix that bug, improve test coverage, and refactor
the bulk decryption code for simplicity, clarity, and separation of
concerns.

Related to https://github.com/pulumi/home/issues/1842.

### DIFF
--- a/pkg/resource/stack/deployment.go
+++ b/pkg/resource/stack/deployment.go
@@ -232,7 +232,6 @@ func DeserializeDeploymentV3(deployment apitype.DeploymentV3, secretsProv Secret
 		if err != nil {
 			return nil, err
 		}
-		dec = config.NewCachedDecrypter(d)
 
 		// Do a first pass through state and collect all of the secrets that need decrypting.
 		// We will collect all secrets and decrypt them all at once, rather than just-in-time.
@@ -240,15 +239,16 @@ func DeserializeDeploymentV3(deployment apitype.DeploymentV3, secretsProv Secret
 		// wait times in stacks with a large number of secrets.
 		var ciphertexts []string
 		for _, res := range deployment.Resources {
-			ciphertexts = append(ciphertexts, scanResource(res)...)
+			collectCiphertexts(&ciphertexts, res.Inputs)
+			collectCiphertexts(&ciphertexts, res.Outputs)
 		}
 
-		// we call BulkDecrypt to Prime the cache with the decrypted cipertexts
-		// we don't actually use the response here at this point
-		_, err = dec.BulkDecrypt(ciphertexts)
+		// Decrypt the collected secrets and create a decrypter that will use the result as a cache.
+		cache, err := config.BulkDecrypt(d, ciphertexts)
 		if err != nil {
 			return nil, err
 		}
+		dec = newMapDecrypter(d, cache)
 
 		e, err := secretsManager.Encrypter()
 		if err != nil {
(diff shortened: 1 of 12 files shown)


pgavlin:
One note: bulk decryption will not see secrets nested _inside other secrets_. A secret is stored as the encrypted, JSON-serialized text of its plain value, so the contents of the plain value are not visible until the secret is decrypted. So the inner secrets in an example like this aren't visible to the scanner: `secret(secret(secret(....)))`. This should be fine, though, as those inner secrets are not encrypted; they're just marked as secrets (technically, they're encrypted using a no-op encrypter).

pkg/resource/stack/secrets.go
```diff
@@ -154,3 +150,35 @@ func (c *cachingCrypter) encryptSecret(secret *resource.Secret, plaintext string
 func (c *cachingCrypter) insert(secret *resource.Secret, plaintext, ciphertext string) {
 	c.cache[secret] = cacheEntry{plaintext, ciphertext}
 }
+
+// mapDecrypter is a Decrypter with a preloaded cache. This decrypter is used specifically for deserialization,
+// where the deserializer is expected to prime the cache by scanning each resource for secrets, then decrypting all
+// of the discovered secrets en masse. Although each call to Decrypt _should_ hit the cache, a mapDecrypter does
+// carry an underlying Decrypter in the event that a secret was missed.
+type mapDecrypter struct {
```
Frassle, line 158: Feels like this is similar to the cachingCrypter idea. Wonder if all the caching logic could just be done once, although this caches by cipertext and cachingCrypter is by Secret pointers so maybe makes no sense to combine.

sdk/go/common/resource/config/crypt.go
```diff
@@ -36,6 +36,10 @@ type Encrypter interface {
 // Decrypter decrypts encrypted ciphertext to its plaintext representation.
 type Decrypter interface {
 	DecryptValue(ciphertext string) (string, error)
+}
+
+// BulkDecrypter is a Decrypter that also supports bulk decryption of secrets.
+type BulkDecrypter interface {
```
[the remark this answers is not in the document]
pgavlin: Ah, yep--good catch.

pkg/resource/stack/secrets.go
pgavlin: > this caches by cipertext and cachingCrypter is by Secret pointers so maybe makes no sense to combine.

This is what led me to keep the two separate. My first instinct was to combine them, but the differences in key type and semantics led me down the proposed approach.

pgavlin: Added a comment re: the distinction

pkg/resource/stack/deployment.go
```diff
@@ -456,34 +456,24 @@ func SerializePropertyValue(prop resource.PropertyValue, enc config.Encrypter,
...
+			collectCiphertexts(ciphertexts, v)
+		}
+	case map[string]interface{}:
+		if prop[resource.SigKey] == resource.SecretSig {
+			if ciphertext, cipherOk := prop["ciphertext"].(string); cipherOk {
+				*ciphertexts = append(*ciphertexts, ciphertext)
+			}
+		} else {
+			for _, v := range prop {
+				collectCiphertexts(ciphertexts, v)
```
dixler, line 473: nice refactor

Frassle (approved):

PR #9044, opened 2022-02-24
Fix DeprecationWarning  fs.rmdir (take 2)

jitbasemartin:
# Description

Take 2 of https://github.com/pulumi/pulumi/pull/7872

Fix following deprecation warning

```txt
(node:49420) [DEP0147] DeprecationWarning: In future versions of Node.js, fs.rmdir(path, { recursive: true }) will be removed. Use fs.rm(path, { recursive: true }) instead
```


Frassle (approved):

Frassle:
Thanks for this.

PR #8452, opened 2021-11-18
Add tsconfig option to specify tsconfig path

iwahbe:
# Description
Adds a runtime option to specify the path to `tsconfig.json`. 
```yaml
name: pulumi-project
runtime:
  name: nodejs
  options:
    tsconfig: mytsconfig.json
```

Fixes #8415

## Checklist

- [x] I have added tests that prove my fix is effective or that my feature works

- [x] I have updated the [CHANGELOG-PENDING](https://github.com/pulumi/pulumi/blob/master/CHANGELOG_PENDING.md) file with my change

- [ ] Yes, there are changes in this PR that warrants bumping the Pulumi Service API version

Issue #8415, opened 2021-11-15
tsconfig "module" conflict: project needs "esnext", pulumi needs "commonjs"


chrishoward:
## Hello!
- Vote on this issue by adding a 👍 reaction
- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details
We are using pulumi (typescript) in a repository which also has a create-react-app application. Our tsconfig file has `"module": "esnext"` as required by create-react-app. We had no issues using pulumi v2.x.x as it ignored our tsconfig and used its own internal config [**here**](https://github.com/pulumi/pulumi/blob/272c4643b20d40113689475fd2f9fe3134e6d431/sdk/nodejs/cmd/run/run.ts#L171).

In pulumi v3.x.x it was changed to use the tsconfig found in the project root and override the internal pulumi options with `...compilerOptions`. Now our pulumi throws the error:

```pulumi:pulumi:Stack (roads-vicroads-ui-dev):
    error: Running program '/home/coder/my-repository' failed with an unhandled exception:
    /home/coder/my-repository/pulumi.ts:52
    import * as mime from "mime";
    ^^^^^^

    SyntaxError: Cannot use import statement outside a module
        at wrapSafe (internal/modules/cjs/loader.js:988:16)
        at Module._compile (internal/modules/cjs/loader.js:1036:27)
        at Module.m._compile (/home/coder/my-repository/node_modules/ts-node/src/index.ts:439:23)
        at Module._extensions..js (internal/modules/cjs/loader.js:1101:10)
        at Object.require.extensions.<computed> [as .ts] (/home/coder/my-repository/node_modules/ts-node/src/index.ts:442:12)
        at Module.load (internal/modules/cjs/loader.js:937:32)
        at Function.Module._load (internal/modules/cjs/loader.js:778:12)
        at Module.require (internal/modules/cjs/loader.js:961:19)
        at require (internal/modules/cjs/helpers.js:92:18)
        at Object.<anonymous> (/home/coder/my-repository/node_modules/@pulumi/pulumi/cmd/run/run.js:219:31)
```

Attempts to change our tsconfig "module" value to "commonjs" see it forced back to "esnext" by create-react-app. 

Does pulumi support other module options? Is it possible to keep the module as commonjs for pulumi?
I'll create a PR for this approach for discussion.

### Steps to reproduce
1. Set up pulumi typescript
2. Set "module" to "esnext" in tsconfig.json
[...]

PR #7828, opened 2021-08-23
Bump ts-node version 7.0.1=>10.2.0

iwahbe:
# Description

Bump `ts-node` version from 7 to 10.

Fixes #4876

We are using the CI to test this. We can also look at the time it took for test's to run to check for performance regressions. 

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works

- [ ] I have updated the [CHANGELOG-PENDING](https://github.com/pulumi/pulumi/blob/master/CHANGELOG_PENDING.md) file with my change

- [ ] Yes, there are changes in this PR that warrants bumping the Pulumi Service API version

PR #9049, opened 2022-02-25
Use SWC to transpile typescript projects

dmattia:
# Description

I just went down an interesting rabbit hole following the history of the run.ts file that sets up the ts-node running of typescript pulumi programs, and believe that simplifying the existing code will better future-proof pulumi, while providing better performance, with lower maintenance, while supporting more Typescript setups.

Here's the history of this file as I understand it:

-  Once upon a time, Luke Hoban noticed that having a tsconfig.json file in their home directory caused pulumi to hang, as ts-node was by default looking up the chain for any tsconfig.json file it could find, and was then attempting to typecheck all code in the `files` of that config file (https://github.com/pulumi/pulumi/issues/1772)
- As a solution, https://github.com/pulumi/pulumi/pull/1857 came along and said that pulumi would only check for tsconfig.json files in the project directory, solving Luke's problem, but creating an incompatibility with some fairly standard typescript layouts where the tsconfig file is up the chain.
- Issues and comments came in from users with confusing error messages where their tsconfig files up the directory chain were being ignored, and the error messages were non-obvious as to what was wrong.
  - https://github.com/pulumi/pulumi/pull/1857#issuecomment-677694535 has not been answered
  - Issue https://github.com/pulumi/pulumi/issues/8415 was opened
- PR https://github.com/pulumi/pulumi/pull/8452 patched the situation by adding a new config option that would allow users to specify their tsconfig file. Of note, they would likely have to search through the existing issues to even find that their error messages were related to tsconfig files, as Pulumi's handling here is non-standard
- Performance-related issues began popping up around ts-node, making it hard to support upgrading ts-node or typescript versions:
  - At one point, ts-node was updated from version 7 to version 8: https://github.com/pulumi/pulumi/pull/3627/files
  - An issue opened that startup times were slow for ts pulumi: https://github.com/pulumi/pulumi/issues/3671
  - Joe Duffy found that the lines reading in tsconfig files are non-performant: https://github.com/pulumi/pulumi/issues/3671#issuecomment-584718728
  - Luke Hoban found that the ts-node bump caused the perf decrease: https://github.com/pulumi/pulumi/issues/3671#issuecomment-592810579
  - A PR was merged moving the ts-node version back down: https://github.com/pulumi/pulumi/pull/4007
  - An issue was created to revisit the ts-node version: https://github.com/pulumi/pulumi/issues/4876
  - A PR was opened to bump up ts-node to version 10: https://github.com/pulumi/pulumi/pull/7828
  - That PR was also closed due to performance reasons: https://github.com/pulumi/pulumi/pull/7828#issuecomment-904994057
  - A suggestion for using something like esbuild without typechecking was made: https://github.com/pulumi/pulumi/issues/4876#issuecomment-1028616028. Of note, ts-node can be used without typechecking as well
  - PR https://github.com/pulumi/pulumi/pull/8981 added support to turn off typechecking. This still does not fix all of the problems though, as ts-node can still fail to run the code if skipBuilds was turned on, which was fixed to be on anytime tsconfig.json was not in the project dir.

This PR does a few things:
- Removes some of the patching that has happened over time (while still supporting custom tsconfig.json locations that are not in the parent chain of the project dir in the same way as before).
- Bumps up the ts-node version to major version 10
- Turns off typechecking
- Turns on transpiling typescript using SWC, which is sort of like esbuild (https://typestrong.org/ts-node/docs/transpilers/)

I created a test repo that uses these changes, comparing the performance on a plain `pulumi new aws-typescript` repo using the latest pulumi/pulumi and my fork: https://github.com/dmattia/pulumi-fork-perf-test

Ignoring the CI setup time and yarn installation time, the `pulumi preview` step with pulumi v3.25.0 took:
- 21s: https://github.com/dmattia/pulumi-fork-perf-test/runs/5335693814?check_suite_focus=true
- 16s: https://github.com/dmattia/pulumi-fork-perf-test/runs/5335693888?check_suite_focus=true
- 20s: https://github.com/dmattia/pulumi-fork-perf-test/runs/5335693955?check_suite_focus=true

and using the fork took:
- 9s: https://github.com/dmattia/pulumi-fork-perf-test/runs/5335694020?check_suite_focus=true
- 7s: https://github.com/dmattia/pulumi-fork-perf-test/runs/5335694060?check_suite_focus=true
- 8s: https://github.com/dmattia/pulumi-fork-perf-test/runs/5335694110?check_suite_focus=true

This is a significant performance increase, and I would encourage anyone interested to fork that repo and test for themselves on more operating systems, node versions, etc. if desired. Or, feel free to comment here what systems you're interested in and I will gladly update the job matrix to test using those options.

As for the lack of typechecking, this is where it gets a lot more into a matter of opinion, but my thought is this: If pulumi wants to typecheck by default, it should use standard typescript handling without custom handling of tsconfig.json files. Doing both of those things by default makes for a challenging onboarding experience for anyone in a custom-setup typescript repo that is not using `pulumi new` and doesn't just happen to have their tsconfig file in the correct place.

There is also a middle ground here where we do not use the swc transpiler, turn on typechecking by default, and use `scope` and `scopeDir` in ts-node to tell typescript to only typecheck the files in the pulumi working directory (see: https://github.com/TypeStrong/ts-node/blob/main/src/index.ts#L209-L218). These options, only available in ts-node v10.x and higher, would make the case where someone had a rogue tsconfig.json file in their home directory a non-issue (although they should really remove that 😬) as only the entry file (and other files in the same dir) would be typechecked.

Fixes https://github.com/pulumi/pulumi/issues/4876

Suggested reviewers: @iwahbe, @lukehoban, as they have been involved in many of the above discussions

## Checklist

### DIFF
--- a/sdk/nodejs/cmd/run/run.ts
+++ b/sdk/nodejs/cmd/run/run.ts
@@ -152,37 +152,13 @@ export function run(
     // If this is a typescript project, we'll want to load node-ts.
     const typeScript: boolean = process.env["PULUMI_NODEJS_TYPESCRIPT"] === "true";
 
-    // We provide reasonable defaults for many ts options, meaning you don't need to have a tsconfig.json present
-    // if you want to use TypeScript with Pulumi. However, ts-node's default behavior is to walk up from the cwd to
-    // find a tsconfig.json. For us, it's reasonable to say that the "root" of the project is the cwd,
-    // if there's a tsconfig.json file here. Otherwise, just tell ts-node to not load project options at all.
-    // This helps with cases like pulumi/pulumi#1772.
-    const defaultTsConfigPath = "tsconfig.json";
-    const tsConfigPath: string = process.env["PULUMI_NODEJS_TSCONFIG_PATH"] ?? defaultTsConfigPath;
-    const skipProject = !fs.existsSync(tsConfigPath);
-
-    const transpileOnly = (process.env["PULUMI_NODEJS_TRANSPILE_ONLY"] ?? "false") === "true";
-
-    let compilerOptions: object;
-    try {
-        const tsConfigString = fs.readFileSync(tsConfigPath).toString();
-        const tsConfig = parseConfigFileTextToJson(tsConfigPath, tsConfigString).config;
-        compilerOptions = tsConfig["compilerOptions"] ?? {};
-    } catch (e) {
-        compilerOptions = {};
-    }
 
     if (typeScript) {
         tsnode.register({
-            transpileOnly,
-            skipProject: skipProject,
-            compilerOptions: {
-                target: "es6",
-                module: "commonjs",
-                moduleResolution: "node",
-                sourceMap: "true",
-                ...compilerOptions,
-            },
+            typeCheck: false,
+            transpileOnly: true,
+            swc: true,
+            project: process.env["PULUMI_NODEJS_TSCONFIG_PATH"], // will be undefined most of the time
         });
     }
 
@@ -239,15 +215,6 @@ ${defaultMessage}`);
     process.on("exit", runtime.disconnectSync);
 
     programStarted();
-
-    // This needs to occur after `programStarted` to ensure execution of the parent process stops.
-    if (skipProject && tsConfigPath !== defaultTsConfigPath) {
-        return new Promise(() => {
-            const e = new Error(`tsconfig path was set to ${tsConfigPath} but the file was not found`);
-            e.stack = undefined;
-            throw e;
-        });
-    }
 
     const runProgram = async () => {
         // We run the program inside this context so that it adopts all resources.
--- a/sdk/nodejs/cmd/run/run.ts
+++ b/sdk/nodejs/cmd/run/run.ts
@@ -157,7 +157,6 @@ export function run(
 
     if (typeScript) {
         tsnode.register({
-            typeCheck: !transpileOnly,
             transpileOnly,
             swc: transpileOnly,
             project: process.env["PULUMI_NODEJS_TSCONFIG_PATH"], // will be undefined most of the time
--- a/sdk/nodejs/cmd/run/run.ts
+++ b/sdk/nodejs/cmd/run/run.ts
@@ -150,14 +150,16 @@ export function run(
     }
 
     // If this is a typescript project, we'll want to load node-ts.
-    const typeScript: boolean = process.env["PULUMI_NODEJS_TYPESCRIPT"] === "true";
+    const typeScript = process.env["PULUMI_NODEJS_TYPESCRIPT"] === "true";
 
+    // Users can optionally disable typechecking when using typescript projects 
+    const transpileOnly = process.env["PULUMI_NODEJS_TRANSPILE_ONLY"] === "true";
 
     if (typeScript) {
         tsnode.register({
-            typeCheck: false,
-            transpileOnly: true,
-            swc: true,
+            typeCheck: !transpileOnly,
+            transpileOnly,
+            swc: transpileOnly,
             project: process.env["PULUMI_NODEJS_TSCONFIG_PATH"], // will be undefined most of the time
         });
     }
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,3 +1,6 @@
 ### Improvements
 
+[sdk/nodejs] - Use SWC to transpile Typescript code without typechecking
+  [#9049](https://github.com/pulumi/pulumi/pull/9049)
+
 ### Bug Fixes
--- a/sdk/nodejs/package.json
+++ b/sdk/nodejs/package.json
@@ -21,8 +21,9 @@
         "require-from-string": "^2.0.1",
         "semver": "^6.1.0",
         "source-map-support": "^0.4.16",
-        "ts-node": "^7.0.1",
-        "typescript": "~3.7.3",
+        "ts-node": "^10.5.0",
+        "@swc/core": "^1.2.145",
+        "typescript": "~3.8.3",
         "upath": "^1.1.0"
     },
     "devDependencies": {


lukehoban:
> Turns off typechecking

This is the one biggest question I have here.  This is a meaningful user experience change.  Today, users get errors early in their deployment if there are TypeScript-level issues.  With this change, they would not.  We'd be relying on IDE tooling to capture errors before running `pulumi up`.  That may be worth it for the performance win, but there is a bit of a tradeoff here.  I think at the very least we should let users opt in/out of this in case they want the alternative to the default.

> Removes some of the patching that has happened over time (while still supporting custom tsconfig.json locations that are not in the parent chain of the project dir in the same way as before).

Previously there were "good defaults" for the case where there was not a tsconfig.json - what happens with these changes to the defaults being used for compilation?  Do we still support projects with no tsconfig.json?  What default compilation options get used?

Frassle:
We've only just recently added the `PULUMI_NODEJS_TRANSPILE_ONLY` option, echoing Lukes comments I see no reason to remove that option as part of this PR.

michaelfarrell76:
> This is the one biggest question I have here. This is a meaningful user experience change. Today, users get errors early in their deployment if there are TypeScript-level issues. With this change, they would not. We'd be relying on IDE tooling to capture errors before running pulumi up. That may be worth it for the performance win, but there is a bit of a tradeoff here. I think at the very least we should let users opt in/out of this in case they want the alternative to the default.

@lukehoban maybe there could be an option to include or not include this? We do a similar thing with webpack. disabling type checking during compilation makes a major different in speed. and most of the time the build step doesnt even run on a CI if the type checking fails

dmattia:
> We've only just recently added the PULUMI_NODEJS_TRANSPILE_ONLY option, echoing Lukes comments I see no reason to remove that option as part of this PR.

Yeah I kinda wish I'd waited on merging that PR until I went through the full history here 😬 

But for sure, I can update to make transpiling optional. So long as standard tsconfig.json setups are supported, that seems fine to me.

> We'd be relying on IDE tooling to capture errors before running pulumi up

This is correct, with the additional comment that users can run `tsc` directly in addition to relying on IDE tooling.

> I think at the very least we should let users opt in/out of this in case they want the alternative to the default

Totally fair I think. Do you have a preference on which is the default? Maybe if we want to have typechecking off by default, we leave it as on by default in this PR, and then change over the default during the next major version release?

> Previously there were "good defaults" for the case where there was not a tsconfig.json - what happens with these changes to the defaults being used for compilation? Do we still support projects with no tsconfig.json? What default compilation options get used?

I believe this section of ts-node's documentation covers this: https://github.com/TypeStrong/ts-node#tsconfigbases, where ts-node will selectively pick compiler option defaults based on the runtime being used, such as https://github.com/tsconfig/bases/blob/main/bases/node14.json for node14.

I have not tested this fully though other than reading those docs.

Frassle:
> Totally fair I think. Do you have a preference on which is the default? 

I think we'd want to keep the current behavior of typechecking by default. If you can think of a good way we could better advertise the `PULUMI_NODEJS_TRANSPILE_ONLY` option to those who could make use it we'd be all ears.

dmattia:
@Frassle Makes sense, I'll leave the default as it was then.

I just pushed up the code to make transpilation optional. You can see https://github.com/dmattia/pulumi-fork-perf-test/actions/runs/1900613729 for the new CI tests. If you look into the results, you can see that the time to run `pulumi preview` with three trials on each setup was:

Using pulumi v3.25.0, PULUMI_NODEJS_TRANSPILE_ONLY=true: 7s, 8s, 9s
Using pulumi v3.25.0, PULUMI_NODEJS_TRANSPILE_ONLY=false: 20s, 21s, 18s
Using this branch, PULUMI_NODEJS_TRANSPILE_ONLY=true: 7s, 7s, 8s
Using this branch, PULUMI_NODEJS_TRANSPILE_ONLY=false: 17s, 21s, 20s

From this, we can see that even with the ts-node v10, this branch should mantain at least even performance with previous branches. Of note, I would expect that SWC would show it's strength more with larger projects, which I have verified in my company's monorepo, but providing test cases for that is a bit trickier.

sdk/nodejs/cmd/run/run.ts
```diff
@@ -150,39 +150,17 @@ export function run(
...
-        compilerOptions = tsConfig["compilerOptions"] ?? {};
-    } catch (e) {
-        compilerOptions = {};
-    }
+    // Users can optionally disable typechecking when using typescript projects 
+    const transpileOnly = process.env["PULUMI_NODEJS_TRANSPILE_ONLY"] === "true";
 
     if (typeScript) {
         tsnode.register({
+            typeCheck: !transpileOnly,
```
Frassle, line 160: `typeCheck` is a deprecated option to register replaced by `transpileOnly` which is set on the next line.

sdk/nodejs/cmd/run/run.ts
```diff
@@ -150,39 +150,16 @@ export function run(
     }
 
     // If this is a typescript project, we'll want to load node-ts.
-    const typeScript: boolean = process.env["PULUMI_NODEJS_TYPESCRIPT"] === "true";
+    const typeScript = process.env["PULUMI_NODEJS_TYPESCRIPT"] === "true";
 
-    // We provide reasonable defaults for many ts options, meaning you don't need to have a tsconfig.json present
```
Frassle, line 155: I wonder if this needs to be saved for a major change version. There are potentially users who currently use pulumi with no tsconfig.json present and so get the default compiler options we currently set. With this change we are instead going to get ts-nodes default compiler options which I suppose is potentially breaking.

dmattia: Sure, I think https://github.com/pulumi/pulumi/pull/8199 is on the docket for release during the next major release as well, and the changes should build nicely off of each other.

dmattia:
This can be potentially closed in favor of https://github.com/pulumi/pulumi/pull/10122 now that newer versions of ts-node support `swc` as a built-in option: https://github.com/TypeStrong/ts-node/releases/tag/v10.5.0

PR #9051, opened 2022-02-25
improve no project found error message

susanev:
# Description

if someone uses `pulumi stack init` but is not in a pulumi project we display an error and suggest that they may be using `--stack` incorrectly. but we believe that may be less common, and it may be more common that folks are in the wrong directory, or haven't yet created a pulumi project. this updates the error message to include that info. super open to feedback!! also if anyone feels strongly we should not lose the stack flag message, id be happy to keep it in there, just let me know.

## Checklist

### DIFF
--- a/sdk/go/common/workspace/paths.go
+++ b/sdk/go/common/workspace/paths.go
@@ -142,7 +142,7 @@ func DetectProjectAndPath() (*Project, string, error) {
 		return nil, "", err
 	} else if path == "" {
 		return nil, "", errors.Errorf("no Pulumi project found in the current working directory. " +
-			"Move to a directory with a Pulumi project or try creating a project first with `pulumi new`."
+			"Move to a directory with a Pulumi project or try creating a project first with `pulumi new`.")
 	}
 
 	proj, err := LoadProject(path)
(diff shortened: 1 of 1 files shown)


sdk/go/common/workspace/paths.go
```diff
@@ -142,7 +142,7 @@ func DetectProjectAndPath() (*Project, string, error) {
 		return nil, "", err
 	} else if path == "" {
 		return nil, "", errors.Errorf("no Pulumi project found in the current working directory. " +
-			"If you're using the `--stack` flag, make sure to pass the fully qualified name (org/project/stack)")
+			"Move to a directory with a Pulumi project or try creating a project first with `pulumi new`."
```
susanev, line 145:
```suggestion
			"Move to a directory with a Pulumi project or try creating a project first with `pulumi new`.")
```

pgavlin (approved):

joeduffy:
For what it's worth, we explicitly added the org/project/stack language because of the ease of forgetting to include the qualified name; see https://github.com/pulumi/pulumi/pull/7708.

susanev:
@joeduffy ill open up a separate issue, thx!

Frassle:
I think that should be caught in httpstate.ParseStackReference. If we try to auto fill in project and can't find a project then wrap the error with the extra info to fully specify the stack.

PR #9083, opened 2022-03-01
CI: revamp, use an explicit matrix of test commands

AaronFriel:
This vastly increases the number of test jobs, but should have the benefit of ensuring each performs a smaller amount of work and completes more quickly, improving test time when there is low contention for runners.

Work done in this PR, by commit and impact:

Re-enables linting on PRs: the GitHub Action lint job by setting the matrix parameters that steps within the job required; it's now working and the linting for Python is fixed.

Improves readability of the test.yml, test-fast.yml GitHub Action workflows the test matrix an explicit set of shell commands that can be run locally, making it easier to determine which commands to run to reproduce integration test failures.

Improves verification of what test will run for a given makefile by adding a dryrun mode to all go tests. If the environment variable is set `PULUMI_TEST_DRYRUN=true`, go tests will print the command line it would run and exit without running tests.

Improves reliability of running the the monolithic `tests/integration` package by partitioning it into 5 tests, one for each SDK and one for the rest. This is done by generating a `-Run ...` command with regular expressions matching tests for each SDK. You can observe in this in the CLI by running:

```
PULUMI_TEST_DRYRUN=true make test_integration
```

Removes special cases to skip certain tests on Windows. 

Halves the time to run the pkg tests by separating out the `pkg/codegen/nodejs` test into its own job. This package's tests take approximately 15 minutes on their own.

Fixes the sdk/nodejs sxs test by updating the test to work correctly on Linux & Windows, it was testing an obsolete version of Pulumi (=1.6.0) and it was updated and simplified to enable testing multiple versions of TypeScript, presently it tests ~3.7.3, ^3, and ^4.

Fixes Python linting to 100% success.

Improves GitHub Actions caching in all of our jobs by updating setup steps for Python, Go, and Node to enable caching when available.

Adds a heartbeat that logs to the GitHub Actions every 10 seconds - on Windows "heartbeat" and on *nix a nice 💓 emoji.

Lastly, parallelizes almost all of our tests and adds a lint rule to ensure future tests are parallelized. This helps keep Windows jobs in particular within a reasonable time to complete.


AaronFriel:
@t0yv0 in order to help this merge cleanly, I'd want to merge this _and then_ revert these two commits:

* runs tests using the workflow on the branch & uses the "full" test suite on PR tests: https://github.com/pulumi/pulumi/pull/9083/commits/620028bb0da8a8df1c782e520396638101c3e371
* disables MacOS tests: https://github.com/pulumi/pulumi/pull/9083/commits/8331cc4740060892d40698f6e6d12c0928e52bb5

I'd keep the "test-parallelism" change on the 2nd commit though, so just a partial revert there.

As a follow up PR, I think I'd make Windows & Mac integration tests only run on a daily schedule. What do you think?

.github/workflows/build.yml
```diff
@@ -53,8 +53,8 @@ jobs:
     strategy:
       matrix:
         include:
-          - platform: macos-latest
-            os: darwin
+          # - platform: macos-latest
```
t0yv0, line 56: Is this intended to remove macos-latest?

.github/workflows/run-build-and-acceptance-tests.yml
```diff
@@ -136,14 +161,14 @@ jobs:
 
   build:
     name: Build
-    uses: pulumi/pulumi/.github/workflows/build.yml@master
+    uses: pulumi/pulumi/.github/workflows/build.yml@aaronfriel/ci-revamp
```
t0yv0, line 164: Remember to fixup before merging.

.github/workflows/run-build-and-acceptance-tests.yml
```diff
@@ -136,14 +161,14 @@ jobs:
...
     with:
       enable-coverage: true
 
   test:
     name: Test
     needs: build
-    uses: pulumi/pulumi/.github/workflows/test-fast.yml@master
+    uses: pulumi/pulumi/.github/workflows/test.yml@aaronfriel/ci-revamp
```
t0yv0, line 171: Ditto

.github/workflows/test-fast.yml
```diff
@@ -83,17 +114,27 @@ jobs:
...
+          mkdir -p "${{ runner.temp }}/tmp"
       - name: Enable code coverage
         if: ${{ inputs.enable-coverage && (matrix.platform != 'windows-latest') }}
         run: |
           echo "PULUMI_TEST_COVERAGE_PATH=$(pwd)/coverage" >> $GITHUB_ENV
-      - name: Set up Make (windows)
-        if: ${{ matrix.platform == 'windows-latest' }}
-        run: |
-          choco install make
+      # See: https://github.com/actions/virtual-environments/issues/2642#issuecomment-774988591
```
t0yv0, line 130: Interesting

scripts/integration_test_subsets.py
```diff
@@ -0,0 +1,48 @@
+"""Defines test subsets of integration tests.
```
t0yv0, line 1: Ah interesting, so it's regex-based. 

Earlier I was looking at actually moving Go files around so these tests sit in separately named packages. I don't recall why I didn't finish that up but it certainly looked doable.

sdk/go/common/workspace/plugins_test.go
```diff
@@ -379,18 +395,19 @@ func TestPluginDownload(t *testing.T) {
...
-			Name:              "aws",
+			Name:              "mock-latest",
 			Kind:              PluginKind("resource"),
 		}
 		expectedVersion := semver.MustParse("4.37.5")
 		source := info.GetSource()
 		getHTTPResponse := func(req *http.Request) (io.ReadCloser, int64, error) {
 			assert.Equal(t,
-				"https://api.github.com/repos/pulumi/pulumi-aws/releases/latest",
+				"https://api.github.com/repos/pulumi/pulumi-mock-latest/releases/latest",
```
t0yv0, line 410: https://github.com/pulumi/pulumi-mock-latest does not exist but that's OK i guess?

scripts/integration_test_subsets.py
AaronFriel: I refrained because there are an awful lot of relative paths, didn't feel confident (at that point) moving these to separate packages.

.github/workflows/build.yml
AaronFriel: Temporary commit so I don't clog up runners. We should run the full suite at least one time with this change removed.

Separately I'll propose in Slack to move Windows and Mac integration tests to a scheduled job on the master branch.

sdk/nodejs/Makefile
```diff
@@ -52,34 +54,43 @@ install_plugin:: build
...
+	yarn run mocha --timeout 120000 'bin/tests/runtime/closureLoader.spec.js'
+	yarn run nyc -s mocha 'bin/tests_with_mocks/**/*.spec.js'
+
+test_auto:: $(TEST_ALL_DEPS)
+	yarn run nyc -s mocha --timeout 120000 'bin/tests/automation/**/*.spec.js'
+
+TSC_SUPPORTED_VERSIONS = ~3.7.3 ^3 ^4
+
+version=$(subst sxs_test_,,$(word 1,$(subst !, ,$@)))
+sxs_test_%:
```
t0yv0, line 70: These ones may be very slow good they're separate. There's something compiling a C++ node_gyp library as part of Node modules installation.

sdk/go/common/workspace/plugins_test.go
AaronFriel: I'm scratching my head here as to why I did this, but I think it was to make it clear it was a unit test with a fake URL.

sdk/nodejs/cmd/pulumi-language-nodejs/main_test.go
```diff
@@ -4,7 +4,7 @@
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
 //
-//     http://www.apache.org/licenses/LICENSE-2.0
+//     htp://www.apache.org/licenses/LICENSE-2.0
```
t0yv0, line 7: This intentional?

AaronFriel: will fix up!

t0yv0 (approved):
LGTM - skimmed through, left some comments. Only concern is that if we run out of throughput and latency goes through the roof on PRs, we'll need to this big PR out. We can do that though.

scripts/integration_test_subsets.py
t0yv0: Yeah, understood.

sdk/go/common/workspace/plugins_test.go
t0yv0: That's legit.

.github/workflows/build.yml
t0yv0: Excellent. Thanks.

sdk/nodejs/Makefile
AaronFriel: updating the test target from Pulumi 1.6.0 fixed the node_gyp slowness and build error on Windows! 🎉

PR #9098, opened 2022-03-03
Implement state rename command for renaming a resource

### LINKED ISSUE #2060: Add 'pulumi state' command for renaming resources
It's currently difficult and error-prone to rename a resource when editing the checkpoint directly. This scenario has come up a few times and it would be useful to have a `pulumi state` command capable of renaming a resource.

https://github.com/pulumi/pulumi/issues/1909

Zaid-Ajaj:
# Description

Implements `state rename` command. Fixes #2060 

## Checklist

### DIFF
--- a/pkg/cmd/pulumi/state_rename.go
+++ b/pkg/cmd/pulumi/state_rename.go
@@ -1,0 +1,92 @@
+// Copyright 2016-2018, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/pulumi/pulumi/pkg/v3/resource/deploy"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/diag"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/util/cmdutil"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/util/result"
+
+	"github.com/spf13/cobra"
+)
+
+func newStateRenameCommand() *cobra.Command {
+	var stack string
+	var yes bool
+	var force bool
+
+	cmd := &cobra.Command{
+		Use:   "rename <resource URN> <new name>",
+		Short: "Renames a resouce from a stack's state",
+		Long: `Renames a resouce from a stack's state
+
+This command renames a resource from a stack's state. The resource is specified 
+by its Pulumi URN (use ` + "`pulumi stack --show-urns`" + ` to get it) and the new name of the resource.
+
+Make sure that URNs are single-quoted to avoid having characters unexpectedly interpreted by the shell.
+
+Example:
+pulumi state rename 'urn:pulumi:stage::demo::eks:index:Cluster$pulumi:providers:kubernetes::eks-provider' new-name-here
+`,
+		Args: cmdutil.ExactArgs(2),
+		Run: cmdutil.RunResultFunc(func(cmd *cobra.Command, args []string) result.Result {
+			yes = yes || skipConfirmations()
+			urn := resource.URN(args[0])
+			newResourceName := args[1]
+			// Show the confirmation prompt if the user didn't pass the --yes parameter to skip it.
+			showPrompt := !yes
+
+			res := runStateEdit(stack, showPrompt, urn, func(snap *deploy.Snapshot, resource *resource.State) error {
+
+				// the resource is protected but the user didn't use --force
+				if !force && resource.Protect {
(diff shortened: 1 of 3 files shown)


pkg/cmd/pulumi/state_rename.go
```diff
@@ -0,0 +1,93 @@
...
+			yes = yes || skipConfirmations()
+			urn := resource.URN(args[0])
+			newResourceName := args[1]
+			// Show the confirmation prompt if the user didn't pass the --yes parameter to skip it.
+			showPrompt := !yes
+
+			res := runStateEdit(stack, showPrompt, urn, func(snap *deploy.Snapshot, resource *resource.State) error {
+
+				// the resource is protected but the user didn't use --force
+				if !force && resource.Protect {
```
Frassle, line 59: I don't think "Protect" needs to be taken into consideration for renames. Protect is that you won't delete it, not that you won't change anything about it.

pkg/cmd/pulumi/state_rename.go
```diff
@@ -0,0 +1,93 @@
...
+Example:
+pulumi state rename 'urn:pulumi:stage::demo::eks:index:Cluster$pulumi:providers:kubernetes::eks-provider' new-name-here
+`,
+		Args: cmdutil.ExactArgs(2),
+		Run: cmdutil.RunResultFunc(func(cmd *cobra.Command, args []string) result.Result {
+			yes = yes || skipConfirmations()
+			urn := resource.URN(args[0])
+			newResourceName := args[1]
+			// Show the confirmation prompt if the user didn't pass the --yes parameter to skip it.
+			showPrompt := !yes
```
Frassle, line 54: Before running any stateEdit we should do a non-mutating pass of the current state to check the renamed URN isn't already present, and error if it is.

pkg/cmd/pulumi/state_rename.go
```diff
@@ -0,0 +1,93 @@
...
+				}
+
+				if resource.Protect {
+					cmdutil.Diag().Warningf(diag.RawMessage("" /*urn*/, "renaming protected resource due to presence of --force"))
+					resource.Protect = false
+				}
+
+				// update the URN with only the name part changed
+				resource.URN = resource.URN.Rename(newResourceName)
+
```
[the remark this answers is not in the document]
Zaid-Ajaj: is this only a matter of changing `runStateEdit` to `runTotalStateEdit` and it will take care of the rest? I thought I actually needed to look for _dependants_ rather than dependencies since only dependants would reference an existing resource

pkg/cmd/pulumi/state_rename.go
Zaid-Ajaj: That makes sense, I'll try to figure how to do that 👍

pkg/cmd/pulumi/state_rename.go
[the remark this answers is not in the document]
Frassle: Yes anything that is dependent on the resource being renamed will need a fix up. You don't actually need to build a dependency graph though, you can just do a straight forward iterate through everything and replace anywhere you see the old URN with the new URN I think.

pkg/cmd/pulumi/state_rename.go
```diff
@@ -0,0 +1,121 @@
+// Copyright 2016-2018, Pulumi Corporation.
```
Frassle, line 1:
```suggestion
// Copyright 2016-2022, Pulumi Corporation.
```

pkg/cmd/pulumi/state_rename.go
```diff
@@ -0,0 +1,121 @@
...
+
+	cmd := &cobra.Command{
+		Use:   "rename <resource URN> <new name>",
+		Short: "Renames a resource from a stack's state",
+		Long: `Renames a resource from a stack's state
+
+This command renames a resource from a stack's state. The resource is specified 
+by its Pulumi URN (use ` + "`pulumi stack --show-urns`" + ` to get it) and the new name of the resource.
+
+Make sure that URNs are single-quoted to avoid having characters unexpectedly interpreted by the shell.
```
Frassle, line 56: UX ❤️

pkg/cmd/pulumi/state_rename.go
```diff
@@ -0,0 +1,121 @@
...
+
+Example:
+pulumi state rename 'urn:pulumi:stage::demo::eks:index:Cluster$pulumi:providers:kubernetes::eks-provider' new-name-here
+`,
+		Args: cmdutil.ExactArgs(2),
+		Run: cmdutil.RunResultFunc(func(cmd *cobra.Command, args []string) result.Result {
+			yes = yes || skipConfirmations()
+			urn := resource.URN(args[0])
```
Frassle, line 64: Should probably add a check here for `urn.IsValid()` and error out if false

pkg/cmd/pulumi/state_rename.go
```diff
@@ -0,0 +1,121 @@
...
+				inputResource := existingResources[0]
+				oldUrn := inputResource.URN
+				// update the URN with only the name part changed
+				newUrn := oldUrn.Rename(newResourceName)
+				// Check whether the new URN _does not_ correspond to an existing resource
+				candidateResources := edit.LocateResource(snap, newUrn)
+				if len(candidateResources) > 0 {
+					return errors.New("The chosen new name for the state corresponds to an already existing resource")
+				}
+
```
Frassle, line 85: Ideally the checks above could come before the user prompt that we're going to modify state. I think this is fine for this PR but would be nice to have a think if there's a way to keep the code this simple but split the check/prompt/do work up.

Frassle (approved):
Couple of minor changes to do but otherwise LGTM

PR #9101, opened 2022-03-03
Remove type test from `pulumi console`

Frassle:
Move ConsoleURL to just httpstate.Backend and don't type test on httpstate.Stack.


Zaid-Ajaj (approved):
Ready to go 🚀

PR #9105, opened 2022-03-03
Move `StackIdentifier` to httpstate.Backend

Frassle:
The stack method just called into the backend method anyway and this lets us remove a few more type tests where we can just check for httpstate.Backend not for that and httpstate.Stack.

Eventually all StackRefernces will be the same so there won't even be any chance of an error from passing the wrong type of StackReference to StackIdentifier.


Frassle:
Not keen on this actually. Going to rethink another way to remove these type tests.

PR #9113, opened 2022-03-04
Don't type test for stack tags in pulumi new

Frassle:
Move Tags to just be on Stack. Saves a type test in `pulumi new`.

Given Tags doesn't actually do a fetch the only error you can really get back from it is that the backed doesn't support tags.


Zaid-Ajaj (approved):

pkg/cmd/pulumi/new.go
```diff
@@ -508,8 +507,10 @@ func getStack(stack string, opts display.Options) (backend.Stack, string, string
 	name := ""
 	description := ""
 	if s != nil {
-		if cs, ok := s.(httpstate.Stack); ok {
-			tags := cs.Tags()
+		tags, err := s.Tags()
```
iwahbe, line 510: Could we check for that error and propagate all others. If a new error is added, or something unexpected happens, I don't want to silently eat errors. 

If the only possible error for `Tags` is "not supported", maybe change the return type from `error` to a bool indicating if tags are supported.

iwahbe (changes requested):

iwahbe (approved):

PR #9119, opened 2022-03-04
Debug print incoming protos

t0yv0:
# Description

Fixes # (issue)

## Checklist

### DIFF
--- a/sdk/go/common/resource/plugin/rpc.go
+++ b/sdk/go/common/resource/plugin/rpc.go
@@ -19,7 +19,9 @@ import (
 	"reflect"
 	"sort"
 
+	proto "github.com/golang/protobuf/proto"
 	structpb "github.com/golang/protobuf/ptypes/struct"
+
 	"github.com/pkg/errors"
 
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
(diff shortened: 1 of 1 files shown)


Frassle:
Huh I thought everywhere already logged these.

t0yv0:
Perhaps I'm not sure where to look? How do you look at these?

Sharing this without intention to merge as debug help for @dixler

t0yv0:
If I recall correctly, the default stuff I saw printed in the logs is a lossy form it does not distinguish if "true" is a boolean or string literal "true" and it caused some insanity for our debugging; this form is very explicit.

Frassle:
Ah right yeh a few lines down from your change there's:
```
logging.V(9).Infof("Unmarshaling property for RPC[%s]: %s=%v", opts.Label, key, v)
```
but that's the value after unmarshaling, not the protobuf value.

I think there's value in tidying this up so we can see the proto value and the unmarshaled value.

Maybe changing that logging to something like:
```
protoValue := props.Fields[key]
logging.V(9).Infof("Unmarshaling property for RPC[%s]: %s=%v", opts.Label, key, proto.MarshalTextString(protoValue))
v, err := UnmarshalPropertyValue(pk, protoValue, opts)
		if err != nil {
			return nil, err
		} else if v != nil {
			logging.V(9).Infof("Unmarshaled property for RPC[%s]: %s=%v", opts.Label, key, v)
...
```

t0yv0:
Closing in favor of https://github.com/pulumi/pulumi/pull/10193

Issue #9033, opened 2022-02-23
pulumi stack disappears during a pulumi up when using the GCP bucket backend


jsravn:
## Hello!

- Vote on this issue by adding a 👍 reaction
- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details

During a `pulumi up` the stack effectively disappears from the stack list for other users. `pulumi stack select <stack>` fails saying stack does not exist. Similarly StackReferences can't look up output values, again with a not found error.

At minimum, it should return an error that the stack is in use. At best, it should support retrieving output values from the last finished run.

### Steps to reproduce

1. Create a pulumi project with a GCP storage backend.
2. Try doing pulumi up
3. Try to do a `pulumi stack output -s <stackname>` while the up is running

Expected: Should find the stack and print its outputs, or report the stack is currently locked (not ideal though, for reads).
Actual: Pulumi reports the stack can't be found.

mikhailshilkov:
cc @Frassle to keep on your radar for the local backends improvements

silasdavis:
I'm not sure what conclusions to draw from this but it feels like I need to rethink using this in production. Experiencing this locally. Is there a recovery?

Frassle:
Reading the issue the recovery sounds like wait for the `up` to finish. I'm working on the local backend at the moment and I 
 think our lock code needs a look at anyway so I'll keep this mind while looking at that. See if I can figure out how this could report as just missing, as @jsravn said at the least this should pick up the lock file.

Frassle:
OK I think what's happening here is two fold. Firstly when an update is running we repeatedly save out the snapshot file but that's done in three steps:
1. Copy "<stack>.json" to "<stack>.json.bak" overwriting if it already exists
2. Delete "<stack>.json"
3. Write a new "<stack>.json"

Secondly we always check for a "<stack>.json" before looking for a lock file. So if your trying to lookup stack outputs between step 2 and 3 above the backend just thinks the stack doesn't exist (because there's no "<stack>.json").

I think we can just stop doing that delete operation and let the write just overwrite it instead (which it looks like go-cloud does atomically) so the "<stack>.json" file will always exists allowing other bits of the system to pick up on it.

PR #9122, opened 2022-03-05
Ensure filestate stack.json file are atomically replaced

Frassle:
# Description

When saving a new version of the stack checkpoint we don't delete the current file, instead we let WriteAll atomically replace it. This ensures that if any concurrent attempts are made to list or read the stack while the write is happening they can still pick up the existing about to be replaced version.

Also inlined the renaming logic to `backupObject` and so can now log a warning if the backup copy fails.

Fixes https://github.com/pulumi/pulumi/issues/9033

## Checklist

### DIFF
--- a/pkg/backend/filestate/backend.go
+++ b/pkg/backend/filestate/backend.go
@@ -423,7 +423,7 @@ func (b *localBackend) RenameStack(ctx context.Context, stack backend.Stack,
 
 	// To remove the old stack, just make a backup of the file and don't write out anything new.
 	file := b.stackPath(stackName)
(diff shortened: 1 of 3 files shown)


mikhailshilkov (approved):
This makes sense to me. Probably a larger topic than this change - but should we start adding some test harness around file backends that could cover changes like this one?

Frassle:
> but should we start adding some test harness around file backends that could cover changes like this one?

Yes we should! I did look at if it made sense to write a test for this specific case but it's very very racy so pretty much always passed even without the fix, so I didn't deem it useful enough to keep.

PR #9127, opened 2022-03-06
allow to override makefile variables

teto:
# Description

I want to package the pulumi python sdk for my linux distribution www.nixos.org. 
One difficulty is that the shell is not in /bin/bash on that system (there is only /bin/sh, bash is at /nix/store/3pa0xk3mgmx7hqskg63gxviyw7f217i6-bash-interactive-5.1-p12/bin/bash for instance)
When running the tests I would like to be able to override the SHELL value without having to patch all the Makefiles, hence the change.

I also added a way to bypass the `pulumictl get version`. I dont know how it works but it seems like it needed a git repo ? I already know the version from the nix package so I wanted to skip that altogether.

Fixes # (issue)

## Checklist

### DIFF
--- a/build/common.mk
+++ b/build/common.mk
@@ -90,7 +90,7 @@
 #
 # Importing common.mk should be the first thing your Makefile does, after
 # optionally setting SUB_PROJECTS, PROJECT_NAME and NODE_MODULE_NAME.
-SHELL       := /bin/bash
+SHELL       ?= /bin/bash
 .SHELLFLAGS := -ec
 
 STEP_MESSAGE = @echo -e "\033[0;32m$(shell echo '$@' | tr a-z A-Z | tr '_' ' '):\033[0m"
--- a/sdk/go/Makefile
+++ b/sdk/go/Makefile
@@ -1,6 +1,6 @@
 PROJECT_NAME     := Pulumi Go SDK
 LANGHOST_PKG     := github.com/pulumi/pulumi/sdk/v3/go/pulumi-language-go
-VERSION 		 := $(if ${PULUMI_VERSION},${PULUMI_VERSION}, $(shell cd ../../ && pulumictl get version))
+VERSION                 := $(if ${PULUMI_VERSION},${PULUMI_VERSION}, $(shell cd ../../ && pulumictl get version))
 TEST_FAST_PKGS   := $(shell go list ./pulumi/... ./pulumi-language-go/... ./common/... | grep -v /vendor/ | grep -v templates)
 TEST_AUTO_PKGS   := $(shell go list ./auto/... | grep -v /vendor/ | grep -v templates)
 TESTPARALLELISM  := 10
--- a/sdk/go/Makefile
+++ b/sdk/go/Makefile
@@ -1,6 +1,6 @@
 PROJECT_NAME     := Pulumi Go SDK
 LANGHOST_PKG     := github.com/pulumi/pulumi/sdk/v3/go/pulumi-language-go
-VERSION          := $(shell cd ../../ && pulumictl get version)
+VERSION 		 := $(if ${PULUMI_VERSION},${PULUMI_VERSION}, $(shell cd ../../ && pulumictl get version))
 TEST_FAST_PKGS   := $(shell go list ./pulumi/... ./pulumi-language-go/... ./common/... | grep -v /vendor/ | grep -v templates)
 TEST_AUTO_PKGS   := $(shell go list ./auto/... | grep -v /vendor/ | grep -v templates)
 TESTPARALLELISM  := 10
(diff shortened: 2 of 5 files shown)


build/common.mk
```diff
@@ -105,9 +105,9 @@ endif
 PYTHON ?= python3
 PIP ?= pip3
 
-PULUMI_BIN          := $(PULUMI_ROOT)/bin
-PULUMI_NODE_MODULES := $(PULUMI_ROOT)/node_modules
-PULUMI_NUGET        := $(PULUMI_ROOT)/nuget
+PULUMI_BIN          ?= $(PULUMI_ROOT)/bin
```
Frassle, line 108: These should all be created in PULUMI_ROOT which can already be overriden. I don't see the need for the user to be able to set these aswell?

sdk/python/Makefile
```diff
@@ -1,7 +1,7 @@
 PROJECT_NAME     := Pulumi Python SDK
 LANGHOST_PKG     := github.com/pulumi/pulumi/sdk/v3/python/cmd/pulumi-language-python
-VERSION          := $(shell cd ../../ && pulumictl get version)
-PYPI_VERSION 	 := $(shell cd ../../ && pulumictl get version --language python)
+VERSION          ?= $(shell cd ../../ && pulumictl get version)
```
Frassle, line 3: Ah this used to be supported in the get-version script (https://github.com/pulumi/pulumi/blob/master/scripts/get-version#L14) but wasn't translated over to pulumictl. I think the right fix here is that https://github.com/pulumi/pulumictl/blob/master/cmd/pulumictl/version/cli.go should be updated to check for the PULUMI_VERSION envvar.

build/common.mk
teto: Indeed I removed that.

sdk/python/Makefile
teto: I dont know go so I would rather not patch pulumictl. I am in no hurry so I can wait for an external patch. Meanwhile is it that bad to allow the override here ? Thanks for reviewing this btw.

Frassle: I think we'd want to write this such that PYPI_VERSION (and likewise for the other languages) will either use the expected repo version, or the same string as VERSION. And then there's the question of if we want to call the envvar VERSION or if we should have what get-version script has PULUMI_VERSION.

Frassle: Given the only thing the --language option is doing in pulumictl is handling some edge cases of pypi and pre-release versions I think it might be ok to just have the override login here in the make files.

So VERSION should be set to PULUMI_VERSION if defined, else to shelling out to pulumictl. Likewise for things like PYPI_VERSION and DOTNET_VERSION in the other makefiles.

sdk/python/Makefile
```diff
@@ -1,7 +1,7 @@
 PROJECT_NAME     := Pulumi Python SDK
 LANGHOST_PKG     := github.com/pulumi/pulumi/sdk/v3/python/cmd/pulumi-language-python
-VERSION          := $(shell cd ../../ && pulumictl get version)
-PYPI_VERSION 	 := $(shell cd ../../ && pulumictl get version --language python)
+VERSION          := $(if ${PULUMI_VERSION},${PULUMI_VERSION}, $(shell cd ../../ && pulumictl get version))
```
Frassle, line 3: Can you make a similar change to the other Makefiles that set VERSION?

sdk/nodejs/Makefile
```diff
@@ -1,6 +1,6 @@
 PROJECT_NAME     := Pulumi Node.JS SDK
 NODE_MODULE_NAME := @pulumi/pulumi
-VERSION 		 := $(shell cd ../../ && pulumictl get version --language javascript)
+VERSION 		 := $(if ${PULUMI_VERSION},${PULUMI_VERSION}, $(shell cd ../../ && pulumictl get version --language javascript))
```
Frassle, line 3:
```suggestion
VERSION                        := $(if ${PULUMI_VERSION},${PULUMI_VERSION}, $(shell cd ../../ && pulumictl get version --language javascript))
```

sdk/go/Makefile
```diff
@@ -1,6 +1,6 @@
 PROJECT_NAME     := Pulumi Go SDK
 LANGHOST_PKG     := github.com/pulumi/pulumi/sdk/v3/go/pulumi-language-go
-VERSION          := $(shell cd ../../ && pulumictl get version)
+VERSION 		 := $(if ${PULUMI_VERSION},${PULUMI_VERSION}, $(shell cd ../../ && pulumictl get version))
```
Frassle, line 3:
```suggestion
VERSION                 := $(if ${PULUMI_VERSION},${PULUMI_VERSION}, $(shell cd ../../ && pulumictl get version))
```

Frassle: Don't use tabs.

teto:
Is it good then ?

Frassle (approved):
Seems sane to me, we used to support PULUMI_VERSION for other packing systems so no reason not to add it back.

PR #9132, opened 2022-03-07
Remove last few uses of get-version script

Frassle:
VERSION had already been changed to be set via pulumictl instead of the local get-version script a while ago. This cleans up the last few references to that local script to instead use pulumictl and removes the get-version script.


AaronFriel (approved):
Looks good to me, but a cursory google and read of [Homebrew's versioning docs](https://docs.brew.sh/Versions) didn't say anything about semver, prerelease versions in particular.

Approving nonetheless as IIRC, we don't ever publish prerelease versions.

Frassle:
> Approving nonetheless as IIRC, we don't ever publish prerelease versions.

Yeh I think we would of hit this already with the get-version script. But if we do need to support this probably easier to add to pulumictl then more bash magic in get-version!

PR #9140, opened 2022-03-08
testing: allow mock contexts to get resource options

AaronFriel:
For use in other provider / plugin projects, allows unit testing that resource options such as parent, provider, etc. are passed through correctly.


iwahbe (approved):
LGTM

Issue #9134, opened 2022-03-07
Add a flag to disable code generation to pulumi import


jaxxstorm:
Getting Pulumi into the state is easy with `pulumi import` and we rather handily generate the code you need to add to your stack.

As per this comment: https://github.com/pulumi/pulumi/issues/6032#issuecomment-1059386589

It would be useful to have the mechanism to not generate the code when running `pulumi import` in some situations

Frassle:
I think we should solve the problem in #6032 in a better way as well, but this is a really easy flag to just add.

mikhailshilkov:
What's the downside of code generation that users need to disable it?

Frassle:
> What's the downside of code generation that users need to disable it?

In #6032 its that they wanted to use the same name for resources of different types, but our PCL binder didn't like that because it also uses the given name for variable names which need to be unique. @pgavlin had some thoughts on how to best allow that situation properly.

So we should fix that properly, but the user also requested if they could just turn the code generation off because they don't really need it.

PR #9141, opened 2022-03-08
Add a --generate-code flag to import

Frassle:
# Description

Adds a flag `--generate-code` to pulumi import. By default this is true, but users can pass `--generate-code=false` to skip the code generation step of import.

Fixes https://github.com/pulumi/pulumi/issues/9134

## Checklist

### DIFF
--- a/pkg/cmd/pulumi/import.go
+++ b/pkg/cmd/pulumi/import.go
@@ -380,7 +380,7 @@ func newImportCmd() *cobra.Command {
 			}
 
 			if !generateCode && outputFilePath != "" {
-				fmt.Print("Output file will not be used as --generate-code is false.\n")
+				fmt.Fprintln(os.Stderr, "Output file will not be used as --generate-code is false.")
 			}
 
 			var outputResult bytes.Buffer
--- a/pkg/cmd/pulumi/import.go
+++ b/pkg/cmd/pulumi/import.go
@@ -279,6 +279,7 @@ func newImportCmd() *cobra.Command {
 	var providerSpec string
 	var importFilePath string
 	var outputFilePath string
+	var generateCode bool
 
 	var debug bool
 	var message string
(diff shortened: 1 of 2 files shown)


pkg/cmd/pulumi/import.go
```diff
@@ -496,37 +501,39 @@ func newImportCmd() *cobra.Command {
 				Scopes:             cancellationScopes,
 			}, imports)
 
-			deployment, err := getCurrentDeploymentForStack(s)
-			if err != nil {
-				return result.FromError(err)
-			}
+			if generateCode {
```
pgavlin, line 504: Minor nit: maybe pull the body of this `if` out into its own function? Normally I'd suggest using an early return here to be more idiomatic and avoid the rightward movement, but that isn't convenient with the `res != nil` bits later.

pkg/cmd/pulumi/import.go
```diff
@@ -378,6 +379,10 @@ func newImportCmd() *cobra.Command {
 				importFile = f
 			}
 
+			if !generateCode && outputFilePath != "" {
+				return result.Errorf("an output file may not be specified in conjunction with --generate-code=false")
+			}
```
pgavlin, line 384: Nit: seems like we could just make this a warning rather than an error.

pgavlin (approved):
Two nits, LGTM otherwise. Thanks for turning this around so fast!

pkg/cmd/pulumi/import.go
```diff
@@ -378,6 +379,10 @@ func newImportCmd() *cobra.Command {
 				importFile = f
 			}
 
+			if !generateCode && outputFilePath != "" {
+				fmt.Print("Output file will not be used as --generate-code is false.\n")
```
iwahbe, line 383: I would print this message to stderr.

pkg/cmd/pulumi/import.go
Frassle: I'd generally lean towards keeping the code inline here, we're just hitting 120 cols which doesn't seem unreasonable.

iwahbe (approved):

Issue #8476, opened 2021-11-22
Update grpc to grpc-dotnet


emiliza:
For .NET6 support we'll need to consider updating to  grpc-dotnet v2.41.0-pre1 or later - this should also fix the issue for missing libgrpc for Apple silicon, which https://github.com/pulumi/pulumi/pull/7527 addresses.

Zaid-Ajaj:
Resolved as of #9149

Issue #8447, opened 2021-11-18
Dotnet 6 issues on m1 & m1max cpus - running pulumi up fails


stejsj:
## Hello!

- Vote on this issue by adding a 👍 reaction
- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details

We are unable to run pulumi up on our infrastructure projects on our new macbook pro m1max machines with dotnet 6 installed via homebrew. 

We get the following error:

Type                 Name                                  Plan     Info
     pulumi:pulumi:Stack  services-orchestration-dev-stejsj-t2           1 error; 5 messages

Diagnostics:
  pulumi:pulumi:Stack (services-orchestration-dev-stejsj-t2):
    A fatal error occurred. The required library libhostfxr.dylib could not be found.
    If this is a self-contained application, that library should exist in [/Users/sejensen/dev/chipper-orchestration/infra/bin/Debug/netcoreapp3.1/].
    If this is a framework-dependent application, install the runtime in the global location [/usr/local/share/dotnet/x64] or use the DOTNET_ROOT environment variable to specify the runtime location or register the runtime location in [/etc/dotnet/install_location].
    The .NET Core runtime can be found at:
      - https://aka.ms/dotnet-core-applaunch?missing_runtime=true&arch=x64&rid=osx.10.17-x64

    error: an unhandled error occurred: Program exited with non-zero exit code: 131
--------

Setting DOTNET_ROOT results in this error instead
  pulumi:pulumi:Stack (services-orchestration-dev-stejsj-t2):
    Failed to load A, error: dlopen(/usr/local/share/dotnet/host/fxr/6.0.0/libhostfxr.dylib, 0x0001): tried: '/usr/local/share/dotnet/host/fxr/6.0.0/libhostfxr.dylib' (mach-o file, but is an incompatible architecture (have 'arm64', need 'x86_64')), '/usr/local/lib/libhostfxr.dylib' (no such file), '/usr/lib/libhostfxr.dylib' (no such file)
    The library libhostfxr.dylib was found, but loading it from /usr/local/share/dotnet/host/fxr/6.0.0/libhostfxr.dylib failed
      - Installing .NET Core prerequisites might help resolve this problem.
         https://go.microsoft.com/fwlink/?linkid=2063366

    error: an unhandled error occurred: Program exited with non-zero exit code: 130

Expected: being able to run pulumi up with dothet 6 and m1 or m1 max cpus
Actual: fails to run

[...]

PR #9142, opened 2022-03-08
Migrating the dotnet sdk to net6.0

Zaid-Ajaj:
# Description

 - #8476
 - #8447

First of all, replaces `Grpc` package which is now in maintenance mode (until 2022 May) with `grpc-dotnet`, specifically using the `Grpc.Net.Client` package instead. 

We are still using the `Grpc.Tools` package as the main way to generate C# types and clients from our proto files. Though updated this one from 2.37.0 to 2.44.0 (latest)

Resolved remaining compile-time issues when respect to nullable references as a result of using APIs from `net6.0` target framework. 

Using a local copy of updated `Pulumi` nuget package and a local, updated version of `Pulumi.Aws` package, I've managed to successfully deploy resources to AWS using `pulumi up` and a console application also running on `net6.0` target framework.

Finally, for the CI I've introduced a new input called `dotnet-package-version` which is the version of the Pulumi packages to be published to nuget, rather than using `DOTNET_VERSION` which is the dotnet toolchain version

## Checklist

### DIFF
--- a/sdk/dotnet/Pulumi/Pulumi.csproj
+++ b/sdk/dotnet/Pulumi/Pulumi.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
     <Nullable>enable</Nullable>
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
     <Authors>Pulumi</Authors>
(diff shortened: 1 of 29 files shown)


sdk/dotnet/Pulumi/Pulumi.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
```
mikhailshilkov, line 4: Does it mean that .NET Core 3.1 and .NET 5 users won't be able to use Pulumi anymore? Or will that still work?

Zaid-Ajaj: Yeah this is a breaking change. As discussed, i will be reverting this to maintain backward-compat

PR #9147, opened 2022-03-08
Fix RetainOnDelete for Go SDK

jdartigalongue:
# Description
This MR fixes a bug in the Go SDK where the RetainOnDelete configuration is not stored in the state file and therefore, not applied. 

## Checklist
- [x] I have updated the [CHANGELOG-PENDING]


mattfenwick:
+1 nice one @jdartigalongue !!

PR #9149, opened 2022-03-09
Use grpc-dotnet instead of grpc

Zaid-Ajaj:
# Description

 - #8476
 - #8447

First of all, replaces `Grpc` package which is now in maintenance mode (until 2022 May) with `grpc-dotnet`, specifically using the `Grpc.Net.Client` package instead. 

We are still using the `Grpc.Tools` package as the main way to generate C# types and clients from our proto files. Though updated this one from 2.37.0 to 2.44.0 (latest)

Using a local copy of updated `Pulumi` nuget package and a local, updated version of `Pulumi.Aws` package, I've managed to successfully deploy resources to AWS using `pulumi up` and a console application also running on `net6.0` target framework on my M1 machine.

## Checklist


CHANGELOG_PENDING.md
```diff
@@ -18,6 +18,8 @@
 - [cli/import] - Code generation in `pulumi import` can now be disabled with the `--generate-code=false` flag.
   [#9141](https://github.com/pulumi/pulumi/pull/9141)
 
+- [language/dotnet] - updated Pulumi dotnet packages to use grpc-dotnet instead of grpc [#9149](https://github.com/pulumi/pulumi/pull/9149)
```
Frassle, line 21:
```suggestion
- [language/dotnet] - Updated Pulumi dotnet packages to use grpc-dotnet instead of grpc.
   [#9149](https://github.com/pulumi/pulumi/pull/9149)
```

sdk/dotnet/Pulumi/Deployment/GrpcEngine.cs
```diff
@@ -3,6 +3,7 @@
 using System.Collections.Generic;
 using System.Threading.Tasks;
 using Grpc.Core;
```
Frassle, line 5: This isn't the old package reference still being used via the Tools dep is it?

Frassle (approved):

sdk/dotnet/Pulumi/Deployment/GrpcEngine.cs
Zaid-Ajaj: I think so, i will remove it

sdk/dotnet/Pulumi/Deployment/GrpcEngine.cs
```diff
@@ -1,22 +1,40 @@
...
+using Grpc.Net.Client;
 using Pulumirpc;
 
 namespace Pulumi
 {
     internal class GrpcEngine : IEngine
     {
         private readonly Engine.EngineClient _engine;
+        private static GrpcChannel? _engineChannel = null;
+        private readonly object _engineChannelLock = new object();
```
justinvp, line 16: Why is `_engineChannel` static? Should `_engineChannelLock` be static too? (Same questions for `GrpcMonitor` below).

Zaid-Ajaj: @justinvp According to the docs, creating a `GrpcChannel` is expensive whereas creating the gRPC client that uses the channel is not: hence I was keeping a single instance of the channel to be used across client instances to avoid recreating the channels every time. 

As for `_engineChannelLock` it doesn't need to be static since it is only used for locking. 

The lock however is no longer needed! I refactored the code to use a (concurrent) dictionary to keep track of the channels created by different Engine and Monitor addresses. This dictionary is thread-safe and doesn't need special locking

sdk/dotnet/Pulumi/Deployment/GrpcEngine.cs
```diff
@@ -1,22 +1,42 @@
...
+using Grpc.Net.Client;
 using Pulumirpc;
 
 namespace Pulumi
 {
     internal class GrpcEngine : IEngine
     {
         private readonly Engine.EngineClient _engine;
+        // Using a static dictionary to keep track of and re-use gRPC channels
+        // According to the docs, creating GrpcChannels is expensive so we keep track of a bunch of them here
```
Frassle, line 18:
```suggestion
        // According to the docs(https://docs.microsoft.com/en-us/aspnet/core/grpc/performance?view=aspnetcore-6.0#reuse-grpc-channels), creating GrpcChannels is expensive so we keep track of a bunch of them here
```

sdk/dotnet/Pulumi/Deployment/GrpcMonitor.cs
```diff
@@ -1,22 +1,37 @@
...
         private readonly ResourceMonitor.ResourceMonitorClient _client;
-
+        private static ConcurrentDictionary<string, GrpcChannel> _monitorChannels = new ConcurrentDictionary<string, GrpcChannel>();
         public GrpcMonitor(string monitor)
         {
             // maxRpcMessageSize raises the gRPC Max Message size from `4194304` (4mb) to `419430400` (400mb)
             var maxRpcMessageSize = 400 * 1024 * 1024;
-            var grpcChannelOptions = new List<ChannelOption> { new ChannelOption(ChannelOptions.MaxReceiveMessageLength, maxRpcMessageSize)};
-            this._client = new ResourceMonitor.ResourceMonitorClient(new Channel(monitor, ChannelCredentials.Insecure, grpcChannelOptions));
+            if (!_monitorChannels.ContainsKey(monitor))
```
Frassle, line 21: Wonder if we should have a "GrpcPool" class to share between GrpcEngine and GrpcMonitor.

sdk/dotnet/Pulumi/Deployment/GrpcEngine.cs
Zaid-Ajaj: Added links to the docs, thanks!

sdk/dotnet/Pulumi/Deployment/GrpcMonitor.cs
Zaid-Ajaj: i think for this use case it is simple enough to keep it as imo

sdk/dotnet/Pulumi/Deployment/GrpcEngine.cs
```diff
@@ -1,22 +1,42 @@
...
 using Pulumirpc;
 
 namespace Pulumi
 {
     internal class GrpcEngine : IEngine
     {
         private readonly Engine.EngineClient _engine;
+        // Using a static dictionary to keep track of and re-use gRPC channels
+        // According to the docs (https://docs.microsoft.com/en-us/aspnet/core/grpc/performance?view=aspnetcore-6.0#reuse-grpc-channels), creating GrpcChannels is expensive so we keep track of a bunch of them here
+        private static ConcurrentDictionary<string, GrpcChannel> _engineChannels = new ConcurrentDictionary<string, GrpcChannel>();
```
justinvp, line 19: Nit: `readonly`

sdk/dotnet/Pulumi/Deployment/GrpcEngine.cs
```diff
@@ -1,22 +1,42 @@
...
+                // Inititialize the engine channel once for this address
+                _engineChannels[engineAddress] = GrpcChannel.ForAddress(new Uri($"http://{engineAddress}"), new GrpcChannelOptions
+                {
+                    MaxReceiveMessageSize = maxRpcMessageSize,
+                    MaxSendMessageSize = maxRpcMessageSize,
+                    Credentials = Grpc.Core.ChannelCredentials.Insecure,
+                });
+            }
+
+            this._engine = new Engine.EngineClient(_engineChannels[engineAddress]);
```
justinvp, line 39: Should we use `GetOrAdd`?

```c#
var engineChannel = _engineChannels.GetOrAdd(engineAddress, address =>
{
    // maxRpcMessageSize raises the gRPC Max Message size from `4194304` (4mb) to `419430400` (400mb)
    var maxRpcMessageSize = 400 * 1024 * 1024;

    // Allow for insecure HTTP/2 transport (only needed for netcoreapp3.x)
    // https://docs.microsoft.com/en-us/aspnet/core/grpc/troubleshoot?view=aspnetcore-6.0#call-insecure-grpc-services-with-net-core-client
    AppContext.SetSwitch("System.Net.Http.SocketsHttpHandler.Http2UnencryptedSupport", true);
    // Inititialize the engine channel once for this address
    return GrpcChannel.ForAddress(new Uri($"http://{address}"), new GrpcChannelOptions
    {
        MaxReceiveMessageSize = maxRpcMessageSize,
        MaxSendMessageSize = maxRpcMessageSize,
        Credentials = Grpc.Core.ChannelCredentials.Insecure,
    });
});

_engine = new Engine.EngineClient(engineChannel);
```

Zaid-Ajaj: That would be another way of implementing it. `GetOrAdd` does exactly the same as the above: checks if it exists and return it or if the key does not exist, adds to the dictionary. Do you want me to update the code with `GetOrAdd` instead?

Issue #8747, opened 2022-01-13
Creating a new project with the `-y` arg when the project name already exists doesn't return helpful error message


pierskarsenbarg:
## Hello!

- Vote on this issue by adding a 👍 reaction
- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details

If I already have a project with the name `project-123` and I create an empty folder with the same name and run `pulumi new {provider}-{language} -y` then I get an error message `Sorry, '' is not a valid project name. A project name may not be empty.`

### Steps to reproduce

1. Create a new folder called `project-123`
2. Change to that folder and run `pulumi new typescript -y` (creates a project called `project-123` because of the folder name)
3. Change to the parent folder (`cd ..`)
4. Delete folder called `project-123` and create a new one with the same name
5. Change to that folder and run `pulumi new typescript -y` again
6. You'll get the error message

Expected:  An error message suggesting that a project with that name already exists
Actual: 

If you were to just run `pulumi new typescript` a second time (without the `-y` argument) then you'd see that Pulumi asks you for a project name and doesn't allow you to just hit enter, because it's detected that the project already exists.

nimbinatus:
Literally just ran into this when doing a video 😅

PR #9156, opened 2022-03-09
Fix error message in `pulumi new`

Frassle:
# Description

When running with --yes and letting pulumi select the default project name if it was invalid you got an error message like:
`Sorry, '' is not a valid project name.`

This fixes that case to instead return the actual project name that caused the issue.

Fixes https://github.com/pulumi/pulumi/issues/8747

## Checklist


mikhailshilkov (approved):
LGTM

Issue #9120, opened 2022-03-04
Support self-signed certs for `pulumi login`


MitchellGerdisch:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

When deploying a self-hosted Pulumi Service for testing, I may want to just use a self-signed cert.
But, `pulumi login https://MY_API_ENDPOINT` does not work.
It throws: 
```
error: problem logging in: getting user info from https://api.mitch.pulumi-ce.team: performing HTTP request: Get "https://api.mitch.pulumi-ce.team/api/user": x509: certificate signed by unknown authority
```

Frassle:
Figured this would be a good test case for https://github.com/pulumi/pulumi/issues/4665.

Considering adding something like the below to Pulumi.yaml.
```
name: ssl_test
runtime: nodejs
backend:
  url: https://localhost:8080
  config:
    service:disable-tls: true
```

Of course that would depend on you having a project file in your working directory. I wonder if we need some other way of passing config that lives above projects for setting backend settings?

lukehoban:
I'm not sure we want to put this in the project file.  This is configuration about the backend itself - which we generally want to set machine-scoped.  To the extent we want to support backend configuration explicitly, I think we actually want to support it at the stack level, vs. the project level (since it's really stacks that live in backends, not projects).

I'd be inclined to support this as a flag to `pulumi login`, and to then serialize this into the `~/.pulumi/credentials.json` metadata associated with the backend for use in other backend operations.

Frassle:
> This is configuration about the backend itself - which we generally want to set machine-scoped.

Project files already have a backend.URL configuration. Is the problems that's trying to solve also solved by making backends machine-wide? Or is this just that configuring a given backend URL is machine-wide but then it makes sense for individual projects to opt-in to which configured backend they use?

> I think we actually want to support it at the stack level

We'd have to load stack.json files before building a backend to parse the stack reference in that case. Which I think we could hack around today, but given issues like https://github.com/pulumi/pulumi/issues/2005#issuecomment-431913734 we'd have to be able to tell if "foo/bar" is intended to be owner "foo" or project "foo" and that varies based on if the backend supports organisations or not.

lukehoban:
> Is the problems that's trying to solve also solved by making backends machine-wide?

[...]

PR #9159, opened 2022-03-09
Add --insecure option to `pulumi login`

Frassle:
# Description

Add --insecure option to `pulumi login` which disables https certificate checks.

Fixes https://github.com/pulumi/pulumi/issues/9120

## Checklist

### DIFF
--- a/pkg/cmd/pulumi/login.go
+++ b/pkg/cmd/pulumi/login.go
@@ -35,6 +35,7 @@ func newLoginCmd() *cobra.Command {
 	var cloudURL string
 	var defaultOrg string
 	var localMode bool
+	var insecure bool
 
 	cmd := &cobra.Command{
 		Use:   "login [<url>]",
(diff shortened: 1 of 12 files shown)


lukehoban (approved):
Is there any reasonable way to add test coverage for this?

pkg/cmd/pulumi/login.go
```diff
@@ -171,6 +172,7 @@ func newLoginCmd() *cobra.Command {
 	cmd.PersistentFlags().StringVar(&defaultOrg, "default-org", "", "A default org to associate with the login. "+
 		"Please note, currently, only the managed and self-hosted backends support organizations")
 	cmd.PersistentFlags().BoolVarP(&localMode, "local", "l", false, "Use Pulumi in local-only mode")
+	cmd.PersistentFlags().BoolVarP(&localMode, "insecure", "k", false, "Allow insecure server connections when using SSL")
```
justinvp, line 175:
```suggestion
	cmd.PersistentFlags().BoolVarP(&insecure, "insecure", "k", false, "Allow insecure server connections when using SSL")
```

justinvp, line 175: Also, why `-k` for the shorthand char? Do we even need a shorthand for this one?

justinvp (changes requested):
I was thinking of force pushing a squashed commit to your branch that addresses my suggestions (fixes the bug, and removes the `-k` shorthand), and adds a changelog, to get it merged today. We can follow-up on adding test coverage.

justinvp (approved):
I force pushed a squashed commit w/ feedback addressed.

justinvp:
bors merge

PR #9179, opened 2022-03-11
Implement pulumi stack unselect

### LINKED ISSUE #9070: pulumi stack select: no way to unselect
## Hello!
Once i have `pulumi stack select` (via explicitly or some other operations select it), there doesn't appear to be a way to "de-select` the stack.

## Issue details

I would like to have a way to "un-select" or "de-select" the currently active stack.

Don't see a mechanism for this mentioned here: https://www.pulumi.com/docs/intro/concepts/stack/

The only work around I cans see right now, If I manually edit or remove files in `rm -rf ~/.pulumi/workspaces/*`, my stack selection is reset.

Zaid-Ajaj:
# Description

Fixes #9070 by implementing a `pulumi stack unselect` command. This command removes the selected stack from the current workspace

## Checklist


pkg/cmd/pulumi/stack_unselect.go
```diff
@@ -0,0 +1,61 @@
...
+			if err != nil {
+				return err
+			}
+
+			settings := currentWorkspace.Settings()
+			if !settings.IsEmpty() {
+				// a stack is selected
+				// reset it
+				settings.Stack = ""
+				saveResult := currentWorkspace.Save()
```
Frassle, line 47:
```suggestion
				err := currentWorkspace.Save()
```
We nearly always Error results `err`

pkg/cmd/pulumi/stack_unselect.go
```diff
@@ -0,0 +1,61 @@
...
+				if saveResult == nil {
+					// saving changes was successful
+					fmt.Println("Stack was unselected")
+				} else {
+					fmt.Println("Something went wrong when unselecting the current stack from the workspace")
```
Frassle, line 52:
```suggestion
					fmt.Println("Could not unselect the current stack from the workspace: %s", err)
```
Include the error text here so the user knows what went wrong

pkg/cmd/pulumi/stack_unselect.go
```diff
@@ -0,0 +1,61 @@
...
+			"from.\n",
+		Args: cmdutil.NoArgs,
+		Run: cmdutil.RunFunc(func(cmd *cobra.Command, args []string) error {
+			currentWorkspace, err := workspace.New()
+			if err != nil {
+				return err
+			}
+
+			settings := currentWorkspace.Settings()
+			if !settings.IsEmpty() {
```
Frassle, line 43: I know `IsEmpty` only looks at Stack now, but I think I'd rather this was an explicit check that settings.Stack is "", rather than using IsEmpty. If we add other workspace settings you could end up that settings isn't empty but Stack is still "".

Zaid-Ajaj: Done ✅

pkg/cmd/pulumi/stack_unselect.go
Zaid-Ajaj: Done ✅ I assumed the returned error would be printed at a later stage but I guess we include it here too

Frassle (approved):

Issue #8246, opened 2021-10-18
[codegen] Suppress creation of a `Provider` type when there is no configuration.


lukehoban:
For component packages in particular, there is typically no package level configuration, and as a result, the `Provider` type that is code generated by default for all packages is not useful or even usable.  

Since the `Provider` resource is not needed when there is no package config that can be constructed independently of another provider, we should consider suppressing the generation of a `Provider` resource for packages that do not have config.  Or else, if there is some reason that might be undesirable, give package authors a way to opt-out of generating a `Provider` resource for their package (which would likely be the default for all component boilerplates).

See the docs for our `aws-apigateway` component for an example of where this looks particularly "awkward": https://www.pulumi.com/registry/packages/aws-apigateway/api-docs/

PR #9183, opened 2022-03-12
[codegen/go] Add support for specifying internal dependencies

### LINKED ISSUE #9180: [codegen/go] Add support for specifying Go dependencies
## Hello!

## Issue details

There are situations where an MLC provider may make use of other resource providers inside its implementation (e.g. `pulumi-random`, `pulumi-docker`, etc.), but not exposed publicly. Such dependencies _aren't_ listed in the provider's associated Go SDK's `go.mod` since nothing in the SDK itself refers to them. Which can be problematic for plugin discovery. Since these dependencies are not listed in the Go SDK's `go.mod`, the associated plugins will not be installed automatically for users of the Go SDK.

Note that the Go SDK codegen does not generate the `go.mod` file -- this must be created manually. And manually adding such dependencies to the SDK's `go.mod` is not sufficient to address this because `go mod tidy` will remove any unused dependencies. To avoid this, we should add a new capability to the Go codegen that emits internal dummy imports (referencing some well known exported type) for any Go dependencies listed in the schema. That way, the dependencies can be manually added to `go.mod` and won't be stripped when `go mod tidy` is run.

justinvp:
There are situations where an MLC provider may use other resource providers inside its implementation. Such dependencies are not exposed through public APIs and aren't used at all from the generated Go SDK. Since these dependencies aren't used in the Go SDK, there's no way to specify them in the SDK's `go.mod`, as they'd be removed by `go mod tidy`. This is problematic for plugin discovery. Since the dependencies aren't able to be listed in the SDK's `go.mod`, the associated plugins will not be installed automatically for users of the SDK.

This change introduces a new Go schema option (`InternalDependencies`) that can be used to specify a set of packages that will be emitted as blank imports in the generated SDK so that the associated dependencies in `go.mod` will not be removed by `go mod tidy`.

Fixes #9180


pkg/codegen/testing/test/testdata/internal-references-go/go/example/init.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	r := &Provider{}
+	err := ctx.RegisterResource(typ, name, nil, r, pulumi.URN_(urn))
+	return r, err
+}
+
+func init() {
+	var _ random.Provider
+	var _ tls.Provider
```
justinvp, line 35: These internal references are emitted because of the following in the schema:

```json
      "internalReferences": [
         "github.com/pulumi/pulumi-random/sdk/v4/go/random.Provider",
         "github.com/pulumi/pulumi-tls/sdk/v4/go/tls.Provider"
       ]
```

justinvp:
I considered adding a more general `Dependencies` schema option for Go (along the lines of `Dependencies`, `Requires`, and `PackageReferences` for Node.js, Python, and .NET respectively), but that has some downsides and would be a more involved feature to implement:

- We don't currently codegen the `go.mod` for the Go SDK (that is something that needs to be maintained manually), so having a general `Dependencies` option would be a bit deceiving.
- What would actually be specified in `Dependencies`? Would it be a map of module to version? Or a map of plugin names to plugin versions? The former would align more with the similar options for the other languages. But then we wouldn't necessarily know what well-known type to reference. We could try to reference the `Provider` type, but that might not exist (see #8246). We'd probably have to load the externally referenced schema to find an exported resource or type to reference (so then would `Dependencies` specify the plugin version?) and take the externally reference schema's import base path into account.

In the end, it seemed like the approach in this PR provides the simplest way to achieve what's needed. You specify a reference to a well known type (most likely `Provider`) using `importpath.type` format (similar to referencing a variable with `ldflags -x`), and the codegen emits a dummy variable referencing it.

iwahbe (commented):
I think `internalReferences` is a helpful addition. Have you explored importing with `_`:
```go
import (
         _ "github.com/pulumi/pulumi-random/sdk/v4/go/random",
         _ "github.com/pulumi/pulumi-tls/sdk/v4/go/tls"
)
```
This tells go we wish to import a package but don't intend to reference it. 

Nit: I think that `internalDependencies` is more clear.

viveklak (approved):
LGTM! Nit - could we update the PR description to reference `InternalDependencies` instead?

justinvp:
I'm going to merge this. Everything passed except the downstream codegen tests which all failed with:

> API rate limit exceeded for installation ID 807028

And this change could not affect any of the downstream providers.

PR #9184, opened 2022-03-12
Improve error message when provided stack path is not fully qualified

### LINKED ISSUE #9131: message error when folks use --stack without fully qualified name
## overview
in https://github.com/pulumi/pulumi/pull/9051 i removed the `If you're using the --stack flag, make sure to pass the fully qualified name (org/project/stack)` error message as folks either rarely hit that or never hit that in that situation. we should detect when folks are using that flag and not using a fully qualified name and display this error message then.

there's a bit more convo on #9051 about this.

Zaid-Ajaj:
# Description

Fixes #9131 (issue)

## Checklist

### DIFF
--- a/sdk/go/common/workspace/paths.go
+++ b/sdk/go/common/workspace/paths.go
@@ -142,7 +142,8 @@ func DetectProjectAndPath() (*Project, string, error) {
 		return nil, "", err
 	} else if path == "" {
 		return nil, "", errors.Errorf("no Pulumi project found in the current working directory. " +
-			"Move to a directory with a Pulumi project or try creating a project first with `pulumi new`.")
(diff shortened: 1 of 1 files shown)


susanev (changes requested):
Hi @Zaid-Ajaj i think there may have been some miscommunication on this issue, as it looks like you are adding the fully qualified message back into the error message i removed it from in https://github.com/pulumi/pulumi/pull/9051. folks will rarely or never hit it in that situation, we need to detect specifically when they are using that flag without a qualified name and display only that msg then. happy to sync u with you on Monday if that would be helpful.

Zaid-Ajaj:
Hi @susanev, I was tracking down where the issue occurs to report the error but didn't realise I ended up in the same place you were at before 😅 I will revisit it. Thanks for the heads up!

PR #9185, opened 2022-03-12
Support plugins from private Pulumi repos

Frassle:
# Description

This makes a few changes to our github authentication flow.
1) We only look at the GITHUB_TOKEN environment variable now. GITHUB_ACTOR and GITHUB_PERSONAL_ACCESS_TOKEN are no longer used. 
2) The token is sent via the Authorization header instead of Authentication, as documented at https://docs.github.com/en/rest/overview/resources-in-the-rest-api#authentication
3) We try to use this env var even when accessing the pulumi repos.

This allows us to develop plugins in private repos but continue to use our standard plugin flow.

## Checklist

### DIFF
--- a/sdk/go/common/workspace/plugins.go
+++ b/sdk/go/common/workspace/plugins.go
@@ -183,7 +183,7 @@ func (source *getPulumiSource) Download(
 		serverURL,
 		url.QueryEscape(fmt.Sprintf("pulumi-%s-%s-v%s-%s-%s.tar.gz", source.kind, source.name, version.String(), opSy, arch)))
 
-	req, err := buildHTTPRequest(endpoint, "", "", "", "")
+	req, err := buildHTTPRequest(endpoint, "", "", "")
 	if err != nil {
 		return nil, -1, err
 	}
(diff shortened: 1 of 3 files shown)


sdk/go/common/workspace/plugins.go
```diff
@@ -196,47 +196,36 @@ type githubSource struct {
 	name         string
 	kind         PluginKind
 
-	tokenType string
```
Frassle, line 199: `tokenType` was always "token" so I've just trimmed it out

Frassle:
cc @patriziobruno, you may want to test this is ok for your workflows.

sdk/go/common/workspace/plugins.go
```diff
@@ -196,47 +196,22 @@ type githubSource struct {
...
-// Creates a new github source from authentication data in the environment, if it exists
-func newAuthenticatedGithubSource(organization, name string, kind PluginKind) (*githubSource, error) {
-	ghToken := os.Getenv("GITHUB_TOKEN")
-	paToken := ""
-	actor := ""
-	tokenType := "token"
-	if ghToken == "" {
-		tokenType = ""
-		paToken = os.Getenv("GITHUB_PERSONAL_ACCESS_TOKEN")
-		actor = os.Getenv("GITHUB_ACTOR")
```
t0yv0, line 222: Maybe consider warning if these env vars are detected as set? To help users transition that might rely on this.

t0yv0 (approved):
LGTM, very timely.

sdk/go/common/workspace/plugins_test.go
```diff
@@ -281,7 +282,11 @@ func newMockReadCloserString(data string) (io.ReadCloser, int64, error) {
 
 //nolint:paralleltest // mutates environment variables
 func TestPluginDownload(t *testing.T) {
+	expectedBytes := []byte{1, 2, 3}
+	token := "RaNd0m70K3n_"
```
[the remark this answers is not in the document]
Frassle: Lol no, but we don't actually care about the format of the token we just pass it straight through to the http header

patriziobruno:
@Frassle, I love the new version, but it doesn't build on my machine and can't test it. I'm not sure I can use this workflow outside a github action without basic auth.

Frassle:
> I'm not sure I can use this workflow outside a github action without basic auth.

I tested setting GITHUB_TOKEN to my personal access token and sending in the Authorization header and that let me download our private releases so I think this should work. This will be in the release this week so if it does break you just open an issue and we investigate quickly.

sdk/go/common/workspace/plugins.go
Frassle: GITHUB_PERSONAL_ACCESS_TOKEN is an envvar we made up so I can warn on that pretty safely.

patriziobruno:
@Frassie, I concur, I've just tried curling api.github.com with my personal access token used as GITHUB_TOKEN and it worked.

PR #9191, opened 2022-03-14
[auto] Add --path flag for set config

mariospas:
# Description

Addition support for `--path` flag while using set config cmd

Fixes # (issue)

## Checklist


sdk/go/auto/local_workspace.go
```diff
@@ -160,10 +160,14 @@ func (l *LocalWorkspace) SetConfig(ctx context.Context, stackName string, key st
 	if val.Secret {
 		secretArg = "--secret"
 	}
+	pathArg := ""
```
iwahbe, line 163: I'm not sure if this works, and if it does it will be fragile. If `!val.IsPath`, we will construct a command like:
```sh
pulumi config set 'foo[0]' --plaintext "" --stack dev -- bar
```
which will error since `""`  is not a known argument.

Since `path` is a flag, we can just set it to its value. 
```suggestion
        pathArg := fmt.Sprintf("--path=%t", val.IsPath)
```

iwahbe (changes requested):
Thanks for the contribution! I think the change is positive. 

I would love to see a test included.

mikhailshilkov:
Hi @mariospas - do you still want to adjust your PR based on the comments from @iwahbe here?

sdk/go/auto/local_workspace.go
```diff
@@ -160,13 +160,16 @@ func (l *LocalWorkspace) SetConfig(ctx context.Context, stackName string, key st
 	if val.Secret {
 		secretArg = "--secret"
 	}
-	pathArg := ""
+
+	var args []string
 	if val.IsPath {
-		pathArg = "--path"
+		args = []string{"config", "set", secretArg, "--path", key, "--stack", stackName,
```
iwahbe, line 165: We need to ensure that the values of `args` match except for path. I would like to see this done without duplicating the entire arg list. Every time I look at this, I need to manually verify that the only difference between `val.IsPath` and `!val.IsPath` is the extra `"--path"` in the arg list.

iwahbe (changes requested):
@mariospas It looks like you changed all instances of `github.com/pulumi` to `github.com/mariospas`. Could you revert that changes?

lunaris:
I think this was taken care of in https://github.com/pulumi/pulumi/pull/12265; closing.

Issue #4676, opened 2020-05-21
Consider publishing pulumi to winget packages repository


rodrigoscna:
It would be nice to have pulumi installed by just running `winget install pulumi`.

jaxxstorm:
I had a look at this because I thought it might be interesting as well!

Adding a package seems relatively straightforward, you just need to add a version config to the packages repo:
https://github.com/microsoft/winget-pkgs/tree/master/manifests

However, what's currently not clear is what the schema definition for those packages is. For example, [here's](https://docs.microsoft.com/en-us/windows/package-manager/package/manifest?tabs=minschema%2Ccompschema#complete-schema) the complete YAML schema, however it seems that the assumption here is that the winget package _executes_ an installer, rather than just copy a binary

rodrigoscna:
Yeah, it's basically an auto-installer which downloads the setup file and runs it.

I believe you would need to add to the CI a pull request to their repository for every new version released that would have the URL for the newest installer.

adriantorrie:
This would be good to have

adriantorrie:
`winget` released at v1.0 now:
https://devblogs.microsoft.com/commandline/windows-package-manager-1-0/

Makes `winget` available for us in a corporate environment to use as a supported CLI now. Having pulumi as an available 
 `winget` package would be great.

Zaid-Ajaj:
Currently blocked by [winget-cli: Support installing .zip files #140](https://github.com/microsoft/winget-cli/issues/140)

Zaid-Ajaj:
OK, this _was_ blocked because `winget` doesn't know how to install zip files which is what pulumi publishes to the releases. However, I was able to programmatically generate an [MSI installer](https://github.com/pulumi/pulumi-winget/releases/tag/v3.26.1) from the corresponding zip files (retrieved from pulumi releases). 

After releasing the MSI, a winget manifest is generated from it and sent as an [automated PR to the winget](https://github.com/microsoft/winget-pkgs/pull/54644) registry

All that is left is for the winget registry to accept the PR to make it available from `winget install pulumi`.

jedieaston:
Thanks for setting this up!

Looks like @pulumi-bot needs to sign the Microsoft CLA for the PR to pass CI. [Once that's fixed](https://github.com/microsoft/winget-pkgs/pull/54644#issuecomment-1069781619), this should be good to go.

Zaid-Ajaj:
[...]

PR #9193, opened 2022-03-14
Dispatch winget-deploy event on new releases

Zaid-Ajaj:
# Description

This PR adds a new workflow which triggers a `winget-deploy` dispatch event from pulumictl. This triggers the [pulumi-winget](https://github.com/pulumi/pulumi-winget) repository to generate a manifest of the latest pulumi release and submits it to [winget-pkgs](https://github.com/microsoft/winget-pkgs) using [winget-create](https://github.com/microsoft/winget-create)

This is an attempt at fixing #4676 though it requires more testing

## Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -79,6 +79,22 @@ jobs:
           pulumictl create choco-deploy $(pulumictl get version --language generic -o)
         env:
           GITHUB_TOKEN: ${{ secrets.PULUMI_BOT_TOKEN}}
+  winget:
+    name: Winget Update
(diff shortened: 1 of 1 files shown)


mikhailshilkov (approved):
LGTM

Zaid-Ajaj:
@mikhailshilkov Is there a way to trigger this event manually? This way, it will attempt to submit latest pulumi CLI to the winget registry

mikhailshilkov:
Run just one action from the workflow? No I don't think so.

PR #9196, opened 2022-03-14
Suggest to use fully qualified stack name with --stack

### LINKED ISSUE #9131: message error when folks use --stack without fully qualified name

Zaid-Ajaj:
# Description

Fixes #9131

## Checklist


Frassle (commented):
I think this should trigger the error in the places we care about. cc @susanev to confirm the wording.

susanev (approved):
thank you!!

Issue #9192, opened 2022-03-14
[sdk/go] Panic using StringPtr as input to pulumi.All


lukehoban:
The following program panics.

```go
package main

import (
	"fmt"

	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
)

var defaulted = "<defaulted>"

func main() {
	pulumi.Run(func(ctx *pulumi.Context) error {
		aStringOutput := pulumi.String("Test").ToStringOutput()
		aStringPtrInput := pulumi.StringPtr("Hello World")

		derived := pulumi.All(aStringOutput, aStringPtrInput).ApplyT(func(vs []interface{}) string {
			first := vs[0].(string)
			second := vs[1].(*string)

			if second == nil {
				second = &defaulted
			}

			return fmt.Sprintf("%s %s", first, *second)
		}).(pulumi.StringOutput)

		ctx.Export("result", derived)
		return nil
	})
}
```

```
  pulumi:pulumi:Stack (repro-string-ptr-dev):
    panic: reflect: Elem of invalid type interface {}
    goroutine 25 [running]:
    reflect.(*rtype).Elem(0x1719900)
    	/opt/brew/Cellar/go/1.17.8/libexec/src/reflect/type.go:954 +0x134
    github.com/pulumi/pulumi/sdk/v3/go/pulumi.awaitInputs({0x19314f8, 0xc00003c150}, {0x17514e0, 0xc0005b4590, 0x0}, {0x1719900, 0xc00068e080, 0x194})
    	/Users/James/Code/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.26.1/go/pulumi/types.go:723 +0xbfe
    github.com/pulumi/pulumi/sdk/v3/go/pulumi.awaitInputs({0x19314f8, 0xc00003c150}, {0x1719900, 0xc000528090, 0xc00051c000}, {0x1719900, 0xc00068e080, 0x194})
    	/Users/James/Code/go/pkg/mod/github.com/pulumi/pulumi/sdk/v3@v3.26.1/go/pulumi/types.go:742 +0x1125
    github.com/pulumi/pulumi/sdk/v3/go/pulumi.awaitInputs({0x19314f8, 0xc00003c150}, {0x16e1160, 0xc0001820c0, 0x0}, {0x16e1160, 0xc0001820d8, 0x197})
[...]

PR #9197, opened 2022-03-14
Support pointers in pulumi.All

Frassle:
# Description

In awaitInputs we didn't correctly handle the case of the input value being a pointer and the resolved destination being `interface{}`. We now handle this by allocating a new pointer of the same type of the input, and assiging that to the resolved location if resolved.Type is `any`.

Else we expect resolved.Type() to be some pointer type that the exisiting logic can handle.

Fixes https://github.com/pulumi/pulumi/issues/9192

## Checklist


justinvp (approved):
LGTM

PR #9198, opened 2022-03-14
[dotnet/sdk] Refactor usage of concurrent dictionary

Zaid-Ajaj:
# Description

The PR applies a refactoring to the way `ConcurrentDictionary` is used from the dotnet SDK that keeps track of the gRPC channels as suggested by @justinvp 

## Checklist

### DIFF
--- a/sdk/dotnet/Pulumi/Deployment/GrpcEngine.cs
+++ b/sdk/dotnet/Pulumi/Deployment/GrpcEngine.cs
@@ -16,27 +16,27 @@ namespace Pulumi
         private readonly Engine.EngineClient _engine;
         // Using a static dictionary to keep track of and re-use gRPC channels
         // According to the docs (https://docs.microsoft.com/en-us/aspnet/core/grpc/performance?view=aspnetcore-6.0#reuse-grpc-channels), creating GrpcChannels is expensive so we keep track of a bunch of them here
-        private static ConcurrentDictionary<string, GrpcChannel> _engineChannels = new ConcurrentDictionary<string, GrpcChannel>();
+        private static readonly ConcurrentDictionary<string, GrpcChannel> _engineChannels = new ConcurrentDictionary<string, GrpcChannel>();
 
         public GrpcEngine(string engineAddress)
         {
             // maxRpcMessageSize raises the gRPC Max Message size from `4194304` (4mb) to `419430400` (400mb)
             var maxRpcMessageSize = 400 * 1024 * 1024;
-            if (!_engineChannels.ContainsKey(engineAddress))
+            var engineChannel = _engineChannels.GetOrAdd(engineAddress, address => 
             {
                 // Allow for insecure HTTP/2 transport (only needed for netcoreapp3.x)
                 // https://docs.microsoft.com/en-us/aspnet/core/grpc/troubleshoot?view=aspnetcore-6.0#call-insecure-grpc-services-with-net-core-client
                 AppContext.SetSwitch("System.Net.Http.SocketsHttpHandler.Http2UnencryptedSupport", true);
                 // Inititialize the engine channel once for this address
-                _engineChannels[engineAddress] = GrpcChannel.ForAddress(new Uri($"http://{engineAddress}"), new GrpcChannelOptions
+                return GrpcChannel.ForAddress(new Uri($"http://{address}"), new GrpcChannelOptions
                 {
                     MaxReceiveMessageSize = maxRpcMessageSize,
                     MaxSendMessageSize = maxRpcMessageSize,
                     Credentials = Grpc.Core.ChannelCredentials.Insecure,
                 });
-            }
+            });
 
-            this._engine = new Engine.EngineClient(_engineChannels[engineAddress]);
+            this._engine = new Engine.EngineClient(engineChannel);
         }
         
         public async Task LogAsync(LogRequest request)
--- a/sdk/dotnet/Pulumi/Deployment/GrpcMonitor.cs
+++ b/sdk/dotnet/Pulumi/Deployment/GrpcMonitor.cs
@@ -15,26 +15,26 @@ namespace Pulumi
         private readonly ResourceMonitor.ResourceMonitorClient _client;
         // Using a static dictionary to keep track of and re-use gRPC channels
         // According to the docs (https://docs.microsoft.com/en-us/aspnet/core/grpc/performance?view=aspnetcore-6.0#reuse-grpc-channels), creating GrpcChannels is expensive so we keep track of a bunch of them here
-        private static ConcurrentDictionary<string, GrpcChannel> _monitorChannels = new ConcurrentDictionary<string, GrpcChannel>();
+        private static readonly ConcurrentDictionary<string, GrpcChannel> _monitorChannels = new ConcurrentDictionary<string, GrpcChannel>();
         public GrpcMonitor(string monitorAddress)
         {
             // maxRpcMessageSize raises the gRPC Max Message size from `4194304` (4mb) to `419430400` (400mb)
             var maxRpcMessageSize = 400 * 1024 * 1024;
-            if (!_monitorChannels.ContainsKey(monitorAddress))
+            var monitorChannel = _monitorChannels.GetOrAdd(monitorAddress, address =>
             {
                 // Allow for insecure HTTP/2 transport (only needed for netcoreapp3.x)
                 // https://docs.microsoft.com/en-us/aspnet/core/grpc/troubleshoot?view=aspnetcore-6.0#call-insecure-grpc-services-with-net-core-client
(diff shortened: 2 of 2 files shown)


sdk/dotnet/Pulumi/Deployment/GrpcEngine.cs
```diff
@@ -16,27 +16,27 @@ internal class GrpcEngine : IEngine
         private readonly Engine.EngineClient _engine;
         // Using a static dictionary to keep track of and re-use gRPC channels
         // According to the docs (https://docs.microsoft.com/en-us/aspnet/core/grpc/performance?view=aspnetcore-6.0#reuse-grpc-channels), creating GrpcChannels is expensive so we keep track of a bunch of them here
-        private static ConcurrentDictionary<string, GrpcChannel> _engineChannels = new ConcurrentDictionary<string, GrpcChannel>();
+        private static readonly ConcurrentDictionary<string, GrpcChannel> _engineChannels = new ConcurrentDictionary<string, GrpcChannel>();
 
         public GrpcEngine(string engineAddress)
         {
             // maxRpcMessageSize raises the gRPC Max Message size from `4194304` (4mb) to `419430400` (400mb)
             var maxRpcMessageSize = 400 * 1024 * 1024;
```
justinvp, line 24: I was going to suggest moving this inside the lambda, to avoid the closure allocation, but changing it to `const int` rather than `var` achieves that as well:

```suggestion
            const int maxRpcMessageSize = 400 * 1024 * 1024;
```

sdk/dotnet/Pulumi/Deployment/GrpcMonitor.cs
```diff
@@ -15,26 +15,26 @@ internal class GrpcMonitor : IMonitor
         private readonly ResourceMonitor.ResourceMonitorClient _client;
         // Using a static dictionary to keep track of and re-use gRPC channels
         // According to the docs (https://docs.microsoft.com/en-us/aspnet/core/grpc/performance?view=aspnetcore-6.0#reuse-grpc-channels), creating GrpcChannels is expensive so we keep track of a bunch of them here
-        private static ConcurrentDictionary<string, GrpcChannel> _monitorChannels = new ConcurrentDictionary<string, GrpcChannel>();
+        private static readonly ConcurrentDictionary<string, GrpcChannel> _monitorChannels = new ConcurrentDictionary<string, GrpcChannel>();
         public GrpcMonitor(string monitorAddress)
         {
             // maxRpcMessageSize raises the gRPC Max Message size from `4194304` (4mb) to `419430400` (400mb)
             var maxRpcMessageSize = 400 * 1024 * 1024;
```
justinvp, line 22:
```suggestion
            const int maxRpcMessageSize = 400 * 1024 * 1024;
```

justinvp (approved):
One suggestion, but LGTM otherwise. Thanks for cleaning this up!

Frassle:
`GetOrAdd` doesn't take a lock for the key, so this isn't really any different then what was before.

sdk/dotnet/Pulumi/Deployment/GrpcEngine.cs
Zaid-Ajaj: done ✅

sdk/dotnet/Pulumi/Deployment/GrpcMonitor.cs
Zaid-Ajaj: done ✅

justinvp:
> so this isn't really any different then what was before

There's a minor benefit that it avoids looking up the key twice. (And it's more clear, to me at least, what's going on at a glance).

Zaid-Ajaj:
> this isn't really any different then what was before.

@Frassle Yeah, this refactoring doesn't change the functionality. Just makes it look better, I guess. These changes were requested by @justinvp from my previous PR and I wanted to address them here :smile:

Frassle:
Yeh I agree the changes are an improvement, just wanted to make clear that `GrpcChannel.ForAddress` can still get called multiple times for the same address and wondered if we wanted to do a more coarse lock to prevent that.

Zaid-Ajaj:
TIL `GetOrAdd` is _not thread-safe_ 😱 I was just reading the [docs](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.getoradd?view=net-6.0). 

Well, there goes the nice API, I think a combination of `ContainsKey` and `TryAdd` will do the trick without introducing locks. That is, assuming that indexing/reading is atomic

Frassle:
Just to be clear, `GetOrAdd` is thread-safe. It just that it does not lock around the valueFactory and may call that function multiple-times (it assumes that valueFactory is threadsafe). Using `GetOrAdd` has the benefit that you won't concurrently use two channels because GetOrAdd will always return the one single thing in the dictionary.

The only way to prevent ForAddress being called multiple times here is some sort of lock.

The common suggestion I've seen for this sort of thing is to have a `ConcurrentDictionary<TKey, Lazy<TValue>>` where the lazy is created with it's thread safety mode set to [ExecutionAndPublication](https://docs.microsoft.com/en-us/dotnet/api/system.threading.lazythreadsafetymode?view=net-6.0#system-threading-lazythreadsafetymode-executionandpublication).

Zaid-Ajaj:
@Frassle Looks like I am mixing things up 😅 yeah only the `valueFactory` is not necessarily called once. I added the lock back to be sure that a channel is only created once per address. I saw the trick with using Lazy but it seems too fancy for this use-case so I went back to basics with a simple lock

sdk/dotnet/Pulumi/Deployment/GrpcEngine.cs
```diff
@@ -16,24 +16,29 @@ internal class GrpcEngine : IEngine
...
         public GrpcEngine(string engineAddress)
         {
+            // Allow for insecure HTTP/2 transport (only needed for netcoreapp3.x)
+            // https://docs.microsoft.com/en-us/aspnet/core/grpc/troubleshoot?view=aspnetcore-6.0#call-insecure-grpc-services-with-net-core-client
+            AppContext.SetSwitch("System.Net.Http.SocketsHttpHandler.Http2UnencryptedSupport", true);
             // maxRpcMessageSize raises the gRPC Max Message size from `4194304` (4mb) to `419430400` (400mb)
-            var maxRpcMessageSize = 400 * 1024 * 1024;
-            if (!_engineChannels.ContainsKey(engineAddress))
+            const int maxRpcMessageSize = 400 * 1024 * 1024;
+            lock (_channelsLock)
```
Frassle, line 28: We should do a TryGetValue before taking the lock. In the happy case that the channel is already there that saves a lock

Zaid-Ajaj: Like [this](https://github.com/pulumi/pulumi/pull/9198/commits/d6204c039edbb1ebea9ed0d474d5f84777ec5637)?

sdk/dotnet/Pulumi/Deployment/GrpcEngine.cs
```diff
@@ -16,27 +16,36 @@ internal class GrpcEngine : IEngine
...
-                    Credentials = Grpc.Core.ChannelCredentials.Insecure,
-                });
+                // A channel already exists for this address
+                this._engine = new Engine.EngineClient(engineChannel);
             }
+            else 
+            {
+                lock (_channelsLock)
+                {
+                    // Inititialize the engine channel once for this address
```
Frassle, line 37: Between TryGetValue and taking the lock someone else could of added to the dictionary so we need another TryGetValue here.

sdk/dotnet/Pulumi/Deployment/GrpcEngine.cs
```diff
@@ -16,27 +16,36 @@ internal class GrpcEngine : IEngine
...
+                    var channel = GrpcChannel.ForAddress(new Uri($"http://{engineAddress}"), new GrpcChannelOptions
+                    {
+                        MaxReceiveMessageSize = maxRpcMessageSize,
+                        MaxSendMessageSize = maxRpcMessageSize,
+                        Credentials = Grpc.Core.ChannelCredentials.Insecure,
+                    });
 
-            this._engine = new Engine.EngineClient(_engineChannels[engineAddress]);
+                    _engineChannels.TryAdd(engineAddress, channel);
```
Frassle, line 45: This TryAdd should always be true, I'd consider throwing if it's false.

Zaid-Ajaj: > This TryAdd should always be true, I'd consider throwing if it's false

Isn't this a bit overkill? Since we are already using `TryGetValue` to know if a channel already exists, I opted for  `_engineChannels[engineAddress] = channel` to add the channel to the dictionary. Don't think we care too much about the result of `TryAdd`

Frassle (approved):

PR #9203, opened 2022-03-15
ci: tolerate and retry fluke test failures

AaronFriel:
Uses `--rerun-fails` with `gotestsum` to improve reliability of flakey tests.

#### Miscellany

I noticed a build error creating `/node_modules` and saw locally that if PULUMI_HOME isn't set, in the makefile the expression `$(realpath $$HOME/.pulumi-dev` doesn't do what I would expect:

makefile:
```
test_foo:
	echo $(realpath $$HOME/.pulumi-dev/)
	echo $(realpath $$HOME/.pulumi-dev)
```

output:

```
$ make test_foo
echo 

echo 

```

Fortunately, $HOME is very rarely not an absolute path, so I don't know what we're guarding against with realpath. If we do need to use realpath, we can change that line in common.mk to `$(realpath $(shell echo "$$HOME/.pulumi-dev"))` or even `$(shell mkdir -p $$HOME/.pulumi-dev && readlink -f $$HOME/.pulumi-dev)`, but that actually introduces another oddity: MacOS doesn't bundle readlink because it uses a decade+ old version of GNU coreutils. But I digress.

### DIFF
--- a/build/common.mk
+++ b/build/common.mk
@@ -96,9 +96,7 @@ SHELL       := /bin/bash
 STEP_MESSAGE = @echo -e "\033[0;32m$(shell echo '$@' | tr a-z A-Z | tr '_' ' '):\033[0m"
 
 # Our install targets place items item into $PULUMI_ROOT.
-ifeq ($(PULUMI_ROOT),)
-	PULUMI_ROOT:=$(realpath "$$HOME/.pulumi-dev")
-endif
+PULUMI_ROOT ?= $$HOME/.pulumi-dev
 
 # Use Python 3 explicitly vs expecting that `python` will resolve to a python 3
 # runtime.
--- a/scripts/go-test.py
+++ b/scripts/go-test.py
@@ -3,6 +3,7 @@ Wraps `go test`.
 """
 
 from datetime import datetime
+from typing import List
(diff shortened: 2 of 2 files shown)


t0yv0 (approved):
Did you have a chance to test it before merging? IF so lgtm.

AaronFriel:
Here's what a successful test run looks like:

```
TEST_ALL_DEPS="" make test_integration_subpkgs 
gotestsum --jsonfile /home/friel/go/src/github.com/pulumi/pulumi/test-results/5676503b-36fe-4fb9-9440-9a459e88ecb6.json --junitfile /home/friel/go/src/github.com/pulumi/pulumi/test-results/5676503b-36fe-4fb9-9440-9a459e88ecb6.xml --rerun-fails=2 --packages github.com/pulumi/pulumi/tests -- -count=1 -cover -timeout 1h -tags=all -parallel 10 -v
💓
💓
✓  . (27.562s)

=== Skipped
=== SKIP: . TestStackCommands/StackUnselect (0.00s)
    stack_test.go:117: 
    --- SKIP: TestStackCommands/StackUnselect (0.00s)

=== SKIP: . TestLocalStateLocking (0.00s)
    stack_test.go:457: 

DONE 26 tests, 2 skipped in 28.745s
```

PR #9209, opened 2022-03-15
ci: disable dotnet telemetry, emit version

AaronFriel:
n/t

CI errors on the PR may be disregarded as tests use workflow defined on @master by default, and this won't change those until it's merged. (A catch-22 we need to resolve, but separately.)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,6 +16,7 @@ env:
   PYPI_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
   TRAVIS_PUBLISH_PACKAGES: true
   PYTHON: python
+  DOTNET_CLI_TELEMETRY_OPTOUT: "true"
 
 jobs:
   templates_smoke_test:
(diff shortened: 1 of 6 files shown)


.github/workflows/run-build-and-acceptance-tests.yml
```diff
@@ -119,7 +120,11 @@ jobs:
         with:
           dotnet-version: ${{ matrix.dotnet-version }}
       - run: mkdir -p ${{ runner.temp }}/opt/pulumi/nuget
-      - run: dotnet nuget add source ${{ runner.temp }}/opt/pulumi/nuget
+      - name: Add NuGet packages as a local NuGet source
```
t0yv0, line 123: This is needed in lint job?

AaronFriel: Not sure, didn't want to make a bigger change than necessary, just wanted to log the version for posterity.

mikhailshilkov (approved):
This looks fine - but don't we already know the version is .NET 5?

.github/workflows/release.yml
```diff
@@ -387,7 +388,11 @@ jobs:
         with:
           dotnet-version: ${{ matrix.dotnet-version }}
       - run: mkdir -p ${{ runner.temp }}/opt/pulumi/nuget
-      - run: dotnet nuget add source ${{ runner.temp }}/opt/pulumi/nuget
+      - name: Add NuGet packages as a local NuGet source
+        run: |
+          echo $(which dotnet)
```
t0yv0, line 393: Is `echo $(which dotnet)` the same as `which dotnet`?

t0yv0: Or this is to avoid non-0 exit code terminating the workflow.

AaronFriel: As a habit yeah, I put diagnostics in echo statements.

PR #9217, opened 2022-03-16
Lower windows test parallelism

Frassle:
Try and fix the memory limits we seem to be hitting


mikhailshilkov (commented):
I feel like @t0yv0 tried this in the past with no improvement?

AaronFriel:
@Frassle looks good to me to retry here.

AaronFriel (approved):

PR #9236, opened 2022-03-17
Readd "Make StackReference.Name a tokens.Name (#9088)"

Frassle:
This reverts commit f0aa4df149b6d0afa9169ed59d91c92fb12e208d.

This also removes the AsName asserting casts for stack names. We do want to add them in at some point to be sure that bad names don't slip in somehow but they don't need adding with this.

### DIFF
--- a/sdk/go/common/util/fsutil/qname.go
+++ b/sdk/go/common/util/fsutil/qname.go
@@ -25,3 +25,10 @@ import (
 func QnamePath(nm tokens.QName) string {
 	return strings.Replace(string(nm), tokens.QNameDelimiter, string(os.PathSeparator), -1)
 }
+
+// NamePath just cleans a name and makes sure it's appropriate to use as a path.
+func NamePath(nm tokens.Name) string {
+	// Currently we assume this is a no-op because the set of chars allowed in a Name [A-Za-z0-9_.-] are also
+	// valid file names
+	return string(nm)
+}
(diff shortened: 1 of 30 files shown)


iwahbe:
Can you remind me why we are doing this? Are the assertions dangerous?

Frassle:
We're doing this so that our types try to match what we expect in terms of string formats. I'd like to get to the point where we can assert IsName on every function that takes one to ensure we're validating user strings coming into the system. Having the correct types helps with that.

The asserts however are dangerous if anyone currently has data that wouldn't validate. I don't think that's the case anymore but to reduce risk I'm going to put them into the release _after_ the one that has this.

sdk/go/common/util/fsutil/qname.go
```diff
@@ -25,3 +25,10 @@ import (
 func QnamePath(nm tokens.QName) string {
 	return strings.Replace(string(nm), tokens.QNameDelimiter, string(os.PathSeparator), -1)
 }
+
+// NamePath just cleans a name and makes sure it's appropriate to use as a path.
+func NamePath(nm tokens.Name) string {
+	// Currently we assume this is a no-op because the set of chars allowed in a Name [A-Za-z0-9_.-] are also
```
iwahbe, line 31: NIt:
```suggestion
	// Currently this is a no-op because the set of chars allowed in a Name [A-Za-z0-9_.-] are also
```

iwahbe (approved):
LGTM

PR #9238, opened 2022-03-17
PCL: Use resource aliases when resolving a type

iwahbe:
# Description

We want valid PCL to be as high fidelity as possible. That means we should support alias resolution for resource types. 

## Checklist


Makefile
```diff
@@ -128,3 +128,21 @@ tidy::
 
 validate_codecov_yaml::
 	curl --data-binary @codecov.yml https://codecov.io/validate
+
+name=$(subst schema-,,$(word 1,$(subst !, ,$@)))
```
Frassle, line 132: Maybe a quick comment on what this magic incantation does?

Frassle (approved):

Issue #9018, opened 2022-02-19
pulumi stack commands don't work correctly with a default org set


leezen:
Example:

```
❯ pulumi stack ls
NAME  LAST UPDATE  RESOURCE COUNT  URL
aws   n/a          n/a             https://app.pulumi.com/leezen/scratch-typescript/aws
test  n/a          n/a             https://app.pulumi.com/leezen/scratch-typescript/test
```

```
❯ pulumi stack select aws                                                                                                                                                                               error: no stack named 'pulumi/aws' found
```

```
❯ pulumi stack select leezen/aws                           
```

The above appears to succeed. But then, there's no selected stack:

```
❯ pulumi stack ls
NAME  LAST UPDATE  RESOURCE COUNT  URL
aws   n/a          n/a             https://app.pulumi.com/leezen/scratch-typescript/aws
test  n/a          n/a             https://app.pulumi.com/leezen/scratch-typescript/test
```

Moreover, when I go to `pulumi up`, it prompts with with a list of possible stacks:

```
❯ pulumi up
Please choose a stack, or create a new one:  [Use arrows to move, enter to select, type to filter]
> aws
  test
  <create a new stack>
```

Upon choosing one, I receive:

```
❯ pulumi up
Please choose a stack, or create a new one: aws
error: no stack named 'pulumi/aws' found
```

Removing the default org configuration from `~/.pulumi/config.json` resolves the issues above.

PR #9240, opened 2022-03-17
Fix cloudBackendReference.String

Frassle:
# Description

`cloudBackendReference.String` didn't use the same logic for infering the default owner as `cloudBackend.ParseStackReference`. This was partically evident during `pulumi stack select` where if you had two stacks "user/foo" and "org/bar" then `stack ls` would print them as:
```
foo
org/bar
```

If you then set the workspace default-org `stack ls` would still print them in the same way. But when passing "foo" to `ParseStackReference` it would pick up the implicit owner of "org" not "user" (because the default-org is set), and then you'd get a an error that the stack you selected didn't exist.

`String` now matches `ParseStackReference` and checks if a default-org is set and elides that name, not the user name if so.

Fixes https://github.com/pulumi/pulumi/issues/9018

## Checklist


Frassle:
> Are there any scenarios when this can break users?

I don't think so, its only going to change for people using `org set-default` and I think the only difference that could effect people is `pulumi stack --show-name` will print differently but it should print in a way that is now parseable.

Frassle:
>  Do you want to add some tests for `String()`?

Testing these parts of the system is really awkward right now, but I plan on changing this around some more as part of the backend work and should be able to add tests then.

mikhailshilkov:
> Testing these parts of the system is really awkward right now,

Could you expand on this? I guess I'm naive but why not create a `cloudBackendReference` and call a `String()` on it?

Frassle:
> Could you expand on this? I guess I'm naive but why not create a cloudBackendReference and call a String() on it?

Because that needs a whole `cloudBackend` as well. I've got plans to simplify references such that that won't be needed.

mikhailshilkov (approved):
LGTM

PR #9241, opened 2022-03-17
Add missing whitespace in error string

Frassle:
Before:
`If you're using the --stack flag,pass the fully qualified name (org/project/stack):`

After:
`If you're using the --stack flag, pass the fully qualified name (org/project/stack):`


susanev (approved):
Thx!

PR #9243, opened 2022-03-18
Read needs a non-empty ID

Frassle:
# Description

```
I'm getting an internal assertion thrown here: https://github.com/pulumi/pulumi/blob/8b516bb05434ca8c6dac94b76a5583e855ffd0b6/sdk/go/common/resource/plugin/provider_plugin.go#L808
(Actually, it's line 807 in the stack trace, but presumably either the URN or ID is empty..)
This is in a brand new project/stack. Never been deployed. Any ideas on what I should be looking for? (edited) 
```

This was due to accidentally passing `""` as the ID in a get call. This PR fixes it so we now report an error like `Expected an ID for urn:pulumi:test::test::pkgA:m:typA::resA` rather than a panic due to an assert.

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,6 +1,6 @@
 ### Improvements
 
-- [area/cli] - Implement `pulumi stack unselect`.
+- [cli] - Implement `pulumi stack unselect`.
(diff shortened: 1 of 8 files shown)


iwahbe (approved):
Improving error messages is always great. LGTM

PR #9244, opened 2022-03-18
Add version to runtime in Pulumi.yaml

Frassle:
Nothing uses this yet, but as we support real plugins for languages we'll need some way to specify the version.

This supports the version by either adding "@vX.Y.Z" to the runtime string (e.g. `runtime: dotnet@v1.0.0`) or by using the version key:
```
runtime:
  name: dotnet
  version: v1.0.0
```

### DIFF
--- a/sdk/go/auto/local_workspace_test.go
+++ b/sdk/go/auto/local_workspace_test.go
@@ -662,7 +662,7 @@ func TestNewStackRemoteSourceWithSetup(t *testing.T) {
 	}
 	project := workspace.Project{
 		Name: tokens.PackageName(pName),
-		Runtime: workspace.NewProjectRuntimeInfo("go", map[string]interface{}{
+		Runtime: workspace.NewProjectRuntimeInfo("go", nil, map[string]interface{}{
 			"binary": binName,
 		}),
 	}
@@ -772,7 +772,7 @@ func TestUpsertStackRemoteSourceWithSetup(t *testing.T) {
 	}
 	project := workspace.Project{
 		Name: tokens.PackageName(pName),
-		Runtime: workspace.NewProjectRuntimeInfo("go", map[string]interface{}{
+		Runtime: workspace.NewProjectRuntimeInfo("go", nil, map[string]interface{}{
 			"binary": binName,
 		}),
 	}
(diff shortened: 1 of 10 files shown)


sdk/go/common/workspace/project.go
```diff
@@ -202,12 +205,14 @@ func (ps *ProjectStack) Save(path string) error {
 
 type ProjectRuntimeInfo struct {
 	name    string
+	version *semver.Version
 	options map[string]interface{}
 }
 
-func NewProjectRuntimeInfo(name string, options map[string]interface{}) ProjectRuntimeInfo {
+func NewProjectRuntimeInfo(name string, version *semver.Version, options map[string]interface{}) ProjectRuntimeInfo {
```
stack72, line 212: Wouldn't it be better if we now created a struct that we can grow with the new options required rather than have a break of the signature each time we pass something in?

Frassle: Well at that point I think this is just the ProjectRuntimeInfo struct. 
I'm wondering if we should just leave this method as it was to not change the public API here and maybe at some point just remove it altogether. Is it a good go-ism to have a "NewX" method and an exported "X struct"?

AaronFriel (approved):
lgtm, I agree with exporting the constructor and leaving the internal fields private.

AaronFriel:
@Frassle yeah, as-is.

mikhailshilkov (changes requested):
As discussed, let's not merge this until we settle on the design doc

sdk/go/auto/local_workspace_test.go
```diff
@@ -772,7 +772,7 @@ func TestUpsertStackRemoteSourceWithSetup(t *testing.T) {
 	}
 	project := workspace.Project{
 		Name: tokens.PackageName(pName),
-		Runtime: workspace.NewProjectRuntimeInfo("go", map[string]interface{}{
+		Runtime: workspace.NewProjectRuntimeInfo("go", nil, map[string]interface{}{
```
t0yv0, line 775: Having to edit here it seems like its a minor breaking change to the API that users might use?

Frassle: Yeh I have no idea if anyone is calling this. Our "common" code is in a weird place of mostly kinda internal for the cli, but maybe people use it. We never call this with a version, so we could just leave it as is and let it default to nil in the method.

t0yv0 (approved):

Frassle:
Decision made not to support this yet.

PR #9248, opened 2022-03-19
Replace deprecated String.prototype.substr()

CommanderRoot:
# Description

[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with functions which work similarily but aren't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

## Checklist


Frassle (approved):
/run-acceptance-tests

PR #9249, opened 2022-03-20
[area/nodejs] Take engines property into account when engine-strict appear in .npmrc file

### LINKED ISSUE #9201: Pulumi not respecting engines property and npmrc file
## Hello!

## Issue details

Pulumi doesn't appear to respect the `engines` setting in `package.json` or the `engine-strict` config option in `.npmrc`

I've built a Custom Resource in Pulumi with a pre-deploy step that copies files from one location to another before creating a function. It uses the `fs.cpSync` method, which was only introduced to node in  version `16.7`. 

I need to ensure that Pulumi fails if the wrong version of Node is being used. I've set the correct `engines` property in `package.json` and set `engine-strict=true` in my `.npmrc` file.

### Steps to reproduce

Repo to reproduce - [](https://github.com/antstanley/pulumi-repro)

1. Specify minimum Node version on a Pulumi project

add below to `package.json`

```json
    "engines": {
        "node": ">=16.7.0"
    }
```

add below a `.npmrc` file

```sh
engine-strict=true
```

2. Change Node version to earlier version

If you have `nvm` configured run the command below to set your project to use node v14

```sh
nvm use 14
```

3. Run `pulumi up` or `pulumi preview` to run your project

Expected: Would hope to get an error like below, that I get when running `jest` or `ts-node` on the code base

```sh

Your Node version is incompatible with "/code/pulumi-repro".

Expected version: >=16.7.0
Got: v14.19.0

This is happening because the package's manifest has an engines.node field specified.
To fix this issue, install the required Node version.

```

Actual: Pulumi runs as per normal, even though wrong engine is specified

In my specific use case, because I'm using a newer feature of node (`fs.cpSync`) I get an error saying `fs.cpSync is not a function`, that is being generated by a Custom Component I've created used by other teams who might not be fully aware they need the latest LTS version of node.

Zaid-Ajaj:
# Description

Fixes #9201

## Checklist

### DIFF
--- a/sdk/nodejs/cmd/run/run.ts
+++ b/sdk/nodejs/cmd/run/run.ts
@@ -17,6 +17,8 @@ import * as url from "url";
 import * as minimist from "minimist";
 import * as path from "path";
 import * as tsnode from "ts-node";
+import * as ini from "ini";
+import * as semver from "semver";
 import { parseConfigFileTextToJson } from "typescript";
 import { ResourceError, RunError } from "../../errors";
 import * as log from "../../log";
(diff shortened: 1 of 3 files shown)


sdk/nodejs/cmd/run/run.ts
```diff
@@ -59,6 +61,28 @@ function packageObjectFromProjectRoot(projectRoot: string): Record<string, any>
...
+        if (!fs.existsSync(npmRcPath)) {
+            return emptyConfig;
+        }
+        // file .npmrc exists, read its contents
+        const npmRc = fs.readFileSync(npmRcPath, "utf-8");
+        // Use ini to parse the contents of the .npmrc file
+        // This is what node does as described in the npm docs
+        // https://docs.npmjs.com/cli/v8/configuring-npm/npmrc#comments
+        return ini.parse(npmRc);
+    } catch {
```
Frassle, line 79: Wonder if we should just allow this error to bubble up. It's probably something like bad permissions and having that explict error might be better than silently ignoring if engine-strict is set?

Zaid-Ajaj: I followed the same approach as with `packageObjectFromProjectRoot` where we just return an empty object if we are unable to read or parse it. Do you think it is likely that an error would occur when we are only _reading_ the file?

Frassle: Only errors I can think likely are permission issues. It's probably rare enough to not really worry either way.

danielrbradley: Would we have a way of raising a warning here instead - to help the user if it is some kind of permissions issue?

danielrbradley (approved):
This looks good from a nodejs point of view 👍

Zaid-Ajaj:
thanks @danielrbradley for the review 🙏

Issue #9247, opened 2022-03-18
Experimental Update plans fails with "panic: fatal: An assertion has failed" when using StackReference


tedchang77:
## Hello!

- Vote on this issue by adding a 👍 reaction
- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details

using a plan to do an pulumi up that has a pulumi stackreference fails with a panic message.

### Steps to reproduce

1. pulumi preview --save-plan plan.json
2. pulumi up --plan=plan.json --logtostderr -v9

pulumi program to reproduce:
```
import * as k8s from "@pulumi/kubernetes";
import * as kx from "@pulumi/kubernetesx";
import * as pulumi from "@pulumi/pulumi";

const gkeStack = new pulumi.StackReference("telushealth/gke/dev");
const appLabels = { app: "nginx" };
const deployment = new k8s.apps.v1.Deployment("nginx", {
    spec: {
        selector: { matchLabels: appLabels },
        replicas: 1,
        template: {
            metadata: { labels: appLabels },
            spec: { containers: [{ name: "nginx", image: "nginx" }] }
        }
    }
});
export const name = deployment.metadata.name;
```

Expected: plan to be applied to pulumi up
Actual: cli crashes with panic message

```
panic: fatal: An assertion has failed

goroutine 47 [running]:
github.com/pulumi/pulumi/sdk/v3/go/common/util/contract.failfast(...)
        /Users/runner/work/pulumi/pulumi/sdk/go/common/util/contract/failfast.go:23
github.com/pulumi/pulumi/sdk/v3/go/common/util/contract.Assert(...)
        /Users/runner/work/pulumi/pulumi/sdk/go/common/util/contract/assert.go:26
github.com/pulumi/pulumi/pkg/v3/resource/deploy.(*ResourcePlan).checkGoal(0x184ce40, 0xc001e464c0, 0xc001e464c0, 0x3d)
        /Users/runner/work/pulumi/pulumi/pkg/resource/deploy/plan.go:479 +0xf94
github.com/pulumi/pulumi/pkg/v3/resource/deploy.(*stepGenerator).generateSteps(0xc000859d40, {0x7fef240b3168, 0xc00139b4f0})
        /Users/runner/work/pulumi/pulumi/pkg/resource/deploy/step_generator.go:471 +0x16cb
[...]

PR #9254, opened 2022-03-21
Don't check update plans against invalid resources

Frassle:
# Description

If `Check` returns failures it won't return any new inputs that we can check plans against. The resource will failed anyway because of the check failure so we can just skip checking against the update plan.

Fixes https://github.com/pulumi/pulumi/issues/9247

## Checklist


Zaid-Ajaj (approved):
LGTM

PR #9262, opened 2022-03-21
Support projects in the file state backend

Frassle:
# Description

This is a first pass at supporting projects for the file state backend.

Projects must be given explictly, that is unlike the service backend where the stack reference "teststack" would implictly parse as the stack "teststack" in the current project in the current organisation.
"teststack" with the filestate backend will continue to parse as the stack "teststack" not associated with a project, but you can now give the stack reference now as "testproj/teststack" to get a project scoped stack.

This makes no effort to try automatically moving stack files to be associated with projects. Users can use `pulumi stack rename` to do that if they so wish.

Fixes the project part of https://github.com/pulumi/pulumi/issues/2522.

## Checklist

### DIFF
--- a/pkg/backend/filestate/backend.go
+++ b/pkg/backend/filestate/backend.go
@@ -23,7 +23,6 @@ import (
 	"os"
 	"path"
 	"path/filepath"
-	"regexp"
 	"strings"
 	"sync"
 	"time"
@@ -89,19 +88,27 @@ type localBackend struct {
 }
 
 type localBackendReference struct {
-	name tokens.Name
+	name    tokens.Name
+	project tokens.Name
 }
 
 func (r localBackendReference) String() string {
-	return string(r.name)
+	return r.FullyQualifiedName().String()
 }
 
 func (r localBackendReference) Name() tokens.Name {
 	return r.name
 }
 
+func (r localBackendReference) Project() tokens.Name {
+	return r.project
+}
+
 func (r localBackendReference) FullyQualifiedName() tokens.QName {
-	return r.Name().Q()
+	if r.project == "" {
+		return r.name.Q()
+	}
+	return tokens.QName(fmt.Sprintf("%s/%s", r.project, r.name))
 }
 
 func IsFileStateBackendURL(urlstr string) bool {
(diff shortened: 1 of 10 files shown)


pkg/backend/filestate/backend.go
```diff
@@ -253,8 +265,28 @@ func (b *localBackend) SupportsOrganizations() bool {
...
+		name = split[0]
+	case 2:
+		project = split[0]
+		name = split[1]
+	default:
+		return nil, fmt.Errorf("could not parse stack reference '%s'", stackRef)
+	}
+
+	if project != "" && !tokens.IsName(project) {
+		return nil, errors.New("project names may only contain alphanumerics, hyphens, underscores, and periods")
```
t0yv0, line 282: Include here in the error message the given `project` name and/or `stackRef` it was parsed from.

pkg/backend/filestate/backend.go
```diff
@@ -827,27 +870,83 @@ func (b *localBackend) getLocalStacks() ([]tokens.Name, error) {
...
+
 		if file.IsDir {
-			continue
+			projName := objectName(file)
+			// If this isn't a valid Name it won't be a project directory, so skip it
+			if !tokens.IsName(projName) {
+				continue
+			}
+
+			projectFiles, err := listBucket(b.bucket, filepath.Join(path, projName))
```
t0yv0, line 881: If there was an API to listBucket selectively with a wildcard would that be more appropriate? Thinking of buckets with many objects. Perhaps we never have those in the backend.

sdk/go/common/tokens/names.go
```diff
@@ -94,7 +94,7 @@ func IntoQName(s string) QName {
 
 // AsQName converts a given string to a QName, asserting its validity.
 func AsQName(s string) QName {
-	contract.Assertf(IsQName(s), "Expected string '%v' to be a name (%v)", s, QNameRegexpPattern)
+	contract.Assertf(IsQName(s), "Expected string '%v' to be a qname (%v)", s, QNameRegexpPattern)
```
t0yv0, line 97: In case this can ever surface to the user, expand that a bit, what's a qname? "Expected name '%v' to match $QNameRegexpPattern regular expression" or something like this?

t0yv0 (approved):
The changes look good to me, I'm a little under-familiar with the backends so not super sure if this change-set is comprehensive or bw-safe but seems like some thought has been given to this.

pkg/backend/filestate/backend.go
Frassle: It looks like you can ask for everything underneath (just don't pass a delimiter to the query, only a prefix), then we could do all the filtering and splitting client side. I think I'd rather keep this simple version for now but I'll add a TODO to revisit this to see if we can do less bucket API calls.

sdk/go/common/tokens/names.go
Frassle: This is an assert, so we don't really expect it to be surfaced to users unless they were using this in their own code. At which point they've called "AsQName" so they must know what a qname is.

t0yv0: Ah yeah it's fine then.

benlongo:
I attempted to get this even with master in #12012, but may have missed things as I'm unfamiliar with the code.

Frassle:
bors merge

Frassle:
bors merge

Frassle:
bors retry

Frassle:
bors retry

PR #9270, opened 2022-03-22
Implement "pulumi state clear-pending-operations"

### LINKED ISSUE #4265: Easier way to remove/clear pending operations

Zaid-Ajaj:
# Description

This PR implements `pulumi state clear-pending-operations` as an easier to way to clear pending operations from checkpoint files instead of exporting, editing and importing the stack files.

Fixes #4265. In that issue it is also discussed whether or not pulumi should internally and automatically decide to clear the pending operations. I am personally not sure either about the solution but for the time being at least the users will be able to easily clear the operations themselves without having to edit JSON files by hand. 

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -21,6 +21,9 @@
 - [sdk/nodejs] - Take engines property into account when engine-strict appear in npmrc file 
   [#9249](https://github.com/pulumi/pulumi/pull/9249)
 
+- [cli] Implement `pulumi state clear-pending-operations`
+  [#9270](https://github.com/pulumi/pulumi/pull/9270)
+
 ### Bug Fixes
 
 - [sdk/nodejs] - Fix uncaught error "ENOENT: no such file or directory" when an error occurs during the stack up.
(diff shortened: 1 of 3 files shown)


lukehoban:
My thinking was that we would address this via the approach outlined in https://github.com/pulumi/pulumi/issues/4265#issuecomment-721340725.  Any reason to prefer this approach over that?  It seems that the approach noted there takes a lot more of the friction away from users proactively.

Zaid-Ajaj:
> Any reason to prefer this approach over that?

@lukehoban I liked that approach personally but didn't realise it was the consensus as the final solution  so I fell back to what the users initially asked for. Also just found out that [[engine] Clear pending operations with refresh.](https://github.com/pulumi/pulumi/pull/8435) partially addresses this issue but it has been open for a while. 

I am thinking the following:
 - Will resolve the merge conflicts of [[engine] Clear pending operations with refresh.](https://github.com/pulumi/pulumi/pull/8435) and get it in
 - Turn pending operations error into a warning as per your comment

lukehoban:
> I am thinking the following:
> 
> * Will resolve the merge conflicts of https://github.com/pulumi/pulumi/pull/8435 and get it in
> * Turn pending operations error into a warning and explain that refresh clears pendings operations as per your comment

Perfect - sounds like a good plan!

Zaid-Ajaj:
Closing in favor of https://github.com/pulumi/pulumi/pull/9293

PR #9271, opened 2022-03-23
[cli] Add a replay-events debug command.

pgavlin:
This command renders a recorded stream of engine events using either the
progress or diff renderer. This is useful for debugging problems with
these renderers.

### DIFF
--- a/pkg/backend/display/display.go
+++ b/pkg/backend/display/display.go
@@ -56,21 +56,21 @@ func ShowEvents(
 
 	switch opts.Type {
 	case DisplayDiff:
-		ShowDiffEvents(op, action, events, done, opts)
+		ShowDiffEvents(op, events, done, opts)
 	case DisplayProgress:
 		ShowProgressEvents(op, action, stack, proj, events, done, opts, isPreview)
 	case DisplayQuery:
 		contract.Failf("DisplayQuery can only be used in query mode, which should be invoked " +
 			"directly instead of through ShowEvents")
 	case DisplayWatch:
-		ShowWatchEvents(op, action, events, done, opts)
+		ShowWatchEvents(op, events, done, opts)
 	default:
 		contract.Failf("Unknown display type %d", opts.Type)
 	}
 }
 
 func logJSONEvent(encoder *json.Encoder, event engine.Event, opts Options, seq int) error {
-	apiEvent, err := ConvertEngineEvent(event)
+	apiEvent, err := ConvertEngineEvent(event, false /* showSecrets */)
 	if err != nil {
 		return err
 	}
--- a/pkg/backend/httpstate/state.go
+++ b/pkg/backend/httpstate/state.go
@@ -168,7 +168,7 @@ func (u *cloudUpdate) recordEngineEvents(startingSeqNumber int, events []engine.
 
 	var apiEvents apitype.EngineEventBatch
 	for idx, event := range events {
-		apiEvent, convErr := display.ConvertEngineEvent(event)
+		apiEvent, convErr := display.ConvertEngineEvent(event, false /* showSecrets */)
 		if convErr != nil {
 			return fmt.Errorf("converting engine event: %w", convErr)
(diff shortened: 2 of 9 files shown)


pkg/backend/httpstate/state.go
```diff
@@ -168,7 +168,7 @@ func (u *cloudUpdate) recordEngineEvents(startingSeqNumber int, events []engine.
 
 	var apiEvents apitype.EngineEventBatch
 	for idx, event := range events {
-		apiEvent, convErr := display.ConvertEngineEvent(event)
+		apiEvent, convErr := display.ConvertEngineEvent(event, false)
```
Frassle, line 171:
```suggestion
		apiEvent, convErr := display.ConvertEngineEvent(event, false /* showSecrets */)
```

pkg/backend/display/display.go
```diff
@@ -56,21 +56,21 @@ func ShowEvents(
...
-		ShowWatchEvents(op, action, events, done, opts)
+		ShowWatchEvents(op, events, done, opts)
 	default:
 		contract.Failf("Unknown display type %d", opts.Type)
 	}
 }
 
 func logJSONEvent(encoder *json.Encoder, event engine.Event, opts Options, seq int) error {
-	apiEvent, err := ConvertEngineEvent(event)
+	apiEvent, err := ConvertEngineEvent(event, true)
```
Frassle, line 73:
```suggestion
	apiEvent, err := ConvertEngineEvent(event, true /* showSecrets */)
```

pkg/cmd/pulumi/replay_events.go
```diff
@@ -0,0 +1,187 @@
...
+			for _, e := range events {
+				eventChannel <- e
+			}
+			<-doneChannel
+
+			return nil
+		}),
+	}
+
+	cmd.PersistentFlags().BoolVarP(
```
Frassle, line 120: Its a bit sad this is needed but looks to be a gnarly untangling to get rid of it.

pgavlin: Yeah. We might be able to scrape this out of the events if we have enough of them--I think that the resource events include a bit that indicates whether or not they're part of a preview. Feels like a better solution would be to issue a "start" event that contains this info, though.

pgavlin:
> > These changes also switch the JSON event log to show secrets
>
> I am very open to making this configurable--event logs are still useful without rich secrets, they're just lower-fidelity. In the case that motivated this change, I needed high-fidelity secrets in order to debug our diff rendering logic.

Curious if anybody has thoughts on this?

iwahbe:
> > > These changes also switch the JSON event log to show secrets
> > 
> > 
> > I am very open to making this configurable--event logs are still useful without rich secrets, they're just lower-fidelity. In the case that motivated this change, I needed high-fidelity secrets in order to debug our diff rendering logic.
> 
> Curious if anybody has thoughts on this?

I think this is fine. Since this is a debug command, we can presume users know that this will expose secrets. That being said, making the behavior configurable would expand the set of users we could ask to run this command, and should add only marginal complexity.

pgavlin:
> > > > These changes also switch the JSON event log to show secrets
> > > 
> > > 
> > > I am very open to making this configurable--event logs are still useful without rich secrets, they're just lower-fidelity. In the case that motivated this change, I needed high-fidelity secrets in order to debug our diff rendering logic.
> > 
> > 
> > Curious if anybody has thoughts on this?
> 
> I think this is fine. Since this is a debug command, we can presume users know that this will expose secrets.[^1] That being said, making the behavior configurable would expand the set of users we could ask to run this command, and should add only marginal complexity.
> 
> [1]: If this assumption is invalid, we should definitely put exposing secrets behind a flag.

I ended up reverting this aspect of the change. Retaining plaintext ends up being pretty much useless because the plaintext is removed upstream by the engine.

Frassle (approved):

PR #9274, opened 2022-03-23
[codegen/go] Update Go SDK function output to check for errors

guineveresaenger:
Fixes https://github.com/pulumi/pulumi-aws/issues/1872.

This should result in the following sample output in the Go SDK:

```
func GetPolicyDocumentOutput(ctx *pulumi.Context, args GetPolicyDocumentOutputArgs, opts ...pulumi.InvokeOption) GetPolicyDocumentResultOutput {
	return pulumi.ToOutputWithContext(context.Background(), args).
		ApplyT(func(v interface{}) (GetPolicyDocumentResult, error) {
			args := v.(GetPolicyDocumentArgs)
			r, err := GetPolicyDocument(ctx, &args, opts...)
			if err != nil {
				return nil, err
			}
			if r == nil {
				return nil, fmt.Errorf("expected either result or error to be nil, not both")
			}
			return *r, err
		}).(GetPolicyDocumentResultOutput)
}
```

# Description

Fixes # (issue)

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -39,10 +39,13 @@
 
 - [cli/engine] - Fix a panic due to passing `""` as the ID for a resource read.
   [#9243](https://github.com/pulumi/pulumi/pull/9243)
-  
+
 - [cli/engine] - Fix a panic due to `Check` failing while using update plans.
   [#9254](https://github.com/pulumi/pulumi/pull/9254)
-  
+
+- [codegen/go] - Fix Go SDK function output to check for errors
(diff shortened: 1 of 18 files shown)


t0yv0:
Minor corrections here: https://github.com/pulumi/pulumi/pull/9274

Thanks so much. This is the right place for the fix.

t0yv0 (approved):

ghostsquad:
I'm running pulumi cli 3.37.2, and it appears this is still an issue. Do you know what's going on?

PR #9279, opened 2022-03-24
Lock golangci-lint at v1.44, the last 1.17 go image

Frassle:


.github/workflows/run-build-and-acceptance-tests.yml
```diff
@@ -60,7 +60,7 @@ jobs:
             [1]: ${{ steps.vars.outputs.run-url }}
 
   go-lint:
-    container: golangci/golangci-lint:latest
+    container: golangci/golangci-lint:v1.44
```
mikhailshilkov, line 63: Do you want to leave a comment about why 1.44? Should we open a follow-up issue too?

mikhailshilkov (approved):

PR #9275, opened 2022-03-23
Strip Aliases from state

Frassle:
Fixes #9089

PR #9283, opened 2022-03-24
Alternative fix for aliases, pass a global alias map around

Frassle:


Frassle:
This needs to go on top of the changes in https://github.com/pulumi/pulumi/pull/9275. It's an "alternative" in the sense of today we needed a fix for the build and it wasn't clear if we should just revert 9275 or merge this.

pgavlin:
Just for the sake of posterity:

The after a resource operation finishes, the snapshotter needs to write out a new copy of the snapshot. However, at the time we write the snapshot, there may be resources that have not yet been registered that refer to the just-registered resources by a different URN due to aliasing. Those references need to be fixed up prior to writing the snapshot in order to preserve the snapshot's integrity (in particular, the property that all URNs refer to resources that exist in the snapshot).

For example, consider the following simple dependency graph: `A <-- B`. When that graph is serialized, `B` will contain a reference to `A` in its dependency list. Let the next run of the program produces the graph `A' <-- B` where `A'` is aliased to `A`. After `A'` is registered, the snapshotter needs to write a snapshot that contains its state, but `B` must also be updated so it references `A'` instead of `A`, which will no longer be in the snapshot.

AaronFriel:
I prefer #9288 here over introducing global state.

PR #9284, opened 2022-03-24
Fix generated code panic on *nil

t0yv0:
# Description

@guineveresaenger how about this version? The `fmt.Errorf()` ran into some issues with having to import `fmt` which unfortunately are still really pesky in our go codegen, we need to work out automatic `import()` section fixup. The alternate version simply coerces `nil` to empty struct. I don't think it's much of a problem as this should not arise really in practice - the underlying code should `err` or else return non-nil. The original error is fixed in this version by guarding `*r` by a nil check.

I've then auto-updated the test cases to match:

```
cd pkg/codegen/go
PULUMI_ACCEPT=1 go test -test.v
```

Fixes # (issue)

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -34,3 +34,6 @@
 
 - [cli/engine] - Fix a panic due to `Check` failing while using update plans.
   [#9254](https://github.com/pulumi/pulumi/pull/9254)
+
+- [codegen/go] - Fix Go SDK function output to check for errors
+  [pulumi-aws#1872](https://github.com/pulumi/pulumi-aws/issues/1872)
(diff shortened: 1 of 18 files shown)


t0yv0:
Very strange. This is a confusing lint failure https://github.com/pulumi/pulumi/runs/5678715762?check_suite_focus=true#step:4:14 which I cannot reproduce locally. 

Let's see if it's flaky or not. I've merged `master`.

guineveresaenger:
aha, I wasn't sure on how to autoupdate the tests! thanks for running that! ❤️ 

I just rebased off of latest `master` so let's see what happens to the diff here!

guineveresaenger:
This looks a lot cleaner - thank you for this suggestion! 😁

Issue #9089, opened 2022-03-02
Stack size issues with parented resources and aliases


DSTAlan:
## Hello!

- Vote on this issue by adding a 👍 reaction
- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details

When trying to update the names of rules and subscriptions for an existing Azure Native stack I am experiencing issues where the pulumi up is taking hours and the stack is growing to a huge size.  I have a stack json file (stored in Azure storage) that currently is 1.3 MB.  When I run a pulumi up this grows in size while Pulumi up is running.  The latest attempt grew from 1.3 MB to over 400MB in size.  When I look in the stack file I see a lot of lines similar to the following...

"urn:pulumi:<stack>::<project>::azure-native:resources/v20190701:ResourceGroup$azure-native:servicebus/v20150801:Namespace$azure-native:servicebus:Topic$azure-native:servicebus:Subscription$azure-native:servicebus/v20210101preview:Rule::<rulename>",
"urn:pulumi:<stack>::<project>::azure-native:resources/v20190701:ResourceGroup$azure-native:servicebus/v20150801:Namespace$azure-native:servicebus:Topic$azure-native:servicebus:Subscription$azure-native:servicebus/v20210601preview:Rule::<rulename>",
"urn:pulumi:<stack>::<project>::azure-native:resources/v20190701:ResourceGroup$azure-native:servicebus/v20150801:Namespace$azure-native:servicebus:Topic$azure-native:servicebus:Subscription$azure-native:servicebus/v20211101:Rule::<rulename>",
"urn:pulumi:<stack>::<project>::azure-native:resources/v20190701:ResourceGroup$azure-native:servicebus/v20150801:Namespace$azure-native:servicebus:Topic$azure-native:servicebus/v20140901:Subscription$azure-native:servicebus:Rule::<rulename>",
"urn:pulumi:<stack>::<project>::azure-native:resources/v20190701:ResourceGroup$azure-native:servicebus/v20150801:Namespace$azure-native:servicebus:Topic$azure-native:servicebus/v20140901:Subscription$azure-native:servicebus/v20170401:Rule::<rulename>",
"urn:pulumi:<stack>::<project>::azure-native:resources/v20190701:ResourceGroup$azure-native:servicebus/v20150801:Namespace$azure-native:servicebus:Topic$azure-native:servicebus/v20140901:Subscription$azure-native:servicebus/v20180101preview:Rule::<rulename>",
[...]

PR #9288, opened 2022-03-24
[engine] Only record a resource's chosen alias.

pgavlin:
As we discovered when removing aliases from the state entirely, the
snapshotter needs to be alias-aware so that it can fix up references to
resources that were aliased. After a resource operation finishes, the
snapshotter needs to write out a new copy of the snapshot. However, at
the time we write the snapshot, there may be resources that have not yet
been registered that refer to the just-registered resources by a
different URN due to aliasing. Those references need to be fixed up
prior to writing the snapshot in order to preserve the snapshot's
integrity (in particular, the property that all URNs refer to resources
that exist in the snapshot).

For example, consider the following simple dependency graph: A <-- B.
When that graph is serialized, B will contain a reference to A in its
dependency list. Let the next run of the program produces the graph A'
<-- B where A' is aliased to A. After A' is registered, the snapshotter
needs to write a snapshot that contains its state, but B must also be
updated so it references A' instead of A, which will no longer be in the
snapshot.

These changes take advantage of the fact that although a resource can
provide multiple aliases, it can only ever resolve those aliases to a
single resource in the existing state. Therefore, at the time the
statefile is fixed up, each resource in the statefile could only have
been aliased to a single old resource, and it is sufficient to store
only the URN of the chosen resource rather than all possible aliases. In
addition to preserving the ability to fix up references to aliased
resources, retaining the chosen alias allows the history of a logical
resource to be followed across aliases.

Fixes https://github.com/pulumi/pulumi/issues/9089.

### DIFF
--- a/pkg/engine/journal.go
+++ b/pkg/engine/journal.go
@@ -28,7 +28,7 @@ type JournalEntry struct {
 
 type JournalEntries []JournalEntry
 
-func (entries JournalEntries) Snap(base *deploy.Snapshot) *deploy.Snapshot {
+func (entries JournalEntries) Snap(base *deploy.Snapshot) (*deploy.Snapshot, error) {
 	// Build up a list of current resources by replaying the journal.
 	resources, dones := []*resource.State{}, make(map[*resource.State]bool)
 	ops, doneOps := []resource.Operation{}, make(map[*resource.State]bool)
@@ -130,8 +130,13 @@ func (entries JournalEntries) Snap(base *deploy.Snapshot) *deploy.Snapshot {
 
 	manifest := deploy.Manifest{}
 	manifest.Magic = manifest.NewMagic()
-	return deploy.NewSnapshot(manifest, secretsManager, resources, operations)
 
+	snap := deploy.NewSnapshot(manifest, secretsManager, resources, operations)
+	err := snap.NormalizeURNReferences()
+	if err != nil {
+		return snap, err
+	}
+	return snap, snap.VerifyIntegrity()
 }
 
 type Journal struct {
@@ -189,7 +194,7 @@ func (j *Journal) RecordPlugin(plugin workspace.PluginInfo) error {
 	return nil
 }
 
-func (j *Journal) Snap(base *deploy.Snapshot) *deploy.Snapshot {
+func (j *Journal) Snap(base *deploy.Snapshot) (*deploy.Snapshot, error) {
 	return j.entries.Snap(base)
 }
 
--- a/pkg/engine/lifeycletest/pulumi_test.go
+++ b/pkg/engine/lifeycletest/pulumi_test.go
@@ -1302,7 +1302,7 @@ func TestAliases(t *testing.T) {
 	}
 
 	updateProgramWithResource := func(
-		snap *deploy.Snapshot, resources []Resource, allowedOps []deploy.StepOp) *deploy.Snapshot {
+		snap *deploy.Snapshot, resources []Resource, allowedOps []deploy.StepOp, expectFailure bool) *deploy.Snapshot {
 		program := deploytest.NewLanguageRuntime(func(_ plugin.RunInfo, monitor *deploytest.ResourceMonitor) error {
 			err := registerResources(t, monitor, resources)
 			return err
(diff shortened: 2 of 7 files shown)


pkg/engine/journal.go
```diff
@@ -119,8 +119,10 @@ func (entries JournalEntries) Snap(base *deploy.Snapshot) *deploy.Snapshot {
 
 	manifest := deploy.Manifest{}
 	manifest.Magic = manifest.NewMagic()
-	return deploy.NewSnapshot(manifest, secretsManager, resources, operations)
 
+	snap := deploy.NewSnapshot(manifest, secretsManager, resources, operations)
+	err := snap.NormalizeURNReferences()
+	return snap, err
```
Frassle, line 125: Worth adding a call to validate here as well?

pkg/engine/lifeycletest/pulumi_test.go
```diff
@@ -1476,26 +1476,6 @@ func TestAliases(t *testing.T) {
 		},
 	}}, []deploy.StepOp{deploy.OpSame, deploy.OpReplace, deploy.OpCreateReplacement, deploy.OpDeleteReplaced})
 
-	// ensure failure when different resources use duplicate aliases
```
Frassle, line 1479: This should still test ok now?

pgavlin: This is an error we can no longer detect. The setup for this test is... odd. It starts from an empty base state and wants to validate that two resources cannot mention the same alias. The only way to do that from the statefile alone is to always store all of the aliases in the statefile. In order to support this check, we'd need to move the handling of that specific case into the step generator. I did this in a different branch, but I got a little bit more ambitious there and fixed a few other bugs I found along the way: https://github.com/pulumi/pulumi/compare/pgavlin/choose-single-alias

IMO, this specific check is relatively low-value, as it's really only covering the case where resources that _have no base state_ cannot be mention the same alias. If the resources are both aliased to the same base state, then we will still issue an error (just added a test for that).

There is one interesting case that we don't catch in either case: given the base state `[A]`, it is possible to register a resource `A` and a resource `B` that is aliased to `A`. This is bad, and we should fix it, but that requires fixing a nasty bug in provider aliasing. That bug is an easy fix, I think (a possible fix is included in that other branch), but I'd like to separate it out nonetheless.

pkg/engine/journal.go
pgavlin: Done.

pgavlin:
Not sure why the checks are showing up as mostly failed, but [the latest run](https://github.com/pulumi/pulumi/actions/runs/2036880757) was successful. The only thing that this needs is a CHANGELOG entry.

pgavlin:
I do want to elaborate on the idea here, too.

As we discovered when removing aliases from the state entirely, the snapshotter needs to be alias-aware so that it can fix up references to resources that were aliased. After a resource operation finishes, the snapshotter needs to write out a new copy of the snapshot. However, at the time we write the snapshot, there may be resources that have not yet been registered that refer to the just-registered resources by a different URN due to aliasing. Those references need to be fixed up prior to writing the snapshot in order to preserve the snapshot's integrity (in particular, the property that all URNs refer to resources that exist in the snapshot).

For example, consider the following simple dependency graph: A <-- B. When that graph is serialized, B will contain a reference to A in its dependency list. Let the next run of the program produces the graph A' <-- B where A' is aliased to A. After A' is registered, the snapshotter needs to write a snapshot that contains its state, but B must also be updated so it references A' instead of A, which will no longer be in the snapshot.

These changes take advantage of the fact that although a resource can provide multiple aliases, it can only ever _resolve_ those aliases to a single resource in the existing state. Therefore, at the time the statefile is fixed up, each resource in the statefile could only have been aliased to a single old resource, and it is sufficient to store only the URN of the chosen resource rather than all possible aliases. In addition to preserving the ability to fix up references to aliased resources, retaining the chosen alias allows the history of a logical resource to be followed across aliases.

AaronFriel (approved):

pkg/resource/deploy/step_generator.go
```diff
@@ -305,6 +306,8 @@ func (sg *stepGenerator) generateSteps(event RegisterResourceEvent) ([]Step, res
 
 				// register the alias with the provider registry
 				sg.deployment.providers.RegisterAlias(urn, urnOrAlias)
+
+				alias = []resource.URN{urnOrAlias}
```
justinvp, line 310: Nit: Worth a comment here describing why we're only saving the single alias?

justinvp (approved):
LGTM

pkg/resource/deploy/step_generator.go
pgavlin: Done.

PR #9290, opened 2022-03-25
Add an EnumType to the PCL model

### LINKED ISSUE #6263: [programgen] strict enums not being generated in programgen
See example here for dotnet: https://github.com/pulumi/arm2pulumi/issues/16

iwahbe:
Introduce a new type in enum type in PCL.

Languages implementation:
- [X] C#
- [x] Python
- [x] TypeScript
- [x] Go

# Description
Fixes #6263

## Checklist


pkg/codegen/hcl2/model/type_enum.go
```diff
@@ -0,0 +1,229 @@
...
+	// Elements are the possible values of the enum.
+	Elements []cty.Value
+
+	// The type of the enum's values.
+	Type Type
+
+	// The associated schema type. It is necessary to include this, since
+	// otherwise identical types are distinct if they have a different schema
+	// type.
+	Token string
```
pgavlin, line 41: The types in `hcl2/model` aren't intended to deal with schema types directly. Instead, the intent is that schema types are associated with `hcl2/model` types via annotations (for example, see the `Annotations` field in [`model.ObjectType`](https://github.com/pulumi/pulumi/blob/master/pkg/codegen/hcl2/model/type_object.go#L30-L44) and the way it is used [by the `pcl` package](https://github.com/pulumi/pulumi/blob/master/pkg/codegen/pcl/binder_schema.go#L302-L305)). Although the annotations approach is a little janky, I'd prefer that we remain consistent throughout this package until we find a better alternative.

iwahbe: I'm good with moving the `src *schema.EnumType` field into an annotation. It's a direct reference to the schema (at the type level), and while it's necessary for codegen, it is not inherently necessary for specifying a `EnumType`. That makes sense to me. 

The `Token` (I'm fine calling it something else, GUID comes to mind) is an intrinsic part of defining an `EnumType`, because `EnumType {Elements: [1,2], Type: Number, Token: "foo"}` is an entirely different type then `EnumType {Elements: [1,2], Type: Number, Token: "bar"}`. It seems very strange to have basic methods like `Equals` rely on an untyped annotation. I can ensure that we don't rely on any property of `Token` in `hcl2/model` besides uniqueness.

pkg/codegen/hcl2/model/type_enum.go
```diff
@@ -0,0 +1,164 @@
...
+
+	// Annotations records any annotations associated with the object type.
+	Annotations []interface{}
+
+	s string
+}
+
+func NewEnumType(token string, typ Type, elements []cty.Value, annotations ...interface{}) *EnumType {
+	if len(elements) != 0 {
```
AaronFriel, line 56: Also illegal to have enum with 0 elements?

AaronFriel (commented):
lgtm with nit, I don't think we have any enums declared with zero variants anyhow.

pkg/codegen/hcl2/model/type_enum.go
pgavlin: > The Token (I'm fine calling it something else, GUID comes to mind) is an intrinsic part of defining an EnumType, because EnumType {Elements: [1,2], Type: Number, Token: "foo"} is an entirely different type then EnumType {Elements: [1,2], Type: Number, Token: "bar"}

So I'm not sure that I agree with the premise--there's nothing about enum types in the abstract that requires them to have nominative semantics--but I see how it's helpful for PCL. Let's ship this as-is and consider following up with separate nominative types later (which might also be useful as a replacement for the fragile annotations system).

pgavlin (approved):

AaronFriel (approved):

pkg/codegen/hcl2/model/type_enum.go
iwahbe: That makes sense. I think I need to read more about type systems. I tend to think the value of an `Enum` as its name, but that isn't the only way to think about it. I'm curious what you imagine. Something like a `Named<T>` wrapper type? 

I'll leave a comment suggesting the refactor.

PR #9293, opened 2022-03-25
Warn users when there are pending operations but proceed with deployment

### LINKED ISSUE #4265: Easier way to remove/clear pending operations

Zaid-Ajaj:
# Description

Fixes #4265

This PR changes the deployment execution such that it no longer errors out when there are pending operations. Instead, it only warns the users and explains that they can use `pulumi refresh` to clear the operations and they will no longer see the warning (unless there are pending `create` operations)

## Checklist

### DIFF
--- a/pkg/engine/lifeycletest/pulumi_test.go
+++ b/pkg/engine/lifeycletest/pulumi_test.go
@@ -642,10 +642,8 @@ func TestRefreshPreservesPendingCreateOperations(t *testing.T) {
 	// Notice here, we have two pending operations: update and create
 	// After a refresh, only the pending CREATE operation should
 	// be in the updated snapshot
-
 	resA := newResource(urnA, "0", false)
 	resB := newResource(urnB, "0", false)
-
 	old := &deploy.Snapshot{
 		PendingOperations: []resource.Operation{
 			{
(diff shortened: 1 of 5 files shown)


rawkode (approved):
Very welcomed change! ❤️

pkg/engine/lifeycletest/pulumi_test.go
```diff
@@ -605,11 +600,6 @@ func TestRefreshWithPendingOperations(t *testing.T) {
 	options := UpdateOptions{Host: deploytest.NewPluginHost(nil, nil, program, loaders...)}
 	project, target := p.GetProject(), p.GetTarget(t, old)
 
-	// Without refreshing, an update should fail.
```
Frassle, line 608: As above

pkg/resource/deploy/deployment_executor.go
```diff
@@ -146,8 +175,11 @@ func (ex *deploymentExecutor) Execute(callerCtx context.Context, opts Options, p
 		if opts.RefreshOnly {
 			return nil, nil
 		}
-	} else if ex.deployment.prev != nil && len(ex.deployment.prev.PendingOperations) != 0 && !preview {
-		return nil, result.FromError(PlanPendingOperationsError{ex.deployment.prev.PendingOperations})
+	} else if ex.deployment.prev != nil && len(ex.deployment.prev.PendingOperations) > 0 && !preview {
+		// Print a warning for users that there are pending operations.
+		// Explain that these operations can be cleared using pulumi refresh (except for CREATE operations)
+		// since these require user intevention:
+		ex.printPendingOperationsWarning()
```
Frassle, line 182: Similarly to what we did with refresh I think we might want to do something to ensure that after doing the update our snapshot still tracks that there were pending operations that were not resovled.

Zaid-Ajaj:
Thanks @Frassle, I will revisit the points and try add some tests 😄

Zaid-Ajaj:
@Frassle Added tests
 - TestRefreshPreservesPendingCreateOperations
 - TestUpdateShowsWarningWithPendingoOperations

Can you please have a look? 🙏

pkg/engine/lifeycletest/pulumi_test.go
```diff
@@ -550,11 +550,6 @@ func TestPreviewWithPendingOperations(t *testing.T) {
 	// A preview should succeed despite the pending operations.
 	_, res := op.Run(project, target, options, true, nil, nil)
 	assert.Nil(t, res)
-
-	// But an update should fail.
-	_, res = op.Run(project, target, options, false, nil, nil)
```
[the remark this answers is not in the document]
Frassle: Ah I see this is now just for preview and we have tests elsewhere for updates.

pkg/engine/lifeycletest/pulumi_test.go
```diff
@@ -626,6 +616,150 @@ func TestRefreshWithPendingOperations(t *testing.T) {
...
+					return nil
+				}
+				return result.Errorf("Unexpected warning diag message: %s", payload.Message)
+			}
+		}
+		return result.Error("Expected a diagnostic message, got none")
+	}
+
+	_, res := op.Run(project, target, options, false, nil, validate)
+	assert.Nil(t, res)
```
Frassle, line 760: Like the refresh test above can we assert that the pending operations are still present (and we should test both PendingUpdates and PendingCreates are retained)

Frassle: Hmm that might be ok, I don't think it's ideal but probably good enough for this work

Zaid-Ajaj: Alright, added asserts for retaining pending create operations, this way the behaviour is documented until we revisit a better solution

Frassle (approved):

PR #9295, opened 2022-03-25
Consistently use actions/setup-python@v3

Frassle:


t0yv0:
LGTM, let's see if CI passes.

PR #9298, opened 2022-03-25
sec: address dependabot issue in docker dep

AaronFriel:
We depend on a fairly old version of the Docker dependency, which was flagged with a a CVE on another repo that depends on pulumi/pulumi.

https://github.com/advisories/GHSA-g7v2-2qxx-wjrw

This updates Docker related dependencies to supported versions, and follows the deprecation warning on the term package that it was refactored to a separate repo.

### DIFF
--- a/pkg/backend/display/progress.go
+++ b/pkg/backend/display/progress.go
@@ -27,7 +27,7 @@ import (
 	"unicode"
 	"unicode/utf8"
 
-	"github.com/docker/docker/pkg/term"
+	"github.com/moby/term"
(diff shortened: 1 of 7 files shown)


iwahbe (approved):
Do we want to get read of our last "github.com/docker/docker/*" dependency while we are at it? That would be `"github.com/docker/docker/api/types/time"` in `github.com/pulumi/pulumi/pkg/cmd/pulumi/logs.go`. It is equally out of date.

PR #9310, opened 2022-03-28
Add internal abstractions in backend/httpstate/client

justinvp:
This change introduces internal `httpClient` and `restClient` abstractions in the `httpstate` backend, along with a `newClient` function variable that can be used to customize the `*Client` returned from `NewClient`.

### DIFF
--- a/pkg/backend/httpstate/client/api.go
+++ b/pkg/backend/httpstate/client/api.go
@@ -127,9 +127,37 @@ func intPtr(i int) *int {
 	return &i
 }
 
+// httpClient is an HTTP client abstraction, used by defaultRESTClient.
+type httpClient interface {
+	Do(req *http.Request, retryAllMethods bool) (*http.Response, error)
+}
+
+// defaultHTTPClient is an implementation of httpClient that provides a basic implementation of Do
+// using the specified *http.Client, with retry support.
+type defaultHTTPClient struct {
+	client *http.Client
+}
+
+func (c *defaultHTTPClient) Do(req *http.Request, retryAllMethods bool) (*http.Response, error) {
+	if req.Method == "GET" || retryAllMethods {
+		// Wait 1s before retrying on failure. Then increase by 2x until the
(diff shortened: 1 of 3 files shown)


pkg/backend/httpstate/client/api.go
```diff
@@ -127,9 +127,37 @@ func intPtr(i int) *int {
 	return &i
 }
 
+// httpClient is an HTTP client abstraction, used by defaultRESTClient.
+type httpClient interface {
+	Do(req *http.Request, retryAllMethods bool) (*http.Response, error)
```
AaronFriel, line 132: Nit/bike-shedding: rename method to "Request"? `client.Request(...)` Or is `Do` a common go-ism?

And for my idiomatic Go edification, is it unusual to have an exported method name on a non-exported interface name? I imagine the interface should be `HttpClient` or the method should be `request`?

AaronFriel (approved):

pkg/backend/httpstate/client/api.go
justinvp: > Or is `Do` a common go-ism?

In this case, I'm aligning with `net/http`'s `Client`'s `Do` method: https://pkg.go.dev/net/http#Client.Do

> And for my idiomatic Go edification, is it unusual to have an exported method name on a non-exported interface name? I imagine the interface should be `HttpClient` or the method should be `request` or `do`?

In my experience, I've seen plenty of non-exported types with exported methods, so it doesn't seem too unusual to me, but other Go developers may feel otherwise. I went with exported `Do` to align with the naming in `net/http`, but didn't export the interface itself because I was aiming to avoid introducing new public APIs as part of this change. Although, since the intent is to allow forks to make use of this internal abstraction, perhaps we should consider exporting it. Unless there are any objections, I'll plan on merging this as-is and we can consider exporting the types as a subsequent change.

AaronFriel: Makes sense, appreciate the background there & thought toward keeping the public interface small.

PR #9323, opened 2022-03-31
Monitor.Invoke and Provider.Invoke take different arguments

Frassle:
Monitor.Invoke and Provider.Invoke were both using the same protobuf message as their input, but they actually take different arguments. 
`provider`, `version`, `acceptResources`, and `pluginDownloadURL` are only used when sending messages to the monitor. None of these made sense to send to a provider. We used to set `acceptResources` based on the providers configuration, but providers never looked at this and the engine doesn't care because it can always accept resources.

Our calls to `Invoke` in source_eval.go _should_ have been using `req.GetAcceptResources()` but weren't. This also fixes that.


sdk/proto/provider.proto
```diff
@@ -106,10 +106,6 @@ message ConfigureErrorMissingKeys {
 message InvokeRequest {
     string tok = 1;                  // the function token to invoke.
     google.protobuf.Struct args = 2; // the arguments for the function invocation.
-    string provider = 3;             // an optional reference to the provider version to use for this invoke.
-    string version = 4;              // the version of the provider to use when servicing this request.
-    bool acceptResources = 5;        // when true operations should return resource references as strongly typed.
-    string pluginDownloadURL = 6;    // an optional reference to the provider url to use for this invoke.
```
AaronFriel, line 112: Add a comment here that we've removed fields 3-6, and add a [reserved](https://developers.google.com/protocol-buffers/docs/proto3#reserved) line like `reserved "provider", "version", "acceptResources", "pluginDownloadURL"`? We may be able to just use `3-6`, too, but to be safe reserving the names will keep those fields from being read by an old plugin.

sdk/proto/resource.proto
```diff
@@ -131,3 +131,12 @@ message RegisterResourceOutputsRequest {
...
 }
+
+message ResourceInvokeRequest {
+    string tok = 1;                  // the function token to invoke.
+    google.protobuf.Struct args = 2; // the arguments for the function invocation.
+    string provider = 3;             // an optional reference to the provider version to use for this invoke.
+    string version = 4;              // the version of the provider to use when servicing this request.
+    bool acceptResources = 5;        // when true operations should return resource references as strongly typed.
+    string pluginDownloadURL = 6;    // an optional reference to the provider url to use for this invoke.
+}
```
AaronFriel, line 142: nit: add newline, prevents awkward diffs later

AaronFriel (changes requested):
Looks good modulo the requested change.

AaronFriel (approved):
lgtm

PR #9325, opened 2022-03-31
Add /run-integration-tests command

Frassle:
Add a "/run-integration-tests" command we can run for pull requests where we want to run a full integration check before merging to master.


.github/workflows/pr.yml
```diff
@@ -17,6 +17,7 @@ jobs:
             **Note for the maintainer:**  Commands available:
 
             * `/run-acceptance-tests` - used to test run the acceptance tests for the project
+            * `/run-integration-tests` - used to run the full integration tests for the project
```
t0yv0, line 20: What'd be the difference between acceptance and integration tests for us? Do we document this?

PR #9334, opened 2022-04-01
Ensure that FormatOptions.Color is set before use in diag.defaultSink

Frassle:
defaultSink calls `Color.Colorize` which will panic if Color isn't one of the expected options. We just trying to catch the most egregious failings here where Color has been left blank because of a default initialized `FormatOptions` struct.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -6,6 +6,7 @@
         "build.directoryFilters": [
             "-sdk/dotnet/Pulumi.Automation.Tests/bin",
             "-sdk/nodejs/bin",
+            "-sdk/nodejs/tests",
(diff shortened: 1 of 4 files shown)


Zaid-Ajaj (approved):
The error in the CI looks unrelated to the changes here (dotnet automation tests failing because of a login):
```
Expected: typeof(Pulumi.Automation.Commands.Exceptions.StackNotFoundException)
Actual:   typeof(Pulumi.Automation.Commands.Exceptions.CommandException): code: 255
stdout: 
stderr: error: could not get cloud url: failed to read Pulumi credentials file. Please re-run `pulumi login` to reset your credentials file: unexpected end of JSON input
```

PR #9337, opened 2022-04-02
Unpin protobuf, teach pylint about generated members

Frassle:
Follow on from https://github.com/pulumi/pulumi/pull/9336

### DIFF
--- a/sdk/python/lib/pulumi/_utils.py
+++ b/sdk/python/lib/pulumi/_utils.py
@@ -27,7 +27,6 @@ def _empty():
 
 def _empty_doc():
     """Empty function docstring."""
-    ...
 
 
 _empty_lambda = lambda: None
--- a/sdk/python/mypy.ini
+++ b/sdk/python/mypy.ini
@@ -2,8 +2,6 @@
 
 [mypy]
 
-namespace_packages = True
-
 # Per-module options:
 
 [mypy-dill]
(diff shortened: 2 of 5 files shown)


sdk/python/lib/pulumi/_utils.py
```diff
@@ -27,7 +27,6 @@ def _empty():
 
 def _empty_doc():
     """Empty function docstring."""
-    ...
```
Frassle, line 30: update to pylint it started complaining that this ellipsis was unneeded

sdk/python/mypy.ini
```diff
@@ -2,8 +2,6 @@
 
 [mypy]
 
-namespace_packages = True
```
Frassle, line 5: This was put in because protobuf broke namespacing a while back, they've since backed that change out so we can revert this as well.

sdk/python/requirements.txt
```diff
@@ -1,16 +1,14 @@
...
-protobuf>=3.6.0,<3.20.0
+protobuf>=3.6.0
 grpcio>=1.33.2,<1.44.0
 dill>=0.3.0
 six>=1.12.0
 semver>=2.8.1
 pyyaml>=5.3.1
 
 # Dev packages only needed during development.
-pylint==2.10.2
```
Frassle, line 13: Not sure why this was pinned? Latest version seems to work fine.

Zaid-Ajaj (approved):
LGTM

Issue #7911, opened 2021-09-07
.NET 6 support


mikhailshilkov:
.NET 6 will ship in November 2021. It's going to be the next LTS version.

- [x] #8153 
- [x] https://github.com/pulumi/pulumi/issues/8476
- [x] https://github.com/pulumi/pulumi/issues/8447
- [x] https://github.com/pulumi/templates/issues/269
- [x] https://github.com/pulumi/protobuf/pull/1
- [x] https://github.com/pulumi/templates/issues/292
- [x] https://github.com/pulumi/pulumi-docker-containers/issues/85
- [x] Consider switching our docs to default to .NET 6
- [x] https://github.com/pulumi/pulumi/issues/10114
- [x] https://github.com/pulumi/pulumi-hugo/issues/1724

t0yv0:
Possibly blocked on grpc update.

mikhailshilkov:
All the improvements related to .NET 6 that we planned in this epic are now shipped! .NET 6 is the default .NET version in Pulumi now, and we also rely on some more modern C# features. Read more in our blog [Improved Pulumi experience with .NET 6](https://www.pulumi.com/blog/pulumi-targets-dotnet-6/).

PR #9343, opened 2022-04-04
[sdk/dotnet] Introduce deployment outputs to minimise friction in NET6 programs

Zaid-Ajaj:
# Description
As part of #7911 this PR introduces a new overload for `Deployment.RunAsync` in the dotnet SDK to be used in combination with .NET 6 top-level statements to simplify C# pulumi programs:
```cs
using Pulumi;
using Pulumi.Aws.S3;

await Deployment.RunAsync(() =>
{
    var bucket = new Bucket("my-bucket");
    return new Outputs 
    {
        { "BucketName", bucket.BucketName }
    };
});

```
This overload of `RunAsync` takes a lambda of type `Func<Outputs>` where `Outputs` uses a collection initialiser to add type-safe outputs. For the initial implementation, `Outputs` will only allow `string`, `int`, `bool`, `double` and their `Output<T>` counter-part. More types can be added as necessary in subsequent releases. 

This is better than `Func<IDictionary<string, object?>>` where the values are not typed (objects) and require users to _know_ that they need a concrete implementation of `IDictionary`. This can be confusing for beginners who are just getting started. 
 

The plan is in making .NET 6 examples and docs using this idiom

## Checklist


CHANGELOG_PENDING.md
```diff
@@ -2,6 +2,8 @@
 
 - [cli] - Installing of language specific project dependencies is now managed by the language plugins, not the pulumi cli.
   [#9294](https://github.com/pulumi/pulumi/pull/9294)
+- [sdk/dotnet] Introduce deployment outputs to minimise friction in NET6 programs
```
mikhailshilkov, line 5: Nit
```suggestion
- [sdk/dotnet] Introduce deployment outputs to minimise friction in .NET 6 programs
```

sdk/dotnet/Pulumi/Deployment/DeploymentOutputs.cs
```diff
@@ -0,0 +1,48 @@
+using System;
```
mikhailshilkov, line 1: Copyright is missing. (Btw, don't we have a linter for this?)

sdk/dotnet/Pulumi/Deployment/Deployment_Run.cs
```diff
@@ -22,6 +22,43 @@ public static Task<int> RunAsync(Action action)
...
+        /// <see cref="Deployment.RunAsync(Action)"/> overloads.  Because cloud Resource construction is
+        /// inherently asynchronous, the result of this function is a <see cref="Task{T}"/> which should
+        /// then be returned or awaited.  This will ensure that any problems that are encountered during
+        /// the running of the program are properly reported.  Failure to do this may lead to the
+        /// program ending early before all resources are properly registered.
+        /// </para>
+        /// 
+        /// </summary>
+        /// <param name="contextHandler">Callback that creates stack resources.</param>
+        public static async Task<int> RunAsync(Action<DeploymentOutputs> contextHandler)
```
mikhailshilkov, line 52: What about an overload that would allow async callbacks? (so that users could await invokes, or are we consciously de-emphasizing those in favor of output-based invokes?)

sdk/dotnet/Pulumi/Deployment/DeploymentOutputs.cs
```diff
@@ -0,0 +1,48 @@
...
+        {
+            outputs = new Dictionary<string, object?>();
+        }
+
+        public static DeploymentOutputs Create()
+        {
+            return new DeploymentOutputs();
+        }
+
+        public void Export<T>(string outputName, Output<T> value)
```
mikhailshilkov, line 23: This still allows passing an arbitrary object as a value, right?

sdk/dotnet/Pulumi/Deployment/DeploymentOutputs.cs
```diff
@@ -0,0 +1,48 @@
...
+    public class DeploymentOutputs
+    {
+        private readonly Dictionary<string, object?> outputs;
+
+        private DeploymentOutputs()
+        {
+            outputs = new Dictionary<string, object?>();
+        }
+
+        public static DeploymentOutputs Create()
```
mikhailshilkov, line 18: Nit: Docs are missing for public APIs.

sdk/dotnet/Pulumi/Deployment/DeploymentOutputs.cs
```diff
@@ -0,0 +1,48 @@
...
+    /// <summary>
+    /// A type-safe wrapper around <see cref="IDictionary{String, Object}"/> for generating stack outputs from created resources. 
+    /// </summary>
+    public class DeploymentOutputs
+    {
+        private readonly Dictionary<string, object?> outputs;
+
+        private DeploymentOutputs()
+        {
+            outputs = new Dictionary<string, object?>();
```
mikhailshilkov, line 15: So what's the main advantage of this class over a dictionary? Preventing users from passing objects as values? Would it be better to actually fix that problem and allow passing (a subset of) arbitrary classes?

sdk/dotnet/Pulumi/PublicAPI.Unshipped.txt
```diff
@@ -0,0 +1,8 @@
+Pulumi.DeploymentOutputs
+static Pulumi.DeploymentOutputs.Create() -> Pulumi.DeploymentOutputs
+Pulumi.DeploymentOutputs.Export<T>(string outputName, Pulumi.Output<T> value) -> void
+Pulumi.DeploymentOutputs.Export(string outputName, string value) -> void
+Pulumi.DeploymentOutputs.Export(string outputName, int value) -> void
+Pulumi.DeploymentOutputs.Export(string outputName, bool value) -> void
+Pulumi.DeploymentOutputs.AsDictionary() -> System.Collections.Generic.IDictionary<string, object>
+static Pulumi.Deployment.RunAsync(System.Action<Pulumi.DeploymentOutputs> contextHandler) -> System.Threading.Tasks.Task<int>
```
mikhailshilkov, line 8: I'm a bit nervous to introduce the 3rd way to declare outputs. Let's make sure it has enough upside.

Frassle:
My main worry with this is how it's going to behave if people call outputs.Export inside an apply. The dictionary method forces them to return the keys and values together at the end of the stack method, which means we have a safe point to call RegisterResourceOutputs.

I do wonder if we could maybe take the type safety ideas from this but still keep the single return point of the dictionary approach. My brain immediately when to thinking "we need a `HDictionary<string>`" 😜 which lets not call it that, but I think that's probably the right direction to take this. You'd could have the type safety of `Add<T>(string outputName, Output<T> value)` but still have a single return point. This idea probably works well in other languages as well.

sdk/dotnet/Pulumi/Deployment/DeploymentOutputs.cs
Zaid-Ajaj: The main advantage is syntactic: make writing the programs more beginner-friendly. Not having to define anything to get going just like python/go/typescript programs

I know we can just say "Look at the types and you will know what to do!" but I believe that if a beginner looks at this
```cs
RunAsync(Func<IDictionary<string, object>>)
```
That 
 - they will be confused if they don't know what `Func<...>` is
 - will probably try to `new IDictionary<...>` and get a compile error
 - will wonder why the values are just any `object` where we have explicit `Output<T>` that come from created resources. 

There is also `RunAsync(Action)` which isn't as useful because it doesn't return outputs 

IMO this is the version that people would actually want to use instead of defining `Stack`s

Zaid-Ajaj:
My main worry with this is how it's going to behave if people call outputs.Export inside an apply. The dictionary method forces them to return the keys and values together at the end of the stack method

@Frassle Returning at the end doesn't necessarily mean users will not attempt to export outputs from within an `Apply`. I can still change the signature from `Action<DeploymentOutputs>` to `Func<DeploymentOutputs, DeploymentOutputs>` As follows:
```cs
await Deployment.RunAsync(outputs =>
{
    var bucket = new Bucket("my-bucket");
    outputs.Export("BucketID", bucket.Id);
    return outputs;
});
```
If you _really_ want to make it hard to export from `Apply` calls, I can go the fluent-syntax route:
```cs
await Deployment.RunAsync(() =>
{
    var bucket = new Bucket("my-bucket");
    
    return Outputs
      .Export("BucketID", bucket.Id)
      .Export("BucketName", bucket.BucketName);
});
```
The type would be `Func<DeploymentOutputs>` and `Outputs.Export` starts the method call chain. 

What do you think?

sdk/dotnet/Pulumi/Deployment/Deployment_Run.cs
Zaid-Ajaj: An async callback can be added when we agree on how the sync version looks like :smile: I don't know much about invokes but if these are async calls then we should definitely have an async version (also because many .NET APIs are async too)

sdk/dotnet/Pulumi/Deployment/DeploymentOutputs.cs
Zaid-Ajaj: Added the copyright

sdk/dotnet/Pulumi/Deployment/DeploymentOutputs.cs
Zaid-Ajaj: @mikhailshilkov You were right: any arbitrary value could be implicitly converted to `Output<T>` but now I've removed this from the implementation. Only a select bunch of types can be used with this new `Outputs` type. See updated PR please 🙏

sdk/dotnet/Pulumi/Deployment/Outputs.cs
```diff
@@ -0,0 +1,65 @@
...
+{
+    /// <summary>
+    /// A type-safe wrapper around <see cref="IDictionary{String, Object}"/> for generating stack outputs from created resources. 
+    /// </summary>
+    public class Outputs : IEnumerable<KeyValuePair<string, object?>>
+    {
+        // Internally, we keep track of a dictionary which keeps track of untyped values (objects)
+        private readonly Dictionary<string, object?> outputs = new Dictionary<string, object?>();
+
+        public void Add(string outputName, string value)
```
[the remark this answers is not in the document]
Zaid-Ajaj: @Frassle that's the thing, I am not sure on this one from the users perspective 🤔 if you add `Output<T>` overload, users are able to do this:
```cs
return new Outputs 
{
   { "Bucket", new Bucket("my-bucket") } 
};
```
And yes, users _can_ serialize resources as I've seen in the code but do we want to expose this capability when outputting stack values? I went through the examples to see if there is anything using this pattern but couldn't find anything where users need the full scope of outputs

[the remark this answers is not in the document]
Frassle: Yeh it's an odd space. If we had extensible traits we could mark this as any pulumi serializable value, but given the type system isn't that strong I don't think we can do any better than just `object` which renders this just a slightly nicer IDictionary<string,object>?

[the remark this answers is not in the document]
Zaid-Ajaj: > I don't think we can do any better than just object which renders this just a slightly nicer IDictionary<string,object>

Yeah I agree. I think using a couple of types for now will be more than enough to cover tons of use cases without going with "outputs can be anything". I am still in favor of adding `Outputs` and actually pushing its use with a lambda in `RunAsync` when migrating examples/docs to .NET 6 because right now dotnet samples _stand out_ with the extra boilerplate when compared to the other languages

sdk/dotnet/Pulumi/Deployment/Outputs.cs
```diff
@@ -0,0 +1,65 @@
...
+
+using System.Collections;
+using System.Collections.Generic;
+
+namespace Pulumi
+{
+    /// <summary>
+    /// A type-safe wrapper around <see cref="IDictionary{String, Object}"/> for generating stack outputs from created resources. 
+    /// </summary>
+    public class Outputs : IEnumerable<KeyValuePair<string, object?>>
```
mikhailshilkov, line 11: Can we make it implement `IDictionary` and reuse the existing `RynAsync` overloads? What am I missing?

Zaid-Ajaj: I tried implementing `IDictionary` but it forces `Outputs` to implement `Add(string, object)` and untyped set indexer which is not exactly the idea here. Also tried implementing `IReadOnlyDictionary` but it could not be implicitly converted to `IDictionary` when using the existing `RunAsync` methods.

mikhailshilkov: What about an implicit conversion operator to IDictionary?

Frassle:
> I'm a bit nervous to introduce the 3rd way to declare outputs. Let's make sure it has enough upside.

I like the idea behind doing a change like this but I'm feeling like given the current state of things we're not really going to be able to do any better than the current `new Dictionary<string, object>`.

It was good to explore the space here, but I don't think there's enough value here unless we can get a system that fully supports everything that could be a valid output, and does it in a safe way (i.e. not just `object`).

Zaid-Ajaj:
> I don't think there's enough value here unless we can get a system that fully supports everything that could be a valid output

@Frassle @mikhailshilkov Thanks for providing feedback. The final call is yours. I don't have strong feelings about this either, thought we could do better especially when going to write .NET 6 examples and docs

mikhailshilkov:
Thanks for the ideas! Independently, I suppose we can still contemplate re-writing .NET programs to a top-level `RunAsync` call some time soon-ish?

Zaid-Ajaj:
>  I suppose we can still contemplate re-writing .NET programs to a top-level RunAsync call some time soon-ish?

That's the plan 🙏

PR #9345, opened 2022-04-04
fix: enable codegen for resources named, e.g., Environment

AaronFriel:
Modifies the template for `Utilities.cs` to remove `using` statements and fully qualify all names in it, enabling codegen for resources with names like `Environment` or `System`, or even `Pulumi`.

There are likely still gaps with this, as resources include a number of Using statements, but looking at our examples many of these are unused. e.g.: in one of our unit tests this example only uses the Pulumi namespace:

https://github.com/pulumi/pulumi/blob/a1e18dae4dc00142c086fcae0a90171f278db166/pkg/codegen/testing/test/testdata/other-owned/dotnet/Resource.cs#L4:L9

Screenshot of VS Code showing all but the Pulumi `using` as unnecessary.


Makefile
```diff
@@ -108,6 +111,14 @@ test_pkg_nodejs: get_schemas
 test_pkg_python: get_schemas
 	@cd pkg && $(GO_TEST) ${PKG_CODEGEN_PYTHON}
 
+test_pkg_dotnet: get_schemas
+# invoked as part of "test_pkg_rest", listed separately to update codegen just for dotnet
```
Frassle, line 115: Same comment copied to nodejs and python above?

pkg/codegen/dotnet/templates.go
```diff
@@ -90,11 +85,11 @@ namespace {{.Namespace}}
         private readonly static string version;
         public static string Version => version;
 
-        static Utilities()
+        static {{.ClassName}}()
```
Frassle, line 88: How was this working before?!

pkg/codegen/dotnet/templates.go
```diff
@@ -90,11 +85,11 @@ namespace {{.Namespace}}
...
         {
-            var assembly = typeof(Utilities).GetTypeInfo().Assembly;
+            var assembly = global::System.Reflection.IntrospectionExtensions.GetTypeInfo(typeof({{.ClassName}})).Assembly;
```
Frassle, line 90: `System.Type` already has an `Assembly` property. I'm not sure what this indirection through `GetTypeInfo` is doing?

Makefile
AaronFriel: `test_pkg_rest` does not invoke python or nodejs, it runs everything but those.

```
PROJECT_PKGS    := $(shell cd ./pkg && go list ./... | grep -v -E '^(${PKG_CODEGEN_NODEJS}|${PKG_CODEGEN_PYTHON})$$')
```

pkg/codegen/dotnet/templates.go
AaronFriel: We had hardcoded the classname to Utilities:

https://github.com/pulumi/pulumi/blob/f2d88584444c45808485ced5e47f1e06f04d14c9/pkg/codegen/dotnet/gen.go#L1968

pkg/codegen/dotnet/templates.go
AaronFriel: Not sure, but I didn't want to change the semantics.

Makefile
Frassle: Maybe comment that instead 😨 that's very non-obvious

pkg/codegen/dotnet/templates.go
Frassle: Fair enough, one to re-visit but I'm like 99.9% certain it won't change the semantics here.

Frassle (approved):
Approved cause I need to go sleep, but I'd add some more comments to the Makefile before merging.

Makefile
AaronFriel: Adding that and then merging!

PR #9347, opened 2022-04-05
Fix typo in `pulumi about`

aetheryx:
# Description

Fix typo in `pulumi about` ([pkg/cmd/pulumi/about.go](https://github.com/pulumi/pulumi/compare/master...aetheryx:patch-1#diff-0e269f8c30430af0c9555b9bf13c3940d3befb38488a55dd36e330e990dd6b0e)): `Puluimi` -> `Pulumi`

## Checklist

  Reason: shouldn't need a test

  Reason: I don't think typo fixes make it into the changelog, but can update the pending changelog if they should


susanev (approved):
thank you!!

PR #9351, opened 2022-04-05
[events] Retain `Secret` values.

pgavlin:
Instead of replacing `Secret` values with the text `[secret]`, retain
the `Secret` value and replace its element with the text `[secret]`.
This allows downstream consumers (e.g. the CLI's diff renderer) to
understand that the contents of the value is secret if necessary.

Concretely, this fixes an issue in which differences in nested
properties of secret object values would not render correctly.

### DIFF
--- a/pkg/engine/events.go
+++ b/pkg/engine/events.go
@@ -544,7 +544,7 @@ func filterAsset(v *resource.Asset, debug bool) *resource.Asset {
 	// we only include the text if it represents a user's serialized program code, as
 	// that is something we want the receiver to see to display as part of
 	// progress/diffs/etc.
-	text := v.Text
+	var text string
 	if v.IsUserProgramCode() {
 		// also make sure we filter this in case there are any secrets in the code.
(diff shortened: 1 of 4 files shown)


pkg/engine/events.go
```diff
@@ -605,3 +487,99 @@ func (e *eventEmitter) diagErrorEvent(d *diag.Diag, prefix, msg string, ephemera
 func (e *eventEmitter) diagWarningEvent(d *diag.Diag, prefix, msg string, ephemeral bool) {
 	diagEvent(e, d, prefix, msg, diag.Warning, ephemeral)
 }
+
+func filterResourceProperties(m resource.PropertyMap, debug bool) resource.PropertyMap {
```
pgavlin, line 491: Happy to make a more targeted fix than this refactoring if reviewers feel that constraining these changes is warranted. I felt that the refactor clarified the code given that it removes some reflection and allows us to be more precise in what we expect here.

Frassle (approved):

Issue #9136, opened 2022-03-07
[diffs] Improve diff rendering for long strings


pgavlin:
Displaying diffs between long strings in the style of  `"old" => "new"` is not terribly helpful if only a small portion of the string has changed. We should consider displaying these changes as textual diffs (n.b. I think that we already do this for e.g. the body of serialized closures).

Issue #5831, opened 2019-05-22
Display better diffs for JSON literal properties


jen20:
We could do a better job of displaying diffs in (for example) IAM policies which are JSON strings.

```typescript
  pulumi:pulumi:Stack: (same)
    [urn=urn:pulumi:dev::iamdiff::pulumi:pulumi:Stack::iamdiff-dev]
    ~ aws:iam/policy:Policy: (update)
        [id=arn:aws:iam::153052954103:policy/testpolicy-8b49b63]
        [urn=urn:pulumi:dev::iamdiff::aws:iam/policy:Policy::testpolicy]
      ~ policy: "{\"Version\":\"2012-10-17\",\"Statement\":[{\"Effect\":\"Allow\",\"Action\":[\"s3:CreateBucket\"],\"Resource\":[\"*\"]}]}" => "{\"Version\":\"2012-10-17\",\"Statement\":[{\"Effect\":\"Allow\",\"Action\":[\"s3:CreateBucket\",\"s3:DeleteBucket\"],\"Resource\":[\"*\"]}]}"
    --aws:iam/rolePolicyAttachment:RolePolicyAttachment: (delete-replaced)
        [id=testrole-d8a65d0-20190522144955669700000001]
        [urn=urn:pulumi:dev::iamdiff::aws:iam/rolePolicyAttachment:RolePolicyAttachment::attach]
    +-aws:iam/rolePolicyAttachment:RolePolicyAttachment: (replace)
        [id=testrole-d8a65d0-20190522144955669700000001]
        [urn=urn:pulumi:dev::iamdiff::aws:iam/rolePolicyAttachment:RolePolicyAttachment::attach]
      ~ policyArn: "arn:aws:iam::153052954103:policy/testpolicy-8b49b63" => output<string>
    ++aws:iam/rolePolicyAttachment:RolePolicyAttachment: (create-replacement)
        [id=testrole-d8a65d0-20190522144955669700000001]
        [urn=urn:pulumi:dev::iamdiff::aws:iam/rolePolicyAttachment:RolePolicyAttachment::attach]
      ~ policyArn: "arn:aws:iam::153052954103:policy/testpolicy-8b49b63" => output<string>
```

We could consider use of a JSON patch library to show the actual diff here instead of a complete string, if we have some way of marking that a property is a JSON object.

manospasj:
[...]

PR #9363, opened 2022-04-07
[display] Refactor diff translation and display.

pgavlin:
- Move the code that deals in translating detailed diffs into object
  diffs out of package `backend/display` and into package `engine`
- Move the code that deals in displaying diffs out of package `engine`
  and into package `backend/display`


pgavlin:
This is in advance of some further changes to the renderer I'm working on (see https://github.com/pulumi/pulumi/issues/5831 and https://github.com/pulumi/pulumi/issues/9136)

Frassle (approved):

PR #9369, opened 2022-04-08
fix: Automation API Color flag should pass unquoted string

paiyar:
Background
---
Automation API Color flag should pass unquoted string:

```
error: unsupported color option: '\"auto\"'
```

Since we're using ["os/exec"](https://pkg.go.dev/os/exec) there is no shell, I think we should not quote arguments?

> Unlike the "system" library call from C and other languages, the os/exec package intentionally does not invoke the system shell


iwahbe (approved):
Thanks for the fix.

Issue #9342, opened 2022-04-04
Typescript deprecation message not escaped


danielrbradley:
## Hello!

- Vote on this issue by adding a 👍 reaction
- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details

Adding a deprecation messge including quotation marks breaks compilation of the Typescript SDK.

### Steps to reproduce

1. Add a deprecation message such as `Deprecated in "v1"`
2. Generate and compile Typescript SDK
3. Warning generated as: `pulumi.log.warn("Resource is deprecated: Deprecated in "v1"")`
4. Typescript SDK build will fail

Expected: `pulumi.log.warn("Resource is deprecated: Deprecated in \"v1\"")`
Actual:`pulumi.log.warn("Resource is deprecated: Deprecated in "v1"")`

PR #9371, opened 2022-04-08
Escape deprecation messages before interpolating into source text

Frassle:
# Description

Fixes https://github.com/pulumi/pulumi/issues/9342

## Checklist


danielrbradley (approved):
This looks great thank you!

Frassle:
> Are there any other strings we generate that should be escaped? 

It looks like other strings are using module/type/property names which should already be string safe.

Issue #9350, opened 2022-04-05
CLI Performance Issues when using stackReferences


Kamaradeivanov:
## Hello!

- Vote on this issue by adding a 👍 reaction
- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details

For several months I have been experiencing poor performance when using the `pulumi up` command. I suppose that the number of resources in several of my stack was in cause.

I already check the following closed issue : 

- https://github.com/pulumi/pulumi/issues/3178
- https://github.com/pulumi/pulumi/issues/3257

### Steps to reproduce

Recently I start a new project with just 12 resources : 

- 1 kubernetes Provider
- 3 kubernetes Secret
- 6 pulumi StackReference

Here is the trace aggregated View :

![image](https://user-images.githubusercontent.com/1730415/161802520-f301cbe0-cb8f-4d8a-a587-f85f87561da2.png)

And the trace for the pulumi job :

![image](https://user-images.githubusercontent.com/1730415/161804794-3bc0ee3a-23cb-4c5e-a733-de0fc53501c2.png)

Traces show me almost only api/decryptValue (more than 1000) and some api/getStack (12)

A pulumi up on that stack need 5 minutes to apply ... and 5min for the preview, so 10min to create 3 poor secrets. 

Expected: The pulumi up command should take 1 or 2 minutes.
Actual: It take 10 minutes

Kamaradeivanov:
Sorry forgot to mention that : 

pulumi version : v3.28.0

And typescript versions: 

```json
 "dependencies": {
    "@pulumi/kubernetes": "3.18.0",
    "@pulumi/pulumi": "3.28.0",
  }
```

mikhailshilkov:
Is it possible to remove some/all StackReferences temporarily just to confirm they are the culprit? Does the time grow linearly with the number of references?

Kamaradeivanov:
I have simplify stacks used as reference to have only one secret at the top of each output. Now performance are far better.

Kamaradeivanov:
I've done modification to switch all stack to this principle : if there is more than one secret into an output, I apply the method pulumi.secret() on that output. 

After doing that. When I apply the job that took 14min the first time (the one with 6 StackReference resources) it now took a little more than 4min. 

![image](https://user-images.githubusercontent.com/1730415/162195096-c5b28564-b5cd-4e27-b7bd-aa6d32e5cb50.png)

Frassle:
[...]

PR #9373, opened 2022-04-08
Change BulkDecrypt to not rely on type tests

Frassle:
# Description

`BulkDecrypt` only made use of the bulk decryption if the `Decrypter` passed in was itself a `BulkDecrypter`. Only the `serviceCrypter` implemented this interface but most of the time it's wrapped with a `cachingCrypter`.

Rather than just adding BulkDecrypt to `cachingCrypter` I've just moved it to the `Decrypter` interface and ensured that every  implementation either does something efficient if it can, or calls `DefaultBulkDecrypt` to just loop over 1-by-1. 

Fixes https://github.com/pulumi/pulumi/issues/9350

## Checklist

### DIFF
--- a/pkg/resource/stack/secrets.go
+++ b/pkg/resource/stack/secrets.go
@@ -129,6 +129,10 @@ func (c *cachingCrypter) DecryptValue(ciphertext string) (string, error) {
 	return c.decrypter.DecryptValue(ciphertext)
 }
 
+func (c *cachingCrypter) BulkDecrypt(ciphertexts []string) (map[string]string, error) {
+	return c.decrypter.BulkDecrypt(ciphertexts)
+}
+
 // encryptSecret encrypts the plaintext associated with the given secret value.
 func (c *cachingCrypter) encryptSecret(secret *resource.Secret, plaintext string) (string, error) {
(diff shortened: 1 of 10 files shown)


pkg/resource/stack/secrets.go
```diff
@@ -188,3 +192,40 @@ func (c *mapDecrypter) DecryptValue(ciphertext string) (string, error) {
...
+		c.cache = make(map[string]string)
+	}
+
+	for ct, pt := range decrypted {
+		secretMap[ct] = pt
+		c.cache[ct] = pt
+	}
+
+	return secretMap, nil
+}
```
pgavlin, line 231: Do we actually need this?

pgavlin (commented):
It makes me sad that we need to do this, but it does seem like the right choice given the bug. LGTM.

pgavlin (approved):

pkg/resource/stack/secrets.go
Frassle: I think current data flow is such that no this will never be hit, but it was pretty easy to write and means we won't trip up on a bad method if we refactor things slightly in the future that cause it to start being hit.

PR #9379, opened 2022-04-11
Replace plugin.Proc with Kill

Frassle:
# Description

While today all plugins are processes we have plans to also run plugins via other methods. The only thing the process object was used for on the plugin struct was to kill it, so I've weakened the interface to just require a `Kill` method. We'll be able to implement that for our other plugin runners as well.

## Checklist

### DIFF
--- a/sdk/go/common/resource/plugin/plugin.go
+++ b/sdk/go/common/resource/plugin/plugin.go
@@ -91,9 +91,10 @@ type plugin struct {
 	Args []string
 	// Env specifies the environment of the plugin in the same format as go's os/exec.Cmd.Env
 	// https://golang.org/pkg/os/exec/#Cmd (each entry is of the form "key=value").
-	Env    []string
-	Conn   *grpc.ClientConn
-	Proc   *os.Process
+	Env  []string
+	Conn *grpc.ClientConn
+	// Function to trigger the plugin to be killed. If the plugin is a process this will just SIGKILL it.
+	Kill   func() error
 	Stdin  io.WriteCloser
 	Stdout io.ReadCloser
 	Stderr io.ReadCloser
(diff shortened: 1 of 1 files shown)


t0yv0:
That method is very worrisome, having watched it operate on different platforms while debugging tests. I know it's a lot more work but can we consider reworking things to terminate gracefully without kill? Some heartbeat scheme perhaps where subordinate processes check in with the process monitor and if they can't reach it, gracefully shut down.

t0yv0:
That method is very worrisome, having watched it operate on different platforms while debugging tests. I know it's a lot more work but can we consider reworking things to terminate gracefully without kill? Some heartbeat scheme perhaps where subordinate processes check in with the process monitor and if they can't reach it, gracefully shut down.

t0yv0:
Linux and Mac are different in how they work with orphan process PIDs; whether the process sits in the process table after terminated or not, and Windows-specific things are really different they scan the process table... These race conditions associated with the fact that scanning process children may race with these processes creating children or stopping them. Also for GPID, setgpid accidentally or through library in a subprocess may disrupt the scheme of tracking "descendants" implied here.

t0yv0:
Linux and Mac are different in how they work with orphan process PIDs; whether the process sits in the process table after terminated or not, and Windows-specific things are really different they scan the process table... These race conditions associated with the fact that scanning process children may race with these processes creating children or stopping them. Also for GPID, setgpid accidentally or through library in a subprocess may disrupt the scheme of tracking "descendants" implied here.

t0yv0 (approved):
Approving, with reservations.

Frassle:
> That method is very worrisome, having watched it operate on different platforms while debugging tests.

This PR isn't changing the current method. Just changing the interface to call it, but yes I agree we can probably do this in a cleaner way.

PR #9380, opened 2022-04-11
[cli] Show rich diffs for JSON/YAML objects/arrays

### LINKED ISSUE #5831: Display better diffs for JSON literal properties

pgavlin:
When displaying a diff between textual values, first try to decode the
old and new values as JSON and YAML. If both values decode successfully,
render the diff between the decoded values rather than the diff between
the text.

Fixes #5831.


pgavlin:
I have one reservation about this change as currently written, which is that the kind of diff observed by the object diff may not match the kind of diff observed by the textual diff. It feels like this will be most confusing for providers that are themselves diffing these properties as text. In that case, it is possible for the provider to indicate that a property has changed while the renderer will show it as an update with no actual changes to the value. In these cases, we could opt to continue rendering the textual diff--thoughts?

AaronFriel (approved):

PR #9390, opened 2022-04-13
gosdk: add API to retrieve raw property map of outputs

AaronFriel:
This enables a Pulumi program in Go that directly constructs resources by type tokens to get access to the full property map, including querying the known-ness and secret-ness of properties without needing to marshal them into an appropriately tagged data structure.

For example, in a Go program where we wish to dynamically access properties of a resource, we might perform this step while evaluating keys at each level:

```go
rawOutputs := x.GetRawOutputs(ctx.ctx)
return evaluateAccessF(rawOutputs, accessors)
```

Is there a better way to plumb this API, and are we satisfied with the naming?

### DIFF
--- a/sdk/go/pulumi/context.go
+++ b/sdk/go/pulumi/context.go
@@ -314,6 +314,7 @@ func (ctx *Context) Invoke(tok string, args interface{}, result interface{}, opt
 			KeepResources: true,
 		}),
 	)
+
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


Frassle:
Why can't we just expose access to the output map? Something like `Resource.Outputs() map[string]Output` where that blocks till it's resolved, or explicitly returning a channel or promise like object instead?

AaronFriel:
@Frassle updated the API to be scarier, and also made it return an output rather than block on it because for my use case, performing `ApplyT` on it ensures that I can use the Pulumi engine to handle concurrency rather than block on the output.

Frassle (approved):

Issue #9385, opened 2022-04-12
Add update plan support to Automation API


EvanBoyle:
We should add support for update plans to Automation API. Ideally you should be able to:

1. use Automation API to run a preview and get a strongly typed update plan (make sure types are available in all SDKs) as the output
2. manipulate the plan programmatically
3. run an update using that plan

This request has come up from several users over the last year or so. Here's a note from the original plan issue: 
- https://github.com/pulumi/pulumi/issues/2318#issuecomment-1009486815
- https://github.com/pulumi/pulumi/issues/2318#issuecomment-1089882330

Frassle:
> manipulate the plan programmatically

Not sure what's expected here? We don't currently define the format for an update plan so these are just opaque files to the user.

EvanBoyle:
@Frassle as a user in an automated setting I may want to generate a plan, and then apply some business logic on the plan to either relax or expand the generated constraints. I may just want to read the generated plan and make some assertions on it. 

I don't think transparent string is the right choice here. Not nearly as useful unless we add the types to the Automation API SDKs.

Frassle:
I'm not sure we're ready to expose that level of detail as something user could take a dependency on. There's also the pain point that these types are easy to define for the go sdk (just reuse what's in sdk/go/common/apitype/plan.go) but I didn't think automation api duplicated this many types into other languages for anything else.

PR #9391, opened 2022-04-13
Add update plans to automation api

Frassle:
# Description

Fixes https://github.com/pulumi/pulumi/issues/9385

## Checklist


EvanBoyle (commented):
Generally the shape of the changes looks correct, but I think we should publish the types of the plan in the SDK as a part of this change: 

https://github.com/pulumi/pulumi/issues/9385#issuecomment-1098167901

EvanBoyle:
Also, we do have the ability to add tests here. See:https://github.com/pulumi/pulumi/blob/c2483ecce9f262cc39b24e82265dc09ef0f3c354/sdk/go/auto/local_workspace_test.go#L55

sdk/dotnet/Pulumi.Automation/PreviewOptions.cs
```diff
@@ -20,6 +20,8 @@ public sealed class PreviewOptions : UpdateOptions
 
         public PulumiFn? Program { get; set; }
 
+        public string? Plan {get; set; }
```
iwahbe, line 23: Since this is public facing, it really should have a comment explaining what it does. I had to remember how the CLI worked before I figured out that this represents the path where the plan is stored, not the plan itself.

sdk/go/auto/optpreview/optpreview.go
```diff
@@ -101,6 +101,13 @@ func UserAgent(agent string) Option {
 	})
 }
 
+// Plan specifies the path to save the update plan from the preview to.
```
iwahbe, line 104:
```suggestion
// Plan specifies the path where the update plan should be saved.
```
Should we call this `PlanPath` instead of `Plan`?

iwahbe (approved):
Id love to see some tests. Beyond that and a few nits, LGTM.

Frassle:
Added a couple of tests just to sanity check things. I'll loop back and re-think testing here though, it's pain having to write every test 4 or more times.

Issue #9381, opened 2022-04-11
python dynamic resource providers suddenly stopped working


sammck:
### What happened?

it looks a little like a github repo pulumi/pulumi-pulumi-python disappeared?

(
```bash
.venv) sam@sammck-ubuntu:~/projects/xpulumi/xp/project/dev-box$ pulumi up
Previewing update (dev):
 
error: could not load plugin for pulumi-python provider 'urn:pulumi:dev::dev-box::pulumi:providers:pulumi-python::default': no resource plugin 'pulumi-resource-pulumi-python' found in the workspace or on your $PATH, install the plugin using `pulumi plugin install resource pulumi-python`
(.venv) sam@sammck-ubuntu:~/projects/xpulumi/xp/project/dev-box$ pulumi plugin install resource pulumi-python
error: 404 HTTP error fetching plugin from https://api.github.com/repos/pulumi/pulumi-pulumi-python/releases/latest. If this is a private GitHub repository, try providing a token via the GITHUB_TOKEN environment variable. See: https://github.com/settings/tokens
```

### Steps to reproduce

From a clean machine:

```bash
sam@sammck-ubuntu:~$ pulumi plugin install resource pulumi-python
pulumi: command not found
sam@sammck-ubuntu:~$ nano ~/.bashrc
sam@sammck-ubuntu:~$ ~/.pulumi/bin/pulumi plugin install resource pulumi-python
error: 404 HTTP error fetching plugin from https://api.github.com/repos/pulumi/pulumi-pulumi-python/releases/latest. If this is a private GitHub repository, try providing a token via the GITHUB_TOKEN environment variable. See: https://github.com/settings/tokens
```

### Expected Behavior

successful installation of plugin

### Actual Behavior

HTTP error accessing the pulumi github repo

### Versions used

v.3.28.0

### Additional context

This was working just a few days ago. no changes to my code, now I can neither update nor destroy my stacks.

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

Frassle:
The pulumi-python resource plugin isn't actually a real plugin. It's a file that's packaged with the pulumi binary.

Can you check where you are running `pulumi` from and make sure that `pulumi-resource-pulumi-python` is also in that folder and that folder is on $PATH (N.B. This is arguably a bug, but dynamic providers only work if they're on PATH as opposed to every other provider and plugin)

[...]

PR #9396, opened 2022-04-13
Fix dynamic providers not on $PATH

Frassle:
# Description

Dynamic providers (e.g pulumi-resource-pulumi-nodejs) are bundled with the pulumi binary like language plugins. As such if the pulumi/bin folder isn't on path we need to look them up using the same "look next to the current executable" logic that language plugins have been using.

Fixes https://github.com/pulumi/pulumi/issues/9381

## Checklist


iwahbe (approved):
Assuming tests are ✅ , LGTM

PR #9402, opened 2022-04-13
Handle IgnoreChanges in the engine, don't send to providers

Frassle:
# Description

One of the main uses cases of IgnoreChanges is to work around provider bugs where they report changes from `Diff` even though nothing has really changed. Text normalisation can cause this for example, where for example the user input is "   whitespace   " and the provider internally normalises to "whitespace" but continues to report a change from `Diff`.

We currently (i.e. pre this PR) send the IgnoreChanges list to the provider as part of diff and update. This was only used in tfbridge's diff code to clear out any ignored attributes that were still returned by terraforms diff logic.

This PR is pretty much just moving that logic up to the engine instead. The engine will no longer send the IgnoreChanges list to providers (old providers will see an empty list thanks to grpc compatibility), and instead fix up the result of the providers diff if it reports any changes to the properties that should be ignored.

## Checklist


Frassle:
TODO This needs some more tests to make sure nested property paths are handled correctly, but want to check this general approach is sane before spending time finishing it off.

Frassle:
@pgavlin pointed out that even if we handle Diff in the engine, the tfbridge still needs it passed to Update _and_ it needs to continue to have all it's "diff with ignore changes" logic kept because it re-diffs as part of Update.

Given that, there isn't really much to save moving this to the engine.

PR #9404, opened 2022-04-14
Don't disable ambient providers during tests

Frassle:
I had wanted this to ensure that dynamic providers were working correctly, but we also have test providers that we don't install as proper plugins.


iwahbe (approved):
LGTM

PR #9407, opened 2022-04-15
Return errors on parse failures for `config.Try*`

iwahbe:
# Description

`config.Try*` returns an error message if a config key is missing, but there is currently no way to guarantee that a key is pulled from the config. When a non-string primitive key is tried, a failure to parse leads returns the type appropriate zero value. This is super counterintuitive to me. Especially since `config.TryObject` does return a parse error when appropriate. 

This is a breaking change, but it is hard to imagine a situation where `config.Try*` is used correctly. Therefore I believe the change is appropriate. 

This PR changes the behavior of `config.Try*` to return an error if either the requested key is missing in the config or if the key is invalid for the requested type. 

## Example
```go
cfg := config.New(ctx, "dev")
v, err := cfg.TryInt("key")
if errors.Is(err, config.ErrMissingVar) {
    // Set a default
    v = 42
} else if err != nil {
    // We set the value incorrectly
    panic(fmt.Errorf(`Attempted to pass default value "%s" as float64: %w`, cfg.Require("key"), err))
}
```

PS: We do the same thing for `config.Require*`, and I made the same change. 

## Checklist


CHANGELOG_PENDING.md
```diff
@@ -5,5 +5,16 @@
 
 ### Bug Fixes
 
+<<<<<<< HEAD
```
pgavlin, line 8: This file needs a merge

sdk/go/pulumi/config/require.go
```diff
@@ -53,7 +53,11 @@ func RequireObject(ctx *pulumi.Context, key string, output interface{}) {
 
 func requireBool(ctx *pulumi.Context, key, use, insteadOf string) bool {
 	v := require(ctx, key, use, insteadOf)
-	return cast.ToBool(v)
+	o, err := cast.ToBoolE(v)
+	if err != nil {
+		contract.Failf("unable to parse required configuration variable '%s'; %s", key, err.Error())
```
pgavlin, line 58: Huh, weird that we use `contract.Fail` in the config APIs. Might consider just changing these to raw `panic`s.

pgavlin (changes requested):
CHANGELOG_PENDING needs to be cleaned up; LGTM otherwise.

sdk/go/pulumi/config/require.go
```diff
@@ -18,17 +18,21 @@ package config
...
+	"fmt"
 
 	"github.com/spf13/cast"
 
-	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
 )
 
+func failf(format string, a ...interface{}) {
+	panic(fmt.Sprintf(format, a...))
```
pgavlin, line 29: Nit: I'd panic with an error here to make it a bit easier to recover if necessary.

```suggestion
	panic(fmt.Errorf(format, a...))
```

pgavlin (approved):

pkg/codegen/pcl/rewrite_convert.go
```diff
@@ -288,10 +287,45 @@ func convertLiteralToString(from model.Expression) (string, bool) {
 	return "", false
 }
 
+// lowerConversion performs the main logic of LowerConversion. nil, false is
```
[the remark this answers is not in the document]
iwahbe: Yes it did. Changed to a new branch https://github.com/pulumi/pulumi/pull/9426. I'll revert it here.

PR #9418, opened 2022-04-18
Remove package-lock.json in abount integration test

stack72:
Dependabot was complaining we had pinned to a security vulnerable
version of minimist in our abount integration tests

https://github.com/pulumi/pulumi/security/dependabot/43

We don't check the package-lock.json in on other tests so we are
removing that lock file to remove the security vulnerability

# Description

Fixes # (issue)

## Checklist


iwahbe (approved):
🚢

Issue #9011, opened 2022-02-18
Possible grpcio issues in Python SDK


jaxxstorm:
A customer reported the following:

This was just released [today](https://pypi.org/project/grpcio/#history).  Since the pulumi module requires `grpcio >= 1.33.2` when I installed a new venv today it installed `1.44` which broke my code.  It can be fixed by `pip install grpcio==1.43.0`  Thought this might help save someone the hour I spent troubleshooting my problem.

```
Diagnostics:
  pulumi:pulumi :Stack (monorepo-azure-rglock-sandbox):
    E0217 21:24:58.269558767  269972 fork_posix.cc:70]           Fork support is only compatible with the epoll1 and poll polling strategies
```

We might consider pinning this version until we confirm the fix?

jaxxstorm:
see also #9133 9133

jaxxstorm:
Backstory:

https://github.com/grpc/grpc/blob/master/doc/fork_support.md#111

PR #9423, opened 2022-04-19
Unpin python grpcio version

### LINKED ISSUE #9421: grpcio pinned <1.44 causes issues on ARM64 platforms
https://github.com/pulumi/pulumi/blob/b31c5b136c6e6626c6b5bac8b2b8a777135586bb/sdk/python/requirements.txt#L4

I'm having trouble getting different packages to work together with Pulumi due to the upper bound on grpcio version. Ubuntu 20.04 ARM64 really wants 1.44, and the wheel for 1.43 can't seem to build from source. Looking online, there might have been a reason for the pin in the past, but maybe it is resolved?

I'm running on Apple M1 silicon but in an Ubuntu VM.

iwahbe:
# Description

We pinned the python grpcio version because of https://github.com/grpc/grpc/issues/29044 (https://github.com/pulumi/pulumi/issues/9011). Since the underlying issue has been fixed, we can unpin.

Fixes #9421

## Checklist


Frassle (approved):

Zaid-Ajaj:
Re-ran the tests and now they are green. They were failing because of GA API rate limit

PR #9443, opened 2022-04-21
Return 0 exitcode from runPulumiCommandSync if no error

Frassle:
# Description

If `runPulumiCommandSync` exits without an error it should also report that the exit code was 0, not -2 which we use to indicate an error happened but we don't know the exit code.

## Checklist


Zaid-Ajaj (approved):
LTGM

PR #9444, opened 2022-04-21
pulumi convert command

Frassle:
# Description

First pass at a pulumi convert command to go from one "language" to another. Currently for internal experimentation.

## Checklist

### DIFF
--- a/pkg/cmd/pulumi/convert.go
+++ b/pkg/cmd/pulumi/convert.go
@@ -1,4 +1,4 @@
-// Copyright 2016-2020, Pulumi Corporation.
+// Copyright 2016-2022, Pulumi Corporation.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


pkg/cmd/pulumi/convert.go
```diff
@@ -0,0 +1,101 @@
...
+			case langNodejs:
+				programGenerator = nodejs.GenerateProgram
+			case langPython:
+				programGenerator = python.GenerateProgram
+			default:
+				return result.Errorf("cannot generate programs for %v", language)
+			}
+
+			parser := syntax.NewParser()
+			err := parser.ParseFile(os.Stdin, "main.pp")
```
iwahbe, line 62: This will be used in error messages, so I don't think we should reference a file that doesn't exist. 
```suggestion
                        err := parser.ParseFile(os.Stdin, "<stdin>")
```

AaronFriel (approved):

PR #9462, opened 2022-04-24
convert: add language aliases for --language arg

AaronFriel:
Ensures that if a user types `--language csharp` we get the intended behavior. We may want to refine this to only permit `csharp`, to be determined.

### DIFF
--- a/pkg/cmd/pulumi/convert.go
+++ b/pkg/cmd/pulumi/convert.go
@@ -32,9 +32,8 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/result"
 )
 
-var outDir string
-
 func newConvertCmd() *cobra.Command {
+	var outDir string
 	var language string
 
 	cmd := &cobra.Command{
@@ -49,7 +48,7 @@ func newConvertCmd() *cobra.Command {
 
 			var programGenerator programGeneratorFunc
 			switch language {
-			case "c#":
+			case "csharp", "c#":
 				programGenerator = dotnet.GenerateProgram
 			case langGo:
 				programGenerator = gogen.GenerateProgram
@@ -109,7 +108,7 @@ func newConvertCmd() *cobra.Command {
 
 	cmd.PersistentFlags().StringVar(
 		//nolint:lll
-		&outDir, "out", ".", "Which language plugin to use to generate the pulumi program")
+		&outDir, "out", ".", "The output directory to write the convert project to")
 
 	return cmd
 }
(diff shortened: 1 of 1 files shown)


pkg/cmd/pulumi/convert.go
```diff
@@ -46,11 +49,11 @@ func newConvertCmd() *cobra.Command {
 
 			var programGenerator programGeneratorFunc
 			switch language {
-			case langDotnet:
+			case "c#":
```
Frassle, line 52: This should definitely allow "csharp" as well. We use that else where for template and example names.

pkg/cmd/pulumi/convert.go
```diff
@@ -46,11 +49,11 @@ func newConvertCmd() *cobra.Command {
...
 				programGenerator = dotnet.GenerateProgram
 			case langGo:
 				programGenerator = gogen.GenerateProgram
-			case langNodejs:
```
Frassle, line 53: We're a little inconsistent about if we really mean languages or runtimes 😞 
go and python are easy, langauge == runtime. But nodejs will run javascript or typescript but it codegens typescript, and dotnet will run fsharp, csharp, or vb (or anything else dotnet run will run!) but codegens csharp.
Worth giving some thoughts about this going forward...

pkg/cmd/pulumi/convert.go
```diff
@@ -97,5 +107,9 @@ func newConvertCmd() *cobra.Command {
 		//nolint:lll
 		&language, "language", "", "Which language plugin to use to generate the pulumi program")
 
+	cmd.PersistentFlags().StringVar(
+		//nolint:lll
+		&outDir, "out", ".", "Which language plugin to use to generate the pulumi program")
```
Frassle, line 112:
```suggestion
		&outDir, "out", ".", "The output directory to write the convert project to")
```

pkg/cmd/pulumi/convert.go
```diff
@@ -31,6 +32,8 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/result"
 )
 
+var outDir string
```
Frassle, line 35: This can be moved down in scope with "var language string"

Frassle (approved):

pkg/cmd/pulumi/convert.go
AaronFriel: Definitely. At one point I also had in the file "cs", "ts", and "py" as well. What do you think?

Frassle: Probably not needed for now. Not even sure if we even really need "c#", one name for each thing will probably suffice.

PR #9464, opened 2022-04-24
codegen: preserve externally visible names of a resources and outputs

AaronFriel:
Lexical names, or variable/declaration names, are invisible to the Pulumi API at runtime. They exist for the programmer to give useful names to objects to reason about their program. Modifying lexical names to fit the target language does not change the [denotational semantics](https://en.wikibooks.org/wiki/Haskell/Denotational_semantics) of the program, e.g.: renaming the _variable name_ `myBucket` to `my_bucket` in the example below results in the same behavior on `pulumi up`. 

```ts
const myBucket = new aws.s3.Bucket("my-bucket", {website: {
//    ~~~~~~~~ lexical name        ~~~~~~~~~~~ unique name
    indexDocument: "index.html",
}});
```

By contrast, the first argument to a resource constructor and the argument to name a stack output is meaningful to the Pulumi API and can affect the resources. We'll call this name the UniqueName, and this PR maps a `__uniqueName` string attribute in our intermediate language on a resource or output to a UniqueName on their AST nodes, which is used as-is (modulo escaping) in code generation.

In codegen, we've typically used the lexical name for this argument, this PR modifies our codegen to instead use the UniqueName.

### DIFF
--- a/pkg/codegen/pcl/binder.go
+++ b/pkg/codegen/pcl/binder.go
@@ -29,6 +29,8 @@ import (
 	"github.com/zclconf/go-cty/cty"
 )
 
+const UniqueNamePropertyKey = "__uniqueName"
+
 type bindOptions struct {
 	allowMissingVariables  bool
 	allowMissingProperties bool
@@ -166,6 +168,7 @@ func (b *binder) declareNodes(file *syntax.File) (hcl.Diagnostics, error) {
 	var diagnostics hcl.Diagnostics
 
 	// Declare body items in source order.
+Items:
 	for _, item := range model.SourceOrderBody(file.Body) {
 		switch item := item.(type) {
 		case *hclsyntax.Attribute:
@@ -213,8 +216,15 @@ func (b *binder) declareNodes(file *syntax.File) (hcl.Diagnostics, error) {
 					diagnostics = append(diagnostics, labelsErrorf(item, "resource variables must have exactly two labels"))
 				}
 
+				uniqueName, diagErr := popStringAttr(item.Body.Attributes, UniqueNamePropertyKey)
+				if diagErr != nil {
+					diagnostics = append(diagnostics, diagErr)
+					break Items
+				}
+
 				resource := &Resource{
-					syntax: item,
+					syntax:     item,
+					uniqueName: uniqueName,
 				}
 				declareDiags := b.declareNode(item.Labels[0], resource)
 				diagnostics = append(diagnostics, declareDiags...)
@@ -238,10 +248,16 @@ func (b *binder) declareNodes(file *syntax.File) (hcl.Diagnostics, error) {
 				}
 
 				// TODO(pdg): check body for valid contents
+				uniqueName, diagErr := popStringAttr(item.Body.Attributes, UniqueNamePropertyKey)
+				if diagErr != nil {
+					diagnostics = append(diagnostics, diagErr)
+					break Items
+				}
 
 				v := &OutputVariable{
-					typ:    typ,
-					syntax: item,
+					typ:        typ,
+					syntax:     item,
+					uniqueName: uniqueName,
 				}
 				diags := b.declareNode(name, v)
 				diagnostics = append(diagnostics, diags...)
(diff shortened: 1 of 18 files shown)


Frassle (commented):
Looks fine to me, but I don't touch codegen much

pkg/codegen/pcl/binder.go
```diff
@@ -256,6 +272,33 @@ func (b *binder) declareNodes(file *syntax.File) (hcl.Diagnostics, error) {
 	return diagnostics, nil
 }
 
+// Pop an attribute that's internal to Pulumi, e.g.: "__uniqueName" on a resource or output.
+func popStringAttr(attrs hclsyntax.Attributes, key string) (string, *hcl.Diagnostic) {
```
iwahbe, line 276: Is this how we handle other pulumi special attributes? I'm concerned about destructively mutating the program as we bind it. Do we mention somewhere that a program cannot be bound twice?

iwahbe (changes requested):
In general, this looks super good. 

I have an issue with `Name` and `UniqueName`. Would it be possible to change them to `LexicalName` and `LogicalName` (or something similar). `Name` requires looking at the documentation to figure out which name we mean.  [_logical name_](https://www.pulumi.com/docs/intro/concepts/resources/#autonaming) is what our documentation calls `UniqueName`.  `UniqueName` does not need to be unique (only unique up to kind and stack), and only for useful programs.

pkg/codegen/pcl/binder.go
AaronFriel: I couldn't find anywhere else where we used a side channel like this, like `__default`. I think that's a good point about mutating the program. Will look a bit deeper.

Agree on the renames.

AaronFriel:
@iwahbe 

I renamed and modified `popStringAttr` to `getStringAttrValue`, instead of popping the attribute it just parses the syntax and extracts an exact string value.

In `bindResourceBody`, all attributes were bound to the `Inputs` field as keys, this PR special cases `__logicalName`, instead binding it to the `logicalName` field on the node.

In `bindOutputVariable`, that code only specifically looks for a `value` attribute presently, so this added a check for `__logicalName` and binds that to the `logicalName` field on the node.

Lastly, I added a codegen test. I don't think these are valid because logical names have some rules, but it's a good starting point (out of scope right now) to ensure we're correctly quoting the space of valid logical names.

```hcl
resource resourceLexicalName "random:index/randomPet:RandomPet" {
  // not necessarily a valid logical name, just testing that it passes through to codegen unmodified
  __logicalName = "aA-Alpha_alpha.🤯⁉️"
}

output outputLexicalName {
  __logicalName = "bB-Beta_beta.💜⁉"
  value = resourceLexicalName.id
}
```

```ts
package main

import (
	"github.com/pulumi/pulumi-random/sdk/v4/go/random"
)

func main() {
		resourceLexicalName, err := random.NewRandomPet(ctx, "aA-Alpha_alpha.🤯⁉️", nil)
		if err != nil {
			return err
		}
		ctx.Export("bB-Beta_beta.💜⁉", resourceLexicalName.ID())
		return nil
	})
}
```

pkg/codegen/pcl/resource.go
```diff
@@ -89,11 +94,21 @@ func (r *Resource) Traverse(traverser hcl.Traverser) (model.Traversable, hcl.Dia
 	return r.VariableType.Traverse(traverser)
 }
 
-// Name returns the name of the resource.
+// Deprecated: Name returns the variable or declaration name of the resource.
 func (r *Resource) Name() string {
 	return r.Definition.Labels[0]
 }
```
iwahbe, line 100: Are we depreciating this now? We still use it in our code. See `+python/gen_program.go:358`. What happened to `LexicalName`?

I don't think we should depreciate this, since it's an internal API. I think we should change the name to `LexicalName`.

iwahbe (changes requested):
The change to how we retrieve the tag and increased testing is great. I don't understand the depreciated comment.

pkg/codegen/pcl/resource.go
AaronFriel: I think `Name()` since it's exported might be used by tfgen et al. This was a snafu while I was in the process of adding `LexicalName()`, and I temporarily had two methods, a deprecated `Name()` and a new `LexicalName()`.

The refactor to rename Name to LexicalName seems out of scope. I think the comments are fine, _for now_, but we should revisit in a few weeks. I'll create an issue and assign it to myself.

AaronFriel: I've force-pushed a commit removing this.

pkg/codegen/pcl/binder.go
AaronFriel: https://github.com/pulumi/pulumi/issues/9469

pkg/codegen/pcl/resource.go
AaronFriel: - https://github.com/pulumi/pulumi/issues/9469

iwahbe (approved):

PR #9478, opened 2022-04-26
Monitor.Call and Provider.Call take different arguments

Frassle:
# Description

In a similar vein to https://github.com/pulumi/pulumi/pull/9323, we where reusing the same protobuf message for sending to Monitor.Call and Provider.Call.

The majority of fields on this message we're only set for the provider call, while some were only set for the monitor call.

## Checklist


Frassle:
Talking to Dan and Paul might see if we can just kill Call instead...

PR #9481, opened 2022-04-27
Check parent URN is correct

Frassle:
# Description

Baby steps to moving more logic to the engine and out of the SDKs. This will check that that the root stack does not have a parent set, that other resource's parents have been created, and if they don't set a parent and a root stack exists it will set that to the parent.

## Checklist


RobbieMcKinstry (commented):
From a style and design perspective, this LGTM. I don't have the context to review for correctness.

PR #9484, opened 2022-04-27
[cli] Render JSON/YAML encoded values as objects.

pgavlin:
Reuse the support for decoding JSON/YAML strings into objects to display
such values as objects during create, same, and delete operations (we
already do this for updates).

These changes also unexport a number of methods that should be internal
to the display package.


iwahbe (approved):
LGTM

PR #9491, opened 2022-04-29
Replace Hugo shortcodes

cnunciato:
This change does a couple of things to improve our website build-processing performance:

* Replaces all `{{% md %}}` shortcodes with a custom Markdown-rendering function. The `{{% md %}}` shortcode exists to allow these resource templates to leave the job of rendering Markdown snippets to Hugo, which is okay in principle, but expensive in terms of the CPU and memory resources it requires when applied across all of our providers. Since we already have Goldmark (the same Markdown renderer used by Hugo) available as a dependency here, it's much more performant to do this work outside the Hugo process as part of docs generation.

* Replaces all `{{% chooser %}}`, `{{% choosable %}}`, `{{% examples %}}`, and `{{% example %}}` shortcodes as well for similar reasons. 

See https://github.com/pulumi/docs/pull/7420 for PR previews (using GitHub hosted runners).


jkisk (approved):

iwahbe (approved):
LGTM

PR #9496, opened 2022-04-29
Provider support for dotnet

Frassle:
# Description

This adds a new namespace to the dotnet sdk "Provider" with classes to support the writing of a native dotnet provider.

The Provider class will internally deal with our grpc protocol and call into virtual methods given by the user which use only dotnet native domain models.

It is a bit sad all this is manually written, I feel like there _might_ be a good way to codegen this from the protobuf specs such that if they ever change this file could stay in sync.

Also the PropertyValue Marshal/Unmarshal duplicates a lot of the existing serialization logic, and I think that logic could be rewritten as a layer ontop of this (that is we'd have "objects -> PropertyValues" and "PropetyValues -> grpc" rather than what we currently have which is "objects -> grpc" and now "PropertyValues -> grpc")

## Checklist


sdk/dotnet/Pulumi/Provider/Provider.cs
```diff
@@ -0,0 +1,1335 @@
...
+        {
+            Tok = tok;
+            Args = args;
+        }
+    }
+
+    public sealed class InvokeResponse
+    {
+
+        public IDictionary<string, PropertyValue>? Return { get; set; }
```
Zaid-Ajaj, line 610: With the work being done in #11418, it should be possible to define an invoke response that returns a single `PropertyValue` rather than a bag of these

Frassle: Yeh I'll probably rejig this, few other things I've copied from the go SDK which are more legacy than really needed.

Frassle:
I'll move this over to pulumi-dotnet

PR #9513, opened 2022-05-03
[sdk/go] Enable obtaining resource outputs off a ResourceOutput

### LINKED ISSUE #6073: ApplyString fails when applier return type is StringOutput
ApplyString (and all similar functions) fails when the return type of the applier is a StringOutput, even though it could succeed.  It incorrectly decides to return an `AnyOutput`, and then fails to convert that to a `StringOutput`.  At the very least, a better error saying this is not allowed should be presented instead of defaulting to `AnyOutput` and having that panic as a conversion.  But I see no reason we can't correctly support this as well.

```go
x := pulumi.String("hello").ToStringOutput().ApplyString(func(x string) pulumi.StringOutput {
    return pulumi.String(x).ToStringOutput()
})
```

```
    panic: interface conversion: pulumi.Output is pulumi.AnyOutput, not pulumi.StringOutput
    goroutine 1 [running]:
    github.com/pulumi/pulumi/sdk/v2/go/pulumi.(*OutputState).ApplyString(...)
        /Users/lukehoban/go/pkg/mod/github.com/pulumi/pulumi/sdk/v2@v2.9.2/go/pulumi/types_builtins.go:1015
    main.main.func1(0xc0001aa500, 0x247dfb4, 0x13)
        /Users/lukehoban/dd/pulumidemos/goexamplecam/main.go:19 +0xb26
    github.com/pulumi/pulumi/sdk/v2/go/pulumi.RunWithContext(0xc0001aa500, 0x24ae2e0, 0xc0000381ef, 0xc)
        /Users/lukehoban/go/pkg/mod/github.com/pulumi/pulumi/sdk/v2@v2.9.2/go/pulumi/run.go:101 +0x1fb
    github.com/pulumi/pulumi/sdk/v2/go/pulumi.RunErr(0x24ae2e0, 0x0, 0x0, 0x0, 0x0, 0x0)
        /Users/lukehoban/go/pkg/mod/github.com/pulumi/pulumi/sdk/v2@v2.9.2/go/pulumi/run.go:82 +0x239
    github.com/pulumi/pulumi/sdk/v2/go/pulumi.Run(0x24ae2e0, 0x0, 0x0, 0x0)
        /Users/lukehoban/go/pkg/mod/github.com/pulumi/pulumi/sdk/v2@v2.9.2/go/pulumi/run.go:39 +0x59
    main.main()
        /Users/lukehoban/dd/pulumidemos/goexamplecam/main.go:14 +0x3e
    exit status 2
```

### LINKED ISSUE #9506: Go SDK codegen for resource-valued outputs declares the wrong output types or one that is not chain-able
In awsx, the Go SDK generates this for the `ApplicationLoadBalancer` resource's type. This is a component resource whose outputs are resources, but going from `ResourceOutput` to the resource's outputs is not possible without ApplyT. I'll use the LoadBalancer output for my examples below.

https://github.com/pulumi/pulumi-awsx/blob/e3c920d342be3736be37a22936974cfc255b0718/sdk/go/awsx/lb/applicationLoadBalancer.go#L16-L30

```go
// Provides an Application Load Balancer resource with listeners, default target group and default security group.
type ApplicationLoadBalancer struct {
...
	// Underlying Load Balancer resource
	LoadBalancer lb.LoadBalancerOutput `pulumi:"loadBalancer"`
```

The output `loadBalancer` should actually be a `LoadBalancer` resource:

https://github.com/pulumi/pulumi-awsx/blob/c3c7a654b2d20ce22cbd291e49d37a1214f8ec6b/awsx/schema.json#L2407-L2410

```json
                "loadBalancer": {
                    "$ref": "/aws/v5.3.0/schema.json#/resources/aws:lb%2floadBalancer:LoadBalancer",
                    "description": "Underlying Load Balancer resource"
                },
```

So that we can use Output property chaining in Go, given `lb: ApplicationLoadBalancer`, we should be able to obtain `lb.LoadBalancer.DnsName` without using `ApplyT`.

AaronFriel:
This enables generating output chaining methods such as this if we regenerated the AWS SDK:

```go
// The DNS name of the load balancer.
func (o LoadBalancerOutput) DnsName() pulumi.StringOutput {
	return o.ApplyT(func(v *LoadBalancer) pulumi.StringOutput { return v.DnsName }).(pulumi.StringOutput)
}
```

Enabling ergonomic access to outputs of resource-outputs:

```go
		lb, err := lb.NewApplicationLoadBalancer(ctx, "lb", nil)
		ctx.Export("dnsName", lb.LoadBalancer.DnsName())
```

Fixes #9506
Fixes #6073

### DIFF
--- a/sdk/go/pulumi/types.go
+++ b/sdk/go/pulumi/types.go
@@ -196,19 +196,28 @@ func (o *OutputState) fulfillValue(value reflect.Value, known, secret bool, deps
 			// We didn't get any new dependencies, so no need to merge.
 			return
 		}
-		depSet := make(map[Resource]struct{})
-		for _, d := range o.deps {
-			depSet[d] = struct{}{}
-		}
-		for _, d := range deps {
-			depSet[d] = struct{}{}
-		}
-		mergedDeps := make([]Resource, 0, len(depSet))
-		for d := range depSet {
-			mergedDeps = append(mergedDeps, d)
-		}
-		o.deps = mergedDeps
+		o.deps = mergeDependencies(o.deps, deps)
 	}
+}
+
+func mergeDependencies(ours []Resource, theirs []Resource) []Resource {
+	if theirs == nil {
+		return ours
+	} else if ours == nil {
+		return theirs
+	}
+	depSet := make(map[Resource]struct{})
+	for _, d := range ours {
+		depSet[d] = struct{}{}
+	}
+	for _, d := range theirs {
+		depSet[d] = struct{}{}
+	}
+	mergedDeps := make([]Resource, 0, len(depSet))
+	for d := range depSet {
+		mergedDeps = append(mergedDeps, d)
+	}
+	return mergedDeps
 }
 
 func (o *OutputState) resolve(value interface{}, known, secret bool, deps []Resource) {
@@ -247,11 +256,13 @@ func (o *OutputState) await(ctx context.Context) (interface{}, bool, bool, []Res
 		//
 		// NOTE: this isn't exactly type safe! The element type of the inner output really needs to be assignable to
 		// the element type of the outer output. We should reconsider this.
-		ov, ok := o.value.(Output)
-		if !ok {
+		if ov, ok := o.value.(Output); ok {
+			deps := o.deps
+			o = ov.getState()
+			o.deps = mergeDependencies(o.deps, deps)
+		} else {
 			return o.value, true, o.secret, o.deps, nil
 		}
-		o = ov.getState()
 	}
 }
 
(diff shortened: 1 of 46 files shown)


sdk/go/pulumi/types.go
```diff
@@ -247,11 +256,13 @@ func (o *OutputState) await(ctx context.Context) (interface{}, bool, bool, []Res
...
 		// the element type of the outer output. We should reconsider this.
-		ov, ok := o.value.(Output)
-		if !ok {
+		if ov, ok := o.value.(Output); ok {
+			deps := o.deps
+			o = ov.getState()
+			o.deps = mergeDependencies(o.deps, deps)
+		} else {
 			return o.value, true, o.secret, o.deps, nil
 		}
```
iwahbe, line 265: Nit: I think this style is easier to read. Feel free to ignore. 

```suggestion
		ov, ok := o.value.(Output)
		if !ok {
			return o.value, true, o.secret, o.deps, nil
		}
		deps := o.deps
		o = ov.getState()
		o.deps = mergeDependencies(o.deps, deps)
```

iwahbe (approved):
LGTM.

PR #9514, opened 2022-05-03
Fix when we add default org to stack name

Frassle:
# Description

Fixes an issue in pulumi new where if you tried to give a stack name of "org/stack" and had a default org set it would result in a stack name of "defaultOrg/org/stack" and then complain your org didn't match your project.

## Checklist


Zaid-Ajaj (approved):
LGTM

PR #9515, opened 2022-05-03
Stop using Get/SetRootResource in dotnet

Frassle:
# Description

There's no reason to use this engine call in dotnet, the engine address and the root stack resource are both global variables. No point roundtripping through the engine.

Plan is to also remove this from nodejs (node does have true global variables, so we'll just use that) and then eventually kill this from the engine (probably a 4.0 change).

## Checklist


Zaid-Ajaj (approved):
Good one, let's get this in ✅

PR #9516, opened 2022-05-03
Reads must have unique URNs

Frassle:
# Description

RegisterResource and ReadResource both use the same URN namespace, and so duplicates are not allowed across both and either of them.

## Checklist


pkg/resource/deploy/step_generator.go
```diff
@@ -112,7 +112,15 @@ func (sg *stepGenerator) Errored() bool {
 // GenerateReadSteps is responsible for producing one or more steps required to service
 // a ReadResourceEvent coming from the language host.
 func (sg *stepGenerator) GenerateReadSteps(event ReadResourceEvent) ([]Step, result.Result) {
+	// Generate a URN for this new resource, confirm we haven't seen it before in this deployment.
 	urn := sg.deployment.generateURN(event.Parent(), event.Type(), event.Name())
+	if sg.urns[urn] {
+		// TODO[pulumi/pulumi-framework#19]: improve this error message!
```
t0yv0, line 118: That issue is closed, is this the right ref?

t0yv0:
There is this:

```
	// Walk the root resource's dependents in order and build up the set of resources that require replacement.
	//
	// NOTE: the dependency graph we use for this calculation is based on the dependency graph from the last snapshot.
	// If there are resources in this graph that used to depend on the root but have been re-registered such that they
	// no longer depend on the root, we may make incorrect decisions. To avoid that, we rely on the observation that
	// dependents can only have been _removed_ from the base dependency graph: for a dependent to have been added,
	// it would have had to have been registered prior to the root, which is not a valid operation. This means that
	// any resources that depend on the root must not yet have been registered, which in turn implies that resources
	// that have already been registered must not depend on the root. Thus, we ignore these resources if they are
	// encountered while walking the old dependency graph to determine the set of dependents.
	impossibleDependents := sg.urns
	for _, d := range sg.deployment.depGraph.DependingOn(root, impossibleDependents, false) {
		replace, keys, res := requiresReplacement(d)
		if res != nil {
			return nil, res
		}
		if replace {
			toReplace, replaceSet[d.URN] = append(toReplace, dependentReplace{res: d, keys: keys}), true
		}
	}
```

It uses sg.urns also. I am wondering if the assumptions made in that code still hold with this change. TBH I don't understand this fully, a worked example could be nice. Would ReadResource ever contribute to an entry in `sg.urns` for a resource that actually depends on root, and make some subtle problems here?

pkg/resource/deploy/step_generator.go
Frassle: It's a copy of this code from RegisterResource, I didn't check the issue out. Looks like it's not really fixed, but not sure if we ever will...

t0yv0: Lift to a common func to highlight the sharing?

Frassle: Sure

Frassle:
> It uses sg.urns also. I am wondering if the assumptions made in that code still hold with this change. 

I think that should be fine, the assumptions it makes about resources apply to registers and reads.

PR #9517, opened 2022-05-03
Fix works-when-passphrase-set test

Frassle:
# Description

This test was checking that we could get a secret provider when setting PULUMI_CONFIG_PASSPHRASE, but it set it to the empty string! That's now treated the same as not setting the variable so the test started failing.

## Checklist


Zaid-Ajaj (approved):
Good to go

PR #9544, opened 2022-05-04
Modify make brew to include Java and YAML

### LINKED ISSUE #9538: Brew formula missing pulumi-language-java and pulumi-language-yaml binaries
### What happened?

We expect the homebrew formula to ship pulumi-language-java and pulumi-language-yaml providers together with Pulumi CLI but it does not at the moment. As of 3.32.0.

### Steps to reproduce

Install pulumi view homebrew, run pulumi up on a Java project

### Expected Behavior

Expect pulumi-language-java to be included

### Actual Behavior

No pulumi-language-java included

### Versions used

_No response_

### Additional context

_No response_

### Contributing

t0yv0:
# Description

Fixes #9538

How I tested this:

https://github.com/t0yv0/pulumi has essentially the same change

https://github.com/t0yv0/homebrew-tap has a Formula that's derived from https://github.com/Homebrew/homebrew-core/blob/master/Formula/pulumi.rb which we're trying to make work, but disables checksums and redirects source to t0yv0/pulumi

Then one can test with:

`anton@Antons-MacBook-Pro> brew install t0yv0/tap/pulumi2`

This executes the modified Formula against this change, and you get 

```
which pulumi-language-java
which pulumi-language-yaml
```

Note that I removed the reliance on `gh` CLI on concerns that the environments where the official Brew formula runs do not have it installed. Assume curl instead. Note that computing github release paths in curl is not pretty, and if `gh release install` did checksum validation we're losing that in this version.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,9 +72,8 @@ dist:: build
 	cd pkg && go install -ldflags "-X github.com/pulumi/pulumi/pkg/v3/version.Version=${VERSION}" ${PROJECT}
 
 # NOTE: the brew target intentionally avoids the dependency on `build`, as it does not require the language SDKs.
-brew:: BREW_VERSION := $(shell scripts/get-version HEAD)
 brew::
-	cd pkg && go install -ldflags "-X github.com/pulumi/pulumi/pkg/v3/version.Version=${BREW_VERSION}" ${PROJECT}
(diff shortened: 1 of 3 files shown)


Frassle (approved):
Looks reasonable to me, and testing done sounds fine

PR #9558, opened 2022-05-05
[sdk/nodejs] Avoid wrapping nested resource refs as Output in provider

justinvp:
# Description

TODO: I still need to write a detailed description for this change, but in short, I suspect this relatively simple workaround in the Node.js SDK would address https://github.com/pulumi/pulumi-yaml/issues/193

Fixes # (issue)

## Checklist


lunaris:
Closing this for now as https://github.com/pulumi/pulumi-yaml/issues/193 appears to have been fixed by other means?

PR #9559, opened 2022-05-06
Stop using Get/SetRootResource in nodejs

Frassle:
# Description

Follow on from https://github.com/pulumi/pulumi/pull/9515

Just use globalThis to make the stack resource a global rather than a module variable. This ensures that even if we end up with multiple Pulumi modules they can all grab the stack object and read URN off it.

This is also a temporary fix, we're going to move a lot of the parenting logic to the engine so SDKs won't need this global state.

We've kept the SetRootResource calls in this version of the SDK so that if a user ends up with an old SDK version (that was using the engine calls and module variables) loaded side by side with this version of the SDK both should work together fine.

At the point we change the engine to set parents to the stack if they're parent option is unset we can rip all these calls out completely. Old SDKs that were using Get/SetRootResource will start seeing "undefined" for the GetRootResource call (because it will return NotImplemented and their fallback logic is to then read the module variable that won't have been set) but that won't break anything as they'll just send a RegisterResource call with parent=undefined and the engine will fill it in to the root stack resource.

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,3 +1,6 @@
 ### Improvements
 
+- [nodejs] No longer roundtrips requests for the stack URN via the engine.
+  [#9559](https://github.com/pulumi/pulumi/pull/9559)
+
 ### Bug Fixes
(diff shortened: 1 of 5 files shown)


Frassle:
Thought about this, and think it would be better to do the default parent work in the engine first and then all this can just get ripped out completely all at once.

Issue #9541, opened 2022-05-04
Changed PULUMI_CONFIG_PASSPHRASE behaviour can delete entire state file.


wannessels:
### What happened?

After upgrading to 3.31.1 the state file was emptied for every stack we ran.

We are using pulumi automation (python) and storing our state in an s3 bucket with kms key and appropriate policies. As such we don't need the pulumi passphrase and set it to empty.

Root cause is probably bad error handling in stack.create_or_select , in this case triggered by changed behaviour of empty PULUMI_CONFIG_PASSPHRASE env var, see https://github.com/pulumi/pulumi/pull/9490

### Steps to reproduce

[pulumi_sample.py](https://github.com/pulumi/pulumi/files/8622400/pulumi_sample.py.txt)

first run with 3.30.0
```
>pulumi version
v3.30.0
warning: A new version of Pulumi is available. To upgrade from version '3.30.0' to '3.31.1', visit https://pulumi.com/docs/reference/install/ for manual instructions and release notes.

>python pulumi_sample.py
Updating (stack):

+  pulumi:pulumi:Stack project-stack creating
+  random:index:RandomString random creating
+  random:index:RandomString random created
+  pulumi:pulumi:Stack project-stack created

Resources:
+ 2 created

Duration: 1s
```
then upgrade to 3.31.1
```
>pulumi version
v3.31.1

>python pulumi_sample.py  
Traceback (most recent call last):
  File "pulumi_sample.py", line 24, in <module>
    stack = Stack.create_or_select("stack", workspace)
  File "venv\lib\site-packages\pulumi\automation\_stack.py", line 160, in create_or_select
    return Stack(stack_name, workspace, StackInitMode.CREATE_OR_SELECT)
  File "venv\lib\site-packages\pulumi\automation\_stack.py", line 186, in __init__
    workspace.create_stack(name)
  File "venv\lib\site-packages\pulumi\automation\_local_workspace.py", line 214, in create_stack
    self._run_pulumi_cmd_sync(args)
  File "venv\lib\site-packages\pulumi\automation\_local_workspace.py", line 300, in _run_pulumi_cmd_sync
    return _run_pulumi_cmd(args, self.work_dir, envs, on_output)
  File "venv\lib\site-packages\pulumi\automation\_cmd.py", line 75, in _run_pulumi_cmd
    raise create_command_error(result)
pulumi.automation.errors.CommandError:
 code: 4294967295
 stdout: Created stack 'stack'
[...]

PR #9561, opened 2022-05-06
Add --secrets-provider passthrough option

Frassle:
# Description

This can be used for users who don't want _any_ encryption.

Currently user in this situtation use the passphrase option with an empty passphrase. That's been made more difficult because we've started treating PULUMI_CONFIG_PASSPHRASE="" the same as it not being set at all.

This users can now use passthrough which then doesn't require any environment variables set to work.

If users try to make a passphrase stack with an empty passphrase we'll warn that it's not recomnded:
```
pulumi stack init --secrets-provider=passphrase
Please enter your desired stack name.
To create a stack in an organization, use the format <org-name>/<stack-name> (e.g. `acmecorp/dev`).
stack name: (dev) secrets
Created stack 'secrets'
Enter your passphrase to protect config/secrets:
warning: empty passphrases are not recommended, you probably want to use --secret-provider=passthrough.
```

Two design questions came up while putting this together.
1) Is "passthrough" the right name for this?
2) Should we just leave it in plain text or should we do something to blind the values in the state file like base64 encoding them. I like the former for making it _very_ clear that there is *NO* security with this option, I like the later so you don't have to worry about accidently cat'ing a secret to your screen looking at state files.

Fixes https://github.com/pulumi/pulumi/issues/9541

## Checklist

### DIFF
--- a/pkg/cmd/pulumi/crypto_passthrough.go
+++ b/pkg/cmd/pulumi/crypto_passthrough.go
@@ -1,0 +1,50 @@
+// Copyright 2016-2022, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"github.com/pulumi/pulumi/pkg/v3/secrets"
+	"github.com/pulumi/pulumi/pkg/v3/secrets/passthrough"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/tokens"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/workspace"
+)
+
+func newPassthroughSecretsManager(stackName tokens.Name, configFile string) (secrets.Manager, error) {
+	contract.Assertf(stackName != "", "stackName %s", "!= \"\"")
+
+	if configFile == "" {
+		f, err := workspace.DetectProjectStackPath(stackName.Q())
+		if err != nil {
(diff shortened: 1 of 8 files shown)


pkg/cmd/pulumi/stack_change_secrets_provider.go
```diff
@@ -34,7 +34,7 @@ func newStackChangeSecretsProviderCmd() *cobra.Command {
 		Args:  cmdutil.ExactArgs(1),
 		Short: "Change the secrets provider for the current stack",
 		Long: "Change the secrets provider for the current stack. " +
-			"Valid secret providers types are `default`, `passphrase`, `awskms`, `azurekeyvault`, `gcpkms`, `hashivault`.\n\n" +
+			"Valid secret providers types are `default`, `passphrase`, `awskms`, `azurekeyvault`, `gcpkms`, `hashivault`, 'passthrough'.\n\n" +
```
iwahbe, line 37: I would love to see this turned into a centrally stored list instead of copying these values around and hoping to remember them all for each update.

pkg/cmd/pulumi/crypto_passthrough.go
```diff
@@ -0,0 +1,50 @@
...
+
+	info, err := workspace.LoadProjectStack(configFile)
+	if err != nil {
+		return nil, err
+	}
+
+	// If there are any other secrets providers set in the config, remove them.
+	info.EncryptionSalt = ""
+	info.EncryptedKey = ""
+	info.SecretsProvider = "passthrough"
```
iwahbe, line 44:
```suggestion
	info.SecretsProvider = passthrough.Type
``

iwahbe (commented):
We have a default secrets provider, so maybe instead of "passthrough" it should be "none"? I think the name is non-ideal, but I can't think of a better one so I won't object to it. It makes sense to me to store the "passthrough" state in plain text. I assume that not using a secrets provider indicates a desire to have no security on you secrets (presumably because you will secure it another way).

AaronFriel:
Is choosing "" actually choosing an empty string as a cipher?

If people are choosing this, is this something we want to encourage and encode as an option? I feel like we should just add a `PULUMI_UNSAFE_ALLOW_EMPTY_PASSPHRASE` env var rather than adding an officially sanctioned trivially reversed encryption mode.

Frassle:
> Is choosing "" the same as not encrypting, or is it using a key that's all empty bytes?

It's using a key of zero bytes.

PR #9563, opened 2022-05-06
Fix the function template to show YAML under the YAML tab

cnunciato:
Right now, YAML examples are being shown under the Java tab:

image

Th change fixes that.


pkg/codegen/docs/templates/function.tmpl
```diff
@@ -93,7 +93,7 @@ func </span>{{ .FunctionName.go }}Output<span class="p">(</span>{{ htmlSafe .Fun
 
 <!-- YAML -->
 <div>
-<pulumi-choosable type="language" values="java">
+<pulumi-choosable type="language" values="yaml">
```
t0yv0, line 96: This change looks good the section says YAML.

t0yv0 (approved):

PR #9568, opened 2022-05-07
Allow PULUMI_CONFIG_PASSPHRASE to be empty

Frassle:
# Description

This partially reverts 273fa3343dbbef8a7551b3d8cf87af19bd356e89

I've left the behaviour that PULUMI_CONFIG_PASSPHRASE_FILE being empty is the same as unset because an empty file path isn't valid anyway.

Fixes https://github.com/pulumi/pulumi/issues/9541

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,3 +1,6 @@
 ### Improvements
 
 ### Bug Fixes
+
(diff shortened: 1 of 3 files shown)


mikhailshilkov (approved):
LGTM

PR #9572, opened 2022-05-09
[schema] Move the binder into its own file.

pgavlin:
Just what it says on the tin.


pkg/codegen/schema/bind.go
```diff
@@ -0,0 +1,1321 @@
+package schema
```
mikhailshilkov, line 1: Don't we lint for copyright headers?

mikhailshilkov (approved):
LGTM, I assume it's all mechanical move of code

pkg/codegen/schema/bind.go
pgavlin: I don't think so. Fixed, in any case.

pgavlin:
Azure failure is unrelated.

PR #9591, opened 2022-05-13
Bump versions of semver and OneOf

khellang:
# Description

Both OneOf (https://github.com/mcintyre321/OneOf/pull/102) and semver (https://github.com/maxhauser/semver/pull/66) has recently added a `netstandard2.0` target to their packages, which means they no longer drag along the NETStandard.Library dependency when targeting non-.NET Framework platforms:

image

image

This PR bumps those packages to their latest versions to get rid of the NETStandard.Library dependency and subsequently some vulnerability warnings in Snyk (and other analyzers):

image

## Checklist

### DIFF
--- a/sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
+++ b/sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
@@ -36,7 +36,7 @@ namespace Pulumi.Automation
         private readonly LocalSerializer _serializer = new LocalSerializer();
         private readonly bool _ownsWorkingDir;
         private readonly Task _readyTask;
-        private static readonly SemVersion _minimumVersion = SemVersion.Parse("3.1.0");
+        private static readonly SemVersion _minimumVersion = new SemVersion(3, 1, 0);
 
         /// <inheritdoc/>
         public override string WorkDir { get; }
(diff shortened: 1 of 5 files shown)


khellang:
> * I have added tests that prove my fix is effective or that my feature works

I've assumed that the existing tests already cover the features used in the two libraries affected 😅

sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
```diff
@@ -36,7 +36,7 @@ public sealed class LocalWorkspace : Workspace
         private readonly LocalSerializer _serializer = new LocalSerializer();
         private readonly bool _ownsWorkingDir;
         private readonly Task _readyTask;
-        private static readonly SemVersion _minimumVersion = SemVersion.Parse("3.1.0");
+        private static readonly SemVersion _minimumVersion = new SemVersion(3, 1, 0);
```
khellang, line 39: All these `SemVersion`-related changes were made because `SemVersion.Parse` was made obsolete. The equivalent is passing `SemVersionStyles.Any`, which is the most permissive format.

sdk/dotnet/Pulumi.Automation/LocalWorkspace.cs
```diff
@@ -384,7 +384,7 @@ private async Task PopulatePulumiVersionAsync(CancellationToken cancellationToke
 
         internal static SemVersion? ParseAndValidatePulumiVersion(SemVersion minVersion, string currentVersion, bool optOut)
         {
-            if (!SemVersion.TryParse(currentVersion, out SemVersion? version))
+            if (!SemVersion.TryParse(currentVersion, SemVersionStyles.Any, out SemVersion? version))
```
khellang, line 387: All these `SemVersion`-related changes were made because `SemVersion.Parse(string)` and `SemVersion.Parse(string, bool)` was made obsolete. The equivalent is passing `SemVersionStyles.Any`, which is the most permissive format.

Frassle (approved):

PR #9598, opened 2022-05-13
Cleanup make ensure

iwahbe:
# Description

Part of a general effort to clean up our build process. 

Builds on https://github.com/pulumi/pulumi/pull/9595

## Checklist


t0yv0:
If we continue with Make I'd love one day to work out so we don't run recursive make (call make from make) but have one make. The recursive make is a source of evaluating targets twice.

iwahbe:
> Skimming lightly, looks good. I trust the py changes are autoformatter and no semantic changes there.

The python change is just changing the winner for if the protobuf generator or `black` gets to format the code. No semantic changes. 

> If we continue with Make I'd love one day to work out so we don't run recursive make (call make from make) but have one make. The recursive make is a source of evaluating targets twic

I noticed that. I have some ideas.

PR #9600, opened 2022-05-15
[codegen] Improve handling of resources with outputs that are resources.

AaronFriel:
Code generation fails to bind types on scope traversals through outputs when the output is a resource type and not an object/map/etc.

Needed to fix: https://github.com/pulumi/pulumi-yaml/issues/185

The code generation is currently subpar as we aren't using lifted/proxied property access in each language, and I'm happy to take feedback if it's a small fix. As an example, the Go example should be:

```go
// This:
ctx.Export("simple", rt.Res1.Obj1())
// Not this:
ctx.Export("simple", rt.Res1.ApplyT(func(res1 *resourceproperties.Res1) (resourceproperties.Obj1, -error) {
	return res1.Obj1, nil
}).(resourceproperties.Obj1Output))
```

But I think that fixing this requires a new rewriting pass, akin to the one implemented in Node.

### DIFF
--- a/pkg/codegen/pcl/binder_schema.go
+++ b/pkg/codegen/pcl/binder_schema.go
@@ -254,22 +254,7 @@ func (b *binder) schemaTypeToType(src schema.Type) model.Type {
 				typ = &schema.OptionalType{ElementType: typ}
 			}
 
-			t := b.schemaTypeToType(typ)
-			if prop.ConstValue != nil {
-				var value cty.Value
-				switch v := prop.ConstValue.(type) {
-				case bool:
-					value = cty.BoolVal(v)
-				case float64:
-					value = cty.NumberFloatVal(v)
-				case string:
-					value = cty.StringVal(v)
-				default:
-					contract.Failf("unexpected constant type %T", v)
-				}
-				t = model.NewConstType(t, value)
-			}
-			properties[prop.Name] = t
+			properties[prop.Name] = b.schemaTypeToTypeOrConst(typ, prop)
 		}
 		return objType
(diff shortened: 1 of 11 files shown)


pkg/codegen/pcl/binder_schema.go
```diff
@@ -301,6 +286,23 @@ func (b *binder) schemaTypeToType(src schema.Type) model.Type {
...
+
+		properties := map[string]model.Type{}
+		objType := model.NewObjectType(properties, src)
+		b.schemaTypes[src] = objType
+		for _, prop := range src.Resource.Properties {
+			typ := prop.Type
+			if !prop.IsRequired() {
+				typ = &schema.OptionalType{ElementType: typ}
+			}
+
```
iwahbe, line 302: Do we need to do any input conversions (`model.InputType`)? I don't understand where the `schema.Propertie.Plain` flag is encoded.

iwahbe (changes requested):
I have 1 question. The rest looks great.

pkg/codegen/pcl/binder_schema.go
AaronFriel: I cannot honestly say I know enough about our binding and types to be sure, please let me know if this helps you answer:

This code path and the types we generate here should only appear in output positions: scope traversals/accessors off of a resource node.

The properties we enumerate over / the `typ` here should encode the `plain` flag as well as `default`, `const` fields.

iwahbe (approved):
Thanks for the explanation.

PR #9603, opened 2022-05-16
Pass through resource replacement in the schema

### LINKED ISSUE #9017: Add information to the schema that indicates which resource properties cause replacement
## Hello!

## Issue details

Add information to the schema that indicates which resource properties cause replacement. this is in support of adding an indicator of this to our docs, see https://github.com/pulumi/registry/issues/518 for a bit more context.

iwahbe:
# Description

Fixes #9017 

## Checklist


kimberleyamackenzie (commented):
I don't feel like I have enough context in pulumi/pulumi to give a full approve, but the new class added conditionally works for me on the frontend!  Thank you for taking this on! 🎉 🎉 🎉

t0yv0:
Is it possible to not check in pytest_cache? Otherwise LGTM. Seems like indeed a docs-only feature so unlikely break anything.

t0yv0 (approved):

pkg/codegen/schema/schema.go
```diff
@@ -352,8 +352,10 @@ type Property struct {
 	Secret bool
 	// ReplaceOnChanges specifies if the property is to be replaced instead of updated (default false).
 	ReplaceOnChanges bool
-
-	Plain bool
+	// WillReplaceOnChanges indicates that the provider will replace the resource when
+	// this property is changed. This property is used exclusively for docs.
+	WillReplaceOnChanges bool
```
pgavlin, line 357: Nit: given that this is docs-specific, I might have added this in a docs-specific entry in `Language` to more cleanly separate target-specific info from generic info. This does seem generic enough that I think it's fine where it is, even if it's only used by docs.

PR #9607, opened 2022-05-17
Use append for single strings

abennett:
# Description

Fixes automation API bug where `preview` and `up` were passing incorrect args to `pulumi`.

## Checklist


abennett:
Just a four-line diff. 👍

mikhailshilkov (approved):
LGTM, thanks a lot for this fix!

PR #9610, opened 2022-05-17
Add gzip flag to filestate backend

iwahbe:
# Description

This PR branches from #7153, merging master back in and move where the gzip logic takes place. 

Fixes https://github.com/pulumi/pulumi/issues/5469

## Checklist


pyoio:
Is there an ETA on this being reviewed -> released? We're stuck using GCS backend for the foreseeable future and local executions of pulumi are uploading gigabytes per invocation (the state JSON is about 14mb) when updating only a couple of parameters 😭

tests/stack_test.go
```diff
@@ -524,6 +524,91 @@ func TestLocalStateLocking(t *testing.T) {
...
+	e.RunCommand("pulumi", "up", "--non-interactive", "--yes", "--skip-preview")
+	assertStackFilesOk(true)
+
+	e.RunCommand("pulumi", "up", "--non-interactive", "--yes", "--skip-preview")
+	assertStackFilesOk(true)
+
+	// Disable Gzip compression
+	e.Setenv(filestate.PulumiFilestateGzipEnvVar, "0")
+	e.RunCommand("pulumi", "up", "--non-interactive", "--yes", "--skip-preview")
+	assertStackFilesOk(false)
```
AaronFriel, line 598: Am I reading this right that once a user opts into gzip, disabling causes a failure?

AaronFriel (approved):
I think we might want a flag to indicate they want to decompress/ungzip, otherwise folks are stuck or, I suppose they could:

1. set the flag
2. stack export
3. unset
4. stack import

If that workflow works, I think this is fine if the intent is to surface this as an env var for community evaluation.

tests/stack_test.go
awoimbee: No, this checks that when we re-run `pulumi up` with gzip compression disabled, pulumi reads the stack (whether `.json.gz` or `.json`) and writes `.json` (check the definition of `assertStackFilesOk`).

pkg/backend/filestate/state.go
```diff
@@ -302,16 +321,38 @@ func (b *localBackend) backupStack(name tokens.Name) error {
 	stackFile := filepath.Base(stackPath)
 	ext := filepath.Ext(stackFile)
 	base := strings.TrimSuffix(stackFile, ext)
+	if ext2 := filepath.Ext(base); ext2 != "" && ext == encoding.GZIPExt {
+		// base: stack-name.json, ext: .gz
+		// ->
+		// base: stack-name, ext: .json.gz
+		ext = ext2 + ext
+		base = strings.TrimSuffix(base, ext2)
```
awoimbee, line 329:
```suggestion
		base = strings.TrimSuffix(base, ext2)
		byts = encoding.Gzip(byts)
```
Pulumi works great without this fix as it checks the magic header instead of the file extension to know if a file is gzip'ed. But currently `~/.pulumi/backups/dummy/dummy.1645191346109283284.json.gz` would be written as plaintext json.

awoimbee (approved):
Thanks for this PR, sorry that mine wasn't flawless (the test I wrote needs some work...).
It's been a year since I started using my fork and that's the only issues I found.

tests/stack_test.go
AaronFriel: @iwahbe: A nit then on renaming `assertStackFilesOk` to something like `assertStackFileFormat` and taking a string argument.

sdk/go/common/encoding/marshal.go
```diff
@@ -118,3 +124,65 @@ func (m *yamlMarshaler) Unmarshal(data []byte, v interface{}) error {
...
+	}
+	if err := reader.Close(); err != nil {
+		return err
+	}
+	return m.inner.Unmarshal(inflated, v)
+}
+
+// IsCompressed returns if data is zip compressed.
+func IsCompressed(data []byte) bool {
+	return data[0] == 31 && data[1] == 139
```
t0yv0, line 179: Ah! This is very interesting.

Is there a lib function we can use?

Does this panic on `len(data)< 2`?

t0yv0 (approved):
Looks good modulo a few nits! Thank you!

sdk/go/common/encoding/marshal.go
iwahbe: There is a library function we can repurpose, but will technically violate its contract. This is the check that `compress/gzip` uses internally (constants translated). 
```go
// IsCompressed returns if data is zip compressed.
func IsCompressed(buf []byte) bool {
	// Taken from compress/gzip/gunzip.go
	return len(buf) >= 3 && buf[0] == 31 && buf[1] == 139 && buf[2] == 8
}
```

It is a superset of the previous check.

pkg/backend/filestate/state.go
iwahbe: @awoimbee I don't understand. I'm going to explain my though process. Can you tell me where I went wrong?

 We read from the bucket a file named `stackPath`. We then write it to `backupDir/stackPath`. If the file is `myStack.json.gz`, we then write it directly to `stackPath/myStack.json.gz`. We shouldn't need to compress it since it is already a zip file (since it is named `myStack.json.gz`).

sdk/go/common/encoding/marshal.go
```diff
@@ -175,8 +175,9 @@ func (m *gzipMarshaller) Unmarshal(data []byte, v interface{}) error {
 }
 
 // IsCompressed returns if data is zip compressed.
-func IsCompressed(data []byte) bool {
-	return data[0] == 31 && data[1] == 139
+func IsCompressed(buf []byte) bool {
```
t0yv0, line 178: 👍

pkg/backend/filestate/state.go
awoimbee: Oh yeah, I think you're right. I have some weird files with `.gz` that aren't gzip, but that might be caused by a debug build I ran (and anyways pulumi can read it fine).

PR #9613, opened 2022-05-18
allow `pulumi destroy -s <stack>` if not in a pulumi project dir

### LINKED ISSUE #2440: Can `pulumi destroy -s <owner>/<project>/<name>` work even when Pulumi.yaml is not present?
It would be nice if this could be made to work, and I believe it is now possible.

The current implementation requires that a Pulumi.yaml file be present for a few reasons:

1. It needed a project name, to pass to the engine.  In the past, we picked this off of the `Pulumi.yaml` file. But with the above command, we know the name of the project, because it was passed in as part of the stack identifier.

2. We needed to find the `Pulumi.<stack-name>.yaml` file to pull configuration information out, in order to configure the default providers.  But we now have `pulumi config refresh` which means we should be able to get the configuration for the last deployment from the service and use that instead of needing the file.

3. We needed to pass the root directory of the project to the engine.  This likely makes sense for cases where the engine is going to launch a language provider to actually do some work (i.e. `update` and `preview`) but it is not immediately obvious to me that this would be needed for the `destroy` operation.  It's possible this is needed so we can figure out the set of plugins we needed to create by default.  In fact, this information is retained in the checkpoint itself, so we shouldn't actually need this (and, as I think about it, we had to support a case like this in the past with the PPC and we were able to make things kinda work)

So, I think that with some careful work, we could actually make this work, which would be a pretty nice win.

It would require that you had installed the plugins for all the resource providers you were going to need to do the `destroy` operation, but that feels okay.  Again, we have plugin information in the checkpoint, so it should be possible to discover this without actually having any Pulumi project artifacts present on your disk.

dixler:
# Description

Fixes #2440 

## Checklist

### DIFF
--- a/pkg/backend/httpstate/backend.go
+++ b/pkg/backend/httpstate/backend.go
@@ -107,6 +107,7 @@ type Backend interface {
 
 	CloudURL() string
 
+	GetCloudStackIdentifier(stackRef backend.StackReference) (client.StackIdentifier, error)
 	StackConsoleURL(stackRef backend.StackReference) (string, error)
 	Client() *client.Client
 }
@@ -569,7 +570,7 @@ func (b *cloudBackend) ParseStackReference(s string) (backend.StackReference, er
 		return nil, errors.New("stack names may only contain alphanumeric, hyphens, underscores, and periods")
 	}
 
-	return cloudBackendReference{
+	return CloudBackendReference{
 		owner:   qualifiedName.Owner,
 		project: qualifiedName.Project,
 		name:    tokens.Name(qualifiedName.Name),
@@ -712,6 +713,22 @@ func (b *cloudBackend) GetStack(ctx context.Context, stackRef backend.StackRefer
 	return newStack(stack, b), nil
 }
 
+// Confirm the stack identity matches the environment. e.g. stack init foo/bar/baz shouldn't work
+// if the project name in Pulumi.yaml is anything other than "bar".
+func currentProjectMatchesWorkspace(stack client.StackIdentifier) bool {
+	projPath, err := workspace.DetectProjectPath()
+	if err != nil {
+		return false
+	}
+
+	proj, err := workspace.LoadProject(projPath)
+	if err != nil {
+		return false
+	}
+
+	return proj.Name.String() != stack.Project
+}
+
 func (b *cloudBackend) CreateStack(
 	ctx context.Context, stackRef backend.StackReference, _ interface{} /* No custom options for httpstate backend. */) (
 	backend.Stack, error) {
@@ -725,11 +742,9 @@ func (b *cloudBackend) CreateStack(
 		return nil, fmt.Errorf("error determining initial tags: %w", err)
 	}
 
-	// Confirm the stack identity matches the environment. e.g. stack init foo/bar/baz shouldn't work
-	// if the project name in Pulumi.yaml is anything other than "bar".
-	projNameTag, ok := tags[apitype.ProjectNameTag]
-	if ok && stackID.Project != projNameTag {
-		return nil, fmt.Errorf("provided project name %q doesn't match Pulumi.yaml", stackID.Project)
+	if !currentProjectMatchesWorkspace(stackID) {
+		return nil, fmt.Errorf("provided project name %q doesn't match Pulumi.yaml."+
+			"if this is intentional re-run this command outside of a Pulumi project directory", stackID.Project)
 	}
 
 	apistack, err := b.client.CreateStack(ctx, stackID, tags)
@@ -1313,8 +1328,13 @@ func cleanProjectName(projectName string) string {
 }
 
 // getCloudStackIdentifier converts a backend.StackReference to a client.StackIdentifier for the same logical stack
+func (b *cloudBackend) GetCloudStackIdentifier(stackRef backend.StackReference) (client.StackIdentifier, error) {
+	return b.getCloudStackIdentifier(stackRef)
+}
+
+// getCloudStackIdentifier converts a backend.StackReference to a client.StackIdentifier for the same logical stack
 func (b *cloudBackend) getCloudStackIdentifier(stackRef backend.StackReference) (client.StackIdentifier, error) {
-	cloudBackendStackRef, ok := stackRef.(cloudBackendReference)
+	cloudBackendStackRef, ok := stackRef.(CloudBackendReference)
 	if !ok {
 		return client.StackIdentifier{}, errors.New("bad stack reference type")
 	}
--- a/pkg/cmd/pulumi/destroy.go
+++ b/pkg/cmd/pulumi/destroy.go
@@ -23,13 +23,32 @@ import (
 
 	"github.com/pulumi/pulumi/pkg/v3/backend"
 	"github.com/pulumi/pulumi/pkg/v3/backend/display"
+	"github.com/pulumi/pulumi/pkg/v3/backend/httpstate"
 	"github.com/pulumi/pulumi/pkg/v3/engine"
 	"github.com/pulumi/pulumi/pkg/v3/resource/graph"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/tokens"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/cmdutil"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/result"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/workspace"
 )
+
+func readProjectStackRef(stack string, s backend.Stack) (*workspace.Project, string, error) {
+	proj, root, err := readProject()
+	if stack == "" {
+		return proj, root, err
+	}
+
+	cloudBackendStackRef, ok := s.Ref().(httpstate.CloudBackendReference)
+	if !ok {
+		return nil, "", errors.New("destroying stacks by reference is currently only supported on pulumi service backends")
+	}
+
+	return &workspace.Project{
+		Name: tokens.PackageName(cloudBackendStackRef.Project()),
+	}, "", nil
+}
 
 func newDestroyCmd() *cobra.Command {
 	var debug bool
@@ -124,7 +143,7 @@ func newDestroyCmd() *cobra.Command {
 			if err != nil {
 				return result.FromError(err)
 			}
-			proj, root, err := readProject()
+			proj, root, err := readProjectStackRef(stack, s)
 			if err != nil {
 				return result.FromError(err)
 			}
@@ -135,12 +154,12 @@ func newDestroyCmd() *cobra.Command {
 			}
 
 			sm, err := getStackSecretsManager(s)
-			if err != nil {
+			if err != nil && stack == "" {
 				return result.FromError(fmt.Errorf("getting secrets manager: %w", err))
 			}
 
 			cfg, err := getStackConfiguration(s, sm)
-			if err != nil {
+			if err != nil && stack == "" {
 				return result.FromError(fmt.Errorf("getting stack configuration: %w", err))
 			}
 
--- a/pkg/backend/httpstate/stack.go
+++ b/pkg/backend/httpstate/stack.go
@@ -39,14 +39,14 @@ type Stack interface {
 	StackIdentifier() client.StackIdentifier
 }
 
-type cloudBackendReference struct {
+type CloudBackendReference struct {
 	name    tokens.Name
 	project string
 	owner   string
 	b       *cloudBackend
 }
 
-func (c cloudBackendReference) String() string {
+func (c CloudBackendReference) String() string {
 	// If the project names match, we can elide them.
 	if c.b.currentProject != nil && c.project == string(c.b.currentProject.Name) {
 
@@ -69,14 +69,22 @@ func (c cloudBackendReference) String() string {
 	return fmt.Sprintf("%s/%s/%s", c.owner, c.project, c.name)
 }
 
-func (c cloudBackendReference) Name() tokens.Name {
+func (c CloudBackendReference) Name() tokens.Name {
 	return c.name
+}
+
+func (c CloudBackendReference) Project() string {
+	return c.project
+}
+
+func (c CloudBackendReference) Owner() string {
+	return c.owner
 }
 
 // cloudStack is a cloud stack descriptor.
 type cloudStack struct {
 	// ref is the stack's unique name.
-	ref cloudBackendReference
+	ref CloudBackendReference
 	// orgName is the organization that owns this stack.
 	orgName string
 	// currentOperation contains information about any current operation being performed on the stack, as applicable.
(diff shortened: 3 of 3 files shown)


pkg/cmd/pulumi/destroy.go
```diff
@@ -23,14 +23,33 @@ import (
...
+	cloudBackendStackRef, ok := s.Ref().(httpstate.CloudBackendReference)
+	if !ok {
+		return nil, "", errors.New("destroying stacks by reference is currently only supported on pulumi service backends")
+	}
+
+	return &workspace.Project{
+		Name: tokens.PackageName(cloudBackendStackRef.Project()),
+	}, "", nil
+}
+
```
dixler, line 52: I feel like there is a better approach.

pkg/backend/httpstate/backend.go
```diff
@@ -712,6 +712,28 @@ func (b *cloudBackend) GetStack(ctx context.Context, stackRef backend.StackRefer
 	return newStack(stack, b), nil
 }
 
+// Confirm the stack identity doesn't contradict the environment.
+// e.g. stack init foo/bar/baz shouldn't work
+// if the project name in Pulumi.yaml is anything other than "bar".
+// if the CWD is not a Pulumi project, it doesn't contradict.
```
dixler, line 718: Refactor comments

pkg/cmd/pulumi/destroy.go
```diff
@@ -26,11 +26,43 @@ import (
...
+type panicSecretsManager struct {
+	crypter config.Crypter
+}
+
+func (m *panicSecretsManager) Type() string                         { return "" }
+func (m *panicSecretsManager) State() interface{}                   { return nil }
+func (m *panicSecretsManager) Encrypter() (config.Encrypter, error) { return m.crypter, nil }
+func (m *panicSecretsManager) Decrypter() (config.Decrypter, error) { return m.crypter, nil }
+func (m *panicSecretsManager) EncryptedKey() []byte                 { return []byte{} }
+
```
dixler: delete

pkg/cmd/pulumi/destroy.go
```diff
@@ -159,6 +181,13 @@ func newDestroyCmd() *cobra.Command {
 				return nil
 			}
 
+			sm := snap.SecretsManager
```
dixler: just reference directly

sdk/go/common/workspace/paths.go
```diff
@@ -125,9 +125,17 @@ func DetectProjectStackPath(stackName tokens.QName) (string, error) {
 // DetectProjectPathFrom locates the closest project from the given path, searching "upwards" in the directory
 // hierarchy.  If no project is found, an empty path is returned.
 func DetectProjectPathFrom(path string) (string, error) {
-	return fsutil.WalkUp(path, isProject, func(s string) bool {
+	path, err := fsutil.WalkUp(path, isProject, func(s string) bool {
```
dixler: this doesn't return an error and is intended not to. Handle the error in the caller since the caller is incorrectly assuming errors if not found.

pkg/cmd/pulumi/destroy.go
```diff
@@ -29,8 +29,23 @@ import (
...
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/result"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/workspace"
 )
 
+func readProjectStackRef(stack string, s backend.Stack) (*workspace.Project, string, error) {
+	proj, root, err := readProject()
+	if stack == "" {
+		return proj, root, err
+	}
```
dixler: move 36 into conditional

pkg/cmd/pulumi/destroy.go
```diff
@@ -29,8 +29,23 @@ import (
...
+
+	proj, err = s.Project()
+	if err != nil {
+		return nil, "", err
+
+	}
+	return proj, "", nil
```
dixler, line 45: there is no root since we're specifying the stack.

pkg/cmd/pulumi/destroy.go
```diff
@@ -134,20 +149,11 @@ func newDestroyCmd() *cobra.Command {
 				return result.FromError(fmt.Errorf("gathering environment metadata: %w", err))
 			}
 
-			sm, err := getStackSecretsManager(s)
-			if err != nil {
-				return result.FromError(fmt.Errorf("getting secrets manager: %w", err))
-			}
-
```
dixler, line 141: We now get the secretsManager directly from the snapshot that the backend recorded rather than the current project.

unsure if there are edge cases where we want users to be able to have the local directory dictate destroy behavior rather than using the backend's state to unwind.

pkg/cmd/pulumi/destroy.go
```diff
@@ -134,20 +149,11 @@ func newDestroyCmd() *cobra.Command {
...
-			cfg, err := getStackConfiguration(s, sm)
-			if err != nil {
-				return result.FromError(fmt.Errorf("getting stack configuration: %w", err))
-			}
-
```
dixler, line 146: moved lower to use the snapshot's secretsManager

pkg/cmd/pulumi/destroy.go
```diff
@@ -134,20 +149,11 @@ func newDestroyCmd() *cobra.Command {
...
 			snap, err := s.Snapshot(commandContext())
 			if err != nil {
 				return result.FromError(err)
 			}
+
```
dixler: extraneous line

pkg/cmd/pulumi/crypto.go
```diff
@@ -49,6 +50,16 @@ func getStackDecrypter(s backend.Stack) (config.Decrypter, error) {
...
+		switch s := s.(type) {
+		case filestate.Stack:
+			return newPassphraseSecretsManager(s.Ref().Name(), stackConfigFile,
+				false /* rotatePassphraseSecretsProvider */)
+		case httpstate.Stack:
+			client := s.Backend().(httpstate.Backend).Client()
+			id := s.StackIdentifier()
+
+			return service.NewServiceSecretsManager(client, id)
+		}
```
dixler: no longer using, clean up.

pkg/backend/httpstate/backend.go
dixler, line 718: rewrite comment

pkg/backend/stack.go
```diff
@@ -171,7 +172,8 @@ func GetEnvironmentTagsForCurrentStack() (map[apitype.StackTagName]string, error
 	// Tags based on Pulumi.yaml.
 	projPath, err := workspace.DetectProjectPath()
 	if err != nil {
-		return nil, err
```
dixler, line 174: Maybe wrap this function in `GetEnvironmentTagsForStack(stackRef StackReference)` then fallback to `GetEnvironmentTagsForCurrentStack()`

pkg/backend/stack.go
```diff
@@ -34,6 +34,7 @@ import (
 // Stack is used to manage stacks of resources against a pluggable backend.
 type Stack interface {
 	Ref() StackReference                                    // this stack's identity.
+	Project() (*workspace.Project, error)                   // this stack's identity.
```
dixler, line 37: I don't like this.

pkg/backend/stack.go
```diff
@@ -171,7 +172,8 @@ func GetEnvironmentTagsForCurrentStack() (map[apitype.StackTagName]string, error
 	// Tags based on Pulumi.yaml.
 	projPath, err := workspace.DetectProjectPath()
 	if err != nil {
-		return nil, err
+		// No current stack return empty
+		return make(map[apitype.StackTagName]string), nil
```
dixler, line 176: has more logical error handling imo than having the caller determine empty tagset.

sdk/go/common/workspace/paths.go
```diff
@@ -85,6 +85,12 @@ func DetectProjectPath() (string, error) {
 	path, err := DetectProjectPathFrom(dir)
 	if err != nil {
 		return "", err
+
+	}
+	if path == "" {
+		return "", fmt.Errorf(
+			"no Pulumi.yaml project file found (searching upwards from %s). If you have not "+
+				"created a project yet, use `pulumi new` to do so", dir)
```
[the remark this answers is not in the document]
dixler: Also fixes assumption that `DetectProjectPathFrom(dir)` returns an error if a match isn't found.

pkg/cmd/pulumi/destroy.go
```diff
@@ -159,6 +158,11 @@ func newDestroyCmd() *cobra.Command {
 				return nil
 			}
 
+			cfg, err := getStackConfiguration(s, snap.SecretsManager)
+			if err != nil {
+				return result.FromError(fmt.Errorf("getting stack configuration: %w", err))
+			}
+
```
dixler, line 165: We now get the secretsManager directly from the snapshot that the backend recorded rather than the current project.

unsure if there are edge cases where we want users to be able to have the local directory dictate destroy behavior rather than using the backend's state to unwind.

pkg/backend/stack.go
dixler: Should project be exposed at such a high level? It seems to be an abstraction by the pulumi service on top of their stack naming conventions :woman_shrugging: 

I wonder what apply would look like if we were to remove workspace.Project from it.

sdk/go/common/workspace/paths.go
```diff
@@ -125,9 +125,18 @@ func DetectProjectStackPath(stackName tokens.QName) (string, error) {
...
 	})
+	if err != nil {
+		return "", err
+	}
+	if path == "" {
+		return "", fmt.Errorf(
+			"no Pulumi.yaml project file found (searching upwards from %s). If you have not "+
+				"created a project yet, use `pulumi new` to do so", path)
+	}
+	return path, nil
```
dixler, line 139: added error handling

pkg/backend/httpstate/backend.go
```diff
@@ -712,6 +712,29 @@ func (b *cloudBackend) GetStack(ctx context.Context, stackRef backend.StackRefer
...
+	projPath, err := workspace.DetectProjectPath()
+	if err != nil {
+		return false
+	}
+
+	if projPath == "" {
+		return false
+	}
+
+	proj, err := workspace.LoadProject(projPath)
```
t0yv0, line 730: Is there a reason we can't use `b.currentProject` of type `*workspace.Project` and we need to re-detect and load it again from ambient information?

pkg/cmd/pulumi/config.go
```diff
@@ -913,17 +913,24 @@ func looksLikeSecret(k config.Key, v string) bool {
...
 // it is uses instead of the default configuration file for the stack
 func getStackConfiguration(stack backend.Stack, sm secrets.Manager) (backend.StackConfiguration, error) {
+	var cfg config.Map
 	workspaceStack, err := loadProjectStack(stack)
 	if err != nil {
-		return backend.StackConfiguration{}, fmt.Errorf("loading stack configuration: %w", err)
+		// if project is unavailable, fallback to get the project's configuration from the backend
+		cfg, err = backend.GetLatestConfiguration(commandContext(), stack)
+		if err != nil {
+			return backend.StackConfiguration{}, fmt.Errorf("loading stack configuration: %w", err)
```
t0yv0, line 922: Nit: perhaps combine the two errors somehow (from loadProjectStack and backend.GetLatestConfiguration). Currently only the second error is reported. Or if there's logging around could use verbose logging for the ignored error.

pkg/cmd/pulumi/destroy.go
t0yv0: Is it that `getStackSecretsManager(s)` fails in the `pulumi destroy -s <stack>` scenario out of dir?

t0yv0:
Need a test covering the desired functionality. 

The changes look workable but I hesitate a little - not very familiar with this part of the codebase yet, and if the ordering of loading effects matters.

Also - to clarify - the desired functionality only works for the http backend? Or it works for filestate backend also?

pkg/cmd/pulumi/destroy.go
dixler: I think we need to decrypt secrets in order to get the config to run the engine.

dixler:
> Need a test covering the desired functionality. 
> 
> 
> 
> The changes look workable but I hesitate a little - not very familiar with this part of the codebase yet, and if the ordering of loading effects matters.
> 
> 
> 
> Also - to clarify - the desired functionality only works for the http backend? Or it works for filestate backend also?
> 
> 

Thanks. Will do. The filestate backend also is able to delete stacks via `-s`

iwahbe:
@dixler Thanks for the explanation.

pkg/backend/httpstate/backend.go
```diff
@@ -904,6 +924,10 @@ func (b *cloudBackend) createAndStartUpdate(
 	if err != nil {
 		return client.UpdateIdentifier{}, 0, "", err
 	}
+	if currentProjectContradictsWorkspace(stackID) {
+		return client.UpdateIdentifier{}, 0, "", fmt.Errorf(
+			"provided project name %q doesn't match Pulumi.yaml", stackID.Project)
```
t0yv0, line 929: ```%q	a single-quoted character literal safely escaped with Go syntax.```

Learning something new here.

t0yv0 (approved):
Looking good to me. If there's another reviewer more familiar with this part of the code I'd appreciate that.

pkg/cmd/pulumi/config.go
```diff
@@ -913,17 +913,25 @@ func looksLikeSecret(k config.Key, v string) bool {
 // getStackConfiguration loads configuration information for a given stack. If stackConfigFile is non empty,
 // it is uses instead of the default configuration file for the stack
 func getStackConfiguration(stack backend.Stack, sm secrets.Manager) (backend.StackConfiguration, error) {
+	var cfg config.Map
 	workspaceStack, err := loadProjectStack(stack)
 	if err != nil {
-		return backend.StackConfiguration{}, fmt.Errorf("loading stack configuration: %w", err)
+		// if project is unavailable, fallback to get the project's configuration from the backend
+		cfg, err = backend.GetLatestConfiguration(commandContext(), stack)
```
Frassle, line 920: This is the most worrying change for me. Currently we do not treat config stored in the backend as authoritative, if the user is missing config locally we error. With this change we'll now carry on with the config stored in the backend.

It's probably a horrible amount of re-plumbing but I wonder if we could somehow just not hit this code path for getStackConfiguration when calling "destroy"?

pkg/cmd/pulumi/destroy.go
Frassle: I think we could maybe not load config here for destroy and just default to an empty config map?

pkg/cmd/pulumi/config.go
dixler: I'm not really that familiar with this yet, but why isn't the config stored in the backend authoritative if there is no alternative? Is it not the last configuration that was used to deploy the stack?

I could double check if calling destroy with an empty configuration works, but would it have issues calculating things if config was blank?

Frassle: I can't recall the reasoning I asked this _years_ ago. @lukehoban or @pgavlin might have comments on if it's ok or not for us to start using backend config like this.

pgavlin: I think that using this if and only if we're not under the stack's project's directory is fine.

If local configuration is available, then we treat it as authoritative for two reasons:
1. it allows the user to change configuration for the destroy (e.g. the user can replace short-lived authentication tokens)
2. w.r.t. the model, a destroy is essentially just an `update` with a null pulumi program, and the config remains an input into the `update`.

PR #9615, opened 2022-05-18
Fix python automation api text coloring option

moralex1:
# Description

I want to have text coloring using automation API and Python in order to help understand what is changed in the CI. I saw that the color option was ignored in automation API stack operations. I added it.

Fixes # (issue)

## Checklist

    * I could not find any test that tested automation API argument parsing

    * This is a minor change that does not affect the service.


Zaid-Ajaj:
Hi @moralex1, thanks a lot for the contribution 🙏 the changes look good and it is good to go ✅

PR #9617, opened 2022-05-18
Fixes checksum validation issues and factors out publish-binaries.yml

### LINKED ISSUE #9602: Master builds are failing with checksum errors
A few recent master builds have failed with

image

https://github.com/pulumi/pulumi/runs/6431805010?check_suite_focus=true#step:15:29

t0yv0:
# Description

Fixes #9602

It seems that we do not pin goreleaser and upgrades to 1.8.3 changed the binary naming, which defeated the `go-wrapper.sh` prebuilt logic. As a result we stopped reusing prebuilt binaries and sometimes would rebuild binaries in the publish step. This also slowed down the publish step significantly. Due to differences in go cache this would result in mismatch.

These kinds of changes are difficult to test but I've tested up to matching checksums by invoking the master.yml-like configuration from the `run-build-and-acceptance-tests.yml` script.

As a bonus I tried factoring out the publishing into a reusable workflow. This way there is a bit more hope that whatever we fix there goes out to all of master.yml, prerelease.yml, release.yml.

## Checklist

### DIFF
--- a/.github/workflows/publish-binaries.yml
+++ b/.github/workflows/publish-binaries.yml
@@ -89,6 +89,7 @@ jobs:
       - name: Inspect goreleaser-prebuilt
         run: |
           find goreleaser-prebuilt
+
       # Section 3: dry-run goreleaser to verify checksums
       - name: Download pulumi-windows-checksums
         uses: actions/download-artifact@v2
--- a/.github/workflows/publish-binaries.yml
+++ b/.github/workflows/publish-binaries.yml
@@ -133,7 +133,7 @@ jobs:
         uses: goreleaser/goreleaser-action@v2
         with:
           version: latest
-          args: -p 3 -f ${{ inputs.goreleaser-config }} ${{ inputs.goreleaser-flags}}
+          args: -p 3 -f ${{ inputs.goreleaser-config }} ${{ inputs.goreleaser-flags }} --rm-dist
       - uses: actions/upload-artifact@v3
         with:
           name: goreleaser-artifacts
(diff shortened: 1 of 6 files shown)


.github/workflows/publish-binaries.yml
```diff
@@ -0,0 +1,148 @@
...
+      - name: Install goreleaser-filter
+        uses: jaxxstorm/action-install-gh-release@v1.5.0
+        with:
+          repo: t0yv0/goreleaser-filter
+
+      # Section 1: configure
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
```
t0yv0, line 47: Will this be able to access secrets or do we need to thread through secrets: in the on: section I wonder.

scripts/go-wrapper.sh
```diff
@@ -28,8 +28,13 @@ case "$1" in
         ARGS=( "$@" )
         BUILDDIR=${ARGS[${#ARGS[@]}-1]}
         OUTPUT=${ARGS[${#ARGS[@]}-2]}
+
         PREBUILT="${OUTPUT/goreleaser/goreleaser-prebuilt}"
 
+        # Since at least goreleaser 1.8.3 binaries are named
```
t0yv0, line 34: This is the fix.

.github/workflows/publish-binaries.yml
t0yv0: According to the docs `secrets: inherit` should do the trick.

dixler (approved):

PR #9620, opened 2022-05-18
[schema] Add support for on-demand binding.

pgavlin:
These changes extend the public API of `pkg/codegen/schema` to support
on-demand binding of package members. On-demand binding is appropriate for
scenarios that do not require the entire package, especially those such as
program code generation or the YAML LSP server that require only specific
types/functions/etc.

The extensions to the public API consist of two new types and several new
methods. The most notable of these are `PackageReference` and
`Loader.LoadPackageReference`. The former provides the on-demand binding
interface, while the latter creates instances of the former (n.b. it was
my intent to make a breaking change to the signature of `Loader.LoadPackage`
s.t. it returns a `PackageReference`, but the circular dependency between
this Go module and those for YAML and Java prevented that change).

These changes _dramatically_ reduce the memory required to interace with
Pulumi Packages, and only require memory proportional to the number of
accessed package members. We may be able to improve on this in the future
by removing type/resource/function interning, which would allow those
values to be garbage collected at a granaular level rather than at a
package level. That is a more radical change, though, as it requires new
equality semantics for each of the affected types (some of which are
currently used as map keys).

### DIFF
--- a/pkg/codegen/go/gen_program.go
+++ b/pkg/codegen/go/gen_program.go
@@ -429,7 +429,7 @@ func (g *generator) collectConvertImports(
 }
 
 func (g *generator) getVersionPath(program *pcl.Program, pkg string) (string, error) {
-	for _, p := range program.Packages() {
+	for _, p := range program.PackageReferences() {
 		if p.Name() == pkg {
 			if ver := p.Version(); ver != nil && ver.Major > 1 {
 				return fmt.Sprintf("/v%d", ver.Major), nil
@@ -907,7 +907,7 @@ func (g *generator) getModOrAlias(pkg, mod string) string {
 //
 // TODO: naming decisions should really be encoded statically so that they can be decided locally.
 func programPackageDefs(program *pcl.Program) ([]*schema.Package, error) {
-	refs := program.Packages()
+	refs := program.PackageReferences()
 	defs := make([]*schema.Package, len(refs))
 	for i, ref := range refs {
 		def, err := ref.Definition()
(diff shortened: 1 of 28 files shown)


pkg/codegen/go/gen_program.go
```diff
@@ -895,3 +902,19 @@ func (g *generator) getModOrAlias(pkg, mod string) string {
 	}
 	return mod
 }
+
+// Go needs complete package definitions in order to properly resolve names.
```
AaronFriel, line 906: What info is missing?

AaronFriel (approved):
Looks good to me if tests pass, but I think with the release issues we should hold off on merging until after the next release.

iwahbe (approved):
I'm exited for this to land.

pgavlin:
Okay, the codegen tests are failing due to [rate limit issues when creating the `Diff for...` comment](https://github.com/pulumi/pulumi/runs/6562534114?check_suite_focus=true#step:8:7287), but the results themselves look clean so far (see e.g. [aws](https://github.com/pulumi/pulumi-aws/compare/master...pulumi-bot:integration/pulumi-terraform-bridge/551071edcee1565ace178f6c0912e51494d94477)).

The current set of changes does not include a few breaks that I'd like to make:
- I'd like `schema.Loader.LoadPackage` to return a `PackageReference`. This is blocked by the circular dependency between p/p and YAML/Java. It would also temporarily break codegen tests.
- I'd like `pcl.Program.Packages` to return a `[]schema.PackageReference`. Also blocked by the circular dependency.
- I'd like `schema.PackageSpec` to embed a `schema.PackageInfoSpec` instead of manually inlining its fields in order to remove duplication. This would temporarily break codegen tests.

I am fine with making those breaking changes later on.

t0yv0:
Nice! Are the downstream failures because of the braking changes?

We should try to remove the dependency loop soon.

pgavlin:
> Nice! Are the downstream failures because of the braking changes?

The downstream failures are failing due to [rate limit issues when creating the Diff for... comment](https://github.com/pulumi/pulumi/runs/6562534114?check_suite_focus=true#step:8:7287), but the results themselves look clean so far (see e.g. [aws](https://github.com/pulumi/pulumi-aws/compare/master...pulumi-bot:integration/pulumi-terraform-bridge/551071edcee1565ace178f6c0912e51494d94477)).

PR #9641, opened 2022-05-20
added confirmation string to `pulumi logout`

### LINKED ISSUE #9450: It would be best to have "Successfully logged out" message when logged out
## Hello!

## Issue details

I'm using **v3.30.0**.

`pulumi logout` gives us no message that tells us if we are logged out or not.

Current behaviour:

```
me@MacBook-Pro ~/r/m/myfolder> pulumi logout
me@MacBook-Pro ~/r/m/myfolder>
```

Better behaviour that I hope:

```
me@MacBook-Pro ~/r/m/myfolder> pulumi logout
Successfully logged out of https://app.pulumi.com/matoruru!
me@MacBook-Pro ~/r/m/myfolder>
```

dixler:
# Description
Adds a confirmation string to `pulumi logout` to add more clarity/confidence in the state after the command exits.

Fixes #9450 

## Checklist

### DIFF
--- a/pkg/cmd/pulumi/logout.go
+++ b/pkg/cmd/pulumi/logout.go
@@ -74,6 +74,7 @@ func newLogoutCmd() *cobra.Command {
 			var be backend.Backend
 			var err error
 			if filestate.IsFileStateBackendURL(cloudURL) {
+				fmt.Printf("Logged out of %s\n", cloudURL)
 				return workspace.DeleteAccount(cloudURL)
 			}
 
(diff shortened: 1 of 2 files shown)


iwahbe (approved):

susanev (approved):

pkg/cmd/pulumi/logout.go
```diff
@@ -74,6 +74,7 @@ func newLogoutCmd() *cobra.Command {
 			var be backend.Backend
 			var err error
 			if filestate.IsFileStateBackendURL(cloudURL) {
+				fmt.Printf("Logged out of %s\n", cloudURL)
```
joeduffy, line 77: Didn’t notice this at the time, but informational messages like this really ought to go to stderr.

dixler: Semi-related, but I was talking with @susanev around what the CLI pipe experience might be like since every change to output could be a breaking change from that perspective.

PR #9645, opened 2022-05-20
more confident CLI strings

### LINKED ISSUE #9643: make CLI more self-confident
## Hello!

## Issue details
We should remove `successfully` from strings we output to the user as it's redundant. If it's unsuccessful, we print an error message.

dixler:
# Description
Ideation on making the CLI more confident.

Fixes #9643 

## Checklist

### DIFF
--- a/pkg/engine/plugins.go
+++ b/pkg/engine/plugins.go
@@ -203,7 +203,7 @@ func installPlugin(plugin workspace.PluginInfo) error {
 
 	}
 
-	logging.V(7).Infof("installPlugin(%s, %s): successfully installed", plugin.Name, plugin.Version)
+	logging.V(7).Infof("installPlugin(%s, %s): installed", plugin.Name, plugin.Version)
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


pkg/engine/plugins.go
```diff
@@ -203,7 +203,7 @@ func installPlugin(plugin workspace.PluginInfo) error {
 
 	}
 
-	logging.V(7).Infof("installPlugin(%s, %s): successfully installed", plugin.Name, plugin.Version)
+	logging.V(7).Infof("installPlugin(%s, %s): installation complete", plugin.Name, plugin.Version)
```
dixler, line 206: when not if

Frassle (commented):
Message changes so far look good to me :)
You keeping this as a draft to change more or do you want to get this set in as is?

dixler:
@Frassle I want to leave it as a draft for now. Thanks for asking.

t0yv0 (approved):

PR #9646, opened 2022-05-20
Enable cache on gotestsum

t0yv0:
# Description

Missed one more cache opportunity.

Fixes # (issue)

## Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -220,7 +220,7 @@ jobs:
           tag: v0.0.31
           cache: enable
       - name: Install gotestsum
-        uses: aaronfriel/action-install-gh-release@v1.4.0-beta.4
+        uses: jaxxstorm/action-install-gh-release@v1.7.1
         with:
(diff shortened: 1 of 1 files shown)


dixler (approved):

.github/workflows/test.yml
```diff
@@ -224,6 +224,7 @@ jobs:
         with:
           repo: gotestyourself/gotestsum
           tag: v1.8.1
+          cache: enable
```
t0yv0, line 227: Actually something is off here, why are we using aaronfriel/action-install-gh-release@v1.4.0-beta.4 and not jaxxstorm/action-install-gh-release@v1.7.1

@AaronFriel  would you know ?

t0yv0: Discussed that this change was upstreamed, so we do not need to reference the fork anymore.

AaronFriel: needed to grab .zip files for gotestsum on Windows, IIRC. it's upstreamed now.

PR #9651, opened 2022-05-21
[codegen] Potpourri of codegen improvements for `pulumi convert`

AaronFriel:
* [codegen] Fix ImportLanguages map usage in python, nodejs
* [codegen] Add typescript dependency to packages during convert
* [testing] Add non-destructive TypeCheck operations for pulumi convert testing

These are used in:
- pulumi/pulumi-yaml#195


iwahbe (approved):
Looks good to me. LGTM

Issue #9457, opened 2022-04-23
Multi-Language Component Cannot be a Parent of Another Resource - Causes Promise Leak


phillipedwards:
### What happened?

Using a multi-language component as the `parent` property to another resource causes a fatal error (promise leak).

### Steps to reproduce

1. Install a Pulumi component, in  your Pulumi program, which is generated from a MLC. Eg- https://www.pulumi.com/registry/packages/aws-s3-replicated-bucket/.
2. Create another resource, such as a bucket object.
3. Set the bucket object's `parent` resourceOption to the MLC.
4. See promise leak

### Expected Behavior

MLC should be able to be set as parent of another resource.

### Actual Behavior

MLC as a parent of another resource causes promise leak and fatal error.

### Versions used

CLI Version                        3.30.0
aws                                    5.2.0
aws-s3-replicated-bucket  0.0.6

### Additional context

Sample program to reproduce error:
```typescript
import * as pulumi from "@pulumi/pulumi";
import * as aws from "@pulumi/aws";
import * as s3 from "@pulumi/aws-s3-replicated-bucket";
import { PathLike, promises as fs } from "fs";

const uploadObjectsToBucket = async (bucket: s3.ReplicatedBucket, dir: PathLike) => {
    const objects = await fs.readdir(dir, { withFileTypes: true });
    for (const object of objects) {
        const path = `${dir}/${object.name}`;
        new aws.s3.BucketObject(object.name, {
            bucket: bucket.sourceBucket.id,
            source: new pulumi.asset.FileAsset(path)
        }, {
            parent: bucket // commenting 'parent' will stop error
        });
    }
}

export = async () => {

    const mlcBucket = new s3.ReplicatedBucket("rep-bucket", {
        destinationRegion: "us-east-1"
    });

    await uploadObjectsToBucket(mlcBucket, "./content")
}
```

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

Frassle:
I think this might just be a manifestation of the more general issues with setting parent to any component resource. It looks like the BucketObject is trying to register before the component is constructed and then a load of properties are just being filled in as the system also tries to shutdown.

PR #9665, opened 2022-05-24
Don't add children of remote componet resources

### LINKED ISSUE #9071: Python hangs on specific inputs
Thanks @iwahbe @justinvp – both of those workarounds are OK for the repro example. In practice, I've tried to use a new Kubernetes provider to create cluster resources and have come up against another issue. This seems to be because cluster resources like `Namespace` have `parent=cluster` and this is interacting with providers. When we run `pulumi preview` it just hangs.

To reproduce:
1. With `pulumi==3.22.1` (before #8770) run `pulumi up` with the following program:
     ```python
    import pulumi
    import pulumi_eks as eks
    import pulumi_kubernetes as k8s
    
    cluster = eks.Cluster(resource_name="test")
    
    namespace = k8s.core.v1.Namespace(
        resource_name="test", opts=pulumi.ResourceOptions(provider=cluster.provider, parent=cluster)
    )
    ```
2. Upgrade to `pulumi==3.25.0`.
3. Implement the workaround and run `pulumi preview`:
    ```python
    import json
    
    import pulumi_eks as eks
    import pulumi_kubernetes as k8s
    import pulumi
    
    cluster = eks.Cluster(resource_name="test")
    
    provider = k8s.Provider(
        resource_name="test", kubeconfig=cluster.kubeconfig.apply(lambda k: json.dumps(k))
    )
    
    namespace = k8s.core.v1.Namespace(
        resource_name="test",
        opts=pulumi.ResourceOptions(provider=provider, parent=cluster),
    )
    ```

_Originally posted by @calumabarnett in https://github.com/pulumi/pulumi/issues/9024#issuecomment-1053499137_

iwahbe:
# Description

Fixes #9071

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -13,3 +13,6 @@
 
 - [sdk/python] Fix spurious diffs causing an "update" on resources created by dynamic providers.
   [#9656](https://github.com/pulumi/pulumi/pull/9656)
+
+- [sdk/python] Do not depend on the children of remote component resources.
+  [#9665](https://github.com/pulumi/pulumi/pull/9665)
--- a/sdk/python/lib/pulumi/runtime/resource.py
+++ b/sdk/python/lib/pulumi/runtime/resource.py
@@ -799,11 +799,10 @@ async def _add_dependency(
 
     from .. import ComponentResource  # pylint: disable=import-outside-toplevel
 
-    if isinstance(res, ComponentResource):
+    if isinstance(res, ComponentResource) and not res._remote:
         for child in res._childResources:
             await _add_dependency(deps, child, from_resource)
-        if not res._remote:
-            return
+        return
 
     no_cycles = declare_dependency(from_resource, res) if from_resource else True
     if no_cycles:
--- a/sdk/python/lib/test/langhost/remote_component_dependencies/__init__.py
+++ b/sdk/python/lib/test/langhost/remote_component_dependencies/__init__.py
@@ -1,0 +1,13 @@
+# Copyright 2016-2021, Pulumi Corporation.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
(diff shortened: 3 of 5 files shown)


t0yv0:
I'm a bit confused here.

Can we first have a by-design spec of the dependency relation? @pgavlin we discussed late last year that you were going to provide a spec, time to return to this perhaps?

The spec I have in mind as "current" was using cycle breaking but did not have special rules for remote components. It's partially implemented (say .NET pending, probably Python pending) across SDKs.

pgavlin (approved):
I agree that we need a spec here. I think we're going to have to work together to come up with one. I'm pretty sure that we do eventually to port the cycle-breaking code that @t0yv0 added last year to the other SDKs. However, I think that what you have here works, and we should probably port it to the other SDKs as well.

If I had to give an _informal_ spec of the expected behavior as of today, I'd say that it's something like this. The dependencies of a resource `R` is the set of custom and remote component resources reachable through the union of `R`'s implicit and explicit dependencies (i.e. resources referred to by `R`'s inputs and those present in `R`'s `dependsOn` option, if any. A resource `A` is _reachable_ from another resource `B` if `B` is a non-remote component resource and `A` is one of its descendants.

I think that the changes in this PR agree with that informal statement.

t0yv0:
The PR makes sense with the extended comment, great! 

Thinking out loud. Unlike remote=False, SDK cannot observe remote=True children. SDK is trying though to assist the engine with ordering of submitted RegisterResource calls, in particular if R depends on RemoteC, then it's trying to make sure RegisterResource(R) does not hit the engine until all of RemoteC's children are done. It appears waiting on `URN` of RemoteC to resolve is what guarantees it. The engine does not finish resolving the RemoteC.URN until all the children are done. 

@Frassle this will throw a wrench into trying to fiddle with remote-ness of URNs - this subtlety here is important, we use URN resolution as completion as signaling in the cooperative ordering game.

@iwahbe, the cycle breaking is totally problematic but without it you can get SDK hangs. The cycle breaking is the less problematic solution in practice over the set of programs people are trying to write.

We need to work out something better eventually. The key challenge is that the set of children of a resource is not known to the SDK when it's remote, and not known to the SDK when it's local either because the set is expanded as the program continues executing. There's no solid way to synchronize and do "please wait until all children are known", but we can design something perhaps.

t0yv0 (approved):

Frassle:
> @Frassle this will throw a wrench into trying to fiddle with remote-ness of URNs - this subtlety here is important, we use URN resolution as completion as signaling in the cooperative ordering game.

Yeh I'm not planning on making URN get resolved by SDKs, I was just mulling over the idea of somehow typing it to show that it's an Output that will never be "unknown". Having thought more on that I don't think we actually need to do anything here.

> We need to work out something better eventually. The key challenge is that the set of children of a resource is not known to the SDK when it's remote, and not known to the SDK when it's local either because the set is expanded as the program continues executing. There's no solid way to synchronize and do "please wait until all children are known", but we can design something perhaps.

Yes. Seeing something very similar with https://github.com/pulumi/pulumi/issues/9457

Issue #9670, opened 2022-05-24
CLI crashed when renaming stack


mikefrancis:
### What happened?

Tried to rename stack a second time

### Steps to reproduce

```bash
pulumi stack rename foo
pulumi stack rename bar
```

### Expected Behavior

Stack is renamed, no crash

### Actual Behavior

Stack is renamed, however CLI crashed with the following message:

```bash
Pulumi Version:   v3.33.1
Go Version:       go1.18.2
Go Compiler:      gc
Architecture:     arm64
Operating System: darwin
Panic:            runtime error: slice bounds out of range [-1:]

goroutine 1 [running]:
runtime/debug.Stack()
	/opt/homebrew/Cellar/go/1.18.2/libexec/src/runtime/debug/stack.go:24 +0x68
main.panicHandler()
	/private/tmp/pulumi-20220519-95273-1x30i3a/pkg/cmd/pulumi/main.go:29 +0x38
panic({0x1041b9700, 0x14000257b60})
	/opt/homebrew/Cellar/go/1.18.2/libexec/src/runtime/panic.go:838 +0x204
github.com/pulumi/pulumi/pkg/v3/backend/filestate.(*localBackend).renameHistory(0x14000a87380, {0x14000186798, 0x16}, {0x16d7b78f4, 0x3})
	/private/tmp/pulumi-20220519-95273-1x30i3a/pkg/backend/filestate/state.go:419 +0x37c
github.com/pulumi/pulumi/pkg/v3/backend/filestate.(*localBackend).RenameStack(0x14000a87380, {0x1042e3d98, 0x1400013e008}, {0x1042ef548, 0x14000b65740}, {0x16d7b78f4, 0x3})
	/private/tmp/pulumi-20220519-95273-1x30i3a/pkg/backend/filestate/backend.go:434 +0x21c
github.com/pulumi/pulumi/pkg/v3/backend.RenameStack({0x1042e3d98, 0x1400013e008}, {0x1042ef548?, 0x14000b65740?}, {0x16d7b78f4, 0x3})
	/private/tmp/pulumi-20220519-95273-1x30i3a/pkg/backend/stack.go:73 +0x70
github.com/pulumi/pulumi/pkg/v3/backend/filestate.(*localStack).Rename(0x14000186798?, {0x1042e3d98?, 0x1400013e008?}, {0x16d7b78f4?, 0x0?})
[...]

PR #9673, opened 2022-05-24
Fix a panic in renameHistory when encountering unexpected files

Frassle:
# Description

Fixes https://github.com/pulumi/pulumi/issues/9670

## Checklist


iwahbe (approved):
LGTM

Issue #6214, opened 2021-01-28
registerStackTransformations not registering


liamawhite:
When I try to add a set of tags to all resources in the stack, nothing happens.

## Expected Behavior
Tagging all AWS specific items that can be tagged.

## Current Behavior
Nothing, it doesn't even appear to be called as I don't see the console.log appear in the output. It does work if I pass the same function as a resource level transformation.

## Steps to Reproduce

**Transformation function**: `isTaggable` matches [this example](https://github.com/joeduffy/aws-tags-example/tree/master/autotag-ts).

```
export function defaultTags(): pulumi.ResourceTransformation {
    return (args) => {
        const metadata = new pulumi.Config("tetrate").requireObject<metadata>("metadata")
        console.log(args.type)
        if (isTaggable(args.type)) {
            args.props["tags"] = { ...args.props["tags"], ...{
                "pulumi:project": pulumi.getProject(),
                "pulumi:stack": pulumi.getStack(),
                "tetrate:owner": metadata.owner,
                "tetrate:source": metadata.source,
            } }
            return { props: args.props, opts: args.opts }
        }
        return undefined
    }
}

interface metadata {
    owner: string
    source: string
}
```

**Project `index.ts`**: (`export class VPC extends awsx.ec2.Vpc`)
```
import * as pulumi from "@pulumi/pulumi"
import { VPC } from '@tetrate/aws'
import { PrivateSubnet, PublicSubnet, IsolatedSubnet } from '@tetrate/aws'
import { defaultTags } from '@tetrate/aws'

pulumi.runtime.registerStackTransformation(defaultTags())

const opsVpc = new VPC("cloudops", {
    cidrBlocks: ["10.10.0.0/16"],
    subnets: [
        new PrivateSubnet(22),
        new PublicSubnet(24),
        new IsolatedSubnet(24)
    ]
})

export const cloudops = opsVpc.output
```

## Context (Environment)
Typescript: 2.19.0

leezen:
I'm unable to reproduce this with the following:

**index.ts**
```
import * as pulumi from "@pulumi/pulumi";
import * as aws from "@pulumi/aws";
import * as awsx from "@pulumi/awsx";
import { defaultTags } from "./tagging";

pulumi.runtime.registerStackTransformation(defaultTags());

const vpc = new awsx.ec2.Vpc("test", {});
```

**tagging.ts**
```
import * as pulumi from "@pulumi/pulumi";

// isTaggable copied from the same link as you pointed to

[...]

PR #9680, opened 2022-05-25
Stop using Get/SetRootResource in nodejs

Frassle:
# Description

Just use globalThis to make the stack resource a global rather than a module variable. This ensures that even if we end up with multiple Pulumi modules they can all grab the stack object and read URN off it.

We'll keep calling SetRootResource so old modules SxS continue to pickup the stack URN, at some point (4.0 probably) we can clean all this up.

Note that we also need the stack resource to be global for stack transformations, the roundtripping of the URN via the engine didn't preserve those.

## Checklist

### DIFF
--- a/sdk/nodejs/runtime/resource.ts
+++ b/sdk/nodejs/runtime/resource.ts
@@ -49,8 +49,6 @@ import {
 import {
     excessiveDebugOutput,
     getMonitor,
-    getProject,
-    getRootResource,
     getStack,
     isDryRun,
     isLegacyApplyEnabled,
@@ -58,9 +56,9 @@ import {
     serialize,
     terminateRpcs,
 } from "./settings";
+import { getStackResource } from "./stack";
 
 const gstruct = require("google-protobuf/google/protobuf/struct_pb.js");
-const providerproto = require("../proto/provider_pb.js");
 const resproto = require("../proto/resource_pb.js");
 
 interface ResourceResolverOperation {
@@ -521,9 +519,8 @@ async function prepareResource(label: string, res: Resource, custom: boolean, re
 
         // Wait for the parent to complete.
         // If no parent was provided, parent to the root resource.
-        const parentURN = opts.parent
-            ? await opts.parent.urn.promise()
-            : await getRootResource();
+        const parent = opts.parent ?? getStackResource();
+        const parentURN = parent ? await parent.urn.promise() : undefined;
 
         let providerRef: string | undefined;
         let importID: ID | undefined;
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -9,6 +9,9 @@
 - [sdk/nodejs] Lazy load inflight context to remove module import side-effect.
   [#9375](https://github.com/pulumi/pulumi/issues/9375)
 
+- [nodejs] No longer roundtrips requests for the stack URN via the engine.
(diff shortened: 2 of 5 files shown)


lukehoban:
> Note that we also need the stack resource to be global for stack transformations, the roundtripping of the URN via the engine didn't preserve those.

Does this fix https://github.com/pulumi/pulumi/issues/6214?

Frassle:
> Does this fix https://github.com/pulumi/pulumi/issues/6214?

Yup probably, going to try and write a test to demonstrate that but SxS are always a bit awkward to write.

sdk/nodejs/runtime/resource.ts
```diff
@@ -92,7 +91,7 @@ interface ResourceResolverOperation {
 /**
  * Get an existing resource's state from the engine.
  */
-export function getResource(res: Resource, props: Inputs, custom: boolean, urn: string): void {
+export function getResource(res: Resource, parent: Resource | undefined, props: Inputs, custom: boolean, urn: string): void {
```
t0yv0, line 94: We believe nothing outside the code already updated in this PR would call this func? So we don't care if this is breaking?

sdk/nodejs/runtime/resource.ts
```diff
@@ -519,9 +518,7 @@ async function prepareResource(label: string, res: Resource, custom: boolean, re
 
         // Wait for the parent to complete.
         // If no parent was provided, parent to the root resource.
-        const parentURN = opts.parent
-            ? await opts.parent.urn.promise()
-            : await getRootResource();
+        const parentURN = parent ? await parent.urn.promise() : undefined;
```
t0yv0, line 521: This `undefined` here. Tracing the code a bit, it seems it will just go straight to gRPC so things like ReadResourceRequest will have parent empty. The engine then assumes it's the `RootResource` so we're all good with this?

t0yv0 (approved):
Seems like a good change, LGTM. Though underlying cooperation of SetRootResource, GetRootResource and parent tracking is a bit convouted, you mention you want to drive this eventually to a "cleanup" but it's not obvious to me, spelling it out would help. What's the responsibility of the SDK vs the ResourceMonitor.

sdk/nodejs/runtime/resource.ts
Frassle: Yeh we don't expect things outside the SDK to call into runtime.

sdk/nodejs/runtime/resource.ts
Frassle: Yup, this should be fine. getRootResource should always be able to return the stack resource now so this shouldn't be undefined, but if it somehow is the engine will fill it in anyway.

Frassle:
> Though underlying cooperation of SetRootResource, GetRootResource and parent tracking is a bit convouted, you mention you want to drive this eventually to a "cleanup" but it's not obvious to me, spelling it out would help. What's the responsibility of the SDK vs the ResourceMonitor.

SetRootResource/GetRootResource will die eventually. We need to keep them for back compatibility with old SDKs for now. 

I'm not totally sure what we want the split of SDK vs engine responsibility to be here. On one hand we could push to eventually make it an error to send a resource without a parent, everything should at least be parented to the root stack. On the other, the engine can find and fill in the root stack, so we don't need the SDK to track that. I'm leaning towards the latter for now, but we might change direction.

t0yv0:
The later SGTM, more work on engine less on SDK is better.

PR #9685, opened 2022-05-25
Use goreleaser prebuilt

### LINKED ISSUE #8885: Investigate Gorelaser PRO prebuilt binary support
## Hello!

## Issue details

According to https://goreleaser.com/customization/build/ there is a prebuilt builder option:

```
    # Builder allows you to use a different build implementation.
    # This is a GoReleaser Pro feature.
    # Valid options are: `go` and `prebuilt`.
    # Defaults to `go`.
    builder: prebuilt
```

In our CI pipeline binaries are built by goreleaser in build.yml step, then rebuilt for release in release.yml step (in a different container). To make goreleaser use prebuilt binaries we trick it by wrapping the Go toolchain (go-wrapper.sh and unpack.sh scripts) which may become error-prone in the future. The task is to investigate if the same result can be accomplished cleanly with the `prebuilt` builder.

t0yv0:
# Description

**Before merging**: make sure GORELEASER_KEY secret is added to repo secrets.

Refactors and simplifies goreleaser setup using the PRO prebuilt feature.

- removed dep on goreleaser-filter, manually checked in variations of goreleaser config that were previously invisible
- when releasing, use prebuilt always; removes the need for checksum checks
- use ubuntu runners wherever possible for cross-building; only use macos builders for darwin builds (required)
- go-wrapper is retained for now to keep supporting quirky coverage-enabled builds; but prebuilt support is removed
- unpack.sh is removed, instead goreleaser config teaches where to find prebuilt binaries; less indirection
- builds are a little faster primarily through more precise cache keys
- do NOT call `go mod tidy` when building/releasing - as that can modify go.sum; would prefer to have go.sum modifications checked in through commits/PRs

Fixes #8885 

## Checklist


.goreleaser.build.darwin.yml
```diff
@@ -0,0 +1,83 @@
...
+
+release:
+  disable: true
+
+changelog:
+  skip: true
+
+builds:
+# Darwin builds
+- id: pulumi-unix
```
AaronFriel, line 13: Should this be -unix on the .darwin.yml?

AaronFriel (commented):
I do wonder if we have gotten ourselves into some accidental complexity here with having the 3 separate builders, which seems to result in the 3 separate goreleaser.build.$os.yml configs.

Would moving to a single build job for all combinations simplify the matrix and # of files to maintain? There's so much duplication, it seems like.

t0yv0:
> I do wonder if we have gotten ourselves into some accidental complexity here with having the 3 separate builders, which seems to result in the 3 separate goreleaser.build.$os.yml configs.
> 
> Would moving to a single build job for all combinations simplify the matrix and # of files to maintain? There's so much duplication, it seems like.

Putting the builds on the same container would down the builds and put more pressure on mac workers (all the builds then have to happen on macos). In by book that's not worth it. 

There might be other solutions that squash the 3 `.goreleaser.build.*.yml` files into one.

AaronFriel:
I'm wondering if that's true that it slows down the build, but I don't know what is cacheable in the go module cache - can `go build` reuse intermediate outputs from builds targeting another OS?

t0yv0:
It's quite true that building in parallel is better even given cache hits, but of course test.yml wait times dominate.

AaronFriel:
That's true, I was more interested in trying to reduce the complexity of maintaining these goreleaser configs if possible. But that's a non-blocking nit, we can pay that tax next time we need to make a change, when that change is valuable enough to justify it.

t0yv0:
I found something that reduces it slightly, will append here once I get past the nasty problem with tests not finding tall the scripts.

AaronFriel (approved):

.github/workflows/run-build-and-acceptance-tests.yml
```diff
@@ -174,9 +174,9 @@ jobs:
   build:
     name: Build
     uses: ./.github/workflows/build.yml
-    needs: versions
```
t0yv0, line 177: Bad merge here.

t0yv0:
@stack72 could you confirm GORELEASER_KEY is set to repo secrets? I do not have admin access to set this up. Then I can merge. TYVM.

PR #9686, opened 2022-05-25
[schema] Faster schema loading via caching alongside plugins

AaronFriel:
Implements caching of schemas in package loader. Package loader searches for plugins & before calling GetSchema, attempts to read a `schema-${pkg}.json` file. On cache miss or when the plugin is newer than the cache, the file is updated via GetSchema call.

Tested against locally installed packages as well as plugin cache located packages, e.g.: when `pulumi-resource-azure-native` is in $GOBIN, this caches the schema to `$GOBIN/schema-azure-native.json`.

A micro benchmark shows substantial improvements in wall time and bytes of memory allocated/copied, and a combined benchmark of this PR along with PRs to Azure Native & YAML reduced the time for a `pulumi preview` on an Azure Native stack down to 2.5 seconds from 12, and down to 340MiB peak memory usage from ~2GiB.

A microbenchmark shows the reduction in `loader_test.go`:

```
enchmarkLoadPackageReference/no-cache-8      1   2231342910 ns/op  351691352 B/op   7063 allocs/op
BenchmarkLoadPackageReference/file-cache-8   4    281257858 ns/op  631195948 B/op   2998 allocs/op
BenchmarkLoadPackageReference/mmap-cache-8   7    157173553 ns/op   15967628 B/op   2934 allocs/op
```
no cache: starting a plugin process and calling GetSchema() on it in every iteration. the memory usage is misleading as a result - it can't count the subprocess.

file cache: using a schema.json file colocated with the plugin. (this benchmark is pre-warmed, but it would be created if this cache misses)

mmap cache: instead of using io.ReadAll (copying) bytes, use mmap with O_RDONLY

Related PRs:
- https://github.com/pulumi/pulumi/pull/9620
- https://github.com/pulumi/pulumi-azure-native/pull/1689/
- https://github.com/pulumi/pulumi-yaml/pull/230

### DIFF
--- a/pkg/codegen/schema/loader.go
+++ b/pkg/codegen/schema/loader.go
@@ -46,9 +46,9 @@ type pluginLoader struct {
 type pluginLoaderCacheOptions struct {
 	// useEntriesCache enables in-memory re-use of packages
 	disableEntryCache bool
-	// useFileCache enables skipping plugin loading when possible and caching JSON schemas to files.
+	// useFileCache enables skipping plugin loading when possible and caching JSON schemas to files
 	disableFileCache bool
-	// useMmap enables the use of memory mapped IO to avoid copying the JSON schema when using
+	// useMmap enables the use of memory mapped IO to avoid copying the JSON schema
 	disableMmap bool
 }
 
--- a/pkg/codegen/schema/loader.go
+++ b/pkg/codegen/schema/loader.go
@@ -12,7 +12,6 @@ import (
 	"time"
 
 	"github.com/edsrzf/mmap-go"
-	"github.com/hashicorp/go-multierror"
 	"github.com/natefinch/atomic"
 
 	"github.com/blang/semver"
@@ -30,7 +29,6 @@ type Loader interface {
 
 type ReferenceLoader interface {
 	Loader
-	io.Closer
 
 	LoadPackageReference(pkg string, version *semver.Version) (PackageReference, error)
 }
@@ -40,8 +38,6 @@ type pluginLoader struct {
 
 	host    plugin.Host
 	entries map[string]PackageReference
-	files   []*os.File
-	mmaps   []mmap.MMap
 
 	cacheOptions pluginLoaderCacheOptions
 }
@@ -177,27 +173,6 @@ func (l *pluginLoader) ensurePlugin(pkg string, version *semver.Version) error {
 	return nil
 }
 
-func (l *pluginLoader) Close() error {
-	var result error
-	for _, m := range l.mmaps {
-		if err := m.Unmap(); err != nil {
-			result = multierror.Append(result, err)
-		}
-	}
-
-	for _, f := range l.files {
-		if err := f.Close(); err != nil {
-			result = multierror.Append(result, err)
-		}
-	}
-
-	if err := l.host.Close(); err != nil {
-		result = multierror.Append(result, err)
-	}
-
-	return result
-}
-
 func (l *pluginLoader) LoadPackage(pkg string, version *semver.Version) (*Package, error) {
 	ref, err := l.LoadPackageReference(pkg, version)
 	if err != nil {
@@ -328,17 +303,21 @@ func (l *pluginLoader) loadPluginSchemaBytes(pkg string, version *semver.Version
 	return schemaBytes, provider, nil
 }
 
+var mmapedFiles = make(map[string]mmap.MMap)
+
 func (l *pluginLoader) loadCachedSchemaBytes(pkg string, path string, schemaTime time.Time) ([]byte, bool) {
 	if l.cacheOptions.disableFileCache {
 		return nil, false
 	}
 
+	if schemaMmap, ok := mmapedFiles[path]; ok {
+		return schemaMmap, true
+	}
+
 	success := false
 	schemaFile, err := os.OpenFile(path, os.O_RDONLY, 0644)
 	defer func() {
-		if success {
-			l.files = append(l.files, schemaFile)
-		} else {
+		if !success {
 			schemaFile.Close()
 		}
 	}()
@@ -370,7 +349,6 @@ func (l *pluginLoader) loadCachedSchemaBytes(pkg string, path string, schemaTime
 		return nil, success
 	}
 	success = true
-	l.mmaps = append(l.mmaps, schemaMmap)
 
 	return schemaMmap, success
 }
--- a/pkg/codegen/schema/loader.go
+++ b/pkg/codegen/schema/loader.go
@@ -1,6 +1,7 @@
 package schema
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io"
@@ -9,6 +10,10 @@ import (
 	"regexp"
 	"sync"
 	"time"
+
+	"github.com/edsrzf/mmap-go"
+	"github.com/hashicorp/go-multierror"
+	"github.com/natefinch/atomic"
 
 	"github.com/blang/semver"
 	"github.com/segmentio/encoding/json"
@@ -25,6 +30,7 @@ type Loader interface {
 
 type ReferenceLoader interface {
 	Loader
+	io.Closer
 
 	LoadPackageReference(pkg string, version *semver.Version) (PackageReference, error)
 }
@@ -34,6 +40,20 @@ type pluginLoader struct {
 
 	host    plugin.Host
 	entries map[string]PackageReference
+	files   []*os.File
+	mmaps   []mmap.MMap
+
+	cacheOptions pluginLoaderCacheOptions
+}
+
+// Caching options intended for benchmarking or debugging:
+type pluginLoaderCacheOptions struct {
+	// useEntriesCache enables in-memory re-use of packages
+	disableEntryCache bool
+	// useFileCache enables skipping plugin loading when possible and caching JSON schemas to files.
+	disableFileCache bool
+	// useMmap enables the use of memory mapped IO to avoid copying the JSON schema when using
+	disableMmap bool
 }
 
 func NewPluginLoader(host plugin.Host) ReferenceLoader {
@@ -43,12 +63,35 @@ func NewPluginLoader(host plugin.Host) ReferenceLoader {
 	}
 }
 
+func newPluginLoaderWithOptions(host plugin.Host, cacheOptions pluginLoaderCacheOptions) ReferenceLoader {
+	return &pluginLoader{
+		host:    host,
+		entries: map[string]PackageReference{},
+
+		cacheOptions: cacheOptions,
+	}
+}
+
 func (l *pluginLoader) getPackage(key string) (PackageReference, bool) {
-	l.m.RLock()
-	defer l.m.RUnlock()
+	if l.cacheOptions.disableEntryCache {
+		return nil, false
+	}
 
 	p, ok := l.entries[key]
 	return p, ok
+}
+
+func (l *pluginLoader) setPackage(key string, p PackageReference) PackageReference {
+	if l.cacheOptions.disableEntryCache {
+		return p
+	}
+
+	if p, ok := l.entries[key]; ok {
+		return p
+	}
+
+	l.entries[key] = p
+	return p
 }
 
 // ensurePlugin downloads and installs the specified plugin if it does not already exist.
@@ -134,6 +177,27 @@ func (l *pluginLoader) ensurePlugin(pkg string, version *semver.Version) error {
 	return nil
 }
 
+func (l *pluginLoader) Close() error {
+	var result error
+	for _, m := range l.mmaps {
+		if err := m.Unmap(); err != nil {
+			result = multierror.Append(result, err)
+		}
+	}
+
+	for _, f := range l.files {
+		if err := f.Close(); err != nil {
+			result = multierror.Append(result, err)
+		}
+	}
+
+	if err := l.host.Close(); err != nil {
+		result = multierror.Append(result, err)
+	}
+
+	return result
+}
+
 func (l *pluginLoader) LoadPackage(pkg string, version *semver.Version) (*Package, error) {
 	ref, err := l.LoadPackageReference(pkg, version)
 	if err != nil {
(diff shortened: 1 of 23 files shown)


AaronFriel:
Ah, it looks like I need to update depoytest and/or this loader to make the caching compatible with what deploytest does.

AaronFriel:
@iwahbe I ran into a really strange issue with fixing the binder on this one - we have some unit tests that reference `google-native` types via $ref. Not just that, but there's a few different versions of the plugin in the json pointers. Somehow these tests pass on a clean system without these changes, but fail after. Installing the google native resource plugin causes tests to pass, but I also don't see where that could be occurring.

The strange part is that I can't find where in our code we're getting the google native schema _or_ where there's a branch that's allowing the $ref to fail _or_ where the plugin is installed.

iwahbe:
> @iwahbe I ran into a really strange issue with fixing the binder on this one - we have some unit tests that reference `google-native` types via $ref. Not just that, but there's a few different versions of the plugin in the json pointers. Somehow these tests pass on a clean system without these changes, but fail after. Installing the google native resource plugin causes tests to pass, but I also don't see where that could be occurring.
> 
> The strange part is that I can't find where in our code we're getting the google native schema _or_ where there's a branch that's allowing the $ref to fail _or_ where the plugin is installed.

That sounds super weird. Is it possible that we are doing something with pointer equality: attempt to load several versions of the google-native, fail and assign everything to nil pointers. With changes results in failure because you can't load the package. That would be my best guess, but I would need to see where the errors are to be more helpful.

pgavlin (commented):
> A micro benchmark shows substantial improvements in wall time and bytes of memory allocated/copied, and a combined benchmark of this PR along with PRs to Azure Native & YAML reduced the time for a pulumi preview on an Azure Native stack down to 2.5 seconds from 12, and down to 340MiB peak memory usage from ~2GiB.

What do things look like end-to-end without the changes in this PR?

AaronFriel:
@pgavlin Without schema caching, but with your PR to YAML here https://github.com/pulumi/pulumi-yaml/pull/230

755 MiB peak resident memory
2.95s wall time

Re-running the experiment after checking this branch out but with an empty plugin cache (all I've done is `pulumi plugin install ...`):

734 MiB 
3.05s

And second run after the schema is cached:

342 MiB peak
2.13s

AaronFriel:
Timing and memory usage is very consistent, over 3 runs without this PR, I saw 755, 754, 758 MiB. Wall times are less consistent, 2.9-3.7s.

With this PR, over 3 runs: 340, 334, 342 MiB and a range of 2.170-2.191s.

pgavlin (commented):
Okay. There's one other thing I'd like to try in the engine--would you mind running the numbers with the changes in https://github.com/pulumi/pulumi/compare/pgavlin/get-schema-zero-copy ?

AaronFriel:
Building YAML & pulumi with that branch:

| RSS (MiB) | time (s) |
|----------:|---------:|
|       751 |    2.911 |
|       738 |    2.927 |
|       734 |    3.064 |

AaronFriel:
Here's process tree memory usage for the 774 MiB run:

```
🌊 gotta-watch-em-all: Reached a new high water mark of 774 MiB, 37 MiB greater than before!
🌊 process                                                                          private MiB    total MiB
🌊 pulumi (22359)                                                                         61MiB       774MiB
🌊   pulumi-language-yaml (22380)                                                        250MiB       704MiB
🌊     pulumi-resource-azure-native (22392)                                              424MiB       424MiB
🌊     pulumi-resource-docker (22417)                                                     28MiB        28MiB
🌊   pulumi-resource-azure-native (22428)                                                  9MiB         9MiB
```

The bulk of the usage is in the azure native provider - which is also built with a zero-copy GetSchema using the embedded string:
- https://github.com/pulumi/pulumi-azure-native/pull/1689/

pkg/codegen/schema/loader.go
```diff
@@ -32,6 +36,20 @@ type pluginLoader struct {
...
 	host    plugin.Host
 	entries map[string]PackageReference
+	files   []*os.File
+	mmaps   []mmap.MMap
+
+	cacheOptions PluginLoaderCacheOptions
+}
+
+// Caching options intended for benchmarking or debugging:
+type PluginLoaderCacheOptions struct {
```
pgavlin, line 46: How are these options set? All of the fields are unexported.

pkg/codegen/schema/loader.go
```diff
@@ -132,6 +179,26 @@ func (l *pluginLoader) ensurePlugin(pkg string, version *semver.Version) error {
 	return nil
 }
 
+func (l *pluginLoader) Close() error {
+	for _, m := range l.mmaps {
+		if err := m.Unmap(); err != nil {
+			return err
```
pgavlin, line 185: We might want to collect these errors rather than returning early s.t. we unmap/close as many files as possible.

pkg/codegen/schema/loader.go
```diff
@@ -210,3 +250,106 @@ func LoadPackageReference(loader Loader, pkg string, version *semver.Version) (P
...
+			// Nonfatal
+		}
+		version = info.Version
+	}
+
+	return schemaBytes, version, nil
+}
+
+func (l *pluginLoader) loadPluginSchemaBytes(pkg string, version *semver.Version) ([]byte, plugin.Provider, error) {
+	if err := l.ensurePlugin(pkg, version); err != nil {
```
pgavlin, line 294: Although this is pre-existing code, I admit that I'm a little surprised that it works for our tests. I suppose we must use alternative `Loader`s rather than alternative `plugin.Host`s, as there are certainly implementations of the latter that do not use out-of-process plugins (mainly used for by `lifecycletest` for testing the engine). I would have expected this to live in the host implementation.

pkg/codegen/schema/loader.go
```diff
@@ -210,3 +250,106 @@ func LoadPackageReference(loader Loader, pkg string, version *semver.Version) (P
 	}
 	return p.Reference(), nil
 }
+
+func (l *pluginLoader) loadSchemaBytes(pkg string, version *semver.Version) ([]byte, *semver.Version, error) {
+	pluginInfo, err := workspace.GetPluginInfo(workspace.ResourcePlugin, pkg, nil, false /* skipMetadata */)
```
pgavlin, line 255: I'm a bit surprised by this. Is this to avoid the overhead of loading the provider plugin? If we loaded the plugin, we could just ask it for its info. I'm interested primarily because I don't love letting the `workspace` concept leak into the package loader--I'd much rather leave that logic for the plugin host. If we loaded the provider plugin, I think we could ask it for its info directly via `GetPluginInfo`.

pkg/codegen/schema/loader.go
```diff
@@ -210,3 +250,106 @@ func LoadPackageReference(loader Loader, pkg string, version *semver.Version) (P
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	if version == nil {
+		version = pluginInfo.Version
+	}
+
+	cachedSchemaPath := filepath.Join(pluginInfo.Path, fmt.Sprintf("schema-%v.json", pkg))
```
pgavlin, line 264: ~I am pretty nervous about writing data out to an unverified location. Despite what I mentioned about workspaces earlier, I'd really rather that we only read/write from/to cached schemas that live in the workspace's plugin cache.~

I see that you call this out later--thanks!

pkg/codegen/schema/loader.go
```diff
@@ -210,3 +250,106 @@ func LoadPackageReference(loader Loader, pkg string, version *semver.Version) (P
...
+		return nil, false
+	}
+
+	stat, err := schemaFile.Stat()
+	if err != nil {
+		return nil, false
+	}
+	cachedAt := stat.ModTime()
+
+	if cachedAt.Before(pluginCreatedAt) {
```
pgavlin, line 333: I am a little nervous about using file times here, but I don't have a better option besides e.g. checksumming the provider.

pgavlin (commented):
I admit that I'm worried that this might be a bridge too far. It feels somewhat more reliable to me to have plugins distribute the schema file next to the binary (though that does have its own set of issues). Maybe it would be enough to have the plugin installer remove any cached schema during installation, but I'm not sure.

AaronFriel:
> It feels somewhat more reliable to me to have plugins distribute the schema file next to the binary (though that does have its own set of issues).

Let's chat about this internally, I agree that a change like this has ramifications outside the engine & providers need to be included in that.

> Maybe it would be enough to have the plugin installer remove any cached schema during installation, but I'm not sure.

We wouldn't need to, but with the above, we need to agree on a naming convention for the file:

```
--- /home/friel/.pulumi/plugins/resource-aws-v4.36.0 ---------
                         /..
  343.3 MiB [##########]  pulumi-resource-aws
   15.8 MiB [          ]  schema-aws.json                            
   44.0 KiB [          ]  CHANGELOG.md
   12.0 KiB [          ]  LICENSE
   12.0 KiB [          ]  README.md
```

pkg/codegen/schema/loader.go
AaronFriel: I'm inclined to agree. It was very handy though to be able to `make install` a provider and have this automatically work, but... I'm not sure we should.

pkg/codegen/schema/loader.go
AaronFriel: This was so that we could get the plugin version & directory, otherwise when the version is nil (quite common for YAML and our tests) and we lack a directory, we don't have a unique name for the cache file. It leads to some peculiar errors when using the EKS provider (uses aws v4) and the awsx provider (uses aws v5).

I've refactored this to a `ResolvePlugin` method on the `plugin/host.go Host` interface to support `deploytest`. 

This ensures we put the plugin in a versioned directory alongside the plugin executable without having to host it.

pkg/codegen/schema/loader.go
AaronFriel: It turns out our integration tests in this repo install 3 different aws plugins, 3 google native plugins, 2 random plugins, a kubernetes plugin, and an azure-native plugin.

Our TestGeneratePackage tests use schemas that refer to external packages. In `GeneratePackageFilesFromSchema` we call `schema.ImportSpec` which doesn't take a loader as an argument, so we rely on the default plugin loader. 

Moving this down here was why my tests were failing (my comment to @iwahbe.) because I was asking the host to resolve the plugin before we had ensured it. My more recent commits move this to right above where we call `l.host.ResolvePlugin` (formerly `workspace.GetPluginInfo`).

😭

pkg/codegen/schema/loader.go
AaronFriel: Good point.

pkg/codegen/schema/loader.go
AaronFriel: A benchmark test in the same package.

I'll make this type and the `NewPluginLoaderWithOptions` constructor private to the package.

AaronFriel:
I have a fixup commit for the lint failure, waiting for tests to complete to avoid tying up more GHA resources.

pkg/codegen/schema/loader.go
AaronFriel: I've changed this comparison to use an expiration "`SchemaTime`" provided by the host on PluginInfo. For `deploytest`, the time is zero, for file based plugins it's the modified-at timestamp.

This is reliable, as building a new binary or touching the script that runs it - component provider implemented in another language - invalidates the cache.

I found this a bit easier than doing a hash or checksum.

pkg/codegen/schema/loader.go
AaronFriel: I've pushed this logic up into the host, which _may_ compute a `SchemaPath` property and put it on the PluginInfo. The path follows the same convention here, not repeating the version.

I think the only remaining questions here are whether we should put the version in this filename, and if so, should we disallow this optimization when the plugin is found via $PATH to avoid cluttering up other directories with this file. I could go either way on this, for 99.99% of users this cache will improve performance. Is it worth it for local engineering? Probably not.

pkg/codegen/schema/loader.go
AaronFriel: I've pushed several of the implementation details of the filesystem into fields on PluginInfo returned from this method. Schema Loader is now agnostic to the filesystem details and lets the host determine where and how to cache.

pkg/codegen/schema/loader.go
AaronFriel: I've removed support for caching locally installed packages in this PR to limit scope.

AaronFriel:
@pgavlin this is ready for review now, I've pared down the caching to only work on plugins installed in the plugin dir (and deploytest) and pushed decisions about where and when to invalidate the cache out to the workspace / plugin host. Whatever the plugin host returns from `ResolvePlugin` is authoritative.

AaronFriel:
Rebased to address the conflict.

pkg/codegen/schema/loader.go
```diff
@@ -34,6 +38,18 @@ type pluginLoader struct {
...
+	cacheOptions pluginLoaderCacheOptions
+}
+
+// Caching options intended for benchmarking or debugging:
+type pluginLoaderCacheOptions struct {
+	// useEntriesCache enables in-memory re-use of packages
+	disableEntryCache bool
+	// useFileCache enables skipping plugin loading when possible and caching JSON schemas to files.
+	disableFileCache bool
+	// useMmap enables the use of memory mapped IO to avoid copying the JSON schema when using
```
iwahbe, line 51: I think this comment is missing the end of the sentence.

pkg/codegen/schema/loader.go
```diff
@@ -228,3 +242,113 @@ func LoadPackageReference(loader Loader, pkg string, version *semver.Version) (P
...
+		err = atomic.WriteFile(pluginInfo.SchemaPath, bytes.NewReader(schemaBytes))
+		if err != nil {
+			return nil, nil, fmt.Errorf("Error writing schema from plugin to cache: %w", err)
+		}
+	}
+
+	if version == nil {
+		info, err := provider.GetPluginInfo()
+		if err != nil {
+			// Nonfatal
```
iwahbe, line 282: I think this error check is wrong. Do you mean
```go
if err == nil {
    version = info.Version
}
```

Right now you assign from `info.Version` even though it could be nil. See [`providerServer.GetPluginInfo`](https://github.com/pulumi/pulumi/blob/f3e430a1da2b665f29d711b9f1bfd9b9c7f240cc/sdk/go/common/resource/plugin/provider_server.go#L138-L144).

FYI: We have `contract.IgnoreError` if we just want to swallow the error.

iwahbe (approved):
I have a couple of nits, but this looks good in general. LGTM

pkg/codegen/schema/loader.go
AaronFriel: Yep, that's what I meant :)

PR #9688, opened 2022-05-25
Pre-compute versions as an artifact

### LINKED ISSUE #6634: Calculate build version once and reuse it
A potential hole was found in our builds due to https://github.com/pulumi/docs/issues/5507

Wr effectively found that when we tag a release, if an existing build is running, the publish step of that release will pull all the tags from the repository and then use those tags to calculate the version

therefore, we could be running tests of the master build as part of an "alpha" release and if we pushed a release tag (e.g. v2.21.2), the master build could use that v2.21.2 tag and then publish an artifact of that name. That would have a specific checksum. When the release build kicked in, it would recreate a new version of that binary and publish that + those checksums to S3 as well as GitHub. That could put us in a situation where we are serving incorrect binaries

**It looks like this only happened once - v2.23.1**

We should change the build process to calculate the release versions (for each language + Go binaries) at the start of the process and then re-use those versions throughout the build process

t0yv0:
# Description

Fixes #6634 - make CI version robust to new tags being pushed in the middle of a CI run.

## Checklist


AaronFriel:
@t0yv0 I don't see how `pulumictl` would return different results if a new tag were pushed - is it because pulumictl looks at the git repo and computes the version string based off the most recent tag?

t0yv0:
Yes, correct. It's been happening in practice.

t0yv0:
This is not an elegant fix but this is what we do in pulumi-java, does seem to work for this problem.

AaronFriel (approved):
I'm approving & I think the ultimate solution here is to adopt merge queues (aka merge trains in GitLab) once they are generally available:

https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue

Instead of running jobs on pushes to master and reacting to failures, we should be enforcing an always-green PR pipeline such that every commit merged to master is green.

Merge trains/queues/etc follow the philosophy of Uber's paper, Kepping master green at scale: https://eng.uber.com/research/keeping-master-green-at-scale/

stack72:
the issue we face is that the actions/checkout doesn't checkout tags and therefore we need to follow up and ask for all tags. so we start the GHA with checking out the tags, if a build finds new tags half way through then the different steps means it will calculate based on that new tag

PR #9691, opened 2022-05-26
Fix unsafe html emitted for Java properties

viveklak:
# Description

This change fixes unsafe html emitted for Java properties for template types which results in mangling API docs for multiple resources, see https://www.pulumi.com/registry/packages/command/api-docs/local/command/ and https://www.pulumi.com/registry/packages/kubernetes/api-docs/helm/v3/chart/ for instance.

The general expectation based on [this template]( https://github.com/pulumi/pulumi/blob/master/pkg/codegen/docs/templates/constructor_args.tmpl) is that the type name is already safe (`htmlSafe` marks literal strings to be safe) but this was not true before the change for template types (`e.g. List<T>`), resulting in unescaped html. 

I can certainly make the escaping parameterized on the language but I suspect this should be safe across all languages. Happy to make this more targeted if necessary.

Fixes # (issue)
https://github.com/pulumi/pulumi-kubernetes/issues/1982
https://github.com/pulumi/registry/issues/949

## Checklist

### DIFF
--- a/pkg/codegen/docs/gen.go
+++ b/pkg/codegen/docs/gen.go
@@ -611,18 +611,20 @@ func (mod *modContext) typeString(t schema.Type, lang string, characteristics pr
 		displayName = mod.cleanTypeString(t, langTypeString, lang, modName, characteristics.input)
 	}
 
-	// If word-breaks need to be inserted, then the type string should be html-encoded first in order to avoid confusing
-	// the Hugo rendering where the word-break tags are inserted. This is required for languages with template types.
-	if insertWordBreaks {
-		displayName = html.EscapeString(displayName)
-		displayName = wbr(displayName)
-	}
-
 	displayName = cleanOptionalIdentifier(displayName, lang)
 	langTypeString = cleanOptionalIdentifier(langTypeString, lang)
 
+	// Name and DisplayName should be html-escaped to avoid throwing off rendering for template types in languages like
+	// csharp, Java etc. If word-breaks need to be inserted, then the type string should be html-escaped first.
+	if insertWordBreaks {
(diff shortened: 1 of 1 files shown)


pkg/codegen/docs/gen.go
```diff
@@ -611,14 +611,10 @@ func (mod *modContext) typeString(t schema.Type, lang string, characteristics pr
...
-	// should be html-encoded first if the language is C# in order
-	// to avoid confusing the Hugo rendering where the word-break
-	// tags are inserted.
+	// If word-breaks need to be inserted, then the type string should be html-encoded first in order to avoid confusing
+	// the Hugo rendering where the word-break tags are inserted. This is required for languages with template types.
 	if insertWordBreaks {
-		if lang == "csharp" {
-			displayName = html.EscapeString(displayName)
-		}
+		displayName = html.EscapeString(displayName)
```
dixler, line 617: should we just move this to the end of the function to ensure that the output is escaped and not only if there's a word break?

dixler: #9693 has an example

viveklak: Left a comment on #9693 - I don't think that is quite right. I did update this a bit though.

dixler: You're right. Could we move the escaping before we get into the if statement so that we can reduce duplication and adding an extra branch?

viveklak: doh ofcourse. Not enough ☕ .

dixler (approved):
LGTM. make sure to wait until @Frassle is done with the patch release.

PR #9693, opened 2022-05-26
moved escapes

dixler:
# Description
moved escapes lower

Fixes # (issue)

## Checklist


pkg/codegen/docs/gen.go
```diff
@@ -611,23 +611,18 @@ func (mod *modContext) typeString(t schema.Type, lang string, characteristics pr
...
 	}
 
 	displayName = cleanOptionalIdentifier(displayName, lang)
 	langTypeString = cleanOptionalIdentifier(langTypeString, lang)
 
 	return propertyType{
-		Name:        langTypeString,
-		DisplayName: displayName,
+		Name:        html.EscapeString(langTypeString),
+		DisplayName: html.EscapeString(displayName),
```
viveklak, line 625: Won't this escape the wbr tags inserted which you don't want?

PR #9700, opened 2022-05-26
Inline protobuf generation docker image

Frassle:
Inline the dockerfile we use to do protobuf code generation from https://github.com/pulumi/pulumi-build-container

This is the only place that image is used. There's little point in having a whole pipeline and separate repo for it.


stack72 (approved):

sdk/python/requirements.txt
```diff
@@ -1,6 +1,6 @@
 # Packages needed by the library.
 # Keep this list in sync with setup.py.
-protobuf>=3.6.0,<4
+protobuf>=4.21.0
```
jspiewak, line 3:
```suggestion
protobuf>=4.21.0,<5
```

sdk/proto/build-container/Dockerfile
```diff
@@ -0,0 +1,59 @@
+# syntax = docker/dockerfile:experimental
+FROM ubuntu:bionic
+
+# This image was copied from https://github.com/pulumi/pulumi-build-container
+# It can probably be simplifed as it is now purely used to generate the protobuf files
+
+LABEL description="Pulumi Build Container"
+LABEL maintainer="james@jen20.com"
```
AaronFriel, line 8: Update maintainer? :)

PR #9708, opened 2022-05-26
Set unspecified explicit provider version to default provider version

### LINKED ISSUE #9482: Default and Explicit Provider is Selecting Different Plugin Versions
### What happened?

In a go program, with multiple versions of the same Pulumi plugin installed, the default provider selects the appropriate version based on the SDK version, while the explicit provider selects the newest available version of the plugin. For example, if versions `4.38.1` and `5.1.3` of the `aws` plugin are installed and the SDK is pinned to `github.com/pulumi/pulumi-aws/sdk/v4/go/aws` in the go program, the default provider will correctly select the `4.38.1` version while the explicit provider will incorrectly select the `5.1.3` version.

Not only is this incorrect, but this is problematic as there are some breaking changes seen in `5.1.3`. Plugin versions should match the SDK's version unless the user defines otherwise. Note, a workaround is to specify the version of the `plugin` for the provider in the provider's resource options, however, this is cumbersome.

### Steps to reproduce

1. Clone https://github.com/phillipedwards/aws-plugin-issue
2. Ensure AWS plugin `4.38.1` and `5.1.2 | 5.1.3` are installed. `pulumi plugin install resource aws ....`
3. Execute `pulumi up`
4. Pay close attention to the `eks.LookupCluster` calls. (they have different outcomes with different providers).

### Expected Behavior

Explicit and default providers, in the same program, should select the correct plugins as dictated by the language SDK. In this case, that should be `pulumi aws 4.38.1` as aws is pinned to `github.com/pulumi/pulumi-aws/sdk/v4/go/aws` in the Pulumi program.

### Actual Behavior

Explicit provider disregards the pinned version in the program and selects the latest available version of the aws plugin, which is `5.1.3` in this case. The default provider correctly selects version `4.38.1`. 

NOTE: if only version `4.38.1` is installed, the explicit provider will select the correct plugin version.

### Versions used

CLI: 3.30.0
AWS: 4.38.1, 5.1.3
OS: Ubuntu - 20.04

### Additional context

Viewing log data, `GetPluginPath` is initially called with a semver included (`4.38.1`) which correctly selects the plugin version. Later in the execution, `GetPluginPath()` is called _without_ a semver included which defaults plugin selection to "the latest version". This data can be seen executing with `pulumi up|pre --logtostderr -v=9 2> logs.txt`.

### Contributing

iwahbe:
# Description

Fixes #9482 

## Checklist

### DIFF
--- a/pkg/engine/plugins.go
+++ b/pkg/engine/plugins.go
@@ -56,6 +56,33 @@ func (p pluginSet) Union(other pluginSet) pluginSet {
 	return newSet
 }
 
+// Removes less specific entries.
+//
+// For example, the plugin aws would be removed if there was an already existing plugin
+// aws-5.4.0.
+func (p pluginSet) Deduplicate() pluginSet {
+	existing := map[string]workspace.PluginInfo{}
+	newSet := newPluginSet()
+	add := func(p workspace.PluginInfo) {
+		prev, ok := existing[p.Name]
+		if ok && p.Version == nil && p.PluginDownloadURL == "" {
+			// no new information
+			return
+		}
+
+		if prev.Version == nil && prev.PluginDownloadURL == "" {
+			// New plugin is more specific then the old one
+			delete(newSet, prev.String())
+		}
+		newSet.Add(p)
+		existing[p.Name] = p
+	}
+	for _, value := range p {
+		add(value)
+	}
+	return newSet
+}
+
 // Values returns a slice of all of the plugins contained within this set.
 func (p pluginSet) Values() []workspace.PluginInfo {
 	var plugins []workspace.PluginInfo
--- a/pkg/engine/update.go
+++ b/pkg/engine/update.go
@@ -237,7 +237,7 @@ func installPlugins(
 		return nil, nil, err
 	}
 
-	allPlugins := languagePlugins.Union(snapshotPlugins)
+	allPlugins := languagePlugins.Union(snapshotPlugins).Deduplicate()
 
 	// If there are any plugins that are not available, we can attempt to install them here.
 	//
--- a/pkg/resource/deploy/source_eval.go
+++ b/pkg/resource/deploy/source_eval.go
@@ -1010,6 +1010,8 @@ func (rm *resmon) RegisterResource(ctx context.Context,
 				return nil, fmt.Errorf("%s: passed invalid version: %w", label, err)
 			}
 			providers.SetProviderVersion(props, &version)
+		} else if prov, ok := rm.defaultProviders.defaultProviderInfo[providers.GetProviderPackage(t)]; ok && prov.Version != nil {
+			providers.SetProviderVersion(props, prov.Version)
 		}
 		if req.GetPluginDownloadURL() != "" {
 			providers.SetProviderURL(props, req.GetPluginDownloadURL())


pkg/engine/plugins.go
```diff
@@ -56,6 +56,33 @@ func (p pluginSet) Union(other pluginSet) pluginSet {
...
+	existing := map[string]workspace.PluginInfo{}
+	newSet := newPluginSet()
+	add := func(p workspace.PluginInfo) {
+		prev, ok := existing[p.Name]
+		if ok && p.Version == nil && p.PluginDownloadURL == "" {
+			// no new information
+			return
+		}
+
+		if prev.Version == nil && prev.PluginDownloadURL == "" {
```
Frassle, line 73: Little confusing that this line works because prev will be a zero value struct when ok is false. Maybe both this if conditions should be wrapped in an "if ok {" block to be clear that we're only doing comparisons if there's something to compare to.

pkg/engine/plugins.go
```diff
@@ -56,6 +56,33 @@ func (p pluginSet) Union(other pluginSet) pluginSet {
...
+			// New plugin is more specific then the old one
```
Frassle, line 74: What if the old just had version set and the new just has downloadURL set? What if we get two with version set but to different versions (5.0 and 6.0 for example). It's not clear there's a correct total ordering here?

pkg/engine/update.go
```diff
@@ -237,7 +237,7 @@ func installPlugins(
 		return nil, nil, err
 	}
 
-	allPlugins := languagePlugins.Union(snapshotPlugins)
+	allPlugins := languagePlugins.Union(snapshotPlugins).Deduplicate()
```
Frassle, line 240: computeDefaultProviderPlugins sorts and selects the default provider by greatest version anyway, why do we need to deduplicate?

pkg/resource/deploy/source_eval.go
```diff
@@ -1010,6 +1010,8 @@ func (rm *resmon) RegisterResource(ctx context.Context,
 				return nil, fmt.Errorf("%s: passed invalid version: %w", label, err)
 			}
 			providers.SetProviderVersion(props, &version)
+		} else if prov, ok := rm.defaultProviders.defaultProviderInfo[providers.GetProviderPackage(t)]; ok && prov.Version != nil {
```
Frassle, line 1013: Can you have downloadURL set but version not set? If somehow the user tries to use a plugin without a version but with a download url won't this line then cause that plugin to not be found?

pkg/engine/plugins.go
iwahbe: If either `pluginDownloadURL`, `Version` or both are set we consider the plugin fully specified and keep it. It is ok to keep `pkg v1.2.3` `pkg v2.3.4` and `pkg example.com` in a single set. What we don't want to do is keep `pkg` in that same set since there are more specific versions used. In general, there will be a `pky vx.y.y` in the plugin set because the user depended on a language package `pulumi-pkg` with version `x.y.z`.

pkg/engine/update.go
iwahbe: We need to deduplicate to avoid installing plugins which will never be used. I could move the `Deduplicate` call to only effect the set passed into `ensurePluginsAreInstalled` to limit the impact of the change. You are correct that it won't matter elsewhere.

pgavlin (commented):
Left some additional notes in the linked issue.

pgavlin (commented):
The thing that worries me about these changes is that I’m not entirely sure how the “default” default provider version is computed. The statement I’d like to be able to make is “if a resource registration does not request a specific version, it will either use the latest version of the plugin or the version indicated by GetRequiredPlugins. If the “default” default provider version matches those semantics, then I think that we can move forward.

iwahbe:
> The thing that worries me about these changes is that I’m not entirely sure how the “default” default provider version is computed. The statement I’d like to be able to make is “if a resource registration does not request a specific version, it will either use the latest version of the plugin or the version indicated by GetRequiredPlugins. If the “default” default provider version matches those semantics, then I think that we can move forward.

The new scheme mostly complies with ☝️, because default plugins comply with ☝️. We only source all changes introduced in this PR from the default provider information. Here is a more detailed explanation:

The new defaults are provisioned from the `defaultProviderVersions` that are handed to new `deploy.EvalSource`s [here](https://github.com/pulumi/pulumi/blob/34e30d83d0eb2f0670520fb6d15c028cbc0a6bb5/pkg/engine/update.go#L418-L424). This is computed by taking the most up to date version plugin version sent by `GetRequiredPlugins` (or the state if `GetRequiredPlugins` is empty). This is a link to the in-code description:
https://github.com/pulumi/pulumi/blob/34e30d83d0eb2f0670520fb6d15c028cbc0a6bb5/pkg/engine/plugins.go#L251-L268

Example of violating the above rules:

Consider the following go program using the aws 1.0.0 SDK:
```go
aws.NewProvider("explicit", &emptyArgs)
s3.NewBucket("unparented")
```
 
If the program is launched on a computer with only `aws 1.0.0` installed, that would be recorded in the state file, and both `explicit` and `unparented` would be created with the `aws 1.0.0` plugin. If the same resources were created with the program factored out to a component resource in another module and pulumi up was run again with `aws 1.0.0` and `aws 1.1.0` installed, this PR would have both `explicit` and `unparented` created with `aws 1.0.0`, contrary to the above requirement. 

Without this PR, the second `pulumi up` would create `explicit` with plugin `aws 1.1.0` but `unparented` with plugin `aws 1.0.0` (which I think is more confusing).

pgavlin (commented):
I think that as long as the statement is "if a register resource request is missing a version (whether or not the request is for a provider type), then the engine will select the default package version (i.e. the "default" default provider version)", then I'm pretty comfortable. As long as that is consistent, then this becomes a matter of documenting how we compute the default package version (sounds like "GetRequiredPlugins if possible, otherwise from the state").

I think we need additional tests for version selection here before we can consider merging. I'd love to see new tests in `pkg/engine/lifecycletest` that cover the various scenarios. It might even be best to add those tests in a separate PR prior to merging this to help us clarify the existing behavior and the changes we'll see with this PR.

pgavlin (approved):
Okay, this looks good to me. One last request: before merging, can you update the PR description with a description of the problem, the current behavior, and the new behavior, and then make sure that the PR description is taken for the squashed changes? I think that will make it easier to understand these changes when going through the history.

PR #9715, opened 2022-05-27
add additional CI provider detection

EvanBoyle:
Detect an additional CI provider, relying on the default implementation of [DetectVars](https://github.com/pulumi/pulumi/blob/29eb7bd3300f974e42791ad049a7c821ecf69c8b/sdk/go/common/util/ciutil/systems.go#L109) that just populates the provider name.


pgavlin (approved):

pgavlin (commented):
LGTM. Python failures are unrelated.

AaronFriel:
Would it make sense to add a `CI` env var fallback? 

It's not unknown to do so in other tools: https://github.com/percy/cli/blob/HEAD/packages/env/src/environment.js#L46-L47

PR #9721, opened 2022-05-27
Expose SDK generation from `pulumi`

iwahbe:
# Description

This is an updated and expanded version of https://github.com/pulumi/pulumi/pull/7876. 

With regards to versioning:
We build codegen as part of pulumi/pulumi, so our codegen would be part of the same versioning scheme. This should be ok for the vast majority of users. If users are concerned about versioning SDKgen (codegen) separately from the CLI, they are able to write their own binary, pinning specific versions of pulumi/pulumi and pulumi/pulumi-java as they feel necessary. While I think that this system can be used by power users, adding `pulumi schema generate-sdk` does not restrict more invested users from building out SDKgen (codegen) as their own application.

## Checklist

### DIFF
--- a/pkg/cmd/pulumi/import.go
+++ b/pkg/cmd/pulumi/import.go
@@ -461,7 +461,7 @@ func newImportCmd() *cobra.Command {
 				programGenerator = gogen.GenerateProgram
 			case "nodejs":
 				programGenerator = nodejs.GenerateProgram
-			case "python":
+			case "python": //nolint:goconst
 				programGenerator = python.GenerateProgram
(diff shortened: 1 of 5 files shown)


stack72 (commented):
I think we still need to talk about this - I feel really strongly about trying to not make our CLI the kitchen sink. I am a fan of having separate tools for separate things. I'd like to try and understand the user experience around the separation of users / builders

iwahbe:
Closed in favor of https://github.com/pulumi/pulumi/pull/10711

PR #9731, opened 2022-05-30
Send smart aliases via gRPC to engine

Frassle:
# Description

This add an new field on the RegisterResource and ReadResource protobuf messages to the resource monitor. As well as the `repeated string aliases` field, we now have a `repeated Alias smartAliases` field (name open to scrutiny). The `Alias` object is a protobuf encoding to match the alias objects we have in the SDKs where users can just set "name", or "type". Those objects worked by the SDK then resolving down to a URN, that resolution logic is now also in the engine and so eventually we'll be able to remove all that logic from every SDK (we can't remove it straight away because of back compatibility, a user may have a new SDK but old engine).

Fixes https://github.com/pulumi/pulumi/issues/9734

## Checklist


AaronFriel:
Curious how big of a change in the engine would it be to rename the existing field (keeping tag ID the same) to `alias_urns` and naming the new field `aliases`?

I'd then add a `[deprecated = true]` to the old field, which is a no-op in most languages but might add an annotation or IDE hint.

Frassle:
> Curious how big of a change in the engine would it be to rename the existing field (keeping tag ID the same) to alias_urns and naming the new field aliases? I believe protobufs allow these sort of renames as the thing that matters are the type and tag numbers?
> 
> I'd then add a [deprecated = true] to the old field, which is a no-op in most languages but might add an annotation or IDE hint.

Done. Interestingly a discovery fell out of that rename that we didn't look at aliases for resource reads, and only the Go SDK tried to send them. I'm not sure if aliases make sense for reads anyway so I've just removed that field for now.

Would be good to get @pgavlin check on if alias for reads is/isn't needed.

pkg/resource/deploy/deploytest/resourcemonitor.go
```diff
@@ -104,6 +104,7 @@ type ResourceOptions struct {
 	Remote                  bool
 	Providers               map[string]string
 	AdditionalSecretOutputs []resource.PropertyKey
+	SmartAliases            []resource.Alias
```
AaronFriel, line 107: Able to rename this field as well?

AaronFriel (approved):
My only comment is that this makes me long for codegen-ing resource options. It's a lot of little changes to review in each SDK.

Looking forward to deleting "inheritedChildAlias" and all the other machinery in the SDKs in 4.0.

Frassle:
> My only comment is that this makes me long for codegen-ing resource options. It's a lot of little changes to review in each SDK.

Working on a design for that today 😄 

> Able to rename this field as well?

Yes will do!

AaronFriel:
LGTM

PR #9732, opened 2022-05-30
Fix `join` equivalent operation when nesting ApplyT

### LINKED ISSUE #9506: Go SDK codegen for resource-valued outputs declares the wrong output types or one that is not chain-able

AaronFriel:
An Output and its ApplyT method has behavior akin to NodeJS' Promise and then method, in that nested outputs are unwrapped. A previous PR, to fix #9506, lacked rigorous tests of the desired behavior of the unwrapping.

This adds those tests and solves some bugs caused by mutation in the `await` function. Instead, a fresh Output is created and we properly implement a Monadic "join" like behavior where we accumulate known-ness/errors, secret-ness, and dependencies. If not for known-ness, these would be a commutative monoid and order-independent. Two states however are non-commutative: when not known or the error field is set. This behaves similar to throwing an exception in NodeJS, we early exit and additional Then methods don't have their arguments called.

Without this PR, the added tests will fail as follows:

This line of `TestApplyTOutputJoinDeps`, due to mutation of `outB` adding `rA` to the dependencies:

https://github.com/pulumi/pulumi/blob/ef8af8843061be632b574744210aa2c253cede59/sdk/go/pulumi/types_test.go#L864

These lines of `TestApplyTOutputJoin`, due to mutation of `out2` adding `r1` to the dependencies:

https://github.com/pulumi/pulumi/blob/ef8af8843061be632b574744210aa2c253cede59/sdk/go/pulumi/types_test.go#L908-L910

This line of `TestApplyTOutputJoin`, due to `await` not accumulating secret-ness. 

https://github.com/pulumi/pulumi/blob/ef8af8843061be632b574744210aa2c253cede59/sdk/go/pulumi/types_test.go#L913

### DIFF
--- a/sdk/go/pulumi/types.go
+++ b/sdk/go/pulumi/types.go
@@ -201,13 +201,15 @@ func (o *OutputState) fulfillValue(value reflect.Value, known, secret bool, deps
 }
 
 func mergeDependencies(ours []Resource, theirs []Resource) []Resource {
-	mergedDeps := make([]Resource, 0, len(ours)+len(theirs))
-	if theirs == nil {
-		return append(mergedDeps, ours...)
-	} else if ours == nil {
-		return append(mergedDeps, theirs...)
+	if len(ours) == 0 && len(theirs) == 0 {
+		return nil
+	} else if len(theirs) == 0 {
+		return append(make([]Resource, 0, len(ours)), ours...)
+	} else if len(ours) == 0 {
+		return append(make([]Resource, 0, len(ours)), theirs...)
 	}
 	depSet := make(map[Resource]struct{})
+	mergedDeps := make([]Resource, 0, len(ours)+len(theirs))
 	for _, d := range ours {
 		depSet[d] = struct{}{}
 	}
--- a/sdk/go/pulumi/types.go
+++ b/sdk/go/pulumi/types.go
@@ -255,11 +255,11 @@ func (o *OutputState) await(ctx context.Context) (interface{}, bool, bool, []Res
 		o.mutex.Unlock()
 
 		deps = mergeDependencies(deps, o.deps)
-		if !o.known || o.err != nil {
-			return nil, false, false, deps, o.err
-		}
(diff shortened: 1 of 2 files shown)


AaronFriel:
@pgavlin Should we accumulate this metadata here? What convinced me is that otherwise there's some unsoundness in our dependency tracking if the inner output is constructed in this way:

```go
outputA.applyT(func (x int) IntOutput {
  innerOutput := outputB.applyT(func (y int) IntOutput {
    return x * y
  }).(IntOutput) 
  // this output would only have the dependencies, known-ness, secret-ness of B
  return innerOutput
}).(IntOutput)
```

PR #9735, opened 2022-05-31
Poll gRPC healthchecks to auto-cancel sub-processes

Frassle:
# Description

This hooks the rpcutils cancellation to a polling process that asks for the health of the engine.

If the engine dies, the health check fails, rpcutil gets cancelled, sub-process quits.
This is hooked into both the langauge plugins, and the shared provider main method.

N.B. Providers started in "attach" mode do not poll the health check as it is expected that these processes we're explictly started by the user and the user will shut them down.

Fixes https://github.com/pulumi/pulumi/issues/8696

## Checklist


sdk/go/common/util/rpcutil/health.go
```diff
@@ -0,0 +1,67 @@
...
+	healthpb "google.golang.org/grpc/health/grpc_health_v1"
+	"google.golang.org/grpc/status"
+)
+
+// Healthcheck will poll the address at duration intervals and then call cancel once it reports unhealthy
+func Healthcheck(context context.Context, addr string, duration time.Duration, cancel context.CancelFunc) error {
+	conn, err := grpc.Dial(
+		addr,
+		grpc.WithInsecure(),
+		grpc.WithUnaryInterceptor(OpenTracingClientInterceptor()),
```
t0yv0, line 32: Interesting. For most analysis I'd want to do on the trace, having these calls in the trace wouldn't be that helpful. But it seems reasonable to include it for completeness. On the fence about this.

t0yv0:
It looks like it would work but did you get a chance to try it out? Terminate the CLI process and observe the lang process self-terminating?

The approach looks pretty good but perhaps a partial solution since only installed in lang providers. 

- resource provider processes are definitely subject to the same problem; how do we roll out the change to those? 

- do we have other shims or processes, such as Python and Node processes for example, that need special treatment? or since those processes are purely gRPC clients and essentially do not host any servers, they would die naturally when the gRPC servers they talk to are down?

I'd merge this but log the remaining work in the backlog.

t0yv0:
Ah I see it's hooked into provider main. That's good. Do we believe all providers call that? Checking.

t0yv0:
Yes looks like a bit convoluted but bottoms out here for pulumi-aws:

https://github.com/pulumi/pulumi-terraform-bridge/blob/f9399c91f0fe22cb6e6754e8a63556e054e8e114/pkg/tfbridge/serve.go#L28

https://github.com/pulumi/pulumi-aws/blob/master/provider/cmd/pulumi-resource-aws/main.go#L26

This should work then 🤞

t0yv0:
We need to roll it out to Java and YAML.

t0yv0:
https://github.com/pulumi/pulumi-java/issues/640
https://github.com/pulumi/pulumi-yaml/issues/242

t0yv0:
I think once this propagates everywhere we can remove the SIGKILL sending code right?

Frassle:
> I think once this propagates everywhere we can remove the SIGKILL sending code right?

I mean it still means things shutdown sooner so I think we'd keep it in?

Frassle:
> It's dangerous though - it bypasses graceful shutdown and finalizers. I don't know how much we rely on those yet but we probably do to some extent? Make sure all logs and flushed out, etc.

Ah right yes, we should probably rework them to sigterm and catch that in the plugins to gracefully quit.

pkg/resource/provider/main.go
```diff
@@ -52,12 +55,27 @@ func Main(name string, provMaker func(*HostClient) (pulumirpc.ResourceProviderSe
...
+			cancelChannel = make(chan bool)
+			go func() {
+				<-ctx.Done()
+				close(cancelChannel)
+			}()
+			err = rpcutil.Healthcheck(ctx, args[0], 5*time.Minute, cancel)
+			if err != nil {
+				return fmt.Errorf("could not start health check host RPC server: %w", err)
+			}
+		}
```
AaronFriel, line 72: style nit: it doesn't follow for me here is that we check host != nil then don't use host in the consequent. If host == nil is a possibility (even with err == nil), is that an error we should handle separately?

Might be cleaner to do: 

```go
if host == nil { 
  /* error */ 
}
// host != nil case, unindented
```

sdk/go/common/util/rpcutil/health.go
AaronFriel: Better to trace than not to, IMO. Might help solve why a provider is failing if the trace shows something strange in healthcheck. Unlikely, but possible?

AaronFriel (approved):

pkg/resource/provider/main.go
Frassle: It's not an error for host to be nil. That's what we get if we're doing an provider attach.

Frassle: Although the host == nil check shouldn't actually be needed, we assume NewHostClient won't return nil if err was nil

Frassle:
> I've approved modulo the conversation on SIGTERM and graceful termination.

I'll pick that up as a separate PR.

PR #9747, opened 2022-06-01
Special case the error specifing app.pulumi.com

### LINKED ISSUE #9410: poor error message when using pulumi login app.pulumi.com
_orginally reported through community slack_

## problem
a new user in community slack was using `pulumi login https://app.pulumi.com/<individual account-name>` instead of `pulumi login`
which first requests a token, and then results in the error `error: problem logging in: getting user info from https://app.pulumi.com/susanev: unmarshalling response object: invalid character '<' looking for beginning of value`

this error is unhelpful, and does not give information to the user that will help them fix the problem.

currently, folks need to only use `pulumi login` when using the service as their backend.

## thoughts & questions
- would it be possible to assume they mean pulumi login in this case? or support people using app.pulumi.com in this case?
    - possibly identify if they are using app.pulumi.com instead of api.pulumi.com
- if not, can we respond with a better error message letting them know to use `pulumi login`?

## ref
[internal slack convo](https://pulumi.slack.com/archives/C9SGX9QA1/p1650038927550389)
[community slack convo](https://pulumi-community.slack.com/archives/C01PF3E1B8V/p1650038002997409)

iwahbe:
# Description

Fixes #9410

I think we should just respond with an error message here. If you type `https://api.pulumi.com` then everything works as expected. We shouldn't perform the automatic conversion of `app` to `api`. 

## Checklist

### DIFF
--- a/pkg/cmd/pulumi/login.go
+++ b/pkg/cmd/pulumi/login.go
@@ -121,9 +121,10 @@ func newLoginCmd() *cobra.Command {
 					return fmt.Errorf("could not determine current cloud: %w", err)
 				}
 			} else if url := strings.TrimPrefix(strings.TrimPrefix(
-				cloudURL, "https://"), "http://"); strings.HasPrefix(url, "app.pulumi.com/") {
+				cloudURL, "https://"), "http://"); strings.HasPrefix(url, "app.pulumi.com/") ||
+				strings.HasPrefix(url, "pulumi.com") {
 				return fmt.Errorf("%s is not a valid self-hosted backend, "+
-					"use `pulumi login` without any arguments to log into the Pulumi service backend", cloudURL)
+					"use `pulumi login` without arguments to log into the Pulumi service backend", cloudURL)
 			} else {
 				// Ensure we have the correct cloudurl type before logging in
 				if err := validateCloudBackendType(cloudURL); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/cmd/pulumi/login.go
```diff
@@ -120,6 +120,10 @@ func newLoginCmd() *cobra.Command {
 				if err != nil {
 					return fmt.Errorf("could not determine current cloud: %w", err)
 				}
+			} else if url := strings.TrimPrefix(strings.TrimPrefix(
+				cloudURL, "https://"), "http://"); strings.HasPrefix(url, "app.pulumi.com/") {
```
Frassle, line 124: Wonder if we should just do contains check for pulumi.com? Would also catch if people tried to do like `pulumi login pulumi.com`

pkg/cmd/pulumi/login.go
```diff
@@ -120,6 +120,10 @@ func newLoginCmd() *cobra.Command {
...
+				return fmt.Errorf("%s is not a valid self-hosted backend, "+
+					"use `pulumi login` without any arguments to log into the Pulumi service backend", cloudURL)
```
susanev, line 126: this is kinda nity, so feel free to ignore if you feel strongly

```suggestion
					"use only `pulumi login` to log into the Pulumi service backend", cloudURL)
```

susanev (approved):

PR #9750, opened 2022-06-01
Warn about missing AdditionalSecretOutputs

Frassle:
# Description

Currently if a user specifies a key in additional secret outputs that doesn't match any of the resources property keys the engine simply ignores it. This could be _very_ confusing, especially given the mismatches between snakeCase and camel_case property keys. 

This PR changes the engine to emit a warning diagnostic if a given additional secret key doesn't match any of the resources property keys.

## Checklist


AaronFriel:
@Frassle Should we go a step further here and for Python in particular, handle the casing for the user?

Relatedly, what do we do if the user gives us snake cased `ignoreChanges` paths in Python?

iwahbe (approved):
Looks good

PR #9760, opened 2022-06-02
[engine] Remove a copy of schema bytes, use regexp over string.

AaronFriel:
Found another place we copy the schema.

### DIFF
--- a/pkg/codegen/schema/loader.go
+++ b/pkg/codegen/schema/loader.go
@@ -149,15 +149,12 @@ func (f getSchemaNotImplemented) Error() string {
 	return fmt.Sprintf("it looks like GetSchema is not implemented")
 }
 
+var schemaIsEmptyRE = regexp.MustCompile(`\s*\{\s*\}\s*$`)
+
 func schemaIsEmpty(schemaBytes []byte) bool {
 	// We assume that GetSchema isn't implemented it something of the form "{[\t\n ]*}" is
 	// returned. That is what we did in the past when we chose not to implement GetSchema.
-	if matched, err := regexp.Match(`\s*\{\s*\}\s*$`, schemaBytes); err != nil || matched {
-		// err != nil if the regexp is invalid, unit tests validate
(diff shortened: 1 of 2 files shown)


pkg/codegen/schema/loader.go
```diff
@@ -171,9 +182,7 @@ func (l *pluginLoader) LoadPackageReference(pkg string, version *semver.Version)
 	if err != nil {
 		return nil, err
 	}
-	// We assume that GetSchema isn't implemented it something of the form "{[\t\n ]*}" is
-	// returned. That is what we did in the past when we chose not to implement GetSchema.
-	if s := string(schemaBytes); strings.TrimSpace(strings.TrimSuffix(strings.TrimPrefix(s, "{"), "}")) == "" {
```
AaronFriel, line 176: Gotta love casts that copy.

iwahbe: The cast should not copy, but research tells me they do... 😒.

pkg/codegen/schema/loader.go
```diff
@@ -149,6 +149,17 @@ func (f getSchemaNotImplemented) Error() string {
 	return fmt.Sprintf("it looks like GetSchema is not implemented")
 }
```
iwahbe, line 151: We can move the regex outside the function:
```suggestion
var schemaIsEmptyRE = regexp.MustCompile(`\s*\{\s*\}\s*$`)
```
and then `schemaIsEmpty` becomes `return schemaIsEmptyRE.matches(schemaBytes)`. This will clean up the code and ensure that the regex is valid.

iwahbe (approved):

pkg/codegen/schema/loader.go
AaronFriel: Good change - implemented and will merge with that.

mikhailshilkov:
Why have we merged code that does pass linter and breaks the build?

PR #9767, opened 2022-06-02
Fix test-macos-cron.yml

t0yv0:
# Description

Fix another cron job neglected by the Goreleaser refactor.

Fixes # (issue)

## Checklist


.github/workflows/test-macos-cron.yml
```diff
@@ -1,24 +1,25 @@
-name: Test Windows (Daily)
+name: Test Mac OS (Daily)
```
mikhailshilkov, line 1: I wondered why we have two windows jobs...

PR #9769, opened 2022-06-02
[python,nodejs] Speed up codegen tests

iwahbe:
# Description

Two codegen optimizations on the longest running languages:
- Python: Don't run pyTestCheck unless there were mixins that add tests. On my machine this changes the python test time from ~340s to ~55s.
- TypeScript: Add the `tsc` flags '--skipLibCheck true` and `--skipDefaultLibCheck true'. This cuts ~50s off of the
total time needed in github.com/pulumi/pulumi/pkg/v3/codegen/nodejs. This is valid if we assume that packages which we depend on are valid. 

## Checklist

### DIFF
--- a/pkg/codegen/python/gen_test.go
+++ b/pkg/codegen/python/gen_test.go
@@ -150,6 +150,11 @@ func buildVirtualEnv() error {
 }
 
 func pyTestCheck(t *testing.T, codeDir string) {
+	extraDir := filepath.Join(filepath.Dir(codeDir), "python-extras")
+	if _, err := os.Stat(extraDir); os.IsNotExist(err) {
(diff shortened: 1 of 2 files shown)


pkg/codegen/python/gen_test.go
```diff
@@ -150,6 +150,11 @@ func buildVirtualEnv() error {
 }
 
 func pyTestCheck(t *testing.T, codeDir string) {
+	extraDir := filepath.Join(filepath.Dir(codeDir), "python-extras")
```
t0yv0, line 153: Thought about this one, but probably reasonable. It really does nothing but run the tests. So if no tests, don't run.

PR #9770, opened 2022-06-02
[codegen] isolate and make concurrency-safe the opaque types in binder

AaronFriel:
A test failure on the PR #9686 exposed a data race that could cause a spurious error or a panic. The data race occurs if two threads both do a Get of the same name from the global map, both find that the entry is absent, and one succeeds in inserting before the other, resulting in NewOpaqueType erroneously returning an error. The panic occurs if both threads end up in the map write at the same time.

https://github.com/pulumi/pulumi/blob/36cbf572f45790f6859a8da3f8d67f6b32620fca/pkg/codegen/pcl/binder_schema.go#L237-L239

### DIFF
--- a/pkg/codegen/hcl2/model/type_opaque.go
+++ b/pkg/codegen/hcl2/model/type_opaque.go
@@ -1,4 +1,4 @@
-// Copyright 2016-2020, Pulumi Corporation.
+// Copyright 2016-2022, Pulumi Corporation.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -16,6 +16,7 @@ package model
 
 import (
 	"fmt"
+	"sync"
 
 	"github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/hclsyntax"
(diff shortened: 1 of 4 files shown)


pkg/codegen/hcl2/model/type_opaque.go
```diff
@@ -34,33 +35,74 @@ type OpaqueType struct {
...
-// The set of opaque types, indexed by name.
-var opaqueTypes = map[string]*OpaqueType{}
+// A global type map for built-in types, e.g.: "bool"
+var globalTypes = make(map[string]*OpaqueType)
 
-// GetOpaqueType fetches the opaque type for the given name.
-func GetOpaqueType(name string) (*OpaqueType, bool) {
-	t, ok := opaqueTypes[name]
-	return t, ok
+type OpaqueTypeMap interface {
```
AaronFriel, line 41: An interface was chosen here to ensure that anything that constructs this, does so by pointer ref and thus the mutex is copied by reference and not value.

AaronFriel:
@iwahbe going to revise this based on conversation with @pgavlin and remove the interning of opaque types entirely.

AaronFriel:
@iwahbe none of the fields of the OpaqueType were used except for Name, this simplifies it down to a string alias.

iwahbe (approved):

PR #9771, opened 2022-06-02
[pkg] Enable main_test to compile on go 1.17 and go 1.18

AaronFriel:
This file previously compiled only on go 1.17 as the signature for `MainStart` changed. This adds the necessary methods and via conditional compilation, targets pre-1.18 and 1.18 methods.

### DIFF
--- a/pkg/cmd/pulumi/main_test.go
+++ b/pkg/cmd/pulumi/main_test.go
@@ -20,13 +20,24 @@ import (
 	"io"
 	"os"
 	"path"
+	"reflect"
 	"testing"
+	"time"
 
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 	"github.com/spf13/pflag"
 )
 
 type noTestDeps int
+
+type corpusEntry = struct {
+	Parent     string
+	Path       string
+	Data       []byte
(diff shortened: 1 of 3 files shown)


pkg/cmd/pulumi/main_test.go
```diff
@@ -20,14 +20,25 @@ import (
...
 type noTestDeps int
 
+type corpusEntry = struct {
+	Parent     string
+	Path       string
+	Data       []byte
+	Values     []interface{}
+	Generation int
+	IsSeed     bool
+}
```
AaronFriel, line 40: unexported type from testing package, we need an alias definition to implement the methods below.

t0yv0 (approved):
LGTM. Something very tricky is happening with noTestDeps, do you know what's that for?

AaronFriel:
@t0yv0 I only know what we've done with it before & that MainStart has an internal type that behaves a bit like ours.

The doc comment on the testDeps interface:

```
// testDeps is an internal interface of functionality that is
// passed into this package by a test's generated main package.
// The canonical implementation of this interface is
// testing/internal/testdeps's TestDeps.
```

And there's this internal implementation called `matchStringOnly` that has similarly default implementations.

Issue #8436, opened 2021-11-16
Automation API create_or_select_stack ignores work_dir parameter


felixwang9817:
## Hello!

- Vote on this issue by adding a 👍 reaction
- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

### Steps to reproduce

This [line](https://github.com/pulumi/pulumi/blob/83e24765f3d3ddc99d26e8785b6c38d13eb4cfe3/sdk/python/lib/pulumi/automation/_local_workspace.py#L434) does not use the `work_dir` parameter, which seems to be a mistake (at the very least, it is inconsistent with the behavior described in the docstring).

Expected: 
Actual:

emiliza:
Thanks for filing @felixwang9817 - this is interesting. It's unclear why `work_dir` isn't explicitly called out there as a top level option, but this does exist in `opts`

felixwang9817:
@emiliza yes, I got the intended result by setting `work_dir` in `opts`. I just wanted to report this since the docs are incorrect, which may confuse folks that don't want to look at the source code.

squaremo:
I think this is a case of misleading API. The *_stack procedures behaves modally, depending on which keyword arguments you supply:

 - if `program` and `project_name` are in the keyword args (see https://github.com/pulumi/pulumi/blob/06ba63bb57e90706c1550861b785075ae860144a/sdk/python/lib/pulumi/automation/_local_workspace.py#L386), the resulting stack refers to an inline program;
 - otherwise if `work_dir` is in the keyword args (see https://github.com/pulumi/pulumi/blob/06ba63bb57e90706c1550861b785075ae860144a/sdk/python/lib/pulumi/automation/_local_workspace.py#L541), the stack refers to a program constructed from that directory.

The `work_dir` _keyword argument_ means "this is where the program is". There's _also_ a field in the `opts` argument called `work_dir`, which means "use this as the current directory". For a local program, the keyword argument overrides the option (https://github.com/pulumi/pulumi/blob/06ba63bb57e90706c1550861b785075ae860144a/sdk/python/lib/pulumi/automation/_local_workspace.py#L552); for an inline program, the `work_dir` keyword argument is ignored and the `work_dir` option is defaulted.

The docstring says (emphasis mine) "unless a `work_dir` **option** is specified" -- so it's accurate, but a bit of a trap.

It would be less confusing if the keyword argument had a different name like `project_dir`. But changing it would break existing code; so I suggest three steps:

[...]

PR #9794, opened 2022-06-07
Further explain kwargs for Python stack funcs

squaremo:
The *_stack functions in the pulumi.automation package take keyword arguments that control where the program used in the stack comes from. In particular, the keyword argument work_dir (and absence of program and project_name) indicates that the program should be constructed using the given directory. There's _also_ a field with the same name accepted in the opts argument, leading to confusion in at least one case (see https://github.com/pulumi/pulumi/issues/8436).

This commit tries to mitigate the confusion by making the docstrings for those functions more explicit about which arguments are required and ignored in each case.

Addresses #8436.

## Checklist

No additional or adapted tests -- Python docstring changes only.


Frassle (approved):
LGTM

PR #9798, opened 2022-06-07
Embed the respected version in default opts

iwahbe:
# Description

When `languages.go.respectSchemaVersion = true` make sure that resource and invoke requests have the correct version embedded. 
## Checklist


t0yv0 (approved):
Lacking context for these but looks reasonable!

PR #9804, opened 2022-06-08
Clone and test pulumi/templates from the current working branch

Zaid-Ajaj:
# Description

Right now, when we are raising a PR in pulumi/templates, it uses [workspace.RetrieveTemplates](https://github.com/pulumi/templates/blob/master/tests/template_test.go#L70) which in turn calls [retrievePulumiTemplates](https://github.com/pulumi/pulumi/blob/master/sdk/go/common/workspace/templates.go#L311) in pulumi/pulumi that clones and tests the templates from `master` branch of pulumi/templates instead of cloning and testing the template changes from the raised PR 

> see [pulumi/templates#294](https://github.com/pulumi/templates/pull/294) where the problem comes to light: github actions building outdated templates from master branch and not the new ones

This PR makes sure that if we are running the tests of pulumi/templates that we are also cloning the right branch, not always defaulting to `master`

## Checklist

### DIFF
--- a/sdk/go/common/workspace/templates.go
+++ b/sdk/go/common/workspace/templates.go
@@ -327,6 +327,20 @@ func retrievePulumiTemplates(templateName string, offline bool, templateKind Tem
 		return TemplateRepository{}, err
 	}
 
+	currentGithubRepo := os.Getenv("GITHUB_REPOSITORY")
+
+	if !offline && currentGithubRepo == "pulumi/templates" {
+		// means we are running this function from pulumi/templates
+		// then make sure we are retrieving the templates from the branch
+		// that had triggered this code so that we are testing the right changes
+		currentGithubBranch := os.Getenv("GITHUB_BRANCH")
+		templatesBranch := plumbing.NewBranchReferenceName(currentGithubBranch)
+		err := gitutil.GitCloneOrPull(currentGithubRepo, templatesBranch, templateDir, false /*shallow*/)
+		if err != nil {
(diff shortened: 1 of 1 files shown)


iwahbe (changes requested):
Won't this pull in both the current branch and the default branch (since the `if !offline && currentGithubRepo == ...` block won't return without an error. Is that the intention? Otherwise, could we simplify by embedding something like the following in the `if !offline {` block:
```go
if b := os.Getenv("GITHUB_BRANCH"); os.Getenv("GITHUB_REPOSITORY") == "pulumi/templates" && b != "" {
    pulumiTemplatesBranch = b
}
```

Zaid-Ajaj:
@iwahbe nice catch! I've simplified it now

sdk/go/common/workspace/templates.go
```diff
@@ -326,7 +326,7 @@ func retrievePulumiTemplates(templateName string, offline bool, templateKind Tem
 	if err := os.MkdirAll(templateDir, 0700); err != nil {
 		return TemplateRepository{}, err
 	}
-
+
```
Frassle, line 329:
```suggestion

```

Frassle (approved):

sdk/go/common/workspace/templates.go
Zaid-Ajaj: Done!

iwahbe (approved):

PR #9810, opened 2022-06-08
[utils] Handle nils in mapper encoding

pgavlin:
The mapper was not handling nil values correctly during encoding. When
attempting to encode a nil `reflect.Value`, the mapper was converting it
to a nil `interface{}`, then taking the `reflect.ValueOf` that interface
value, leading to an invalid `reflect.Value` that tripped an assertion.
These changes avoid the intermediate `interface{}` and consequent
invalid `reflect.Value`.

Contributes to https://github.com/pulumi/pulumi-command/issues/41 and
https://github.com/pulumi/pulumi-command/issues/68.


pgavlin:
I'll be adding tests later today. Once these changes are in, they'll need to be rolled out to `pulumi-command` in order to fix the referenced issues.

jkisk (approved):

PR #9838, opened 2022-06-10
Allow a path to a file when constructing an archive

squaremo:
# Description

It's legitimate to construct an archive with a single file (rather than a directory) by mentioning the file's name:

    zip handler.zip handler/main

In fact, that's what the AWS Lambda docs suggest: https://docs.aws.amazon.com/lambda/latest/dg/golang-package.html

This PR lets you do the equivalent when making an `asset.FileArchive`. It needs a little rearrangement of `directoryArchiveReader` so that the individual paths are relative to a base rather than sharing a prefix; otherwise, a single-file archive (in which the name is the entire file path) cannot be constructed. I've sign-posted this generalisation by renaming the helper struct in question to `filesystemArchiveReader`.

In the context of making aws.lambda.Function objects, this is useful for uploading a binary (e.g., using the `go1.x` runtime) conveniently, rather than making your own ZIP file out of band, or including irrelevant files.

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,3 +1,6 @@
 ### Improvements
 
+- [sdk] Allow constructing archives with just a file
+  [#9838](https://github.com/pulumi/pulumi/pull/9838)
+
 ### Bug Fixes
--- a/sdk/go/common/resource/asset.go
+++ b/sdk/go/common/resource/asset.go
@@ -769,13 +769,13 @@ func (a *Archive) readAssets() (ArchiveReader, error) {
 	return r, nil
 }
 
-// directoryArchiveReader is used to read an archive that is represented by a directory in the host filesystem.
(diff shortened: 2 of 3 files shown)


Frassle:
Why do we want this implicit rather than explicitly via an asset archive?
```
NewAssetArchive({"file.txt": NewFileAsset("file.txt")})
```

squaremo:
> Why do we want this implicit rather than explicitly via an asset archive?

People familiar with tar or ZIP, or just taking the AWS docs as a guide, will expect to be able to make an archive with a file directly. Yes, it is possible to get the same result in the API as it works now. It's also possible to list out the files in a directory explicitly, and construct an AssetArchive from those; yet FileArchive is a useful part of the API, because it saves that labour and works the way people expect archives to work. This change makes that more so, by removing an unexpected error case.

PR #9840, opened 2022-06-10
Update protoc-gen-go to 1.5.2

Frassle:
Continuing the baby steps to get grpc generation updated.

The new version of gen-go requires a go_package option so that import paths can be done correctly. Not that our generation needs full import paths because it's all one go package.


AaronFriel:
@Frassle lot of codegenerated changes including some binary strings in this PR, going to pull it down and check the diff.

AaronFriel (approved):

PR #9842, opened 2022-06-10
[sdk/go] Ensure stackref getters localize secret-ness to the key retrieved

### LINKED ISSUE #9564: Pulumi output with a single secret taints all other outputs when using Go SDK
### What happened?

When using the Go SDK, if a stack has any secret in the output, and we reference any of the outputs (secret or not), Pulumi will treat that output as a secret. This is problematic since `pulumi preview` will mask the output even though it isn't secret and we'd like to see the change. Additionally, if the output is used in the context of something like a Kubernetes env var, this issue in conjunction with https://github.com/pulumi/pulumi-kubernetes/issues/1576 causes the entire env var array to be labeled as secret.

I believe that there was a fix for this for the Node and Python SDKs in https://github.com/pulumi/pulumi/issues/2744, but no such fix was created for the Go SDK. In the meantime, I've just been wrapping all the outputs in `pulumi.Unsecret` - let me know if there's a better workaround here. If this is indeed an issue, I can also go ahead and start working on a fix. 

### Steps to reproduce

In my case, I created a Kubernetes deployment, which had the following:
```
Env: core.EnvVarArray{
	core.EnvVarArgs{
		Name: pulumi.String("some-name"),
		Value: stackRef.GetStringOutput(pulumi.String("some-output")),
	},
}
```
where `some-output` is a non-secret output from another stack.

Then, after running `pulumi preview`, I see something like 
```
~ spec: {
  ~ template: {
      ~ spec: {
          ~ containers: [
              ~ [0]: {
                      ~ env: [
                          ~ [0]: {
                                  ~ name: [secret] => [secret]
                                }
                        ]
                    }
            ]
        }
    }
}
```

### Expected Behavior

Pulumi not to treat the output as a secret (and causing the env var to be considered a secret) 

### Actual Behavior

Pulumi considered the output to be a secret which caused the preview to show `[secret] => [secret]`

### Versions used

```
pulumi about
CLI
Version      3.32.1
Go Version   go1.18.1
Go Compiler  gc

Plugins
NAME        VERSION
gcp         6.20.0
go          unknown
kubernetes  3.18.3

Host
OS       darwin
Version  12.3.1
Arch     arm64

This project is written in go (/opt/homebrew/bin/go v1.18.1 darwin/arm64)

...

NAME                                        VERSION
github.com/MakeNowJust/heredoc/v2           v2.0.1
github.com/pkg/errors                       v0.9.1
github.com/pulumi/pulumi-gcp/sdk/v6         v6.20.0
github.com/pulumi/pulumi-kubernetes/sdk/v3  v3.18.3
github.com/pulumi/pulumi/sdk                v1.14.1
github.com/pulumi/pulumi/sdk/v3             v3.30.0
k8s.io/apimachinery                         v0.23.6
```

### Additional context

_No response_

### Contributing

AaronFriel:
Fixes #9564

### DIFF
--- a/sdk/go/pulumi/stack_reference.go
+++ b/sdk/go/pulumi/stack_reference.go
@@ -3,6 +3,8 @@ package pulumi
 import (
 	"fmt"
 	"reflect"
+
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
 )
 
 // StackReference manages a reference to a Pulumi stack.
@@ -22,21 +24,30 @@ type StackReference struct {
 // GetOutput returns a stack output keyed by the given name as an AnyOutput
 // If the given name is not present in the StackReference, Output<nil> is returned.
 func (s *StackReference) GetOutput(name StringInput) AnyOutput {
-	return All(name, s.Outputs).
-		ApplyT(func(args []interface{}) interface{} {
-			n, outs := args[0].(string), args[1].(map[string]interface{})
-			v, ok := outs[n]
-			if ok {
-				return v
+	return All(name, s.rawOutputs).
+		ApplyT(func(args []interface{}) (interface{}, error) {
+			n, stack := args[0].(string), args[1].(resource.PropertyMap)
+			if !stack["outputs"].IsObject() {
+				return Any(nil), fmt.Errorf("failed to convert %T to object", stack)
 			}
-			if s.ctx.DryRun() {
-				// It is a dry run, so it is safe to return an unknown output.
-				return UnsafeUnknownOutput([]Resource{s})
+			outs := stack["outputs"].ObjectValue()
+			v, ok := outs[resource.PropertyKey(n)]
+			if !ok {
+				if s.ctx.DryRun() {
+					// It is a dry run, so it is safe to return an unknown output.
+					return UnsafeUnknownOutput([]Resource{s}), nil
+				}
+
+				// We don't return an error to remain consistent with other SDKs regarding missing keys.
+				return nil, nil
 			}
 
-			// We don't return an error to remain consistent with other SDKs regarding
-			// missing keys.
-			return nil
+			ret, secret, _ := unmarshalPropertyValue(s.ctx, v)
+
+			if secret {
+				ret = ToSecret(ret)
+			}
+			return ret, nil
 		}).(AnyOutput)
 }
 
--- a/sdk/go/pulumi/stack_reference_test.go
+++ b/sdk/go/pulumi/stack_reference_test.go
@@ -2,6 +2,7 @@ package pulumi
 
 import (
 	"fmt"
+	"strings"
 	"testing"
 
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
@@ -56,6 +57,9 @@ func TestStackReference(t *testing.T) {
 		zed1, _, _, _, err := await(ref1.GetOutput(String("zed")))
 		assert.NoError(t, err)
 		assert.Equal(t, outputs["zed"], zed1)
+		nonexistant, _, _, _, err := await(ref0.GetOutput(String("nonexistant")))
+		assert.NoError(t, err)
+		assert.Equal(t, nil, nonexistant)
 		numf, _, _, _, err := await(ref1.GetFloat64Output(String("numf")))
 		assert.NoError(t, err)
 		assert.Equal(t, outputs["numf"], numf)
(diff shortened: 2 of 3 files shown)


sdk/go/pulumi/stack_reference.go
```diff
@@ -13,14 +15,26 @@ type StackReference struct {
 	Name StringOutput `pulumi:"name"`
 	// Outputs resolves with exports from the named stack
 	Outputs MapOutput `pulumi:"outputs"`
+
+	// Engine context in which resolve resource references in outputs.
+	Context *Context
```
AaronFriel, line 20: I was reluctant to add this, as I'm unsure of the mechanics here. Can stacks export a resource as an output, and that would _work_?

`unmarshalPropertyValue` requires a valid Pulumi engine Context to handle that case. Is it worthwhile to split that function to support this use case?

AaronFriel, line 20: I was reluctant to add this because it's adding a field to an exported type in the SDK, but I don't expect people are constructing this except via NewStackReference.

`unmarshalPropertyValue` requires a valid Pulumi engine Context to handle resource references. Can stacks export a resource as an output, and that would _work_? I don't expect it would.

sdk/go/pulumi/stack_reference_test.go
```diff
@@ -72,3 +73,85 @@ func TestStackReference(t *testing.T) {
...
+		assert.Equal(t, expected, outs1)
+
+		for k, v := range expected {
+			shouldSecret := strings.HasPrefix(k, "secret-")
+
+			for _, ref := range []*StackReference{ref0, ref1} {
+				outputV, known, secret, _, err := await(ref.GetOutput(String(k)))
+				assert.NoError(t, err)
+				assert.True(t, known)
+				assert.Equal(t, shouldSecret, secret)
```
AaronFriel, line 149: TDDed this.

sdk/go/pulumi/stack_reference.go
iwahbe: Is there a reason we couldn't make `Context` private. I don't think it's breaking to add a private field to an exported type.

AaronFriel: I don't see a reason why not. I think it's a breaking change only in the sense that the type is non-constructible by third party packages, but I think that was already the case with the embedded type.

iwahbe: I might be wrong, but I believe that all types are constructible. You just might not be able to construct a valid type on your own. For example, the following is valid:
```go
var f os.File
f.Chdir()
````
This is true even though the std definition of `os.File` is 
```go
// File represents an open file descriptor.
type File struct {
	*file // os specific
}
```

I'm pretty sure that given package foo
```go
// Pkg foo
type Foo struct {
Public string
private string
}
```
You can still do ` foo.Foo { Public: "foo" }` in another package. You just can't name `foo.Foo.private`.

AaronFriel: Some day I will internalize Go's fascination with meaningful-zero-values-by-default.

sdk/go/pulumi/stack_reference.go
```diff
@@ -95,5 +109,7 @@ func NewStackReference(ctx *Context, name string, args *StackReferenceArgs,
 		return nil, err
 	}
 
+	ref.context = ctx
```
iwahbe, line 112: Nit: I think the following is cleaner:
```go
ref := StackReference{ context: ctx }
```

sdk/go/pulumi/stack_reference.go
```diff
@@ -13,14 +15,26 @@ type StackReference struct {
...
+			if !stack["outputs"].IsObject() {
+				return Any(nil), fmt.Errorf("failed to convert %T to object", stack)
+			}
+			outs := stack["outputs"].ObjectValue()
+			ret, secret, _ := unmarshalPropertyValue(s.context, outs[resource.PropertyKey(n)])
+
+			if secret {
+				ret = ToSecret(ret)
+			}
+			return Any(ret), nil
```
iwahbe, line 37: This is all conflict with https://github.com/pulumi/pulumi/pull/9855. 😄

iwahbe (approved):
Looks good. LGTM

sdk/go/pulumi/stack_reference.go
AaronFriel: Rebased on your PR

sdk/go/pulumi/stack_reference.go
AaronFriel: Yours looked good, I tested the nonexistant key behavior, merged & then included tests in this PR :)

PR #9855, opened 2022-06-14
[sdk/go] Improve unknown stack references preview

### LINKED ISSUE #9365: Issues with previews of array values from StackReferences
Notes from feedback from a user:

> I have a stack which pulls a StackReference, which will eventually have an output that isn’t there yet. In a preview, I see this:
> image
>  the second is from a stack reference, the first was created in the program
> 1) the null is wrong (and that value is known)
> 2) the “” is definitely wrong, at best it should be unknown, it should 100% not be an empty string.

iwahbe:
# Description

When a valid stack reference is asked to retrieve an absent key as a string, the default value is returned. This results in previews displaying `""` as the value of the stack reference key. See the [linked issue](https://github.com/pulumi/pulumi/issues/9365) for an example. 

This PR makes the following changes:
1. When the value is absent during a preview, return an unknown output. This fixes the linked issue.
2. When the value is absent during a `pulumi up`, error when a string is requested. This brings the behavior of `string` inline with `Int` and `Float64`.

Fixes #9365

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -7,3 +7,6 @@
 
 - [sdk/go] Correctly parse nested git projects in GitLab.
   [#9354](https://github.com/pulumi/pulumi/issues/9354)
+
+- [sdk/go] Mark StackReference keys that don't exist as unknown. Error when converting unknown keys to strings.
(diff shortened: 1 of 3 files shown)


AaronFriel (approved):
lgtm

PR #9857, opened 2022-06-14
Fix downstream codegen test failures

AaronFriel:
Downstream tests will fail as they update their provider binaries to use go 1.18: https://github.com/pulumi/pulumi/runs/6872372597?check_suite_focus=true#step:8:49

This tactically only updates the Go version for these tests, as we don't want to inadvertently depend on any 1.18 behavior in the Go SDK yet.

Fixes: #9862

### DIFF
--- a/.github/workflows/run-codegen-test.yml
+++ b/.github/workflows/run-codegen-test.yml
@@ -53,7 +53,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        provider: ["aws", "gcp", "azure", "azuread", "random", "kubernetes", "azure-native"]
+        provider: ["aws", "gcp", "azure", "azuread", "random", "kubernetes"]
       fail-fast: false
     steps:
       - name: Install Go
--- a/.github/workflows/run-codegen-test.yml
+++ b/.github/workflows/run-codegen-test.yml
@@ -15,17 +15,34 @@ env:
 jobs:
   comment-notification:
     runs-on: ubuntu-latest
-    if: ${{ always() }}
+    if: github.event_name == 'repository_dispatch'
     steps:
       - name: Create URL to the run output
         id: vars
-        run: echo ::set-output name=run-url::https://$GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID
+        run: echo ::set-output name=run-url::$GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID
       - name: Update with Result
         uses: peter-evans/create-or-update-comment@v2
         with:
-          token: ${{ secrets.GITHUB_TOKEN }}
+          token: ${{ secrets.PULUMI_BOT_TOKEN }}
           repository: ${{ github.event.client_payload.github.payload.repository.full_name }}
+          comment-id: ${{ github.event.client_payload.github.payload.comment.id }}
           issue-number: ${{ github.event.client_payload.github.payload.issue.number }}
+          body: |
+            Please view the results of the Downstream Codegen Tests [Here][1]
+
+            [1]: ${{ steps.vars.outputs.run-url }}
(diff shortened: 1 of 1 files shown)


Frassle:
We had the same thought https://github.com/pulumi/pulumi/pull/9803

Unfortunately we also need to update the call to "go mod tidy" in https://github.com/pulumi/action-test-provider-downstream before this will work. See [my branch](https://github.com/pulumi/action-test-provider-downstream/tree/fraser/v8) for what I think will fix it.

AaronFriel:
@Frassle @iwahbe this looks ready to review:

The slash command failed to make a comment due to rate limits (not surprising) but the run is proceeding here: https://github.com/pulumi/pulumi/actions/runs/2498702244

The tests that ran as a result of a push now generate summary results, which don't require a GitHub API call here: https://github.com/pulumi/pulumi/actions/runs/2498461501

It looks like Azure Native has issues with it OOMing, so I'll ping providers on this PR again to reduce memory usage:
- https://github.com/pulumi/pulumi-azure-native/pull/1689

.github/workflows/run-codegen-test.yml
```diff
@@ -65,14 +82,13 @@ jobs:
         with:
           ref: ${{ env.PR_COMMIT_SHA }}
       - name: Test Downstream
-        uses: pulumi/action-test-provider-downstream@releases/v7
+        uses: pulumi/action-test-provider-downstream@friel/v8
```
Frassle, line 85: Can we get this merged to the downstream repo proper rather than running off a branch. Also we really ought to tighten that repo up to run of main and have tags.

Frassle (changes requested):

.github/workflows/run-codegen-test.yml
AaronFriel: I can make the PR and then make a tag

iwahbe (approved):
I agree with @Frassle's comment, but otherwise LGTM. 

BTW it looks like azure-native is actually failing, not OOMing.

AaronFriel:
Perhaps we should remove azure-native from the testing harness if it takes 1h37m.

AaronFriel:
@iwahbe I've removed azure-native from the list of tests to run, as I don't think we can get good signal on it and I don't want to block this PR on optimizing an 1h30m job.

Once you or @Frassle reviews this PR on the action:
- https://github.com/pulumi/action-test-provider-downstream/pull/11

I'll push a tag and update this PR to reference the tag.

.github/workflows/run-codegen-test.yml
AaronFriel: Now targeting `@v1.0.0` released tag.

.github/workflows/run-codegen-test.yml
```diff
@@ -14,35 +14,52 @@ env:
...
+
             [1]: ${{ steps.vars.outputs.run-url }}
   downstream-test:
     name: Test ${{ matrix.provider }} Downstream
     if: github.event_name == 'repository_dispatch' || github.event.pull_request.head.repo.full_name == github.repository
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        provider: ["aws", "gcp", "azure", "azuread", "random", "kubernetes", "azure-native"]
+        provider: ["aws", "gcp", "azure", "azuread", "random", "kubernetes"]
```
Frassle, line 56: https://github.com/pulumi/pulumi/issues/9874 to track adding this back.

.github/workflows/run-codegen-test.yml
```diff
@@ -65,14 +82,13 @@ jobs:
         with:
           ref: ${{ env.PR_COMMIT_SHA }}
       - name: Test Downstream
-        uses: pulumi/action-test-provider-downstream@releases/v7
+        uses: pulumi/action-test-provider-downstream@v1.0.0
```
Frassle, line 85: Hey hey tagged releases ❤️

Frassle (approved):

Issue #9789, opened 2022-06-07
failed to compute archive hash for "code": archive/tar: missed writing 2026611 bytes


bajtos:
### What happened?

I am managing my AWS Lambda function using Pulumi, using `pulumi.asset.AssetArchive` to provide list of files to upload.

Two to three weeks ago, I started to observe problems when running `pulumi up` locally on my macOS M1 machine. The archive is uploaded to AWS, but some of the larger JavaScript files are truncated in the middle. (The file sizes are between 500KB to 5MB). Fortunately, this problem was not happening when running `pulumi up` from GitHub Actions.

Occasionally, I was receiving the following error on my machine, but again not in GitHub Actions:

Formatted for clarity:
```
Diagnostics:
  pulumi:pulumi:Stack (my-stack-name):
    error: Running program '/home/runner/work/my-project/my-project/infra'
      failed with an unhandled exception:
    Error: failed to register new resource 
      my-api-handler [aws:lambda/function:Function]: 2 
      UNKNOWN: failed to compute archive hash for "code": 
      archive/tar: missed writing 2026611 bytes
```

Full error:

```

Diagnostics:
  pulumi:pulumi:Stack (my-stack-name):
    error: Running program '/home/runner/work/my-project/my-project/infra' failed with an unhandled exception:
    Error: failed to register new resource my-api-handler [aws:lambda/function:Function]: 2 UNKNOWN: failed to compute archive hash for "code": archive/tar: missed writing 2026611 bytes
        at Object.registerResource (/home/runner/work/my-project/my-project/node_modules/.pnpm/@pulumi+pulumi@3.33.2/node_modules/@pulumi/runtime/resource.ts:295:27)
        at new Resource (/home/runner/work/my-project/my-project/node_modules/.pnpm/@pulumi+pulumi@3.33.2/node_modules/@pulumi/resource.ts:393:13)
        at new CustomResource (/home/runner/work/my-project/my-project/node_modules/.pnpm/@pulumi+pulumi@3.33.2/node_modules/@pulumi/resource.ts:772:9)
        at new Function (/home/runner/work/my-project/my-project/node_modules/.pnpm/@pulumi+aws@5.6.0/node_modules/@pulumi/lambda/function.ts:505:9)
        at new ZebraDashboardApiLambda (file:///home/runner/work/my-project/my-project/infra/dashboard-lambda.js:88:20)
[...]

PR #9859, opened 2022-06-14
Add some more checks that io.Copy copies as much as expected

Frassle:
More error checks for https://github.com/pulumi/pulumi/issues/9789

### DIFF
--- a/sdk/go/common/resource/asset.go
+++ b/sdk/go/common/resource/asset.go
@@ -391,9 +391,12 @@ func (a *Asset) EnsureHash() error {
 		defer contract.IgnoreClose(blob)
 
 		hash := sha256.New()
-		_, err = io.Copy(hash, blob)
+		n, err := io.Copy(hash, blob)
 		if err != nil {
 			return err
+		}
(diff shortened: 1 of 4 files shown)


sdk/go/common/resource/asset.go
```diff
@@ -940,7 +943,7 @@ func (a *Archive) Archive(format ArchiveFormat, w io.Writer) error {
 		if err != nil {
 			return err
 		}
-		_, err := io.Copy(w, ss)
+		_, err = io.Copy(w, ss)
```
bajtos, line 946: Missing size check?

```suggestion
		n, err = io.Copy(w, ss)
```

bajtos: (I am not sure if it's possible to find out what is the size of an `ArchiveFormat` though.)

Frassle: Yes, I think I started changing this then realized that there was no expected size to check against.

iwahbe (approved):
LGTM

PR #9860, opened 2022-06-14
Use await not Task.Assign

Frassle:
# Description

Noticed this while looking around the Output code checking for async bugs. I don't think this is a bug, the current logic looks like it handles cancellations and exceptions ok but there's no need for it. We can just use the compilers await logic to do the mapping from a task with a value to an OutputData structure. Also saves the extra TaskCompletionSource object.


Zaid-Ajaj (approved):
Nice refactor ✅

Issue #9704, opened 2022-05-26
Do not expose IgnoreChanges on ComponentResourceOptions


guineveresaenger:
### What happened?

A community member was[ attempting to use `IgnoreChanges` on a Component Resource](https://github.com/pulumi/pulumi-eks/issues/709), in this case `eks.ManagedNodeGroup` and found it to be not working as they expected.

The difference between Resources and Components is subtle from a user perspective, and it seems reasonable to want to turn that knob. However, due to the specific implementation of schematized component resources, it currently is not possible to use the `IgnoreChanges` option, and currently, as per @pgavlin, this is intentional.

In the future, we may implement this option, but to expose it currently is confusing, since it doesn't actually work on Component resources. 

### Steps to reproduce

Use a component and watch your program compile with `opts=pulumi.ResourceOptions(ignore_changes=["root.nested"]),` but fail to ignore changes on `pulumi up`.

### Expected Behavior

This is really more of a UX bug, but due to this option being available, I expect changes to be ignored on all resources, including component resources.

### Actual Behavior

Setting the option does nothing to a component resource.

### Versions used

```
CLI          
Version      3.33.1
Go Version   go1.17.10
Go Compiler  gc

Plugins
NAME        VERSION
aws         5.4.0
eks         0.40.0
kubernetes  3.19.2
python      unknown

Host     
OS       darwin
Version  11.6
Arch     x86_64
```

### Additional context

While we may in the future be able to have this option on component resources, we do not now and this is intentional.

[Internal Slack thread](https://pulumi.slack.com/archives/C02VCJEBT2N/p1653524516656529)

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

AaronFriel:
@guineveresaenger I think what we want to do here, rather than change our types (which could be a breaking change & require a 4.0 release) is to improve our docs and consider warning in the engine on setting "ignore_changes" to a component.

Frassle:
Docs update + warning + keeping a TODO comment to remind us to clean this up as part of 4.0. 

I think the warning should be pretty easy to add. Just add some diagnostic calls just before the `if remote` block at https://github.com/pulumi/pulumi/blob/master/pkg/resource/deploy/source_eval.go#L1079 if the RegisterResourceRequest had any settings that we ignore for ComponentResources. 

[...]

PR #9863, opened 2022-06-14
Add warning when using ignoreChange on Component.

RobbieMcKinstry:
# Description

This PR adds a warning when component resources have the `ignoreChanges` option applied, since that option has no effect and is ignored by the engine.

**Reviewers:** Looking for guidance on how to test Diagnostic messages. If this change warrants a test, I'm happy to add one. Could you point me to an example of how I could inspect diagnostics and confirm my change works?

@Frassle [in the original issue](https://github.com/pulumi/pulumi/issues/9704#issuecomment-1139460514), you mentioned there could be other options that are ignored by components. If that's true, could you help me determine what those options are?

Fixes https://github.com/pulumi/pulumi/issues/9704

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -3,6 +3,8 @@
 - [auto/*] Add `--policy-pack` and `--policy-pack-configs` options to automation API.
   [#9872](https://github.com/pulumi/pulumi/pull/9872)
 
+- [cli] The engine now produces a warning when the 'ignoreChanges' option is applied to a Component resource.
+
 - [sdk/python] Changed `Output[T].__str__()` to return an informative message rather than "<pulumi.output.Output object at 0x012345ABCDEF>".
   [#9848](https://github.com/pulumi/pulumi/pull/9848)
 
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -4,6 +4,7 @@
   [#9872](https://github.com/pulumi/pulumi/pull/9872)
 
 - [cli] The engine now produces a warning when the 'ignoreChanges' option is applied to a Component resource.
+  [#9863](https://github.com/pulumi/pulumi/pull/9863)
 
 - [sdk/python] Changed `Output[T].__str__()` to return an informative message rather than "<pulumi.output.Output object at 0x012345ABCDEF>".
   [#9848](https://github.com/pulumi/pulumi/pull/9848)
--- a/pkg/resource/deploy/source_eval.go
+++ b/pkg/resource/deploy/source_eval.go
@@ -470,7 +470,7 @@ func (d *defaultProviders) getDefaultProviderRef(req providers.ProviderRequest) 
 // resmon implements the pulumirpc.ResourceMonitor interface and acts as the gateway between a language runtime's
 // evaluation of a program and the internal resource planning and deployment logic.
 type resmon struct {
-	ctx                       *plugin.Context                    // the plugin context
+	diagostics                diag.Sink                          // logger for user-facing messages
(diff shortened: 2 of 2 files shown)


Frassle:
> Reviewers: Looking for guidance on how to test Diagnostic messages. If this change warrants a test, I'm happy to add one. Could you point me to an example of how I could inspect diagnostics and confirm my change works?

Look at the tests in "pkg/engine/lifeycletest" e.g. `TestAdditionalSecretOutputs` in "pkg/engine/lifeycletest/pulumi_test.go"

> @Frassle https://github.com/pulumi/pulumi/issues/9704#issuecomment-1139460514, you mentioned there could be other options that are ignored by components. If that's true, could you help me determine what those options are?

I don't think `additionalSecretOutputs` will really work, and anything around delete/replace/retain I think is meaningless for a ComponentResource.

CHANGELOG_PENDING.md
```diff
@@ -1,5 +1,7 @@
 ### Improvements
 
+- [cli] The engine now produces a warning when the 'ignoreChanges' option is applied to a Component resource.
+
```
Frassle, line 4:
```suggestion
  [#9863](https://github.com/pulumi/pulumi/pull/9863)

```

pkg/resource/deploy/source_eval.go
```diff
@@ -1159,6 +1162,16 @@ func (rm *resmon) RegisterResource(ctx context.Context,
 		outputs = filtered
 	}
 
+	// TODO(@platform): Currently component resources ignore the "ignoreChanges" option.
+	// Revisit these semantics in Pulumi v4.0
+	// See this issue for more: https://github.com/pulumi/pulumi/issues/9704
+	if !custom && len(ignoreChanges) > 0 {
+		rm.ctx.Diag.Warningf(diag.Message(
+			result.State.URN,
+			"The option 'ignoreChanges' has no effect on components. This may be changed in a future version.",
```
Frassle, line 1171: I'd drop the comment about changing in future versions.

RobbieMcKinstry: Thought that might be divisive :)
Sounds good!

RobbieMcKinstry:
Sending this back to draft, since I've identified a couple of other options that aren't recognized on components. Going to be a slightly larger code change, so for now this PR has been benched until I'm done adding those.

Frassle:
> Sending this back to draft, since I've identified a couple of other options that aren't recognized on components. Going to be a slightly larger code change, so for now this PR has been benched until I'm done adding those.

Small changes that slightly improve things are very acceptable, in fact probably preferable, to large changes that try to fix everything at once. I think this change is good as is and could be merged. You can then raise more patches to fix more options.

RobbieMcKinstry:
@Frassle Ah, okay! I asked around and got a slightly different response, but that's is more inline with what I've been told by our bossman :)
I'll reopen the PR and reassign you for review. Once I get the hammer stamp, I'll merge and move my local commits to a new branch, then rebase. 👍🏻

pkg/resource/deploy/source_eval.go
```diff
@@ -470,6 +470,7 @@ func (d *defaultProviders) getDefaultProviderRef(req providers.ProviderRequest)
 // resmon implements the pulumirpc.ResourceMonitor interface and acts as the gateway between a language runtime's
 // evaluation of a program and the internal resource planning and deployment logic.
 type resmon struct {
+	ctx                       *plugin.Context                    // the plugin context
```
Frassle, line 473: Dunno if you need Context for later changes but if not maybe just pass the diag.Sink object instead.

Frassle (approved):

PR #9872, opened 2022-06-15
add support for local policy packs in auto api

### LINKED ISSUE #6757: Allow local policy packs to be run with Automation API
At the moment only cross-organisation policy packs can be run with automation API, which requires the use of the Pulumi backend and an enterprise subscription.

Non-Automation API can use the `--policy-pack` option.

jaxxstorm:
# Description

Adds support for policy packs in automation API so we can address https://github.com/pulumi/actions/issues/212

Fixes #6757

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,5 +1,8 @@
 ### Improvements
 
+- [auto/*] Add `--policy-pack` and `--policy-pack-configs` options to automation API.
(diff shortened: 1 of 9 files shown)


AaronFriel (approved):
LGTM, the API is a bit weird here in terms of requiring configs to be either empty or the same length as the list of packs, but that's exactly what the CLI expects.

PR #9873, opened 2022-06-15
Remove the go mod replace directives

### LINKED ISSUE #9529: Investigate why packages needed to be pinned
While launching new languages, I had to pin `hashicorp/hcl/v2` and `sergi/go-diff` to avoid test failures: https://github.com/pulumi/pulumi/blob/master/pkg/go.mod#L7-L9

Let's figure out why minor version upgrades break us and get rid of the pinned versions.

iwahbe:
# Description

Fixes #9529 

## Checklist


Frassle:
Codegen tests look very sad.
```
panic: fatal: A failure has occurred: unexpected expression node of type *hclsyntax.ParenthesesExpr 
```

iwahbe:
> Codegen tests look very sad.
> 
> ```
> panic: fatal: A failure has occurred: unexpected expression node of type *hclsyntax.ParenthesesExpr 
> ```

It all looks sad. This PR was more a "run all the tests in the background and see what breaks" PR. Now the question is why is it broken.

Frassle (approved):

PR #9882, opened 2022-06-17
Fix dotnet alias spec

Frassle:
Protobuf string fields can't be set to null. As we're building the Spec object all the fields are optional and we were indicating that with nullness which then tripped the null check in the property setter. This changes the code to just check nullness first and only set the spec property if it's non-null.

### DIFF
--- a/sdk/dotnet/Pulumi/Deployment/Deployment_Prepare.cs
+++ b/sdk/dotnet/Pulumi/Deployment/Deployment_Prepare.cs
@@ -160,10 +160,14 @@ namespace Pulumi
                         }
 
                         var aliasSpec = new Pulumirpc.Alias.Types.Spec();
-                        aliasSpec.Name = await MaybeAwait(aliasVal.Name);
-                        aliasSpec.Type = await MaybeAwait(aliasVal.Type);
-                        aliasSpec.Stack= await MaybeAwait(aliasVal.Stack);
-                        aliasSpec.Project = await MaybeAwait(aliasVal.Project);
+                        var aliasName = await MaybeAwait(aliasVal.Name);
+                        if (aliasName != null) { aliasSpec.Name = aliasName; }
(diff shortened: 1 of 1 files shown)


sdk/dotnet/Pulumi/Deployment/Deployment_Prepare.cs
```diff
@@ -160,10 +160,14 @@ private static void ThrowAliasPropertyConflict(string name)
...
-                        aliasSpec.Stack= await MaybeAwait(aliasVal.Stack);
-                        aliasSpec.Project = await MaybeAwait(aliasVal.Project);
+                        var aliasName = await MaybeAwait(aliasVal.Name);
+                        if (aliasName != null) { aliasSpec.Name = aliasName; }
+                        var aliasType = await MaybeAwait(aliasVal.Type);
+                        if (aliasType != null) { aliasSpec.Type = aliasType; }
+                        var aliasStack= await MaybeAwait(aliasVal.Stack);
+                        if (aliasStack != null) { aliasSpec.Stack = aliasStack; }
+                        var aliasProject = await MaybeAwait(aliasVal.Project);
+                        if (aliasProject != null) { aliasSpec.Project = aliasProject; }
```
mikhailshilkov, line 170: What happens if an old SDK without these changes runs against the new engine?

mikhailshilkov (commented):
Could you please add some description to the PR? I'm not sure about the causes and impact of it. Thank you!

sdk/dotnet/Pulumi/Deployment/Deployment_Prepare.cs
Frassle: It will send the old `repeated string` of urns and the engine will accept that. The engine will in fact accept both the old string urns and new alias specs in the same request and just concatenate them all together.

mikhailshilkov (approved):

PR #9886, opened 2022-06-17
Moving previewDigest to sdk/go/common/display, and exporting it. 

### LINKED ISSUE #9851: Make previewDigest into exported type PreviewDigest
## Hello!

## Issue details

To enable the parsing of the preview digest in other programs it would be nice to be able to reference the previewDigest type directly from other programs instead of having to copy/paste it. Can we export that in future versions. 

### Affected area/feature

pulumi/pkg

rshade:
# Description

This moves previewDigest to sdk/go/common/display, and exports it and dependent types.  

Fixes #9851

## Checklist


Frassle (approved):
Fine by me, just fix up the linter complaints.

iwahbe:
I would prefer a change that did not result in the `sdk` depending on `pkg`. AFAIK the only types that are being used are `engine.ResourceChanges` and `deploy.StepOp`. 
- `engine.ResourceChanges` looks like it can be safely moved into the `sdk` (assuming `StepOp` can be moved). 
- `deploy.StepOp` is trickier, since it cannot be easily moved into the `sdk`. I recommend that we move the `StepOp` type (`type StepOp string`) to the `sdk` and redefine it in `pkg`: `type StepOp sdk.StepOp`. This keeps them closely bound but doesn't have the same lock in effect as exporting from `pkg`.

Frassle:
> I would prefer a change that did not result in the sdk depending on pkg. AFAIK the only types that are being used are engine.ResourceChanges and deploy.StepOp.

Oh I did not notice that. Yes the transitive set of all types needed here all need to move to the SDK. We can't have the sdk depend on pkg.

> deploy.StepOp is trickier, since it cannot be easily moved into the sdk

Why? It doesn't look like it depends on much else.

rshade:
So `pkg` can depend on `sdk` but not vice versa, or circular? Sorry, just making sure before I updating it.

iwahbe:
>>deploy.StepOp is trickier, since it cannot be easily moved into the sdk
>
> Why? It doesn't look like it depends on much else.

I thought that `color` (which `StepOp` depends on via methods) was in `pkg`, but it is in `sdk`. In that case I think we can move `StepOp` to the `sdk`. Sorry for the confusion.

> So pkg can depend on sdk but not vice versa, or circular? Sorry, just making sure before I updating it.

`pkg` can and does depend on `sdk`. `sdk` does not (and should not) depend on `pkg`.

Frassle (approved):

PR #9909, opened 2022-06-21
Set DeleteBeforeReplace for forced replaces

Frassle:
# Description

This sets DeleteBeforeReplace to true when doing a targeted replace, or a replace due to a provider change.

This is so that if doing a replace where we aren't changing a unique field (such as the name) we can't safely create the new resource until we've delete the old. We don't currently have an interface to ask the provider "given this old state and new state, assuming we do a replace is it safe to do a CreateBeforeDelete?".

Part of fixing https://github.com/pulumi/pulumi-aws/issues/2009

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -24,6 +24,9 @@
 - [sdk/dotnet] The dotnet SDK will now send alias objects rather than URNs to the engine.
   [#9731](https://github.com/pulumi/pulumi/pull/9731)
 
+- [engine] The engine will default to delete before create for replacements when it has no information from the provider that create before delete is safe.
+  [#9909](https://github.com/pulumi/pulumi/pull/9909)
+
 ### Bug Fixes
 
 - [sdk/go] Handle nils in mapper encoding.
--- a/pkg/resource/deploy/step_generator.go
+++ b/pkg/resource/deploy/step_generator.go
@@ -1462,7 +1462,7 @@ func (sg *stepGenerator) diff(urn resource.URN, old, new *resource.State, oldInp
 	newInputs resource.PropertyMap, prov plugin.Provider, allowUnknowns bool,
 	ignoreChanges []string) (plugin.DiffResult, error) {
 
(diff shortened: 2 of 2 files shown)


mikhailshilkov:
> This is so that if doing a replace where we aren't changing a unique field (such as the name) we can't safely create the new resource until we've delete the old.

This sounds surprising - any auto-named resource should be fine with create-before-delete in this situation? Does this PR regress the experience for all auto-named resources?

Frassle:
> This sounds surprising - any auto-named resource should be fine with create-before-delete in this situation? Does this PR regress the experience for all auto-named resources?

Yes because the engine can't tell which resources are auto-named and which are explicitly named. So we have the choice of:
A) Give the best experience for autonamed resources but error in create on explicitly named resources.
B) Treat all resources the same and delete them before creating their replacements.

mikhailshilkov:
I feel like this choice should be the provider's responsibility. I believe we need to describe in more details why the existing mechanism for this is not enough and proposals to enable the providers do so.

Frassle:
> The workaround for the user is to opt into [deleteBeforeReplace](https://www.pulumi.com/docs/intro/concepts/resources/options/deletebeforereplace/) for resources that need that?

Yes we could choose to not take this PR and just tell users to set deleteBeforeReplace manually in these cases.

iwahbe:
We should investigate the effect of setting `deleteBeforeReplace = false`. Letting the user explicitly opt out of `deleteBeforeReplace` would make me much more comfortable with this change. I suspect that the engine takes `deleteBeforeReplace` as a binary, and we would need to change it to a trinary (`true`, `false`, `unset`).

Frassle:
While this behavior isn't ideal, the team have decided that as long as we're taking https://github.com/pulumi/pulumi/pull/9903 in some form the worst this can cause is an error during create that can be worked around by setting DeleteBeforeReplace. We'll work with the providers team to fix this in https://github.com/pulumi/pulumi/issues/9924.

PR #9919, opened 2022-06-21
Typescript TF 0.12 module support

T-Vova:
Fully implementing Terraform 0.12 Module conversion for all languages so that support for Terraform 0.13 and beyond can be added (#186).


justinvp:
Closing as no longer relevant. Support for this was added in https://github.com/pulumi/pulumi/pull/12398 and various other similar changes.

PR #9928, opened 2022-06-22
[auto] add verbose and logtostderr options

jaxxstorm:
This is a WIP to add the `--logtostderr` and `--verbose` options to automation API

We potentially need to flow the stderr events through to the engine events as well, but I'll figure that out later


sdk/python/lib/pulumi/automation/_stack.py
```diff
@@ -211,6 +211,8 @@ def up(
         on_output: Optional[OnOutput] = None,
         on_event: Optional[OnEvent] = None,
         program: Optional[PulumiFn] = None,
+        verbose: Optional[int] = None,
+        log_to_stderr: Optional[bool] = None,
```
iwahbe, line 215: Does this need to be `Optional[bool]`, can it not just be a `bool` with default `False`>

iwahbe (commented):
This looks good, is it still a WIP?

PR #9933, opened 2022-06-22
Add external enums all languages

### LINKED ISSUE #6236: [codegen/(all languages)] Support referencing external enums
The codegen for each language now supports referencing external resources and types, but referencing external enums needs the same treatment. This will look very similar to how external types are referenced.

iwahbe:
# Description

Fixes #6236

When we are referencing an inputty type (like `FooPtrInput`)  from an external package in go, we need to know if that input type was generated. For this, we need the `.pkg` backlink of the external type to be accurate. This PR relies includes https://github.com/pulumi/pulumi/pull/9936 to facilitate keeping external schema information.

## Checklist


pkg/codegen/go/gen.go
```diff
@@ -676,13 +699,12 @@ func (pkg *pkgContext) contextForExternalReference(t schema.Type) *pkgContext {
...
 	}
 
-	extPkgCtx := &pkgContext{
-		pkg:              extPkg,
-		importBasePath:   goInfo.ImportBasePath,
-		pkgImportAliases: pkgImportAliases,
-		modToPkg:         goInfo.ModuleToPackage,
-	}
-	return extPkgCtx
+	maps := generatePackageContextMap(pkg.tool, extPkg, goInfo)
```
iwahbe, line 702: I'm concerned that this will have performance implications. I'll cache the result, but unfortunately global analysis is necessary for correctness.

pkg/codegen/testing/test/testdata/external-resource-schema/go/example/workload.go
```diff
@@ -149,8 +149,8 @@ func (o WorkloadOutput) ToWorkloadOutputWithContext(ctx context.Context) Workloa
 	return o
 }
 
-func (o WorkloadOutput) Pod() corev1.PodPtrOutput {
-	return o.ApplyT(func(v *Workload) corev1.PodPtrOutput { return v.Pod }).(corev1.PodPtrOutput)
+func (o WorkloadOutput) Pod() corev1.PodTypePtrOutput {
```
AaronFriel, line 152: I don't see it, but are we now outputting this as a type on sdkgen with this PR? I don't see `PodTypePtrOutput` in the exported types of `kubernetes/core/v1`:

https://pkg.go.dev/github.com/pulumi/pulumi-kubernetes/sdk/v3/go/kubernetes/core/v1

AaronFriel (commented):
Code changes look good, but I'm unsure of the type change in the comment below. 

I'm not worried about the performance cost right now, I expect that if we've already loaded the other schema, the bulk of the expensive operations have been performed and generating a new map on each external reference (even one with ~10k types) isn't an issue. We can memoize later?

pkg/codegen/testing/test/testdata/external-resource-schema/go/example/workload.go
iwahbe: Yeah, that type doesn't exist. Neither `corev1.PodPtrOutput` or `corev1.PodTypePtrOutput` exist (this test doesn't compile). The appending of type is correct. Kubernetes `core/v1` has both a `Pod` defined in both [resources](https://github.com/pulumi/pulumi-kubernetes/blob/7547fb6ae397a8592c20ea9b5ddbdaa8605ecc33/provider/cmd/pulumi-resource-kubernetes/schema.json#L38531) and [types](https://github.com/pulumi/pulumi-kubernetes/blob/7547fb6ae397a8592c20ea9b5ddbdaa8605ecc33/provider/cmd/pulumi-resource-kubernetes/schema.json#L16662), and workload references the type:
https://github.com/pulumi/pulumi/blob/0e64874da720de95c2ae02761d8aced498765cee/pkg/codegen/testing/test/testdata/external-resource-schema/schema.json#L62-L69

We are correctly appending the type because we are generating the external "Pod" type with the knowledge that another "Pod" from resources exists.

iwahbe: It does point out that this fix won't operate correctly for output types (which is why we are not generating `*corev1.PodTypeOutput`). I'll add that fix.

pkg/codegen/go/gen.go
t0yv0: Yeah it sounds like it should be not that hard to cache the result at least inside `pkgInfo`, from skimming I didn't see how exactly you did it, but caching the global anaylsis is good. 

You can try to regen the azure-native with these changes and se how much longer it took (actually downstream checks may have done that for us) so that's a data point on perf.

Expecting it to be quite acceptable.

t0yv0 (approved):
A bit more involved case analysis but looking correct to me.

Logging some very helpful out-of-band comments from Ian here:

```
We look at the external schema (the one that generates the package for Foo) and run the generator pre-processor which determines what types are generated.
```

This is good. So we are improving on not generating dangling references. 

However backwards-compatibiltiy remains a bit fragile, there are unexpected constraints. To cite Ian's example:

```
Adding a new resource that references an old type in an optional way is breaking if if the type was not previously referenced in an optional way.  That is confusing, so here is a concrete example:
Given a schema:
resources:
  fooConfigurator:
    inputProperties:
      input:
        $ref: #/types/foo
        required: true
types:
  foo:
    prop1:
      type: string
    prop2:
      type: string
Adding the new resource:
resources:
  takesOptionalFoo:
    input:
      $ref: #/types/foo
is a breaking change to downstream packages.
If there is a downstream package fooConsumer with schema:
resources:
  fooConsumer:
    inputProperties:
      input:
        $ref: foo#/types/foo
Before the schema of foo is updated, fooConsumer will generate the following go
type FooConsumerArgs struct {
  Input *foo.FooInput
}
but after the change (and consuming the new schema) fooConsumer will generate
type FooConsumerArgs struct {
  Input foo.FooPtrInput
}
This changes the public interface of fooConsumer in a breaking way.
```

We do not have a solution to the fundamental problem here, namely that we do global schema analysis to try to avoid generating unused helper types, but we miss that downstream packages may decide to use them; and therefore global changes to the schema may result in breaking changes in the downstream packages. But the PR is no worse than before so we think we should take it.

Issue #9932, opened 2022-06-22
Bound schema types hold inaccurate package data


iwahbe:
### What happened?

Bound schema types such as `package.schema` have an internal reference to their host package. This reference can be inaccurate.

We supply a `*schema.Package` for partially bound packages. It does not become fully accurate until `schema.PackageReference.Definition()` is called. During codegen, we start with a promptly bound `*schema.Package`, but external `"$ref"` types are lazily bound. We have seen bugs introduced when relying on global analysis runs for external packages (necessary for naming when accounting for duplicates). Removing the `*schema.Package` type prevents these types of bugs by enforcing that data is bound before it is observed.

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

AaronFriel:
@iwahbe did we close this out in #9936, or what remains to do here?

iwahbe:
To complete the issue (and remove the footgun) we should remove all references to the `*Package` fields, and then remove those fields.

iwahbe:
Updating pulumi-java will require the utility functions defined in https://github.com/pulumi/pulumi/pull/11578. We will need to wait for the next release (> 3.48.0) to update pulumi-java.

PR #9936, opened 2022-06-22
Include pkg reference in bound schema types

iwahbe:
# Description

We include a `PackageReference` which is guaranteed to be correct with schema types.

Helps with #9932

## Checklist


AaronFriel (approved):
The lighter weight, on-demand Package References are a nice workaround here.

PR #9943, opened 2022-06-23
Avoid realpath Makefile func that seems flaky on Win

### LINKED ISSUE #9942: Flaky Windows tests
### What happened?

Periodically:

https://github.com/pulumi/pulumi/runs/7023093966?check_suite_focus=true#step:35:21

```
C:\hostedtoolcache\windows\Python\3.9.13\x64\python.exe: can't open file 'D:\a\pulumi\pulumi\pkg\apulumipulumi\scripts\go-test.py': [Errno 2] No such file or directory
mingw32-make: *** [makefile:121: test_pkg_nodejs] Error 2
```

### Steps to reproduce

-

### Expected Behavior

-

### Actual Behavior

-

### Versions used

_No response_

### Additional context

_No response_

### Contributing

t0yv0:
# Description

Fixes #9942

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 PROJECT_NAME := Pulumi SDK
-PROJECT_ROOT := $(realpath .)
+PROJECT_ROOT := .
 SDKS         := dotnet nodejs python go
 SUB_PROJECTS := $(SDKS:%=sdk/%)
 
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 PROJECT_NAME := Pulumi SDK
-PROJECT_ROOT := .
+PROJECT_ROOT := ..
 SDKS         := dotnet nodejs python go
 SUB_PROJECTS := $(SDKS:%=sdk/%)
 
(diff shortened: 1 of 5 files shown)


RobbieMcKinstry (commented):
Out of curiousity, what was the purpose of using `realpath` in the first place? It looks from this PR that it can just be ignored.

Makefile
```diff
@@ -1,5 +1,5 @@
 PROJECT_NAME := Pulumi SDK
-PROJECT_ROOT := $(realpath .)
+PROJECT_ROOT := ..
```
iwahbe, line 2: I'm surprised this works, since `.` is very different then `..`.

dixler (approved):
LGTM

Makefile
t0yv0: This is indeed not pretty. It's just we run `GO_TEST` from subfolder like `pkg`, So relative path has to be `pkg/..`.

With realpath the idea was to use an absolute path instead. But something became flaky on how Windows computes that.

iwahbe: I think we should add a comment or maybe just replace `PROJECT_ROOT` with `..`. As it is, this is actively confusing since `PROJECT_ROOT` is neither stable nor the project root.

Issue #3045, opened 2019-08-07
Deprecate PULUMI_TEST_MODE


joeduffy:
We are going to leverage the techniques demonstrated in https://www.pulumi.com/blog/unit-testing-infrastructure-in-nodejs-and-mocha/ instead.

ericrudder:
Can we annotate the "old" (4.19) blog entry with a link/note to the new one
as well, since it will still come up in searches?

dustinfarris:
Will the new method support unit testing libraries that are not necessarily part of a pulumi program?

joeduffy:
Thanks for the questions Dustin. I know this is an important scenario for your team so I want to make sure whatever we do here will work for you. I also think the way you're testing is a close approximation for what I've heard other end users doing, so your feedback is super valuable.

> Will the new method support unit testing libraries that are not necessarily part of a pulumi program?

Any libraries that don't depend on Pulumi runtime functions -- like allocating resources and reading their resolved output properties -- can still be safely tested using existing mechanisms.

The main change here is that `PULUMI_TEST_MODE` tries to execute in a "half runtime" state, where there is no CLI or engine to communicate with, which is what leads to hangs. We have found that this is quite error prone and confusing. Basically, 50% of the time, your test isn't going to actually succeed at doing what you want it to do.

With the new approach, there is a full CLI and engine available, so everything works just as it would during typical Pulumi evaluation.

> The new method feels close to integration testing to me — which is great — but does it come at the cost of unit testing?

Unit testing is still possible. The idea is to run the unit tests during `pulumi preview`. Do you think that will work for your use cases?

> Also, will there be support for utilities like mocha watch that reruns unit tests as you write them?

This is one of the major losses, I think. The existing `mocha` CLI can't be used. You can, however, use the built-in programmatic capabilities of Mocha, which [covers a lot of ground](https://github.com/mochajs/mocha/wiki/Using-Mocha-programmatically). It appears "watch" is one that wouldn't be possible -- unless @lukehoban sees something here that I'm missing.

https://github.com/pulumi/pulumi/issues/2287 will explore capabilities such as this.

dustinfarris:
Thanks for the reply, Joe.

We see the benefits of unit testing during preview and post-update when the entire shape of resources is known.  We are excited to play with this new feature which should be prioritized soon.

[...]

PR #9959, opened 2022-06-24
Fix out-of-date docs for `runtime.isDryRun` et al

### LINKED ISSUE #9957: The docs for the NodeJS `runtime.isDryRun` include an obsolete reference to “test mode”
### What happened?

I read the docs for this method in the API reference, and they currently read in their entirety:

>  Returns true if we're currently performing a dry-run, or false if this is a true update. *Note that we always consider executions in test mode to be "dry-runs", since we will never actually carry out an update, and therefore certain output properties will never be resolved.*

Emphasis mine.

As I described in detail in [this related PR](https://github.com/pulumi/pulumi-hugo/pull/1642), I didn’t really know what “test mode” was but I assumed (incorrectly) that it was related to unit testing a Pulumi project. I assumed that it was automatically enabled when invoking `setMocks`. And I also assumed that therefore if my project code called this method in the context of my unit tests, the method would return `true`. But that was incorrect.

In fact, as per the discussion in that PR, “test mode” is a vestigial mechanism related to testing the runtime itself, and is not relevant to writing or running unit tests for a Pulumi project.

And as far as I can tell from reading the code, and from running my unit tests, the emphasized sentence in the method description is obsolete, incorrect, and therefore inadvertently misleading. So I think it should be removed.

### Steps to reproduce

Try running unit tests that use `setMocks` and in the project code check the value returned by `isDryRun` — it will, by default, always be `false`.

### Expected Behavior

Based on the description of the method, I expected `isDryRun` to return `true` when running unit tests.

### Actual Behavior

It returned `false`.

### Versions used

```
CLI          
Version      3.35.0
Go Version   go1.17.11
Go Compiler  gc

Plugins
NAME    VERSION
aws     5.6.0
nodejs  unknown

Host     
OS       darwin
Version  12.4
Arch     arm64

This project is written in nodejs: executable='/opt/homebrew/bin/node' version='v18.4.0'

Backend        
Name           pulumi.com
URL            https://app.pulumi.com/aviflat
User           aviflat
Organizations  aviflat, latacora

Dependencies:
NAME            VERSION
shadow-cljs     2.19.1
@pulumi/aws     5.6.0
@pulumi/pulumi  3.33.2

Pulumi locates its logs in /var/folders/pp/l138dt0s605b5t04y1mj98w00000gn/T/ by default

```

### Additional context

I have a fork with a branch with a change that addresses this, I will open a PR for it shortly.

### Contributing

aviflat:
As described in #9957 the removed text is no longer correct, relevant, or useful. It is actively misleading because “test mode” in fact has nothing to do with the modality of preview/dryRun.

See also #3045

Fixes #9957

# Description

Fixes #9957

## Checklist

- [ ] ~I have added tests that prove my fix is effective or that my feature works~ — seems unnecessary

- [ ] ~I have updated the [CHANGELOG-PENDING](https://github.com/pulumi/pulumi/blob/master/CHANGELOG_PENDING.md) file with my change~ — seems unnecessary

- [ ] ~Yes, there are changes in this PR that warrants bumping the Pulumi Service API version~ — nope

### DIFF
--- a/sdk/nodejs/runtime/settings.ts
+++ b/sdk/nodejs/runtime/settings.ts
@@ -113,9 +113,9 @@ export function _setIsDryRun(val: boolean) {
 }
 
 /**
- * Returns true if we're currently performing a dry-run, or false if this is a true update. Note that we
- * always consider executions in test mode to be "dry-runs", since we will never actually carry out an update,
- * and therefore certain output properties will never be resolved.
+ * Returns whether or not we are currently doing a preview.
+ *
+ * When writing unit tests, you can set this flag via either `setMocks` or `_setIsDryRun`.
  */
 export function isDryRun(): boolean {
     return options().dryRun === true;
@@ -138,6 +138,10 @@ export function _setFeatureSupport(key: string, val: boolean) {
 
 /**
  * Returns true if test mode is enabled (PULUMI_TEST_MODE).
+ *
+ * NB: this test mode has nothing to do with preview/dryRun modality, and it is not automatically
+ * enabled by calling `setMocks`. It is a vestigial mechanism related to testing the runtime itself,
+ * and is not relevant to writing or running unit tests for a Pulumi project.
  */
 export function isTestModeEnabled(): boolean {
     return options().testModeEnabled === true;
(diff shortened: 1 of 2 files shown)


dixler (approved):
thanks for the contribution

aviflat:
My pleasure!

PR #9965, opened 2022-06-24
Automate release chores

t0yv0:
# Description

This extends `scripts/release.sh` to partially automate our entire release sequence. I am always a little nervous messing things up doing it manually. And perhaps most importantly, taking 3 PRs to do the release is a lot.

The manual sequence is documented in https://github.com/pulumi/home/wiki/Producing-the-Pulumi-CLI-and-SDKs

With this helper script, the idea is to reorder merging and tagging so that only 1 PR is needed. I believe this should work. Three commits are created, and should me merged without squashing on top of master. Only after that all the tags are added. Once the tags are added, release.yml automation kicks off from the tags.

I have test-driven it on a t0yv0/pulumi fork.

As an alternative we could also tag the commits before they land on master. This is a bit dangerous if there is any history rewriting during PR merge.

Fixes # (issue)

## Checklist

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -2,34 +2,95 @@
 
 set -euo pipefail
 
-VERSION="@1"
-
-die()
-{
-    echo "$1"
-    exit 1
-}
-
 usage()
 {
-    echo "Usage: ./scripts/release.sh update-sdk-ref VERSION"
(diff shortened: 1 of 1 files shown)


Frassle (approved):
b-e-a-utiful ❤️

PR #9968, opened 2022-06-27
Refine the description of destroy command to be more accurate

### LINKED ISSUE #9901: pulumi cli description for destroy is wrong
pulumi cli description for destroy is wrong
[internal slack convo](https://pulumi.slack.com/archives/C85BS3LJZ/p1655230372394299)

file to fix up: https://github.com/pulumi/pulumi/blob/master/pkg/cmd/pulumi/destroy.go#L63-L70

mikhailshilkov:
# Description

Fixes #9901

## Checklist

### DIFF
--- a/pkg/cmd/pulumi/destroy.go
+++ b/pkg/cmd/pulumi/destroy.go
@@ -60,12 +60,14 @@ func newDestroyCmd() *cobra.Command {
 		Use:        "destroy",
 		Aliases:    []string{"down"},
 		SuggestFor: []string{"delete", "kill", "remove", "rm", "stop"},
-		Short:      "Destroy an existing stack and its resources",
-		Long: "Destroy an existing stack and its resources\n" +
+		Short:      "Destroy all the existing resources in the stack",
+		Long: "Destroy all the existing resources in the stack\n" +
 			"\n" +
-			"This command deletes an entire existing stack by name.  The current state is\n" +
+			"This command deletes all the resources in the selected stack.  The current state is\n" +
 			"loaded from the associated state file in the workspace.  After running to completion,\n" +
 			"all of this stack's resources and associated state will be gone.\n" +
+			"\n" +
+			"The stack itself is not deleted. Use `pulumi stack rm` to delete the stack.\n" +
 			"\n" +
 			"Warning: this command is generally irreversible and should be used with great care.",
 		Args: cmdutil.NoArgs,
--- a/pkg/cmd/pulumi/destroy.go
+++ b/pkg/cmd/pulumi/destroy.go
@@ -60,12 +60,12 @@ func newDestroyCmd() *cobra.Command {
 		Use:        "destroy",
 		Aliases:    []string{"down"},
 		SuggestFor: []string{"delete", "kill", "remove", "rm", "stop"},
-		Short:      "Destroy all the existing resources in the stack",
-		Long: "Destroy all the existing resources in the stack\n" +
+		Short:      "Destroy all existing resources in the stack",
+		Long: "Destroy all existing resources in the stack, but not the stack itself\n" +
 			"\n" +
-			"This command deletes all the resources in the selected stack.  The current state is\n" +
+			"Deletes all the resources in the selected stack.  The current state is\n" +
 			"loaded from the associated state file in the workspace.  After running to completion,\n" +
-			"all of this stack's resources and associated state will be gone.\n" +
+			"all of this stack's resources and associated state are deleted.\n" +
 			"\n" +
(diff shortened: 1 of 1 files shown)


pkg/cmd/pulumi/destroy.go
```diff
@@ -60,13 +60,15 @@ func newDestroyCmd() *cobra.Command {
...
 		SuggestFor: []string{"delete", "kill", "remove", "rm", "stop"},
-		Short:      "Destroy an existing stack and its resources",
-		Long: "Destroy an existing stack and its resources\n" +
+		Short:      "Destroy all the existing resources in the stack",
+		Long: "Destroy all the existing resources in the stack\n" +
 			"\n" +
-			"This command deletes an entire existing stack by name.  The current state is\n" +
+			"This command deletes all the resources in the selected stack.  The current state is\n" +
 			"loaded from the associated state file in the workspace.  After running to completion,\n" +
 			"all of this stack's resources and associated state will be gone.\n" +
```
Frassle, line 68: Wonder if this should be rephrased "all ... associated state" sounds like all the state for the stack, but it's not its just all the state for the resources.

pkg/cmd/pulumi/destroy.go
```diff
@@ -60,13 +60,15 @@ func newDestroyCmd() *cobra.Command {
 		Use:        "destroy",
 		Aliases:    []string{"down"},
 		SuggestFor: []string{"delete", "kill", "remove", "rm", "stop"},
-		Short:      "Destroy an existing stack and its resources",
-		Long: "Destroy an existing stack and its resources\n" +
+		Short:      "Destroy all the existing resources in the stack",
+		Long: "Destroy all the existing resources in the stack\n" +
 			"\n" +
-			"This command deletes an entire existing stack by name.  The current state is\n" +
+			"This command deletes all the resources in the selected stack.  The current state is\n" +
```
susanev, line 66:
```suggestion
			"Deletes all the resources in the selected stack.  The current state is\n" +
```

pkg/cmd/pulumi/destroy.go
susanev, line 68:
```suggestion
			"all of this stack's resources and associated state are deleted.\n" +
```

pkg/cmd/pulumi/destroy.go
```diff
@@ -60,13 +60,15 @@ func newDestroyCmd() *cobra.Command {
 		Use:        "destroy",
 		Aliases:    []string{"down"},
 		SuggestFor: []string{"delete", "kill", "remove", "rm", "stop"},
-		Short:      "Destroy an existing stack and its resources",
-		Long: "Destroy an existing stack and its resources\n" +
+		Short:      "Destroy all the existing resources in the stack",
+		Long: "Destroy all the existing resources in the stack\n" +
```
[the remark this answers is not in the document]
susanev: thoughts on adding this at the top as well?

susanev (approved):

pkg/cmd/pulumi/destroy.go
```diff
@@ -60,13 +60,15 @@ func newDestroyCmd() *cobra.Command {
...
 			"\n" +
-			"This command deletes an entire existing stack by name.  The current state is\n" +
+			"This command deletes all the resources in the selected stack.  The current state is\n" +
 			"loaded from the associated state file in the workspace.  After running to completion,\n" +
 			"all of this stack's resources and associated state will be gone.\n" +
 			"\n" +
+			"The stack itself is not deleted. Use `pulumi stack rm` to delete the stack.\n" +
+			"\n" +
 			"Warning: this command is generally irreversible and should be used with great care.",
```
susanev, line 72: what does generally irreversible mean? if it means irreversible we should say only that and not generally. if there's some exception we should say what the exception is.

pkg/cmd/pulumi/destroy.go
mikhailshilkov: Not sure I get this - what's your suggested wording?

pkg/cmd/pulumi/destroy.go
mikhailshilkov: I guess it means that some resources may be recreated with no harm, while others may be lost forever (e.g. data in a database).

pkg/cmd/pulumi/destroy.go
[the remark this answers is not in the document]
mikhailshilkov: I feel like it's better as it is...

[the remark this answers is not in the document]
susanev: Okay, let's keep it as is then. If more folks report confusion though we should consider raising the prominence of this detail.

PR #9970, opened 2022-06-27
Improve wording for the stack removal error when stack still has resources

### LINKED ISSUE #9166: Update console output for `pulumi stack rm`
Stack deletion does not actually delete resources and requires the user to execute a `pulumi destroy` first.

However, attempting to delete a stack with resources still attached to it yields this message:

```
casey@caseys-mbp aws-go % pulumi stack rm tearmedown                  
This will permanently remove the 'tearmedown' stack!
Please confirm that this is what you'd like to do by typing ("tearmedown"): tearmedown
error: 'tearmedown' still has resources; removal rejected; pass --force to override
```

Which reads to me (as a person who tries first and regrets later), meant that I should force delete the stack... before I realized the sins of my ways. Suggestion to update wording to encourage user to run `destroy` before suggesting a `--force` flag, which would orphan those resources.

mikhailshilkov:
# Description

Fixes #9166, please make sure the wording makes sense to you

Example of an output:

```
This will permanently remove the 'dev' stack!
Please confirm that this is what you'd like to do by typing ("dev"): dev
error: 'dev' still has resources; removal rejected. Possible actions:
- Make sure that 'dev' is the stack that you want to destroy
- Run `pulumi destroy` to delete the resources, then run `pulumi stack rm`
- Run `pulumi stack rm --force` to override this error
```

## Checklist

### DIFF
--- a/pkg/cmd/pulumi/stack_rm.go
+++ b/pkg/cmd/pulumi/stack_rm.go
@@ -75,10 +75,10 @@ func newStackRmCmd() *cobra.Command {
 			if err != nil {
 				if hasResources {
 					return result.Errorf(
-						"'%s' still has resources; removal rejected. Possible actions:\n" +
-						"- Make sure the correct stack is selected\n" +
-						"- Run `pulumi destroy` to delete the resources, then run `pulumi stack rm` again\n" +
-						"- Run `pulumi stack rm --force` to override this error and remove the stack anyway", s.Ref())
+						"'%s' still has resources; removal rejected. Possible actions:\n"+
+							"- Make sure the correct stack is selected\n"+
+							"- Run `pulumi destroy` to delete the resources, then run `pulumi stack rm` again\n"+
+							"- Run `pulumi stack rm --force` to override this error and remove the stack anyway", s.Ref())
 				}
 				return result.FromError(err)
 			}
--- a/pkg/cmd/pulumi/stack_rm.go
+++ b/pkg/cmd/pulumi/stack_rm.go
@@ -75,7 +75,10 @@ func newStackRmCmd() *cobra.Command {
 			if err != nil {
 				if hasResources {
 					return result.Errorf(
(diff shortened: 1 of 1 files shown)


pkg/cmd/pulumi/stack_rm.go
```diff
@@ -75,7 +75,10 @@ func newStackRmCmd() *cobra.Command {
 			if err != nil {
 				if hasResources {
 					return result.Errorf(
-						"'%s' still has resources; removal rejected; pass --force to override", s.Ref())
+						"'%s' still has resources; removal rejected. Possible actions:\n"+
+							"- Make sure the correct stack is selected\n"+
```
susanev, line 79: for this line, can we state the name of the stack they currently have selected?

pkg/cmd/pulumi/stack_rm.go
```diff
@@ -75,7 +75,10 @@ func newStackRmCmd() *cobra.Command {
...
+							"- Run `pulumi destroy` to delete the resources, then run `pulumi stack rm` again\n"+
```
susanev, line 80:
```suggestion
							"- Run `pulumi destroy` to delete the resources, then run `pulumi stack rm`\n"+
```

pkg/cmd/pulumi/stack_rm.go
```diff
@@ -75,7 +75,10 @@ func newStackRmCmd() *cobra.Command {
...
+							"- Run `pulumi stack rm --force` to override this error and remove the stack anyway", s.Ref())
```
susanev, line 81:
```suggestion
							"- Run `pulumi stack rm --force` to override this error", s.Ref())
```

susanev (approved):

pkg/cmd/pulumi/stack_rm.go
mikhailshilkov: Thank you! I added the stack name and reworded. Not sure I love the result - do you have a suggestion to improve further?

PR #9972, opened 2022-06-27
[cli] Do not require --yes if --skip-preview is set

### LINKED ISSUE #4569: `--yes` shouldn't be required for commands that won't have any `yes/no` prompts

mikhailshilkov:
# Description

`--skip-preview` means there's nothing to confirm for up/destroy/refresh/import, so we shouldn't be asking for `--yes` in non-interactive mode.

Fixes #4569

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -3,6 +3,9 @@
 - [sdk/go] Added `PreviewDigest` for third party tools to be able to ingest the preview json
   [#9886](https://github.com/pulumi/pulumi/pull/9886)
 
+- [cli] Do not require the `--yes` option if the `--skip-preview` option is set.
+  [#9972](https://github.com/pulumi/pulumi/pull/9972)
+
 ### Bug Fixes
 
 - [engine] Filter out non-targeted resources much earlier in the engine cycle.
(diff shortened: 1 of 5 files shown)


dixler:
I would suggest against this since in my head --yes is so heavily associated with committing to doing an update rather than --skip-preview. --skip-preview just means that you want to skip a preview. I was initially thinking we could automatically --skip-preview when --yes is given, but I'd have to look at what kind of info we give the user in non-interactive mode since time estimates and plans make users more patient. Thought I'd write down these initial thoughts before going further.

mikhailshilkov:
> I would suggest against this since in my head --yes is so heavily associated with committing to doing an update rather than --skip-preview. --skip-preview just means that you want to skip a preview. 

But if you run `pulumi up --skip-preview` in interactive mode with the current CLI, you will get the update done without any further questions. Why should non-interactive mode fail then?

dixler (approved):

PR #9974, opened 2022-06-27
Mark convert language as a required parameter

### LINKED ISSUE #9888: Running `pulumi convert` without `--language` yields funny error
If I run `pulumi convert`, with no arguments, I get an error:

`error: cannot generate programs for`

I assume `--language` is a required argument. If so, we should either

* Require that flag to be passed
* Turn it into an argument (usually funny to unconditionally require flags)

mikhailshilkov:
# Description

Fixes #9888

## Checklist

### DIFF
--- a/pkg/cmd/pulumi/convert.go
+++ b/pkg/cmd/pulumi/convert.go
@@ -66,7 +66,7 @@ func newConvertCmd() *cobra.Command {
 			case "yaml": // nolint: goconst
 				projectGenerator = yamlgen.GenerateProject
 			default:
-				return result.Errorf("cannot generate programs for %v", language)
+				return result.Errorf("cannot generate programs for %q language", language)
 			}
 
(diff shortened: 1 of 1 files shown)


iwahbe (approved):
This is a breaking change, but won't change the interface of users who are using `pulumi convert` correctly. Is the reason for not making language a positional argument to avoid further breaking?

mikhailshilkov:
@iwahbe I guess I miss the way it's breaking. What is the scenario that used to work that won't work anymore?

iwahbe:
> @iwahbe I guess I miss the way it's breaking. What is the scenario that used to work that won't work anymore?

Something that was optional is now required, so it meets the obvious definition for a breaking API change. Despite that, there is no scenario where it used to work and now doesn't. Sorry for adding confusion with that comment.

AaronFriel (approved):

PR #9980, opened 2022-06-27
Fix --tracing flag propagation to resource plugins

t0yv0:
# Description

Not including the test case yet - but my informal test case is a program obtained from `pulumi new aws-go` invoked as: 

```
pulumi preview --tracing=file:./preview.tace
```

If one looks at how providers are invoked in that example it's like this:

```
EXEC /home/pulumit0yv0/bin/pulumi-language-go [-root=/home/pulumit0yv0/tmp/tracing-example -tracing=tcp://127.0.0.1:41793 127.0.0.1:39649]
EXEC /home/pulumit0yv0/.pulumi/plugins/resource-aws-v5.7.2/pulumi-resource-aws [127.0.0.1:39649]
```

`tcp://127.0.0.1:41793` is the proxy server that appends to `./preview.trace` started by `pulumi` process.

I think the right fix is to ignore TracingToFile var. Once I did that I started having `-tracing` arg populated twice for the pulumi-language-go command so I cleaned up the place where that happens also.

 With the fix both providers get `-tracing=tcp://127.0.0.1:41793` flag but I'm yet to verify that the traces land where they should.

Fixes # (issue)

## Checklist

### DIFF
--- a/sdk/go/common/resource/plugin/langruntime_plugin.go
+++ b/sdk/go/common/resource/plugin/langruntime_plugin.go
@@ -89,10 +89,6 @@ func buildArgsForNewPlugin(host Host, ctx *Context, options map[string]interface
 
 	args = append(args, fmt.Sprintf("-root=%s", filepath.Clean(root)))
 
-	if cmdutil.IsTracingEnabled() {
-		args = append(args, fmt.Sprintf("-tracing=%s", cmdutil.TracingEndpoint))
-	}
-
 	// NOTE: positional argument for the server addresss must come last
 	args = append(args, host.ServerAddr())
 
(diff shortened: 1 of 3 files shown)


viveklak:
Thanks! I will take this for a spin and report back. LGTM though!

viveklak (approved):

t0yv0:
I've tested out the effect of these changes locally:

- changes only affect `--tracing file:./my.trace` path that uses AppDash
- with the changes it starts collecting traces from providers, so more information is present 
- a great deal of un-parented gRPC server Log calls disappear - these are now parented to `pulumi` trace, which is much better for AppDash users
- can confirm that ResourceProvider:Configure for example span has gRPC server and client spans so this is flowing through gRPC boundary

I think we should take this but not bother with automated tests around it for now as the appdash is not strategic and we're steering toward opentelemetry and replacing the file format.

PR #9984, opened 2022-06-28
Add github:// download url

Frassle:
# Description

This will allow pulumiverse plugins (or others) to use the github releases api:
```
% pulumi plugin install resource astra --server github://api.github.com/pulumiverse
[resource plugin astra-1.0.25] installing
Downloading plugin: 15.12 MiB / 15.12 MiB [=========================] 100.00% 5s
```

## Checklist

### DIFF
--- a/sdk/go/common/workspace/plugins.go
+++ b/sdk/go/common/workspace/plugins.go
@@ -202,7 +202,9 @@ func (source *getPulumiSource) Download(
 
 // githubSource can download a plugin from github releases
 type githubSource struct {
+	host         string
 	organization string
+	repository   string
 	name         string
 	kind         PluginKind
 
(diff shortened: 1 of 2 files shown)


iwahbe (approved):
This looks great.

AaronFriel:
Does this also work with the `pluginDownloadUrl` resource option, i.e.: with sdkgen tweaks, could I codegen `github://api.github.com/aaronfriel` to download pulumi-wordle?

Frassle:
> Does this also work with the pluginDownloadUrl resource option, i.e.: with sdkgen tweaks, if the url was github://api.github.com/aaronfriel would it download pulumi-wordle?

Yes it should, the point of this was for @ringods to try doing exactly that for pulumiverse plugins

PR #10005, opened 2022-06-29
More alias tests

Frassle:


pkg/engine/lifecycletest/pulumi_test.go
```diff
@@ -1740,6 +1740,93 @@ func TestAliases(t *testing.T) {
...
+		for j := 0; j < 10; j++ {
+			n2Aliases = append(n2Aliases, resource.URN(
+				fmt.Sprintf("urn:pulumi:test::test::pkgA:index:t1-v%d$pkgA:index:t2-v%d::n1-sub", i, j)))
+
+			n3Aliases = append(n3Aliases, resource.URN(
+				fmt.Sprintf("urn:pulumi:test::test::pkgA:index:t1-v%d$pkgA:index:t2-v%d$pkgA:index:t3::n1-sub-sub", i, j)))
+		}
+	}
+
+	snap = updateProgramWithResource(snap, []Resource{{
```
t0yv0, line 1814: So "nXAliases" here is "really big" but the engine somehow dedups them or uses a more efficient algo? The tests regresses something that got fixed? Would it still pass on a machine with a ton of memory?

t0yv0 (approved):
Looks reasonable, is there a link to the bug and the fix this is testing? Is there a need to cross-test in every language, the fix is joint responsibility of SDKs and engine?

Does Java need a fix or this test?

pkg/engine/lifecycletest/pulumi_test.go
Frassle: Yeh so the engine was (before we reverted the changes) using more than expected amounts of memory trying to duplicate all these aliases as it did the same multiply out that we're doing here in the test. So yes this could pass on a machine with more memory, not sure how best to do resource usage tests in go to always catch that.

Frassle:
> Looks reasonable, is there a link to the bug and the fix this is testing? Is there a need to cross-test in every language, the fix is joint responsibility of SDKs and engine?

The bug was engine side, but yes we need to test all languages to make sure that regardless of how the SDK works out aliases (current or the new way) everything works as expected.

> Does Java need a fix or this test?

Java should have this test.

PR #10012, opened 2022-06-30
Attempt to connect to the bucket while logging in to a local backend

### LINKED ISSUE #8144: login to a non-existent bucket appears successful but isn't
## Hello!

## Issue details

if I `pulumi login s3://$(uuidgen)`, the command will return success but the user is not actually logged in to anything and their next stack operation will report a confusing error.

### Steps to reproduce

1. `pulumi new aws-typescript -y`
2. `pulumi login s3://$(uuidgen)`
2. `pulumi stack init new-stack`

Expected: the user should receive an error that the bucket does not exist or is not useable
Actual: appears successful

#### Actual output

```
% pulumi login s3://$(uuidgen)
Logged in to Camerons-MacBook-Pro.local as clstokes (s3://81028f9c-1e6e-40f3-817a-bbe16011642f)
# 
# 
# 
% pulumi stack init new-stack
error: could not create stack: An IO error occurred while writing the new snapshot file: blob (key ".pulumi/stacks/new-stack.json") (code=NotFound): NoSuchBucket: The specified bucket does not exist
	status code: 404, request id: YF8PG2RG935H5FTK, host id: 4M1ieQ7clVwQH27Y3fJ0edyHBcKATW31rgutNC7aURk3zcOhKY2IEyJFE4YQ2yi9h4XT/1oq0VI=
aws-ts-webserver %
```

mikhailshilkov:
# Description

Attempt to connect to the bucket while logging in to a local backend to make sure that the bucket exists and can be used under current user settings.

Not logged in to AWS:

```
$pulumi login s3://$(uuidgen)               
error: problem logging in: unable to access bucket s3://85765016-FCBB-4F81-B8B5-E0F5C6269026: blob (code=Unknown): InvalidAccessKeyId: The AWS Access Key Id you provided does not exist in our records.
	status code: 403, request id: <redacted>, host id: <redacted>
```

Non-existing S3 bucket:

```
$pulumi login s3://$(uuidgen)
error: problem logging in: bucket s3://90ABFB7B-3C5D-4BAC-A557-23688A0CF9C6 is not accessible
```

Existing bucket:

```
pulumi login s3://api-f3144
Logged in to Bla as mikhailshilkov (s3://api-f3144)
```

Do we have any automated tests for local backends?

Fixes #8144

## Checklist

### DIFF
--- a/pkg/backend/filestate/backend.go
+++ b/pkg/backend/filestate/backend.go
@@ -153,6 +153,14 @@ func New(d diag.Sink, originalURL string) (Backend, error) {
 		}
 	}
 
+	isAcc, err := bucket.IsAccessible(context.TODO())
+	if err != nil {
+		return nil, fmt.Errorf("unable to access bucket %s: %w", u, err)
+	}
+	if !isAcc {
+		return nil, fmt.Errorf("bucket %s is not accessible", u)
+	}
+
 	// Allocate a unique lock ID for this backend instance.
 	lockID, err := uuid.NewV4()
 	if err != nil {
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -3,6 +3,9 @@
 - [plugins] Plugin download urls now support GitHub as a first class url schema. For example "github://api.github.com/pulumiverse".
   [#9984](https://github.com/pulumi/pulumi/pull/9984)
 
+- [backends] When logging in to a file backend, validate that the bucket is accessible.
+  [#10012](https://github.com/pulumi/pulumi/pull/10012)
+
 ### Bug Fixes
 
 - [cli] `pulumi convert` supports provider packages without a version.


mikhailshilkov:
> A few, see pkg/backend/filestate/backend_test.go and pkg/backend/filestate/bucket_test.go but they just test filesystem buckets, not aws.

Cool, I added a simple test with a file backend. The problem is that the test passes even with the code on master. `blob.OpenBucket` seems to be inconsistent around error handling between backends... I opened https://github.com/google/go-cloud/issues/3140 to clarify.

mikhailshilkov:
Alright... the upstream issue is closed with a recommendation to use `IsAccessible` which this PR does.

pkg/backend/filestate/backend_test.go
```diff
@@ -395,3 +395,12 @@ func TestRenameWorks(t *testing.T) {
 	assert.Len(t, history, 1)
 	assert.Equal(t, apitype.DestroyUpdate, history[0].Kind)
 }
+
+func TestLoginToNonExistingFolderFails(t *testing.T) {
```
iwahbe, line 399: Does this test actually test the new behavior? I'm unsure from your PR comments.

mikhailshilkov: No, it doesn't. Only remote file backends are currently problematic - but I wasn't sure if we have any prior art of testing against the S3 backend specifically - I think I wasn't able to spot any.

Frassle: I'm worried this might be showing that we've regressed here? Did this used to work before the IsAccessible call? Might be gocloud would of just mkdir'd the folder when needed?

You've manually tested this with an s3 bucket right? I'm not overly worried about getting an s3 test in place just for this, although we should probably have some sort of integration test for it added at some point.

pkg/backend/filestate/backend.go
```diff
@@ -153,6 +153,14 @@ func New(d diag.Sink, originalURL string) (Backend, error) {
 		}
 	}
 
+	isAcc, err := bucket.IsAccessible(context.TODO())
+	if err != nil {
+		return nil, fmt.Errorf("unable to access bucket %s: %w", u, err)
```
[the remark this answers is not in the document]
Frassle: yeh Ians suggestion is a better message, this err is that we couldn't check if the bucket was accessible (like a network timeout) rather than we got a code back saying it couldn't be accessed.

pkg/backend/filestate/backend_test.go
mikhailshilkov: > I'm worried this might be showing that we've regressed here? 

I'm sorry, I haven't caught this comment. "this" what and where have we regressed? This test is green as far as I can tell, and (somewhat unfortunately) it would be green without my change.

> You've manually tested this with an s3 bucket right?

Yes, I did. Logging into an existing S3 bucket succeeds and logging into a random guid bucket fails.

pkg/backend/filestate/backend.go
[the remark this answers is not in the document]
mikhailshilkov: 👍 I'll change the comment

Frassle (approved):

PR #10013, opened 2022-06-30
Simplify and harden escaping in Python doc strings SDK gen

### LINKED ISSUE #9441: Python: Invalid docstring generated
A pulumi-azuredevops build is failing with what appears to be an invalid docstring: https://github.com/pulumi/pulumi-azuredevops/pull/102

Per @nimbinatus:

> Here: https://github.com/pulumi/pulumi/blob/master/pkg/codegen/python/gen.go#L324 <-- that needs a better special character set. I'd consider using the set from https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals (in the table a bit further down) to be more complete, and perhaps see if the system takes regex (replace n many forward slashes with twice as many)

mikhailshilkov:
# Description

I feel like this is simpler, more universal, and more correct than what we have? Let me know if you see counter-cases. It's the last test case that trips our current escaping and the AzureDevOps provider.

Fixes #9441

## Checklist

### DIFF
--- a/pkg/codegen/python/gen.go
+++ b/pkg/codegen/python/gen.go
@@ -319,9 +319,7 @@ func printComment(w io.Writer, comment string, indent string) {
 	}
 
 	// Known special characters that need escaping.
-	// TODO: Consider replacing the docstring with a raw string which may make sense if this
-	// list grows much further.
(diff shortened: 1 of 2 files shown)


Zaid-Ajaj (approved):
Looks good!

t0yv0 (approved):

t0yv0 (approved):

Issue #9905, opened 2022-06-21
Don't print all of very long stack outputs


Frassle:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

The CLI should default to only printing the start of very long stack outputs and then an ellipses rather than the whole very long output. See https://github.com/pulumi/pulumi/issues/9885 for context.

PR #10021, opened 2022-07-01
Truncate long stack outputs 

aq17:
# Description

Addresses [9905](https://github.com/pulumi/pulumi/issues/9905)

## Checklist


pkg/backend/display/object_diff.go
```diff
@@ -806,7 +805,7 @@ func (p *propertyPrinter) printPrimitivePropertyValue(v resource.PropertyValue)
 			p.printPropertyValue(vv)
 			return
 		}
-		p.write("%q", v.StringValue())
```
AaronFriel, line 809: Because users might depend on scripting `pulumi stack output` CLI commands, would this change affect the behavior of `pulumi stack output --json` or `pulumi stack output kubeconfig` in a program like the following?

```yaml
name: whatever
runtime: yaml
outputs:
  kubeconfig: |
    contexts:
    - context:
        cluster: docker-desktop
        user: docker-desktop
      name: docker-desktop
    current-context: docker-desktop
    kind: Config
    preferences: {}
    users: []
    # other stuff here
```

I'd expect in both cases, I should be able to get the full YAML document.

Frassle: Yes, this should just be for summary views. Tests to check that would be good

pkg/backend/display/object_diff.go
```diff
@@ -1268,3 +1267,14 @@ func (p *propertyPrinter) translateYAMLValue(v interface{}) (interface{}, bool)
...
 	}
 }
+
+// if string exceeds three lines, truncate and add "..."
+func (p *propertyPrinter) truncatePropertyString(propertyString string) string {
+	const contextLines = 3
+
+	lines := strings.Split(propertyString, "\n")
+	if len(lines) > contextLines {
+		return strings.Join(lines[:3], "") + "..."
```
AaronFriel, line 1277: I think you'd want to join them with "\n" after splitting on "\n"?

AaronFriel (commented):
Left a couple questions.

pkg/backend/display/object_diff.go
aq17: `pulumi stack output` should still show the full output, but let's definitely add tests

aq17: @Frassle realized this change only applies to `pulumi up` when the object diff is computed, so added a flag `show-full-output` in case a user wants it on `pulumi up`, but it does not touch `pulumi stack output`

pkg/backend/display/object_diff.go
```diff
@@ -1268,3 +1274,14 @@ func (p *propertyPrinter) translateYAMLValue(v interface{}) (interface{}, bool)
...
 	}
 }
+
+// if string exceeds three lines, truncate and add "..."
+func (p *propertyPrinter) truncatePropertyString(propertyString string) string {
+	const contextLines = 3
+
+	lines := strings.Split(propertyString, "\n")
+	if len(lines) > contextLines {
+		return strings.Join(lines[:3], "\n") + "..."
```
iwahbe, line 1284: I think this would look better on a new line, but it's up to you.
```suggestion
		return strings.Join(lines[:3], "\n") + "\n..."
```

pkg/backend/display/object_diff.go
```diff
@@ -1268,3 +1274,14 @@ func (p *propertyPrinter) translateYAMLValue(v interface{}) (interface{}, bool)
 		return v, true
 	}
 }
+
+// if string exceeds three lines, truncate and add "..."
+func (p *propertyPrinter) truncatePropertyString(propertyString string) string {
```
iwahbe, line 1279: We should truncate on very long lines too. For example, if the output is a giant JSON blob created for computers, it will be 1 line but a very large number of characters. The exact number is debatable, but I would truncate any line with > 150 characters with a `...`.

pkg/backend/display/diff_test.go
```diff
@@ -78,6 +78,7 @@ func testDiffEvents(t *testing.T, path string, accept bool) {
 		ShowReplacementSteps: true,
 		ShowSameResources:    true,
 		ShowReads:            true,
+		ShowFullOutput:       true,
```
iwahbe, line 81: This is a fine standard, but we should make sure there are tests for where this is false.

iwahbe (changes requested):
Does the diff code get called when a new stack is stood up? We still want long outputs to be truncated when we don't have a previous output.

pkg/backend/display/object_diff.go
aq17: Great point

aq17:
> Does the diff code get called when a new stack is stood up? We still want long outputs to be truncated when we don't have a previous output.

Ah yeah it does not, also not yet when comparing a long update

aq17:
> > Does the diff code get called when a new stack is stood up? We still want long outputs to be truncated when we don't have a previous output.
> 
> Yes it does when a stack is stood up and destroyed, but not yet when comparing a long update

Truncating when comparing a long update breaks all the tests 😢

pkg/backend/display/diff_test.go
aq17: @iwahbe @AaronFriel I've verified the following:

- diff code gets called when a new stack is initialized and destroyed
- diff code gets called when `pulumi up` is run and there is a new long output, a deleted long output, or a modified long output

I'm not sure why the tests are now failing; `ceb79f2` passed, but that printed the output value as is if there were fewer than three lines, even if one of the lines was > 150 chars. I also specified `true` for `ShowFullOutput` (line 81), meaning `truncatePropertyString` shouldn't be called, so the existing tests shouldn't be impacted

aq17:
Added test case files for truncated outputs @iwahbe

pkg/backend/display/object_diff.go
```diff
@@ -1268,3 +1282,31 @@ func (p *propertyPrinter) translateYAMLValue(v interface{}) (interface{}, bool)
...
+	for i := 0; i < numLines; i++ {
+		if len(lines[i]) > maxLineLength {
+			lines[i] = lines[i][:maxLineLength] + "..."
+		}
+	}
+
+	if numLines == 1 {
+		return lines[0]
+	} else if numLines <= contextLines {
+		strings.Join(lines[:numLines], "\n")
```
iwahbe, line 1308: You already set `numLines = contextLines` if `numLines > contextLines`. This branch will always be taken.
```suggestion
	} else if len(lines) <= contextLines {
		return strings.Join(lines, "\n")
```

pkg/backend/display/diff_test.go
```diff
@@ -117,7 +119,11 @@ func TestDiffEvents(t *testing.T) {
 		t.Run(entry.Name(), func(t *testing.T) {
 			t.Parallel()
 
-			testDiffEvents(t, path, accept)
+			if strings.Contains(entry.Name(), "truncate") {
```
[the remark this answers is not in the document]
aq17: Subdirectories under `testdata`, i.e. `testdata/truncate` and `testdata/not-truncate` ?

aq17:
> We should make sure that there are tests where `truncateOutput = true` but nothing is truncated. A logic error in `object_diff.go` was ensuring that `"\n..."` was added to anything with more then one line, even when it wasn't truncating.

Actually, our test case already covers this – an example is the `urn` output as `"urn":"urn:pulumi:dev::eks::pulumi:pulumi:Stack::eks-dev"`, which is 1 line and <150 characters

pkg/backend/display/object_diff.go
```diff
@@ -1268,3 +1281,31 @@ func (p *propertyPrinter) translateYAMLValue(v interface{}) (interface{}, bool)
...
+		if len(lines[i]) > maxLineLength {
+			lines[i] = lines[i][:maxLineLength] + "..."
+		}
+	}
+
+	if numLines == 1 {
+		return lines[0]
+	} else if len(lines) <= contextLines {
+		return strings.Join(lines, "\n")
+	}
```
iwahbe, line 1308: Joining a single line just returns it. 
```suggestion
	if len(lines) <= contextLines {
 		return strings.Join(lines, "\n")
 	}
```

iwahbe:
> > We should make sure that there are tests where `truncateOutput = true` but nothing is truncated. A logic error in `object_diff.go` was ensuring that `"\n..."` was added to anything with more then one line, even when it wasn't truncating.
> 
> Actually, our test case already covers this – an example is the `urn` output as `"urn":"urn:pulumi:dev::eks::pulumi:pulumi:Stack::eks-dev"`, which is 1 line and <150 characters

The error I discussed would only occur if it was more then one line.

aq17:
> > > We should make sure that there are tests where `truncateOutput = true` but nothing is truncated. A logic error in `object_diff.go` was ensuring that `"\n..."` was added to anything with more then one line, even when it wasn't truncating.
> > 
> > 
> > Actually, our test case already covers this – an example is the `urn` output as `"urn":"urn:pulumi:dev::eks::pulumi:pulumi:Stack::eks-dev"`, which is 1 line and <150 characters
> 
> The error I discussed would only occur if it was more then one line.

Ah I see what you mean

iwahbe (approved):
LGTM

aq17:
> LGTM

Thanks for all the feedback 🙌🏼

Issue #9892, opened 2022-06-18
`pulumi convert` help text is wrong


joeduffy:
The help text for `pulumi convert` mentions PCL:

```
Convert resource declarations into a pulumi program.

The PCL program to convert should be supplied on stdin.
```

I think it should say YAML and probably describe the command in a bit more depth.

It also says the program is supplied on `stdin`, which I don't think is true. Instead, it uses the current directory's project. Maybe it's possible to pass it in `stdin` too? I didn't try it, but if so, this deserves a bit more articulation too.

Two other things don't look quite right:

```
Flags:
  -d, --description string   The project description; if not specified, a prompt will request it
  -n, --name string          The project name; if not specified, a prompt will request it
```

The is no prompting for these, they are just taken from the current project.

Zaid-Ajaj:
Fixed in #10036

PR #10036, opened 2022-07-01
Update `pulumi convert`  help text

aq17:
# Description

Addresses #9892 

Fixes # (issue)

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -2,3 +2,5 @@
 
 ### Bug Fixes
 
+- [cli] `pulumi convert` help text is wrong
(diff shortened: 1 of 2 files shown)


stack72 (approved):
Thanks @aq17 - this is way more useful

Just to check that the removals are not breaking changes though??

aq17:
> Thanks @aq17 - this is way more useful
> 
> Just to check that the removals are not breaking changes though??

I don't believe so, was able to build successfully and checked that [those flags aren't actually used in t2fpulumi](https://github.com/pulumi/kube2pulumi/blob/master/cmd/kube2pulumi/all/cli.go#L19)

PR #10037, opened 2022-07-01
Fix orphan bulk decrypt traces

t0yv0:
# Description

Currently opentracing spans generated by the HTTP client when calling BulkDecrypt endpoint are not parented correctly. This seems to be an issue with not propagating `context.Context` through the code correctly. I took a stab at propagating it and I can see the spans snapping back into place.

Since we're changing interfaces under `sdk/` this is a breaking change. It can be fixed by injecting `context.Background()` at the call sites. 

Example project:

```
package main

import (
	"github.com/pulumi/pulumi-aws/sdk/v5/go/aws/s3"
)

func main() {
		bucket, err := s3.NewBucket(ctx, "my-bucket", nil)
		if err != nil {
			return err
		}

		return nil
	})
}
```

```
pulumi up --tracing file:./up.trace --yes
PULUMI_DEBUG_COMMANDS=1 pulumi view-trace ./up.trace
# observe orphan traces around bulkdecrypt calls
```

Fixes # (issue)

## Checklist

### DIFF
--- a/pkg/backend/filestate/state.go
+++ b/pkg/backend/filestate/state.go
@@ -87,17 +87,22 @@ func (u *update) GetTarget() *deploy.Target {
 	return u.target
 }
 
-func (b *localBackend) newQuery(ctx context.Context,
+func (b *localBackend) newQuery(
+	ctx context.Context,
 	op backend.QueryOperation) (engine.QueryInfo, error) {
 
 	return &localQuery{root: op.Root, proj: op.Proj}, nil
 }
 
-func (b *localBackend) newUpdate(stackName tokens.Name, op backend.UpdateOperation) (*update, error) {
+func (b *localBackend) newUpdate(
+	ctx context.Context,
+	stackName tokens.Name,
+	op backend.UpdateOperation) (*update, error) {
 	contract.Require(stackName != "", "stackName")
 
 	// Construct the deployment target.
-	target, err := b.getTarget(stackName, op.StackConfiguration.Config, op.StackConfiguration.Decrypter)
+	target, err := b.getTarget(ctx, stackName,
+		op.StackConfiguration.Config, op.StackConfiguration.Decrypter)
 	if err != nil {
 		return nil, err
 	}
@@ -111,8 +116,12 @@ func (b *localBackend) newUpdate(stackName tokens.Name, op backend.UpdateOperati
 	}, nil
 }
 
-func (b *localBackend) getTarget(stackName tokens.Name, cfg config.Map, dec config.Decrypter) (*deploy.Target, error) {
-	snapshot, _, err := b.getStack(stackName)
+func (b *localBackend) getTarget(
+	ctx context.Context,
+	stackName tokens.Name,
+	cfg config.Map,
+	dec config.Decrypter) (*deploy.Target, error) {
+	snapshot, _, err := b.getStack(ctx, stackName)
 	if err != nil {
 		return nil, err
 	}
@@ -124,7 +133,9 @@ func (b *localBackend) getTarget(stackName tokens.Name, cfg config.Map, dec conf
 	}, nil
 }
 
-func (b *localBackend) getStack(name tokens.Name) (*deploy.Snapshot, string, error) {
+func (b *localBackend) getStack(
+	ctx context.Context,
+	name tokens.Name) (*deploy.Snapshot, string, error) {
 	if name == "" {
 		return nil, "", errors.New("invalid empty stack name")
 	}
@@ -137,7 +148,7 @@ func (b *localBackend) getStack(name tokens.Name) (*deploy.Snapshot, string, err
 	}
 
 	// Materialize an actual snapshot object.
-	snapshot, err := stack.DeserializeCheckpoint(chk)
+	snapshot, err := stack.DeserializeCheckpoint(ctx, chk)
 	if err != nil {
 		return nil, "", err
 	}
--- a/pkg/cmd/pulumi/import.go
+++ b/pkg/cmd/pulumi/import.go
@@ -1,4 +1,4 @@
-// Copyright 2016-2020, Pulumi Corporation.
+// Copyright 2016-2022, Pulumi Corporation.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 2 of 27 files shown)


sdk/go/common/resource/config/crypt.go
```diff
@@ -31,14 +32,19 @@ import (
 // Encrypter encrypts plaintext into its encrypted ciphertext.
 type Encrypter interface {
 	EncryptValue(plaintext string) (string, error)
+	EncryptValueWithContext(ctx context.Context, plaintext string) (string, error)
```
t0yv0, line 35: This is a breaking change for code that implements the interface but not code that consumes it.

pkg/resource/stack/deployment.go
```diff
@@ -165,7 +166,9 @@ func SerializeDeployment(snap *deploy.Snapshot, sm secrets.Manager, showSecrets
 // from it. DeserializeDeployment will return an error if the untyped deployment's version is
 // not within the range `DeploymentSchemaVersionCurrent` and `DeploymentSchemaVersionOldestSupported`.
 func DeserializeUntypedDeployment(
```
t0yv0, line 168: Wondering if it's fair game to change the signature here or need to add a helper `*WithContext` to not break the API.

pkg/backend/filestate/state.go
```diff
@@ -137,7 +137,8 @@ func (b *localBackend) getStack(name tokens.Name) (*deploy.Snapshot, string, err
 	}
 
 	// Materialize an actual snapshot object.
-	snapshot, err := stack.DeserializeCheckpoint(chk)
+	// TODO context.Background() seems wrong, keep refactoring ?
+	snapshot, err := stack.DeserializeCheckpoint(context.Background(), chk)
```
t0yv0, line 141: TBD trying harder here.

pkg/cmd/pulumi/import.go
```diff
@@ -184,11 +184,12 @@ func parseImportFile(f importFile, protectResources bool) ([]deploy.Import, impo
 }
 
 func getCurrentDeploymentForStack(s backend.Stack) (*deploy.Snapshot, error) {
-	deployment, err := s.ExportDeployment(context.Background())
+	ctx := context.Background() // TODO this seems wrong, should pass in from commandContext()?
```
t0yv0, line 187: TBD trying harder here.

pkg/resource/stack/checkpoint.go
```diff
@@ -107,10 +108,11 @@ func SerializeCheckpoint(stack tokens.Name, snap *deploy.Snapshot,
 
 // DeserializeCheckpoint takes a serialized deployment record and returns its associated snapshot. Returns nil
 // if there have been no deployments performed on this checkpoint.
-func DeserializeCheckpoint(chkpoint *apitype.CheckpointV3) (*deploy.Snapshot, error) {
+func DeserializeCheckpoint(ctx context.Context, chkpoint *apitype.CheckpointV3) (*deploy.Snapshot, error) {
```
t0yv0, line 111: Same question whether we can change the signature here.

pkg/testing/integration/program.go
```diff
@@ -573,7 +573,8 @@ func GetLogs(
 	stackInfo RuntimeValidationStackInfo,
 	query operations.LogQuery) *[]operations.LogEntry {
 
-	snap, err := stack.DeserializeDeploymentV3(*stackInfo.Deployment, stack.DefaultSecretsProvider)
+	snap, err := stack.DeserializeDeploymentV3(context.Background(),
```
t0yv0, line 576: Assuming safe to do context.Background() here but could not figure out where this func is used, in testing?

sdk/go/common/resource/config/crypt.go
iwahbe: If we want to be backward compatible, I think we need a new interface. It's not an uncommon pattern to add a `EncrypterWithContext` interface that requires `Encrypter` and then try down casting from `Encrypter` to `EncrypterWithContext` when you need to specify the context. This is how the `error` module does error chaining.

iwahbe (commented):
This looks good to me in general. See my comments for details.

sdk/go/common/resource/config/crypt.go
t0yv0: I was fearing that as it seems to get baroque quickly, but I'll try to work it out. The benefit in the PR is not worth imposing a breaking change.

Frassle: I'm not sure why this stuff is even in common. It's probably fine to just change it.

Frassle: So 
A) This is already a breaking change because you've added methods to an interface
B) I don't think we care about breaking changes in common unless they're likely used by programs, so I'd rather take the "simpler" breaking change of just adding context to the existing Encrypt/Decrypt methods.

t0yv0: That sounds good to me! If we're pretty sure user code does not call Encrypter. I will edit and add a changelog.

pkg/cmd/pulumi/import.go
```diff
@@ -510,7 +514,7 @@ func newImportCmd() *cobra.Command {
 			}, imports)
 
 			if generateCode {
-				deployment, err := getCurrentDeploymentForStack(s)
+				deployment, err := getCurrentDeploymentForStack(ctx, s)
```
Frassle, line 517: s.Import above can take ctx instead of commandContext() as well

pkg/cmd/pulumi/stack_import.go
```diff
@@ -76,7 +77,7 @@ func newStackImportCmd() *cobra.Command {
 			// We do, however, now want to unmarshal the json.RawMessage into a real, typed deployment.  We do this so
 			// we can check that the deployment doesn't contain resources from a stack other than the selected one. This
 			// catches errors wherein someone imports the wrong stack's deployment (which can seriously hork things).
-			snapshot, err := stack.DeserializeUntypedDeployment(&deployment, stack.DefaultSecretsProvider)
+			snapshot, err := stack.DeserializeUntypedDeployment(ctx, &deployment, stack.DefaultSecretsProvider)
```
Frassle, line 80: Again, there's a call to s.ImportDeployment below which can take ctx instead of commandContext()

Frassle (approved):

Issue #10042, opened 2022-07-04
[codegen/dotnet] `pulumi convert` outputs unexpected InputParameters when language-specific settings exist


kpitzen:
### What happened?

When the provider schema contains a `Schema.Language.Name` value, codegen is currently not renaming the corresponding `InputParameter` in accordance with that specification (despite the corresponding resource expecting that name in its `<resource>Args` constructor).

### Steps to reproduce

Create an AWS IAM Policy resource in `yaml` with a corresponding JSON policy document:
```yaml
# Pulumi.yaml
resources:
  policy:
    type: aws:iam:Policy
    properties:
      path: /
      description: My test policy
      policy:
        Fn::ToJSON:
          Version: 2012-10-17
          Statement:
            - Effect: Allow
              Principal: '*'
              Action:
                - s3:GetObject
              Resource:
                - arn:aws:s3:::some-aws-bucket/*
              Condition:
                Foo:
                  Bar:
                    - iamuser-admin
                    - iamuser2-admin
                Baz:
                  Qux:
                    - iamuser3-admin
```

### Expected Behavior

`pulumi convert` outputs corresponding C# code with correctly-named arguments:
```csharp
// MyStack.cs
class MyStack : Stack
{
    public MyStack()
    {
        var createPolicy = new Aws.Iam.Policy("createPolicy", new Aws.Iam.PolicyArgs
        {
            Path = "/",
            Description = "My test policy",
            PolicyDocument = JsonSerializer.Serialize(new Dictionary<string, object?>
            {
                { "Version", "2012-10-17" },
                { "Statement", new[]
                    {
                        new Dictionary<string, object?>
                        {
                            { "Action", new[]
                                {
                                    "s3:GetObject",
                                }
                             },
                            { "Effect", "Allow" },
                            { "Resource", "arn:aws:s3:::some-aws-bucket/*" },
                            { "Condition", new Dictionary<string, object?>
                            {
                                { "Foo", new Dictionary<string, object?>
                                {
                                    { "Bar", new[]
                                        {
                                            "iamuser-admin",
                                            "iamuser2-admin",
                                        }
                                     },
                                } },
                                { "Baz", new Dictionary<string, object?>
                                {
                                    { "Qux", new[]
                                        {
                                            "iamuser3-admin",
                                        }
                                     },
                                } },
                            } },
                        },
                    }
                 },
            }),
        });
    }

}

```

### Actual Behavior

[...]

Issue #10033, opened 2022-07-01
pulumi convert panics when converting iam.Policy to Go


stack72:
### What happened?

When trying to convert a simple program to go, the convert command throws a panic:

```
goroutine 1 [running]:
runtime/debug.Stack()
	/Users/runner/hostedtoolcache/go/1.17.11/x64/src/runtime/debug/stack.go:24 +0x65
main.panicHandler()
	/Users/runner/work/pulumi/pulumi/pkg/cmd/pulumi/main.go:29 +0x3d
panic({0x5713280, 0xc006b73ee0})
	/Users/runner/hostedtoolcache/go/1.17.11/x64/src/runtime/panic.go:1038 +0x215
github.com/pulumi/pulumi/pkg/v3/codegen/go.GenerateProgramWithOptions(0xc0004227c0, {0x40})
	/Users/runner/work/pulumi/pulumi/pkg/codegen/go/gen_program.go:121 +0x7b3
github.com/pulumi/pulumi/pkg/v3/codegen/go.GenerateProgram(...)
	/Users/runner/work/pulumi/pulumi/pkg/codegen/go/gen_program.go:55
github.com/pulumi/pulumi/pkg/v3/codegen/go.GenerateProject({0x7ff7bfeff3eb, 0x2}, {{0xc00082f9b0, 0x10}, {{0xc00082f9d0, 0x4}, 0x0}, {0x0, 0x0}, 0xc0003c1dc0, ...}, ...)
	/Users/runner/work/pulumi/pulumi/pkg/codegen/go/gen_program.go:131 +0x4b
main.newConvertCmd.func1(0xf, {0xc000830e00, 0xc000b57d90, 0x41f4225})
	/Users/runner/work/pulumi/pulumi/pkg/cmd/pulumi/convert.go:89 +0x3bc
github.com/pulumi/pulumi/sdk/v3/go/common/util/cmdutil.RunResultFunc.func1(0xc000932280, {0xc000439400, 0x0, 0x4})
	/Users/runner/work/pulumi/pulumi/sdk/go/common/util/cmdutil/exit.go:112 +0x4f
github.com/spf13/cobra.(*Command).execute(0xc000932280, {0xc0004393c0, 0x4, 0x4})
	/Users/runner/go/pkg/mod/github.com/spf13/cobra@v1.4.0/command.go:860 +0x5f8
github.com/spf13/cobra.(*Command).ExecuteC(0xc000740c80)
[...]

PR #10043, opened 2022-07-04
fix: Adds support for C#-specific InputProperty names

kpitzen:
# Description

Fixes [#10042  (issue)](https://github.com/pulumi/pulumi/issues/10042)

## Checklist

### DIFF
--- a/pkg/codegen/testing/test/program_driver.go
+++ b/pkg/codegen/testing/test/program_driver.go
@@ -75,6 +75,12 @@ var PulumiPulumiProgramTests = []ProgramTest{
 		// Flaky in go: TODO[pulumi/pulumi#8123]
 	},
 	{
+		Directory:   "aws-iam-policy",
+		Description: "AWS IAM Policy",
+		SkipCompile: codegen.NewStringSet("go"),
+		// Blocked on go
+	},
+	{
 		Directory:   "aws-optionals",
 		Description: "AWS get invoke with nested object constructor that takes an optional string",
 		// Testing Go behavior exclusively:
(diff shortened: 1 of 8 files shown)


pkg/codegen/testing/test/program_driver.go
```diff
@@ -74,6 +74,12 @@ var PulumiPulumiProgramTests = []ProgramTest{
 		// Blocked on nodejs: TODO[pulumi/pulumi#8068]
 		// Flaky in go: TODO[pulumi/pulumi#8123]
 	},
+	{
+		Directory:   "aws-iam-policy",
+		Description: "AWS IAM Policy",
+		SkipCompile: codegen.NewStringSet("go"),
+		// Blocked on go
```
kpitzen, line 81: I found this issue while looking into #10033 - I'm still working on that one, and once it's resolved, this can be removed.

kpitzen: Looking at this more closely, I think we can actually _not_ skip `go` here.  The issue in #10033 appears to actually be related to the resulting pcl generated by `pulumi-yaml` rather than the pcl -> go codegen process.

Zaid-Ajaj (approved):
Thanks a lot @kpitzen for the contribution. The code and test case added look great! This is good to go ✅ just needs a rebase from `master` to fix the failure of downstream tests

kpitzen:
> Thanks a lot @kpitzen for the contribution. The code and test case added look great! This is good to go ✅ just needs a rebase from `master` to fix the failure of downstream tests

Thanks @Zaid-Ajaj !  I've just rebased, so tests should be good to go. :)

PR #10073, opened 2022-07-08
Unparent deleted resources when rebuilding state

### LINKED ISSUE #3583: Parents deleted by refresh corrupt Pulumi state file
When state integrity goes sour, `pulumi stack rm` should continue to function.

Steps to reproduce:
1. Deploy k8s deployment and service
2. manually delete deployment
3. `pulumi refresh`. This should not be allowed as it breaks the stack. A warning is shown that the service is now unhealthy (does not target pod)
4. `pulumi destroy` or `pulumi stack rm`
```
error: .pulumi/stacks/dev.json: snapshot integrity failure; refusing to use it: child resource urn:pulumi:dev::gitkit::kubernetes:apps/v1:Deployment$kubernetes:core/v1:Service::gitea refers to missing parent urn:pulumi:dev::gitkit::kubernetes:apps/v1:Deployment::gitea
```

iwahbe:
# Description

Fixes #3583

## Checklist

### DIFF
--- a/pkg/resource/deploy/deployment_executor.go
+++ b/pkg/resource/deploy/deployment_executor.go
@@ -645,7 +645,7 @@ func (ex *deploymentExecutor) rebuildBaseState(resourceToStep map[*resource.Stat
 
 		if new == nil {
 			if refresh {
-				contract.Assert(old.Custom)
+				contract.Assertf(old.Custom, "Expected custom resource")
 				contract.Assert(!providers.IsProviderType(old.Type))
 			}
 			continue
@@ -660,6 +660,12 @@ func (ex *deploymentExecutor) rebuildBaseState(resourceToStep map[*resource.Stat
 				}
 			}
 			new.Dependencies = deps
+		}
+
+		if new.Parent != "" {
+			if _, ok := olds[new.Parent]; !ok {
+				new.Parent = ""
+			}
 		}
 
 		// Add this resource to the resource list and mark it as referenceable.
--- a/pkg/resource/deploy/deployment_executor.go
+++ b/pkg/resource/deploy/deployment_executor.go
@@ -662,12 +662,6 @@ func (ex *deploymentExecutor) rebuildBaseState(resourceToStep map[*resource.Stat
 			new.Dependencies = deps
 		}
 
-		if new.Parent != "" {
-			if _, ok := olds[new.Parent]; !ok {
-				new.Parent = ""
-			}
-		}
-
 		// Add this resource to the resource list and mark it as referenceable.
 		resources = append(resources, new)
 		referenceable[new.URN] = true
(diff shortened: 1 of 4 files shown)


pkg/resource/deploy/deployment_executor.go
```diff
@@ -662,6 +662,12 @@ func (ex *deploymentExecutor) rebuildBaseState(resourceToStep map[*resource.Stat
 			new.Dependencies = deps
 		}
 
+		if new.Parent != "" {
```
t0yv0, line 665: This function before the changes is pretty tough to follow and I'm having a hard time trusting it does the right thing,  with the addition increasing the conceptual burden further.

- is there any way to add the new code in a separate loop and function dedicated to cleaning up parents and with its own docstring?

- if we keep this code in here, could we comment a bit on what's it doing and why, and add a comment to the docstring of the function that it also takes care of parent links

- why `olds[new.Parent]` vs `referenceable[d]`?  
 
- it is concerning that both `olds` and `referenceable` are constructed incrementally over `range ex.deployment.prev.Resources` that uses the top-sort assumption for dependencies; does Parent patch up also need the top-sort assumption or is it somehow compatible with it because a Parent link is also a dependency link?

t0yv0:
Reading the original ticket, it *seems* that we have a snapshot integrity check that is applied on read but not on write; therefore we write a snapshot without integrity to the backend for some reason.

```
error: .pulumi/stacks/dev.json: snapshot integrity failure; refusing to use it: child resource urn:pulumi:dev::gitkit::kubernetes:apps/v1:Deployment$kubernetes:core/v1:Service::gitea refers to missing parent urn:pulumi:dev::gitkit::kubernetes:apps/v1:Deployment::gitea
```

Would it make sense to backlog something about verifying the integrity before writing?

I was curious and found this bit that makes me wonder if VerifyIntegrity is too strict and the code intentionally writes out snapshots that do not pass, but it's not super clear why.

```
	if !DisableIntegrityChecking {
		// Finally, *after* writing the checkpoint, check the integrity.  This is done afterwards so that we write
		// out the checkpoint file since it may contain resource state updates.  But we will warn the user that the
		// file is already written and might be bad.
		if verifyerr := snap.VerifyIntegrity(); verifyerr != nil {
			return "", fmt.Errorf(
				"%s: snapshot integrity failure; it was already written, but is invalid (backup available at %s): %w",
				file, bck, verifyerr)

		}
	}
```

pkg/resource/deploy/deployment_executor.go
iwahbe: I'm going to address each point in turn:
- Yes, and it reads much cleaner. See https://github.com/pulumi/pulumi/pull/10073/commits/c2cd1313170be05230f8a0b314e8e1a76fc51739.
- I added a docstring to the new function I introduced.
- Because olds doesn't include resources with pending deletes. https://github.com/pulumi/pulumi/blob/c2cd1313170be05230f8a0b314e8e1a76fc51739/pkg/resource/deploy/deployment_executor.go#L669-L672
- Parent is part of the requirement of topological sort, and we do rely on this property.

iwahbe:
I'm concerned that by the time we can verify the integrity of a snapshot, we need to write it since it reflects the new state of the world. This would mandate writing our best effort snapshots, even if they are not valid.

t0yv0 (approved):
Thank you for elaborating!

Issue #10066, opened 2022-07-07
Filename `plugin.proto` is too generic, leading to conflict


ringods:
### What happened?

I bumped into this error as part of my hackaton work on contract based testing of a provider, using `pact-go` together with the Go pulumi sdk, both registering `plugin.proto`:

```
panic: proto: file "plugin.proto" is already registered
        previously from: "github.com/pact-foundation/pact-go/v2/internal/native"
        currently from:  "github.com/pulumi/pulumi/sdk/v3/proto/go"
See https://developers.google.com/protocol-buffers/docs/reference/go/faq#namespace-conflict
```

### Steps to reproduce

I have my hackaton work in this branch:

https://github.com/ringods/pulumi-github/tree/pact-testing

The first commit contains the plain setup with which I encountered the problem using the latest tagged version of the pulumi SDK, together with the then released beta version of `pact-go`.

After bumping into the problem, I [reported this](https://github.com/pact-foundation/pact-plugins/issues/4) with the `pact-go` team, they updated their code and released a new beta version on the same day. Later commits on my branch take in this new beta version, which resolves it now between Pulumi and Pact codebases.

### Expected Behavior

Multiple projects should be able to register their protobuf definitions.

### Actual Behavior

Since [protobuf-go v1.26](https://github.com/protocolbuffers/protobuf-go/releases/tag/v1.26.0), such a filename collision generates a panic.

### Additional context

Although both files have different `go_package` directives, as you can see here:

*  https://github.com/pulumi/pulumi/blob/master/sdk/proto/plugin.proto
*  https://github.com/pact-foundation/pact-plugins/blob/f0fdfefef0099518635329a6deab424961690b62/proto/plugin.proto

this doesn't seem to be enough. The filenames need to be unique as well, according this comment on the `protobuf` project:

https://github.com/golang/protobuf/issues/1122#issuecomment-853158260

I strongly suggest to that we also rename our protobuf file to e.g. pulumi-plugin.proto or pulumi/plugin.proto to prevent future filename collisions.

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

Frassle:
I'm not sure what knock on doing a rename here will have on providers. I'd like to move the proto files to be something like "/proto/pulumi/plugin.proto" and to then build them as "pulumi/plugin.proto" so we get the namespacing.

ringods:
[...]

PR #10074, opened 2022-07-08
Compile protobufs with "pulumi" namespace

Frassle:
# Description

This rebuilds the protobufs with a relative path including "pulumi/", that is instead of building "plugin.proto" we now build "pulumi/plugin.proto". Protoc takes that relative path into account when building the message descriptors such that we are now less likely to conflict with messages declared by other projects.

To try and keep the exisiting file structure as much as possible we're doing some sed replacements over the node and python generated code to keep imports relative as if everything was not namespaced. The go code generation only uses the relative path for the descriptor, go imports use the declared `option go_package`.

Fixes https://github.com/pulumi/pulumi/issues/10066

## Checklist


Frassle:
N.B Java might need updating to take this new path to the proto files into account as well.

t0yv0 (approved):

Frassle:
> So presumably this is just a renaming that's fully internal to the consumers. Can't think of anything that'd break.

Yes, I think the only thing that might of been using these is provider code but that will be go and the go types haven't changed at all. It's really awkward how protoc inconsistently uses the protobuf namespace in type generation across the languages, would probably have been nicer if node and python didn't change as well. But I didn't find an option for that, and it shouldn't matter given protobuf stuff is pretty internal.

Frassle:
> But I didn't find an option for that, and it shouldn't matter given protobuf stuff is pretty internal.

Actually even further "this should be ok" the way these are imported for something like the policy SDK means that it will continue to import fine even after these changes. Its just a load of internal imports in the code here look different.

PR #10081, opened 2022-07-09
Address data races detected by `go test -race`

AaronFriel:
These races were detected via Go's race detection tool and are separated by commit.

```
cd pkg && GOWORK=off go test -race -p 1 --parallel 4 -v ./engine/...
```


sdk/go/pulumi/types.go
```diff
@@ -1217,3 +1219,67 @@ func init() {
 	RegisterOutputType(ResourceOutput{})
 	RegisterOutputType(ResourceArrayOutput{})
 }
+
```
RobbieMcKinstry, line 1222: Personally, I would remove this stack trace comment and adjust line 289 to indicate how to reproduce the stack trace. I think that's preferable to including a trace that could become stale and grabs a lot of screen real estate.

RobbieMcKinstry (approved):
This LGTM! Can we adjust the Makefile to run the race detector as part of CI @AaronFriel? That way we don't let new races pop up in this section of the codebase.

sdk/go/pulumi/types.go
Frassle: +1

RobbieMcKinstry:
I'm forgetting how this is implemented. IIRC it's doing runtime detection, which is why tests slow down. I never understood why the Go team didn't implement an interprocedural analysis for this feature instead.

Thanks for volunteering to add the issue to the quality bar epic. IMO its perfectly fine to add it later when we have more opportunity to optimize its execution around CI bandwidth constraints.

AaronFriel:
@RobbieMcKinstry Created:

- https://github.com/pulumi/pulumi/issues/10091
- https://github.com/pulumi/pulumi/issues/10092

Issue #10057, opened 2022-07-06
Increase the concurrency of the S3 state backend


dmattia:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Some self-managed backends, such as the S3 backend, are quite slow in comparison to other backends, like the Pulumi Cloud backend. There are comments on larger issues (https://github.com/pulumi/pulumi/issues/4605#issuecomment-904477833) that reference the slowness of the S3 backend in particular, as well as a comment from a maintainer in https://github.com/pulumi/pulumi/issues/6074 that "the S3 backend is not as optimized currently as the Pulumi Service backend".

Making this more specific for this issue, I believe that one of the root causes of the slowness is that the operations when using an S3/azure state backend are written to the backend in sequence, rather than in parallel like seems to happen for the Pulumi backend.

From looking at the blob library used for talking to many of the backends, it looks like the library exposes MaxConcurrency field: https://github.com/google/go-cloud/blob/60d79299da9fb31d6bd022ae834b6766f45b55b1/blob/driver/driver.go#L65 that could be passed through in the *blob.WriterOptions of the WriteAll method: https://github.com/pulumi/pulumi/blob/ca3b9a57276751c71a4ca7fd10710edcdd72cfd3/pkg/backend/filestate/bucket.go#L22 that is used to write state checkpoint files.

Would you be open to a change where pulumi up could have a new flag named --max-state-concurrency where if set, that value would be passed through to the blob for whatever state is used? Looking through the code, it looks like Azure and S3 are the only blob backends that support it right now.

My understanding of the logic here gets pretty fuzzy though, as I can't seem to find out if that change alone would improve anything, or if there are other differences between the state backends that need to be addressed first.

Link to original Slack thread I created: https://pulumi-community.slack.com/archives/C84L4E3N1/p1657074321389129

### Affected area/feature

This would affect the S3 and Azure state backends, specifically when they call this method: https://github.com/pulumi/pulumi/blob/ca3b9a57276751c71a4ca7fd10710edcdd72cfd3/pkg/backend/filestate/bucket.go#L22

RobbieMcKinstry:
Huh, very interesting, thank you for the write-up! I think it's unlikely we'd accept a flag for `--max-state-concurrency`. My thought is that the flag wouldn't be common enough to warrant a slot as a CLI argument.
[...]

PR #10083, opened 2022-07-09
Increase s3 concurrency options in go-cloud

dmattia:
# Description

I am still drafting this message as I test the new go-cloud functionality

Fixes https://github.com/pulumi/pulumi/issues/10057

## Checklist


lunaris:
Thanks for raising this and sorry it went stale. Given the age of this and the lack of follow-up, I'm going to close this. That said, if there is a need to pursue this and the energy to do so, I'd recommend we look at the following approach:

* Rather than hardcoding this, we probably need to make it configurable, as it seems unlikely that 30 connections (say) will work for everyone in every scenario.
* One option here is to think about extending the `backend` configuration read from `Pulumi.yaml` or the like, so that it has a subfield called `maxConcurrency` that we then thread through to the backend/gocloud. There may be other better options though.
* With all that in, we'd want some tests, documentation and in this case, potentially some rough benchmarking to understand that this feature does provide a benefit worth taking on the complexity for.

I hope this makes sense :pray:

PR #10085, opened 2022-07-10
Add `As*Output` methods to `AnyOutput`

### LINKED ISSUE #6356: [sdk/go] Add AnyOuput type assertion methods?
The `AnyOutput` type represents an `Output` whose value is an `interface{}`.  

It is an important type because `Apply` returns an `AnyOutput`.  But it is unfortunately rather hard to use - as you can't pass it to anything directly.  Instead, you have to do `anyout.ApplyInt(func (v interface{}) int { return v.(int) }).  

In practice, that makes `.Apply` close to unusable - because you ~always must follow an `.Apply` with a `.ApplyT`.  So it's basically always preferable to just use `ApplyT` directly.

We could try to make `AnyOutput` more usable by offering type assertion methods on it - like `anyoutout.AsInt()` via :

```go
func (o AnyOutput) AsInt() IntOutput {
  return o.ApplyInt(func (v interface{}) int {
    return v.(int)
  })
}
```

This method can of course panic because it is a type assertion.  But it would make `AnyOutput` and `Apply` more reasonable to work with.  It would only require N new methods total (not N^2, as these are only needed on `AnyOutput`).

lukehoban:
`AnyOutputs` are outputs wrapping underlying values of type `interface{}`.  This are fairly hard to use today, as noted in #6356.  This PR adds new `As*Output` methods to `AnyOutput` for all builtin types, allowing the `interface{}` value to be asserted to be any of the builtin element types.

As an example - the workaround in https://github.com/pulumi/pulumi/issues/6356#issuecomment-919821721 and be written instead as:

```go
publicSubnets := vpcStack.GetOutput(pulumi.String("publicSubnetIDs")).AsStringArrayOutput()
```

Note that we could go further and add similar overloads of `GetOutput` directly to `StackReference`, but this API ensures that *any* source of `AnyOutput` values can be used more easily.

Two primary API design questions:
1. Should the name be `ToStringOutput`, `ToString`, `AsStringOutput` or `AsString` (or something else)?  I believe `AsStringOutput` is best as it makes clear the result is a StringOutput, and `As` is slightly more descriptive of this being an assertion instead of conversion (and thus that it may panic).
2. Should this panic, or return an error into the Output value?  I'm not sure how much additional value there is in returning an error into the Output resolution, as that still won't be directly catchable by user code.  In practice, it feels like users have to treat this like a potentially-panicing type assertion at the point it is used.

Fixes #6356.


iwahbe (approved):
I think we should take this change.

 That being said, we need to be careful not to rely on it, since we can not assume that all type conversions will be via a method builtin. I'm thinking of types like `StringArrayArrayArrayOutput`.

Issue #10087, opened 2022-07-11
Not all properties are documented in the JSON schema or on the website


ringods:
### What happened?

PR #7631 added support for 2 properties in the code generation of the NodeJS specific language info:

- `pluginName`
- `pluginVersion`

These properties are not documented in the [Pulumi JSON Schema definition](https://raw.githubusercontent.com/pulumi/pulumi/master/pkg/codegen/schema/pulumi.json) file.

These properties are also not documented on the [Schema page](https://www.pulumi.com/docs/guides/pulumi-packages/schema/#packagelanguage) of the website.

### Steps to reproduce

Configure your IDE with JSON Schema support like this:

https://github.com/pulumi/pulumi-provider-boilerplate/#writing-the-schema

In the NodeJS language info section of your provider, try to execute completion on one of the 2 properties.

### Expected Behavior

When leveraging the `pluginName` property in the provider schema file and having the code from #7631, a third-party provider (Pulumiverse and also reported by a customer), this sets the `name` property under the `pulumi` section of the generated `package.json` file. When using the NodeJS language runtime, this executes the following code path:

https://github.com/pulumi/pulumi/blob/7827eb26bc1a64aa56ad450d35bccde0ceadd78e/sdk/nodejs/cmd/pulumi-language-nodejs/main.go#L350-L353

We will also have proper code assist on both of these properties when writing the schema file by hand in an IDE.

### Actual Behavior

Without the `name` property under the `pulumi` section in `package.json`, one ends up in this code path:

https://github.com/pulumi/pulumi/blob/7827eb26bc1a64aa56ad450d35bccde0ceadd78e/sdk/nodejs/cmd/pulumi-language-nodejs/main.go#L361-L369

This could lead to errors as reported in https://github.com/pulumi/pulumi/issues/9606#issuecomment-1140317647, point 4:

```
[resource plugin jaxxstorm/pulumi-scaleway-0.2.1] installing error: [resource plugin jaxxstorm/pulumi-scaleway-0.2.1] downloading from : 403 HTTP error fetching plugin from https://get.pulumi.com/releases/plugins/pulumi-resource-jaxxstorm%2Fpulumi-scaleway-v0.2.1-windows-amd64.tar.gz
```

Because only the `@` symbol is stripped from calculating the plugin binary name, but not the NodeJS namespace.

Given they are not part of the JSON schema, no Intellisense like support is offered, rendering no clue this is possible.

### Additional context

PR #7631
Issue #9606 

[...]

PR #10088, opened 2022-07-11
Set the package.json pulumi name even if PluginName isn't set

Frassle:
# Description

Related to https://github.com/pulumi/pulumi/issues/10087

Currently the nodejs runtime will use "pulumi.name" from package.json as the plugin name, if it's not set it falls back to the npm package name (i.e. "name" in package.json). This is nearly always wrong as npm names look like "@pulumiverse/sentry" while the pulumi package name (and thus plugin name) would be something like "sentry".

This changes nodejs codegen to always set "pulumi.name" to either the user given "PluginName", or if that is not set to the pulumi package name. This should mean most users don't even need to set "PluginName".

## Checklist

### DIFF
--- a/pkg/codegen/nodejs/gen.go
+++ b/pkg/codegen/nodejs/gen.go
@@ -2135,6 +2135,13 @@ func genNPMPackageMetadata(pkg *schema.Package, info NodePackageInfo) string {
 		scriptVersion = pluginVersion
 	}
 
+	pluginName := info.PluginName
+	// Default to the pulumi package name if PluginName isn't set by the user. This is different to the npm
+	// package name, e.g. the npm package "@pulumiverse/sentry" has a pulumi package name of just "sentry".
+	if pluginName == "" {
+		pluginName = pkg.Name
+	}
+
 	// Create info that will get serialized into an NPM package.json.
 	npminfo := npmPackage{
 		Name:        packageName,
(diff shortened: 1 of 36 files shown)


ringods:
This will not completely fix #10087, but it will consistently set:

```json
  "pulumi": {
     "name": <plugin name>
  }
```

in the generated `package.json` files, hereby preventing to execute the faulty code path for third-party providers.

pkg/codegen/nodejs/gen.go
```diff
@@ -2135,6 +2135,13 @@ func genNPMPackageMetadata(pkg *schema.Package, info NodePackageInfo) string {
 		scriptVersion = pluginVersion
 	}
 
+	pluginName := info.PluginName
+	// Default to the pulumi package name if PluginName isn't set by the user. This is different to the npm
```
Frassle, line 2139: Note we could be a bit smarter here and only set name if needed. A lot of "@pulumi" packages would work even without name set here because we strip it off the npm name. That is a plugin "azure" with a npm package name of "@pulumi/azure" doesn't need the plugin name explicitly set. Similarly if the plugin name == the npm package name then this doesn't need to be set.

That would reduce the diff of this change a bit because all the test packages would fit this style, and also means things like actual azure codegen won't change from this.

But then maybe it's better to just be simple and explicit.

iwahbe: I'm happy to take the bigger diff if we get simpler logic.

PR #10094, opened 2022-07-12
Allow installing non-tar files or folders

iwahbe:
This expands the capabilities of `pulumi plugin install resource PKG
VERSION -f FILE` to support a single executable.

# Description

This makes it much easier to develop plugins, since we can trivial install them without wrapping the plugin in a zip directory.

Fixes # (issue)

## Checklist


Frassle:
> This makes it much easier to develop plugins, since we can trivial install them without wrapping the plugin in a zip directory.

Can we look into making this install from a folder rather than a single file. I like the idea of making plugins easier to develop locally (me and @danielrbradley have some other related ideas to this as well which wouldn't even require "installing"), but we've also got a lot of ideas floating around at the moment that will be moving us further and further away from plugins as single file executables. I think this would be much more future-proof if it took a folder with a plugin and copied that folder across to the plugins directory. This is pretty much a glorified copy but it can do some validation (does this folder even look like a plugin, etc) and means the user doesn't need to know the path ~/.pulumi/plugins/<kind>-<name>-<version>.

danielrbradley:
Yup, folder would be a nice addition here, though these could be added later – allow it to work with gzip or single binary now, then could also add directory after. There's a few conversations happening around moving the schema JSON out from being embedded in the binary for performance, compression and packaging process reasons – so the binary might also other files installed alongside it (though this could also just fall back to using the gzip option here).

danielrbradley (approved):

sdk/go/common/workspace/plugins.go
```diff
@@ -795,7 +796,109 @@ func (info PluginInfo) installLock() (unlock func(), err error) {
...
 // Install installs a plugin's tarball into the cache. See InstallWithContext for details.
 func (info PluginInfo) Install(tgz io.ReadCloser, reinstall bool) error {
-	return info.InstallWithContext(context.Background(), tgz, reinstall)
+	return info.InstallWithContext(context.Background(), tarPlugin{tgz}, reinstall)
+}
+
+type PluginContent interface {
+	io.Closer
+
+	WriteToDir(info PluginInfo) error
```
Frassle, line 805: I'd expect WriteToDir to take the directory path, not a PluginInfo. It also looks like things below are inconsistent about how they resolve from a PluginInfo to a directory.

Frassle (changes requested):

sdk/go/common/workspace/plugins.go
```diff
@@ -873,15 +976,14 @@ func (info PluginInfo) InstallWithContext(ctx context.Context, tgz io.ReadCloser
 		return err
 	}
 
-	// Uncompress the plugin.
-	if err := archive.ExtractTGZ(tgz, finalDir); err != nil {
+	if err := content.WriteToDir(info); err != nil {
```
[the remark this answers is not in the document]
iwahbe: `finalDir` is used 7 other times in this function. It is simply a cache of `info.DirPath`.

sdk/go/common/workspace/plugins.go
Frassle: I see because your not assuming the file is already correctly named and so you rename the file to match the expected pulumi-kind-name format. I wonder if that information should be part of creating the FileSource instead? You have a PluginInfo on hand when calling `getFileSource` and so could pass the info and file into SingleFileSource with it just saving a copy of the kind and name.

Then WriteToDir could really just take a directory, rather than each source having to call info.PluginDir().

iwahbe: That works. Done in https://github.com/pulumi/pulumi/pull/10094/commits/b1b61b842180008f33e5a1e9e154357d69d7fd8d.

Frassle (approved):

PR #10099, opened 2022-07-12
Fix docker in sdk/proto/generate.sh on M1

Zaid-Ajaj:
# Description

Needed to provide `--platform linux/x86_64` as described [here](https://stackoverflow.com/questions/68630526/lib64-ld-linux-x86-64-so-2-no-such-file-or-directory-error) in order to get `build-proto` MAKE target to run properly on my M1 machine. 

## Checklist


aq17 (approved):

aq17:
I had a different PR that was going to add this in but separately is probably better 🙌🏼

Issue #10092, opened 2022-07-11
Prevent data races from being released


AaronFriel:
As evidenced by #10081, we have some evidence of a few data races in the Pulumi engine, and we'd like to prevent races.

Unfortunately the race detection flag `-race` dramatically reduces the speed of tests, and implementing this today would negatively impact CI throughput and cycle time.

We could:

- Use the `-race` flag on a subset of tests we think have good coverage.
- Enable the `-race` flag for all tests after completing #10091 
- Use a scheduled GitHub action to run tests comprehensively with `-race`

---

Related issues: #8466, #1605, #11727, #11790

Frassle:
> Use a scheduled GitHub action to run tests comprehensively with -race

I'd like this more generally. I always feel there's three sets of tests you want:
1. Quick PR tests
2. Slower more thorough main branch tests (which https://github.com/pulumi/pulumi/issues/10091 would cover)
3. Very slow, very thorough nightly tests (Not sure when's "night" for us...).

RobbieMcKinstry:
I personally prefer:

> enable the `-race` flag for all tests

Because it's possible (and easy) to introduce races during day-to-day work, I think race detection should be on for all PRs. IMO races are dangerous enough that we should prefer them from landing on `master` at all. In short, the value provided by the `-race` flag on PR outweighs the cost to CI IMHO.

This isn't a hill I'll die on, so if the team disagrees I'm happy to support a different decision.

AaronFriel:
I think I agree with @RobbieMcKinstry for the most part. Fortunately I think there are large classes of tests that won't be impacted by adding `-race` or which we could remove `-race`, and doubly fortunately those are are slowest sets of tests. I still wouldn't want to turn this on by default until we have #10091, but it would allow us to run them on every PR.

1. Integration tests, to a first approximation 100% of their time is waiting for the `pulumi` CLI or copying files in the OS.

   I'd be surprised if adding `-race` impacted this at all.

3. Codegen & docs tests, which again, to a first approximation spend 100% of their time copying bytes in memory or files in the OS, and have _zero_ impact on the average users' workflow.

   We could have a `-race` enabled test that runs a subset of these, but there's no need to run most of these under `-race`. If we have non-deterministic output as a result of a data race, that fails because our codegen tests are snapshot tests, comparing their output to a committed artifact.

Frassle:
[...]

PR #10100, opened 2022-07-12
Fix additional data races in Go SDK

AaronFriel:
This addresses 3 additional data races discovered by running:

```
cd sdk && GOWORK=off go test -race -p 1 --parallel 4 -v ./go/...
```

Related issues:
- Customer report of data race panic that resulted in this work to reproduce: 

  #10060
- Races addressed in previous PR: #10081
- Enable race detection in CI: #10092

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -40,3 +40,8 @@
 - [go] Fix panic when returning pulumi.String.Bool, .String, .Int, and .Float64 in the argument to
   ApplyT and casting the result to the corresponding output, e.g.: BoolOutput.
   [#10103](https://github.com/pulumi/pulumi/pull/10103)
+
+- [engine] Fix data races discovered in CLI and Go SDK that could cause nondeterministic behavior
+  or a panic.
+  [#10081](https://github.com/pulumi/pulumi/pull/10081),
+  [#10100](https://github.com/pulumi/pulumi/pull/10100)


sdk/go/pulumi/types.go
```diff
@@ -547,8 +547,6 @@ func UnsafeUnknownOutput(deps []Resource) Output {
 func ToSecretWithContext(ctx context.Context, input interface{}) Output {
 	x := true
 	o := toOutputWithContext(ctx, nil, input, &x)
-	// set immediate secretness ahead of resolution/fufillment
-	o.getState().secret = true
```
AaronFriel, line 551: The 4th parameter of `toOutputWithContext` is `forceSecretVal *bool`, and it forces the value to be a secret on resolution:

https://github.com/pulumi/pulumi/blob/0b13e67b0a89f3ee03ca3e243e70a93386653a8e/sdk/go/pulumi/types.go#L895-L914

However, if this feels race-y still, line 905-907 could be moved above the `go func()` here and directly set output.getState().secret. 

That said, I think `secret`, `known`, and `deps` should all be considered illegal to read from until awaited. Any code that reads those fields before the await will have incomplete information in the case of nested outputs:

```go
// unwrap an Output[Output[T]]
someOutput.ApplyT(func (x AnyOutput) {
  return x
})
```

Or a concrete example, summing two int outputs without using .All():

```go
xOutput.ApplyT(func (x int) IntOutput {
  return yOutput.ApplyT(func (y int) IntOutput {
    return x + y
  }).(IntOutput)
}).(IntOutput)
```

This latter output has the combined known-ness, secret-ness, and dependencies of both xOutput and yOutput, but that can't be known until the outer output is awaited (which will await the inner and accumulate.)

sdk/go/pulumi/types.go
```diff
@@ -564,14 +564,13 @@ func AllWithContext(ctx context.Context, inputs ...interface{}) ArrayOutput {
 	return ToOutputWithContext(ctx, inputs).(ArrayOutput)
 }
 
-func gatherDependencies(v interface{}) ([]Resource, workGroups) {
+func gatherJoins(v interface{}) workGroups {
```
AaronFriel, line 567: See my other comment for a scenario where this falls down.

In PR #9506 and #9732, I modified `await` to properly implement a monadic `join` operation that accumulated the fields on output state such as `dependencies`, and introduced tests on that PR.

In general we shouldn't be inspecting the dependencies property until we await the output.

sdk/go/pulumi/types.go
RobbieMcKinstry: My concern with this change is less about the raciness of the boolean toggle, but more on relying on `toOutputWithContext` to guarantee the value is secret. I'd hate for someone to change the behavior of `toOutputWithContext` and inadvertently render `ToSecretWithContext` useless. 

I have to imagine catching a bug like that is the job of unit testing and we shouldn't worry about that here 😅

RobbieMcKinstry: But I also do think you're on to something with moving line 905-907 above the` go func()`. I agree that it should be illegal to read from a value before its awaited, but I also think users will do all kinds of screwy things with `Apply`, and the SecOps team would probably prefer we keep secrets secret even when their developers cause trouble. My gut says its less of a foot-gun to set the value outside of the Goroutine.

sdk/go/pulumi/types.go
RobbieMcKinstry: Awesome! Sounds like good engineering pays off! Thanks for the explanation here, that was very helpful in getting context for these changes.

RobbieMcKinstry (changes requested):
Overall looks good! Just want that one note on setting `getState().secret = true` that I'm interested in getting your feedback on before approving. 👍🏻

sdk/go/pulumi/types.go
AaronFriel: I checked and we do have tests for this that fail if the line is removed - secrets that fail to warn that they're secret.

Even so, I agree with "initializing" the output with secret=true before the goroutine call, because I cannot prove there aren't other places in our code where we've illegally read from the secret field before awaiting.

RobbieMcKinstry: That's a sufficient resolution for me then! Just re-request me when you make that change and I'll give you the hammer stamp! 🔨  Thanks Friel, this is super nice, definitely helps contribute towards our company's Q3 mission. :D

AaronFriel: I'll also add a comment on toOutputWithContext that the forceSecret argument is load bearing and ensures that the output is marked secret before any other code could possibly await or read that field.

AaronFriel:
It appears that I've made a change that failed a test @RobbieMcKinstry, moving this to a draft

AaronFriel:
Ah, I inadvertently pushed the changes you requested to my other open PR. 🤦 

Resolved.

AaronFriel:
Now that my other PR is in, I'm adding a commit with changelog info for this and #10081.

AaronFriel:
Rebased.

RobbieMcKinstry (approved):
LGTM! Thanks!

PR #10107, opened 2022-07-12
Insert deep copies prior to mutating objects held by EngineEvents.

RobbieMcKinstry:
# Description

Issue https://github.com/pulumi/pulumi/issues/10060 uncovered some insidious data races. This PR resolves two races specific to emitting Diagnostics.

`EngineEvents` are read from a channel and displayed to the user on a separate thread from ongoing engine work. When the events are read, a deepcopy is made of their `payload`. During the deepcopy, all nested fields are read from memory on the goroutine emitting diagnostics.

In Go, iterating over a map in one thread while writing to a map in another thread results in a panic, because writing to a map could cause a bucket rebalance, and the iterator on the reading thread could be reading from freed memory. Therefore, any `EngineEvent` passed to the `Diagnostics` channel cannot be mutated from another thread, because doing so could result in panic when the diagnostics thread deep-copies a map that's being concurrently written to.

As a complication, it's insufficient to deep-copy the `EngineEvent` prior to writing it to the channel, since _that_ deepcopy could be concurrent with a mutation in a separate (third) thread. Instead, we must locate any instance of mutation, and perform the deepcopy at the site of mutation. This pull request uncovered two such sites:

1. `deploymentExecutor.undangleParentResources`

2. `Snapshot.NormalizeURNReferences`

I attempted to insert deepcopies at more convenient locations farther up the stack, but I was unable to remove the data races with this approach. I opted to insert the copies as close to the mutation as possible.

It's possible this change could result in an (unacceptably large) increase in memory. I think that's unlikely since the objects being copied are not particularly numerous, and the locations where the copies are occurring seem to be called only once per object. Without actually profiling the memory usage of pulumi, those are just assumptions.

Fixes https://github.com/pulumi/pulumi/issues/10060

## Checklist

**Note:** You can't exactly write a test for this change. But you can validate the race detector no longer finds a race! `cd` into `/tests` and run...

```bash
PULUMI_TEST_ORG=robbie GOWORK=off go test -timeout 1h -tags all -run '^TestPrintfGo$' github.com/pulumi/pulumi/tests/integration -v
```

### DIFF
--- a/pkg/resource/deploy/deployment_executor.go
+++ b/pkg/resource/deploy/deployment_executor.go
@@ -26,6 +26,7 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/common/display"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/util/deepcopy"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/logging"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/result"
 )
@@ -671,11 +672,15 @@ func (ex *deploymentExecutor) rebuildBaseState(resourceToStep map[*resource.Stat
 			olds[new.URN] = new
 		}
 	}
+	// Make a deep copies before mutating in undangleParentResources,
+	// since olds and resources may be held by another thread as part of an engine event.
+	var oldsCopy = deepcopy.Copy(olds).(map[resource.URN]*resource.State)
+	var resourcesCopy = deepcopy.Copy(resources).([]*resource.State)
 
-	undangleParentResources(olds, resources)
+	undangleParentResources(oldsCopy, resourcesCopy)
 
-	ex.deployment.prev.Resources = resources
-	ex.deployment.olds, ex.deployment.depGraph = olds, graph.NewDependencyGraph(resources)
+	ex.deployment.prev.Resources = resourcesCopy
+	ex.deployment.olds, ex.deployment.depGraph = oldsCopy, graph.NewDependencyGraph(resourcesCopy)
 }
 
 func undangleParentResources(undeleted map[resource.URN]*resource.State, resources []*resource.State) {
--- a/pkg/resource/deploy/snapshot.go
+++ b/pkg/resource/deploy/snapshot.go
@@ -24,6 +24,7 @@ import (
 	"github.com/pulumi/pulumi/pkg/v3/secrets"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/util/deepcopy"
 )
 
 // Snapshot is a view of a collection of resources in an stack at a point in time.  It describes resources; their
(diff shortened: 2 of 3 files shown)


RobbieMcKinstry:
No guarantee that all data races with `ShowDiagnostics` are fixed with this PR, just the ones detected during certain integ tests. Would be nice to run `-race` more broadly in CI even just once and identify all tests which produce races. Dump those into a megaissue and follow up on them one at a time.

pkg/resource/deploy/snapshot.go
```diff
@@ -57,6 +58,10 @@ func NewSnapshot(manifest Manifest, secretsManager secrets.Manager,
 //
 // Note: This method modifies the snapshot (and resource.States in the snapshot) in-place.
 func (snap *Snapshot) NormalizeURNReferences() error {
+	// Make a deep copy prior to mutating, since this object's pointer
+	// may be held across thread boundaries as a field on an EngineEvent.
+	var snapCopy = deepcopy.Copy(snap).(*Snapshot)
+	*snap = *snapCopy
```
AaronFriel, line 64: This is a non-atomic write that could appear as a tearing read to another thread, right? We're writing to the same memory location as the original snapshot, but I imagine this appears as a non-atomic write.

Could we modify callers to clone the snapshot into a new memory location before calling `snap.NormalizeURNReferences()` and add a comment that this function mutates and should be provided with an owned copy of the snapshot?

AaronFriel (commented):
I'm curious about the semantics of doing `*snap = *snapCopy` in a method and I think it might introduce an even subtler memory race.

pkg/resource/deploy/snapshot.go
AaronFriel: I found that this also allowed me to run the test without a race.

It looks like in the two places we call NormalizeURNReferences outside of a test, it's immediately after initializing a new snapshot. If we do the deepcopy inside `NewSnapshot`, that meshes with the name of the method (it's a snapshot!) and I think is logically equivalent to these two lines.

There's one test where we call `NormalizeURNReferences` at the end and verify no error.

```go
// NewSnapshot creates a snapshot from the given arguments.  The resources must be in topologically sorted order.
// This property is not checked; for verification, please refer to the VerifyIntegrity function below.
func NewSnapshot(manifest Manifest, secretsManager secrets.Manager,
	resources []*resource.State, ops []resource.Operation) *Snapshot {

	return &Snapshot{
		Manifest:          deepcopy.Copy(manifest).(Manifest),
		SecretsManager:    deepcopy.Copy(secretsManager).(secrets.Manager),
		Resources:         deepcopy.Copy(resources).([]*resource.State),
		PendingOperations: deepcopy.Copy(ops).([]resource.Operation),
	}
}
```

AaronFriel: Though I suppose the above is a bit more aesthetically written as:

```go
	snap := &Snapshot{
		Manifest:          manifest,
		SecretsManager:    secretsManager,
		Resources:         resources,
		PendingOperations: ops,
	}
	return deepcopy.Copy(snap).(*Snapshot)
```

pkg/resource/deploy/deployment_executor.go
```diff
@@ -671,11 +672,15 @@ func (ex *deploymentExecutor) rebuildBaseState(resourceToStep map[*resource.Stat
 			olds[new.URN] = new
 		}
 	}
+	// Make a deep copies before mutating in undangleParentResources,
```
Frassle, line 675: This is concerning, snapshots can get fairly large (MBs to GBs) and having to deep copy them for every resource is going to result in N^N memory usage.

pkg/resource/deploy/snapshot.go
RobbieMcKinstry: Nice! I tried that with the snapshot manager, but without success. I thought perhaps the snapshot was leaking from a private field somewhere (i.e. it was assigned from the field into another variable). However, @Frassle 's point still stands. We might not be able to use a deepcopy at all to solve this problem.

RobbieMcKinstry:
I'm going to close this PR in favor of a PR for each of the two data races.

PR #10110, opened 2022-07-13
Handle input types in return position of ApplyT

AaronFriel:
As identified by a build failure on master here: https://github.com/pulumi/pulumi/runs/7313397089?check_suite_focus=true#step:35:100

The previous PR, #10103 failed to handle `IntInput`, and in creating a new zero value (nil) value for the interface failed.

This adds a check to see if the return type is an input type, then it attempts to look up the associated concrete type to operate on that instead. This handles BoolInput, IntInput, et al. And otherwise attempting to look up the element type in the concrete type to output type map.

Manually re-running the failing test succeeds. The test is a bit idiosyncratic, using `GetResource` in an Apply to sequence it after creating a RandomPet, but it makes a good test case: we should expect unusual Go programs in the wild and handle them well.

https://github.com/pulumi/pulumi/blob/73a66f48ead301bacf1d7d2833c016cf803b57c1/tests/integration/get_resource/go/main.go#L43-L49

### DIFF
--- a/sdk/go/pulumi/types.go
+++ b/sdk/go/pulumi/types.go
@@ -25,11 +25,6 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 )
 
-// wrapper enables retrieving the inner type of newtype wrappers such as pulumi.String
-type wrapper interface {
-	ElementType() reflect.Type
-}
-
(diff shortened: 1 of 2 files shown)


iwahbe (approved):
Assuming that tests go ✅: LGTM!

PR #10118, opened 2022-07-13
Accept short branch names in GitRepo.Branch

squaremo:
Users often give short names for branches -- e.g., `dev`, rather than `refs/heads/dev`. This commit makes that work rather than produce an error, by adopting the heuristic that anything that doesn't look like a full branch name (i.e., not starting with refs/heads/) must be a short name.


squaremo:
Test failure looks like a flake to me -- @Frassle wdyt?

squaremo:
> Test failure looks like a flake to me

Reran, and it passed. So yes, a flake :-S

Issue #7653, opened 2021-07-24
Lazily export nodejs sub-modules


joeduffy:
The Azure Native package is quite large, and the transitive closure of source code required to load if you import the base package in at least TypeScript can incur sizable startup penalties.

For example, the `pulumi new azure-typescript` template takes just a few seconds to preview on my laptop (not bad):

```typescript
import * as pulumi from "@pulumi/pulumi";
import * as resources from "@pulumi/azure-native/resources";
import * as storage from "@pulumi/azure-native/storage";

// Create an Azure Resource Group
const resourceGroup = new resources.ResourceGroup("resourceGroup");

// Create an Azure resource (Storage Account)
const storageAccount = new storage.StorageAccount("sa", {
    resourceGroupName: resourceGroup.name,
    sku: {
        name: storage.SkuName.Standard_LRS,
    },
    kind: storage.Kind.StorageV2,
});

// Export the primary key of the Storage Account
const storageAccountKeys = pulumi.all([resourceGroup.name, storageAccount.name]).apply(([resourceGroupName, accountName]) =>
    storage.listStorageAccountKeys({ resourceGroupName, accountName }));
export const primaryStorageKey = storageAccountKeys.keys[0].value;
```

```
real	0m4.859s
user	0m3.937s
sys	0m1.231s
```

An innocuous variant, however, explodes that time to 50 seconds to preview:

```typescript
import * as pulumi from "@pulumi/pulumi";
import * as az from "@pulumi/azure-native";

// Create an Azure Resource Group
const resourceGroup = new az.resources.ResourceGroup("resourceGroup");

// Create an Azure resource (Storage Account)
const storageAccount = new az.storage.StorageAccount("sa", {
    resourceGroupName: resourceGroup.name,
    sku: {
        name: az.storage.SkuName.Standard_LRS,
    },
    kind: az.storage.Kind.StorageV2,
});

// Export the primary key of the Storage Account
const storageAccountKeys = pulumi.all([resourceGroup.name, storageAccount.name]).apply(([resourceGroupName, accountName]) =>
    az.storage.listStorageAccountKeys({ resourceGroupName, accountName }));
export const primaryStorageKey = storageAccountKeys.keys[0].value;
```

```
real	0m49.224s
user	0m44.315s
sys	0m9.584s
```

Notice that this imports the root package instead of submodules directly. Many of our API examples do such things, and a lot of the idiomatic Pulumi TypeScript code we've written over the years does this. It's actually my personally preferred style.

[...]

PR #10122, opened 2022-07-13
Make ts-node and typescript peer dependencies

dmattia:
# Description

Fixes https://github.com/pulumi/pulumi/issues/8177
Related to https://github.com/pulumi/pulumi/issues/4876

Newer ts-node and typescript versions can make configurations not work with Pulumi at all right now. In my case, I am developing in a monorepo that uses newer versions of ts-node and typescript, and these older versions cause my code to not be able to run. See the more detailed descriptions I left in those two Issues linked above for what exactly is breaking

## Checklist

This is not an application code change

I'm not sure where to update this version, but this deserves probably a minor bump in the node sdk

### DIFF
--- a/sdk/nodejs/package.json
+++ b/sdk/nodejs/package.json
@@ -21,8 +21,6 @@
         "require-from-string": "^2.0.1",
         "semver": "^6.1.0",
         "source-map-support": "^0.5.6",
-        "ts-node": "^7.0.1",
-        "typescript": "~3.7.3",
         "upath": "^1.1.0"
     },
     "devDependencies": {
@@ -43,6 +41,10 @@
         "mocha": "^9.0.0",
         "mockpackage": "file:tests/mockpackage",
         "nyc": "^15.1.0"
+    },
+    "peerDependencies": {
+        "ts-node": ">7.0.1",
+        "typescript": ">3.7.3"
     },
     "pulumi": {
         "comment": "Do not remove. Marks this as as a deployment-time-only package"
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,5 +1,7 @@
 ### Improvements
 
+- [sdk/nodejs] - Updates SDK to allow using any newer version of ts-node and typescript
+
 - [cli] Display outputs during the very first preview.
   [#10031](https://github.com/pulumi/pulumi/pull/10031)
 


dmattia:
This is a copy of https://github.com/pulumi/pulumi/pull/8199, which unfortunately was closed because I force pushed to my repo's `master` branch and I guess that is not allowed.

Of note, this code was previously reviewed and is in consideration for inclusion during the next major release: https://github.com/pulumi/pulumi/pull/8199#issuecomment-941254998

sdk/nodejs/package.json
```diff
@@ -21,8 +21,6 @@
         "require-from-string": "^2.0.1",
         "semver": "^6.1.0",
         "source-map-support": "^0.5.6",
-        "ts-node": "^7.0.1",
-        "typescript": "~3.7.3",
```
dixler, line 25: discussion in https://github.com/pulumi/pulumi/pull/10122 around breaking changes.

dmattia:
Thanks for the review @dixler! Yeah this is definitely a breaking change. It was already included in the 4.0 Changes Tracking Issue (https://github.com/pulumi/pulumi/issues/7346#issuecomment-1061292570) I think.

I noticed you requested changes in the PR review, is there anything in particular you'd like me to change?

dixler:
> 

Closing it makes more sense. We can reopen it based on 4.0.

dmattia:
@dixler @RobbieMcKinstry  how do we reopen this based on 4.0? I don't see a branch for 4.0 yet 😕 

I'm also not sure why we couldn't just leave this open and then adjust the target to the 4.0 branch when the time comes?

RobbieMcKinstry:
Sorry, for the miscommunication, @dmattia . We're still noodling on ways to get this in prior to v4.0. No action is required on your part :D

alvis:
@RobbieMcKinstry @dixler My 2c, but every approach has a trade off.

On one hand, making ts-node and typescript as peer dependencies is a breaking change technically, but on the other hand, it's a breaking change only for those using npm < 7.0. Since npm@7, which comes with node 15, peer dependencies are automatically installed as normal dependencies. With node 16 already entered LTS, typescript 4.8 already on the horizon and yet pulumi 4.0 still has a long way to go.... holding super old dependencies doesn't bring any good to pulumi.

If breaking change is a concern, at least we can specify a high version of `typescipt` (e.g. `>=3.7.3 <4.8`) and `ts-node`? It doens't break anything.

Oyelowo:
Might it be possible to also add support for `ES Module` in the process?

RobbieMcKinstry:
> Might it be possible to also add support for `ES Module` in the process?

Hi @Oyelowo Would you mind opening an issue to track this separately (unless one such issue already exists)? I'm curious to hear more about the request, but I think we need to separate that conversation out from this one.

RobbieMcKinstry:
> On one hand, making ts-node and typescript as peer dependencies is a breaking change technically, but on the other hand, it's a breaking change only for those using npm < 7.0. Since npm@7, which comes with node 15, peer dependencies are automatically installed as normal dependencies. With node 16 already entered LTS, typescript 4.8 already on the horizon and yet pulumi 4.0 still has a long way to go.... holding super old dependencies doesn't bring any good to pulumi.

I certainly agree with the sentiment. However, to my knowledge, I don't think we've established a minimum supported version of NPM (or Node or TypeScript). 

**Node:** We recently hit an issue where we [upgraded the GitHub Action](https://github.com/pulumi/actions/pull/678) to use LTS Node, but we had to revert the change to support customers who were using EOL Node. I'd want to get a feel for how many customers are using old versions of runtimes before I think we could upgrade Node. 

**TypeScript:** TypeScript v4.0 introduces a breaking change to the Node language runtime. I recently tried to upgrade it myself but came across these errors. From a breaking API standpoint, it might be possible to resolve the breaks in the runtime while TypeScript is a non-peer dependency. Then, we could see about marking TypeScript as a peer dependency. However...

**NPM:** AFAIK we don't have a minimum required version of NPM, so we can't promise users aren't running npm@6 or earlier. Since we aren't able to require npm@^7, we can't guarantee peer dependencies are automatically installed by NPM. This complication makes migrating TS to a peer dependency breaking, in my eyes.

I've recently looked into other ways to upgrade TypeScript and TSNode without breaking. Just as an example workaround (I'm not suggesting this seriously, just one such example of a workaround), we could allow users who have enabled the `PULUMI_EXPERIMENTAL` environment variable to use a different `package.json` in the NodeJS runtime, allowing us to trial a `package.json` with newer tsnode and TypeScript versions.

I'd be curious to hear other thoughts on my assessment of the minimum version support for NPM, TypeScript, and Node. I think we're going to make big improvements to this process in the near future to allow the Pulumi tool to more freely upgrade dependencies and allow users to pick the tools they want to use instead of what we comes straight out of the box. But it's not totally clear to me how we'll get there.

RobbieMcKinstry:
Correction: we do list the minimum version for particular tools! We list them on the [language runtime pages](https://www.pulumi.com/docs/intro/languages/javascript/).

> Pulumi supports JavaScript programs running on Node.js using any of the [Current, Active and Maintenance LTS versions](https://nodejs.org/en/about/releases/).

Not clear if this also holds true for NPM. It appears only v7 is legacy and no longer supported. https://docs.npmjs.com/cli/v8

alvis:
> Correction: we do list the minimum version for particular tools! We list them on the [language runtime pages](https://www.pulumi.com/docs/intro/languages/javascript/).
> 
> > Pulumi supports JavaScript programs running on Node.js using any of the [Current, Active and Maintenance LTS versions](https://nodejs.org/en/about/releases/).
> 
> Not clear if this also holds true for NPM. It appears only v7 is legacy and no longer supported. https://docs.npmjs.com/cli/v8



> Correction: we do list the minimum version for particular tools! We list them on the [language runtime pages](https://www.pulumi.com/docs/intro/languages/javascript/).
> 
> > Pulumi supports JavaScript programs running on Node.js using any of the [Current, Active and Maintenance LTS versions](https://nodejs.org/en/about/releases/).
> 
> Not clear if this also holds true for NPM. It appears only v7 is legacy and no longer supported. https://docs.npmjs.com/cli/v8

The current supported engines are node 14, 16, 18. npm is always just a bundle and user can freely upgrade to a higher version without any issue. i.e. node 14 users can upgrade to npm 8 without any issue.

alvis:
> **TypeScript:** TypeScript v4.0 introduces a breaking change to the Node language runtime. I recently tried to upgrade it myself but came across these errors. From a breaking API standpoint, it might be possible to resolve the breaks in the runtime while TypeScript is a non-peer dependency. Then, we could see about marking TypeScript as a peer dependency. However...

I agree it can be a breaking change to some users. There are always some small changes to the language from time to time.

Generally it's not an issue to anyone since anyone works on a typescript project would install typescript as a devDependency, so the version is always fixed at the project's end. But somehow pulumi chose to bundle locked-version typescript & ts-node... for the benefit that the end user doesn't need to declare what they should in their devDependency... and hence the discussion we have today.

In retrospect, I'd consider the decision a mistake, but from an IaC & typescript veteran point of view, I reluctantly tend to agree that we probably shouldn't upgrade it in case there are some typescript developers don't have typescript in their devDependency.

Another way to fix it without any breaking change nor the necessity to upgrade the `typescript` version bundled in `@pulumi/pulumi` is to dynamically determine the typescript version installed in the end user project and employ the relevant `ts-node` that support that typescript version use it instead. If the end user doesn't have `typescript` installed, we can use the fallback -- version 3.7.

RobbieMcKinstry:
Hi Alvis,

> But somehow pulumi chose to bundle locked-version typescript & ts-node... In retrospect, I'd consider the decision a mistake.

I'm inclined to agree with you here! I bet this decision was an oversight spawned from early iteration.

> Another way to fix it without any breaking change nor the necessity to upgrade the typescript version bundled in @pulumi/pulumi is to dynamically determine the typescript version installed in the end user project and employ the relevant ts-node that support that typescript version use it instead. If the end user doesn't have typescript installed, we can use the fallback -- version 3.7.

That's a great idea! Unless you want to take a stab at it, I'll explore this next week and report back. :)

dmattia:
@RobbieMcKinstry Any update on your exploration?

My personal view is that this just should be a breaking change that waits until v4, but if there is an interim solution that isn't breaking that would be cool. Also, any idea if a branch could be made for v4 anytime soon?

RobbieMcKinstry:
Hello! Thanks for following up on this! I appreciate the nudge :)

Unfortunately, I haven't had time to look into this yet. Sorry to have promised sooner action!

This quarter my squad has been prioritizing work that gives a guaranteed performance boost to users without needing them to opt-in or change their configurations. My current belief is this issue doesn't hit that objective as well as some other issues, so we've been prioritizing those instead (e.g. #7653). 

I still think we want to address this, but I'm not sure I'll get to it in the next month, although I'll try my best.

I'm going to ask about v4.0 and respond back.

RobbieMcKinstry:
RE: Pulumi v4, it's still not actively worked on, although we're tracking planned features [here](https://github.com/pulumi/pulumi/issues/7346). Since it's far enough off, we can't start a branch for it yet, since it would be quite a lot of effort to keep that branch up to date with the active branch of v3.

RobbieMcKinstry:
> Might it be possible to also add support for ES Module in the process?
@Oyelowo could you point me to an issue or describe in greater detail what the feature is that you're asking for? :D Thank you :)

RobbieMcKinstry:
I'm happy to say we're working on making TypeScript and TS-Node peer-dependencies this quarter 🎉 https://github.com/pulumi/pulumi/pull/10122

dmattia:
Wonderful news :)

tedvanderveen:
@RobbieMcKinstry how are you guys getting in with this PR?

tedvanderveen:
@RobbieMcKinstry Thanks for sharing!

ffMathy:
Who else is then taking over this work? Currently need this as well.

ffMathy:
@dmattia any chance you can look at the conflicts?

julienp:
ts-node and typescript are now *optional* peer dependencies. The npm package now includes bundled versions of these to fallback to if no peer dependencies are installed. If you add your own versions, these will be prioritised over the bundled versions.

https://github.com/pulumi/pulumi/pull/15622

See also https://www.pulumi.com/blog/typescript-versions/

Issue #10078, opened 2022-07-08
`pulumi new` show additional templates is unusable


joeduffy:
If you select "Show additional templates", you get a super long list, and trying to move the cursor beyond the top one doesn't scroll the screen so you ca't actually see what you're trying to select:

![image](https://user-images.githubusercontent.com/3953235/178048610-cc836179-7ad8-469b-ada8-a2d0d47d80fc.png)

This makes it basically unusable for selecting anything that doesn't fit on the screen.

iwahbe:
It looks like like this a problem with the library we use to display choice lists. We should explore choosing a new library or dumping out of interactive mode once a certain number of options is displayed.

AaronFriel:
Related, the length of this list causes awful flickering on re-rendering in PowerShell on Windows:
- #7823

iwahbe:
Is causes flickering on the default terminal on Mac too.

dixler:
I have some comments on help output here https://github.com/pulumi/home/pull/2164#discussion_r912160412__

dixler:
I'm surprised that changing the prompt library is on the table.

@iwahbe i think @aq17 was trying to tag you

AaronFriel:
@aq17 I'm OK with changing the prompt library if it flickers less. Could we do a side by side comparison?

dixler:
Can we keep the output/coloration identical to the current?

AaronFriel:
@aq17 Have you checked whether it solves the flickering issue for `pulumi new` first? If it does, then I'd say tentatively yes to replacing it everywhere.

iwahbe:
The flickering is a fundamental property of most terminal emulators (Including Terminal.app). They can only process so many new writes at once. It can be avoided in some cases by taking full control of the screen and using cursor manipulation instead of rewriting sections. I think the only solution here is to limit the number of templates we list one command.

aq17:
> The flickering is a fundamental property of most terminal emulators (Including Terminal.app). They can only process so many new writes at once. It can be avoided in some cases by taking full control of the screen and using cursor manipulation instead of rewriting sections. I think the only solution here is to limit the number of templates we list one command.
> 
> That being said, I'm not opposed to bumping v1 -> v2 if it fixes problems for us.

Thanks for confirming @iwahbe , I don't believe v1->v2 changes anything of consequence then

AaronFriel:
I think to close this out, we should pick a maximum number of templates to display and allow that to be good enough.

[...]

Issue #8169, opened 2021-10-07
pulumi new selector behaves poorly when lines wrap


svangordon-fruit:
## Hello!

- Vote on this issue by adding a 👍 reaction
- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details

The `pulumi new` template selector behaves in weird ways when the terminal is sized such that lines wrap.

### Steps to reproduce

Pulumi version: `v3.13.2` 
Terminal: `iTerm2 3.4.10` 
Shell: `zsh 5.8`

1. Check the terminal window width and height. Make it high, so you can see the weirdness. Any width that causes the lines to wrap (96 cols or less) will cause the issue.
```
❯ tput cols
80
❯ tput lines
33
```
2. Run a `pulumi new` command using the selector:
```
pulumi new --generate-only --force
```
3. Use the up/down keys to toggle between options

Expected: Selector element output is clean
Actual: Toggling between options results in duplicate `Please choose a template` prompts and multiple selector arrows.

### Duplicate prompts
```
Please choose a template:  [Use arrows to move, enter to select, type to filter]
Please choose a template:  [Use arrows to move, enter to select, type to filter]
Please choose a template:  [Use arrows to move, enter to select, type to filter]
Please choose a template:  [Use arrows to move, enter to select, type to filter]
Please choose a template:  [Use arrows to move, enter to select, type to filter]
Please choose a template:  [Use arrows to move, enter to select, type to filter]
Please choose a template:  [Use arrows to move, enter to select, type to filter]
  aws-csharp                   A minimal AWS C# Pulumi program
  aws-go                       A minimal AWS Go Pulumi program
  aws-javascript               A minimal AWS JavaScript Pulumi program
  aws-python                   A minimal AWS Python Pulumi program
  aws-typescript               A minimal AWS TypeScript Pulumi program
> azure-csharp                 A minimal Azure Native C# Pulumi program
  azure-go                     A minimal Azure Native Go Pulumi program
  azure-javascript             A minimal JavaScript Pulumi program with the native Azure provider
  azure-python                 A minimal Azure Native Python Pulumi program
  azure-typescript             A minimal Azure Native TypeScript Pulumi program
```

### Multiple selector arrows
```
  aws-csharp                   A minimal AWS C# Pulumi program
  aws-go                       A minimal AWS Go Pulumi program
  aws-javascript               A minimal AWS JavaScript Pulumi program
[...]

PR #10130, opened 2022-07-13
Paginate template options

aq17:
Addresses #10078 

## Checklist


pkg/cmd/pulumi/new.go
```diff
@@ -776,7 +776,7 @@ func chooseTemplate(templates []workspace.Template, opts display.Options) (works
 		if err := survey.AskOne(&survey.Select{
 			Message:  message,
 			Options:  options,
-			PageSize: len(options),
+			PageSize: 10,
```
AaronFriel, line 779: That's surprisingly easy.

AaronFriel:
@aq17 going to expand on this to ask some UX questions, perhaps you could find time with someone on @pulumi/user-experience team to go over options?

Some of these are "current day" problems, but a bit exacerbated by only showing 10 items.

I don't see an affordance that there are more templates after the 10th item:
image

I have to select "Show additional templates" to see all, and _then_ I see a list of additional options and the prompt appears... twice?

image

And lastly, without selecting "Show additional templates", filtering only operates on the present list:

image

iwahbe:
If I understand correctly, this should prevent us from needing to use "Show additional templates", since we can just show all templates. The user can just scroll down to see the rest.

aq17:
> If I understand correctly, this should prevent us from needing to use "Show additional templates", since we can just show all templates. The user can just scroll down to see the rest.

Great point, also an easy change 😄

dixler:
@AaronFriel 
> I have to select "Show additional templates" to see all, and _then_ I see a list of additional options and the prompt appears... twice?
> 
> 
> 
> image

https://github.com/pulumi/pulumi/issues/8169

I fixed this before by **upgrading survey by a major version** and it got reverted.

aq17:
> @AaronFriel
> 
> > I have to select "Show additional templates" to see all, and _then_ I see a list of additional options and the prompt appears... twice?
> > image
> 
> #8169
> 
> I fixed this before by **upgrading survey by a major version** and it got reverted because of major bugs

Might be worth upgrading to v2 everywhere we use the package, it doesn't look like there are too many breaking changes

pkg/cmd/pulumi/new.go
```diff
@@ -756,7 +756,7 @@ func chooseTemplate(templates []workspace.Template, opts display.Options) (works
 	message := "\rPlease choose a template:"
 	message = opts.Color.Colorize(colors.SpecPrompt + message + colors.Reset)
 
-	showAll := false
+	showAll := true
```
iwahbe, line 759: We only set `showAll` to `true`, so if it starts as `true`, we can just omit the variable.

dixler:
@aq17 can we list the number of templates that are on the list inside of the prompt? A user might not realize that the list scrolls.

pkg/cmd/pulumi/policy_new.go
```diff
@@ -297,7 +297,7 @@ func choosePolicyPackTemplate(templates []workspace.PolicyPackTemplate,
 	if err := survey.AskOne(&survey.Select{
 		Message:  message,
 		Options:  options,
-		PageSize: len(options),
+		PageSize: 10,
```
dixler, line 300: we should use the height of the user's terminal to determine the size of the page

it looks good inside vscode, but leaves a bit to be desired in a full sized terminal.
2022-07-15-165446_2560x1051_scrot
.

dixler (changes requested):

pkg/cmd/pulumi/policy_new.go
aq17: Just tested it, definitely looks better but there's the issue with the `Please choose a template (#/#) shown:` prompt appearing numerous times as you scroll again

aq17: Actually, it looks like providing a small buffer (terminal height - 5) solves that, but the flickering is 😠 
Screen Shot 2022-07-15 at 5 07 24 PM

AaronFriel: I would be inclined to set a lower limit to reduce flickering. Is there an upstream issue to track that flickering?

AaronFriel: I meant with the CLI library - is this something other folks have noticed or the upstream repo is aware of? The flickering is very noticeable - even more-so on Windows - and I think it'd be good to flag that with the maintainer.

aq17: Ah no, it's also still there even after upgrading to their v2; I'll open an issue now.

pkg/cmd/pulumi/new.go
```diff
@@ -768,6 +769,8 @@ func chooseTemplate(templates []workspace.Template, opts display.Options) (works
 		if height > len(options) {
 			height = len(options)
 		}
+
+		height = height - buffer
```
AaronFriel, line 794: If we change this height to a much smaller value, is the flickering as severe?

aq17: Don't think there's a clear maximum, even something as small as 10 on a max sized terminal causes pretty bad flickering on my mac default terminal

dixler (approved):
I think this PR does resolve the issue that it's addressing. We should :ship: it since we have another issue that tracks the flickering.

aq17:
@dixler [upstream issue](https://github.com/AlecAivazis/survey/issues/436#issuecomment-1203111689) had a few other people commenting on ideas regarding the flickering  – can we still merge this and address flickering in a separate PR?

PR #10128, opened 2022-07-13
Non-mutating snapshot normalization

t0yv0:
# Description

Fixes # (issue)

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works

- [ ] I have updated the [CHANGELOG-PENDING](https://github.com/pulumi/pulumi/blob/master/CHANGELOG_PENDING.md) file with my change

- [ ] Yes, there are changes in this PR that warrants bumping the Pulumi Service API version

PR #10137, opened 2022-07-14
fix mutation of resources by moving alias fixup

AaronFriel:
An attempt to resolve data races identified by #10060, this moves the alias fixup logic to the persistence step. Like #10128, but moving the logic further away from the snapshot. Instead, snapshots do not perform any alias fixup at all, but the Resource "v3" structs that are created in `SerializeResource` and `SerializeOperation` are fixed up incrementally in serializing the deployment.

This is a draft PR to let the larger test suite run async while the team evaluates various solutions. Apologies to @RobbieMcKinstry as I keep returning to this after thinking I've left it alone.

### DIFF
--- a/pkg/backend/display/json.go
+++ b/pkg/backend/display/json.go
@@ -1,4 +1,4 @@
-// Copyright 2016-2018, Pulumi Corporation.
+// Copyright 2016-2022, Pulumi Corporation.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -176,7 +176,7 @@ func ShowPreviewDigest(events <-chan engine.Event, done chan<- bool, opts Option
 
 				if m.Old != nil {
 					oldState := stateForJSONOutput(m.Old.State, opts)
-					res, err := stack.SerializeResource(oldState, config.NewPanicCrypter(), false /* showSecrets */)
+					res, err := stack.SerializeResource(oldState, config.NewPanicCrypter(), false /* showSecrets */, nil /* aliases */)
 					if err == nil {
 						step.OldState = &res
 					} else {
@@ -185,7 +185,7 @@ func ShowPreviewDigest(events <-chan engine.Event, done chan<- bool, opts Option
 				}
 				if m.New != nil {
 					newState := stateForJSONOutput(m.New.State, opts)
-					res, err := stack.SerializeResource(newState, config.NewPanicCrypter(), false /* showSecrets */)
+					res, err := stack.SerializeResource(newState, config.NewPanicCrypter(), false /* showSecrets */, nil /* aliases */)
 					if err == nil {
 						step.NewState = &res
 					} else {
--- a/pkg/resource/stack/deployment.go
+++ b/pkg/resource/stack/deployment.go
@@ -1,4 +1,4 @@
-// Copyright 2016-2018, Pulumi Corporation.
+// Copyright 2016-2022, Pulumi Corporation.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -24,6 +24,7 @@ import (
 	"strings"
 
 	"github.com/pulumi/pulumi/pkg/v3/resource/deploy"
+	"github.com/pulumi/pulumi/pkg/v3/resource/deploy/providers"
 	"github.com/pulumi/pulumi/pkg/v3/secrets"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/apitype"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/apitype/migrate"
@@ -101,6 +102,10 @@ func ValidateUntypedDeployment(deployment *apitype.UntypedDeployment) error {
 func SerializeDeployment(snap *deploy.Snapshot, sm secrets.Manager, showSecrets bool) (*apitype.DeploymentV3, error) {
 	contract.Require(snap != nil, "snap")
 
+	if snap.Manifest.Magic != snap.Manifest.NewMagic() {
+		return nil, fmt.Errorf("magic cookie mismatch; possible tampering/corruption detected")
+	}
+
 	// Capture the version information into a manifest.
 	manifest := snap.Manifest.Serialize()
 
@@ -120,10 +125,24 @@ func SerializeDeployment(snap *deploy.Snapshot, sm secrets.Manager, showSecrets 
 		enc = config.NewPanicCrypter()
 	}
 
+	aliases := make(aliases)
+	for _, state := range snap.Resources {
+		// Add to aliased maps
+		for _, alias := range state.Aliases {
+			// For ease of implementation, some SDKs may end up creating the same alias to the
+			// same resource multiple times.  That's fine, only error if we see the same alias,
+			// but it maps to *different* resources.
+			if otherUrn, has := aliases[alias]; has && otherUrn != state.URN {
+				return nil, fmt.Errorf("Two resources ('%s' and '%s') aliased to the same: '%s'", otherUrn, state.URN, alias)
+			}
+			aliases[alias] = state.URN
+		}
+	}
+
 	// Serialize all vertices and only include a vertex section if non-empty.
 	var resources []apitype.ResourceV3
 	for _, res := range snap.Resources {
-		sres, err := SerializeResource(res, enc, showSecrets)
+		sres, err := SerializeResource(res, enc, showSecrets, aliases)
 		if err != nil {
 			return nil, fmt.Errorf("serializing resources: %w", err)
 		}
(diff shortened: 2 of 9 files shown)


pkg/resource/stack/deployment.go
```diff
@@ -120,10 +125,24 @@ func SerializeDeployment(snap *deploy.Snapshot, sm secrets.Manager, showSecrets
...
+			// For ease of implementation, some SDKs may end up creating the same alias to the
+			// same resource multiple times.  That's fine, only error if we see the same alias,
+			// but it maps to *different* resources.
+			if otherUrn, has := aliases[alias]; has && otherUrn != state.URN {
+				return nil, fmt.Errorf("Two resources ('%s' and '%s') aliased to the same: '%s'", otherUrn, state.URN, alias)
+			}
+			aliases[alias] = state.URN
+		}
+	}
+
```
RobbieMcKinstry, line 141: Is it possible to pull this into a method? Perhaps something like `snap.GenerateAliasesMap()`. Separating the logic might make it easier to unit test, and it appears this block is total.

RobbieMcKinstry (commented):
Just a couple of comments and questions!

RobbieMcKinstry:
@AaronFriel We believe this approach is sufficient to resolve #10060, right? Just confirming you're not expecting an additional PR from me.

pkg/resource/stack/deployment.go
```diff
@@ -101,6 +102,10 @@ func ValidateUntypedDeployment(deployment *apitype.UntypedDeployment) error {
 func SerializeDeployment(snap *deploy.Snapshot, sm secrets.Manager, showSecrets bool) (*apitype.DeploymentV3, error) {
 	contract.Require(snap != nil, "snap")
 
+	if snap.Manifest.Magic != snap.Manifest.NewMagic() {
+		return nil, fmt.Errorf("magic cookie mismatch; possible tampering/corruption detected")
+	}
+
```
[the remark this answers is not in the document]
AaronFriel: This was copied from snapshot.go

pkg/resource/stack/deployment.go
```diff
@@ -332,8 +478,9 @@ func SerializeResource(res *resource.State, enc config.Encrypter, showSecrets bo
 	return v3Resource, nil
 }
 
-func SerializeOperation(op resource.Operation, enc config.Encrypter, showSecrets bool) (apitype.OperationV2, error) {
-	res, err := SerializeResource(op.Resource, enc, showSecrets)
+func SerializeOperation(
+	op resource.Operation, enc config.Encrypter, showSecrets bool, aliases aliases) (apitype.OperationV2, error) {
```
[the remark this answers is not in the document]
AaronFriel: Trying to make the linter happy. Definitely fixable.

AaronFriel:
@RobbieMcKinstry I think so - but pending discussion with @t0yv0 and the team.

t0yv0:
Happy to discuss today - @RobbieMcKinstry I think you have the most context on the user issue so I'd index on your opinion which fix is the most suitable. The approach here works for me but let me review and point out a few small worries.

pkg/backend/display/json.go
```diff
@@ -176,7 +176,7 @@ func ShowPreviewDigest(events <-chan engine.Event, done chan<- bool, opts Option
 
 				if m.Old != nil {
 					oldState := stateForJSONOutput(m.Old.State, opts)
-					res, err := stack.SerializeResource(oldState, config.NewPanicCrypter(), false /* showSecrets */)
+					res, err := stack.SerializeResource(oldState, config.NewPanicCrypter(), false /* showSecrets */, nil /* aliases */)
```
t0yv0, line 179: Is this `nil` allright? It's under `display/show` so not as worried semantically but will it change how things are displayed?

pkg/resource/stack/deployment.go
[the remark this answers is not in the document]
t0yv0: I've read the implementation and it's confusing but I guess we don't have to dig into that now. `m.NewMagic()` is sha256 of `m.Version` which is described as `the pulumi command version` so maybe this prevents Manifest entries created by different CLI versions from interoperating but it's not clear.

pkg/resource/stack/deployment.go
```diff
@@ -153,12 +172,95 @@ func SerializeDeployment(snap *deploy.Snapshot, sm secrets.Manager, showSecrets
...
-	return &apitype.DeploymentV3{
+	deployment := &apitype.DeploymentV3{
 		Manifest:          manifest,
 		Resources:         resources,
 		SecretsProviders:  secretsProvider,
 		PendingOperations: operations,
-	}, nil
+	}
+
+	err := VerifyIntegrity(deployment)
```
t0yv0, line 182: This is my biggest worry.


Context:

```
// saveSnapshot persists the current snapshot and optionally verifies it afterwards.
func (sm *SnapshotManager) saveSnapshot() error {
	snap := sm.snap()
	if err := snap.NormalizeURNReferences(); err != nil {
		return fmt.Errorf("failed to normalize URN references: %w", err)
	}
	if err := sm.persister.Save(snap); err != nil {
		return fmt.Errorf("failed to save snapshot: %w", err)
	}
	if sm.doVerify {
		if err := snap.VerifyIntegrity(); err != nil {
			return fmt.Errorf("failed to verify snapshot: %w", err)
		}
	}
	return nil
}

func (persister *cloudSnapshotPersister) Save(snapshot *deploy.Snapshot) error {
	token, err := persister.tokenSource.GetToken()
	if err != nil {
		return err
	}
	deployment, err := stack.SerializeDeployment(snapshot, persister.sm, false /* showSecrets */)
	if err != nil {
		return fmt.Errorf("serializing deployment: %w", err)
	}
	return persister.backend.client.PatchUpdateCheckpoint(persister.context, persister.update, deployment, token)
}
```

Before the changes, an error in VerifyIntegrity happens after
`sm.persister.Save` succeeds to persist the snapshot.

After the changes, an error in the equivalent of VerifyIntegrity fails
stack.SerializeDeployment so PatchUpdateCheckpoint does not run.

We were discussing this with @iwahbe the other day and we think that this previous behavior was very intentional - we intended to save the state even if it's not verified, as it might have info about what we did to the cloud (even if it was corrupted) and can assist recovery.

t0yv0: Is this change instrumental to what we're doing here with alias rewriting or can it be considered as a separate PR?

pkg/resource/stack/deployment.go
```diff
@@ -153,12 +172,95 @@ func SerializeDeployment(snap *deploy.Snapshot, sm secrets.Manager, showSecrets
...
+// snapshot construction.
+//
+// This function verifies a number of invariants:
+//  1. Provider resources must be referenceable (i.e. they must have a valid URN and ID)
+//  2. A resource's provider must precede the resource in the resource list
+//  3. Parents must precede children in the resource list
+//  4. Dependents must precede their dependencies in the resource list
+//  5. For every URN in the snapshot, there must be at most one resource with that URN that is not pending deletion
+//  6. The magic manifest number should change every time the snapshot is mutated
+func VerifyIntegrity(deployment *apitype.DeploymentV3) error {
```
t0yv0, line 201: This is exactly like snap.VerifyIntegrity() but on *apitype.DeploymentV3 type?

pkg/resource/stack/deployment.go
t0yv0: Perhaps an alternative would be to do `snap.VerifyIntegrity(aliases)` and have that "auto-apply" the alias rewrite as part of verifying. Without changing the control flow.

t0yv0:
Sanity check on call sites of SerializeDeployment that's changing. Are these OK? httpstate/filestate call sites are def ok.

```
pkg/cmd/pulumi/stack_change_secrets_provider.go:        reserializedDeployment, err := stack.SerializeDeployment(snap, newSecretsManager, false /*showSecrets*/)
pkg/cmd/pulumi/stack_export.go:                         serializedDeployment, err := stack.SerializeDeployment(snap, snap.SecretsManager, true)
pkg/cmd/pulumi/stack_import.go:                 sdp, err := stack.SerializeDeployment(snapshot, snapshot.SecretsManager, false /* showSecrets */)
pkg/cmd/pulumi/state.go:        sdep, err := stack.SerializeDeployment(snap, snap.SecretsManager, false /* showSecrets */)
pkg/resource/stack/checkpoint.go:               dep, err := SerializeDeployment(snap, sm, showSecrets)
```

state.go seems not ok since `!stackIsAlreadyHosed` is not going to prevent VerifyIntegrity anymore for example:

```
	// If the stack is already broken, don't bother verifying the integrity here.
	if !stackIsAlreadyHosed {
		contract.AssertNoErrorf(snap.VerifyIntegrity(), "state edit produced an invalid snapshot")
	}

	sdep, err := stack.SerializeDeployment(snap, snap.SecretsManager, false /* showSecrets */)
	if err != nil {
		return result.FromError(fmt.Errorf("serializing deployment: %w", err))
	}
```

Same goes for `func (b *localBackend) saveStack`. It calls stack.SerializeCheckpoint which calls SerializeDeployment. But then this logic is busted:

```
	if !DisableIntegrityChecking {
		// Finally, *after* writing the checkpoint, check the integrity.  This is done afterwards so that we write
		// out the checkpoint file since it may contain resource state updates.  But we will warn the user that the
		// file is already written and might be bad.
		if verifyerr := snap.VerifyIntegrity(); verifyerr != nil {
			return "", fmt.Errorf(
				"%s: snapshot integrity failure; it was already written, but is invalid (backup available at %s): %w",
				file, bck, verifyerr)

		}
	}
```

Issue #10139, opened 2022-07-14
GitHub authentication headers can leak to logs


Frassle:
### What happened?

When installing a Pulumi plugin (custom provider) that's hosted on a private Github repository, a Github Token is required. That token is inserted as a header in the GET query for Github to authenticate the request.

However, if the CLI is set to verbosity level is set to 9 or above *and* the plugin hasn't been installed yet, then `plugins.go` will leak the Github token into the logs as part of logging the request headers.

### Steps to reproduce

Download a plugin from github with GITHUB_TOKEN set.

### Expected Behavior

Log messages to not include the actual GITHUB_TOKEN.

### Actual Behavior

Log message includes the token:
```
I0713 14:54:46.243492    8779 plugins.go:750] plugin install request headers: map[Accept:[application/json] Authorization:[token ghp_Ys1f<REDACTED>] User-Agent:[pulumi-cli/1 (; linux)]]
```

### Versions used

3.35

### Additional context

This was introduced in https://github.com/pulumi/pulumi/pull/9185.

We should either blind the Authorization before writing to logs, or move this to log level 11 which intentionally leaks credentials to help with debugging.

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

aureq:
@Frassle I'm reopening the issue as I think we bring added clarity to the end-user.

One thing that wasn't obvious and as it was reported by @contrarianchris is that setting the log level doesn't mean the user knows the implications of doing so.

As a possible improvements, I think the CLI should display a clear (yellow colored?) warning when the log level is 10 or above, and clear state that such level may reveal sensitive information such as token and request headers.

I also think we should update our documentation https://www.pulumi.com/docs/support/troubleshooting/#verbose-logging to reflect the changes as well.

```
These logs may include sensitive information that is provided from your execution environment to your cloud provider (and which Pulumi may not even itself be aware of) so be careful to audit before sharing.
```

Frassle:
> I also think we should update our documentation https://www.pulumi.com/docs/support/troubleshooting/#verbose-logging to reflect the changes as well.

Yup Luke had also made this comment on the PR and I'd made a note to myself to change that today, but I do like the explicit warning at startup as well. I'll add that.

PR #10140, opened 2022-07-14
Only log request/response headers at level 11

Frassle:
# Description

Fixes https://github.com/pulumi/pulumi/issues/10139

## Checklist


sdk/go/common/workspace/plugins.go
```diff
@@ -747,14 +747,18 @@ func buildHTTPRequest(pluginEndpoint string, token string) (*http.Request, error
 
 func getHTTPResponse(req *http.Request) (io.ReadCloser, int64, error) {
 	logging.V(9).Infof("full plugin download url: %s", req.URL)
-	logging.V(9).Infof("plugin install request headers: %v", req.Header)
+	// This logs at level 11 because it could include authentication headers, we reserve log level 11 for
```
dixler, line 750: Docs? Would love to see what else is in it.

Frassle: We don't really have docs for this, but the sass api client also logs headers at level 11.

Frassle: I'll update that

PR #10146, opened 2022-07-14
Plugin Link

harribo11:
This PR implements Plugin Linking. Using this feature, you can link to local plugin binaries in your `pulumi.yaml` file as follows:

```
plugins:
  providers:
  - name: aws
    path: ../.../bin
  languages:
  - name: yaml
     path: ../../../pulumi-yaml/bin
     version: 1.2.3
```
`plugins` accepts three PluginInfo arrays named `providers`, `languages`, and `analyzer`. A PluginInfo takes a `name`, a `path`, and a `version`. The path links to a directory containing a binary `pulumi-kind-name`.

Manually linked plugins supersede plugins installed at the default location and plugins installed on PATH. During integration testing, when copying files into a temporary directory, relative paths to linked plugins will be replaced by absolute ones.

This PR adds the function `GetProjectPlugins() []*workspace.PluginInfo` to `plugin.Host`. This function is implemented in `plugin.DefaultHost` by adding an argument `workspace.Plugins`. Outside of testing, this is always a field of Project, so we additionally make modifications to allow the field to be passed in.


sdk/go/common/workspace/project.go
```diff
@@ -63,6 +63,10 @@ type ProjectOptions struct {
 	Refresh string `json:"refresh,omitempty" yaml:"refresh,omitempty"`
 }
 
+type ProviderOptions struct {
+	LocalPath string `json:"localPath,omitempty" yaml:"localPath,omitempty"`
```
Frassle, line 67: Probably want to just call this "path"

sdk/go/common/workspace/plugins.go
```diff
@@ -1173,9 +1173,28 @@ func GetPluginInfo(kind PluginKind, name string, version *semver.Version) (*Plug
 //  * if found in the pulumi dir's installed plugins, a PluginInfo and path to the executable
 //  * an error in all other cases.
 func getPluginInfoOrPath(
-	kind PluginKind, name string, version *semver.Version, skipMetadata bool) (*PluginInfo, string, error) {
+	kind PluginKind, name string, version *semver.Version, skipMetadata bool, projectPlugins map[string]*PluginInfo) (*PluginInfo, string, error) {
 	var filename string
 
+	if projectPlugins != nil {
+		plugin, ok := projectPlugins[name]
```
Frassle, line 1180: Not sure we can be as simple as a map keyed by name. For one there are different kinds of plugins and it would be good to have both a "resource-aws" and "analyser-aws". Secondly we need to think how to handle version when using "link". @danielrbradley might have better ideas here on if link should be treated as versionless, or if there's a use case for being able to specify both aws v4  and aws v5 in one project (that sounds reasonable to me).

sdk/go/common/workspace/project.go
```diff
@@ -102,6 +106,8 @@ type Project struct {
 
 	// Options is an optional set of project options
 	Options *ProjectOptions `json:"options,omitempty" yaml:"options,omitempty"`
+
+	Providers map[string]*ProviderOptions `json:"providers,omitempty" yaml:"providers,omitempty"`
```
Frassle, line 110: Again based on how we handle version, this might end up being a list rather than a map. 
```
providers:
- name: aws
  version: 4.0
  path: ../../awsv4/bin
- name: aws
  version: 5.0
  path: ../../awsv5/bin
```
Again @danielrbradley to comment

sdk/go/common/workspace/plugins.go
```diff
@@ -1173,9 +1173,28 @@ func GetPluginInfo(kind PluginKind, name string, version *semver.Version) (*Plug
...
+		if ok {
+			if plugin.Name != name {
+				return nil, "", fmt.Errorf("plugin name mismatch: %s != %s", plugin.Name, name)
+			}
+			if plugin.Kind != kind {
+				return nil, "", fmt.Errorf("plugin kind mismatch: %s != %s", plugin.Kind, kind)
+			}
+			//Comparing pointers? Seems wrong, but I get problems when I try to dereference
```
Frassle, line 1188: Yeh either version can be nil so you need to check for that first. If you look down around line 1280  there's some "if nil" checks before calling Version.GTE()

danielrbradley:
My two cents on @Frassle's feedback...

I don't think differentiating by major version is a must - for the scenario of local development it's pretty safe to say that we are only working with the single current version of the plugin.

I agree on making the plugins a list is a good shout.

Looks like good progress all-round!

harribo11:
I'm not sure how we're going to pass the Project information into the language providers. I think they may have to read it separately.

sdk/go/common/resource/plugin/host.go
```diff
@@ -88,7 +91,42 @@ type Host interface {
 
 // NewDefaultHost implements the standard plugin logic, using the standard installation root to find them.
 func NewDefaultHost(ctx *Context, config ConfigSource, runtimeOptions map[string]interface{},
-	disableProviderPreview bool) (Host, error) {
```
Frassle, line 91: actually given the only thing this is pulling from project is the providers map I think it would better to just pass that map as the parameter, decouples this from Projects and makes it a bit easier to set in tests.

sdk/go/common/workspace/project.go
```diff
@@ -63,6 +63,13 @@ type ProjectOptions struct {
 	Refresh string `json:"refresh,omitempty" yaml:"refresh,omitempty"`
 }
 
+type ProviderOptions struct {
```
Frassle, line 66: rename to PluginOptions

sdk/go/common/workspace/project.go
```diff
@@ -102,6 +109,8 @@ type Project struct {
 
 	// Options is an optional set of project options
 	Options *ProjectOptions `json:"options,omitempty" yaml:"options,omitempty"`
+
+	Providers []*ProviderOptions `json:"providers,omitempty" yaml:"providers,omitempty"`
```
Frassle, line 113: No need for pointer here, we don't expect anything to be nil in this array.
```suggestion
	Providers []ProviderOptions `json:"providers,omitempty" yaml:"providers,omitempty"`
```

sdk/go/common/resource/plugin/context.go
```diff
@@ -50,13 +52,22 @@ func NewContext(d, statusD diag.Sink, host Host, cfg ConfigSource,
 	parentSpan opentracing.Span) (*Context, error) {
 
 	root := ""
-	return NewContextWithRoot(d, statusD, host, cfg, pwd, root, runtimeOptions, disableProviderPreview, parentSpan)
+	projPath, err := workspace.DetectProjectPath()
```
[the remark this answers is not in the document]
Frassle: Urgh yaml uses this 😭 I'm not sure it's safe to error on lack of project here so I'd just try and load the project if you can, but don't error if it fails.

sdk/go/common/workspace/plugins.go
```diff
@@ -1266,9 +1276,42 @@ func GetPluginInfo(kind PluginKind, name string, version *semver.Version) (*Plug
...
+
+	for _, plugin := range projectPlugins {
+		if plugin.Kind != kind {
+			continue
+		}
+		if plugin.Name != name {
+			continue
+		}
+		if plugin.Version != nil && version != nil {
+			if !plugin.Version.Equals(*version) {
```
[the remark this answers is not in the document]
harribo11: I think that is the current behavior? It `continue`s only if the plugin version and requested version are both non-nil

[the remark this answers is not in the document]
Frassle: Hmm must of tripped over something else when I didn't have version set, your right this looks like it should handle it.

sdk/go/common/workspace/plugins.go
```diff
@@ -1219,13 +1222,19 @@ func getPlugins(dir string, skipMetadata bool) ([]PluginInfo, error) {
...
-	info, path, err := getPluginInfoOrPath(kind, name, version, true /* skipMetadata */)
+func GetPluginPath(kind PluginKind, name string, version *semver.Version,
+	projectPlugins []*PluginInfo) (string, string, error) {
+	info, path, err := getPluginInfoOrPath(kind, name, version, true /* skipMetadata */, projectPlugins)
 	if err != nil {
 		return "", "", err
 	}
 
+	//This function is caled GetPluginPath. It calls getPluginInfoOrPath, which returns info, path.
+	//But path *is not used* in this function when it returns correctly. What gives?
```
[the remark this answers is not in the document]
AaronFriel: The shared logic is in `getPluginInfoOrPath` because it performs resolution for ambient and installed plugins, and may not always return both.

sdk/go/common/workspace/project.go
```diff
@@ -109,7 +109,9 @@ type Project struct {
 	// Options is an optional set of project options
 	Options *ProjectOptions `json:"options,omitempty" yaml:"options,omitempty"`
 
-	Providers []PluginOptions `json:"providers,omitempty" yaml:"providers,omitempty"`
+	Providers       []PluginOptions `json:"providers,omitempty" yaml:"providers,omitempty"`
+	LanguagePlugins []PluginOptions `json:"languagePlugins,omitempty" yaml:"languagePlugins,omitempty"`
```
Frassle, line 113: I think we should be consistent, and either name all of these Plugins, or just call this languages. Probably the former, these are relatively advanced use fields, it's ok for them to have longer names.

sdk/go/common/resource/plugin/host.go
Frassle: ProjectPlugins struct with Languages/Providers/Analyzers?
Might even make sense for the yaml to look like that as well?
```
plugins:
  providers:
  - name: aws
    path: ../../bin
```
?

sdk/go/common/workspace/plugins.go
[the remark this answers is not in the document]
Frassle: It _always_ returns path.

sdk/go/common/workspace/loaders.go
```diff
@@ -99,6 +99,7 @@ func (singleton *projectLoader) load(path string) (*Project, error) {
 		return nil, err
 	}
 
+	//TODO: configure project.Plugins
```
Frassle, line 102: Anything actually need doing here? I think the plain unmarshal covers it.

sdk/go/common/workspace/project.go
```diff
@@ -63,6 +63,18 @@ type ProjectOptions struct {
 	Refresh string `json:"refresh,omitempty" yaml:"refresh,omitempty"`
 }
 
+type PluginOptions struct {
+	Name    string `json:"name" yaml:"name"`
+	Version string `json:"version" yaml:"version"`
```
Frassle, line 68: I think this could have "omitempty" added to it for Version

sdk/go/common/resource/plugin/context.go
```diff
@@ -50,13 +51,22 @@ func NewContext(d, statusD diag.Sink, host Host, cfg ConfigSource,
...
+	return NewContextWithRoot(d, statusD, host, cfg, pwd, root, runtimeOptions,
+		disableProviderPreview, parentSpan, project)
 }
 
-// Variation of NewContext that also sets known project Root.
+// Variation of NewContext that also sets known project Root. Additionally accepts Project
 func NewContextWithRoot(d, statusD diag.Sink, host Host, cfg ConfigSource,
 	pwd, root string, runtimeOptions map[string]interface{}, disableProviderPreview bool,
-	parentSpan opentracing.Span) (*Context, error) {
+	parentSpan opentracing.Span, project *workspace.Project) (*Context, error) {
```
Frassle, line 69:
```suggestion
	parentSpan opentracing.Span, projectPlugins []*workspace.PluginInfo) (*Context, error) {
```

Given otherwise you have to deal with "project" being nil below anyway.

Frassle (approved):

PR #10149, opened 2022-07-14
[ci] Reduce PR cycle time, MacOS runner demand

AaronFriel:
Run a sampling of integration tests on MacOS and Windows only, and adds that same set of tests to Linux.

This changes the number of test jobs:
* MacOS from 7 to 1
* Windows from 7 to 1
* Linux from 7 to 8

The build jobs which run on MacOS are not changed.

Each commit should be descriptive of its content. The change to the goreleaser configs aims to reduce duplication and aid brevity. The biggest change to goreleaser configuration is the removal of the separate "build IDs" and archives for nix-like and Windows. Instead, the `prepare-for-goreleaser` script prepares `./bin/${GOOS}` directories for the final assembly of extra files into the archives. This simplifies the archive config to the following:

```yaml
archives:
- wrap_in_directory: pulumi
  format_overrides:
    - goos: windows
      format: zip
  replacements:
    amd64: x64
  files:
    - src: bin/{{ .Os }}/*
      dst: '.'
      strip_parent: true
  name_template: "{{ .ProjectName }}-{{ .Tag }}-{{ .Os }}-{{ .Arch }}"
```

The `prepare-for-goreleaser` script takes an optional GOOS argument now, which if set will put _only_ that target's dependencies in `./bin`.

Attempts to further alter or simplify GoReleaser config were not successful - there isn't an easy way to run a "single target" step that produces the final archives in multiple jobs. An upstream discussion was created here: https://github.com/goreleaser/goreleaser/discussions/3242

### DIFF
--- a/.github/workflows/test-windows-cron.yml
+++ b/.github/workflows/test-windows-cron.yml
@@ -19,7 +19,6 @@ jobs:
     name: Test Windows
     needs: build
     uses: ./.github/workflows/test.yml
-    if: github.event_name == 'repository_dispatch' || github.event.pull_request.head.repo.full_name == github.repository
     with:
       enable-coverage: true
       platform: windows-latest
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -236,21 +236,19 @@ jobs:
         with:
           name: pulumi-${{ runner.os }}-${{ runner.arch }}
           path: artifacts/go
-      - name: Install pulumi-language-python-exec
-        run: |
-          mkdir -p bin
-          cp sdk/python/cmd/pulumi-language-python-exec bin/
       - name: Install Pulumi Go Binaries
         run: |
+          echo "Running prep script to put artifacts in bin"
+          ./scripts/prep-for-goreleaser.sh $(go env GOOS)
+
           echo "Checking contents of artifacts/go"
           find artifacts/go
-          mkdir -p bin
           mv artifacts/go/pulumi*/* $PWD/bin/
           chmod a+x $PWD/bin/*
+
           echo "Checking contents of $PWD/bin"
           find $PWD/bin
-      - name: Add ./bin to PATH
-        run: |
+
           LOCAL_PATH=$(./scripts/normpath "${{ github.workspace }}/bin")
           echo "Adding LOCAL_PATH=$LOCAL_PATH to PATH"
           echo $LOCAL_PATH >> $GITHUB_PATH
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -4,6 +4,7 @@ project_name: pulumi
 
 before:
   hooks:
+    - ./scripts/get-language-providers.sh
     - ./scripts/prep-for-goreleaser.sh
 
 blobs:
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -11,180 +11,58 @@ blobs:
 - bucket: get.pulumi.com
   folder: releases/sdk/
   ids:
-    - pulumi-unix
-    - pulumi-windows
+    - pulumi
   provider: s3
   region: us-west-2
 
 builds:
-- id: pulumi-unix
+- &pulumibin
+  id: pulumi
   binary: pulumi
   builder: prebuilt
-  goos: ['linux', 'darwin']
+  goos: ['linux', 'darwin', 'windows']
   goarch: ['amd64', 'arm64']
   goamd64: ['v1']
+  ignore:
+    - goos: windows
+      goarch: arm64
   prebuilt:
-    path: goreleaser-{{ .Os }}/pulumi_{{ .Os }}_{{ .Arch }}{{ with .Amd64 }}_{{ . }}{{ end }}/pulumi
+    path: goreleaser-{{ .Os }}/{{ .Name }}_{{ .Os }}_{{ .Arch }}{{ with .Amd64 }}_{{ . }}{{ end }}/{{ .Name }}{{ .Ext }}
   mod_timestamp: '{{ .CommitTimestamp }}'
-- id: pulumi-windows
-  binary: pulumi
-  builder: prebuilt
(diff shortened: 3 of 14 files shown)


.goreleaser.yml
```diff
@@ -4,186 +4,65 @@ project_name: pulumi
...
-  format: zip
   files:
-    - pulumi-resource-pulumi-nodejs.cmd
-    - pulumi-resource-pulumi-python.cmd
-    - pulumi-python3-shim.cmd
-    - pulumi-python-shim.cmd
-    - pulumi-analyzer-policy.cmd
-    - pulumi-analyzer-policy-python.cmd
-    - pulumi-language-python-exec
+    - src: bin/{{ .Os }}/*
```
t0yv0, line 63: Curious, was it not less error prone to specify all these scripts by name? Are we now sure they are actually included in the release build? If something goes per-shaped will they accidentally slide out of the release? Or that was the case before anyway?

.goreleaser.yml
```diff
@@ -4,186 +4,65 @@ project_name: pulumi
...
   ids:
-    - pulumi-unix
-    - pulumi-windows
+    - pulumi
   provider: s3
   region: us-west-2
 
 builds:
-- id: pulumi-unix
+- &pulumibin
```
t0yv0, line 19: Clever!

scripts/prep-for-goreleaser.sh
```diff
@@ -1,22 +1,53 @@
...
 
+# When run by goreleaser, FILTER_OS and DEST_DIR are blank, so files are installed in:
+#
+# * ./bin/darwin
+# * ./bin/linux
+# * ./bin/windows
+#
+# Allowing us to customize the archives for each.
+#
+# When run by GitHub Actions in tests, we set the first arg, FILTER_OS, so that
```
t0yv0, line 11: Clever.. Wondering if way too clever, perhaps could've forked the script instead. But I'm ok with this I appreciate where the need for this munging is coming from.

.github/workflows/test.yml
```diff
@@ -236,21 +236,19 @@ jobs:
         with:
           name: pulumi-${{ runner.os }}-${{ runner.arch }}
           path: artifacts/go
-      - name: Install pulumi-language-python-exec
-        run: |
-          mkdir -p bin
-          cp sdk/python/cmd/pulumi-language-python-exec bin/
       - name: Install Pulumi Go Binaries
         run: |
+          echo "Running prep script to put artifacts in bin"
```
t0yv0, line 241: A bit confused so wanted to clarify:

- this will install from sdk/python/cmd/pulumi-language-python-exec source to bin/ as before

- this will NOT use the "released" copy that got included in the goreleaser output during build.yml
   - because the scripts are not modeled as "binaries"

.github/workflows/test-windows-cron.yml
```diff
@@ -19,7 +19,6 @@ jobs:
     name: Test Windows
     needs: build
     uses: ./.github/workflows/test.yml
-    if: github.event_name == 'repository_dispatch' || github.event.pull_request.head.repo.full_name == github.repository
```
t0yv0, line 22: 👍

t0yv0 (approved):
This all looks good just have some questions above!

.github/workflows/test.yml
AaronFriel: Correct, but the change to GoReleaser here ensures this folder will be populated the same as the archives are prepared:


```yaml
archives:
  ...
  files:
    - src: bin/{{ .Os }}/*
```

As long as this is load bearing, we can't add a file to our release tarball without adding it to the tests PATH.

scripts/get-language-providers.sh
```diff
@@ -3,48 +3,58 @@
 set -eo pipefail
 set -x
 
+get_version() {
+  repo="$1"
+  (
+    cd pkg
+    GOWORK=off go list -m all | grep "${repo}" | cut -d" " -f2
```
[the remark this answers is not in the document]
AaronFriel: How did you get that to work with a relative path!?!??!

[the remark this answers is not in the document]
AaronFriel: I think there's something fishy about how `go list` works and it's not reliable when using path arguments. Using go1.18.4 I cannot reproduce your output:

```
❯ go list -m github.com/pulumi/pulumi-java/pkg ~/c/github.com/pulumi/pulumi/pkg
go: module /home/friel/c/github.com/pulumi/pulumi/pkg: not a known dependency

❯ go list -m github.com/pulumi/pulumi-java/pkg ./pkg
go: cannot use relative path ./pkg to specify module

❯ go list -m github.com/pulumi/pulumi-java/pkg $(realpath ./pkg); echo $(realpath pkg)
go: module /home/friel/c/github.com/pulumi/pulumi/pkg: not a known dependency
/home/friel/c/github.com/pulumi/pulumi/pkg
```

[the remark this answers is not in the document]
AaronFriel: When I disable workspaces via `GOWORK=off`:

```
❯ GOWORK=off go list -m github.com/pulumi/pulumi-java/pkg ./pkg
go: cannot match "github.com/pulumi/pulumi-java/pkg" without -versions or an explicit version: go.mod file not found in current directory or any parent directory; see 'go help modules'

❯ GOWORK=off go list -m github.com/pulumi/pulumi-java/pkg $(realpath pkg)
go: cannot match "github.com/pulumi/pulumi-java/pkg" without -versions or an explicit version: go.mod file not found in current directory or any parent directory; see 'go help modules'
```

.goreleaser.yml
AaronFriel: We use separate names for Windows vs Linux and MacOS.

This replaces using separate archive specifications for each, and instead pushes the prep work into a shared script. Allows us to name these files in one place (the script) and not in two places (the script **and** GoReleaser config).

PR #10155, opened 2022-07-15
Support PULUMI_PYTHON_CMD with venvs

Frassle:
# Description

When not using a virtual environment we look at the PULUMI_PYTHON_CMD to control what python binary to run (defaulting to looking for "python" or "python3" if it's not set).

This adds a check for this variable when using a virtual environment allowing a user to change from "python", if for example they want to force something like "python39".

## Checklist


Zaid-Ajaj (approved):
LGTM

Issue #10052, opened 2022-07-06
Document that the `As<Type>` token functions should not be used


Frassle:
The `As<Type>` functions in the tokens module assert that the input string is of the expected format. 

We have historically not done any validation of user inputs for these types and so there are many cases of user code in the wild that do not match our documented specifications for names. We can not start asserting these inputs are valid without breaking some users. 

While we may do this work for 4.0, for now it is safer to just be careful not to inadvertently add asserts for these.

We can't just remove these functions as their part of sdk/ which is treated as a public API and they may be used in other codebases.

Frassle:
N.B we should if possible just remove these functions, but we need to check if they are used by any other codebases first.

PR #10156, opened 2022-07-15
Remove As[Q]Name functions

Frassle:
# Description

Doesn't look like these are used by anything, so preferring to just remove them rather than documenting them as dangerous.

Fixes https://github.com/pulumi/pulumi/issues/10052

## Checklist

### DIFF
--- a/pkg/resource/deploy/source_eval.go
+++ b/pkg/resource/deploy/source_eval.go
@@ -350,7 +350,7 @@ func (d *defaultProviders) handleRequest(req providers.ProviderRequest) (provide
 	}
 	if denyCreation {
 		logging.V(5).Infof("denied default provider request for package %s", req)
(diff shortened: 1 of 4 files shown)


pkg/resource/deploy/source_eval.go
```diff
@@ -350,7 +350,7 @@ func (d *defaultProviders) handleRequest(req providers.ProviderRequest) (provide
 	}
 	if denyCreation {
 		logging.V(5).Infof("denied default provider request for package %s", req)
-		return providers.NewDenyDefaultProvider(tokens.AsQName(string(req.Package().Name()))), nil
+		return providers.NewDenyDefaultProvider(tokens.QName(string(req.Package().Name()))), nil
```
Frassle, line 353: This would have asserted if we'd ever been passed a package name that wasn't a valid qname.

Zaid-Ajaj (approved):

PR #10159, opened 2022-07-15
WIP on moving DeepCopy to right-before-channel-send

t0yv0:
# Description

`engine.Event` had `Payload()` accessor apply a deep copy. But in cases where *ResourceState is embedded in the `Payload()` this may have caused panics over map races:

- owner (engine) is editing a map inside *ResourceState
- display code is running Payload() that runs deep copy that is reading the same map simultaneously

This PR explores a solution to prevent sharing mutable *ResourceState across goroutines. What if we agreed to do a deep copy only around channel sends, every time `ch <- evt` is done we instead do `ch <- evt.DeepCopy()`:

- producer (owner) goroutine stops everything while it does the copying
- when the copy is sent on the channel, the consuming goroutine owns the copy and can modify it

This conceptually sounds like a good idea but it's unclear if the deep copy in Payload() is needed for other reasons also; treading carefully here and experimenting, not suggesting to merge.

Fixes # (issue)

## Checklist


t0yv0:
Curiously, this removes the races previously detected by running `TestPrintfGo` test against `pulumi` CLI compiled with `-race` flag on.

Issue #198, opened 2017-06-04
Array valued output properties cause crash serializing Lumi state


lukehoban:
When I introduce an array-valued output property, and return an empty array from `Get`, I notice two issues.  First, the value is printed differently than values of other types (see `binaryMediaTypes` and `warnings` below):

```
      apiName         : "frontend"
      binaryMediaTypes: output<[]>
      createdDate     : "2017-06-04 05:23:38 +0000 UTC"
      id              : "wizhtblt26"
      version         : "1.0"
      warnings        : output<[]>
```

Second, after succesfully creating all resources, `lumi deploy` fails while trying to serialize out the deployment state:

```
F0603 22:24:08.078460   67725 failfast.go:37] An assertion has failed
goroutine 1 [running]:
github.com/golang/glog.stacks(0xc420186100, 0xc420166fc0, 0x46, 0x108)
	/Users/luke/go/src/github.com/golang/glog/glog.go:769 +0xa7
github.com/golang/glog.(*loggingT).output(0x190ee40, 0xc400000003, 0xc4200dd080, 0x18c003c, 0xb, 0x25, 0x0)
	/Users/luke/go/src/github.com/golang/glog/glog.go:720 +0x36b
github.com/golang/glog.(*loggingT).printDepth(0x190ee40, 0x3, 0x1, 0xc423e815a8, 0x1, 0x1)
	/Users/luke/go/src/github.com/golang/glog/glog.go:646 +0x12d
github.com/golang/glog.(*loggingT).print(0x190ee40, 0x3, 0xc423e815a8, 0x1, 0x1)
	/Users/luke/go/src/github.com/golang/glog/glog.go:637 +0x5a
github.com/golang/glog.Fatal(0xc423e815a8, 0x1, 0x1)
	/Users/luke/go/src/github.com/golang/glog/glog.go:1128 +0x53
github.com/pulumi/lumi/pkg/util/contract.failfast(0x1633ba4, 0x17)
	/Users/luke/go/src/github.com/pulumi/lumi/pkg/util/contract/failfast.go:37 +0x11f
github.com/pulumi/lumi/pkg/util/contract.Assert(0x0)
	/Users/luke/go/src/github.com/pulumi/lumi/pkg/util/contract/assert.go:27 +0x49
github.com/pulumi/lumi/pkg/resource.serializeProperty(0x159b5c0, 0xc422228ff0, 0x1626b4c, 0x4, 0xc4229ea2f8, 0xc4221c8b80)
[...]

PR #10163, opened 2022-07-15
Support stack reference as a resource

aq17:
# Description

Addresses `pulumi-yaml` [#198](https://github.com/pulumi/pulumi-yaml/issues/198)

To support stack references as a resource (rather than invoking via Fn::StackReference), we need to provide a mock `pulumi` package schema + package reference

## Checklist


aq17:
@AaronFriel I tried defining the package schema/reference in `builtins.go`, but wasn't able to set `Package.resourceTable` since it isn't exported, so I put it in `schema.go`

PR #10168, opened 2022-07-18
Remove api/renewLease from startup crit path

t0yv0:
# Description

Before the changes, the trace of `pulumi preview` on new AWS project shows `api/renewLease` call on the critical path-to-first-resource, something we are trying to optimize. After the changes the call either never happens on short programs. This seems like a small win. For me this is about 120ms.

Relevant context on where tokens come from and how it flows through the code:

```go
   func (b *cloudBackend) apply(...) {

        // createAndStartUpdate makes the initial token
        // it is Seen as "api/startUpdate" operation in the trace
        // example:
        //
        // POST /api/stacks/t0yv0/perf-js-test/dev/update/6c834f3e-2fba-4286-b317-1ec66cb8e697

	..., token, ... := b.createAndStartUpdate(...) // POST /update/..
        b.runEngineAction(... token ...)
   }

   func (b *cloudBackend) runEngineAction(... token ...) {
	.. := b.newUpdate(... token ...)
   }

   func (b *cloudBackend) newUpdate(... token ...) {
	// Create a token source for this update if necessary.
	var tokenSource *tokenSource
	if token != "" {
		ts, err := newTokenSource(ctx, token, b, update, 5*time.Minute)
		if err != nil {
			return nil, err
		}
		tokenSource = ts
	}
   }
```

It seems that newTokenSource should not refresh the token right away it should be fresh still.

Fixes # (issue)

## Checklist

### DIFF
--- a/pkg/backend/httpstate/state.go
+++ b/pkg/backend/httpstate/state.go
@@ -48,14 +48,8 @@ type tokenSource struct {
 	done     chan bool
 }
 
-func newTokenSource(ctx context.Context, token string, backend *cloudBackend, update client.UpdateIdentifier,
+func newTokenSource(ctx context.Context, initialToken string, backend *cloudBackend, update client.UpdateIdentifier,
 	duration time.Duration) (*tokenSource, error) {
-
-	// Perform an initial lease renewal.
-	newToken, err := backend.client.RenewUpdateLease(ctx, update, token, duration)
-	if err != nil {
-		return nil, err
-	}
 
 	requests, done := make(chan tokenRequest), make(chan bool)
 	go func() {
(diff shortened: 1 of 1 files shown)


pkg/backend/httpstate/state.go
```diff
@@ -48,48 +48,51 @@ type tokenSource struct {
...
+				currentError = fmt.Errorf("renewing lease: %w", err)
+
+				ticker.Stop()
+			} else {
+				token = newToken
+			}
+		}
+
+		// Is an initial lease renewal?
+		// renew()
```
t0yv0, line 79: This one is tricky? The original code was doing `RenewUpdateLease` call but ignoring the resulting `newToken` completely. It seems that the `token string` this func receives as a parameter is already a valid token coming out of the API. So maybe we should not do the initial lease renewal at all?

RobbieMcKinstry: Ah, it's possible that the call was to extend the lifetime of the initial token.
Say `initialToken` is guaranteed valid, but it expires soon. The initial `renew` call might be to extend its expiry? That might be a question for the Service team?

pkg/backend/httpstate/state.go
```diff
@@ -48,48 +48,51 @@ type tokenSource struct {
...
+			// the next request for a lease token.
+			if err != nil {
+				logging.V(3).Infof("error renewing lease: %v", err)
+				currentError = fmt.Errorf("renewing lease: %w", err)
+
+				ticker.Stop()
+			} else {
+				token = newToken
+			}
+		}
```
RobbieMcKinstry, line 76: I wouldn't use a closure to set stack locals. I think that's a little departed from idiomatic Go. I'd prefer if the function returned the two values instead.
Then, we can move them closer to where they're used, inside the `for` loop.

Maybe a pattern like this:

```golang
for {
    var token, err = initialToken, nil
    select {
    case <-ticker.C:
        token, err = renew()
```

I'm not sure this is much better tbh, since `renew` still closes over `ctx`, `update`, `token`, `duration`, `ticker`, and `backend`. I'd prefer another top-level function for `renew` but that's a _lot_ of parameters, so maybe not!

RobbieMcKinstry (changes requested):

pkg/backend/httpstate/state.go
t0yv0: As far as I understand from the trace and digging in the code, the token was *just acquired* couple hundred ms ago from another service call. So it seems like we're ok not renewing right away? If we need to though, I can uncomment the lines here and it will start by renewing, but still async (goroutine). This preserves the speedup we want because this call becomes concurrent with another service call.

pkg/backend/httpstate/state.go
t0yv0: I'd like to learn more idiomatic Go but I worry about the suggestion:

```
for {
    var token, err = initialToken, nil
```

Would this not reset token to initialToken on *every* iteration? That's not what we want really - we want the state to progress `token1, token2, token3, etc`. 

I can tinker with moving this out to a helper struct.

pkg/backend/httpstate/state.go
```diff
@@ -48,53 +48,61 @@ type tokenSource struct {
...
+	initialToken string,
+	client *client.Client,
+	update client.UpdateIdentifier,
+	duration time.Duration) {
+
+	// We will renew the lease after 50% of the duration has elapsed to allow time for retries.
+	renewTicker := time.NewTicker(duration / 2)
+	defer renewTicker.Stop()
+
+	state := struct {
```
t0yv0, line 75: @RobbieMcKinstry how about an anon struct?

RobbieMcKinstry: Ahah this might be overkill for two local variables! I think you solved my issue by moving most of the work into a top-level function!

RobbieMcKinstry (approved):
Approved with the nit that I think you were right to use two locals initially instead of a struct. Once we encapsulated a logic in a new top-level function, we were no longer replying on a closure to update stack local variables. I think that was the biggest issue initially. LGTM 👍🏻

PR #10173, opened 2022-07-18
Add tracing around os.Exec in node provider

t0yv0:
# Description

pulumi-language-nodejs starts node sub-processes to manage Pulumi projects. With these changes the subprocesses are now visible as spans in the [Pulumi trace](https://www.pulumi.com/docs/support/troubleshooting/#performance). The tagging convention is taken from pulumi-language-go. The change is small enough we might not need to highlight it in changelog.

Fixes # (issue)

## Checklist

### DIFF
--- a/sdk/nodejs/cmd/pulumi-language-nodejs/main.go
+++ b/sdk/nodejs/cmd/pulumi-language-nodejs/main.go
@@ -139,9 +139,19 @@ func main() {
 }
 
 // locateModule resolves a node module name to a file path that can be loaded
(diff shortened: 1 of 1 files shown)


t0yv0:
Yes! The https://opentelemetry.io/docs/instrumentation/js/getting-started/nodejs/#setup can be added to the node SDK and then Node can add traces too. We might need to migrate/upgrade opentracing to opentelemetry in the Go SDK first. This starts to sound like a large detour but given how difficult it seems to get an idea what Node is spending time on, maybe we should consider doing it?

PR #10177, opened 2022-07-19
Tidy up some help text

squaremo:
- the help for `pulumi import --yes` and `... --skip-preview` mentioned "refresh" as the operation, rather than "import"; I think this was probably leftover copypasta, so I've changed it to "import"; and,
 - the help text for `pulumi {destroy,up,import,refresh} --skip-preview` has "perform ... performing"; I've changed the first "perform" to "calculate" so it flows better.


squaremo:
(picking a couple of suggested reviewers arbitrarily ...)

ericrudder (approved):

PR #10184, opened 2022-07-19
fixes `pulumi preview` hangs in python

### LINKED ISSUE #10176: `pulumi preview` hangs on failure in chained Outputs
### What happened?

With a particular sequence of chained Outputs, a failure can cause `pulumi preview` (or `up`) to hang rather than crash with a traceback.

### Steps to reproduce

```bash
pulumi new aws-python
```

```python
# __main__.py
import pulumi
import pulumi_aws as aws

output = pulumi.Output.from_input("dummy")
# This fails as the given attribute does not exist.
# Removing this line produces a valid configuration
# that succeeds on `pulumi preview`.
output = output.apply(lambda x: x.bad_attribute)

bucket = aws.s3.Bucket("my-bucket", tags={"dummy": output})
# This line is required to induce the failure.
# Removing it causes `pulumi preview` to crash as expected.
bucket = pulumi.Output.from_input(bucket)

aws.s3.Bucket("my-other-bucket", tags={"dummy": bucket.arn})

# This line is required to induce the failure.
# Removing it causes `pulumi preview` to crash as expected.
aws.get_region()
```

```
pulumi preview
```

### Expected Behavior

The preview crashes with a traceback ending in `AttributeError: 'str' object has no attribute 'bad_attribute'`.

### Actual Behavior

The preview hangs indefinitely until killed via Ctrl+C.

### Versions used

```
CLI
Version      3.36.0
Go Version   go1.17.12
Go Compiler  gc

Plugins
NAME    VERSION
aws     5.10.0
python  unknown

Host
OS       Microsoft Windows 10 Pro
Version  10.0.19043 Build 19043
Arch     x86_64

This project is written in python: executable='C:\Program Files\Python39\python.exe' version='3.9.5'

Current Stack: dev

Found no resources associated with dev

Found no pending operations associated with dev

Dependencies:
NAME        VERSION
pip         22.1.2
pulumi-aws  5.10.0
setuptools  63.2.0
wheel       0.37.1
```

### Additional context

_No response_

### Contributing

dixler:
# Description

Fixes #10176 

## Checklist

### DIFF
--- a/sdk/python/lib/pulumi/runtime/stack.py
+++ b/sdk/python/lib/pulumi/runtime/stack.py
@@ -67,6 +67,8 @@ async def wait_for_rpcs(await_all_outstanding_tasks=True) -> None:
         # We await each RPC in turn so that this loop will actually block rather than busy-wait.
         while len(RPC_MANAGER.rpcs) > 0:
             await asyncio.sleep(0)
+            if RPC_MANAGER.unhandled_exception is not None:
+                break
             log.debug(
                 f"waiting for quiescence; {len(RPC_MANAGER.rpcs)} RPCs outstanding"
             )


t0yv0 (changes requested):
I'd like to see a repro test case checked in and a bit more of an explanation why this is the right fix.

sdk/python/lib/pulumi/runtime/stack.py
```diff
@@ -67,6 +67,8 @@ async def wait_for_rpcs(await_all_outstanding_tasks=True) -> None:
         # We await each RPC in turn so that this loop will actually block rather than busy-wait.
         while len(RPC_MANAGER.rpcs) > 0:
             await asyncio.sleep(0)
+            if RPC_MANAGER.unhandled_exception is not None:
```
t0yv0, line 70: So this terminates the `while len(RPC_MANAGER.rpcs) > 0` loop early in case there're exceptions and jumps to `raise` handler. This *might* make sense but naively this would only fix a hang if there were infinite entries in `RPC_MANAGER.rpcs` being added or else some of these entries never fulfilled their promises, making `await RPC_MANAGER.rpcs.pop()` take indefinite time. Is that the case, then why is it happening?

evanandrews-xrd:
I can't really understand why this loop is written the way it is:

https://github.com/pulumi/pulumi/blob/f2a3c417b7fa2ec6cebae19c33dee588b0c5fd51/sdk/python/lib/pulumi/runtime/stack.py#L61-L73

This awaits the completion of each RPC individually in the *reverse* order they were appended. I'm struggling to debug this effectively, but I would guess that one of the later RPCs is depending on an earlier one, which has not yet completed. It seems like changing `pop()` to `pop(0)` in order to await them in the order they were enqueued fixes the hang:

```python
 while len(RPC_MANAGER.rpcs) > 0: 
     await asyncio.sleep(0) 
     log.debug( 
         f"waiting for quiescence; {len(RPC_MANAGER.rpcs)} RPCs outstanding" 
     ) 
     await RPC_MANAGER.rpcs.pop(0)  # <-- This gets the first entry rather than the last
```

However I can't understand why they need to be awaited individually at all. Is the intention not to simply wait for all RPCs to finish?

```python
rpcs = RPC_MANAGER.rpcs.copy()
RPC_MANAGER.rpcs.clear()
await asyncio.gather(*rpcs)
```

Note the remaining code appears to allow for new RPCs to be created during the execution of the existing ones, hence the need to copy and clear the list before allowing any existing ones to run.

It's possible there are cases I don't understand, particularly since I can't work out what the large comment in the existing code is talking about, but this much simpler version seems to work both in my MRE and my actual code.

evanandrews-xrd:
> I'd still like to see a repro to contextualize the solutions 🙏

Do you mean a repro beyond the one provided in #10176? That was the absolute simplest example I could make that caused the hang.

> The hypothesis that there is a deadlock because this loop waits on Task1 that depends on Task2 and starves Task2 is a bit unconvincing to me, as this code here creates a new callback and should not block?
> `await RPC_MANAGER.rpcs.pop(0)`
> Unless the Task1, Task2 depend themselves on this code (several instances of this code end up being run concurrently), is that what's happening? Need that repro...

After digging into this a bit further, it appears that the task to register the `my-other-bucket` resource is hanging indefinitely. It's not dependent on the completion of a previous task; it will never complete. Running the tasks in the other order (using e.g. `pop(0)`) causes an exception when the `my-bucket` task is awaited where the traceback goes through that `await` in `wait_for_rpcs`. Suppressing that exception then causes the stack to hang forever again:

```python
while len(RPC_MANAGER.rpcs) > 0:
    await asyncio.sleep(0)
    log.debug(
        f"waiting for quiescence; {len(RPC_MANAGER.rpcs)} RPCs outstanding"
    )
    try:
        await RPC_MANAGER.rpcs.pop(0)
    except Exception:
        pass
```

So the issue is that `my-bucket` will explicitly terminate, but `my-other-bucket` won't, and the code as currently written waits for `my-other-bucket` first.

Given all this, perhaps the original fix of checking for exceptions each time around the loop makes the most sense? There's no point continuing if an RPC has already failed. Alternatively someone more familiar with Pulumi's internals may be able to stop the `my-other-bucket` task from hanging.

dixler:
@t0yv0 I think it would be reasonable to add this 2 line diff as it solves the initial repro provided and looks like a very defensively written piece of code like the python equivalent of
```go
if err != nil {
    return err
}
```
We have an initial repro/example, but I'm not sure if we want to put the initial repro into our tests because it has a dependency on AWS and it appears really complex and time consuming to build a provider to repro this behavior.

t0yv0:
@evanandrews-xrd  ah, thanks a lot to pointing me to the repro code, I missed that we have that in the ticket! That is great.

@dixler technically we must have a test for every change we introduce to the SDK; we also should strive to understand the root cause (and simplify the test if we can). There's no real dep here on AWS as any resource would do (test resources included).

So with the repro I did a deep dive into the test case. It seems that:

- "my-bucket" register resource RPC fails with an exception (as expected)
- "my-other-bucket" register resource RPC hangs indefinitely waiting on `Output.from_input(bucket).arn.is_known()`
- surprisingly `bucket.arn.is_known()` completes just fine, the `Output.from_input` part is crucial
- the suggested fix here is trying to short-circuit and fail the program earlier when we become aware that one of the RPCs failed; in this case even if there's unfulfilled futures, the program at least terminates
- but this may leave other hangs/modularity issues with unfulfilled futures

This got me really curious and I got to this:

```python

from pulumi.runtime.rpc_manager import RPC_MANAGER
import asyncio
import pulumi
import pulumi.runtime.sync_await as sa
import pulumi_aws as aws
import sys


O = pulumi.Output


def _sync_await(x):
    try:
        return str(sa._sync_await(asyncio.wait_for(x, timeout=1.0)))
    except Exception as e:
        return f'err:{type(e)}'


def debug_output(name, o):
    print(f'..\n..\nDebugging output {name}')
    print('value:     ', _sync_await(o._future))
    print('is_secret: ', _sync_await(o._is_secret))
    print('is_known:  ', _sync_await(o._is_known))
    print('resources: ', _sync_await(o._resources))


errf = asyncio.Future()
errf.set_exception(AttributeError('!'))

bad = O.from_input("foo").apply(lambda x: x.bad_attr)
debug_output('bad', bad)

bucket = aws.s3.Bucket("my-bucket", tags={"dummy": bad})
debug_output('bucket.arn', bucket.arn)

bucket_output = O.from_input(bucket)
debug_output('bucket_output', bucket_output)
debug_output('bucket_output.arn', bucket_output.arn)
debug_output('bucket_output.apply(..)', bucket_output.apply(lambda x: x.arn))

```

The output is:

```
    ..
    ..
    Debugging output bad
    value:      err:<class 'AttributeError'>
    is_secret:  False
    is_known:   False
    resources:  set()
    ..
    ..
    Debugging output bucket.arn
    value:      err:<class 'AttributeError'>
    is_secret:  err:<class 'AttributeError'>
    is_known:   err:<class 'AttributeError'>
    resources:  {<pulumi_aws.s3.bucket.Bucket object at 0x10db41f70>}
    ..
    ..
    Debugging output bucket_output
    value:      <pulumi_aws.s3.bucket.Bucket object at 0x10db41f70>
    is_secret:  False
    is_known:   True
    resources:  set()
    ..
    ..
    Debugging output bucket_output.arn
    value:      err:<class 'asyncio.exceptions.InvalidStateError'>
    is_secret:  err:<class 'asyncio.exceptions.TimeoutError'>
    is_known:   err:<class 'asyncio.exceptions.TimeoutError'>
    resources:  {<pulumi_aws.s3.bucket.Bucket object at 0x10db41f70>}
    ..
    ..
    Debugging output bucket_output.apply(..)
    value:      err:<class 'asyncio.exceptions.InvalidStateError'>
    is_secret:  err:<class 'asyncio.exceptions.TimeoutError'>
    is_known:   err:<class 'asyncio.exceptions.TimeoutError'>
    resources:  {<pulumi_aws.s3.bucket.Bucket object at 0x10db41f70>}
```

So there's a couple underlying bugs here. There is a bug in `apply` that results in InvalidStateError, TimeoutError (proxy for hang) introduced. I think I've isolated it in https://github.com/pulumi/pulumi/pull/10334 - I pass the original program with 10334.

The second issue is how `bucket.arn` future copied the AttributeError not only into the value field but also into is_secret, is_known fields. It seems like Output combinators like apply and maybe more of them are not prepared to handle this state. There is a code smell here, perhaps something to backlog further.

PR #10187, opened 2022-07-20
[ci] Fix extra extension in goreleaser templates

AaronFriel:
The `{{ .Name }}` template string in GoReleaser includes the extension, which must be stripped to use it in a `prebuilt.path`. 

Verified locally by running all three goreleaser configs with `--debug` for verbose output. The command to run the `.goreleaser.build.yml` against Windows x86-64 is:

```bash
#!/bin/bash

GOWORK=off GORELEASER_CURRENT_TAG=v3.35.0-alpha.1 GOOS=windows GOARCH=amd64 \
  goreleaser build -f .goreleaser.build.yml \
  -p 1 --skip-validate --rm-dist --snapshot --single-target

# rename dir to what .goreleaser.prerelease.yml expects as it copes in artifacts from the build job
mv goreleaser goreleaser-windows

# This will succeed with this PR's commit, where it previously would fail 🎉
GOWORK=off GORELEASER_CURRENT_TAG=v3.35.0-alpha.1 GOOS=windows GOARCH=amd64 \
  goreleaser build -f .goreleaser.prerelease.yml \
  -p 1 --skip-validate --rm-dist --snapshot --single-target
```

### DIFF
--- a/.goreleaser.build.yml
+++ b/.goreleaser.build.yml
@@ -56,13 +56,15 @@ builds:
 - <<: *pulumibin
   id: pulumi-language-java
   binary: pulumi-language-java
+  builder: prebuilt
   prebuilt:
-    path: goreleaser-lang/java/{{ .Os }}-{{ replace .Arch "amd64" "x64" }}/{{ .Name }}{{ .Ext }}
+    path: goreleaser-lang/java/{{ .Os }}-{{ replace .Arch "amd64" "x64" }}/{{ .Name }}
 - <<: *pulumibin
   id: pulumi-language-yaml
   binary: pulumi-language-yaml
+  builder: prebuilt
   prebuilt:
-    path: goreleaser-lang/yaml/{{ .Os }}-{{ replace .Arch "amd64" "x64" }}/{{ .Name }}{{ .Ext }}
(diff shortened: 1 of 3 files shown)


.goreleaser.build.yml
```diff
@@ -56,13 +56,15 @@ builds:
...
   prebuilt:
-    path: goreleaser-lang/java/{{ .Os }}-{{ replace .Arch "amd64" "x64" }}/{{ .Name }}{{ .Ext }}
+    path: goreleaser-lang/java/{{ .Os }}-{{ replace .Arch "amd64" "x64" }}/{{ .Name }}
 - <<: *pulumibin
   id: pulumi-language-yaml
   binary: pulumi-language-yaml
+  builder: prebuilt
   prebuilt:
-    path: goreleaser-lang/yaml/{{ .Os }}-{{ replace .Arch "amd64" "x64" }}/{{ .Name }}{{ .Ext }}
+    path: goreleaser-lang/yaml/{{ .Os }}-{{ replace .Arch "amd64" "x64" }}/{{ .Name }}
```
AaronFriel, line 67: Verified locally that without `builder: prebuilt`, it was not using this key and thus not doing what was expected.

PR #10193, opened 2022-07-20
Add a gRPC tracing interceptor

t0yv0:
# Description

I have been debugging provider issues in a complicated project with remote component resources and it has been quite difficult to make sense of the log output as well as the tracing output in that scenario. What I wanted was primarily a trail of gRPC conversations where requests are correlated to responses and are easy to read and filter/query. This PR adds a debug facility that helped me solve the issue.

Example entry from the log:

```
{
  "method": "/pulumirpc.ResourceMonitor/RegisterResource",
  "request": {
    "type": "aws-apigateway:index:RestAPI",
    "name": "helloWorldApi",
    "parent": "urn:pulumi:tdev::java-repro::pulumi:pulumi:Stack::java-repro-tdev",
    "object": {
      "routes": [
        {
          "eventHandler": {
            "4dabf18193072939515e22adb298388d": "5cf8f73096256a8f31e491e813e4eb8e",
            "id": "helloWorldFunction-91d476c",
            "urn": "urn:pulumi:tdev::java-repro::aws:lambda/function:Function::helloWorldFunction"
          },
          "method": "GET",
          "path": "/"
        }
      ]
    },
    "dependencies": [
      "urn:pulumi:tdev::java-repro::aws:lambda/function:Function::helloWorldFunction"
    ],
    "propertyDependencies": {
      "routes": {
        "urns": [
          "urn:pulumi:tdev::java-repro::aws:lambda/function:Function::helloWorldFunction"
        ]
      }
    },
    "version": "0.0.8",
    "acceptSecrets": true,
    "customTimeouts": {},
    "deleteBeforeReplaceDefined": true,
    "remote": true,
    "acceptResources": true
  },
  "response": {
    "urn": "urn:pulumi:tdev::java-repro::aws-apigateway:index:RestAPI::helloWorldApi",
    "object": {
      "api": {
        "4dabf18193072939515e22adb298388d": "5cf8f73096256a8f31e491e813e4eb8e",
        "id": "",
        "urn": "urn:pulumi:tdev::java-repro::aws-apigateway:index:RestAPI$aws:apigateway/restApi:RestApi::helloWorldApi"
      },
      "deployment": {
        "4dabf18193072939515e22adb298388d": "5cf8f73096256a8f31e491e813e4eb8e",
        "id": "",
        "urn": "urn:pulumi:tdev::java-repro::aws-apigateway:index:RestAPI$aws:apigateway/deployment:Deployment::helloWorldApi"
      },
      "stage": {
        "4dabf18193072939515e22adb298388d": "5cf8f73096256a8f31e491e813e4eb8e",
        "id": "",
        "urn": "urn:pulumi:tdev::java-repro::aws-apigateway:index:RestAPI$aws:apigateway/stage:Stage::helloWorldApi"
      }
    },
    "propertyDependencies": {
      "api": {
        "urns": [
          "urn:pulumi:tdev::java-repro::aws:lambda/function:Function::helloWorldFunction",
          "urn:pulumi:tdev::java-repro::aws-apigateway:index:RestAPI$aws:apigateway/restApi:RestApi::helloWorldApi"
        ]
      },
      "deployment": {
        "urns": [
          "urn:pulumi:tdev::java-repro::aws:lambda/function:Function::helloWorldFunction",
          "urn:pulumi:tdev::java-repro::aws-apigateway:index:RestAPI$aws:apigateway/deployment:Deployment::helloWorldApi"
        ]
      },
      "stage": {
        "urns": [
          "urn:pulumi:tdev::java-repro::aws:lambda/function:Function::helloWorldFunction",
          "urn:pulumi:tdev::java-repro::aws-apigateway:index:RestAPI$aws:apigateway/stage:Stage::helloWorldApi"
        ]
      },
      "url": {
        "urns": [
          "urn:pulumi:tdev::java-repro::aws:lambda/function:Function::helloWorldFunction",
          "urn:pulumi:tdev::java-repro::aws-apigateway:index:RestAPI$aws:apigateway/deployment:Deployment::helloWorldApi"
        ]
      }
    }
  }
}
``` 

Fixes # (issue)

## Checklist

### DIFF
--- a/sdk/go/common/util/rpcutil/debug_interceptor.go
+++ b/sdk/go/common/util/rpcutil/debug_interceptor.go
@@ -1,0 +1,144 @@
+// Copyright 2016-2022, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package rpcutil
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"sync"
+
+	"google.golang.org/grpc"
+
+	"github.com/golang/protobuf/jsonpb"
+	"github.com/golang/protobuf/proto"
+
+	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
+)
+
+// Logs all gRPC converations in JSON format.
+//
+// To enable, PULUMI_DEBUG_GRPC must be set, pointing to a file to
+// append logs to.
+//
+// For example:
+//
+//     PULUMI_DEBUG_GPRC=grpc.json pulumi preview
+//     jq . grpc.json
+//
+func debugServerInterceptor() grpc.UnaryServerInterceptor {
+	return debugInterceptorInstance
+}
+
+type debugInterceptorLogEntry struct {
+	Method   string          `json:"method"`
+	Request  json.RawMessage `json:"request"`
+	Response json.RawMessage `json:"response"`
+	Errors   []string        `json:"errors,omitempty"`
+}
+
+type debugInterceptor struct {
+	logFile string
+	mutex   sync.Mutex
+}
+
+func newDebugInterceptor() *debugInterceptor {
+	logFile := os.Getenv("PULUMI_DEBUG_GRPC")
+	i := &debugInterceptor{logFile: logFile}
+	if logFile != "" {
+		i.clearLogFile()
+	}
+	return i
+}
+
+func (i *debugInterceptor) interceptor() grpc.UnaryServerInterceptor {
+	if i.logFile == "" {
+		return i.noInterceptor()
+	}
+	return i.loggingInterceptor()
+}
+
+func (i *debugInterceptor) loggingInterceptor() grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req interface{},
+		info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		var err error
+		log := debugInterceptorLogEntry{
+			Method: info.FullMethod,
+		}
+		if log.Request, err = i.transcode(req); err != nil {
+			i.track(&log, err)
+		}
+		resp, err := handler(ctx, req)
+		if err != nil {
+			i.track(&log, err)
+		}
+		if log.Response, err = i.transcode(resp); err != nil {
+			i.track(&log, err)
+		}
+		i.record(log)
+		return resp, err
+	}
+}
+
+func (i *debugInterceptor) clearLogFile() {
+	i.mutex.Lock()
+	defer i.mutex.Unlock()
+
+	err := ioutil.WriteFile(i.logFile, []byte{}, 0644)
+	contract.AssertNoError(err)
+}
+
+func (i *debugInterceptor) record(log debugInterceptorLogEntry) {
+	i.mutex.Lock()
+	defer i.mutex.Unlock()
+
+	f, err := os.OpenFile(i.logFile, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+	contract.AssertNoError(err)
+	defer contract.IgnoreClose(f)
+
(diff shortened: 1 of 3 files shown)


AaronFriel (approved):

t0yv0:
opentracing ships its own interceptor, that's how I found the interceptor concept. 

opentracing interceptor is way more complex, while it does capture grpc payloads inside it it's a bit painful to fish those out; and also we might need to change that code when we upgrade to opentelemetry anyway; it sounds like a separate interceptor like this is simpler? 

Is concern extra code, missing features, or runtime overhead? I'm expecting this to be near-0 overhead when not turned on.

sdk/go/common/util/rpcutil/debug_interceptor.go
```diff
@@ -0,0 +1,144 @@
...
+type debugInterceptorLogEntry struct {
+	Method   string          `json:"method"`
+	Request  json.RawMessage `json:"request"`
+	Response json.RawMessage `json:"response"`
+	Errors   []string        `json:"errors,omitempty"`
+}
+
+type debugInterceptor struct {
+	logFile string
+	mutex   sync.Mutex
```
t0yv0, line 57: Note to self: we might have multiple OS processes starting gRPC servers via Serve hitting this code (is this true? Or only pulumi process start servers and the rest are clients?). If that is the case, sync.Mutex is not enough, an to write to a single file there need to be file locks. I can throw one just in case. For the use cases I have in mind here terrible locking overhead is OK to get started with this.

Frassle: Pulumi is a Resource and Engine server, the language plugins are Language servers, providers are Provider servers, automation api programs are Language servers. So yes there are multiple processes starting servers.

Frassle:
So my comment about what makes grpc servers also prompted that we make grpc servers from every language for inline automation programs, and in the future all of automation api will be grpc'd. 

But _every_ connection goes through the engine, so could we not just change it so the engine logs grpc on clients and servers. Then you don't have to worry about locks so much (it's only one process logging) and we don't have to worry about some grpc traces showing up, and others not (because we haven't bothered adding this code to python auto api for example)

t0yv0:
I love the idea. So then, debug tracing should only happen for ResourceMonitor and Engine servers started by `pulumi` process, and not the others; plus we add a client debug interceptor to all the gRPC clients in the `pulumi` process, then we have every gRPC conversation on hand. That sounds perfect.

pgavlin:
> Is concern extra code, missing features, or runtime overhead? I'm expecting this to be near-0 overhead when not turned on.

Just extra code + maintenance, yes. In principle we should already have the data the new interceptor dumps within the OT traces.

t0yv0:
Technically true, but there are couple issues with longevity and applicability of that code:

- We will need to upgrade opentracing to opentelemetry to keep up with the ecosystem
- As we do so we might phase out AppDash, and change the format of file traces if we still support them
- The use cases around that will continue to be oriented toward sending data into Jaeger (or similar) and exploring it there

So, if I want a facility to query the gRPC conversations with `jq`, I'd want to either have a server that impersonates an OT endpoint and extracts the data from there, or something that can query e.g. Jaeger, or parse the file format we produce (currently AppDash but likely changing) to extract the data and reformat it for querying.

Given that I think the separate interceptor(s) approach is more maintainable, if we want this kind of functionality. 

If we don't want it in pulumi/pulumi then I'd take it to a personal project (separate CLI) that does parse the trace format and try to upgrade that when Pulumi changes the trace format.

pgavlin:
> parse the file format we produce (currently AppDash but likely changing) to extract the data and reformat it for querying.
> 
> Given that I think the separate interceptor(s) approach is more maintainable, if we want this kind of functionality.

Yeah, parsing the file format is what I was thinking of. If you think that the separate interceptor is more maintainable, then let's go with that.

AaronFriel:
s/tracing interceptor/debug interceptor/ in the PR title?

I was confused when I saw the PR and read it as I thought this was opentelemetry related, but this is a useful facility on its own to dump logs.

> @t0yv0  opentracing interceptor is way more complex, while it does capture grpc payloads inside it it's a bit painful to fish those out; and also we might need to change that code when we upgrade to opentelemetry anyway; it sounds like a separate interceptor like this is simpler?

> @pgavlin Just extra code + maintenance, yes. In principle we should already have the data the new interceptor dumps within the OT traces.

I'd be inclined to accept this as a lighter weight alternative to full tracing with a `--debug-grpc` command, and leave the opentelemetry work to a separate PR. This is useful functionality, and later when we support opentelemetry, if a user wants to plumb it to something else they'd set an arg like `--opentelemetry-collector=:55680" which could be any arbitrary opentelemetry service.

t0yv0:
Closing to pick back up later (hopefully soon) - tracked in https://github.com/pulumi/pulumi/issues/10232 - did not have time to complete this.

Issue #10196, opened 2022-07-20
Add option to not install dependencies on `pulumi convert`


aq17:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Add option to not install dependencies when running `pulumi convert` and to simply generate the converted program

dixler:
@aq17 left some comments on your merged PR.

Frassle:
Also left a comment

Frassle:
Done by https://github.com/pulumi/pulumi/pull/10205

PR #10198, opened 2022-07-21
Add flag to specify whether to install dependencies on `pulumi convert`

aq17:
# Description

Add `install-dependencies` flag to allow user to opt out of installing dependencies on `pulumi convert` if they only want the generated files. Defaults to `true.

Addresses #10196 

## Checklist

### DIFF
--- a/pkg/cmd/pulumi/convert.go
+++ b/pkg/cmd/pulumi/convert.go
@@ -38,6 +38,7 @@ type projectGeneratorFunc func(directory string, project workspace.Project, p *p
 func newConvertCmd() *cobra.Command {
 	var outDir string
 	var language string
+	var installDeps bool
 
 	cmd := &cobra.Command{
 		Use:   "convert",
@@ -109,8 +110,10 @@ func newConvertCmd() *cobra.Command {
 
 			defer ctx.Close()
 
-			if err := installDependencies(ctx, &proj.Runtime, pwd); err != nil {
-				return result.FromError(err)
+			if installDeps {
+				if err := installDependencies(ctx, &proj.Runtime, pwd); err != nil {
+					return result.FromError(err)
+				}
 			}
 
 			return nil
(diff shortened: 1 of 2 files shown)


AaronFriel (approved):
Short and sweet.

pkg/cmd/pulumi/convert.go
```diff
@@ -128,5 +131,9 @@ func newConvertCmd() *cobra.Command {
 		//nolint:lll
 		&outDir, "out", ".", "The output directory to write the convert project to")
 
+	cmd.PersistentFlags().BoolVar(
+		//nolint:lll
+		&installDeps, "install-dependencies", true, "Whether to install dependencies specified in the pulumi project")
```
dixler, line 136: maybe consider --no-install-dependencies? since the flag should change behavior.

dixler: this flag might imply to the user that dependencies are not installed by default, but they are.

Frassle: Yeh I feel this should of copied 'new' https://github.com/pulumi/pulumi/blob/master/pkg/cmd/pulumi/new.go#L483 and be 'generate-only'

aq17: @dixler @Frassle agreed, will update 😃

mikhailshilkov:
You may know if - but just in case: the text `Addresses https://github.com/pulumi/pulumi/issues/10196` doesn't automatically close the linked issue, while `Fixes https://github.com/pulumi/pulumi/issues/10196` (or `Fixes #10196`) would.

aq17:
> You may know it - but just in case: the text `Addresses https://github.com/pulumi/pulumi/issues/10196` doesn't automatically close the linked issue, while `Fixes https://github.com/pulumi/pulumi/issues/10196` (or `Fixes #10196`) would.

I did not know this before 🙌🏼

Issue #10120, opened 2022-07-13
Update how naming is done across providers


Frassle:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

The current approach to autonames in update plans has found a number of problem cases. All the code that was added to try and use deterministic names based on sequence numbers has been backed out of the providers. This issue is to track adding a new type of "deterministic" name back to providers. Providers will generate deterministic names based on an input hash to `Check`, but that hash will be randomly generated by the engine. When using update plans the engine will save the hash to the plan, and re-use it when running the update based on that plan.

1. Add a new name function that generates a random suffix for a name based on a hash (or falling back to the random source if the hash is empty, for backwards compatibility with old engines)
2. Update the provider interface to start sending a "random seed/hash" value as part of Check, initially the engine will just fill this with 32 bytes of random data on each call.
3. Update all providers to make use of the above two points.
4. Update the engine to save the hashes to update plans and re-use them when calling Check in the update run.

N.B. 1 is probably already covered by https://github.com/pulumi/pulumi/pull/9852. 3 and 4 can be done in parallel.

PR #10203, opened 2022-07-21
Remove sequence numbers

Frassle:
# Description

As part of https://github.com/pulumi/pulumi/issues/10120 we're replacing sequence numbers with saved seeds. This PR simply cleans up all the existing sequence number code and starts replacing it with `randomSeed []byte` on the Check interface.

Note that currently these are always nil, actually generating and saving and loading these from plans will be done in a separate PR.

I've removed two tests as part of this:
1) TestProviderDeterministicPreview because we don't have deterministic previews anymore, and even when we plumb seeds in we'll only have determinism when using plans.
2) TestSequenceNumberResetsAfterReplace, because well there's no sequence numbers anymore.

## Checklist


Zaid-Ajaj (approved):
Looks good ✅

PR #10204, opened 2022-07-21
Change GetPluginPath to assert the paths are as expected and use the paths calculated 

Frassle:
Just a couple of sanity changes based on a comment from Harry on https://github.com/pulumi/pulumi/pull/10146


Zaid-Ajaj (approved):
LGTM

Issue #10189, opened 2022-07-20
TestPythonTypes fails on Windows


mikhailshilkov:
See https://github.com/pulumi/pulumi/runs/7430501050?check_suite_focus=true#step:33:243

```
command.go:83: Invoke 'D:\a\pulumi\pulumi\bin\pulumi.exe stack export --file C:\Users\RUNNER~1\AppData\Local\Temp\p-it-fv-az192-6-declared-e8d3fa071148438160\stack.json' failed: exit status 2
panic: Failed to load user32.dll: A dynamic link library (DLL) initialization routine failed.
goroutine 1 [running]:
syscall.MustLoadDLL(...)
	/opt/hostedtoolcache/go/1.17.12/x64/src/syscall/dll_windows.go:106
github.com/xanzy/ssh-agent.winAPI({0x254dca6, 0x3824eb0}, {0x254e721, 0xb})
	/home/runner/go/pkg/mod/github.com/xanzy/ssh-agent@v0.3.1/pageant_windows.go:68 +0xb4
github.com/xanzy/ssh-agent.init()
	/home/runner/go/pkg/mod/github.com/xanzy/ssh-agent@v0.3.1/pageant_windows.go:62 +0x1ac
    command.go:103: Wrote output to C:\Users\RUNNER~1\AppData\Local\Temp\p-it-fv-az192-6-declared-e8d3fa07-2679856236\command-output\pulumi-export.20220720-141819.f0baa.log
    program.go:1221: Destroying stack
    command.go:34: **** Invoke 'D:\a\pulumi\pulumi\bin\pulumi.exe destroy --non-interactive --yes --skip-preview' in 'C:\Users\RUNNER~1\AppData\Local\Temp\p-it-fv-az192-6-declared-e8d3fa07-2679856236'
    command.go:83: Invoke 'D:\a\pulumi\pulumi\bin\pulumi.exe destroy --non-interactive --yes --skip-preview' failed: exit status 2
panic: Failed to load user32.dll: A dynamic link library (DLL) initialization routine failed.
goroutine 1 [running]:
syscall.MustLoadDLL(...)
	/opt/hostedtoolcache/go/1.17.12/x64/src/syscall/dll_windows.go:106
github.com/xanzy/ssh-agent.winAPI({0x254dca6, 0x3824eb0}, {0x254e721, 0xb})
	/home/runner/go/pkg/mod/github.com/xanzy/ssh-agent@v0.3.1/pageant_windows.go:68 +0xb4
[...]

PR #10207, opened 2022-07-21
[windows] replace ssh-agent with one that lazily loads dlls

AaronFriel:
We frequently see Windows integration tests fail with a panic like in this issue:
- https://github.com/pulumi/pulumi/issues/10189

```
panic: Failed to load user32.dll: A dynamic link library (DLL) initialization routine failed.
goroutine 1 [running]:
syscall.MustLoadDLL(...)
	/opt/hostedtoolcache/go/1.17.12/x64/src/syscall/dll_windows.go:106
github.com/xanzy/ssh-agent.winAPI({0x254dca6, 0x3824eb0}, {0x254e721, 0xb})
	/home/runner/go/pkg/mod/github.com/xanzy/ssh-agent@v0.3.1/pageant_windows.go:68 +0xb4
github.com/xanzy/ssh-agent.init()
	/home/runner/go/pkg/mod/github.com/xanzy/ssh-agent@v0.3.1/pageant_windows.go:62 +0x1ac
```

This forks and replaces `xanzy/ssh-agent` with one that lazily loads `user32.dll`. The hypothesis is that we don't always need to load this DLL, and perhaps something unusual occurs when - during integration tests - we repeatedly load `user32.dll`? Either way, this reduces the work done by the library and removes synchronous startup-time DLL loading such that we won't load the DLL at all for most code paths.

The code path that would cause `user32.dll` to be loaded would be on use of `go-git`, which occurs during template refreshing, and only if there is a process named `pageant` running in the user's session.

https://github.com/xanzy/ssh-agent/compare/v0.3.1...AaronFriel:ssh-agent:v0.5.0


t0yv0:
Afraid to ask, but why do we link in ssh-agent? Is this to support one of the state backends?

t0yv0:
```
go mod why github.com/xanzy/ssh-agent                                                                                                     ~/pulumi/pkg
# github.com/xanzy/ssh-agent
github.com/pulumi/pulumi/pkg/v3/cmd/pulumi
gopkg.in/src-d/go-git.v4
gopkg.in/src-d/go-git.v4/plumbing/transport/client
gopkg.in/src-d/go-git.v4/plumbing/transport/ssh
github.com/xanzy/ssh-agent
```

t0yv0:
Poked around, the usage of go-git seems fairly entrenched, as in addGitMetadata. It makes sense that it would not always fire up ssh dependencies, only when it's trying to say harvest metadata from SSH remotes. Interesting.

pkg/go.mod
```diff
@@ -7,6 +7,8 @@ replace github.com/pulumi/pulumi/sdk/v3 => ../sdk
 // Working around https://github.com/sergi/go-diff/issues/123
 replace github.com/sergi/go-diff => github.com/sergi/go-diff v1.1.0
 
+replace github.com/xanzy/ssh-agent => github.com/aaronfriel/ssh-agent v0.5.0
```
t0yv0, line 10: Bring the fork into pulumi/ssh-agent?

t0yv0:
How confident we are it solves our issue or we want to try to merge and have it go for a bit?

Is there a test we that triggers the pathway that's changing here (the go-git use +ssh use) that still works?

AaronFriel:
@t0yv0 the failure is non-deterministic, so I don't know if "pageant" support is just perpetually broken or what, but this should drive our Windows tests to green more reliably as the code path that loads `user32.dll` will be skipped.

I can try re-running the Windows tests a few times, but I'd love to get this in to improve build reliability.

t0yv0 (approved):

PR #10213, opened 2022-07-21
 Fix x-plat issues with watch, use watchexec-derived binary

### LINKED ISSUE #9326: Cross compiling from Windows/Linux to MacOS fails due to `notify` dependency
Several Pulumi packages have had to update their build systems to use MacOS as the publish/builder OS to address cross-platform build issues with the https://github.com/rjeczalik/notify dependency. Example: pulumi/pulumi-aws-native#421

Solving this could take the form of an upstream change to that dependency to enable cross-compilation, or switching to an alternative filesystem watcher library.

Candidate libraries:

- 6,700+ stars, CI system builds and cross-compiles from linux to 6 OS targets, does not use CGo: https://github.com/fsnotify/fsnotify, see: https://github.com/fsnotify/fsnotify/blob/main/.github/workflows/build.yml#L10
- 1,200+ stars, uses pure Go APIs (not fsevents) for cross-platform support: https://github.com/radovskyb/watcher

AaronFriel:
Replaces library used to implement recursive file watching with https://github.com/pulumi/pulumi-watch, which is itself a shim around https://github.com/watchexec/watchexec.

This solves our cross-platform build issues, and allows us to build our CLIs on any platform and cross-compile to macOS.

Fixes #9326

### DIFF
--- a/pkg/backend/watch.go
+++ b/pkg/backend/watch.go
@@ -18,18 +18,23 @@
 package backend
 
 import (
+	"bufio"
 	"context"
 	"fmt"
+	"os"
+	"os/exec"
 	"path"
+	"path/filepath"
+	"runtime"
 	"strings"
 	"time"
-
-	"github.com/rjeczalik/notify"
 
 	"github.com/pulumi/pulumi/pkg/v3/backend/display"
 	"github.com/pulumi/pulumi/pkg/v3/operations"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/apitype"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/diag/colors"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/util/cmdutil"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/logging"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/result"
 )
@@ -70,23 +75,12 @@ func Watch(ctx context.Context, b Backend, stack Stack, op UpdateOperation,
 		}
 	}()
 
-	events := make(chan notify.EventInfo, 1)
-
-	for _, p := range paths {
-		// Provided paths can be both relative and absolute.
-		watchPath := ""
-		if path.IsAbs(p) {
-			watchPath = path.Join(p, "...")
-		} else {
-			watchPath = path.Join(op.Root, p, "...")
-		}
-
-		if err := notify.Watch(watchPath, events, notify.All); err != nil {
-			return result.FromError(err)
-		}
+	// Provided paths can be both relative and absolute.
+	events, stop, result := watchPaths(op.Root, paths)
+	if result != nil {
+		return result
 	}
-
-	defer notify.Stop(events)
+	defer stop()
 
 	fmt.Printf(op.Opts.Display.Color.Colorize(
 		colors.SpecHeadline+"Watching (%s):"+colors.Reset+"\n"), stack.Ref())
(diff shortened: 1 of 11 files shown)


pkg/backend/watch.go
```diff
@@ -108,8 +99,52 @@ func Watch(ctx context.Context, b Backend, stack Stack, op UpdateOperation,
...
+		if path.IsAbs(p) {
+			watchPath = p
+		} else {
+			watchPath = path.Join(root, p)
+		}
+
+		args = append(args, "--watch", watchPath)
+	}
+
+	watchCmd, err := exec.LookPath("pulumi-watch")
```
Frassle, line 121: So
A) We should probably envvar this so people can override it PULUMI_WATCH_BINARY?
B) We should support the "watch.exe" being next to "pulumi.exe" but not on path.

AaronFriel: Does `PULUMI_WATCH_BINARY` make sense if we require specific command line args?

Finding side-by-side `pulumi-watch` makes sense as a fallback, in case the user is invoking Pulumi by installing it to, e.g.: /opt/pulumi and invoking it via `/opt/pulumi up`. Will add that.

AaronFriel: I've refactored out the `exec.LookPath` to support finding the .exe on Windows & look for a side by side binary. After installing `pulumi`, `pulumi-watch`, and `pulumi-language-yaml` to ~/.pulumi-dev/bin, I was able to run:

```
$ PATH="" ~/.pulumi-dev/bin/pulumi watch
Watching (dev):
10:53:09.646[                    ] Updating...
10:53:10.624[                    ] Update complete.
10:53:15.029[                    ] Updating...
10:53:15.566[                 bar] create random:index/randomUuid:RandomUuid
10:53:15.664[                 bar] done create random:index/randomUuid:RandomUuid
10:53:15.865[                    ] Update complete.
```

If I removed either of the YAML or watch binary, it would fail. Sans watch:

```
$ PATH="" ~/.pulumi-dev/bin/pulumi watch
error: Could not locate pulumi-watch binary
```

Sans YAML:

```
$ PATH="" ~/.pulumi-dev/bin/pulumi watch --debug -v9 --logflow
Watching (dev):
10:57:23.752[                    ] Updating...
10:57:24.117[                    ] Update failed.
```

Not a great UX, but I think that's nothing new.

AaronFriel:
With the CI changes resolved, this is in a good state for review @Frassle.

pkg/backend/watch.go
```diff
@@ -108,8 +102,83 @@ func Watch(ctx context.Context, b Backend, stack Stack, op UpdateOperation,
 			display.PrintfWithWatchPrefix(time.Now(), "",
 				op.Opts.Display.Color.Colorize(colors.SpecImportant+"Update complete."+colors.Reset+"\n"))
 		}
-
 	}
 
 	return nil
 }
+
+func watchPaths(root string, paths []string) (chan string, func(), result.Result) {
```
Frassle, line 110:
```suggestion
func watchPaths(root string, paths []string) (chan string, func(), error) {
```

This isn't ever returning results apart from errors

pkg/backend/watch.go
```diff
@@ -108,8 +102,83 @@ func Watch(ctx context.Context, b Backend, stack Stack, op UpdateOperation,
...
+
+	exePath, exeErr := os.Executable()
+	if exeErr == nil {
+		fullPath, fullErr := filepath.EvalSymlinks(exePath)
+		if fullErr == nil {
+			candidate := filepath.Join(filepath.Dir(fullPath), program)
+
+			// Let's see if the file is executable. On Windows, os.Stat() returns a mode of "-rw-rw-rw" so on
+			// on windows we just trust the fact that the .exe can actually be launched.
+			if stat, err := os.Stat(candidate); err == nil &&
```
Frassle, line 170: Should we give a better error here that the file was found but wasn't executable? Rather than just "Could not locate"

pkg/go.mod
```diff
@@ -37,7 +37,6 @@ require (
 	github.com/opentracing/opentracing-go v1.2.0
 	github.com/pgavlin/goldmark v1.1.33-0.20200616210433-b5eb04559386
 	github.com/pulumi/pulumi/sdk/v3 v3.38.0
-	github.com/rjeczalik/notify v0.9.2
```
Frassle, line 40: 🎉  any left over CGO stuff we can clean up as part of removing this?

AaronFriel: This is the last piece - I've got a follow up PR we can discuss this week that switches our build jobs to Linux only and removes CGO_ENABLED=1 and it did pass checks.

pkg/backend/watch.go
AaronFriel: Added that case:

```
fmt.Errorf("Could not locate an executable pulumi-watch, found %v without execute bit", fullPath)
```

pkg/backend/watch.go
AaronFriel: Yep, updated accordingly.

pkg/go.mod
AaronFriel: PR here, though it won't pass checks until this PR is in:
- https://github.com/pulumi/pulumi/pull/10427

Frassle (approved):

PR #10214, opened 2022-07-22
Make sure import pulumi/pulumi does not necessarily import typescript

t0yv0:
# Description

Continues the earlier work of trying not to pay for importing typescript when unused. Unfortunately further changes are needed. This statement, common in our JavaScript templates and many programs:

```
require("@pulumi/pulumi");
```

Pulls in "./runtime" which in turn pulls in "./closure/serializeClosure" which pulls in "typescript". 

This is only done so "./runtime" can re-export these two functions:

```
export {
    serializeFunctionAsync,
    serializeFunction,
    SerializedFunction,
    SerializeFunctionArgs,
}  from "./closure/serializeClosure";
```

I thought about making a lazy re-export here but I'm a bit rusty with my node to figure out how to safely do that. So instead I just mechanically switched "./closure" code to lazy-import either "typescript" or parts of itself until the problem went away.

Fixes # (issue)

## Checklist

### DIFF
--- a/sdk/nodejs/runtime/closure/createClosure.ts
+++ b/sdk/nodejs/runtime/closure/createClosure.ts
@@ -1,4 +1,4 @@
-// Copyright 2016-2018, Pulumi Corporation.
+// Copyright 2016-2022, Pulumi Corporation.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


sdk/nodejs/runtime/closure/createClosure.ts
```diff
@@ -19,7 +19,8 @@ import { ResourceError } from "../../errors";
 import { Input, isSecretOutput, Output } from "../../output";
 import * as resource from "../../resource";
 import { hasTrueBooleanMember } from "../../utils";
-import { CapturedPropertyChain, CapturedPropertyInfo, CapturedVariableMap, parseFunction } from "./parseFunction";
+import { CapturedPropertyChain, CapturedPropertyInfo, CapturedVariableMap } from "./parseFunction";
+import * as parseFunctionModule from "./parseFunction";
```
Frassle, line 23: Your importing here and calling require below? Doesn't this import trigger an un-conditional require?
Same with typescript in the other files.

t0yv0: It doesn't. Same technique as in https://github.com/pulumi/pulumi/pull/10185 

I was learning from https://www.typescriptlang.org/docs/handbook/modules.html#optional-module-loading-and-other-advanced-loading-scenarios and some experimentation.

When an import is "compile time only" like importing Interface types, it gets elided during translation to JavaScript.

So the pattern is:

```
import * as foo from "foo";  // compile-time only import IFF no foo.func() calls are made; erased from JS code

....
    if (cond) {
         var actualFoo: typeof foo = require("foo");  // require() gets compiled to actual JS code
         // since the typeof declaration above, actualFoo.callFunc() gets type-checked properly, it's not <any>
    }


```

Frassle (approved):

PR #10217, opened 2022-07-22
[dotnet] Set environment exit code during RunAsync

### LINKED ISSUE #8942: Set Environment.ExitCode = -1 when exception thrown in stack results in resources marked for deletion
## Hello!

## Issue details

I would like to suggest setting `Environment.ExitCode = -1` when Pulumi caught exception during stack results.

Background:
Pulumi can write with C# Top-level statement but if user forget to `return` then generated code is not `Task<int>` but `Task`.
This easily reproduce #7050.

Yes, I met this issue and deleted my resources.

```cs
using Pulumi; // and so on...

// bad example. Exception in stack result but ExitCode is 0, resources will be deleted.
await Deployment.RunAsync<MyStack>();

// good. Exception in stack result and ExitCode is 1, stop execution right after.
return await Deployment.RunAsync<MyStack>();
```

Sharplab without return. 
https://sharplab.io/#v2:EYLgZgpghgLgrgJwgZwLQGUCWBbADgGwgAUEB7AcwSm2QBoATEAagB8ABAJgEYBYAKDYAGAARsuAVgDc/fmwCcwgGKlSAOmWkAgsgCeAOwDGACgCU0vrIDMojkpX8A3v2EvR1sQDZRADlEeAPJh6MAB8dlq6hqbOrk58rgmiCmweqgAiEPhQOkYAKjgQ6LhQeupk2OgQBqR69MhGliZmMYmiAOzCXOYJAL78PUA=

Sharplab with return.
https://sharplab.io/#v2:EYLgZgpghgLgrgJwgZwLQGUCWBbADgGwgAUEB7AcwSm2QBoATEAagB8ABAJgEYBYAKDYAGAARsuAVgDc/fmwDsogJzCAYqVIA6NaQCCyAJ4A7AMYAKAJTS+sgMyiOq9fwDe/Ye9F2xANlEAOUW8AHkxDGAA+R10DEws3D1c+D2SlQI0AEQh8KH1TABUcCHRcKEMtMmx0CGNSQ3pkUxtzS3iU0QUuK2SAX35uoA==

### Affected area/feature

* sdk/dotnet

Zaid-Ajaj:
# Description

Fixes #8942 

Tested locally with a pulumi dotnet project that references the Pulumi dotnet SDK via a local reference and the output is now the same for `return await Deployment.RunAsync(...)` and just `await Deployment.RunAsync(...)`

## Checklist


Frassle:
What's this look like for inline automation api programs?

Zaid-Ajaj:
> What's this look like for inline automation api programs?

@Frassle Haven't looked into that yet. Will check it out...

sdk/dotnet/Pulumi/Deployment/Deployment_Inline.cs
```diff
@@ -38,7 +38,11 @@ private Deployment(InlineDeploymentSettings settings)
             Monitor = new GrpcMonitor(settings.MonitorAddr);
             deploymentLogger.LogDebug("Created deployment monitor");
 
-            _runner = new Runner(this, deploymentLogger);
+            // Tell the runner that we are running inside an inline automation program
```
Frassle, line 41: Does this mean that inline deployments that don't "return await" correctly can just hit the same issue this is trying to fix for normal deployments?

Frassle (approved):

PR #10220, opened 2022-07-22
Use plugin link instead of PATH manipulation in programtest

Frassle:
Little experiment to use plugin links rather than PATH manipulation.


Zaid-Ajaj (approved):
Looks good!

PR #10228, opened 2022-07-22
Add proto interfaces to move about logic to plugins

Frassle:
This adds the gRCP declarations to enable language plugins to return results to "pulumi about" rather than having each language hardcoded in the about code.

This _only_ adds the interfaces, it doesn't yet change "about" to use them. The intention is to ensure all the languages (including yaml/java) are updated to return results to these messages, and then to completely replace the current CLI about code with just calls to theses methods. As we version languages with the CLI we don't really need to be concerned with plugins returning unimplemented after that point.


Frassle:
I've got a branch ready to do built on top of this that fills these methods in properly for all the languages in pulumi/pulumi and guts `pulumi about` to just call the plugin. Comparison on top of this PR at https://github.com/pulumi/pulumi/compare/fraser/languageAboutRpc...fraser/languageAbout

Just to show that this rpc interface is sufficient to replace `about`.

Zaid-Ajaj (approved):
LGTM

t0yv0:
Modularity win here, thank you.

PR #10231, opened 2022-07-22
Matrix Testing

hliuson:
# Description

This pull request implements matrix testing, which extends integration testing in a handful of ways:
- Given a Pulumi YAML or PCL program, matrix testing automatically converts it into Pulumi programs in different languages and perform integration testing on them.
- Matrix testing can make use of the plugin link feature and ensure that the tested projects use the correct plugin binary. It additionally performs codegen using the plugin's schema and tests the generated projects using the generated SDKs from the schema.

Codegen is modified minorly to make this happen, as well as to enable specific testing of the core pulumi/pulumi language SDKs.

It additionally adds auxiliary stacks to integration tests which may be used to test a pulumi stack which depends on another stack via a stack reference.

## Checklist


pkg/codegen/go/gen_program.go
```diff
@@ -145,21 +146,34 @@ func GenerateProject(directory string, project workspace.Project, program *pcl.P
...
+				}
+			}
+		}
+	}
+
 	gomod.WriteString(`
-go 1.17
 
 require (
 	github.com/pulumi/pulumi/sdk/v3 v3.30.0
```
Frassle, line 173: Oh god's this is hard coded to 3.30! Pretty sure this should be floating latest like for dotnet.

pkg/codegen/go/gen_program.go
```diff
@@ -145,21 +146,34 @@ func GenerateProject(directory string, project workspace.Project, program *pcl.P
...
 	var gomod bytes.Buffer
 	gomod.WriteString("module " + project.Name.String() + "\n")
+	gomod.WriteString("go 1.17\n")
+	for name, sdk := range localProjects {
+		if name == "go" {
+			gomod.WriteString(fmt.Sprintf("replace github.com/pulumi/pulumi/sdk/v3/go => %s\n", sdk))
+		} else {
+			for _, p := range packages {
+				if p.Name == name {
+					gomod.WriteString(fmt.Sprintf("replace github.com/pulumi/pulumi-%s/sdk/go => %s/go\n", name, sdk))
```
Frassle, line 164: Similar to dotnet, I think it should be on the caller of this function to provide the full package paths

pkg/codegen/nodejs/gen.go
```diff
@@ -2183,6 +2182,10 @@ func genNPMPackageMetadata(pkg *schema.Package, info NodePackageInfo) string {
 		},
 	}
 
+	if !pkg.Test {
```
Frassle, line 2185: Add a comment about why this isn't included when Test is set

pkg/codegen/python/gen.go
```diff
@@ -1981,34 +1981,43 @@ func genPackageMetadata(
 	pluginVersion := version
 	info, ok := pkg.Language["python"].(PackageInfo)
 	if pkg.Version != nil && ok && info.RespectSchemaVersion {
-		version = pypiVersion(*pkg.Version)
+		version = pypiVersion(*pkg.Version) //Does not generate correct version names for pre-release versions.
```
Frassle, line 1984: Oh? Can we fix that? What's the problem?

pkg/testing/integration/program.go
```diff
@@ -1060,9 +1073,36 @@ func (pt *ProgramTester) TestCleanUp() {
 	}
 }
 
+//TODO: Allow matrix tester to pre-initialize stacks to avoid redundant pulumi ups.
```
Frassle, line 1076: Still TODO here?

pkg/testing/matrix/matrix.go
```diff
@@ -0,0 +1,457 @@
...
+	}
+	assert.NotNil(t, proj)
+	assert.NotNil(t, pclProgram)
+
+	localProjects := map[string]string{}
+
+	//Here we are overriding the plugin links to use our own plugin links.
+	proj.Plugins = &workspace.Plugins{}
+
+	//This needs further testing
```
Frassle, line 140: ?

pkg/testing/matrix/matrix.go
```diff
@@ -0,0 +1,457 @@
...
+		loader := schema.NewPluginLoader(ctx.Host)
+		_, pclProgram, err = yamlgen.Eject(pwd, loader)
+		assert.NoError(t, err)
+	}
+	assert.NotNil(t, proj)
+	assert.NotNil(t, pclProgram)
+
+	localProjects := map[string]string{}
+
+	//Here we are overriding the plugin links to use our own plugin links.
```
Frassle, line 137: Assert it's empty before setting it?

pkg/testing/matrix/matrix.go
```diff
@@ -0,0 +1,457 @@
...
+	proj.Plugins = &workspace.Plugins{}
+
+	//This needs further testing
+	for lang, path := range opts.PulumiSDKs {
+		localProjects[lang] = path
```
Frassle, line 142:
```suggestion
		localProjects["pulumi"] = path
```

pkg/testing/matrix/matrix.go
```diff
@@ -0,0 +1,457 @@
...
+		var spec schema.PackageSpec
+		err = json.Unmarshal(schemaBytes, &spec)
+		assert.NoError(t, err)
+
+		pkg, diags, err := schema.BindSpec(spec, nil)
+		assert.NoError(t, err)
+		assert.Empty(t, diags)
+		assert.NotNil(t, pkg)
+
+		//Kludge
```
Frassle, line 226: Not needed anymore?

pkg/testing/matrix/matrix.go
```diff
@@ -0,0 +1,457 @@
...
+	}
+
+	//Execute build commands and add plugin links
+	for _, plugin := range opts.Plugins {
+		for _, cmd := range plugin.Build {
+			err := cmd.Run()
+			assert.NoError(t, err)
+		}
+
+		localProjects[plugin.Name] = fmt.Sprintf("%s/%s-sdk", dir, plugin.Name)
```
Frassle, line 152: I think we want to do this per langauge, we know here that we're going to build SDKs into "/dotnet" folders etc because we do that explicitly below. No need to assume that all dotnet SDKs are inside a "dotnet" folder inside GenerateProgram.

pkg/testing/matrix/matrix.go
```diff
@@ -0,0 +1,457 @@
...
+				err = NodeConfigurePkg(pkg)
+				assert.NoError(t, err)
+				files, err = jsgen.GeneratePackage(pkgName, pkg, files)
+				assert.NoError(t, err)
+			case DOTNET:
+				err = DotnetConfigurePkg(pkg)
+				assert.NoError(t, err)
+				files, err = dotnetgen.GeneratePackage(pkgName, pkg, files)
+				assert.NoError(t, err)
+			//In the future we should support java but I hava no idea where to even start with that.
```
Frassle, line 261: That's fine we can circle back to java later

pkg/testing/matrix/matrix.go
```diff
@@ -0,0 +1,457 @@
...
+			default:
+				fmt.Printf("Unknown language: '%s'", lang)
+				continue
+			}
+
+			sdkDir := filepath.Join(dir, fmt.Sprintf("%s-sdk", pkgName), lang)
+			for p, file := range files {
+				// TODO: full conversion from path to filepath
```
Frassle, line 269: Still TODO?

pkg/testing/matrix/matrix.go
```diff
@@ -0,0 +1,457 @@
...
+				err = os.MkdirAll(filepath.Join(sdkDir, path.Dir(p)), 0700)
+				assert.NoError(t, err)
+				err = os.WriteFile(filepath.Join(sdkDir, p), file, 0600)
+				assert.NoError(t, err)
+			}
+			if lang == "go" {
```
Frassle, line 275: Why are these needed? This feels like something GeneratePackage ought to do? Do we need a new "generateOnly" flag to GeneratePackage so that this logic can live in the language code not here?

pkg/testing/matrix/matrix.go
```diff
@@ -0,0 +1,457 @@
...
+		})*/
+	}
+}
+
+func PythonConfigurePkg(pkg *schema.Package) error {
+	pkg.Version = &semver.Version{
+		Major: pkg.Version.Major,
+		Minor: pkg.Version.Minor,
+		Patch: pkg.Version.Patch,
+	} // Prune patch and build versions, since semver and python don't seem to agree on formatting.
```
Frassle, line 389: Hah they don't, playing with pulumictl for this over at https://github.com/pulumi/pulumictl/pull/51

pkg/testing/matrix/matrix.go
```diff
@@ -0,0 +1,457 @@
...
+	} else if notnil && !ismessage {
+		info, ok := raw.(jsgen.NodePackageInfo)
+		if !ok {
+			return fmt.Errorf("invalid nodejs package info")
+		}
+		nodePkg = info
+	} else if !notnil {
+		nodePkg = jsgen.NodePackageInfo{}
+	}
+	nodePkg.RespectSchemaVersion = true
```
Frassle, line 431: Why do node and python need this but nothing else? That seems inconsistent.

pkg/testing/matrix/matrix.go
```diff
@@ -0,0 +1,457 @@
...
+		langPtOpts = langPtOpts.With(i.ProgramTestOptions{
+			Dir: subdir,
+		})
+		t.Run(langOpt.Language, func(t *testing.T) {
+			i.ProgramTest(t, &langPtOpts)
+		})
+		//clean up subdir
+
+		//Including this seems to cause all tests to fail.
+		/*t.Cleanup(func() {
```
Frassle, line 378: Odd? What was the error?

pkg/testing/matrix/matrix_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+	"github.com/blang/semver"
+	i "github.com/pulumi/pulumi/pkg/v3/testing/integration"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/workspace"
+)
+
+func TestAll(t *testing.T) {
+	t.Parallel()
+	opts := []TestOptions{
+		/*{
```
Frassle, line 29: Can these all be enabled now?

pkg/testing/matrix/matrix_test.go
```diff
@@ -0,0 +1,122 @@
...
+						Initialized: false,
+					},
+				},
+			},
+			Languages: allLanguages(),
+		},*/
+		{
+			Program: &i.ProgramTestOptions{
+				Dir:              "tests/provider",
+				PulumiBin:        "/Users/hliuson/.pulumi-dev/bin/pulumi",
```
Frassle, line 62: Probably leave this unset for now, we might want to look at having the makefiles go build to ./bin so we can hardcode to that path here maybe 
(golangs lets just dump everything across all projects in to one big ~/go folder by default is a mad design)

pkg/testing/matrix/tests/empty/Pulumi.yaml
```diff
@@ -0,0 +1,4 @@
+name: empty
```
Frassle, line 1: The .dev.yaml files shouldn't need to be checked in.

sdk/go/common/workspace/plugins.go
```diff
@@ -1342,11 +1342,9 @@ func getPluginInfoAndPath(
 		for j, p2 := range projectPlugins {
 			if j < i {
 				if p2.Kind == p1.Kind && p2.Name == p1.Name {
-					if p1.Version != nil && p2.Version != nil && p2.Version.Equals(*p1.Version) {
-						return nil, "", fmt.Errorf(
-							"multiple project plugins with kind %s, name %s, version %s",
-							p1.Kind, p1.Name, p1.Version)
-					}
+					return nil, "", fmt.Errorf(
```
Frassle, line 1345: Not allowing aws@v4 and aws@v5 anymore? Per version stuff seemed to be working ok before.

pkg/codegen/go/gen_program.go
```diff
@@ -145,21 +146,34 @@ func GenerateProject(directory string, project workspace.Project, program *pcl.P
...
+	if err != nil {
+		return err
+	}
+
 	// Build a go.mod based on the packages used by program
 	var gomod bytes.Buffer
 	gomod.WriteString("module " + project.Name.String() + "\n")
+	gomod.WriteString("go 1.17\n")
+	for name, sdk := range localProjects {
+		if name == "go" {
```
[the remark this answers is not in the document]
harribo11: I've opted to use the language instead of "pulumi" for the language SDKs, because otherwise either 1. "pulumi" would point to different places for different tests (doable, imo slightly confusing but not the worst), or 2. "pulumi" would just point to p/p/sdk and then the rest of the path would have to be appended (makes assumptions about where the SDKs are). I think that it makes more sense to just do it this way, particularly if developing only one language SDK.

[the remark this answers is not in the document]
Frassle: Ah right but see comments further down that I think the matrix testing should be building this map per-language, and so both of these issues would be solved

pkg/testing/matrix/matrix.go
harribo11: The issue is that we need to have plugin links in the base YAML sometimes otherwise it can't be read into the program.

pkg/testing/matrix/matrix_test.go
harribo11: Stack references still cause the whole thing to panic, otherwise yes

sdk/go/common/workspace/plugins.go
harribo11: Maybe just match major versions?. When I was testing, the exact version matching was fairly brittle and would often cause the plugin to be missed.

harribo11: That can be a separate PR though

pkg/testing/matrix/matrix.go
harribo11: Seems to have resolved itself

pkg/testing/matrix/matrix.go
harribo11: I added this at some point but I don't remember why! Seems to work okay if I remove it.

pkg/testing/matrix/matrix.go
harribo11: Otherwise it just says ${VERSION} and pip/node freak out

pkg/codegen/dotnet/gen_program.go
```diff
@@ -157,7 +158,17 @@ func GenerateProgram(program *pcl.Program) (map[string][]byte, hcl.Diagnostics,
...
+	local, ok := localProjects[pkg]
+	if !ok {
+		return "", false
+	}
+	dotnet, ok := local["dotnet"]
+	return dotnet, ok
+}
+
+func GenerateProject(directory string, project workspace.Project, program *pcl.Program,
+	localProjects map[string]map[string]string) error {
```
Frassle, line 171:
```suggestion
	localProjects map[string]string) error {
```
Resolve to the "dotnet" map in the caller of this method. There's no point passing a map here that we're only going to look at the "dotnet" key in.

pkg/testing/matrix/matrix.go
Frassle: Right, should we not maintain those links though?

harribo11: Ah, figured out how to use the specified plugins in code to load from YAML. So now plugins should only be specified in one place.

sdk/go/common/workspace/plugins.go
AaronFriel: This looks like it would affect Pulumi YAML and plugin resolution in general.

harribo11: This specific check only looks at linked plugins - it was introduced in #10146

pkg/codegen/dotnet/gen_program.go
```diff
@@ -185,34 +187,55 @@ func GenerateProject(directory string, project workspace.Project, program *pcl.P
...
 
 	<ItemGroup>
-		<PackageReference Include="Pulumi" Version="3.*" />
 `)
 
-	// For each package add a PackageReference line
+	packageTemplate := "		<PackageReference Include=\"%s\" Version=\"%s\" />\n"
+	projectTemplate := "		<ProjectReference Include=\"%s\" />\n"
+
+	if localProjects == nil {
```
Frassle, line 195: This isn't needed? A lookup in a nil map just returns nil

pkg/codegen/go/gen_program.go
```diff
@@ -153,21 +154,35 @@ func GenerateProject(directory string, project workspace.Project, program *pcl.P
...
```
Frassle, line 167:
```suggestion
		if name == "pulumi" {
```

pkg/codegen/nodejs/gen_program.go
```diff
@@ -155,16 +157,24 @@ func GenerateProject(directory string, project workspace.Project, program *pcl.P
 	}
 	files["Pulumi.yaml"] = projectBytes
 
+	if localProjects == nil {
+		localProjects = map[string]string{}
+	}
+
 	// Build the pacakge.json
 	var packageJSON bytes.Buffer
+	pulumi := `@pulumi/pulumi": "^3.0.0`
```
Frassle, line 166:
```suggestion
	pulumi := `"@pulumi/pulumi": "^3.0.0"`
```
I thought this was a bugged at first glance, simpler to keep all the quotes together rather than wrapping the outer quotes as part of the fmt string below.

pkg/codegen/nodejs/gen_program.go
```diff
@@ -155,16 +157,24 @@ func GenerateProject(directory string, project workspace.Project, program *pcl.P
...
+	if localPulumi, ok := localProjects["pulumi"]; ok {
+		pulumi = fmt.Sprintf("@pulumi/pulumi\": \"file:%s", localPulumi)
```
Frassle, line 168:
```suggestion
		pulumi = fmt.Sprintf("\"@pulumi/pulumi\": \"file:%s\"", localPulumi)
```

pkg/codegen/nodejs/gen_program.go
```diff
@@ -155,16 +157,24 @@ func GenerateProject(directory string, project workspace.Project, program *pcl.P
...
+	}
 	packageJSON.WriteString(fmt.Sprintf(`{
 		"name": "%s",
 		"devDependencies": {
 			"@types/node": "^14"
 		},
 		"dependencies": {
 			"typescript": "^4.0.0",
-			"@pulumi/pulumi": "^3.0.0"`, project.Name.String()))
+			"%s"`, project.Name.String(), pulumi))
```
Frassle, line 177:
```suggestion
			%s`, project.Name.String(), pulumi))
```

pkg/testing/matrix/matrix.go
```diff
@@ -226,7 +226,12 @@ func Test(t *testing.T, opts TestOptions) {
 		assert.NoError(t, err)
 		assert.NotNil(t, provider)
 
-		schemaBytes, err := provider.GetSchema(int(plugin.Version.Major))
+		max := plugin.Version.Major
+		if max > 1 {
+			max = 1
+		}
+
+		schemaBytes, err := provider.GetSchema(int(max))
```
Frassle, line 234: Isn't this just always 1?

harribo11: I think it can be zero? Otherwise why would GetSchema take any params at all?

Frassle: @AaronFriel or @iwahbe might know this for sure

harribo11:
See https://github.com/pulumi/pulumi-digitalocean/pull/328/ for an example use of this.

pkg/testing/matrix/matrix.go
iwahbe: As far as I'm aware, it is always 0. This is how we generally call `GetSchema`:
https://github.com/pulumi/pulumi/blob/f618cf8c61a4df51e5e9a2741b3b5d05bbdd21d8/pkg/codegen/schema/loader.go#L246-L247

I assume that the parameter is just to allow us to future proof against breaking schema changes.

harribo11:
Outstanding Issues: @stack72 

- CI currently failing tests when fetching plugins, not entirely sure why. This might be one that we want to cut as it introduces external dependencies into tests. 
- May want to give users some configuration options, such as where to put the generated files, whether or not to cleanup afterwards, etc.
- Current method of building plugins is somewhat janky and we may just want to leave that to the user.
- Some languages fail pretty simple codegen tests, (see tests/scalar and tests/structured). tests/reference was previously broken but I think #10419 fixed it.
- A couple random //nolint statements

danielrbradley (commented):
I'd recommend closing this as this specific implementation of matrix testing is now unlikely to land due to staleness.

PR #10234, opened 2022-07-24
test: use `T.Setenv` to set env vars in tests

Juneezee:
# Description

This PR replaces `os.Setenv` with `t.Setenv`. Starting from Go 1.17, we can use `t.Setenv` to set environment variable in test. The environment variable is automatically restored to its original value when the test and all its subtests complete. This ensures that each test does not start with leftover environment variables from previous completed tests.

This saves us at least 2 lines (error check, and unsetting the env var) on every instance.

Reference: https://pkg.go.dev/testing#T.Setenv

## Checklist

- [ ] ~~I have added tests that prove my fix is effective or that my feature works~~: Tests refactoring only, no new features added.

- [ ] ~~I have updated the [CHANGELOG-PENDING](https://github.com/pulumi/pulumi/blob/master/CHANGELOG_PENDING.md) file with my change~~: This PR is a non user-facing change.

- [ ] ~~Yes, there are changes in this PR that warrants bumping the Pulumi Service API version~~: N/A

### DIFF
--- a/pkg/backend/filestate/backend_test.go
+++ b/pkg/backend/filestate/backend_test.go
@@ -155,15 +155,14 @@ func TestListStacksWithMultiplePassphrases(t *testing.T) {
 	assert.NoError(t, err)
 	assert.NotNil(t, aStack)
 	defer func() {
-		err = os.Setenv("PULUMI_CONFIG_PASSPHRASE", "abc123")
+		t.Setenv("PULUMI_CONFIG_PASSPHRASE", "abc123")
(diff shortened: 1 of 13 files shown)


Frassle (approved):
LGTM thanks for the contribution

AaronFriel:
@Juneezee thank you!

@Frassle this has another benefit: `t.SetEnv` panics in parallel tests, helps keep us from writing flaky tests.

PR #10235, opened 2022-07-24
Remove config from host, was unused

Frassle:
Simple change, this stuff just simply wasn't used so rip it all out.

### DIFF
--- a/sdk/go/common/resource/plugin/context.go
+++ b/sdk/go/common/resource/plugin/context.go
@@ -46,7 +46,7 @@ type Context struct {
 // that the host is "owned" by this context from here forwards, such
 // that when the context's resources are reclaimed, so too are the
 // host's.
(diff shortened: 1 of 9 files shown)


sdk/go/common/resource/plugin/context.go
```diff
@@ -46,7 +46,7 @@ type Context struct {
 // that the host is "owned" by this context from here forwards, such
 // that when the context's resources are reclaimed, so too are the
 // host's.
-func NewContext(d, statusD diag.Sink, host Host, cfg ConfigSource,
+func NewContext(d, statusD diag.Sink, host Host, _ ConfigSource,
```
Frassle, line 49: Note that pulumi-yaml calls this function so we can't remove the parameter.

Zaid-Ajaj (approved):

PR #10239, opened 2022-07-25
fix unmarshalling secrets raw

### LINKED ISSUE #10172: Encrypted fields are decrypted in Component Resource outputs
### What happened?

Pulumi is automatically decrypting fields that belong to a ComponentResource output. 

I have ComponentResources that create multiple AWS resources within them. Some of these resources return encrypted fields (an RDS instance for example returns a password that is encrypted). If my output is of type RDS (the schema is shown below), the password field is decrypted and is visible in plaintext.

Schema.JSON
```
"db": {
          "$ref": "/aws/v4.0.0/schema.json#/resources/aws:rds%2Finstance:Instance",
  }
```

ComponentResource output:
```
{
        address                              : "fakevalue.fakevalue.fakevalue.rds.amazonaws.com"
        .....
        password                             : foobartext
}
```

Pulumi-AWS output:
```
{
        address                              : "fakevalue.fakevalue.fakevalue.rds.amazonaws.com"
        .....
        password                             : [secret]
}
```

What's interesting here is that this only happens when the encrypted field is wrapped in a resource. If I return an encrypted field as a string output - it stays encrypted.    

### Steps to reproduce

Create a component resource.
Ensure it returns an output that is of the Type of another AWS resource.
Ensure one of the fields of the output resource is encrypted.

### Expected Behavior

The encrypted field stays encrypted

### Actual Behavior

The encrypted field is decrypted when returned by the Component Resource

### Versions used

_No response_

### Additional context

_No response_

### Contributing

dixler:
# Description

During Invoke, secrets are stripped and returned back in cleartext. This PR keeps them encrypted when returning them.

I've added tests and checked `nodejs` `python` sdk's v3.0.0, but was having difficulty checking the go sdk and dotnet SDK at versions v3.0.0

Fixes #10172 

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,3 +1,8 @@
+### Breaking
+
+- [engine] Engine always encrypts secret return values in Invoke
+  [#10239](https://github.com/pulumi/pulumi/pull/10239)
+
 ### Improvements
 
 - [auto/go] Adds the ability to capture incremental `stderr`
--- a/pkg/resource/deploy/source_eval.go
+++ b/pkg/resource/deploy/source_eval.go
@@ -670,6 +670,7 @@ func (rm *resmon) Invoke(ctx context.Context, req *pulumirpc.ResourceInvokeReque
 	mret, err := plugin.MarshalProperties(ret, plugin.MarshalOptions{
 		Label:         label,
 		KeepUnknowns:  true,
+		KeepSecrets:   true,
 		KeepResources: req.GetAcceptResources(),
 	})
 	if err != nil {
--- a/tests/integration/get_resource/dotnet/Program.cs
+++ b/tests/integration/get_resource/dotnet/Program.cs
@@ -7,6 +7,9 @@ using Pulumi.Random;
 
 class GetResource : CustomResource
 {
+    [Output("secret")]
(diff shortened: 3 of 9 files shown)


Frassle:
Note that `/run-integration-tests` is borked at the moment and won't run against PRs.

dixler:
@mikhailshilkov recommends I add a **Breaking** section to the changelog since we are forcing Invoke to return secrets where it previously was not.

Frassle (approved):
It looks like the SDKs have been ok with secrets since before 2.0 so just hardcoding this to true should be fine.

CHANGELOG_PENDING.md
```diff
@@ -1,3 +1,8 @@
+### Breaking
+
+- [engine] Engine always encrypts secret return values in Invoke
+  [#10239](https://github.com/pulumi/pulumi/pull/10239)
```
mikhailshilkov, line 4: I feel like this may not be clear enough for end users. Given the scary "Breaking" header, it would be nice to explain what exactly is broken.

dixler: I thought so too so I wrote and merged a second PR to fix my changelog entry. https://github.com/pulumi/pulumi/pull/10252

Issue #9824, opened 2022-06-09
Retry on `error: tls: bad record MAC` during plugin installation


lukehoban:
We've heard some reports of users hitting the following error occasionally:

```
error: installing [resource plugin aws-4.38.1] from : untarring file /root/.pulumi/plugins/resource-aws-v4.38.1/pulumi-resource-aws: local error: tls: bad record MAC
```

We likely should retry this failure if possible.

leviwolfe:
We just ran into this. We have been using Pulumi in GitHub actions successfully for months now. One build today failed, and I noticed this output during the `pnpm install` step:

```
.../node_modules/@pulumi/cloudflare install$ node scripts/install-pulumi-plugin.js resource cloudflare v4.0.0
.../node_modules/@pulumi/cloudflare install: [resource plugin cloudflare-4.0.0] installing
.../node_modules/@pulumi/cloudflare install: error: installing [resource plugin cloudflare-4.0.0] from : untarring file /home/runner/.pulumi/plugins/resource-cloudflare-v4.0.0/pulumi-resource-cloudflare: local error: tls: bad record MAC
.../node_modules/@pulumi/cloudflare install: There was an error installing the resource provider plugin. You may try to manually installing the plugin by running `pulumi plugin install resource cloudflare v4.0.0`
.../node_modules/@pulumi/cloudflare install: Done
```

Later in the build during the `pulumi/actions@v3` step, it reported:

```
   stderr: [resource plugin cloudflare-4.0.0] installing
  error: could not load plugin for cloudflare provider 'urn:pulumi:staging::XXXX::pulumi:providers:cloudflare::default_4_0_0': no resource plugin 'pulumi-resource-cloudflare' found in the workspace at version v4.0.0 or on your $PATH, install the plugin using `pulumi plugin install resource cloudflare v4.0.0`
  
   err?: 
```

Restarting our entire workflow a 2nd time succeeded.

Frassle:
We see this error occasionally in our CI testing as well

JoeStead:
This has also been happening to our org yesterday and today

PR #10248, opened 2022-07-26
Retry plugin downloads

Frassle:
# Description

We occasionally see network issues (most notably "tls: bad record MAC")
during plugin downloads. When grabbing plugins for codegen we already
had some retry logic. This shares that retry logic to the `pulumi plugin
install` command and the engine.

Fixes https://github.com/pulumi/pulumi/issues/9824

Note, plumbing in the progress bar and support for printing about
retries is a little bit odd but we're going to have to totally rethink
this interface at some point when the engine gets put behind a gRPC
interface and all the display logic has to be ripped out. So it works
good enough for now.

## Checklist

### DIFF
--- a/sdk/go/common/workspace/plugins.go
+++ b/sdk/go/common/workspace/plugins.go
@@ -809,6 +809,81 @@ func (info PluginInfo) Install(tgz io.ReadCloser, reinstall bool) error {
 	return info.InstallWithContext(context.Background(), tarPlugin{tgz}, reinstall)
 }
 
+// DownloadToFile downloads the given PluginInfo to a temporary file and returns that temporary file.
+// This has some
+func DownloadToFile(
+	pkgPlugin PluginInfo,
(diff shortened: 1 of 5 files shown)


Zaid-Ajaj (approved):
LGTM

sdk/go/common/workspace/plugins.go
```diff
@@ -809,6 +809,81 @@ func (info PluginInfo) Install(tgz io.ReadCloser, reinstall bool) error {
 	return info.InstallWithContext(context.Background(), tarPlugin{tgz}, reinstall)
 }
 
+// DownloadToFile downloads the given PluginInfo to a temporary file and returns that temporary file.
+// This has some
```
svangordon-fruit, line 813: Nit: incomplete comment?

Frassle: Huh, must of context switched halfway through writing that.

PR #10259, opened 2022-07-27
[sdk/nodejs] removed ts stack trace

### LINKED ISSUE #10224: [sdk/nodejs] simplify TypeScript compilation errors
## Hello!

## Issue details

It is a very common error to run `pulumi preview` or `pulumi up` on a program that is not a valid TypeScript program. Please consider reporting this situation concisely to the user instead of printing an irrelevant stack trace.

### Reproduce

- `pulumi new typescript`
- `echo fu >> index.ts`
- `echo bar >> index.ts`
- `pulumi preview`

### Current behavior

```
Previewing update (dev)

View Live: https://app.pulumi.com/t0yv0/node-err/dev/previews/0067e32c-68b7-45fa-901c-7e5b38ee7094

     Type                 Name          Plan       Info
 +   pulumi:pulumi:Stack  node-err-dev  create     1 error
 
Diagnostics:
  pulumi:pulumi:Stack (node-err-dev):
    error: Running program '/Users/anton/tmp/48h/node-err' failed with an unhandled exception:
    TSError: ⨯ Unable to compile TypeScript:
    index.ts(4,1): error TS2304: Cannot find name 'fu'.
    index.ts(6,1): error TS2304: Cannot find name 'bar'.
    
        at createTSError (/Users/anton/tmp/48h/node-err/node_modules/ts-node/src/index.ts:261:12)
        at getOutput (/Users/anton/tmp/48h/node-err/node_modules/ts-node/src/index.ts:367:40)
        at Object.compile (/Users/anton/tmp/48h/node-err/node_modules/ts-node/src/index.ts:558:11)
        at Module.m._compile (/Users/anton/tmp/48h/node-err/node_modules/ts-node/src/index.ts:439:43)
        at Module._extensions..js (node:internal/modules/cjs/loader:1157:10)
        at Object.require.extensions.<computed> [as .ts] (/Users/anton/tmp/48h/node-err/node_modules/ts-node/src/index.ts:442:12)
        at Module.load (node:internal/modules/cjs/loader:981:32)
        at Function.Module._load (node:internal/modules/cjs/loader:822:12)
        at Module.require (node:internal/modules/cjs/loader:1005:19)
        at require (node:internal/modules/cjs/helpers:102:18)
```

Note how the stack trace relates to Pulumi strategy of compiling TypeScript on the fly with `ts-node`. This stack trace is not in terms of the user program and is not related to recovering from the error (fixing the program).

For comparison, here is what `tsc` reports:

```
index.ts:4:1 - error TS2304: Cannot find name 'fu'.

4 fu
  ~~

index.ts:6:1 - error TS2304: Cannot find name 'bar'.

6 bar
  ~~~

Found 2 errors in the same file, starting at: index.ts:4
```

### Desired behavior

```
Previewing update (dev)

View Live: https://app.pulumi.com/t0yv0/node-err/dev/previews/0067e32c-68b7-45fa-901c-7e5b38ee7094

     Type                 Name          Plan       Info
 +   pulumi:pulumi:Stack  node-err-dev  create     1 error
 
Diagnostics:
  pulumi:pulumi:Stack (node-err-dev):
    error: Unable to compile TypeScript:
    index.ts(4,1): error TS2304: Cannot find name 'fu'.
    index.ts(6,1): error TS2304: Cannot find name 'bar'.
```

It definitely makes sense to remove the exception stack trace - it is noise. It makes sense to special-case the error handling code to make err message shorter to only retain useful information.

There are some approaches to displaying actual TypeScript errors: display none, display only the first one (+1 more), or display all as before.  It is difficult to compete with TypeScript itself in how to report the errors. But on the other hand displaying them in Pulumi output can save the user a hop (running `tsc`) and gives actionable context.  Therefore those should probably stay.

dixler:
# Description

This PR removes the stack trace output when a Typescript pulumi program has a compilation error.

example:
[asciicast](https://asciinema.org/a/NOXyPzY6Vu6OJEnT25m8d1Lxy)

Fixes #10224 

## Checklist


iwahbe (approved):
I have a question, but the result is much better. LGTM

sdk/nodejs/cmd/run/run.ts
```diff
@@ -235,7 +235,7 @@ export function run(
         const defaultMessage = err.stack || err.message || ("" + err);
 
         // First, log the error.
-        if (RunError.isInstance(err)) {
+        if (RunError.isInstance(err) || typeof err ==  typeof tsnode.TSError) {
```
iwahbe, line 238: I'm not super familiar with TS, but shouldn't it be:
```suggestion
        if (RunError.isInstance(err) || tsnode.TSError.isInstance(err)) {
```

dixler: https://github.com/pulumi/pulumi/blob/a92a005d6897506d3ea40307dfd37788c51a4dcf/sdk/nodejs/errors.ts#L38-L40

isInstance appears to be a specific method on the RunError class. I tried that one first :)

PR #10263, opened 2022-07-27
refactor: append to env var list instead of overwriting.

svangordon-fruit:
# Description

Refactor the `testing.Environment` `SetEnvVars` method to append to the slice of env vars, rather than overwriting it with a new slice. As per the behavior of `Cmd.Env`, later env vars will take precedence over newer env vars, and the `Env` field is public anyway so if you really wanted to clear the slice you still can. This enables the possibility of setting env vars in `NewEnvironment` without needing to worry that subsequent calls to `SetEnvVars` will erase those env vars.

Fixes # (issue)

## Checklist

### DIFF
--- a/sdk/go/common/testing/environment.go
+++ b/sdk/go/common/testing/environment.go
@@ -106,9 +106,13 @@ func (e *Environment) SetBackend(backend string) {
 	e.Backend = backend
 }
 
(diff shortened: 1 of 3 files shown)


Frassle (approved):
Seems reasonable to me

PR #10271, opened 2022-07-28
Error that version is needed if using --file

Frassle:
Just a small error improvement. Previously we would take the given name and kind and try and look it up on github, but that probably doesn't exist if we're installing from a file.


iwahbe (approved):
Assuming you have tested, LGTM.

PR #10272, opened 2022-07-28
Save random seeds to update plans

Frassle:
# Description

This generates and saves random seeds to update plans. When running an update with the plan the same seed that was used at preview time is passed to Check.

## Checklist


pkg/engine/lifecycletest/pulumi_test.go
```diff
@@ -5053,3 +5053,103 @@ func TestDefaultParents(t *testing.T) {
 	assert.Equal(t, snap.Resources[0].URN, snap.Resources[1].Parent)
 	assert.Equal(t, snap.Resources[0].URN, snap.Resources[2].Parent)
 }
+
+func TestProviderDeterministicPreview(t *testing.T) {
```
Frassle, line 5057: This is basically the same test I removed in https://github.com/pulumi/pulumi/pull/10203 but using randomSeeds for deterministic naming rather than sequence numbers.

Zaid-Ajaj (approved):
Had another go at the code, looks good to me

PR #10280, opened 2022-07-28
Update pulumi SDK version in go.mod in integration tests

### LINKED ISSUE #6807: Clean up go integration testing files
As part of the 3.0 beta program, we had to add go.mods with references to a single go commit in the testing/integration folder - we need to clean these up

mikhailshilkov:
# Description

Fixes #6807

## Checklist


mikhailshilkov:
Do these tests run in a PR?

Frassle:
> Do these tests run in a PR?

I don't think so.

iwahbe:
Instead of pinning a modern version of pulumi, which we will have to manually bump later, should we add a `replace` and have integration tests always run against the current SDK?

Frassle:
> Instead of pinning a modern version of pulumi, which we will have to manually bump later, should we add a replace and have integration tests always run against the current SDK?

Yes

mikhailshilkov:
Should I close this PR then? Do you want to adjust the original issue?

iwahbe:
> As part of the 3.0 beta program, we had to add go.mods with references to a single go commit in the testing/integration folder - we need to clean these up

The original issue just says "cleanup", so I think the `replace` statements qualify. You can add the `replace` statements, or you can close the PR and I can take it.

mikhailshilkov:
I won't have time to do that today, so I'll close it.

Issue #9677, opened 2022-05-24
help for pulumi stack isnt super informative


susanev:
this came up in a usability study today

the person deployed their first stack and then wanted to review the state of the stack in the cli. they looked at both pulumi and pulumi stack help and the docs but could not figure out what command would list out their resources.

could we provide more details than just `manage stacks` to make it clear to folks that using `pulumi stack` will list out their resources and outputs?

i can share a lil video clip and an email from the user with more deets about their experience if it would be useful to you

dixler:
help for `pulumi import` doesn't explain how to use it as well. we might want to repurpose this issue to look over all of the help text.

t0yv0:
@cnunciato could your team help here perhaps? This is what we have:

```
Manage stacks

A stack is a named update target, and a single project may have many of them.
Each stack has a configuration and update history associated with it, stored in
the workspace, in addition to a full checkpoint of the last known good update.

Usage:
  pulumi stack [flags]
  pulumi stack [command]

Available Commands:
  change-secrets-provider Change the secrets provider for a stack
  export                  Export a stack's deployment to standard out
  graph                   Export a stack's dependency graph to a file
  history                 Display history for a stack
  import                  Import a deployment from standard in into an existing stack
  init                    Create an empty stack with the given name, ready for updates
  ls                      List stacks
  output                  Show a stack's output properties
  rename                  Rename an existing stack
  rm                      Remove a stack and its configuration
  select                  Switch the current workspace to the given stack
  tag                     Manage stack tags
  unselect                Resets stack selection from the current workspace

Flags:
  -h, --help           help for stack
  -i, --show-ids       Display each resource's provider-assigned unique ID
      --show-name      Display only the stack name
      --show-secrets   Display stack outputs which are marked as secret in plaintext
  -u, --show-urns      Display each resource's Pulumi-assigned globally unique URN
  -s, --stack string   The name of the stack to operate on. Defaults to the current stack

Global Flags:
      --color string                 Colorize output. Choices are: always, never, raw, auto (default "auto")
  -C, --cwd string                   Run pulumi as if it had been started in another directory
      --disable-integrity-checking   Disable integrity checking of checkpoint files
  -e, --emoji                        Enable emojis in the output (default true)
      --logflow                      Flow log settings to child processes (like plugins)
      --logtostderr                  Log to stderr instead of to files
      --non-interactive              Disable interactive mode for all commands
[...]

PR #10287, opened 2022-07-29
[refactor] moved all command short+long description into its own file

dixler:
# Description

This PR moves all of the documentation strings for commands into its own directory to decouple the likelihood for documentation enhancements to impact command behavior.

tool to check for CLI help regressions
https://github.com/pulumi/home/tree/dixler/cli/help-output/ideation/engineering/48h/cli-ux

part of https://github.com/pulumi/pulumi/issues/10288
part of https://github.com/pulumi/pulumi/issues/9677

## Checklist

### DIFF
--- a/pkg/cmd/pulumi/about.go
+++ b/pkg/cmd/pulumi/about.go
@@ -61,22 +61,12 @@ func newAboutCmd() *cobra.Command {
 	var jsonOut bool
 	var transitiveDependencies bool
 	var stack string
-	short := "Print information about the Pulumi environment."
 	cmd :=
 		&cobra.Command{
 			Use:   "about",
-			Short: short,
-			Long: short + "\n" +
-				"\n" +
-				"Prints out information helpful for debugging the Pulumi CLI." +
-				"\n" +
-				"This includes information about:\n" +
-				" - the CLI and how it was built\n" +
-				" - which OS Pulumi was run from\n" +
-				" - the current project\n" +
-				" - the current stack\n" +
-				" - the current backend\n",
-			Args: cmdutil.MaximumNArgs(0),
+			Short: aboutText.Short,
+			Long:  aboutText.Long,
+			Args:  cmdutil.MaximumNArgs(0),
 			Run: cmdutil.RunFunc(func(cmd *cobra.Command, args []string) error {
 				ctx := context.Background()
 				summary := getSummaryAbout(ctx, transitiveDependencies, stack)
--- a/pkg/cmd/pulumi/cancel.go
+++ b/pkg/cmd/pulumi/cancel.go
@@ -32,15 +32,8 @@ func newCancelCmd() *cobra.Command {
 	var cmd = &cobra.Command{
 		Use:   "cancel [<stack-name>]",
 		Args:  cmdutil.MaximumNArgs(1),
-		Short: "Cancel a stack's currently running update, if any",
-		Long: "Cancel a stack's currently running update, if any.\n" +
-			"\n" +
-			"This command cancels the update currently being applied to a stack if any exists.\n" +
-			"Note that this operation is _very dangerous_, and may leave the stack in an\n" +
-			"inconsistent state if a resource operation was pending when the update was canceled.\n" +
-			"\n" +
-			"After this command completes successfully, the stack will be ready for further\n" +
-			"updates.",
+		Short: cancelText.Short,
+		Long:  cancelText.Long,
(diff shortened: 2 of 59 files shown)


Frassle:
I am not convinced that this is a good change. These strings are single use strings, there's no need for them to be named at all. Better to be inlined at the one single point of use. It also means any command change is even _less_ likely to update the help text because you won't even see the text when opening the file to work on it.

mikhailshilkov:
> decouple the likelihood for documentation enhancements to impact command behavior.

Could you please elaborate on what that means?

susanev:
i can speak to why this is helpful from one point of view...
- this would allow us to easily maintain consistency across messages
- this would allow ux designers to easily make usability improvements to these messages

i understand this introduces complexity for engineers working in specific files, as they now have to go to a different file to make a change to the help text. is there some kind of balance we can find between the two?

pgavlin:
In general I prefer leaving the documentation close to the commands as well, as it helps the readability of the source itself and helps ensure that the help text is updated. However, if we are to separate it out, I wonder if there's a way to make editing these strings a bit more convenient--even using raw string literals might suffice.

dixler:
Thanks. I appreciate the feedback because this is obviously a very contentious suggestion and may never see the light of day :)

> > decouple the likelihood for documentation enhancements to impact command behavior.
> 
> Could you please elaborate on what that means?

If I want to make a simple change to the help text, I know that the changes are scoped to just that file, I can make CI understand that it doesn't have to run all of the tests in order to check this.

> I am not convinced that this is a good change. These strings are single use strings, there's no need for them to be named at all. Better to be inlined at the one single point of use. 

If the single point of use, is 1 person 4 years later proofreading our documentation, then it would be best to put it in a single file as it would limit merge conflicts. I think how I organized it could be improved, but this was just an initial suggestion to create a discussion.

> It also means any command change is even _less_ likely to update the help text because you won't even see the text when opening the file to work on it.

> In general I prefer leaving the documentation close to the commands as well, as it helps the readability of the source itself and helps ensure that the help text is updated.

unfortunately in practice, 
`pulumi help` header hasn't been modified for [4 years](https://github.com/pulumi/pulumi/blame/10d4ef48f082450a6679c309fa308c547f40b341/pkg/cmd/pulumi/destroy.go#L63-L70)
`pulumi up --help` text hasn't been modified for 4 years. https://github.com/pulumi/pulumi/blame/master/pkg/cmd/pulumi/up.go
props to @mikhailshilkov for updating `pulumi destroy` last month as it hadn't been updated for 4-6 years

I'd like to rewrite the import help text. The long text is a useful comment on what `pulumi import` does, but after reading it, I'm still unsure how to use the command.
https://github.com/pulumi/pulumi/blob/master/pkg/cmd/pulumi/import.go#L311-L364

> However, if we are to separate it out, I wonder if there's a way to make editing these strings a bit more convenient--even using raw string literals might suffice.

> i can speak to why this is helpful from one point of view...
>
> this would allow us to easily maintain consistency across messages
this would allow ux designers to easily make usability improvements to these messages
i understand this introduces complexity for engineers working in specific files, as they now have to go to a different file to make a change to the help text. is there some kind of balance we can find between the two?

If we were to move the strings into markdown, it'd make it easier for us to style them and render them and make sure that it doesn't contain any logic in the program making testing simpler.

It should be very easy for users to suggest improvements to the CLI docs in the same way that users provide feedback and enhancements to the main docs and the majority of our users being nodejs/python and our codebase being in go doesn't necessarily help that.

I use the pulumi CLI much more than I read the site, the documentation, or the blogposts and I imagine a lot of users are like this since (imo) the selling point of the SDK is that the user never has to leave their IDE, but I still have to use the CLI in order to deploy. We should make sure that our help text accomplishes our goals.

*n.b.* I could write more, but I'm going to timebox myself as I think I've said most of what I could say.

AaronFriel:
@susanev & @dixler, your quotes here are the most compelling reason to move to this:

> i can speak to why this is helpful from one point of view...
> 
> * this would allow us to easily maintain consistency across messages
> * this would allow ux designers to easily make usability improvements to these messages

> It's a common strategy for video game developers who have to maintain a lot of text information and also helps with localization.

@dixler Out of curiosity, do you know of other terminal UIs/console applications written in Go that have localized messages that we could learn from for best practices? I18n support in our CLI is an ambitious goal, but maybe an i18n library would address @susanev's points even if we used it for just English support.

PR #10289, opened 2022-07-29
Update integration go mod

### LINKED ISSUE #6807: Clean up go integration testing files

iwahbe:
# Description

Fixes #6807

## Checklist


Frassle:
This is probably fine, but I'd suggest waiting till we come back from refresh week before merging just in case integration test workflows don't like it.

Frassle (approved):
Safe time to merge now

PR #10307, opened 2022-08-02
Pulumi bash support

deikzel:
# Description

I found it impractical to have to spin up a project myself every time that I want to create a cloud resource so I put together a command that allows me to create pulumi resources similar to `helm`.

This creates a single tool to provision any service in any cloud provider rather than having to know `aws-cli` `gcloud` `az` `kubectl` etc.

A nice extension to this would be adding `gen_completion` support to suggest fields for resources.

I thought I'd share. It still needs a lot more polish. Maybe y'all might want to adopt it given that YAML seems to be a huge success with the ops crowd.

Feel free to close this if it's not in the direction of your project.

## Syntax

```bash
pulumi install <resource-type> <resource-name> <prop1> ... <propN>
pulumi uninstall <resource-type> <resource-name>
```
## Example Usage
```bash
pulumi install provider:module:Resource my-resource foo=1 bar=2 baz={[,1,2,3,]} obj={[[,foo=2,bar={[,1,2,3,]},]]}
```

**Note**: unfortunately `zsh` overloads `[` and `[[`, but you can escape them with `'['` or `'[['`

shell brace expansion converts it into

```bash
foo=1
bar=2
baz=[
baz=1
baz=2
baz=3
baz=]
obj=[[
obj=foo=2
obj=bar=[
obj=bar=1
obj=bar=2
obj=bar=3
obj=bar=]
obj=]]
```

which is parsed into a Pulumi Yaml program

```yaml
name: global-stack_bXktcmVzb3VyY2U_cHJvdmlkZXI6bW9kdWxlOlJlc291cmNl
runtime: yaml
resources:
    my-resource:
        type: provider:module:Resource
        properties:
            bar: 2
            baz:
                - 1
                - 2
                - 3
            foo: 1
```

and saved to a directory on the user's machine and up'd. the stack names are deterministic based on the resource type and name so `pulumi destroy -s <stack-name>` can clean it up, but I wrote a shorthand for it `pulumi uninstall <resource-type> <resource-name>`.

Some annoyances are having to specify redundant values for every resource that can be configured via providers(gcp:project, gcp:location).

## Syntax
This explanation is somewhat vague. The syntax targets bash brace expansion rather than positional args.

### Assignment
key=value

### Lists
`[` indicates the following keys are part of a list
**Note:** they must all share the same prefix, because we expect shell expansion
`]` closes the list

It's oriented around shell expansion
`{[,` opens the list
`,]}` closes the list

example list: `lst={[,1,2,three,four]}` -> `[1,2,"three","four"]`
nesting a list in a list `{[,1,2,{[,5,6,]},]}` -> `[1,2,[5,6]]`

### Objects
`[[` indicates the following keys are part of an object
**Note:** they must all share the same prefix, because we expect shell expansion
`]]` closes the list

It's oriented around shell expansion
`{[[,` opens the object `,]]}` closes the object

example list: `lst={[,1,2,three,four]}` -> `[1,2,"three","four"]`
nesting a list in a list `{[,1,2,{[,5,6,]},]}` -> `[1,2,[5,6]]`

Fixes # (issue)

## Checklist


AaronFriel:
Hey @deikzel, I don't know what to say! That is a very clever, very powerful command.

We're definitely going to take a look at this and I'm going to take it back to the platform team to discuss adding this with some slight tweaks.

No promises, though we really appreciate the thought here.

blampe:
> The syntax targets bash brace expansion rather than positional args.

Alternatively, could you accept JSON and map that to a YAML manifest for the resource?

dixler:
surprise! I'm actually the author of this PR :confetti_ball:

dixler:
> > The syntax targets bash brace expansion rather than positional args.
> 
> Alternatively, could you accept JSON and map that to a YAML manifest for the resource?

100% it would make it much more portable and there could easily be an option to accept inline JSON.

I was orienting the command syntax around making it comfortable for operators to declaratively configure and imperatively deploy one-off resources with little/no friction. This PR attempts to accomplish this by
- not needing to create and manage a pulumi project.
- not needing to be in that project directory.
- minimizing the need for quoting and escaping because having to quote and escape things is not enjoyable.

Fortunately pulumi tracks state for us on managed backends so it'd likely be easy to migrate these resources to existing stacks to convert the imperative commands you did to set up an environment into reproducible IaC and lowers the barrier to entry of adopting IaC for imperative users.

I wouldn't necessarily recommend someone use this at scale as it doubles the resources under management since every resource is a stack.

a major limitation is that everything is a string in bash and differentiating between a string and a bool and a number can be ambiguous and that's where JSON would come in.

I'll be putting a full design doc together.

dixler:
I'm not currently working on this, but I'll leave it open in case there's more interest.

lunaris:
I'm going to close this as I don't think it's something we can look at right now (even though it's _very_ cool). There are definitely bits we could pick from this though, such as a command-line interface to providers, though we have since used `pulumi install` to mean a different thing, so this would need some tweaking.

PR #10317, opened 2022-08-05
codegen/go: Remove superfluous double forward slash in doc.go

stack72:
This is causing golint failures in new version of golangci-lint

```
doc.go:2: File is not `gofmt`-ed with `-s` (gofmt)
//
```

There's nothing else in the file

```
// A Pulumi package for creating and managing artifactory cloud resources.
//
package artifactory
```

# Description

Fixes # (issue)

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -13,3 +13,6 @@
 
 - Revert [Remove api/renewLease from startup crit path](pulumi/pulumi#10168) to fix #10293.
   [#10294](https://github.com/pulumi/pulumi/pull/10294)
+
+- [codegen/go] Remove superfluous double forward slash from doc.go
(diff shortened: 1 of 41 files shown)


mikhailshilkov (approved):
LGTM. This code has existed for years - is this is a new Go linter rule?

stack72:
> LGTM. This code has existed for years - is this is a new Go linter rule?

I'll be honest with you - I am not sure but we have started having builds fail...

ekini:
Looks like it's part of 1.19 release https://tip.golang.org/doc/go1.19#go-doc and `go fmt` now reformats docs, which it didn't do before, hence the error.

stack72:
> Looks like it's part of 1.19 release https://tip.golang.org/doc/go1.19#go-doc and `go fmt` now reformats docs, which it didn't do before, hence the error.

Ah that’s very useful to know! Thank you :)

PR #10323, opened 2022-08-08
Change pulumi schema to use PackageSpec

Frassle:
This ensures the schema is fully validated by the spec parser, the old
schema wasn't even codegen-able because of various fields not being
linked up correctly.


aq17 (approved):
LGTM, much cleaner 😄

PR #10334, opened 2022-08-09
Fix Python SDK apply causing program hangs

### LINKED ISSUE #10176: `pulumi preview` hangs on failure in chained Outputs

t0yv0:
# Description

Fixes #10176

## Checklist

### DIFF
--- a/sdk/python/lib/pulumi/output.py
+++ b/sdk/python/lib/pulumi/output.py
@@ -1,4 +1,4 @@
-# Copyright 2016-2018, Pulumi Corporation.
+# Copyright 2016-2022, Pulumi Corporation.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import asyncio
+import contextlib
 from functools import reduce
 from inspect import isawaitable
 from typing import (
@@ -219,19 +220,14 @@ class Output(Generic[T_co]):
                 result_is_secret.set_result(is_secret)
                 return cast(U, transformed)
             finally:
-                # Always resolve the future if it hasn't been done already.
-                if not result_is_known.done():
-                    # Try and set the result. This might fail if we're shutting down,
-                    # so swallow that error if that occurs.
-                    try:
-                        if not result_resources.done():
-                            result_resources.set_result(resources)
-                        if not result_is_known.done():
-                            result_is_known.set_result(False)
-                        if not result_is_secret.done():
-                            result_is_secret.set_result(False)
-                    except RuntimeError:
-                        pass
+                with contextlib.suppress(asyncio.InvalidStateError):
+                    result_resources.set_result(resources)
+
+                with contextlib.suppress(asyncio.InvalidStateError):
+                    result_is_known.set_result(False)
+
+                with contextlib.suppress(asyncio.InvalidStateError):
+                    result_is_secret.set_result(False)
 
         run_fut = asyncio.ensure_future(run())
         return Output(result_resources, run_fut, result_is_known, result_is_secret)
--- a/sdk/python/lib/pulumi/output.py
+++ b/sdk/python/lib/pulumi/output.py
@@ -224,9 +224,12 @@ class Output(Generic[T_co]):
                     # Try and set the result. This might fail if we're shutting down,
                     # so swallow that error if that occurs.
                     try:
-                        result_resources.set_result(resources)
-                        result_is_known.set_result(False)
-                        result_is_secret.set_result(False)
+                        if not result_resources.done():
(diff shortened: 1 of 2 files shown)


sdk/python/lib/pulumi/output.py
```diff
@@ -224,9 +224,12 @@ async def run() -> U:
                     # Try and set the result. This might fail if we're shutting down,
                     # so swallow that error if that occurs.
                     try:
-                        result_resources.set_result(resources)
-                        result_is_known.set_result(False)
-                        result_is_secret.set_result(False)
+                        if not result_resources.done():
+                            result_resources.set_result(resources)
+                        if not result_is_known.done():
```
t0yv0, line 229: Check on L223 `if not result_is_known.done():` is perhaps redundant here now.

sdk/python/lib/pulumi/output.py
```diff
@@ -224,9 +224,12 @@ async def run() -> U:
                     # Try and set the result. This might fail if we're shutting down,
```
t0yv0, line 224: I don't understand the "This might fail if we're shutting down" comment, perhaps we can try dropping the try..except now that we have extra .done() guards.

t0yv0: L202 await on _is_known was causing an exception after result_resources was already set; result_resources.set_result(resources) threw another exception short-circuiting backup setters for is_known and Is_secret.

sdk/python/lib/pulumi/output.py
Frassle: Perhaps not because this is all async. And on that all of these checks are probably unsafe and should be written in EAFP style instead:
```
try:
   result_resources.set_result(resources)
except asyncio.InvalidStateError:
   pass
```

Frassle:
So grim that this is 3 futures not one with 3 values 😬 but still good workaround for now

sdk/python/lib/pulumi/output.py
t0yv0: Ah in case a concurrent thread would do a set_result in between the lines? I think that shouldn't happen as these futures are owned by the `def` so there's no concurrent code that sets them.

Frassle: I think I'd still rather EAFP style, more pythonic

evanandrews-xrd: Irrespective of whether you take the EAFP or LBYL approach, there shouldn't be any race conditions since only one task is ever executing (in the Python interpreter) at a time, and the synchronisation points are explicitly specified where you `await`. I agree the check on L223 is redundant since there is no `await` between that check and this new one.

Regarding EAFP, I usually advocate for [contextlib.suppress](https://docs.python.org/3/library/contextlib.html#contextlib.suppress):

```python
with contextlib.suppress(asyncio.InvalidStateError):
    result_resources.set_result(resources)
with contextlib.suppress(asyncio.InvalidStateError):
    result_is_known.set_result(False)
with contextlib.suppress(asyncio.InvalidStateError):
    result_is_secret.set_result(False)
```

Frassle (approved):
LGTM

PR #10265, opened 2022-07-27
[sdk] deduplicate python error logs

dixler:
# Description

## Foreword

The approach in this PR, moves Program Error diagnostic logging logic into the SDK per each language rather than the engine itself. This handles the failure as close to the error as possible meaning that SDKs have more context for where in the program the failure occurred rather than bubbling the message up and can provide more useful insights to developers as they author their programs.

We currently do not differentiate between internal pulumi diagnostic logs and user program diagnostic logs.

The general pattern for Program error messages are as follows

```
  error: <Brief Summary of the Error type>
  user program error line 1
  ...
  user program error line N
```

Putting the error message on a new line, allows users to easily glance at the error text to see the error experiences that they are used to from running native programs instead of having to parse a line of errors. In academic terms, the parsing experience was proportional to `O(NumErrorLines*TermWidth) -> O(NumErrorLines)` which is a significant improvement.

Below are screen captures of the error experience.

### exception
[asciicast](https://asciinema.org/a/jbrNu2O5s495oZTmexWt2tSEF)

### provider error(this provider is from a POC of error messages)
[asciicast](https://asciinema.org/a/C9rrVpbHGp1ZwLNjesURxfzCA)

### compilation error
[asciicast](https://asciinema.org/a/w25z7vyZWimAWE8A5pNnxrYh0)

**Notes:**
- these are peculiar and are likely output at the top due to #10274 
### multi error 1
[asciicast](https://asciinema.org/a/iiTDtotYrDza8rIInC7WcQ0MP)

**Notes:**
- `exit status 1` is output by `go run` and it would require editing stdout or changing how we compile go programs to fix this

### hashicorp multierror
[asciicast](https://asciinema.org/a/btELGeJXBU2fk8pxnNBzHEJqn)

**Notes:**
- `exit status 1` is output by `go run` and it would require editing stdout or changing how we compile go programs to fix this

### provider error
[asciicast](https://asciinema.org/a/HrJHvXCLXO4F0C72O9WTmHKMy)

Fixes pulumi/pulumi#10222 
Fixes pulumi/pulumi#10227 

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works

- [ ] I have updated the [CHANGELOG-PENDING](https://github.com/pulumi/pulumi/blob/master/CHANGELOG_PENDING.md) file with my change

- [ ] Yes, there are changes in this PR that warrants bumping the Pulumi Service API version

PR #10336, opened 2022-08-09
[sdk/python] minimize internal pulumi stacks in python stacktrace

### LINKED ISSUE #10230: [sdk/python] remove pulumi callstacks from pulumi program Exception traceback
## Hello!

## Issue details

We currently report long stack traces from Pulumi Python programs. These stack traces contain frames with implementation details that may be helpful for the maintainers to understand what is happening but are not helpful to users, especially first time users.

Please consider removing the extraneous stack frames. This can be done with an escape hatch, perhaps when Troubleshooting (-v=9) with verbose logging as explained in https://www.pulumi.com/docs/support/troubleshooting/#verbose-logging Pulumi could still emit full stack traces.

### Repro

A very common user issue is a missing module - either venv was not setup correctly or pip has not been run. 

- ```pulumi new python```
- Edit __main__.py:

```
import pulumi
import pulumi_oci
```

- ```pulumi preview```

### Current behavior

```
Previewing update (dev)

View Live: https://app.pulumi.com/t0yv0/python-err/dev/previews/5b1762b3-3261-4cca-aa7f-43b76f8a83b8

     Type                 Name            Plan       Info
 +   pulumi:pulumi:Stack  python-err-dev  create     3 errors
 
Diagnostics:
  pulumi:pulumi:Stack (python-err-dev):
      File "/Users/anton/.nix-profile/bin/pulumi-language-python-exec", line 107, in <module>
        loop.run_until_complete(coro)
        return future.result()
      File "/Users/anton/tmp/48h/python-err/venv/lib/python3.8/site-packages/pulumi/runtime/stack.py", line 126, in run_in_stack
      File "/Users/anton/tmp/48h/python-err/venv/lib/python3.8/site-packages/pulumi/runtime/stack.py", line 49, in run_pulumi_func
        func()
      File "/Users/anton/tmp/48h/python-err/venv/lib/python3.8/site-packages/pulumi/runtime/stack.py", line 126, in <lambda>
      File "/Users/anton/tmp/48h/python-err/venv/lib/python3.8/site-packages/pulumi/runtime/stack.py", line 149, in __init__
        func()
      File "/Users/anton/.nix-profile/bin/pulumi-language-python-exec", line 106, in <lambda>
        coro = pulumi.runtime.run_in_stack(lambda: runpy.run_path(args.PROGRAM, run_name='__main__'))
      File "/nix/store/7q3qsarzyffcm2sjb2yhjgy0d3hzdi7y-python3-3.8.8/lib/python3.8/runpy.py", line 282, in run_path
        return _run_code(code, mod_globals, init_globals,
      File "/nix/store/7q3qsarzyffcm2sjb2yhjgy0d3hzdi7y-python3-3.8.8/lib/python3.8/runpy.py", line 87, in _run_code
        exec(code, run_globals)
      File "./__main__.py", line 2, in <module>
        import pulumi_oci
    ModuleNotFoundError: No module named 'pulumi_oci'
```

### Desired behavior

Basically remove all frames above __main__.py in the reported traceback. Those do not help the user.

```
View Live: https://app.pulumi.com/t0yv0/python-err/dev/previews/5b1762b3-3261-4cca-aa7f-43b76f8a83b8

     Type                 Name            Plan       Info
 +   pulumi:pulumi:Stack  python-err-dev  create     3 errors
 
Diagnostics:
  pulumi:pulumi:Stack (python-err-dev):
      File "./__main__.py", line 2, in <module>
        import pulumi_oci
    ModuleNotFoundError: No module named 'pulumi_oci'
```

### Implementation notes

I've dug a but into Python facilities for stack trace manipulation. It seems that `traceback` module can print traces and can remove frames from these traces. Here is an example program:

```python
import sys
import types
import traceback


class MyError(Exception):
    pass


def foo():
    bar()


def bar():
    baz()


def baz():
    raise MyError()


if __name__ == '__main__':
    try:
        foo()
    except:

        t, e, tb = sys.exc_info()
        traceback.print_exception(t, e, tb)

        print('*'*20 +' TRIMMED ' + '*'*20)

        s = traceback.extract_tb(tb)
        s = traceback.StackSummary.from_list(s[3:])
        print('Traceback (most recent call last):')
        print(''.join(s.format()).rstrip('\n'))
        traceback.print_exception(t, tb=None, value=e)
```

Perhaps this technique could be applicable here, though Pulumi might need to support Python 3.6 still so care is needed when testing on older versions so we do not break those. More Python expertise welcome here.

dixler:
# Description
This PR removes internal pulumi stack traces from the traceback output when a user's python pulumi program raises an Exception.

New:
[asciicast](https://asciinema.org/a/AdW7D8GcSeGsnckWer8v45QMV)

Old:
[asciicast](https://asciinema.org/a/TpqqgNE2kEp0A2qinB9L77LRZ)

Program:
```python

import pulumi
from pulumi_aws import s3

# Create an AWS resource (S3 Bucket)
bucket = s3.Bucket('my-bucket')
raise Exception("foobar")

pulumi.export('bucket_name', bucket.id)
```

Fixes #10230 
## Checklist


sdk/python/cmd/pulumi-language-python-exec
```diff
@@ -103,14 +104,40 @@ if __name__ == "__main__":
...
+                    # if the last module to run was runpy and the current one is a __main__.py
+                    break
+                count -= 1
+                last_module = cur_module
+            # keep only the count of last lines
+            return traceback.format_exc(limit=-count)
+
+        error_msg = "Program failed with an unhandled exception:\n" + clean_stacktrace()
+        pulumi.log.error(error_msg)
+        exit_code = 32
```
dixler, line 140: depends on #10265 for deduplicated output

sdk/python/cmd/pulumi-language-python-exec
```diff
@@ -107,14 +108,37 @@ if __name__ == "__main__":
...
+            # find the first occurrence of the module file name
+            runpy_filename = spec.loader.get_filename()
+            for extract in extracts:
+                cur_module = extract[0]
+                if cur_module.endswith('/__main__.py') and last_module == runpy_filename:
+                    # if the last module to run was runpy and the current one is a __main__.py
+                    break
+                count -= 1
+                last_module = cur_module
+            # keep only the count of last lines
```
dixler, line 138: I need to make this more defensive in the case that we don't detect anything.

sdk/python/cmd/pulumi-language-python-exec
```diff
@@ -107,14 +108,40 @@ if __name__ == "__main__":
...
+            spec = importlib.util.find_spec(runpy.__name__)
+            extracts = traceback.extract_tb(sys.exc_info()[2])
+            count = len(extracts)
+
+            last_module = None
+            # find the first occurrence of the module file name
+            runpy_filename = spec.loader.get_filename()
+            for extract in extracts:
+                cur_module = extract[0]
+                if cur_module.endswith('/__main__.py') and last_module == runpy_filename:
```
dixler, line 133: There may be safer ways to do this using reflection. I'll take a look.

t0yv0:
Could you resolve conflicts please to let tests run?

dixler:
@t0yv0 tests passed

sdk/python/cmd/pulumi-language-python-exec
```diff
@@ -107,14 +108,38 @@ if __name__ == "__main__":
...
+
         coro = pulumi.runtime.run_in_stack(lambda: runpy.run_path(args.PROGRAM, run_name='__main__'))
         loop.run_until_complete(coro)
         exit_code = 0
     except pulumi.RunError as e:
         pulumi.log.error(str(e))
     except Exception:
         # provide the user with traceback information for troubleshooting
-        error_msg = "Program failed with an unhandled exception:\n" + traceback.format_exc()
+        def get_user_stacktrace() -> str:
```
t0yv0, line 121: Style nit, could this be please lifted into top-level `def` prefixed with `_` to indicate internal; Python supports infinite nesting but I'd love us to get away from that style, it's easy to abuse. The new definition should carry a doc comment explaining what it's trying to do as it's highly unusual.

sdk/python/cmd/pulumi-language-python-exec
```diff
@@ -107,14 +108,38 @@ if __name__ == "__main__":
...
+
+            tb = traceback.extract_tb(sys.exc_info()[2])
+
+            for frame_index, frame in enumerate(tb):
+                # loop over stack frames until we reach the main program
+                # then return the traceback truncated to the user's code
+                cur_module = frame[0]
+
+                if get_abs_module_path(user_program_abspath) == get_abs_module_path(cur_module):
+                    # we have detected the start of a user's stack trace
```
t0yv0, line 133: The attempt to detect the start of the user stack trace seems quite fragile. It it always __main__? Are there other ways we could run the detection?

t0yv0:
If I specify in Pulumi.yaml

```
main: a.py
```

Then __main__ is not the startup module anymore.

t0yv0:
One idea that seems good to me is instead to look at a stack frame from `runpy.py` that does `run_path` because that's what we call right here, `pulumi.runtime.run_in_stack(lambda: runpy.run_path(args.PROGRAM, run_name='__main__'))`; if we found that stack trace, we skip everything up to that stack trace plus any subsequent stack traces from `runpy`. This detection would not need to worry about `__main__` naming.

sdk/python/cmd/pulumi-language-python-exec
dixler: > If I specify in Pulumi.yaml
> 
> ```
> main: a.py
> ```
> 
> Then `__main__` is not the startup module anymore.

I'm unsure because from the `Pulumi.yaml` file reference:
> main: (optional) an override for the main program’s location. By default, the program’s working directory is assumed to be the location of Pulumi.yaml. To choose a different location, use the main property. For example, if your Pulumi program is in a subdirectory infra/, use main: infra/.

https://www.pulumi.com/docs/reference/pulumi-yaml/

> One idea that seems good to me is instead to look at a stack frame from `runpy.py` that does `run_path` because that's what we call right here, `pulumi.runtime.run_in_stack(lambda: runpy.run_path(args.PROGRAM, run_name='__main__'))`; if we found that stack trace, we skip everything up to that stack trace plus any subsequent stack traces from `runpy`. This detection would not need to worry about `__main__` naming.

this binds our correctness to the internal implementation of `runpy` which I think is more fragile because it's controlled by python and is internal implementation and could change whereas we have control/knowledge of the function that we're trying to run.

> Could we also think of a test here of some sort? The improvement is very nice but the detector is something that can easily go out of whack over time as Python changes. A test (esp unit test) would be very helpful here.

A unit test would be good for this yes and could detect if the runpy implementation drifts if we go with that, but could also ensure that Pulumi.yaml and the program detection are in step.

sdk/python/cmd/pulumi-language-python-exec
```diff
@@ -108,37 +108,38 @@ if __name__ == "__main__":
...
-            # https://stackoverflow.com/questions/45771299/remove-several-lines-from-the-top-of-a-traceback
-
-            spec = importlib.util.find_spec(runpy.__name__)
-            extracts = traceback.extract_tb(sys.exc_info()[2])
-            count = len(extracts)
-
-            last_module = None
-            # find the first occurrence of the module file name
-            runpy_filename = spec.loader.get_filename()
-            for extract in extracts:
```
dixler: @t0yv0 for context, my old approach that depends on `runpy`

sdk/python/cmd/pulumi-language-python-exec
t0yv0: I'm pretty sure because I ran a test program. main: a.py renames entry-point. I also recall fixing bugs in this area. There's users doing this (renaming entrypoint).

Note that only your additions assume the entrypoint file is called __main__.py; in this code run_name parameter reconfigures the global var __name__ in the script, not the path to the script.

```
        coro = pulumi.runtime.run_in_stack(lambda: runpy.run_path(args.PROGRAM, run_name='__main__')
```

If you could please run this logic through a test program that species `main: a.py` and see if your changes still work with that, that'd be great.

> this binds our correctness to the internal implementation of runpy

True, but I bet runpy changes very slowly and would mostly call itself. The unit test would catch when this changes. I can concede to the current solution of scanning for __main__.py if it happens to work for programs that modify the entrypoint.

> A unit test would be good 

We'd want that regardless what we do here. 👍

dixler: Ok I can fix the `main: a.py` details. I'll also update the hugo documentation to include this detail.

dixler: added tests

tests/integration/python/stack_truncate/main_dir_specified/bar/requirements.txt
```diff
@@ -0,0 +1 @@
+pulumi
```
t0yv0, line 1: You might not need requirements.txt everywhere; this one is misplaced and ignored but the test still works I think.

t0yv0 (approved):
Perfect, this does the trick across all the scenarios we could think of.

tests/integration/python/stack_truncate/main_dir_specified/bar/requirements.txt
dixler: surprisingly, pulumi didn't recognize `requirements.txt` in the base directory for that example and crashed.

t0yv0: Huh, weird. OK then.

PR #10340, opened 2022-08-10
Expose external package cache

### LINKED ISSUE #10308: Potential Codegen regression between v3.33.1 and v3.36.0
When upgrading the pulumi-aws provider to v3.26.1 of the bridge, we effectively changed the version of pulumi from v3.33.1 to v3.36.0. The change also brought in the following PR https://github.com/pulumi/pulumi-terraform-bridge/pull/568 

This regression has caused the pulumi-aws build to go from ~20mins to approx ~80minutes. The publish stage of our binary happens 8 times (different OS / Arch combinations) so this makes it about 300-400 minutes which is causing timeouts

https://github.com/pulumi/pulumi-aws/actions/runs/2761874991

iwahbe:
# Description

As of #9933 we maintain a cache of globally analyzed external packages necessary for name resolution. To avoid unacceptable performance degradations, we cache that set of packages. 

This worked fine for SDK generation, but the cache didn't persist between ProgramGen calls (used to generate examples), which make up the vast majority of time spent.

This PR exposes that cache as a option to allow keeping the cache between invocations of ProgramGen. 

I'm including the new performance graph below for comparison. It doesn't render well here but can be downloaded to view.
profile001

Fixes #10308 

The performance improvement does require this patch to TF-bridge:
```patch
 pkg/tf2pulumi/convert/convert.go | 6 +++++-
 pkg/tfgen/docs.go                | 1 +
 pkg/tfgen/generate.go            | 2 ++
 3 files changed, 8 insertions(+), 1 deletion(-)

diff --git a/pkg/tf2pulumi/convert/convert.go b/pkg/tf2pulumi/convert/convert.go
index b5ba4db..a29888c 100644
--- a/pkg/tf2pulumi/convert/convert.go
+++ b/pkg/tf2pulumi/convert/convert.go
@@ -136,7 +136,9 @@ func Convert(opts Options) (map[string][]byte, Diagnostics, error) {
 		generatedFiles, genDiags, err = hcl2dotnet.GenerateProgram(program)
 		diagnostics = append(diagnostics, genDiags...)
 	case LanguageGo:
-		generatedFiles, genDiags, err = hcl2go.GenerateProgram(program)
+		generatedFiles, genDiags, err = hcl2go.GenerateProgramWithOptions(program, hcl2go.GenerateProgramOptions{
+			ExternalPackageCache: opts.GoPkgCache,
+		})
 		diagnostics = append(diagnostics, genDiags...)
 	case LanguageYaml:
 		generatedFiles, genDiags, err = hcl2yaml.GenerateProgram(program)
@@ -198,6 +200,8 @@ type Options struct {
 
 	// TargetOptions captures any target-specific options.
 	TargetOptions interface{}
+
+	GoPkgCache hcl2go.ExternalPackages
 }
 
 // logf writes a formatted message to the configured logger, if any.
diff --git a/pkg/tfgen/docs.go b/pkg/tfgen/docs.go
index 8fd40e7..cf56631 100644
--- a/pkg/tfgen/docs.go
+++ b/pkg/tfgen/docs.go
@@ -1162,6 +1162,7 @@ func (g *Generator) convert(input afero.Fs, languageName string) (files map[stri
 		ProviderInfoSource:       g.infoSource,
 		SkipResourceTypechecking: true,
 		TerraformVersion:         g.terraformVersion,
+		GoPkgCache:               g.goPkgCache,
 	})
 
 	return
diff --git a/pkg/tfgen/generate.go b/pkg/tfgen/generate.go
index 5b93f83..7e7c66f 100644
--- a/pkg/tfgen/generate.go
+++ b/pkg/tfgen/generate.go
@@ -68,6 +68,7 @@ type Generator struct {
 	skipDocs         bool
 	skipExamples     bool
 	coverageTracker  *CoverageTracker
+	goPkgCache       gogen.ExternalPackages
 
 	convertedCode map[string][]byte
 }
@@ -716,6 +717,7 @@ func NewGenerator(opts GeneratorOptions) (*Generator, error) {
 		skipDocs:         opts.SkipDocs,
 		skipExamples:     opts.SkipExamples,
 		coverageTracker:  opts.CoverageTracker,
+		goPkgCache:       gogen.ExternalPackages{},
 	}, nil
 }
 
```

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -22,3 +22,6 @@
 
 - [codegen/go] Remove superfluous double forward slash from doc.go
   [#10317](https://github.com/pulumi/pulumi/pull/10317)
+
+- [codegen/go] Add an external package cache option to `GenerateProgramOptions`.
+  [#10340](https://github.com/pulumi/pulumi/pull/10340)
--- a/pkg/codegen/go/doc.go
+++ b/pkg/codegen/go/doc.go
@@ -98,7 +98,7 @@ func (d DocLanguageHelper) GetLanguageTypeString(pkg *schema.Package, moduleName
 
 // GeneratePackagesMap generates a map of Go packages for resources, functions and types.
 func (d *DocLanguageHelper) GeneratePackagesMap(pkg *schema.Package, tool string, goInfo GoPackageInfo) {
-	d.packages = generatePackageContextMap(tool, pkg, goInfo)
+	d.packages = generatePackageContextMap(tool, pkg, goInfo, nil)
 }
 
 // GetPropertyName returns the property name specific to Go.
--- a/pkg/codegen/go/gen.go
+++ b/pkg/codegen/go/gen.go
@@ -747,9 +747,7 @@ func (pkg *pkgContext) contextForExternalReference(t schema.Type) (*pkgContext, 
 	}
 
 	var maps map[string]*pkgContext
-	if pkg.externalPackages == nil {
-		pkg.externalPackages = globalCache
-	}
+
 	hash := newExternalPackageHash(extPkg)
 	if extMap, ok := pkg.externalPackages.lookupContextMap(hash); ok {
 		maps = extMap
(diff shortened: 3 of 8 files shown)


AaronFriel (changes requested):
I'd like to see an interface here where we do something safe & performant by default for external callers, and handle the edge case where there are multiple external packages that would alias via the option you added to set the external package map.

We should guard the default (global) map by a mutex. I don't think codegen programs are ever long-lived and tend to involve repeatedly querying the same packages over and over. If memory usage is an issue, we can add some code to evict unused entries later.

AaronFriel (approved):

Issue #10310, opened 2022-08-03
AWS plugin download fails


pratikgajjar:
### What happened?

Pulumi keeps trying to download aws plugin and it fails every time.

### Steps to reproduce

No idea

### Expected Behavior

Pulumi should download plugin and start deployment 

### Actual Behavior

It failed

### Versions used

```
CLI
Version      3.37.2
Go Version   go1.17.12
Go Compiler  gc

Plugins
NAME    VERSION
aws     4.38.1
docker  3.2.0
nodejs  unknown

Host
OS       darwin
Version  12.4
Arch     arm64

This project is written in nodejs: executable='/nix/store/7zdpqkc149lxnakicprrw728px5gi1hk-nodejs-16.14.2/bin/node' version='v16.14.2'

Current Stack: prod

TYPE                                               URN
pulumi:pulumi:Stack                                urn:pulumi:prod::tyrion::pulumi:pulumi:Stack::tyrion-prod
awsx:ecr:Repository                                urn:pulumi:prod::tyrion::awsx:ecr:Repository::tyrion-cr
pulumi:providers:aws                               urn:pulumi:prod::tyrion::pulumi:providers:aws::default_4_15_0
awsx:x:ec2:Vpc                                     urn:pulumi:prod::tyrion::awsx:x:ec2:Vpc::default-vpc
awsx:x:ecs:Cluster                                 urn:pulumi:prod::tyrion::awsx:x:ecs:Cluster::prod-cluster
awsx:x:ecs:FargateService                          urn:pulumi:prod::tyrion::awsx:x:ecs:FargateService::tyrion-fs
awsx:x:ecs:FargateTaskDefinition                   urn:pulumi:prod::tyrion::awsx:x:ecs:FargateTaskDefinition::tyrion-fs
awsx:x:ec2:SecurityGroup                           urn:pulumi:prod::tyrion::awsx:x:ec2:SecurityGroup::tyrion-fs-0
aws:ec2/securityGroup:SecurityGroup                urn:pulumi:prod::tyrion::aws:ec2/securityGroup:SecurityGroup::tyrion-fs-0
aws:ecs/cluster:Cluster                            urn:pulumi:prod::tyrion::aws:ecs/cluster:Cluster::prod-cluster-4af2876
aws:ecr/repository:Repository                      urn:pulumi:prod::tyrion::awsx:ecr:Repository$aws:ecr/repository:Repository::tyrion-cr
[...]

PR #10341, opened 2022-08-10
Don't retry plugin write errors

Frassle:
# Description

No point retrying errors due to writing the plugin to a temporary file. This splits the error tracking up so we only retry read errors.

Fixes https://github.com/pulumi/pulumi/issues/10310

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -22,3 +22,6 @@
 
 - [codegen/go] Remove superfluous double forward slash from doc.go
   [#10317](https://github.com/pulumi/pulumi/pull/10317)
+
+- [cli/plugins] Don't retry plugin downloads that failed due to local file errors.
+  [#10341](https://github.com/pulumi/pulumi/pull/10341)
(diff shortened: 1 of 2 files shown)


Frassle:
> I have added tests that prove my fix is effective or that my feature works

I'd really like to re-write the plugin code to use something like `io/fs.FS` or https://github.com/spf13/afero to make testing this possible.

iwahbe (approved):
I would probably move this to its own function and introducing a 
```go
type readWriteError interface {
  error
  readError() error
  writeError() error
  asError() error { readError() or writeError() }
}
```
abstraction to wrap the concept. 

I think that is more a style preference, and you may take it or leave it. 

The code looks good and I appreciate the clarifying comment. LGTM.

Frassle:
> I think that is more a style preference, and you may take it or leave it.

I'd normally lean towards not adding an abstraction for a single use. I don't think this is going to crop up anywhere else (where else do we do downloads) so I'll just keep it local for now.

mikhailshilkov:
I fixes the retry part - but do we still need to address the write "file too large" aspect of #10310?

Frassle:
Yeh I'll leave the issue open to look into that aspect

Issue #10342, opened 2022-08-10
[sdks/python] Support optional config values in pulumi.Config


kpitzen:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Currently, all of the `config.get_x` methods accept only a string key, and return `Optional[<type>]` where `None` is returned if the config key is not set.  This requires users to add additional logic to provide default values to this configuration depending on the return value of the function.  Instead, it would be ideal if each of these methods could support an optional `default` parameter of their corresponding type to be returned if no config value is set for a given key.

ex:
Instead of:
```python
some_value = config.get("someKey") if config.get("someKey") is not None else my_default
```
we provide an interface that allows for:
```python
some_value = config.get("someKey", my_default)
```

### Affected area/feature

Pulumi Python SDK
This issue also blocks https://github.com/pulumi/pulumi/issues/10268

RobbieMcKinstry:
@kpitzen Forgive me if my Python-foo is out of date, but how would you update the signature to be non-breaking with the current signature?

I was originally picturing the API to look like this:
```python
config.get("someKey", default=my_default)
```

Are you intending to accept the second parameter with `*args`, accepting whatever the first unnamed argument is as the default?

kpitzen:
Obv haven't updated docstrings or anything yet, but you get the idea :)

RobbieMcKinstry:
Ah yes! Okay thank you, that answers my question! SGTM!

Issue #10268, opened 2022-07-27
[codegen/python] Non idiomatic default values for config


lukehoban:
The following YAML:

```yaml
configuration:
  containerPort:
    type: Number
    default: 80
  cpu:
    type: Number
    default: 512
  memory:
    type: Number
    default: 128
```

Currently emits this:

```python
config = pulumi.Config()
container_port = config.get_float("containerPort")
if container_port is None:
    container_port = 80
cpu = config.get_float("cpu")
if cpu is None:
    cpu = 512
memory = config.get_float("memory")
if memory is None:
    memory = 128
```

It should emit this:

```python
config = pulumi.Config()
container_port = config.get_int("containerPort") or 80
cpu = config.get_int("cpu") or 512
memory = config.get_int("memory") or 128
````

Frassle:
I'm not sure this is equivalent. `container_port = config.get_int("containerPort") or 80` will assign 80 if the config value is 0, while really if the config is set to 0 it should use 0.

It would be nice if `get_int` could take a default value to use though, like `container_port = config.get_int("containerPort", 80)`

PR #10344, opened 2022-08-10
feat: Supports optional arguments for pulumi.Config getters

kpitzen:
# Description

Fixes [#10342 ](https://github.com/pulumi/pulumi/issues/10342)

This adds optional arguments to each `pulumi.Config.get<_object|_int|_float|_bool>` method in the python SDK to handle fallback default values.  This interface is a bit more "pythonic" in that most `get` methods provide a default argument (see `dict.get` for a canonical example), and will allow us to generate both more simple and more idiomatic python code in the future (see https://github.com/pulumi/pulumi/issues/10268)

## Checklist

### DIFF
--- a/sdk/python/lib/pulumi/config.py
+++ b/sdk/python/lib/pulumi/config.py
@@ -66,21 +66,28 @@ class Config:
         #         f"use `{use.__name__}` instead of `{instead_of.__name__}`")
         return get_config(full_key)
 
-    def get(self, key: str) -> Optional[str]:
+    def get(self, key: str, default: Optional[str] = None) -> Optional[str]:
         """
-        Returns an optional configuration value by its key, or None if it doesn't exist.
+        Returns an optional configuration value by its key,
+        a default value if that key is unset and a default is provided,
+        or None if it doesn't exist.
 
         :param str key: The requested configuration key.
+        :param Optional[str] default: An optional fallback value to use if the given configuration key is not set.
         :return: The configuration key's value, or None if one does not exist.
         :rtype: Optional[str]
         """
-        return self._get(key, self.get_secret, self.get)
+        config_candidate = self._get(key, self.get_secret, self.get)
+        return config_candidate if config_candidate is not None else default
 
     def get_secret(self, key: str) -> Optional[Output[str]]:
         """
-        Returns an optional configuration value by its key, marked as a secret, or None if it doesn't exist.
+        Returns an optional configuration value by its key, marked as a secret,
+        a default value if that key is unset and a default is provided,
+        or None if it doesn't exist.
 
         :param str key: The requested configuration key.
+        :param Optional[str] default: An optional fallback value to use if the given configuration key is not set.
         :return: The configuration key's value, or None if one does not exist.
         :rtype: Optional[str]
         """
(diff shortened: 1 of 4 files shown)


sdk/python/lib/pulumi/config.py
```diff
@@ -66,21 +66,29 @@ def _get(
         #         f"use `{use.__name__}` instead of `{instead_of.__name__}`")
         return get_config(full_key)
 
-    def get(self, key: str) -> Optional[str]:
+    def get(self, key: str, default: Optional[Any] = None) -> Optional[str]:
```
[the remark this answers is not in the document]
kpitzen: Absolutely correct!

sdk/python/lib/pulumi/config.py
```diff
@@ -231,23 +266,41 @@ def _get_object(
...
+        :param Optional[Any] default: An optional fallback value to use if the given configuration key is not set.
+        :return: The configuration key's value, or None if one does not exist.
+        :rtype: Optional[Any]
+        :raises ConfigTypeError: The configuration value existed but couldn't be coerced to float.
         """
-        v = self._get_object(key)
-        if v is None:
-            return None
+        config_candidate = self._get_object(key)
+        v = config_candidate if config_candidate is not None else default
```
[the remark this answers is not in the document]
kpitzen: My curiosity got the better of me here:
image
image

`Optional[Any]` collapses down to `Any`, so mypy had no idea this would be an issue!

sdk/python/lib/pulumi/config.py
```diff
@@ -66,21 +66,28 @@ def _get(
...
-    def get(self, key: str) -> Optional[str]:
+    def get(self, key: str, default: Optional[str] = None) -> Optional[str]:
         """
-        Returns an optional configuration value by its key, or None if it doesn't exist.
+        Returns an optional configuration value by its key,
+        a default value if that key is unset and a default is provided,
+        or None if it doesn't exist.
 
         :param str key: The requested configuration key.
+        :param Optional[Any] default: An optional fallback value to use if the given configuration key is not set.
```
iwahbe, line 76:
```suggestion
        :param Optional[str] default: An optional fallback value to use if the given configuration key is not set.
```

sdk/python/lib/pulumi/config.py
```diff
@@ -66,21 +66,28 @@ def _get(
...
         :return: The configuration key's value, or None if one does not exist.
         :rtype: Optional[str]
         """
-        return self._get(key, self.get_secret, self.get)
+        config_candidate = self._get(key, self.get_secret, self.get)
+        return config_candidate if config_candidate is not None else default
 
     def get_secret(self, key: str) -> Optional[Output[str]]:
```
iwahbe, line 83: Where is the default parameter? I see it as a `:param` in the comments but not in the function signature.

iwahbe (approved):
I found 2 other things to fix. After that, LGTM!

sdk/python/lib/pulumi/config.py
kpitzen: Hey @iwahbe !  It's on line 69 here - `default: Optional[str] = None`

PR #10346, opened 2022-08-10
feat(codegen/python): Generate config getters with optional default arguments

### LINKED ISSUE #10268: [codegen/python] Non idiomatic default values for config

kpitzen:
# Description

Fixes #10268 

Modifies the existing python codegen logic to support the new `pulumi.Config.get` optional default arguments.  This should result in simpler looking python code when providing a default value.

Note: this PR requires https://github.com/pulumi/pulumi/pull/10344 to be merged first, otherwise we will generate invalid python from the SDK's perspective.

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -6,6 +6,9 @@
 - [sdk/python] Support optional default parameters in pulumi.Config
   [#10344](https://github.com/pulumi/pulumi/pull/10344)
 
+- [codegen/python] Use optional default arguments when generating `pulumi.Config.get` methods
+  [#10346](https://github.com/pulumi/pulumi/pull/10346)
+
 ### Bug Fixes
 
 - [cli] Paginate template options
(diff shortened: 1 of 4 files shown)


iwahbe:
This will mean that newly generated SDKs won't work with older versions of the Pulumi SDK, and thus moving to a new version of an SDK will be a breaking change. I don't know if we are ok with this. 

CC @mikhailshilkov 

The feature implementation itself looks good.

kpitzen:
@iwahbe Moving to a new version of the SDK should be non-breaking, since the old function signature is compatible with the new.  This does mean that generating new python code will require moving to the new SDK, however.

PR #10347, opened 2022-08-10
[sdk/go] deduplicate go error logs

### LINKED ISSUE #10227: [sdk/go] simplify Go program errors
## Hello!

## Issue details

Consider tidying up redundant information when displaying errors from Go programs.

### Reproduce

- ```pulumi new go```
- Edit main.go:
       
```go
package main

import (
	"fmt"

)

func main() {
		return fmt.Errorf("!")
	})
}
```

- `pulumi preview`

### Current behavior

```
Previewing update (dev)

View Live: https://app.pulumi.com/t0yv0/go-err/dev/previews/33a80085-858c-4439-929c-a59b59e92027

     Type                 Name        Plan       Info
 +   pulumi:pulumi:Stack  go-err-dev  create     1 error; 3 messages
 
Diagnostics:
  pulumi:pulumi:Stack (go-err-dev):
    error: program failed: 1 error occurred:
        * !
    exit status 1
 
```

### Desired behavior

It would be lovely to reduce "1 error; 3 messages" to just "1 error" as in other languages like TypeScript.

The `exit status 1` message is not helpful as it is duplicative with  `error: an unhandled error occurred: program exited with non-zero exit code: 1`. We should remove the duplication.

Copying earlier comment from Pat, we should only retain messaging around `exit code 1` when there is no other error information that's making it obvious that the Go program failed

So in the case of the code above we'd want: 

```
Previewing update (dev)

View Live: https://app.pulumi.com/t0yv0/go-err/dev/previews/33a80085-858c-4439-929c-a59b59e92027

     Type                 Name        Plan       Info
 +   pulumi:pulumi:Stack  go-err-dev  create     1 error
 
Diagnostics:
  pulumi:pulumi:Stack (go-err-dev):
    error: program failed: 1 error occurred:
        * !
```

dixler:
# Description
This PR deduplicates error messages for the Go SDK and uses exit status 32 for pulumi programs to indicate that an error has been logged and that the engine should `Bail` and not print any more output.

Fixes #10227 

#### program returned an error(the pulumi Run wrapper will os.Exit(32))
```go
package main

import (
	"fmt"

)

func main() {
		err := fmt.Errorf("!")
		return err
	})
}
```
[asciicast](https://asciinema.org/a/s5h7zHYWT7yqXWBrGvFd301hW)

### program exit code(!=32)
[asciicast](https://asciinema.org/a/4lIkHZVthYb3Cl7fpKJdJvEqG)

## Checklist


CHANGELOG_PENDING.md
```diff
@@ -3,6 +3,10 @@
 - [cli] Updated to the latest version of go-git.
   [#10330](https://github.com/pulumi/pulumi/pull/10330)
 
+- [sdk]
+   - Go SDK now properly outputs concise diagnostic error logs
+   [#10347](https://github.com/pulumi/pulumi/pull/10347)
+
```
dixler, line 9: clean up

sdk/go/pulumi-language-go/main.go
```diff
@@ -352,7 +352,25 @@ func (host *goLanguageHost) Run(ctx context.Context, req *pulumirpc.RunRequest)
...
+
+					// `go run` outputs details to stderr
+					err = errors.Errorf("problem executing program (Go compilation error)")
+				default:
+					// `go run` currently has only 3 exit codes. handle changes gracefully.
+					// - 0: success
+					// - 1: program exits with non-zero exit code(panic or pulumi.RunErr returned non-nil error)
+					// - 2: compilation error
+					err = errors.Errorf("program exited with non-zero exit code: %d", status.ExitStatus())
+				}
```
dixler, line 373: resolved by fixes in #10375

CHANGELOG_PENDING.md
```diff
@@ -54,6 +57,9 @@
 - [auto/go] Clone non-default branches (and tags).
   [#10285](https://github.com/pulumi/pulumi/pull/10285)
 
+- [cli/plugins] Don't retry plugin downloads that failed due to local file errors.
+  [#10341](https://github.com/pulumi/pulumi/pull/10341)
+
```
dixler, line 62:
```suggestion
```

sdk/go/pulumi-language-go/main.go
```diff
@@ -409,7 +415,7 @@ func (host *goLanguageHost) Run(ctx context.Context, req *pulumirpc.RunRequest)
 
 	program, err = compileProgramCwd(req.GetProject())
 	if err != nil {
-		return nil, errors.Wrap(err, "error in compiling Go")
+		return &pulumirpc.RunResponse{Error: "error in compiling Go"}, nil
```
dixler, line 418: this should probably be a run response error rather than just an error.

dixler:
Going to make screen recordings before requesting review.

sdk/go/pulumi/run.go
```diff
@@ -38,15 +38,22 @@ type RunOption func(*RunInfo)
 // to register resources and orchestrate deployment activities.  This connects back to the Pulumi engine using gRPC.
 // If the program fails, the process will be terminated and the function will not return.
 func Run(body RunFunc, opts ...RunOption) {
-	if err := RunErr(body, opts...); err != nil {
-		if err != ErrPlugins {
-			fmt.Fprintf(os.Stderr, "error: program failed: %v\n", err)
-			os.Exit(1)
-		}
+	err := RunErr(body, opts...)
```
t0yv0, line 41: So I'm reading this diff here and the only thing that really changed it seems like was substituting

```
fmt.Fprintf(os.Stderr, "error: program failed: %v\n", err)
```

with

```
ctx.Log.Error(fmt.Sprintf("an unhandled error occurred: program failed: \n%v", err), nil)
```

This change makes sense!

As to reordering if branches, it's nitty/style thing but from the sense of style in Go I've developed so far, the code was better before the change. We use this style all the time:

```
if <error-cond> {
   <handle error>
}
if <error-cond> {
   <handle error>
}
<handle happy path>
``` 

The new style here is

```
if <no err> {
    <handle happy path>
}
<handle error>
```

Seems a bit weird.

sdk/go/pulumi/run.go
```diff
@@ -38,15 +38,22 @@ type RunOption func(*RunInfo)
...
+		return
+	}
 
+	if err == ErrPlugins {
 		printRequiredPlugins()
 		os.Exit(0)
 	}
+
+	// Log the error message
+	if ctx, e := NewContext(context.TODO(), getEnvInfo()); e == nil {
```
t0yv0, line 52: Why is this code trying to recover from err in `NewContext` response? Is this something actually expected to happen? Perhaps `contract.IgnoreError` on it or even `contract.AssertNoError` - it would seem it should never really happen.

sdk/go/pulumi-language-go/main.go
```diff
@@ -353,8 +353,14 @@ func execProgramCmd(cmd *exec.Cmd, env []string) error {
 		return errors.Wrapf(err, "program exited unexpectedly")
 	}
 
+	if status.ExitStatus() == 1 {
```
t0yv0, line 356: Ugh, this does not feel right. Granted, up until recently Pulumi used `go run` therefore ExitStatus was always just either 1 or 2, with the real exit status printed to stderr, so existing programs don't rely on exit codes propagating out. Still. At least Pulumi used to always fail on non-0 exit codes from user code.

This can now swallow exit codes from unrelated conditions quite unexpectedly. 

Can we think of other options? Can we walk through what happens if we remove this and let the func return `return errors.Errorf("program exited with non-zero exit code: %d", status.ExitStatus())`?

If we cannot work this out cleanly, perhaps one crazy option is to overload the exit codes to use a <CUSTOM-INGEGER> reserved by Pulumi to indicate a special exit, that's unlikely to overlap with user code codes.

sdk/go/pulumi/run.go
dixler: Glad to be able to generate discussion around code style.

The current code looks like this
```
if <stmt>; <error> {
    if <special error> {
        <handle error>
        exit()
    }
    <handle error>
    exit()
}
<happy path>
```

The original code has no code in its happy path except `<stmt>`, but that's indented into the if statement so the conditions where the branch runs is confusing.

All the code for Run is error handling as it outsources the logic to `RunErr`

So it looks like `err != ErrPlugins` could possibly be `nil` but it's already guarded by the parent if statement so it doesn't exit at this location so the error handling logic would end up reading out of order since the error handling is at the bottom.

sdk/go/pulumi/run.go
t0yv0: Well the code is still swallowing the err, so it handles it. This seems to be not great, because there is no explanation why this err should be swallowed and not treated normally.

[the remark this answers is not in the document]
t0yv0: >  I wasn't sure about how to share the same result

Usually you do this by factoring out the parts of public functions into private function and recombining them back .

sdk/go/pulumi-language-go/main.go
t0yv0: I'm not sure I follow where RunResponse comes into play, but this is a blocker issue to me. Much more important not to swallow code 1 than to not print double error messages. What else can we do here?

Surfacing some discussion we had with @RobbieMcKinstry  also, what about having Run return exit code 0 if it reported the err to the engine? This seems to check out superficially, but you have more context on this:


```S1:

  - RunFunc returns non-nil err

    Context: func Run(body RunFunc, opts...RunOption)
   
    [[file:~/pulumi/sdk/go/pulumi/run.go::func Run(body RunFunc, opts ...RunOption) {]]

  - Run logs the error to the engine

  - Run finishes with exit code 0

  - pulumi-language-go finishes normally

  - but pulumi finishes with an error since error count != 0

  - only one error

S2:    

  - User code calls os.Exit(1) or panic() /// code=2 too, etc

  - Run does not even have a chance to run finalizers
    
  - pulumi-language-go finishes with exit code != 0

  - pulumi-language-go finishes with exit code != 0
      OR maybe it uses ctx.Log.Error to send it back to engine and finishes with exit code 0

  - only one error
```

t0yv0: Upon discussion @dixler pointed out that Pulumi does not consider it an error when the engine receives several ERROR-level logs. So this approach does not work. @Frassle pointed out we might need protocol changes for the user program to indicate error exit. In the meanwhile we decided with @dixler to be pragmatic here and finish off this PR by using a dedicated non-1 pulumi-only exit code to indicate a failure that already has sent logs to the engine.

t0yv0:
Ping me when ready to review. TYVM.

sdk/go/pulumi/run.go
```diff
@@ -81,7 +83,13 @@ func RunErr(body RunFunc, opts ...RunOption) error {
...
 	defer contract.IgnoreClose(ctx)
 
-	return RunWithContext(ctx, body)
+	err = RunWithContext(ctx, body)
+	// Log the error message
+	if err != nil {
+		err := ctx.Log.Error(fmt.Sprintf("an unhandled error occurred: program failed: \n%v", err), nil)
+		contract.AssertNoError(err)
+	}
+	return err
```
dixler, line 92: This is now in `RunErr` and can reuse the context.

tests/integration/integration_go_test.go
```diff
@@ -39,6 +39,27 @@ func TestNoEmitExitStatus(t *testing.T) {
...
+		Dependencies: []string{
+			"github.com/pulumi/pulumi/sdk/v3",
+		},
+		Stdout:        stdout,
+		Stderr:        stderr,
+		Quick:         true,
+		ExpectFailure: true,
+		ExtraRuntimeValidation: func(t *testing.T, stackInfo integration.RuntimeValidationStackInfo) {
+			// ensure exit status is not emitted by the program
+			errorCount := strings.Count(stderr.String()+stdout.String(), "  error: ")
```
dixler, line 56: the previous implementation logged directly to stdout hence why the `error:` text was not red. To catch and fail on this old behavior, I instead concatenated stdout/stderr and check it for error lines.

CHANGELOG_PENDING.md
```diff
@@ -1,4 +1,7 @@
 ### Improvements
 
+- [sdk/go] Go SDK now properly outputs concise diagnostic error logs
```
t0yv0, line 3: This is just a bit imprecise; the change affects just the failing Go programs right? Perhaps something like "Failing Go programs now display only one error message in Pulumi CLI"

sdk/go/pulumi-language-go/main.go
```diff
@@ -331,31 +332,54 @@ func (host *goLanguageHost) GetRequiredPlugins(ctx context.Context,
 	}, nil
 }
 
-func execProgramCmd(cmd *exec.Cmd, env []string) error {
+func runCmdStatus(cmd *exec.Cmd, env []string) (int, error) {
 	cmd.Env = env
 	cmd.Stdout, cmd.Stderr = os.Stdout, os.Stderr
 
 	err := cmd.Run()
```
t0yv0, line 339: This err handling is incomplete. It only catches subclasses of exec.ExitError. Is that ok?

sdk/go/pulumi-language-go/main.go
```diff
@@ -331,31 +332,54 @@ func (host *goLanguageHost) GetRequiredPlugins(ctx context.Context,
...
+	if status == 0 {
+		return &pulumirpc.RunResponse{}
 	}
 
-	// If the program ran, but exited with a non-zero error code. This will happen often, since user
-	// errors will trigger this.  So, the error message should look as nice as possible.
-	return errors.Errorf("program exited with non-zero exit code: %d", status.ExitStatus())
+	// If the program ran, but returned an error,
+	// the error message should look as nice as possible.
+	if status == constant.ExitStatusLoggedError {
```
t0yv0, line 371: Good

sdk/go/pulumi-language-go/main.go
```diff
@@ -331,31 +332,54 @@ func (host *goLanguageHost) GetRequiredPlugins(ctx context.Context,
...
+		// program failed but sent an error to the engine
+		return &pulumirpc.RunResponse{
+			Bail: true,
+		}
+	}
+
+	// If the program ran, but exited with a non-zero and non-reserved error code.
+	// indicate to the user which exit code the program returned.
+	return &pulumirpc.RunResponse{
+		Error: fmt.Sprintf("program exited with non-zero exit code: %d", status),
```
t0yv0, line 381: Stylistically (nit) it would be better for happy case to go to the bottom of the func, but this reads correct!

sdk/go/pulumi-language-go/main.go
```diff
@@ -395,7 +414,9 @@ func (host *goLanguageHost) Run(ctx context.Context, req *pulumirpc.RunRequest)
 		}
 
 		cmd := exec.Command(gobin, "run", cwd)
-		if err := execProgramCmd(cmd, env); err != nil {
+
+		status, err := runCmdStatus(cmd, env)
```
t0yv0, line 418: This looks right but would benefit from a comment that we use custom err handling here because go run mangles exit codes right.. So we don't expect e.g. 213 code from go run.

sdk/go/pulumi/run.go
```diff
@@ -38,15 +39,16 @@ type RunOption func(*RunInfo)
...
 // If the program fails, the process will be terminated and the function will not return.
 func Run(body RunFunc, opts ...RunOption) {
-	if err := RunErr(body, opts...); err != nil {
-		if err != ErrPlugins {
-			fmt.Fprintf(os.Stderr, "error: program failed: %v\n", err)
-			os.Exit(1)
-		}
+	err := RunErr(body, opts...)
+	if err == nil {
+		return
```
t0yv0, line 44: This is a change in behavior since the code used to do os.Exit(0). Do we need this change?

sdk/go/pulumi/run.go
```diff
@@ -38,15 +39,16 @@ type RunOption func(*RunInfo)
...
+	}
 
+	if err == ErrPlugins {
 		printRequiredPlugins()
```
t0yv0, line 48: This branch that does printRequiredPlugins changed to exit 213 from exit 0 but do we want that?

sdk/go/pulumi/run.go
```diff
@@ -38,15 +39,16 @@ type RunOption func(*RunInfo)
...
-		os.Exit(0)
 	}
+
+	os.Exit(constant.ExitStatusLoggedError)
```
t0yv0, line 51: OK so this is the new case! Good, but sounds like something else is doing logging, it's RunERr it seems.

sdk/go/pulumi/run.go
```diff
@@ -81,7 +83,13 @@ func RunErr(body RunFunc, opts ...RunOption) error {
 	}
 	defer contract.IgnoreClose(ctx)
 
-	return RunWithContext(ctx, body)
+	err = RunWithContext(ctx, body)
+	// Log the error message
+	if err != nil {
+		err := ctx.Log.Error(fmt.Sprintf("an unhandled error occurred: program failed: \n%v", err), nil)
```
t0yv0, line 89: Do we know who calls RunErr? This is a change in this function that now makes it log to ctx.Log; small change but still.

t0yv0: Sounds like folks use it in mock tests e.g. https://github.com/pulumi/examples/blob/ca40203279f393c0c159dadcadc97c6007122997/testing-unit-go/main_test.go#L36

t0yv0: We ask users to use RunErr e.g. in here https://www.pulumi.com/docs/guides/testing/unit/

t0yv0: So the change might be ok if it surfaces ok in those mock testing examples, but I'd be a bit worried without double-checking; it might be just easier to arrange code so that RunErr does not have the changed behavior.

t0yv0:
LGTM modulo nits and we need a bit of due diligence on RunErr change! TYVM.

t0yv0 (approved):

sdk/go/common/constant/exec_kind.go
```diff
@@ -11,3 +11,7 @@ const ExecKindAutoInline = "auto.inline"
 // ExecKindCLI is a flag used to indentify a command as originating
 // from the CLI using a traditional Pulumi project.
 const ExecKindCLI = "cli"
+
+// ExitStatusLoggedError is the exit status to indicate that a pulumi program
+// has failed, but successfully logged an error message to the engine
+const ExitStatusLoggedError = 213
```
dixler, line 17: I just recalled that when working on the other languages almost a month ago, I reserved 32 as an exit status. We should reuse this number for consistency's sake.

https://github.com/pulumi/pulumi/blob/490be1cc52673b7afe26cc4365e199005ed65d4c/sdk/python/cmd/pulumi-language-python/main.go#L67-L70

PR #10355, opened 2022-08-11
feat(codegen/python): Generate config getters with optional default arguments

kpitzen:
# Description

Fixes https://github.com/pulumi/pulumi/issues/10268

Modifies the existing python codegen logic to support the new pulumi.Config.get optional default arguments. This should result in simpler looking python code when providing a default value.

## Checklist

### DIFF
--- a/pkg/codegen/python/gen_program.go
+++ b/pkg/codegen/python/gen_program.go
@@ -553,12 +553,10 @@ func (g *generator) genConfigVariable(w io.Writer, v *pcl.ConfigVariable) {
 	g.genTemps(w, temps)
 
 	name := PyName(v.Name())
-	g.Fgenf(w, "%s%s = config.%s%s(\"%s\")\n", g.Indent, name, getOrRequire, getType, v.LogicalName())
 	if defaultValue != nil {
-		g.Fgenf(w, "%sif %s is None:\n", g.Indent, name)
-		g.Indented(func() {
-			g.Fgenf(w, "%s%s = %.v\n", g.Indent, name, defaultValue)
-		})
+		g.Fgenf(w, "%s%s = config.%s%s(\"%s\", %.v)\n", g.Indent, name, getOrRequire, getType, v.Name(), defaultValue)
+	} else {
+		g.Fgenf(w, "%s%s = config.%s%s(\"%s\")\n", g.Indent, name, getOrRequire, getType, v.Name())
 	}
 }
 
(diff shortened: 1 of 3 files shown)


kpitzen:
Reopening this due to my mistake re: order of operations for deleting old branches.

mikhailshilkov:
Can anybody guess why we haven't done this originally? Was there no `get` accepting a default value?

iwahbe:
> Can anybody guess why we haven't done this originally? Was there no `get` accepting a default value?

`get` accepting a default value was just added: https://github.com/pulumi/pulumi/pull/10344. I had some concerns about merging this, since any provider that takes this change then requires users to upgrade the version of the pulumi python SDK they are using, and I'm not sure what our breaking change policy is.

mikhailshilkov:
> get accepting a default value was just added: https://github.com/pulumi/pulumi/pull/10344.

Oh! I saw this PR but forgot, thank you.

AaronFriel:
@kpitzen @mikhailshilkov If we modified codegen to change this line in setup.py:

https://github.com/pulumi/pulumi-aws/blob/master/sdk/python/setup.py#L63

```python
      install_requires=[
          'parver>=0.2.1',
          'pulumi>=3.0.0,<4.0.0', # ⬅️
          'semver>=2.8.1'
      ],
```

To:


```python
      install_requires=[
          'parver>=0.2.1',
          'pulumi>=$codegenVersion',
          'semver>=2.8.1'
      ],
```

That seems like it would give us more flexibility like with our other languages.

Barring that, I'm not sure how we go about simultaneously adding new APIs to our SDKs and then rolling them out. Do we have a "must be supported for N versions" policy?

AaronFriel (approved):

pkg/codegen/python/gen_program.go
```diff
@@ -553,12 +553,14 @@ func (g *generator) genConfigVariable(w io.Writer, v *pcl.ConfigVariable) {
 	g.genTemps(w, temps)
 
 	name := PyName(v.Name())
+<<<<<<< HEAD
```
iwahbe, line 556: I don't think this merge was finished.

kpitzen: Oops - I was rebasing this branch by mistake and ended up pushing it half-way.  Let me tidy it up.

lunaris:
Closing this as I think we need a discussion about when/how we require users to bump the core SDK when they update a provider SDK. It came up recently when we introduced `@deprecated` as a decorator in #16436 and at that point we made the decision not to break.

PR #10370, opened 2022-08-12
fix: Replaces current fallback nodejs error messaging

kpitzen:
# Description

Currently, when Node encounters an error we haven't anticipated, our error messaging could potentially cause more frustration than necessary.  We should be sure to offer helpful error messaging when possible, and when not possible, we should strive to reduce frustration as much as possible.

## Checklist
I have not added tests, as this is purely a text-copy change.

### DIFF
--- a/sdk/nodejs/cmd/run-policy-pack/run.ts
+++ b/sdk/nodejs/cmd/run-policy-pack/run.ts
@@ -134,8 +134,8 @@ function throwOrPrintModuleLoadError(program: string, error: Error): void {
         }
     }
 
-    console.error("  * Yowzas, our sincere apologies, we haven't seen this before!");
-    console.error(`    Here is the raw exception message we received: ${error.message}`);
+    console.error("  * Pulumi encountered an unexpected error.");
+    console.error(`    Raw exception message: ${error.message}`);
     return;
 }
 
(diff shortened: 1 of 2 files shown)


kpitzen:
@susanev - I've added you for review as I'm not sure what our conventions are for this kind of error messaging.  Any input re: what would be best here would be appreciated!

Frassle (approved):
😬

sdk/nodejs/cmd/run-policy-pack/run.ts
```diff
@@ -134,7 +134,7 @@ function throwOrPrintModuleLoadError(program: string, error: Error): void {
         }
     }
 
-    console.error("  * Yowzas, our sincere apologies, we haven't seen this before!");
```
dixler, line 137: LMAO what? hahaha. boy our tone has changed.

sdk/nodejs/cmd/run/run.ts
```diff
@@ -157,7 +157,7 @@ function throwOrPrintModuleLoadError(program: string, error: Error): void {
         }
     }
 
-    console.error("  * Yowzas, our sincere apologies, we haven't seen this before!");
+    console.error("  * We've encountered an unexpected error.");
```
kpitzen, line 160: @susanev bringing the discussion from #10374 in here.  I'm not strongly opinionated on _what_ the wording actually is, so I'm happy to make the change here.  I could also see us messaging something like `Pulumi encountered an unexpected error.   Raw exception message:`.  Thoughts?

PR #10375, opened 2022-08-12
[sdk/go] enable direct compilation via `go build`

### LINKED ISSUE #10350: Consider explicit `go build` instead of `go run` for running Go programs
## Hello!

## Issue details

Consider removing `go run` style code execution in favor of `go build` followed by running the generated program.

As of pulumi 3.37.2 `pulumi-language-go` has two modes of executing user code, with `go run .` being by far the most common:

- `go run .`
- `./$binary` if .runtime.binary option is set in Pulumi.yaml

The execution happens every time Pulumi interacts with the user program (preview, up, etc), every command ends up calling into this.

The suggestion is to replace `go run .` with two steps, managing a local binary file explicitly:

```go build -o $program```
```./$program```

### Benefits
 
1. Better diagnostics. Currently `go run` hijacks exit codes, see for example https://github.com/golang/go/issues/17813 ; namely it returns exit code 0 for ok, 1 for program error, and 2 for compilation error. If the program exits with a non-0 exit code, instead of returning that code it prints it "exit code 3" for example. Not using `go run` would make it easy for Pulumi to distinguish between compile and runtime errors and print them nicely, and suppress redundant "exit code 3" messages from output (original discussion in https://github.com/pulumi/pulumi/pull/10265).

2. Better execution speed information in the metrics. We would be able to put separate spans and measure compilation/linking vs raw execution time to better identify where slowness is introduced.

3. Unlock path to a Go Build Cache https://github.com/pulumi/pulumi/issues/10473 that would speed up Go programs

### Design

- If PULUMI_GO_USE_RUN=true is set, use default behavior
- If it is not set (by default), use `go build -o $tempFile` to build and `exec $tempFile` to run
- Clean up $tempFile when done
- Note that on Windows the $tempFile must end in `.exe`

### Affected area/feature

Go language support

### PR DESCRIPTION
**Pull request by dixler**
# Description
This PR modifies `pulumi-language-go` to attempt to compile and run user programs first. If env var `PULUMI_GO_USE_RUN=true` is set, we use the old behavior of using `go run`.

Fixes #10350 

## Checklist

### CODE REVIEW

File: sdk/go/pulumi-language-go/main.go [hunk 1]
```diff
@@ -48,6 +49,10 @@ import (
 	pulumirpc "github.com/pulumi/pulumi/sdk/v3/proto/go"
 )
 
+const (
+	PULUMI_BINARY_NAME = ".pulumi.out"
```
dixler, line 53: opinions on whether it should be hidden or not?

sdk/go/pulumi-language-go/main.go
```diff
@@ -77,7 +82,17 @@ func findProgram(binary string) (*exec.Cmd, error) {
 		return nil, errors.Errorf("Failed to find go files for 'go run' matching %s", goFileSearchPattern)
 	}
 
-	return exec.Command(program, "run", cwd), nil
+	binaryToRun := path.Join(cwd, PULUMI_BINARY_NAME)
+	buildCmd := exec.Command(program, "build", "-o", binaryToRun, cwd)
+	if err := buildCmd.Run(); err != nil {
+		logging.V(5).Infof("Unable to `go build` falling back to `go run`")
+		// fallback to old behavior if failure. i.e.
+		// - compilation failure
```
dixler, line 90: trying to limit the number of changes in this PR. `go run` will fail with its typical behavior on compilation failures.

sdk/go/pulumi-language-go/main.go
```diff
@@ -77,7 +82,17 @@ func findProgram(binary string) (*exec.Cmd, error) {
...
+		// - directory is not writable
```
dixler, line 91: it was a pattern at my previous company that directories and files were not writable unless necessary. we can fallback on `go run` to avoid this.

sdk/go/pulumi-language-go/main.go
```diff
@@ -77,7 +82,17 @@ func findProgram(binary string) (*exec.Cmd, error) {
 		return nil, errors.Errorf("Failed to find go files for 'go run' matching %s", goFileSearchPattern)
 	}
 
-	return exec.Command(program, "run", cwd), nil
+	binaryToRun := path.Join(cwd, pulumiBinaryName)
```
Frassle, line 85: Write to temp, not cwd. Use https://pkg.go.dev/io/ioutil#TempFile to decide the full file path.

sdk/go/pulumi-language-go/main.go
Frassle: In what scenario could go run possibly work where go build doesn't? I think we can probably simplify and just remove it.

sdk/go/pulumi-language-go/main.go
Frassle: temp should always be writable, go run already writes it's compiled object to temp, we should do the same.

sdk/go/pulumi-language-go/main.go
dixler: SGTM. fixes a lot of limitations.

mikhailshilkov:
Is there a chance this is breaking to some users? If so, should we consider guarding this change behind a feature flag?

sdk/go/pulumi-language-go/main.go
```diff
@@ -339,10 +337,38 @@ func (host *goLanguageHost) Run(ctx context.Context, req *pulumirpc.RunRequest)
...
+		pulumiHome, err := workspace.GetPulumiHomeDir()
+		if err != nil {
+			return nil, err
+		}
+
+		file, err := ioutil.TempFile(pulumiHome, ".pulumi.out")
+		if err != nil {
+			return nil, errors.Wrap(err, "error in compiling Go")
+		}
+		defer os.Remove(file.Name())
```
t0yv0, line 356: So the os.Remove here negates the perf benefits asked for in the ticket https://github.com/pulumi/pulumi/issues/10350

Several Pulumi operations on the same program without editing source code will *still* relink the program and be slow in presence of os.Remove.

Why are we including it?


Why is it included?

sdk/go/pulumi-language-go/main.go
```diff
@@ -44,40 +45,37 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/logging"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/rpcutil"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/version"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/workspace"
 	"github.com/pulumi/pulumi/sdk/v3/go/pulumi"
 	pulumirpc "github.com/pulumi/pulumi/sdk/v3/proto/go"
 )
 
 func findProgram(binary string) (*exec.Cmd, error) {
-	// we default to execution via `go run`
```
t0yv0, line 52: Is this function now empty-body?

sdk/go/pulumi-language-go/main.go
```diff
@@ -339,10 +337,38 @@ func (host *goLanguageHost) Run(ctx context.Context, req *pulumirpc.RunRequest)
...
+		}
+
+		file, err := ioutil.TempFile(pulumiHome, ".pulumi.out")
+		if err != nil {
+			return nil, errors.Wrap(err, "error in compiling Go")
+		}
+		defer os.Remove(file.Name())
+
+		program = file.Name()
+		compileProgram(program)
```
t0yv0, line 359: Ignoring err here?

sdk/go/pulumi-language-go/main.go
```diff
@@ -339,10 +337,38 @@ func (host *goLanguageHost) Run(ctx context.Context, req *pulumirpc.RunRequest)
...
+	if program == "" {
+		// user did not specify a binary and we will compile and run the binary on-demand
+		logging.V(5).Infof("No prebuilt executable specified, attempting invocation via compilation")
+
+		pulumiHome, err := workspace.GetPulumiHomeDir()
+		if err != nil {
+			return nil, err
+		}
+
+		file, err := ioutil.TempFile(pulumiHome, ".pulumi.out")
```
t0yv0, line 352: Deprecated: As of Go 1.17, this function simply calls os.CreateTemp.

I think based on our discussion it should be "" indicating system temp folder, not pulumiHome; hogging ~/.pulumi/ with 100s of temp files is not great; we could use a subfolder like `path.Join(pulumiHome, "go-compiler-cache")` but then we'd need to include commands to manage/prune that into Pulumi; since that's currently out of scope perhaps just system TEMP is best.

dixler:
@t0yv0 sorry. was about to try reviewing #10394 and pushed some of my cleanup here. I forgot that I still had your review requested. I was working on this PR while the feedback was being made on Friday and I only saw that there was discussion EoD and I'm catching up on it.

sdk/go/pulumi-language-go/main.go
Frassle: I'd recommend just using TEMP for now. As I said on the issue maybe there's something clever we could do here to benefit from compiler caching but I wouldn't worry about that for a first pass (I don't think "go run" is benefiting from it?)

sdk/go/pulumi-language-go/main.go
dixler: I think there are unexpected risks around how users may be depending on `go run` over `go build` it'd be easier to ship a small change to test the waters and then build on top of it.

I've been told that changes to Go can have unexpected surprises given how our power users use it.

sdk/go/pulumi-language-go/main.go
dixler: will clean up.

t0yv0:
I'd love us to get on the same page.

With os.Remove the PR is not very interesting IMO, the chief perf motivation is not attained, and the disk space leak theory is not tested.

> I think there are unexpected risks around how users may be depending on go run over go build

I think what was suggested by @mikhailshilkov is:

> guarding this change behind a feature flag

That sounds prudent, I'm 👍  on that. If we go for a feature flag, we should just get on the same page whether we'd like to default to the new feature on (aggressive) or default to the feature off (conservative). I'm leaning to being aggressive here (default on). WDYT @mikhailshilkov @Frassle ?

The trouble around go run -> go build change would be mostly visible in either changed exit code behavior (but I think that should be internal), or disk space leak in TEMP (but I think OS should be cleaning this up)

mikhailshilkov:
>  If we go for a feature flag, we should just get on the same page whether we'd like to default to the new feature on (aggressive) or default to the feature off (conservative). I'm leaning to being aggressive here (default on). WDYT?

If we go conservative, do we think we can get mileage from this feature? If we know how to make some users give it a try, I'll vote conservative. If not, I guess we have to be aggressive, otherwise it effectively never ships.

Frassle:
> The trouble around go run -> go build change would be mostly visible in either changed exit code behavior (but I think that should be internal), or disk space leak in TEMP (but I think OS should be cleaning this up)

AFAIK "go run" just writes compiled files to TEMP and leaves them there, also it's unlikely anyone depends on the exit codes given it was the language runtime itself that was acting on them and translating into just an error for to send on grpc.

t0yv0:
Yeah with the conservative option we'd have to ask Cosmic and possibly announce on Pulumi Slack to solicit feedback based on opt-in. I'm with you Fraser that risks seem pretty low. So let's take the aggressive option. We'll have an ENV var to switch back to previous behavior, but default to the new behavior. We'll call out the change in CHANGELOG. Then ship and see if we get feedback.

dixler:
Some thoughts:

@Frassle if we want to take advantage of caching, I suggest putting the output binary in the root of the project directory rather than using TMP/PULUMI_HOME. We can know it ahead of time and
- PULUMI_HOME won't have to manage build artifacts
- /tmp isn't persisted across reboots so compilation perf benefits are lost

@mikhailshilkov As for breaking users:
- feature flagging SGTM. I'm currently using the name `PULUMI_USE_GO_RUN` and would like nitpicks on feature flag envvar names.
- I would like to link a GitHub issue in the error message for users to provide feedback if they run into problems and I have opened #10441 in anticipation.

t0yv0:
Placing files in current working directory may be a breaking change in practice (lack of permissions, interaction with .gitignore, breaking CI pipelines that check for pristine source tree, etc). I don't think we can do this. I believe we discussed it and ruled it out but we should have captured it in the reqs, looks like we missed that.

Worrying about reboots is out of scope. It will recompile after /tmp is cleaned up - that's fine.

Frassle:
> @Frassle if we want to take advantage of caching, I suggest putting the output binary in the root of the project directory rather than using TMP/PULUMI_HOME. We can know it ahead of time and

For this first pass we pretty much just want to copy what "go run" does (write to TEMP and leave it) but with an explicit build so we can get better error splits and exit codes.

We can re-visit this to see about getting some caching wins here on a second pass, but as Anton said there's a lot more risks around that.

> PULUMI_USE_GO_RUN 

I would do PULUMI_GO_USE_RUN, this matches other language envvars like PULUMI_PYTHON_CMD where it's PULUMI_<LANGUAGE>_<COMMAND>

dixler:
recording that shows

1. `ls PULUMI_HOME` doesn't contain the `go-compiler-cache` at the start
1. `pulumi pre` takes 1.7s without the cached binary
1. rerunning `pulumi pre` takes 1.1 seconds with the cached binary
1. the `go-compiler-cache` directory is created

https://asciinema.org/a/LqGvxrr3khnEtLhs9zJatH3ze

dixler:
Amending the error message to mention our changes to compilation doesn't seem worth it as it may create a lot of extra noise since differentiating between compilation errors and other errors seems like more effort than it's worth and users will make other Bug Reports that we can correlate.

sdk/go/pulumi-language-go/main.go
```diff
@@ -328,6 +344,39 @@ func (host *goLanguageHost) GetRequiredPlugins(ctx context.Context,
...
+		// Go program completed successfully
+		return nil
+	}
+
+	// error handling
+	exiterr, ok := err.(*exec.ExitError)
+	if !ok {
+		// We didn't even get to run the program.  This ought to never happen unless there's
+		// a bug or system condition that prevented us from running the language exec.  Issue a scarier error.
+		return errors.Wrapf(err, "problem executing program (could not run language executor)")
```
Frassle, line 362: " (could not run language executor)" isn't really accurate here anymore.

sdk/go/pulumi-language-go/main.go
```diff
@@ -337,33 +386,54 @@ func (host *goLanguageHost) Run(ctx context.Context, req *pulumirpc.RunRequest)
...
 	}
 
-	cmd, err := findProgram(host.binary)
-	if err != nil {
-		return nil, err
+	if os.Getenv("PULUMI_GO_USE_RUN") != "" {
+		// feature flag to enable old behavior and use `go run`
+		gobin, err := executable.FindExecutable("go")
+		if err != nil {
+			return nil, err
```
Frassle, line 393:
```suggestion
			return nil, fmt.Errorf("could not run language executor: %w", err)
```

sdk/go/pulumi-language-go/main.go
```diff
@@ -42,40 +43,55 @@ import (
...
+		return "", errors.Errorf("Failed to find go files for 'go build' matching %s", goFileSearchPattern)
+	}
+
+	pulumiHomeDir, err := workspace.GetPulumiHomeDir()
+	if err != nil {
+		return "", errors.Wrap(err, "failed to compile")
 	}
 
-	return exec.Command(program, "run", cwd), nil
+	buildDir := path.Join(
```
Frassle, line 72: Don't build to PULUMI_HOME, build to temp

sdk/go/pulumi-language-go/main.go
```diff
@@ -46,36 +53,65 @@ import (
...
-		return nil, errors.Errorf("Failed to find go files for 'go run' matching %s", goFileSearchPattern)
+		return "", errors.Errorf("Failed to find go files for 'go build' matching %s", goFileSearchPattern)
+	}
+
+	if err != nil {
+		return "", errors.Wrap(err, "failed to compile")
 	}
 
-	return exec.Command(program, "run", cwd), nil
+	hash := hashDir(cwd)
```
Frassle, line 92: I would rather we don't try to be smart here for the first pass. Just write to a fresh file in temp, we're not going to be any worse perf wise as that's all "go run" does.

Content hashing like this requires us to understand _all_ content that could possibly feed into the go compilers build process. That includes things like referenced go modules which *do not* live under the source tree of program being built.

dixler: This is just for naming the output binary destination. The compiler will overwrite the existing binary if it needs to.

cc @t0yv0

Frassle: So don't waste proc time building a hash, just let the temp file API do its job of returning an unused name.

sdk/go/pulumi-language-go/main.go
```diff
@@ -337,33 +400,55 @@ func (host *goLanguageHost) Run(ctx context.Context, req *pulumirpc.RunRequest)
...
+	program := host.binary
+	if program == "" {
+		// user did not specify a binary and we will compile and run the binary on-demand
+		logging.V(5).Infof("No prebuilt executable specified, attempting invocation via compilation")
+
+		program, err = compileProgramCwd(req.GetProject())
+		if err != nil {
+			return nil, errors.Wrap(err, "error in compiling Go")
 		}
+		//defer os.RemoveAll(program)
```
Frassle, line 436: After checking out "go run" some more, it seems it does delete the program after running. So I think we do want a Remove call.

t0yv0: Deleting the program after running negates the raison d'etre of this PR - faster execution benefits listed in https://github.com/pulumi/pulumi/issues/10350

We could ship it as a separate change, but I'm for shipping it in as part of this. It's about risk. The PR compounds the risks of changed exit codes and possible TEMP saturation. But so far both appear quite tolerable. Shall we just ship as is?

Frassle: You still get other benefits for measurements and exit codes just doing the simple thing here and building then deleting. I think _caching_ builds needs some more thinking and should be a separate PR probably under an opt-in flag to start.

sdk/go/pulumi-language-go/main.go
```diff
@@ -46,36 +53,65 @@ import (
 	pulumirpc "github.com/pulumi/pulumi/sdk/v3/proto/go"
 )
 
-func findProgram(binary string) (*exec.Cmd, error) {
-	// we default to execution via `go run`
-	// the user can explicitly opt in to using a binary executable by specifying
-	// runtime.options.binary in the Pulumi.yaml
-	if binary != "" {
-		program, err := executable.FindExecutable(binary)
+func hashDir(path string) string {
```
AaronFriel, line 56: `go build` automatically recompiles if dependencies change, including file contents. If we hash the dir, we will spend a lot of time in the `.git` folder, non-content files.

My `pulumi` dir is 8.5GiB and contains 399,072 files. We don't want to hash all of that.

sdk/go/pulumi-language-go/main.go
```diff
@@ -337,33 +400,55 @@ func (host *goLanguageHost) Run(ctx context.Context, req *pulumirpc.RunRequest)
 		return nil, errors.Wrap(err, "failed to prepare environment")
 	}
 
-	cmd, err := findProgram(host.binary)
-	if err != nil {
-		return nil, err
+	if os.Getenv("PULUMI_GO_USE_RUN") != "" {
+		// feature flag to enable old behavior and use `go run`
+		gobin, err := executable.FindExecutable("go")
```
AaronFriel, line 405: On Windows, this will have a `.exe` ending.

sdk/go/pulumi-language-go/main.go
```diff
@@ -46,36 +53,65 @@ import (
...
+	if runtime.GOOS == "windows" && !strings.HasSuffix(filename, ".exe") {
+		filename = fmt.Sprintf("%s.exe", filename)
+	}
+
+	outfile := path.Join(
+		os.TempDir(),
+		filename,
+	)
+
+	gobin, err := executable.FindExecutable("go")
```
AaronFriel, line 103: Likewise, `.exe` ending on Windows.

sdk/go/pulumi-language-go/main.go
AaronFriel: The difference and reason for this PR as I understand it is that we can save time linking (what `go build` can optimize for) vs `go run`.

AaronFriel (commented):
@dixler @t0yv0 could we use a `runtime.options.build = "./path/to/bin"` option as an opt in and preserve all existing functionality? Then we can test it more widely and ask some of our customers to try it out before making a change that will affect all users of automation API & go runtime.

It derisks the approach and enables users with performance sensitive workflows to opt in to a path that avoids relinking.

CHANGELOG_PENDING.md
```diff
@@ -16,6 +16,9 @@
 - [codegen] Reduce time to execute `pulumi convert` and some YAML programs, depending on providers used, by up to 3 seconds.
   [#10444](https://github.com/pulumi/pulumi/pull/10444)
 
+- [sdk/go] enable direct compilation via go build(set `PULUMI_USE_GO_RUN=true` to opt out)
+  [#10375](https://github.com/pulumi/pulumi/pull/10375)
```
AaronFriel, line 20: I strongly believe the polarity of this should be flipped, and we should use a runtime option or env var to opt in to using `go build` and the new behavior.

As I don't currently see the env var used in this PR, requesting changes.

AaronFriel (changes requested):

CHANGELOG_PENDING.md
dixler: took feedback and forgot to update the CHANGELOG_PENDING. it's now `PULUMI_GO_USE_RUN`. nice catch, will fix.

sdk/go/pulumi-language-go/main.go
t0yv0: Folks we're not setting up this PR  and @dixler  for success by moving the goalposts.

At the point of this morning we're trying to agree on having two changes:

1. https://github.com/pulumi/pulumi/issues/10350 will only use `go build -o $tempFile`, no caching, ship-by-default, feature-flag-to-turn-off, unblock some UX changes

2. https://github.com/pulumi/pulumi/issues/10473 to use `go build -o $cachedFile`, ship-under-flag-first, caching to improve the speed, depends on 10350

Are we good with this? If you need to comment on fixing the design let's take the discussion into those tickets and resolve it there. Then we need to give @dixler time to get the PR lined up.

Perhaps @dixler close this PR and let's start afresh? 🤔

AaronFriel (dismissed):

AaronFriel (commented):
Err, commenting to remove my "request changes"

sdk/go/common/util/executable/executable.go
```diff
@@ -27,7 +27,7 @@ func FindExecutable(program string) (string, error) {
 	}
 
 	cwdProgram := filepath.Join(cwd, program)
-	if fileInfo, err := os.Stat(cwdProgram); !os.IsNotExist(err) && !fileInfo.Mode().IsDir() {
+	if fileInfo, err := os.Stat(cwdProgram); err == nil && !fileInfo.Mode().IsDir() {
```
t0yv0, line 30: Why was this change needed?

sdk/go/pulumi-language-go/main.go
```diff
@@ -46,36 +47,47 @@ import (
...
+		return "", errors.Wrap(err, "failed to compile")
 	}
 
-	cwd, err := os.Getwd()
+	filePrefix := fmt.Sprintf("pulumi-go.%s", buildID)
+	if runtime.GOOS == "windows" {
+		filePrefix = fmt.Sprintf("pulumi-go.%s.exe", buildID)
+	}
+
+	f, err := os.CreateTemp("", filePrefix+".*")
```
t0yv0, line 70: This will not do what you want, it will be /tmp/boo.exe.352141780 not /tmp/boo.352141780.exe on Windows

sdk/go/pulumi-language-go/main.go
```diff
@@ -46,36 +47,47 @@ import (
...
+	if matches, err := filepath.Glob(goFileSearchPattern); err != nil || len(matches) == 0 {
+		return "", errors.Errorf("Failed to find go files for 'go build' matching %s", goFileSearchPattern)
 	}
 
-	// Fall back to 'go run' style executions
-	logging.V(5).Infof("No prebuilt executable specified, attempting invocation via 'go run'")
-	program, err := executable.FindExecutable("go")
 	if err != nil {
-		return nil, errors.Wrap(err, "problem executing program (could not run language executor)")
+		return "", errors.Wrap(err, "failed to compile")
```
t0yv0, line 62: I don't think this does indicate "failed to compile" at all, the err seems to be coming from os.Getwd() likely dead code here.

sdk/go/pulumi-language-go/main.go
```diff
@@ -46,36 +47,47 @@ import (
...
 
-	cwd, err := os.Getwd()
+	filePrefix := fmt.Sprintf("pulumi-go.%s", buildID)
+	if runtime.GOOS == "windows" {
+		filePrefix = fmt.Sprintf("pulumi-go.%s.exe", buildID)
+	}
+
+	f, err := os.CreateTemp("", filePrefix+".*")
+	if err != nil {
+		return "", err
```
t0yv0, line 72: This and L75 err seems good to annotate like all the other errs returned here.

sdk/go/pulumi-language-go/main.go
```diff
@@ -46,36 +47,47 @@ import (
...
+	}
+	if err := f.Close(); err != nil {
+		return "", err
+	}
+	outfile := f.Name()
+
+	gobin, err := executable.FindExecutable("go")
 	if err != nil {
-		return nil, errors.Wrap(err, "unable to get current working directory")
+		return "", errors.Wrap(err, "problem executing program (could not run language executor)")
```
t0yv0, line 81: Or perhaps "could not find 'go' executable".

sdk/go/pulumi-language-go/main.go
```diff
@@ -46,36 +47,47 @@ import (
...
 	}
+	buildCmd := exec.Command(gobin, "build", "-o", outfile, cwd)
+	buildCmd.Stdout, buildCmd.Stderr = os.Stdout, os.Stderr
```
t0yv0, line 84: I'm wondering if this will inherit ENV from system stdenv. It seems good not to send Pulumi environment to the compilation phase but it's also good not to shadow so if the users sets a GO_SPECIFIC_ENV_VAR it gets passed to compilation.. That might be happening just fine as is. Interesting.

sdk/go/pulumi-language-go/main.go
```diff
@@ -337,33 +376,50 @@ func (host *goLanguageHost) Run(ctx context.Context, req *pulumirpc.RunRequest)
...
 	}
 
-	cmd, err := findProgram(host.binary)
-	if err != nil {
-		return nil, err
+	if os.Getenv("PULUMI_GO_USE_RUN") != "" {
+		// feature flag to enable old behavior and use `go run`
+		gobin, err := executable.FindExecutable("go")
+		if err != nil {
+			return nil, err
```
t0yv0, line 383: Perhaps same as above here for consistency, ""could not find 'go' executable"."

sdk/go/pulumi-language-go/main.go
```diff
@@ -337,33 +376,50 @@ func (host *goLanguageHost) Run(ctx context.Context, req *pulumirpc.RunRequest)
...
+	// the user can explicitly opt in to using a binary executable by specifying
+	// runtime.options.binary in the Pulumi.yaml
+	program := host.binary
+	if program == "" {
+		// user did not specify a binary and we will compile and run the binary on-demand
+		logging.V(5).Infof("No prebuilt executable specified, attempting invocation via compilation")
+
+		program, err = compileProgramCwd(req.GetProject())
+		if err != nil {
+			return nil, errors.Wrap(err, "error in compiling Go")
```
t0yv0, line 410: Ah never mind,  we wrap the error here 👍

sdk/go/pulumi-language-go/main.go
```diff
@@ -337,33 +376,50 @@ func (host *goLanguageHost) Run(ctx context.Context, req *pulumirpc.RunRequest)
...
 		}
+		defer os.RemoveAll(program)
```
t0yv0, line 412: RemoveAll and not Remove? It can have subdirectories?

sdk/go/pulumi-language-go/main.go
```diff
@@ -337,33 +376,50 @@ func (host *goLanguageHost) Run(ctx context.Context, req *pulumirpc.RunRequest)
 		return nil, errors.Wrap(err, "failed to prepare environment")
 	}
 
-	cmd, err := findProgram(host.binary)
-	if err != nil {
-		return nil, err
+	if os.Getenv("PULUMI_GO_USE_RUN") != "" {
```
t0yv0, line 379: I was thinking if binary: is specified and PULUMI_GO_USE_RUN is specified, it will use run. But perhaps that's more intuitive! Good as-is I guess.

t0yv0:
It's looking great! Had a few nits - mostly concerned about misspelled exe files on Windows - but if Windows tests pass this should be good to go!

t0yv0 (approved):

t0yv0:
I'd love to have this tested on Windows beyond the CI tests.  Anyone has a Windows machine handy? The failure mode that can squeeze through the test is if the code compiles and works but os.Remove removes the wrong file (.exe suffix magic) and ignores the error, so we leak in TEMP.

t0yv0:
For the truly paranoid, ran a targeted Windows test here https://github.com/t0yv0/temp-check-go-windows-behavior/runs/8003274453?check_suite_focus=true#step:5:10 - it works out fine. Seems like on Windows you don't need the file to be named '.exe' for it to be able to execute. So this works out.

dixler:
thought of an edge case where a user provides a binary. the current code path will try to `go run` if the envvar is supplied.

also linux tests were failing since it needs `executable.FindExecutable()` if the program is supplied.

sdk/go/pulumi-language-go/main.go
dixler: this should be a bug. I just understood what you meant.

dixler (commented):
have a lot of feedback from prior commits. may be outdated, but posting.

Issue #10262, opened 2022-07-27
Azure DevOps Pipeline Fails


MitchellGerdisch:
### What happened?

Running Azure Devops Pipeline is suddenly failing  pulumi cli when using a windows runner image and installing the v3.37.0 of Pulumi.
Running with ubuntu does not have a problem.

The following error is thrown and the pipeline fails:
(See pipeline code below.)
```
Starting: Pulumi Setup
==============================================================================
Task         : Pulumi Azure Pipelines Task
Description  : Azure Pipelines task extension for running Pulumi apps.
Version      : 1.0.16
Author       : Pulumi
Help         : Join us on Slack at https://slack.pulumi.com./
==============================================================================
Downloading: https://get.pulumi.com/releases/sdk/pulumi-v3.37.0-windows-x64.zip
Extracting archive
C:\Windows\system32\chcp.com 65001
Active code page: 65001
C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe -NoLogo -Sta -NoProfile -NonInteractive -ExecutionPolicy Unrestricted -Command "$ErrorActionPreference = 'Stop' ; try { Add-Type -AssemblyName System.IO.Compression.FileSystem } catch { } ; [System.IO.Compression.ZipFile]::ExtractToDirectory('D:\a\_temp\4f043eaf-28c9-470d-b6dd-ece43ddbf47d', 'D:\a\_temp\c0b62a5c-431b-4bed-885c-e05f7f525230')"
##[error]Pulumi installation failed with error 'Directory does not exist: D:\a\_temp\c0b62a5c-431b-4bed-885c-e05f7f525230\pulumi\bin'.
```

**Pipeline code:**
```
variables:
  # Override which service connection to use if environment variable is set.
  serviceConnection: REDACTED
  # Override which Pulumi Organization to use if environment variable is set.
  pulumiOrg: REDACTED
  # Returns the root branch name. e.g. for branch "dev/goo" this is "goo"
  pulumiStack: $(Build.SourceBranchName) 

trigger:
- dev/*

pool:
  # vmImage: ubuntu-latest
  vmImage: windows-latest

# # Use Pulumi's published python image. It contains the pulumi command line and a python environment.
# container: pulumi/pulumi-python:latest

steps:
# Just some debugging stuff
- script: |
    pulumi version
    echo "serviceConnection: $(serviceConnection)"
    echo "pulumiOrg: $(pulumiOrg)"
[...]

PR #10383, opened 2022-08-12
Reduce macOS jobs on PRs, replace with partial integration tests

AaronFriel:
Reimplement #10149 to reduce CI workloads again. Includes fixes from PRs #10187 and #10206.

Unblocks PR #10213, as the above PRs are the foundation for including an additional binary reliably in builds.

Fixes the regression #10262 by placing Windows files in a `pulumi/bin` folder in the archive, matching what install scripts expect on Windows.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -174,8 +174,9 @@ jobs:
       - name: Upload pulumi.whl
         uses: actions/upload-artifact@v2
         with:
-            name: pulumi.whl
-            path: sdk/python/lib/dist/*.whl
+          name: pulumi.whl
+          path: sdk/python/lib/dist/*.whl
+          retention-days: 2
 
   build_node_sdk:
     name: Build Pulumi Node SDK tarball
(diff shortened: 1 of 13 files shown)


AaronFriel:
@t0yv0 this is ready to review, it is largely (almost identically) the same as the mentioned PRs that were reverted, except for this change:

```diff
-  wrap_in_directory: pulumi
+  wrap_in_directory: pulumi{{ if eq .Os "windows" }}/bin{{ end }}
  format_overrides:
```

This should solve the regression #10262. Unfortunately [an issue I've raised is that goreleaser doesn't make it feasible to generate archives without a release](https://github.com/goreleaser/goreleaser/discussions/3242). I can run the script locally though:

```shell
GORELEASER_CURRENT_TAG=v3.35.0-alpha.1 \
  goreleaser release -f .goreleaser.build.yml -p 5 --skip-validate --rm-dist --snapshot
#                       ^ substitute for any of the three files
#                         as each has the same archive section now
```
(Due to #9326 you will need to remove `goos: darwin` as an option to run this on Windows or Linux.)

My plan to double-check this is:
1. Merge after today's release, watch the `master` pipeline build prerelease artifacts, and inspect the .zip and .tar.gz files it produced manually. 
2. If the .tar.gz and .zip files do not have the folder structure our install scripts accept and which I observed when running locally, I'll revert again.

AaronFriel:
Rebased to force all checks to run against master being green.

RobbieMcKinstry (approved):
LGTM! Strange that the `if` syntax isn't documented anywhere (that I can find) but good on you for snoofin' it out.

PR #10392, opened 2022-08-15
Move about information to language plugins

Frassle:
# Description

This moves all the language specific "about" information into the language plugins. This is part of the continued chipping away of having the engine have to know anything about specific languages.

## Checklist


Zaid-Ajaj (approved):
The code looks good, I was looking at the integration tests and seeing the only thing being tested is that the command doesn't error (stderr is empty?). Given that those are passing, this is good to go for now we probably need better coverage around those tests

PR #10393, opened 2022-08-15
Thread context through cmd more consistently

Frassle:
Pretty simple change, just fix up a load of places through /cmd where we either used context.Background, or assumed commandContext.

Now all helper methods take an explicit `ctx context.Context` and all cobra commands use `cmd.Context()` to get a context. The correct command context is set once in main.go.


Zaid-Ajaj (approved):
Looks good!

Frassle:
Going for a lunch time swim I think this may break the tracing flag, will look into that.

Frassle:
> Going for a lunch time swim I think this may break the tracing flag, will look into that.

Sigh yes, we can't set the context up front like this because the context depends on the parsing of flags. All a bit of a fun global mutable state party, but I'll leave the other changes in here.

PR #10394, opened 2022-08-15
Enable `pulumi refresh` to solve pending creates

### LINKED ISSUE #4265: Easier way to remove/clear pending operations

iwahbe:
# Description

This PR changes `pulumi refresh` so it can deal with creates. When run interactively, `pulumi refresh` will prompt the user for how to deal with pending creates. 
Interactive options are 
- import
- clear
- skip

A flag has been added (`--skip-pending-creates`) to disable this new behavior. 

To assist with the common case of removing all pending creates, `--clear-pending-creates` has been added.

To allow scripting to resolve pending creates via refresh, a flag (`--import-pending-creates`) has been added. It takes a list of `URN ID`.

Regardless of if these flags are passed, a refresh is still run.

Fixes #4265 

This implements the agreed upon design and replaces https://github.com/pulumi/pulumi/pull/9778.

## Checklist


CHANGELOG_PENDING.md
```diff
@@ -3,9 +3,8 @@
 - [cli] Updated to the latest version of go-git.
   [#10330](https://github.com/pulumi/pulumi/pull/10330)
 
-- [sdk]
-   - merge python error message and traceback into single error message.
-   [#10348](https://github.com/pulumi/pulumi/pull/10348)
```
dixler, line 8: ty

pkg/cmd/pulumi/refresh.go
```diff
@@ -244,3 +342,57 @@ func newRefreshCmd() *cobra.Command {
 
 	return cmd
 }
+
+type editPendingOp = func(op resource.Operation) (*resource.Operation, error)
+
+func editPendingCreates(s backend.Stack, opts display.Options, yes bool, f editPendingOp) result.Result {
```
dixler, line 348: editPendingCreates->filter(/+?)mapPendingCreates

dixler (commented):
reviewed with @iwahbe in a call. looking forward to this.

dixler:
I think we might want to update the error message to direct users to use `pulumi refresh` with this PR
https://github.com/pulumi/pulumi/blob/7827eb26bc1a64aa56ad450d35bccde0ceadd78e/pkg/resource/deploy/deployment_executor.go#L116-L120

also I got confused in the flow when trying to work with my stack.
https://asciinema.org/a/YsO3wkjshq8M7IkiTQytjx04w

I thought I cleared the resource so when I try to run `pulumi up` it doesn't redeploy? I must be missing something?

iwahbe:
@dixler We should definitely update the error message (https://github.com/pulumi/pulumi/pull/10394/commits/cbdc3d8948a393b880d5f1a4faa7e5eb43b940c6). I just tried the same workflow you did, and it worked as expected for me. My guess is that when you corrupted your stack, it already met the desired configuration. If you could show me that process, I could provider a better answer.

dixler:
@iwahbe can we add more information to the command options. the commands aren't very clear about what they're going to do which is why my screen recording has me Ctrl+C'ing out of one of them because I wasn't expecting a prompt from `import`

so instead of 
```
> import
  clear
  skip
```
it might look like
```
> import(short description)
  clear(short description)
  skip(short description)
```

pkg/cmd/pulumi/refresh.go
```diff
@@ -140,6 +150,55 @@ func newRefreshCmd() *cobra.Command {
 				return result.FromError(fmt.Errorf("getting stack configuration: %w", err))
 			}
 
+			if skipPendingCreates && clearPendingCreates {
```
Frassle, line 153: Same check for skipPending and importPending?

iwahbe:
@dixler Bumping the `survey` version to `v2.3.5` gives a description option:
Screen Shot 2022-08-17 at 12 06 03 PM

pkg/cmd/pulumi/refresh.go
iwahbe: `--skip-pending-creates` implies don't show me existing pending creates.
`--clear-pending-creates` implies remove pending creates.
It doesn't make sense to provide a command that ensures there will be no pending creates *and* that you shouldn't show pending creates (since there will never be any).

`--import-pending-creates` turns some set of pending creates into imports (finishing the create). 
It makes sense to run `--import-pending-creates` & `--skip-pending-creates`, telling the CLI to import a set of resources and not prompt for the rest.

It even makes sense to run `--import-pending-creates` & `--clear-pending-creates`, telling the CLI to import a set of resources and remove the rest.

> Same check for skipPending and importPending?

I think it is valid to combine `--skip-pending-creates` with `--import-pending-creates`.

dixler:
> @dixler Bumping the `survey` version to `v2.3.5` gives a description option:
> 
> Screen Shot 2022-08-17 at 12 06 03 PM
> 
> 
> 
> 

I remember my first reverted PR was due to some serious issues when I tried bumping survey.

Also, in my recording, it looked like the pending create was cleared, but it had succeeded so that description may not always be accurate.

iwahbe:
> I remember my first reverted PR was due to some serious issues when I tried bumping survey.

I didn't bump survey globally, just for `pulumi refresh`. Do you remember what problem the bump introduced for you?

> Also, in my recording, it looked like the pending create was cleared, but it had succeeded so that description may not always be accurate.

In your recording, it looks like you had a pending create as well as a finished create. Unless you hacked your state, they correspond to 2 different resources. You cleared the pending create, but your state was already happy, so an additional create was not necessary. If you finished the import, pulumi would then delete the imported resource on the next `pulumi up`.

This is what it looks like when you run `pulumi up` after you killed pulumi while creating a bucket. Notice that it both has a pending create *and* creates a new bucket.
Screen Shot 2022-08-18 at 11 27 07 AM
The associated pulumi program is:
```yaml
name: yamyam
runtime: yaml
resources:
  bucket:
    type: aws:s3:BucketV2
```

dixler:
> > I remember my first reverted PR was due to some serious issues when I tried bumping survey.
> 
> I didn't bump survey globally, just for `pulumi refresh`. Do you remember what problem the bump introduced for you?
> 
yeah. it was a while back, but we had a fun thread about it https://pulumi.slack.com/archives/CBUP505U7/p1643713912052759

> > Also, in my recording, it looked like the pending create was cleared, but it had succeeded so that description may not always be accurate.
> 
> The description is accurate in describing the intended use of the command. I'm still not sure what happened in your recording, but I can't replicate the behavior. Can you tell me what you did to corrupt your state.

I believe I ran. the implementation of this shouldn't be an issue since that subcommand creates a yaml project and ups it.

```
$ pulumi install gcp:container:Cluster my-cluster initialNodeCount=2 minMasterVersion=1.22.10-gke.600 nodeConfig={[[,machineType=n1-standard-1,oauthScopes={[,"https://www.googleapis.com/auth/compute","https://www.googleapis.com/auth/devstorage.read_only","https://www.googleapis.com/auth/logging.write","https://www.googleapis.com/auth/monitoring",]},]]} project=pulumi-development location=us-west2
```
and I `pkill`'d pulumi

GKE clusters take a while to spin up, but I had a lot of stuff that got in my way that day, so the cluster likely finished provisioning by the time that I got back to checking this.

yaml that was generated:
```yaml
name: global-stack_bXktY2x1c3Rlcg_Z2NwOmNvbnRhaW5lcjpDbHVzdGVy
runtime: yaml
resources:
    my-cluster:
        type: gcp:container:Cluster
        properties:
            initialNodeCount: 2
            location: us-west2
            minMasterVersion: 1.22.10-gke.600
            nodeConfig:
                machineType: n1-standard-1
                oauthScopes:
                    - https://www.googleapis.com/auth/compute
                    - https://www.googleapis.com/auth/devstorage.read_only
                    - https://www.googleapis.com/auth/logging.write
                    - https://www.googleapis.com/auth/monitoring
            project: pulumi-development
```

iwahbe:
@dixler I'm still unable to repro your problem. I tried spinning up a GKE cluster and then killing pulumi several times. Each time I ran the experiment I got one of the expected (and correct) responses:
- No resource, no pending -> attempt to create new
- No resource, pending -> attempt to create new
  - Clear -> attempt to create new
  - Import -> Don't attempt to create new
- Resource, no pending -> don't attempt to create new

I dropped survey back to v1.

dixler (approved):
the error state I ran into had a pending create while it was deploying the cluster. it shouldn't hold back this PR.

Frassle (approved):

mikhailshilkov:
🎉 Let's file an issue to write a blog post about this change in M78 please.

cromega:
I'm sorry to poke a dead thread but what's the solution using the Automation API?

iwahbe:
> I'm sorry to poke a dead thread but what's the solution using the Automation API?

As far as I'm aware, this hasn't been added to the Automation API yet.

PR #10269, opened 2022-07-27
[cli] Allow Pulumi.yaml files to contain additional, preserved keys

AaronFriel:

PR #10397, opened 2022-08-15
Revert "[cli] Allow Pulumi.yaml files to contain additional, preserved keys"

AaronFriel:
Without some changes to `pulumi new` behavior, this has the surprising and incorrect behavior of causing the `resources` block to be duplicated in templates.

Reverts pulumi/pulumi#10269


AaronFriel:
@RobbieMcKinstry it's a bit sillier than that. The way `pulumi new` works is that it opens the template project file, modifies it with the user's changes (name of project, etc.), then marshals it and saves it, then concatenates the `.append` file without unmarshaling.

Well, the previous PR fixed a bug where when we marshal/unmarshal a `Pulumi.yaml` file, we delete any unknown keys, e.g. `resources`, `variables` because they aren't defined in the struct. But we still concatenated the `.append` file, hence this revert.

PR #10399, opened 2022-08-15
Skip codegen tests that import az-nature. 

RobbieMcKinstry:
# Description

Some of our Codegen tests import azure-native, which is a known massive project. AZ Native has finally exceeded the max size permitted by `go mod` for import and has started failing builds as a result. **This issue is blocking release**.

This PR skips the failing tests that use AZ-Native. I tried to leave as much enabled as possible -- I didn't disable tests in other languages that use AZ Native. Unfortunately, I wasn't able to use `SkipCompile` instead of `Skip`, because `go mod tidy` is producing the error, and that happens at generation-time, which is before compile-time.

Also, I didn't disable all tests that use AZ Native -- only the ones that were failing on master.

Marking this PR as "draft" for now because a better solution would be to pin an earlier version of AZ Native, which I might be able to do.

### DIFF
--- a/pkg/codegen/testing/test/program_driver.go
+++ b/pkg/codegen/testing/test/program_driver.go
@@ -162,6 +162,7 @@ var PulumiPulumiProgramTests = []ProgramTest{
 	{
 		Directory:   "typed-enum",
 		Description: "Supply strongly typed enums",
+		Skip:        codegen.NewStringSet(golang),
 	},
 	{
 		Directory:   "python-resource-names",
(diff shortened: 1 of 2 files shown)


RobbieMcKinstry:
I have no clue how this change could have broken downstream codegen tests. That makes no sense to me.

Frassle:
> I have no clue how this change could have broken downstream codegen tests. That makes no sense to me.

They've been broken for ages because of an api change that hasn't been applied to tfbridge yet.

Frassle (approved):

PR #10400, opened 2022-08-15
chore(sdks): Deprecate PULUMI_TEST_MODE

### LINKED ISSUE #3045: Deprecate PULUMI_TEST_MODE

kpitzen:
# Description

Fixes #3045 

This PR removes `PULUMI_TEST_MODE` from both nodejs and python SDKs.  I couldn't find any references to this functionality in other SDKs, so I restricted my work to these.

## Checklist

Technically, I only _removed_ tests.


Frassle (approved):
Looks reasonable.

PR #10401, opened 2022-08-16
[cli] Enable single-file Pulumi YAML templates

AaronFriel:
Expands on #10269 with utility functions to preserve existing YAML document structure, including comments, while mutating key fields.

And coincidentally simplifying our templates, changes to how we install dependencies has removed the need to check for the "python" runtime and set a venv field.

With these changes, with or without a `Pulumi.yaml.append` file, `pulumi new aws-yaml` creates a file like so:

```yaml
name: tmp-941fe618-e0bf-4149-ab8b-54ca557202ed
description: A minimal AWS Pulumi YAML program
runtime: yaml
resources:
    # Create an AWS resource (S3 Bucket)
    my-bucket:
        type: aws:s3:Bucket
outputs:
    bucketName: ${my-bucket.id}
```

The new implementation only removes the `.append` file, it does not concatenate it with `Pulumi.yaml`.

Fixes pulumi/pulumi-yaml#303

### DIFF
--- a/pkg/cmd/pulumi/new.go
+++ b/pkg/cmd/pulumi/new.go
@@ -279,6 +279,7 @@ func runNew(ctx context.Context, args newArgs) error {
 			return err
 		}
 
+		contract.Assert(len(workspaceDocument.Content) == 1)
 		projFile, err := yaml.Marshal(workspaceDocument.Content[0])
 		if err != nil {
 			return err
--- a/pkg/cmd/pulumi/new.go
+++ b/pkg/cmd/pulumi/new.go
@@ -27,6 +27,8 @@ import (
 	"strings"
 	"unicode"
 
+	"gopkg.in/yaml.v3"
+
 	"github.com/spf13/cobra"
 	"golang.org/x/crypto/ssh/terminal"
 	survey "gopkg.in/AlecAivazis/survey.v1"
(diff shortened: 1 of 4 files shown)


iwahbe:
It looks like tests are failing, but the changes to `pkg/cmd/pulumi/new.go` look good.

Frassle:
> And coincidentally simplifying our templates, changes to how we install dependencies has removed the need to check for the "python" runtime and set a venv field.

Huh I think that's actually a bug. If venv isn't set I don't think we should be assuming "venv" in general, it just has been happening to work because we only called InstallDependencies from "new" where we have been assuming that. 
GetRequiredPlugins at least only tries to make and install a venv if the property is set.

pkg/cmd/pulumi/new.go
```diff
@@ -246,49 +249,54 @@ func runNew(ctx context.Context, args newArgs) error {
...
-		} else if !os.IsNotExist(err) {
-			return fmt.Errorf("could not get %s: %w", appendFileName, err)
+
+		err = os.WriteFile(path, projFile, 0600)
+		if err != nil {
+			return err
 		}
 	}
 
+	appendFileName := "Pulumi.yaml.append"
```
Frassle, line 293: This is just deleting the .append file? Doesn't the above need to read it and merge it in for correct yaml templates?

pkg/cmd/pulumi/new.go
```diff
@@ -246,49 +249,54 @@ func runNew(ctx context.Context, args newArgs) error {
...
+		}
+		err = yamlutil.Delete(&workspaceDocument, "template")
+		if err != nil {
+			return err
+		}
 
-	if proj.Runtime.Name() == "yaml" {
-		projFile := filepath.Join(root, "Pulumi.yaml")
-		f, err := ioutil.ReadFile(projFile)
+		projFile, err := yaml.Marshal(workspaceDocument.Content[0])
```
Frassle, line 282: contract.Assert(len(workspaceDocument.Content) == 1)

pkg/cmd/pulumi/new.go
AaronFriel: We generated the .Append files by splitting the project file and taking the latter half, I believe @iwahbe had a script that did this.

See: 

https://github.com/pulumi/templates/blob/master/aws-yaml/Pulumi.yaml

We don't want to append the YAML file, we want to gracefully migrate away from needing the `.append` file in templates.

Frassle: Ah ok so the .yaml and .append.yaml both had all the resources in

AaronFriel:
@Frassle I've updated this to reinstate the `venv` workaround.

pkg/cmd/pulumi/new.go
```diff
@@ -246,47 +250,70 @@ func runNew(ctx context.Context, args newArgs) error {
...
+			return err
+		}
+		proj.Template = nil
+		err = yamlutil.Delete(&workspaceDocument, "template")
+		if err != nil {
+			return err
+		}
+		if proj.Runtime.Name() == "python" {
+			// If the template does give virtualenv use it, else default to "venv"
+			if len(proj.Runtime.Options()) == 0 {
```
Frassle, line 287: dicey, virtualenv is the only option allowed on python at the moment... I think we should remove this hack well before we add a new option though

Frassle (approved):

iwahbe (approved):
Tests are 🟢 and the feature looks good. LGTM

Issue #8131, opened 2021-10-04
Drop python 3.6 support


komalali:
Python 3.6 will be EOL'd in [December 2021](https://www.python.org/downloads/#:~:text=2016-12-23-,2021-12-23,-PEP%20494). We currently [some code](https://github.com/pulumi/pulumi/blob/0b1414dc6d59908a600116470bc30d2450828296/sdk/python/lib/pulumi/runtime/sync_await.py#L20) that behaves differently when using python 3.6 vs other versions due to changes in the python asyncio API between these two versions. Dropping support for python 3.6 would help in simplifying this code. When we do this, we should make sure to update the [language docs](https://www.pulumi.com/docs/intro/languages/python/) to reflect the new reality.

**Update: 11/2** This will be deprecated in December 2021

komalali:
Looks like the deprecation message being printed isn't correct. It should say Python `3.6` and not `3.7`
```
Python 3.7 is approaching EOL and will not be supported in Pulumi soon. Check https://github.com/pulumi/pulumi/issues/8131 for more details
```

PR #2638, opened 2019-04-11
Enable unit testing for Pulumi programs

joeduffy:
This change enables rudimentary unit testing of your Pulumi programs, by introducing a `PULUMI_TEST_MODE` envvar that, when set, allows programs to run without a CLI. That includes

* Just being able to import your Pulumi modules, and test ordinary functions -- which otherwise would have often accidentally triggered the "Not Running in a CLI" error message
* Being able to verify a subset of resource properties and shapes, with the caveat that outputs are not included, due to the fact that this is a perpetual "dry run" without any engine operations occurring

In principle, this also means you can attach a debugger and step through your code.

PR #10402, opened 2022-08-16
fix(sdk/python): Makes SETTINGS attributes context-aware

### LINKED ISSUE #6052: [Automation API] Support parallel execution of python inline programs
The Pulumi python runtime implementation uses quite a bit of global state: 
* [SETTINGS](https://github.com/pulumi/pulumi/blob/master/sdk/python/lib/pulumi/runtime/settings.py#L94)
* [ROOT](https://github.com/pulumi/pulumi/blob/master/sdk/python/lib/pulumi/runtime/settings.py#L189)
* [CONFIG](https://github.com/pulumi/pulumi/blob/master/sdk/python/lib/pulumi/runtime/config.py#L24)

We'll need to refactor all usage of global state out of the python SDK to allow Automation API to execute concurrent programs on different stacks that use the inline/hosted execution model.

Related: https://github.com/pulumi/pulumi/issues/5449

kpitzen:
# Description

Fixes #6052 

I'm throwing this up a bit early to expose my current thoughts on the issue.  I don't love how boilerplate-y this approach is, and I'd appreciate a couple 👀 to give me a sanity check around ways to make this make a little more sense.

## Checklist


sdk/python/lib/pulumi/runtime/settings.py
```diff
@@ -98,6 +95,73 @@ def __init__(
...
+    def legacy_apply_enabled(self, v: Optional[bool]):
+        return self._legacy_apply_enabled.set(v).var.get()
+
+    @property
+    def feature_support(self):
+        return self._feature_support.get()
+
+    @feature_support.setter
+    def feature_support(self, v: Optional[dict]):
+        return self._feature_support.set(v).var.get()
```
kpitzen, line 163: This will need to be a bit smarter.  I think this will need to be an object which overrides `__getitem__` to adapt the dict interface to a ContextVar'd dict

sdk/python/lib/pulumi/runtime/settings.py
```diff
@@ -98,6 +95,73 @@ def __init__(
         else:
             self.engine = None
 
+    def __repr__(self):
+        return f"Settings: {self.engine}:{self.monitor}:{self.project}:{self.stack}"
```
kpitzen, line 99: I'll likely remove this, but if we have a good idea of what we'd _actually_ want a `__repr__` to look like, I'd love to hear it.  Inspecting some of these objects is currently not very useful.

sdk/python/lib/pulumi/runtime/settings.py
```diff
@@ -98,6 +95,73 @@ def __init__(
...
+
+    @property
+    def engine(self):
+        return self._engine.get()
+
+    @engine.setter
+    def engine(self, v: Optional[Union[engine_pb2_grpc.EngineStub, Any]]):
+        return self._engine.set(v).var.get()
```
kpitzen, line 107: `.set()` returns a `Token` which references the original `ContextVar` via `.var`, so we have to then `.get()` the new value :)

kpitzen:
Thinking about this a bit more, maybe we can provide a neat little decorator to just `ContextVar` a property and handle all this unfortunate boilerplate for us.

sdk/python/lib/pulumi/runtime/settings.py
```diff
@@ -92,6 +88,33 @@ def __init__(
...
+    def dry_run(self) -> Optional[bool]:
+        ...
+
+    @contextproperty("legacy_apply_enabled")
+    def legacy_apply_enabled(self) -> Optional[bool]:
+        ...
+
+    @contextproperty("feature_support")
+    def feature_support(self) -> Optional[dict]:
+        ...
```
kpitzen, line 113: This interface feels a lot more ergonomic than the initial pass.  There's still a bit of weirdness but it feels much more on the right track

sdk/python/lib/pulumi/_utils.py
```diff
@@ -119,3 +120,59 @@ def lazy_import(fullname):
...
+        prop = type(self)(fget, self.fset, self.fdel, name=self._name, doc=self.__doc__)
+        return prop
+
+    def setter(self, fset):
+        prop = type(self)(self.fget, fset, self.fdel, name=self._name, doc=self.__doc__)
+        return prop
+
+    def deleter(self, fdel):
+        prop = type(self)(self.fget, self.fset, fdel, name=self._name, doc=self.__doc__)
+        return prop
```
kpitzen, line 178: This is essentially a reimplementation of the builtin `property` but wrapping values in `ContextVar`

sdk/python/.pylintrc
```diff
@@ -68,6 +68,7 @@ disable=raw-checker-failed,
         useless-suppression,
         deprecated-pragma,
         use-symbolic-message-instead,
+        method-hidden,
```
kpitzen, line 71: Bringing in annotations ends up throwing linting errors about hiding methods (which is a lie!  Pylint just isn't smart enough to realize the attributes set in the constructor _ARE_ the methods being "hidden")

sdk/python/lib/test/runtime/test_localized_global_state.py
```diff
@@ -0,0 +1,16 @@
...
+
+    async def set_and_retrieve_settings_value(stack_name: str):
+        default_settings_instance.stack = stack_name
+        assert default_settings_instance.stack == stack_name
+        return default_settings_instance.stack
+
+    foo_name = await set_and_retrieve_settings_value("foo")
+    bar_name = await set_and_retrieve_settings_value("bar")
+    assert foo_name != bar_name != "project"
+    assert default_settings_instance.project == "project"
```
kpitzen, line 16: If I have concerns about this PR, this test likely sums them up.  We will be isolating value changes to local contexts which will enable parallel pulumi programs to run without bleeding state.  However, any async tasks which depend on this formerly-global state will _also_ be isolated.  Tests pass with the current configuration but I think it's worth having a conversation here.

kpitzen: I have verified that, at runtime for local stacks, settings values are preserved and work just fine, so it _could_ easily be weirdness with how our tests are implemented that cause us to need an escape hatch from the context-var-y-ness implemented in this PR.

sdk/python/lib/pulumi/_utils.py
```diff
@@ -119,3 +120,87 @@ def lazy_import(fullname):
     sys.modules[fullname] = module
     loader.exec_module(module)
     return module
+
+
+def contextproperty(_fn=None, *, default: typing.Optional[typing.Any] = None):
```
Frassle, line 125: I wouldn't normally use underscores in param names. Is there a reason to do so?

sdk/python/lib/pulumi/_utils.py
```diff
@@ -119,3 +120,87 @@ def lazy_import(fullname):
...
+        )
+        inner.__doc__ = new_property.__doc__
+        return new_property
+
+    if _fn is None:
+        return inner
+    return inner(_fn)
+
+
+class ContextProperty:
```
Frassle, line 138: doc comments on what this is for would be good, it's not going to be immediatly clear why we're not just using contextvars  directly

kpitzen: Sure!  I'll definitely document this.

sdk/python/lib/pulumi/_utils.py
kpitzen: Not _particularly_ - it's generally used to indicate params not to access directly, but I don't mind removing it here.

sdk/python/lib/pulumi/_utils.py
kpitzen: For completeness, the added doctests _do_ pass should we ever want to start supporting them:
image

Frassle (approved):
Looks good, only thing I'd like to see is an automation api test showing two updates in parallel. 
There's a few automation tests already at sdk/python/lib/test/automation.

kpitzen:
@Frassle Awesome!  I'll add those tests in the followup PR.

sdk/python/lib/pulumi/runtime/settings.py
```diff
@@ -92,9 +86,44 @@ def __init__(
...
+    def dry_run(self) -> Optional[bool]:
+        ...
+
+    @contextproperty
+    def legacy_apply_enabled(self) -> Optional[bool]:
+        ...
+
+    @contextproperty
+    def feature_support(self) -> Optional[dict]:
+        ...
```
justinvp, line 119: Why do these individual attributes need to use contextvars? Instead, why not store the top-level `SETTINGS` var as a `ContextVar`, similar to what you're doing in #10472 for `CONFIG`, `ROOT`, and `_SECRET_KEYS`?

sdk/python/lib/pulumi/runtime/settings.py
```diff
@@ -125,12 +154,7 @@ def get_project() -> str:
...
-    project = SETTINGS.project
-    if not project:
-        raise RunError(
-            "Missing project name; for test mode, please call `pulumi.runtime.set_mocks`"
-        )
-    return project
+    return SETTINGS.project
 
 
 def _set_project(v: Optional[str]):
```
justinvp, line 160: It looks like `_set_project` and `_set_stack` are the only functions that modify the attributes of the global `SETTINGS` variable.

These were introduced in #2638 to support `PULUMI_TEST_MODE`, which was deprecated in #10400. Since this functionality was deprecated, I wonder if we can simply remove `_set_project` and `_set_stack`. It doesn't look like anything else in the repo is calling them.

Removing them may allow for a simpler implementation here, if we only make the top-level `SETTINGS` a `ContextVar`.

sdk/python/lib/pulumi/runtime/settings.py
kpitzen: The reason I didn't want to do that was throughout our codebase, the way and frequency in which we access SETTINGS is different than the others.  `CONFIG`, `ROOT`, and `_SECRET_KEYS` are simple objects which are accessed relatively infrequently.  In the case of settings, however, we don't use the getter functions universally, and access its attributes directly, so modifying the way in which our code interacts with SETTINGS throughout felt like a higher risk change than to maintain its interface and hide the complexity within the object itself.

justinvp:
Looks like contextvars is new in Python 3.7, and since Python 3.6 is EOL, this seems fine for us to make use of now. As a follow-up we probably should revisit #8131 and probably add `python_requires='>=3.7',` to `setup.py`.

Frassle:
> As a follow-up we probably should revisit https://github.com/pulumi/pulumi/issues/8131 and probably add python_requires='>=3.7', to setup.py.

Yeh I've started the process to get our supported versioned documented better, https://github.com/pulumi/pulumi-hugo/pull/1860. Updating reqs like those in setup.py will come after that.

sdk/python/lib/pulumi/runtime/settings.py
justinvp: OK, fair enough. 😄

justinvp (approved):

PR #10410, opened 2022-08-16
[sdk/nodejs] emit colorized stack traces

### LINKED ISSUE #10379: [sdk/nodejs] improve nodejs stacktrace error message
## Hello!

## Issue details
currently we output large stack traces which makes it hard for users to determine where their errors are coming from. is there a way that we could reduce these without confusing users.

normally running a program via node/ts-node looks better as they both use color to draw attention to relevant details.

pulumi:
[asciicast](https://asciinema.org/a/aNwrthFasG38qoXdTFcrAar2l)

nodejs:
[asciicast](https://asciinema.org/a/qAfVccIYk8ob5pCecOe515TMn)

ts-node:
[asciicast](https://asciinema.org/a/D2esRjjXByUTPvt9RxuxxY0vc)

#### Solution

Pass through error message coloring to the error message output.

dixler:
# Description
This PR changes pulumi nodejs error logging to emit colorized stack traces to match nodejs's default behavior.

Fixes #10379 

New:
[asciicast](https://asciinema.org/a/wh2tFUZXqm8cXqVo5WcHEexii)

Current:
[asciicast](https://asciinema.org/a/IvTzhNHfsAj82FT115bXT4XSN)

## Checklist

### DIFF
--- a/pkg/backend/display/events.go
+++ b/pkg/backend/display/events.go
@@ -3,6 +3,7 @@ package display
 import (
 	"errors"
 	"fmt"
+	"regexp"
 	"time"
 
 	"github.com/pulumi/pulumi/pkg/v3/engine"
@@ -17,6 +18,8 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/common/tokens"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 )
+
+var matchAnsiControlCodes = regexp.MustCompile(`\x1b\[[0-9;]*[mK]`)
 
 // ConvertEngineEvent converts a raw engine.Event into an apitype.EngineEvent used in the Pulumi
 // REST API. Returns an error if the engine event is unknown or not in an expected format.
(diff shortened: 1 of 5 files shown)


mikhailshilkov:
Could you please add an example of a stacktrace with a few layers in a user program? (e.g. when an error happens in a callback in a function in a class )

dixler:
@mikhailshilkov they look identical

New:
https://asciinema.org/a/qph5eRlYkVAEQcnvTZtwwV4DI

Current:
https://asciinema.org/a/928KfZrEO5P1ezmYlKOqZKsjE

mikhailshilkov:
Oh! Where is the node garbage in this case? Does it only come up if the error is at the top level?

dixler:
> Oh! Where is the node garbage in this case? Does it only come up if the error is at the top level?

it seems that node truncates stack traces and the node garbage was truncated.

Frassle:
Does this behave well with piped output? I.e. if stdout is not a pty

dixler:
> Does this behave well with piped output? I.e. if stdout is not a pty

I doubt it. I saw that pulumi python calls pip and it has color display so I was curious if we're concerned about this since I'm not sure if we optimize pulumi pipe experience. I know non-interactive still outputs colors, but I'd have to check what it does if it's outputting to a file.

dixler:
Added code to strip out ANSI control codes before forwarding the DiagnosticEvent to the service.

pkg/backend/display/events.go
```diff
@@ -18,6 +19,8 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
 )
 
+var matchAnsiControlCodes = regexp.MustCompile(`\x1b\[[0-9;]*[mK]`)
```
abhinav, line 22: 👍 
This looks like it'll cover at least all color cases.
We can expand this if we need to cover more escape codes in the future.

abhinav (approved):

dixler:
bors merge

PR #10413, opened 2022-08-16
[cli] make `pulumi stack rm` more usable

### LINKED ISSUE #9663: pulumi stack rm prompt is confusing
when deleting a stack we ask the person to type the stack name, for example if the stack name is dev we ask them to type "dev", but if you type "dev" with the quotes it does not work.

suggest dropping both the quotes and the parens from that prompt

```
This will permanently remove the 'dev' stack!
Please confirm that this is what you'd like to do by typing ("dev"): 
```

this came up in a usability study today

dixler:
# Description
This PR clarifies highlighting of confirmation text in `confirmPrompt`.

Fixes #9663

New:
[asciicast](https://asciinema.org/a/m8E9OK7TVtSWRvIJAWtHEivlv)

Old:
[asciicast](https://asciinema.org/a/G0ScAU2SYLnx1WYGwnBMT6dpK)

## Checklist


pkg/cmd/pulumi/pulumi.go
```diff
@@ -649,5 +649,6 @@ func confirmPrompt(prompt string, name string, opts display.Options) bool {
 
 	reader := bufio.NewReader(os.Stdin)
 	line, _ := reader.ReadString('\n')
-	return strings.TrimSpace(line) == name
+	input := strings.TrimSpace(line)
+	return strings.Trim(input, "\"") == name
```
Frassle, line 653: I worry about putting this so deep in input parsing, there _may_ be cases where '"' is a valid char to input and this will confuse the hell out of us if we forget about it.
Also are they're really enough people adding quotes to their prompt answers that this is needed?

Frassle: OK I re-read the issue, this is _very_ specific to "pulumi stack rm" so I'd be ok with the trim('"') there.

dixler: I'll recheck usage and name this explicitly for stackRm if it's the only user.

Frassle (approved):

PR #10415, opened 2022-08-16
fix: handle .net and other domains in VCS detection

blampe:
# Description

VCS detection logic was hard-coded to only look for .com and .org domains; loosen it to accept any domain. Please double check my regex!

Fixes https://github.com/pulumi/pulumi-service/issues/8883

## Checklist


susanev:
wow, thx for the extremely quick fix!

CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 CHANGELOG
 =========
 
+## Unreleased
```
Frassle, line 4: Don't edit CHANGELOG.md in PRs, our workflow is the edit CHANGELOG_PENDING.md instead.

blampe: My bad, fixed!

Frassle (approved):

PR #10419, opened 2022-08-16
Fix StackReference code gen

### LINKED ISSUE #10260: Inaccurate StackReference code generation via `pulumi convert`
Code generation assumes that the `pulumi` package should be interpreted like others, such as `aws`, and imported as `@pulumi/pulumi-pulumi` (nodejs) or via `pulumi_pulumi` (python), and so on.

The schema is also inaccurate for Python at least, as the `name` property in the Python SDK is `stack_name`.

This issue will track successfully generating code for each of our target languages via PCL.

aq17:
# Description

Fixes #10260 

## Checklist


aq17:
[genPreamble](https://github.com/pulumi/pulumi/blob/master/pkg/codegen/go/gen_program.go#L288) takes care of importing the top-level pulumi packages, so we should ignore any further imports when converting type `pulumi:pulumi:StackReference`

pkg/codegen/go/gen_program.go
```diff
@@ -354,9 +354,14 @@ func (g *generator) collectImports(
...
 		if r, isResource := n.(*pcl.Resource); isResource {
 			pkg, mod, name, _ := r.DecomposeToken()
-			if pkg == "pulumi" && mod == "providers" {
-				pkg = name
-				mod = ""
+			if pkg == "pulumi" {
+				if mod == "providers" {
+					pkg = name
+					mod = ""
+				} else if mod == "pulumi" {
```
iwahbe, line 361: Where would `pulumi:pulumi:*` be a valid token?

I'll definitely want to see a test for this PR.

Frassle: Stack references are "pulumi:pulumi:StackReference"

pkg/codegen/schema/mock_pulumi_schema.go
```diff
@@ -22,11 +22,20 @@ func newPulumiPackageReference() *Package {
 				},
 			},
 		},
+		Provider: ResourceSpec{
+			InputProperties: map[string]PropertySpec{
```
iwahbe, line 26: We should probably add a description here, since this stuff is exposed via the LSP server.

pkg/codegen/pcl/binder_resource.go
```diff
@@ -59,19 +59,29 @@ func (b *binder) bindResourceTypes(node *Resource) hcl.Diagnostics {
...
-		return hcl.Diagnostics{unknownPackage(pkg, tokenRange)}
+	var pkgSchema *packageSchema
+	if pkg != "pulumi" && module != "pulumi" {
+		var ok bool
+		pkgSchema, ok = b.options.packageCache.entries[pkg]
+		if !ok {
+			return hcl.Diagnostics{unknownPackage(pkg, tokenRange)}
+		}
+	} else {
+		isProvider = true
```
iwahbe, line 70: Is is possible to assign `pkgSchema` a value here so we can prevent the conditional on line 76?

pkg/codegen/pcl/binder_resource.go
```diff
@@ -59,19 +59,29 @@ func (b *binder) bindResourceTypes(node *Resource) hcl.Diagnostics {
 		pkg, isProvider = name, true
 	}
 
-	pkgSchema, ok := b.options.packageCache.entries[pkg]
-	if !ok {
-		return hcl.Diagnostics{unknownPackage(pkg, tokenRange)}
+	var pkgSchema *packageSchema
+	if pkg != "pulumi" && module != "pulumi" {
```
iwahbe, line 63: Could we switch the order of this conditional so it matches the conditional on line 76.

iwahbe (commented):
I'm don't see the test for C#. I'm also missing the change to `program_driver.go` which would actually run this test.

pkg/codegen/nodejs/gen_program.go
```diff
@@ -362,6 +362,9 @@ func outputRequiresAsyncMain(ov *pcl.OutputVariable) bool {
 func resourceTypeName(r *pcl.Resource) (string, string, string, hcl.Diagnostics) {
 	// Compute the resource type from the Pulumi type token.
 	pkg, module, member, diagnostics := r.DecomposeToken()
+	if pkg == "pulumi" && module == "pulumi" {
+		module = ""
+	}
```
AaronFriel, line 367: Should we refactor this out to a top level method alongside `MapProvidersAsResources`? We need to do this for each language, may as well export it as a method. If we add any future special cases & built-in methods, that'll be the place to do it.

AaronFriel (commented):
Left one nit, what do you think about centralizing that token remapping for `pulumi:pulumi:* => pulumi::*` in `codegen/pcl/utilities.go`?

pkg/codegen/nodejs/gen_program.go
aq17: For collecting imports, I think it may still be helpful to still have `mod` information to distinguish between needing to import `pulumi:Providers:<name>` versus ignoring `pulumi:pulumi:*`, like for [here](https://github.com/pulumi/pulumi/pull/10419/files#diff-d6817c1c54a334e663ae7a8ef2bcc697641fcddf412b0d2b97c884a9e297a5f3R364)

AaronFriel: I think you've found some dead code since I introduced `MapProvidersAsResources`. We'll have already mapped `pulumi:providers:$name` to `$name::Provider`, so I don't think we'll ever end up in this branch:

```
if pkg == "pulumi" && mod == "providers" {
```

Let me put my thinking another way: I'd like any sort of "special case" logic about what the `pulumi` package does that _isn't_ like other providers to live in `codegen/pcl/utilities.go` because eventually we will want language plugins & codegen to be split out into separate repos, and we won't want to have to update N repos with new special cases. (e.g.: Pulumi-Java)

If we move the mapping here into a `func FixupPulumiPackageTokens` that's shared, then we can update that in one place and as our language plugins update, we don't need to find and update the same thing in each of them.

aq17: Moved the logic to reorg pulumi package tokens to `utilities.go`; however, we still currently special case ignoring importing the `pulumi` package since `genpreamble` does that already, and it attempts to do it incorrectly i.e. `pulumi_pulumi` (the original issue from the ticket)

iwahbe (approved):
This looks good to me. You need to fix a test, but once the PR goes green, LGTM.

pkg/codegen/nodejs/gen_program.go
AaronFriel: That makes sense.

PR #10422, opened 2022-08-17
Improve lock error for filestate

Frassle:
# Description

No need to manually delete lock files anymore, `pulumi cancel` will do this.


RobbieMcKinstry (approved):
Nit: Unlikely the backticks will render in markdown. Don't think we have a CLI style guide, but I wonder if quotes is more common.

Frassle:
Hmm good point I'll see if we have cases elsewhere of printing command suggestions

Frassle:
It looks like we're a little inconsistent, seems mostly to be ` but some '. I'll do a pass to be consistent about using ` in another PR.

PR #10423, opened 2022-08-17
Move default secret manager to stack

Frassle:
This removes a type test for stack type, allowing us to use MockStacks in this area of code.


Zaid-Ajaj (approved):
Looks good!

Issue #10091, opened 2022-07-11
Reduce PR blocking and keep trunk green with merge queues


AaronFriel:
We would like to prevent false results in our testing system and increase CI velocity by moving many of our CI checks to a step _between_ clicking merge PR and the result being placed onto the master branch.

This could occur via the [bors bot](https://bors.tech/) or likely later this year, [GitHub merge queues](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue).

AaronFriel:
This is also blocked, as noted in a design meeting, by having a method of producing release notes / a changelog that doesn't require each PR commit to the same file.

iwahbe:
We have this as a separate issue: https://github.com/pulumi/pulumi/issues/9945.

PR #10427, opened 2022-08-17
ci: Build CLI in test using Linux runners, production using macOS

AaronFriel:
Updated based on team discussion, we will use Linux runners for PR tests and continue building release binaries on macOS.

------

Blocked on #10213, this switches our CI system to build all binaries using cross-compilation from Ubuntu Linux runners, reducing macOS contention.

There is one prominent downside to this that is relevant:

Building on macOS with CGO_ENABLED=1 is required for the CLI to use split-tunnel VPNs on macOS that use per-DNS-name tunneling. This could impact enterprise clients with self-hosted Pulumi services. See:
- https://github.com/golang/go/issues/12524

For that reason, we may want to discuss running builds on Linux in pre-merge checks, and when implementing merge queues in #10091, ensure that merge checks & release builds use macOS builders.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,6 +32,11 @@ on:
         default: false
         required: false
         type: boolean
+      macos-build-platform:
+        description: 'Platform to use when building macOS binaries'
+        default: macos-latest
+        required: false
+        type: string
     secrets: {}
 
 env:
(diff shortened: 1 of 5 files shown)


t0yv0:
The code looks OK of course it's just a matter of deciding what's the right course of action. Continuing to release native Mac builds seems prudent. Question is if they're tested and we can still get a gain here.

This sounds good:

> For that reason, we may want to discuss running builds on Linux in pre-merge checks, and when implementing merge queues in https://github.com/pulumi/pulumi/issues/10091, ensure that merge checks & release builds use macOS builders.

In current setup, this can be good to:

PRs check linux builds
master.yml checks Mac builds
release.yml checks Mac builds

We believe the difference is going to be negligible wrt our current test suite right?

AaronFriel:
@t0yv0 yes - although in my conception of #10091 these are the same thing and will be replacing by a single `staging.yml` that runs on merge queues that builds, tests, and if all tests pass, then prepares a draft release that is only a couple clicks away from being published.

> master.yml checks Mac builds
> release.yml checks Mac builds

iwahbe (approved):
LGTM

PR #10429, opened 2022-08-17
Terminate test if RunCommand fails

Frassle:


RobbieMcKinstry:
I checked to see if the caller was perhaps inspecting stderr or something. Looks like those values are just disregarded. This LGTM.

RobbieMcKinstry (approved):

Issue #10439, opened 2022-08-18
Ampersand in output being converted to \u0026 in state file


eli88fine:
### What happened?

In the console output of the `up`, the output shows up as expected:

`auto-eng-appstream-relay-endpoint-url: "https://appstream2.us-east-1.aws.amazon.com/saml?stack=auto-eng-appstream--appstream--prod-4e3f21b&accountId=111111111111"`

However, when I go to the state file to look at it to grab a bunch of outputs (I'm grabbing them manually from the state file because I need to give them to our SSO IT provider so that they can configure SSO access for this appstream), the ampersand has been converted to escaped unicode (\u0026)...which definitely does **not** work if you try and use it in a URL:
`"auto-eng-appstream-relay-endpoint-url": "https://appstream2.us-east-1.aws.amazon.com/saml?stack=auto-eng-appstream--appstream--prod-4e3f21b\u0026accountId=111111111111"`

I'm using an S3 bucket to store the state, but hopefully that doesn't have anything to do with the issue

### Steps to reproduce

`export("auto-eng-appstream-relay-endpoint-url", "https://appstream2.us-east-1.aws.amazon.com/saml?stack=auto-eng-appstream--appstream--prod-4e3f21b&accountId=111111111111")`

### Expected Behavior

the output in the console matches the output in the state file

### Actual Behavior

the state file converted the ampersand to \u0026

### Output of `pulumi about`

`pulumi about` does not work when using self-hosted Pulumi.

But here's `pulumi plugin ls`

pulumi plugin ls
NAME        KIND      VERSION  SIZE    INSTALLED    LAST USED
aws         resource  5.10.0   391 MB  2 weeks ago  2 weeks ago
aws-native  resource  0.19.0   82 MB   2 weeks ago  2 weeks ago
command     resource  0.4.1    24 MB   2 weeks ago  2 weeks ago

and here's `pip freeze` (I'm using python-flavored pulumi):

Arpeggio==1.10.2
astroid==2.11.7
attrs==21.4.0
boto3==1.24.44
boto3-stubs==1.24.44
botocore==1.27.45
botocore-stubs==1.27.33
cattrs==22.1.0
cfgv==3.3.1
click==8.1.3
colorama==0.4.5
copier==6.1.0
dill==0.3.5.1
distlib==0.3.5
dunamai==1.12.0
exceptiongroup==1.0.0rc8
filelock==3.7.1
grpcio==1.47.0
iam-floyd==0.407.0
[...]

PR #10440, opened 2022-08-18
Consistently disable HTML escaping in JSON output

Frassle:
# Description

Fixes https://github.com/pulumi/pulumi/issues/10439

## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -1,6 +1,6 @@
 ### Improvements
 
-- [provider/python]: Improved exception display. The traceback is now shorter and it always starts with user code.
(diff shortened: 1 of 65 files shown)


Frassle:
ah of course this horribly breaks config, will need to be a bit more targeted about what is or isn't indented.

Frassle:
ah this changed new line behaviour as well... I _think_ that's ok and we can just change the tests but will double check.

Frassle:
Welp, this is fun.
`json.Marshal` doesn't give any control over the escapeHTML option.
`json.NewEncoder().Encode` doesn't give any control over it writing out a new line character after every `Encode` call.

Go has the most annoying design decisions.

Zaid-Ajaj (approved):
Nice one!

PR #10443, opened 2022-08-18
refactor: simplify error handling in readProject API

AaronFriel:
Simplifies some error handling by bubbling up a named & wrapped error which `pulumi destroy` can handle as a special case. Removes the need for handling a return value with an empty project and no error in commands that cannot handle that case.

### DIFF
--- a/pkg/cmd/pulumi/destroy.go
+++ b/pkg/cmd/pulumi/destroy.go
@@ -28,6 +28,7 @@ import (
 	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/cmdutil"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/contract"
+	"github.com/pulumi/pulumi/sdk/v3/go/common/util/logging"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/util/result"
 	"github.com/pulumi/pulumi/sdk/v3/go/common/workspace"
 )
@@ -132,7 +133,8 @@ func newDestroyCmd() *cobra.Command {
 
 			proj, root, err := readProject()
 			if err != nil && errors.Is(err, workspace.ErrProjectNotFound) {
-				// Allow destroy outside of a project using backend to resolve.
+				logging.Warningf("failed to find current Pulumi project, continuing using stack %v from backend %v",
+					s.Ref().Name(), s.Backend().Name())
 				proj = &workspace.Project{}
 				root = ""
 			} else if err != nil {
--- a/pkg/cmd/pulumi/destroy.go
+++ b/pkg/cmd/pulumi/destroy.go
@@ -133,7 +133,7 @@ func newDestroyCmd() *cobra.Command {
 
 			proj, root, err := readProject()
 			if err != nil && errors.Is(err, workspace.ErrProjectNotFound) {
-				logging.Warningf("failed to find current Pulumi project, continuing using stack %v from backend %v",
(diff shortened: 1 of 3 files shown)


pkg/cmd/pulumi/destroy.go
```diff
@@ -130,7 +131,11 @@ func newDestroyCmd() *cobra.Command {
 			}
 
 			proj, root, err := readProject()
-			if err != nil {
+			if err != nil && errors.Is(err, workspace.ErrProjectNotFound) {
+				// Allow destroy outside of a project using backend to resolve.
+				proj = &workspace.Project{}
+				root = ""
```
Frassle, line 137: logging.Warningf the text about we couldn't find the project but we're carrying on anyway

pkg/cmd/pulumi/destroy.go
```diff
@@ -130,7 +131,11 @@ func newDestroyCmd() *cobra.Command {
 			}
 
 			proj, root, err := readProject()
-			if err != nil {
+			if err != nil && errors.Is(err, workspace.ErrProjectNotFound) {
```
Frassle, line 134: This isn't right? The error won't directly be an ErrProjectNotFound, because you wrapped it another error

AaronFriel: I learned that this works! `fmt.Errorf` with a `%w` argument wraps errors and allows you to add helpful error messages around basic errors like `fs.ErrWhatever`.

https://github.com/golang/go/blob/release-branch.go1.17/src/fmt/print.go#L574

It's quite powerful.

pkg/cmd/pulumi/destroy.go
AaronFriel: 👍 good call

pkg/cmd/pulumi/destroy.go
Frassle: And Is picks up the inner wrapped error?

AaronFriel: Yep. https://go.dev/play/p/CCkYSjwVvuR

pkg/cmd/pulumi/destroy.go
```diff
@@ -130,7 +132,12 @@ func newDestroyCmd() *cobra.Command {
...
+				logging.Warningf("failed to find current Pulumi project, continuing using stack %v from backend %v",
```
Frassle, line 136:
```suggestion
				logging.Warningf("failed to find current Pulumi project, continuing continuing with an empty project using stack %v from backend %v",
```

iwahbe (approved):
LGTM

pkg/cmd/pulumi/destroy.go
AaronFriel: Applying the suggestion & merging after tests complete.

pkg/cmd/pulumi/destroy.go
```diff
@@ -130,7 +132,12 @@ func newDestroyCmd() *cobra.Command {
 			}
 
 			proj, root, err := readProject()
-			if err != nil {
+			if err != nil && errors.Is(err, workspace.ErrProjectNotFound) {
+				logging.Warningf("failed to find current Pulumi project, continuing continuing with an empty project using stack %v from backend %v",
```
iwahbe, line 136:
```suggestion
				logging.Warningf("failed to find current Pulumi project, continuing with an empty project using stack %v from backend %v",
```

iwahbe (commented):
Typo

Frassle:
Would be good to merge this, the current state of things is making "pulumi about" report odd errors.

PR #10447, opened 2022-08-19
Fix non-pulumi owned provider import alias

aq17:
# Description

Fixes https://github.com/pulumi/pulumi-yaml/issues/309

## Checklist


pkg/codegen/nodejs/utilities.go
```diff
@@ -99,6 +99,9 @@ func makeValidIdentifier(name string) string {
 		}
 	}
 	name = builder.String()
+	if strings.HasPrefix(name, "pulumi_") {
+		name = strings.TrimPrefix(name, "pulumi_")
+	}
```
AaronFriel, line 104: Putting this here in `makeValidIdentifier` seems like it would affect more than just imports. It looks like the package name is inconsistently handled. I'd look at `genPreamble` to figure out why the import is generated as `pulumi_eventstorecloud`. 

The schema for Pulumi package says its name is `eventstorecloud`: https://github.com/EventStore/pulumi-eventstorecloud/blob/main/provider/cmd/pulumi-resource-eventstorecloud/schema.json

But the NPM package is `@eventstore/pulumi-eventstorecloud`. 

Is `genPreamble` generating the `import * as` name from the name of the NPM package instead of the name of the Pulumi package? That'd be my hunch.

AaronFriel (changes requested):

pkg/codegen/nodejs/utilities.go
aq17: The package name is correctly `@eventstore/pulumi-eventstorecloud`. We strip it to `pulumi-eventstorecloud` with [Base()](https://github.com/pulumi/pulumi/blob/master/pkg/codegen/nodejs/gen_program.go#L322), then swap the `-` for `_`. It looks like we'll need a way of determining this provider originates from a non-pulumi repo, similar to the change [here](https://github.com/pulumi/pulumi/pull/10485)

iwahbe (commented):
I think that `makeValidIdentifier` is (and should remain) special purposed to take an identifier and deterministically make it valid. It should not be used for correcting pulumi names. 

Pulumi does not reserve the `pulumi_` namespace, so this is not a reliable way to determine if a package is Pulumi owned. The schema allows you to override package import names, so the only valid way to get an import name is to look at the schema:
https://github.com/pulumi/pulumi/blob/1cc5bdc91a2baef50160ac440373354e3041d070/pkg/codegen/nodejs/importer.go#L29-L32

The only way to correctly determine if a package is Pulumi owned is to check the `Package.Publisher == "Pulumi"` field. I'm scared of code that relies on that for correctness though.

pkg/codegen/nodejs/utilities.go
iwahbe, line 104: Trim prefix includes its own existence check, so you don't need to duplicate it.
```suggestion
	name = strings.TrimPrefix(name, "pulumi_")
```

This is the implementation of `strings.TrimPrefix`:
```go
func TrimPrefix(s, prefix string) string {
	if HasPrefix(s, prefix) {
		return s[len(prefix):]
	}
	return s
}
```

AaronFriel: Right - I think the issue is that we're working with the wrong string here. The issue isn't detecting whether it's "Pulumi owned", it's that we're doing:

```
makeValidIdentifier(NPM package name after the /)
```

Instead of:

```
makeValidIdentifier(Pulumi package name)
```

Where:

* NPM package name: `@eventstore/pulumi-eventstorecloud`
* Pulumi package name: `eventstorecloud`

The NPM package name ending in the same string as the Pulumi package name here is a coincidence and a misdirect.

iwahbe (approved):
I would love to see a test here. Otherwise this looks simple and sweet.

aq17:
> I would love to see a test here. Otherwise this looks simple and sweet.

Double checked this for other languages, python and dotnet look fine, but unfortunately in go, we incorrectly import it as `"github.com/pulumi/pulumi-eventstorecloud/sdk/go/eventstorecloud"` whereas the[ actual import is github.com/EventStore/...](https://www.pulumi.com/registry/packages/eventstorecloud/installation-configuration/). However, it looks like our [GoPackageInfo](https://github.com/pulumi/pulumi/blob/master/pkg/codegen/go/gen_program.go#L217) has the `ImportBasePath` as `github.com/pulumi/pulumi-eventstorecloud/sdk/go/eventstorecloud`.

pkg/codegen/nodejs/gen_program.go
```diff
@@ -319,7 +321,12 @@ func (g *generator) genPreamble(w io.Writer, program *pcl.Program, preambleHelpe
 		if pkg == "@pulumi/pulumi" {
 			continue
 		}
-		as := makeValidIdentifier(path.Base(pkg))
+		var as string
+		if puPkg, ok := npmToPuPkgName[pkg]; ok {
+			as = makeValidIdentifier(path.Base(puPkg))
+		} else {
+			as = makeValidIdentifier(path.Base(pkg))
```
AaronFriel, line 328: Would this work if it was just:

`as = makeValidIdentifier(pkg)`?

AaronFriel (approved):

Issue #10339, opened 2022-08-09
[cli] Add a CLI command to migrate stacks between backends (`pulumi stack migrate`?)


lukehoban:
It's common to need to migrate stacks between backends.  A few examples:
1. From a filestate backend (e.g. S3) to the Pulumi Service
2. From the Pulumi Service to the Self-Hosted Pulumi Service
3. From the Self-Hosted Pulumi Service to the Pulumi Service

It's possible to do this today via `pulumi login A` + `pulumi stack export` + `pulumi login B` + `pulumi stack init`  + `pulumi stack import`.  However - that doesn't include a few things:
a. It only includes the most recent state file, not all the historical updates
b. It doesn't record the correct update times for any of the updates
c. It doesn't include update logs for Service to Service migrations
d. It requires manually switching between backends, and is thus hard to automate for bulk migration of stacks

We should add a feature to the Pulumi CLI to address at least (a) and (d), and ideally also (b) and (c).  To be able to support high-fidelity Service-to-Service migrations (for to/from self-hosted in particular), this may require some new APIs in the Pulumi Service.

Frassle:
https://github.com/pulumi/pulumi/pull/10455 adds a first basic version of this. We'll need to work with the service team to accomplish anything more.

tmehlinger:
I would add an item (e) to the list above, export/import requires rewriting URNs. See https://github.com/pulumi/pulumi/pull/10455#issuecomment-1227825148.

Frassle:
Fixed by https://github.com/pulumi/pulumi/pull/22902

PR #10455, opened 2022-08-19
Basic stack migrate command

Frassle:
# Description

This isn't anything fancy, just a glorifed way to do stack export, login to new backend, stack export.
But this lays the groundwork for a better migrate system as and when the service can support it.

Part of https://github.com/pulumi/pulumi/issues/10339

## Checklist


pkg/cmd/pulumi/stack_migrate.go
```diff
@@ -0,0 +1,152 @@
...
+				ctx, opts,
+				getBackend,
+				stackName, version, target,
+				args[0], secretProvider)
+		}),
+	}
+	cmd.PersistentFlags().StringVarP(
+		&stackName, "stack", "s", "", "The name of the stack to operate on. Defaults to the current stack")
+	cmd.PersistentFlags().StringVarP(
+		&stackName, "target", "t", "", "The name of the new stack to create. Defaults to the name of source stack")
```
kpitzen, line 148: should this be `&target`?

pkg/cmd/pulumi/stack_migrate_test.go
```diff
@@ -0,0 +1,207 @@
...
+		},
+		ExportDeploymentF: func(ctx context.Context) (*apitype.UntypedDeployment, error) {
+			return mockDeployment, nil
+		},
+		RefF: func() backend.StackReference {
+			return &mockStackReference{name: "some_stack"}
+		},
+		DefaultSecretManagerF: func(configFile string) (secrets.Manager, error) {
+			if configFile == "" {
+				f, err := workspace.DetectProjectStackPath("some_stack")
```
kpitzen, line 117: Just for my own clarity - what's the significance of `"some_stack"` here?

pkg/cmd/pulumi/stack_migrate.go
```diff
@@ -0,0 +1,152 @@
...
+			secretProvider := ""
+			if len(args) > 1 {
+				secretProvider = args[1]
+			}
+
+			return migrateStack(
+				ctx, opts,
+				getBackend,
+				stackName, version, target,
+				args[0], secretProvider)
```
Zaid-Ajaj, line 142: Maybe check that `args` is not empty and report an error if that was the case before getting first element here? Just in case users don't provide any args for `pulumi stack migrate` and get a panic

Frassle: Ah yes, that's what `cmdutil.RangeArgs(1, 2)` is for. Will fix

pkg/cmd/pulumi/stack_migrate.go
Frassle: Yes

kpitzen (approved):

pkg/cmd/pulumi/stack_migrate.go
```diff
@@ -0,0 +1,152 @@
...
+
+	if currentConfig.HasSecureValue() {
+		dec, decerr := getStackDecrypter(s)
+		if decerr != nil {
+			return decerr
+		}
+		decrypter = dec
+	} else {
+		decrypter = config.NewPanicCrypter()
+	}
```
dixler, line 81: nit because I always like the challenge of eliminating `else`
```suggestion
	getDeployment := s.ExportDeployment
	if version != "" {
		// Check that the stack and its backend supports the ability to export previous deployments
		be := s.Backend()
		specificExpBE, ok := be.(backend.SpecificDeploymentExporter)
		if !ok {
			return fmt.Errorf("the current backend (%s) does not provide the ability to export previous deployments",
				be.Name())
		}
		getDeployment = func(ctx context.Context) (*apitype.UntypedDeployment, error) {
			return specificExpBE.ExportDeploymentForVersion(ctx, s, version)
		}
	}

	deployment, err := getDeployment(ctx)
	if err != nil {
		return err
	}

	// Build decrypter based on the existing secrets provider
	decrypter := config.NewPanicCrypter().(config.Decrypter)
	currentConfig := currentProjectStack.Config

	if currentConfig.HasSecureValue() {
		dec, decerr := getStackDecrypter(s)
		if decerr != nil {
			return decerr
		}
		decrypter = dec
	}
```

pkg/cmd/pulumi/stack_migrate.go
```diff
@@ -0,0 +1,152 @@
...
+	var target string
+	var version string
+
+	cmd := &cobra.Command{
+		Use:   "migrate <target backend url> [target secret provider]",
+		Args:  cmdutil.RangeArgs(1, 2),
+		Short: "Migrate a stack to a different backend",
+		Long: "Migrate a stack to a different backend.\n" +
+			"\n" +
+			"This will export the current (or selected stack) from the backend currently logged into, and then " +
```
dixler, line 125: Comment: it'd be nice if we could dynamically generate this string to specify the currently selected stack and current backend, but definitely not necessary for this.

Adding to https://github.com/pulumi/home/issues/2181

dixler (commented):
- would like more details in the help description
- have a dumb refactoring nit

mikhailshilkov:
What's the plan of shipping this vs. the more advanced version of the same command? Will you just change it in place as you go? Are you sure that there is enough upside of shipping this basic version and then having to think of backwards compatibility? Are we aligned with product about the messaging, docs, launch content around it? Can we please review the design doc with the overall plan?

TLDR; I wouldn't ship it lightly.

pkg/cmd/pulumi/stack_migrate.go
```diff
@@ -0,0 +1,152 @@
...
+			"import that stack into the new target backend. This will also ",
```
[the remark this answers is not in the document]
Frassle: > Some questions I might ask are whether commands require being in a Pulumi project to run

This currently does depend on being in a project because it also handles re-encrypting config for the new secret backend.

pkg/cmd/pulumi/stack_migrate.go
Frassle: the current is inline with what these code blocks look like elsewhere so I'm inclined to match the current style. 
In a perfect world go would be a good language with expression based "if" and else blocks would make perfect sense, but alas.

pkg/cmd/pulumi/stack_migrate.go
Frassle: I think that _might_ be possible, I'll give it try.

Frassle:
> What's the plan of shipping this vs. the more advanced version of the same command? Will you just change it in place as you go?

That's the current plan, I think the command will stay pretty much the same in terms of UX. We'll just be able to internally copy across more and more data as the backend interface is improved.

> Are you sure that there is enough upside of shipping this basic version and then having to think of backwards compatibility? 

Yes, we already have people asking how to move stacks from filestate to httpstate and this is a good start to covering that. As above, I don't expect the UX to change and I don't think backwards comparability will be an issue. If today you just get the latest stack version copied across, but tomorrow it correctly pulls all the history with it as well I don't think that will break anyone. Keep in mind this is _only_ for creating new stacks, it doesn't edit existing ones in the new backend.

> Are we aligned with product about the messaging, docs, launch content around it? Can we please review the design doc with the overall plan?

That needs doing. We need to work with service to see how/what can be extended on the service API to support more here.

> TLDR; I wouldn't ship it lightly.

Understood, but as above I don't think there are any major concerns around adding this as a starting point for migration. Will sync with everyone before merging though.

tmehlinger:
Since this relies on exporting and importing stacks, it needs to address the current challenge with doing so manually. Importing stack state tends to produce warnings like these:

```
warning: resource 'urn:pulumi:STACK::PROJECT::pulumi:pulumi:Stack::PROJECT-STACK' is from a different stack (STACK != NEW_STACK_NAME)
```
The importing code needs to rewrite URNs to account for the differences in project and stack names. It also needs to account for URN differences in stack references, and outputs from referent stacks.

Frassle:
> The importing code needs to rewrite URNs to account for the differences in project and stack names. It also needs to account for URN differences in stack references, and outputs from referent stacks.

This is only moving from one backend to another, it does not rename the project or stack in the process of doing so. `stack rename` can be used separately for that purpose. As such this command doesn't need to handle URN rewrites.

PR #10456, opened 2022-08-19
Accept options.version on pulumi convert

aq17:
# Description

Unfinished change to unblock failing tests for https://github.com/pulumi/pulumi-yaml/pull/291
## Checklist

### DIFF
--- a/pkg/codegen/pcl/binder_resource.go
+++ b/pkg/codegen/pcl/binder_resource.go
@@ -331,7 +331,11 @@ func (b *binder) bindResourceBody(node *Resource) hcl.Diagnostics {
 					t = model.NewListType(ResourcePropertyType)
 					resourceOptions.IgnoreChanges = item.Value
 				case "version":
-					continue
+					t = model.StringType
+					resourceOptions.Version = item.Value
+				case "pluginDownloadURL":
+					t = model.StringType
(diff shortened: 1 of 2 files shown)


pkg/codegen/pcl/binder_resource.go
```diff
@@ -330,6 +330,8 @@ func (b *binder) bindResourceBody(node *Resource) hcl.Diagnostics {
 				case "ignoreChanges":
 					t = model.NewListType(ResourcePropertyType)
 					resourceOptions.IgnoreChanges = item.Value
+				case "version":
+					continue
```
iwahbe, line 334: Shouldn't we assert that this is a string.
We should probably do the same for `pluginDownloadURL` (which always appears in the same context as `version`, since (name, version, pluginDownloadURL) is the triple used to uniquely define a provider.)

aq17: This PR is mostly intended to unblock[ PR 291](https://github.com/pulumi/pulumi-yaml/pull/291), as we'll implement the logic for the binder to handle versions in a separate PR. Right now, it is still being ignored, so we could do the assertion in the separate PR?

iwahbe (approved):

PR #10458, opened 2022-08-22
[dotnet/sdk] Implement Deployment.TestAsync overloads which accept functions

### LINKED ISSUE #10438: [dotnet/sdk] Implement Deployment.TestAsync overloads that don't require a Stack definition
Right now, when using `Deployment.RunAsync` with callback functions instead of Stacks, users cannot easily use those functions in unit tests without having to implement a dummy test stack in order to run `Deployment.TestAsync`. 

It would be great to implement overloads of `Deployment.TestAsync` that accept the same signatures that `Deployment.RunAsync` does. 

I propose that the callback overloads return not only a list of created resources but also a dictionary of keyed outputs:
```cs
ImmutableArray<Resource> TestAsync(
    IMocks testMocks,
    TestOptions options,
    Action createResources
);

(ImmutableArray<Resource>, Dictionary<string, object>) TestAsync(
    IMocks testMocks,
    TestOptions options,
    Action<Dictionary<string, object>> createResources
);
```
And their async counter parts as well. 

The `TestAsync<'Stack>` overload returns a list of created resources, _including_ the type of the `Stack` itself as one of created resources. I think with the overloads above, they shouldn't return the stack as a created resource since users used a lambda and not a stack explicitly. i.e.
```cs
var resources = Deployment.TestAsync(mocks, options, () =>
{
    new Bucket("test-bucket");
});

Assert.That(resources.Length == 1, "Only one resource has been created");
```

### Affected area/feature

dotnet/sdk

Zaid-Ajaj:
Fixes #10438 

The implementation uses an `internal TestStack` which creates the resources in the constructor and makes the outputs available. This stack is then filtered out from the created resources during `Deployment.TestAsync` and the outputs are retained in the following API:
```cs
var (resources, outputs) = await Deployment.TestAsync(mocks, options, () =>
{
    // Create resources here
    // return outputs
});
```
and without returning outputs
```cs
var resources = await Deployment.TestAsync(mocks, options, () =>
{
    // Create resources here
});
```

## Checklist


sdk/dotnet/Pulumi/Deployment/Deployment_Run.cs
```diff
@@ -166,6 +166,128 @@ public static Task<ImmutableArray<Resource>> TestAsync<TStack>(IMocks mocks, Tes
...
+        /// <param name="createResources">The function which creates resources and returns outputs.</param>
+        /// <returns>Test result containing created resources and outputs, if any.</returns>
+        public static async Task<(ImmutableArray<Resource> Resources, IDictionary<string, object?> Outputs)> TestAsync(
+            IMocks testMocks,
+            TestOptions testOptions,
+            Func<Task<IDictionary<string, object?>>> createResources)
+        {
+            var createdResources = await TestAsync(
+                mocks: testMocks,
+                runAsync: runner => runner.RunAsync(() => new TestStack(createResources)),
```
Frassle, line 216: Why does this need TestStack? Can't we just pass createResources to RunAsync?

sdk/dotnet/Pulumi/Deployment/Deployment_Run.cs
```diff
@@ -166,6 +166,128 @@ public static Task<ImmutableArray<Resource>> TestAsync<TStack>(IMocks mocks, Tes
             where TStack : Stack, new()
             => TestAsync(mocks, runner => runner.RunAsync<TStack>(), options);
 
+        /// <summary>
+        /// Used inside TestAsync overloads where users have a function that creates resources
+        /// in which case an internal TestStack is used to create the resources.
+        ///
+        /// This function takes the created resources from the TestStack and filters it out of the created resources
```
Frassle, line 173: I'm not sure we should filter out the implicitly created stack. The stack resource will be created even for lambda based deployments, test framework should probably reflect that.

sdk/dotnet/Pulumi/Deployment/Deployment_Run.cs
Zaid-Ajaj: Still need the `TestStack` since this is how we obtain the returned dictionary containing outputs

sdk/dotnet/Pulumi/Deployment/Deployment_Run.cs
Frassle: Ok lets go with that for now, I'm not 100% convinced but this change is still worth having

Frassle (approved):

PR #10460, opened 2022-08-22
fix: Sporadic initialization panics in ssh-agent library on Windows

AaronFriel:
Follow on from #10207 to ensure pulumi-language-* are built without using this ssh-agent library.

To address systemically, as `replace` directives only apply to the current Go module and not to any users of it, I've introduced a PR upstream here: https://github.com/xanzy/ssh-agent/pull/10

If that is merged and a release cut to say v0.4.0, we can depend on xanzy/ssh-agent@v0.4.0 and [go version selection](https://go.dev/ref/mod#minimal-version-selection) will ensure that all binaries that depend on pkg or sdk will include the lazy initialization routine that mitigates this panic.


t0yv0 (approved):
Nice, let us try it! 🙇‍♂️

PR #10461, opened 2022-08-22
add info message on empty update

### LINKED ISSUE #9229: detect when an _empty_ update is being run
## overview
can we detect when someone is running an update with nothing being deployed beyond the stack itself and warn the user about this?

folks may hit this when they forget to save the file.

more than likely, folks don't mean to do this, but a lot of new users may do this by accident; if we can message them ahead of time, it may help them better understand what is happening and adjust ahead of running the update

dixler:
# Description
This PR provides an INFO message to indicate to users if the update that they are trying to run is not deploying any resources(other than the `pulumi:pulumi:Stack` resource).

Fixes #9229 

No resources:
[asciicast](https://asciinema.org/a/PaYwBCJqCiqtDywvy1Cmp3JdT)

## Checklist


susanev (commented):
yay!! thank you! do you mind if i give some feedback on the message?

pkg/backend/apply.go
```diff
@@ -125,6 +125,16 @@ func PreviewThenPrompt(ctx context.Context, kind apitype.UpdateKind, stack Stack
 		return plan, changes, nil
 	}
 
+	if kind == apitype.UpdateUpdate &&
```
Frassle, line 128: Should probably be for previews as well

pkg/backend/apply.go
```diff
@@ -230,6 +240,28 @@ func PreviewThenPromptThenExecute(ctx context.Context, kind apitype.UpdateKind,
...
+	count := 0
+	seen := make(map[resource.URN]engine.StepEventMetadata)
+
+	displayOpts.SummaryDiff = true
+
+	for _, e := range events {
+		if e.Type == engine.SummaryEvent {
+			continue
+		}
+		msg := display.RenderDiffEvent(e, seen, displayOpts)
```
Frassle, line 253: Can't we look at the type token directly?

pkg/backend/apply.go
```diff
@@ -125,6 +125,16 @@ func PreviewThenPrompt(ctx context.Context, kind apitype.UpdateKind, stack Stack
 		return plan, changes, nil
 	}
 
+	if kind == apitype.UpdateUpdate &&
+		countResources(events, op.Opts.Display) == 0 {
+
+		emptyStackWarningMessage := "\b" + op.Opts.Display.Color.Colorize(
+			fmt.Sprintf("%sinfo:%s it appears that there are no resources in this update\n",
```
susanev, line 132:
```suggestion
			fmt.Sprintf("%sinfo:%s there are no resources in this update\n",
```

is _resources_ the right word here? or should it be _resource changes_?

dixler: The issue that was opened only mentioned deploying  a stack that only contains a stack resource and nothing else.

> can we detect when someone is running an update with nothing being deployed beyond the stack itself and warn the user about this?

I may have misunderstood your ask.

pkg/backend/apply.go
dixler: I've looked through the codebase and `preview` actually invokes `apply` directly and doesn't go through this pathway as it returns a `*deploy.Plan` the changes to include this on preview, would be much larger and would change a lot of function signatures.

pkg/backend/apply.go
susanev: @dixler ah yah, thx for reminding me! this is confusing! i tried an edit, what do you think?

dixler:
I have updated the flow to include stacks where no changes are detected. Here is a recording of hitting both flows.

https://asciinema.org/a/z0geCZ7yYfEUvFNrDdZ1xhIdV

cc @susanev

pkg/backend/apply.go
dixler: imo users don't specify the stack resource so it's an internal detail that can be neglected

mikhailshilkov:
Is there a better way to visualize this? In the demo above, there are two elements that display ~the same thing:

```
Resources:                                                                                                                       
   1 unchanged                                                                                                                                                                                                                                           

info: There are no resources found in this update.
```

Should we combine "1 unchanged" and the info message into a single coherent indication? I'm not exactly sure about the wording - but the info message feels a bit alien to me. Also, does it solve the original goal - warning users about a potential mistake they are making?

pkg/backend/apply.go
Frassle: Ugh backend code.
We should fix this, a vast amount of the value here will be pointing this out to people at preview time.

susanev (approved):
thx Kyle and sry for the delay!

id suggest putting a newline after the info message because it kinda gets lost when there's no space between it and the question about doing the update. but up to ya.

pkg/backend/apply.go
```diff
@@ -125,6 +125,17 @@ func PreviewThenPrompt(ctx context.Context, kind apitype.UpdateKind, stack Stack
 		return plan, changes, nil
 	}
 
+	infoPrefix := "\b" + op.Opts.Display.Color.Colorize(colors.SpecWarning+"info: "+colors.Reset)
```
dixler, line 128: modified to use the `warning` yellow color as we use it to indicate that there's a new version of pulumi.

pkg/backend/apply.go
```diff
@@ -125,6 +125,17 @@ func PreviewThenPrompt(ctx context.Context, kind apitype.UpdateKind, stack Stack
...
+	if kind != apitype.UpdateUpdate {
+		// If not an update, we can skip displaying warnings
+	} else if countResources(events) == 0 {
+		// This is an update and there are no resources being CREATED
+		fmt.Print(infoPrefix, "There are no resources in your stack(other than the stack resource).\n\n")
```
dixler, line 133: @susanev I've incorporated your feedback into this version of the error message.
- add newline
- mention stack resource

@mikhailshilkov does this read better?

pkg/backend/apply.go
```diff
@@ -125,23 +125,15 @@ func PreviewThenPrompt(ctx context.Context, kind apitype.UpdateKind, stack Stack
...
 	} else if !hasChanges(events) {
 		// This is an update and there are no resources being UPDATED
-		noChangesWarningMessage := "\b" + op.Opts.Display.Color.Colorize(
-			fmt.Sprintf("%sinfo:%s There are no changes found in this update.\n",
-				colors.SpecInfo,
-				colors.Reset))
-		fmt.Printf(noChangesWarningMessage)
-
+		fmt.Print(infoPrefix, "This update does not contain any changes to resources.\n\n")
```
dixler, line 136: Does this read better than the previous

> info: There are no changes found in this update.

dixler (commented):
Hopefully this version is better.

I can incorporate tests. I just would like to iron out the specifics around the message.

cc @mikhailshilkov @susanev

pkg/backend/apply.go
mikhailshilkov: I think it does read better.

It still does feel repetitive to me. Feel free to ignore me, but reading the goals in the issue

> more than likely, folks don't mean to do this, but a lot of new users may do this by accident; if we can message them ahead of time, it may help them better understand what is happening and adjust ahead of running the update

> can we message them that this was probably not attended in clear language, along with some suggestions for what they should do next, e.g., save. happy to work with someone to figure out the details of how we want to solve the issue.

I don't feel like we addressed them? Should the message focus on a CTA?

cnunciato: Apologies if I'm reading this wrong, but I wouldn't consider "no changes" something we should warn users about or even message as something out of the ordinary. Quite often "no changes" is just what you want to see.

dixler: > Apologies if I'm reading this wrong, but I wouldn't consider "no changes" something we should warn users about or even message as something out of the ordinary. Quite often "no changes" is just what you want to see.

I think the message is to provide confirmation to novice users that the diff summary means what they think it might mean.

I marked it as `info`, but used the warn coloring to indicate where the error message starts and to draw attention and the purple doesn't show up well on a black terminal.

The reason I thought to even use the warn coloring was from:
```
warning: A new version of Pulumi is available. To upgrade from version '3.35.3' to '3.39.0', run 
   $ curl -sSL https://get.pulumi.com | sh
```

but I never realized that it actually said `warning`. should I open an issue to change that to an info message and we can use the info coloring in this location and then we can start to use the info coloring more?

dixler:
I was looking into modifying an integration test on this, but the codepath to trigger this is via interactive mode. I could add unit tests to test the function that I put together, but they'd be incredibly contrived and I don't believe that they'd be very useful.

PR #10462, opened 2022-08-22
Reintroduce renew lease optimization

t0yv0:
# Description

Re-introduce the optimization that avoids calling renewLease on critical startup path. Previous attempt caused issues and was reverted (see [Postmortem](https://docs.google.com/document/d/1rgeMg8bvEf1w3Ysv3GjGBVZrjb2dzTPPD1HLeH8a9j4/edit?usp=sharing)). The underlying backend issue with a short expiration on initial token has been patched, so this is safe to reintroduce. Besides reintroducing the original code, the PR addresses @RobbieMcKinstry concern that we cannot solely rely on background ticker to keep the token fresh (up-to-date). To address this, we chose to block the caller of GetToken and refresh the token right then rather than risking returning a stale token. 

Fixes https://github.com/pulumi/home/issues/2200

## Checklist


t0yv0:
@RobbieMcKinstry WDYT? I'm considering maybe lifting the refresh logic out to spell out some unit tests as this is getting more and more tricky.

t0yv0:
Ah looks like our integration tests are flagging this as invalid, working it out.

pkg/backend/httpstate/state.go
```diff
@@ -48,53 +48,79 @@ type tokenSource struct {
...
 		}
-	}()
 
-	return &tokenSource{requests: requests, done: done}, nil
+		now := time.Now()
+
+		// We will renew the lease after 50% of the duration
+		// has elapsed to allow time for retries.
+		stale := now.Add(duration / 2).After(state.expires)
+		if !stale {
```
RobbieMcKinstry, line 88:
```suggestion
		if stale := now.Add(duration / 2).After(state.expires); !stale {
```

RobbieMcKinstry (approved):
Looks good overall. I feel less confident with my review after the last time however ;D

pkg/backend/httpstate/token_source.go
```diff
@@ -0,0 +1,135 @@
...
+			renewUpdateLeaseIfStale()
+		case c, ok := <-ts.requests:
+			if !ok {
+				close(ts.done)
+				return
+			}
+			// If ticker has not kept up, block on
+			// renewing rather than risking returning a
+			// stale token.
+			renewUpdateLeaseIfStale()
```
t0yv0, line 115: Oof - i've tried adding a test for this case, but it seems like it's not possible to reach here. The problem is that we need an active token to renew a lease. If we're detecting we've not kept up with refreshing tokens, we have a stale token, then we won't be able to review it either.  There is some grace here perhaps that the staleness check indicates a need to renew slightly before actual expiration. So if the ticker fails completely, this safeguard might be beneficial to keep. How to test this though..

pkg/backend/httpstate/state.go
t0yv0: Prefer variable name as docu on what this is.

RobbieMcKinstry (approved):

PR #10471, opened 2022-08-23
Schema for project files

Frassle:
# Description

Can be added to vscode for validation while writing project files.

Not using to validate loaded project files yet, but we may want to do that in the future. I expect this to be _much_ more useful when users start writing config entries in project files.

image

### DIFF
--- a/sdk/go/common/apitype/project.json
+++ b/sdk/go/common/apitype/project.json
@@ -1,0 +1,144 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "https://github.com/pulumi/pulumi/blob/master/sdk/go/common/apitype/project.json",
+    "title": "Pulumi Project",
+    "description": "A schema for Pulumi project files.",
+    "type": "object",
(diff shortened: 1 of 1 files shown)


AaronFriel:
I prefer we use the LSP to provide this in VS Code, as we can publish the extension and have it be recommended for Pulumi.yaml files.

@iwahbe is this schema helpful for that or is the implementation for project keys different?

Frassle:
https://code.visualstudio.com/docs/languages/json#_json-schemas-and-settings
> VS Code extensions can also define schemas and schema mapping.

So I'd guess we could just write an extension to point at the schema

AaronFriel (approved):
No objection to merging the schema file, only to the suggestion that we publish this independently from the LSP server plugin. That plugin be the recommended way to edit project files in VS Code. We can deliver a better UX if the LSP server handles both `runtime: yaml` and plain project files.

iwahbe:
The LSP completion for top level keys is manually implemented. It's not schema driven.

PR #10472, opened 2022-08-23
 fix: Ports Python CONFIG, ROOT, and _SECRET_KEYS to ContextVars 

### LINKED ISSUE #6052: [Automation API] Support parallel execution of python inline programs

kpitzen:
# Description

This PR makes our global values that are not more complicated classes (like SETTINGS) ContextVars, and updates their interfaces where appropriate.

Fixes #6052 
## Checklist

### DIFF
--- a/CHANGELOG_PENDING.md
+++ b/CHANGELOG_PENDING.md
@@ -19,6 +19,9 @@
 - [sdk/python] Makes global SETTINGS values context-aware to not leak state between Pulumi programs running in parallel
   [#10402](https://github.com/pulumi/pulumi/pull/10402)
 
+- [sdk/python] Makes global ROOT, CONFIG, _SECRET_KEYS ContextVars to not leak state between parallel inline Pulumi programs
+  [#10472](https://github.com/pulumi/pulumi/pull/10472)
+
 ### Bug Fixes
 
 - [codegen/go] Fix StackReference codegen.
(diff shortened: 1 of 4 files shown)


kpitzen:
This PR also depends on #10402

justinvp:
Presumably, once this lands, we can remove the locking we have in the following places:
- https://github.com/pulumi/pulumi/blob/89b08f3f1f0f366f93be3674c50fe220ed312822/sdk/python/lib/pulumi/provider/server.py#L66-L96
- https://github.com/pulumi/pulumi/blob/89b08f3f1f0f366f93be3674c50fe220ed312822/sdk/python/lib/pulumi/provider/server.py#L197-L225

kpitzen:
The approach here is a bit more straightforward given how simple these specific globals are.

Also to your point @justinvp - that seems possible given no context should be able to contend for its own resources, and contextvars will prevent contexts from accessing other contexts' globals!

justinvp (approved):
LGTM

Issue #10450, opened 2022-08-19
Issue unprotecting a gcp app engine


jchs:
### What happened?
```
================================================================================
The Pulumi CLI encountered a fatal error. This is a bug!
We would appreciate a report: https://github.com/pulumi/pulumi/issues/
Please provide all of the below text in your report.
================================================================================
Pulumi Version:   v3.38.0
Go Version:       go1.19
Go Compiler:      gc
Architecture:     arm64
Operating System: darwin
Panic:            runtime error: slice bounds out of range [:-2]


goroutine 1 [running]:
runtime/debug.Stack()
        /Users/runner/hostedtoolcache/go/1.19.0/x64/src/runtime/debug/stack.go:24 +0x64
main.panicHandler()
        /Users/runner/work/pulumi/pulumi/pkg/cmd/pulumi/main.go:29 +0x34
panic({0x104085880, 0x1400047a0d8})
        /Users/runner/hostedtoolcache/go/1.19.0/x64/src/runtime/panic.go:884 +0x204
gopkg.in/AlecAivazis/survey%2ev1.paginate(0x140003e16fa?, {0x140004f27a0?, 0x14000859438?, 0x102873fe0?}, 0x14000859408?)
        /Users/runner/go/pkg/mod/gopkg.in/!alec!aivazis/survey.v1@v1.8.9-0.20200217094205-6773bdf39b7f/survey.go:254 +0xe0
gopkg.in/AlecAivazis/survey%2ev1.(*Select).Prompt(0x1400067e5b0)
        /Users/runner/go/pkg/mod/gopkg.in/!alec!aivazis/survey.v1@v1.8.9-0.20200217094205-6773bdf39b7f/select.go:199 +0xf8
gopkg.in/AlecAivazis/survey%2ev1.Ask({0x140008596f0, 0x1, 0x5?}, {0x103d87520?, 0x14000635300}, {0x0, 0x0, 0x1400067e5b0?})
        /Users/runner/go/pkg/mod/gopkg.in/!alec!aivazis/survey.v1@v1.8.9-0.20200217094205-6773bdf39b7f/survey.go:148 +0x1b0
gopkg.in/AlecAivazis/survey%2ev1.AskOne({0x1041ab5f0?, 0x1400067e5b0?}, {0x103d87520?, 0x14000635300?}, 0x1?, {0x0?, 0x0?, 0x0?})
[...]

PR #10475, opened 2022-08-23
Fix negative page size panic

t0yv0:
# Description

https://github.com/pulumi/pulumi/pull/10130 introduced a Terminal UI fix to optimize the pageSize parameter when displaying large selectors via the `survey` library. Unfortunately under certain conditions the logic results in a negative pageSize which causes `survey` library to panic. The fix here centralizes the logic introduced in 10130 into a helper function and makes sure that its return values are non-negative.

Fixes [#10540](https://github.com/pulumi/pulumi/issues/10450)

## Checklist

### DIFF
--- a/pkg/cmd/pulumi/terminal.go
+++ b/pkg/cmd/pulumi/terminal.go
@@ -1,0 +1,44 @@
+// Copyright 2016-2022, Pulumi Corporation.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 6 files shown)


pkg/cmd/pulumi/terminal.go
```diff
@@ -0,0 +1,44 @@
...
+		pageSize = opts.terminalHeight
+	} else if _, height, err := terminal.GetSize(0); err == nil {
+		pageSize = height
+	}
+	if pageSize > opts.nopts {
+		pageSize = opts.nopts
+	}
+	const buffer = 5
+	if pageSize > buffer {
+		pageSize = pageSize - buffer
```
aq17, line 41: Thanks for this 😄

aq17 (approved):
LGTM!

PR #10482, opened 2022-08-24
Add token expiration fields

t0yv0:
# Description

The new fields will enable servers to decide on and indicate token expiration intervals, which in the future will allow the clients to make fewer assumptions about the token lifetime and optimize renewing tokens based on the server-provided expiration.

Fixes # (issue)

## Checklist

### DIFF
--- a/sdk/go/common/apitype/updates.go
+++ b/sdk/go/common/apitype/updates.go
@@ -189,7 +189,7 @@ type RenewUpdateLeaseResponse struct {
 	// The renewed token.
 	Token string `json:"token"`
 
-	// TokenExpiration is the UNIX timestamp by which the token will expire.
+	// TokenExpiration is a UNIX timestamp by which the token will expire.
 	TokenExpiration int64 `json:"tokenExpiration,omitempty"`
 }
 
--- a/sdk/go/common/apitype/updates.go
+++ b/sdk/go/common/apitype/updates.go
@@ -16,7 +16,6 @@ package apitype
 
 import (
 	"encoding/json"
-	"time"
 
 	"github.com/pulumi/pulumi/sdk/v3/go/common/diag/colors"
 )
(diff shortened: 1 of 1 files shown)


sdk/go/common/apitype/updates.go
```diff
@@ -95,6 +96,9 @@ type StartUpdateResponse struct {
 
 	// Token is the lease token (if any) to be used to authorize operations on this update.
 	Token string `json:"token,omitempty"`
+
+	// TokenExpiration is the point in time by which the token will expire.
+	TokenExpiration *time.Time `json:"tokenExpiration,omitempty"`
```
Frassle, line 101: Everywhere else we send times from the service we send them as unix timestamps (int64). This would be different as go formats time.Time like " 2014-05-16T08:28:06.801064-04:00" 
Is that intended?

t0yv0: I just was not paying attention. There seem to be both versions in use? In addition there's int64+*int64 in use.

```go
// ManifestV1 captures meta-information about this checkpoint file, such as versions of binaries, etc.
type ManifestV1 struct {
	// Time of the update.
	Time time.Time `json:"time" yaml:"time"`

// OperationStatus describes the state of an operation being performed on a Pulumi stack.
type OperationStatus struct {
	Kind    UpdateKind `json:"kind"`
	Author  string     `json:"author"`
	Started int64      `json:"started"`
}
// LogEntry is the individual entries in a JSON response to a Logs operation.
type LogEntry struct {
	ID        string `json:"id"`
	Timestamp int64  `json:"timestamp"`
	Message   string `json:"message"`
}
// StackSummary describes the state of a stack, without including its specific resources, etc.
type StackSummary struct {
	// LastUpdate is a Unix timestamp of the start time of the stack's last update, as applicable.
	LastUpdate *int64 `json:"lastUpdate,omitempty"`
```

I'm fine with using `*in64` or whatever makes sense.

t0yv0: @isabeldepapel in case service team has a preference here.

isabeldepapel: No, no preference--whatever you all thinks makes the most sense works for us

Frassle: I think int64 would be more consistent. The service doesn't actually write ManifestV1's.

t0yv0:
@Frassle PTAL, using int64.

Frassle (approved):
👍 looks fine to me
One day it would be really nice to get an openapi spec from the service team, so they can drive the evolution of these types and we just codegen them from openapi.

t0yv0:
> One day it would be really nice to get an openapi spec from the service team, so they can drive the evolution of these types and we just codegen them from openapi.

CC @EvanBoyle - one day but no rush :)

PR #10485, opened 2022-08-24
Fix import path for non-pulumi owner providers

### LINKED ISSUE #314: Need support for more file system primitives
In order to support mirroring static files from a Pulumi Application into the Pulumi Cloud (e.g. static resource files for a website), we need the ability to crawl the local file system at deployment time.

Single files can be exposed via `lumi.assets.File`, or zipping entire directories via `lumi.assets.Archive`. But it appears we need an entire File System interface: `ls`, `stat`, `sha1sum `,  etc. (For referenceNode's see [fs module](https://nodejs.org/api/fs.html).)

https://github.com/pulumi/pulumi-fabric/blob/master/lib/lumi/asset/asset.ts

Since this isn't strictly blocking any current milestone work on my plate, I was planning on defering this until the runtime change to V8 lands. Tracked with #311 . Since then we could just use the actual Node FileSystem module, rather than recreating some subset of it.

Though we'd still want to use some interfaced subset, to reduce the surface area we'd need to port when we support other languages. Right?

**Question**: Where is this implemented in Lumi? I assume the source code for handling the runtime implications of this object (reading file contents) is in https://github.com/pulumi/pulumi-fabric/tree/master/cmd/lumi somewhere?

Related to https://github.com/pulumi/pulumi-framework/issues/20.

/cc @lukehoban

aq17:
# Description

Fixes [pulumi/yaml 314](https://github.com/pulumi/pulumi-yaml/issues/314)

## Checklist


AaronFriel (commented):
I think if we changed this file to create a provider resource, we could add a test for this:

https://github.com/pulumi/pulumi/blob/b1557bc5ecca1bf56a3847b0c9d8784b04f67cf1/pkg/codegen/testing/test/testdata/third-party-package-pp/third-party-package.pp#L1-L8

pkg/codegen/testing/test/testdata/third-party-package-pp/go/third-party-package.go
```diff
@@ -20,6 +19,12 @@ func main() {
 		if err != nil {
 			return err
 		}
+		_, err := other.NewProvider(ctx, "Provider", &other.ProviderArgs{
+			Username: pulumi.String("foo"),
+		})
```
AaronFriel, line 24: Interesting. This has revealed another bug.

We're not importing the name `other` anywhere here. I'd expect one of two things:

1. This is `thirdparty.NewProvider`, and we keep the import `git.example.org/thirdparty`
2. This is `other.NewProvider` and we alias the import as `other`.

pkg/codegen/go/gen_program.go
```diff
@@ -512,6 +512,9 @@ func (g *generator) getPulumiImport(pkg, vPath, mod, name string) string {
 	if len(modSplit) > 1 {
 		if modSplit[0] == "" || modSplit[0] == "index" {
 			imp = fmt.Sprintf("github.com/pulumi/pulumi-%s/sdk%s/go/%s", pkg, vPath, pkg)
+			if info.ImportBasePath != "" {
+				imp = info.ImportBasePath
+			}
 		} else {
 			imp = fmt.Sprintf("github.com/pulumi/pulumi-%s/sdk%s/go/%s/%s", pkg, vPath, pkg, modSplit[0])
 		}
```
AaronFriel, line 520: The `else` block looks wrong here, too.

pkg/codegen/testing/test/testdata/third-party-package-pp/go/third-party-package.go
aq17: [The pkg name](https://github.com/pulumi/pulumi/blob/0287fb0a5eceb405adf2f7329f31a1065c239c9a/pkg/codegen/testing/test/testdata/other.json) is `other` and the import is `thirdparty`, so #1 is expected. Would we expect to translate `other:index:Thing` -> `import git.example.org/thirdparty/index` and have `index.NewThing`?

pkg/codegen/go/gen_program.go
aq17: `imp = fmt.Sprintf("%s/%s", info.ImportBasePath, mod)` ?

AaronFriel: Yeah, or `modSplit[0]`?

pkg/codegen/testing/test/testdata/third-party-package-pp/go/third-party-package.go
aq17: [The pkg name](https://github.com/pulumi/pulumi/blob/8e9f405fc777932d6737e06efcdd9fa100163b85/pkg/codegen/testing/test/testdata/other.json) is `other` but the import path is "git.example.org/thirdparty", so I think we need #2

aq17: I don't think this is a bug – in [other.json](https://github.com/pulumi/pulumi/blob/8e9f405fc777932d6737e06efcdd9fa100163b85/pkg/codegen/testing/test/testdata/other.json), we didn't have any alias, and we set the mod of `other:index:Thing` to be `other` correctly, hence `other.NewThing`, while the basepath is `git.example.org/thirdparty`. I'm not sure what the intention of setting the module to be `index` was to hit this special case, but in the meantime I added the `other` alias to the resource schema cc @iwahbe

iwahbe: I don't think we should care about the package name. We should see #1 (`third party.NewProvider`).

Regardless, this is unrelated to aliases in the schema, those should never be required for correct imports (any good schema should generate valid code). If we want to care about the package name, the aliasing would be 
```go
package sdk
import (
    other "git.example.org/thirdparty"
)

other.NewProvider(...)
```

aq17: Our [expected examples](https://github.com/pulumi/pulumi/tree/master/pkg/codegen/testing/test/testdata/third-party-package-pp) for all other languages import using the alias `other`. I'd expect consistent behavior, so if we expect #1, is there an issue with those languages' codegen?

pkg/codegen/go/gen_program.go
aq17: @AaronFriel actually, we just want this to be `info.ImportBasePath`, if the mod is equal to `index` or empty. Otherwise, we import something like `github.com/muhlba91/pulumi-proxmoxve/sdk/v2/go/proxmoxve/index` unnecessarily.

pkg/codegen/go/gen_program.go
```diff
@@ -510,8 +512,11 @@ func (g *generator) getPulumiImport(pkg, vPath, mod, name string) string {
 	modSplit := strings.Split(mod, "/")
 	// account for mods like "eks/ClusterVpcConfig" index...
 	if len(modSplit) > 1 {
-		if modSplit[0] == "" || modSplit[0] == "index" {
+		if modSplit[0] == "" || modSplit[0] == IndexToken {
 			imp = fmt.Sprintf("github.com/pulumi/pulumi-%s/sdk%s/go/%s", pkg, vPath, pkg)
+			if info.ImportBasePath != "" {
+				imp = info.ImportBasePath
```
aq17, line 518: @iwahbe Friel and I decided that we should not alias imports unless there is an explicit alias mapping in the schema, so this should be `thirdparty.NewProvider`. There was an inconsistency where we here we should be importing the `ImportBasePath` if `mod == "" || mod == "index"` (so the import would be `.../thirdparty`), but then before calling `getModOrAlias`, setting `mod = pkg` on line 625 which results in `modOrAlias` being `other`.

pkg/codegen/testing/test/testdata/third-party-package-pp/go/third-party-package.go
AaronFriel: For posterity here, the channel consensus and some intentional design choices (such as with aws-native) lean toward using the Go package name (#1) without an alias, and not the Pulumi package name.

pkg/codegen/testing/test/testdata/other.json
```diff
@@ -16,6 +16,13 @@
           "type": "number"
         }
       }
+    }, 
+    "pulumi:providers:other": {
+      "inputProperties": {
+        "name": {
+          "type": "string"
+        }
+      }
```
iwahbe, line 25: There is a `"provider"` section of the JSON. Providers are not standard resources.

iwahbe (commented):
This is looking good. I think we can merge once tests go 🟢.

pkg/codegen/testing/test/testdata/third-party-package-pp/go/third-party-package.go
```diff
@@ -8,7 +8,7 @@ import (
 
 func main() {
 	pulumi.Run(func(ctx *pulumi.Context) error {
-		_, err := other.NewThing(ctx, "Other", &other.ThingArgs{
+		_, err := thirdparty.NewThing(ctx, "Other", &thirdparty.ThingArgs{
 			Idea: pulumi.String("Support Third Party"),
 		})
```
AaronFriel, line 13: Where does this arg come from? I don't see it in the .pp file.

AaronFriel (commented):
Is the `.pp` file and codegen for the other 3 packages up to date?

AaronFriel (approved):
Disregard that, I was reading the wrong code.

Could we construct the provider with some args like the underlying issue does? Just to see if there's anything else that arises. If you can add an arg, LGTM.

Issue #10431, opened 2022-08-18
Panic when Aborting Operation with Ctrl-C 


RobbieMcKinstry:
### What happened?

As reported by @stack72, canceling a running program resulted in the following stack trace:

```go
panic: send on closed channel

goroutine 762 [running]:
github.com/pulumi/pulumi/pkg/v3/engine.diagEvent(0x140001744c8, 0x14000f1de00, {0x14000709940, 0x1c}, {0x14001d617a0, 0x51}, {0x10355ca13, 0x5}, 0x0)
	/Users/runner/work/pulumi/pulumi/pkg/engine/events.go:461 +0x150
github.com/pulumi/pulumi/pkg/v3/engine.(*eventEmitter).diagErrorEvent(...)
	/Users/runner/work/pulumi/pulumi/pkg/engine/events.go:485
github.com/pulumi/pulumi/pkg/v3/engine.(*eventSink).Errorf(0x140001744c8, 0x7c?, {0x0?, 0x2d?, 0x140014eff08?})
	/Users/runner/work/pulumi/pulumi/pkg/engine/eventsink.go:88 +0xf8
github.com/pulumi/pulumi/pkg/v3/resource/deploy.(*stepExecutor).executeChain(0x140017d4b40, 0x103eedbf0?, {0x14000c198a0?, 0x1, 0x0?})
	/Users/runner/work/pulumi/pulumi/pkg/resource/deploy/step_executor.go:262 +0x264
github.com/pulumi/pulumi/pkg/v3/resource/deploy.(*stepExecutor).worker.func1()
	/Users/runner/work/pulumi/pulumi/pkg/resource/deploy/step_executor.go:439 +0xb4
created by github.com/pulumi/pulumi/pkg/v3/resource/deploy.(*stepExecutor).worker
	/Users/runner/work/pulumi/pulumi/pkg/resource/deploy/step_executor.go:436 +0x21c
```

### Steps to reproduce

* (With Ctrl-C) Cancel a running Pulum Program…

* …right as the step executor is sending an engine event.

### Expected Behavior

Graceful shutdown to occur, with no panic

### Actual Behavior

Program panic

### Additional context

[Internal link](https://pulumi.slack.com/archives/C02VCJEBT2N/p1660774877880689) to Slack thread

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

t0yv0:
[...]

PR #10489, opened 2022-08-24
Do not panic when sending events to closed emitter

t0yv0:
# Description

Fixes [#10431](https://github.com/pulumi/pulumi/issues/10431)

## Checklist


pkg/engine/events.go
```diff
@@ -345,42 +356,68 @@ func makeStepEventStateMetadata(state *resource.State, debug bool) *StepEventSta
...
+			e, event)
+	}
+}
+
+func (e *eventEmitter) sendEvent(event Event) {
+	if e.isClosed() {
+		e.logUnableToSendEvent(event)
+		return
+	}
+	e.ch <- event
```
t0yv0, line 386: Tiny possibility of a race here, might need to use a mutex after all.

t0yv0:
This is not quite right (though it passes the repro), need to work out the synchronization logic better.

pkg/engine/events.go
```diff
@@ -278,7 +285,11 @@ func queueEvents(events chan<- Event, buffer chan Event, done chan bool) {
 				if !ok {
 					// If the event source has been closed, flush the queue.
 					for _, e := range queue {
-						events <- e
+						if !emitter.isClosed() {
+							events <- e
```
t0yv0, line 289: This isn't right but tricky. emitter.isClosed() is always true here. However if we don't guard here we have another panic in the repro. The events channel gets closed outside of the event emitter code.

pkg/engine/events.go
```diff
@@ -225,31 +226,37 @@ func makeEventEmitter(events chan<- Event, update UpdateInfo) (eventEmitter, err
...
-	go queueEvents(events, buffer, done)
-
-	return eventEmitter{
-		done: done,
-		ch:   buffer,
-	}, nil
+	closed := new(int32)
+	*closed = 0
+	emitter := eventEmitter{
+		closed: closed,
```
AaronFriel, line 245: Is there a reason that would be preferred over this?

```suggestion
	closed := 0
	emitter := eventEmitter{
		closed: &closed,
```

AaronFriel (commented):
Having refactored the event emitter, would it make sense to:

1. Move it to a separate package to keep the fields private
2. Use a `defer func() { if r := recover(); r != nil { ... }}` function in `sendEvents` rather than guards that check if it's closed?

As you note, a mutex is needed or you'll race the "is closed" check race.

t0yv0:
> Move it to a separate package to keep the fields private

That's not really possible in go due to import cycles, like this:

```
package github.com/pulumi/pulumi/pkg/v3/backend
        imports github.com/pulumi/pulumi/pkg/v3/backend/display
        imports github.com/pulumi/pulumi/pkg/v3/engine
        imports github.com/pulumi/pulumi/pkg/v3/engine/internal/event_emitter
        imports github.com/pulumi/pulumi/pkg/v3/engine: import cycle not allowed
package github.com/pulumi/pulumi/pkg/v3/backend
        imports github.com/pulumi/pulumi/pkg/v3/backend/display
        imports github.com/pulumi/pulumi/pkg/v3/engine
        imports github.com/pulumi/pulumi/pkg/v3/engine/internal/event_emitter
        imports github.com/pulumi/pulumi/pkg/v3/engine: import cycle not allowed
```

We'd need to introduce abstractions to break the cycles; that'd take the PR off-track for a very small marginal gain - I feel like won't tackle it here.

t0yv0:
Recovering from a chan send panic sounds like a good idea in our situation.. I'll try that. It's a code smell for sure I'll have to put heavy comments on, but I don't think I can confidently get the lifetimes right because *something* is closing engine events chan on termination but stepExecutor keeps sending new diag events.

Frassle:
I think a code smell here is fine for now. A lot of this logic is probably gonna get gutted and re-done as part of splitting the engine and cli up.

t0yv0:
@Frassle @AaronFriel PTAL - the last version still passes my repro test.

AaronFriel (approved):
lgtm, I appreciate scoping & documenting the defer/recover block in `trySendEvent`

PR #10491, opened 2022-08-25
Enabled Terraform modules to be converted to Python and added tests

T-Vova:


justinvp:
Closing as no longer relevant. Support for this was implemented in #12555.

PR #10492, opened 2022-08-25
Split PluginInfo in Info and Spec

Frassle:
# Description

PluginSpec is used to specifiy a plugin, and is what is passed to things like "Install". PluginInfo is used to refer to an installed plugin, and so has extra data like file sizes, and time stamps, but does not include things like plugin download url.

## Checklist

### DIFF
--- a/pkg/cmd/pulumi/about.go
+++ b/pkg/cmd/pulumi/about.go
@@ -223,15 +223,13 @@ type pluginAbout struct {
 }
 
 func getPluginsAbout() ([]pluginAbout, error) {
-	var pluginInfo []workspace.PluginInfo
-	var err error
-	pluginInfo, err = getProjectPluginsSilently()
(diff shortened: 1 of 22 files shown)


Frassle:
We'll need to fix yaml and tfbridge up after this, as they do they're own plugin loading.

Frassle:
This was interesting to do, seeing what was actually a "spec" and what was an "info". I've started looking into checksums now, and that's raising more questions about specs and plugin sources. I think "info" is correct from this PR now, it's an entry for either a plugin folder, or a single file on PATH.

iwahbe (approved):
This looks good to me. Do you have any idea why the windows test is failing? I'm nervous to merge unless I know why tests are failing.

Frassle:
> Do you have any idea why the windows test is failing?

No I've seen that a few time recently an re-runs have fixed it. The lack of logs on the action makes it a bit tricky to tell why.

PR #10494, opened 2022-08-25
Add support for SHA256 checksums in PluginSpec

Frassle:
# Description

Not yet plumbed all the way through, but this add support for Download to verify that what it downloads matches a checksum.

## Checklist


iwahbe (approved):
Looks good to me (pending a 🟢 test suite)

t0yv0:
Curious when plumbed all the way through, how will Pulumi CLI code know the "PluginSpec" checksums for every version of every plugin? Will it be baked into the CLI?

t0yv0:
That is, this seems to be verifying the crypto checksums against known checksums, but the database of known checksums and access to it is a bit of a complicated beast to manage. Have we considered and rejected the approach of verifying publisher signature instead like https://pkg.go.dev/crypto/ecdsa#VerifyASN1 or similar (I'm not up to date on best Go crypto libraries). This approach requires managing publisher pub keys, of which we'll have fewer (one for Pulumi and a few for trusted partners), instead of a full database for every version of every plugin.

iwahbe:
My understanding is that the checksums will be known by the published SDKs via embedding them in `pulumi-plugin.json` (or `package.json` for nodejs). This will verify that the downloaded binary was published along with the SDK requesting its download.

t0yv0:
That makes sense then! Just to be sure, so this will verify *automatic* downloads initiated via package managers like PIP, but not manual downloads such as `pulumi plugin install resource aws` etc?

Frassle:
> but not manual downloads such as pulumi plugin install resource aws etc?

@iwahbe's description is correct. Although I do plan on adding a --checksum option on `plugin install` as well, but yes it won't be automatic.

t0yv0 (approved):

justinvp:
Are we going to want to emit the checksum in all our generated SDKs, like we do with the download URL? https://github.com/pulumi/pulumi/issues/7242

Frassle:
> Are we going to want to emit the checksum in all our generated SDKs, like we do with the download URL? https://github.com/pulumi/pulumi/issues/7242

Yes

AaronFriel:
This introduces a sequencing constraint on building and publishing SDKs. The current process publishes the binaries at the end, and `goreleaser`'s workflow makes it a bit frustrating to create the final archives _prior to_ publishing them as part of a release.

I'm looking at this diagram here:

https://github.com/pulumi/pulumi-aws/actions/runs/2882584492

One way to solve this would be to use a separate channel for conveying integrity.

Frassle:
cc @stack72 for Friel's comment above. If we put checksums in package.json etc can providers build and get those archive checksums first to feed into sdk gen?

Frassle:
For the record the plan is for publishing sdks to come after binary building at some point anyway so this ordering constraint should be fine.

iwahbe:
To some degree, we already have this constraint. If we publish an SDK `vX.Y.Z` pointing to plugin `vX.Y.Z`, we have an invalid SDK until we publish the plugin, since it will point to a binary that does not exist.

PR #10496, opened 2022-08-25
Revert "Split PluginInfo in Info and Spec"

t0yv0:
Reverts pulumi/pulumi#10492

This change seems to be causing TestInstallCleansOldFiles test failure on Windows, which is only qualifying master builds. Reverting temporarily.


t0yv0:
https://github.com/t0yv0/pulumi/runs/8023094373?check_suite_focus=true passing with this commit taken out, failing otherwise.

aq17 (approved):

PR #10498, opened 2022-08-25
[sdk/nodejs] Use loopback addresses for providers & automation API

AaronFriel:
Improves security and should avoid tripping Windows & macOS firewall to allow `pulumi-resource-*` to accept network connections.


scottslowe (approved):
I'm not a strong coder (yet) but LGTM

Frassle (approved):

PR #10499, opened 2022-08-25
test: fix regex used to test Go version output in about command.

svangordon-fruit:
# Description

The issue that https://github.com/pulumi/pulumi/pull/10378 was addressing was not fixed by that change. The regex still does not match the expected output, causing the `TestAboutCommands/json` test to fail. This PR updates that regex, so that it's correctly checking that the Go version output is a valid core semver.

Fixes # (issue)

## Checklist

### DIFF
--- a/tests/about_test.go
+++ b/tests/about_test.go
@@ -42,7 +42,7 @@ func TestAboutCommands(t *testing.T) {
 		stdout, _ := e.RunCommand("pulumi", "about", "--json")
 		var res interface{}
 		assert.NoError(t, json.Unmarshal([]byte(stdout), &res), "Should be valid json")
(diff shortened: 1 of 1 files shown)


tests/about_test.go
```diff
@@ -42,7 +42,7 @@ func TestAboutCommands(t *testing.T) {
 		stdout, _ := e.RunCommand("pulumi", "about", "--json")
 		var res interface{}
 		assert.NoError(t, json.Unmarshal([]byte(stdout), &res), "Should be valid json")
-		assert.Regexp(t, `"goVersion": "go\d+.\d+"`, stdout)
+		assert.Regexp(t, `"goVersion": "go(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)"`, stdout)
```
AaronFriel, line 45:
```suggestion
		assert.Regexp(t, `"goVersion": "go1\..*"`, stdout)
```

AaronFriel (approved):

PR #10502, opened 2022-08-25
[sdk/python] Use loopback addresses for providers & automation API

AaronFriel:
Same as #10498, for Python.


justinvp (approved):
LGTM. Thanks!

PR #10503, opened 2022-08-25
[sdk/dotnet] Use loopback addresses for automation API

AaronFriel:
Like #10498, for dotnet. Dotnet uses the same class to provide a language host for Automation API inline & local programs, and lacks a dynamic provider, hence fewer lines changed.


justinvp (approved):
LGTM. Thanks!

Issue #2800, opened 2019-06-04
Should be possible to get the "qualified" stack name


hausdorff:
`getStack`, as implemented in both the Python and Node.js SDKs return only the stack name itself. They do not return the organization name.

I'm not sure what we're using this for, but it seems to me that it would be _much_ more useful for this to return this information as well. Another option is to have a separate `getFQStack` or something, but I have a hard time imagining when you'd want to get a stack name that does not include org name as well.

@ellismg I will leave it to you to triage, but I believe we should fix this before we move `pulumi query` out from behind the debug flag.

eropple:
There's a `name` field in `Pulumi.yaml` that seems like it should also be part of the qualified stack name. (This is particularly painful when trying to use the S3 backend, as right now all stacks in a single `pulumi login` are globally namespaced.)

joeduffy:
This is related to https://github.com/pulumi/pulumi/issues/2622.

I'll admit that I was surprised where we landed after refactoring the engine to support the new project model. I assumed we would treat `org/stack` as the "stack name" everywhere, and then simply support a way of getting the simple name if that's all you needed. This pops up in more places than just `getStack`, like whenever we print the stack name in the CLI. But if we're going to stick with the opposite, we definitely do need a way of getting the fully qualified name.

I can imagine a myriad of options

1. Make `getStack` return the fully qualified `org/stack` name (breaking change)
2. Make `getStack` return an object with two properties, `.org` and `.stack` (breaking change)
3. Same as #2, but also add a few complementary functions (breaking change)
    - `getStackName` that returns `stack`
    - `getStackOrganization` that returns `org`
    - `getFullStackName` that returns `stack/org`
4. Add a new `getFullStackName` function that returns `stack/org`
5. Add a new `getOrganization` that returns `org`
6. _... eliding others ..._

I'll just mention passing by, as the one who made the mistake, that the `getStack` and `getProject` names are unfortunate because they sound like they return rich objects, but actually just return names.

eropple:
The thing that I find awkward is that "org" is a Pulumi SaaS concept and doesn't map to a file-store backend (or similar, like S3).

I would expect, as a naive user, that a user would have SaaS logins that surfaced to them projects that they have visibility into, and projects would have stacks. S3 backends could then ignore the first step.

"Orgs" being a meaningful division (as opposed to a way to collect projects and apply ACLs to them, i.e. GitHub itself) is kinda odd.

joeduffy:
[...]

PR #10504, opened 2022-08-25
Add getOrganization to nodejs and python

Frassle:
# Description

NodeJS and Python support for https://github.com/pulumi/pulumi/issues/2800

## Checklist


Zaid-Ajaj (approved):
Went through the code, looks good! Can merge when CI goes ✅

PR #10509, opened 2022-08-26
[cli] fix `pulumi console` command

### LINKED ISSUE #10508: `pulumi console` no longer opens to the current stack
### What happened?

From Pulumi CLI docs, `pulumi console` is described as follows:

https://www.pulumi.com/docs/reference/cli/pulumi_console/ 
> Opens the current stack in the Pulumi Console

however `pulumi console` now opens to my personal app.pulumi.com org regardless of if a stack is selected or not.

### Steps to reproduce

1. from a directory containing a valid pulumi program, with a selected stack.
2. `pulumi console`
3. see landing page in Pulumi SaaS Service.

### Expected Behavior

I should be redirected to the specific stack eg- https://app.pulumi.com/phillipedwards/my-project/dev

### Actual Behavior

I am redirected to my personal org eg- https://app.pulumi.com/phillipedwards

### Output of `pulumi about`

_No response_

### Additional context

_No response_

### Contributing

dixler:
# Description
This PR fixes the `pulumi console` command to redirect the user to the stack's page in the pulumi service UI as documented.

Fixes #10508 

## Checklist

### DIFF
--- a/pkg/cmd/pulumi/console.go
+++ b/pkg/cmd/pulumi/console.go
@@ -73,13 +73,13 @@ func newConsoleCmd() *cobra.Command {
 				// Open the stack specific URL (e.g. app.pulumi.com/{org}/{project}/{stack}) for this
 				// stack if a stack is selected and is a cloud stack, else open the cloud backend URL
 				// home page, e.g. app.pulumi.com.
-				if consoleURL, err := cloudBackend.StackConsoleURL(stack.Ref()); err != nil {
-					launchConsole(consoleURL)
-				} else {
+				url, err := cloudBackend.StackConsoleURL(stack.Ref())
+				if err != nil {
 					// Open the cloud backend home page if retrieving the stack
 					// console URL fails.
-					launchConsole(cloudBackend.URL())
+					url = cloudBackend.URL()
 				}
+				launchConsole(url)
 				return nil
 			}
 			fmt.Println("This command is not available for your backend. " +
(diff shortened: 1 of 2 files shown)


pkg/cmd/pulumi/console.go
```diff
@@ -73,13 +73,13 @@ func newConsoleCmd() *cobra.Command {
 				// Open the stack specific URL (e.g. app.pulumi.com/{org}/{project}/{stack}) for this
 				// stack if a stack is selected and is a cloud stack, else open the cloud backend URL
 				// home page, e.g. app.pulumi.com.
-				if consoleURL, err := cloudBackend.StackConsoleURL(stack.Ref()); err != nil {
-					launchConsole(consoleURL)
-				} else {
```
dixler, line 78: logic was flipped.

susanev:
thank you!!

aq17 (approved):
🎉

pkg/cmd/pulumi/console.go
dixler: I could make a 1 char replacement diff, but I think rewriting it would avoid the easy misreading of `==` in the first place.

Frassle:
This change makes no sense! If there's no error fetching the stack URL then use the backend URL, else if there is an error fetching the stack URL use it. This wouldn't even be writable using result types, there is no way this is the correct fix.

aq17:
> It's too early in the morning to be reading code 😅

Had me re-reading the change like ten times before I saw your updated comment 😌

PR #10513, opened 2022-08-27
Enabled modules to be translated to Dotnet and added tests

T-Vova:


pkg/codegen/testing/test/testdata/terraform-child-module-pp/dotnet/terraform-child-module.cs
```diff
@@ -0,0 +1,27 @@
...
+            BucketName = "s3BucketId",
+        });
+
+        var childModule = new Aws_s3_static_website_bucket.Index("childModule", new()
+        {
+            CustomModuleParameter = s3Bucket.Id,
+        });
+
+    }
+}
```
Zaid-Ajaj, line 23: The generated code here looks good! One thing though would be great to add is
```cs
new CustomResourceOptions { Parent = this }
```
to each created resource

Zaid-Ajaj (changes requested):
Looks good but it would be good to set the parent to `this` for each resource created within. a component resource

RobbieMcKinstry:
Hey @T-Vova , would you have any interest in picking this up again? It looks like it's pretty close to the finish line! :)

